PR #1956, opened 2022-11-17
ci: split out zone tarballs in package job

iliana:
[Does what it says on the tin.](https://github.com/oxidecomputer/omicron/runs/9540802770) This is so we can more easily grab the zone images to shove into an updates repository of some fashion.

### DIFF
--- a/.github/buildomat/jobs/package.sh
+++ b/.github/buildomat/jobs/package.sh
@@ -6,6 +6,7 @@
 #: rust_toolchain = "nightly-2022-09-27"
 #: output_rules = [
 #:	"=/work/package.tar.gz",
+#:	"=/work/zones/*.tar.gz",
 #: ]
 #:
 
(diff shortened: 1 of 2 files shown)


.github/buildomat/jobs/package.sh
```diff
@@ -6,6 +6,7 @@
 #: rust_toolchain = "nightly-2022-09-27"
 #: output_rules = [
 #:	"=/work/package.tar.gz",
+#:	"=/work/zones/*.tar.gz",
```
smklein, line 9: There is a single tarball that currently is just a `.tar`, and not a `.tar.gz` -- sled agent itself.

We might wanna grab that one too, since we'll need it when constructing the Helios image?

smklein (approved):
Nice; this is way more succinct than I thought it would be!

.github/buildomat/jobs/package.sh
iliana: There's also `out/maghemite.tar`. I figured I'd deal with both of those later, it's simple enough to tweak this pair of scripts.

PR #1957, opened 2022-11-17
v1 instance APIs as per RFD-322

zephraph:
This PR establishes the new pattern for how the v1 of oxide's external API will be shaped. The details here are driven by the determinations of [RFD-322](https://rfd.shared.oxide.computer/rfd/0322).

To recap RFD-322, we want our API to be more ID focused but to also support names. Through the resolution of that RFD we've decided to go with a flatter API that uses query parameters to support names if needed. 

The notion of being ID first is something we'd like to flow through the rest of Nexus too. Where possible, IDs should always be favored over names. 

## Overview
- Instead of breaking the old API the new, flattened endpoints are being added under the `/v1/` path which we agreed to introduce in [RFD-299](https://rfd.shared.oxide.computer/rfd/0299). 
- Establishes the practice that names don't flow beyond the `http_entrypoints` file. New `<resource>_lookup` methods are introduced that resolve to the authz identifier of a given resource and that's passed around instead. 
- Introduces a notion of selectors (i.e. `ProjectSelector` and `InstanceSelect`) which is a data structure that stores all possible representations of parameters required to lookup a resource. 

### Example

Old way of fetching instance

```
GET /organizations/<org_name>/projects/<project_name>/instances/<instance_name>
GET /by-id/instances/<instance_id>
```

New ways of fetching instance

```
GET /v1/instances/<instance_id>
GET /v1/instances/<instance_name>?project=<project_id>
GET /v1/instances/<instance_name>?project=<project_name>&organization=<org_id>
GET /v1/instances/<instance_name>?project=<project_name>&organization=<org_name>
```

Old way of creating an instance

```
POST /organizations/<org_name>/projects/<project_name>/instances/<instance_name> { ...body }
```

New way of creating an instance

```
POST /v1/instances?project=<project_id> { ...body }
POST /v1/instances?project=<project_name>&organization=<org_id> { ...body }
POST /v1/instances?project=<project_name>&organization=<org_name> { ...body }
```

Note that query parameters are _always_ used to scope down an API call regardless of its method.


common/src/api/external/mod.rs
```diff
@@ -268,6 +268,47 @@ impl Name {
     }
 }
 
+#[derive(Serialize, Deserialize)]
+#[serde(try_from = "String")]
+pub enum ResourceIdentifier {
```
zephraph, line 273: Might just change this to be `NameOrID` because it's not _really_ a resource identifier. A resource identifier would be uuid or path, not just uuid and name.

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -2078,6 +2080,66 @@ async fn instance_create(
...
+                    query.organization_name.ok_or_else(|| {
+                        Error::InvalidRequest { message: "organization_name is required when using instance name".to_string() }
+                    })?;
+                let project_name = query.project_name.ok_or_else(|| {
+                    Error::InvalidRequest {
+                        message:
+                            "project_name is required when using instance name"
+                                .to_string(),
+                    }
+                })?;
```
zephraph, line 2124: @ahl would appreciate your input on if this is the right way to validate these params. _Technially_ this implementation will ignore it if you provide params for project/org and a UUID.

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -2065,19 +2095,69 @@ async fn instance_create(
...
-                &new_instance_params,
+                params::ProjectSelector::ProjectAndOrg {
+                    project_name: project_name.clone().into(),
+                    organization_name: organization_name.clone().into(),
+                },
             )
             .await?;
+        let instance = nexus
+            .project_create_instance(&opctx, &project_id, &new_instance_params)
+            .await?;
```
zephraph, line 2109: One of the implicit goals of RFD-322 is to reduce our overall reliance on names in Nexus. Effectively we only want names to be utilities to simplify lookups for humans but to not flow deeper down than API level. With this new selector lookup approach, we resolve names to relevant ids as the first step. Internal nexus apis should therefore _only_ take IDs.

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -2065,19 +2095,69 @@ async fn instance_create(
...
+    let path = path_params.into_inner();
+    let query = query_params.into_inner();
+    let handler = async {
+        let opctx = OpContext::for_external_api(&rqctx).await?;
+        let instance_id = nexus
+            .instance_lookup_id(
+                &opctx,
+                query.selector.to_instance_selector(path.instance.into()),
+            )
+            .await?;
```
zephraph, line 2154: Even though these top level instance endpoints actually take in the project selector via query params we transform it into an instance selector for more uniform matching. This does mean that the `instance_lookup_id` method isn't exactly aware of where it's being called from and is thus can only really return a generic error message. We should probably have a custom error type for this method such that we can do a better job from the callsite here of giving the user accurate error feedback (E.g. you're missing a query param).

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -2065,19 +2095,69 @@ async fn instance_create(
...
+    /// - `project_name`, `organization_name`
+    ///
+    /// If Id is used the above qualifiers are will be ignored
+    instance: NameOrId,
+}
+
+#[derive(Deserialize, JsonSchema)]
+struct InstanceQueryParams {
+    #[serde(flatten)]
+    selector: params::ProjectSelector,
```
zephraph, line 2130: I think the `flatten` here causes there to be duplicates in the output because the selector is an untagged enum. Need to figure out how to fix that.

zephraph: Actually, I believe I was incorrect. I think this is likely a schemars issue. I was able to work around it in [d0a19c6](https://github.com/oxidecomputer/omicron/pull/1957/commits/d0a19c6669e6427c24989a4d79fe84ac687a9bf4) by using `#[schemars(skip)]` for the duplicate names, but we should dig into where this is failing and make the relevant PR.

zephraph:
Once `nexus.json` is updated the build is failing with dropshot errors. This is what I'm seeing:

```
error[E0599]: `NameOrId` doesn't implement `std::fmt::Display`
   --> oxide-client/src/lib.rs:7:1
    |
7   | / progenitor::generate_api!(
8   | |     spec = "../openapi/nexus.json",
9   | |     interface = Builder,
10  | |     tags = Separate,
11  | | );
    | | ^
    | | |
    | | method `to_string` not found for this enum
    | | `NameOrId` cannot be formatted with the default formatter
    | |_doesn't satisfy `NameOrId: ToString`
    |   doesn't satisfy `NameOrId: std::fmt::Display`
    |
    = note: the following trait bounds were not satisfied:
            `NameOrId: std::fmt::Display`
            which is required by `NameOrId: ToString`
note: the following trait must be implemented
   --> /Users/just-be/.rustup/toolchains/nightly-2022-09-27-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/fmt/mod.rs:755:1
    |
755 | pub trait Display {
    | ^^^^^^^^^^^^^^^^^
    = note: this error originates in the macro `progenitor::generate_api` (in Nightly builds, run with -Z macro-backtrace for more info)
```

I went through in [f439f59](https://github.com/oxidecomputer/omicron/pull/1957/commits/f439f592f8abe46dcd33fd15f2ff8d2e7d82f1b8) and implemented `Display` for `NameOrId`, but it doesn't seem like that's actually what's required here. Dropshot is generating types that are failing and I've got to do something else to work around that, I'm just not sure what yet.

common/src/api/external/mod.rs
```diff
@@ -268,6 +268,47 @@ impl Name {
...
+        "NameOrId".to_string()
+    }
+
+    fn json_schema(
+        gen: &mut schemars::gen::SchemaGenerator,
+    ) -> schemars::schema::Schema {
+        schemars::schema::SchemaObject {
+            subschemas: Some(Box::new(schemars::schema::SubschemaValidation {
+                one_of: Some(vec![
+                    gen.subschema_for::<Name>(),
```
ahl, line 301: check out `label_schema`

ahl (commented):
still need to look at the meat of this PR, but wanted to post this so I didn't forget

common/src/api/external/mod.rs
zephraph: It's not clear to me what the implications for this are. Is it just hints for clients?

common/src/api/external/mod.rs
```diff
@@ -268,6 +268,47 @@ impl Name {
...
+    fn try_from(value: String) -> Result<Self, Self::Error> {
+        if let Ok(id) = Uuid::parse_str(&value) {
+            Ok(NameOrId::Id(id))
+        } else {
+            Ok(NameOrId::Name(Name::try_from(value)?))
+        }
+    }
+}
+
+impl JsonSchema for NameOrId {
```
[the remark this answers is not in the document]
zephraph: Could you provide some more context here? Should I have also added an impl for something else to keep it in sync, is the impl here incorrect, can this type _not_ have this impl for some reason I'm not seeing?

nexus/types/src/external_api/params.rs
```diff
@@ -18,6 +18,88 @@ use serde::{
...
+        project_name: Name,
+        organization_id: Uuid,
+    },
+    ProjectAndOrg {
+        // FIXME: There's a bug in schemars or serde which causes project_name to be emitted twice
+        #[schemars(skip)]
+        project_name: Name,
+        organization_name: Name,
+    },
+    None {},
```
david-crespo, line 37: Does this `None` need to be built into the enum because of how deserialization works? Otherwise it seems like a natural spot for `Option` or `Result`. Alternatively, I feel like in the `None` case the endpoint should automatically 400 at the parsing, which would mean you don't need the `None` in the enum itself.

zephraph: So an untagged enum tries to match all of the options and errors out if it can't find a match. I _think_ wrapping it in an option might be viable. I didn't take that tack because we always need to match against the enum anyway and having a None option (that essentially means there were not matches in the untagged enum) makes the handling easier. 

If a route has a path parameter and that path parameter is not an ID then the selector having a matching arm is required. Given that two types are involved here the optional isn't really valid. In this case we can actually document `None` as a part of the API where it's understood at what point you shouldn't include a selector.

nexus/tests/output/nexus_tags.txt
```diff
@@ -48,6 +48,16 @@ instance_start                           /organizations/{organization_name}/proj
...
+v1_instance_create                       /v1/instances
+v1_instance_delete                       /v1/instances/{instance}
+v1_instance_list                         /v1/instances
+v1_instance_migrate                      /v1/instances/{instance}/migrate
+v1_instance_reboot                       /v1/instances/{instance}/reboot
+v1_instance_serial_console               /v1/instances/{instance}/serial-console
+v1_instance_serial_console_stream        /v1/instances/{instance}/serial-console/stream
+v1_instance_start                        /v1/instances/{instance}/start
+v1_instance_stop                         /v1/instances/{instance}/stop
+v1_instance_view                         /v1/instances/{instance}
```
karencfv, line 60: I'm not entirely convinced about the `v1` prefix for the operation IDs 🤔 Maybe I'm missing something? This would make our generators create function names like `V1InstanceCreate()`.

karencfv (commented):
Thanks a bunch for working on this!!! Super stoked to see this coming alive :)

nexus/tests/output/nexus_tags.txt
zephraph: It would, yeah. I added the version so that it didn't break existing operations. I went back and forth on if it should be a prefix or a postfix and just settled on pre because it matches the url structure. I don't mind either way.

karencfv: Ah gotcha, so then is this only temporary for the PoC then?

zephraph: Well, I think we likely should keep either a pre-fix or a postfix. It's also possible that we have to rev an endpoint and it'd break the clients if we didn't have a separate op name.

david-crespo: For the foreseeable future, we will only have one version, so I think we should avoid adding `v1` to the operation IDs.

nexus/src/app/instance.rs
```diff
@@ -407,15 +430,11 @@ impl super::Nexus {
...
-        organization_name: &Name,
-        project_name: &Name,
-        instance_name: &Name,
+        instance_id: Uuid,
     ) -> UpdateResult<db::model::Instance> {
         let (.., authz_instance, db_instance) =
             LookupPath::new(opctx, &self.db_datastore)
-                .organization_name(organization_name)
-                .project_name(project_name)
-                .instance_name(instance_name)
```
david-crespo, line 418: This is all a big improvement, thanks for showing what this would look like.

david-crespo:
Don't have much useful to say, but I've looked over it and everything is surprisingly straightforward. I expect we'll find ways to abstract more when convert more resources and endpoints, but as it is, I like the simplicity.

david-crespo:
Rev 2 looks great, loving it.

nexus/tests/output/nexus_tags.txt
zephraph: While I agree that may be the case, my preference at this point is to make the changes additive so our current clients still work until we've migrated them. I moved the version to a postfix on the operation which feels a bit better. Once we've converted all the clients we can revisit the decision of removing the postfix.

zephraph:
This PR is blocked by https://github.com/oxidecomputer/progenitor/issues/253 which depends on https://github.com/oxidecomputer/typify/issues/144

nexus/src/app/instance.rs
```diff
@@ -53,16 +54,80 @@ use uuid::Uuid;
...
+    pub async fn instance_lookup_id(
+        &self,
+        opctx: &OpContext,
+        instance_selector: params::InstanceSelector,
+    ) -> LookupResult<Uuid> {
+        match instance_selector {
+            params::InstanceSelector { instance: NameOrId::Id(id), .. } => {
+                // TODO: 400 if project or organization are present
+                Ok(id)
+            }
```
zephraph, line 66: Question (maybe for @davepacheco). Right now the ID is just a pass through, but if we did a lookup here it would ensure anytime a resource lookup happened you'd always know the user had read permissions. It's somewhat silly given that you're looking up something you potentially already have, but it would guarantee that after this point IDs are valid and at _least_ support read access.

zephraph: I chatted with folks in chat and my take away was that read access is the least permissible role. I went ahead and implemented this change which sort of guarantees the minimal access is read only.

nexus/tests/integration_tests/endpoints.rs
```diff
@@ -1280,15 +1283,6 @@ lazy_static! {
...
 
-        VerifyEndpoint {
-            url: "/by-id/instances/{id}",
-            visibility: Visibility::Protected,
-            unprivileged_access: UnprivilegedAccess::None,
-            allowed_methods: vec![
-                AllowedMethod::Get,
-            ],
-        },
-
```
zephraph, line 1291: Given that the endpoints are now collapsed this mechanism of having two `VerifyEndpoint` definitions won't work. Essentially any endpoint that is verified here will be consumed from the spec so if you try to verify the same endpoint twice you'll end up with a situation where it says the spec doesn't contain the other instance of the endpoint (because the previous verify consumed it). 

After the API is converted this suite may need to be revisited.

nexus/tests/integration_tests/unauthorized_coverage.rs
```diff
@@ -138,7 +138,20 @@ fn test_unauthorized_coverage() {
...
+            unexpected_uncovered_endpoints
+                .push_str(&format!("\t{}\n", endpoint));
+            has_uncovered_endpoints = true;
+        }
+    }
+    assert_eq!(
+        has_uncovered_endpoints, false,
+        "{}\nMake sure you've added a test for this endpoint in unauthorized.rs.",
+        unexpected_uncovered_endpoints
+    )
```
zephraph, line 154: I have been on the other end of debugging this test one too many times. If you click through to the comment it's helpful sort of in explaining what's happening, but the test output most of the time is entirely illegible. This just updates the test output to tell you a list of things you need to write tests for instead of an inscrutable mismatch of file contents.

nexus/tests/output/uncovered-authz-endpoints.txt
```diff
@@ -7,3 +7,16 @@ login_spoof                              (post   "/login")
 login_local                              (post   "/login/{silo_name}/local")
 login_saml                               (post   "/login/{silo_name}/saml/{provider_name}")
 logout                                   (post   "/logout")
+
+Deprecated API endpoints to be removed at the end of the V1 migration
```
zephraph, line 11: Because I updated the test _not_ to be an exact file match we can add extra lines of text in here.

nexus/tests/integration_tests/endpoints.rs
```diff
@@ -224,29 +225,31 @@ lazy_static! {
...
     pub static ref DEMO_INSTANCE_NICS_URL: String =
-        format!("{}/network-interfaces", *DEMO_INSTANCE_URL);
+        format!("/organizations/{}/projects/{}/instances/{}/network-interfaces", *DEMO_ORG_NAME, *DEMO_PROJECT_NAME, *DEMO_INSTANCE_NAME);
     pub static ref DEMO_INSTANCE_EXTERNAL_IPS_URL: String =
-        format!("{}/external-ips", *DEMO_INSTANCE_URL);
-    pub static ref DEMO_INSTANCE_SERIAL_URL: String =
-        format!("{}/serial-console", *DEMO_INSTANCE_URL);
-    pub static ref DEMO_INSTANCE_SERIAL_STREAM_URL: String =
-        format!("{}/serial-console/stream", *DEMO_INSTANCE_URL);
+        format!("/organizations/{}/projects/{}/instances/{}/external-ips", *DEMO_ORG_NAME, *DEMO_PROJECT_NAME, *DEMO_INSTANCE_NAME);
```
zephraph, line 252: The new URL structure doesn't lend itself well to the same composition as before. You have parts of the hierarchy that come _after_ whatever we're appending to the structure. Instead of taking that tack I'm planning to just flatten this whole tree of variables. I think ultimately it'll make it clearer as to what's actually being defined.

zephraph:
Unless there's strong feedback my plan is to merge this PR by EOD. I will, of course, address any issues that come up after merge. There's some urgency around getting this API work done so that clients and be updated (and new client features worked on) before our next customer demo.

smklein (approved):
I find this new structure really readable, and the separation of "lookup" vs "authenticated usage" is great.

There are a couple areas where clearly we need to follow-up, but I understand why we're not doing them in this PR:
- Marking the non-v1 endpoints as deprecated (presumably waiting on https://github.com/oxidecomputer/dropshot/pull/503)
- *Deleting* the non-v1 endpoints
- Getting v1 endpoints up for all APIs, not just instances

Can we get up issues (or at least a larger, tracking issue) to make sure we're covering all these things? I know it's probably obvious in your head right now, but I wanna make sure we're recording this just in case the last 10% takes us some time to complete.

Again, great job with this PR, I'm excited to not need to puzzle over the codebase to figure out whether I should or should not have an "authz" object or a name.

ahl (approved):

zephraph:
> Can we get up issues (or at least a larger, tracking issue) to make sure we're covering all these things? I know it's probably obvious in your head right now, but I wanna make sure we're recording this just in case the last 10% takes us some time to complete.

Absolutely, I'll create that now.

zephraph:
I'm going to merge this PR and move forward for now. I would still like post merge feedback from @davepacheco. Even though I'm merging, we can still revisit decisions made here if necessary.

nexus/src/app/instance.rs
```diff
@@ -378,15 +403,11 @@ impl super::Nexus {
...
-        project_name: &Name,
-        instance_name: &Name,
+        authz_instance: &authz::Instance,
     ) -> UpdateResult<db::model::Instance> {
         let (.., authz_instance, db_instance) =
             LookupPath::new(opctx, &self.db_datastore)
-                .organization_name(organization_name)
-                .project_name(project_name)
-                .instance_name(instance_name)
+                .instance_id(authz_instance.id())
```
davepacheco, line 410: Same here as above -- we already did a lookup in the caller.  It'd be nice to just use that here instead of doing another lookup.  (You could even wind up with a different result here than the caller got.)

nexus/src/app/sagas/instance_create.rs
```diff
@@ -46,8 +46,6 @@ use uuid::Uuid;
 #[derive(Clone, Debug, Deserialize, Serialize)]
 pub struct Params {
     pub serialized_authn: authn::saga::Serialized,
-    pub organization_name: Name,
```
davepacheco, line 49: 🎉 (I love dropping names from saga stuff)

nexus/tests/integration_tests/instances.rs
```diff
@@ -123,6 +123,65 @@ async fn test_instances_access_before_create_returns_not_found(
     );
 }
 
+#[nexus_test]
```
davepacheco, line 126: I'm assuming this isn't as exhaustive a test as what we had for the old APIs.  Is there a plan for converting them?

nexus/tests/integration_tests/unauthorized_coverage.rs
```diff
@@ -138,7 +138,20 @@ fn test_unauthorized_coverage() {
     let expected_uncovered_endpoints =
         std::fs::read_to_string("tests/output/uncovered-authz-endpoints.txt")
             .expect("failed to load file of allowed uncovered endpoints");
-    assert_eq!(expected_uncovered_endpoints, uncovered_endpoints);
+    let mut unexpected_uncovered_endpoints = "These endpoints were expected to be covered by the unauthorized_coverage test but were not:\n".to_string();
```
davepacheco, line 141: It seems like we're no longer checking whether there are things in the allowlist file that _are_ covered.  That seems bad?  It feels like it undermines the point of an allowlist.  (why wouldn't we just put every endpoint into the allowlist?)

nexus/types/src/external_api/params.rs
```diff
@@ -18,6 +18,34 @@ use serde::{
 use std::{net::IpAddr, str::FromStr};
 use uuid::Uuid;
 
+#[derive(Deserialize, JsonSchema)]
```
davepacheco, line 21: At one point I think we had talked about describing more precisely the set of things that was allowed, so that it would be impossible to express for example an InstanceSelector that had an instance name and no project part.  Did we decide not to do that because it can't be expressed using query parameters?

nexus/tests/output/uncovered-authz-endpoints.txt
```diff
@@ -7,3 +7,16 @@ login_spoof                              (post   "/login")
...
+instance_delete                          (delete "/organizations/{organization_name}/projects/{project_name}/instances/{instance_name}")
```
davepacheco, line 12: Why aren't we covering these any more?

davepacheco (commented):
Thanks for taking this on!  Sorry this took a little bit but it sounded like the feedback was still wanted.

I found a bunch of things (noted above) that I think would be resolved if the `foo_lookup()` functions returned `lookup::Foo` instead of `authz::Foo`:

- The lookup functions themselves could directly use the parent lookup function (i.e., `instance_lookup()` could call `project_lookup()`) rather than duplicating the logic.  (I'm worried this logic duplication (and boilerplate) grows with the depth of the tree.  It'll also be more than it looks like here once we handle the extra error cases. )
- The caller of the lookup function (or, more likely, subsequent code) can decide which authz check to do.  This preserves the previous coupling between lookup and authz check that I hope makes it harder to get the authz check wrong (or forget one).  It would also avoid doing multiple redundant authz checks in every path.  (There are still some code paths with extra checks (just like before), but with the current pattern, I think _every_ path would have redundant checks.)
- The caller of the lookup function (or subsequent code) can decide whether they want the database part (`fetch()`) or not (`lookup()`).  Otherwise, are many code paths that do multiple lookups just to re-fetch the database part that was just fetched and thrown away.

I'm curious what you think.

Thanks again for taking this on.  Long live ids!

nexus/tests/integration_tests/instances.rs
zephraph: Yes, absolutely.

nexus/tests/integration_tests/unauthorized_coverage.rs
zephraph: This is bending my brain a little. So, this would happen when you've set something in the allowlist that you expect to _not_ be covered by tests but the tests cover it anyway. So the `uncovered_endpoints` would be _smaller_ than the `expected_uncovered_endpoints` meaning you might need to remove something from the allowlist. 

That's fair. I think the thing I was trying to solve here is there failure experience is _really bad_. I've hit up against this test a lot and it always takes me a while to figure out what went wrong and what I need to do. Ultimately my goal here is just to make it more explicit as to what needs to be done. I'll see if I can work backwards and test both ways without making it messier.

nexus/types/src/external_api/params.rs
zephraph: In a previous iteration of this I defined the shape of the selector as an enum of only valid combinations. There were a few challenges with this (like a bug in schemars that caused duplicate schema output). There was added implementation complexity there that ultimately still yielded the same schema output. I had floated some ideas in the RFD about how to make the schema more expressive in terms of typing, but generating that output is a highly non-trivial endeavor that would eat up more time than I have to give to this initiative.

nexus/src/app/instance.rs
```diff
@@ -53,18 +54,84 @@ use uuid::Uuid;
 const MAX_KEYS_PER_INSTANCE: u32 = 8;
 
 impl super::Nexus {
+    pub async fn instance_lookup(
```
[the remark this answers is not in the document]
zephraph: Yeah, I like that idea a lot. I'll explore it a little more. There've been some iterations on this that have slowly started to feel better. It went from names, to ids, to authz objects. Maybe the lookups are the last step to _really_ make it compositional. Thanks for the suggestion.

nexus/tests/integration_tests/unauthorized_coverage.rs
davepacheco: That sounds good.  Yeah, I'm all for making this easier.  We could also use expectorate here like we do in other places.  I think the suggestion _not_ to was based on not making it seem like the right thing for people to _add_ things to this list.

Issue #1960, opened 2022-11-18
Bootstrap Agent: Responsibility for Launching Switch zone needs to happen earlier


smklein:
See 11/18 Control Plane Sync for details.

TL:DR:
- Current launching of the "Switch Zone" is managed by the sled agent (nit: NOT the bootstrap agent). The sled agent monitors devinfo for the `/dev/tofino` device, and creates the switch zone with a corresponding underlay address for the switch zone to use.
- **This is wrong**. The switch zone needs to launch before trust quorum has been established, and before the sled agent is up-and-running. As such, this hardware monitoring and device launching needs to happen within the bootstrap agent itself.
- Furthermore, we will want to supply some mechanism for the underlay address to be "added later" to the switch zone so Nexus can communicate with switch zone services (i.e., MGS, Dendrite) after underlay addressing has been  established. This is tracked by https://github.com/oxidecomputer/omicron/issues/1961

Issue #1961, opened 2022-11-18
Switch Zone: Create "late-binding" mechanism for provisioning underlay address


smklein:
See: https://github.com/oxidecomputer/omicron/issues/1960 for context

- We want to launch the switch zone before underlay addresses are established
- When the underlay comes up, we'd like to add the "switch zone address" to the already-running "switch zone"

smklein:
@Nieuwejaar I wanted to get your feedback on this, here's my proposal:

1. The bootstrap service launches the switch zone *without* an underlay IP address. Services in the switch zone will need to be updated to ensure that this address is an optional argument.
2. The underlay is established as a part of RSS, eventually sled agents can boot.
3. Once the sled agent starts, it **halts and destroys** the switch zone, and then **creates a new switch zone** using the underlay address.

WDYT? This is definitely doable from the sled-agent side of things, and seems like the easiest path forward (it seems doable in https://github.com/oxidecomputer/omicron/pull/1970 ). Does stopping / re-starting the switch zone pose any problems from your point-of-view (aside from https://www.illumos.org/issues/15167 , perhaps)?

Nieuwejaar:
For dpd and tfportd, the address is already optional.  If one isn't provided, they use localhost and the default port number.

Tearing down and restarting the switch zone should be fine - modulo the illumos kernel panic.  If you can `svcadm disable` tfport before halting the zone, I can have it clean up all of its links before exiting.  If all goes well, that _should_ avoid the kernel panic.

bnaecker:
@Nieuwejaar I'm surprised that this is workable. Is there really _no_ state at all maintained by the programs running in the switch zone? It seems like restarting either `dpd` or the BF SDE would walk back through a bunch of link initialization that might interrupt things. Is that not true?

As an alternative, we could include in the `dpd` Dropshot API an endpoint for starting a _new_ server listening on some underlay address. Those would be duplicates really, modulo that endpoint for starting a new server, each providing the same API just on different addresses. IIRC, the inner state of the main `dpd` server object (or the `Switch` type maybe?) is behind a lock already, so I don't think there should be an additional hit for synchronizing requests handled by different servers.

Nieuwejaar:
Sorry - I didn't mean to imply that this would be seamless.  The switch would essentially go offline during the restart and the links would all need to come back up.  By "fine" I just meant that this would work, not that it was ideal.  Thanks for catching me.

[...]

Issue #1934, opened 2022-11-11
sled agent must ensure a dataset for zpools (currently `rpool/zone`) exists


smklein:
This issue has been broadened in scope a bit.

OLD:

Currently, [sled agent] assumes the `rpool/zone` dataset exists - it should *ensure* it exists. This won't necessarily be present on all machines.

NEW:

Sled Agent must be responsible for management of U.2s, M.2s, and the partitions, zpools, and datasets within.

This issue tracks Sled Agent performing the following tasks:
- [x] Identify new U.2 / M.2 devices from `libdevinfo`
- [x] Query the devices to understand if they have already been formatted with a zpool
  - [x] If "no", `zpool create` a new zpool on the device
- [x] Compare the list of "imported zpools" with "the list of devices from `libdevinfo`. Import any new devices
- [x] Ensure (at least one) dataset for zones exists (Completed by https://github.com/oxidecomputer/omicron/commit/789fac409ab5de0c75e64b71ba82eb47c75a5808)

smklein:
Related to this: `omicron-package` uses the utility `get_all_omicron_datasets` to delete all pre-installed zones when doing an install. If `rpool/zone` does not exist, this iteration currently fails - we should handle this case more gracefully (It's probably fair to assume these datasets do not exist if `ZONE_ZFS_DATASET` does not exist).

leftwo:
The actual error I see on an install of a "fresh" system (meaning omicron was not installed yet, a fresh boot on a ramdisk).

```
Running install
Nov 16 23:20:58.008 DEBG target: Target({"switch_variant": "stub"})
Nov 16 23:20:58.014 INFO Installing service, dst: /opt/oxide/internal_dns.tar.gz, src: out/internal-dns.tar.gz
Nov 16 23:20:58.014 INFO Installing service, dst: /opt/oxide/propolis-server.tar.gz, src: out/propolis-server.tar.gz
Nov 16 23:20:58.014 INFO Installing service, dst: /opt/oxide/mg-ddm.tar, src: out/maghemite.tar
Nov 16 23:20:58.014 INFO Installing service, dst: /opt/oxide/crucible.tar.gz, src: out/crucible.tar.gz
Nov 16 23:20:58.014 INFO Installing service, dst: /opt/oxide/switch.tar.gz, src: out/switch-stub.tar.gz
[...]

PR #1970, opened 2022-11-21
[bootstrap][sled-agent] Launch the switch zone earlier

smklein:
This PR adds the ability for the bootstrap agent to launch the switch zone, without a corresponding underlay network. This is necessary, as the bootstrap agent must be capable of initializing the switch zone in order to run the bootstrap network, which allows for rack initialization.

However, this is also complicated for a few reasons:
- Sled Agent also wants to be in charge of service initialization, once it's capable of executing
- Once an underlay address exists, the switch zone needs to be re-initialized with the new address (so it may be accessible to other control plane entities, such as Nexus).

TODOs before taking out of draft:
- [x] (Done in 043d580e45aa457069d56fcdfa101caed484d5f8, https://github.com/oxidecomputer/omicron/pull/1970/commits/2dd1a6294e9e9537f99d576b29268cd5607053df)  Decide whether we should "kill and restart" the switch zone, or "refresh the services within". See https://github.com/oxidecomputer/omicron/issues/1961 for context; current implementation is ~~"kill and restart"~~ update SMF properties and refresh service.
- [x] The SMF services for the zones being launched all still *expect* that they'll receive a non-optional IP address for the underlay. This needs to be updated to be actually optional.
- [x] (Moved to https://github.com/oxidecomputer/omicron/issues/1934 ; we are currently simply using `rpool/zone`) I left a comment in the bootstrap agent about this - do we need to initialize ZFS filesystem for the zones? In Sled Agent, we call `zfs create -o zoned=on -o mountpoint=/zone rpool/zone` before creating any zones, and all our zone filesystems exist within `rpool/zone`. Should this be the responsibility of the bootstrap agent to perform this operation? The switch zone would have a local filesystem, no?

Fixes https://github.com/oxidecomputer/omicron/issues/1960
Part of https://github.com/oxidecomputer/omicron/issues/1961


sled-agent/src/bootstrap/agent.rs
```diff
@@ -76,6 +88,151 @@ impl From<BootstrapError> for ExternalError {
...
+        exit_rx: oneshot::Receiver<()>,
+        hardware: HardwareManager,
+        services: ServiceManager,
+    ) -> Self {
+        Self { log, exit_rx, hardware, services }
+    }
+
+    async fn run(
+        mut self,
+    ) -> Result<(HardwareManager, ServiceManager), BootstrapError> {
```
smklein, line 114: A very similar type of worker exists in `sled_agent.rs`, which also monitors for the switch zone.

However, these "workers" are distinct -- the bootstrap agent version *only* looks for the switch zone, but the sled agent one is presumably going to monitor for a lot more signals from hardware.

Additionally, the bootstrap agent is activating the switch zone without an underlay IP address, while the sled agent *always* activates the switch zone with an underlay IP address.

sled-agent/src/bootstrap/agent.rs
```diff
@@ -236,91 +408,154 @@ impl Agent {
...
+                        }
+                    }
+                }
+                let restarter = RestartMonitor {
+                    run: true,
+                    log: self.log.clone(),
+                    hardware: Some(hardware),
+                    services: Some(services.clone()),
+                    monitor: hardware_monitor,
+                };
```
smklein, line 470: I fully admit, this is a little weird, but it felt like a safe way of solving this problem.

Basically: we need to stop the hardware monitor and take resources away from the bootstrap server to hand them off to the sled agent. But if this fails, we need to put them back.

I did consider changing the initialization of the sled agent, to make the `ServiceManager` an `Option`, but then sled agent would need error-handling for an "optional" `ServiceManager` while serving requests on its HTTP service, which seems grosser (and more visibly bad).

sled-agent/src/bootstrap/agent.rs
```diff
@@ -236,91 +408,154 @@ impl Agent {
...
+                tokio::fs::write(&path, &serialized_request).await.map_err(
+                    |err| BootstrapError::Io {
+                        message: format!(
+                            "Recording Sled Agent request to {path:?}"
+                        ),
+                        err,
+                    },
+                )?;
+
+                restarter.cancel();
```
smklein, line 511: This is the point where we say: We aren't going to try to put the resources back; we're all-in on the sled agent starting.

sled-agent/src/bootstrap/agent.rs
bnaecker: I think I understand what's going on here. To check:

- `restarter` is capturing the hardware monitor (and some other stuff)
- Importantly, the `run` field is `true` at construction.
- If this method completes successfully, we `cancel()` the `restarter`, which sets `run` to false, and continue with...other stuff.
- If the method fails, `restarter` is dropped. Because `run` is true at construction, this will ensure the hardware monitor continues in that case.

So far so good. I think there's one thing I'm missing: don't we need to put the monitor "back" somewhere? If this method fails, how does the monitor not get dropped anyway?

bnaecker: Ok, I think I see now. It's a separate Tokio task, which we signal to exit when we're satisfied.

sled-agent/src/services.rs
```diff
@@ -456,6 +405,21 @@ impl ServiceManager {
                         }
                     }
                 }
+                ServiceType::Tfport { pkt_source } => {
+                    // The tfport service requires a MAC device to/from which sidecar
+                    // packets may be multiplexed.  If the link isn't present, don't
+                    // bother trying to start the zone.
```
bnaecker, line 411: I'm not really able to follow all the logic here, but will this be retried? The `tfpkt` driver can be reloaded, or otherwise manually manipulated, especially during development. If there is no `pkt_source` at one point in time, this will be retried, right?

sled-agent/src/services.rs
```diff
@@ -883,65 +863,168 @@ impl ServiceManager {
         Ok(())
     }
 
-    pub async fn ensure_switch(&self, request: Option<ServiceZoneRequest>) {
+    /// Ensures that a switch zone exists with the provided IP adddress.
+    pub async fn activate_switch(
+        &self,
+        switch_ip: Option<Ipv6Addr>,
```
bnaecker, line 869: Kind of a nit, but can we rename this to `service_ip` or something like that? IIUC, this is the link-local IPv6 we use for `dpd`, MGS, and `tfportd`, and not an IP address on the switch itself.

bnaecker (approved):
Looks OK to me, with the caveat that I'm not in a position to take a deep look at much of it. I think you've used it successfully, and we're in a bit of a time crunch, so 👍 !

A couple of questions  / nits, but merge at your discretion. Thanks Sean!

sled-agent/src/bootstrap/agent.rs
smklein: I'm going to just add this explicitly as a comment: 24a03b37e4d71bd8af502955e1ba111f8370aa04

sled-agent/src/illumos/running_zone.rs
```diff
@@ -307,22 +309,39 @@ impl RunningZone {
     pub fn opte_ports(&self) -> &[Port] {
         &self.inner.opte_ports
     }
+
+    /// Halts and removes the zone, awaiting its termination.
+    ///
+    /// Allows callers to synchronously stop a zone, and inspect an error.
```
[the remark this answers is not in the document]
smklein: Yeah, before, stop happened on `drop`, so it wasn't possible to get a `Result` out of it.

With this new version, it's at least possible to see if stopping the zone succeeds or fails.

sled-agent/src/services.rs
smklein: Here's the callstack, starting from the `pub` API used elsewhere:

`pub activate_switch` <- Called when we detect the tofino driver was loaded
`ensure_switch_zone`
`start_switch_zone`
`initialize_zone_loop` <- Called in a newly spawned `tokio::task`, loops until the zone launches or we're explicitly told to stop.
`initialize_zone`
`link_needed` <- This is the function where we're verifying the `pkg_source`

So yeah, we'll keep retrying to call `initialize_zone` forever until we're told otherwise (e.g., if the tofino driver is unloaded).

sled-agent/src/services.rs
smklein: IMO `service_ip` may be ambiguous in the context of the service manager, and I kinda like that the name is uniform between files. WDYT about `switch_zone_ip`? (See: e42b4689823c8e4064937ec557f5da3bd9c75406)

bnaecker: That's good! Just something to be clear it's for software running in the switch zone on the host, not anything particularly related to the switch itself.

bnaecker:
Thanks @smklein, all your updates and comments are super helpful as usual. Everything LGTM, feel free to merge when you're ready!

PR #1985, opened 2022-11-28
[readme] add instructions to install prerequisites

sunshowers:


README.adoc
```diff
@@ -29,6 +29,22 @@ Folks with access to Oxide RFDs may find RFD 48 ("Control Plane Requirements") a
 
 == Build and run
 
+On Linux, Mac or illumos, to install prerequisites for a xref:docs/how-to-run-simulated.adoc[simulated version of omicron], run:
+
+[source,text]
+----
+$ ./tools/install_prerequisites.sh
+----
+
```
smklein, line 38: Can we add this, for context?

```suggestion
This script expects that you are both attempting to compile code and execute it on the same machine. If you’d like to have a different machine for a "builder" and a "runner", you can use the two more fine-grained scripts:

[source,text]
----
# To be invoked on the machine building Omicron
$ ./tools/install_builder_prerequisites.sh
# To be invoked on the machine running Omicron
$ ./tools/install_runner_prerequisites.sh
----
```


This is copied from here: https://github.com/oxidecomputer/omicron/blob/main/docs/how-to-run.adoc#installing-prerequisite-software but I think the distinction can be important, depending on how multi-machine deployments are set up.

sunshowers: Thanks for the review! I think I'll do what @davepacheco mentioned and reorder the sections to avoid DRY.

davepacheco (approved):

PR #1986, opened 2022-11-28
Add direnv setup

sunshowers:
Make it easy for [direnv](https://direnv.net) users like myself to have clickhouse and cockroachdb added automatically to the PATH.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,4 @@
+# For use with direnv: https://direnv.net
+
+PATH_add out/cockroachdb/bin
+PATH_add out/clickhouse


smklein (approved):

davepacheco:
Thank you for the comment in the file!  (I wonder what else it needs to be kept in sync with.  I guess some of install_prerequisites.sh and/or the how-to-run docs.)

sunshowers:
> Thank you for the comment in the file! (I wonder what else it needs to be kept in sync with. I guess some of install_prerequisites.sh and/or the how-to-run docs.)

I think so! Maybe we could also suggest using direnv in the how to run docs, if more people start using it.

PR #1987, opened 2022-11-28
[nexus] reduce MIN_EXPECTED_PASSWORD_VERIFY_TIME to 650ms

sunshowers:
On sufficiently fast machines like my Ryzen 7950x, we can go under the
900ms limit.

I tested this by running it a few times, and it appears to work.

### DIFF
--- a/nexus/passwords/src/lib.rs
+++ b/nexus/passwords/src/lib.rs
@@ -53,7 +53,7 @@ pub const MAX_PASSWORD_LENGTH: usize = 512;
 // takes as long as we think it should on whatever machine the test suite is
 // running on.
 pub const MIN_EXPECTED_PASSWORD_VERIFY_TIME: std::time::Duration =
-    std::time::Duration::from_millis(900);
+    std::time::Duration::from_millis(650);
 
 /// Returns an [`Argon2`] context suitable for hashing passwords the same way
 /// we do for external authentication


davepacheco:
Based on the output you pasted in chat (1.3s for two iterations), this seems pretty safe.  For reference, [here's my rationale from last time we reduced this](https://github.com/oxidecomputer/omicron/pull/1841#discussion_r1006243996).  I think it still holds for a value of 650ms.  This is basically saying: it should take at least 650ms on all development and test machines, which is still within the "0.5s-1s" target mentioned in RFD 321.  But it's getting close.  Much beyond this and I'd feel more comfortable bumping the Argon parameters instead (so it takes longer).

I'm not an expert in this area and if other folks do feel it's safer to bump the parameters now rather than reduce the minimum expected time, that'd work for me too.

davepacheco (approved):

sunshowers:
OK -- I looked at bumping the Argon params but I was worried that it would require passwords to be reset (depending on the design of the database) so I didn't touch it.

sunshowers:
Looking at that discussion, I guess the other option here would be to have separate minimum required times for dev machines and gimlets. But maybe let's get this landed for now and iterate on it.

davepacheco:
> OK -- I looked at bumping the Argon params but I was worried that it would require passwords to be reset (depending on the design of the database) so I didn't touch it.

For the record, it shouldn't cause passwords to be reset (there are tests to ensure that), and even if it did, we're at a stage where that's okay.

> Looking at that discussion, I guess the other option here would be to have separate minimum required times for dev machines and gimlets. But maybe let's get this landed for now and iterate on it.

True.  I explained in that comment why I don't think that's the way to go.  I don't think there's any particular reason to differentiate based on _hardware_.  We could differentiate dev/prod, but I'm afraid that makes it easier to accidentally deploy insecure configuration to prod.

sunshowers:
Ah, got it -- thanks!

PR #1995, opened 2022-11-30
RFD-322 PoC Revision 2

zephraph:
This is an updated approach to #1957 that was spawned from feedback from @ahl. The main feedback here is that we'd like the CLI and API be consistent. In the cli we'll probably use `--project` and `--organization` not `--project_id`, `--project_name`. The over-specification of params and their types didn't actually make the resulting API any simpler... in fact, it introduced extra failure modes where you could have combinations of selectors for the same resource. 

I'm putting this up as a separate PR to merge on the first just so folks can take a peek at it and voice any feedback/concerns before I merge them together. 

## Differences from #1957

- Selectors (`InstanceSelector`, `ProjectSelector`) are structs now instead of enums. 
- Uses `NameOrId` for query parameters reducing `project_id` and `project_name` down to just `project`
- DB newtype wrapper for `NameOrId` is no longer necessary

### DIFF
--- a/nexus/src/app/instance.rs
+++ b/nexus/src/app/instance.rs
@@ -32,6 +32,7 @@ use omicron_common::api::external::InstanceState;
 use omicron_common::api::external::InternalContext;
 use omicron_common::api::external::ListResultVec;
 use omicron_common::api::external::LookupResult;
+use omicron_common::api::external::NameOrId;
 use omicron_common::api::external::UpdateResult;
 use omicron_common::api::external::Vni;
 use omicron_common::api::internal::nexus;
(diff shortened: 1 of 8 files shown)


karencfv (commented):
This is looking great! I like it, a lot of the complexity has been reduced

nexus/src/app/instance.rs
```diff
@@ -74,10 +78,10 @@ impl super::Nexus {
...
                 Ok(authz_instance.id())
             }
-            params::InstanceSelector::InstanceProjectAndOrgId {
-                instance_name,
-                project_name,
-                organization_id,
+            params::InstanceSelector {
+                instance: NameOrId::Name(instance_name),
+                project: Some(NameOrId::Name(project_name)),
+                organization: Some(NameOrId::Id(organization_id)),
```
david-crespo, line 84: I see, tbe optional parent identifiers aren’t a problem because you were already doing this whole pattern match on the valid combinations anyway. So encoding them into the type was redundant and way harder for spec generation.

david-crespo:
Great improvement!

PR #1998, opened 2022-11-30
[wicketd] add stub API for fetching artifacts

sunshowers:
Add a basic stub API for fetching artifacts, and update the OpenAPI
schema.

The implementation hasn't been filled out yet.

### DIFF
--- a/openapi/wicketd.json
+++ b/openapi/wicketd.json
@@ -10,6 +10,60 @@
     "version": "0.0.1"
   },
   "paths": {
+    "/artifacts/{name}/{version}/{hash}": {
+      "get": {
+        "summary": "Fetch an artifact from the in-memory cache.",
+        "operationId": "get_artifact",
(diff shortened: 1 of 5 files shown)


openapi/wicketd.json
```diff
@@ -10,6 +10,60 @@
     "version": "0.0.1"
   },
   "paths": {
+    "/artifacts/{name}/{version}/{hash}": {
```
andrewjstone, line 13: I don't think we'd want to require a hash for all artifacts. It's really only needed for phase2 images. I'm not sure how those will be versioned. Ideally, we would just target by name and version. But in some cases we may want name and hash. For now, I'd just exclude the hash.

sunshowers: Done

andrewjstone (approved):
Looks great. Glad to have you in this code base @sunshowers

PR #2001, opened 2022-12-01
[wicketd] add to default-members, fix up clippy, docs and test issues

sunshowers:
Apparently I'd forgotten to update the spec, but it wasn't caught in CI.

Adding wicketd to default-members causes the wicketd tests to be run in CI. It also catches a few clippy and docs-related issues.


smklein (approved):
I know you're still working through the clippy issue (which is now being raised because clippy is running where it should be) but the rest of this PR LGTM

PR #2005, opened 2022-12-02
dedup progenitor, and update crucible and propolis revs

sunshowers:
This is PR 3 of N in this series. For the earlier ones, see:

* https://github.com/oxidecomputer/crucible/pull/552
* https://github.com/oxidecomputer/propolis/pull/262

Currently, we specify the progenitor dependency as both:

```toml
progenitor-client = { git = "https://github.com/oxidecomputer/progenitor", branch = "main" }
```

And

```toml
progenitor-client = { git = "https://github.com/oxidecomputer/progenitor" }
```

Cargo (due to https://github.com/rust-lang/cargo/issues/7497) doesn't dedup these dependencies even
though they resolve to the same branch and hash. This causes issues with
`cargo doc` colliding with itself, as it tries to document both:

```
Documenting progenitor-impl v0.2.1-dev (https://github.com/oxidecomputer/progenitor#4b5cef4b)
Documenting progenitor-impl v0.2.1-dev (https://github.com/oxidecomputer/progenitor?branch=main#4b5cef4b)
```

Fix this by always using the branch name. (We choose this option over
not specifying the branch so to maintain uniformity in cases where a
non-default branch must be used.)

This also requires a crucible update, since that didn't specify a branch
either. And the crucible update necessitates a propolis update to get the versions aligned.

In the future it would be nice to have a lint which ensures that git
dependencies always have the branch name in them.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1057,18 +1057,6 @@ dependencies = [
 [[package]]
 name = "crucible-client-types"
 version = "0.1.0"
-source = "git+https://github.com/oxidecomputer/crucible?rev=144d8dafa41715e00b08a5929cc62140ff0eb561#144d8dafa41715e00b08a5929cc62140ff0eb561"
-dependencies = [
- "base64",
- "schemars",
- "serde",
- "serde_json",
- "uuid",
-]
-
-[[package]]
-name = "crucible-client-types"
-version = "0.1.0"
 source = "git+https://github.com/oxidecomputer/crucible?rev=861a09c350fd0e127b6105c8099e31a0bc3b711a#861a09c350fd0e127b6105c8099e31a0bc3b711a"
 dependencies = [
  "base64",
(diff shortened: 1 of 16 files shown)


sunshowers:
This is going to need https://github.com/oxidecomputer/propolis/pull/262 as well.

leftwo (commented):
You will also need to update the packaging side as well, with new links to the CI produced 
archives for propolis and crucible, as we want the compiled Omicron to match the tar.gz file that becomes part of the installed image.

Check out the `package-manifest.toml` file for details.
You will have to wait for both crucible and propolis to finish their CI runs on main/master before you will be able to get the zip archive and SHA hash needed here.

ahl (commented):
looks good

zephraph:
Is the helios failure just a flake?

sunshowers:
> Is the helios failure just a flake?

I think it might be genuine. I'm not entirely sure. I'm going to dig into it later today.

luqmana:
Looks like https://github.com/bytecodealliance/rustix/issues/467

sunshowers:
Ah OK, that's easy enough to patch in omicron. Will put up a fix for it.

luqmana (approved):
Thanks!

PR #2008, opened 2022-12-05
RFD-322 PoC -- Disks API

zephraph:
This PR is a continuation of #1957 with a focus on updating Disk APIs

## Changes

Old way of fetching disks

```
GET /organizations/<org_name>/projects/<project_name>/disks/<disk_name>
GET /by-id/disks/<disk_id>
```

New ways of fetching instance

```
GET /disks/<disk_id>
GET /disks/<disk_name>?project=<project_id>
GET /disks/<disk_name>?project=<project_name>&organization=<org_id>
GET /disks/<disk_name>?project=<project_name>&organization=<org_name>
```

Also includes the disk metrics endpoint

Before

```
GET /organizations/<org_name>/projects/<project_name>/disks/<disk_name>/metrics/<metric_name>
```

After

```
GET /disks/<disk_id>/metrics/<metric_name>
...
```

> **Note**
> `metric_name` here is a key to an enum not a `Name` type. That's why this API breaks conventions with some of the other endpoints by being deeper.


common/src/api/external/http_pagination.rs
```diff
@@ -45,7 +45,7 @@ use crate::api::external::Name;
 use crate::api::external::ObjectIdentity;
 use crate::api::external::PaginationOrder;
 use dropshot::HttpError;
-use dropshot::PaginationParams;
+pub use dropshot::PaginationParams;
```
zephraph, line 48: This is necessary due to the metrics params using `PaginationParams` directly. See that addition here: https://github.com/oxidecomputer/omicron/pull/2008/files#diff-aee01ef600c8eb547421397e5d89c41a67dd6e6ca6033676cac47b7909e1e344R123-R130

nexus/src/app/instance.rs
```diff
@@ -712,25 +706,13 @@ impl super::Nexus {
...
-            LookupPath::new(opctx, &self.db_datastore)
-                .project_id(authz_project.id())
-                .instance_name(instance_name)
-                .fetch()
-                .await?;
-
+        let (.., authz_instance) =
+            instance_lookup.lookup_for(authz::Action::Modify).await?;
+        let (.., authz_disk) =
+            disk_lookup.lookup_for(authz::Action::Modify).await?;
```
zephraph, line 715: The permissions here before were wrong I believe. It was doing a `fetch` call, which was unnecessary, and not using the right permissions.

nexus/src/app/instance.rs
```diff
@@ -760,25 +742,13 @@ impl super::Nexus {
...
```
zephraph, line 751: Same as the attach case. Doing an unnecessary fetch and not using the right permissions.

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -2307,11 +2315,75 @@ async fn ip_pool_service_range_remove(
 
 // Disks
 
+#[endpoint {
+    method = GET,
+    path = "/v1/disks",
+    tags = ["disks"],
+}]
```
zephraph, line 2322: > **Warning**
> I've taken liberties here. This code warrants extra attention. 

Previously we had
- `GET /.../projects/{project}/disks`
- `GET /.../instances/{instance}/disks`


Essentially this allows listing all disks in the first case and only disks attached to a particular instance in the second case. I've collapsed this down into a single endpoint. 

- `GET /v1/disks?project={project_id}`
- `GET /v1/disks?instance={instance_id}`

If you provide the instance selector then it filters down to disks attached to the instance. Otherwise it gives you all disks in the projects. That makes this endpoint have slightly different requirement semantics on its selectors. In general the selector for the parent resource (`project` in this case) is _required_ for list endpoints. Because the selector here can be _either_ an instance or project (and their subsequent paths) all the selectors are optional. 

Overall I think the _usage_ of this is simplified and it gives one place to go to list disks. Let me know what you think.

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -2449,24 +2609,96 @@ async fn disk_delete(
...
     };
     apictx.external_latencies.instrument_dropshot_handler(&rqctx, handler).await
 }
 
+/// Attach a disk to an instance
+#[endpoint {
+    method = POST,
+    path = "/v1/disks/{disk}/attach",
+    tags = ["disks"],
+}]
```
zephraph, line 2631: > **Warning**
> This section warrants extra attention. 

With attach and detach actions I took the same sort of approach as I did with listing in that I collapsed it down to the disks endpoint. 

Before:
- `POST /.../instances/{instance}/disks/attach { disk: Name }`
- `POST /.../instances/{instance}/disks/detach { disk: Name}`

After
- `/v1/disks/attach { instance: NameOrId }`
- `/v1/disks/detach { instance: NameOrId }`

Given the overall flatter structure of the API I think this mapping makes slightly more sense. Input welcome.

karencfv: Unfortunately I don't think this approach will be as clear as the previous one. You generally attach a disk to an instance(vm), and not the other way around. Flipping it around can potentially confuse people and the endpoint itself could be misleading. 

What do you think about something like: `/v1/instance/{instance}/attach { disk: NameOrId }` instead of `/v1/disks/{disk}/attach { instance: NameOrId }`

We can even go a bit further by having a very explicit endpoint like: `/v1/instance/{instance}/attach-disk { disk: NameOrId }`

karencfv (commented):
Nice! Another group of endpoints to tick off the list 🎉 

Left a few comments/questions let me know what you think

nexus/src/external_api/http_entrypoints.rs
karencfv: In line with https://github.com/oxidecomputer/omicron/pull/2008#discussion_r1053966938

nexus/src/app/disk.rs
```diff
@@ -20,26 +21,58 @@ use omicron_common::api::external::Error;
...
+                let disk =
+                    LookupPath::new(opctx, &self.db_datastore).disk_id(*id);
+                Ok(disk)
+            }
+            params::DiskSelector {
+                disk: NameOrId::Name(name),
+                project_selector: Some(project_selector),
+            } => {
+                let disk = self
+                    .project_lookup(opctx, project_selector)?
```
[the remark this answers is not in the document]
zephraph: It shouldn't be too much work. It's fairly formulaic.

nexus/src/external_api/http_entrypoints.rs
zephraph: I'm not sure what you mean here.

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -2449,24 +2609,96 @@ async fn disk_delete(
...
+        project_selector: query.project_selector,
+    };
+    let handler = async {
+        let opctx = OpContext::for_external_api(&rqctx).await?;
+        let instance_lookup =
+            nexus.instance_lookup(&opctx, &instance_selector)?;
+        let disk_lookup = nexus.disk_lookup(&opctx, &disk_selector)?;
+        let disk = nexus
+            .instance_attach_disk(&opctx, &instance_lookup, &disk_lookup)
+            .await?;
```
zephraph, line 2658: Note to self: This won't work. The lookup is used for the thing in the path (the disk in this case) not the thing in the body (instance).

nexus/src/external_api/http_entrypoints.rs
zephraph: I see what you're saying. The challenge with attach and detach is it's a two way action. It acts on both the disk and the instance. The benefit of the implementation I have is that there's only a single place to look for disks within the API. If clarification is the main sticking point it could just as well be updated to be `/v1/disks/{disk}/attach-to-instance` and `/v1/disks/{disk}/detach-from-instance`. If we go back to hanging it off of instances then I believe we should keep with the previous model of `/instance/{instance}/disks/*`.

zephraph: I'm going to roll this back into the original style. I found an issue in my implementation that would require updates at the Nexus app layer to address which I'd rather not do.

nexus/src/external_api/http_entrypoints.rs
davepacheco: I can see the appeal.  I'm not sure I grok the implications but it sounds reasonable!

nexus/src/app/instance.rs
```diff
@@ -712,25 +706,13 @@ impl super::Nexus {
...
-                .fetch()
-                .await?;
-        let (.., authz_instance, _) =
-            LookupPath::new(opctx, &self.db_datastore)
-                .project_id(authz_project.id())
-                .instance_name(instance_name)
-                .fetch()
-                .await?;
-
+        let (.., authz_instance) =
```
davepacheco, line 712: What happens if the disk is in a different Project?  (Also, do we have a test for that?)

Before, it wasn't possible to specify this.  Now it seems like it is.

nexus/src/external_api/http_entrypoints.rs
zephraph: Ha, well, I did an about face. This no longer applies. The endpoint is now `/v1/instances/{instance}/disks` which maps to what we had previously.

nexus/src/external_api/http_entrypoints.rs
zephraph: Okay, this is reverted to

- `/v1/instances/{instance}/disks/attach`
- `/v1/instances/{instance}/disks/detach`

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -3226,11 +3443,44 @@ async fn instance_disk_list(
...
+#[endpoint {
+    method = POST,
+    path = "/v1/instances/{instance}/disks/attach",
+    tags = ["instances"],
+}]
+async fn instance_disk_attach_v1(
+    rqctx: Arc<RequestContext<Arc<ServerContext>>>,
+    path_params: Path<params::InstancePath>,
+    query_params: Query<params::OptionalProjectSelector>,
+    disk_to_attach: TypedBody<params::DiskPath>,
```
zephraph, line 3455: I want to acknowledge that using `DiskPath` here is weird. My plan is to eventually convert all the `*Path` structs into `*Identifier` for consistency. Whether they're used in the path or provided via the body doesn't really matter given it's just different places to specify a resource.

nexus/tests/output/uncovered-authz-endpoints.txt
```diff
@@ -1,7 +1,9 @@
 API endpoints with no coverage in authz tests:
```
zephraph, line 1: I know this is a mess. I will address it.

nexus/src/app/instance.rs
zephraph: I've updated the code here. It now uses a lookup for the instance and gets the `authz_project` then does a lookup for the disk with the project id as the root of the lookup. I think what you're saying still may apply though. If you try to attach a disk by UUID then the project id in the lookup would be ignored and the disk UUID would be used instead. I'll add a check for this, good catch.

zephraph: Fix is added in [8f79dc6](https://github.com/oxidecomputer/omicron/pull/2008/commits/8f79dc630002f082eab9f655e2a03a194ebe52ce). I only added it to the attach case because the detach case only operates on whats already attached so it's an unnecessary check there. In the error message I noticed that they must be in the same project. This technically tells the user something about the relationship of the resources they may not have known, but I think that's fine because the step before should ensure they have modify permissions on both resources.

nexus/src/external_api/http_entrypoints.rs
ahl: I guess I'm fine with either. You had me convinced (and enthusiastic!) about a single endpoint with multiple search criteria, but I'm fine with the "about face". What changed your mind?

nexus/src/external_api/http_entrypoints.rs
ahl: Agreed with where you landed on this: I'm "talking to" the instance and the object of the attach/detach "verb" is the disk.

nexus/src/app/disk.rs
```diff
@@ -372,17 +373,10 @@ impl super::Nexus {
 
     pub async fn project_delete_disk(
```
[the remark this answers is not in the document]
zephraph: I've added a note in the parent issue to tackle this as a follow up.

nexus/types/src/external_api/params.rs
```diff
@@ -83,6 +89,44 @@ pub struct OptionalProjectSelector {
...
+    #[serde(flatten)]
+    pub project_selector: ProjectSelector,
+    #[serde(flatten)]
+    pub pagination: PaginatedByName,
+}
+
+#[derive(Deserialize, JsonSchema)]
+pub struct DiskMetricsList {
+    #[serde(flatten)]
+    pub pagination: PaginationParams<ResourceMetrics, ResourceMetrics>,
```
ahl, line 124: I don't understand how this works. I would expect that the second type parameter to `PaginationParams` would need to have some way to encode `disk_selector` in `disk_metrics_list_v1` or else subsequent iterations would need to continue to pass in the properties of `ProjectSelector` (FYI, progenitor assumes that this is not necessary).

I'd suggest

```suggestion
pub struct DiskMetricsListParams {
    #[serde(flatten)]
    pub metrics: ResourceMetrics,
    
    #[serde(flatten)]
    pub project_selector: Option<ProjectSelector>,
}

pub struct DiskMetricsListSelector {
    pub metrics: ResourceMetrics,
    pub disk_select: DiskSelector,
}
```

Then `disk_metrics_list_v1` would do `query_params: Query<PaginationParams<params::DiskMetricsListParams, params::DiskMetricsListSelector>>,`

ahl: I'd also note, that I think this would be easier to understand if it lived with the endpoint. The organization of endpoint/param/view means that related items are not contiguous.

nexus/types/src/external_api/params.rs
```diff
@@ -83,6 +89,44 @@ pub struct OptionalProjectSelector {
...
+        }
+    }
+}
+
+#[derive(Deserialize, JsonSchema)]
+pub struct DiskList {
+    #[serde(flatten)]
+    pub project_selector: ProjectSelector,
+    #[serde(flatten)]
+    pub pagination: PaginatedByName,
```
ahl, line 118: as below: I'm not sure how this works.

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -2307,11 +2315,49 @@ async fn ip_pool_service_range_remove(
...
+    rqctx: Arc<RequestContext<Arc<ServerContext>>>,
+    query_params: Query<params::DiskList>,
+) -> Result<HttpResponseOk<ResultsPage<Disk>>, HttpError> {
+    let apictx = rqctx.context();
+    let nexus = &apictx.nexus;
+    let query = query_params.into_inner();
+    let handler = async {
+        let opctx = OpContext::for_external_api(&rqctx).await?;
+        let project_lookup =
+            nexus.project_lookup(&opctx, &query.project_selector)?;
```
ahl, line 2333: this is invalid when query.pagination is.page matches the WhichPage::Next variant.

nexus/src/external_api/http_entrypoints.rs
zephraph: A few reasons:
- Keeping `/disks` on the end of instances is consistent with the current API so this is the minimal viable change. This is the main one. 
- A consequence of being able to select on `instance` or `project` is that they both had to be optional. That just means there's an extra failure mode here. There's also a bit more to document.

I will say that I like the single endpoint better (what I had) than I like having two endpoints returning the same type of things at different levels of the API (what's there now and before this PR). If we already have a place that we're interacting with disks associated with an instance though, it probably just makes sense to add the listing there. 

Overall I would be good to move back to the single `disk` list endpoint, but we'd need to do something with the attach/detach APIs. I don't think we'd want to have `instances/{}/disks/(attach|detach)` because the `/disks` part would be hanging with no valid methods on it. If we dropped `/disks` like @karencfv original suggestion then `/instances/{}/(attach|detach)` may be too vague. What are you attaching or detaching? Lastly I just wasn't a fan of chaining names together where we don't have to `/disks-attach`. So, that's why I ended up just going back to the original.

nexus/types/src/external_api/params.rs
```diff
@@ -83,6 +89,44 @@ pub struct OptionalProjectSelector {
     pub project_selector: Option<ProjectSelector>,
 }
 
+#[derive(Deserialize, JsonSchema)]
+pub struct DiskSelector {
+    #[serde(flatten)]
+    pub project_selector: Option<ProjectSelector>,
```
ahl, line 95: do we prefer this style to 
```suggestion
    pub project: Option<NameOrId>,
    pub organization: Option<NameOrId>,
```

Does it allow for additional code sharing for these nested, flattened selector structures?

ahl (commented):
looks good; I think there's a problem with the disk list and disk list metrics endpoints

nexus/types/src/external_api/params.rs
zephraph: By encoding a selector at every level (which itself is a resource + possible parent selector) we can flatten both the creation of selectors and the lookup functions down to only concerning itself with two cases. If you look at the endpoint handlers they're always using using a selector and potentially a path param. My thought here is that this general shape will be easier to encode in a macro. Once organization goes away and project is flattened a bit of the structure can be reduced though.

nexus/types/src/external_api/params.rs
zephraph: Good catch. Yeah, this is wrong. Let me do some house keeping.

nexus/src/app/instance.rs
davepacheco: Great.  I'm not sure what the plan is for TODO-v1 but this seems like a test we want sooner rather than later because it feels easy to break without noticing and it seems really bad if we accidentally ship it that way.

zephraph: All the `TODO-v1` comments will be addressed before switching to the v1 endpoints.

common/src/api/external/http_pagination.rs
david-crespo: Not necessary anymore?

nexus/types/src/external_api/params.rs
zephraph: Fixed in [7f608d1](https://github.com/oxidecomputer/omicron/pull/2008/commits/7f608d1d79151efb1ab0c21056f31fd688a92d9c)

nexus/src/external_api/http_entrypoints.rs
zephraph: Fixed in Fixed in [7f608d1](https://github.com/oxidecomputer/omicron/pull/2008/commits/7f608d1d79151efb1ab0c21056f31fd688a92d9c). I also updated the integration tests to verify

zephraph:
I'm going to go ahead and merge this for the sake of forward progress. @ahl if you notice anything else let me know and I'll follow up on it.

PR #2009, opened 2022-12-05
Add packaging for SoftNPU variant of Dendrite

smklein:
I'll need to spin the revs to avoid conflicting with https://github.com/oxidecomputer/omicron/pull/2007 , but this should make the SoftNPU flavor of Dendrite ready to be used.

This PR *does not add support for launching this flavor of Dendrite* -- that's follow-up work to be written in the Sled Agent! However, it does put the right flavor of Dendrite into the switch zone, which should appear in `/opt/oxide/switch.tar.gz` on the target machine.

Part of https://github.com/oxidecomputer/omicron/issues/1832

### DIFF
--- a/package-manifest.toml
+++ b/package-manifest.toml
@@ -168,8 +168,8 @@ only_for_targets.switch_variant = "softnpu"
 #
 # 1. Build the zone image manually
 #   1a. cd <dendrite tree>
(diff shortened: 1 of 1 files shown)


smklein:
Idk if this is needed anymore? been a couple months, up to you if you wanna merge something like this

PR #2020, opened 2022-12-06
update CockroachDB to v22.2.0

davepacheco:
**This cannot be merged as-is.**

- [ ] oxidecomputer/garbage-compactor#7
- [ ] need to update the expected checksums

but I did test this locally with v22.2.0 on illumos and it passed `cargo test`.


davepacheco:
We'll definitely want to go to a more recent version in this release train given these technical advisories:
https://www.cockroachlabs.com/docs/advisories/a96465
https://www.cockroachlabs.com/docs/advisories/a97178

davepacheco:
Also this one:
https://www.cockroachlabs.com/docs/advisories/a93398

ahl:
Is this a priority for MVP or something we intend to defer. If the former, we should probably do it ASAP. If the latter... so much the better?

davepacheco:
I've set this to milestone MVP+1.  My hope was that we could squeeze this in between other work, but it's fair to be worried about updating this dependency late in the game.

Candidly: it makes me nervous to ship with a CockroachDB that's a year old.  Like dependabot updates and Rust toolchain updates, this sort of thing is basically not that important (potentially for years) until it suddenly becomes very important, but at that point it's a much larger amount of suddenly-urgent work.  That said, I can't see doing this more urgently than anything else on the MVP board.

For what it's worth, I think this isn't that much work.  It's basically just blocked on oxidecomputer/garbage-compactor#7, which is essentially just blocked on getting 1-2 new IPS packages.  (It's possible there are breaking changes between what I tested in that ticket and what we'd actually be moving to, but given we're still talking about the same major release of CockroachDB, I hope that won't be the case.)

davepacheco:
Closing per RFD 508.

PR #1907, opened 2022-11-01
Use Oxide pkg repo for installing / uninstall OPTE and kernel bits

bnaecker:

PR #2021, opened 2022-12-07
move xde onu to stlouis onu

rcgoodfellow:
We are trying to migrate workflows from the xde and integration illumos branches to the stlouis branch. This commit updates the opte install script to grab the stlouis onu/p5p build from the os-build repo instead of the xde branch build.

I've tested this by setting up a RAP environment and the OPTE/XDE plumbing comes up successfully.


bnaecker (approved):
LGTM, thanks! Would be good to give this the same treatment as https://github.com/oxidecomputer/omicron/pull/1907, pulling from a hosted repo rather than installing the p5p archive locally. But that can wait!

luqmana (approved):

PR #2022, opened 2022-12-07
Remove instance serial-console buffer from sled-agent

lifning:
This functionality has been migrated into propolis-server: https://github.com/oxidecomputer/propolis/pull/266

The sled-agent-sim -backed integration test for nexus' serial console API now works against a [mock propolis-server](https://github.com/oxidecomputer/propolis/pull/279) running on localhost.

(This is all in service of making it reasonable to implement migration of serial console connections and history along with instance state)


nexus/tests/integration_tests/instances.rs
```diff
@@ -2713,6 +2713,7 @@ async fn test_instances_memory_not_divisible_by_min_memory_size(
     );
 }
 
+#[ignore] // API no longer in sled-agent(-sim). TODO: another way to test w/o propolis?
```
smklein, line 2716: We do some Shenanigans to test similar things with crucible from the integration test suite.

If you're running this test suite in a simulated environment, obviously crucible downstairs doesn't exist, because you don't have disks.

Instead, the test suite [directly calls to the sled agent](https://github.com/oxidecomputer/omicron/blob/27dfb62275b92b960a3e2a1c0644cfca492f3faf/nexus/test-utils/src/resource_helpers.rs#L466-L494), calling methods like [`create_zpool` and `create_crucible_dataset`](https://github.com/oxidecomputer/omicron/blob/27dfb62275b92b960a3e2a1c0644cfca492f3faf/sled-agent/src/sim/sled_agent.rs#L269-L281), which just lands in a [simulated implementation of storage](https://github.com/oxidecomputer/omicron/blob/27dfb62275b92b960a3e2a1c0644cfca492f3faf/sled-agent/src/sim/storage.rs). This "simulated implementation" basically just stupidly keeps track of all resources provisioned (it's basically just a few hashmaps of "UUID" -> "what was the request that created this object"). Obviously you can't use this for any real storage, but this means that we can test all the "disk provisioning" stuff in Nexus, and verify what requests *should* have made it down to the sled agent.

Here's the part that I think might be relevant for you:
- In the *real* implementation of Crucible, we expect that each downstairs will be sitting on an HTTP server, responding to requests from Nexus directly....
- ... So in our *simulated* implementation, we [create a server](https://github.com/oxidecomputer/omicron/blob/27dfb62275b92b960a3e2a1c0644cfca492f3faf/sled-agent/src/sim/storage.rs#L384) which [re-implements the crucible endpoints](https://github.com/oxidecomputer/omicron/blob/27dfb62275b92b960a3e2a1c0644cfca492f3faf/sled-agent/src/sim/http_entrypoints_storage.rs) by calling into the "stub implementation".

I suspect we could do something similar for instances:
- When the simulated sled agent should be creating an instance, we could spin up a "simulated" implementation of the propolis server.
- We could implement whatever endpoints would help you write this test, returning whatever data is convenient
- We could expose APIs out of the simulated sled agent to "get direct access to the stub propolis server", so you could check that structure during test.

lifning: I see! That makes a lot of sense, thanks :D

nexus/tests/integration_tests/instances.rs
```diff
@@ -2921,6 +2925,9 @@ async fn test_instance_serial(cptestctx: &ControlPlaneTestContext) {
             .await;
 
     // Now, simulate completion of instance boot and check the state reported.
+    // TESTABILITY HACK: since we told sled-agent-sim to run a mock propolis,
+    // it will also tell the nexus that its propolis IP is localhost as part
+    // of this exchange. (nexus's call to sled-agent instance_put)
```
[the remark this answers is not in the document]
lifning: I think I'll rephrase it to make it more clear, but basically: prior to this step, nexus's database is "wrong" about propolis's IP, because nexus "assigned" it one with its usual logic.
specifically how in the call to instance_simulate -> instance_finish_transition:
  - nexus makes API call to sled-agent-sim's instance_poke to tell it to pretend the instance finished booting
  - sled-agent-sim makes API call to nexus's cpapi_instances_put with the updated InstanceRuntimeState that *now also contains the mock propolis server's IP* which overwrites the one nexus had allocated via its usual instance-creating saga
  - *now* we can successfully connect to nexus's serial console endpoints.
 
This felt unusual enough to me that I wanted to write something down here to call it out, as `instance_simulate` otherwise would strike me as only being used for changing the runtime state of the instance, and not hiding any additional load-bearing magic like this -- and of course this is specific to the test use case where Nexus *would* be immediately wrong about the IP it allocated for the instance in the first place.

sled-agent/src/sim/sled_agent.rs
```diff
@@ -232,6 +238,66 @@ impl SledAgent {
                 .await?;
         }
 
+        // if we're making our first instance and a mock propolis-server
+        // is running, interact with it, and patch the instance's
+        // reported propolis-server IP for reports back to nexus.
+        let mock_lock = self.mock_propolis.lock().await;
```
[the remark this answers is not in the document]
lifning: this constraint is due to the aforementioned issue where the control plane is hardcoded to expect propolis-server to run on `:12400`. we should absolutely run one per UUID after fixing that

nexus/src/app/instance.rs
```diff
@@ -857,53 +857,54 @@ impl super::Nexus {
...
         Ok(params::InstanceSerialConsoleData {
-            data: sa_data.data,
-            last_byte_offset: sa_data.last_byte_offset,
+            data: data.data,
+            last_byte_offset: data.last_byte_offset,
         })
     }
 
+    // TODO: Technically the stream is two way so the user of this method can
+    // modify the instance in some way. Should we use different permissions?
```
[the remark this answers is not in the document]
lifning: I think rather than move it I'll just do the tiny refactor to add the check to propolis_client_for_instance

sled-agent/src/sim/sled_agent.rs
```diff
@@ -445,4 +419,57 @@ impl SledAgent {
 
         Ok(())
     }
+
+    /// Used for integration tests that require a component to talk to a
+    /// mocked propolis-server API.
+    // TODO: fix schemas so propolis-server's port isn't hardcoded in nexus
```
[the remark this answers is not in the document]
lifning: https://github.com/oxidecomputer/omicron/issues/2457

smklein (approved):
Sorry for the delay on this! This LGTM, and we should merge, and IMO those follow-ups will make testability a lot cleaner.

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -3392,20 +3392,22 @@ async fn instance_serial_console(
 async fn instance_serial_console_stream_v1(
     rqctx: RequestContext<Arc<ServerContext>>,
     path_params: Path<params::InstancePath>,
-    query_params: Query<params::OptionalProjectSelector>,
+    query_params: Query<params::InstanceSerialConsoleRequest>,
```
david-crespo, line 3395: On `main` this is fixed by taking both sets of query params:

```rs
query_params: Query<params::InstanceSerialConsoleRequest>,
selector_params: Query<params::OptionalProjectSelector>,
```

lifning: dropshot doesn't currently like that for this function, though - too many params.
> the trait `dropshot::handler::HttpHandlerFunc<_, _, _>` is not implemented for fn item `fn(RequestContext<Arc<context::ServerContext>>, dropshot::Path<InstancePath>, dropshot::Query<InstanceSerialConsoleRequest>, dropshot::Query<OptionalProjectSelector>, WebsocketUpgrade) -> impl futures::Future<Output = Result<http::Response<hyper::Body>, dropshot::HttpError>> {external_api::http_entrypoints::<impl From<instance_serial_console_stream> for ApiEndpoint<<RequestContext<Arc<context::ServerContext>> as RequestContextArgument>::Context>>::from::instance_serial_console_stream}`

however, for reasons that currently escape me, adding `impl_HttpHandlerFunc_for_func_with_params!((0, T1), (1, T2), (2, T3), (3, T4));` to https://github.com/oxidecomputer/dropshot/blob/main/dropshot/src/handler.rs#L357 was insufficient to resolve this.

david-crespo: Weird! That will be an @ahl question I think.

lifning: it occurs to me now that the removal of the non-v1 API call means there's no reason not to just throw it in with the rest of InstanceSerialConsoleRequest. so i'm doing that

Issue #1482, opened 2022-07-22
Do not delete Projects with child resources


bnaecker:
Promoting [this TODO](https://github.com/oxidecomputer/omicron/blob/8d13230e1f50106818e262cae12ec9caf6e6f514/nexus/src/db/datastore/project.rs#L69) to an issue. We need to check for child resources in a Project before deleting it, by adding an `rcgen` field to the database and checking it [similar to other resources](https://github.com/oxidecomputer/omicron/blob/main/nexus/src/db/datastore/ip_pool.rs#L122-L165).

Issue #1334, opened 2022-06-30
Implement Floating IPs


bnaecker:
See https://github.com/oxidecomputer/omicron/issues/1333 for some general background on external IPs for instances. Floating IPs are the second form of IP addresses that allow inbound connectivity to guests. In contrast to Ephemeral IPs, which only live as long as instances, Floating IPs are API resources in their own right. They can be moved between objects, and are not deleted when the instance they're attached to is stopped. Similar to Ephemeral IPs, they allow inbound and outbound connectivity, and reserve the whole port range for the guest. Also, to Ephemeral IPs, Floating IPs can reference a specific IP Pool at creation time.

This work tracks adding the database representation and HTTP API for Floating IPs, and assigning them to guests for instance NAT when requested.

NOTE: None of this takes into account the current workaround in OPTE to allow inbound connections via the external IPs meant for source NAT. That will go away, hopefully very soon, once boundary services can be stood up to perform the decapsulation of guest traffic from OPTE.

bnaecker:
Closed by #1458

bnaecker:
Woops, closed the wrong issue.

FelixMcFelix:
I'm currently working on the bits to handle this in [opte#420](https://github.com/oxidecomputer/opte/pull/420), after which I'm happy to tackle the API/DB side.

Issue #1015, opened 2022-05-05
Implement Project-level defaults


bnaecker:
As part of [RFD 21](https://rfd.shared.oxide.computer/rfd/0021#_concepts) and the discussion on #1003, we'd like to implement defaults at the level of a Project. The canonical example from RFD 21 is a default VPC Subnet, in which we create NICs for a guest if one is not explicitly provided in the client request. Defaults should be configurable resources, at the Project scope, for example via CRUD on an endpoint like `/organizations/{org}/projects/{project}/defaults`. We'll need a new table in CRDB to manage these. Below is a preliminary list of the defaults we'd like to include, but more should be added as they make sense.

- VPC
- VPC Subnet
- Image from which guest instances are created
- Whether guest NICs default to dual IPv4/v6, or exactly one of those. As an example, if the default is to provide both versions and the client requests only a specific IPv4 address, an IPv6 address is added automatically.

bnaecker:
@rmustacc Another thing occurs to me getting back up to speed on IP pools and instance NAT. In order to meet the goal that every instance should be able to reach the internet by default (assuming FW and routing rules allow it), we'll need to provide a unique (IP, port-range) tuple for each instance, which OPTE would use for that instance's outbound connectivity. If that's correct (a man-sized "if"), it seems we might want to reflect part of this in the project-level defaults. Something like the size of the port-range might make sense, to represent the default NAT quota. Let me know if you have thoughts here.

bnaecker:
Something about this concept doesn't quite make sense to me. It's been rattling around in the back of my head, and I think I know what it is now. Consider this quote from RFD 21:

> When a project is created, a default VPC is created as well.

If that's true, how is that information populated? It can't refer to the project-level defaults, because those defaults don't exist yet. The user may be able to change them, but only after the project (and default VPC) have been created. It seems like we need a bunch of default values for these Project-level defaults, _or_ we need to have a way to change them at a higher level than a project, such as an Organization.

I guess an alternative is to provide a flag to the Project creation endpoint that _prevents_ creating that default VPC. The user could set that, change the project-level defaults, and then create the default VPC again. Though at that point, why have the defaults at all?

bnaecker:
[...]

Issue #1168, opened 2022-06-07
[db] Audit usage of "ON CONFLICT DO NOTHING RETURNING ..." 


smklein:
(here's some [background](https://stackoverflow.com/questions/34708509/how-to-use-returning-with-on-conflict-in-postgresql) that covers this use-case)

There are a handful of spots in the datastore where we:
- INSERT INTO a table,
- ON CONFLICT, DO NOTHING
- and have a RETURNING clause

Example: https://github.com/oxidecomputer/omicron/blob/3ae28c77cbd2355abeab07e358c5ccbeb035f641/nexus/src/db/datastore.rs#L880-L884

This is valid SQL, but it's a little suspect:

- If there are no conflicts (we're inserting the object with the UUID for the first time) we get the value that we intended to insert.
- However, if there *are* conflicts, this group of statements does not return any rows. **Instead, it results in a `NotFound` error from the database**.

This means that, especially in situations where we expect idempotency, this statement may return different results if invoked multiple times.

smklein:
Avoiding `do_nothing()`, and replacing it with `do_update().set(...)`, for *any* field, seems to be a reasonable workaround?

davepacheco:
Maybe?  The top answer in the linked Stackoverflow post explains a bunch of unfortunate side effects of that, though some of those may not apply in CockroachDB.

I feel like we had a chunk of code for doing exactly this a long time ago.  Does that ring a bell?

smklein:
I googled "ON CONFLICT DO NOTHING RETURNING" and found [this](https://stackoverflow.com/questions/34708509/how-to-use-returning-with-on-conflict-in-postgresql); I'm assuming this is what you're referencing?

This answer mentions:

> The [approach of updating anyway] seems ok for a single conflict target, few conflicts, small tuples and no triggers.

- Our conflict target is nearly always a UUID, so it's unique
- So there are few conflicts
- This data is relatively small
- CRDB is not using triggers

I do not recall of any code to help in this case of "insert or read the record idempotently".

Their proposal of:

- Use a CTE to generate all the potential values to be inserted
- Attempt to INSERT into the table, DO NOTHING on return
- SELECT the values which were potentially inserted, and...
- ... UNION those values with all the potential values to be inserted

Definitely could also work, and avoid the write contention.

They mention:

[...]

PR #2024, opened 2022-12-07
[nexus] Safer project deletion

smklein:
## Previously

Project deletion was very problematic - even though projects contain many resources, they could be deleted while their children objects still exist. This would effectively "orphan" those resources, making them inaccessible.

For example: Create an instance in a project, delete the project. Now the instance exists, but cannot be deleted.

## This PR

Adds `rcgen` to projects, and ensures that they can  no longer be deleted while containing child resources.

- Implements `DatastoreCollectionConfig<ChildResource>` for `Project`, for many child resources
- Uses `Project::insert_resource` to bump `rcgen` while creating child resources
- Checks for child resources existing during project deletion
- Adds tests

Areas worth a particular close review:
- **Coping with the default VPC**. When projects are created, they are initialized with a [default VPC](https://github.com/oxidecomputer/omicron/blob/eb15e326976b4eea7e3e2545251290fdcb743344/nexus/src/app/project.rs#L46-L72) named `default`, as described in [RFD 21](https://rfd.shared.oxide.computer/rfd/21#_concepts). This default VPC may additionally contain subnets - by default, a subnet named `default` is also created, but more may exist. **At the time that we delete a project, we need to either delete these resources ourselves, or verify that they have been deleted by the user**.
  - This PR currently explicitly deletes the default VPC / Subnet from the client-side, but I'm open towards performing that deletion automatically within the "delete project" request. However, it would be difficult to avoid doing so in a transaction - presumably if we succeed in deleting the default VPC, but fail in deleting the project, we'd want to "undo" deletion of the VPC.
  - Reading https://github.com/oxidecomputer/omicron/issues/1015 , it seems like users will have explicit control over the creation of the default VPC -- this makes me lean slightly more towards "deletion of the default VPC must be done manually; it's just another VPC".
- **IP Pool Ranges and External IPs**. These get special treatment:
  - IP pool ranges should only exist within IP pools, so they are not treated as "direct children of projects". VPC subnets are also like this, but IP pool ranges contain a direct reference to `project_id`, which makes this seem a little quirkier to me.
  - Floating external IPs are not implemented (see: https://github.com/oxidecomputer/omicron/issues/1334), so we cannot fully implement them safely bumping `rcgen` on creation. Regardless, we do check for their absence when deleting projects. 
 
Before taking out of draft:
- [x] Deal with the default VPC in projects never being deleted
- [x] Fix the end-to-end test, which seems broken by this: https://github.com/oxidecomputer/omicron/pull/2024/commits/6400f95723a59cbfd7c7f83bc3fb81a9776448f6
- [x] Add tests for projects still containing these resources
  - [x] Add external IP test: https://github.com/oxidecomputer/omicron/pull/2024/commits/bef880a71385cdd38fdced3e6d4b2460462d2734
  - [x] See if we can stop the instance during creation; that would simplify testing: https://github.com/oxidecomputer/omicron/pull/2024/commits/7605115e4c559ca3e1d94ea9ad1172229baf946d
- [x] Implement (or check if it's necessary?) the rcgen bump on `ExternalIp` creation: https://github.com/oxidecomputer/omicron/pull/2024/commits/bef880a71385cdd38fdced3e6d4b2460462d2734

Fixes https://github.com/oxidecomputer/omicron/issues/1482


nexus/src/db/datastore/snapshot.rs
```diff
@@ -8,66 +8,62 @@ use super::DataStore;
...
+        use db::schema::snapshot::dsl;
+        let project_id = snapshot.project_id;
+        let snapshot: Snapshot = Project::insert_resource(
+            project_id,
+            diesel::insert_into(dsl::snapshot)
+                .values(snapshot)
+                .on_conflict((dsl::project_id, dsl::name))
+                .filter_target(dsl::time_deleted.is_null())
+                .do_update()
+                .set(dsl::time_modified.eq(Utc::now())),
```
smklein, line 54: This is one example case of https://github.com/oxidecomputer/omicron/issues/1168 .

The old code issued a query to ensure the snapshot exists, and then proceeded to perform a lookup to find the snapshot -- two queries instead of one.

nexus/src/db/datastore/project.rs
```diff
@@ -65,33 +66,241 @@ impl DataStore {
...
+
         use db::schema::project::dsl;
 
         let now = Utc::now();
-        diesel::update(dsl::project)
+        let updated_rows = diesel::update(dsl::project)
             .filter(dsl::time_deleted.is_null())
             .filter(dsl::id.eq(authz_project.id()))
+            .filter(dsl::rcgen.eq(db_project.rcgen))
             .set(dsl::time_deleted.eq(now))
```
sunshowers, line 287: Does this not also update the rcgen?

nexus/tests/integration_tests/projects.rs
```diff
@@ -72,3 +85,366 @@ async fn test_projects(cptestctx: &ControlPlaneTestContext) {
...
+        StatusCode::BAD_REQUEST,
+        Method::DELETE,
+        &url,
+    )
+    .authn_as(AuthnMode::PrivilegedUser)
+    .execute()
+    .await
+    .unwrap()
+    .parsed_body()
+    .unwrap();
```
sunshowers, line 131: In the past I've found it a bit hard to figure out test failures with multiple `unwrap`s in a single statement. Could I convince you to replace these with `expect` calls? :)

nexus/tests/integration_tests/basic.rs
```diff
@@ -242,16 +242,33 @@ async fn test_projects_basic(cptestctx: &ControlPlaneTestContext) {
     assert_eq!(project.identity.description, expected.identity.description);
     assert!(project.identity.description.len() > 0);
 
-    // Delete "simproject2".  We'll make sure that's reflected in the other
-    // requests.
+    // Delete "simproject2", but first delete:
+    // - The default subnet within the default VPC
+    // - The default VPC
+    NexusRequest::object_delete(
```
sunshowers, line 248: Out of curiosity, do we have (and/or does this PR add) tests for the sorts of races that `rcgen` is meant to prevent?

sunshowers (commented):
Thanks for adding me to the review! I spent quite a bit of time looking at this and it was really helpful to me. I've put in a few comments where I had further questions but overall this makes sense to me.

nexus/db-model/src/project.rs
```diff
@@ -40,6 +48,61 @@ impl From<Project> for views::Project {
...
+// but needed for project deletion safety.
+// TODO(https://github.com/oxidecomputer/omicron/issues/1334): Cannot be
+// utilized until floating IPs are implemented.
+impl DatastoreCollectionConfig<ExternalIp> for Project {
+    type CollectionId = Uuid;
+    type GenerationNumberColumn = project::dsl::rcgen;
+    type CollectionTimeDeletedColumn = project::dsl::time_deleted;
+    type CollectionIdColumn = external_ip::dsl::project_id;
+}
+
```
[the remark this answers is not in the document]
smklein: Ehhh that's kinda a blurry line. We do sometimes (and I wouldn't be opposed to having one!), but generally I've favored "feel the initial pain of writing it out a lot" rather than "be as brief as possible", until the benefit of the macro is immediately obvious.

I think right now we have a few attribute macros for DB queries and metrics-querying support, and a handful of macros for authz stuff? `nexus/db-macros` and `nexus/authz-macros` are examples. 

But yeah, that question of readability is a bit subjective -- do you think this would benefit from a macro? My opinion *had* been that this `impl` is specifying a handful of fields we'd want to supply through a macro anyway, so I figured a macro would just obscure the intent, but I'm curious if you have an example in mind where this might seem clearer.

nexus/src/db/datastore/project.rs
smklein: Nah, it doesn't need to. The whole point of the rcegen number is that it allows us to do:

1. Check if children exist
2. Try to update project as "time_deleted.eq(now)"

As if those two operations were **atomic**.

- If a new child *was* inserted when we do (2) the rcgen should change. This is the only spot we "read" the rcgen.
- If no new children are inserted when we do (2), the operation can know that (because it already checked) no children previously existed, and no children exist now.

So we don't need to update the value in this code path -- if we filter this record and see any project, it's gonna be deleted anyway!

nexus/tests/integration_tests/basic.rs
smklein: Good question, and honestly, not *really*?

We do have a test on the general "rcgen should increment" behavior: https://github.com/oxidecomputer/omicron/blob/5fe955d946ed2f962f56a5278e8939c1d62ef688/nexus/src/db/collection_insert.rs#L631-L639

But we don't have a reliable way to:
- Start the delete operation
- Check that "no children exist"
- (Separate thread of execution) Insert a new child resource
- Show that deletion fails because of the rcgen mismatch

(Should we? Probably!)

nexus/db-model/src/project.rs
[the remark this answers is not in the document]
sunshowers: Oh I was thinking of a private macro here, not a public API. Though the primary benefit I've found is if we expect many more cases and it enhances code clarity (e.g. [this is a place where macros make life better](https://github.com/camino-rs/camino/blob/61c16e917ac365b0ebecf5559f69a0c75ec9236b/src/lib.rs#L2796-L2806)). I don't think this crosses that bar.

nexus/src/app/project.rs
```diff
@@ -153,12 +153,15 @@ impl super::Nexus {
...
-            .organization_name(organization_name)
-            .project_name(project_name)
-            .lookup_for(authz::Action::Delete)
-            .await?;
-        self.db_datastore.project_delete(opctx, &authz_project).await
+        let (.., authz_project, db_project) =
+            LookupPath::new(opctx, &self.db_datastore)
+                .organization_name(organization_name)
+                .project_name(project_name)
+                .fetch()
```
davepacheco, line 160: super nitty (and up to you): I think you can use `fetch_for(authz::Action::Delete)` here, though it probably makes no difference since `DataStore::project_delete()` will have to check anyway.

nexus/src/db/datastore/disk.rs
```diff
@@ -64,19 +67,29 @@ impl DataStore {
...
         let name = disk.name().clone();
-        let disk: Disk = diesel::insert_into(dsl::disk)
-            .values(disk)
-            .on_conflict(dsl::id)
-            .do_nothing()
-            .returning(Disk::as_returning())
-            .get_result_async(self.pool())
-            .await
-            .map_err(|e| {
+        let project_id = disk.project_id;
```
davepacheco, line 70: Really how neat it is that this was so straightforward (at least this part), the way `collection_insert` was built.

nexus/src/db/datastore/ip_pool.rs
```diff
@@ -158,17 +159,45 @@ impl DataStore {
         };
         let pool = IpPool::new(&new_pool.identity, project_id, rack_id);
         let pool_name = pool.name().as_str().to_string();
-        diesel::insert_into(dsl::ip_pool)
-            .values(pool)
-            .returning(IpPool::as_returning())
-            .get_result_async(self.pool_authorized(opctx).await?)
+
+        if let Some(project_id) = project_id {
+            Project::insert_resource(
```
davepacheco, line 164: This is probably not a new issue, but do we not need to do an authz check here?

If we did, then at L171 I think you could use `authz_resource.not_found()`.

nexus/src/db/datastore/instance.rs
```diff
@@ -66,22 +69,32 @@ impl DataStore {
...
+            project_id,
+            diesel::insert_into(dsl::instance)
+                .values(instance)
+                .on_conflict(dsl::id)
+                .do_nothing(),
+        )
+        .insert_and_get_result_async(self.pool())
+        .await
+        .map_err(|e| match e {
+            AsyncInsertError::CollectionNotFound => Error::ObjectNotFound {
```
davepacheco, line 84: How is there no authz check here?  If there were, you could use `authz_project.not_found()` instead of making the error by hand here.

nexus/src/db/datastore/disk.rs
```diff
@@ -64,19 +67,29 @@ impl DataStore {
...
+            project_id,
+            diesel::insert_into(dsl::disk)
+                .values(disk)
+                .on_conflict(dsl::id)
+                .do_nothing(),
+        )
+        .insert_and_get_result_async(self.pool())
+        .await
+        .map_err(|e| match e {
+            AsyncInsertError::CollectionNotFound => Error::ObjectNotFound {
```
davepacheco, line 82: Mentioned elsewhere, but it's surprising there's no authz check here.  If there were, you could use `authz_disk.not_found()` to generate the `ObjectNotFound` error.

nexus/src/db/datastore/project.rs
```diff
@@ -65,33 +66,241 @@ impl DataStore {
         })
     }
 
+    async fn ensure_no_instances_in_project(
```
davepacheco, line 69: Feels like a lot of boilerplate around these functions but I don't have any particular suggestions to improve it.

nexus/src/db/datastore/project.rs
```diff
@@ -65,33 +66,241 @@ impl DataStore {
...
+                .filter(instance::dsl::time_deleted.is_null())
+                .select(instance::dsl::id)
+                .limit(1)
+                .first_async::<Uuid>(self.pool_authorized(opctx).await?)
+                .await,
+        )
+        .map_err(|e| public_error_from_diesel_pool(e, ErrorHandler::Server))?;
+        if instance_found.is_some() {
+            return Err(Error::InvalidRequest {
+                message: "project to be deleted contains an instance"
```
davepacheco, line 89: Would it be possible (or, easy) to include information about the thing that we found in the internal message?  I'm thinking about debugging a code path like this (e.g., we find a case where it's unexpectedly failing to delete a Project because some Instance is still around, but we don't think there's an Instance around).  It could make a world of difference if we knew which instance Nexus thought was keeping this Project alive, particularly if the problem winds up being transient.

Same with the other functions.

nexus/src/db/datastore/snapshot.rs
```diff
@@ -8,66 +8,62 @@ use super::DataStore;
...
-
     pub async fn project_ensure_snapshot(
         &self,
         opctx: &OpContext,
         authz_silo: &authz::Silo,
         snapshot: Snapshot,
     ) -> CreateResult<Snapshot> {
-        let snapshot_id = snapshot.id();
         let gen = snapshot.gen;
+        opctx.authorize(authz::Action::CreateChild, authz_silo).await?;
```
davepacheco, line 43: We're potentially creating a Snapshot inside a Project, right?  Why is this checking for permission to create something inside a Silo?

nexus/src/db/datastore/snapshot.rs
davepacheco: Did you mean this to set the time modified to itself rather than `now()`?  It seems like this will adjust the modification time even though no modification has been made.

davepacheco (commented):
Nice!

nexus/src/db/datastore/project.rs
smklein: Done in 0249d63ed7b7317167ab8bfb961bfd87ca603ae8 via a macro; I'm not sure if this is better or worse

nexus/src/db/datastore/project.rs
smklein: Done in 31ec7651c429503deb11d06b966da9eb5592fc11

smklein:
I think I've addressed all comments requesting changes - if y'all have a chance, I'd appreciate another look

nexus/tests/integration_tests/projects.rs
```diff
@@ -146,12 +146,12 @@ async fn test_project_deletion(cptestctx: &ControlPlaneTestContext) {
     // Project deletion will fail while the subnet & VPC remain.
     create_project(&client, &org_name, &name).await;
     assert_eq!(
-        "project to be deleted contains a vpc",
+        "project to be deleted contains a vpc: default",
```
davepacheco, line 149: ooh, this is nice

davepacheco (approved):
Thanks for fixing the authz issues!

nexus/tests/integration_tests/basic.rs
sunshowers: This is where tokio's [`time::pause`](https://docs.rs/tokio/latest/tokio/time/fn.pause.html) can come in handy -- you can use it to insert arbitrary interleavings between tasks. This can also be used with property-based tests -- we should maybe consider this at some point in the future :)

sunshowers (commented):
Looks good to me, though I'm more of a learner than anything else here!

PR #2025, opened 2022-12-07
[wicketd-client] don't depend on omicron-common

sunshowers:
omicron-common depends on dropshot, and wicketd-client (e.g.
installinator) doesn't need an HTTP server.


andrewjstone (approved):
LGTM!  I have no idea why I put it there in the first place :)

PR #2027, opened 2022-12-07
[installinator] initial version

sunshowers:
Implement the initial version of the "installinator" described in [RFD 345](https://rfd.shared.oxide.computer/rfd/0345#_host_phase_2_install).

Per the RFD, the installinator is "[t]he key process in the Phase-2 Trampoline Image that fetches data over the bootstrap network to install on an M-2 device."

> 2.6 Host Phase-2 Install
>
> At the start of this step, the host has booted, it’s running user-land processes, the bootstrap network (and not the underlay network) is available (meaning that T6 and Maghemite are both up). At the conclusion of this step, the Phase-2 Host Image will be present on an M.2 device (of the two).
>
> The Trampoline image contains a single-purpose program called Installinator that downloads the Phase-2 Host Image and places it on an M.2. Installinator discovers peers on the bootstrap network which includes sled-agent and wicketd instances; it connects to each and requests the desired Phase-2 Host Image by ID and/or checksum.
>
> Every sled-agent serves data from its own, current M.2; wicketd serves additional images that have been uploaded to it.

This isn't complete yet; the TODOs are tracked in https://github.com/oxidecomputer/omicron/issues/2061.

Questions:

- [x] Is it OK to buffer the downloaded artifact to memory?
  - Answer is *yes* for V1. ~~In the future, we might want to write directly out to disk as we verify it (assuming we do streaming verification).~~ (This is not an issue for a long time because we have orders of magnitude more RAM than the size of the host image). 
- [x] Currently we attempt fetches from all peers and take the first one that can serve us the entire artifact. Do we need to first check for the presence of the artifact then attempt fetches one-by-one? (This is going to depend on how often the installinator runs and how many peers are expected to have the artifact, I guess).
  - Resolved in retry/fetch loop.


installinator/src/peers.rs
```diff
@@ -0,0 +1,161 @@
...
+        let download_futs: FuturesUnordered<_> = self
+            .imp
+            .peers()
+            .iter()
+            .map(|peer| async {
+                let result =
+                    self.imp.fetch_artifact_from_peer(peer, artifact_id).await;
+                (*peer, result)
+            })
+            .collect();
```
sunshowers, line 41: One thing we may want to do here is rather than using `FuturesUnordered` which bundles everything up into a single task, use `tokio::spawn` + mpsc and broadcast channels. Which is better depends on the size of the artifacts and whether we want to try downloading sequentially or in parallel, I think.

smklein:
Solid start! The use of mocks hopefully means we can add tests for the expected behaviors (and failure handling), even without running in a fully realistic environment. 

>  Is it OK to buffer the downloaded artifact to memory?

Probably yes? At this point, we should *definitely* not be relying on U.2 storage, so the M.2s are basically the only place we *could* use to write. But the whole point of this program is to stream an artifact into the M.2 *anyway*, on the slice for the host OS. So if we *were* going to buffer it to a file in local storage, we might as well just write it directly to the target slice.

We could definitely stream it, for example, keeping a 1 GiB in-memory buffer, and writing it out gradually to the M.2 slice. That might not be necessary for the v1 implementation though.

>  Currently we attempt fetches from all peers and take the first one that can serve us the entire artifact. Do we need to first check for the presence of the artifact then attempt fetches one-by-one? (This is going to depend on how often the installinator runs and how many peers are expected to have the artifact, I guess).

IMO this operation:
- Is rare
- Only needs one peer to succeed (and all *functional* peers should have the artifact, though another sled running the trampoline will not)
- *Really* needs to succeed when it can

With that in mind, I'd consider a policy like:
1. Populate a list of peers, `P`
2. Select one peer `p` ∈ `P`, try to download the artifact
3. If we can get the artifact, great, move on. Verify / write the host OS, boot it.
4. If we can't (for *whatever* reason - it's bad, the peer doesn't have it, network drops) repeat step (2) for a new `p`  ∈  `P`
5. If we can't download the artifact, repeat step (1), and try again.

If / how we choose to make these operations concurrent is up to you, but I think we should functionally fall back to this type of behavior if we happened to fail accessing the artifact.

sunshowers:
> We could definitely stream it, for example, keeping a 1 GiB in-memory buffer, and writing it out gradually to the M.2 slice. That might not be necessary for the v1 implementation though.

Not going to worry too much about this for now, but if we want to stream and write in chunks while verifying that all the bytes we receive are valid, we could probably use a verified streaming crate such as [bao](https://docs.rs/bao).

I've implemented the strategy you mentioned -- I think this should work fine for now. I've added a bunch of retries and timeouts to various spots. Going to focus on testing this out now.

installinator/src/peers.rs
```diff
@@ -0,0 +1,334 @@
...
+    pub(crate) attempt: usize,
+    pub(crate) addr: Ipv6Addr,
+    pub(crate) artifact: BufList,
+}
+
+impl FetchedArtifact {
+    /// In a loop, discover peers, and fetch from them.
+    ///
+    /// This only produces an error if the discover function errors out. In normal use, it is expected
+    /// to never error out.
```
smklein, line 32: IMO this might be more clear as:

```suggestion
    /// If `discover_fn` returns an error other than `DiscoverPeersError::Retry`, that error is returned.
    /// Otherwise, this function continues looping until it succeeds.
```

installinator/src/peers.rs
```diff
@@ -0,0 +1,334 @@
...
+            attempt += 1;
+            let peers = match discover_fn().await {
+                Ok(peers) => peers,
+                Err(DiscoverPeersError::Retry(error)) => {
+                    slog::debug!(
+                        log,
+                        "(attempt {attempt}) failed to discover peers, retrying: {}",
+                        DisplayErrorChain::new(AsRef::<dyn std::error::Error>::as_ref(&error)),
+                    );
+                    // XXX: add a delay here?
```
smklein, line 53: FWIW a fixed-duration async sleep on the order of milliseconds seems reasonable as a first pass, just to avoid slamming the CPU?

installinator/src/peers.rs
```diff
@@ -0,0 +1,334 @@
...
+            .finish()
+    }
+}
+
+#[derive(Debug)]
+pub(crate) struct Peers {
+    log: slog::Logger,
+    imp: Box<dyn PeersImpl>,
+    timeout: Duration,
+}
```
smklein, line 104: I'm a little confused by the layering here.

We have:

- `struct Peers`: Looks... mostly real? but exposes some "mock" methods? But also *wraps* a `dyn` trait which could be used to inject a mock peer?
- `trait PeersImpl`: Seems like a decent spot to split between mock / not-mocked peers?
- `struct MockPeers` (in `peers.rs`): Contains a `Vec<Ipv6Addr>`, and I don't actually know how "mocked" this is supposed to be? There are comments inside about reaching out over the boostrap network?
- `struct MockPeers` (in `mock_peers.rs`): Contains a `BTreeMap<Ipv6Addr, MockPeer>`, and seems to be an actually mocked implementation?

What's the intended layering? Is one of the "MockPeers" supposed to be a "RealPeers"? In that case, is the layering supposed to be:
- "struct Peers" contains a "dyn PeersImpl" which might be a mock or real peer implementation?

Or something else? If that's the case, why is a `mock_discover` method being exposed from "struct Peers"? Shouldn't the "top-level" peers be agnostic to the realness / fakenss of the underlying peer impl?

installinator/src/peers.rs
```diff
@@ -0,0 +1,334 @@
...
+            peer,
+            artifact_id,
+            sender,
+            cancel_receiver,
+        );
+
+        let mut artifact_bytes = BufList::new();
+
+        loop {
+            match tokio::time::timeout(self.timeout, receiver.recv()).await {
```
smklein, line 202: Nice, great to see we're coping with timeouts already

installinator/src/mock_peers.rs
```diff
@@ -0,0 +1,504 @@
...
+                            })
+                            .collect();
+                        Self::new(artifact, peers, attempt_bitmaps)
+                    },
+                )
+        }
+    }
+
+    #[derive(Debug, Arbitrary)]
+    enum MockResponse_ {
```
smklein, line 339: Is this like a proptest thing? Why have structs and enums that look nearly identical to existing ones? Couldn't we just have these derives on `MockResponse`, rather than creating a distinct `MockResponse_`?

(This is less of a suggestion, more of a question; I'm not super familiar with proptest)

installinator/src/dispatch.rs
```diff
@@ -0,0 +1,150 @@
...
+        let artifact = FetchedArtifact::loop_fetch_from_peers(
+            &log,
+            || {
+                // TODO: discover nodes via the bootstrap network
+                async {
+                    Peers::mock_discover(&log)
+                        .await
+                        .context("error discovering peers")
+                        .map_err(DiscoverPeersError::Retry)
+                }
```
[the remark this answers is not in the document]
jgallagher: 1. I've been thinking of installinator as doing everything sled-agent (or really bootstrap agent) would do to get to the point of communicating on the bootstrap network, so I'm inclined to say yes, the installinator should launch maghemite (since normally bootstrap-agent [launches it](https://github.com/oxidecomputer/omicron/blob/ac88471f4c20540122b5ced344cede295b8c41bd/sled-agent/src/bootstrap/maghemite.rs#L29)). I'm fuzzy on other requirements to get to the bootstrap network - at a minimum, starting maghemite requires passing it the names of the interfaces it should be using, so those have to exist (and possibly have link-local addresses already?).

2. I was testing locally on a Linux host with multiple [helios-engvm](https://github.com/oxidecomputer/helios-engvm) VMs talking to each other. This required a bit of finagling (although not as much as I expected, I think) with libvirt networking settings to make sure the VMs could see each other. It's been many months since I tried that, but it might still work. After getting the VMs starting up and setting up link-local addresses, I had to add routes out in the linux host to let them see each other. For the two VMs I had running and the virtual bridge device libvirt created, that was:

```
ip -6 route add fdb0:5254:e4:589f::1/64 via 1de::1f7 dev virbr1
ip -6 route add fdb0:5254:8a:8c1c::1/64 via 1de::1e6 dev virbr1
```

This felt fragile at the time and still does. I'm not sure if there's a better way to fake a bootstrap network locally for testing though.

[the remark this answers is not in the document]
andrewjstone: You should be able to use [falcon](https://github.com/oxidecomputer/falcon) for this if you have a helios box. I created an [example two-node toplolgy](https://github.com/oxidecomputer/falcon-omicron) that I used to test trust quourm. Unfortunately (like the falcon instructions)t his is somewhat out of date. Primarily I think this is just because of the images that should be used. Ry or Levon should be able to help you here. I think Falcon is a much more reliable way to do this type of testing personally.

installinator/src/lib.rs
```diff
@@ -0,0 +1,10 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+mod dispatch;
+mod errors;
+mod mock_peers;
```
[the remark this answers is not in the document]
sunshowers: Yeah, my original plan was to have a mock mode, but the more I've written this the more I think it makes sense to just make it test-only.

installinator/src/errors.rs
```diff
@@ -0,0 +1,26 @@
...
+    Timeout { peer: Ipv6Addr, timeout: Duration, bytes_fetched: usize },
+}
+
+#[derive(Debug, Error)]
+pub(crate) enum DiscoverPeersError {
+    #[error("failed to discover peers (will retry)")]
+    Retry(#[source] anyhow::Error),
+
+    #[error("failed to discover peers (no more retries left, will abort)")]
+    #[allow(unused)]
```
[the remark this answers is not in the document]
sunshowers: Ooh thanks for catching this!

installinator/src/dispatch.rs
```diff
@@ -0,0 +1,150 @@
...
+        let drain = slog_async::Async::new(drain).build().fuse();
+        Ok(slog::Logger::root(drain, slog::o!()))
+    }
+}
+
+#[derive(Debug, Subcommand)]
+enum InstallinatorCommand {
+    /// Discover peers on the bootstrap network.
+    Discover(DiscoverOpts),
+    /// Perform the installation.
```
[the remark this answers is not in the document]
sunshowers: Yeah, it's a debugging-only command. What do you think of calling it something like debug-discover?

installinator/src/peers.rs
sunshowers: Thanks -- the comment was outdated; a previous version didn't have the retry/abort logic.

installinator/src/peers.rs
```diff
@@ -0,0 +1,334 @@
...
+        artifact_id: &ArtifactId,
+        sender: mpsc::Sender<Result<Bytes, progenitor_client::Error>>,
+        cancel_receiver: oneshot::Receiver<()>,
+    );
+}
+
+#[derive(Clone, Debug)]
+struct MockPeers {
+    peers: Vec<Ipv6Addr>,
+}
```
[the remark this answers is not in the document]
sunshowers: This one is meant to be deleted once we have all the integration work done. Will rename it to something appropriate.

installinator/src/peers.rs
sunshowers: > "struct Peers" contains a "dyn PeersImpl" which might be a mock or real peer implementation?

Yeah, basically this.

> If that's the case, why is a mock_discover method being exposed from "struct Peers"? Shouldn't the "top-level" peers be agnostic to the realness / fakenss of the underlying peer impl?

I agree, will fix! (This code is all meant to be deleted anyway.)

installinator/src/mock_peers.rs
sunshowers: So the `MockResponse_` data structures are basically identical to `MockResponse`, except they contain a [`proptest::sample::Index`](https://altsysrq.github.io/rustdoc/proptest/latest/proptest/sample/struct.Index.html) to indicate indexes rather than a `usize`.

The alternative would be to use a generic here -- I'm not entirely sure if that works with the derive macros I'm using here, but I can look!

As to why this is required, I'll write up a thing about some of the theory behind proptest at some point. The tl;dr is that we want to try and do our best to not use `prop_flat_map` (monadic bind). There are several ways to work around it and using indexes that get resolved later is one of them.

installinator/src/peers.rs
sunshowers: Fixed. I've also renamed the `MockPeers` in `peers.rs` to `ExamplePeers` -- the plan is for `ExamplePeers` to go away before this PR lands.

installinator/src/peers.rs
sunshowers: Done -- added a 10ms delay.

installinator/src/peers.rs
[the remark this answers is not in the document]
sunshowers: Renamed to `ExamplePeers`.

installinator/src/dispatch.rs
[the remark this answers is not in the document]
sunshowers: Renamed to `debug-discover`.

installinator/src/dispatch.rs
```diff
@@ -0,0 +1,150 @@
...
+    Install(InstallOpts),
+}
+
+#[derive(Debug, Args)]
+#[command(version)]
+struct DiscoverOpts {
+    //
```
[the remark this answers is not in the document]
sunshowers: Fixed, thanks.

installinator/src/lib.rs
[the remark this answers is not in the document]
sunshowers: Done.

installinator/src/errors.rs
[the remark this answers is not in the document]
sunshowers: Ah actually `DiscoverPeersError::Abort` is *used* but is not currently *created* (in non-test code). So this has to be in place for now, but we can remove it after we've hooked things up.

installinator/src/dispatch.rs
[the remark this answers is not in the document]
sunshowers: Thanks everyone. I do have a Helios VM -- I'm going to play around with falcon.

installinator/src/peers.rs
```diff
@@ -0,0 +1,424 @@
...
+    client: wicketd_client::Client,
+}
+
+impl ArtifactClient {
+    fn new(addr: SocketAddrV6, log: &slog::Logger) -> Self {
+        let endpoint = format!("http://[{}]:{}", addr.ip(), addr.port());
+        let log = log.new(
+            slog::o!("component" => "ArtifactClient", "peer" => addr.to_string()),
+        );
+        let client = wicketd_client::Client::new(&endpoint, log.clone());
```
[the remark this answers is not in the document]
smklein: FYI @andrewjstone

installinator/src/peers.rs
```diff
@@ -264,37 +325,70 @@ impl PeersImpl for ExamplePeers {
...
+
+        slog::debug!(
+            &self.log,
+            "preparing to receive {:?} bytes from artifact",
+            artifact_bytes.content_length(),
+        );
+
+        let mut bytes = artifact_bytes.into_inner_stream();
+        while let Some(item) = bytes.next().await {
+            _ = sender.send(item.map_err(Into::into)).await;
```
smklein, line 387: Nitpick, but why do we expect the receiver to remain open for the duration of this function?

- If we don't expect that: Shouldn't we bail on error?
- If we do expect that: Should we assert it?

sunshowers: Good question! There's two separate tasks here so the receiver staying open isn't guaranteed. Guess we should just exit in that case.

sunshowers: (We can also get rid of the cancel oneshot channel in that case, I think! Dropping the receiver indicates that the task should be cancelled.)

sunshowers: Done.

installinator/src/dispatch.rs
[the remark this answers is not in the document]
sunshowers: Managed to get a setup going on atrium with a ton of help from @rcgoodfellow. I've updated the PR with basic discovery that seems to work, will make some notes on the updated version.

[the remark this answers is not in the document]
sunshowers: So I basically got this working by:

**Using physical hardware rather than a VM**, because there appear to be issues with kvm/nested virtualization on Linux. I used atrium for this.

**Using a falcon template** which copies over a ton of scripts from falcon-omicron: https://github.com/sunshowers/falcon-bootstrap

I didn't use falcon-omicron because I couldn't get it to build, so I started from a new template and copied over the config and scripts from falcon-omicron.

**Changing the firstboot.sh script to statically assign IPs**, since DHCP wasn't working for me in the falcon environment on atrium (could be just an IP misconfiguration somewhere):

```bash
case $(hostname) in
        rain_om1)
                v4_addr="192.168.3.101"
                ;;
        rain_om2)
                v4_addr="192.168.3.102"
                ;;
        *)
                echo -n "unknown hostname $(hostname)"
                exit 1
                ;;
esac
```

**Downloading ddmadm and ddmd** from [this CI job](https://github.com/oxidecomputer/maghemite/runs/9492743660) and scping them over.

**Spinning up ddmd** in both VMs with `pfexec ./ddmd 8000 :: vioif0/v6 server`.

**Using ddmadm get-peers** to verify that both ddmd instances could see each other:

```
rain@rain_om2:~$ ./ddmadm get-peers
{
    "3": PeerInfo {
        addr: fe80::8:20ff:fe9e:326e,
        host: Some(
            "rain_om1",
        ),
        kind: Some(
            Server,
        ),
        status: Active,
    },
}
```

**Creating addresses and advertising them**:

* on rain-om1: `ipadm create-addr -T static -a fd00:1::1/64 vioif0/u6 && ddmadm advertise-prefix fd00:1::1/64`
* on rain-om2: `ipadm create-addr -T static -a fd00:2::1/64 vioif0/u6 && ddmadm advertise-prefix fd00:2::1/64`

(There was an issue here with the ddmds not adding these addresses to the kernel routing tables as inspected with `netstat -nr -f inet6`, which was fixed by restarting the ddmd servers.)

**Copying over wicketd to rain-om1 and running it** with `./wicketd run --address '[fd00:1::1]:14000' ~/config.toml`

**Copying over installinator to rain-om2 and running it** with `./installinator install __installinator-test:100000000 test.out`

This copied over a 100MB file from wicketd as expected. (Yay!)

sunshowers:
Notes about testing bootstrap discovery:

**Using physical hardware rather than a VM**, because there appear to be issues with kvm/nested virtualization on Linux. I used atrium for this.

**Using a falcon template** which copies over a ton of scripts from falcon-omicron: https://github.com/sunshowers/falcon-bootstrap

I didn't use falcon-omicron because I couldn't get it to build, so I started from a new template and copied over the config and scripts from falcon-omicron.

**Changing the firstboot.sh script to statically assign IPs**, since DHCP wasn't working for me in the falcon environment on atrium (could be just an IP misconfiguration somewhere):

```bash
case $(hostname) in
        rain_om1)
                v4_addr="192.168.3.101"
                ;;
        rain_om2)
                v4_addr="192.168.3.102"
                ;;
        *)
                echo -n "unknown hostname $(hostname)"
                exit 1
                ;;
esac
```

**Downloading ddmadm and ddmd** from [this CI job](https://github.com/oxidecomputer/maghemite/runs/9492743660) and scping them over.

**Spinning up ddmd** in both VMs with `pfexec ./ddmd 8000 :: vioif0/v6 server`.

**Using ddmadm get-peers** to verify that both ddmd instances could see each other:

```
rain@rain_om2:~$ ./ddmadm get-peers
{
    "3": PeerInfo {
        addr: fe80::8:20ff:fe9e:326e,
        host: Some(
            "rain_om1",
        ),
        kind: Some(
            Server,
        ),
        status: Active,
    },
}
```

**Creating addresses and advertising them**:

* on rain-om1: `ipadm create-addr -T static -a fd00:1::1/64 vioif0/u6 && ddmadm advertise-prefix fd00:1::1/64`
* on rain-om2: `ipadm create-addr -T static -a fd00:2::1/64 vioif0/u6 && ddmadm advertise-prefix fd00:2::1/64`

(There was an issue here with the ddmds not adding these addresses to the kernel routing tables as inspected with `netstat -nr -f inet6`, which was fixed by restarting the ddmd servers.)

**Copying over wicketd to rain-om1 and running it** with `./wicketd run --address '[fd00:1::1]:14000' ~/config.toml`

**Copying over installinator to rain-om2 and running it** with `./installinator install __installinator-test:100000000 test.out`

This copied over a 100MB file from wicketd as expected. (Yay!)

installinator/src/peers.rs
```diff
@@ -0,0 +1,435 @@
...
+pub(crate) enum DiscoveryMechanism {
+    /// The default discovery mechanism: hit the bootstrap network.
+    Bootstrap,
+
+    /// A list of peers is manually specified.
+    List(Vec<SocketAddrV6>),
+}
+
+// TODO: This currently hardcodes this wicketd port, will probably want to sync up on this.
+const WICKETD_PORT: u16 = 14000;
```
sunshowers, line 34: @andrewjstone I've currently hardcoded the wicketd port here.

One thing I'm wondering is if we should have a separate dropshot server for serving artifacts, that both wicketd and sled-agent both spin up. sled-agent doesn't currently have an HTTP server exposed on the bootstrap network at all, so this "artifact server" could be shared by both services. Installinator wouldn't need to know about wicketd vs sled-agent: it could just hit the artifact port.

andrewjstone: @sunshowers I think that makes a lot of sense. Let's do it!

Also, great work on this whole PR and all the stuff you got working today. I'm sorry I haven't had a chance to dig in deeper on this this week. From my cursory looks periodically everything appears to be going swimmingly :)

sunshowers:
OK, since I'm going on vacation today my plan is to land what we have here, and create an issue tracking the remaining work that needs to be done.

sunshowers:
Remaining work is tracked in #2061.

smklein (approved):
Great work pulling together all these pieces together, and getting a well-tested foundation for this component.

There's more work to-do (as you documented in #2061 ), but you should feel free to land this as-is, and we can iterate on top of this codebase!

installinator/src/ddm_admin_client.rs
```diff
@@ -0,0 +1,84 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! This file is mostly copied from sled-agent. It's a TODO to unify them.
```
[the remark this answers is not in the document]
sunshowers: Done, added to #2061!

installinator/src/mock_peers.rs
```diff
@@ -0,0 +1,513 @@
...
+enum ResponseAction_ {
+    #[weight(19)]
+    Response {
+        #[strategy((0..1000u64).prop_map(Duration::from_millis))]
+        after: Duration,
+        count: prop::sample::Index,
+    },
+    #[weight(1)]
+    Error,
+}
```
sunshowers, line 395: @faithanalog this is an example for how to use `proptest::sample::Index`.

Issue #485, opened 2021-12-06
Enable retrieving and modifying a resource by `id`


zephraph:
Currently we have no way of retrieving or acting on a resource if we have its `id`. To remedy that we should implement a suite of endpoints to act on resources given the `id`. Having an `id` is unique in that you don't have to do a path resolution to find the resource (given that `id`s are globally unique to a resource).

As originally suggested by @davepacheco, we've agreed on `/by_id/<resource>/<id>` as the general shape of the API. It makes sense to have GET, PUT, and DELETE actions for each of these resources. 

<details>
<summary>Previous approach</summary>

When calling an API endpoint we should allow `name` or `id` to be used interchangeably. We've had several discussions about this in the past. @davepacheco mentioned using `id:<uuid>` as a pattern for specifying `id`s in the route. 

I'd personally like to add a restriction to names such that a name _can't_ be a valid UUID in order to avoid confusion there. We could go further by creating a wrapping enum `Identifier` type which looks like

```rs
enum Identifier {
  Id(Uuid),
  Name(String)
}
```

which would be type that the api expects as a resource identifier. That would then be converted from the api callsite down into an id to be passed to the rest of the internal calls. 

Regardless of the implementation I do feel strongly we should have _some_ mechanisms for being able to make requests by id especially given that we have no mechanism for tracing resource renames.  

</details>

david-crespo:
The relevant meeting recording is here: [Omicron + Oso + Review (2021-11-22 at 10:02 GMT-8)](https://drive.google.com/file/d/1t_lQyPycGzExShqFVfL4aEyxMU8gUyii/view)

The part where we get into the above questions starts at 1:00:10. The discussion is kind of jumpy but here are some notes I wrote down afterward:

---

While going over [Dave's second authz PR](https://github.com/oxidecomputer/omicron/pull/405), We realized that the `(org id, project id, disk id)` needed to authorize actions on a disk is basically the same as what you produce when just trying to look up a disk by the names. We also realized you need all three IDs even when you’re doing a lookup with a UUID because you need the parent IDs for authz. 

Because you need all these IDs no matter what, you might as well do the lookups that produce those IDs as early as possible in the call chain. That way you can deal with a universal resource identifier thing as much of the time as possible. So, rather than passing around models (e.g., to and from the `nexus.rs` functions), you pass around what I’m thinking we should call a `Resource`, which looks something like

```rust
struct DiskResource {
  disk: db::models::Disk,
[...]

PR #2042, opened 2022-12-09
Update lookup fns introduced in the V1 API to return `lookup::*` over `authz::*`

zephraph:
This PR is mostly in response to @davepacheco's feedback on #1957. Special thanks to @smklein for helping work through some rust lifetime issues here. 

Previously the `instance_lookup` and `project_lookup` functions returned a `authz::Instance` or `authz::Project` respectively that was authorized with read permissions. The challenge here is that this meant that in several places we required duplicate lookups to be able to resolve the instance with the right permissions. To avoid that we're passing around the lookup step itself instead.

### DIFF
--- a/nexus/src/app/instance.rs
+++ b/nexus/src/app/instance.rs
@@ -15,6 +15,7 @@ use crate::cidata::InstanceCiData;
 use crate::context::OpContext;
 use crate::db;
 use crate::db::identity::Resource;
+use crate::db::lookup;
 use crate::db::lookup::LookupPath;
 use crate::db::queries::network_interface;
 use crate::external_api::params;
@@ -36,6 +37,7 @@ use omicron_common::api::external::NameOrId;
 use omicron_common::api::external::UpdateResult;
 use omicron_common::api::external::Vni;
 use omicron_common::api::internal::nexus;
+use ref_cast::RefCast;
 use sled_agent_client::types::InstanceRuntimeStateMigrateParams;
 use sled_agent_client::types::InstanceRuntimeStateRequested;
 use sled_agent_client::types::InstanceStateRequested;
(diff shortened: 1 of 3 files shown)


david-crespo (approved):
Awesome. It feels very clearly like the way it's meant to be done. We've been working our way here for [at least a year](https://github.com/oxidecomputer/omicron/issues/485)!

augustuswm:
Overall this makes sense to me. I don't know enough about Nexus to know if this would work or fit with existing patterns, but following this pattern we are likely to end up with a lot of match cases for project scoped resources that mirror:

```rust
ResourceSelector {
    resource: NameOrId::Name(resource_name),
    project: Some(NameOrId::Name(project_name)),
    organization: Some(NameOrId::Id(organization_id)),
} => {
    // ...
}
ResourceSelector {
    resource: NameOrId::Name(resource_name),
    project: Some(NameOrId::Name(project_name)),
    organization: Some(NameOrId::Name(organization_name)),
} => {
    // ...
}
```

Outside the scope of this PR, but maybe `InstanceSelector` could look like:

```rust
struct InstanceSelector {
    pub instance: NameOrId,
    pub project: Option<ProjectSelector>,
}
```

which would reduce the needed branches per selector. Following this path though would unfortunately start replicating a lot of the logic already encoded in the the `db::lookup` enums.

This shouldn't block this PR, but from my limited reading, it seems like with Selectors we are starting to encode the hierarchy of resources in multiple places.

augustuswm:
Right, I should have been clearer (and used a different name) on that. It would end up as a different struct, such that the chain became `Params -> Selector -> Lookup`

zephraph:
I agree generally with the feedback here. I'm going to merge this PR and do a few more endpoints (definitely those lower in the tree like project and org). After those are all in, I'll start investigating scoping things down for composability.

zephraph:
I agree generally with the feedback here. I'm going to merge this PR and do a few more endpoints (definitely those lower in the tree like project and org). After those are all in, I'll start investigating scoping things down for composability.

Issue #2028, opened 2022-12-08
Update API to match resolution of RFD-322


zephraph:
In [RFD-322](https://rfd.shared.oxide.computer/rfd/0322) we reached consensus to update our API design to be ID first[^1]. This issue is to track the work necessary to complete that transition. 

- [ ] Update RFD-322's determination section for clarity
- [ ] Update `instance` endpoints first as a PoC of the approach #1957 
- [ ] Update `disk` endpoints #2008 
- [ ] Update `organization` endpoints
- [ ] Update `project` endpoints
- [ ] Update `snapshot` and `image` endpoints
- [ ] Update networking endpoints
- [ ] Update `system` endpoints
- [ ] Update remaining endpoints
- [ ] Mark non-v1 endpoints as deprecated (depends on https://github.com/oxidecomputer/dropshot/pull/503)
- [ ] Delete non-V1 endpoints

---

[^1]: I'm working on synthesizing the discussions and decisions into an an actual determination section

zephraph:
Something to follow up on..

In #2347 the `snapshot_create` method in the nexus layer requires that the `project_lookup` it uses to be passed by value instead of reference. That's because when calling a function hanging off of a lookup (like `disk_name` in this case) it actually _consumes_ the lookup. Meaning you can't do multiple things on a lookup. Given that that's the case, I'm not sure it makes sense to pass lookups by reference. You almost want to guarantee that the lookup is being moved to whatever Nexus function is calling it. 

This is more of a logical correctness thing than anything. I would like to have the API calls being mostly consistent though.

karencfv:
Do we expect people to be able to change their usernames? If not maybe we should keep `GET /system/user/{user_name}` with name instead of ID? 🤔

david-crespo:
@karencfv I think that's right. Those are built-in users whose names do not change.

https://github.com/oxidecomputer/omicron/blob/73941804cc3d0be57f481cc25188c2faefc24b46/common/src/sql/dbinit.sql#L1711-L1717

PR #2050, opened 2022-12-13
RFD-322: Add `/v1/` endpoints for organizations and projects

zephraph:
WIP

### DIFF
--- a/nexus/db-model/src/name.rs
+++ b/nexus/db-model/src/name.rs
@@ -37,6 +37,8 @@ use serde::{Deserialize, Serialize};
 #[display("{0}")]
 pub struct Name(pub external::Name);
 
+// impl Into<external::Name> for Name {}
+
 NewtypeFrom! { () pub struct Name(external::Name); }
 NewtypeDeref! { () pub struct Name(external::Name); }
 
--- a/nexus/src/app/instance.rs
+++ b/nexus/src/app/instance.rs
@@ -62,56 +62,26 @@ impl super::Nexus {
         instance_selector: &'a params::InstanceSelector,
     ) -> LookupResult<lookup::Instance<'a>> {
         match instance_selector {
-            params::InstanceSelector { instance: NameOrId::Id(id), .. } => {
-                // TODO: 400 if project or organization are present
+            params::InstanceSelector(NameOrId::Id(id), ..) => {
                 let instance =
                     LookupPath::new(opctx, &self.db_datastore).instance_id(*id);
                 Ok(instance)
             }
-            params::InstanceSelector {
-                instance: NameOrId::Name(instance_name),
-                project: Some(NameOrId::Id(project_id)),
-                ..
-            } => {
-                // TODO: 400 if organization is present
-                let instance = LookupPath::new(opctx, &self.db_datastore)
-                    .project_id(*project_id)
-                    .instance_name(Name::ref_cast(instance_name));
-                Ok(instance)
+            params::InstanceSelector(
+                NameOrId::Name(name),
+                project_selector,
+            ) => {
+                if let Some(project) = project_selector {
+                    let instance = self
+                        .project_lookup(opctx, project)?
+                        .instance_name(Name::ref_cast(name));
+                    Ok(instance)
+                } else {
+                    Err(Error::InvalidRequest {
+                        message: "Unable to resolve instance by name without instance".to_string(),
+                    })
+                }
             }
-            params::InstanceSelector {
-                instance: NameOrId::Name(instance_name),
-                project: Some(NameOrId::Name(project_name)),
-                organization: Some(NameOrId::Id(organization_id)),
-            } => {
-                let instance = LookupPath::new(opctx, &self.db_datastore)
-                    .organization_id(*organization_id)
-                    .project_name(Name::ref_cast(project_name))
-                    .instance_name(Name::ref_cast(instance_name));
-                Ok(instance)
-            }
-            params::InstanceSelector {
-                instance: NameOrId::Name(instance_name),
-                project: Some(NameOrId::Name(project_name)),
-                organization: Some(NameOrId::Name(organization_name)),
-            } => {
-                let instance = LookupPath::new(opctx, &self.db_datastore)
-                    .organization_name(Name::ref_cast(organization_name))
-                    .project_name(Name::ref_cast(project_name))
-                    .instance_name(Name::ref_cast(instance_name));
-                Ok(instance)
-            }
-            // TODO: Add a better error message
-            _ => Err(Error::InvalidRequest {
-                message: "
-                Unable to resolve instance. Expected one of
-                    - instance: Uuid 
-                    - instance: Name, project: Uuid
-                    - instance: Name, project: Name, organization: Uuid
-                    - instance: Name, project: Name, organization: Name
-                "
-                .to_string(),
-            }),
         }
     }
 
--- a/nexus/src/app/organization.rs
+++ b/nexus/src/app/organization.rs
@@ -7,6 +7,7 @@
 use crate::authz;
 use crate::context::OpContext;
 use crate::db;
+use crate::db::lookup;
 use crate::db::lookup::LookupPath;
 use crate::db::model::Name;
 use crate::external_api::params;
@@ -18,40 +19,36 @@ use omicron_common::api::external::DeleteResult;
 use omicron_common::api::external::Error;
 use omicron_common::api::external::ListResultVec;
 use omicron_common::api::external::LookupResult;
+use omicron_common::api::external::NameOrId;
 use omicron_common::api::external::UpdateResult;
+use ref_cast::RefCast;
 use uuid::Uuid;
 
 impl super::Nexus {
+    pub fn organization_lookup<'a>(
+        &'a self,
+        opctx: &'a OpContext,
+        organization_selector: &'a params::OrganizationSelector,
+    ) -> LookupResult<lookup::Organization<'a>> {
+        match organization_selector {
+            params::OrganizationSelector(NameOrId::Id(id)) => {
+                let organization = LookupPath::new(opctx, &self.db_datastore)
+                    .organization_id(*id);
+                Ok(organization)
+            }
+            params::OrganizationSelector(NameOrId::Name(name)) => {
+                let organization = LookupPath::new(opctx, &self.db_datastore)
+                    .organization_name(Name::ref_cast(name));
+                Ok(organization)
+            }
+        }
+    }
     pub async fn organization_create(
         &self,
         opctx: &OpContext,
         new_organization: &params::OrganizationCreate,
     ) -> CreateResult<db::model::Organization> {
         self.db_datastore.organization_create(opctx, new_organization).await
-    }
-
-    pub async fn organization_fetch(
-        &self,
-        opctx: &OpContext,
-        organization_name: &Name,
-    ) -> LookupResult<db::model::Organization> {
-        let (.., db_organization) = LookupPath::new(opctx, &self.db_datastore)
-            .organization_name(organization_name)
-            .fetch()
-            .await?;
-        Ok(db_organization)
-    }
-
-    pub async fn organization_fetch_by_id(
-        &self,
-        opctx: &OpContext,
-        organization_id: &Uuid,
-    ) -> LookupResult<db::model::Organization> {
-        let (.., db_organization) = LookupPath::new(opctx, &self.db_datastore)
-            .organization_id(*organization_id)
-            .fetch()
-            .await?;
-        Ok(db_organization)
     }
 
     pub async fn organizations_list_by_name(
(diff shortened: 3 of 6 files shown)


nexus/types/src/external_api/params.rs
```diff
@@ -511,7 +590,7 @@ pub struct NetworkInterfaceUpdate {
 // Type used to identify a Project in request bodies, where one may not have
 // the path in the request URL.
 #[derive(Debug, Clone, Deserialize, Serialize, JsonSchema)]
-pub struct ProjectPath {
+pub struct OldProjectPath {
```
zephraph, line 593: An unfortunate naming conflict, but this'll go away completely once we do #2055

nexus/tests/output/uncovered-authz-endpoints.txt
```diff
@@ -1,22 +1,36 @@
...
+organization_list                        (get    "/organizations")
+organization_view                        (get    "/organizations/{organization_name}")
+organization_policy_view                 (get    "/organizations/{organization_name}/policy")
+project_list                             (get    "/organizations/{organization_name}/projects")
+project_view                             (get    "/organizations/{organization_name}/projects/{project_name}")
+instance_list                            (get    "/organizations/{organization_name}/projects/{project_name}/instances")
+instance_view                            (get    "/organizations/{organization_name}/projects/{project_name}/instances/{instance_name}")
+instance_serial_console                  (get    "/organizations/{organization_name}/projects/{project_name}/instances/{instance_name}/serial-console")
+instance_serial_console_stream           (get    "/organizations/{organization_name}/projects/{project_name}/instances/{instance_name}/serial-console/stream")
+project_policy_view                      (get    "/organizations/{organization_name}/projects/{project_name}/policy")
```
zephraph, line 18: There's work I could do here to get this test passing for all the old endpoints but I'm opting _not_ to do that in favor of making process. The authz test is non-trivial and I'm trying to update it as little as possible until I can get a critical mass of APIs converted. I _am_ going through and ensuring the new API endpoints are correctly tested though. 

On that note, there is a slight issue in that you can only test and endpoint from the spec once. If you try to verify it more than once it freaks out on you. I don't have the exact error handy but it does this because it "consumes" routes from the spec and if it encounters something its already consumed then that's considered a failure. We actually need to be able to test a route at least twice (name path, id). Anyway, I'll add that to the top level issue to resolve.

common/src/api/external/mod.rs
```diff
@@ -288,6 +288,18 @@ impl TryFrom<String> for NameOrId {
...
+    fn from(name: Name) -> Self {
+        NameOrId::Name(name)
+    }
+}
+
+impl From<Uuid> for NameOrId {
+    fn from(id: Uuid) -> Self {
+        NameOrId::Id(id)
+    }
+}
```
zephraph, line 301: These greatly reduce a lot of conversion noise. We might not need them post v1 migration.

nexus/db-model/src/name.rs
```diff
@@ -37,6 +37,12 @@ use serde::{Deserialize, Serialize};
 #[display("{0}")]
 pub struct Name(pub external::Name);
 
+impl From<Name> for external::NameOrId {
+    fn from(name: Name) -> Self {
+        Self::Name(name.0)
+    }
+}
```
zephraph, line 44: Same as above

nexus/src/app/organization.rs
```diff
@@ -30,30 +53,6 @@ impl super::Nexus {
...
-        &self,
-        opctx: &OpContext,
-        organization_id: &Uuid,
-    ) -> LookupResult<db::model::Organization> {
-        let (.., db_organization) = LookupPath::new(opctx, &self.db_datastore)
-            .organization_id(*organization_id)
-            .fetch()
-            .await?;
-        Ok(db_organization)
-    }
```
zephraph, line 55: @davepacheco you mentioned in #1957 that we could probably just ditch the fetch endpoints now that we're passing around lookups. I've been doing that.

nexus/src/app/instance.rs
```diff
@@ -62,56 +62,28 @@ impl super::Nexus {
         instance_selector: &'a params::InstanceSelector,
     ) -> LookupResult<lookup::Instance<'a>> {
         match instance_selector {
```
zephraph, line 64: You'll notice this selector logic is greatly reduced. It checks the name and ID case of the resource being selected against and delegates selection of sub resources to that resource's lookup.

nexus/types/src/external_api/params.rs
```diff
@@ -18,34 +19,109 @@ use serde::{
...
+    fn from(name: Name) -> Self {
+        OrganizationSelector { organization: name.into() }
+    }
+}
+
+#[derive(Deserialize, JsonSchema)]
+pub struct OptionalOrganizationSelector {
+    #[serde(flatten)]
+    pub organization_selector: Option<OrganizationSelector>,
+}
```
zephraph, line 52: The only reason this optional selector has to exist is because dropshot chokes on `Query<Option<OrganizationalSelector>>`. If that worked we could delete this.

zephraph: The types actually work out fine, but the `EXPECTORATE` update fails. 

<details>
<summary>Example error</summary>

```
thread 'integration_tests::commands::test_nexus_openapi' panicked at 'expected normal process exit with code 0, got Signaled(6)

process stderr:thread 'main' panicked at 'invalid type Object(
    SchemaObject {
        metadata: Some(
            Metadata {
                id: None,
                title: Some(
                    "Nullable_ProjectSelector",
                ),
                description: None,
                default: None,
                deprecated: false,
                read_only: false,
                write_only: false,
                examples: [],
            },
        ),
        instance_type: None,
        format: None,
        enum_values: None,
        const_value: None,
        subschemas: Some(
            SubschemaValidation {
                all_of: Some(
                    [
                        Object(
                            SchemaObject {
                                metadata: None,
                                instance_type: None,
                                format: None,
                                enum_values: None,
                                const_value: None,
                                subschemas: None,
                                number: None,
                                string: None,
                                array: None,
                                object: None,
                                reference: Some(
                                    "#/components/schemas/ProjectSelector",
                                ),
                                extensions: {},
                            },
                        ),
                    ],
                ),
                any_of: None,
                one_of: None,
                not: None,
                if_schema: None,
                then_schema: None,
                else_schema: None,
            },
        ),
        number: None,
        string: None,
        array: None,
        object: None,
        reference: None,
        extensions: {
            "nullable": Bool(true),
        },
    },
)', /Users/just-be/.cargo/git/checkouts/dropshot-a4a923d29dccc492/2529126/dropshot/src/handler.rs:932:20
', nexus/tests/integration_tests/commands.rs:106:38
```

</details>

nexus/types/src/external_api/params.rs
```diff
@@ -18,34 +19,109 @@ use serde::{
 use std::{net::IpAddr, str::FromStr};
 use uuid::Uuid;
 
+#[derive(Deserialize, JsonSchema)]
+pub struct OrganizationPath {
+    pub organization: NameOrId,
+}
```
zephraph, line 25: Important thing to note is that I've relocated _all_ of the `/v1/` params here.

karencfv (commented):
I'll leave the review of the internals for people who are more acquainted with the code base, but I played around with the progenitor generated client and it works beautifully!!!

Thanks for taking this on, the API is looking great! 🙇‍♀️

nexus/types/src/external_api/params.rs
augustuswm: I haven't tried converting all of it but you might be able to represent this as:

```rust
pub struct OrganizationSelector {
    pub organization: Option<NameOrId>,
}

pub struct ProjectSelector {
    #[serde(flatten)]
    pub organization_selector: Option<OrganizationSelector>,
    pub project: Option<NameOrId>,
}
```

It would mean that the error message in `_ => Err(Error::invalid_request(...)` would need updating.

nexus/src/app/instance.rs
augustuswm: That looks a lot cleaner. The error cases will likely need to be expanded depending on how that gets spec'd. Specifically the over-specification case will want a different error message.

For example:
```rust
InstanceSelector { project_selector: Some(project_selector), instance: NameOrId::Id(id) }
```

zephraph: Right, yeah. @david-crespo  mentioned in the RFD how he thought that case shouldn't be an error and I'm still undecided. Should we just ignore that case?

nexus/types/src/external_api/params.rs
zephraph: That's true. I made it more explicit because there are cases when we know the user must provide a path param (like the parent resource of a list endpoint) vs when it could be represented by the path. I think the extra code (which can be generated) is worth the trade off of a more correct spec.

nexus/src/app/instance.rs
zephraph: Fair point. I think disallowing invalid combinations is the lowest common denominator.

zephraph: I've updated this in https://github.com/oxidecomputer/omicron/pull/2050/commits/69ac00809a03f69a323a74986f9e8deb3af12abe to have expanded error checks. Essentially I just added an extra case for instance and project.

nexus/types/src/external_api/params.rs
```diff
@@ -18,34 +19,109 @@ use serde::{
...
+    }
+}
+
+#[derive(Deserialize, JsonSchema)]
+pub struct ProjectList {
+    #[serde(flatten)]
+    pub pagination: PaginatedByNameOrId,
+    #[serde(flatten)]
+    pub organization: OrganizationSelector,
+}
```
david-crespo, line 78: I'd almost rather define this next to the endpoint handler since that's the only place it's called.

david-crespo (approved):
Looks great. A bit noisy, but the noise is _extremely_ straightforward so it's not a big deal. We'll have to see whether macros or openapi-lint is the better approach to avoiding mistakes.

nexus/types/src/external_api/params.rs
zephraph: Yeah, I just grouped them together because I'm hoping to macroize it at some point.

zephraph:
All right folks, I'm going to go ahead and merge. Your reviews are still welcome/greatly desired. I'll address any issues in a followup.

nexus/tests/integration_tests/unauthorized_coverage.rs
```diff
@@ -135,23 +135,16 @@ fn test_unauthorized_coverage() {
...
+    // let expected_uncovered_endpoints =
+    //     std::fs::read_to_string("tests/output/uncovered-authz-endpoints.txt")
+    //         .expect("failed to load file of allowed uncovered endpoints");
+
+    // TODO: Update this to remove overwrite capabilities
+    // See https://github.com/oxidecomputer/expectorate/pull/12
+    assert_contents(
+        "tests/output/uncovered-authz-endpoints.txt",
+        uncovered_endpoints.as_str(),
+    );
```
ahl, line 147: what's the plan?

nexus/types/src/external_api/params.rs
```diff
@@ -18,34 +19,109 @@ use serde::{
...
+impl From<Name> for OrganizationSelector {
+    fn from(name: Name) -> Self {
+        OrganizationSelector { organization: name.into() }
+    }
+}
+
+#[derive(Deserialize, JsonSchema)]
+pub struct OptionalOrganizationSelector {
+    #[serde(flatten)]
+    pub organization_selector: Option<OrganizationSelector>,
```
ahl, line 51: why not
`pub organization: Option<NameOrId>,`

nexus/types/src/external_api/params.rs
```diff
@@ -18,34 +19,109 @@ use serde::{
...
+}
+
 #[derive(Deserialize, JsonSchema)]
 pub struct ProjectSelector {
+    #[serde(flatten)]
+    pub organization_selector: Option<OrganizationSelector>,
```
ahl, line 57: why not just `organization: Option<NameOrId>`

nexus/types/src/external_api/params.rs
ahl: @david-crespo I thought you were the one who split these out by type rather than by use?

ahl (commented):
adding these as new endpoints rather than modifying existing ones make errors much harder to spot. When are we going to get rid of the non-v1 operations? presumably we'll be changing the operation names to remove `v1`?

zephraph:
> adding these as new endpoints rather than modifying existing ones make errors much harder to spot. When are we going to get rid of the non-v1 operations? 

That's true, but it's generally also less disruptive. My goal is to not block folks working on other things while the transition is in progress. The loose order of operations is 

1. Add the new endpoints
2. Update the docs and tests
3. Update the clients
4. Remove the old endpoints

You can see (or add) more specifics in the parent issue: https://github.com/oxidecomputer/omicron/issues/2028

> presumably we'll be changing the operation names to remove `v1`?

Tentatively, yes. That seems to be the popular opinion. I think that that because we can have `/v1/` and `/v2/` of an endpoint then we should likely have that mirrored via the operation names. Even though it's more noisy, it's externally consistent. I'm flexible though.

nexus/types/src/external_api/params.rs
zephraph: Consistency mostly. I'll flatten it down.

Issue #2055, opened 2022-12-14
Updates to IP Pools


smklein:
We had a sync on 12/14/22 on this: https://drive.google.com/file/d/1ffOrgbYsoomSaNeo284FX1EmUpWMJqFR/view?usp=share_link

- [x] Remove "project ID" field: https://github.com/oxidecomputer/omicron/pull/2056
- [x] Remove "rack ID" field: https://github.com/oxidecomputer/omicron/pull/2056
- [x] (Later) add bit to distinguish between internal / external visibility?: https://github.com/oxidecomputer/omicron/pull/2056
- [x] Ensure Nexus can point to the one with "internal visibility" only: https://github.com/oxidecomputer/omicron/pull/2056
- [ ] (Later) add field for AZ?

smklein:
Summary from @rmustacc :

- External IPs are best thought of like a hardware resource due to their finite nature and is much more like DRAM than say VMs or volumes in nature.
- Fundamentally an IP pool represents a disjoint group of IPs with the same connectivity properties. For example, all just Internet facing public IPs or say a private VLAN.
- Customers eventually want to be able to constrain groups of projects to specific IP pools. While this isn't part of the MVP, the reality is that it's more than one.
- Internal services want to get their configuration for how to find addresses initially via IP Pools (e.g. nexus, external DNS, NTP, routing though that's a bit more nuanced).
- The system level IP Pools API is going to be simplified right now to mostly represent a fleet (though kind of AZ) resource and the only real acl style bit is whether it's internal or external.
- Eventually we'd like a project-specific API to list the IP pools that are accessible in that context separate from the the system level API to make it obvious what is and isn't usable. This may be worth it for MVP.
- Longer term (e.g. not for the MVP) we want to do better with defaults at a project level or similar and figure out what more fine-grained ACLs might look like on IP Pools, but that's not today.

david-crespo:
Closing in favor of issues I'm creating for the followup work to #3985

Issue #1276, opened 2022-06-24
[nexus] Managing local rack -> managing local fleet


smklein:
Some operations within Nexus are implemented as "manage the state which may exist within my local rack". This includes:

- Awaiting handoff from RSS
- Ensuring a rack-wide CRDB instance exists
- Ensuring sufficient redundancy for services exists with a rack

However, longer-term, we would ideally migrate many of these operations to be "fleet-wide" instead of "rack-wide". This way, one nexus could control multiple racks simultaneously, ensure that CRDB nodes are distributed within an AZ, and ensure that service redundancy suffices for multi-rack failure scenarios.

For additional context, see: https://github.com/oxidecomputer/omicron/pull/1234/files/28d87f51ab88cce3d8ff2560a8996904e8c78f81#diff-5a93a4691987ea1b28d848375a2728abcb26cec85d477d051243cb1863198392

andrewjstone:
Subsumed by more specific issues.

PR #2056, opened 2022-12-14
[nexus] Remove project_id, rack_id from IP pools

smklein:
## Before this PR

- IP Pools could exist in at most one project. IP allocation during instance creation occurred by [either by requesting an IP pool belonging to a project, or by "just looking for any unreserved IP Pool"](https://github.com/oxidecomputer/omicron/blob/79765a4e3b39a29bc9940c0e4a49c4364fbcc9e3/nexus/src/db/queries/external_ip.rs#L186-L212). As discussed in https://github.com/oxidecomputer/omicron/issues/2055 , our intention is for IP pools to be used across multiple projects, and for projects to be able to use multiple IP pools.
- "Service" IP pools were indexed by rack ID, though (as documented in https://github.com/oxidecomputer/omicron/issues/1276 ), they should probably be accessed by AZ instead.

## This PR

- Adds a default IP pool named `default`, which is used for address allocation unless a more specific IP pool is provided
- Removes "project ID" from IP pools (and external IP addresses)
- Removes "rack ID" from IP pool API and DB representation

## In the future

- This PR doesn't provide the many-to-many connection between projects and IP pools that we eventually want, where projects can be configured to use different IP pools for different purposes. However, by removing the not-quite-accurate relationship that an IP pool must belong to a *single* project, the API moves closer towards this direction.
- We probably should access the `service_ip_pool` API with the AZ UUID used for the query, but since AZs don't exist in the API yet, this has been omitted.

Part of #2055

### DIFF
--- a/common/src/sql/dbinit.sql
+++ b/common/src/sql/dbinit.sql
@@ -1109,16 +1109,7 @@ CREATE TABLE omicron.public.ip_pool (
     time_modified TIMESTAMPTZ NOT NULL,
     time_deleted TIMESTAMPTZ,
 
-    /* Optional ID of the project for which this pool is reserved. */
-    project_id UUID,
-
-    /*
-     * Optional rack ID, indicating this is a reserved pool for internal
-     * services on a specific rack.
-     * TODO(https://github.com/oxidecomputer/omicron/issues/1276): This
-     * should probably point to an AZ or fleet, not a rack.
-     */
-    rack_id UUID,
+    internal_only BOOL NOT NULL,
 
     /* The collection's child-resource generation number */
     rcgen INT8 NOT NULL
@@ -1130,15 +1121,6 @@ CREATE TABLE omicron.public.ip_pool (
 CREATE UNIQUE INDEX ON omicron.public.ip_pool (
     name
 ) WHERE
-    time_deleted IS NULL;
-
-/*
- * Index ensuring uniqueness of IP pools by rack ID
- */
-CREATE UNIQUE INDEX ON omicron.public.ip_pool (
-    rack_id
-) WHERE
-    rack_id IS NOT NULL AND
     time_deleted IS NULL;
 
 /*
@@ -1155,20 +1137,11 @@ CREATE TABLE omicron.public.ip_pool_range (
     /* The range is inclusive of the last address. */
     last_address INET NOT NULL,
     ip_pool_id UUID NOT NULL,
-    /* Optional ID of the project for which this range is reserved.
-     *
-     * NOTE: This denormalizes the tables a bit, since the project_id is
-     * duplicated here and in the parent `ip_pool` table. We're allowing this
-     * for now, since it reduces the complexity of the already-bad IP allocation
-     * query, but we may want to revisit that, and JOIN with the parent table
-     * instead.
-     */
-    project_id UUID,
     /* Tracks child resources, IP addresses allocated out of this range. */
     rcgen INT8 NOT NULL
 );
 
-/* 
+/*
  * These help Nexus enforce that the ranges within an IP Pool do not overlap
  * with any other ranges. See `nexus/src/db/queries/ip_pool.rs` for the actual
  * query which does that.
@@ -1183,14 +1156,6 @@ CREATE UNIQUE INDEX ON omicron.public.ip_pool_range (
 )
 STORING (first_address)
 WHERE time_deleted IS NULL;
-
-/*
- * Index supporting allocation of IPs out of a Pool reserved for a project.
- */
-CREATE INDEX ON omicron.public.ip_pool_range (
-    project_id
-) WHERE
-    time_deleted IS NULL;
 
 
 /* The kind of external IP address. */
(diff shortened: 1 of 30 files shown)


nexus/types/src/external_api/params.rs
```diff
@@ -508,23 +508,13 @@ pub struct NetworkInterfaceUpdate {
...
 // IP POOLS
 
-// Type used to identify a Project in request bodies, where one may not have
-// the path in the request URL.
-#[derive(Debug, Clone, Deserialize, Serialize, JsonSchema)]
-pub struct ProjectPath {
-    pub organization: Name,
-    pub project: Name,
-}
-
```
zephraph, line 518: Thanks! I'll be able to remove this hacky thing I did today: https://github.com/oxidecomputer/omicron/pull/2050/files#r1048986084

smklein:
I suspect we'll want a follow-up to this PR to add APIs for the following:

- Given a project, use ACLs to describe which IP pools are *visible* to a user
- Given a project, configure a set of IP pools that it should use for address allocation

As mentioned on https://github.com/oxidecomputer/omicron/issues/2055

nexus/db-model/src/ip_pool.rs
```diff
@@ -28,13 +28,7 @@ pub struct IpPool {
...
     pub identity: IpPoolIdentity,
 
-    /// An optional ID of the project for which this pool is reserved.
-    pub project_id: Option<Uuid>,
-
-    /// An optional ID of the rack for which this pool is reserved.
-    // TODO(https://github.com/oxidecomputer/omicron/issues/1276): This
-    // should probably point to an AZ or fleet, not a rack.
-    pub rack_id: Option<Uuid>,
+    pub internal_only: bool,
```
david-crespo, line 31: Would be nice to have a comment explaining what this means.

smklein: Done: 1ea2005001863d89da8d37316314f0109b196caa

common/src/sql/dbinit.sql
```diff
@@ -1109,16 +1109,7 @@ CREATE TABLE omicron.public.ip_pool (
...
-    project_id UUID,
-
-    /*
-     * Optional rack ID, indicating this is a reserved pool for internal
-     * services on a specific rack.
-     * TODO(https://github.com/oxidecomputer/omicron/issues/1276): This
-     * should probably point to an AZ or fleet, not a rack.
-     */
-    rack_id UUID,
+    internal_only BOOL NOT NULL,
```
zephraph, line 1112: Minor bikeshedding here, no action required unless you feel like it. 

This could just be called `internal` as _only_ seems to me to be implied. 

It'd probably be good to copy the comment from the datastore struct or just note to check that file. Do `@see` comments work in rust? Maybe that's just a jsdoc thing. Again, not required. Just a nice to have.

nexus/src/app/ip_pool.rs
```diff
@@ -28,16 +28,19 @@ impl super::Nexus {
         opctx: &OpContext,
         new_pool: &params::IpPoolCreate,
     ) -> CreateResult<db::model::IpPool> {
-        self.db_datastore.ip_pool_create(opctx, new_pool, None).await
+        self.db_datastore
+            .ip_pool_create(opctx, new_pool, /* internal_only= */ false)
+            .await
     }
 
     pub async fn ip_pool_services_create(
```
zephraph, line 36: It's outside the scope of this PR, but I wonder if _IP Pool Services_ is the right way to frame this now. Should it just be internal ip pools?

nexus/src/app/sagas/instance_create.rs
```diff
@@ -626,13 +626,15 @@ async fn sic_allocate_instance_snat_ip(
         OpContext::for_saga_action(&sagactx, &saga_params.serialized_authn);
     let instance_id = sagactx.lookup::<Uuid>("instance_id")?;
     let ip_id = sagactx.lookup::<Uuid>("snat_ip_id")?;
+
+    let (.., pool) = datastore
+        .ip_pools_lookup_by_name_no_auth(&opctx, "default")
```
zephraph, line 631: Is this really just for looking up the `default` IP pool? If so, I wonder if we shouldn't just have a `ip_pools_lookup_default` that's documented to not require authz but otherwise doesn't take arbitrary names.

smklein: I did this, and added authz checks, within 91415a4571c916c88c9432fc5ad7620722191afe

nexus/src/app/ip_pool.rs
smklein: I hear you on this; maybe `s/service/internal` would be more consistent. When I originally wrote this, I think I was writing "service" as shorthand for "internal service".

common/src/sql/dbinit.sql
smklein: Done in b4dd0b1d7d16b92a54c9bb872401712334281d5d

I dunno about the linked comments, but I included a summary here.

zephraph (approved):
This looks great! Thanks for tackling it @smklein.

zephraph:
See also the meeting with @rmustacc that shaped these changes: https://drive.google.com/file/d/1ffOrgbYsoomSaNeo284FX1EmUpWMJqFR/view?usp=sharing

PR #2057, opened 2022-12-16
RFD-322 Networking

zephraph:
See #2028 

This PR adds V1 versions of the developer facing networking APIs. Firewall rules aren't in this PR. I'll do that in a follow up.

## New Endpoints

All endpoints here are scoped via query parameters by their parent selectors. So for `vpc-subnets` if you provide the vpc by name, you'll also need to provide the project. If the projected is specified by name you'll need the org. 

So `/v1/vpc-subnet?vpc={vpc_id}`, `/v1/vpc-subnet?vpc={vpc_name}&project={project_id}`, etc

```shell
# Scoped via  ?project={project}
GET    /v1/vpcs       # List VPCs
POST   /v1/vpcs       # Create a VPC
GET    /v1/vpcs/{vpc} # Get a VPC
DELETE /v1/vpcs/{vpc} # Delete a VPC

# Scoped via  ?vpc={vpc}
GET    /v1/vpc-subnets          # List VPC subnets
POST   /v1/vpc-subnets          # Create a VPC subnet
GET    /v1/vpc-subnets/{subnet} # Get a VPC subnet
DELETE /v1/vpc-subnets/{subnet} # Delete a VPC subnet

# Scoped via  ?vpc={vpc}
GET    /v1/vpc-routers          # List VPCs
POST   /v1/vpc-routers          # Create a VPC
GET    /v1/vpc-routers/{router} # Get a VPC
DELETE /v1/vpc-routers/{router} # Delete a VPC

# Scoped via  ?router={router}
GET    /v1/vpc-router-routes         # List VPC router routes
POST   /v1/vpc-router-routes         # Create a VPC router route
GET    /v1/vpc-router-routes/{route} # Get a VPC router route
DELETE /v1/vpc-router-routes/{route} # Delete a VPC router route

# Scoped via ?instance={instance}
GET    /v1/network-interfaces             # List network interfaces
POST   /v1/network-interfaces             # Create network interface
GET    /v1/network-interfaces/{interface} # Get a network interface
DELETE /v1/network-interfaces/{interface} # Delete network interface
```

For a quick overview of the entire API (and the changes introduced in this PR) check out the [nexus_tags.txt file](https://github.com/oxidecomputer/omicron/pull/2057/files#diff-482331ae638fae81e451868307fcd2f8836600a2ac552c965da987a402d94e54)

## Notable changes

- I pulled app level network interface functions out into their [own file](https://github.com/oxidecomputer/omicron/pull/2057/files#diff-4d675c5f8ef51a7a12255d0a6cd593b4bfb7b77dbf5e791c97abf9e874a588bf). 
- I've slowly started to update the naming convention of app and datastore level functions to be less nested (and closer to the operation names) as per [this comment](https://github.com/oxidecomputer/omicron/pull/2008#discussion_r1054725623). It's a little bit inconsistent now but the intent is to do a larger follow up to clean these all up. At present I'm just updating them as I come across them. 
- I've begun moving all possible contents of an http entrypoint into its handler for more timing information as per [this comment](https://github.com/oxidecomputer/omicron/pull/2206#discussion_r1085612867)


common/src/api/external/mod.rs
```diff
@@ -1557,24 +1557,6 @@ pub struct RouterRoute {
...
-
-/// Updateable properties of a [`RouterRoute`]
-#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema)]
-pub struct RouterRouteUpdateParams {
-    #[serde(flatten)]
-    pub identity: IdentityMetadataUpdateParams,
-    pub target: RouteTarget,
-    pub destination: RouteDestination,
-}
-
```
zephraph, line 1577: I moved this over to the params file to be co-located with the other params.

nexus/types/src/external_api/params.rs
```diff
@@ -35,6 +36,31 @@ pub struct InstancePath {
...
+#[derive(Deserialize, JsonSchema)]
+pub struct RouterPath {
+    pub router: NameOrId,
+}
+
+#[derive(Deserialize, JsonSchema)]
+pub struct RoutePath {
+    pub route: NameOrId,
+}
+
```
david-crespo, line 63: Not something to be addressed in this PR, but it's a little silly that these are all the same. If we called it `resource` in all cases maybe they could share the struct. Something like:

```rs
pub struct NameOrIdPath {
    pub resource: NameOrId,
}
```

david-crespo (approved):
Whew, this is a BIG one

nexus/types/src/external_api/params.rs
zephraph: Yeah, the ultimate intent here I think is to have it be generated by a macro. Unfortunately the name matters as it shows up in documentation, the API spec, etc.

nexus/src/app/instance.rs
```diff
@@ -799,220 +797,6 @@ impl super::Nexus {
         Ok(disk)
     }
 
-    /// Create a network interface attached to the provided instance.
```
zephraph, line 802: I relocated the network interface logic in its own file. We break this resource at the data store layer, might as well be consistent with it.

nexus/src/app/vpc_router.rs
```diff
@@ -162,24 +138,50 @@ impl super::Nexus {
...
+                let route = self
+                    .vpc_router_lookup(opctx, selector)?
+                    .router_route_name(Name::ref_cast(name));
+                Ok(route)
+            }
+            params::RouteSelector {
+                route: NameOrId::Id(_),
+                router_selector: Some(_),
+            } => Err(Error::invalid_request(
+                "when providing subnet as an ID, vpc should not be specified",
```
bnaecker, line 168: This looks like a copy-paste from the updates to the VPC Subnet?

nexus/src/app/vpc_subnet.rs
```diff
@@ -8,39 +8,70 @@ use crate::authz;
...
+                let subnet = self
+                    .vpc_lookup(opctx, selector)?
+                    .vpc_subnet_name(Name::ref_cast(name));
+                Ok(subnet)
+            }
+            params::SubnetSelector {
+                subnet: NameOrId::Id(_),
+                vpc_selector: Some(_),
+            } => Err(Error::invalid_request(
+                "when providing subnet as an ID, vpc should not be specified",
```
bnaecker, line 59: Same nit here, maybe this should be "vpc_selector".

nexus/src/app/vpc_subnet.rs
```diff
@@ -8,39 +8,70 @@ use crate::authz;
...
+            )),
+            _ => Err(Error::invalid_request(
+                "subnet should either be an ID or vpc should be specified",
```
bnaecker, line 62: Same nit here, maybe this should be "vpc_selector".

nexus/src/app/vpc_subnet.rs
```diff
@@ -184,69 +215,25 @@ impl super::Nexus {
         }
     }
 
-    pub async fn vpc_list_subnets(
+    pub async fn vpc_subnet_list(
```
bnaecker, line 218: More a question than anything. I noticed changing the ordering of this name to reflect the public API method names, but the `vpc_update_subnet` function below (and others) weren't changed the same way. What's the reasoning behind the naming at this layer?

nexus/types/src/external_api/params.rs
```diff
@@ -150,6 +176,140 @@ impl InstanceSelector {
...
+    pub vpc: NameOrId,
+}
+
+#[derive(Deserialize, JsonSchema)]
+pub struct OptionalVpcSelector {
+    #[serde(flatten)]
+    pub vpc_selector: Option<VpcSelector>,
+}
+
+// TODO-v1: delete this post migration
```
bnaecker, line 221: Not totally sure what this means, but should we track in an issue? I should add, my confusion comes from the fact that the `NetworkInterfaceSelector::new` method has no such note. Should it?

nexus/types/src/external_api/params.rs
```diff
@@ -150,6 +176,140 @@ impl InstanceSelector {
...
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema, PartialEq)]
+pub struct OptionalInstanceSelector {
+    #[serde(flatten)]
+    pub instance_selector: Option<InstanceSelector>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema, PartialEq)]
+pub struct NetworkInterfaceSelector {
```
bnaecker, line 186: I'm sure it's annoying and duplicative, but it would personally really help to have docstrings on these. Since those are likely to be nearly the same for all these `*Selector` types, it'd be enough to have a good docstring on one and refer to it in the others. A block comment in the code describing what these mean would be fine too.

nexus/types/src/external_api/params.rs
```diff
@@ -150,6 +176,140 @@ impl InstanceSelector {
...
+    ) -> Self {
+        VpcSelector {
+            project_selector: project
+                .map(|p| ProjectSelector::new(organization, p)),
+            vpc,
+        }
+    }
+}
+
+#[derive(Deserialize, JsonSchema)]
```
bnaecker, line 236: Just curious really, but why do these `*Selector` types differ in what traits they derive?

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -220,13 +224,26 @@ pub fn external_api() -> NexusApiDescription {
         api.register(vpc_subnet_update)?;
         api.register(vpc_subnet_list_network_interfaces)?;
 
+        api.register(vpc_subnet_list_v1)?;
+        api.register(vpc_subnet_view_v1)?;
+        api.register(vpc_subnet_create_v1)?;
+        api.register(vpc_subnet_delete_v1)?;
+        api.register(vpc_subnet_update_v1)?;
+        // api.register(vpc_subnet_list_network_interfaces_v1)?;
```
bnaecker, line 232: Is this dead code? It looks like there's no such endpoint function itself, so I'd expect we'd just delete this.

zephraph:
I'm going to push forward here. Please, if folks have any feedback don't hesitate to add it after the merge. I'm more than happy to follow up with any fixes or re-work that may be needed.

bnaecker:
Sorry for the delay! I'm nearly done with my comments!

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -4278,57 +4405,81 @@ async fn instance_network_interface_delete(
...
+        let (.., interface) = nexus
+            .network_interface_lookup(&opctx, &interface_selector)?
+            .fetch()
+            .await?;
+        Ok(HttpResponseOk(interface.into()))
+    };
+    apictx.external_latencies.instrument_dropshot_handler(&rqctx, handler).await
+}
+
 /// Fetch a network interface
```
bnaecker, line 4456: Maybe add a note about which new method to use instead, similar to the others.

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -4392,6 +4549,39 @@ async fn instance_network_interface_update(
     apictx.external_latencies.instrument_dropshot_handler(&rqctx, handler).await
 }
 
+/// Update a network interface
+#[endpoint {
+    method = PUT,
+    path = "/organizations/{organization_name}/projects/{project_name}/instances/{instance_name}/network-interfaces/{interface_name}",
+    tags = ["instances"],
```
bnaecker, line 4556: Is this not `deprecated = true`?

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -5398,6 +6103,43 @@ async fn vpc_router_update(
...
+            routes,
+            &marker_for_name_or_id,
+        )?))
+    };
+    apictx.external_latencies.instrument_dropshot_handler(&rqctx, handler).await
+}
+
 // Vpc Router Routes
 
 /// List routes
```
bnaecker, line 6145: I think this method needs `deprecated = true` and a reference to the new API to use.

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -5453,6 +6229,7 @@ struct RouterRoutePathParam {
 }
 
 /// Fetch a route
+/// Use `GET /v1/vpc-router-routes/{route}` instead
 #[endpoint {
     method = GET,
     path = "/organizations/{organization_name}/projects/{project_name}/vpcs/{vpc_name}/routers/{router_name}/routes/{route_name}",
```
bnaecker, line 6235: Needs `deprecated = true`.

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -5493,17 +6271,54 @@ async fn vpc_router_route_view_by_id(
...
+                &create,
+            )
+            .await?;
+        Ok(HttpResponseCreated(route.into()))
+    };
+    apictx.external_latencies.instrument_dropshot_handler(&rqctx, handler).await
+}
+
 /// Create a router
 #[endpoint {
```
bnaecker, line 6323: Needs the `deprecated = true` treatment, I think.

bnaecker (commented):
Hey Justin! This is an enormous amount of work. Thank you so much for taking it on, this looks like a big improvement. I have a few comments, all related to documentation and comments. There's nothing in the code itself that appears out of order. Thanks again, great work!

nexus/src/app/network_interface.rs
```diff
@@ -0,0 +1,210 @@
...
+              instance_selector: Some(_),
+              network_interface: NameOrId::Id(_),
+            } => {
+              Err(Error::invalid_request(
+                "when providing network_interface as an id, instance_selector should not be specified"
+              ))
+            }
+            _ => {
+              Err(Error::invalid_request(
+                "network_interface should either be a UUID or instance should be specified"
```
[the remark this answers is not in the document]
zephraph: It's somewhat formatted to match the name of the query parameter the users would see in the doc. I have a TODO in the parent PR to improve error messages though and this is definitely one of those areas that can use it. For now this is consistent with the others.

nexus/src/app/vpc.rs
```diff
@@ -39,6 +41,42 @@ use uuid::Uuid;
...
+                let vpc = self
+                    .project_lookup(opctx, selector)?
+                    .vpc_name(Name::ref_cast(name));
+                Ok(vpc)
+            }
+            params::VpcSelector {
+                vpc: NameOrId::Id(_),
+                project_selector: Some(_),
+            } => Err(Error::invalid_request(
+                "when providing vpc as an ID, project should not be specified",
```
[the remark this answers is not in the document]
zephraph: `project_selector` itself means `project: ID` or `project: Name, Organization: NameOrId` which is why it's calling out project specifically. It's true that this isn't really complete though, because if `project` was removed but `organization` was left that would probably still be permissible (even if not strictly correct). I'll add a note to create some test cases around these failure modes too.

nexus/src/app/vpc_router.rs
zephraph: Yep! good catch.

nexus/src/app/vpc_subnet.rs
zephraph: I've got a follow up to fully correct all of these I just didn't do it all in this one PR to reduce the noise. Overall my intention here is to have a consistent naming scheme. I'll open up a broader PR for the renames and we can discuss in detail if it's the right move there.

nexus/types/src/external_api/params.rs
zephraph: It's tracked in the parent issue. All the `::new` interfaces should have that note so I just missed it there.

nexus/src/app/vpc.rs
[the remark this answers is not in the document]
bnaecker: Ah, I see, that helps. I think my confusion began with `instance_selector`, used above. If that were changed to `instance` then these would all be consistent, which seems a marked improvement!

nexus/src/app/vpc_subnet.rs
bnaecker: Sounds good, and like you're on top of it. Thanks!

nexus/types/src/external_api/params.rs
bnaecker: Ah, yeah, I didn't mean for OpenAPI consumers, but for developers working on Omicron itself. I realize that's pretty annoying since you've written it all and know exactly what everything means, so I definitely think the block-comment-and-references approach is sufficient!

nexus/src/external_api/http_entrypoints.rs
zephraph: Yep, good catch!

nexus/src/external_api/http_entrypoints.rs
bnaecker: Maybe add a TODO item to the parent issue? I can't tell if it's already covered, so ignore me if so :)

nexus/types/src/external_api/params.rs
zephraph: If a selector is the parent of another selector then it has to implement extra traits to be used in the lookup chain. If it's a leaf selector though, it doesn't really need to. Ultimately the plan is to have these all generated by macros and not individually implemented so it'll likely just be consistent in the end.

nexus/src/app/vpc.rs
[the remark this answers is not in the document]
zephraph: I'll add a follow up for that.

nexus/types/src/external_api/params.rs
bnaecker: Sounds good, thanks for the explanation!

nexus/types/src/external_api/params.rs
zephraph: Ah, no, that makes sense. I didn't understand what you were saying in the beginning. I'm going to take another pass in this file post the V1 migration and I'll make sure there are good docs for everything here.

nexus/src/app/vpc_router.rs
bnaecker: FYI, looks like this one didn't get fixed up before the merge. Probably want to fix it up later.

nexus/types/src/external_api/params.rs
```diff
@@ -35,6 +36,31 @@ pub struct InstancePath {
     pub instance: NameOrId,
 }
 
+#[derive(Deserialize, JsonSchema)]
+pub struct NetworkInterfacePath {
+    pub interface: NameOrId,
```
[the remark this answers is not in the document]
bnaecker: Same here, looks like these didn't get comments attached attached prior to merge, hopefully we can add those at some point.

nexus/src/app/vpc_router.rs
zephraph: Yeah, I'll make a docs sweep.

nexus/types/src/external_api/params.rs
karencfv: We did that at "previous job" (have a common "resource" struct), and it was such a pain when we needed to add additional fields for a new version, it ended up being a spaghetti nightmare 😅

PR #2058, opened 2022-12-16
[sagas] Make a macro to simplify declaring saga actions

smklein:
Writing sagas involves a lot of boilerplate:

- When creating `NexusAction` nodes, it's convention to re-state the name of the saga in each label (and also name the label after the variable itself). This is easily machine-generated.
- After defining all the `NexusAction` nodes, they must then all be re-listed in `register_actions`.
- When actually creating the DAG for the saga, each node is re-stated. This is the spot where node "output names" can be defined, but the label is another variation of the `NexusAction` node name (usually camelcased).

This PR introduces (and uses) a macro called `declare_saga_actions`.

This macro aims to reduce this boilerplate, by requiring only the following:
- The name of the saga
- The name of each action
- The output of each action
- The "forward" action function
- (Optional) The "undo" action function

It then generates nodes with names that are automatically "the right format", and produces some helper functions:
- `fn <saga name>_register_actions(registry: &mut ActionRegistry)` to help implement `NexusSaga::register_actions`
- `fn <node>_action() -> Node` to produce a default `Node` implementation when generating the DAG

### DIFF
--- a/nexus/src/app/sagas/mod.rs
+++ b/nexus/src/app/sagas/mod.rs
@@ -154,7 +154,7 @@ macro_rules! __action_name {
 ///
 /// For this input:
 ///
-/// ```
+/// ```ignore
 /// declare_saga_actions! {
 ///     my_saga;
 ///     SAGA_NODE1 -> "output1" {
--- a/nexus/src/app/sagas/mod.rs
+++ b/nexus/src/app/sagas/mod.rs
@@ -151,6 +151,13 @@ macro_rules! __action_name {
 
 /// A macro intended to reduce boilerplate when writing saga actions.
 ///
+/// This macro aims to reduce this boilerplate, by requiring only the following:
+/// - The name of the saga
+/// - The name of each action
+/// - The output of each action
+/// - The "forward" action function
+/// - (Optional) The "undo" action function
+///
 /// For this input:
 ///
 /// ```ignore
--- a/nexus/src/app/sagas/mod.rs
+++ b/nexus/src/app/sagas/mod.rs
@@ -10,7 +10,7 @@
 // easier it will be to test, version, and update in deployed systems.
 
 use crate::saga_interface::SagaContext;
-use lazy_static::lazy_static;
+use once_cell::sync::Lazy;
 use std::sync::Arc;
 use steno::new_action_noop_undo;
 use steno::ActionContext;
@@ -79,12 +79,11 @@ impl From<SagaInitError> for omicron_common::api::external::Error {
     }
 }
 
-lazy_static! {
-    pub(super) static ref ACTION_GENERATE_ID: NexusAction =
-        new_action_noop_undo("common.uuid_generate", saga_generate_uuid);
-    pub static ref ACTION_REGISTRY: Arc<ActionRegistry> =
-        Arc::new(make_action_registry());
-}
+pub(super) static ACTION_GENERATE_ID: Lazy<NexusAction> = Lazy::new(|| {
+    new_action_noop_undo("common.uuid_generate", saga_generate_uuid)
+});
+pub static ACTION_REGISTRY: Lazy<Arc<ActionRegistry>> =
+    Lazy::new(|| Arc::new(make_action_registry()));
 
 fn make_action_registry() -> ActionRegistry {
     let mut registry = steno::ActionRegistry::new();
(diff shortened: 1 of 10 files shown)


zephraph:
This is so much better! I'm gonna leave it to other folks to review because I don't have faith in my macro foo, but thanks for doing this.

nexus/src/app/sagas/mod.rs
```diff
@@ -116,3 +116,129 @@ pub(super) async fn saga_generate_uuid<UserType: SagaType>(
...
+    //  S = SagaName <old nodes> <> ...
+    //
+    // Basically, everything to the left of "<>" is just us propagating state
+    // through the macro, and everything to the right of it is user input.
+    (S = $saga:ident $($nodes:ident),* <> $node:ident -> $out:literal { + $a:ident - $u:ident } $($tail:tt)*) => {
+        lazy_static::lazy_static! {
+            static ref $node: crate::app::sagas::NexusAction = ::steno::ActionFunc::new_action(
+                crate::app::sagas::__action_name!($saga, $node), $a, $u,
+            );
+        }
```
sunshowers, line 210: Can we use `once_cell::Lazy` here (maybe in a followup since this just refactors code?) `lazy_static` is mostly deprecated at this point I think.

nexus/src/app/sagas/snapshot_create.rs
```diff
@@ -173,14 +161,7 @@ impl NexusSaga for SagaSnapshotCreate {
...
     fn register_actions(registry: &mut ActionRegistry) {
-        registry.register(Arc::clone(&*REGIONS_ALLOC));
-        registry.register(Arc::clone(&*REGIONS_ENSURE));
-        registry.register(Arc::clone(&*CREATE_DESTINATION_VOLUME_RECORD));
-        registry.register(Arc::clone(&*CREATE_SNAPSHOT_RECORD));
-        registry.register(Arc::clone(&*SEND_SNAPSHOT_REQUEST));
-        registry.register(Arc::clone(&*START_RUNNING_SNAPSHOT));
-        registry.register(Arc::clone(&*CREATE_VOLUME_RECORD));
-        registry.register(Arc::clone(&*FINALIZE_SNAPSHOT_RECORD));
+        snapshot_create_register_actions(registry);
```
sunshowers, line 164: Being less error-prone here is really nice!

nexus/src/app/sagas/snapshot_create.rs
```diff
@@ -207,58 +188,17 @@ impl NexusSaga for SagaSnapshotCreate {
...
-            "snapshot_request",
-            "SendSnapshotRequest",
-            SEND_SNAPSHOT_REQUEST.as_ref(),
-        ));
-
+        builder.append(regions_alloc_action());
+        builder.append(regions_ensure_action());
+        builder.append(create_destination_volume_record_action());
+        builder.append(create_snapshot_record_action());
+        builder.append(send_snapshot_request_action());
```
sunshowers, line 195: Can this use the same pattern as `register_actions` above where you don't have to specify all the actions?

nexus/src/app/sagas/mod.rs
```diff
@@ -116,3 +116,129 @@ pub(super) async fn saga_generate_uuid<UserType: SagaType>(
...
+///     - `fn saga_node1_action() -> steno::Node` referencing this node, with an
+///     output named "output1".
+/// - For `SAGA_NODE2`:
+///     - A `NexusAction` labeled "my_saga.saga_node2" (containing "do2").
+///     - `fn saga_node2_action() -> steno::Node` referencing this node, with an
+///     output named "output2".
+/// - For `my_saga`:
+///     - `fn my_saga_register_actions(...)`, which can be called to implement
+///     `NexusSaga::register_actions`.
+macro_rules! declare_saga_actions {
```
sunshowers, line 182: Do we expect to make this a public macro (`macro_export`) in the future, maybe for other parts of the code that use sagas?

sunshowers (approved):
I think this macro does a good job enhancing the clarity of the code. Just a few questions but nothing blocking.

leftwo (approved):
Love it.

Would it be a good idea to put a block comment somewhere to help new saga authors
figure out what they need to provide to the macro, much like the write up you have
here for explaining the PR itself?   Somethign like this list:

```
This macro aims to reduce this boilerplate, by requiring only the following:

The name of the saga
The name of each action
The output of each action
The "forward" action function
(Optional) The "undo" action function
```

nexus/src/app/sagas/mod.rs
smklein: Done in 6abce0865dfb225b3581f05df7a544b2561168f8 ! We are using `lazy_static` elsewhere in Omicron, but that can be converted too

nexus/src/app/sagas/snapshot_create.rs
smklein: Ah, good question. This is a part I didn't feel comfortable automating yet.

In the case of *this particular saga* - `snapshot_create` - we just call all the actions that we declared, in order.

However, not all our sagas are this simple - the DAG in `make_saga_dag` describes the control flow of execution, which may be more complex than "all the actions we declared in the order of declaration".

For example...

... Constants can be embedded into the DAG:

https://github.com/oxidecomputer/omicron/blob/2f453125d0ecb95f091706976368589c073a7d35/nexus/src/app/sagas/instance_create.rs#L142-L147

... Nodes that are shared among multiple sagas can be called (such as generating UUIDs):

https://github.com/oxidecomputer/omicron/blob/2f453125d0ecb95f091706976368589c073a7d35/nexus/src/app/sagas/instance_create.rs#L149-L153

... Some nodes can be called in parallel, or called as subsagas:

https://github.com/oxidecomputer/omicron/blob/2f453125d0ecb95f091706976368589c073a7d35/nexus/src/app/sagas/instance_create.rs#L223-L245

nexus/src/app/sagas/mod.rs
smklein: Good question! Right now, I expected to keep this "Nexus-internal", since my short-term goal was to just make it easier to add new sagas.

However, if we like the structure of this macro, I could see us making it public, or moving it into the `steno` repo itself.

smklein:
> Love it.
> 
> Would it be a good idea to put a block comment somewhere to help new saga authors figure out what they need to provide to the macro, much like the write up you have here for explaining the PR itself? Somethign like this list:
> 
> ```
> This macro aims to reduce this boilerplate, by requiring only the following:
> 
> The name of the saga
> The name of each action
> The output of each action
> The "forward" action function
> (Optional) The "undo" action function
> ```

I extended the macro's documentation to list this within https://github.com/oxidecomputer/omicron/pull/2058/commits/29a3ab5a00001777ef2a1a738c81a549051fb37b

davepacheco:
Thanks for doing this.  I'm sorry I didn't get to the review before it landed.  It's really nice to have a lot less repetition here.

One thing to keep in mind for the future is that there are reasons the action name, node label, and output name might be different.  The only example I found today was with the common uuid.generate action (which is kind of obviated after dynamic DAGs anyway).  The idea is you might have a common piece of code (the action) that's used in multiple places in the graph or multiple different graphs to generate different outputs (e.g., instance_id, disk_id, etc.).  I don't think this changes much here.  I think it just means that if someone decides it would be useful to have different action name / node label / node name, they can't use this macro or they need to generalize the macro first somehow.  (It's also possible that allowing these to be different isn't that useful an idea and Steno shouldn't require so many strings that in practice are going to be all the same.)

Issue #2034, opened 2022-12-09
[nexus] Instance deletion is not a saga, but it should be


smklein:
It has multiple non-atomic steps that could independently fail; if we don't retry we'll risk leaking resources.

PR #2060, opened 2022-12-16
[nexus] Instance Deletion is now a saga

smklein:
Fixes https://github.com/oxidecomputer/omicron/issues/2034


nexus/src/app/sagas/instance_delete.rs
```diff
@@ -0,0 +1,106 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
```
zephraph, line 3: This is totally an aside... are we supposed to have license headers on every file?

smklein: I believe so? Seems to be the case for nearly every rust file in Omicron

nexus/src/app/sagas/instance_delete.rs
```diff
@@ -0,0 +1,106 @@
...
+    let osagactx = sagactx.user_data();
+    let params = sagactx.saga_params::<Params>()?;
+    let opctx = OpContext::for_saga_action(&sagactx, &params.serialized_authn);
+    osagactx
+        .datastore()
+        .project_delete_instance(&opctx, &params.authz_instance)
+        .await
+        .map_err(ActionError::action_failed)?;
+    Ok(())
+}
```
zephraph, line 75: It looks like part of what `project_delete_instance` is doing is going through an detaching disks. I guess that's really just updating the db record for the disks to unassociate them with the instance (via `Instance::detach_resource`). Looking in the detach disk call it's using the same `Instance::detach_resource`. I was wondering if that action, detaching disks, should be hoisted to the saga. Given that it's basically using the same thing under the hood I suppose not?

nexus/src/app/sagas/instance_delete.rs
```diff
@@ -0,0 +1,106 @@
...
+    INSTANCE_DELETE_RECORD -> "no_result1" {
+        + sid_delete_instance_record
+    }
+    DELETE_NETWORK_INTERFACES -> "no_result2" {
+        + sid_delete_network_interfaces
+    }
+    DEALLOCATE_EXTERNAL_IP -> "no_result3" {
+        + sid_deallocate_external_ip
+    }
+}
```
zephraph, line 36: Should there be any rollback cases represented here? What happens if a step fails?

zephraph (approved):
Overall this looks good. The new saga macro really helps make this easier to follow. Had a question about rollbacks and somewhat of a question about disk detachment but otherwise I'm 👍

nexus/authz-macros/src/lib.rs
```diff
@@ -278,7 +278,7 @@ fn do_authz_resource(
 
     Ok(quote! {
         #[doc = #doc_struct]
-        #[derive(Clone, Debug)]
+        #[derive(Clone, Debug, Serialize, Deserialize)]
```
davepacheco, line 281: I think we talked about this offline or I at least raised the question.  This makes these internal structures a committed interface because they'll be stored into the database.  Is that not a problem?

nexus/db-model/src/update_artifact.rs
```diff
@@ -7,14 +7,16 @@ use crate::schema::update_available_artifact;
...
+use serde::Serialize;
 use std::io::Write;
 
 impl_enum_wrapper!(
     #[derive(SqlType, Debug, QueryId)]
     #[diesel(postgres_type(name = "update_artifact_kind"))]
     pub struct UpdateArtifactKindEnum;
 
-    #[derive(Clone, Copy, Debug, Display, AsExpression, FromSqlRow, PartialEq, Eq)]
+    #[derive(Clone, Copy, Debug, Display, AsExpression, FromSqlRow, PartialEq, Eq, Serialize, Deserialize)]
```
davepacheco, line 19: Same here -- is this now more of a committed interface than it was?  Maybe not, since it seems like this was already in the database.

nexus/src/authz/api_resources.rs
```diff
@@ -150,7 +150,7 @@ impl<T: ApiResource + oso::PolarClass + Clone> AuthorizedResource for T {
 /// This object is used for authorization checks on a Fleet by passing it as the
 /// `resource` argument to [`crate::context::OpContext::authorize()`].  You
 /// don't construct a `Fleet` yourself -- use the global [`FLEET`].
-#[derive(Clone, Copy, Debug)]
+#[derive(Clone, Copy, Debug, Serialize, Deserialize)]
```
davepacheco, line 153: How does this get serialized? `{}`?

nexus/src/app/sagas/instance_delete.rs
```diff
@@ -0,0 +1,106 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
```
davepacheco, line 4: Is there a reason this is a saga and not just a transaction that does all three of these things?   Is it just the lack of good support for non-interactive transactions?

davepacheco (commented):
Thanks for fixing this.

nexus/src/app/sagas/instance_delete.rs
smklein: I do not think it should be.

When I wrote `DatastoreDetachManyTarget::detach_resources` - which `Instance` objects implement, so they could detach disks - the whole point of it was that the detach operation (on disks) could be issued as part of the same statement that deletes instances, within a [common table expression](https://www.cockroachlabs.com/docs/stable/common-table-expressions.html) being issued to CRDB.

By existing within a single CTE, the statement is transactional, even though it modifies multiple tables and only sends a single request to CRDB. This follows the advice of https://www.cockroachlabs.com/docs/v22.2/performance-best-practices-overview#reduce-transaction-contention

Using CTEs, where possible, is generally our "best performance" way to combine multiple SQL statements. My usage of sagas elsewhere is a worse-performance attempt to get similar properties, in a non-atomic way.

smklein: Related: Ideally, each function exposed from the `datastore()` would only issue a single DB statement, so that it either "fully completes" or "does nothing at all".

`project_delete_instance`, by virtue of being a CTE, already abides by this principle.

nexus/src/app/sagas/instance_delete.rs
smklein: There *could* be; I'm not 100% sure there *should* be? If we're able to delete the instance record, I think we should be "best-effort" rolling forward deletion, rather than attempting to undo the deletion. 

For example, suppose we cannot re-allocate the deleted resource: either the resource itself was exhausted, user-data was deleted, or a sled cooperating in the allocation is no longer around. What should we do?

By having deletion be a saga, it's at least *durable* - even if Nexus crashes, it'll finish the work when it comes back online - but I'm not sure about the value of "making deletion undoable" vs "making sure it rolls forward, whatever that means".

TL;DR: *Maybe*? But there is a lot of complexity down the road of undoing deletions that I'm not really sure how to handle yet.

nexus/authz-macros/src/lib.rs
smklein: I thought this was the topic of RFD 289: [Steno Upgrade](https://rfd.shared.oxide.computer/rfd/289).

> [Determinations](https://rfd.shared.oxide.computer/rfd/289#_determinations)
> For a given saga, execution is only ever carried out by one version of Nexus. Failover may still cause execution to move between different instances of Nexus.
>
> There are two pieces to this: draining sagas during upgrade and having Nexus recover only sagas that were created with the same version as itself.

If we *don't* do this, then Nexus will need to redo lookup of relevant resources repeatedly within each saga step. In addition to the cost of looking up / recording saga states, this seems like it amplifies the total number of database operations needed to complete the "real work" within a saga.

nexus/db-model/src/update_artifact.rs
smklein: I was relying on the determination of RFD 289 for this rationale too. Changing the structs within `db-model` would *already* be a breaking change - requiring the old version of Nexus to perform saga draining - so adding the serde derives here doesn't seem like a new constraint.

nexus/src/app/sagas/instance_delete.rs
smklein: Yeah, basically. Any saga steps that purely talk to CRDB could theoretically be combined, but my main goals were:

1. Make sure we have crash-consistency (before: we weren't durable, now, we are)
2. Avoid interactive transactions. They're slow, and even worse, their contention has resulted in CRDB errors like https://www.cockroachlabs.com/docs/stable/transaction-retry-error-reference.html#retry_write_too_old . Those errors are a bummer, especially for high-volume operations, because they consume compute time and *don't necessarily make forward progress*. 

I'm very open to smushing these DB-only steps together, but putting them in a saga felt like a reasonable intermediate step.

(Also, when we add dendrite integration, I think we're going to have some non-DB-only steps here *anyway*, so having this operation be a saga is something we will probably want)

smklein: I was also doing some reading on https://www.cockroachlabs.com/docs/stable/transactions.html#batched-statements , which seems great, but has the caveat: the ability for CRDB to perform an automatic retry without another network round-trip has some caveats:
- The size of results must be less than 16 KiB, or automatic retry is disabled
- If any part of the transaction relies on conditional logic, or automatic retry is disabled

I still think it's worth it to pursue this path, but it felt like an "optimization" above multiple saga steps. In contrast, the following felt like correctness issues:
- Lack of durability across crashes
- Throwing errors unnecessarily back to the client because of database contention

nexus/src/authz/api_resources.rs
smklein: I think that depends on the serializer. For `serde-json`, which is used by steno, I believe the name of the struct is used:

https://serde.rs/json.html

nexus/authz-macros/src/lib.rs
smklein: I also view this as a major tension between:

- `authz` objects are becoming more-and-more a required input to datastore operations...
- ... and many saga operations want to issue exactly one datastore operation.

So this leaves us with a few options:
1. **No serialization, authz is still required for datastore ops**: This means we need to repeatedly lookup each resource we're acting upon, and inflates the total number of requests to the db. This also means that a permission change mid-operation could cause some odd failure modes.
2. **Do authorization checks earlier**. Basically, rather than passing an `authz` object to datastore for evaluation, pass around an object that implies "the caller is *already authorized to perform the operation*". This new object would need to be serializable and provided as input to datastore functions, but the `authz::Resource` objects could remain outside the saga log.
3. **Serialize the `authz::Resource` objects, use them in the datastore operations as they exist today**. This is basically what this PR does, by adding the derives.

davepacheco: > I thought this was the topic of RFD 289

Yeah, good call.  I think as long as we only ever serialize these into saga state (and not elsewhere in the database, for example, or over HTTP interfaces), I think you're right that this is fine.

davepacheco (approved):

Issue #2059, opened 2022-12-16
More Sagas


smklein:
Some operations in Nexus should become sagas. This issue tracks any and all operations that should become sagas:

- [ ] [Instance Deletion](https://github.com/oxidecomputer/omicron/issues/2034)
- [ ] Snapshot Deletion
- [ ] Project Creation
- [ ] Project Deletion
- [ ] Org Creation (Unless we remove orgs?)
- [ ] Org Deletion (Unless we remove orgs?)
- [ ] Silo Creation
- [ ] Silo Deletion

See also: https://github.com/oxidecomputer/omicron/issues/2052

PR #2081, opened 2022-12-19
[nexus] Project and VPC creation are now sagas

smklein:
- Adds a VPC creation saga
- Adds a project creation saga, which uses the VPC creation as a subsaga
- Does not yet add undo actions; that's follow-up work

Part of https://github.com/oxidecomputer/omicron/issues/2059


plotnick (approved):
Sorry I missed this; I saw the "review requested" notification for #2086, but somehow not this one. In conjunction with that, this looks great.

PR #2082, opened 2022-12-19
Include extent information in Crucible VolumeConstructionRequests

gjcolombo:
Pick up the latest definition of `crucible_client_types::VolumeConstructionRequest` and the corresponding Propolis bits. The new definition requires a `VolumeConstructionRequest::Region` to specify the number of extents in the region and the number of blocks in each extent. The disk creation saga already has all this data (it needs it to ask Crucible to create the regions in the first place), so this is just a matter of plumbing that data into the serialized VCR that Nexus saves in its database.

This change is a prerequisite to live migrating instances. See [crucible#557](https://github.com/oxidecomputer/crucible/pull/557) and [propolis#275](https://github.com/oxidecomputer/propolis/pull/275) for more context.

Tests: `cargo test`; installed Omicron locally and boot-tested a couple of instances.

### DIFF
--- a/nexus/tests/integration_tests/volume_management.rs
+++ b/nexus/tests/integration_tests/volume_management.rs
@@ -1741,6 +1741,8 @@ fn create_region(
 ) -> VolumeConstructionRequest {
     VolumeConstructionRequest::Region {
         block_size,
+        blocks_per_extent: 1,
+        extent_count: 1,
         gen,
         opts: CrucibleOpts {
             id,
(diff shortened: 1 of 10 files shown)


nexus/tests/integration_tests/volume_management.rs
```diff
@@ -1741,6 +1741,8 @@ fn create_region(
 ) -> VolumeConstructionRequest {
     VolumeConstructionRequest::Region {
         block_size,
+        blocks_per_extent: 1,
+        extent_count: 1,
```
leftwo, line 1745: lol.   I don't know why but yeah, a 1 extent sized region makes me laugh.

leftwo (approved):

nexus/tests/integration_tests/volume_management.rs
gjcolombo: a stocking stuffer-sized extent

gjcolombo:
Thanks for the review!

Issue #2052, opened 2022-12-13
Tracking issue for saga unwind safety


smklein:
- [x] Instance create: https://github.com/oxidecomputer/omicron/pull/1843
- [x] Disk create: https://github.com/oxidecomputer/omicron/pull/1835
- [ ] Disk Delete
- [ ] Instance Migration
- [ ] Snapshot Create
- [ ] Volume Delete
- [ ] Volume Remove ROP

If other sagas are added without unwind safety tests, they should be added to this list.

Related to https://github.com/oxidecomputer/omicron/issues/1799

askfongjojo:
We've seen at least two incidents of failed `instance_create` unwind on rack2 resulting in panic. One was #3212 and another was caused by a networking issue which prevented the vnic from being cleaned up:
```
BRM42220031 # cat /var/svc/log/oxide-sled-agent\:default.log | looker -l error
WARNING: Failed to delete OPTE port overlay VNIC while dropping port. The VNIC will not be cleaned up properly, and the xde device itself will not be deleted. Both the VNIC and the xde device must be deleted out of band, and it will not be possible to recreate the xde device until then. Error: DeleteVnicError { name: "vopte4", err: CommandFailure(CommandFailureInfo { command: "/usr/sbin/dladm delete-vnic vopte4", status: ExitStatus(unix_wait_status(256)), stdout: "", stderr: "dladm: vnic deletion failed: link busy\n" }) }
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: AdmError { op: Uninstall, zone: "oxz_propolis-server_2df3f0da-9b07-47bd-85de-a89fc640383f", err: CommandOutput(CommandOutputError("exit code 1\nstdout:\n\nstderr:\nzoneadm: zone 'oxz_propolis-server_2df3f0da-9b07-47bd-85de-a89fc640383f': uninstall operation is invalid for shutting_down zones.")) }', sled-agent/src/instance.rs:476:64
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
[ May 26 17:59:33 Stopping because all processes in service exited. ]
[ May 26 17:59:33 Executing stop method (:kill). ]
[ May 26 17:59:33 Executing start method ("ctrun -l child -o noorphan,regent /opt/oxide/sled-agent/sled-agent run /opt/oxide/sled-agent/pkg/config.toml &"). ]
```

PR #2086, opened 2022-12-21
[nexus] Make the VPC creation saga unwind-safe, add tests

smklein:
- Defines undo actions for VPC creation
- Adds test for the VPC creation saga

Part of #2052


nexus/src/app/sagas/vpc_create.rs
```diff
@@ -315,6 +374,21 @@ async fn svc_update_firewall(
...
+    sagactx: NexusActionContext,
+) -> Result<(), anyhow::Error> {
+    let osagactx = sagactx.user_data();
+    let params = sagactx.saga_params::<Params>()?;
+    let opctx = OpContext::for_saga_action(&sagactx, &params.serialized_authn);
+    let (authz_vpc, _) =
+        sagactx.lookup::<(authz::Vpc, db::model::Vpc)>("vpc")?;
+    osagactx
+        .datastore()
+        .vpc_update_firewall_rules(&opctx, &authz_vpc, vec![])
```
plotnick, line 387: With the caveat that I don't really understand Sagas yet, this empty vector seems odd to me. Wouldn't we want an undo to restore the old set of rules, or at least the default rules? I don't think we default to an empty set anywhere else.

nexus/src/app/sagas/vpc_create.rs
```diff
@@ -124,11 +129,6 @@ async fn svc_create_vpc(
     let vpc_id = sagactx.lookup::<Uuid>("vpc_id")?;
     let system_router_id = sagactx.lookup::<Uuid>("system_router_id")?;
 
-    // TODO: This is both fake and utter nonsense. It should be eventually
```
plotnick, line 127: I am very glad this TODO can go away.

plotnick (approved):
Thank you for taking this on! It's great that we have the tooling in place to make these flows more robust. Modulo my concern about empty firewall rules in the undo action, this looks great to me.

nexus/src/app/sagas/vpc_create.rs
smklein: I believe the implementation of `vpc_update_firewall_rules` does the following:

- Deletes all firewall rules attached to the supplied `vpc_id`
- Inserts new rules, which reference `vpc_id`

This is why we're passing the empty vec - it clears out the old rules, and doesn't insert anything new.

Since this is an "undo" step along the pathway of VPC creation, I don't *think* it matters that we chuck the firewall rules -- we're making this call to delete all resources that might reference the VPC before we remove it.

TL;DR: I cleared it out to clean our SQL tables before removing the associated VPC. Does that seem okay?

nexus/src/app/sagas/vpc_create.rs
```diff
@@ -333,3 +407,323 @@ async fn svc_notify_sleds(
...
+                datastore.pool_for_tests().await.unwrap(),
+            )
+            .await
+            .optional()
+            .unwrap()
+            .map(|rule| {
+                eprintln!("Firewall rule exists: {rule:?}");
+            })
+            .is_none()
+    }
```
smklein, line 649: FYI, if I *don't* remove the firewall rules, this fails -- unwinding the saga would still leave old firewall rules around.

nexus/src/app/sagas/vpc_create.rs
plotnick: Thank you for that very helpful explanation; I see now. Yes, that seems fine.

PR #2087, opened 2022-12-21
[nexus] Make project creation unwind safe, add tests

smklein:
- Defines undo actions for project creation saga
- Adds test for the project creation saga

Part of #2052


nexus/src/app/sagas/project_create.rs
```diff
@@ -88,25 +88,42 @@ async fn spc_create_record(
...
 
+async fn spc_create_record_undo(
+    sagactx: NexusActionContext,
+) -> Result<(), anyhow::Error> {
+    let osagactx = sagactx.user_data();
+    let params = sagactx.saga_params::<Params>()?;
+    let opctx = OpContext::for_saga_action(&sagactx, &params.serialized_authn);
+
+    let (_authz_project, project) =
+        sagactx.lookup::<(authz::Project, db::model::Project)>("project")?;
```
zephraph, line 99:
```suggestion
    let (.., project) =
        sagactx.lookup::<(authz::Project, db::model::Project)>("project")?;
```
Would this work?

smklein: They're functionally the same; I was just being explicit about "what is the unused value" in this case since deserialization is particularly important on the `lookup` function, which is generic based on the type parameters.

smklein: Also, to be pedantic in the context of `let` binding:

1. `let (.., foo) = ...` means "destructure and bind one or more values to the anonymous "...", but only bind the last value to `foo`"
2. `let (_, foo) = ...` means "destructure and bind exactly one value to `_`, and bind the second value to `foo`"
3. `let (_unused, foo) = ...` means "destructure and bind exactly one value to `_unused`, and bind the second value to `foo`"

Technically in cases (1) and (2), since the bind is anonymous, it also goes out of scope, and `drop` is called immediately on the unnamed values. In contrast, in (3), `drop` is only called on `_unused` when it actually goes out of scope.

However, I don't think `authz::Project` has a `drop` implementation, hence why this is the same.

So it's not *exactly* the same, but it's *basically* the same - I just used this for readability, since the types of values being pulled out of the `sagactx.lookup` are important to get right, and **caller-specified**.

nexus/src/app/project.rs
```diff
@@ -85,8 +85,10 @@ impl super::Nexus {
                 saga_params,
             )
             .await?;
-        let db_project = saga_outputs
-            .lookup_node_output::<db::model::Project>("project")
+        let (_authz_project, db_project) = saga_outputs
+            .lookup_node_output::<(authz::Project, db::model::Project)>(
```
plotnick, line 89: Question: is it _safe_ to store `authz` objects in the Saga like this? I (think I) understand the motivation to do so, I'm just wondering (probably too informally) if they could somehow be extracted and misused, or if having `authz` instances be deserializable opens the door to attacks like starting with a fake serialization and getting a real but unauthorized `authz` object from that.

nexus/src/app/sagas/project_create.rs
```diff
@@ -125,3 +142,167 @@ async fn spc_create_vpc_params(
...
+
+    #[nexus_test(server = crate::Server)]
+    async fn test_saga_basic_usage_succeeds(
+        cptestctx: &ControlPlaneTestContext,
+    ) {
+        let client = &cptestctx.external_client;
+        let nexus = &cptestctx.server.apictx.nexus;
+        let org_id = create_org(&client).await;
+
+        // Before running the test, confirm we have no records of any VPCs.
```
plotnick, line 248:
```suggestion
        // Before running the test, confirm we have no records of any projects.
```

plotnick (approved):
Left a (non-blocking) security related question, but otherwise looks great, especially the tests; `verify_clean_slate` is awesome and I'll be stealing that idea in the future.

nexus/src/app/project.rs
smklein: I discussed something similar with @davepacheco last week: https://github.com/oxidecomputer/omicron/pull/2060#discussion_r1054890025

Let's suppose we don't do this, and that we refuse to serialize any authz objects. In that case, every saga step acting on an object would need to look it up again, and re-authorize. One particularly weird fall-out of this is that changing authorizations for a particular user mid-saga could result in some gnarly edge-cases. For example:

- Suppose a user is authorized to create / modify disks
- Suppose the user is partway through a disk-creation saga
- Suppose that saga *fails*, for whatever reason (Out of space? Crucible bug? Whatever)
- Suppose that **concurrently, the authorizations granted to that user are modified to disallow disk modification**.
- Now, when unwinding the saga operations, we **cannot actually remove the partially-created disk, because our authorization check would fail**. 

This seems quite bad to me. I understand that there are other latent concerns around authz objects, such as "can we guarantee they are constructed safely", but an inability to pass them across saga nodes implies that "the whole authz world might change" between saga operations, which seems like it could make unwinding fail arbitrarily.

plotnick: Thanks; I always feel better knowing I'm thinking along similar lines as Dave :smile: Your arguments make total sense from a practical standpoint, and the concerns raised are so far entirely hypothetical. But I will paste a quote from the linked comment here just for emphasis:
> as long as we only ever serialize these into saga state (and not elsewhere in the database, for example, or over HTTP interfaces), I think you're right that this is fine.

One of the things that I take from your example above is that using sagas implies a kind of temporal snapshotting, rather like a closure: when we create a saga, we need to capture the state associated with the relevant database objects at that instant in time. That necessarily includes authorization for the included queries/modifications, and I think you're right that using a new context to undo old actions seems semantically wrong. So, all this to say: thanks for considering the question, I've come around and see now this is the only way to properly handle this with the current machinery.

Issue #1597, opened 2022-08-16
[sled agent] Consider setting uniform coreadm values to extract info from terminating processes?


smklein:
In the 8/16 control plane sync, we discussed the possibility of using https://illumos.org/man/8/coreadm to set a filter to extract core files from crashing non-global zones into the global zone.

Currently, when non-global zone services terminate, Sled Agent stops and deletes the underlying zone. This helps avoid leakage of that resource - we have no further execution-time usage for it - but limits visibility.

By dumping core files into the global zone, we'd be able to inspect errors, even after the zone is destroyed.

rmustacc:
In particular, we want to enable global cores and use the `%z` token to include the zone name for disambiguation. In the past we've done things like `/var/xxx/%z/core.%f.%p`.

jclulow:
A request: probably please don't hard code `/var` paths (or use of `rpool` specifically) for any more large files.  It seems fine as a default but in the ramdisk environment we're going to want to direct those files to specific tmpfs or other mounted pools, etc.  The current use of `/var/oxide` for writing a bunch of larger files is something we'll probably have to unwind so it'll be good to avoid adding more things like that.

leftwo:
> In the past we've done things like /var/xxx/%z/core.%f.%p.

coreadm wants the directory to exist before it will create a core.  If we use `/%z/` as part of the path, then I believe something outside coreadm will need to create that directory.  In previous use cases, was there another subsystem that created the %z directory, or did I just miss the coreadm option that would create on demand?

davepacheco:
I believe that in the case being referenced (Joyent's SmartOS), the path was really `/zones/%z/cores/core.%f.%p`, and yes, that directory was created by the machinery that created the zone (`vmadm(1M)`).

leftwo:
> I believe that in the case being referenced (Joyent's SmartOS), the path was really `/zones/%z/cores/core.%f.%p`, and yes, that directory was created by the machinery that created the zone (`vmadm(1M)`).

https://www.illumos.org/issues/2123
That vmadm?

jclulow:
If we need to create a cores directory we can do that in the brand code.  It has hooks for installing and for booting and so on.

leftwo:
If we do want a `.../%z/...` in the path, then something will need to both create that directory on zone creation, and remove it (if empty) when the zone goes away.  Otherwise we are left with a record of every zone ever created.

davepacheco:
[...]

PR #2088, opened 2022-12-21
set GOTRACEBACK=crash when running cockroach

davepacheco:
By default, when Go encounters a fatal runtime error, it dumps some state (some Goroutine stack traces) and exits the process with a non-zero status code.  It'd be nice to get core files for these fatal failures.  Such core files proved invaluable for debugging #1146 and friends.  Go provides a mechanism to dump core on fatal failures by setting [`GOTRACEBACK=crash`](https://pkg.go.dev/runtime#hdr-Environment_Variables) in the environment.  This causes it to dump its state like normal and then raise `SIGABRT`.   This is how I generated most of the core files that I used to debug those issues.

Getting a core file also requires having configured `coreadm` properly and potentially updating `ulimit -c`.  These are properties of the developer's environment and I think we should not take ownership of these.  In production, I assume we have already configured them or will do so.

There are two places where this needs to be changed:

- in omicron-test-utils: this covers both the test suite and the `omicron-dev db-run` developer tool
- in the SMF manifest: this covers production

This change:

- updates the dev/db CockroachDB starter infrastructure:
    - add support for setting environment variables, preserving the current default behavior of propagating the current environment into the `cockroach` process
    - uses that to set GOTRACEBACK=crash by default (this can be overridden)
    - print out the environment variables used
    - tests the above on illumos systems only (because testing this is platform-specific 
-- I'm using `pargs -e PID` to get the environment of the child process)
    - separately adds GOTRACEBACK=crash to the one-off invocation that we use to run `cockroach version`
- updates the SMF manifest to set GOTRACEBACK=crash on deployed systems

### DIFF
--- a/test-utils/src/dev/db.rs
+++ b/test-utils/src/dev/db.rs
@@ -9,6 +9,7 @@ use anyhow::anyhow;
 use anyhow::bail;
 use anyhow::Context;
 use omicron_common::postgres_config::PostgresConfigWithUrl;
+use std::collections::BTreeMap;
 use std::ffi::{OsStr, OsString};
 use std::fmt;
 use std::ops::Deref;
@@ -68,6 +69,8 @@ pub struct CockroachStarterBuilder {
     store_dir: Option<PathBuf>,
     /// optional value for the listening port
     listen_port: u16,
+    /// environment variables, mirrored here for reporting
+    env: BTreeMap<String, String>,
     /// command-line arguments, mirrored here for reporting
     args: Vec<String>,
     /// describes the command line that we're going to execute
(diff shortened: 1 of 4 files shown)


smklein:
> Getting a core file also requires having configured coreadm properly and potentially updating ulimit -c. These are properties of the developer's environment and I think we should not take ownership of these. In production, I assume we have already configured them or will do so.

Should we priority boost https://github.com/oxidecomputer/omicron/issues/1597 ? Is this ultimately the sled agent's job?

davepacheco:
> > Getting a core file also requires having configured coreadm properly and potentially updating ulimit -c. These are properties of the developer's environment and I think we should not take ownership of these. In production, I assume we have already configured them or will do so.
> 
> Should we priority boost #1597 ?

I do think this is critical for MVP.  It's valuable as soon as it can be done.  I'm not sure whether it's urgent enough to displace other work.

> Is this ultimately the sled agent's job?

Will move this to #1597.

davepacheco:
Also just to be clear this PR is not in any way blocked on #1597.

test-utils/src/dev/db.rs
```diff
@@ -1101,11 +1152,16 @@ mod test {
...
+        let mut builder = new_builder().store_dir(&data_dir);
+        builder.env("GOTRACEBACK", "bogus").env("OMICRON_DUMMY", "dummy");
+        let starter = builder.build().unwrap();
 
         // This common function will verify that the entire temporary directory
         // is cleaned up.  We do not need to check that again here.
-        test_setup_database(starter, &data_dir, false).await;
+        let mut env_overrides = BTreeMap::new();
+        env_overrides.insert("GOTRACEBACK", "bogus");
+        env_overrides.insert("OMICRON_DUMMY", "dummy");
```
smklein, line 1163: Nitpick, can we declare this first, and use it to call `builder.env`? Just to "set the values once"

smklein (approved):

test-utils/src/dev/db.rs
davepacheco: Good call.

davepacheco: Fixed in e71b56b2.

PR #2089, opened 2022-12-21
Integrate SoftNPU as virtual hardware

### LINKED ISSUE #1465: Initial integration with Dendrite
This issue tracks a first-pass at integrating Omicron with Dendrite. The actual services needed are already packaged as part of `omicron-package`, which includes a stubbed no-op implementation of `dpd`, the data-plane daemon responsible for shuttling traffic between the sleds (e.g., for guests) and the broader network. `dpd`'s Dropshot API needs to used during initial programming of the switch ports (as part of bootstrapping) and also keeping track of the mapping between external IP addresses (again, mostly for guests) to the sleds in the rack those IP addresses actually "reside" on.

- [ ] https://github.com/oxidecomputer/omicron/issues/1463
- [ ] https://github.com/oxidecomputer/omicron/issues/1464

internet-diglett:
- [x] Automate creation of `softnpu` zone
- [x] Automate startup up `softnpu`
- [x] Automate configuration of `softnpu`
  - [x]  <s> Configure link addresses and gateway parameters via `scadm` / `softnpuadm` upon `softnpu` startup </s>
  - [x] Enable communication between `dendrite` and `softnpu`
    - [x] Have switch zone mount `softnpu` socket 
    - [x] package `dendrite` with `softnpu` feature
    - [x] When `dendrite` asic type is `Softnpu`, have `dendrite` configure switch via `softnpu` socketfile
  - [x] Configure arp and nat entries via `dendrite` upon guest VM creation
    - [x] Add logic to `dendrite-softnpu` for programming nat entries
  - [x] cleanup nat entries upon guest VM deletion
 
Closes #1465 
Closes https://github.com/oxidecomputer/opte/issues/236


tools/scrimlet/softnpu-init.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+
+set -e
+set -x
+
+# Gateway ip is automatically configured based on the default route on your development machine
+# Can be overridden by setting GATEWAY_IP
+GATEWAY_IP=${GATEWAY_IP:=$(netstat -rn | grep default | awk -F ' ' '{print $2}')}
```
rcgoodfellow, line 8: If we're assuming IPv4 here, which I think we are given the `arp` call below, the `netstat` call needs to be changed to `netstat -rn -f inet`. Otherwise you'll get two entries and `GATEWAY_IP` will contain two addresses on two lines.

On my machine

```
ry@masaka:~$ netstat -rn | grep default | awk -F ' ' '{print $2}'
192.168.1.1
fe80::92ec:77ff:fe2e:7027
```

versus

```
ry@masaka:~$ netstat -rn -f inet | grep default | awk -F ' ' '{print $2}'
192.168.1.1
```

internet-diglett: Awesome, will add this change.

docs/boundary-services-a-to-z.adoc
```diff
@@ -0,0 +1,179 @@
...
+only way I know how to do this right now is by looking at the web console. Go to
+the instances page and navigate to the network tab. You'll see an External IP
+field in the table. Let's assume that is `10.100.0.6` for this example.
+
+Now we need to go tell boundary services about this information.
+
+This can be accomplished via the following `swadm` command:
+
+----
+./out/softnpu/swadm -h "[fd00:1122:3344:101::2]" nat add \
```
rcgoodfellow, line 150: The values here are not consistent with the `opteadm` output above.

rcgoodfellow:
Just dropping a note that I was able to successfully test this end-to-end with the CI builds brought in via the setup-virtual-hardware scripts. From zero to instances talking to the Internet.

rcgoodfellow:
Looks like destroy virtual hardware needs a bit of love.

```
ry@masaka:~/src/omicron$ pfexec ./tools/destroy_virtual_hardware.sh
+++ dirname ./tools/destroy_virtual_hardware.sh
++ cd ./tools
++ pwd
+ SOURCE_DIR=/home/ry/src/omicron/tools
+ cd /home/ry/src/omicron/tools/..
+ OMICRON_TOP=/home/ry/src/omicron
+ MARKER=/etc/opt/oxide/NO_INSTALL
+ [[ -f /etc/opt/oxide/NO_INSTALL ]]
++ id -u
+ [[ 0 -ne 0 ]]
+ verify_omicron_uninstalled
+ svcs svc:/system/illumos/sled-agent:default
svcs: Pattern 'svc:/system/illumos/sled-agent:default' doesn't match any instances
+ unload_xde_driver
++ modinfo
++ grep xde
++ cut -d ' ' -f 1
+ local ID=239
+ [[ -n 239 ]]
+ modunload -i 239
+ success 'Verified the xde kernel driver is unloaded'
+ set +x
Verified the xde kernel driver is unloaded
+ try_remove_vnics
+ try_remove_address lo0/underlay
+ local ADDRESS=lo0/underlay
++ ipadm show-addr -p -o addr lo0/underlay
ipadm: Address object not found+ [[ -n '' ]]
+ success 'Verified address lo0/underlay does not exist'
+ set +x
Verified address lo0/underlay does not exist
+ try_remove_interface sc0_1
+ local IFACE=sc0_1
++ ipadm show-if -p -o IFNAME sc0_1
ipadm: Could not get interface(s): Interface does not exist+ [[ -n '' ]]
+ success 'Verified IP interface sc0_1 does not exist'
+ set +x
Verified IP interface sc0_1 does not exist
+ try_remove_vnic sc0_1
+ local LINK=sc0_1
++ dladm show-vnic -p -o LINK sc0_1
+ [[ -n sc0_1 ]]
+ dladm delete-vnic sc0_1
dladm: vnic deletion failed: link busy
+ warn 'Failed to delete VNIC link sc0_1'
+ set +x
Failed to delete VNIC link sc0_1
+ success 'Verified VNIC link sc0_1 does not exist'
+ set +x
Verified VNIC link sc0_1 does not exist
+ INDICES=("0" "1")
+ for I in "${INDICES[@]}"
+ try_remove_simnet net0
+ local LINK=net0
++ dladm show-simnet -p -o LINK net0
+ [[ -n net0 ]]
+ dladm delete-simnet -t net0
dladm: simnet deletion failed: link busy
+ warn 'Failed to delete simnet link net0'
+ set +x
Failed to delete simnet link net0
+ success 'Verified simnet link net0 does not exist'
+ set +x
Verified simnet link net0 does not exist
+ try_remove_simnet sc0_0
+ local LINK=sc0_0
++ dladm show-simnet -p -o LINK sc0_0
+ [[ -n sc0_0 ]]
+ dladm delete-simnet -t sc0_0
dladm: simnet deletion failed: link busy
+ warn 'Failed to delete simnet link sc0_0'
+ set +x
Failed to delete simnet link sc0_0
+ success 'Verified simnet link sc0_0 does not exist'
+ set +x
Verified simnet link sc0_0 does not exist
+ try_remove_simnet sr0_0
+ local LINK=sr0_0
++ dladm show-simnet -p -o LINK sr0_0
+ [[ -n sr0_0 ]]
+ dladm delete-simnet -t sr0_0
dladm: simnet deletion failed: link busy
+ warn 'Failed to delete simnet link sr0_0'
+ set +x
Failed to delete simnet link sr0_0
+ success 'Verified simnet link sr0_0 does not exist'
+ set +x
Verified simnet link sr0_0 does not exist
+ try_remove_simnet scr0_0
+ local LINK=scr0_0
++ dladm show-simnet -p -o LINK scr0_0
+ [[ -n scr0_0 ]]
+ dladm delete-simnet -t scr0_0
+ success 'Verified simnet link scr0_0 does not exist'
+ set +x
Verified simnet link scr0_0 does not exist
+ for I in "${INDICES[@]}"
+ try_remove_simnet net1
+ local LINK=net1
++ dladm show-simnet -p -o LINK net1
+ [[ -n net1 ]]
+ dladm delete-simnet -t net1
dladm: simnet deletion failed: link busy
+ warn 'Failed to delete simnet link net1'
+ set +x
Failed to delete simnet link net1
+ success 'Verified simnet link net1 does not exist'
+ set +x
Verified simnet link net1 does not exist
+ try_remove_simnet sc1_0
+ local LINK=sc1_0
++ dladm show-simnet -p -o LINK sc1_0
+ [[ -n sc1_0 ]]
+ dladm delete-simnet -t sc1_0
+ success 'Verified simnet link sc1_0 does not exist'
+ set +x
Verified simnet link sc1_0 does not exist
+ try_remove_simnet sr0_1
+ local LINK=sr0_1
++ dladm show-simnet -p -o LINK sr0_1
+ [[ -n sr0_1 ]]
+ dladm delete-simnet -t sr0_1
dladm: simnet deletion failed: link busy
+ warn 'Failed to delete simnet link sr0_1'
+ set +x
Failed to delete simnet link sr0_1
+ success 'Verified simnet link sr0_1 does not exist'
+ set +x
Verified simnet link sr0_1 does not exist
+ try_remove_simnet scr0_1
+ local LINK=scr0_1
++ dladm show-simnet -p -o LINK scr0_1
+ [[ -n scr0_1 ]]
+ dladm delete-simnet -t scr0_1
+ success 'Verified simnet link scr0_1 does not exist'
+ set +x
Verified simnet link scr0_1 does not exist
+ try_destroy_zpools
+ readarray -t ZPOOLS
++ zfs list -d 0 -o name
++ grep '^oxp_'
+ for ZPOOL in "${ZPOOLS[@]}"
+ VDEV_FILE=/home/ry/src/omicron/oxp_d462a7f7-b628-40fe-80ff-4e4189e2d62b.vdev
+ zfs destroy -r oxp_d462a7f7-b628-40fe-80ff-4e4189e2d62b
+ zfs unmount oxp_d462a7f7-b628-40fe-80ff-4e4189e2d62b
+ zpool destroy oxp_d462a7f7-b628-40fe-80ff-4e4189e2d62b
+ rm -f /home/ry/src/omicron/oxp_d462a7f7-b628-40fe-80ff-4e4189e2d62b.vdev
+ success 'Verified ZFS pool and vdev oxp_d462a7f7-b628-40fe-80ff-4e4189e2d62b does not exist'
+ set +x
Verified ZFS pool and vdev oxp_d462a7f7-b628-40fe-80ff-4e4189e2d62b does not exist
+ for ZPOOL in "${ZPOOLS[@]}"
+ VDEV_FILE=/home/ry/src/omicron/oxp_e4b4dc87-ab46-49fb-a4b4-d361ae214c03.vdev
+ zfs destroy -r oxp_e4b4dc87-ab46-49fb-a4b4-d361ae214c03
+ zfs unmount oxp_e4b4dc87-ab46-49fb-a4b4-d361ae214c03
+ zpool destroy oxp_e4b4dc87-ab46-49fb-a4b4-d361ae214c03
+ rm -f /home/ry/src/omicron/oxp_e4b4dc87-ab46-49fb-a4b4-d361ae214c03.vdev
+ success 'Verified ZFS pool and vdev oxp_e4b4dc87-ab46-49fb-a4b4-d361ae214c03 does not exist'
+ set +x
Verified ZFS pool and vdev oxp_e4b4dc87-ab46-49fb-a4b4-d361ae214c03 does not exist
+ for ZPOOL in "${ZPOOLS[@]}"
+ VDEV_FILE=/home/ry/src/omicron/oxp_f4b4dc87-ab46-49fb-a4b4-d361ae214c03.vdev
+ zfs destroy -r oxp_f4b4dc87-ab46-49fb-a4b4-d361ae214c03
+ zfs unmount oxp_f4b4dc87-ab46-49fb-a4b4-d361ae214c03
+ zpool destroy oxp_f4b4dc87-ab46-49fb-a4b4-d361ae214c03
+ rm -f /home/ry/src/omicron/oxp_f4b4dc87-ab46-49fb-a4b4-d361ae214c03.vdev
+ success 'Verified ZFS pool and vdev oxp_f4b4dc87-ab46-49fb-a4b4-d361ae214c03 does not exist'
+ set +x
Verified ZFS pool and vdev oxp_f4b4dc87-ab46-49fb-a4b4-d361ae214c03 does not exist
+ remove_softnpu_zone
+ zoneadm -z softnpu halt
+ zoneadm -z softnpu uninstall -F
INFO: omicron: uninstalling zone softnpu...
+ zonecfg -z softnpu delete -F
+ rm -rf /opt/softnpu/stuff
+ zfs destroy rpool/softnpu-zone
+ rm -rf /softnpu-zone
ry@masaka:~/src/omicron$ ipadm
igb0/dhcp         dhcp     ok           192.168.1.4/24
igb0/v6           addrconf ok           fe80::aa5e:45ff:fecd:bd94/10
igb0/v6           addrconf ok           2603:3024:b0b:3400::18ba/128
ry@masaka:~/src/omicron$ ipadm show-if
IFNAME     CLASS     STATE    CURRENT      PERSISTENT
lo0        VIRTUAL   ok       -m-v------46 ---
igb0       IP        ok       bm--------46 -46
net0       IP        down     bm--------46 ---
net1       IP        down     bm--------46 ---
ry@masaka:~/src/omicron$ dladm
igb0        phys      1500   up       --         --
net0        simnet    1600   up       --         sc0_0
sc0_0       simnet    1600   up       --         net0
sr0_0       simnet    9000   up       --         --
net1        simnet    1600   up       --         --
sr0_1       simnet    9000   up       --         --
sc0_1       vnic      1500   up       --         igb0
```

rcgoodfellow:
> Looks like destroy virtual hardware needs a bit of love.

This is now working as expected for me.

rcgoodfellow:
I sorted out the networking issues we were seeing today with softnpu in the following 

- https://github.com/oxidecomputer/p4/pull/19
- https://github.com/oxidecomputer/softnpu/pull/4

And pushed an Omicron commit you can cherry-pick into this branch with the updates.

- https://github.com/oxidecomputer/omicron/commit/428a035df49e38aaadad5978f3c0095e31e92302

rcgoodfellow:
Noting that we'll need to update to the current Dendrite API as we are now a few commits behind. This is causing saga failure for me due to API mismatch.

- https://github.com/oxidecomputer/dendrite/pull/87

rcgoodfellow:
I've put together a pile of pull requests to fix softnpu/p4 range keys.

- [ ] P4 https://github.com/oxidecomputer/p4/pull/20
- [ ] Softnpu https://github.com/oxidecomputer/softnpu/pull/5
- [ ] Sidecar-lite https://github.com/oxidecomputer/sidecar-lite/pull/8
- [ ] Dendrite https://github.com/oxidecomputer/dendrite/pull/104

These updates collectively solve the issue I was seeing with proxy arp replying to a much larger range than assigned. I think they may also address the NAT range issues @internet-diglett has been observing.

nexus/src/app/sagas/instance_create.rs
```diff
@@ -299,11 +306,245 @@ impl NexusSaga for SagaInstanceCreate {
...
     }
 }
 
+async fn sic_add_network_config(
+    sagactx: NexusActionContext,
+) -> Result<(), ActionError> {
+    let params = sagactx.saga_params::<Params>()?;
+    let opctx = OpContext::for_saga_action(&sagactx, &params.serialized_authn);
+    let osagactx = sagactx.user_data();
+    let dpd_client: Arc<dpd_client::Client> =
```
bnaecker, line 321: No need to clone, I think, the `dpd-client` methods we're calling take `&self`.

nexus/src/app/sagas/instance_create.rs
```diff
@@ -299,11 +306,245 @@ impl NexusSaga for SagaInstanceCreate {
...
+
+    let sled_ip_address = sled.address();
+
+    debug!(log, "fetching network interfaces");
+
+    // TODO: Correctness
+    // Currently we're only mapping external ip addresses to the first
+    // network interface we find.We'll need a more sophisticated
+    // approach as we expand our use cases.
+    let network_interface = datastore
```
bnaecker, line 352: Summary from chat. We need to fetch the primary interface. If there isn't one, we need to return `Ok(())` here.

nexus/src/app/sagas/instance_create.rs
```diff
@@ -299,11 +306,245 @@ impl NexusSaga for SagaInstanceCreate {
...
+        .first()
+        .ok_or_else(|| {
+            ActionError::action_failed(Error::internal_error(
+                "guest network interface not found",
+            ))
+        })?
+        .clone();
+
+    // Is there a cleaner way to get from sled_agent_types::MacAddr -> MacAddr6
+    // (and ultimately dpd_client::types::MacAddr)?
```
bnaecker, line 365: Not right now. I want to make an `oxide-net` crate that has a bunch of these types we've duplicated in a billions places. But until then, nope.

nexus/src/app/sagas/instance_create.rs
```diff
@@ -299,11 +306,245 @@ impl NexusSaga for SagaInstanceCreate {
...
+
+    let external_ips = &datastore
+        .instance_lookup_external_ips(&opctx, instance_id)
+        .await
+        .map_err(ActionError::action_failed)?;
+
+    // TODO: Correctness
+    // Do we provide a mechanism for binding specific external ip
+    // addresses to specific network interfaces? Currently we're mapping
+    // all external ip addresses to the first interface we find.
```
bnaecker, line 386: Summary from chat: Yes, all external IPs appear on the primary interface.

nexus/src/app/sagas/instance_create.rs
```diff
@@ -299,11 +306,245 @@ impl NexusSaga for SagaInstanceCreate {
...
+                    .nat_set_ipv6(
+                        &network.ip(),
+                        *entry.first_port,
+                        *entry.last_port,
+                        &nat_target,
+                    )
+                    .await
+            }
+        }
+        .map_err(|e| {
```
bnaecker, line 451: So I think the fact that each request can fail means these need to be separate saga actions. Look at the NIC and disk-creation for examples. We basically create N nodes, for N NICs. Each of those creates (or destroys) exactly one item. Otherwise, the action is not really idempotent.

nexus/src/app/sagas/instance_create.rs
```diff
@@ -299,11 +306,245 @@ impl NexusSaga for SagaInstanceCreate {
...
+            Err(_) => {
+                // TODO: correctness
+                // This needs a bit more logic since "not found" is an error,
+                // but in this case it is actually ok. We want to propogate
+                // other errors.
+            }
+        }
+    }
+
+    for entry in entries_to_remove {
```
bnaecker, line 521: Same thing here about idempotency. We need a different forward and undo action for _each_ external IP address.

nexus/src/app/sagas/instance_delete.rs
```diff
@@ -68,6 +75,82 @@ impl NexusSaga for SagaInstanceDelete {
...
+    debug!(log, "fetching external ip addresses");
+
+    let external_ips = &datastore
+        .instance_lookup_external_ips(&opctx, params.authz_instance.id())
+        .await
+        .map_err(ActionError::action_failed)?;
+
+    let mut entries_to_remove: Vec<ExternalIp> = vec![];
+
+    for entry in external_ips {
```
bnaecker, line 97: Same thing here about idempotency. We need a different saga action for each external IP I think.

bnaecker (commented):
Not at all a complete review. Just a few notes that came up during chat today, wanted to get them persisted.

nexus/src/app/sagas/instance_create.rs
internet-diglett: Thinking about this more, I'm not _certain_ this is the case. If any of the requests fail, the saga is supposed to unwind, right? Even if the node is re-attempted, any requests that were successful would not be re-sent because of the logic above this that checks for existing nat entries, so unless I'm missing something, I think this is actually idempotent.

nexus/src/app/sagas/instance_create.rs
internet-diglett: Do we actually need a separate saga node for _each_ ip, or do we need to make sure that we attempt to delete the nat entry for every existing ip _at least once_?

nexus/src/app/sagas/instance_create.rs
internet-diglett: Persisting our chat discussion here:

I had a misunderstanding of Steno's behavior. Since we don't run `undo` actions for failed nodes, the logic in `sic_remove_network_config()` would never be called in the event of a partial failure, which is why the overall behavior would not be idempotent.

nexus/src/app/sagas/instance_delete.rs
internet-diglett: Persisting convo from chat:

It doesn't seem that any of the other saga nodes in `instance_delete.rs` are using subsagas, so here we're going to ensure that we have attempted to nat entries for all external ips before checking errors.

docs/boundary-services-a-to-z.adoc
```diff
@@ -0,0 +1,183 @@
...
+The `softnpu` zone will be configured and launched during the `create_virtual_hardware.sh`
+script.
+
+== 2. Build and install the control plane.
+
+----
+./tools/create_self_signed_cert.sh
+
+cargo build --release --bin omicron-package
+./target/release/omicron-package -t switch_variant=softnpu package
```
rcgoodfellow, line 39: Looks like if we're overriding the default we now need to explicitly set the `image_type` option.

```suggestion
./target/release/omicron-package -t 'image_type=standard switch_variant=softnpu' package
```

docs/boundary-services-a-to-z.adoc
```diff
@@ -0,0 +1,183 @@
...
+pfexec ./target/release/omicron-package -t switch_variant=softnpu install
```
rcgoodfellow, line 40:
```suggestion
pfexec ./target/release/omicron-package -t 'image_type=standard switch_variant=softnpu` install
```

docs/boundary-services-a-to-z.adoc
internet-diglett: Looks like the latest package magic requires
```
./target/release/omicron-package -t <target_name> target create -i standard -m non-gimlet -s softnpu
```

I've updated the `a-to-z` doc accordingly.

installinator/src/hardware.rs
```diff
@@ -22,9 +22,10 @@ impl Hardware {
             .context("failed to detect whether host is a gimlet")?;
         ensure!(is_gimlet, "hardware scan only supported on gimlets");
 
-        let hardware = HardwareManager::new(log, None).map_err(|err| {
-            anyhow!("failed to create HardwareManager: {err}")
-        })?;
+        let hardware =
```
bnaecker, line 25: Is this change necessary? Not a huge deal, but checking I'm not missing anything.

nexus/src/app/mod.rs
```diff
@@ -262,6 +279,7 @@ impl Nexus {
             ),
             samael_max_issue_delay: std::sync::Mutex::new(None),
             resolver,
+            dpd_client: Arc::clone(&dpd_client),
```
bnaecker, line 282: Unnecessary clone, it looks like?

nexus/src/app/sagas/instance_create.rs
```diff
@@ -299,11 +311,255 @@ impl NexusSaga for SagaInstanceCreate {
...
+        })?
+        .to_owned();
+
+    debug!(log, "checking for existing nat mapping for {target_ip:#?}");
+
+    // TODO: currently if we have this environment variable set, we want to
+    // bypass all calls to DPD. This is mainly to facilitate some tests where
+    // we don't have dpd running. In the future we should probably have these
+    // testing environments running dpd-stub so that the full path can be tested.
+    if let Ok(_) = std::env::var("SKIP_ASIC_CONFIG") {
```
bnaecker, line 421: Could we also print a debug message here? I worry we might have this set, but not realize it, during development, and wonder why our IP addresses aren't being forward to Dendrite.

nexus/src/app/sagas/instance_create.rs
```diff
@@ -299,11 +311,255 @@ impl NexusSaga for SagaInstanceCreate {
...
+            if e.status() == Some(http::StatusCode::NOT_FOUND) {
+                debug!(log, "no nat entry found for: {target_ip:#?}");
+            } else {
+                return Err(ActionError::action_failed(Error::internal_error(
+                    &format!("failed to query dpd: {e}"),
+                )));
+            }
+        }
+    }
+
```
bnaecker, line 449: I'm trying to remember the last conversation we had about this, but I thought we decided that the error-handling here would be a bit different. We'd instead try to create the mapping, and ignore an "existing mapping" error, rather than checking if it exists and then creating it.

Is that still worth doing here?

nexus/src/app/sagas/instance_create.rs
```diff
@@ -299,11 +311,255 @@ impl NexusSaga for SagaInstanceCreate {
...
+            )))
+        })?
+        .to_owned();
+
+    // TODO: currently if we have this environment variable set, we want to
+    // bypass all calls to DPD. This is mainly to facilitate some tests where
+    // we don't have dpd running. In the future we should probably have these
+    // testing environments running dpd-stub so that the full path can be tested.
+    if let Ok(_) = std::env::var("SKIP_ASIC_CONFIG") {
+        return Ok(());
```
bnaecker, line 527: Same note here, might be worth a `debug!` message.

nexus/src/app/sagas/instance_delete.rs
```diff
@@ -67,6 +71,79 @@ impl NexusSaga for SagaInstanceDelete {
...
+        .instance_lookup_external_ips(&opctx, params.authz_instance.id())
+        .await
+        .map_err(ActionError::action_failed)?;
+
+    // TODO: currently if we have this environment variable set, we want to
+    // bypass all calls to DPD. This is mainly to facilitate some tests where
+    // we don't have dpd running. In the future we should probably have these
+    // testing environments running dpd-stub so that the full path can be tested.
+    if let Ok(_) = std::env::var("SKIP_ASIC_CONFIG") {
+        return Ok(());
```
bnaecker, line 99: Same note here about `debug!` message.

nexus/src/app/sagas/instance_delete.rs
```diff
@@ -67,6 +71,79 @@ impl NexusSaga for SagaInstanceDelete {
...
+                        ));
+                    errors.push(new_error);
+                }
+            }
+        }
+    }
+
+    if let Some(error) = errors.first() {
+        return Err(error.clone());
+    }
```
bnaecker, line 142: I think returning the first error is probably the best we can do at this point, but it might help to log all of them?

sled-agent/src/params.rs
```diff
@@ -216,7 +216,7 @@ pub struct Zpool {
 
 // The type of networking 'ASIC' the Dendrite service is expected to manage
 #[derive(
-    Clone, Debug, Deserialize, Serialize, JsonSchema, PartialEq, Eq, Copy, Hash,
+    Clone, Debug, Deserialize, Serialize, JsonSchema, PartialEq, Eq, Hash,
```
bnaecker, line 219: It seems like this is still `Copy`, but I might be missing something.

sled-agent/src/services.rs
```diff
@@ -217,6 +218,10 @@ enum SwitchZone {
         request: ServiceZoneRequest,
         // A background task which keeps looping until the zone is initialized
         worker: Option<Task>,
+        // Filesystems for the switch zone to mount
+        // Since Softnpu is currently managed via a UNIX socket, we need to
+        // pass those files in to the SwitchZone so Dendrite can mmanage Softnpu
```
bnaecker, line 223: Spelling nit: `mmanage`

sled-agent/src/services.rs
```diff
@@ -495,6 +500,7 @@ impl ServiceManager {
     async fn initialize_zone(
         &self,
         request: &ServiceZoneRequest,
+        filesystems: &Vec<zone::Fs>,
```
bnaecker, line 503: Style nit: this could take a slice instead of a vec, e.g., `&[zone::Fs]`.

sled-hardware/src/illumos/mod.rs
```diff
@@ -534,11 +535,22 @@ impl HardwareManager {
...
+                ScrimletMode::Disabled => HardwareView::new_stub_tofino(
+                    // active=
+                    false,
+                ),
+                ScrimletMode::Stub => HardwareView::new_stub_tofino(true),
+                // TODO: correctness
+                // I'm not sure whether or not we should be treating softnpu
+                // as a stub or treating it as a real ASIC here, so this
+                // might change.
+                ScrimletMode::Softnpu => HardwareView::new_stub_tofino(true),
```
bnaecker, line 550: This refers to how we look for the driver and devices files. This can't be a "real" Tofino, though we may want to add a variant to `TofinoView` for the SoftNPU-specific stuff.

sled-hardware/src/lib.rs
```diff
@@ -39,6 +40,20 @@ pub enum HardwareUpdate {
     DiskRemoved(UnparsedDisk),
 }
 
+/// Configuration for forcing a sled to run as a Scrimlet
+#[derive(Clone, Debug, Deserialize, Copy)]
+#[serde(rename_all = "snake_case")]
+pub enum ScrimletMode {
+    /// Force sled to run as a Gimlet
+    /// this is to preserve the old behavior of `scrimlet_override = false`,
+    /// but I haven't found where that logic has actually been leveraged...
```
bnaecker, line 49: Not sure, but this comment might be outdated. I don't see any hits for the string `scrimlet_override` in the latest main. Maybe you mean `"stub_scrimlet"`?

smf/sled-agent/non-gimlet/config.toml
```diff
@@ -31,7 +34,7 @@ zpools = [
 #
 # If empty, this will be equivalent to the first result from:
 # $ dladm show-phys -p -o LINK
-# data_link = "igb0"
+data_link = "igb0"
```
bnaecker, line 37: Should we leave this commented out in `main`?

tools/setup_path.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
bnaecker, line 1: I might make this optional in some way. It's a bit unsporting to automatically adjust folks' `PATH`, especially in a way that would result in many copies of the same directory if it's run multiple times.

bnaecker (commented):
_Tons_ of great work @internet-diglett! This has been a long slog, your persistence is admirable. I've got a few questions and small comments, but this generally seems pretty solid to me!

.github/buildomat/jobs/build-and-test-linux.sh
```diff
@@ -10,6 +10,9 @@
 #:	"!/var/tmp/omicron_tmp/rustc*",
 #: ]
 #:
+#: access_repos = [
+#:   "oxidecomputer/dendrite",
```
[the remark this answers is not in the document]
internet-diglett: Now that we're building from the OpenAPI spec, I don't believe this is being used for anything anymore.

docs/boundary-services-a-to-z.adoc
```diff
@@ -0,0 +1,184 @@
...
+Once your host has been populated with images, you can use the script at
+`tools/quickstart.sh` to quickly create a VM and set up the `proxy-arp`. Please
+be sure to edit the `ip_pool_start` and `ip_pool_end` variables to match your
+desired address ranges.
+
+== 5. Configuring scrimlet/sidecar
+
+A this point we have an instance up and running. At the time of writing there is
+not control plane driven boundary services automation so we're going to
+configure the scrimlet by hand.
```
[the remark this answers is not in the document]
internet-diglett: I think I missed removing this from the doc

.github/buildomat/jobs/build-and-test-linux.sh
```diff
@@ -62,6 +65,11 @@ ptime -m cargo build --locked --all-targets --verbose
 # We also don't use `--workspace` here because we're not prepared to run tests
 # from end-to-end-tests.
 #
+
+# TODO: we are bypassing calls to DPD during these tests for now. This env var
```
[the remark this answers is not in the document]
internet-diglett: That was actually one of my tasks for today. Will drop a link here once the issue is created.

installinator/src/hardware.rs
internet-diglett: There seems to have been a deviation of sorts early in the development of this PR / branch. I think the original branch I started working from used `Logger` as the parameter, meanwhile as `main` progressed, it might have been changed to `&Logger`.

illumos-utils/src/opte/illumos/mod.rs
```diff
@@ -132,36 +120,3 @@ pub fn initialize_xde_driver(
...
-        fs::write(xde_conf, &new)
-            .expect("Failed to modify xde configuration file");
-    }
-
-    // Ensure the driver picks up the updated configuration file, if it's been
-    // loaded previously without the workaround.
-    std::process::Command::new(crate::PFEXEC)
-        .args(&["update_drv", "xde"])
-        .output()
-        .expect("Failed to reload xde driver configuration file");
```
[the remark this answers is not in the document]
smklein: Ah, nevermind, I see this is removed in "create_virtual_hardware" below

tools/quickstart.sh
```diff
@@ -0,0 +1,87 @@
+#!/bin/bash
```
smklein, line 1: I'm kinda opposed to checking this script in - it's *very* environment-specific, may not make sense for all devs, but has a name which implies "you should run me first!"

I think this sorta thing makes sense for developers to have / use, but it's a bit of a landmine for someone to run without a lot of context about their environment

docs/boundary-services-a-to-z.adoc
```diff
@@ -0,0 +1,184 @@
...
+- The address range in the IP pool should be on a subnet in your local network that
+  can NAT out to the Internet.
+- Be sure to set up an external IP for the instance you create.
+- You will need to set up `proxy-arp` if your VM external IP addresses are on the
+  same L2 network as the router or other non-oxide hosts.
+
+Once your host has been populated with images, you can use the script at
+`tools/quickstart.sh` to quickly create a VM and set up the `proxy-arp`. Please
+be sure to edit the `ip_pool_start` and `ip_pool_end` variables to match your
+desired address ranges.
```
smklein, line 107: See my comment below about this script; IMO we should not check it in.

smklein (commented):
1. Thanks for taking this work on, really appreciate you doing it. Herculean effort, which rockets the effectiveness of the product forward.
2. I have a smattering of comments about minor changes before we check in, but when in doubt, pretend I added a "do what he said" comment under everything @bnaecker said.
3. Adding a stub implementation of dpd so `cargo test` can work without custom environment variables should be a pretty high priority for us to fix -- lmk if I can help get this work done!

smf/sled-agent/non-gimlet/config.toml
internet-diglett: Yep, that was a whoopsie.

sled-hardware/src/lib.rs
internet-diglett: Yeah, I meant `stub_scrimlet`.

sled-agent/src/params.rs
internet-diglett: Can you elaborate?

nexus/src/app/sagas/instance_create.rs
internet-diglett: I _think_ the reason we changed it in the various `delete` methods but not here was because we could match on `Some(http::StatusCode::NOT_FOUND)` in the deletion attempts but we didn't get something similar back during creation attempts?

common/src/nexus_config.rs
```diff
@@ -173,6 +173,12 @@ pub struct TimeseriesDbConfig {
     pub address: Option<SocketAddr>,
 }
 
+/// Configuration for the dataplane daemon.
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+pub struct DpdConfig {
+    pub address: SocketAddr,
```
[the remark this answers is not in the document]
rcgoodfellow: On a real system there will be a `dpd` instance per scrimlet, each with a unique underlay address. `DpdConfig` is currently a single-valued member `dpd_api` in `PackageConfig`. This works for single-sidecar setups, but we'll need to evolve toward being able to handle multiple `dpd` instances.

I've [noted this](https://github.com/oxidecomputer/omicron/blob/26a408522894619526037a4dfeb94a49ecd6b916/nexus/src/app/sagas/port_settings_apply.rs#L151-L154) in the external networking plumbing I'm currently working on. I'd be happy to take this issue on in that work if we think that makes sense. There is additional infrastructure in the database in that branch that helps to identify the sidecar of interest through `rack_id` and `switch_location` when operations come through the API that require `dpd` interaction.

.github/buildomat/jobs/build-and-test-linux.sh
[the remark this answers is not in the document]
internet-diglett: https://github.com/oxidecomputer/omicron/issues/2629

nexus/src/app/sagas/instance_create.rs
```diff
@@ -299,11 +311,255 @@ impl NexusSaga for SagaInstanceCreate {
...
+        .ok_or_else(|| {
+            ActionError::action_failed(Error::internal_error(&format!(
+                "failed to find external ip address at index: {which}"
+            )))
+        })?
+        .to_owned();
+
+    debug!(log, "checking for existing nat mapping for {target_ip:#?}");
+
+    // TODO: currently if we have this environment variable set, we want to
```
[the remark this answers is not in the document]
internet-diglett: https://github.com/oxidecomputer/omicron/issues/2629

.cargo/config
```diff
@@ -15,3 +15,7 @@ rustdocflags = "--document-private-items"
 rustflags = [
     "-C", "link-arg=-Wl,-znocompstrtab"
 ]
+
+# We need to enable this to access private repos
+[net]
+git-fetch-with-cli = true
```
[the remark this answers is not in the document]
internet-diglett: I don't _think_ so. Removing it and seeing what happens in CI.

tools/setup_path.sh
internet-diglett: I've removed the reference of this from the `how-to-run` doc

nexus/src/app/sagas/instance_delete.rs
internet-diglett: Added logging above.

nexus/src/app/sagas/instance_create.rs
bnaecker: Ugh, right. See https://github.com/oxidecomputer/dendrite/issues/152

tools/setup_path.sh
bnaecker: Oh good catch. I read that in `how-to-run.adoc` and transposed it into a script. Sorry about that!

tools/quickstart.sh
bnaecker: I think an alternative would be to include the script or substantial portions of it in the documentation, such as `how-to-run.adoc`.

bnaecker (approved):
Thanks for addressing all my questions, @internet-diglett! This all looks OK to me!

sled-agent/src/params.rs
bnaecker: There's a diff here, removing the `Copy` trait. This enum sure looks like it's `Copy` to me, so I'm wondering if we want to put it back. It looks like maybe it was historical -- we added data to one of the enum variants that wasn't `Copy`; had to remove `Copy` from this enum; and then removed the data, but didn't put `Copy` back.

This is all tiny, and I'm just curious.

PR #2090, opened 2022-12-22
[nexus] Make snapshot deletion a saga

smklein:
Part of https://github.com/oxidecomputer/omicron/issues/2059


nexus/src/app/sagas/snapshot_delete.rs
```diff
@@ -0,0 +1,113 @@
...
+            serde_json::to_value(&volume_delete_params).map_err(|e| {
+                super::SagaInitError::SerializeError(
+                    String::from("volume_id"),
+                    e,
+                )
+            })?,
+        ));
+
+        let volume_delete_params = sagas::volume_delete::Params {
+            volume_id: params.snapshot.destination_volume_id.unwrap(),
```
smklein, line 63: Note: Should not be unwrapping, if this can be non-optional

leftwo: I think after PR #2092 it won't be.

nexus/src/app/disk.rs
```diff
@@ -496,8 +496,7 @@ impl super::Nexus {
         // (on-disk snapshots, running read-only downstairs) because disks
         // *could* still be using them (if the snapshot has not yet been turned
         // into a regular crucible volume). It will involve some sort of
-        // reference counting for volumes, and probably means this needs to
-        // instead be a saga.
+        // reference counting for volumes.
```
leftwo, line 499: This is now a saga, but the comment here seems to suggest becoming a saga would
solve some of the `TODO-correctness` issues above.

However, I'm not sure if this saga does solve those issues.

leftwo (approved):
Looks good,
Just a question really about if a comment is still relevant.

nexus/src/app/sagas/snapshot_delete.rs
smklein: Removed in https://github.com/oxidecomputer/omicron/pull/2090/commits/3f37a48aa32ef27d1531b832637850078320b503

nexus/src/app/disk.rs
smklein: Yeah, I removed the part about saga-ification, but left the TODO for refcounting. I'm not changing anything with respect to refcounting, just making sure that it can maintain crash correctness.

I am not confident enough about the current refcounting scheme to remove the rest of the TODO, tbh.

PR #2092, opened 2022-12-22
Make snapshot destination_volume_id not an Option

leftwo:


nexus/db-model/src/schema.rs
```diff
@@ -77,7 +77,7 @@ table! {
         disk_id -> Uuid,
         volume_id -> Uuid,
 
-        destination_volume_id -> Nullable<Uuid>,
+        destination_volume_id -> Uuid,
```
smklein, line 80: We also need to update https://github.com/oxidecomputer/omicron/blob/cdf3207e67241afb4dfc5af17c77a28d75003307/common/src/sql/dbinit.sql#L795 to `NOT NULL`

smklein (approved):

PR #2093, opened 2022-12-23
[nexus] Make snapshot-create saga unwind-safe

smklein:
- Makes snapshot creation unwind-safe
- Adds tests to validate that unwinding the saga does not leave detritus

Part of https://github.com/oxidecomputer/omicron/issues/2052


nexus/src/app/sagas/disk_create.rs
```diff
@@ -581,7 +581,7 @@ fn randomize_volume_construction_request_ids(
 }
 
 #[cfg(test)]
-mod test {
+pub(crate) mod test {
```
smklein, line 584: I'm mostly modifying this so I can re-use some of the test functions which validates "disk creation doesn't leave detritus".

Most of these checks are identical for snapshot creation.

nexus/src/app/sagas/snapshot_create.rs
```diff
@@ -1062,4 +1222,236 @@ mod test {
...
+        let mut dag = create_saga_dag::<SagaSnapshotCreate>(params).unwrap();
+
+        // The saga's input parameters include a disk UUID, which makes sense,
+        // since the snapshot is created from a disk.
+        //
+        // Unfortunately, for our idempotency checks, checking for a "clean
+        // slate" gets more expensive when we need to compare region allocations
+        // between the disk and the snapshot. If we can undo the snapshot
+        // provisioning AND delete the disk together, these checks are much
+        // simpler to write.
```
smklein, line 1404: This detail matters - basically, it's easy to check "nothing exists in the DB", and harder to check "the state which was provisioned for a disk exists and matches, but nothing else was added".

Unwinding the snapshot *AND* deleting the disk gets us back to the simpler "nothing should exist" state.

nexus/src/app/sagas/snapshot_create.rs
```diff
@@ -649,6 +740,60 @@ async fn ssc_start_running_snapshot(
...
+            // ssc_create_volume_record_undo).
+            //
+            // To cope, we treat "running snapshot not found" as "Ok", since it
+            // may just be the result of the volume deletion steps completing.
+            .or_else(|err| match err {
+                ErrorResponse(r) if r.status() == StatusCode::NOT_FOUND => {
+                    Ok(())
+                }
+                _ => Err(err),
+            })?;
```
smklein, line 788: I found this a little quirky; figured I'd call it out.

- If we *haven't* run the later "volume creation" undo step, we need to cleanup this running snapshot.
- If we *have* run the later "volume creation" undo step, it's already gone

leftwo (approved):

Issue #2094, opened 2022-12-26
Tracking issue for saga idempotency safety


smklein:
This issue tracks testing that invoking all actions / undo actions within a saga "twice" is safe, and results in the same output as being called once.

Related to https://github.com/oxidecomputer/omicron/issues/2052, https://github.com/oxidecomputer/omicron/issues/1799

smklein:
Blocked on https://github.com/oxidecomputer/steno/pull/88

bnaecker:
As part of https://github.com/oxidecomputer/omicron/pull/10144, we noted that during VPC creation the UUIDs for Internet Gateways are generated inside the saga node that creates them. To make the VPC create saga completely idemponent, these should be generated at the start of the saga and looked up later. The database query that actually inserts them should have an `ON CONFLICT` clause or otherwise detect this case and handle it gracefully too.

PR #2095, opened 2022-12-28
[nexus] Make instance creation actions/undo actions idempotent

smklein:
- Makes use of https://github.com/oxidecomputer/steno/pull/88 to test idempotency of instance creation actions + undo actions
- Fixes all the ways in which it was *not* idempotent, including...
  - (Action) Inserting the same network interface failed with a VPC subnet error, instead of an "already exists" error.
  - (Action) Inserting the instance record twice fell victim to https://github.com/oxidecomputer/omicron/issues/1168
  - (Undo) Deleting the instance record failed to execute twice, as it could not find the instance record for the second invocation.

Part of https://github.com/oxidecomputer/omicron/issues/2094


nexus/src/app/sagas/instance_create.rs
```diff
@@ -1004,7 +1023,7 @@ mod test {
                         name: DISK_NAME.parse().unwrap(),
                     },
                 )],
-                start: true,
+                start: false,
```
smklein, line 1026: I admittedly did flip this bit - otherwise, I need to poke the instance and update state to "stopped" before I can delete it anyway in the tests below.

nexus/src/app/sagas/instance_create.rs
```diff
@@ -1144,17 +1178,136 @@ mod test {
...
+        let nexus = &cptestctx.server.apictx.nexus;
+        let project_id = create_org_project_and_disk(&client).await;
+
+        // Build the saga DAG with the provided test parameters
+        let opctx = test_opctx(&cptestctx);
+
+        let params = new_test_params(&opctx, project_id);
+        let dag = create_saga_dag::<SagaInstanceCreate>(params).unwrap();
+
+        // The "undo_node" should always be immediately preceding the
```
andrewjstone, line 1202: While perfectly reasonable with our current implementation, it makes me a tad uncomfortable that we are expecting a steno internal Dag structure to look a certain way for this test.

I'm not really sure the best way to handle this, except for perhaps first classing this type of injection in steno. I"m curious to hear @davepacheco's thoughts when he gets back, although I don't think we should hold up the PR for this.

andrewjstone (approved):

nexus/src/app/sagas/instance_create.rs
smklein: We aren't expecting anything of the internal DAG structure, other than "it's ordered". "Which node is which" is entirely up to the test -- the *test* is making the call "whenever we fail, the node preceding the failure node should have an undo action called twice to test idempotency".

We can arbitrarily change the saga DAG, and this test should be able to remain unchanged.

All nodes we're iterating over are either actions, constants, or subsagas, as discussed in https://github.com/oxidecomputer/steno/pull/67 .

> I'm not really sure the best way to handle this, except for perhaps first classing this type of injection in steno.

What would this mean to you? I thought the APIs I added to steno to [iterate over the dag](https://github.com/oxidecomputer/steno/commit/666141615d18dde53f0d00bc4c4f9b474af006e7), [expose indices which could be targets for injection](https://github.com/oxidecomputer/steno/commit/4cebb996217453b79896b53c9c2026007f2e69e8), and add the ability to [inject repetitions](https://github.com/oxidecomputer/steno/commit/53129dd9c011d1f9d2fd781676243198e1e297bc) alongside the already-existing "error injection" counted as first-class support?

andrewjstone: > We aren't expecting anything of the internal DAG structure, other than "it's ordered". "Which node is which" is entirely up to the test -- the test is making the call "whenever we fail, the node preceding the failure node should have an undo action called twice to test idempotency".

> We can arbitrarily change the saga DAG, and this test should be able to remain unchanged.

Yup, thinking about this more, you are 100% correct.

> What would this mean to you? I thought the APIs I added to steno to [iterate over the dag](https://github.com/oxidecomputer/steno/commit/666141615d18dde53f0d00bc4c4f9b474af006e7), [expose indices which could be targets for injection](https://github.com/oxidecomputer/steno/commit/4cebb996217453b79896b53c9c2026007f2e69e8), and add the ability to [inject repetitions](https://github.com/oxidecomputer/steno/commit/53129dd9c011d1f9d2fd781676243198e1e297bc) alongside the already-existing "error injection" counted as first-class support?

I was overthinking this. What you did makes sense. My initial thought was something unnecessarily complicated. We'd inject errors then iterate again, reading the annotated error nodes and getting the prior nodes affected by the errors. However, as you point out that's unnecessary, since we know based on the dag that the node immediately prior to the error node will be undone.

Sorry for unnecessarily getting your hackles up.

smklein: No worries! Just want to make sure I wasn't missing a cleaner solution - probing at what's going on is always appreciated

PR #2096, opened 2022-12-28
[nexus] Make instance deletion actions idempotent

smklein:
- Makes use of https://github.com/oxidecomputer/steno/pull/88 to test idempotency of instance deletion actions
- Fixes the way in which it was not idempotent: deleting the instance record

Part of #2094


andrewjstone (approved):
Looks great!

PR #2097, opened 2022-12-28
[nexus] Make disk creation saga idempotent

smklein:
- Makes use of https://github.com/oxidecomputer/steno/pull/88 to test idempotency of disk creation actions
- Fixes the way in which it was not idempotent: creating the disk record

Part of #2094


leftwo (approved):
Yeah this looks good.

I guess this means we are, as the title suggests, making it idempotent.
I go back and forth about if it is an error when you try to create anything
and what you are trying to create is already there.  I think in this case
it won't be a problem, so yeah, good to go.

PR #2098, opened 2022-12-28
[nexus] Make disk deletion saga idempotent

smklein:
- Makes use of https://github.com/oxidecomputer/steno/pull/88 to test idempotency of disk deletion actions

Part of #2094


nexus/src/app/sagas/disk_create.rs
```diff
@@ -617,21 +617,25 @@ pub(crate) mod test {
         project.identity.id
     }
 
+    pub fn new_disk_create_params() -> params::DiskCreate {
+        params::DiskCreate {
+            identity: IdentityMetadataCreateParams {
+                name: DISK_NAME.parse().expect("Invalid disk name"),
+                description: "My disk".to_string(),
+            },
+            disk_source: params::DiskSource::Blank {
```
leftwo, line 626: This could be expanded to create more `DiskSource`s.  There are other code paths that
we don't get coverage on for the saga when we have a Snapshot or a Global Image.
I created an issue for this: https://github.com/oxidecomputer/omicron/issues/2115

This PR can go ahead without it, as it's improving on what we have now.

leftwo (approved):

Issue #1959, opened 2022-11-17
Collect and transfer x509 cert information into Nexus


smklein:

smklein:
Related: https://github.com/oxidecomputer/dropshot/pull/504 and https://github.com/oxidecomputer/dropshot/pull/502 should make this slightly easier

Issue #1497, opened 2022-07-26
some non-unique indexes in dbinit.sql are fishy


davepacheco:
We have a bunch of non-unique indexes in the database.  I'm a little skeptical of these because:

- They're not good for looking up a specific item in the table because the lookup would be linear in the number of duplicates in the index.
- They're not good for pagination for the same reason.  At best, if a page boundary lands in a run of duplicates, then the next page request will have to do a linear scan of the duplicated items because it has no way to quickly get to the one it needs to start with.  At worst, the database will think it can't use the index at all because the items in it do not necessarily appear to satisfy the ORDER BY that's used for pagination.
- In most cases, you can make these indexes good for both lookup and pagination by adding the primary key (usually `id`) as the last column in the index.  (I believe that CockroachDB implicitly includes the primary key already and it's possible it uses this to address the above problems?  But (a) if we're relying on that, we should verify it, and (b) if so, then we probably still want to be consistent.  Right now, a bunch of our indexes explicitly include the id for this purpose and a bunch of indexes don't.) 

There's at least one good use case for non-unique indexes, which is when you expect to fetch a bunch of rows based on some property, you expect there could be a large number of these, and you don't care which ones you get.  Examples of this include the index on `console_session.time_created` that's intended to be used to expire old sessions.  This is fine because we'll wind up querying for sessions older than, say, a day ago, and we don't care which ones we find.  (In fact, we'll probably keep querying and expiring everything we find until we run out.)

We should take a pass through the non-unique indexes and audit them.

Examples that look intended for pagination or lookup but that I think won't work for the above reasons:

* `sled.rack_id` (needs sled's id too and then it'll be unique)
* `service.sled_id` (needs service's id too and then it'll be unique)
* `region.volume_id` (needs region's id too and then it'll be unique)
* `disk.attach_instance_id` (needs disk's id too and then it'll be unique)

These I think are wrong only because they have the fields in the wrong order:

* `identity_provider.(id, silo_id)` -- but I think this probably needs to be `(silo_id, id)` because otherwise it can't be used to quickly list IdPs in a Silo
* `identity_provider.(name, silo_id)` -- but I think this probably needs to be `(silo_id, name)` for the same reason
* `saml_identity_provider.(id, silo_id)` -- same

These I think are unique, just not constrained to be:

* `metric_producer.(oximeter_id, id)`

Ones I'm not sure about without more investigation:

* `Dataset.size_used`
* `ip_pool_range.project_id`
[...]

PR #2099, opened 2022-12-29
Store x.509 certs in CRDB, auto-load on boot, API for management

smklein:
- Removes x.509 cert from Nexus package
- Transfers certificate to Nexus as a part of the "RSS handoff", if they are available
- Nexus auto-loads certificate from CRDB when booting
- Adds `/system/certificates` API to Nexus

TODO (remaining items postponed until after this PR):
- [x] Fix existing tests
- [x] When x.509 certs are updated via the API, should we refresh them?
  - [ ] Should we refresh neighboring Nexus servers too?
- [x] Tests for the API?
- [x] ... if we *don't* require Nexus to boot by scraping a CRDB cert, we could remove it from the package system right now?
- [x] Certificate validation on input?
  - [ ] ... but should we add more?
- [x] Adding a `/system/...` endpoint this way is surprisingly difficult. There are a ton of steps to keep track of... maybe I could document or improve this process?: https://github.com/oxidecomputer/omicron/pull/2101

Part of https://github.com/oxidecomputer/omicron/issues/1959


nexus/src/lib.rs
```diff
@@ -158,34 +174,15 @@ impl Server {
...
-            .map(|r| r.unwrap_err())
-            .collect::<Vec<String>>();
-
-        if errors.len() > 0 {
-            let msg = format!("errors shutting down: ({})", errors.join(", "));
-            Err(msg)
-        } else {
-            Ok(())
-        }
+        self.apictx.nexus.close_servers().await
```
jgallagher, line 177: Tiny nit - it looks like this and `close()` are the same (except that `close()` unwraps and this doesn't). Would it make sense for `close()` to call this method so if it changes in the future only this one needs updating?

nexus/src/app/mod.rs
```diff
@@ -245,6 +280,55 @@ impl Nexus {
...
+        &self,
+        external_servers: ExternalServers,
+        internal_server: DropshotServer,
+    ) {
+        // If any servers already exist, close them.
+        let _ = self.close_servers().await;
+
+        // Insert the new servers.
+        *self.external_servers.lock().await =
+            ExternalServers { ..external_servers };
```
jgallagher, line 294: Nit - I think this could be

```suggestion
        *self.external_servers.lock().await = external_servers;
```

nexus/src/app/mod.rs
```diff
@@ -245,6 +280,55 @@ impl Nexus {
         }
     }
 
+    // Called to hand off management of external servers to Nexus.
+    pub(crate) async fn set_servers(
+        &self,
+        external_servers: ExternalServers,
+        internal_server: DropshotServer,
```
jgallagher, line 287: Server lifetime question - will the servers passed into this method already be bound to their listening ports, or does that happen later? (If already bound, does that require that they be on different ports than any existing servers, if we close those as a part of this method?)

nexus/src/app/mod.rs
```diff
@@ -58,6 +61,27 @@ pub(crate) const MAX_NICS_PER_INSTANCE: usize = 8;
 // TODO-completness: Support multiple external IPs
 pub(crate) const MAX_EXTERNAL_IPS_PER_INSTANCE: usize = 1;
 
+pub(crate) struct ExternalServers {
+    config: dropshot::ConfigDropshot,
+    https_port: u16,
+    http: Option<DropshotServer>,
```
jgallagher, line 67: Is the `http` server actually optional (given `refresh_tls_config` can fail if it's `None`)?

common/src/sql/dbinit.sql
```diff
@@ -128,6 +128,32 @@ CREATE INDEX ON omicron.public.service (
...
+
+    -- cert.pem file as a binary blob
+    cert BYTES NOT NULL,
+
+    -- key.pem file as a binary blob
+    key BYTES NOT NULL
+);
+
+-- Add an index which lets us look up certificates for a particular service
+-- class.
```
jgallagher, line 152: `certificate_list` is paginated by name - does it need an index also?

nexus/db-model/src/certificate.rs
```diff
@@ -0,0 +1,45 @@
...
+use super::ServiceKind;
+use crate::schema::certificate;
+use db_macros::Resource;
+use nexus_types::external_api::params;
+use nexus_types::external_api::views;
+use nexus_types::identity::Resource;
+use uuid::Uuid;
+
+/// Representation of x509 certificates used by services.
+#[derive(Queryable, Insertable, Debug, Clone, Selectable, Resource)]
```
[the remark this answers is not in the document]
smklein: Done in 6841e232cb616984a1ef8064d6077068f51d8cc2 - implemented manually, redacting the key.

nexus/test-interface/src/lib.rs
```diff
@@ -41,8 +41,8 @@ use uuid::Uuid;
 pub trait NexusServer {
     async fn start_and_populate(config: &Config, log: &Logger) -> Self;
 
-    fn get_http_servers_external(&self) -> Vec<SocketAddr>;
-    fn get_http_server_internal(&self) -> SocketAddr;
+    async fn get_http_servers_external(&self) -> Vec<SocketAddr>;
```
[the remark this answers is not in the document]
smklein: I went ahead and made this non-vec-y in 5f6543ddfc5c07be24052a85af74ad1ae32ba67e , so that there's explicit control over `http` vs `https`

nexus/tests/integration_tests/certificates.rs
```diff
@@ -0,0 +1,395 @@
...
+            .body(hyper::Body::empty())
+            .unwrap();
+
+        match scheme.as_str() {
+            "http" => {
+                let http_client = hyper::Client::builder().build_http();
+                http_client.request(request).await.map(|_| ())
+            }
+            "https" => {
+                let make_pki_verifier = || {
```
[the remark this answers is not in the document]
smklein: Doesn't need to be; it was just a "one-time-use function". Made non-closurey in a98d492aa6966927ccc0d89af4e57e1905e9ed14

nexus/tests/integration_tests/certificates.rs
```diff
@@ -0,0 +1,395 @@
...
+        .await
+        .expect("failed to issue GET requst for certificate")
+        .parsed_body()
+        .expect("failed to parse response for certificate GET request")
+}
+
+async fn cert_get_expect_not_found(client: &ClientTestContext) {
+    let cert_url = format!("{CERTS_URL}/{}", CERT_NAME);
+    let error = NexusRequest::new(
+        RequestBuilder::new(client, Method::GET, &cert_url)
```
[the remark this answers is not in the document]
smklein: Done in da1092c14b71b96f562bf94b757b5170ff04b42a

nexus/src/lib.rs
smklein: Done in b9cfb6ea2d191d9bfaec4030e8d9412cc1232969

nexus/src/app/mod.rs
smklein: Done in 70221549c05b500c2098efeb70619d874e94f62e

nexus/src/app/mod.rs
smklein: The servers passed into this method definitely *can* be bound to listening ports - but if they were, that failure would happen at bind time.

The `set_servers` method is being called during Nexus initialization, so this "close" is just defensive coding, not really a necessary aspect of setup.

nexus/src/app/mod.rs
smklein: `refresh_tls_config` really just wants to get the `ServerContext` from *somewhere* - I refactored this to abff7032de80f20fb2f4020ed917587d895f74d4.

For now: Yes, the `http` server is optional, but that's kinda because of the weird relationship between `ServerContext` and `Nexus`. Namely: `Nexus` is created first (without servers - this is when it truly must be optional), then `ServerContext` is created (referencing `Nexus`), then `Nexus`' servers can be initialized (referencing `ServerContext`, making it non-optional).

I could make it `OnceCell` or something, but I actually do want to eventually allow the http server to be optional, so we can roll with HTTPS only when we have an automated certificate refresh mechanism that's well-tested.

common/src/sql/dbinit.sql
smklein: Done, with test in 027b0e854481259236e37cc502bc858aa583a720

nexus/src/app/mod.rs
jgallagher: Makes sense. I was going to ask if `ExternalServers` (or maybe `Nexus`?) could hold a non-optional `ServerContext`, since we always know we have a single one of those regardless of whether we have an http server / https server / both / neither, but I think that has the same chicken-and-egg problem of creation order and would create an `Arc` cycle (`Nexus` -> `ExternalServers` -> `ServerContext` -> `Nexus`) that would cause all the types involved to never be dropped. The current code has the same cycle, but it's explicitly broken in `close_servers` when we `take()` the servers out of `http`/`https`.

sunshowers: To aid my understanding: can both `http` and `https` be `Some` at any given time?

nexus/src/app/mod.rs
```diff
@@ -58,6 +61,44 @@ pub(crate) const MAX_NICS_PER_INSTANCE: usize = 8;
...
+}
+
+impl ExternalServers {
+    pub fn new(config: dropshot::ConfigDropshot, https_port: u16) -> Self {
+        Self { config, https_port, http: None, https: None }
+    }
+
+    pub fn set_http(&mut self, http: DropshotServer) {
+        self.http = Some(http);
+    }
```
sunshowers, line 78: How about adding a `set_https` here for symmetry and using it?

nexus/src/app/mod.rs
```diff
@@ -58,6 +61,44 @@ pub(crate) const MAX_NICS_PER_INSTANCE: usize = 8;
...
+        } else if let Some(context) =
+            self.http.as_ref().map(|server| server.app_private())
+        {
+            // If an HTTP server is already running, use that server context.
+            Some(context.clone())
+        } else {
+            None
+        }
+    }
+}
```
sunshowers, line 100: This feels a little weird to me. I guess it's because there's really just one `Arc<crate::ServerContext>` that's shared across http and https and we want to return it? Maybe it makes sense to just store a (non-optional) `Arc<crate::ServerContext>` within `ExternalServers` itself?

Is it guaranteed that there's always just one server context?

nexus/src/app/mod.rs
```diff
@@ -245,6 +297,56 @@ impl Nexus {
         }
     }
 
+    // Called to hand off management of external servers to Nexus.
```
sunshowers, line 300: Called by whom?

nexus/src/app/mod.rs
```diff
@@ -245,6 +297,56 @@ impl Nexus {
...
+        external_servers.https.as_ref().map(|server| server.local_addr())
+    }
+
+    pub async fn get_internal_server(&self) -> Option<std::net::SocketAddr> {
+        self.internal_server
+            .lock()
+            .await
+            .as_ref()
+            .map(|server| server.local_addr())
+    }
```
sunshowers, line 348: These seem to return the external server's address rather than a handle to the server itself, as the names suggest? Maybe `get_http_external_server_address` etc?

sunshowers (commented):
(Still reviewing this)

common/src/sql/dbinit.sql
```diff
@@ -128,6 +128,39 @@ CREATE INDEX ON omicron.public.service (
...
+CREATE INDEX ON omicron.public.certificate (
+    service
```
davepacheco, line 154: I'm always worried about non-unique indexes -- see #1497.  Is this one okay?

common/src/sql/dbinit.sql
```diff
@@ -128,6 +128,39 @@ CREATE INDEX ON omicron.public.service (
...
+);
+
+-- Add an index which enforces that certificates have unique names, and which
+-- allows pagination-by-name.
+CREATE UNIQUE INDEX ON omicron.public.certificate (
+    name
```
davepacheco, line 160: Are names intended to be globally unique (not per-service)?

common/src/sql/dbinit.sql
```diff
@@ -128,6 +128,39 @@ CREATE INDEX ON omicron.public.service (
     sled_id
 );
 
+-- x509 certificates which may be used by services
+CREATE TABLE omicron.public.certificate (
+    -- Identity metadata (asset)
```
davepacheco, line 133: nit: this doesn't look like asset metadata.  it looks like resource metadata, since it has a name, mtime, etc.

nexus/Cargo.toml
```diff
@@ -86,15 +86,19 @@ oximeter-producer.workspace = true
 [dev-dependencies]
 criterion.workspace = true
 expectorate.workspace = true
+hyper-rustls = "0.23.2"
```
davepacheco, line 89: What's our policy re: putting versions in here vs. using workspace-level dep?  I would have thought the expectation was to always use a workspace dep.  Otherwise it seems like every time you add a dep here, you need to see if any other crate is using it first so you figure out if it needs to be hoisted to the workspace level?

nexus/src/app/certificate.rs
```diff
@@ -0,0 +1,230 @@
...
+use openssl::x509::X509;
+use ref_cast::RefCast;
+use std::sync::Arc;
+use uuid::Uuid;
+
+fn validate_certs(input: Vec<u8>) -> Result<(), String> {
+    let certs = X509::stack_from_pem(&input.as_slice())
+        .map_err(|err| format!("Failed to parse certificate as PEM: {err}"))?;
+    if certs.is_empty() {
+        return Err("could not parse".to_string());
```
davepacheco, line 29: This failure mode seems rather undebuggable...I guess we don't get any more information than this out of the library?

nexus/src/app/certificate.rs
```diff
@@ -0,0 +1,230 @@
...
+                .certificate_name(Name::ref_cast(name)),
+        }
+    }
+
+    pub async fn certificate_create(
+        self: &Arc<Self>,
+        opctx: &OpContext,
+        params: params::CertificateCreate,
+    ) -> CreateResult<db::model::Certificate> {
+        validate_certs(params.cert.clone()).map_err(|e| {
```
davepacheco, line 61: In other places I think we've defined newtypes whose serde deserialize winds up validating ("parse, don't validate").  Is that an option here (or, is there some reason we didn't do that here)?

nexus/src/app/certificate.rs
```diff
@@ -0,0 +1,230 @@
...
+            Error::InvalidValue { label: String::from("key"), message: e }
+        })?;
+
+        let new_certificate = db::model::Certificate::new(
+            Uuid::new_v4(),
+            db::model::ServiceKind::Nexus,
+            params,
+        );
+
+        // TODO: Saga?
```
davepacheco, line 76: Why?  (when would we have enough information to resolve this TODO?)

nexus/src/app/certificate.rs
```diff
@@ -0,0 +1,230 @@
...
+        match certificate {
+            NameOrId::Id(id) => {
+                LookupPath::new(opctx, &self.db_datastore).certificate_id(*id)
+            }
+            NameOrId::Name(name) => LookupPath::new(opctx, &self.db_datastore)
+                .certificate_name(Name::ref_cast(name)),
+        }
+    }
+
+    pub async fn certificate_create(
```
davepacheco, line 56: Some doc comments would be useful here...I was surprised to see that this only seems to handle Nexus certificates.  Maybe even rename these to `nexus_certificate_*()`?  (not a big deal for sure)

nexus/src/app/certificate.rs
```diff
@@ -0,0 +1,230 @@
...
+
+        Ok(Some(dropshot::ConfigTls::AsBytes {
+            certs: certificate.cert.clone(),
+            key: certificate.key.clone(),
+        }))
+    }
+
+    /// Refreshes the TLS configuration for the currently-running Nexus external
+    /// server. This involves either:
+    /// - Creating a new HTTPS server if once does not exist
```
davepacheco, line 159:
```suggestion
    /// - Creating a new HTTPS server if one does not exist
```

nexus/src/app/certificate.rs
```diff
@@ -0,0 +1,230 @@
...
+    /// - Refreshing an existing HTTPS server if it already exists
+    /// - Tearing down an HTTPS server if no certificates exist
```
davepacheco, line 161: Do we ever actually want to do this?

More generally I'm a little unclear about how we want this sort of management to work.  Right now it looks like there's no way to add a certificate that is not immediately used.  I wonder if it will be important to separate these steps.  Along these lines, if we remove a cert as soon as it's deleted, and someone wants to replace the cert by delete + create, might they accidentally brick the rack by doing it in the wrong order?

nexus/src/app/certificate.rs
```diff
@@ -0,0 +1,230 @@
...
+            }
+            // Refresh an existing server.
+            (Some(tls_config), Some(https)) => {
+                info!(
+                    self.log,
+                    "Refresh TLS: Refreshing HTTPS server at {}",
+                    https.local_addr()
+                );
+                https.refresh_tls(&tls_config).await.map_err(|e| {
+                    Error::internal_error(&format!("Cannot refresh TLS: {e}"))
```
davepacheco, line 207: What state are we in at this point?  Is the server running with the previous set of certs?

nexus/src/app/rack.rs
```diff
@@ -90,8 +93,36 @@ impl super::Nexus {
             })
             .collect();
 
+        let certificates: Vec<_> = request
+            .certs
+            .into_iter()
+            .enumerate()
+            .map(|(i, c)| {
+                // One-indexing for user-visible values.
```
davepacheco, line 101:
```suggestion
                // The indexes that appear in user-visible names for these certificates start from one (e.g., certificate names "default-1", "default-2", etc.).
```

(I initially misread this comment to mean that this index needed to match up with some 1-indexed thing that we exposed to end users, so I suggested this edit to be more explicit.)

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -3327,6 +3333,129 @@ async fn instance_disk_detach(
     apictx.external_latencies.instrument_dropshot_handler(&rqctx, handler).await
 }
 
+// Certificates
+
+/// List system-wide certificates
+///
+/// Returns a list of all the system-wide certificates. System-wide certificates are returned sorted
```
davepacheco, line 3340: From the user's perspective, these are more specifically "TLS certificates for the externally-facing API servers", right?

I wonder if we even want to put that into the URL.  In the future we might manage other kinds of certificates and I imagine we'd want separate endpoints to manage them.

nexus/test-utils/src/lib.rs
```diff
@@ -55,6 +55,36 @@ impl<N: NexusServer> ControlPlaneTestContext<N> {
...
+    pub async fn external_https_client(&self) -> ClientTestContext {
+        self.server
+            .get_https_server_external()
+            .await
+            .map(|addr| {
+                ClientTestContext::new(
+                    addr,
+                    self.logctx
+                        .log
+                        .new(o!("component" => "external client test context")),
```
davepacheco, line 83:
```suggestion
                        .new(o!("component" => "external https client test context")),
```

nexus/test-utils/src/lib.rs
```diff
@@ -55,6 +55,36 @@ impl<N: NexusServer> ControlPlaneTestContext<N> {
...
+    pub async fn external_http_client(&self) -> ClientTestContext {
+        self.server
+            .get_http_server_external()
+            .await
+            .map(|addr| {
+                ClientTestContext::new(
+                    addr,
+                    self.logctx
+                        .log
+                        .new(o!("component" => "external client test context")),
```
davepacheco, line 68:
```suggestion
                        .new(o!("component" => "external http client test context")),
```

nexus/tests/config.test.toml
```diff
@@ -2,6 +2,9 @@
 # Oxide API: configuration file for test suite
 #
 
+# Avoid conflicting with other ports during HTTPS tests
```
davepacheco, line 5: Why is this a separate top-level property?

nexus/tests/integration_tests/unauthorized.rs
```diff
@@ -498,7 +504,10 @@ async fn verify_endpoint(
             if let Some(&AllowedMethod::GetWebsocket) = allowed {
                 request = request.websocket_handshake();
             }
-            let response = request.execute().await.unwrap();
+            let response = request
+                .execute()
+                .await
+                .expect(&format!("Failed making {method} request to {uri}"));
```
davepacheco, line 510: nit: I don't really care but I think clippy usually complains about this and wants `.unwrap_or_else(|| panic!(...))`

(clippy currently doesn't run on test code, I think, but that's a bug)

nexus/tests/output/authz-roles.out
```diff
@@ -950,6 +950,23 @@ resource: GlobalImage id "b46bf5b5-e6e4-49e6-fe78-8e25d698dabc"
   silo1-org1-proj1-viewer          ✘  ✔  ✘  ✔  ✘  ✘  ✘  ✘
   unauthenticated                  !  !  !  !  !  !  !  !
 
+resource: Certificate id "c56bf5b5-e6e4-49e6-fe78-8e25d698dabc"
```
davepacheco, line 953: 🎉

nexus/src/app/mod.rs
```diff
@@ -81,6 +122,12 @@ pub struct Nexus {
     /// Task representing completion of recovered Sagas
     recovery_task: std::sync::Mutex<Option<db::RecoveryTask>>,
 
+    /// External dropshot servers
+    external_servers: Mutex<ExternalServers>,
+
+    /// Internal dropshot server
+    internal_server: Mutex<Option<DropshotServer>>,
```
sunshowers, line 129: It seems like `external_servers` here is a `tokio::sync::Mutex` rather than `std::sync::Mutex` because `refresh_tls_config` sometimes holds the lock open across await points -- is that correct? If so could you document this in a comment?

Also does the same apply to `internal_server`? If not then maybe it makes sense to use a `std::sync::Mutex` instead (similar to `recovery_task` above?)

(I'm sure you're already familiar with it, but for folks who might be wondering about the difference between `std::sync::Mutex` and `tokio::sync::Mutex`: https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use)

nexus/src/app/mod.rs
```diff
@@ -245,6 +297,56 @@ impl Nexus {
...
+        let mut external_servers = self.external_servers.lock().await;
+        if let Some(server) = external_servers.http.take() {
+            server.close().await?;
+        }
+        if let Some(server) = external_servers.https.take() {
+            server.close().await?;
+        }
+        if let Some(server) = self.internal_server.lock().await.take() {
+            server.close().await?;
+        }
```
sunshowers, line 324: Can/should these be parallelized?

sunshowers (commented):
Just a couple more comments.

nexus/src/app/mod.rs
smklein: Yes, it's possible to have both HTTP and HTTPS servers exposed (on different ports).

nexus/src/app/certificate.rs
smklein: I was trying to keep this PR simple - my goal was "simple CRUD API, and we'll use whatever certificate we have for Nexus".

Do you want me to add an explicit endpoint for "certificate refresh" and "certificate promotion / activation"? I could make something more complex, where we add endpoints using the CRUD API, but then can explicitly nominate one for usage? I do think that has the same issue that "eventually, you make a call to the server, where on success it uses a new certificate for future requests", unless we also want to add a timeout or multi-server setup?

Alternatively, I can go back up to dropshot, and add support for handling multiple x.509 certificates simultaneously? That might require some invasive changes to rustls though

nexus/src/app/certificate.rs
```diff
@@ -0,0 +1,246 @@
...
+    }
+    Ok(())
+}
+
+fn validate_private_key(key: Vec<u8>) -> Result<(), String> {
+    let _ = PKey::private_key_from_pem(&key.as_slice())
+        .map_err(|err| format!("Failed to parse private key as PEM: {err}"))?;
+
+    Ok(())
+}
```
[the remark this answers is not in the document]
smklein: Sure, updated to use `thiserror` - a lot of Nexus needs to convert errors to `omicron_common::api::external::Error`, which implies an HTTP-level error, so I was just being a little fast-and-loose with Strings, but I can be more explicit.

See: c2f2dfedf3bcadb007636ff718ca7f7516b8b624

nexus/src/app/certificate.rs
```diff
@@ -0,0 +1,246 @@
...
+            .certificate_list_for(
+                opctx,
+                db::model::ServiceKind::Nexus,
+                pagparams,
+            )
+            .await
+    }
+
+    pub async fn certificate_delete(
+        self: &Arc<Self>,
```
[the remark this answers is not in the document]
smklein: It *would* be if these were using sagas, for semi-questionable reasons (needing to grab another reference to `Nexus`), but fair point, we aren't using them here.

Removed: a616d1a47cacd2a8c7c19e20d27a0502a9c73107

nexus/src/app/mod.rs
smklein: Sure; done in b339ed90d9b36b135d32e7771fb6c49f22a08394

nexus/src/app/mod.rs
smklein: Currently this happens to be called by `nexus::Server::start(...)` - I'm a little hesitant to reference that explicitly in a comment directly though, since that seems prone to bit-rot via refactor.

nexus/src/app/mod.rs
smklein: Done in 8450dbb1aaed31ac1182055eebc42406704b3b99

common/src/sql/dbinit.sql
smklein: I can also add an index by ID, I think that should be fine?

I really just want the ability to have certificates for different classes of service (e.g., "here are all certificates for nexus") so I can filter by that category and pull them out later. This is a case that definitely shouldn't be unique - it's fine to have multiple certs for a service class - but would the following eliminate this problem?

```sql
CREATE INDEX on omicron.public.certificate (
  id,
  service
) WHERE time_deleted IS NULL;
```

(Updated in 24d9421fd07e7f7f391e5b457c4c9726ad8b8ca8)

common/src/sql/dbinit.sql
smklein: Yeah, that was my intent.

common/src/sql/dbinit.sql
smklein: You're right; this was a typo. I started this PR making it "internal only", but when I realized it made sense as part of the API, I added the "name" and other fields. Updated this comment in a4972fd9893d9457c9a78a0d7dc9c1494116de97

nexus/Cargo.toml
smklein: Using workspace deps in 705d6464b72ff25d5df490138800ac96f116fff5

nexus/src/app/certificate.rs
smklein: It' not perfect, but I can move the validation into the model layer: 80791ba5e48b909297d7bdbb383e01ffc9382668

The type I want is the `db_model::Certificate` anyway, so I can make validation part of that constructor.

nexus/src/app/certificate.rs
smklein: Updated in 400d90c8528b936c873df2c9ca317344c079a542

nexus/src/app/certificate.rs
smklein: Updated in 184811fde2594c7d4fc52a4de8cb4728a4457c58 - I'm adding the service type in the API. I expected we'd handle certificates for services other than Nexus eventually, but had originally punted it. I can include it now (as an enum of one variant), to make the intent more clear.

nexus/src/app/certificate.rs
smklein: Fixed in 1269e18634b24b0835e54efc9518c7db4cf51c45

nexus/src/app/certificate.rs
smklein: I'll unwrap explicitly here: 4bafeeefbcc14c81e7bc6852665d4bfb4b57e158 , this error should not happen

nexus/src/app/rack.rs
smklein: Done fb340d7f971b072951da271923156bf58fa313dc

nexus/src/external_api/http_entrypoints.rs
smklein: I'm using a service type in the input and output: fc877e330f7b978265d91996faa596b228e8f786 . It's only nexus now, but hopefully this makes future certificate usage more clear

nexus/test-utils/src/lib.rs
smklein: Done 187f5bd761b3a6b847a0194c5bd6d1d9e34b56f6

nexus/tests/config.test.toml
smklein: - We need to know what this is before the HTTPS server is running; it's possible that the certificate is only known after the "create_certificate" API is invoked
- In production, we want it to be 443
- If it's 443 for our tests, they'll fail, because they're binding to localhost and will conflict

nexus/tests/integration_tests/unauthorized.rs
smklein: Done in 254bf902adf67a46b1ce8332888b0e9c7bd90882

nexus/src/app/mod.rs
smklein: They can be, though I'm not sure they should be - this operation isn't very common.

nexus/src/app/mod.rs
smklein: Yeah, I did this for `external_server`, and only used the tokio mutex on the internal server for parity. It's true, however -- we don't need to use an async mutex here. Updated in e34fb02cba788ff3f9730134fcb6b936b52a87a2

common/src/sql/dbinit.sql
davepacheco: I think the order of columns in the index is wrong now.  I think you want it on (service, id).  Otherwise, you can't use this index to efficiently list the certificates for a particular service because they're sorted by id first.  (But you _do_ need the id in the index or else you can't _page_ through the list of certificates for a particular service using the index.)

nexus/db-model/src/certificate.rs
```diff
@@ -8,8 +8,55 @@ use db_macros::Resource;
...
 
+#[derive(Debug, thiserror::Error)]
+pub enum CertificateError {
+    #[error("Failed to parse certificate: {0}")]
+    BadCertificate(openssl::error::ErrorStack),
+
+    #[error("Certificate exists, but is empty")]
+    CertificateEmpty,
+
+    #[error("Failed to parse private key")]
```
davepacheco, line 24: Is there anything useful in the openssl stack here (that'd be useful to put into the error message the way we do for `BadCertificate`)?

nexus/src/app/certificate.rs
davepacheco: Nice!

nexus/src/app/certificate.rs
davepacheco: Thanks!

nexus/types/src/external_api/shared.rs
```diff
@@ -151,6 +151,16 @@ pub enum UserProvisionType {
...
 }
 
+/// The service intended to use this certificate.
+#[derive(Clone, Copy, Debug, Deserialize, Serialize, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub enum ServiceUsingCertificate {
+    /// This certificate is intended for the "Nexus" service.
+    ///
+    /// This service acts as the main API point for accessing the Oxide rack.
+    Nexus,
```
davepacheco, line 161: nit: I'd suggest we call this `External` or `ExternalApi` instead.  I'm not sure operators should be aware of "Nexus", or that it's the external endpoint (it might not always be! if we add a load balancer in the future).  Plus, these certs are used for only one of Nexus's servers (not the internal API).

davepacheco (approved):
Forgot to hit "approve".  I'm cool with deferring all the items I mentioned.

nexus/src/app/mod.rs
```diff
@@ -126,7 +130,7 @@ pub struct Nexus {
     external_servers: Mutex<ExternalServers>,
 
     /// Internal dropshot server
-    internal_server: Mutex<Option<DropshotServer>>,
+    internal_server: std::sync::Mutex<Option<DropshotServer>>,
```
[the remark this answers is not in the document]
smklein: It was requested that I make this change in https://github.com/oxidecomputer/omicron/pull/2099#discussion_r1067409290 ; @sunshowers , do you have strong opinions?

nexus/tests/config.test.toml
smklein: It's bizarre to pass `ConfigDropshot` here for the HTTPS server - if we [look at the struct](https://docs.rs/dropshot/0.8.0/dropshot/struct.ConfigDropshot.html), it contains the following:

```rust
pub struct ConfigDropshot {
    // Contains an address and port. Whatever address we pass ahead-of-time is a complete fabrication.
    pub bind_address: SocketAddr,
    // This part of the configuration is "true".
    pub request_body_max_bytes: usize,
    // This part of the configuration is a complete fabrication - when Nexus is launched, the certificates are not necessarily known.
    pub tls: Option<ConfigTls>,
}
```

So, basically: If we pass a `ConfigDropshot` for the external server before nexus has booted, it's only half valid, which IMO is confusing.

I'd rather pass configuration values that are "entirely true", and let Nexus fill in the rest (TLS info, addresses, etc) with known-at-runtime information.

nexus/db-model/src/certificate.rs
smklein: I had a request to do roughly the opposite of this in https://github.com/oxidecomputer/omicron/pull/2099#discussion_r1065046652 - I don't have any way to parse the openssl stack, and I haven't validated that it wouldn't leak private key info.

davepacheco: Okay, we can deal with that if we run into this failure mode in the wild.

PR #2100, opened 2022-12-29
System Update API

david-crespo:
[RFD 348](https://rfd.shared.oxide.computer/rfd/0348) is still in progress, but I find it helpful to do this in parallel.

--- 

This is a big PR, but fortunately it is very straightforward: endpoints plus the service functions, models, and tables you'd expect to see backing those endpoints. It's a good time to add "updateable" to your computer's dictionary.

### Included here

- [x] Endpoints
  - [x] System version + status (updating or not + reason)
  - [x] Get tree of updateable components (each w/ version + status)
  - [x] List system updates
  - [x] Get system update 
  - [x] Get tree of component updates for top-level system update
  - [x] Start update (noop, returns placeholder `SystemUpdateDeployment`)
  - [x] Stop update (noop)
  - [x] Change existing refresh endpoint to match new `/v1/system/update/...` convention
- [x] Nexus functions for all that except start and stop
  - [x] Basic tests that create the thing and then retrieve it
- [x] Tables
  - [x] System updates
  - [x] Component updates
  - [x] Join table between system updates and component updates
  - [x] Updateable components
  - [x] Update deployments
- Weird stuff
  - [x] `version_sort` column on `system_update` and `component_update` that lets us hack in sorting by version
  - [x] Enforce maximum major/minor/patch version number because `version_sort` relies on zero-padding
  - [x] Make system update semver version the PK for the system updates table and fetch by version in endpoints
    - Ideally we'd be able to fetch by ID also, but there would been a lot of boilerplate to make it work like `NameOrId` without comparable value due to the latter's reuse in many places. Version is unique and immutable, so this should be good enough to start.

### Conspicuously missing, to be done in followups

- `nexus` functions for updating updateable component rows (possibly also system updates and component updates — not sure if they're supposed to be modified)
- Actually do an update

Do a find for `TODO:` for an embarrassing number of additional bullets.

### Open questions

- What is a good length limit for version strings in the DB? There's no official limit, but the docs suggest a number less than or equal to 255: https://semver.org/#does-semver-have-a-size-limit-on-the-version-string
  - Where/how do we enforce the length limit besides the DB?
- How to represent system status, especially steady reason


nexus/types/src/external_api/views.rs
```diff
@@ -412,3 +413,47 @@ pub struct DeviceAccessTokenGrant {
...
+
+#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema)]
+pub struct SystemVersionRange {
+    // TODO: if we want to use semver::Version, I think we have to newtype it
+    // in order to implement JsonSchema. ew
+    // pub low: semver::Version,
+    // pub high: semver::Version,
+    pub low: String,
+    pub high: String,
+}
```
david-crespo, line 427: @smklein if we want to use `semver::Version` here, which sounds sensible, we need to wrap it and `impl JsonSchema` on that. Correct?

smklein: I believe so - looks like someone already started work on this here: https://github.com/GREsau/schemars/pull/195

david-crespo: Good timing. I looked but apparently in the wrong place.

nexus/types/src/external_api/views.rs
```diff
@@ -479,6 +479,8 @@ pub enum VersionStatus {
 pub struct SystemVersion {
     pub version_range: VersionRange,
     pub status: VersionStatus,
+    // TODO: time_released? time_last_applied? I got a fever and the only
```
smklein, line 482: 🐮🔔

nexus/tests/output/nexus_tags.txt
```diff
@@ -173,13 +173,20 @@ silo_view                                /system/silos/{silo_name}
...
 system_image_view                        /system/images/{image_name}
 system_image_view_by_id                  /system/by-id/images/{id}
+system_update_components_list            /v1/system/update/updates/{update_id}/components
+system_update_list                       /v1/system/update/updates
+system_update_start                      /v1/system/update/updates/{update_id}/start
+system_update_stop                       /v1/system/update/updates/{update_id}/stop
+system_update_view                       /v1/system/update/updates/{update_id}
 system_user_list                         /system/user
 system_user_view                         /system/user/{user_name}
+system_version                           /v1/system/update/version
```
david-crespo, line 189: IF (big IF) we don't have anything else hanging off of `/system/update` other than what's here, seems like we could get away without doing my goofy `/update/updates` idea if we put system version and component version tree under `/system/version`. So:

```
system_update_list                       /v1/system/updates
system_update_view                       /v1/system/updates/{update_id}
system_update_components_list            /v1/system/updates/{update_id}/components
system_update_start                      /v1/system/updates/{update_id}/start
system_update_stop                       /v1/system/updates/{update_id}/stop

system_version                           /v1/system/version
system_component_version_list            /v1/system/version/components
```

This has the disadvantage of putting version under a different top-level key, which makes it less obvious how closely tied it is to the `/system/updates` endpoints.

david-crespo:
@smklein so far it seems like we'll need the following tables. Maybe we can pair tomorrow on throwing something basic in.

- system updates
- component updates
- join table associating system updates and component updates
- component version and status
- update in progress

System version + status response would be put together by looking at the update in progress table for the status and pulling the low and high versions from the list of components.

common/src/sql/dbinit.sql
```diff
@@ -1419,6 +1419,20 @@ CREATE INDEX ON omicron.public.update_available_artifact (
...
+ */
+CREATE TABLE omicron.public.system_update (
+    /* Unique identifier for this update package */
+    id UUID PRIMARY KEY,
+    time_created TIMESTAMPTZ NOT NULL,
+    time_modified TIMESTAMPTZ NOT NULL
+
+    /* Unique semver version */
+    /* TODO: If the version is really supposed to be unique, we could make it the PK? */
+    version STRING(40) NOT NULL,
```
david-crespo, line 1433: We could do ser/de on strings but it's kind of silly. Would be easier to store major, minor, patch as numbers and run them through `SemverVersion::new(major, minor, patch)` on DB read.

nexus/tests/output/nexus_tags.txt
zephraph: I much prefer this. The `/system/update/updates` pattern is fairly different than the rest of our API where typically resources are plural.

david-crespo: I forgot, we already have one:

https://github.com/oxidecomputer/omicron/blob/75bd90550e63940bbed7ef40ee21e18fa1b628dc/nexus/src/external_api/http_entrypoints.rs#L5032-L5040

nexus/db-model/src/system_update.rs
```diff
@@ -0,0 +1,168 @@
...
+    #[diesel(embed)]
+    pub identity: UpdateableComponentIdentity,
+    pub device_id: String,
+    pub component_type: UpdateableComponentType,
+    pub version: SemverVersion,
+    // pub status: VersionStatus,
+    /// ID of the parent component, e.g., the sled a disk belongs to. Value will
+    /// be `None` for top-level components whose "parent" is the rack.
+    pub parent_id: Option<Uuid>,
+}
```
david-crespo, line 156: As mentioned in the "open questions" section of the PR description, I've determined this model is insufficient because it does not allow me to represent nodes like `PSC` or `Host` below (example tree taken from [RFD 334](https://rfd.shared.oxide.computer/rfd/0334#_requirements)), which are parents of updateable components but which are not themselves directly associated with any update artifacts. Displaying this tree in full seems pretty important for the UI, so I'm going to rethink this model a bit. It's probably not a drastic change.

```
Rack─┬─PSC─┬─RoT──Hubris Archive ABCD
     │     └─SP───Hubris Archive DCBA
     ├─Sidecar─┬─RoT──Hubris Archive ABCD
     │         └─SP───Hubris Archive FACE
     ├─Sidecar─┬─RoT──Hubris Archive ABCD
     │         └─SP───Hubris Archive FACE
     ├─Scrimlet─┬─RoT──Hubris Archive ABCD
     │          ├─SP───Hubris Archive BEEF
     │          └─Host─┬─Host Phase 1 Image 1234
     │                 ├─Host Phase 2 Image 1234
     │                 └─Omicron Image 5678
     ├─Scrimlet …
     ├─Gimlet …
     └─Gimlet─┬─RoT──Hubris Archive ABCD
              ├─SP───Hubris Archive BEEF
              └─Host─┬─Host Phase 1 Image 1234
                     ├─Host Phase 2 Image 1234
                     └─Omicron Image 5678
```

common/src/sql/dbinit.sql
```diff
@@ -1419,6 +1419,83 @@ CREATE INDEX ON omicron.public.update_available_artifact (
...
 );
 
+/*
+ * System updates
+ */
+CREATE TABLE omicron.public.system_update (
+    /* Identity metadata (asset) */
+    id UUID PRIMARY KEY,
+    time_created TIMESTAMPTZ NOT NULL,
+    time_modified TIMESTAMPTZ NOT NULL,
```
zephraph, line 1429: What does it mean for an update to be modified? The version updated?

zephraph: Also, is there a need for an index to be created for this or other tables here?

david-crespo: I don't really know why these would be modified. The only reason I have it on here is because it's convenient to make these `Asset`s, which are like `Resource` but without name and description. I don't think it's too bad to have the field there and never use it — the alternative (not being an `Asset`) is probably worse, though I haven't thought about it too hard.

Yes to indexes, will put that on my to-do list.

nexus/src/app/update.rs
```diff
@@ -276,4 +304,295 @@ impl super::Nexus {
...
+            .await
+    }
+
+    pub async fn system_update_fetch_by_id(
+        &self,
+        opctx: &OpContext,
+        update_id: &Uuid,
+    ) -> LookupResult<db::model::SystemUpdate> {
+        // TODO: I don't think this is the right way to do this auth check
+        opctx.authorize(authz::Action::ListChildren, &authz::FLEET).await?;
```
zephraph, line 359: For this, I don't _think_ you need an explicit authorize call. When you call `fetch()` below that checks to ensure you have read access to the `SystemUpdate` resource. That's all you should really need here. The `ListChildren` of the parent is only required on a list endpoint.

nexus/src/app/update.rs
```diff
@@ -276,4 +304,295 @@ impl super::Nexus {
...
+        self.db_datastore.system_updates_list_by_id(opctx, pagparams).await
+    }
+
+    pub async fn system_update_list_components(
+        &self,
+        opctx: &OpContext,
+        update_id: &Uuid,
+    ) -> ListResultVec<db::model::ComponentUpdate> {
+        // TODO: I don't think this is the right way to do this auth check
+        opctx.authorize(authz::Action::ListChildren, &authz::FLEET).await?;
```
zephraph, line 384: It could be! Is fleet the parent?

nexus/src/app/update.rs
david-crespo: I thought that, but when I didn't do this, my authorization test showed it was being more permissive than I expected. That could mean I gave the wrong arguments to the auth resource macro.

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -5048,6 +5058,201 @@ async fn updates_refresh(
...
+/// List all updates
+#[endpoint {
+     method = GET,
+     path = "/v1/system/update/updates",
+     tags = ["system"],
+}]
+async fn system_update_list(
+    rqctx: Arc<RequestContext<Arc<ServerContext>>>,
+    // TODO: pagination should probably be by most recent first, ID is nonsense
+    query_params: Query<PaginatedById>,
```
zephraph, line 5128: That'll just be a sort mode we'll have to add on the ID based pagination. Pagination by ID (or name) denotes how it indicates what the next page is internally but doesn't necessarily define how it's sorted. That'll be left up to the `sort_by` (which does default to some mode). We'll just add a new sort mode to `PaginatedById` which'll cover this case.

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -5048,6 +5058,201 @@ async fn updates_refresh(
...
+    apictx.external_latencies.instrument_dropshot_handler(&rqctx, handler).await
+}
+
+/// Path parameters for SystemUpdate requests
+#[derive(Deserialize, JsonSchema)]
+struct SystemUpdatePathParam {
+    // TODO: do updates have names? Should this therefore be NameOrId? Do we
+    // allow access by some other string identifier (like version) which, like
+    // Name, is disjoint with the set of Uuids? NameOrVersion?
+    id: Uuid,
```
zephraph, line 5157: I definitely think we _wouldn't_ do `NameOrId` here. It doesn't, in my mind, make sense for an update to have a name. `IdOrVersion` could be a thing, but there's actually quite a bit of boilerplate around making something like that work so we should just proceed with caution there.

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -5048,6 +5058,201 @@ async fn updates_refresh(
...
+    apictx.external_latencies.instrument_dropshot_handler(&rqctx, handler).await
+}
+
+/// View system update component tree
+#[endpoint {
+    method = GET,
+    path = "/v1/system/update/updates/{id}/components",
+    tags = ["system"],
+}]
+async fn system_update_components_list(
```
zephraph, line 5188: So we settled on not pagination the component lists? I'd saw that mentioned in the RFD.

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -5048,6 +5058,201 @@ async fn updates_refresh(
...
+/// Start system update
+#[endpoint {
+    method = POST,
+    path = "/v1/system/update/updates/{id}/start",
+    tags = ["system"],
+}]
+async fn system_update_start(
+    rqctx: Arc<RequestContext<Arc<ServerContext>>>,
+    path_params: Path<SystemUpdatePathParam>,
+) -> Result<HttpResponseUpdatedNoContent, HttpError> {
```
zephraph, line 5219: I think you want `HttpResponseAccepted`?

nexus/tests/output/nexus_tags.txt
zephraph: I don't guess dropshot would let you register `/system/updates/refresh` before `/system/updates/{update_id}` to have it automatically be handled, would it? 😅. There's actually no overlap in this case because `refresh` isn't a valid UUID. If that worked, it'd be sort of ideal. 

Beyond that, I'm not really sure. I obviously don't think we should stall on this API decision and I'm fine with moving forward with what's here. Hopefully we'll figure out a better way to do this at some point and update accordingly.

david-crespo: No, Dropshot would have to be modified: https://github.com/oxidecomputer/dropshot/issues/199

And more and more I agree with the comments on that issue arguing it’s a footgun.

nexus/src/external_api/http_entrypoints.rs
david-crespo: Yes, for now, but it's easy to change because it's its own endpoint and it already returns a `ResultsPage`.

common/src/sql/dbinit.sql
```diff
@@ -1419,6 +1419,98 @@ CREATE INDEX ON omicron.public.update_available_artifact (
...
+    -- string as the PK. However, removing it would make this one of the only
+    -- resources without an ID, which means we'd have to do a bunch of special
+    -- stuff that we get more or less for free with other resources. For
+    -- example, paginating by ID (note this is distinct from sorting by ID).
+
+    /* Identity metadata (asset) */
+    id UUID NOT NULL,
+    time_created TIMESTAMPTZ NOT NULL,
+    time_modified TIMESTAMPTZ NOT NULL
+);
```
david-crespo, line 1439: Using the version as the PK while still keeping the ID around is one of the stranger parts of this PR, but I landed on it after trying less weird-sounding alternatives that turned out to be uglier.

smklein: Okay, minor nitpick, that doesn't really matter much but might help with consistency:

1. Can we make the ID the `PRIMARY KEY` if we're gonna have it?
2. Can we then create a unique index on the `version`?

I *think* this should be functionally very similar, but it matches the pattern of other objects with "both IDs and unique names".

common/src/sql/dbinit.sql
```diff
@@ -1419,6 +1419,98 @@ CREATE INDEX ON omicron.public.update_available_artifact (
...
+    'bootloader_for_host_proc',
+    'hubris_for_psc_rot',
+    'hubris_for_psc_sp',
+    'hubris_for_sidecar_rot',
+    'hubris_for_sidecar_sp',
+    'hubris_for_gimlet_rot',
+    'hubris_for_gimlet_sp',
+    'helios_host_phase_1',
+    'helios_host_phase_2',
+    'host_omicron'
```
smklein, line 1459: We're probably gonna add zones-updateable-separately-from-the-host-OS like CRDB + Clickhouse to this list, right? (Not in this PR, just kinda thinking out loud)

nexus/db-model/src/semver_version.rs
```diff
@@ -0,0 +1,64 @@
...
+)]
+#[diesel(sql_type = sql_types::Text)]
+#[display("{0}")]
+pub struct SemverVersion(pub external::SemverVersion);
+
+NewtypeFrom! { () pub struct SemverVersion(external::SemverVersion); }
+NewtypeDeref! { () pub struct SemverVersion(external::SemverVersion); }
+
+impl SemverVersion {
+    pub fn new(major: u64, minor: u64, patch: u64) -> Self {
```
smklein, line 35: Will we ever make it possible to have either "prerelease" or "build metadata" as a string here? 
I'm eyeing the https://docs.rs/semver/latest/semver/struct.Version.html struct for comparison.

(IMO if we say "no for this PR", I'm totally fine with that, just wondering about future expansion)

common/src/sql/dbinit.sql
```diff
@@ -1419,6 +1419,98 @@ CREATE INDEX ON omicron.public.update_available_artifact (
     targets_role_version
 );
 
+/*
+ * System updates
+ */
+CREATE TABLE omicron.public.system_update (
+    /* Unique semver version */
+    version STRING(64) PRIMARY KEY,
```
smklein, line 1427: If we're representing semver versions as STRING types, we won't be able to compare them within DB operations, right? 

For example, let's suppose we have two `system_update` objects, and we want to order them. Maybe we want to paginate by version, maybe we just want to query if `A < B`.

We can't perform a lexicographic comparison here, since that would make a version like `1.11.0` appear smaller than `1.9.0`. Do we need to identify semver versions by subcomponents?

nexus/db-model/src/system_update.rs
```diff
@@ -0,0 +1,178 @@
...
+
+// TODO: As mentioned in the "open questions" section of the PR description,
+// I've determined this model is insufficient because it does not allow me to
+// represent nodes like PSC or Host below (example tree taken from RFD 334),
+// which are parents of updateable components but which are not themselves
+// directly associated with any update artifacts. Displaying this tree in full
+// seems pretty important for the UI, so I'm going to rethink this model a bit.
+// It's probably not a drastic change. Might be as simple as making the artifact
+// field nullable.
```
smklein, line 142: Could we decouple this from system update, and just have this be "Component", and act as a superset of "UpdateableComponent"?

There would be a little bit of duplication in the `component_type` enum, but it should be possible to convert from `component_type` to `component_update_type` without loss.

nexus/src/db/datastore/update.rs
```diff
@@ -56,4 +64,168 @@ impl DataStore {
...
+        opctx.authorize(authz::Action::CreateChild, &authz::FLEET).await?;
+
+        use db::schema::component_update;
+        use db::schema::system_update_component_update as join_table;
+
+        let result = diesel::insert_into(component_update::table)
+            .values(update.clone())
+            .on_conflict(component_update::columns::id) // TODO: should also conflict on version
+            .do_nothing()
+            .returning(ComponentUpdate::as_returning())
```
smklein, line 112: See: https://github.com/oxidecomputer/omicron/issues/1168 - if this already existed, `result` would return `NotFound`

nexus/src/db/datastore/update.rs
```diff
@@ -56,4 +64,168 @@ impl DataStore {
...
+            });
+
+        diesel::insert_into(join_table::table)
+            .values(SystemUpdateComponentUpdate {
+                system_update_id,
+                component_update_id: update.id(),
+            })
+            .on_conflict(join_table::all_columns)
+            .do_nothing()
+            .returning(SystemUpdateComponentUpdate::as_returning())
```
smklein, line 133: Same as the other case of `ON CONFLICT DO NOTHING RETURNING` . There appear to be a few more of these down below too.

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -5075,6 +5083,197 @@ async fn updates_refresh(
...
+#[endpoint {
+     method = GET,
+     path = "/v1/system/update/version",
+     tags = ["system"],
+}]
+async fn system_version(
+    rqctx: Arc<RequestContext<Arc<ServerContext>>>,
+) -> Result<HttpResponseOk<views::SystemVersion>, HttpError> {
+    let apictx = rqctx.context();
+    let _nexus = &apictx.nexus;
```
smklein, line 5096: Pretty sure this is already on your radar, but we should consider a non-stub call before merging

smklein (commented):
Looks like a great start, thanks again for pushing on this.

I have some feedback below, but if you're itching to merge, lemme know, and I can give a view with that lens. I think we should file issues for follow-ups, but we should feel free to merge-and-iterate.

common/src/sql/dbinit.sql
david-crespo: So the problem here was with the lookup macros. I could not figure out how to get a nice lookup-by-version with `lookup_resource!` without making `version` the primary key, because the way the macro is designed, it sounds like it has to be either by primary key or by name, and a non-PK version string is neither. I would love to find out that I understood this wrong, or that it's not that hard to change the macro (I looked but got spooked).

https://github.com/oxidecomputer/omicron/pull/2100/files#diff-6e518912406e7c5a535066ef28084be22fdbdae61895bab565eb656aa4682e01R698-R705

nexus/src/authz/api_resources.rs
```diff
@@ -963,6 +963,14 @@ authz_resource! {
     polar_snippet = FleetChild,
 }
 
+authz_resource! {
+    name = "SystemUpdate",
+    parent = "Fleet",
+    primary_key = Uuid,
```
david-crespo, line 969: just noticed this is wrong if version is the pk

common/src/sql/dbinit.sql
david-crespo: Hadn't thought about it, but yeah, sounds right. I just copied these from the example tree in [RFD 334](https://rfd.shared.oxide.computer/rfd/0334#_requirements).

nexus/db-model/src/semver_version.rs
david-crespo: Yeah, I think so. You could always use the struct constructor directly, though in our case that's pretty annoying, as we have two different wrappers. So we'd probably just add more constructor functions.

common/src/sql/dbinit.sql
david-crespo: Definitely a problem. I talked to Justin about it and mostly decided to punt. I don't really know how to do it. Storing the components as separate columns sounds like a huge pain in the ass, but maybe Diesel can help. Another approach would be to generate a stable, lexicographically (or numerically? probably not possible with the string meta bits) sortable representation to store alongside the original. I think it would require us to commit to limits on how big each number can be. so for example we could store `1.22.3` as `"00000001.00000022.00000003"` or something.

smklein: Clever; that padding approach could probably get us pretty far.

Diesel does have some support for embedded structures - that's what I'm doing all over the place with the `Identity` objects that are derived for assets / resources - but it's definitely a little more cumbersome than just a single value.

nexus/db-model/src/system_update.rs
david-crespo: This question makes me realize the parents don't need to be in the database at all. All you need is a mapping from `updateable_component_type -> parent_component_type`, and that can be hard-coded in software, just like the list updateable_component_type values is. Parent component types could themselves have a parent, like `Host -> Gimlet`.

```
Rack─┬─PSC─┬─RoT──Hubris Archive ABCD
     │     └─SP───Hubris Archive DCBA
     ├─Sidecar─┬─RoT──Hubris Archive ABCD
     │         └─SP───Hubris Archive FACE
     ├─Sidecar─┬─RoT──Hubris Archive ABCD
     │         └─SP───Hubris Archive FACE
     ├─Scrimlet─┬─RoT──Hubris Archive ABCD
     │          ├─SP───Hubris Archive BEEF
     │          └─Host─┬─Host Phase 1 Image 1234
     │                 ├─Host Phase 2 Image 1234
     │                 └─Omicron Image 5678
     ├─Scrimlet …
     ├─Gimlet …
     └─Gimlet─┬─RoT──Hubris Archive ABCD
              ├─SP───Hubris Archive BEEF
              └─Host─┬─Host Phase 1 Image 1234
                     ├─Host Phase 2 Image 1234
                     └─Omicron Image 5678
```

nexus/src/db/datastore/update.rs
david-crespo: Resolved in [6cdfe40](https://github.com/oxidecomputer/omicron/pull/2100/commits/6cdfe40954988d5e5c5e8d3dc33972b35df79075) by not doing `ON CONFLICT` at all. Because these are single row inserts, we can rely on unique constraints on id and version to produce the errors we want.

PR #2101, opened 2022-12-29
[docs][nexus] Document the process for adding new endpoints

smklein:
I've been committing to Omicron for a while, and even I forget all the steps involved here.

I figured I'd write this down after going through the process on https://github.com/oxidecomputer/omicron/pull/2099 - maybe this could also serve as a discussion point for "are some of these steps more complicated than they need to be?".

Feedback welcome

### DIFF
--- a/docs/adding-an-endpoint.adoc
+++ b/docs/adding-an-endpoint.adoc
@@ -16,44 +16,44 @@ this document should act as a jumping-off point.
 * Add endpoints for either the internal or external API
 ** xref:../nexus/src/external_api/http_entrypoints.rs[The External API] is customer-facing, and provides interfaces for both developers and operators
 ** xref:../nexus/src/internal_api/http_entrypoints.rs[The Internal API] is internal, and provides interfaces for services on the Oxide rack (such as the Sled Agent) to call
-** Register endpoints in the `register_endpoints` method (xref:https://github.com/oxidecomputer/omicron/blob/1dfe47c1b3122bc4f32a9c517cb31b1600581ea2/nexus/src/external_api/http_entrypoints.rs#L84[Example])
+** Register endpoints in the `register_endpoints` method (https://github.com/oxidecomputer/omicron/blob/1dfe47c1b3122bc4f32a9c517cb31b1600581ea2/nexus/src/external_api/http_entrypoints.rs#L84[Example])
 ** These endpoints typically call into the *Application* layer, and do not access the database directly
 * Inputs and Outputs
-** Input parameters are defined in xref:https://github.com/oxidecomputer/omicron/blob/main/nexus/types/src/external_api/params.rs[params.rs] (xref:https://github.com/oxidecomputer/omicron/blob/1dfe47c1b3122bc4f32a9c517cb31b1600581ea2/nexus/types/src/external_api/params.rs#L587-L601[Example])
-** Output views are defined in xref:https://github.com/oxidecomputer/omicron/blob/main/nexus/types/src/external_api/views.rs[views.rs] (xref:https://github.com/oxidecomputer/omicron/blob/1dfe47c1b3122bc4f32a9c517cb31b1600581ea2/nexus/types/src/external_api/views.rs#L270-L274[Example])
+** Input parameters are defined in https://github.com/oxidecomputer/omicron/blob/main/nexus/types/src/external_api/params.rs[params.rs] (https://github.com/oxidecomputer/omicron/blob/1dfe47c1b3122bc4f32a9c517cb31b1600581ea2/nexus/types/src/external_api/params.rs#L587-L601[Example])
+** Output views are defined in https://github.com/oxidecomputer/omicron/blob/main/nexus/types/src/external_api/views.rs[views.rs] (https://github.com/oxidecomputer/omicron/blob/1dfe47c1b3122bc4f32a9c517cb31b1600581ea2/nexus/types/src/external_api/views.rs#L270-L274[Example])
 
 === **Lookup & Authorization**
-* Declare a new resource-to-be-looked-up via `lookup_resource!` in xref:../nexus/src/db/lookup.rs[lookup.rs] (xref:https://github.com/oxidecomputer/omicron/blob/1dfe47c1b3122bc4f32a9c517cb31b1600581ea2/nexus/src/db/lookup.rs#L557-L564[Example])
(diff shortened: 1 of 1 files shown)


docs/adding-an-endpoint.adoc
```diff
@@ -0,0 +1,59 @@
...
+** Register endpoints in the `register_endpoints` method (https://github.com/oxidecomputer/omicron/blob/1dfe47c1b3122bc4f32a9c517cb31b1600581ea2/nexus/src/external_api/http_entrypoints.rs#L84[Example])
+** These endpoints typically call into the *Application* layer, and do not access the database directly
+* Inputs and Outputs
+** Input parameters are defined in https://github.com/oxidecomputer/omicron/blob/main/nexus/types/src/external_api/params.rs[params.rs] (https://github.com/oxidecomputer/omicron/blob/1dfe47c1b3122bc4f32a9c517cb31b1600581ea2/nexus/types/src/external_api/params.rs#L587-L601[Example])
+** Output views are defined in https://github.com/oxidecomputer/omicron/blob/main/nexus/types/src/external_api/views.rs[views.rs] (https://github.com/oxidecomputer/omicron/blob/1dfe47c1b3122bc4f32a9c517cb31b1600581ea2/nexus/types/src/external_api/views.rs#L270-L274[Example])
+
+=== **Lookup & Authorization**
+* Declare a new resource-to-be-looked-up via `lookup_resource!` in xref:../nexus/src/db/lookup.rs[lookup.rs] (https://github.com/oxidecomputer/omicron/blob/1dfe47c1b3122bc4f32a9c517cb31b1600581ea2/nexus/src/db/lookup.rs#L557-L564[Example])
```
smklein, line 26: Notes for myself:

- It's kinda weird that the `lookup_resource` and `authz_resource` are separate; I *think* they both need to be used together to be functional
- Populating the list of resources in `oso_generic.rs` feels like it could be automated
- Generating the lookup methods on `LookupPath` feels like it could be automated
- I believe these structures actually depend on a `db::model` object existing (after all, how else would they know what to fetch?), which is another layer of coupling that's subtle here. I wonder if there's a way to consolidate all these tightly coupled pieces a little better, or to actually make them more loosely coupled.

docs/adding-an-endpoint.adoc
```diff
@@ -0,0 +1,59 @@
...
+* Methods to send queries to the database are defined as part of the https://github.com/oxidecomputer/omicron/tree/1dfe47c1b3122bc4f32a9c517cb31b1600581ea2/nexus/src/db/datastore[datastore] structure (https://github.com/oxidecomputer/omicron/blob/1dfe47c1b3122bc4f32a9c517cb31b1600581ea2/nexus/src/db/datastore/ip_pool.rs[Example])
+
+=== **Testing**
+
+* Authorization
+** There exists a https://github.com/oxidecomputer/omicron/blob/main/nexus/src/authz/policy_test[policy test] which compares all Oso objects against an expected policy. New resources are usually added to https://github.com/oxidecomputer/omicron/blob/main/nexus/src/authz/policy_test/resources.rs[resources.rs] to get coverage.
+* openapi
+** Nexus generates a new openapi spec from the dropshot endpoints. If you modify endpoints, you'll usually need to `EXPECTORATE=overwrite cargo test` the following targets...
+*** External API: `test_nexus_openapi`
+*** Internal API: `test_nexus_openapi_internal`
```
smklein, line 54: These tests are good, in that they're getting so much coverage, but they also kinda suck from a developer UX point of view. Testing any API changes feels like it requires re-running `cargo test` with and without `EXPECTORATE` a *bunch* of times, especially since changing the openapi file causes different code generation, which means the next call to `cargo test` might not actually build...

andrewjstone (approved):
Thank you so much for doing this @smklein

david-crespo (approved):

docs/adding-an-endpoint.adoc
luqmana: Was it faster to just run it manually without going via the tests, e.g.

```console
$ cargo run -p omicron-nexus --bin nexus -- -I nexus/examples/config.toml > openapi/nexus-internal.json
$ cargo run -p omicron-nexus --bin nexus -- -O nexus/examples/config.toml > openapi/nexus.json
$ cargo run -p omicron-sled-agent --bin sled-agent -- openapi > openapi/sled-agent.json
```

david-crespo: Definitely seems to build fewer packages, at least. I will be trying this. Normally I run

```
EXPECTORATE=overwrite cargo t integration_tests::commands::test_nexus_openapi
```

I only ever work on the external API.

smklein: Added both the test and non-test options

davepacheco:
This is great.  A bunch of this winds up being basically Nexus architecture.  I wonder if it should live in nexus/src/lib.rs (or some other top-level place directly in the source)?

docs/adding-an-endpoint.adoc
```diff
@@ -0,0 +1,71 @@
...
+** Remember to make the object visible; this is usually defined in xref:../nexus/db-model/src/lib.rs[lib.rs] (https://github.com/oxidecomputer/omicron/blob/1dfe47c1b3122bc4f32a9c517cb31b1600581ea2/nexus/db-model/src/lib.rs#L102[Example]).
+* Methods to send queries to the database are defined as part of the https://github.com/oxidecomputer/omicron/tree/1dfe47c1b3122bc4f32a9c517cb31b1600581ea2/nexus/src/db/datastore[datastore] structure (https://github.com/oxidecomputer/omicron/blob/1dfe47c1b3122bc4f32a9c517cb31b1600581ea2/nexus/src/db/datastore/ip_pool.rs[Example])
+
+=== **Testing**
+
+* Authorization
+** There exists a https://github.com/oxidecomputer/omicron/blob/main/nexus/src/authz/policy_test[policy test] which compares all Oso objects against an expected policy. New resources are usually added to https://github.com/oxidecomputer/omicron/blob/main/nexus/src/authz/policy_test/resources.rs[resources.rs] to get coverage.
+* openapi
+** Nexus generates a new openapi spec from the dropshot endpoints. If you modify endpoints, you'll need to update openapi JSON files.
+*** The following commands may be used to update APIs:
```
davepacheco, line 53: FYI, I think this overlaps with [this section of the README](https://github.com/oxidecomputer/omicron/blob/main/README.adoc#generated-service-clients-and-updating), which has some discussion on how to make changes that affect both a server and client inside Omicron.

PR #2103, opened 2023-01-03
MGS: Don't use ignition results to decide whether to contact an SP

jgallagher:
This PR changes both the "get state of an SP" and the "get state of all SPs" to, primarily, no longer first ask the ignition controller whether the SP on skip talking to the SP if ignition indicates it is not. Instead, this changes `SpInfo` such that it can now indicate ignition status/failure and SP status/failure independently.

Additionally for the "get state of all SPs" endpoint, because we no longer use ignition as a filter, we no longer need to take a timeout; instead, we send requests to all SPs concurrently and wait for them all to respond with our normal rules (i.e., our configured number of attempts / timeout per attempt for individual requests).


gateway/src/management_switch.rs
```diff
@@ -203,14 +203,28 @@ impl ManagementSwitch {
...
+        // every port.
+        self.sockets.get(&port).unwrap()
+    }
+
+    /// Get an iterator providing the ID and handle to communicate with every SP
+    /// we know about.
+    ///
+    /// This function can only fail if we have not yet completed discovery (and
+    /// therefore can't map our switch ports to SP identities).
+    pub(crate) fn all_sps(
```
jgallagher, line 218: This method is the change that I find maybe the most confusing/dubious in this PR: this returns an iterator of all SPs we _potentially_ know about (i.e., every switch port we're configured to use to communicate with an SP). In practice on real hardware this will always return an iterator with 35 items, even if some of the slots are currently unoccupied. This means that if any slot is unoccupied, the "get all SP state" endpoint will not return faster than it takes for the requests to that slot to timeout. This is maybe a bigger conversation than just this PR, but I'd like to address this soon in a followup.

andrewjstone: I think the question here is "How do we know when we shouldn't check all SPs?". Unfortunately, I also think the answer is we can never know, since we can't rely on ignition to tell us. Given that, I don't see how we workaround this. In most cases we expect racks to be full, but there will always be failures and upgrades, and we will timeout some large percentage of the the time. In that case, I think we may have to just accept that and make the timeout as short as possible.

andrewjstone (approved):
Looks great!

gateway/src/management_switch.rs
jgallagher: Fundamentally I agree with you; however, if we have a permanently (or effectively permanently) underpopulated rack, it's very unfortunate that this will always time out. I have similar concerns about detecting removal / replacement / swapping of a gimlet; that is definitely not handled correctly right now.

Thinking out loud, maybe each `SingleSp` should remember the last time it heard from the SP it thinks it should be talking to (which in general should be less than the interval on which we send periodic "discover" packets, which is currently < 60 seconds), and if that time is sufficiently long + ignition says there's no gimlet present we stop waiting for it?

andrewjstone: That sounds reasonable, but only if there is some guaranteed communication from the SP.

PR #2106, opened 2023-01-04
[installinator] split up artifact server

sunshowers:
Move this away from being part of the wicketd artifact server for two
reasons:

1. The wicketd server is going to listen on the technician port, while the artifact server needs to listen on the bootstrap network. We don't want to provide full access to the technician interface on the bootstrap network.

2. The sled agent will also be able to serve artifacts; we'd like to
   make it so that the installinator can just reach out to every peer on
   the boostrap network without having to care about the differences
   between the sled agent and wicketd.

In the future, we'll make it so that:

* the wicketd technician interface can upload artifacts to memory, which will then
  get served by the wicketd artifact server

* sled-agent will be able to serve artifacts directly from disk.


wicketd/src/lib.rs
```diff
@@ -60,22 +59,50 @@ pub async fn run_server(config: Config, args: Args) -> Result<(), String> {
...
+    loop {
+        tokio::select! {
+            res = &mut wicketd_server_fut, if !wicketd_server_done => {
+                wicketd_server_done = true;
+                res.map_err(|err| format!("running wicketd server: {}", err))?;
+            }
+            res = &mut artifact_server_fut, if !artifact_server_done => {
+                artifact_server_done = true;
+                res.map_err(|err| format!("running artifact server: {:?}", err))?;
+            }
```
sunshowers, line 102: Q: is this the right behavior? Or should the wicketd server take priority and if it closes, the entire loop exits?

wicketd/src/bin/wicketd.rs
```diff
@@ -23,6 +23,9 @@ enum Args {
 
         #[clap(short, long, action)]
         address: SocketAddrV6,
+
+        #[clap(long, action)]
+        artifact_address: SocketAddrV6,
```
smklein, line 28: Could we document the different between these two? It's obvious in the context of this PR, less obvious when read in isolation

wicketd/src/lib.rs
smklein: We don't have any reason for either of these servers to "stop running", while expecting that the other needs to continue operating, right?

If that's the case, I *think* it would be fine to bail out on error if either of them fail, since that's sorta a case of either:
1. We're terminating, or
2. Something has gone horribly wrong

smklein (approved):

wicketd/src/lib.rs
sunshowers: Hmm, so do we expect wicketd to generally keep running forever? I guess upgrades and such would be handled by just terminating the process?

installinator-artifactd/src/http_entrypoints.rs
```diff
@@ -0,0 +1,50 @@
...
+    endpoint, ApiDescription, FreeformBody, HttpError, HttpResponseOk, Path,
+    RequestContext,
+};
+use hyper::Body;
+
+use crate::{context::ServerContext, store::ArtifactId};
+
+type ArtifactServerApiDesc = ApiDescription<ServerContext>;
+
+/// Return a description of the wicketd api for use in generating an OpenAPI spec
```
andrewjstone, line 19: s/wicketd/artifact server/ ?

installinator/src/peers.rs
```diff
@@ -359,7 +359,7 @@ struct ArtifactClient {
     // TODO: this currently uses a wicketd client. However, because the standard wicketd server is
```
andrewjstone, line 359: Can probably remove this todo at this point.

wicketd/src/lib.rs
andrewjstone: > We don't have any reason for either of these servers to "stop running", while expecting that the other needs to continue operating, right?
> 
> If that's the case, I _think_ it would be fine to bail out on error if either of them fail, since that's sorta a case of either:
> 
> 1. We're terminating, or
> 2. Something has gone horribly wrong

I think this sounds reasonable.



> Hmm, so do we expect wicketd to generally keep running forever? I guess upgrades and such would be handled by just terminating the process?

Yes, wicketd should run indefinitely, as it's also the emergency escape hatch if the control-plane stops working. The only time it should stop is during upgrades, power outages, or the aformentioned horrible bugs.

andrewjstone (approved):
LGTM.

installinator-artifactd/src/server.rs
```diff
@@ -0,0 +1,57 @@
...
+
+impl ArtifactServer {
+    /// Creates a new artifact server with the given address.
+    pub fn new(address: SocketAddrV6, log: &slog::Logger) -> Self {
+        let log = log.new(slog::o!("component" => "installinator artifactd"));
+        let store = ArtifactStore::new(&log);
+        Self { address, log, store }
+    }
+
+    /// Starts the artifact server.
```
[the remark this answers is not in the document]
sunshowers: Done.

wicketd/src/bin/wicketd.rs
sunshowers: Done

installinator/src/peers.rs
sunshowers: Done.

installinator-artifactd/src/http_entrypoints.rs
sunshowers: Done.

wicketd/src/lib.rs
sunshowers: Thanks, I've made it so that if either server exits, the process exits.

sunshowers: Note that there's a small behavior change I've introduced here:

Previously, if the wicketd server exited without an error (`Ok(())`), we'd return `Ok(())` and exit with code 0.

With this change that's no longer the case -- the server always exits with a non-zero code.

PR #2109, opened 2023-01-04
fix `test_setup_database_bad_listen_url()`

### LINKED ISSUE #2108: `test_setup_database_bad_listen_url()` builds an invalid command and can racily fail
I saw this locally when using `cargo nextest run` to execute the full test suite:

```
--- STDERR:                                               omicron-test-utils dev::db::test::test_setup_database_bad_listen_url ---
will run: bash start-single-node --insecure --http-addr=:0 -c sleep 60 --store=path=/tmp/.tmphSogYR/data,ballast-size=0 --listen-addr 127.0.0.1:0 --listening-url-file /tmp/.tmphSogYR/listen-url
environment:
... snip ...
bash: start-single-node: No such file or directory
error: Exited { exit_code: 127 }
thread 'dev::db::test::test_setup_database_bad_listen_url' panicked at 'unexpected error trying to start database: cockroach exited unexpectedly with status 127 (see error output above)', test-utils/src/dev/db.rs:1196:13
```

but running the test in isolation passes. I think what's happening is:

1. `test_setup_database_bad_listen_url()` is testing our `start()` method, not cockroachdb specifically, so it actually tries to start a `bash -c sleep 60`: https://github.com/oxidecomputer/omicron/blob/main/test-utils/src/dev/db.rs#L1175-L1176
2. However, `CockroachStarterBuilder::new_with_cmd()` assumes it's always running cockroachdb, and inserts other args before the `-c sleep 60`: https://github.com/oxidecomputer/omicron/blob/main/test-utils/src/dev/db.rs#L126-L129
3. This produces the invalid `bash start-single-node ...` command I see in the failure above
4. `start()` then races: if bash exits with the error fast enough, we'll hit https://github.com/oxidecomputer/omicron/blob/main/test-utils/src/dev/db.rs#L362-L364 and fail the test (which I see when running the full test suite). If the test wins the race and checks the listen url file before bash exits (https://github.com/oxidecomputer/omicron/blob/main/test-utils/src/dev/db.rs#L374), we'll see the expected error and pass the test

davepacheco:
Fixes #2108.

Prior to this, the public constructor was `CockroachStarterBuilder::new()`.  That used `CockroachStarterBuilder::new_with_cmd(COCKROACHDB_BIN)` to do most of the work.  A couple of tests (including the one that's broken) used `CockroachStarterBuilder::new_with_cmd(some_other_cmd)`.  The goal was to use as much of the same machinery as possible (to test it) while overriding the command with one that would make it easier to trigger certain failure modes.

Prior to this, `new_with_cmd()` did three things:

1. initializes the fields of `CockroachStarterBuilder` (which involves a few non-trivial decisions)
2. configures the environment
3. configures the command-line arguments

It's step 3 that's making things difficult for `test_setup_database_bad_listen_url()`.  Steps 2 and 3 are not needed by either of the tests that call `new_with_cmd()` directly.  So I've moved those steps to `CockroachStarterBuilder::new()` so that `new_with_cmd()` only initializes the struct fields.

I also renamed `new_with_cmd()` to `new_raw()` to avoid conveying the idea that it was just like `new()`, but with an overridden command.  I'm looking to convey "the caller is responsible for various details that the regular `new()` takes care of".  I'm open to better names.

### DIFF
--- a/test-utils/src/dev/db.rs
+++ b/test-utils/src/dev/db.rs
@@ -1176,7 +1176,8 @@ mod test {
     async fn test_setup_database_bad_listen_url() {
         // We don't need to actually run Cockroach for this test, and it's
         // simpler (and faster) if we don't.  But we do need something that
-        // won't exit before we get a chance to trigger an error.
+        // won't exit before we get a chance to trigger an error and that can
+        // also accept the extra arguments that the builder will provide.
         let mut builder = CockroachStarterBuilder::new_raw("bash");
         builder.arg("-c").arg("sleep 60");
         let starter = builder.build().unwrap();
--- a/test-utils/src/dev/db.rs
+++ b/test-utils/src/dev/db.rs
@@ -83,19 +83,7 @@ pub struct CockroachStarterBuilder {
 
 impl CockroachStarterBuilder {
     pub fn new() -> CockroachStarterBuilder {
-        CockroachStarterBuilder::new_with_cmd(COCKROACHDB_BIN)
-    }
-
-    fn new_with_cmd(cmd: &str) -> CockroachStarterBuilder {
-        let mut builder = CockroachStarterBuilder {
-            store_dir: None,
-            listen_port: COCKROACHDB_DEFAULT_LISTEN_PORT,
-            env: BTreeMap::new(),
-            args: vec![String::from(cmd)],
-            cmd_builder: tokio::process::Command::new(cmd),
-            start_timeout: COCKROACHDB_START_TIMEOUT_DEFAULT,
-            redirect_stdio: false,
-        };
+        let mut builder = CockroachStarterBuilder::new_raw(COCKROACHDB_BIN);
 
         // Copy the current set of environment variables.  We could instead
         // allow the default behavior of inheriting the current process
@@ -128,6 +116,24 @@ impl CockroachStarterBuilder {
             .arg("--insecure")
             .arg("--http-addr=:0");
         builder
+    }
+
+    /// Helper for constructing a `CockroachStarterBuilder` that runs a specific
+    /// command instead of the usual `cockroach` binary
+    ///
+    /// This is used by `new()` as a starting point.  It's also used by the
+    /// tests to trigger failure modes that would be hard to reproduce with
+    /// `cockroach` itself.
+    fn new_raw(cmd: &str) -> CockroachStarterBuilder {
+        CockroachStarterBuilder {
+            store_dir: None,
+            listen_port: COCKROACHDB_DEFAULT_LISTEN_PORT,
+            env: BTreeMap::new(),
+            args: vec![String::from(cmd)],
+            cmd_builder: tokio::process::Command::new(cmd),
+            start_timeout: COCKROACHDB_START_TIMEOUT_DEFAULT,
+            redirect_stdio: false,
+        }
     }
 
     /// Redirect stdout and stderr for the "cockroach" process to files within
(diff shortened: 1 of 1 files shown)


jgallagher (approved):
I have no objection to `new_raw`. It's not the most descriptive, but the function is private and `_raw` really implies "you should look at this in detail if you think you need it", which seems reasonable.

I confirmed this fixes the failure I was seeing. If we somehow get preempted for longer than 60 seconds between starting the subprocess and checking the exit status, we'll still fail, but that seems fine. Do we want to try to craft a better error for that (unlikely) case? If I turn the sleep timer down to 1 second and then manually inject a pause, this is the error we get:

```
will run: bash -c sleep 1 --store=path=/tmp/.tmpmGLaHg/data,ballast-size=0 --listen-addr 127.0.0.1:0 --listening-url-file /tmp/.tmpmGLaHg/listen-url
environment:
error: Exited { exit_code: 0 }
thread 'dev::db::test::test_setup_database_bad_listen_url' panicked at 'unexpected error trying to start database: cockroach exited unexpectedly with status 0 (see error output above)', test-utils/src/dev/db.rs:1204:13
```

davepacheco:
We wordsmithed this offline.

From c07e4bc, a successful run of this test looks like:

```
$ cargo test -p omicron-test-utils -- test_setup_database_bad_listen_url --nocapture
    Finished test [unoptimized + debuginfo] target(s) in 0.36s
     Running unittests src/lib.rs (target/debug/deps/omicron_test_utils-91f1571d01d3dfa8)

running 1 test
will run: bash -c true --store=path=/dangerzone/omicron_tmp/.tmpzmb3GH/data,ballast-size=0 --listen-addr 127.0.0.1:0 --listening-url-file /dangerzone/omicron_tmp/.tmpzmb3GH/listen-url
environment:
error: Unknown { source: checking listen file "/dangerzone/omicron_tmp/.tmpzmb3GH/listen-url"

Caused by:
    Is a directory (os error 21) }
error message was: checking listen file "/dangerzone/omicron_tmp/.tmpzmb3GH/listen-url": Is a directory (os error 21)
test dev::db::test::test_setup_database_bad_listen_url ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 15 filtered out; finished in 0.14s
```

To test the case @jgallagher mentioned, I replaced `sleep 60` with `true`.  It took a couple of runs, but I got:

```
$ cargo test -p omicron-test-utils -- test_setup_database_bad_listen_url --nocapture
    Finished test [unoptimized + debuginfo] target(s) in 0.57s
     Running unittests src/lib.rs (target/debug/deps/omicron_test_utils-91f1571d01d3dfa8)

running 1 test
will run: bash -c true --store=path=/dangerzone/omicron_tmp/.tmp3ZyfmW/data,ballast-size=0 --listen-addr 127.0.0.1:0 --listening-url-file /dangerzone/omicron_tmp/.tmp3ZyfmW/listen-url
environment:
error: Exited { exit_code: 0 }
thread 'dev::db::test::test_setup_database_bad_listen_url' panicked at 'unexpected error from CockroachStarter::start(): expected error checking listen file, found cockroach exited unexpectedly with status 0 (see error output above)', test-utils/src/dev/db.rs:1203:13
test dev::db::test::test_setup_database_bad_listen_url ... FAILED

failures:

failures:
    dev::db::test::test_setup_database_bad_listen_url

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 15 filtered out; finished in 0.61s

error: test failed, to rerun pass `-p omicron-test-utils --lib`
```

which I think we agreed is clear enough.

PR #2110, opened 2023-01-04
[wicketd] add an ArtifactGetter abstraction, use it to upload artifacts over technician interface

sunshowers:
Introduce:

1. An `ArtifactGetter` abstraction that represents a way to get data from the artifact store.
2. `WicketdArtifactStore`, which is shared by the wicketd and artifact servers.

`WicketdArtifactStore` implements `ArtifactGetter` and is used to:

* Upload artifacts over the technician interface
* Download artifacts via the bootstrap network.

In the future, sled-agent will have its own `ArtifactGetter` implementation that serves images directly from disk.

Tested this by running

```
cargo run -p wicketd -- run --address '[::1]:8000' --artifact-address '[::1]:14000' examples/config.toml
```

Then, in another terminal, creating a 1GB file (`head -c 1000000000 /dev/zero > example.bin`) and uploading it with:

```
curl -X PUT --data-binary "@example.bin" 'http://[::1]:8000/artifacts/foo/0.1.0'
```

Finally, downloading the artifact with:

```
cargo run -p installinator -- install --mechanism 'list:[::1]:14000' foo:0.1.0 downloaded.bin
```

And checking that they're the same with:

```
$ b3sum example.bin downloaded.bin
55c6dac98fbc9a388f619f5f4ffc4c9fdd3eb37eab48afd68b65da90ef3070b1  example.bin
55c6dac98fbc9a388f619f5f4ffc4c9fdd3eb37eab48afd68b65da90ef3070b1  downloaded.bin
```

Depends on #2106.


andrewjstone (approved):
Looks good @sunshowers.

PR #2111, opened 2023-01-05
Update pagination to support v1 API patterns

zephraph:
This is a WIP effort to update pagination to work with v1 endpoints.

### DIFF
--- a/common/src/api/external/http_pagination.rs
+++ b/common/src/api/external/http_pagination.rs
@@ -42,6 +42,7 @@
 
 use crate::api::external::DataPageParams;
 use crate::api::external::Name;
+use crate::api::external::NameOrId;
 use crate::api::external::ObjectIdentity;
 use crate::api::external::PaginationOrder;
 use dropshot::HttpError;
@@ -150,11 +151,12 @@ pub fn marker_for_id<S, T: ObjectIdentity>(_: &S, t: &T) -> Uuid {
 pub fn marker_for_name_or_id<T: ObjectIdentity>(
     scan: &ScanByNameOrId,
     item: &T,
-) -> NameOrIdMarker {
+) -> NameOrId {
     let identity = item.identity();
-    match pagination_field_for_scan_params(scan) {
-        PagField::Name => NameOrIdMarker::Name(identity.name.clone()),
-        PagField::Id => NameOrIdMarker::Id(identity.id),
+    match scan.sort_by {
+        NameOrIdSortMode::NameAscending => identity.name.clone().into(),
+        NameOrIdSortMode::NameDescending => identity.name.clone().into(),
+        NameOrIdSortMode::IdAscending => identity.id.into(),
     }
 }
 
@@ -177,10 +179,6 @@ where
 
 /// Given a request and pagination parameters, return a [`DataPageParams`]
 /// describing the current page of results to return
-///
-/// This implementation is used for `ScanByName` and `ScanById`.  See
-/// [`data_page_params_nameid_name`] and [`data_page_params_nameid_id`] for
-/// variants that can be used for `ScanByNameOrId`.
 pub fn data_page_params_for<'a, S, C>(
     rqctx: &'a Arc<RequestContext<C>>,
     pag_params: &'a PaginationParams<S, PageSelector<S, S::MarkerValue>>,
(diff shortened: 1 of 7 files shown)


nexus/src/external_api/http_entrypoints.rs
```diff
@@ -3074,25 +3067,24 @@ async fn instance_stop(
 async fn instance_serial_console_v1(
     rqctx: Arc<RequestContext<Arc<ServerContext>>>,
     path_params: Path<params::InstancePath>,
-    query_params: Query<params::InstanceSerialConsole>,
+    query_params: Query<params::InstanceSerialConsoleRequest>,
+    selector_params: Query<params::OptionalProjectSelector>,
```
zephraph, line 3071: So, this is a bit weird. It's implemented on the same principle as described in the docs here: https://docs.rs/dropshot/latest/dropshot/#advanced-usage-notes. Even though this endpoint doesn't do pagination, we can't flatten the `InstanceSerialConsoleRequest` and `ProjectSelector` structs into a single struct due to https://github.com/nox/serde_urlencoded/issues/33.

common/src/api/external/http_pagination.rs
```diff
@@ -298,15 +296,45 @@ impl ScanParams for ScanById {
 // We include this now primarily to exercise the interface for doing so.
 
 /// Query parameters for pagination by name or id
-pub type PaginatedByNameOrId =
-    PaginationParams<ScanByNameOrId, PageSelectorByNameOrId>;
+pub type PaginatedByNameOrId<Selector> = PaginationParams<
```
ahl, line 299: I think you could do

```suggestion
pub type PaginatedByNameOrId<Selector = ()> = PaginationParams<
```

to avoid changing uses of this that don't require additional selectors

common/src/api/external/http_pagination.rs
```diff
@@ -298,15 +296,45 @@ impl ScanParams for ScanById {
...
+    ScanByNameOrId<Selector>,
+    PageSelectorByNameOrId<Selector>,
```
ahl, line 301: So in general, it seems like the first parameter to PaginationParams is going to want to be the tuple of NameOrIds needed to find the item in question. The second parameter is going to want to be just the ID of the top level container. That is to say, I think we should only be doing name lookups for the first query; subsequent pages should just have the ID of the container that we're iterating through.

ahl (commented):
This deserves far deeper of a look than I've given it, but I wanted to share some high level feedback rather than sitting on it. I will look more tonight and tomorrow.

common/src/api/external/http_pagination.rs
zephraph: I tried this, but it's rather difficult and made the handlers a lot messier. 

Right now we take the query parameters and resolve them to a lookup and pass that to nexus. My understanding of this approach is that we'll need to resolve the lookup to an ID inside the handler as well as passing it along to be further resolved by Nexus (on the first page). That also means we have to have logic to check if it's a selector or id (or if the selector is an id) so we're not doing wasteful lookups. My attempts at this made both the handler much longer and harder to understand as well as increased the complexity of the types around pagination.

I agree that this would be an ideal approach. I also think it's an optimization that we can defer. From a user documentation perspective the same caveats would apply when looking up a resource by name to paginate as would when just looking up a name in general.

common/src/api/external/http_pagination.rs
zephraph: done

common/src/api/external/http_pagination.rs
ahl: > I tried this, but it's rather difficult and made the handlers a lot messier.

That's disappointing; I could imagine it being a way to extract common patterns rather than to complicate each handler.

> Right now we take the query parameters and resolve them to a lookup and pass that to nexus. My understanding of this approach is that we'll need to resolve the lookup to an ID inside the handler as well as passing it along to be further resolved by Nexus (on the first page).

Not necessarily. It could be as simple as having the nexus function that performs the resolution return the resolved ID.

> That also means we have to have logic to check if it's a selector or id (or if the selector is an id) so we're not doing wasteful lookups. My attempts at this made both the handler much longer and harder to understand as well as increased the complexity of the types around pagination.

It could literally be the same types if, for some reason, that's simpler: it could just be that for the page selector **only** the simplest ID form is encoded in the next_token.

> I agree that this would be an ideal approach. I also think it's an optimization that we can defer.

I think of this not as a performance optimization, but a simplification of the handlers. If it's not, agreed, we can defer or just not pursue it.

> From a user documentation perspective the same caveats would apply when looking up a resource by name to paginate as would when just looking up a name in general.

Further caveats apply: with this approach a rename of any intermediate collection will cause the iteration to fail whereas using a resolved ID for the container means that renames (e.g. of the parent) won't interfere.

zephraph: > Not necessarily. It could be as simple as having the nexus function that performs the resolution return the resolved ID.

Yeah, I was thinking the same. We turn the selector into a lookup and pass that along to the function that lists the resources. I tried returning a tuple of the resource list and ID but I had some issues with the return type no longer being a result list. Maybe it just needed to be a result of a tuple where one member was a result list? Uncertain.

ahl (approved):
I'm Adam Leventhal and I approve this PR

PR #2112, opened 2023-01-05
MGS: Remove `Communicator`

jgallagher:
This was a helper type that used to do more, but has evolved into little more than a glorified wrapper around `ManagementSwitch`. This commit removes `Communicator` entirely in favor of using `ManagementSwitch` directly. It has a handful of related cleanup changes: renaming some methods for clarity, making some methods private, and a minor amount of error handling cleanup.

Builds on #2103 and should be merged after it.


gateway/src/http_entrypoints.rs
```diff
@@ -395,8 +392,14 @@ async fn sp_list(
                 sp_state.insert(id, state);
             }
             // Result from our ignition controller.
-            Either::Right(ignition_state) => {
-                bulk_ignition_state = Some(ignition_state);
+            Either::Right(ignition_state_result) => {
+                // If `buulk_ignition_state` succeeded, it returns an iterator
```
andrewjstone, line 396: typo: buulk

andrewjstone (approved):

PR #2113, opened 2023-01-06
Fix ip pool policy for instance allocation

smklein:
Fixes a regression introduced by https://github.com/oxidecomputer/omicron/pull/2056

Updates polar policy to allow instance creation from regular Silo users, and adds a regression test

### DIFF
--- a/nexus/src/authz/omicron.polar
+++ b/nexus/src/authz/omicron.polar
@@ -373,9 +373,10 @@ resource IpPoolList {
 has_relation(fleet: Fleet, "parent_fleet", ip_pool_list: IpPoolList)
 	if ip_pool_list.fleet = fleet;
 
-# Any authenticated user can create a child of a provided IP Pools.
+# Any authenticated user can create a child of a provided IP Pool.
 # This is necessary to use the pools when provisioning instances.
-has_permission(_actor: AuthenticatedActor, "create_child", _ip_pool: IpPool);
+has_permission(actor: AuthenticatedActor, "create_child", ip_pool: IpPool)
+  if actor.silo.fleet = ip_pool.fleet;
 
 # Describes the policy for accessing "/system/images" (in the API)
 resource GlobalImageList {
--- a/nexus/src/authz/omicron.polar
+++ b/nexus/src/authz/omicron.polar
@@ -376,7 +376,7 @@ has_relation(fleet: Fleet, "parent_fleet", ip_pool_list: IpPoolList)
 # Any authenticated user can create a child of a provided IP Pool.
 # This is necessary to use the pools when provisioning instances.
 has_permission(actor: AuthenticatedActor, "create_child", ip_pool: IpPool)
-  if actor.silo.fleet = ip_pool.fleet;
+	if actor.silo.unwrap().fleet = ip_pool.fleet;
 
 # Describes the policy for accessing "/system/images" (in the API)
 resource GlobalImageList {
--- a/nexus/src/authz/omicron.polar
+++ b/nexus/src/authz/omicron.polar
@@ -376,7 +376,7 @@ has_relation(fleet: Fleet, "parent_fleet", ip_pool_list: IpPoolList)
 # Any authenticated user can create a child of a provided IP Pool.
 # This is necessary to use the pools when provisioning instances.
 has_permission(actor: AuthenticatedActor, "create_child", ip_pool: IpPool)
-	if actor.silo.unwrap().fleet = ip_pool.fleet;
+	if silo in actor.silo and silo.fleet = ip_pool.fleet;
 
 # Describes the policy for accessing "/system/images" (in the API)
 resource GlobalImageList {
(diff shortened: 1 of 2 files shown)


nexus/tests/integration_tests/instances.rs
```diff
@@ -2889,6 +2894,98 @@ async fn test_instance_ephemeral_ip_from_correct_pool(
...
+        "/organizations/{}/projects/{}/instances",
+        ORGANIZATION_NAME, PROJECT_NAME
+    );
+    NexusRequest::objects_post(client, &url_instances, &instance_params)
+        .authn_as(AuthnMode::SiloUser(user_id))
+        .execute()
+        .await
+        .expect("Failed to create instance")
+        .parsed_body::<Instance>()
+        .expect("Failed to parse instance");
```
smklein, line 2986: Before updating the polar policy, this failed.

The instance creation pathway checks for `authz::Action::CreateChild` on the target IP pool, which works fine for `AuthnMode::PrivilegedUser`, but fails for a silo user.

For example: https://github.com/oxidecomputer/omicron/blob/75bd90550e63940bbed7ef40ee21e18fa1b628dc/nexus/src/app/sagas/instance_create.rs#L607

nexus/tests/integration_tests/instances.rs
```diff
@@ -2889,6 +2894,98 @@ async fn test_instance_ephemeral_ip_from_correct_pool(
...
+        user_id,
+        AuthnMode::PrivilegedUser,
+    )
+    .await;
+
+    // Populate IP Pool
+    populate_ip_pool(&client, "default", None).await;
+
+    // Create test organization and projects.
+    NexusRequest::objects_post(
```
smklein, line 2925: Not *really* part of this PR, but it would be neat if the `resource_helpers` used a builder pattern to make these calls a little easier to customize. They currently hard-code "privileged user", which means I need to do the request myself.

luqmana: Let's open a bug for that?

nexus/src/authz/omicron.polar
```diff
@@ -373,6 +373,10 @@ resource IpPoolList {
 has_relation(fleet: Fleet, "parent_fleet", ip_pool_list: IpPoolList)
 	if ip_pool_list.fleet = fleet;
 
+# Any authenticated user can create a child of a provided IP Pools.
```
plotnick, line 376:
```suggestion
# Any authenticated user can create a child of a provided IP pool.
```

luqmana (approved):
Thanks, with this I can create an instance while logged in as a silo user (alice)

nexus/src/authz/omicron.polar
```diff
@@ -373,6 +373,10 @@ resource IpPoolList {
...
+# This is necessary to use the pools when provisioning instances.
+has_permission(_actor: AuthenticatedActor, "create_child", _ip_pool: IpPool);
```
plotnick, line 378: I don't think it matters much right now since we don't really support fleets yet, but for consistency with other policy rules I think there should probably be an additional restriction here:
```
has_permission(actor: AuthenticatedActor, "create_child", ip_pool: IpPool)
        if actor.silo.fleet = ip_pool.fleet;
```

plotnick (approved):
Nice catch, and thanks for the regression test!

nexus/src/authz/omicron.polar
smklein: Done

nexus/src/authz/omicron.polar
smklein: Done

nexus/tests/integration_tests/instances.rs
smklein: https://github.com/oxidecomputer/omicron/issues/2114

nexus/src/authz/omicron.polar
smklein: Hrm, this appears to actually break the parsing of the oso file:

```
    error_message_internal: failed to compute authorization: trace (most recent evaluation last):
      009: allow(oso::host::value::PolarValue TYPE `oso::host::value::PolarValue`, oso::host::value::PolarValue TYPE `oso::host::value::PolarValue`, oso::host::value::PolarValue TYPE `oso::host::value::PolarValue`)
      008: actor.authenticated and
        has_permission(actor.authn_actor.unwrap(), action.to_perm(), resource)
        in rule allow at line 21, column 5
      007: actor.authn_actor
        in rule allow at line 22, column 20
      006: actor.authn_actor.unwrap()
        in rule allow at line 22, column 20
      005: action.to_perm()
        in rule allow at line 22, column 48
      004: has_permission(actor.authn_actor.unwrap(), action.to_perm(), resource)
        in rule allow at line 22, column 5
      003: actor.silo.fleet = ip_pool.fleet
        in rule has_permission at line 379, column 6
      002: actor.silo.fleet = ip_pool.fleet
        in rule has_permission at line 379, column 6
      001: actor.silo.fleet
        in rule has_permission at line 379, column 6
     
    Application error: Attribute fleet not found on type Option. at line 379, column 6:
        379:   if actor.silo.fleet = ip_pool.fleet;
```

smklein: I'm unwrapping in a11fc2689fd8c7c262c64710c4fd712751183f91 , though I'm open to a safer way to do this in polar.

PR #2118, opened 2023-01-06
update Steno to crates.io dep

### LINKED ISSUE #2117: omicron is depending on steno via git again
Under #1548 / #1549 I updated Omicron to get Steno from crates.io, with the assumption that we'd publish releases as needed (and I made that part easy under oxidecomputer/steno#37).  This was undone and then redone a few times -- looks like most recently with #2095.  The problem with using a Git dependency here is that at least at one point, the Git dependencies make it really hard to make breaking changes to Steno even when the Git dependencies specify a particular commit.  I think the problem related to the multiple dependencies on Steno that exist in Omicron by virtue of Omicron depending on other things that depend back on other parts of Omicron.  (Maybe this is better now with workspace dependencies?  But I don't think so.  I think the problem last time was related to #1537 -- something in Omicron depended on something in Propolis, which depends back on omicron-common via Git sha, so it would get the old Steno even after Omicron was updated to use the crates.io version.)

I imagine this is probably blocked on me releasing a new version of Steno, which I will try to do shortly.

davepacheco:
Fixes #2117

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5946,8 +5946,9 @@ checksum = "a2eb9349b6444b326872e140eb1cf5e7c522154d69e7a0ffb0fb81c06b37543f"
 
 [[package]]
 name = "steno"
-version = "0.3.1-dev"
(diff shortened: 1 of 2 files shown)


ahl (approved):
Mark 14:38

smklein (approved):

PR #2119, opened 2023-01-07
[nexus] Fix missing endpoint summaries, deprecate more that should be

david-crespo:
I was updating the docs site to the latest spec and noticed some of the new endpoints have no label in the sidebar or title in the content pane.

I also added `deprecated = true` to a few legacy endpoints that had corresponding `/v1` ones. There are now exactly 24 `/v1` and 24 deprecated.

2023-01-06-docs-no-summary


zephraph:
Thanks for the cleanup! I forgot that docstring was being used in the docs.

zephraph (approved):

PR #2123, opened 2023-01-10
Fix simulated sled agent RSS handoff

smklein:


david-crespo:
I’m surprised this didn’t come up sooner. Maybe fewer things are using the simulated sled agent than I thought, or they were pinned to an older working version.

davepacheco:
It looks like this assumes you'll be starting one Sled Agent once.  One of the useful things about the simulated mode is that you can start up a lot of Sled Agents.  I assume they shouldn't all do the handoff?

To be most analogous to the real system, it seems like this should be a separate manual step in https://github.com/oxidecomputer/omicron/blob/main/docs/how-to-run-simulated.adoc, right?  In principle maybe this could even use most of the same mechanism that the real RSS uses?

karencfv (commented):
Thanks for working on this! Just tried it out and it works like a charm 🙇‍♀️

smklein:
> It looks like this assumes you'll be starting one Sled Agent once. One of the useful things about the simulated mode is that you can start up a lot of Sled Agents. I assume they shouldn't all do the handoff?
> 
> To be most analogous to the real system, it seems like this should be a separate manual step in https://github.com/oxidecomputer/omicron/blob/main/docs/how-to-run-simulated.adoc, right? In principle maybe this could even use most of the same mechanism that the real RSS uses?

We're in a weird mid-point with some of these features. In the limit, sleds should report themselves to Nexus. After RSS handoff, *nexus* should be in charge of telling individual sleds to provision datasets on them. 

This isn't how initialization of the zpools within sled agents used to work:

- The "simulated sleds agents" don't coordinate with each other on a bootstrap network, so they can't really coordinate with this handoff.
- The "simulated sled agents" currently assume they're running with a single zpool, which contains a single (crucible) dataset.

Here's my recommendation:

- I can easily plumb through the initialization info for a *single* sled agent. This means the "one simulated sled agent" case works fine, and can provision disks.
- **Later**, when Nexus is fully in charge of dataset provisioning to U.2s, it'll be fine for us to just have each simulated sled agent report "I have a zpool with nothing on it". Nexus will respond by telling them to "provision crucible", and we won't need to coordinate a fake handoff on a non-existent simulated bootstrap network.

davepacheco:
Sounds good.  Is it easy in the meantime to make this optional with a CLI flag?  That way somebody could still run multiple simulated sled agents and things would still work as long as only one of them did the transfer of control?  (If that makes no sense or if it's not a trivial change, feel free to ignore!)

smklein:
> Sounds good. Is it easy in the meantime to make this optional with a CLI flag? That way somebody could still run multiple simulated sled agents and things would still work as long as only one of them did the transfer of control? (If that makes no sense or if it's not a trivial change, feel free to ignore!)

I actually don't think that'll be necessary - subsequent requests to perform RSS -> Nexus handoff will be ignored, and see an "OK" response: https://github.com/oxidecomputer/omicron/blob/27dfb62275b92b960a3e2a1c0644cfca492f3faf/nexus/src/db/datastore/rack.rs#L117-L120

This just means that the 2nd+ sled agents won't be able to tell Nexus about the crucible datasets they're using (yet).

PR #2126, opened 2023-01-10
add missing members

### LINKED ISSUE #2125: dependabot has failed since the conversion to workspace dependencies

ahl:
Fixes #2125 

See https://github.com/dependabot/dependabot-core/issues/5864

I added the missing workspace members to the `members` array in the top level `Cargo.toml`. I also put all members into `sort` order since there were several opinions expressed in the list about the order of ASCII (and the order of the alphabet).


davepacheco (approved):
Thanks for finding this!

luqmana (approved):
thanks for catching this!

PR #1904, opened 2022-11-01
Update OPTE

plotnick:
Pull in [OPTE version `e6e197a08c086cfa4cdb2e58bf71c580f89229ac`]( https://github.com/oxidecomputer/opte/commit/e6e197a08c086cfa4cdb2e58bf71c580f89229ac) (which includes new default firewall actions and [VNI filters](https://github.com/oxidecomputer/opte/issues/281)),
update sled-agent for API changes, and [remove default RDP firewall rule](https://github.com/oxidecomputer/omicron/issues/1821#issuecomment-1279567735).

PR #2146, opened 2023-01-10
rev opte to get to a stable build

ahl:


luqmana:
Unfortunately this will fail CI until the lab-netdev images get upgraded. Working with josh on getting that sorted out.

Also, there's since been some changes to opte that need to be fixed on the omicron side. https://github.com/oxidecomputer/omicron/pull/1904 already has most of those changes. I also have some local changes for upcoming opte changes too.

ahl:
@luqmana sounds like I should close this?

luqmana:
yea, for the moment at least

PR #2164, opened 2023-01-13
[wicket] move wizard into its own file

sunshowers:
Currently, wizard is the main entrypoint for wicket. To support
uploading artifacts, We're going to change it so that it has to go
through a clap command layer.


andrewjstone (approved):
LGTM! Welcome to the world of wicket @sunshowers

PR #2165, opened 2023-01-13
[wicket] add an upload command

sunshowers:
Add a command that can upload data to wicketd, via stdin from wicket.

This also makes wicket behave like a login shell: read arguments passed
in with `-c`.

Tested by creating a user locally with wicket as the login shell, then
running:

```
% cat foo.bin | ssh wicket-test@localhost upload my-artifact 0.1.0
Jan 13 22:12:25.946 INFO read artifact my-artifact:0.1.0 (11957848 bytes) from stdin, uploading to wicketd
Jan 13 22:12:25.972 INFO successfully uploaded my-artifact:0.1.0 (11957848 bytes) to wicketd
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -143,6 +143,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "assert_cmd"
+version = "2.0.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9834fcc22e0874394a010230586367d4a3e9f11b560f469262678547e1d2575e"
+dependencies = [
+ "bstr 1.1.0",
+ "doc-comment",
+ "predicates",
+ "predicates-core",
+ "predicates-tree",
+ "wait-timeout",
+]
+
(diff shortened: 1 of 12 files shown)


andrewjstone (approved):
Looks great. Thanks for fixing the logging too!

jclulow:
When we look at locking down the sshd configuration we might want to consider not doing the login shell pattern, but rather using the sshd configuration directive `ForceCommand` and then having the shell program process the `SSH_ORIGINAL_COMMAND` environment variable that would be set when the forced command is invoked.

sunshowers:
> When we look at locking down the sshd configuration we might want to consider not doing the login shell pattern, but rather using the sshd configuration directive ForceCommand and then having the shell program process the SSH_ORIGINAL_COMMAND environment variable that would be set when the forced command is invoked.

Good point! I don't have a sense of which one's better, but if we decide to do that we can switch over to it pretty easily.

PR #2167, opened 2023-01-13
[wicket] fix call to get_prev_component_id

sunshowers:
Previously, we were showing the next component ID in both the previous
and next spots.

After this change:

image


andrewjstone (approved):
Nice! How the hell did I let that one slip in.

PR #2176, opened 2023-01-17
[sled-agent] Disk Detection, Partition Management, and U.2 formatting

smklein:
# Features

Enables the Sled Agent to identify physical disks, format zpools on U.2 devices, and use those zpools as allocation targets for datasets within Nexus.

For anyone familiar with the [hard-coded list of zpools in sled agent's `config.toml`](https://github.com/oxidecomputer/omicron/blob/3f8e18a07c8711c8fd99e0cf941721525e85c4d9/smf/sled-agent/config.toml#L17-L27), these hook into the same mechanism, but come from real devices. As such, specifying them ahead-of-time is no longer required.

## Mechanism

Implements the mechanism described in [RFD 352](https://rfd.shared.oxide.computer/rfd/0352) to detect disks, partitions, and zpools within.
- Uses [illumos-devinfo](https://github.com/oxidecomputer/illumos-devinfo) to query `libdevinfo` for device information
- Uses [libefi-illumos](https://github.com/oxidecomputer/libefi-illumos) to query `libefi` for partition information
- Uses [fstyp](https://illumos.org/man/8/fstyp) to bridge the gap from "partition path" to "zpool name".

## Testing

Tested manually on `gimlet-sn21` on the following device layout:

```bash
Slot    Device  Path
   0       u.2  /devices/pci@ab,0/pci1022,1483@1,1/pci1b96,0@0/blkdev@w0014EE81000BC153,0
   1       u.2  /devices/pci@ab,0/pci1022,1483@1,2/pci1b96,0@0/blkdev@w0014EE81000BC67A,0
   2       u.2  /devices/pci@ab,0/pci1022,1483@1,3/pci1b96,0@0/blkdev@w0014EE81000BC5F5,0
   3       u.2  /devices/pci@ab,0/pci1022,1483@1,4/pci1b96,0@0/blkdev@w0014EE81000BC5C5,0
   4       u.2  /devices/pci@0,0/pci1022,1483@3,4/pci1b96,0@0/blkdev@w0014EE81000BC5D4,0
   5       u.2  /devices/pci@0,0/pci1022,1483@3,3/pci1b96,0@0/blkdev@w0014EE81000BC5C9,0
   6       u.2  /devices/pci@0,0/pci1022,1483@3,2/pci1b96,0@0/blkdev@w0014EE81000BC5D2,0
   7       u.2  /devices/pci@38,0/pci1022,1483@1,3/pci1b96,0@0/blkdev@w0014EE81000BC5CB,0
  17       m.2  /devices/pci@0,0/pci1022,1483@1,3/pci1344,3100@0/blkdev@w00A07501340802D1,0
```

- Used `nvmeadm secure-erase` to fully wipe U.2 devices, observed that the sled-agent detects and correctly formats them.
- Used `zpool destroy` to wipe the zpool off the U.2 device while leaving the GPT partition, observed that the sled-agent correctly detects the partition and adds the zpool to it.
- Tested that in all cases Nexus is notified of the new zpools.


common/src/sql/dbinit.sql
```diff
@@ -128,6 +128,35 @@ CREATE INDEX ON omicron.public.service (
     sled_id
 );
 
+-- A physical disk which exists inside the rack.
```
smklein, line 131: Probably need to pull this to a separate PR

sled-agent/src/hardware/illumos/mod.rs
```diff
@@ -53,16 +94,69 @@ enum TofinoView {
...
+                    (true, false) => updates.push(TofinoUnloaded),
+                    _ => (),
+                };
+
+                // Update our view of the underlying hardware
+                self.tofino = TofinoView::Real(polled_hw.tofino);
+            }
+            TofinoView::Stub { .. } => (),
+        }
+    }
```
smklein, line 138: This implementation is *basically* the same as before this PR, just moved.

sled-agent/src/hardware/illumos/mod.rs
```diff
@@ -73,56 +167,194 @@ fn node_name(subsystem_vid: i32, subsystem_id: i32) -> String {
...
+    Ok(None)
+}
+
+// Gather a notion of "device identity".
+//
+// NOTE: What exactly comprises a "reliable, unique identifier" for a disk is
+// still a little bit in flux. While that's being sorted out, we simply produce
+// a string that can uniquely identify the device to Nexus.
+//
+// At the moment, this consists of a "Vendor ID, Serial ID, and model ID".
```
smklein, line 246: I'm open to actually returning these individual fields up to Nexus, rather than as a concatenated string.

However, from Nexus' point-of-view, I'd like to use the ~~Verbal, Somatic, and Material Components~~ Vendor, Serial, and Model IDs as an opaque primary key to identify disks.

sled-agent/src/hardware/illumos/mod.rs
```diff
@@ -73,56 +167,194 @@ fn node_name(subsystem_vid: i32, subsystem_id: i32) -> String {
...
+                node = parent;
+
+                if let Some(driver_name) = node.driver_name() {
+                    if driver_name == "nvme" {
+                        device_id = Some(blkdev_device_identity(&format!(
+                            "/devices{}:devctl",
+                            node.devfs_path().map_err(Error::DevInfo)?
+                        ))?);
+                    }
+                }
```
smklein, line 288: To make this a bit more concrete:

We might find a device like:
`blkdev` -> `/devices/pci@ab,0/pci1022,1483@1,1/pci1b96,0@0/blkdev@w0014EE81000BC153,0`

This means we could find a parent `nvme` device at:
`nvme` -> `/devices/pci@ab,0/pci1022,1483@1,1/pci1b96,0@0:devctl`

This is the device that will respond to NVMe-specific ioctls, and can return a unique device identity for us to use as a primary key at a higher layer in the control plane.

sled-agent/src/hardware/illumos/mod.rs
```diff
@@ -73,56 +167,194 @@ fn node_name(subsystem_vid: i32, subsystem_id: i32) -> String {
...
+                    })?;
+
+                    let variant = if let Some(variant) =
+                        slot_to_disk_variant(slot)
+                    {
+                        variant
+                    } else {
+                        warn!(log, "Slot# {slot} is not recognized as a disk: {devfs_path}");
+                        break;
+                    };
```
smklein, line 313: This is, AFAICT, the most reliable way of distinguishing between U.2 and M.2 devices.

sled-agent/src/hardware/illumos/mod.rs
```diff
@@ -73,56 +167,194 @@ fn node_name(subsystem_vid: i32, subsystem_id: i32) -> String {
...
+                    let disk = UnparsedDisk::new(
```
smklein, line 314: The "polling loop" should be pretty tight, since we'll be re-running it periodically until we have a better event-based scheme. To that end: I'm avoiding doing GPT parsing here, which is why we're returning an "UnparsedDisk", rather than just a "Disk".

sled-agent/src/hardware/illumos/nvme.rs
```diff
@@ -0,0 +1,118 @@
...
+}
+
+/// FFI-safe version of the `nvme_ioctl_t` used to communicate with the `nvme` driver
+#[repr(C)]
+#[derive(Debug)]
+struct Ioctl {
+    // Length of `n_buf`
+    n_len: libc::size_t,
+    // Data buffer for command
+    n_buf: *mut libc::c_void,
```
jgallagher, line 77: The C struct defines this field as a `uintptr_t` - should we use `usize` here instead of `*mut c_void`? https://rust-lang.github.io/unsafe-code-guidelines/layout/scalars.html#isize-and-usize notes that usize is guaranteed to be compatible with `uintptr_t`.

sled-agent/src/hardware/illumos/nvme.rs
```diff
@@ -0,0 +1,118 @@
...
+    pub nvme_version: String,
+    pub fru_guid: String,
+}
+
+pub(crate) const NVME_ID_CTRL_BUFSIZE: usize = 4096;
+
+impl From<&[u8]> for ControllerId {
+    fn from(buf: &[u8]) -> ControllerId {
+        assert!(buf.len() >= NVME_ID_CTRL_BUFSIZE);
+        let vendor_id = u16::from_le_bytes(buf[..2].try_into().unwrap());
```
jgallagher, line 34: Should we use `from_ne_bytes` throughout this function? I'm assuming `buf` contains the equivalent of `memcpy`'ing the `nvme_identify_ctrl_t` C structure, which would give us native endianness on all the fields.

jgallagher: Following up on that - if you wanted to run the `nvme_identify_ctrl_t` through bindgen to get a Rust-equivalent definition, we could use [zerocopy](https://docs.rs/zerocopy/latest/zerocopy/index.html) to copy from `buf` into that struct, and then pull all the fields out directly instead of having to hardcode offsets. (We use `zerocopy` in hubris quite a lot, although more for IPC than FFI.)

sled-agent/src/hardware/illumos/gpt.rs
```diff
@@ -0,0 +1,49 @@
...
+//! Enables either real or faked GPT access.
+
+use libefi_illumos::Error;
+use std::path::Path;
+
+/// Trait to sub-in for access to the libefi_illumos::Gpt.
+///
+/// Feel free to extend this interface to exactly match the methods exposed
+/// by libefi_illumos::Gpt for testing.
+pub(crate) trait LibEFIGpt {
```
jgallagher, line 16: Style nit - `LibEfiGpt` maybe (similarly `LibEfiPartition`)?

sled-agent/src/hardware/illumos/disk.rs
```diff
@@ -0,0 +1,407 @@
...
+            });
+        }
+
+        // NOTE: If we wanted to, we could validate additional information about
+        // the size, GUID, or name of the partition. At the moment, however,
+        // we're relying on the index within the partition table to indicate the
+        // "intent" of the partition.
+    }
+
+    Ok(expected_partitions.iter().map(|p| p.clone()).collect())
```
jgallagher, line 71: I think this could be
```suggestion
    Ok(expected_partitions.to_vec())
```
although that might need a `Copy` on `Partition`, not sure.

sled-agent/src/hardware/illumos/disk.rs
```diff
@@ -0,0 +1,407 @@
...
+                    let zpool_name = ZpoolName::new(Uuid::new_v4());
+                    Zpool::create(zpool_name, dev_path)?;
+                    return Ok(vec![Partition::ZfsPool]);
+                }
+                DiskVariant::M2 => {
+                    // TODO: If we see a completely empty M.2, should we create
+                    // the expected partitions? Or would it be wiser to infer
+                    // that this indicates an unexpected error conditions that
+                    // needs mitigation?
+                    todo!("Provisioning M.2 devices not yet supported");
```
jgallagher, line 143: How long do we expect this `todo!()` to be around? Would it be safer to make it an error instead?

sled-agent/src/hardware/illumos/mod.rs
```diff
@@ -73,56 +170,187 @@ fn node_name(subsystem_vid: i32, subsystem_id: i32) -> String {
     format!("pci{subsystem_vid:x},{subsystem_id:x}")
 }
 
+fn slot_to_disk_variant(slot: i64) -> Option<DiskVariant> {
+    match slot {
+        0x00..=0x09 => Some(DiskVariant::U2),
+        0x11..=0x12 => Some(DiskVariant::M2),
```
jgallagher, line 176: Where do these values come from?

sled-agent/src/hardware/illumos/mod.rs
```diff
@@ -73,56 +170,187 @@ fn node_name(subsystem_vid: i32, subsystem_id: i32) -> String {
...
+    let model = ctrl_id.model;
+    Ok(DiskIdentity { vendor, serial, model })
+}
+
+fn poll_blkdev_node(
+    log: &Logger,
+    disks: &mut HashSet<UnparsedDisk>,
+    mut node: Node<'_>,
+) -> Result<(), Error> {
+    if let Some(driver_name) = node.driver_name() {
```
jgallagher, line 261: Could use the new `let else` syntax here to reduce rightward drift, I think? Something like

```rust
let Some(driver_name) = node.driver.name() else {
    return Ok(());
};
```

I'd probably also throw in

```rust
if driver_name != "blkdev" {
    return Ok(());
}
```

to shift this whole function over a couple levels.

sled-agent/src/hardware/illumos/mod.rs
```diff
@@ -73,56 +170,187 @@ fn node_name(subsystem_vid: i32, subsystem_id: i32) -> String {
...
+                    })
+                {
+                    let slot = slot_prop.as_i64().ok_or_else(|| {
+                        Error::UnexpectedPropertyType {
+                            name: slot_prop_name,
+                            ty: "i64".to_string(),
+                        }
+                    })?;
+
+                    let variant = if let Some(variant) =
```
jgallagher, line 302: I think this could also use `let .. else`:

```rust
let Some(variant) = slot_to_disk_variant(slot) else {
    warn!(log, "Slot# {slot} is not recognized as a disk: {devfs_path}");
    break;
};
```

sled-agent/src/illumos/zpool.rs
```diff
@@ -154,6 +162,21 @@ pub struct Zpool {}
...
+    pub fn create(
+        name: ZpoolName,
+        vdev: &std::path::PathBuf,
+    ) -> Result<(), CreateError> {
+        let mut command = std::process::Command::new(PFEXEC);
+        let cmd = command.args(&[
+            ZPOOL,
+            "create",
+            &name.to_string(),
+            &vdev.to_string_lossy(),
```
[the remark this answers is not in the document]
jclulow: I think it's probably best to always use `.arg()`, generally, so that you don't have to worry about string type specifics, and you can group related arguments; i.e.,

```rust
let mut cmd = std::process::Command::new(PFEXEC);
cmd.arg(ZPOOL).arg("create");
cmd.arg(&name.to_string()); /* does this not implement an .as_str() or Display? */
cmd.arg(vdev);
```

sled-agent/src/illumos/zpool.rs
```diff
@@ -25,6 +25,14 @@ enum Error {
     Parse(#[from] ParseError),
 }
 
+#[derive(thiserror::Error, Debug)]
+#[error("Failed to create zpool: {err}")]
+pub struct CreateError {
+    #[from]
+    #[source]
```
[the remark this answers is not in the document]
smklein: Done, thanks for the advice!

sled-agent/src/illumos/zpool.rs
```diff
@@ -154,6 +162,21 @@ pub struct Zpool {}
 
 #[cfg_attr(test, mockall::automock)]
 impl Zpool {
+    pub fn create(
+        name: ZpoolName,
+        vdev: &std::path::PathBuf,
```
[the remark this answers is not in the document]
smklein: Updated

sled-agent/src/illumos/zpool.rs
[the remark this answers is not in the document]
smklein: Ah, thanks, calling `arg` repeatedly does make the conversion a lot easier. Updated.

sled-agent/src/sled_agent.rs
```diff
@@ -400,9 +400,18 @@ impl SledAgent {
...
             }
         }
+
+        if let Err(e) = self
+            .inner
+            .storage
+            .ensure_using_exactly_these_disks(self.inner.hardware.disks())
+            .await
+        {
+            warn!(log, "Failed to add disk: {e}");
```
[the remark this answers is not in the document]
smklein: Failure to ensure any disks results in an error, not a warning. However, we still *try* to insert /remove everything else that needs to be changed.

I didn't want "one bad U.2" to cause a whole sled to eject itself, which is sorta the motivation behind "continuing when we've already seen an error anyway".

sled-agent/src/storage_manager.rs
```diff
@@ -863,13 +880,151 @@ impl StorageManager {
...
+        &self,
+        unparsed_disks: I,
+    ) -> Result<(), Error>
+    where
+        I: IntoIterator<Item = crate::hardware::UnparsedDisk>,
+    {
+        let mut new_disks = HashMap::new();
+
+        // We may encounter errors while parsing any of the disks; keep track of
+        // any errors that occur and return any of them if something goes wrong.
```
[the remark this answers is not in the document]
smklein: This is true, and I've updated the doc comment. At the moment, no, this doesn't change the caller's behavior.

sled-agent/src/storage_manager.rs
```diff
@@ -863,13 +880,151 @@ impl StorageManager {
...
+        let keys_to_be_removed = disks
+            .keys()
+            .filter_map(|key| {
+                if !new_disks.contains_key(key) {
+                    Some(key.clone())
+                } else {
+                    None
+                }
+            })
+            .collect::<Vec<_>>();
```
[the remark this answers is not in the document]
smklein: I updated to mostly use this format, with one wrinkle. I agree that the problem area of "disk was re-inserted, but with new data" is... weird. It's complex. I dunno how to arbitrarily handle it, especially if subtle aspects of the disk somehow change.

I've opted to treat this as a case of "remove the disk", and then "add the disk". Open to suggestions on what else to do, but it would certainly be unexpected if a disk started reporting a new partition layout / zpool / slot while we were in the middle of using it.

sled-agent/src/storage_manager.rs
```diff
@@ -863,13 +880,151 @@ impl StorageManager {
...
+    }
+
+    async fn upsert_disk_locked(
+        &self,
+        disks: &mut tokio::sync::MutexGuard<'_, HashMap<PathBuf, Disk>>,
+        disk: Disk,
+    ) -> Result<(), Error> {
+        let zpool_name = disk.zpool_name().clone();
+        disks.insert(disk.devfs_path().to_path_buf(), disk);
+        self.upsert_zpool(&zpool_name).await?;
```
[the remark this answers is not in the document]
smklein: Yeah, I think that's right. It would still be useful to track the physical disk, even if the zpool can't load for some reason.

sled-agent/src/hardware/mod.rs
```diff
@@ -24,5 +31,305 @@ pub enum HardwareUpdate {
...
+    IoError { path: PathBuf, error: std::io::Error },
+    #[error("Failed to open partition at {path} due to {error}")]
+    Gpt { path: PathBuf, error: anyhow::Error },
+    #[error("Unexpected partition layout at {path}: {why}")]
+    BadPartitionLayout { path: PathBuf, why: String },
+    #[error("Requested partition {partition:?} not found on device {path}")]
+    NotFound { path: PathBuf, partition: Partition },
+    #[error(transparent)]
+    ZpoolCreate(#[from] crate::illumos::zpool::CreateError),
+    #[error("Cannot format {path}: missing a '/dev/ path")]
```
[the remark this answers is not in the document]
smklein: Yupdated

sled-agent/src/hardware/mod.rs
```diff
@@ -24,5 +31,305 @@ pub enum HardwareUpdate {
...
+            "{path}:wd{raw}",
+            path = self.devfs_path.display(),
+            raw = if raw { ",raw" } else { "" },
+        ))
+    }
+
+    // Finds the first 'variant' partition, and returns the path to it.
+    fn partition_device_path(
+        &self,
+        partitions: &Vec<Partition>,
```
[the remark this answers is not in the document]
smklein: Fixed, thanks!

sled-agent/src/hardware/illumos/nvme.rs
```diff
@@ -0,0 +1,118 @@
...
+        let subsystem_vendor_id =
+            u16::from_le_bytes(buf[2..4].try_into().unwrap());
+        let serial =
+            String::from_utf8(buf[4..24].to_vec()).unwrap().trim().to_string();
+        let model =
+            String::from_utf8(buf[24..64].to_vec()).unwrap().trim().to_string();
+        let firmware_rev =
+            String::from_utf8(buf[64..72].to_vec()).unwrap().trim().to_string();
```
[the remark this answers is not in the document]
smklein: Went ahead and made this lossy, to avoid the unwrap.

sled-agent/src/hardware/illumos/nvme.rs
```diff
@@ -0,0 +1,118 @@
...
+// Ioctl command definitions
+const NVME_IOC: i32 =
+    ('N' as i32) << 24 | ('V' as i32) << 16 | ('M' as i32) << 8;
+const NVME_IOC_IDENTIFY_CTRL: i32 = NVME_IOC | 1;
+const NVME_IDENTIFY_CTRL: u64 = 1;
+
+fn ioctl(
+    dev: &PathBuf,
+    cmd: i32,
+    buffer: &[u8],
```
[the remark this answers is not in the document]
smklein: Good catch, updated.

sled-agent/src/hardware/illumos/nvme.rs
smklein: Switched to `usize`, it worked right away.

sled-agent/src/hardware/illumos/nvme.rs
smklein: Updated to `from_ne_bytes`, agreed that makes more sense than assuming the host is little endian.

I went with the non-bindgen approach just to keep the code surface area relatively small, but I'd be open to a bindgen-produced API if we really wanted a more involved interaction with the NVME interface.

sled-agent/src/hardware/illumos/gpt.rs
smklein: Sure, I'm always on the fence with acronyms (sometimes, the letters *should* be capitalized!), but this updated to `LibEfi...`.

sled-agent/src/hardware/illumos/disk.rs
smklein: Updated, thanks!

sled-agent/src/hardware/illumos/disk.rs
smklein: Yeah, I'll throw an error instead. Updated.

sled-agent/src/hardware/illumos/mod.rs
smklein: https://github.com/oxidecomputer/illumos-gate/blob/stlouis/usr/src/uts/oxide/milan/milan_dxio_data.c#L823-L847

Added a link to this in the documentation for this function.

sled-agent/src/hardware/illumos/mod.rs
smklein: Good catch, updated!

sled-agent/src/hardware/illumos/mod.rs
smklein: Updated

sled-agent/src/storage_manager.rs
```diff
@@ -863,13 +880,177 @@ impl StorageManager {
...
+                    // Changed Disk -> Disk should be removed.
+                    new_disk != *old_disk
+                } else {
+                    // Not in the new set -> Disk should be removed.
+                    true
+                }
+            })
+            .map(|(key, _)| key);
+        for key in keys_to_be_removed {
+            if let Err(e) = self.delete_disk_locked(&mut disks, &key).await {
```
jgallagher, line 963: I don't think this is right - at this point `disks` is empty (because we swapped it on line 940), so `delete_disk_locked` won't do anything. Fixing this seems tricky. Playing with it a bit.

sled-agent/src/hardware/illumos/mod.rs
```diff
@@ -73,56 +170,188 @@ fn node_name(subsystem_vid: i32, subsystem_id: i32) -> String {
...
+                slot,
+                variant,
+                device_id.ok_or_else(|| {
+                    Error::NotAnNVMEDevice(PathBuf::from(devfs_path))
+                })?,
+            );
+            disks.insert(disk);
+            break;
+        }
+    }
```
jclulow, line 325: I think rather than a loop, here, which seems like it would admit some tree structures that are not quite what we are expecting, I would do an explicit walk up the tree looking for the precise shape we want; e.g., the sketch below:

```rust
/*
 * Locate nvme parent of blkdev node:
 */
let Some(p) = node.parent()? else {
    bail!("blkdev has no parent device?");
};
if p.driver_name.as_deref() != Some("nvme") {
    bail!("expected nvme parent, got {:?}", p.driver_name);
}
/*
 * XXX Get the inquiry-* properties from the node while we're here.
 */

/*
 * Locate the PCI parent of the nvme node:
 */
let Some(p) = p.parent()? else {
    bail!("nvme has no parent device?");
};
let slot = p.props()
    .filter_map(|res| res.ok())
    .filter(|prop| prop.name() == "physical-slot#")
    .filter_map(|prop| {
        let num: Option<u32> = prop.as_i64().try_into();
        num
    })
    .next();
let Some(slot) = slot else {
    bail!("no physical slot number on PCI bridge?");
};
let Some(v) = slot_to_disk_variant(slot) else {
    bail!("unrecognised disk slot {}", slot);
};

let disk = UnparsedDisk::new(
...
```

jclulow (changes requested):
Hi Sean!

With my apologies for the interminable delay, I have taken an initial look through this.  I think the shape is about right on the Nexus/Omicron end.  There are some things about the interaction with devinfo and NVMe and programs that we're running that I think might merit some changes.

Let me know if anything I've said needs more explanation.  There are a lot of complex moving pieces which I am always happy to talk through if you'd like.

Thanks for picking this up, it's extremely critical!

sled-agent/src/illumos/fstyp.rs
```diff
@@ -0,0 +1,57 @@
...
+#[cfg_attr(test, mockall::automock)]
+impl Fstyp {
+    /// Executes the 'fstyp' command and parses the name of a zpool from it, if
+    /// one exists.
+    pub fn get_zpool(path: &Path) -> Result<ZpoolName, Error> {
+        let mut command = std::process::Command::new(PFEXEC);
+        let cmd = command.arg(FSTYP).arg("-a").arg(path);
+
+        let output = execute(cmd).map_err(Error::from)?;
+        let stdout = String::from_utf8_lossy(&output.stdout);
```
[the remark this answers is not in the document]
smklein: Sure, sounds good. Updated.

sled-agent/src/storage_manager.rs
smklein: My short-term workaround: I'm going back to my old mechanism of updating the set of disks in place.

This desperately needs to be refactored to have some better unit tests without the tight coupling to real hardware

smklein:
(The current CI failures are due to ghcr being down)

We pretty desperately need this PR in as the foundation for some follow-up work, so I'm going to merge. I'm still very open to follow-ups and additional feedback, but I'm prioritizing urgency in this particular case.

sled-agent/src/illumos/fstyp.rs
```diff
@@ -0,0 +1,57 @@
...
+
+/// Wraps 'fstyp' command.
+pub struct Fstyp {}
+
+#[cfg_attr(test, mockall::automock)]
+impl Fstyp {
+    /// Executes the 'fstyp' command and parses the name of a zpool from it, if
+    /// one exists.
+    pub fn get_zpool(path: &Path) -> Result<ZpoolName, Error> {
+        let mut command = std::process::Command::new(PFEXEC);
```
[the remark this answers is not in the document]
jclulow: Yes, I believe `LC_ALL` overrides everything else; [environ(7)](https://illumos.org/man/7/environ) seems to concur.

sled-agent/src/hardware/illumos/mod.rs
jclulow: Looks great!

PR #2182, opened 2023-01-19
Use strings for update artifact versions

iliana:
This is part 1 for getting TUF repo generation done; this changes a common data structure for updates so I can do some other data type shifting on top of it.

I had initially used integer versions for update artifacts as a bit of a strawman. Over the course of a few meetings on updates that culminated in [RFD 334](https://rfd.shared.oxide.computer/rfd/0334), still in prediscussion, I've been convinced that we need something a bit more nuanced than a monotonically-increasing integer. While RFD 334 still shows a determination that a monotonically-increasing integer is okay, the inability to describe an "in-between" version (for the certainly-likely future case of producing a bug fix on an older, still-supported branch while customers move to the latest branch) is unacceptable.

Semver is the likely version format, but as far as Nexus and Sled Agent talking to each other is concerned, the version could be any string we like as long as it's unique for that name and artifact kind. (Strings also keep the use of `#[derive(JsonSchema)]` simple.)

### DIFF
--- a/common/src/api/internal/nexus.rs
+++ b/common/src/api/internal/nexus.rs
@@ -81,7 +81,7 @@ impl ProducerEndpoint {
 #[derive(Clone, Debug, Deserialize, Serialize, JsonSchema)]
 pub struct UpdateArtifact {
     pub name: String,
-    pub version: i64,
+    pub version: String,
     pub kind: UpdateArtifactKind,
 }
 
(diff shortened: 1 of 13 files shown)


plotnick (approved):
I think we're all agreed now that semver (and so strings rather than integers) are the right approach overall, and will give us the flexibility that we need. I will be updating RFD 334 shortly to reflect this. Thanks for updating this part!

PR #2183, opened 2023-01-19
clippy in CI skips lots of code

davepacheco:
We need `--all-targets` to cover tests, examples, etc.

Also by not running this on Helios, I think we're probably not covering the illumos-specific code.

This PR:

- updates the GitHub Actions CI job to run clippy with `--all-targets`
- adds a buildomat job to run clippy the same way
- fixes all the newly-found clippy issues.  I used `--fix` to do a lot of this.

### DIFF
--- a/.github/buildomat/jobs/clippy.sh
+++ b/.github/buildomat/jobs/clippy.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+#:
+#: name = "clippy (helios)"
+#: variety = "basic"
+#: target = "helios-latest"
+#: rust_toolchain = "nightly-2022-09-27"
+#: output_rules = []
+#:
+
+# Run clippy on illumos (not just other systems) because a bunch of our code
+# (that we want to check) is conditionally-compiled on illumos only.
+
+set -o errexit
+set -o pipefail
+set -o xtrace
+
+cargo --version
+rustc --version
+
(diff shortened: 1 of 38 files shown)


.github/buildomat/jobs/clippy.sh
```diff
@@ -0,0 +1,27 @@
...
+rustc --version
+
+# Put tools on our PATH to satisfy install_builder_prerequisites.
+export PATH="$PATH:$PWD/out/cockroachdb/bin:$PWD/out/clickhouse"
+banner prerequisites
+ptime -m bash ./tools/install_builder_prerequisites.sh -y
+
+banner clippy
+# See the corresponding GitHub Actions job for more about these arguments.
+ptime -m cargo +'nightly-2022-09-27' clippy --all-targets -- --deny warnings --allow clippy::style
```
david-crespo, line 27: why does this one need explicit nightly version and the other doesn't?

david-crespo (approved):

.github/buildomat/jobs/clippy.sh
davepacheco: I don't think there's a good reason.  I went for consistency with the toolchain used elsewhere (and actually didn't notice we _weren't_ specifying a toolchain on the other one).

PR #2187, opened 2023-01-20
Bump rust toolchain to nightly-2023-01-19

smklein:
I need this to access a "raw FD" on illumos in this PR: https://github.com/oxidecomputer/omicron/pull/2176

Apparently this was only a module that was made non-private on illumos rather recently.

### DIFF
--- a/.github/buildomat/jobs/build-and-test-linux.sh
+++ b/.github/buildomat/jobs/build-and-test-linux.sh
@@ -3,7 +3,7 @@
 #: name = "build-and-test (ubuntu-20.04)"
 #: variety = "basic"
 #: target = "ubuntu-20.04"
-#: rust_toolchain = "nightly-2022-09-27"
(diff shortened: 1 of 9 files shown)


iliana:
This doesn't seem to be strictly necessary; this import compiles fine on stable (1.65.0):

```
use std::os::unix::io::RawFd;
```

The [`os_fd` tracking issue](https://github.com/rust-lang/rust/issues/98699) indicates the new path, `std::os::fd::RawFd`, is meant to unify unix and wasi under the same path since they both have the same concept, but we're not targeting wasi. Unfortunately now you only find the new path in the generated docs.

(It's also not clear to me when this is landing in stable. The feature gate in the code shows it'll stabilize in 1.66.0 but I didn't see any indication of the stabilization process happening, it almost looks accidental.)

ahl:
nightly is dead; long live nightly

Issue #2186, opened 2023-01-20
thread 'integration_tests::instances::test_instance_metrics' panicked at 'attempt to subtract with overflow'


iliana:
https://github.com/oxidecomputer/omicron/runs/10768352787 / https://buildomat.eng.oxide.computer/wg/0/details/01GQ6DS2KJ0QDM6YWDNRZDTCHZ/B44Zl5JIuOAhkn5GInFdH46LbPhq8Dq00FztPZBK0cQJCm5s/01GQ6DSB1H70J2FKRWK36STEQ6#S11679:

```
---- integration_tests::instances::test_instance_metrics stdout ----
--
log file: /var/tmp/omicron_tmp/test_all-df03734e17da7df2-test_instance_metrics.51359.58.log
note: configured to log to "/var/tmp/omicron_tmp/test_all-df03734e17da7df2-test_instance_metrics.51359.58.log"
thread 'integration_tests::instances::test_instance_metrics' panicked at 'attempt to subtract with overflow', nexus/tests/integration_tests/metrics.rs:33:36
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
WARN: dropped CockroachInstance without cleaning it up first (there may still be a child process running and a temporary directory leaked)
WARN: temporary directory leaked: /var/tmp/omicron_tmp/.tmpm1YIU2
WARN: dropped ClickHouseInstance without cleaning it up first (there may still be a child process running and a temporary directory leaked)
```

https://github.com/oxidecomputer/omicron/blob/d6dc72f6a5339296193f1a47310ced437aa39aa1/nexus/tests/integration_tests/metrics.rs#L26-L39

This panics if `measurements.items.len() == 0`.

smklein:
Almost definitely a result of https://github.com/oxidecomputer/omicron/pull/1782 ; I can look

smklein:
Here's what that test is *supposed* to be doing...

We're supposed to be querying the "project" and "org" we created to see how much space they report to be using.

We first do that by querying the DB for those values, which indicates they've been inserted: https://github.com/oxidecomputer/omicron/blob/d6dc72f6a5339296193f1a47310ced437aa39aa1/nexus/tests/integration_tests/instances.rs#L545-L553

We also want to query the Nexus API, to test that the metrics are being offloaded to Clickhouse and Oximeter properly.

[...]

PR #2188, opened 2023-01-20
Add collection tokens into oximeter

smklein:
Attempts to improve the `force_collect` method on `Oximeter`:

https://github.com/oxidecomputer/omicron/blob/e69a5a34e8f357e71044956172c6c06ef5d38161/oximeter/collector/src/lib.rs#L550-L556

This method triggers the collection of metrics from Producers, but Oximeter is a pipeline, and this just *sends* an initial message, rather than ensuring that those metrics have made their way to Clickhouse.

This PR weaves an optional "token" through this pipeline, so a caller can call `force_collect` and confidently know "if any metrics were ready to be produced before, they'll be in Clickhouse after this call".

Fixes https://github.com/oxidecomputer/omicron/issues/2186

### DIFF
--- a/oximeter/collector/src/lib.rs
+++ b/oximeter/collector/src/lib.rs
@@ -28,7 +28,9 @@ use std::path::Path;
 use std::sync::Arc;
 use std::time::Duration;
 use thiserror::Error;
-use tokio::{sync::mpsc, sync::oneshot, sync::Mutex, task::JoinHandle, time::interval};
+use tokio::{
+    sync::mpsc, sync::oneshot, sync::Mutex, task::JoinHandle, time::interval,
+};
 use uuid::Uuid;
 
 /// Errors collecting metric data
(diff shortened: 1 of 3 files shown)


oximeter/collector/src/lib.rs
```diff
@@ -47,11 +49,16 @@ pub enum Error {
...
+
 // Messages for controlling a collection task
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 enum CollectionMessage {
     // Explicit request that the task collect data from its producer
-    Collect,
+    //
+    // Also sends a oneshot that is signalled once the task scrapes
+    // data from the Producer, and places it in the Oximeter server.
```
bnaecker, line 60: "places it in the ClickHouse server"

oximeter/collector/src/lib.rs
```diff
@@ -368,11 +387,29 @@ impl OximeterAgent {
...
+            // ... and keep track of the token that indicates once the metric
+            // has made it into Clickhouse.
+            collection_oneshots.push(rx);
         }
+        drop(collection_tasks);
+
+        // Only return once all producers finish processing the token we
+        // provided.
+        //
+        // NOTE: This can either mean that the colleciton completed
```
bnaecker, line 410: Spelling nit: "colleciton" -> "collection"

bnaecker (approved):
Nice! Couple of spelling nits, but :+1: otherwise.

PR #2190, opened 2023-01-20
Allow specifying a manual maghemite source

jmpesp:
Previously, only "prebuilt" could be specified as a maghemite source in package-manifest.toml. This commit adds an option to use "manual". Users that wish to use "manual" will have to supply the file out/downloads/ddm-admin-manual.json (where "prebuilt" looks for out/downloads/ddm-admin-{commit}.json).


rcgoodfellow (approved):
LGTM!

PR #2195, opened 2023-01-21
[wicket] add a status bar with liveness information for wicketd and MGS

sunshowers:
Within wicket, it's important to know if we're seeing outdated
information. Add a status bar widget, and use it to display liveness
information for both wicketd and MGS.

In the future, we can use the status bar to show information about
updates and other information.

Here's what it looks like when it's live:

image

And here's when it hasn't received information from wicketd for a little while:

image

Depends on #2193 and #2194.


wicket/src/widgets/status_bar.rs
```diff
@@ -0,0 +1,190 @@
...
+        let row = Row::new(vec![
+            "wicketd".into(),
+            wicketd_spans,
+            // A bit of spacing between wicketd and MGS.
+            " ".into(),
+            "MGS".into(),
+            mgs_spans,
+        ]);
+
+        let table_style = Style::default().fg(OX_GRAY).bg(OX_GREEN_DARKEST);
```
andrewjstone, line 122: While I'm not looking to change it right now, It's nice to have the style settable outside the widget rather than hardcoded, so we can experiment in the caller.  This is typically done with a default / builder interface.

wicket/src/widgets/liveness.rs
```diff
@@ -0,0 +1,89 @@
...
+    Delayed(u64),
+
+    NoResponse,
+}
+
+impl ComputedLiveness {
+    pub fn to_spans(&self) -> Spans<'static> {
+        match self {
+            ComputedLiveness::Live(secs) => Spans::from(vec![
+                Span::styled("live", Style::default().fg(OX_GREEN_LIGHT)),
```
andrewjstone, line 63: The style should probably be configurable through the `StatusBar` widget and passed in via`LivenessState`. This allows easy customization of individual screens. I went through quite a bit of iteration during initial development, such as changing colors of the rack: https://github.com/oxidecomputer/omicron/blob/main/wicket/src/screens/rack.rs#L135-L153

sunshowers:
> Before merge though, can you ensure that when you resize the screen a lot, especially vertically, that the rack rendering still works? 

Thanks for catching this -- I had spotted that but I'll give everything a closer look today.

sunshowers:
> Before merge though, can you ensure that when you resize the screen a lot, especially vertically, that the rack rendering still works?

Done in https://github.com/oxidecomputer/omicron/pull/2230.

wicket/src/widgets/liveness.rs
sunshowers: Done, added a `LivenessStyles` stylesheet.

wicket/src/widgets/status_bar.rs
sunshowers: Ended up doing a bit of a compromise: styles are defined in a stylesheet, but the stylesheet is hardcoded in the status bar widget currently. (It'll be easy to extract that out if we do per-screen status bar stylesheets, though I think having the same style for the status bar throughout provides uniformity which is nice.)

PR #2196, opened 2023-01-21
[wicket] fix help text

sunshowers:
Was missing closing angle brackets in a few places, which stood out a
bit.


andrewjstone (approved):
Thanks!

PR #2206, opened 2023-01-23
Update pagination strategy to prevent duplication

zephraph:
This is an attempt to consolidate pagination functionality such that the datastore layer is the only layer that needs to be aware of `id` or `name` differences. The handlers will treat all list endpoints as the same and only one implementation of listing an endpoint will be required in the app layer. 

There's a small about of verbosity in the handlers where multiple parameter types must be constructed to retrieve different sources of information, but it's largely duplicative and can likely just be refactored into a shared function. I've kept it flat for now. I'm generally unhappy with the names of the different parameter types and how we refer to them in parameters but again that feels like something that could be tidied up later after the base implementation.


nexus/src/external_api/http_entrypoints.rs
```diff
@@ -490,26 +490,18 @@ async fn silo_list(
 ) -> Result<HttpResponseOk<ResultsPage<Silo>>, HttpError> {
     let apictx = rqctx.context();
     let nexus = &apictx.nexus;
+    let query = query_params.into_inner();
```
davepacheco, line 493: nit: a downside of putting this stuff outside the `handler` async block is that the request timers won't account for the time required to do this.  Hopefully it's not much but that's generally why I try to keep almost everything inside that block.

davepacheco (approved):
Nice!  This seems like a lot less boilerplate overall.

nexus/src/external_api/http_entrypoints.rs
zephraph: Cool, I didn't realize that. I'm happy to move stuff back into the handler.

PR #2191, opened 2023-01-20
Update to latest MGS with a single shared UDP socket

jgallagher:
Most of the changes in this PR are miscellaneous fallout from minor type / API changes made in MGS over the past few weeks. The most substantial changes are in `ManagementSwitch::new()`, which now either creates a `SharedSocket` (for real deployments) or a UDP socket per SP (for sim deployments and tests). This is a little messy but seemed like the most straightforward way to continue to support running MGS proper without being in a real switch zone.

PR #2207, opened 2023-01-23
Rework MGS's config schema for switch zone interfaces

jgallagher:
We now have a `[[switch.port]]` list instead of requiring keying by port number, which allows us to remove several `HashMap`s in favor of `Vec`s. Our internal `SwitchPort` type is now purely logical and does not relate to anything physical (it's a newtype around an index into a `Vec<_>` of some type derived from the corresponding `[[switch.port]]` entry). Each `[[switch.port]]` config block now also specifies an explicit `ignition-target` number instead of implicitly assuming the port number and ignition target number matched.

This builds on #2191 and should be merged after it. For review, it may make sense to start by looking at the changes to the `config.toml` files to see the new definitions; all the code changes are fallout from wanting the config to look like it does now.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3620,7 +3620,6 @@ dependencies = [
  "schemars",
  "serde",
  "serde_json",
- "serde_with",
  "slog",
  "slog-dtrace",
  "sp-sim",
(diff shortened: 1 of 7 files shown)


andrewjstone (approved):
LGTM

PR #2212, opened 2023-01-23
[MGS] sled-agent launches and reconfigures MGS in the switch zone

jgallagher:
There are three related changes in this PR:

1. MGS can now start multiple dropshot servers (that all share the same server context), allowing it to listen on multiple addresses. In particular for us, this includes localhost (for talking to wicketd before the rack is setup) and the underlay network (once it exists, for talking to nexus). The addresses on which MGS is listening can be changed while MGS is running.
2. MGS can now read its own SMF properties (there are only two: the UUID identifying our MGS instance, and the list of addresses on which to listen). If sent `SIGUSR1`, MGS will reread its SMF properties and change its listening addresses if they've changed. (Changes to the ID are not allowed.) MGS's SMF manifest now includes `:kill -USR1` as its refresh method, allowing `smf refresh mgs` to work.
3. `sled-agent` sets the relevant properties for MGS to listen on localhost (always) and the underlay network (when it exists).

In combination with other recent changes to MGS (particularly #2207), this PR should allow sled-agent to successfully launch and configure MGS within a real switch zone.


Cargo.toml
```diff
@@ -108,6 +108,7 @@ criterion = { version = "0.4", features = [ "async_tokio" ] }
 crossbeam = "0.8"
 crucible-agent-client = { git = "https://github.com/oxidecomputer/crucible", rev = "0e2bea7741c6bdc34e8e08addb941ee31e4ac9ec" }
 crucible-client-types = { git = "https://github.com/oxidecomputer/crucible", rev = "0e2bea7741c6bdc34e8e08addb941ee31e4ac9ec" }
+crucible-smf = { git = "https://github.com/oxidecomputer/crucible", rev = "5f06c7ebc320c19846670e9d745a4d44f78901b4" }
```
jgallagher, line 111: This `rev` is different from the one used in `crucible-agent-client` and `crucible-client-types`. This seems to be okay, although it means two checkouts of the same repo. I tried bumping the other two to match and ran into multiple type definition problems that looked like propolis needed an update first, maybe? I'm happy to address this in whatever way makes sense:

a) leave as-is, and crucible will be updated the next time it needs to be in the normal course of it being updated
b) leave as-is on this PR, but try to follow up and update crucible ASAP
c) update crucible as a part of this PR, even if that means changes not directly related to MGS

smklein: My vote is for (b), fwiw

gateway/src/bin/mgs.rs
```diff
@@ -51,8 +84,165 @@ async fn do_run() -> Result<(), CmdError> {
...
+
+    let prop_id = config
+        .get_property(PROP_ID)
+        .map_err(|err| Error::GetProperty { prop: PROP_ID, err })?
+        .ok_or_else(|| Error::MissingProperty { prop: PROP_ID })?
+        .value()
+        .map_err(|err| Error::GetValue { prop: PROP_ID, err })?
+        .ok_or(Error::MissingValue { prop: PROP_ID })?
+        .as_string()
+        .map_err(|err| Error::ValueAsString { prop: PROP_ID, err })?;
```
smklein, line 205: Thanks for the rock-solid error handling here.

gateway/src/lib.rs
```diff
@@ -44,14 +50,65 @@ pub fn run_openapi() -> Result<(), String> {
...
+type HttpServer = dropshot::HttpServer<Arc<ServerContext>>;
+
 pub struct Server {
     /// shared state used by API request handlers
     pub apictx: Arc<ServerContext>,
-    /// dropshot server for requests from nexus
-    pub http_server: dropshot::HttpServer<Arc<ServerContext>>,
+    /// dropshot servers for requests from nexus or wicketd, keyed by their bind
+    /// address
+    pub http_servers: HashMap<SocketAddrV6, HttpServer>,
```
smklein, line 63: There's pretty tight coupling between `http_servers` and `all_servers_shutdown`, right?

Should we be exposing `http_servers` as `pub`? Seems like that risks a caller adding a server without waiting on it, or removing a server without awaiting termination.

gateway/src/lib.rs
```diff
@@ -77,34 +138,109 @@ impl Server {
...
+        mem::swap(&mut http_servers, &mut self.http_servers);
+
+        for (addr, server) in http_servers {
+            info!(
+                self.apictx.log,
+                "adjusting dropshot addresses; stopping old server";
+                "addr" => %addr,
+            );
+            server.close().await?;
+        }
```
smklein, line 241: Does this entry still exist in `all_servers_shutdown` when we do this, or does it get removed from that `FuturesUnordered` if someone is `await`-ing it via `wait_for_finish`?

sled-agent/src/services.rs
```diff
@@ -694,12 +694,21 @@ impl ServiceManager {
...
+                    // Always tell MGS to listen on localhost so wicketd can
+                    // contact it even before we have an underlay network.
+                    smfh.addpropvalue(
+                        "config/address",
+                        &format!("[::1]:{MGS_PORT}"),
+                    )?;
+
                     if let Some(address) = request.addresses.get(0) {
-                        smfh.setprop(
+                        smfh.addpropvalue(
```
smklein, line 706: Just to clarify, this second call to `addpropvalue` **appends** to the list of values, rather than replacing it, right?

smf/mgs/manifest.xml
```diff
@@ -12,14 +12,31 @@
...
 
   <property_group name='config' type='application'>
+    <!--
+      `config/id` is expected to have a single value; sled-agent will replace
+      "unknown" with a UUID as a part of starting us.
+    -->
     <propval name='id' type='astring' value='unknown' />
-    <propval name='address' type='astring' value='unknown' />
+    <!--
+      `config/address` is allow to have multiple values, so we do not seed it
```
smklein, line 36:
```suggestion
      `config/address` is allowed to have multiple values, so we do not seed it
```

smklein (approved):
Looks great - solid work integrating all these gnarly pieces together (underlay initialization, SMF services and refresh, juggling multiple dropshot servers, etc)!

A few comments below, but nothing blocking.

gateway/src/lib.rs
jgallagher: There is, yes. I don't think any of `Server`'s fields ought to be pub, generally, but when I tried making them private I bumped into issues with the integration tests poking at things, and I put pub back for expedience. I'll see if I can make these private and expose more robust pub methods that the tests can use.

gateway/src/lib.rs
jgallagher: It does still exist in `all_servers_shutdown`. `wait_for_finish` loops until `all_servers_shutdown` is completely empty, though, so happens is:

0. Some task starts `await`ing `wait_for_finish`. Call this task A.
1. Task B calls this method to reconfigure servers.
2. Any new servers are started above this loop, and close handles for them are inserted into `all_servers_shutdown`. (Similarly, any existing servers are left running and left in `all_servers_shutdown`.)
3. Any servers being shut down are closed here.
4. Task A will see the "wait for finish" futures for the servers from step 3 pop out, but `wait_for_finish` the method still won't return, because the `FuturesUnordered` still contains the futures from step 2.

A caveat is that if any `server.close()` _fails_, the future in the `FuturesUnordered` will also fail with the same error, so `wait_for_finish` will return the error immediately. I think this is okay? If something goes wrong refreshing and we can't shut down a server, the safest thing to do is probably restart.

sled-agent/src/services.rs
jgallagher: Correct! That is the primary difference between `setprop` and `addpropvalue`.

gateway/src/lib.rs
andrewjstone: Great explanation, thanks!

andrewjstone (approved):
This is all super clear. It's great to see this all come to fruition!

gateway/src/lib.rs
jgallagher: Fixed in 876a610a: the fields are now private, and `Server` grew a few methods to support integration tests that don't expose us to someone mucking about with `http_servers`/`all_servers_shutdown` directly.

PR #2230, opened 2023-01-25
[wicket] fix rack heights, fail gracefully if there isn't enough space

sunshowers:
This PR bundles a few things:

* Fix the max_height < 37 case: it should be 20, not 21.
* Fail gracefully when the height is too short: don't panic.
* Add a bottom margin of 2 to prepare for a status bar at the bottom
  (https://github.com/oxidecomputer/omicron/pull/2195).

### DIFF
--- a/wicket/src/widgets/rack.rs
+++ b/wicket/src/widgets/rack.rs
@@ -14,11 +14,14 @@ use crate::screens::TabIndex;
 use slog::Logger;
 use std::collections::BTreeMap;
 use tui::buffer::Buffer;
+use tui::layout::Alignment;
 use tui::layout::Rect;
 use tui::style::Color;
 use tui::style::Style;
+use tui::text::Text;
 use tui::widgets::Block;
 use tui::widgets::Borders;
+use tui::widgets::Paragraph;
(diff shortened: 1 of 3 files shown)


wicket/src/widgets/rack.rs
```diff
@@ -255,27 +264,35 @@ impl RackState {
...
+        match &self.component_rects {
+            ComponentRects::Displayed { rects_map, .. } => {
+                // Find the interesecting component.
+                // I'm sure there's a faster way to do this with percentages, etc..,
+                // but this works for now.
+                for (id, rect) in rects_map {
+                    let mouse_pointer = Rect { x, y, width: 1, height: 1 };
+                    if rect.intersects(mouse_pointer) {
+                        if self.hovered == Some(*id) {
+                            // No chnage
```
andrewjstone, line 276: I know it's my typo, but I just noticed. Feel free to fix or leave for later :)

andrewjstone (approved):
Ship it!

wicket/src/widgets/rack.rs
sunshowers: Fixed.

Issue #2211, opened 2023-01-23
sleds should propagate serial number during "self-reporting" API to nexus


smklein:
```
BRM42220004 # prtconf -v /devices | grep -A1 baseboard-identifier
        name='baseboard-identifier' type=string items=1
            value='BRM42220004'
```

Nabbing this info and plumbing it up to Nexus means that Nexus can correlate "self-reported" information about sleds with "MGS-reported" information about sleds.

jclulow:
NB: The baseboard identifier is not technically sufficient, you also want the model and revision, I believe.  I am not sure that the BRM* portion of the serial number is unique on its own.

smklein:
Will be completed by https://github.com/oxidecomputer/omicron/pull/2258

PR #2231, opened 2023-01-25
Update sled's model to track slot and serial number

zephraph:
I'm working on an early version of inventory and the first requirement is knowing what sleds are present in what slots. The aim of this PR is to just provide the implementation for the data model in CRDB and the API. Ensuring crdb has an accurate view of the world will be handled in another PR. 

---

My current understanding of how this'll all need to work is that Nexus will need to request `GET /sp` from MGS which'll return a list of all SP statues which includes a serial number and other info provided by ignition like if it's present, the power state, etc. Once we have a response from MGS we'll need to do some level of reconciliation with the information already in CRDB for sleds. 

@jgallagher relayed that there's a potentially long delay where `/sp` could take up to 20 seconds with retries. Given that, we'll likely want to have a process polling in the background to keep the data in CRDB up to date. Which instance of nexus that does the polling or how we coordinate that is a whole other can of worms.


jgallagher:
A couple minor comments:

* Josh [pointed out](https://github.com/oxidecomputer/omicron/issues/2211#issuecomment-1401113355) that we also need the model and revision, not just the serial number. MGS `/sp` doesn't return those on `main` right now but it's a tiny change to include them. The model is also a string (for current gimlets, it's `913-0000019`), and the revision is an integer.
* We have control on how long the upper bound for MGS's `/sp` is. With the current config it's closer to 10 seconds than 20, but I don't think 10s vs 20s is super relevant for these purposes. For context: MGS is communicating with the SPs via UDP, so it has retries and timeouts and a cap on eventually giving on up any given request. `/sp` tries to contact _all_ SPs in a full rack; if the rack is underpopulated or fully populated but an SP isn't responding, it will always hit the timeout before returning. We could make this smarter in terms of not trying to contact unpopulated slots (with some care to make sure we don't miss recently-inserted sleds), but we'll always have the potential timeout caused by a present-but-currently-unresponsive SP.

zephraph:
Thanks for the feedback and clarification. Do you think we'll need to read these fields individually? I wonder if we could just join them all into the same string to reduce what we need to compare. Also, do you know off hand the potential length of the serial and model numbers?

rmustacc:
> Thanks for the feedback and clarification. Do you think we'll need to read these fields individually? I wonder if we could just join them all into the same string to reduce what we need to compare. Also, do you know off hand the potential length of the serial and model numbers?

To create a unique ID in the database (e.g. the server's id) we'll want to concatenate them. However, they'll want to be stored separately for easy computation on them. Right now the model numbers are duro CPNs. Right now our model numbers aka CPNs are at least 11 ascii characters `XXX-XXXXXXX` (ref https://rfd.shared.oxide.computer/rfd/0308#_oxv2_guidance but note the image for the 0xv2 scheme is incorrect cc @eanders01). I would probably assume they could be bigger and pad that. I would not assume they will be that way forever, but we will have some knowledge of changes. Similarly, serial numbers use the format discussed in that section and are 11 characters (different design), but I would assume that could grow a little bit.

zephraph:
It seems like we may just want to store the machine versions as described in the RFD. We could add an implementation to grab specific values, but I can't see offhand what computation we'd want to do with any specific value. Regardless, we can add affordances for that (and padding) with a single field.

nexus/db-model/src/sled.rs
```diff
@@ -55,6 +60,11 @@ impl Sled {
             ip: ipv6::Ipv6Addr::from(addr.ip()),
             port: addr.port().into(),
             last_used_address,
+            // TODO: these fields should be populated
```
jgallagher, line 63: Apologies if this is a dumb question, but I'm not sure from the context - at least `slot` of these has to come from MGS, but the `id`/`addr` definitely do not come from MGS. Is it reasonable to expect `Sled::new()` to already have combined info from both sled-agent and MGS?

zephraph: Possibly! I'm not really sure. Maybe a better question for @smklein? I'm running off the default assumption that the information from MGS is populated at a later time. If we had that initial information from when the sled record was created though, that would be nice.

smklein: Well, right now we don't have any integration with MGS, so the naive answer for today is "we only have the self-reported information from the sled".

Admittedly, the sled *can*- and maybe *should* - self-report the part/serial/revision values as part of the call to nexus.

The handoff looks like the following:

- Sled agent boots
- Sled agent [sends a request to Nexus, indicating that it's alive](https://github.com/oxidecomputer/omicron/blob/5c0ac704e81ae79370ee743cb5168d7aeb949e6c/nexus/src/internal_api/http_entrypoints.rs#L66-L86)
- Nexus calls this function, stores the record for the sled in the DB.

In the future, we could make Nexus' implementation of this endpoint call MGS if we don't already know the slot number of the sled - that would give us an opportunity to ensure that "all sleds known to Nexus have a non-optional slot". **Personally, I would prefer if we required some of this information up-front, rather than hoping it gets back-filled later**. Trying to operate normally with sleds that have unknown identifiers and unknown versions seems like it'll just cause us pain.

(related: In the future, it might make sense to split Nexus' awareness of sleds in two. Currently, we only really know about sleds if the host has fully booted. Later, it probably makes sense to track "sleds that are visible to MGS" and "sleds which have fully booted their hosts" as separate entities)

zephraph: I think ultimately it will have to. As I understand it, without having the part/serial/revision we actually have no way of matching Nexus' notion of a sled to that MGS returns. If we know which cubby the sled is in we can make call MGS to get more information about it but because sleds can be physically moved that feels like a weak guarantee to me. I haven't wired it up that way in this PR because none of those mechanisms exist yet.

I know it's an aside, but I'm a little hesitant about having multiple notions of a sled. It seems to me that that would be much harder to reconcile from a product perspective and be a potential source of syncing bugs.

zephraph:
Closing this in favor of #2258

PR #2233, opened 2023-01-25
[MGS] Bring gateway-cli closer to parity with faux-mgs

jgallagher:
This PR reworks `gateway-cli` (which had languished quite a bit) to be closer to in parity with `faux-mgs`, with a few notable exceptions:

1. The SP the user wants to talk to is specified in terms of a logical / physical identifier (e.g., `sled/7` is "the gimlet in slot 7") instead of an IP address or interface. The only network connection made by gateway-cli itself is to MGS.
2. There is no `discover` subcommand. MGS uses discovery packets to find and map out online SPs, but doesn't expose those to its clients.
3. The `IgnitionLinkEvents` and `ClearIgnitionLinkEvents` commands are omitted. My current understanding is that these are for manual ignition debugging only and not events we want the control plane to care about in general. If that changes, they're in the MGS protocol already and would need endpoints added.
4. A few commands have no MGS endpoint; these are left as `todo!()`s that I will address in followup PRs that touch both MGS proper and `gateway-cli`.

### DIFF
--- a/gateway-cli/src/usart.rs
+++ b/gateway-cli/src/usart.rs
@@ -1,0 +1,263 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+// Copyright 2022 Oxide Computer Company
+
+use anyhow::Context;
+use anyhow::Result;
+use gateway_messages::SpComponent;
+use gateway_sp_comms::AttachedSerialConsoleSend;
+use gateway_sp_comms::SingleSp;
+use std::fs::File;
+use std::io;
+use std::io::Write;
+use std::mem;
+use std::os::unix::prelude::AsRawFd;
+use std::path::PathBuf;
+use std::time::Duration;
+use termios::Termios;
+use tokio::io::AsyncReadExt;
+use tokio::sync::mpsc;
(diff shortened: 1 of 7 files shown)


gateway-cli/src/usart.rs
```diff
@@ -66,41 +71,43 @@ pub(crate) async fn run(
...
 
                 flush_delay.start_if_unstarted().await;
             }
 
-            chunk = console_rx.recv() => {
-                let chunk = chunk.unwrap();
+            message = ws.next() => {
+                let chunk = match message {
+                    Some(Ok(Message::Binary(chunk))) => chunk,
+                    Some(Ok(Message::Close(..))) | None => {
```
andrewjstone, line 103: I'm not really familiar with websocket usage. 

Why do we have to close the websocket on line 89, but not here?

andrewjstone (approved):
LGTM.

gateway-cli/src/usart.rs
jgallagher: On line 89, we're getting an indication from stdin that the user of `gateway-cli` wants to stop, so we close the websocket to politely disconnect from MGS. Here, it's the websocket that's told us it's closing (i.e., MGS is kicking us out), so we don't have to close it from our side.

andrewjstone: Thanks for the clarification.

PR #2238, opened 2023-01-25
[MGS] Replace ignition/power-{on,off} with one ignition-command endpoint

jgallagher:
This addresses one "todo missing MGS endpoint" (sending ignition power-reset to an SP) in gateway-cli.

While we're here, remove the unimplemented and currently-unplanned endpoints to power on/off individual components.


andrewjstone (approved):
ship it!

PR #2239, opened 2023-01-25
[MGS] Add endpoints to get/set a component's active firmware slot

jgallagher:


andrewjstone (approved):
Ship it!

gateway/src/http_entrypoints.rs
```diff
@@ -297,6 +297,15 @@ enum PowerState {
     A2,
 }
 
+/// Identifier for an SP's component's firmware slot; e.g., slots 0 and 1 for
+/// the host boom flash.
```
andrewjstone, line 301: Just noticed this typo. Presumably you meant 'host boot rom`

gateway-cli/src/main.rs
```diff
@@ -336,8 +337,25 @@ async fn main() -> Result<()> {
         Command::IgnitionCommand { sp, command } => {
             client.ignition_command(sp.type_, sp.slot, command).await?;
         }
-        Command::ComponentActiveSlot { .. }
-        | Command::StartupOptions { .. } => {
+        Command::ComponentActiveSlot { sp, component, set_slot } => {
+            if let Some(slot) = set_slot {
+                client
+                    .sp_component_active_slot_set(
+                        sp.type_,
```
zephraph, line 344: What's the significance of the trailing `_` on type?

jgallagher: The OpenAPI spec [says this should be `type`](https://github.com/oxidecomputer/omicron/blob/8dc98ac9e6d6fd6f0b177909af7895060eef7f9b/openapi/gateway.json#L975-L977), but `type` is a reserved word, so progenitor [appends an _](https://github.com/oxidecomputer/progenitor/blob/32f2089f4100613a60dd0bc72674e37f0a72966a/progenitor-impl/src/util.rs#L116-L121) when defining the field name.

gateway/src/http_entrypoints.rs
jgallagher: I do! Thanks, fixed.

PR #2241, opened 2023-01-26
[wicketd] fix delayed status of updates from MGS

sunshowers:
Previously, we were only sending notifications over this channel if the
inventory had changed. This meant that we weren't getting newer instants
if the inventory was unchanged, causing wicket to say "delayed" rather
than "live".

Tested by setting up MGS and sp-sim, then ensuring that the status was
always "live".

### DIFF
--- a/gateway/tests/integration_tests/setup.rs
+++ b/gateway/tests/integration_tests/setup.rs
@@ -92,7 +92,7 @@ pub async fn test_setup_with_config(
     test_name: &str,
     sp_port: SpPort,
     mut server_config: omicron_gateway::Config,
-    sp_sim_config: &mut sp_sim::Config,
+    sp_sim_config: &sp_sim::Config,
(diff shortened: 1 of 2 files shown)


andrewjstone (approved):
LGTM

gateway/tests/integration_tests/setup.rs
```diff
@@ -92,7 +92,7 @@ pub async fn test_setup_with_config(
     test_name: &str,
     sp_port: SpPort,
     mut server_config: omicron_gateway::Config,
-    sp_sim_config: &mut sp_sim::Config,
```
sunshowers, line 95: Accidentally put this change in, though I think it makes sense.

andrewjstone: Agree it makes sense, as long as tests still pass, which they should if it builds.

PR #2244, opened 2023-01-28
[tufaceous] split out into a library/binary pair

sunshowers:
Turn the main parts of tufaceous into a library, so that it can be used
by wicketd.

Also add tests to ensure the end-to-end repository creation flow works.


tufaceous-lib/src/date.rs
```diff
@@ -3,7 +3,7 @@ use chrono::{DateTime, Duration, Timelike, Utc};
 
 /// Parser for datelike command line arguments. Can accept a duration (e.g.
 /// "1w") or an ISO8601 timestamp.
-pub(crate) fn parse_duration_or_datetime(s: &str) -> Result<DateTime<Utc>> {
+pub fn parse_duration_or_datetime(s: &str) -> Result<DateTime<Utc>> {
```
david-crespo, line 6: nitpick but since this is only used by the CLI it could go in the other crate. it could even be inlined into `main.rs`

david-crespo (approved):
It's a nice way to chop up the code anyway

sunshowers:
Yeah, I'm planning to use tufaceous-lib in wicketd, for both the main flow and for tests.

PR #2245, opened 2023-01-28
Wicket packaging

andrewjstone:
Package up `wicketd` and `wicket` with omicron-package. I tested this out locally, and it seemed to work.

There are a couple things remaining to do:
 * We need to be able to have the `BOOTSTRAP_ARTIFACT_PORT` listen on a bootstrap address. It's currently hardcoded to use localhost which is wrong. 
 * Should sled-agent construct a config file for wicket so it knows how to contact wicketd? Currently the wicketd port is just hardcoded inside wicket.

The first should probably be done in this PR. The second can likely wait until we figure out how we are going to launch wicket.

### DIFF
--- a/common/src/address.rs
+++ b/common/src/address.rs
@@ -40,7 +40,7 @@ pub const OXIMETER_PORT: u16 = 12223;
 pub const DENDRITE_PORT: u16 = 12224;
 pub const MGS_PORT: u16 = 12225;
 pub const WICKETD_PORT: u16 = 12226;
-pub const WICKETD_ARTIFACT_PORT: u16 = 12227;
+pub const BOOTSTRAP_ARTIFACT_PORT: u16 = 12227;
 
 pub const NEXUS_INTERNAL_PORT: u16 = 12221;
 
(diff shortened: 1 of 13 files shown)


common/src/address.rs
```diff
@@ -39,6 +39,8 @@ pub const CLICKHOUSE_PORT: u16 = 8123;
 pub const OXIMETER_PORT: u16 = 12223;
 pub const DENDRITE_PORT: u16 = 12224;
 pub const MGS_PORT: u16 = 12225;
+pub const WICKETD_PORT: u16 = 12226;
+pub const WICKETD_ARTIFACT_PORT: u16 = 12227;
```
sunshowers, line 43: Neat, I was wondering if we had a port registry :)

Note that this isn't strictly speaking the wicketd artifact port, since sled-agent can also serve artifacts on this port. I've been calling it the "installinator artifact port", though I've been wondering if we should just call it the "bootstrap artifact port" or something (since it runs on the bootstrap network).

andrewjstone: Thanks. Fixed.

andrewjstone:
I had a brief chat in network channel with @rcgoodfellow, and he wasn't expecting to advertise bootstrap prefixes from the switch zone. He said it can be done, but the alternative is to move the artifact server into the sled-agent itself. We were going to do this anyway for recovery, so maybe having it in one place makes sense. More to be discussed tomorrow. 

CC @sunshowers

sunshowers:
Looks good, thanks!

sunshowers (approved):

PR #2246, opened 2023-01-28
[tufaceous] add global = true to global arguments

sunshowers:
This lets users specify these arguments as both:

```
tufaceous --repo foo init
```

(this previously worked)

and as:

```
tufaceous init --repo foo
```

(this didn't work previously)

Depends on #2244.


david-crespo (approved):

sunshowers:
Accidentally landed this on the dependent branch -- put up again as #2247. Since this is already accepted, I'll go ahead and land that one.

PR #2257, opened 2023-01-30
[nexus] Exposing physical disks in operator API

smklein:
Depends on https://github.com/oxidecomputer/omicron/pull/2176

- Restructure sled agent's "storage manager" to be better at reporting notifications to Nexus
- Report notifications of disk attachment and removal to Nexus
- Create DB structures to represent physical disks
- Expose some information about those disks in the public API

TODO, potentially after this PR:
- [ ] Record a better history of "which disks have been attached to which sleds"
- [ ] Make "Physical Disks" the parents of "Zpools", rather than their current "parent" (which is just "sled").

Part of https://github.com/oxidecomputer/omicron/issues/2036


nexus/src/external_api/http_entrypoints.rs
```diff
@@ -5333,6 +5334,44 @@ async fn sled_view(
...
+}]
+async fn physical_disks_list(
+    rqctx: RequestContext<Arc<ServerContext>>,
+    path_params: Path<SledPathParam>,
+    query_params: Query<PaginatedById>,
+) -> Result<HttpResponseOk<ResultsPage<PhysicalDisk>>, HttpError> {
+    let apictx = rqctx.context();
+    let nexus = &apictx.nexus;
+    let path = path_params.into_inner();
+    let query = query_params.into_inner();
```
zephraph, line 5353: Something @davepacheco noted in another PR that I'm starting to do with all the API is to put everything inside the handler that can go there. These are fairly minor, but the handler itself is what's instrumented with a timer. If you see any of the API endpoints that have everything nested in the handler, that's why. You don't need to worry about doing that here, I'll catch it when I go over this for V1.

common/src/sql/dbinit.sql
```diff
@@ -128,6 +128,46 @@ CREATE INDEX ON omicron.public.service (
     sled_id
 );
 
+CREATE TYPE omicron.public.physical_disk_kind AS ENUM (
+  'm2',
+  'u2'
+);
+
+-- A physical disk which exists inside the rack.
+CREATE TABLE omicron.public.physical_disk (
```
zephraph, line 137: I assume aspects like the state of the disk and its position will just be added later when we can get that context?

nexus/test-utils/src/resource_helpers.rs
```diff
@@ -71,6 +72,24 @@ where
         .unwrap()
 }
 
+pub async fn object_put<InputType, OutputType>(
```
zephraph, line 75: I'm surprised this didn't already exist!

sled-agent/src/storage_manager.rs
```diff
@@ -664,7 +682,218 @@ impl StorageWorker {
...
+                    new_disks.insert(devfs_path, disk);
+                }
+                Err(e) => {
+                    warn!(self.log, "Cannot parse disk: {e}");
+                    err = Some(e.into());
+                }
+            };
+        }
+
+        let mut disks = resources.disks.lock().await;
```
zephraph, line 726: Is this lock automatically released when it goes out of scope?

smklein: Yup, locks in rust typically implement a `Drop` which releases it

common/src/sql/dbinit.sql
smklein: Yeah. I didn't add it now because I was pretty explicitly instructed that "slot # != actual physical mapping" But yeah, that would go here if I had it.

zephraph (approved):
This looks good from an API perspective 👍. I'm less confident of my ability to give solid feedback on the `store_manager` with its usage of mutexes. That said, I did try to look through it and assuming the locks resolve out of scope nothing jumped out at me.

sled-agent/src/storage_manager.rs
zephraph: That's very cool.

nexus/src/external_api/http_entrypoints.rs
smklein: Heh I was actually just following the pattern set by other entrypoints - which appear to parse inputs before calling "the main body of the endpoint" in a handler - but you're totally right, that distinction doesn't really make sense, we might as well instrument the whole thing.

Done in 3ff99699e27d669da1704562354a46a56c6df80d

nexus/types/src/external_api/views.rs
```diff
@@ -285,14 +285,36 @@ pub struct Rack {
...
+pub struct PhysicalDisk {
+    #[serde(flatten)]
+    pub identity: AssetIdentityMetadata,
+
+    pub vendor: String,
+    pub serial: String,
+    pub model: String,
+
+    pub disk_type: PhysicalDiskType,
+}
```
zephraph, line 316:
```suggestion
pub struct PhysicalDisk {
    #[serde(flatten)]
    pub identity: AssetIdentityMetadata,

    pub vendor: String,
    pub serial: String,
    pub model: String,

    pub disk_type: PhysicalDiskType,
    pub sled_id: Uuid,
}
```

Including `sled_id` here is helpful. We include the parent's ID on most (though not all) resources. We should probably work to unify that.

smklein: Added, but as an `Option`, since I want to preserve the ability to tell users about disks which may no longer be attached to any sleds for historical purposes.

PR #2258, opened 2023-01-30
Plumb baseboard information from Sled Agent to Nexus

smklein:
Depends on a refactor made in https://github.com/oxidecomputer/omicron/pull/2176

- Uses `libdevinfo` to grab `baseboard-*` fields about Gimlets
- Plumbs that information up to Nexus

Fixes https://github.com/oxidecomputer/omicron/issues/2211

TODO:
- [ ] Add tests
- [ ] Make it easier to "inject fake values" here, possibly via sled agent configuration
- [ ] Actually index on these values in Nexus, rather than storing them as arbitrary strings. E.g., we should probably ensure there aren't duplicates across the fleet.


common/src/sql/dbinit.sql
```diff
@@ -81,6 +81,11 @@ CREATE TABLE omicron.public.sled (
     /* Idenfities if this Sled is a Scrimlet */
     is_scrimlet BOOL NOT NULL,
 
+    /* Baseboard information about the sled */
+    baseboard_identifier STRING(63) NOT NULL,
+    baseboard_model STRING(63) NOT NULL,
+    baseboard_revision INT8 NOT NULL,
```
zephraph, line 87: Does baseboard carry some semantic meaning that customers would understand? I'd based the context of the names off of [RFD-308#3.3](https://rfd.shared.oxide.computer/rfd/308#_human_readable_text) in #2231.

zephraph: Digging around and it seems like this is fairly standard:
- windows: `wmic baseboard`
- linux: `sudo dmidecode -t baseboard`

jclulow: We may want to call it something else in the customer visible view (e.g., Gimlet ID or Compute Sled ID or whatever marketing name we give the sled), but I think it's good that we're calling it one thing throughout the internals of the stack.

leftwo:
It looks like the sled slot/cubby number did not make it into this PR, though it seemed like from https://github.com/oxidecomputer/omicron/pull/2231 it may have been part of the original plan.

PR #2266, opened 2023-01-30
[tufaceous] add commands to archive and extract tuf repositories

sunshowers:
Add two commands:

1. `tufaceous archive`, which archives a repository into a zip file.
2. `tufaceous extract`, which extracts a repository from a zip file.

## Why zip and not tar?

The main reason is that zip supports random access to files and tar does
not.

I was hoping to just be able to read the repository out of the zip file
from memory, but I ran into https://github.com/awslabs/tough/pull/563
while implementing it. Once that is resolved (or we write our own TUF
crate) it should be possible to do that.

Regardless, I don't want to foreclose that option forever, so let's just
use zip for now rather than having to deal with a migration in the
future.

Depends on #2300.


sunshowers:
A thing I realized after today's update sync is that it's not completely clear to me whether a TUF repository managed by tufaceous represents a single, consistent version, or multiple versions.

Based on #2192 I think that a TUF repo hosted on the internet would contain all the versions, each with its own `artifacts.json`, but a TUF repo managed by tufaceous on local disk would describe the components corresponding to a single version.

Assuming that's the case: Do we need to write bridging code to take an artifacts.json from a hosted repo, download the artifacts and make a local repo representing a single update? (Or maybe CI should produce a tufaceous repo as an archive?)

sunshowers:
Another thought occurred to me on the ride home: are we going to have to create multiple archives, each with a different "profile" of included components? For example, the installinator needs host and control plane images, not individual zones, while other users (which ones?) would need to maintain individual zones as images. If so, how are we going to manage that?

Probably worth talking about this in/alongside our sync tomorrow.

iliana:
> A thing I realized after today's update sync is that it's not completely clear to me whether a TUF repository managed by tufaceous represents a single, consistent version, or multiple versions.
> 
> Based on #2192 I think that a TUF repo hosted on the internet would contain all the versions, each with its own `artifacts.json`, but a TUF repo managed by tufaceous on local disk would describe the components corresponding to a single version.

This matches with what I think. (I also think that the multiple-versions repo on the internet should be managed by tufaceous, so that we don't have two different tools maintaining a repository format that needs to be consistent.)

> Assuming that's the case: Do we need to write bridging code to take an artifacts.json from a hosted repo, download the artifacts and make a local repo representing a single update? (Or maybe CI should produce a tufaceous repo as an archive?)

That bridge code would be useful for us; we will need to make those zips for ourselves for use with wicket, and can re-sign the repository data to reflect the files available in the repository. (It may not be useful for people who don't have the signing keys, because they can't remove targets from the repository metadata.)

sunshowers:
> This matches with what I think. (I also think that the multiple-versions repo on the internet should be managed by tufaceous, so that we don't have two different tools maintaining a repository format that needs to be consistent.)

That makes sense! OK, in that case what I've named `OmicronRepo` we should probably call `UpdateRepo` or `SingleUpdateRepo` or something.

We may still want to think about generating multiple archives out of a single update spec. I can see it being implemented either within tufaceous or as a shell script on top of tufaceous. If we decide to do so within tufaceous, I can write a followup PR implementing this.

> That bridge code would be useful for us; we will need to make those zips for ourselves for use with wicket

What priority do you think that should have? I think we definitely don't need it for Manufacturing PVT1 but probably do for MVP.

iliana:
> I think we definitely don't need it for Manufacturing PVT1 but probably do for MVP.

Agree.

iliana (approved):

PR #2267, opened 2023-01-31
Bump russh to 0.36.0

iliana:
#2251 and #2252 need to be applied simultaneously to work.

(I idly wonder if this class of problem will ever work under Dependabot.)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5303,9 +5303,9 @@ dependencies = [
 
 [[package]]
 name = "russh"
-version = "0.35.0-beta.9"
+version = "0.36.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


lifning:
particularly silly because the [commit range](https://github.com/warp-tech/russh/compare/v0.35.0-beta.9...v0.36.0) dependabot shows for the one [*includes* the version bump in the other](https://github.com/warp-tech/russh/compare/v0.35.0-beta.9...v0.36.0#diff-c6ff2fa1eea6ca3500c9887218e568159fa0c609100b512c7e8a9ca734a88f6bR28), whose dependabot PR evidently is confused to the point that it it doesn't give a commit range at all

lifning (approved):

PR #2268, opened 2023-01-31
revert Cargo.lock change from #2263

iliana:
Not sure how #2263 ended up modifying Cargo.lock in that way but a combination of two changes broke CI; this should unbreak it.


zephraph (approved):
Thanks!

david-crespo (approved):
whoops and thanks

PR #2283, opened 2023-01-31
Populate system update tables with mock data at rack init

david-crespo:
At rack init, we will pull information on the available system updates from the TUF repo and use that to populate our tables relating to system update: system updates and update components. We will also populate information about the state of rack-level updateable components like the PSC. On sled init, we will populate the DB with updateable components associated with that sled.

While we figure out how to do those things, this PR populates the DB with mock data representing the above.

### What it looks like in the console

2023-01-31-populated-update

### DIFF
--- a/nexus/src/app/update.rs
+++ b/nexus/src/app/update.rs
@@ -12,7 +12,7 @@ use crate::db::lookup::LookupPath;
 use crate::db::model::UpdateArtifactKind;
 use chrono::Utc;
 use hex;
-use nexus_types::external_api::params;
+use nexus_types::external_api::{params, shared};
 use omicron_common::api::external::{
     self, CreateResult, DataPageParams, Error, ListResultVec, LookupResult,
     PaginationOrder,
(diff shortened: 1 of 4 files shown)


david-crespo:
@davepacheco helpfully pointed me to [this existing system](https://github.com/oxidecomputer/omicron/blob/main/nexus/src/populate.rs) for populating hard-coded data at startup time. I might not have time to convert this today. If not, I'll do it tomorrow.

nexus/src/app/update.rs
```diff
@@ -626,23 +740,24 @@ mod tests {
...
-        let high =
-            nexus.highest_component_system_version(&opctx).await.unwrap_err();
-        assert_matches!(high, external::Error::InternalError { .. });
+        //
+        // let low =
+        //     nexus.lowest_component_system_version(&opctx).await.unwrap_err();
+        // assert_matches!(low, external::Error::InternalError { .. });
+        // let high =
+        //     nexus.highest_component_system_version(&opctx).await.unwrap_err();
+        // assert_matches!(high, external::Error::InternalError { .. });
```
david-crespo, line 760: Because these are populated at startup, there's no longer a chance for these to 500. Not sure how to/whether to test this anymore.

PR #2284, opened 2023-01-31
[MGS] Add endpoints to get/set host startup options

jgallagher:


gateway/src/http_entrypoints/conversions.rs
```diff
@@ -270,3 +272,39 @@ impl From<gateway_sp_comms::SpInventory> for SpComponentList {
         Self { components: inv.devices.into_iter().map(Into::into).collect() }
     }
 }
+
+impl From<HostStartupOptions> for StartupOptions {
```
andrewjstone, line 276: Why create a separate type at all instead of just using `gateway_messages::StartupOptions` directly? Surely they can both be serialized as either hubpack or json, right?

andrewjstone: Hmm, I see that it's because of the bitflags vs bools. I think the conversion makes sense in this case. You can ignore this comment.

andrewjstone (approved):
Looks good.

PR #2286, opened 2023-01-31
[MGS] Add "component clear status" endpoint

jgallagher:
This is currently only useful with the `monorail` component of a sidecar's SP, for which it resets the packet counters for each port.

Builds on #2285 and should be merged after it.


andrewjstone (approved):
Ship it!

PR #2287, opened 2023-01-31
[MGS] Flesh out SpState information

jgallagher:
Adds serial number/model/revision, hubris image information, and RoT information.

Builds on #2286 and should be merged after it.

Fixes https://github.com/oxidecomputer/minimum-upgradable-product/issues/40.


gateway/src/http_entrypoints.rs
```diff
@@ -66,13 +66,78 @@ pub struct SpInfo {
 pub enum SpState {
```
andrewjstone, line 66: Oh hell yes! This is stuff wicket craves!

andrewjstone (approved):
I'm very excited to see this plumbing make it's way north.

PR #2293, opened 2023-02-01
[wicketd] upload TUF repositories and unpack artifacts from them

sunshowers:
TODO:

- [ ] Move repository validation logic to tufaceous-lib (check with iliana about this -- should it be optional? mandatory?)
- [ ] Test repository validation logic
- [x] Check whether we want to support arbitrary update kinds in installinator or just specific ones
  - Just specific kinds for now, we can always update the installinator in the future to support more kinds (if we support arbitrary kinds in the future, it will likely involve downloading a manifest and then applying updates from that)
- [x] Add non-zone kinds to `UpdateArtifactKind` ([maybe from the list here](https://github.com/oxidecomputer/omicron/blob/5c0ac704e81ae79370ee743cb5168d7aeb949e6c/wicket/src/update.rs#L28-L43))
  - Done in #2300. 
- [x] Test out flow with wicket, wicketd and installinator
  - Done by @jgallagher in #2375.
- [ ] Write integration tests for wicket, wicketd and installinator

Depends on #2266.


sunshowers:
This is ready for review -- I haven't done a few of the TODO items but I think we can get this in for now, then build on it to write end-to-end and integration tests later.

installinator/src/peers.rs
```diff
@@ -371,9 +372,16 @@ impl ArtifactClient {
     }
 
     async fn fetch(&self, artifact_id: ArtifactId, sender: FetchSender) {
+        // TODO: should the installinator try and handle artifact kinds that are
+        // currently unknown to it? (Maybe if a destination is also specified?)
+        // If so, we'll want to change this to a string.
```
jgallagher, line 377: My gut feeling is "no, and if it needs to we need to update installinator to handle the new kind first". I don't think we need to be too anticipatory here (and hopefully I don't regret this comment in hindsight!).

wicketd/src/http_entrypoints.rs
```diff
@@ -59,21 +58,26 @@ async fn get_inventory(
...
 
-/// An endpoint used to upload artifacts to the server.
+/// Upload TUF repositories to the server.
+///
+/// At any given time, wicketd will keep at most one repository in memory. Any
+/// previously-uploaded repositories will be discarded.
 #[endpoint {
     method = PUT,
-    path = "/artifacts/{name}/{version}",
+    path = "/repositories/{name}/{version}",
```
jgallagher, line 67: I'm not sure what the name/version are buying us now that we hold at most one repository. If we need to identify it, should that identification come from metadata within the repository itself instead of args from the uploader?

jgallagher (approved):
Looks great!

installinator-artifactd/src/http_entrypoints.rs
```diff
@@ -32,10 +33,13 @@ pub fn api() -> ArtifactServerApiDesc {
...
     method = GET,
-    path = "/artifacts/{name}/{version}"
+    path = "/artifacts/{kind}/{name}/{version}"
 }]
 async fn get_artifact(
     rqctx: RequestContext<ServerContext>,
+    // NOTE: this is an `ArtifactId` and not an `UpdateArtifactId`, because this
+    // code might be dealing with an unknown artifact kind. This can happen
+    // if a new artifact kind is introduced across version changes.
     path: Path<ArtifactId>,
```
sunshowers, line 43: Q for @iliana and @david-crespo: we introduce an `ArtifactId` here that is similar to `UpdateArtifactId`, except its kind is potentially unknown (unlike `KnownArtifactKind` which is an enum, `ArtifactKind` is just a newtype wrapper around a string).

For wicketd it makes sense to use that because we might be introducing new artifact kinds across versions, and wicketd is mostly a simple shoveler of data rather than something that cares about the content of each artifact.

Do the same considerations apply to Nexus and sled-agent? In other words, should Nexus and sled-agent use this `ArtifactId` rather than `UpdateArtifactId`? We'll still want to have `KnownArtifactKind` be around, but it wouldn't be a type that gets transmitted over the wire. As a bonus, it would also eliminate a big chunk of the complexity involved in adding a new known artifact kind.

installinator/src/peers.rs
sunshowers: Yeah, I think the answer being "no" here is fine. Importantly, unlike wicket, the installinator is part of the destination version, so there's no version skew issues there.

wicketd/src/http_entrypoints.rs
sunshowers: Good question! I'd put it in for logging purposes but we don't need it. Agreed that having a file with this would make sense, though we don't currently have a place to store that. (@iliana maybe in `artifacts.json`?)

I guess we should also change the name to `repository` rather than `repositories` in that case.

jgallagher: Hmm, that still has the problem that the name/version here are effectively arbitrary and coming from an unverifiable source outside the rack. How much logging detail do we need if someone uploads a bad repository?

sunshowers: I guess we don't really, no. I've updated the PR to remove the name and version.

installinator/src/peers.rs
sunshowers: Removed the comment.

PR #2299, opened 2023-02-01
Propagate physical CPU count and RAM size up to Nexus

smklein:
Propagate "CPU count" and "Physical RAM" information from Sled Agents up to Nexus, and through the external API.

This information will be used in the future to make provisioning decisions (e.g., do I have enough CPU / RAM to run the services I expect to run on a particular Sled).

It's also currently being plumbed through the operator API for visibility, though that representation of hardware may change to align with a more "holistic view of inventory".


openapi/nexus-internal.json
```diff
@@ -2138,6 +2138,19 @@
               }
             ]
           },
+          "cpus": {
+            "description": "Number of CPUs online on the sled.",
+            "type": "integer",
+            "format": "int64"
+          },
```
rmustacc, line 2145: I know we're trying to get the basics of provisioning going, but I think we don't really want to call this CPUs and we really want to be crisp about what we're tracking and why. Here are a few high-level thoughts with this as it applies to the PR as a whole.

We want to think in terms of the physical units: sockets, cores, and threads. While at the moment we're not performing core-aware scheduling, that is something that we know we need to do as a high-priority post-MVP item. That will change how we want to think about this and how we look at this. It also helps because the term `CPUs` is very ambiguous as to what it means. I think, even if we don't care about physical inventory right now that we want to think about this in terms of what this actually is, which is total threads in the system. If we'd rather call it `logical_cpus`, we could, though I think that term is equally vague and more OS-specific.

I think we likely want to think carefully about online. As what is going to be usable for guests will change, especially in the core aware world. Given that we're not going to be performing core-aware scheduling for a while and don't have a way to , I would consider just using the non-online value to start with. But wen we are, the online value is misleading for what a customer can use for VMs for two reasons:

* The moment a single thread in a core fails in a core-aware world, we can no longer perform VMM core-aware scheduling on that physical core.
* I think the more actionable thing for customers is understanding the total cores that are present, how many threads have failed, and what percent of the resources we're actually using for VMs are used.

openapi/nexus.json
```diff
@@ -13706,22 +13706,37 @@
...
+              }
+            ]
+          },
           "rack_id": {
+            "description": "The rack to which this Sled is currently attached.",
             "type": "string",
             "format": "uuid"
           },
           "service_address": {
+            "description": "The underlay address on which the Sled Agent is listening.",
```
rmustacc, line 13739: I realize this is just being updated to be more descriptive, but why is this in the external API? We can defer to a separate review of the JSON if useful.

openapi/nexus.json
```diff
@@ -13706,22 +13706,37 @@
...
+          "physical_ram": {
+            "description": "The total amount of physical RAM available to the sled.",
+            "allOf": [
+              {
+                "$ref": "#/components/schemas/ByteCount"
+              }
+            ]
+          },
           "rack_id": {
+            "description": "The rack to which this Sled is currently attached.",
```
rmustacc, line 13734: I think we may want to think carefully about the term rack_id as we have no good way of actually relating this to a physical rack. It's really a logical control plane instance id for a rack.

I mention this because this suggests that it can change due to moving the sled, but if we move all the sleds from rack A to rack B, it won't change. So I think we may want to be more precise. This may be something that should come up as a different follow up along with the bit below.

sled-agent/src/hardware/illumos/sysconf.rs
```diff
@@ -0,0 +1,40 @@
...
+            arg: argname.to_string(),
+            e: std::io::Error::last_os_error(),
+        });
+    }
+    Ok(r.try_into()?)
+}
+
+/// Returns the number of physical processors on this sled.
+pub fn online_processor_count() -> Result<u32, Error> {
+    Ok(u32::try_from(sysconf(
```
rmustacc, line 29: Separate to this, `sysconf` returns a long. In an ILP64 this is an i64. I realize we're unlikely to have more than a u32 of processors, but everywhere else you treat cpus as an `i64`.

sled-agent/src/hardware/illumos/sysconf.rs
```diff
@@ -0,0 +1,40 @@
...
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error("sysconf failed accessing {arg}: {e}")]
+    Sysconf { arg: String, e: std::io::Error },
+
+    #[error("Integer conversion error: {0}")]
+    Integer(#[from] std::num::TryFromIntError),
+}
+
+fn sysconf(argname: &str, arg: libc::c_int) -> Result<u64, Error> {
```
rmustacc, line 16: Just want to make sure its intentional that we're doing i64->u64 conversion on the return value here. I assume that we're relying on negative values only being returned for the `-1 case. But if that's not the case, won't we end up converting something to a large unsigned value?

If we're relying on this we should probably be a little defensive as other consumers are going to heartbeat this into the database if we have something wrong.

openapi/nexus.json
```diff
@@ -13706,22 +13706,37 @@
...
+        "description": "An operator's view of a sled.",
         "type": "object",
         "properties": {
           "baseboard": {
             "$ref": "#/components/schemas/Baseboard"
           },
+          "cpus": {
+            "description": "The number of online processors available to the sled.",
+            "type": "integer",
+            "format": "int64"
```
rmustacc, line 13718: Why is the number of CPUs a signed value. What does a negative number of CPUs mean?

rmustacc (commented):
I know that you're still working on this as a draft @smklein, but I wanted to add a bunch of notes about the current data we're grabbing from the system, what it represents, and the corresponding nomenclature. There are a few unrelated comments in the external JSON just because this is being updated in this change. If it's more appropriate to handle that a different way, that's fine. I just noticed it because it was updated in this change.

Happy to talk more if it'd help on the specifics live if that'd help. Thanks for working on this.

sled-agent/src/hardware/illumos/sysconf.rs
rmustacc: I guess the try info may make this safe and I misunderstood that due to make lack of familiarity. I might be explicit about why we're doing the type conversions somewhere to explain to a reader why we want the uniform u64.

nexus/db-model/src/schema.rs
```diff
@@ -390,6 +390,9 @@ table! {
         part_number -> Text,
         revision -> Int8,
 
+        cpus -> Int8,
+        physical_ram -> Int8,
```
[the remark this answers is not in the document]
smklein: Okay, I've updated these to `online_logical_cpus` and `usable_physical_ram`.

smklein (commented):
Thanks for the feedback @rmustacc ; I've tried to update both the names, types, and documentation to be more accurate.

openapi/nexus.json
```diff
@@ -13717,13 +13717,17 @@
             "type": "string",
             "format": "uuid"
           },
+          "online_logical_cpus": {
```
rmustacc, line 13720: I've been struggling with the nomenclature here. This is probably fine, but I'd like to write up the different thoughts here as part of trying to think about future things. Part of my concern is what does the term cpu or processor mean and trying to come up with something that is understandable to lots of people, but I suspect this is a fool's errand in the end.

Fundamentally in the world that we are operating in, we have CPUs (referring in this instance to thing that we buy from someone) that are installed into sockets. They have some number of cores and some number of logical units beneath that that represent schedulable work (trying to avoid terms for a moment). Those logical units are called different things. For example Intel calls it "hyper-threading" and sometimes threads. In cpuid leaf 0xb and 0x1f they describe it as SMT. The AMD PPR is more willing to use the term threads in family 0x17 and 0x19. The amd64 level calls them threads.

But amd64 also had a thing they did in family 0x15 that lead the next level to be "logical cores" because of how they used CMT (clustered multi-threading). Which means the question of what is or isn't a core is weird, though this also isn't a CPU or thing that we expect to see again.

What we currently call a logical CPU corresponds to threads, which is what most folks seem to agree on. I've tried to look this up in ARM, but don't see great consensus. RISC-V does call this same thing a HART, or hardware thread, an says a core is made up of one or more HARTs.

From a provisioning perspective in the short and longer term we're interested in how many usable logical CPUs and cores there are. From an operator inventory perspective and overall capacity planning (versus in the moment), where we assume all hardware is usable, the things I think about are: asking how many sockets does a server have (more informational), how many total cores are there and how many threads/logical CPUs here are.

Finally, the last thing with nomenclature here is what do we think of when we're provisioning guests, which is the fundamental customer-facing abstraction. Right now the API file describes folks as passing `ncpus` and defines it as a CPU which right now maps to logical CPUs / threads in the system.

So playing this out, it seems like we'd expect a server object if we continue this trend to have:

* online_logical_cpus (still not sure if we should maybe call this cores)
* online_physical_cores
* total_logical_cpus
* total_physical_cpus
* total_physical_sockets

We may opt to leave out sockets here because it's not immediately useful. It's also worth noting that in a core-aware scheduling world, `online_logical_cpus` is not the same as what's usable. It's also not clear to me what we want to present with respect to what the actual provisionable CPU is from an operator's perspective, that is what is usable for guests, which is something that they care about, because it's probably not going to be all 128 even on a good day. This almost suggests hat we want a third category of `usable_xxx`, though this is starting to get complicated and more nuanced. Anyways, that's probably more on this than is useful.

sled-agent/src/hardware/illumos/sysconf.rs
smklein: I thought we didn't want to enable negative values for CPUs? By using a u32 here, I validate the value from the sled. If I propagated an i64 all the way up to Nexus, it would be possible to use an "invalid" value for CPU count for much longer. Shouldn't we be throwing an error as early as we reasonably can?

rmustacc: I'm honestly not sure what the right answer is to the fact unsigned types are for chumps in the db. Seems like probably best to just hope the constraint we have works and is never violated and so it can always serialize / de-serialize as a u32 and then what you're saying makes sense. Just whenever I see a constraint on that, my mind goes to what happens if that gets violated because of issue XXX and the wrong thing is on disk. As I said above, this probably isn't worth worrying about.

openapi/nexus.json
zephraph: Meghan actually if there was something we needed to do during manufacturing or setup to include information from the rack's nameplate.

@smklein this came up during the RFD-267 huddle too. Essentially it seems like we're going to be mostly unable to tell what sled occupies which rack (accurately, anyway). I believe this is going to be fairly problematic in a multi-rack deployment in that it both breaks some of the assumptions in the relational hierarchy and it'll also be difficult to communicate to an operator the physical location of a piece of hardware in their overall installation.

common/src/sql/dbinit.sql
```diff
@@ -86,6 +86,10 @@ CREATE TABLE omicron.public.sled (
     part_number STRING(63) NOT NULL,
     revision INT8 NOT NULL,
 
+    /* CPU & RAM summary for the sled */
+    online_logical_cpus INT8 CHECK (online_logical_cpus BETWEEN 0 AND 4294967295) NOT NULL,
```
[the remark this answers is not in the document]
rmustacc: Yes, though this gets nuanced in the future. See https://github.com/oxidecomputer/omicron/pull/2299#discussion_r1099457140 and the fact that we could have online logical cpus that aren't usable for VM guests in a core-aware scheduling world because you wouldn't have a working twin (but it could be usable for others).

openapi/nexus.json
zephraph: I'm worried that if we get too nuanced here it'll make the API harder to document and to reason about. I don't want to derail progress and perhaps it's something we can just put a pin in, but do other platforms define these terms in some way we could borrow? What would an operator typically see if they were using VMWare or Open stack or something? 

I have an uninformed perspective of this so my opinion counts for little here, but I wouldn't understand what `online_logical_cpus` means without reading documentation. Does `online` imply that there are `offline` versions too?

nexus/db-model/src/unsigned.rs
```diff
@@ -0,0 +1,113 @@
...
+    AsExpression,
+    Eq,
+    Ord,
+    PartialEq,
+    PartialOrd,
+    FromSqlRow,
+    Serialize,
+    Deserialize,
+)]
+#[diesel(sql_type = sql_types::Int4)]
```
zephraph, line 29: Should you also delete the `u16.rs` file given that it's defined here now?

openapi/nexus.json
zephraph: Thanks for the detailed info!

nexus/db-model/src/unsigned.rs
smklein: Removed

smklein:
I went ahead and updated the Nexus-level tracking of compute resources as "usable_hardware_threads", since that seems to be the unit we're going to care about in the near-term.

The terms "CPUs" and "Processors" are a little too vague for my liking - perhaps it'll be more clear when we're describing inventory, but from an accounting / overprovisioning point-of-view, I think we care about "how many things can we run at the same time".

We will likely add "usable_cores" in the future, in a core-aware world, but I think it will be reasonable to make that jump when we get there.

smklein:
Whenever y'all have time, this is ready for another look - I think I've addressed all feedback that I can.

smklein:
One more ping on this

rmustacc:
Sorry for the delay in getting back to you @smklein. I think the approach we have for threads makes sense for now. I'm still not really convinced that the DRAM value we're providing is going to be actionable and is going to need a lot more docs to make sense of because it doesn't help an operator answer the question of how much capacity is actually available for VMs on this sled. All that said, I don't see this as a reason to hold this up at all. It's a value we need in nexus, regardless of whether or not it's something we want to expose to the operator.

Issue #2289, opened 2023-01-31
Packaging: Make a decision on the packaging of the host OS vs zones


smklein:

ahl:
We decided to audit all repos to see where things land. Here's that work in progress:
https://docs.google.com/spreadsheets/d/1u249gK9bc9a1W7ri5_10lcum0QVQWr115-pj_YJDaeY/edit#gid=0

PR #2300, opened 2023-02-02
[common] add variants to UpdateArtifactKind

sunshowers:
Add the list of updatable components from [RFD 300](https://rfd.shared.oxide.computer/rfd/0300#_list_of_updatable_components).

Remove "zone" since we determined that individual zones are not
updatable.

This list might change in the future, but it gives enough for me to work
with wrt the installinator and such.

### Adding new artifact kinds

Adding these new update kinds was quite tricky for me, can it be made simpler? (See 
"Adding a new UpdateArtifactKind" in the PR.) Particularly concerned about what seemed to be a circular dependency: the command listed [here](https://github.com/oxidecomputer/omicron/blob/main/docs/adding-an-endpoint.adoc#5-testing) failed to compile.

### Representing artifact kinds as strings

We have a bit of a conflict between the idiomatic ways to represent artifact kinds:

* openapi-lint requires that enum variants have underscores in them (e.g. `gimlet_sp`)
* on the command line, convention typically is to use dashes in them (e.g. `gimlet-sp`).

There are a few ways to handle this:

1. Always use underscores everywhere, making the CLI somewhat less idiomatic. This is what I've currently chosen, but I'm happy to change this.
2. Use dashes in the `FromStr` and `Display` impls, and underscores in the `Serialize`/`Deserialize` impls. This will work but I'm worried about the discrepancy causing new bugs down the road.
3. Use underscores as the canonical representation, but add a separate command-line parser which recognizes both underscores and dashes.
4. Always use dashes everywhere, silencing the OpenAPI lint somehow. This seems to work with progenitor, but other client builders might have trouble with this if they don't accept dashes.


david-crespo:
Whew. Looks like it’s time to hook the update API up to this stuff.

smklein:
> What's in the "control plane" asset type? It seems like it must be a lot of different things like Nexus and Cockroach and DNS zone images and so on.
> 
> If that's true, I'm not sure we should be breaking the host OS ROM image and ramdisk image into two pieces like that?

Related: https://github.com/oxidecomputer/omicron/issues/2289

iliana:
> What's in the "control plane" asset type? It seems like it must be a lot of different things like Nexus and Cockroach and DNS zone images and so on.

Something we were looking at yesterday while chatting about #2289 was where to make a split, if any, in a phase 2 Helios image that also contained all the control plane bits. The theory is that if we ship an image like that, we can teach Nexus to understand how to split those things up later.

The most logical split we could think of immediately was that the Helios image could be separate from the control plane bits, and extracted into a separate ZFS dataset. But I am responding to this before I've had coffee today so I'm not sure I remember any particular benefits from doing that.

jclulow:
Is there a recording I can watch or whatever to try to catch up on what's been discussed already?  I definitely was not expecting to stick all of the control plane software into the OS ramdisk.

sunshowers:
Right, I'm currently assuming based on discussions with @ahl, @iliana and others, as well as RFDs 300 and 345, that we're going to have two separate images:

1. the host phase-2 Helios image
2. the control plane zone bundle image.

smklein:
> Is there a recording I can watch or whatever to try to catch up on what's been discussed already? I definitely was not expecting to stick all of the control plane software into the OS ramdisk.

https://drive.google.com/file/d/1G4y8uh76va-zjA_R5JCISbQ9is9YZzl1/view at 18:00 includes some context as well.

jclulow:
> Right, I'm currently assuming based on discussions with @ahl, @iliana and others, as well as RFDs 300 and 345, that we're going to have two separate images:
> 
>     1. the host phase-2 Helios image
> 
>     2. the control plane zone bundle image.

OK that makes sense.  I think we should just have a `HostOs` image then, which would contain the ROM (aka phase1) and the ramdisk (aka phase2) component.  The former is really just a subset of the files from the second one, and they need to be treated as a single unit with a single version identifier, etc.

sunshowers:
> OK that makes sense. I think we should just have a HostOs artefact then, which would contain the ROM (aka phase1) and the ramdisk (aka phase2) component. The former is really just a subset of the files from the latter, and they need to be treated as a single unit with a single version identifier, etc.

They're still separate updatable components though, right? For example, the installinator will write the Phase 2 Host image but not the Phase 1 Host image.

jclulow:
> They're still separate updatable components though, right? For example, the installinator will write the Phase 2 Host image but not the Phase 1 Host image.

I think it depends on how the system is going to arrange things with respect to update planning, and how we're going to identify that while there are two blobs, it's one _image_ with one version, and that the update is not complete until both blobs are put in the right place -- in the same way that a control plane bundle update will not be complete until the whole thing is extracted into the right parts of the file system on the sled.

sunshowers:
What do you think about having the phase 1 and phase 2 host images be two separate updatable components as listed in `artifacts.json`, but that we require that they always have the same version number?

jclulow:
I think what we probably want to produce as output from the Host OS image build is something like a single tar file that contains the `rom` and `zfs.img` that we produce today, and perhaps other metadata and signatures and so on.  It might also include more than one `rom` with different names that are configured differently; e.g., recently we had occasion to produce a suite of ROM files each with a different RAM frequency cap: `rom.ddr3200`, `rom.ddr2667`, etc.

In future Gimlet revisions, there could also be more or fewer blobs depending on changes we make to the board and how it boots, etc.  It might be sufficient for the whole image to be written to one onboard flash device large enough that it can meet the requirements we currently split between the SPI ROM and the M.2 slice in a Gimlet/C.

Different venues in the system would make use of different parts of the tar, but they can just ignore the parts they do not need; e.g., the update engine would fish out `rom` for the SP update, sled-agent would fish out `zfs.img` for the M.2 update, etc.

sunshowers:
The determination made in today's update sync was to have two images:

1. The host image.
2. The trampoline image.

Other code will be responsible for splitting up the host and trampoline images into phase 1 and 2.

I'll update this PR to reflect that some time tomorrow.

andrewjstone (approved):
LGTM!

common/src/api/internal/nexus.rs
```diff
@@ -113,12 +139,44 @@ pub struct UpdateArtifactId {
...
+mod tests {
+    use super::*;
+    use strum::IntoEnumIterator;
+
+    #[test]
+    fn update_artifact_kind_roundtrip() {
+        for kind in UpdateArtifactKind::iter() {
+            let as_string = kind.to_string();
+            let kind2 = as_string.parse::<UpdateArtifactKind>().unwrap_or_else(
+                |error| panic!("kind {as_string} parsed successfully: {error}"),
```
jgallagher, line 177: Should this be "parsed unsuccessfully" or maybe "failed to parse"?

jgallagher (approved):

common/src/api/internal/nexus.rs
sunshowers: Ah yeah! I'll fix it up in a followup (I'm thinking of redoing this code a bit anyway.)

sunshowers: Fixed in https://github.com/oxidecomputer/omicron/pull/2369.

PR #2316, opened 2023-02-03
[MGS] Use dropshot's web socket support for serial console endpoint

jgallagher:
The primary driver for this PR was removing the `TODO` in gateway-cli where the progenitor MGS client didn't know that the serial console endpoint was a websocket, so we had to manually connect via `tokio_tungstenite`. As a very nice side effect, we get to drop all the manual websocket upgrade / HTTP header stuff we had previously.

Highly recommend turning on "ignore whitespace" when reviewing - it cuts the diff down by about two thirds.


gateway/src/serial_console.rs
```diff
@@ -4,260 +4,150 @@
...
+    let ws_stream =
+        WebSocketStream::from_raw_socket(upgraded, Role::Server, Some(config))
+            .await;
+
+    let (ws_sink, ws_stream) = ws_stream.split();
+
+    // Spawn a task to send any messages received from the SP to the client
+    // websocket.
+    //
+    // TODO-cleanup We have no way to apply backpressure to the SP, and are
```
andrewjstone, line 56: Would it be easier, and perhaps just as useful to use a bounded channel here and drop data directly, rather than tell the SP to drop data?

andrewjstone (approved):
I see I commented on an old todo, but still think it's worth considering. Ship this sucker though :)

gateway/src/serial_console.rs
jgallagher: (Just for clarity - this is not new code and is unchanged by this PR)

We could, yeah. I'm not sure how to pick a bound for this channel - the unit we're sending on it is "packets from the SP" which have somewhere between 1 and ~1024 bytes in them, so maybe it's fine to just pick a pretty big bound?

andrewjstone: Yeah, I'd set it to just something quite large. Even 1MB means we have 1000 max sized packets queued up. So 10MB probably good enough.

jgallagher: Set it to 10,000 and added a bit of logging in 87124c2. Chipping away at those TODOs, thanks!

PR #2319, opened 2023-02-04
Fix fetch update deployment by ID

david-crespo:
I forgot to register the resource. Also updated the authz tests for update now that we're populating data up front.

### DIFF
--- a/nexus/tests/integration_tests/endpoints.rs
+++ b/nexus/tests/integration_tests/endpoints.rs
@@ -1516,28 +1516,12 @@ lazy_static! {
             )],
         },
 
-        /* Metrics */
-
         VerifyEndpoint {
-            url: &DEMO_SYSTEM_METRICS_URL,
+            url: "/v1/system/update/version",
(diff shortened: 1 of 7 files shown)


nexus/tests/integration_tests/endpoints.rs
```diff
@@ -1601,6 +1585,17 @@ lazy_static! {
...
+        /* Metrics */
+
+        VerifyEndpoint {
+            url: &DEMO_SYSTEM_METRICS_URL,
+            visibility: Visibility::Public,
+            unprivileged_access: UnprivilegedAccess::None,
+            allowed_methods: vec![
+                AllowedMethod::Get,
+            ],
+        },
```
zephraph, line 1597: The diff here is confusing, ha. This is just moved down but it wasn't clear at first glance.

zephraph:
This is great, thanks for doing it. Glad to not have `/v1/` endpoints in the authz ignorelist.

PR #2320, opened 2023-02-04
Add Bootstrap vnic and address to switch zone

andrewjstone:
Add a bootstrap vnic (over the bootstrap etherstub in the global zone) to the switch zone, along with a bootstrap IP address that can be listened on by the wicketd artifact server. The wicketd artifact server is used for serving OS and host images, and a follow up change will be made for it to listen on the bootstrap server. 

Currently we don't expect any other zones besides the switch and global zone to have access to the bootstrap network. Therefore we keep address allocation and plumbing minimal for now.

### DIFF
--- a/sled-agent/src/bootstrap/hardware.rs
+++ b/sled-agent/src/bootstrap/hardware.rs
@@ -134,7 +134,6 @@ impl HardwareMonitor {
             HardwareManager::new(log.clone(), sled_config.stub_scrimlet)
                 .map_err(|e| Error::Hardware(e))?;
 
-        // TODO: Add the etherstub for the bootstrap network here
         let service_manager = ServiceManager::new(
             log.clone(),
             underlay_etherstub.clone(),
--- a/sled-agent/src/bootstrap/hardware.rs
+++ b/sled-agent/src/bootstrap/hardware.rs
@@ -124,17 +124,20 @@ impl HardwareMonitor {
     pub async fn new(
         log: &Logger,
         sled_config: &SledConfig,
-        etherstub: Etherstub,
-        etherstub_vnic: EtherstubVnic,
+        underlay_etherstub: Etherstub,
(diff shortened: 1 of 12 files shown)


sled-agent/src/illumos/running_zone.rs
```diff
@@ -301,6 +301,7 @@ impl RunningZone {
                 // Re-initialize guest_vnic state by inspecting the zone.
                 opte_ports: vec![],
                 link: None,
+                bootstrap_vnic: None, // TODO: Change this?
```
smklein, line 304: I recommend looking at `get_control_interface` in `sled-agent/src/illumos/zone.rs` for how we're re-locating the `control_vnic` in this case. Basically, we ask the running zone for VNICs, and filter by prefix..

andrewjstone: Thanks! Fixed in https://github.com/oxidecomputer/omicron/pull/2320/commits/64b600f45337d7c58d54a5c5a25411aa60658dde

sled-agent/src/bootstrap/hardware.rs
```diff
@@ -124,19 +125,24 @@ impl HardwareMonitor {
...
+        underlay_etherstub: Etherstub,
+        underlay_etherstub_vnic: EtherstubVnic,
+        bootstrap_etherstub: Etherstub,
+        switch_zone_bootstrap_address: Ipv6Addr,
     ) -> Result<Self, Error> {
         let hardware =
             HardwareManager::new(log.clone(), sled_config.stub_scrimlet)
                 .map_err(|e| Error::Hardware(e))?;
 
+        // TODO: Add the etherstub for the bootstrap network here
```
smklein, line 137: Aren't we already doing that? Is this out-of-date?

smklein (approved):

PR #2338, opened 2023-02-08
[wicket] Enable arrow navigation for rack

### LINKED ISSUE #2337: wicket: Use arrow keys to navigate the rack screen?
Small feature request: Could we use up/down/left/right to navigate between slots on the rack screen in addition to the current tab/shift-tab/mouse navigation?

andrewjstone:
Fixes #2337


jgallagher (approved):
Thanks, works perfectly!

PR #2340, opened 2023-02-09
[sled-agent] Re-install Omicron Zones, rather than assuming they work

smklein:
This forces deletion and re-installation of "installed-but-not-running" zones which are discovered by the Sled Agent.

This is admittedly a bit of a heavy hammer - we could validate the installation looks like what we'd hope - but it quickly mitigates some issues seen, where a zone failing to install (and retrying later) does not succeed, because the configuration gets out-of-date.

### DIFF
--- a/sled-agent/src/illumos/zone.rs
+++ b/sled-agent/src/illumos/zone.rs
@@ -204,6 +204,11 @@ impl Zones {
         Ok(())
     }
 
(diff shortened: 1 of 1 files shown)


andrewjstone (approved):
As far as I can tell from my work with zones in the omicron code, this seems completely reasonable.

PR #2345, opened 2023-02-09
[sled-agent] Responsibility for cleanup-on-boot moved to bootstrap agent

smklein:
- Previously, the Sled Agent managed zones exclusively. On boot, it took responsibility for a few things:
  - Ensuring that a filesystem for zones exists
  - Deleting all "old" zones
  - Deleting all "old" vnics
  - Deleting all "old" XDE devices
- With the addition of the switch zone, some of the responsibility for launching zones has shifted into the bootstrap agent
  - To avoid race conditions, this "clean slate" behavior now occurs *before* any switch management


andrewjstone (approved):
LGTM.

andrewjstone (approved):

PR #2346, opened 2023-02-10
Merge `version` and `version_sort`

### LINKED ISSUE #2344: Merge `version` and `version_sort` columns
Semver version strings do not sort well lexicographically. In #2100 I hacked in DB sorting of system updates by version by adding a second column to sort by, `version_sort`:

https://github.com/oxidecomputer/omicron/blob/46f2295acd34394660f2576c4161fe47d60d3798/common/src/sql/dbinit.sql#L1575-L1578

https://github.com/oxidecomputer/omicron/blob/46f2295acd34394660f2576c4161fe47d60d3798/common/src/sql/dbinit.sql#L1667-L1672

which uses left zero padding to make the versions sort _almost_ correctly. (Technically, build and prerelease meta are not supposed to be sorted lexicographically, but we have decided to tolerate being wrong in this way for now.)

https://github.com/oxidecomputer/omicron/blob/46f2295acd34394660f2576c4161fe47d60d3798/nexus/db-model/src/semver_version.rs#L59-L73

However! These do not actually need to be two columns. I only did it that way because I didn't feel like figuring out how to deserialize the zero-padded version string back into a `Semver::Version` (built-in `parse` chokes on the zeros). The change that's required here is to update our `ToSql` for `SemverVersion` to do the zero-padding and the `FromSql` to remove the zero-padding before passing it to `Semver::Version`'s `parse`.

https://github.com/oxidecomputer/omicron/blob/46f2295acd34394660f2576c4161fe47d60d3798/nexus/db-model/src/semver_version.rs#L86-L110

That way, we do not have to generate a sortable string manually in the constructor.

https://github.com/oxidecomputer/omicron/blob/46f2295acd34394660f2576c4161fe47d60d3798/nexus/db-model/src/system_update.rs#L42-L54

david-crespo:
Closes #2344. Followup to #2100. Rather than manually adding a sortable `version_sort` column alongside the `version` column, make `SemverVersion` itself automatically sortable by serializing it to the DB in sortable form (with zero padding on the numbers). On deserialization, we remove the padding before trying to parse it.

### DIFF
--- a/common/src/sql/dbinit.sql
+++ b/common/src/sql/dbinit.sql
@@ -1573,18 +1573,13 @@ CREATE TABLE omicron.public.system_update (
     -- this resource different from every other resource for little benefit.
 
     -- Unique semver version
-    version STRING(64) NOT NULL, -- TODO: length
-    -- version string with maj/min/patch 0-padded to be string sortable
-    version_sort STRING(64) NOT NULL -- TODO: length
+    version STRING(64) NOT NULL -- TODO: length
 );
 
 CREATE UNIQUE INDEX ON omicron.public.system_update (
     version
 );
 
-CREATE UNIQUE INDEX ON omicron.public.system_update (
-    version_sort
-);
  
 CREATE TYPE omicron.public.updateable_component_type AS ENUM (
     'bootloader_for_rot',
@@ -1668,8 +1663,6 @@ CREATE TABLE omicron.public.updateable_component (
     -- This may need to be nullable if we are registering components before we
     -- know about system versions at all
     system_version STRING(64) NOT NULL, -- TODO: length
-    -- version string with maj/min/patch 0-padded to be string sortable
-    system_version_sort STRING(64) NOT NULL, -- TODO: length
 
     status omicron.public.update_status NOT NULL
     -- TODO: status reason for updateable_component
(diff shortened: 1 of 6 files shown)


zephraph:
Cool, yeah. I much prefer only having one column representing the version.

common/src/sql/dbinit.sql
```diff
@@ -1573,18 +1573,13 @@ CREATE TABLE omicron.public.system_update (
     -- this resource different from every other resource for little benefit.
 
     -- Unique semver version
-    version STRING(64) NOT NULL, -- TODO: length
-    -- version string with maj/min/patch 0-padded to be string sortable
-    version_sort STRING(64) NOT NULL -- TODO: length
+    version STRING(64) NOT NULL -- TODO: length
```
zephraph, line 1576: One random thing here... I noticed a lot of other strings are defined as 63 characters in the DB and I'm not exactly sure why that is.

zephraph: It seems like CRDB itself imposes that length limit on some of its internal strings: https://www.cockroachlabs.com/docs/stable/schema-design-overview.html#hard-limits

david-crespo: Oh yeah, there's a comment at the top but it is ambivalent about 63 or 64.

https://github.com/oxidecomputer/omicron/blob/302a7307239827e5ac1ca37efafc0649d8f9f003/common/src/sql/dbinit.sql#L11-L16

zephraph (approved):

PR #2348, opened 2023-02-10
Work around flaky apt repo in CI

iliana:
[Something is wrong with the packages.microsoft.com repository](https://github.com/microsoft/linux-package-repositories/issues/34) and it is causing CI flakes whenever they update the repository. We don't need it so we'll remove it.

https://github.com/iliana/there-are-too-many-things-in-the-computer-box/actions/runs/4147694155/jobs/7174875470 is a briefer demonstration of this being run.


andrewjstone (approved):
Thanks @iliana

PR #2358, opened 2023-02-13
[nexus] Populate IP pool, nexus service information, during rack setup

smklein:
My long-term goal is to have Nexus be in charge of provisioning all services.

For that to be possible, Nexus must be able to internalize all input during the handoff from RSS. This PR extends the RSS -> Nexus handoff to include:

- What "Nexus Services" are being launched?
- What external IP addresses are used for those services?

Fixes https://github.com/oxidecomputer/omicron/issues/1958


nexus/src/db/queries/external_ip.rs
```diff
@@ -392,23 +393,34 @@ impl NextExternalIp {
...
-    //     first_address +
-    //         generate_series(0, last_address - first_address)
-    //         AS candidate_ip
-    // FROM
-    //     ip_pool_range
-    // WHERE
-    //     <pool_restriction> AND
-    //     time_deleted IS NULL
+    // SELECT * FROM (
+    //   SELECT
```
smklein, line 397: If you're curious about how this works, there's an example fiddle which shows the logic:

https://www.db-fiddle.com/f/4jyoMCicNSZpjMt4jFYoz5/7203

nexus/src/db/queries/external_ip.rs
```diff
@@ -1023,7 +1052,103 @@ mod tests {
...
+                &context.opctx,
+                id,
+                IpAddr::V4(Ipv4Addr::new(10, 0, 0, 2)),
+            )
+            .await
+            .expect("Failed to allocate service IP address");
+
+        // We return success, but with the address requested the first time.
+        assert_eq!(ip.id, ip_again.id);
+        assert_eq!(ip.ip.ip(), ip_again.ip.ip());
```
bnaecker, line 1113: This test was pretty surprising to me. The queries on line 1071 and 1087 ask for the address `10.0.0.3` twice, and it looks like we're testing that we get back the exact same record we've already inserted. That checks the query is idempotent, and looks reasonable.

The query on 1101 explicitly asks for `10.0.0.2`. I see that we're getting back the same record, but only _because we asked for the same ID_. I would have expected the query to fail, since we're asking for an existing record by ID (presumably using an ID allocated during a saga), but get back a record with a different IP than we asked for. Under what circumstances would we want that? It seems like we'd want to fail the query, because some higher-level software repeated a query, but with a different address. That seems indicative of a bug, but I might be missing something.

I guess it's possible for the software making this query to check that we got the address we actually asked for, but that also seems weird. If there's an explicit address being requested, I'd expect to either get back that address or to fail, not to succeed but with a different address.

nexus/types/src/internal_api/params.rs
```diff
@@ -215,15 +217,13 @@ impl std::fmt::Debug for Certificate {
...
     pub datasets: Vec<DatasetCreateRequest>,
-    // TODO(https://github.com/oxidecomputer/omicron/issues/1530):
-    // While it's true that Nexus will only run with a single address,
-    // we want to convey information about the available pool of addresses
-    // when handing off from RSS -> Nexus.
-
-    // TODO(https://github.com/oxidecomputer/omicron/issues/1528):
-    // Support passing x509 cert info.
+    /// IP pools which may be used for internal services, such as Nexus.
+    pub internal_services_ip_pool: Vec<IpRange>,
```
bnaecker, line 225: This is a single pool, right? If so, the comment is a bit confusing. I might say, "IP ranges that form the pool used for internal services, such as Nexus", or something to that effect. Or might this need to be multiple pools?

sled-agent/src/rack_setup/service.rs
```diff
@@ -480,6 +481,25 @@ impl ServiceInner {
                 for svc in &zone.services {
                     let kind = match svc {
                         ServiceType::Nexus { external_ip, internal_ip: _ } => {
+                            // NOTE: Eventually, this IP pool will be entirely
+                            // user-supplied. For now, however, it's inferred
+                            // based on the input IP addresses.
```
bnaecker, line 486: Just checking -- `external_services_ip_pool` is a `Vec` with exactly one element now, because it's just made from the IP address we provide for Nexus. But it's a `Vec` so that the operator can provide multiple ranges in future, yes?

bnaecker (commented):
Nice work, lots of effort here! I have some questions about the details of the updated query, and a few nits, but overall looks excellent.

nexus/src/db/queries/external_ip.rs
andrewjstone: I agree with @bnaecker that this interface does not follow the principle of least surprise. Semantically, I'd expect failure or return of the address asked for instead.

andrewjstone (commented):
This generally looks good to me. @bnaecker  already did a thorough review and I'll let him take the lead here.

nexus/src/db/queries/external_ip.rs
smklein: So I admittedly struggled a bit with this query, because:

- On the one hand, I really want to have a "unified allocation flow" for IP allocation. I don't want to, for example, forget to update the collection rcgen for service IPs vs instance IPs...
- ... but on the other hand, yeah, scanning through addresses when we know we want to allocate *one* is kinda non-optimal.

I went ahead and used this optimized form in 1637370b20978dab16b5bc0f6efa08d3b6c9a0a4 , which more-or-less matches the fiddle you provided.

nexus/src/db/queries/external_ip.rs
```diff
@@ -392,23 +393,34 @@ impl NextExternalIp {
     // each IP Pool, along with the pool/range IDs.
```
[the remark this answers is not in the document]
smklein: Updated in bc87a4702b5135f48ff28f850186f8980fd5f1cb

nexus/src/db/queries/external_ip.rs
```diff
@@ -976,7 +1005,7 @@ mod tests {
             Ipv4Addr::new(10, 0, 0, 2),
         ))
         .unwrap();
-        context.create_service_ip_pool("for-nexus", ip_range).await;
+        context.initialize_ip_pool("oxide-service-pool", ip_range).await;
```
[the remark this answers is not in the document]
smklein: Done in 64b0969f0506c0175203d5faec9f8a449d2e66da

nexus/src/db/queries/external_ip.rs
bnaecker: Nice, that looks like both a lot of work, and be very helpful! Thanks for taking that on.

bnaecker:
Thanks for all the rework @smklein, this all looks good to me! Feel free to merge at your discretion.

PR #2363, opened 2023-02-15
[wicket] Add List widget

andrewjstone:
This widget will be used to scroll through rack components during update.

### DIFF
--- a/wicket/src/widgets/list.rs
+++ b/wicket/src/widgets/list.rs
@@ -1,0 +1,405 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! List widget, specifically designed for use with wicket
+
+use crate::Point;
+
+use super::{get_control_id, Control, ControlId, HoverResult};
+use std::fmt::Display;
+use tui::layout::Rect;
+use tui::style::Style;
+use tui::text::{Span, Spans, Text};
+use tui::widgets::{Block, BorderType, Borders, Paragraph, Widget};
+
+/// An indicator is a character that represents some facet of an item in
+/// a list.
+///
+/// For example, a list of components being updated may have a symbol which is
+/// styled grey for no update, yellow for partial update, green for updated.
+#[derive(Debug)]
+pub struct Indicator {
+    pub style: Style,
+    pub symbol: &'static str,
+}
+
+/// TODO: Paramaterize this with an item that is `Display`, like `ComponentId`?
+/// It's a tradeoff between time to stringify during rendering, vs parse during input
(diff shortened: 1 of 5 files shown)


wicket/src/widgets/list.rs
```diff
@@ -0,0 +1,405 @@
...
+///
+/// For example, a list of components being updated may have a symbol which is
+/// styled grey for no update, yellow for partial update, green for updated.
+#[derive(Debug)]
+pub struct Indicator {
+    pub style: Style,
+    pub symbol: &'static str,
+}
+
+/// TODO: Paramaterize this with an item that is `Display`, like `ComponentId`?
```
andrewjstone, line 27: Remove this outdated comment.

wicket/src/widgets/list.rs
```diff
@@ -0,0 +1,405 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! List widget, specifically designed for use with wicket
```
sunshowers, line 5: Out of curiosity, how is this different from tui's [`List` widget](https://docs.rs/tui/latest/tui/widgets/struct.List.html)?

sunshowers (commented):
Looks great, thanks!

sunshowers (approved):

wicket/src/widgets/list.rs
andrewjstone: Great question! I started to add this as a comment in the code, but removed it after some time as distracting. 

The most basic difference is that the TUI List widget, like all TUI stateful widgets, mutates its state during rendering. Wicket doesn't allow that (although it could easily enough though it would change the model), because it separates all actions on state into the mutable event handlers, and rendering just uses immutable state. The [Screen](https://github.com/oxidecomputer/omicron/blob/main/wicket/src/screens/mod.rs#L56-L73) trait in wicket defines this. The reason TUI does this is because it has no defined user interaction model and must mutate state during rendering where it knows the current position (`Rect`) to operate on. We don't have to do force this, because our code handles resize events and can mutate state on those events and calculate position before we ever render. We don't even use the `Rect` parameter when rendering our widgets. 

This seems like a relatively trivial difference, except that we also define the state of our widgets to be [Controls](https://github.com/oxidecomputer/omicron/blob/main/wicket/src/widgets/mod.rs#L32-L61), which provide a user interaction model on top of the pure rendering functionality of TUI widgets. This allows us to build abstractions like TabIndex, mouse hover/click, and animation support across **all** our stateful widgets that isn't possible in `tui` because each stateful widget has it's own concrete associated type - [ListState](https://docs.rs/tui/latest/tui/widgets/struct.ListState.html) for `List` which doesn't allow more sophisticated styling  like hover styles.

PR #2369, opened 2023-02-15
[common] make ArtifactKind a newtype wrapper, rename UpdateArtifactKind to KnownArtifactKind

sunshowers:
Making ArtifactKind a newtype wrapper allows equality checks to work properly no matter how it's constructed.

Also use a clearer `KnownArtifactKind` name.


common/src/update.rs
```diff
@@ -30,56 +32,72 @@ pub struct Artifact {
 /// describe artifact kinds it is not yet aware of, this has a fallback
 /// `Unknown` variant.
 #[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize)]
-#[serde(untagged)]
-pub enum ArtifactKind {
-    Known(UpdateArtifactKind),
-    Unknown(String),
-}
+pub struct ArtifactKind(Cow<'static, str>);
```
iliana, line 35: I think `Cow<'static, str>` is fine but I'm curious what the use case is over a regular owned `String`, since `ArtifactKind::from_known` always results in an owned `String`. I didn't see any uses of the `Cow::Borrowed` form outside of tests on first glance.

(Which makes me wish the parse-display crate had an optional `to_static_str` method that did the right thing for enums with only unit variants.)

iliana (approved):

common/src/update.rs
sunshowers: Good question -- in this case there's no advantage I think, but in the general case you might use this to construct strings as constants:

```rust
const CONTROL_PLANE: ArtifactKind = ArtifactKind::from_static("control_plane");
```

If the known artifact kinds were declared as constants rather than an enum, and/or if parse-display had `to_static_str`, this would make more sense. Currently there's no advantage over a string.

common/src/update.rs
```diff
@@ -30,56 +32,72 @@ pub struct Artifact {
 /// describe artifact kinds it is not yet aware of, this has a fallback
 /// `Unknown` variant.
```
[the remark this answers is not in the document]
sunshowers: Done, thanks for catching it!

common/src/update.rs
sunshowers: Changed to a string.

PR #2370, opened 2023-02-15
[ipcc-key-value] Add an `update_id` to InstallinatorImageId

jgallagher:
The primary motivation for this PR is to add an `update_id: Uuid` to the data passed down to installiator via MGS and the SP, which will allow us to track progress / completion over the bootstrap network.

While making this change, I realized `Uuid`s were be serialized by ciborium more efficiently than our `[u8; 32]` arrays, so this PR also optimizes the serialization of the hash fields: they now always serialize as exactly 34 bytes, instead of varying between 34 and 66.

Specifically, serde has an optimized path for [serializing byte sequences](https://docs.rs/serde/latest/serde/trait.Serializer.html#tymethod.serialize_bytes), which ciborium uses (encoding them as a `bytes` CBOR object). However, serde's derived `Serialize` implementation for `[T; N]` cannot be specialized for `[u8; N]`, so it serializes them as arbitrary sequences (which ciborium then encodes as an `array` CBOR object, within which bytes may take 1 or 2 bytes on the wire, depending on the value). [serde_bytes](https://docs.rs/serde_bytes/latest/serde_bytes/) exists to give an easy way to opt into optimizing byte sequences, but it's only defined for byte slices and `Vec<u8>`, not `[u8; N]`, so I adapted [this workaround](https://github.com/serde-rs/bytes/issues/26#issuecomment-902550669), which didn't quite work with ciborium. The overall effect is most visible in the tests: the explicit example is obviously shorter, and the test that previously checked for a `MAX_SIZE` (because the actual on-the-wire size was variable) now checks for an exact `EXPECTED_SIZE` (by using `bytes` instead of `array` we now always serialize to exactly the size we expect).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2872,6 +2872,7 @@ dependencies = [
  "proptest",
  "serde",
  "test-strategy",
+ "uuid",
 ]
 
 [[package]]
--- a/gateway-cli/src/main.rs
+++ b/gateway-cli/src/main.rs
@@ -139,6 +139,13 @@ enum Command {
         /// Clear any previously-set ID.
         #[clap(long)]
         clear: bool,
+        /// ID to assign to this update for progress/completion messages.
+        #[clap(
+            long,
+            conflicts_with = "clear",
+            required_unless_present = "clear"
+        )]
+        update_id: Option<Uuid>,
         /// Hash of the host OS image.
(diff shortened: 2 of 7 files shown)


sunshowers (approved):
neat, looks great!

PR #2371, opened 2023-02-15
use the tofino crate to probe for tofino devices

Nieuwejaar:
The `tofino` create provides the functionality to probe the device tree for tofino ASICs and drivers.  Use that crate rather than doing the search internally.

When configuring a switch zone, set an `smf` property that tells the `dpd` daemon which tofino device it is expected to use.


smklein (approved):
Looks great; I'm happy to have these details handled by an auxiliary crate.

PR #2372, opened 2023-02-16
[wicket] Start of an update screen

andrewjstone:
Currently the update screen uses fake data for per component updates and doesn't include a button to actually perform an update. This intermediate commit is used to prevent a very large commit coming soon.

Also included here is navigation between screens using `ctrl-n` and `ctrl-p`, and use of `CommonScreenState` in the Rack screen. Admittedly some of the common usage could use some cleanup around the abstractions, but I don't think it's critical right now. I'd rather move forward with actual update functionality.


wicket/src/screens/update.rs
```diff
@@ -0,0 +1,192 @@
...
+        }
+    }
+
+    fn handle_key_event(
+        &mut self,
+        state: &mut State,
+        event: KeyEvent,
+    ) -> Vec<Action> {
+        match event.code {
+            KeyCode::Tab => {
```
andrewjstone, line 54: Currently, `rack_state` contains a `TabIndex` to cycle through components. We maintain this state in both the `Rack` and `Components` screen, meaning when you switch between them the same component is selected. This isn't currently maintained across the update screen, and I'm not sure it makes sense there. Currently there is actually no mapping from the list to the rack state `TabIndex`. We should add that support if it makes sense, or remove the tab controls here. We can also allow tab controls just for scroll up/scroll down, which we probably should do for symmetry.

wicket/src/screens/common.rs
```diff
@@ -134,19 +139,53 @@ impl CommonScreenState {
...
-                style::button()
+    pub fn draw_screen_navigation_instructions(&self, f: &mut Frame) {
+        let prev = self.prev_screen_label();
+        let next = self.next_screen_label();
+        if let Some(p) = prev {
+            let width = p.width() as u16;
+            let prev =
+                Paragraph::new(p).style(Style::default().fg(OX_OFF_WHITE));
+            let rect = Rect {
+                x: f.size().width - width - next.width() as u16 - 3,
```
jgallagher, line 150: Constants like ` - 3` make me a little uneasy, but I'm not going to quibble too much. Does this have a meaningful name, or is it relative to something else?

wicket/src/screens/update.rs
jgallagher: Couple comments:

* The tab/backtab on the update screen _appear_ to not do anything at all, until you go back to the previous screen. I'd vote for changing tab controls to scroll up/down on this screen and not affect the previous screen.
* If I switch between the inventory and update screens, each remembers the last sled selected but they're not related to each other; it feels weird. Maybe updates shouldn't be an "after" screen from inventory, but both are reachable from the rack screen? Connecting the selected sled in the update screen with the selected sled on the rack screen seems more reasonable if there's not the inventory screen in between. (This feels like a bigger change - even if you like it I'd defer it and just leave the two screens selections' unconnected for now.)

jgallagher (approved):
Played with it, and it looks great. :+1:

wicket/src/widgets/update.rs
```diff
@@ -0,0 +1,186 @@
...
+                .map(|i| {
+                    let item = if i < 32 {
+                        ComponentId::Sled(i)
+                    } else if i < 34 {
+                        ComponentId::Switch(i - 32)
+                    } else {
+                        ComponentId::Psc(i - 34)
+                    };
+
+                    // TODO: Fix this hardcoding nonsense
```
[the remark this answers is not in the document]
andrewjstone: Yep. I even had a comment about filling in real data that I removed by mistake.

wicket/src/screens/common.rs
andrewjstone: Yeah, it's an offset for spacing, I'll name it. Good call!

wicket/src/screens/update.rs
andrewjstone: Yeah, I agree this isn't right as is. I'm going to discuss this with @benjaminleonard. As mentioned offline, I think it may be better to just move the inventory into the update screen as a separate tab with the list view for scrolling between both update history and inventory.

wicket/src/widgets/update.rs
```diff
@@ -0,0 +1,186 @@
...
+        self.status_rect = mid_h_chunks[1];
+        self.status_rect.width -= 14;
+        self.status_rect.x += 14;
+
+        self.components.resize(list_rect);
+    }
+
+    fn fill_list(&mut self, _: &Inventory) {
+        self.components.init(
+            (0u8..36u8)
```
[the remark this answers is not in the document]
andrewjstone: Both! I should just put a bottom margin and allow the list to be as tall as possible for the largest informational view perhaps. Although, whitespace at the bottom of the list looks weird and I was thinking of cropping it out. On the other hand on small laptop monitors, it' possible we won't have room for the full list. I doubt it, but the code should deal with it. I'll discuss with @benjaminleonard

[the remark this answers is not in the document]
andrewjstone: I'm going to leave it as in this commit in the interest of integrating with wicketd. But I'll come back to this.

wicket/src/screens/update.rs
andrewjstone: I had a good chat with @benjaminleonard. We're likely going restructure the pages altogether. For know I'm going to make tabbing go up/down the list.

andrewjstone: Fixed.

PR #2373, opened 2023-02-16
CI: generate a TUF repo with the control plane assets

iliana:
or: MORE JOBS FOR THE BUILDOMAT GODS

This generates a partial TUF repo (multipart for now since Buildomat has a 1 GiB file size limit) using `tufaceous assemble`. It currently contains the control plane bundle and host OS images; I'm going to go looking for how I can find the other artifacts to include, but I think this is ready enough for review.

This also renames the "build-end-to-end-tests" job to "CI tools", and adds a build for tufaceous there. This is a "how long does it take CI to run" optimization, because this allows us to build the tufaceous binary while we're waiting for the rest of the control plane to build.

### DIFF
--- a/.github/buildomat/jobs/tuf-repo.sh
+++ b/.github/buildomat/jobs/tuf-repo.sh
@@ -29,6 +29,12 @@ rustc --version
 COMMIT=$(git rev-parse HEAD)
 VERSION="1.0.0-alpha+git${COMMIT:0:11}"
 
+#
+# The package job builds two switch zones. The one we need should be named "switch.tar.gz" so sled-agent can find it.
+#
+mv /input/package/work/zones/{switch-asic,switch}.tar.gz
+rm /input/package/work/zones/switch-softnpu.tar.gz
+
 cargo build --locked --release --bin tufaceous
 
 # Generate a throwaway repository key.
--- a/.github/buildomat/jobs/tuf-repo.sh
+++ b/.github/buildomat/jobs/tuf-repo.sh
@@ -5,7 +5,7 @@
 #: target = "helios-latest"
 #: output_rules = [
 #:	"=/work/manifest.toml",
-#:	"=/work/repo.zip",
+#:	"=/work/repo.zip.part*",
 #: ]
 #:
 #: [dependencies.ci-tools]
@@ -83,3 +83,9 @@ EOF
 done
 
 /work/tufaceous assemble --no-generate-key --skip-all-present /work/manifest.toml /work/repo.zip
+
+#
+# XXX: Buildomat currently does not support uploads greater than 1 GiB. This is
+# an awful temporary hack which we need to strip out the moment it does.
+#
+split -a 1 -b 1024m /work/repo.zip /work/repo.zip.part
--- a/.github/buildomat/jobs/tuf-repo.sh
+++ b/.github/buildomat/jobs/tuf-repo.sh
@@ -31,6 +31,24 @@ VERSION="1.0.0-alpha+git${COMMIT:0:11}"
 ptime -m gunzip < /input/ci-tools/work/tufaceous.gz > /work/tufaceous
 chmod a+x /work/tufaceous
 
+#
+# We do two things here:
+# 1. Run `omicron-package stamp` on all the zones.
+# 2. Run `omicron-package unpack` to switch from "package-name.tar.gz" to "service_name.tar.gz".
+#
+mkdir /work/package
+pushd /work/package
+tar xf /input/package/work/package.tar.gz out package-manifest.toml target/release/omicron-package
+target/release/omicron-package -t default target create -i standard -m gimlet -s asic
+ln -s /input/package/work/zones/* out/
+rm out/switch-softnpu.tar.gz  # not used when target switch=asic
+for zone in out/*.tar.gz; do
+    target/release/omicron-package stamp "$(basename "${zone%.tar.gz}")" "$VERSION"
+done
+mv out/versioned/* out/
+OMICRON_NO_UNINSTALL=1 target/release/omicron-package unpack --out install
+popd
+
 # Generate a throwaway repository key.
 python3 -c 'import secrets; open("/work/key.txt", "w").write("ed25519:%s\n" % secrets.token_hex(32))'
 read -r TUFACEOUS_KEY </work/key.txt
@@ -46,23 +64,13 @@ version = "$VERSION"
 kind = "composite-control-plane"
 EOF
 
-# switch-asic needs to be named "switch", skip here.
-# switch-softnpu does not go in TUF repos.
-for zone in $(ls /input/package/work/zones/* | grep -v switch); do
+for zone in /work/package/install/*.tar.gz; do
     cat >>/work/manifest.toml <<EOF
 [[artifact.control_plane.source.zones]]
 kind = "file"
 path = "$zone"
 EOF
 done
-
-# Add switch-asic, but call it switch so sled-agent can find it.
-cp /input/package/work/zones/switch-asic.tar.gz /work/switch.tar.gz
-cat >>/work/manifest.toml <<EOF
-[[artifact.control_plane.source.zones]]
-kind = "file"
-path = "/work/switch.tar.gz"
-EOF
 
 for kind in host trampoline; do
     mkdir -p /work/os/$kind
--- a/.github/buildomat/jobs/tuf-repo.sh
+++ b/.github/buildomat/jobs/tuf-repo.sh
@@ -1,0 +1,51 @@
+#!/bin/bash
+#:
+#: name = "helios / tuf-repo"
+#: variety = "basic"
+#: target = "helios-latest"
+#: rust_toolchain = "1.68.2"
+#: output_rules = [
+#:	"=/work/manifest.toml",
+#:	"=/work/repo.zip",
+#: ]
+#:
+#: [dependencies.package]
+#: job = "helios / package"
+#:
+
+set -o errexit
+set -o pipefail
+set -o xtrace
+
+cargo --version
+rustc --version
+
+COMMIT=$(git rev-parse HEAD)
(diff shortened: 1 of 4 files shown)


.github/buildomat/jobs/tuf-artifact.sh
```diff
@@ -0,0 +1,37 @@
...
+
+set -o errexit
+set -o pipefail
+set -o xtrace
+
+cargo --version
+rustc --version
+
+COMMIT=$(git rev-parse HEAD)
+VERSION="0.0.0+git${COMMIT:0:11}"
```
smklein, line 23: (not... necessarily a recommendation, but interested in your thoughts)

WDYT about using git tags to identify the version?

I *think* `git tag --contains HEAD` would return the tag, if one exists, and we could use that as the semver version?

I think your decision to use "0.0.0+git<commit>" is a *great* idea for development, but obviously, when we actually want to get around to publishing `0.0.1`, we need to store that semver value *somewhere*, for both individual components and for the system version.

.github/buildomat/jobs/tuf-artifact.sh
```diff
@@ -0,0 +1,37 @@
...
+
+pushd /input/package/work/out
+tar cvf "/work/control-plane.tar" ./*
+popd
+
+cargo build --locked --release --bin tufaceous
+
+python3 -c 'import secrets; open("/work/key.txt", "w").write("ed25519:%s\n" % secrets.token_hex(32))'
```
smklein, line 31: What is this key?

.github/buildomat/jobs/tuf-artifact.sh
```diff
@@ -0,0 +1,37 @@
...
+read -r TUFACEOUS_KEY </work/key.txt
+export TUFACEOUS_KEY
+
+target/release/tufaceous -r /work/repo init --no-generate-key
+target/release/tufaceous -r /work/repo add --name control_plane control_plane "/work/control-plane.tar" "$VERSION"
+target/release/tufaceous -r /work/repo archive /work/repo.zip
```
smklein, line 37: So I can see how we're labeling *components* with versions - how do you think this should map to a system-level version?

.github/buildomat/jobs/tuf-artifact.sh
iliana: It's a throwaway key. Perhaps I should make that clearer.

For the purposes of this repository, I don't think we have a good way of establishing trust other than "the repo was spit out of buildomat so it's probably related to this commit".

.github/buildomat/jobs/tuf-artifact.sh
smklein: (Not trying to place the pressure of "making the call about what the system version should be" here - if we use a schema like `0.0.0+git commit`, that totally works too, just wondering how we shove that all into a TUF repo that Nexus could theoretically query later)

.github/buildomat/jobs/tuf-artifact.sh
iliana: [`git describe`](https://git-scm.com/docs/git-describe) is probably the right play here then, because it gives either the tag of a commit, or a tag and the number of commits after that tag (and has some formatting play available).

smklein: Also, while I have the bad idea train rolling through here: 

Should we embed some type of timestamp in here too, so that development versions of software can be lexicographically sorted, and we can automatically prefer "more-recent" ones?

iliana: RFD some-number-i-can't-think-of declares artifact versions to be semver, I think. (I should probably enforce that in tufaceous.)

smklein: Right, I meant as part of the `+...` semvar label

smklein:
I know I jumped straight to comments, but wanted to call this out:

- This is a culmination of so much great work
- It's extremely rad
- I am so hyped for this to land

.github/buildomat/jobs/tuf-artifact.sh
iliana: I think actually this shouldn't map to a system-level version at all, at this stage. That seems like a release engineering process, to decide which artifacts (from all the repositories) become a specific ✨ System Version ✨.

(And that process can be automatic, and can be based on the commit hashes of the various components, so that we can have a nightly system version or the like. But the point is that the ✨ System Version ✨ is a combination of things that's outside the control of (probably) this repo.)

iliana:
`worker: upload error: file "/work/repo.zip" is 1277434891 bytes in size, which is larger than 1GiB and cannot be uploaded at this time.`
oh no

jclulow:
Ah.  Yes.  I guess it's time to fix that then.

.github/buildomat/jobs/tuf-repo.sh
```diff
@@ -0,0 +1,91 @@
...
+[artifact.control_plane]
+name = "control-plane"
+version = "$VERSION"
+[artifact.control_plane.source]
+kind = "composite-control-plane"
+EOF
+
+# switch-asic needs to be named "switch", skip here.
+# switch-softnpu does not go in TUF repos.
+for zone in $(ls /input/package/work/zones/* | grep -v switch); do
```
jgallagher, line 51: A couple things came up today doing this manually to try mupdating the dogfood rack:

1. Some of the zones produced by `omicron-package` are only intermediates that are themselves then included in other zones; e.g., [dendrite](https://github.com/oxidecomputer/omicron/blob/857a9b95a83e7c833822fc29e3b89c4d1be41386/package-manifest.toml#L305) is marked as `intermediate_only` and is then a part of [the switch zone](https://github.com/oxidecomputer/omicron/blob/857a9b95a83e7c833822fc29e3b89c4d1be41386/package-manifest.toml#L360). Adding the intermediates here is relatively harmless, I think, other than making the tuf repo larger? But it would be good to prune them if we could.
2. For zones with multiple words (e.g., `crucible-pantry`), we may have some mixed hyphens/underscores? E.g., on my build VM, I see 

```
/work/zones/crucible-pantry.tar.gz
/work/zones/external-dns.tar.gz
/work/zones/internal-dns.tar.gz
/work/zones/mg-ddm.tar.gz
/work/zones/omicron-nexus.tar.gz
/work/zones/oximeter-collector.tar.gz
/work/zones/propolis-server.tar.gz
```

but on a running system under `/opt/oxide` I see

```
/opt/oxide/clickhouse.tar.gz
/opt/oxide/crucible_pantry.tar.gz
/opt/oxide/external_dns.tar.gz
/opt/oxide/oximeter.tar.gz
/opt/oxide/propolis-server.tar.gz
/opt/oxide/cockroachdb.tar.gz
/opt/oxide/internal_dns.tar.gz
/opt/oxide/nexus.tar.gz
/opt/oxide/ntp.tar.gz
/opt/oxide/crucible.tar.gz
/opt/oxide/mg-ddm.tar.gz
```

where some but not all of the hyphens have been replaced by underscores. I'm not sure what's doing that replacement today, or if I'm just confused?

.github/buildomat/jobs/tuf-repo.sh
```diff
@@ -0,0 +1,91 @@
...
+[[artifact.control_plane.source.zones]]
+kind = "file"
+path = "/work/switch.tar.gz"
+EOF
+
+for kind in host trampoline; do
+    mkdir -p /work/os/$kind
+    pushd /work/os/$kind
+    # https://github.com/oxidecomputer/helios#os-image-archives
+    tar xf /input/$kind/work/helios/image/output/os.tar.gz image/rom image/zfs.img
```
jgallagher, line 71: Untarring these are not strictly necessary because the tarballs produced by helios already have the right structure. Instead of untarring and using `composite-host` below, I think the manifest section could be

```toml
[artifact.$kind]
name = "$kind"
version = "$VERSION"
source = { kind = "file", path = "/input/$kind/work/helios/image/output/os.tar.gz" }
```

Untarring just the specific files we need has the side effect of pruning out any extra stuff helios puts in the tarball that we don't care about today - I don't know whether we like that or not.

jgallagher (commented):
Thank you for doing this!

.github/buildomat/jobs/tuf-repo.sh
iliana: Oh interesting. I'm not sure whether we like it or not either; my concern would mainly be about size but it looks like the extra stuff adds like 9 MiB, which is not really that big.

.github/buildomat/jobs/tuf-repo.sh
iliana: that is... strange. let me stare at package-manifest.toml and some other parts of the code

iliana: OK, so dendrite wasn't ending up in the repo because the package job uses a [hardcoded list of zones](https://github.com/oxidecomputer/omicron/blob/15ae3e942e6eb90fbe6302e71f2367d542648ecb/.github/buildomat/jobs/package.sh#L72-L86).

The hyphens/underscores thing is because the zone names built in out/ and installed in /opt/oxide are inconsistent, e.g. the _package name_ is internal-dns but the _service name_ is internal_dns. Fortunately `omicron-package unpack` does all this logic for us and is trivial enough to get running in this job that we'll just use it.

.github/buildomat/jobs/tuf-repo.sh
iliana: Latest commit uses the tarballs directly, although they don't have any version info stamped in them.

jgallagher: I think the untarring is still (unnecessarily) here and could be removed.

jgallagher (approved):
Thank you so much for this!

.github/buildomat/jobs/tuf-repo.sh
iliana: oh yup

PR #1024, opened 2022-05-06
openapi linting

ahl:
This is still a WIP, but only insofar as it depends on some in-flight changes for openapi-lint, progenitor, and typify.

This makes omicron conform to the updated openapi-lint per https://github.com/oxidecomputer/openapi-lint/pull/7

Most significantly:
- enums all include `#[serde(rename_all = "snake_case")]` (unless they're all caps such as with `DnsRecord`)
- `IpNet` effectively becomes a `string` with various formatting constraints
- I've simplified a bunch of copy-pasta with `..Default::default()`
- Some small changes to oximeter types for consistency
- Bifurcation of oximeter types between persistent structures and API interfaces

PR #2375, opened 2023-02-16
[wicketd] Add endpoints to drive updates (SP only, for now)

jgallagher:
Currently this only supports updates to SPs. Adds endpoints to:

1. Get the list of available artifacts
2. Start updating a target SP
3. Get the progress of updates to a target SP
4. Get the progress of updates to all SPs (which have had at least one update started at some point in wicketd's lifetime)

It also adds a check to the repository upload endpoint that the uploaded repository contains all the artifacts we need to perform updates.

Builds on #2293 and should be merged after it.

### DIFF
--- a/gateway/examples/config.toml
+++ b/gateway/examples/config.toml
@@ -2,6 +2,12 @@
 # Oxide API: example configuration file
 #
 
+# Maximum number of host phase2 trampoline images we're willing to cache. Note
+# that this value is specified in terms of _number of images_, not bytes, and
+# our cache is in-memory. We expect this value to be small in production,
+# potentially even 1 (i.e., only keep the most-recently-uploaded image).
+host_phase2_recovery_image_cache_max_images = 1
+
 [dropshot]
 # We want to allow uploads of host phase 2 recovery images, which may be
 # measured in the (small) hundreds of MiB. Set this to 512 MiB.
--- a/gateway/src/http_entrypoints.rs
+++ b/gateway/src/http_entrypoints.rs
@@ -19,6 +19,7 @@ use dropshot::HttpError;
 use dropshot::HttpResponseOk;
 use dropshot::HttpResponseUpdatedNoContent;
 use dropshot::Path;
+use dropshot::Query;
 use dropshot::RequestContext;
 use dropshot::TypedBody;
 use dropshot::UntypedBody;
@@ -856,20 +857,17 @@ async fn sp_component_serial_console_detach(
     Ok(HttpResponseUpdatedNoContent {})
 }
 
-// TODO: how can we make this generic enough to support any update mechanism?
 #[derive(Deserialize, JsonSchema)]
-pub struct UpdateBody {
+pub struct ComponentUpdateIdSlot {
     /// An identifier for this update.
     ///
     /// This ID applies to this single instance of the API call; it is not an
     /// ID of `image` itself. Multiple API calls with the same `image` should
     /// use different IDs.
     pub id: Uuid,
-    /// The binary blob containing the update image (component-specific).
-    pub image: Vec<u8>,
     /// The update slot to apply this image to. Supply 0 if the component only
     /// has one update slot.
-    pub slot: u16,
+    pub firmware_slot: u16,
 }
 
 #[derive(Deserialize, JsonSchema)]
@@ -927,16 +925,20 @@ async fn sp_reset(
 async fn sp_component_update(
     rqctx: RequestContext<Arc<ServerContext>>,
     path: Path<PathSpComponent>,
-    body: TypedBody<UpdateBody>,
+    query_params: Query<ComponentUpdateIdSlot>,
+    body: UntypedBody,
 ) -> Result<HttpResponseUpdatedNoContent, HttpError> {
     let apictx = rqctx.context();
 
     let PathSpComponent { sp, component } = path.into_inner();
     let sp = apictx.mgmt_switch.sp(sp.into())?;
     let component = component_from_str(&component)?;
+    let ComponentUpdateIdSlot { id, firmware_slot } = query_params.into_inner();
 
-    let UpdateBody { id, image, slot } = body.into_inner();
-    sp.start_update(component, id, slot, image)
+    // TODO-performance: this makes a full copy of the uploaded data
+    let image = body.as_bytes().to_vec();
+
+    sp.start_update(component, id, firmware_slot, image)
         .await
         .map_err(SpCommsError::from)?;
 
--- a/wicketd/src/artifacts.rs
+++ b/wicketd/src/artifacts.rs
@@ -157,7 +157,7 @@ fn extract_and_validate(
     // if time were available, we might want to be able to load older
     // versions of artifacts over the technician port in an emergency.
     //
-    // XXX we aren't checking against a root of trust at this point --
+    // XXX we aren't checking against a trust anchor at this point --
     // anyone can sign the repositories and this code will accept that.
     let repository = OmicronRepo::load_ignore_expiration(temp_path)
         .map_err(RepositoryError::LoadRepository)?;
@@ -282,8 +282,37 @@ enum RepositoryError {
 
 impl RepositoryError {
     fn to_http_error(&self) -> HttpError {
-        // TODO: add better errors than just 503
-        HttpError::for_unavail(None, DisplayErrorChain::new(self).to_string())
+        let message = DisplayErrorChain::new(self).to_string();
+
+        match self {
+            // Errors we had that are unrelated to the contents of a repository
+            // uploaded by a client.
+            RepositoryError::TempDirCreate(_) => {
+                HttpError::for_unavail(None, message)
+            }
+
+            // Errors that are definitely caused by bad repository contents.
+            RepositoryError::DuplicateEntry(_)
+            | RepositoryError::DuplicateArtifactKind(_)
+            | RepositoryError::LocateTarget { .. }
+            | RepositoryError::MissingArtifactKind(_)
+            | RepositoryError::MissingTarget(_) => {
+                HttpError::for_bad_request(None, message)
+            }
+
+            // Gray area - these are _probably_ caused by bad repository
+            // contents, but there might be some cases (or cases-with-cases)
+            // where good contents still produce one of these errors. We'll opt
+            // for sending a 4xx bad request in hopes that it was our client's
+            // fault.
+            RepositoryError::OpenArchive(_)
+            | RepositoryError::Extract(_)
+            | RepositoryError::LoadRepository(_)
+            | RepositoryError::ReadArtifactsDocument(_)
+            | RepositoryError::ReadTarget { .. } => {
+                HttpError::for_bad_request(None, message)
+            }
+        }
     }
 }
 
--- a/wicketd/src/artifacts.rs
+++ b/wicketd/src/artifacts.rs
@@ -45,10 +45,26 @@ impl WicketdArtifactStore {
     pub(crate) fn put_repository(&self, bytes: &[u8]) -> Result<(), HttpError> {
         slog::debug!(self.log, "adding repository"; "size" => bytes.len());
 
-        let new_artifacts = extract_and_validate(&bytes, &self.log)
+        let new_artifacts = extract_and_validate(bytes, &self.log)
             .map_err(|error| error.to_http_error())?;
         self.replace(new_artifacts);
         Ok(())
+    }
+
+    pub(crate) fn artifact_ids(&self) -> Vec<ArtifactId> {
+        self.artifacts.lock().unwrap().keys().cloned().collect()
+    }
+
+    /// Take a snapshot of the contents of the artifact store.
+    ///
+    /// This allows the update planner to grab the current repository (along
+    /// with all its data) to hand off to a potentially long-running update,
+    /// avoiding (some) issues with the artifacts being replaced while the
+    /// update is running.
+    pub(crate) fn snapshot(&self) -> HashMap<ArtifactId, BufList> {
+        // We expect this hashmap to be relatively small (order ~10), and
+        // cloning both ArtifactIds and BufLists are cheap.
+        self.artifacts.lock().unwrap().0.clone()
     }
 
     // ---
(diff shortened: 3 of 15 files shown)


wicketd/src/update_planner.rs
```diff
@@ -0,0 +1,489 @@
...
+            current: log.current.map(Into::into),
+            events: log.events.into_iter().map(Into::into).collect(),
+        }
+    }
+}
+
+#[derive(Debug)]
+pub(crate) struct UpdatePlanner {
+    mgs_client: gateway_client::Client,
+    artifact_store: WicketdArtifactStore,
```
sunshowers, line 87: Seems like the planner doesn't need to store a copy of the artifact store -- it can just receive it in `regenerate_plan`.

wicketd/src/update_planner.rs
```diff
@@ -0,0 +1,489 @@
...
+            // allow this one if that update is no longer running.
+            Entry::Occupied(mut slot) => {
+                if slot.get().is_finished() {
+                    slot.insert(spawn_update_driver().await);
+                    Ok(())
+                } else {
+                    Err(UpdatePlanError::UpdateInProgress(sp))
+                }
+            }
+        }
```
sunshowers, line 183: In what situations would we want an update to be started here after another one finishes?

wicketd/src/update_planner.rs
```diff
@@ -0,0 +1,489 @@
...
+pub(crate) struct UpdatePlanner {
+    mgs_client: gateway_client::Client,
+    artifact_store: WicketdArtifactStore,
+    update_plan: StdMutex<Result<UpdatePlan, UpdatePlanError>>,
+    running_updates: Mutex<BTreeMap<SpIdentifier, JoinHandle<()>>>,
+    // Note: Our inner mutex here is a standard mutex, not a tokio mutex. We
+    // generally hold it only log enough to update its state or push a new
+    // update event into its running log; occasionally we hold it long enough to
+    // clone it.
+    update_logs: Mutex<BTreeMap<SpIdentifier, Arc<StdMutex<UpdateLog>>>>,
```
sunshowers, line 94: So I'm trying to figure out if there's a reason `running_updates` and `update_logs` are stored separately. Could this just be something like:

```rust
struct UpdatePlanner {
    // ...
    sp_update_data: Mutex<BTreeMap<SpIdentifier, SpUpdateData>>,
}

struct SpUpdateData {
    running: Option<JoinHandle<()>>,
    logs: Arc<StdMutex<UpdateLog>>,
}
```

wicketd/src/update_planner.rs
```diff
@@ -0,0 +1,489 @@
...
+        self.set_current_update_state(UpdateStateKind::SendingArtifactToMgs {
+            artifact: artifact.id.clone(),
+        });
+        self.mgs_client
+            .sp_component_update(
+                self.sp.type_,
+                self.sp.slot,
+                SP_COMPONENT,
+                &body,
+            )
```
sunshowers, line 387: I did some digging and it looks like this will send the body as a `TypedBody<UpdateBody>`, which means that the update body will be a JSON-encoded array of bytes. This seems pretty wasteful? Maybe we could upload an untyped body here and pass in the id and slot as query parameters.

wicketd/src/update_planner.rs
```diff
@@ -0,0 +1,489 @@
...
+            tokio::time::sleep(STATUS_POLL_FREQ).await;
+        }
+    }
+}
+
+// Helper function to convert a `BufList` into a `Vec<u8>` for use with
+// `gateway_client::Client`.
+//
+// TODO-performance Can we pass `BufList`s directly (or wrapped) to
+// `gateway_client::Client` endpoints without copying the data?
```
sunshowers, line 482: Definitely should work if we turn the RPC call into an `UntypedBody` and using `wrap_stream`: https://docs.rs/reqwest/latest/reqwest/struct.Body.html#method.wrap_stream

sunshowers (commented):
Looks pretty cool -- have some questions and comments.

wicketd/src/mgs.rs
```diff
@@ -32,22 +32,23 @@ const CHANNEL_CAPACITY: usize = 8;
...
 pub struct MgsHandle {
     tx: tokio::sync::mpsc::Sender<MgsRequest>,
 }
 
-/// The response to a `get_inventory` call: the inventory of artifacts known to wicketd, or a
+/// The response to a `get_inventory` call: the inventory known to wicketd, or a
 /// notification that data is unavailable.
 #[derive(Clone, Debug, JsonSchema, Serialize)]
-#[serde(rename_all = "kebab-case", tag = "type", content = "data")]
+#[serde(rename_all = "snake_case", tag = "type", content = "data")]
```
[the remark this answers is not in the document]
jgallagher: Yeah, I don't remember (or maybe never knew) all the details, but I based this on seeing https://github.com/oxidecomputer/omicron/pull/1024 go by a while back.

wicketd/src/update_planner.rs
jgallagher: In practice, I think only if we uploaded a new TUF repository and wanted to MUPdate to an even newer version than we had previously MUPdated, but since wicketd runs "forever" that seems plausible (if not all that likely).

In development, it's handy to be able to start updates even if you've previously updated the target.

wicketd/src/http_entrypoints.rs
```diff
@@ -68,10 +81,108 @@ async fn put_repository(
...
+    match rqctx.context().update_planner.start(target.into_inner()).await {
+        Ok(()) => Ok(HttpResponseUpdatedNoContent {}),
+        Err(err) => match err {
+            UpdatePlanError::DuplicateArtifacts(_)
+            | UpdatePlanError::MissingArtifact(_) => {
+                // TODO-correctness for_bad_request may not be right - both of
+                // these errors are issues with the TUF repository, not this
+                // request itself.
+                Err(HttpError::for_bad_request(None, err.to_string()))
+            }
```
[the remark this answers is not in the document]
jgallagher: Hmm, so I think fixing this requires changing how the artifact store works. Right now, when we upload a repo, we:

1. Replace all the artifacts in the artifact store with the new repo's contents
2. Generate a plan, returning an error if it has any duplicate or missing artifacts

But if step 2 fails, step 1 means any requests to start an update will also fail for the same reason.

I can see at least a couple options:

1. If an upload fails due to duplicate/missing artifacts, we don't repeat that error here, we just report a "you haven't uploaded a valid repo" error.
2. In the artifact store, we don't replace artifacts until after we know if we can create a valid plan from the new repository.

I think 1 is fine? It's certainly easier to implement. It does mean if you upload a valid repo and then an invalid repo, you can't do updates, but that seems correct.

wicketd/src/update_planner.rs
jgallagher: Maybe! I landed where I did after a lot of churn, and your suggestion does look better. I'll give it a shot and see if I bump into anything unexpected.

wicketd/src/http_entrypoints.rs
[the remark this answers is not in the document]
sunshowers: Hmm, I'm not following what the difference is between 1 and 2. It seems like what we should do is:

1. Unpack the repo.
2. Generate a plan, which also makes sure that the repo is valid as the plan defines it.
3. Replace the artifacts and the plan with the new one.

This seems closer to option 2 than to 1, I think?

wicketd/src/update_planner.rs
sunshowers: Makes sense.

wicketd/src/artifacts.rs
```diff
@@ -83,3 +116,147 @@ impl ArtifactGetter for WicketdArtifactStore {
...
+
+    // XXX: might be worth differentiating between server-side issues (503)
+    // and issues with the uploaded archive (400).
+    extractor.extract(temp_path).map_err(RepositoryError::Extract)?;
+
+    // Time is unavailable during initial setup, so ignore expiration. Even
+    // if time were available, we might want to be able to load older
+    // versions of artifacts over the technician port in an emergency.
+    //
+    // XXX we aren't checking against a root of trust at this point --
```
andrewjstone, line 144: I think the term you want here is "trust anchor", not "root of trust".

wicketd/src/mgs.rs
```diff
@@ -79,6 +80,24 @@ impl MgsHandle {
     }
 }
 
+pub fn make_mgs_client(
+    log: Logger,
+    mgs_addr: SocketAddrV6,
+) -> gateway_client::Client {
+    // TODO-correctness Do all users of this client (including both direct API
+    // calls by `UpdatePlanner` and polling by `MgsManager`) want the same
```
andrewjstone, line 88: Probably not as we discovered on Friday. I'm fine leaving for now though.

andrewjstone (approved):
Looks good, and it works as demo'd on Friday. I think it's safe to merge. Leaving it up to you and @sunshowers to figure out the rest of the comments.

wicketd/src/http_entrypoints.rs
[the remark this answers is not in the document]
jgallagher: This was somewhat nontrivial but is a big improvement: [a026753](https://github.com/oxidecomputer/omicron/pull/2375/commits/a0267530f1d034527a4fa76cad79ab8da96592e1)

wicketd/src/update_planner.rs
jgallagher: No problems making this change: [96135ee](https://github.com/oxidecomputer/omicron/pull/2375/commits/96135ee80d6f4d0ad3db830c600014da512b669a)

wicketd/src/update_planner.rs
jgallagher: Done in [47d1d1d](https://github.com/oxidecomputer/omicron/pull/2375/commits/47d1d1d51c466cf8d2f7d64fc0f1fb2f08002ed2)

wicketd/src/artifacts.rs
```diff
@@ -83,3 +116,147 @@ impl ArtifactGetter for WicketdArtifactStore {
...
+    // Create a temporary directory to hold artifacts in (we'll read them
+    // into memory as we extract them).
+    let dir = tempfile::tempdir().map_err(RepositoryError::TempDirCreate)?;
+    let temp_path = <&Utf8Path>::try_from(dir.path()).map_err(|error| {
+        RepositoryError::TempDirCreate(error.into_io_error())
+    })?;
+
+    slog::info!(log, "extracting uploaded archive to {temp_path}");
+
+    // XXX: might be worth differentiating between server-side issues (503)
```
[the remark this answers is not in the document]
jgallagher: I took a stab at this in [5dff92d](https://github.com/oxidecomputer/omicron/pull/2375/commits/5dff92d7c96b37e04ade6a3a342387b11c99b6ac); we can refine further as needed

gateway/src/http_entrypoints.rs
```diff
@@ -927,16 +925,20 @@ async fn sp_reset(
 async fn sp_component_update(
     rqctx: RequestContext<Arc<ServerContext>>,
     path: Path<PathSpComponent>,
-    body: TypedBody<UpdateBody>,
+    query_params: Query<ComponentUpdateIdSlot>,
+    body: UntypedBody,
```
sunshowers, line 929: nice!

wicketd/src/artifacts.rs
```diff
@@ -33,41 +42,64 @@ pub(crate) struct WicketdArtifactStore {
     log: slog::Logger,
     // NOTE: this is a `std::sync::Mutex` rather than a `tokio::sync::Mutex` because the critical
     // sections are extremely small.
-    artifacts: Arc<Mutex<DebugIgnore<HashMap<ArtifactId, BufList>>>>,
+    artifacts_with_plan: Arc<Mutex<ArtifactsWithPlan>>,
```
sunshowers, line 45: This is going to conflict with #2381 (though the conflicts should be easy to address since they're both directionally doing the same thing)

wicketd/src/artifacts.rs
```diff
@@ -125,7 +157,7 @@ fn extract_and_validate(
     // if time were available, we might want to be able to load older
     // versions of artifacts over the technician port in an emergency.
     //
-    // XXX we aren't checking against a root of trust at this point --
+    // XXX we aren't checking against a trust anchor at this point --
```
sunshowers, line 160: Thanks for catching this!

wicketd/src/artifacts.rs
```diff
@@ -236,11 +272,122 @@ enum RepositoryError {
...
+
+    #[error("multiple artifacts found for kind `{0:?}`")]
+    DuplicateArtifactKind(KnownArtifactKind),
+
+    #[error("missing artifact of kind `{0:?}`")]
+    MissingArtifactKind(KnownArtifactKind),
 }
 
 impl RepositoryError {
     fn to_http_error(&self) -> HttpError {
```
sunshowers, line 284: Yes, thanks for fixing this up!

sunshowers (approved):
Looks fantastic, thanks for addressing all the comments!

gateway/examples/config.toml
```diff
@@ -2,6 +2,12 @@
 # Oxide API: example configuration file
 #
 
+# Maximum number of host phase2 trampoline images we're willing to cache. Note
+# that this value is specified in terms of _number of images_, not bytes, and
+# our cache is in-memory. We expect this value to be small in production,
+# potentially even 1 (i.e., only keep the most-recently-uploaded image).
+host_phase2_recovery_image_cache_max_images = 1
```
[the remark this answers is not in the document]
jgallagher: Trampoline phase 2. This is probably a copy/paste from when I was calling it "recovery image" instead of "trampoline image"; I should make a pass through MGS and replace all of those at some poitn.

wicketd/src/artifacts.rs
jgallagher: Ah, yeah, looks like we can drop in `UpdatePlan::new()` for `        // TODO: build out update planner here`, and add the plan to `TufRepoArtifacts`. Do you want to merge 2381 first?

sunshowers: I think you should just land this first since it's ready to go, and I'll rebase on top of your change.

PR #2378, opened 2023-02-17
[nexus] more validation when adding cert for external api

luqmana:
Had an old cert that I added using the API which succeed only to fail when actually trying to use the API. We can catch those kinds of issues a bit earlier.

(There's discussion about whether we'll even keep this api but in the meanwhile probably doesn't hurt?)

### DIFF
--- a/nexus/db-model/src/certificate.rs
+++ b/nexus/db-model/src/certificate.rs
@@ -9,6 +9,7 @@ use nexus_types::external_api::params;
 use nexus_types::external_api::views;
 use nexus_types::identity::Resource;
 use omicron_common::api::external::Error;
+use openssl::asn1::Asn1Time;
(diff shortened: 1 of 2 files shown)


davepacheco:
Cool.  Out of curiosity, what was the failure mode here?  Was it just that the client didn't trust the cert that was served after the update?

luqmana:
> Cool. Out of curiosity, what was the failure mode here? Was it just that the client didn't trust the cert that was served after the update?

Yup, just the usual expired cert kinda error:

```
error sending request for url (https://pluto.oxide-preview.com//system/silos): error trying to connect: invalid peer certificate contents: invalid peer certificate: CertExpired
```

smklein (approved):
Great additions, thanks!

PR #2380, opened 2023-02-17
Update OPTE and use new dedicated dev node.

luqmana:
The kernel driver portion of OPTE (xde) no longer requires non-upstreamed kernel bits and can run on an updated stock helios. Thus we can remove the os-incorporation shuffling from the install script. The uninstall script remains mostly the same to still allow people with modified kernel bits to get back to stock.

### DIFF
--- a/sled-agent/src/opte/illumos/port_manager.rs
+++ b/sled-agent/src/opte/illumos/port_manager.rs
@@ -267,7 +267,7 @@ impl PortManager {
         // The Port object's drop implementation will clean up both of those, if
         // any of the remaining fallible operations fail.
         let port_name = self.inner.next_port_name();
-        let hdl = OpteHdl::open(OpteHdl::DLD_CTL)?;
+        let hdl = OpteHdl::open(OpteHdl::XDE_CTL)?;
 
         // TODO-completeness: Add support for IPv6.
         let vpc_cfg = VpcCfg {
@@ -480,7 +480,7 @@ impl PortManager {
         &self,
         rules: &[VpcFirewallRule],
     ) -> Result<(), Error> {
-        let hdl = OpteHdl::open(OpteHdl::DLD_CTL)?;
+        let hdl = OpteHdl::open(OpteHdl::XDE_CTL)?;
         for ((_, port_name), port) in self.inner.ports.lock().unwrap().iter() {
             let rules = opte_firewall_rules(rules, port.vni(), port.mac());
             let port_name = port_name.clone();
--- a/tools/uninstall_opte.sh
+++ b/tools/uninstall_opte.sh
@@ -31,6 +31,12 @@ PUBLISHERS=("on-nightly" "helios-netdev")
 HELIOS_PUBLISHER="helios-dev"
 STOCK_CONSOLIDATION="pkg://$HELIOS_PUBLISHER/consolidation/osnet/osnet-incorporation"
 
+# Saved list of publisher origins we've removed if we need to restore them
+REMOVED_PUBLISHERS=()
+
+# The version of OPTE currently installed, if any
+INSTALLED_OPTE_VER=""
+
 # Remove a publisher in $1, assuming it matches a few basic assumptions
 function remove_publisher {
     local PUBLISHER="$1"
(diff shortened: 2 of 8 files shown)


tools/uninstall_opte.sh
```diff
@@ -63,7 +63,14 @@ function to_stock_helios {
...
     pkg set-publisher --sticky --search-first "$HELIOS_PUBLISHER"
-    pkg install --no-refresh -v "$CONSOLIDATION"
+
+    # Some packages which don't exist in stock might've been pulled in and
+    # prevent us from installing the stock consolidation package. So, we'll
+    # do a dry-run and pick out any packages we need to reject.
+    local PKGS_TO_REJECT=($(pkg install -n --no-refresh -v "$CONSOLIDATION" 2>&1 | grep -o "Package '.*' must be uninstalled or upgraded" | tr -d \' | cut -d ' ' -f 2 | sed 's/^/--reject /'))
+
+    # Now install stock consolidation but reject the packages we don't want.
+    pkg install --no-refresh -v ${PKGS_TO_REJECT[@]} "$CONSOLIDATION"
```
luqmana, line 73: @jclulow This is a lil adhoc but seems to work. Or is there a better way to do this?

(The specific packages I've needed to reject are `system/gpio` and `system/library/libispi`.)

jclulow: If you're going to depend on the (unstable, for people) output of the tool like this, I would probably more tightly control the environment for this `pkg` invocation.  At a minimum I would scrub out the locale stuff; e.g.,

```sh
LANG= LC_CTYPE= LC_NUMERIC= LC_TIME= LC_COLLATE= LC_MONETARY= \
    LC_MESSAGES= LC_ALL=C.UTF-8 \
    pkg install ...
```

Otherwise it's possible the string you're looking for will be spelt differently depending on the user environment.

I suspect it may also be preferable to just hard-code the list of packages you have there which works (`system/gpio`, `system/library/libispi`) and if this step fails, print an informative message about what one might do next otherwise (i.e., add more to the list).  That would probably be less likely to result in accidentally removing a lot of packages due to some unexpected behaviour on someone's system.

bnaecker:
@luqmana What's the relationship between the updated installation scripts here and these (now very old) scripts in #1907? I'm hitting some issues now that brought me back here, and wondering which solution to use, or whether to merge those in some way.

tools/uninstall_opte.sh
luqmana: Updated it with the constrained locale env. Also added some confirmation asking the user before just trying to uninstall.

luqmana:
> @luqmana What's the relationship between the updated installation scripts here and these (now very old) scripts in #1907? I'm hitting some issues now that brought me back here, and wondering which solution to use, or whether to merge those in some way.

They're a little orthogonal I guess. This PR gets rid of pulling in `os-consolidation` but still installs `driver/network/opte` from the tarball for a specific CI run. In the longer term I believe we're gonna just switch over helios to use the `helios-netdev` packages which would let use `pkg install driver/network/opte` directly.

luqmana:
Bumped rev to pull in fix for oxidecomputer/opte#335 too

luqmana:
Bump for oxidecomputer/opte#339 fix

sled-agent/src/opte/illumos/port_manager.rs
```diff
@@ -267,7 +267,7 @@ impl PortManager {
         // The Port object's drop implementation will clean up both of those, if
         // any of the remaining fallible operations fail.
         let port_name = self.inner.next_port_name();
-        let hdl = OpteHdl::open(OpteHdl::DLD_CTL)?;
+        let hdl = OpteHdl::open(OpteHdl::XDE_CTL)?;
```
bnaecker, line 270: So, this isn't really part of this PR, but I've always found it a bit odd that consumers of `opte-ioctl` need to know anything about `XDE_CTL` or `DLD_CTL` at all. What do you think about making this API work like so?

```rust
let hdl = OpteHdl::open()?;
```

I.e., we use `XDE_CTL` internally, rather than having it part of the API.

bnaecker (approved):
Looks okay to me! I'd love your thoughts on changing the OPTE API that requires consumers specify the device node at all, but it's not really blocking. We can do that separately if we agree it's useful. Thanks for this!

sled-agent/src/opte/illumos/port_manager.rs
luqmana: Opened https://github.com/oxidecomputer/opte/issues/341

PR #2381, opened 2023-02-17
[installinator] add support for fetching artifacts by hash

sunshowers:
In the installinator, we've decided that we're going to fetch artifacts by (kind, hash) rather than by (kind, name, version). Build out the infrastructure to support this.

Notes:

* The installinator is currently expected to fetch the host phase 2 image, not the full host image. There's no known kind for it, but I think using "host_phase_2" as the kind name is OK.
* Since we can no longer have a `host_phase_2` kind in the TUF archive, testing this is a bit annoying. I put up https://github.com/oxidecomputer/omicron/pull/2386 as a way to test against tufaceous for now. Eventually we'll want to use a real archive, and extract a tarball within wicketd.
* I also added a TODO spot where the update planner in #2375 can go (cc @jgallagher).

Tested the fetch process by hand, going via wicket through to wicketd and then to the installinator.

Depends on #2387.


common/src/update.rs
```diff
@@ -159,6 +192,54 @@ impl FromStr for ArtifactKind {
...
+#[serde(transparent)]
+pub struct ArtifactHash(
+    #[serde(with = "serde_human_bytes::hex_array")]
+    #[schemars(schema_with = "hex_schema::<32>")]
+    pub [u8; 32],
+);
+
+impl fmt::Debug for ArtifactHash {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_tuple("ArtifactHash").field(&hex::encode(self.0)).finish()
```
andrewjstone, line 217: Love it :)

common/src/update.rs
```diff
@@ -65,6 +74,30 @@ pub struct ArtifactId {
...
+    Hash,
+    Ord,
+    PartialOrd,
+    Deserialize,
+    Serialize,
+    JsonSchema,
+)]
+pub struct ArtifactHashId {
+    /// The kind of artifact this is.
+    pub kind: ArtifactKind,
```
andrewjstone, line 95: If we didn't unpack in wicketd, but instead sent the whole image, and had installinator unpack, would this be a "known" type?. It feels silly to send the whole image to installinator but it also feels silly to not be able to specify a very well known type here.

installinator/src/dispatch.rs
```diff
@@ -103,47 +111,97 @@ struct InstallOpts {
...
-        // much more RAM available than the size of the host image. (If this changes in the future,
-        // we'll need to have some sort of streaming here.)
-        let artifact = FetchedArtifact::loop_fetch_from_peers(
+        let artifact_hashes = self.artifact_ids.resolve()?;
+
+        let host_phase_2_id = ArtifactHashId {
+            // TODO: currently we're assuming that wicket will unpack the host
+            // phase 2 image. We may instead have the installinator do it.
+            // kind: KnownArtifactKind::Host.into(),
+            kind: ArtifactKind::HOST_PHASE_2,
```
andrewjstone, line 120: Where is this defined?

andrewjstone (approved):
This looks good to me. I'm still unsure of how `ArtifactId::HostPhase2` actually gets defined. I'm starting to wonder if we should decouple the low level artifacts that may be unpacked from the ArtifactId in TUF. This also seems like a silly solution, but I really dislike strings for what should be well defined enum variants. It's fine to have a separate type to parse into an 'Unknown` variant for backwards compatibility, but otherwise it feels wrong.

installinator/src/dispatch.rs
sunshowers: This is defined in https://github.com/oxidecomputer/omicron/pull/2387.

andrewjstone: Thanks!

common/src/update.rs
andrewjstone: Good point. Well, for now let's stick with this as is and see how it goes :)

PR #2384, opened 2023-02-17
Add some missing endpoint deprecations

david-crespo:
Followup to #2057. Noticed these while upgrading the docs site — deprecated endpoints are filtered out to give priority to the v1 ones, but these were not being filtered out.


zephraph (approved):
Thanks for catching these

PR #2386, opened 2023-02-18
[tufaceous] add the ability to add unknown artifact kinds behind a flag

sunshowers:
Currently, tufaceous only supports adding known artifact kinds. However,
in some situations (e.g. testing) it is helpful to add unknown artifact
kinds. Add support for that to tufaceous.


andrewjstone (approved):
LGTM

PR #2387, opened 2023-02-18
[common] define artifact kinds for host/trampoline phase 1/2

sunshowers:
These aren't part of `KnownArtifactKind` because they aren't serialized
in TUF repositories, but defining them in a central location is still
useful.

I did need to go back and use `Cow<'static, str>` to support these
constants.

Depends on #2386.


andrewjstone (approved):
Ship it!

PR #2388, opened 2023-02-18
Better detection of underlay devices based on system ID

bnaecker:
- Adds a `is_gimlet()` function that returns true on, well, Gimlets.
- Use the `is_gimlet()` to detect whether we're on a developer maching, and thus which NICs to use in two places. The first is `find_chelsio_links`, which has thus far been returning the emulated Chelsios, net0 and net1. On non-Gimlets, that's still the case. On Gimlets it returns the list of devices named `cxgbeN`, all actual Chelsio NICs. The second is the physical link that is currently used to determine the bootstrap address. That's net0 or cxgbe0, for non-Gimlets and Gimlets, respecively.
- Adds a few helper functions for creating IPv6 link-locals in a well-defined way, and factors out the name for all of them, which is now `iface/ll`, rather than `iface/linklocal`.

### DIFF
--- a/sled-agent/src/common/underlay.rs
+++ b/sled-agent/src/common/underlay.rs
@@ -31,7 +31,7 @@ pub enum Error {
     BadAddrObj(#[from] addrobj::ParseError),
 
     #[error("Could not determine if host is a Gimlet: {0}")]
-    SystemDetection(#[from] anyhow::Error),
+    SystemDetection(#[source] anyhow::Error),
 
     #[error("Could not enumerate physical links")]
     FindLinks(#[from] FindPhysicalLinkError),
@@ -56,7 +56,7 @@ pub fn find_nics() -> Result<Vec<AddrObject>, Error> {
 /// developer machine, or generally a non-Gimlet, this will return the
 /// VNICs we use to emulate those Chelsio links.
 pub(crate) fn find_chelsio_links() -> Result<Vec<PhysicalLink>, Error> {
-    if is_gimlet()? {
+    if is_gimlet().map_err(Error::SystemDetection)? {
         Dladm::list_physical().map_err(Error::FindLinks).map(|links| {
             links
                 .into_iter()
--- a/sled-agent/src/common/underlay.rs
+++ b/sled-agent/src/common/underlay.rs
@@ -4,12 +4,20 @@
 
 //! Finding the underlay network physical links and address objects.
 
+use crate::hardware::is_gimlet;
 use crate::illumos::addrobj;
 use crate::illumos::addrobj::AddrObject;
+use crate::illumos::dladm::Dladm;
+use crate::illumos::dladm::FindPhysicalLinkError;
 use crate::illumos::dladm::PhysicalLink;
+use crate::illumos::dladm::CHELSIO_LINK_PREFIX;
 use crate::zone::Zones;
 
 // Names of VNICs used as underlay devices for the xde driver.
+//
+// NOTE: These are only used on non-Gimlet systems. In that case, they are
+// expected to have been created by a run of
+// `./tools/create_virtual_hardware.sh`.
 const XDE_VNIC_NAMES: [&str; 2] = ["net0", "net1"];
 
 #[derive(thiserror::Error, Debug)]
@@ -21,6 +29,12 @@ pub enum Error {
 
     #[error(transparent)]
     BadAddrObj(#[from] addrobj::ParseError),
+
+    #[error("Could not determine if host is a Gimlet: {0}")]
+    SystemDetection(#[from] anyhow::Error),
+
+    #[error("Could not enumerate physical links")]
+    FindLinks(#[from] FindPhysicalLinkError),
 }
 
 pub fn find_nics() -> Result<Vec<AddrObject>, Error> {
@@ -28,7 +42,7 @@ pub fn find_nics() -> Result<Vec<AddrObject>, Error> {
 
     let mut addr_objs = Vec::with_capacity(underlay_nics.len());
     for nic in underlay_nics {
-        let addrobj = AddrObject::new(&nic.0, "linklocal")?;
+        let addrobj = AddrObject::link_local(&nic.0)?;
         Zones::ensure_has_link_local_v6_address(None, &addrobj)?;
         addr_objs.push(addrobj);
     }
@@ -36,13 +50,23 @@ pub fn find_nics() -> Result<Vec<AddrObject>, Error> {
     Ok(addr_objs)
 }
 
+/// Return the Chelsio links on the system.
+///
+/// For a real Gimlet, this should return the devices like `cxgbeN`. For a
+/// developer machine, or generally a non-Gimlet, this will return the
+/// VNICs we use to emulate those Chelsio links.
 pub(crate) fn find_chelsio_links() -> Result<Vec<PhysicalLink>, Error> {
-    // TODO-correctness: This should eventually be determined by a call to
-    // `Dladm` to get the real Chelsio links on a Gimlet. These will likely be
-    // called `cxgbeN`, but we explicitly call them `netN` to be clear that
-    // they're likely VNICs for the time being.
-    Ok(XDE_VNIC_NAMES
-        .into_iter()
-        .map(|name| PhysicalLink(name.to_string()))
-        .collect())
+    if is_gimlet()? {
+        Dladm::list_physical().map_err(Error::FindLinks).map(|links| {
+            links
+                .into_iter()
+                .filter(|link| link.0.starts_with(CHELSIO_LINK_PREFIX))
+                .collect()
+        })
+    } else {
+        Ok(XDE_VNIC_NAMES
+            .into_iter()
+            .map(|name| PhysicalLink(name.to_string()))
+            .collect())
+    }
 }
--- a/sled-agent/src/config.rs
+++ b/sled-agent/src/config.rs
@@ -57,7 +57,7 @@ pub enum ConfigError {
         err: toml::de::Error,
     },
     #[error("Could not determine if host is a Gimlet: {0}")]
-    SystemDetection(#[from] anyhow::Error),
+    SystemDetection(#[source] anyhow::Error),
     #[error("Could not enumerate physical links")]
     FindLinks(#[from] FindPhysicalLinkError),
 }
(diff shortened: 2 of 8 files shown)


bnaecker:
This addresses an issue @leftwo hit during dog-fooding. We've been basically hard-coding the "simulated Chelsios" as VNICs `net0` and `net1` for a long time. Starting the sled-agent on a real Gimlet obviously keels over with that, and rightly so. This adds a method `is_gimlet()` to detect whether the host system is a Gimlet, and uses that in a few places.

I've tested this on a non-Gimlet, and everything seems to come up correctly. We use `net0/ll` and `net1/ll` as the interfaces for DDM, and the corresponding data links when setting up the `xde` driver. I'm in the process of testing it on a Gimlet in Rack2 now, and will report that before merging.

bnaecker:
All of the failures in the `build-and-test` check are `EACCES` when trying to query the devinfo root node to detect whether the system is a Gimlet. I guess I assumed the sled-agent ran with elevated permissions in this scenario, but that's obviously not correct. We're able to do things like create IP addresses because we shell out to `pfexec ipadm`.

bnaecker:
I have a bad way to fix this:

```diff
diff --git a/sled-agent/src/hardware/illumos/mod.rs b/sled-agent/src/hardware/illumos/mod.rs
index 3a0279ec..0f617465 100644
--- a/sled-agent/src/hardware/illumos/mod.rs
+++ b/sled-agent/src/hardware/illumos/mod.rs
@@ -43,6 +43,7 @@ enum Error {
 const GIMLET_ROOT_NODE_NAME: &str = "Oxide,Gimlet";

 /// Return true if the host system is an Oxide Gimlet.
+#[cfg(not(test))]
 pub fn is_gimlet() -> anyhow::Result<bool> {
     let mut device_info = DevInfo::new_force_load()?;
     let mut node_walker = device_info.walk_node();
@@ -52,6 +53,15 @@ pub fn is_gimlet() -> anyhow::Result<bool> {
     Ok(root.node_name() == GIMLET_ROOT_NODE_NAME)
 }

+/// Return true if the host system is an Oxide Gimlet.
+//
+// TODO-testing: This assumes we never test on real Gimlets. That seems like a
+// very bad assumption.
+#[cfg(test)]
+pub fn is_gimlet() -> anyhow::Result<bool> {
+    Ok(false)
+}
+
 // A snapshot of information about the underlying Tofino device
 #[derive(Copy, Clone)]
 struct TofinoSnapshot {
```

sled-agent/src/common/underlay.rs
```diff
@@ -21,28 +29,44 @@ pub enum Error {
 
     #[error(transparent)]
     BadAddrObj(#[from] addrobj::ParseError),
+
+    #[error("Could not determine if host is a Gimlet: {0}")]
+    SystemDetection(#[from] anyhow::Error),
```
jgallagher, line 34: This `#[from]` feels a little dicey to me because `anyhow::Error` is so generic - now any function that returns an `Error` but calls a function returning `anyhow::Error` can use `?`, and it will automatically become `Error::SystemDetection`. I think I'd change this to `#[source]` and use explicit `.map_err(Error::SystemDetection)` calls.

sled-agent/src/config.rs
```diff
@@ -46,6 +56,10 @@ pub enum ConfigError {
         #[source]
         err: toml::de::Error,
     },
+    #[error("Could not determine if host is a Gimlet: {0}")]
+    SystemDetection(#[from] anyhow::Error),
```
jgallagher, line 60: Same comment as above w.r.t. `#[from] anyhow::Error`

sled-agent/src/hardware/illumos/mod.rs
```diff
@@ -40,6 +40,28 @@ enum Error {
...
 
+const GIMLET_ROOT_NODE_NAME: &str = "Oxide,Gimlet";
+
+/// Return true if the host system is an Oxide Gimlet.
+#[cfg(not(test))]
+pub fn is_gimlet() -> anyhow::Result<bool> {
+    let mut device_info = DevInfo::new_force_load()?;
+    let mut node_walker = device_info.walk_node();
+    let Some(root) = node_walker.next().transpose()? else {
+        return Err(anyhow::anyhow!("No nodes in device tree"));
```
jgallagher, line 51: Style nit - anyhow provides [`bail!`](https://docs.rs/anyhow/latest/anyhow/macro.bail.html) for this use:

```suggestion
        anyhow::bail!("No nodes in device tree");
```

sled-agent/src/hardware/illumos/mod.rs
```diff
@@ -40,6 +40,28 @@ enum Error {
...
+    };
+    Ok(root.node_name() == GIMLET_ROOT_NODE_NAME)
+}
+
+/// Return true if the host system is an Oxide Gimlet.
+//
+// TODO-testing: This assumes we never test on real Gimlets. That seems like a
+// very bad assumption.
```
jgallagher, line 59: Agreed :). A couple suggestions:

* Can the permission problem be fixed in CI? (Do we need to `pfexec` the test runs?)
* If not, as a slightly-less-painful bandaid could we change this to call all the same functions that `is_gimlet` calls, but return `Ok(false)` on any error? I.e., the `#[cfg(test)]` version never fails, but on a real gimlet with correct permissions it could still return `Ok(true)`.

bnaecker: >  Can the permission problem be fixed in CI? (Do we need to pfexec the test runs?)

I'm not sure, though I'm a bit hesitant to `pfexec` the _entire_ test suite. It's all in a VM, I guess, but it feels like a big hammer to me.

> If not, as a slightly-less-painful bandaid could we change this to call all the same functions that is_gimlet calls, but return Ok(false) on any error? I.e., the #[cfg(test)] version never fails, but on a real gimlet with correct permissions it could still return Ok(true).

That might make sense, yeah. We're currently running this in three contexts: non-Gimlet developer machines, CI, and potentially real Gimlets. In the first two, `Ok(false)` is _correct_ even if we get the answer for the wrong reason. On the latter, we'll get the right answer since everyone is currently logging in as `root` anyway. That's not great, but it's a separate issue from this one. When we deal with it, we can be sure to configure permissions such that calling `libdevinfo` doesn't fail anyway.

Thanks for the suggestion @jgallagher, very helpful as usual.

jgallagher (approved):
Looks good! :+1:

bnaecker:
Thanks John. Just confirming now on a Gimlet that this did the right thing, and didn't bork anything else!

bnaecker:
Ok, things look right to me. @leftwo picked up this branch and installed Omicron. That was done via `OMICRON_NO_UNINSTALL=1 ./omicron-package install`. That environment variable instructs the tool not to wipe out all the data-links and zones, which is critical since that would destroy the link-local IPv6 address we use to get _into_ the Gimlet. The whole control plane came up as expected. The `cxgbe{0,1}` data links were use for DDM, and we created link-locals on them with the name `ll`. Those links were also provided as the underlay devices to `xde`. Here are the relevant snippets of the sled-agent log:

```
{"msg":"Setting mg-ddm interfaces","v":0,"name":"SledAgent","level":30,"time":"2023-02-21T18:04:31.584941801Z","hostname":"BRM42220070","pid":22393,"interfaces":"(\"cxgbe0/ll\" \"cxgbe1/ll\")"}
```

and

```
{"msg":"using '[AddrObject { interface: \"cxgbe0\", name: \"ll\" }, AddrObject { interface: \"cxgbe1\", name: \"ll\" }]' as data links for xde driver","v":0,"name":"SledAgent","level":30,"time":"2023-02-21T18:04:34.385224563Z","hostname":"BRM42220070","pid":22393,"sled_id":"9a721086-36f5-4a39-af55-1c389fe5eaae","component":"SledAgent"}
```

Here are the links and addresses:

```bash
BRM42220070 # dladm
tfpkt0      phys      1500   down     --         --
cxgbe0      phys      9000   up       --         --
cxgbe1      phys      9000   up       --         --
bootstrap_stub0 etherstub 9000 up     --         --
bootstrap0  vnic      9000   up       --         bootstrap_stub0
underlay_stub0 etherstub 9000 up      --         --
underlay0   vnic      9000   up       --         underlay_stub0
oxControlService0 vnic 9000  up       --         underlay_stub0
oxControlStorage0 vnic 9000  up       --         underlay_stub0
oxControlStorage1 vnic 9000  up       --         underlay_stub0
oxControlStorage2 vnic 9000  up       --         underlay_stub0
oxControlStorage3 vnic 9000  up       --         underlay_stub0
oxControlStorage4 vnic 9000  up       --         underlay_stub0
oxControlStorage5 vnic 9000  up       --         underlay_stub0
oxControlStorage6 vnic 9000  up       --         underlay_stub0
oxControlStorage7 vnic 9000  up       --         underlay_stub0
oxControlStorage8 vnic 9000  up       --         underlay_stub0
oxControlStorage9 vnic 9000  up       --         underlay_stub0
oxControlStorage10 vnic 9000 up       --         underlay_stub0
oxControlService1 vnic 9000  up       --         underlay_stub0
oxControlPublic0 vnic 9000   up       --         cxgbe0
oxControlService2 vnic 9000  up       --         underlay_stub0
oxControlService3 vnic 9000  up       --         underlay_stub0
BRM42220070 # ipadm
cxgbe0/ll         addrconf ok           fe80::aa40:25ff:fe04:154/10
cxgbe1/ll         addrconf ok           fe80::aa40:25ff:fe04:15c/10
bootstrap0/ll     addrconf ok           fe80::8:20ff:fe34:1d04/10
bootstrap0/bootstrap6 static ok         fdb0:a840:2504:154::1/64
underlay0/ll      addrconf ok           fe80::8:20ff:feae:958a/10
underlay0/sled6   static   ok           fd00:1122:3344:101::1/64
underlay0/internaldns static ok         fd00:1122:3344:1::2/64
BRM42220070 #
```

The public address for Nexus is over the `cxgbe0` link. That's not functional, but it is expected at this point.

sled-agent/src/hardware/illumos/mod.rs
jclulow: For what it's worth if you're having devinfo permissions errors it's probably because we're passing a flag to try and force attach stuff, which requires root.  It would be reasonable at least for the "am I a Gimlet" case to _not_ pass that flag, as I believe you're just looking at properties on the root node which is always there and doesn't require any flags to see.

bnaecker: Thanks Josh, that's helpful. I think we _are_ indeed force-attaching. I'll try it out with out that now, and hopefully we can remove the test-only version of this.

PR #2390, opened 2023-02-19
make internal-dns package more generic

davepacheco:
In prep for an external DNS component using the same pieces as internal DNS (see RFD 357), this change:

- renames the `internal-dns` Rust crate to `dns-server`
- renames the `internal-dns-client` Rust crate to `dns-service-client`.  (I'm open to better names here.  I thought `dns-client` would be confusing because this is an _http_ client to _configure_ the DNS server, not a DNS protocol client.)
- updates the package manifest so that we still deliver an `internal-dns` package

Things I still want to do here:

- [x] verify that the built internal-dns tarball is the same as before
- [x] add an `external-dns` package to package-manifest
- [x] review the changes
- [x] move the internal service names to a separate package -- they don't really belong in `dns-service-client` because they really _are_ internal-DNS-specific


davepacheco:
I compared the internal-dns package produced by commit 7bfbfef4f322885e3d602ca6b052c03cf4a20e7b (from this PR, calling this "new") with that produced by commit 424eb9667b11d8940c0c88a116fe0f8709331f36 (from "main" earlier today, calling this "old").

First, a summary of what's in the tarballs (these transcripts are all a bit edited for clarity):

```
dap@ivanova ~ $ mkdir internal-dns-compare
dap@ivanova ~ $ cd internal-dns-compare/
dap@ivanova internal-dns-compare $ tar tzvf $HOME/omicron-clean/out/internal-dns.tar.gz | sort > old-contents
dap@ivanova internal-dns-compare $ tar tzvf $HOME/omicron-iter/out/internal-dns.tar.gz | sort > new-contents
$ diff old-contents new-contents 
2,3c2,3
< -rw-r--r--   0/0      231 Jul 23 18:21 2006 root/var/svc/manifest/site/internal_dns/config.toml
< -rw-r--r--   0/0     1575 Jul 23 18:21 2006 root/var/svc/manifest/site/internal_dns/manifest.xml
---
> -rw-r--r--   0/0      231 Jul 23 18:21 2006 root/var/svc/manifest/site/internal-dns/config.toml
> -rw-r--r--   0/0     1571 Jul 23 18:21 2006 root/var/svc/manifest/site/internal-dns/manifest.xml
8,9c8,9
< -rwxr-xr-x   0/0        0 Jul 23 18:21 2006 root/opt/oxide/internal_dns
< -rwxr-xr-x   0/0        0 Jul 23 18:21 2006 root/opt/oxide/internal_dns/bin
---
> -rwxr-xr-x   0/0        0 Jul 23 18:21 2006 root/opt/oxide/dns-server
> -rwxr-xr-x   0/0        0 Jul 23 18:21 2006 root/opt/oxide/dns-server/bin
14,17c14,17
< -rwxr-xr-x   0/0        0 Jul 23 18:21 2006 root/var/svc/manifest/site/internal_dns
< -rwxr-xr-x   0/0   30746024 Jul 23 18:21 2006 root/opt/oxide/internal_dns/bin/dns-server
< -rwxr-xr-x   0/0   32975584 Jul 23 18:21 2006 root/opt/oxide/internal_dns/bin/dnsadm
< Decompressing '/home/dap/omicron-clean/out/internal-dns.tar.gz' with '/usr/bin/gzcat'...
---
> -rwxr-xr-x   0/0        0 Jul 23 18:21 2006 root/var/svc/manifest/site/internal-dns
> -rwxr-xr-x   0/0   30755496 Jul 23 18:21 2006 root/opt/oxide/dns-server/bin/dns-server
> -rwxr-xr-x   0/0   32994416 Jul 23 18:21 2006 root/opt/oxide/dns-server/bin/dnsadm
> Decompressing '/home/dap/omicron-iter/out/internal-dns.tar.gz' with '/usr/bin/gzcat'...
```

Now the contents:

```
dap@ivanova internal-dns-compare $ mkdir old
dap@ivanova internal-dns-compare $ tar xzf $HOME/omicron-clean/out/internal-dns.tar.gz -C old
dap@ivanova internal-dns-compare $ mkdir new
dap@ivanova internal-dns-compare $ tar xzf $HOME/omicron-iter/out/internal-dns.tar.gz -C new
dap@ivanova internal-dns-compare $ diff -r old new
Common subdirectories: old/root and new/root
Common subdirectories: old/root/opt and new/root/opt
Common subdirectories: old/root/var and new/root/var
Common subdirectories: old/root/opt/oxide and new/root/opt/oxide
Only in new/root/opt/oxide: dns-server
Only in old/root/opt/oxide: internal_dns
Common subdirectories: old/root/var/svc and new/root/var/svc
Common subdirectories: old/root/var/svc/manifest and new/root/var/svc/manifest
Common subdirectories: old/root/var/svc/manifest/site and new/root/var/svc/manifest/site
Only in new/root/var/svc/manifest/site: internal-dns
Only in old/root/var/svc/manifest/site: internal_dns
dap@ivanova internal-dns-compare $ diff -r new/root/opt/oxide/dns-server old/root/opt/oxide/internal_dns/
Common subdirectories: new/root/opt/oxide/dns-server/bin and old/root/opt/oxide/internal_dns//bin
Binary files new/root/opt/oxide/dns-server/bin/dns-server and old/root/opt/oxide/internal_dns//bin/dns-server differ
Binary files new/root/opt/oxide/dns-server/bin/dnsadm and old/root/opt/oxide/internal_dns//bin/dnsadm differ
dap@ivanova internal-dns-compare $ diff -r new/root/var/svc/manifest/site/internal-dns old/root/var/svc/manifest/site/internal_dns
diff -r new/root/var/svc/manifest/site/internal-dns/manifest.xml old/root/var/svc/manifest/site/internal_dns/manifest.xml
16c16
<       exec='ctrun -l child -o noorphan,regent /opt/oxide/dns-server/bin/dns-server --config-file /var/svc/manifest/site/dns-server/config.toml --server-address %{config/server_address} --dns-address %{config/dns_address} --dns-zone %{config/dns_zone} &amp;'
---
>       exec='ctrun -l child -o noorphan,regent /opt/oxide/internal_dns/bin/dns-server --config-file /var/svc/manifest/site/internal_dns/config.toml --server-address %{config/server_address} --dns-address %{config/dns_address} --dns-zone %{config/dns_zone} &amp;'
```

I think this shows that the only changes in the internal-dns package are:

* /opt/oxide/internal_dns got moved to /opt/oxide/dns-server (and the reference in the SMF manifest was updated)
* /var/svc/manifest/site/internal_dns got moved to /var/svc/manifest/site/dns-server
* the built binaries are different

I think that's all expected.  I'm not sure what's up with the hyphens vs. underscores but I tried to use hyphens everywhere because that seems to be what the paths inside the repo use and that's also what some packages use (e.g., `sled-agent`).

davepacheco:
CI keeps failing because I'm mixing up hyphens and underscores.  I'm trying to figure out what to do here.

Earlier (in 7bfbfef4f322885e3d602ca6b052c03cf4a20e7b), the service names had underscores ("internal_dns").  I did that because I saw sled-agent had a hyphen in its service name.  The CI "deploy" job saw this error on that commit:

> Failed to issue SMF command: Failed to do 'importing manifest' by running command in zone: Error running command in zone 'oxz_internal_dns': Command [/usr/sbin/zlogin oxz_internal_dns /usr/sbin/svccfg import /var/svc/manifest/site/internal_dns/manifest.xml] executed and failed with status: exit status: 1  stdout:   stderr: svccfg: Unable to stat file /var/svc/manifest/site/internal_dns/manifest.xml.  No such file or directory\n

I inferred (maybe wrongly?) that the service name should have a hyphen instead.  So I switched to that in 660c703e349f157a1731f7250376763d2b1194b9.  Then I saw this error from the same job:

> Failed to install zone 'oxz_internal_dns' from '/opt/oxide/internal_dns.tar.gz': Failed to execute zoneadm command 'Install' for zone 'oxz_internal_dns': Failed to parse command output: exit code 1\nstdout:\nA ZFS file system has been created for this zone.\nINFO: omicron: installing zone oxz_internal_dns @ "/zone/oxz_internal_dns"...\nINFO: omicron: replicating /usr tree...\nINFO: omicron: replicating /lib tree...\nINFO: omicron: replicating /sbin tree...\nINFO: omicron: pruning SMF manifests...\nINFO: omicron: pruning global-only files...\nINFO: omicron: unpacking baseline archive...\nINFO: omicron: unpacking image "/opt/oxide/internal_dns.tar.gz"...\nstderr:\nError: No such file or directory (os error 2) 

So zone install appears to depend on the image tarball having an underscore.  That may be because the internal service name (in Rust, not the package-manifest) has an underscore?

I'm not clear on what the convention should be:

- In package-manifest.toml, there are 3 service_names that have a hyphen (sled-agent, propolis-server, and mg-ddm) and 2 that have an underscore (crucible_pantry, internal_dns).  But those three names don't correspond to internal service zones.
- In terms of omicron_sled_agent::params::ServiceType, we use snake case (underscores).  Ditto ZoneType in the same module.  Ditto omicron_nexus::db_model::ServiceKind.
- In the "smf" directory, the directories use hyphens.  In the actual SMF manifests there, we use hyphens for the SMF service names.

I think I will try going back to underscores in package-manifest.toml, since that's what it was before this change, and that should match the Rust enum values.  The question remains why I got that first error about the wrong SMF manifest.  I guess it's because I changed the delivered path of the SMF manifest to internal-dns (to match the package name) instead of internal_dns.

.github/buildomat/jobs/deploy.sh
```diff
@@ -5,7 +5,8 @@
 #: target = "lab-opte-0.21"
 #: output_rules = [
 #:	"%/var/svc/log/system-illumos-sled-agent:default.log",
-#:	"%/zone/oxz_nexus/root/var/svc/log/system-illumos-nexus:default.log",
+#:	"%/zone/oxz_*/root/var/svc/log/system-illumos-*.log",
+#:	"!/zone/oxz_propolis-server_*/root/var/svc/log/*.log",
```
smklein, line 9: What's the difference between a line starting with "%" and "!" here?

package-manifest.toml
```diff
@@ -89,10 +89,37 @@ setup_hint = "Run `./tools/ci_download_cockroachdb` to download the necessary bi
...
 service_name = "internal_dns"
+source.type = "composite"
+source.packages = [ "dns-server.tar.gz", "internal-dns-customizations.tar.gz" ]
+output.type = "zone"
+
+[package.external-dns]
+service_name = "external_dns"
+source.type = "composite"
+source.packages = [ "dns-server.tar.gz", "external-dns-customizations.tar.gz" ]
+output.type = "zone"
```
smklein, line 100: At a glance, I think this packaging makes sense. I do want to check-in with you though - does this feel weird / unwieldly from your point-of-view?

(For example, I suppose I could support adding `paths` directly to a `composite` type package, so the intermediate "customizations" package wouldn't be needed?)

smklein (approved):
Looks good!

.github/buildomat/jobs/deploy.sh
davepacheco: `%` causes matching files to be uploaded on a best-effort basis, meaning that if they change (or are removed, after oxidecomputer/buildomat#16), the job won't fail.

`!` negates any other pattern, meaning that if any file matches a `!` pattern, it will not be uploaded even if it also matches a regular pattern (no prefix or `%`).  We have this here for propolis basically to work around buildomat#16.

davepacheco: (deeper reference: https://github.com/oxidecomputer/buildomat#configuration, "output rules")

package-manifest.toml
davepacheco: It feels okay.  (Truthfully, this model was a bit unintuitive to me.  I was surprised that if an Omicron package uses a Rust package, it must have the same name as the Rust package.  That made it seem like they're 1-1 with Rust packages.  But they're not, since there are both Omicron packages with no Rust package and Rust packages with no Omicron package.  So at first I thought I was going to have to extend the packaging system if I wanted to use the same Rust package in two different Omicron packages.  I was glad it turned out to be a lot simpler than that, and it's pretty straightforward.  I think I'd need more experience with this to suggest any concrete improvements.  So that's why I say it feels okay.)

PR #2398, opened 2023-02-21
[installinator] stop spamming cargo test logs

### LINKED ISSUE #2389: Installinator tests are quite verbose
When running `cargo test` at the top-level in Omicron, I've noticed that the installinator tests are quite noisy, spitting out many thousands of log lines. This can make finding failures a bit tricky. Most tests seem to use `omicron_test_utils::dev::test_setup_log()` as a useful way to get all data on failures, but avoid printing on success. I think [`slog_term::TestStdoutWriter`](https://docs.rs/slog-term/latest/slog_term/struct.TestStdoutWriter.html) is another useful tool here, ensuring the test harness captures log output like a normal `println!()`.

sunshowers:
Fixes #2389.

### DIFF
--- a/installinator/src/mock_peers.rs
+++ b/installinator/src/mock_peers.rs
@@ -423,12 +423,11 @@ mod tests {
     use crate::{
         errors::DiscoverPeersError,
         peers::{FetchedArtifact, Peers},
-        stderr_env_drain,
     };
 
     use bytes::Buf;
     use futures::future;
-    use slog::Drain;
+    use omicron_test_utils::dev::test_setup_log;
     use test_strategy::proptest;
 
     use std::future::Future;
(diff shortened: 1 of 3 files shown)


andrewjstone (approved):

installinator/src/mock_peers.rs
```diff
@@ -441,19 +440,21 @@ mod tests {
         #[strategy((0..2000u64).prop_map(Duration::from_millis))]
         timeout: Duration,
     ) {
+        let log = test_setup_log("proptest_fetch_artifact");
```
jgallagher, line 443: This returns a log context, and I think you need to call `logctx.cleanup_successful()` at the end of the test.

sunshowers: Thanks, done.

PR #2400, opened 2023-02-21
Don't force-attach devinfo tree when detecting if system is a Gimlet

bnaecker:


bnaecker:
I tested this on my local machine, where the previous version would indeed fail with `EACCES`. This should be able to detect if the system is a Gimlet with no elevated permissions. Thanks @jclulow!

luqmana (approved):

PR #2404, opened 2023-02-21
package manifest could use more explanation

davepacheco:
It took me longer than I expected to wrap my head around omicron-package so I thought I'd add some comments to the manifest file explaining the basics.  (My apologies if this already existed somewhere else but I didn't find it.)


smklein (approved):
Thanks for this addition! This file ended up growing a bit organically, so I appreciate the notes!

PR #2405, opened 2023-02-22
use ArtifactHash for InstallinatorImageIds

sunshowers:
Now that we have an ArtifactHash type, we can use it throughout for
installinator image IDs. This type serializes to hex strings in
human-readable contexts like JSON and query params, and bytes otherwise
like in CBOR.


openapi/gateway.json
```diff
@@ -1197,24 +1197,12 @@
...
           },
           "host_phase_2": {
-            "type": "array",
-            "items": {
-              "type": "integer",
-              "format": "uint8",
-              "minimum": 0
-            },
-            "minItems": 32,
-            "maxItems": 32
```
sunshowers, line 1217: This API changes in a way that I think makes sense: now you'll be able to read artifact hashes as hex strings in messages.

openapi/gateway.json
```diff
@@ -1197,24 +1197,12 @@
...
+            "type": "string",
+            "format": "hex string (32 bytes)"
```
jgallagher, line 1201: Just making sure I understand - this field is effectively free form / up to clients to interpret, assuming it isn't one of the listed formats in the OpenAPI spec (a very short list)? Should it be something that's easier to parse by a machine? `"hex"` is appealing but loses the length requirement, for example.

sunshowers: Yeah, it's a free-form field. I think we could teach progenitor about a format like "hex32" (to match e.g. uint64) and then switch to that in the future.

sunshowers: (maybe "hex256")

jgallagher: Gotcha, makes sense. Thanks.

jgallagher (approved):

PR #2410, opened 2023-02-22
[wicketd] Updating hosts, part 1: Delivering the trampoline image

jgallagher:
This is an incomplete chunk of driving sled recovery from wicketd, and is the first of several PRs (1 more today, and more to come subsequently). I'm sure this isn't 100% right, but should be roughly the right shape, and I hope the review burden is lighter by staging things this way.

This PR covers:

* Adding additional artifact types to the `UpdatePlan`
* Delivering the trampoline phase 1 image to the SP via MGS
* Uploading the trampoline phase 2 image to MGS (which is a little more complicated because it only needs to be done once per trampoline image, not once per SP)


andrewjstone (approved):
Thanks @jgallagher. Lot's of details to keep together here. Thanks for the great work!

PR #2411, opened 2023-02-22
[wicketd] Updating hosts, part 2: Configuring installinator

jgallagher:
This builds on #2410 and continues implementing sled recovery (although it is still very incomplete!).

This PR covers:

* Hashing the extracted-from-the-host-artifact host phase 2 "artifact" (recorded in `UpdatePlan` and added to the `by_hash` collection of artifacts)
* Adding the control plane artifact hash to `UpdatePlan`
* Generating an `InstallinatorImageId` and passing it down to the target SP
* Configuring host startup options and booting the host


andrewjstone (approved):
LGTM!

sunshowers:
Landing with permission from jgallagher :)

Issue #2406, opened 2023-02-22
sled-agent creates crucible zones on M.2 devices


leftwo:
While trying to install Omicron on the dogfood rack, I found that sled-agent had created
zpools on the M.2 devices, and spun up crucible zones on those zpools:

On `BRM42220070`, The gimlet in cubby 26 in the dogfood rack we have:
```
BRM42220070 # diskinfo 
TYPE    DISK                    VID      PID              SIZE          RMV SSD
NVME    c1t00A0750132753688d0   NVMe     Micron_7300_MTFDHBG1T9TDF 1788.50 GiB   no  yes
NVME    c2t0014EE81000D2E9Dd0   NVMe     WUS4C6432DSP3X3  2980.82 GiB   no  yes
NVME    c3t0014EE81000D2EEAd0   NVMe     WUS4C6432DSP3X3  2980.82 GiB   no  yes
NVME    c4t0014EE81000D3035d0   NVMe     WUS4C6432DSP3X3  2980.82 GiB   no  yes
NVME    c5t0014EE81000D2EE9d0   NVMe     WUS4C6432DSP3X3  2980.82 GiB   no  yes
NVME    c6t0014EE81000D2FFAd0   NVMe     WUS4C6432DSP3X3  2980.82 GiB   no  yes
NVME    c7t0014EE81000D2F5Bd0   NVMe     WUS4C6432DSP3X3  2980.82 GiB   no  yes
NVME    c8t00A0750132753657d0   NVMe     Micron_7300_MTFDHBG1T9TDF 1788.50 GiB   no  yes
NVME    c9t0014EE81000D307Fd0   NVMe     WUS4C6432DSP3X3  2980.82 GiB   no  yes
NVME    c10t0014EE81000D2E50d0  NVMe     WUS4C6432DSP3X3  2980.82 GiB   no  yes
NVME    c11t0014EE81000D2E4Dd0  NVMe     WUS4C6432DSP3X3  2980.82 GiB   no  yes
NVME    c12t0014EE81000D2E4Ad0  NVMe     WUS4C6432DSP3X3  2980.82 GiB   no  yes
```

If I look at zpool list, I see 13 zpools:
```
BRM42220070 # zpool list
NAME                                       SIZE  ALLOC   FREE  CKPOINT  EXPANDSZ   FRAG    CAP  DEDUP    HEALTH  ALTROOT
[...]

PR #2412, opened 2023-02-22
[sled agent] RSS no longer provisions datasets on M.2 devices

smklein:
Fixes https://github.com/oxidecomputer/omicron/issues/2406


sled-agent/src/storage_manager.rs
```diff
@@ -1203,17 +1211,49 @@ impl StorageManager {
...
+        let mut zpools = Vec::with_capacity(pools.len());
+
+        for (name, pool) in pools.iter() {
+            let disk_type = if let Some(disk_path) = &pool.disk_path {
+                if let Some(disk) = disks.get(disk_path) {
+                    disk.variant().into()
+                } else {
+                    // If the zpool claims to be attached to a disk that we
+                    // don't know about, that's an error.
+                    return Err(Error::ZpoolNotFound(
```
leftwo, line 1239: If we bail here, what happens?   Will the sled agent be able to continue working?

sled-agent/src/rack_setup/plan/service.rs
```diff
@@ -154,7 +157,7 @@ impl Plan {
             //
             // Once this responsibility shifts to Nexus, we actually only
             // need enough zpools to provision CRDB.
-            if zpools.len() < MINIMUM_ZPOOL_COUNT {
+            if zpools.len() < MINIMUM_U2_ZPOOL_COUNT {
```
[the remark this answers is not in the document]
smklein: Good call, removed in 17013b0c91e03a8e1a44cdb807cedb3ec64c2a5f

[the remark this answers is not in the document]
smklein: For context, I think this comment came from an era when we explicitly enumerated all u.2s -- this is no longer the case, as they generally show up asynchronously

sled-agent/src/storage_manager.rs
smklein: Yes - this is only an error being returned from a querying function, which is only called by RSS surveying the zpool landscape right now.

I could opt to simply ignore the zpool causing the error, but I'm leaning towards "loud errors, returned as early as possible" while we're still in development.

leftwo: I think being loud about errors is good.
I just was concerned about the rest of sled agent making forward progress if errors happen.

leftwo (approved):

PR #2418, opened 2023-02-23
Add ability to import blocks to disks

jmpesp:
This PR introduces the concept of importing blocks into disks by adding the following:

- in the state machine diagram for a Disk, a disk starts at "creating" and can now transition to either "detached" or a new state "import ready"

- when a disk enters the "import ready" state, it is attached to a Pantry, and will remain attached until "finalized"

- two new operations can now start from state "import ready": importing blocks via bulk writes, and importing blocks from a URL

- importing blocks from a URL will kick off a saga that contacts the associated Pantry and submits an "import from URL" job.

- importing blocks via bulk writes requires:
  1. manually setting the disk's state to "importing from bulk writes"
  2. calling a new bulk write endpoint to POST blocks in
  3. manually setting the state back to "import ready" when done

- when importing blocks is completed, a user will "finalize" a disk, which transitions the state from "import ready" to "detached", and optionally takes a snapshot before doing so

This PR also implements creating global images from snapshots.


nexus/src/app/disk.rs
```diff
@@ -408,4 +352,277 @@ impl super::Nexus {
...
+        opctx: &OpContext,
+        disk_selector: &params::DiskSelector,
+        params: params::ImportBlocksFromUrl,
+    ) -> UpdateResult<()> {
+        let authz_disk: authz::Disk;
+
+        (_, _, _, authz_disk, _) = self
+            .disk_lookup(&opctx, &disk_selector)?
+            .fetch_for(authz::Action::Modify)
+            .await?;
```
zephraph, line 368:
```suggestion
        (_, _, _, authz_disk) = self
            .disk_lookup(&opctx, &disk_selector)?
            .lookup_for(authz::Action::Modify)
            .await?;
```

Given you aren't actually using the db object, may as well only do the lookup.

common/src/api/external/mod.rs
```diff
@@ -958,6 +958,14 @@ pub enum DiskState {
     Creating,
     /// Disk is ready but detached from any Instance
     Detached,
+    /// Disk is ready to receive blocks from an external source
+    ImportReady,
```
zephraph, line 962: Why is it important for `ImportReady` to exist over just having `Detached` be able to transition into one of the importing states?

jmpesp: A few reasons:

1. `Detached` and `ImportReady` represent states from which only a subset of operations can be started. `ImportReady` is the state where the Disk is constantly attached to a Pantry and the operations that start from there all involve that Pantry. To transition from `Detached` straight to `ImportingFromBulkWrite` would mean attaching and detaching from a Pantry each time, which means a saga has to be invoked for that state change. It seemed lighter weight to park the disk in `ImportReady` instead exiting the disk_create saga, and to require the user to "finalize" the disk once all importing was done.
2. To "finalize" a disk means to transition out of `ImportReady` to `Detached`, and there's no way for that disk to again enter `ImportReady`. Once it's detached, it's ready for regular use with instances. Though, this supposes a world where a user will never again want to import blocks into a disk after they've done it the first time.

nexus/src/app/disk.rs
jmpesp: right, changed in 628ff59c

jmpesp:
@leftwo I'd love a second set of eyes on the `volume_checkout_randomize_ids` function, and whether or not I've done things correctly there :)

nexus/src/app/sagas/snapshot_create.rs
```diff
@@ -925,48 +886,28 @@ async fn ssc_call_pantry_attach_for_disk(
...
+        call_pantry_attach_for_disk(
+            &log,
+            &opctx,
+            &osagactx.nexus(),
+            params.disk_id,
+            pantry_address,
+        )
+        .await?;
+    } else {
+        info!(log, "disk {} already attached to a pantry", params.disk_id);
```
leftwo, line 909: When would we expect a disk already attached to a pantry?

nexus/src/db/datastore/volume.rs
```diff
@@ -335,6 +335,99 @@ impl DataStore {
...
+                })
+            }
+        }
+    }
+
+    /// Checkout a copy of the Volume from the database using `volume_checkout`,
+    /// then randomize the UUIDs in the construction request. Because this is a
+    /// new volume, it is immediately passed to `volume_create` so that the
+    /// accounting for Crucible resources stays correct.
+    pub async fn volume_checkout_randomize_ids(
```
leftwo, line 410: Excuse my ignorance, but why do we need new/different volume IDs?

nexus/tests/integration_tests/pantry.rs
```diff
@@ -0,0 +1,821 @@
...
+
+    // Validate disk is in state detached
+    validate_disk_state(client, DiskState::Detached).await;
+
+    // Create an instance to attach the disk.
+    create_instance_and_attach_disk(client, nexus, StatusCode::ACCEPTED).await;
+
+    // Validate that a user cannot start a bulk write
+    bulk_write_start(client, StatusCode::BAD_REQUEST).await;
+}
```
leftwo, line 792: Throw a `bulk_write_bytes` and `bulk_write_stop` in here and cover those cases as well.

leftwo (commented):
Nice work.

Some thoughts:

How long will a disk be ImportReady? forever?
I assume we can delete a disk if an import operation starts, but never finishes or has
an import error?  I don't think I saw a test for that path (giving up on import at the various stages).

We should add the disk import states to the import RFD.  I think I know how they go but it would be nice to formalize it.

common/src/api/external/mod.rs
```diff
@@ -984,6 +992,12 @@ impl TryFrom<(&str, Option<Uuid>)> for DiskState {
         match (s, maybe_id) {
             ("creating", None) => Ok(DiskState::Creating),
             ("detached", None) => Ok(DiskState::Detached),
+            ("import_ready", None) => Ok(DiskState::ImportReady),
+            ("importing_from_url", None) => Ok(DiskState::ImportingFromUrl),
+            ("importing_from_bulk_writes", None) => {
+                Ok(DiskState::ImportingFromBulkWrites)
+            }
+            ("finalizing", None) => Ok(DiskState::Finalizing),
```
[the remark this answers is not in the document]
jmpesp: Yes

nexus/src/app/disk.rs
```diff
@@ -408,4 +352,277 @@ impl super::Nexus {
...
+            // - if the import landed first, then the bulk write would overwrite
+            //   what the import job imported, probably corrupting the data.
+            //
+            // I also believe it's not correct to use a saga here. Really, the
+            // user's cli (or any other program really) is responsible for this
+            // bulk write operation, not Nexus. If the bulk_write call below
+            // fails, then that failure would be propagated up to the user, and
+            // that user's program can act accordingly. In a way, the user's
+            // program is an externally driven saga instead.
+
```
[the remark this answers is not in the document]
jmpesp: That's right, yeah. https://github.com/oxidecomputer/crucible/pull/654 will add a validate command to the Pantry, allowing the user to checksum the whole disk and compare it against their known good one. Once that lands I'll update Nexus' API so that a checksum can be passed during the finalize step.

nexus/src/app/disk.rs
```diff
@@ -408,4 +352,277 @@ impl super::Nexus {
...
+                db_disk.id(),
+            )))
+        }
+    }
+
+    /// Move a disk from the "ImportingFromBulkWrites" state to the
+    /// "ImportReady" state, usually signalling the end of manually importing
+    /// blocks.
+    pub async fn disk_manual_import_stop(
+        self: &Arc<Self>,
```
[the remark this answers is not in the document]
jmpesp: Right - this slightly protects the user from a race where one thread is still bulk writing in, and another calls finalize.

Imagine two threads, one performing bulk writes, and another erroneously calling finalize. If it's possible to transition from `ImportingFromBulkWrites` to `Finalizing`, then thread 1 will error, but thread 2 will succeed: it's interrupting what thread 1 is doing but that state transition is allowed. The disk will be successfully finalized but will fail the checksum validation (only if one is supplied!).

With an explicit transition back to `ImportReady` required, thread 2 will fail to finalize the disk unless it calls `bulk-write-stop` first.

Note if thread 1 is performing bulk writes and thread 2 erroneously calls `bulk-write-stop` then `finalize`, it'll still interrupting what thread 1 is doing and will successfully finalize (still failing the checksum validation if it's supplied).

In the case of importing from a url, a saga is changing the disk's state. It won't be possible to finalize the disk until the pantry job is complete because the saga will set the disk's state back to import ready when the job is done. If we could transition from `ImportingFromUrl` to `Finalizing`, then users could cut the saga off and/or cause the saga to unwind. In this case, the fact that saga transitioned from `ImportReady` to `ImportingFromUrl` means that it's "locked" the disk and has exclusive use of it because all operations start only from state `ImportReady`, including finalizing.

nexus/src/app/sagas/finalize_disk.rs
```diff
@@ -0,0 +1,383 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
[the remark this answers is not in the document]
jmpesp: No idea, though from a quick `ripgrep` of the nexus directory, only nexus/db-macros/src/lib.rs and nexus/types/src/identity.rs have copyright notices at the top!

nexus/src/app/sagas/snapshot_create.rs
jmpesp: In the case that the `snapshot_create` saga is a sub-saga to the `finalize` saga.

nexus/src/db/datastore/volume.rs
jmpesp: This is called when creating an image out of a snapshot. Without an extra volume, the accounting for the underlying crucible resources associated with that snapshot won't be incremented, and if the user deletes that snapshot the underlying crucible resources will be freed. With this new volume, the user can delete the snapshot and still make disks out of this image.

nexus/src/app/sagas/import_blocks_from_url.rs
```diff
@@ -0,0 +1,365 @@
...
+    let pantry_address = sagactx.lookup::<SocketAddrV6>("pantry_address")?;
+    let job_id =
+        sagactx.lookup::<String>("call_pantry_import_from_url_for_disk")?;
+
+    let endpoint = format!("http://{}", pantry_address);
+
+    let client = crucible_pantry_client::Client::new(&endpoint);
+
+    while !client.is_job_finished(&job_id).await.unwrap().job_is_finished {
+        info!(
```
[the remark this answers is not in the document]
jmpesp: done in 19a4e605

nexus/tests/integration_tests/pantry.rs
jmpesp: right, done in 107bb6fe

jmpesp:
> How long will a disk be ImportReady? forever?

This is up to the user's program that's driving the import.

> I assume we can delete a disk if an import operation starts, but never finishes or has
an import error? I don't think I saw a test for that path (giving up on import at the various stages).

Currently we can only delete a disk from the `Detached` state: if the user gives up on an import, it's possible to bulk-write-stop, then finalize, then delete the disk. It's a bit verbose.

leftwo:
> Currently we can only delete a disk from the Detached state: if the user gives up on an import, it's possible to bulk-write-stop, then finalize, then delete the disk. It's a bit verbose.

Can the web users delete a disk that is in this state?  That would be my concern that some CLI user started then abandoned the import process, and a UI user wants to clean up this disk and them not being able to.

jmpesp:
> > Currently we can only delete a disk from the Detached state: if the user gives up on an import, it's possible to bulk-write-stop, then finalize, then delete the disk. It's a bit verbose.
> 
> Can the web users delete a disk that is in this state? That would be my concern that some CLI user started then abandoned the import process, and a UI user wants to clean up this disk and them not being able to.

I don't see why not if the console is taught about the new disk states and the import related endpoints.

leftwo:
> 

Ah, got it.   I thought the work was in this repo, not the actual console repo.

zephraph:
I'm not sure I completely follow but my view is that the logic for deletion should likely be encapsulated in the delete endpoint. If a user has chosen to delete a disk Nexus should be able to take that disk from whatever given state it's in to deleted without requiring extraneous user actions.

jmpesp:
> I'm not sure I completely follow but my view is that the logic for deletion should likely be encapsulated in the delete endpoint. If a user has chosen to delete a disk Nexus should be able to take that disk from whatever given state it's in to deleted without requiring extraneous user actions.

We [already don't allow deleting a disk in any arbitrary state](https://github.com/oxidecomputer/omicron/blob/597d89f01b8a3517ef633a19e41b87d3b6010310/nexus/src/db/datastore/disk.rs#L514-L518) though. It would be bad news to delete an attached disk, just like it'd be bad news to delete a disk where there's an "import-blocks-from-URL" saga going.

zephraph:
As an FYI, we've implemented project based images for the V1 API but we don't yet have a mechanism for making them shared across the silo. There's a follow up issue here: https://github.com/oxidecomputer/omicron/issues/2521.

nexus/src/app/sagas/disk_create.rs
```diff
@@ -155,16 +155,36 @@ async fn sdc_create_disk_record(
...
-    };
+                db_snapshot.block_size
+            }
+            params::DiskSource::Image { image_id } => {
+                let (.., image) =
+                    LookupPath::new(&opctx, &osagactx.datastore())
+                        .image_id(*image_id)
+                        .fetch()
+                        .await
+                        .map_err(ActionError::action_failed)?;
```
leftwo: The `ImportingBlocks` path does a `map_error(|e| .. e.to_string() ...`. Why not do that here and for GlobalImage path too?

leftwo (approved):
Yeah, I just have that one nit here, nothing of substance.

We do need to update the image RFD with details about how image import works from the API side, so let's be sure that work is done.

nexus/src/app/disk.rs
```diff
@@ -456,4 +400,277 @@ impl super::Nexus {
...
+    /// Import blocks from a URL into a disk
+    pub async fn import_blocks_from_url_for_disk(
+        self: &Arc<Self>,
+        opctx: &OpContext,
+        disk_selector: &params::DiskSelector,
+        params: params::ImportBlocksFromUrl,
+    ) -> UpdateResult<()> {
+        let authz_disk: authz::Disk;
+
+        (_, _, authz_disk) = self
```
zephraph, line 413:
```suggestion
        (.., authz_disk) = self
```

Doesn't seem like you need to explicitly destruct these values if they're not being used.

nexus/src/app/disk.rs
```diff
@@ -456,4 +400,277 @@ impl super::Nexus {
...
+    /// blocking any import from URL jobs.
+    pub async fn disk_manual_import_start(
+        self: &Arc<Self>,
+        opctx: &OpContext,
+        disk_selector: &params::DiskSelector,
+    ) -> UpdateResult<()> {
+        let authz_disk: authz::Disk;
+        let db_disk: db::model::Disk;
+
+        (_, _, authz_disk, db_disk) = self
```
zephraph, line 442:
```suggestion
        (.., authz_disk, db_disk) = self
```

Same as above

nexus/src/app/disk.rs
```diff
@@ -456,4 +400,277 @@ impl super::Nexus {
...
+    /// Bulk write some bytes into a disk that's in state ImportingFromBulkWrites
+    pub async fn disk_manual_import(
+        self: &Arc<Self>,
+        opctx: &OpContext,
+        disk_selector: &params::DiskSelector,
+        param: params::ImportBlocksBulkWrite,
+    ) -> UpdateResult<()> {
+        let db_disk: db::model::Disk;
+
+        (_, _, _, db_disk) = self
```
zephraph, line 481:
```suggestion
        (.., db_disk) = self
```
Same

nexus/src/app/disk.rs
```diff
@@ -456,4 +400,277 @@ impl super::Nexus {
...
+    /// blocks.
+    pub async fn disk_manual_import_stop(
+        self: &Arc<Self>,
+        opctx: &OpContext,
+        disk_selector: &params::DiskSelector,
+    ) -> UpdateResult<()> {
+        let authz_disk: authz::Disk;
+        let db_disk: db::model::Disk;
+
+        (_, _, authz_disk, db_disk) = self
```
zephraph, line 607:
```suggestion
        (.., authz_disk, db_disk) = self
```

nexus/src/app/disk.rs
jmpesp: right, addressed in 06471b9a

nexus/src/app/disk.rs
jmpesp: addressed in 06471b9a

nexus/src/app/disk.rs
jmpesp: addressed in 06471b9a

PR #2419, opened 2023-02-24
Use OPTE port to provide nexus external connectivity.

luqmana:
Allow creating "service" OPTE ports (as opposed to "guest" kinds) to give services external connectivity (just nexus currently).

Still with a couple TODO about hardcoding some values that should probably come from a special project/vpc for services but otherwise seems at parity with the existing vnic approach.

We can see an OPTE port created for the Nexus:
```
$ pfexec opteadm list-ports
LINK                             MAC ADDRESS              IPv4 ADDRESS     EXTERNAL IPv4    IPv6 ADDRESS                             EXTERNAL IPv6                            STATE   
opte_nexus0                      A8:40:25:FF:00:01        10.0.0.5         192.168.1.20     None                                     None                                     running 
```

I can reach the external API from another machine:
```
$ curl --head http://192.168.1.20/
HTTP/1.1 405 Method Not Allowed
content-type: application/json
x-request-id: bcd3e89a-54f6-4bbb-ac44-166b92afe186
content-length: 93
date: Thu, 23 Feb 2023 23:59:26 GMT
```

Can even see some stats for it:
```
$ pfexec opteadm dump-layer -p opte_nexus0 firewall
Layer firewall
Inbound Flows
PROTO  SRC IP           SPORT  DST IP           DPORT  HITS     ACTION                
TCP    192.168.1.2      63049  10.0.0.5         80     0        no-op                 
Unknown 0.0.0.0          0      0.0.0.0          0      0        no-op                 

Outbound Flows
PROTO  SRC IP           SPORT  DST IP           DPORT  HITS     ACTION                
TCP    10.0.0.5         80     192.168.1.2      63049  1        no-op                 
Unknown 0.0.0.0          0      0.0.0.0          0      0        no-op                 

Inbound Rules
ID     PRI    HITS   PREDICATES                             ACTION            
0      100    18     inner.ulp.dst=80,443                   "Stateful Allow"
1      65534  81     inner.ether.ether_type=ARP             "Stateful Allow"
DEF    --     1372   --                                     "deny"

Outbound Rules
ID     PRI    HITS   PREDICATES                             ACTION            
DEF    --     0      --                                     "stateful allow"
```


illumos-utils/src/opte/illumos/port_manager.rs
```diff
@@ -454,31 +351,239 @@ impl PortManager {
...
+                return Err(opte_ioctl::Error::InvalidArgument(String::from(
+                    "IPv6 is not yet supported for external addresses",
+                ))
+                .into());
+            }
+        };
+
+        let snat = SNat4Cfg {
+            external_ip,
+            // TODO: Assumes IP is exlusive to this service.
```
jmpesp, line 536: typo here, plus: isn't this true (it is at least for IPv6)?

sled-agent/src/services.rs
```diff
@@ -479,6 +456,60 @@ impl ServiceManager {
...
+            match svc {
+                ServiceType::Nexus { external_ip, .. } => {
+                    // TODO: Arbitrary MAC choice from the "system" portion of the
+                    // virtual MAC address space.
+                    let mac = [0xA8, 0x40, 0x25, 0xFF, 0x00, 0x01].into();
+                    // TODO: Arbitrary VNI choice from Oxide-reserved range.
+                    let vni = 100;
+                    // Allow incoming web traffic for nexus
+                    // TODO: Should come froma service VPC (along with the VNI).
+                    let firewall_rules = &[serde_json::from_str(
```
jmpesp, line 483: why not use the type here?

illumos-utils/src/opte/illumos/port_manager.rs
```diff
@@ -454,31 +351,239 @@ impl PortManager {
...
+        mac: MacAddr6,
+        external_ip: IpAddr,
+        vni: u32,
+        firewall_rules: &[VpcFirewallRule],
+    ) -> Result<(Port, PortTicket), Error> {
+        // TODO: Assume each service zone has a single port.
+        let port_name = format!("{}_{}0", XDE_LINK_PREFIX, svc_name);
+
+        // TODO: Arbitrary subnet, should come from a "service" VPC.
+        let vpc_subnet: Ipv4Network = "10.0.0.0/24".parse().unwrap();
```
jmpesp, line 510: this could be wrapped up in a `opte::params::NetworkInterface` too

jmpesp: What do you think about this PR defining that service VPC, and creating fake network interfaces in there for services? It would reuse a lot of the existing machinery

jmpesp: I deleted another comment and will shove it here instead: if that work is done, users could query for network interfaces in that service VPC to discover the Nexus IP they need to set.

illumos-utils/src/opte/non_illumos/port_manager.rs
```diff
@@ -137,47 +191,50 @@ impl PortManager {
...
+        mac: MacAddr6,
+        _external_ip: IpAddr,
+        vni: u32,
+        firewall_rules: &[VpcFirewallRule],
+    ) -> Result<(Port, PortTicket), Error> {
+        // TODO: Assume each service zone has a single port.
+        let port_name = format!("{}_{}0", XDE_LINK_PREFIX, svc_name);
+
+        // TODO: Arbitrary subnet, should come from a "service" VPC.
+        let vpc_subnet: Ipv4Network = "10.0.0.0/24".parse().unwrap();
```
jmpesp, line 221: same question here - this could be wrapped up in a NetworkInterface

sled-agent/src/services.rs
```diff
@@ -479,6 +456,60 @@ impl ServiceManager {
...
+        for svc in &req.services {
+            match svc {
+                ServiceType::Nexus { external_ip, .. } => {
+                    // TODO: Arbitrary MAC choice from the "system" portion of the
+                    // virtual MAC address space.
+                    let mac = [0xA8, 0x40, 0x25, 0xFF, 0x00, 0x01].into();
+                    // TODO: Arbitrary VNI choice from Oxide-reserved range.
+                    let vni = 100;
+                    // Allow incoming web traffic for nexus
+                    // TODO: Should come froma service VPC (along with the VNI).
```
jmpesp, line 482: typo, plus see previous comment

jmpesp (changes requested):
Here's a first pass. I'm going to be testing this on my setup shortly.

illumos-utils/src/opte/illumos/port_manager.rs
bnaecker: It looks like this uses the same external IP to create the `SNat4Cfg` and also passes that as the `external_ips` field of the `Ipv4Cfg`. That's "fine" in the sense that OPTE will always use the latter if it exists, ignoring the former. I don't see a difference between IPv4 and IPv6 here.

One question I have is where these addresses come from. Eventually, operators will provide the address lots they want to assign to the rack. We'll then take one of these for Nexus, possibly a specific address. It's not part of this PR, but we will want to store that in the DB and explicitly assign the address to Nexus, along with the entire port range. That would realistically reserve the whole port range, since this must allow incoming connections.

I actually think we should consider _not_ providing an `SNatCfg` at all here. It plays no role, since OPTE will ignore it, and I think it might confuse things more.

illumos-utils/src/opte/illumos/port_manager.rs
```diff
@@ -454,31 +351,239 @@ impl PortManager {
...
+            // NOTE: This value of this flag is irrelevant, since the driver
+            // always overwrites it. The field itself is used in the `oxide-vpc`
+            // code though, to determine how to set up the ARP layer, which is
+            // why it's still here.
+            proxy_arp_enable: true,
+            phys_gw_mac: Some(MacAddr::from(
+                self.inner.gateway_mac.into_array(),
+            )),
+            // TODO-completeness (#2153): Plumb domain search list
+            domain_list: vec![],
```
bnaecker, line 567: Is this relevant or desirable for service ports? I thought the intent of that issue was to provided guest DNS clients a domain list when doing resolution for _internal_ DNS names, e.g., of other instances. I don't know that we'd want the same thing for service ports, nor am I really sure what data we would provide here.

davepacheco:
Cool!  I'd just been thinking of this in the context of RFD 357, and in particular [this section](https://github.com/oxidecomputer/rfd/tree/master/rfd/0357#dns-server-and-external-api-networking).  Would we use this mechanism for the external DNS servers as well?

illumos-utils/src/opte/illumos/port_manager.rs
```diff
@@ -73,7 +75,7 @@ struct PortManagerInner {
     // IP address of the hosting sled on the underlay.
     underlay_ip: Ipv6Addr,
 
-    // Map of all ports, keyed on the instance Uuid and the port name.
+    // Map of all ports, keyed on the instance/service Uuid and the port name.
     ports: Mutex<BTreeMap<(Uuid, String), Port>>,
```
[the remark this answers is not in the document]
luqmana: Updated to use id + slot for both. After some reworking decided it's useful to keep "slot" for the service ports as well to support adding multiple to a zone (which we might need to support assigning both an ipv4 and ipv6 ext ip for nexus).

illumos-utils/src/opte/illumos/port_manager.rs
luqmana: Removed the snat config for service ports which exclusively use external IPs right now.

sled-agent/src/services.rs
luqmana: Just was a bit more annoying to fill in when I first wrote it :P Anyways, this bits been removed and moved elsewhere.

illumos-utils/src/opte/illumos/port_manager.rs
luqmana: Moved a lot of these hardcoded values to a `NetworkInterface` as suggested. Also added subsequent commits to create some built-in orgs/proj/vpc for services to stash these values and use a lot of the existing machinery.

One caveat: these service "NIC"s aren't in the normal `network_interface` table as it currently expects a non-NULL FK to the `instance` table. Tried a different approaches to consolidate but those ended up requiring a lot more significant changes which I'd rather do in a follow-up than this PR. So for now there's a new table `service_network_interface` which is basically the same except instead of pointing to the `instance` table it points to the `service` table.

illumos-utils/src/opte/non_illumos/port_manager.rs
luqmana: done

sled-agent/src/services.rs
luqmana: factored out to a fixed `NetworkInterface` structure that gets passed in

illumos-utils/src/opte/illumos/port_manager.rs
luqmana: As-is, probably not relevant? Just left the same comment for guest ports. Anyways, did some deduplication so there's no longer a service-specific TODO here anymore.

luqmana:
> Cool! I'd just been thinking of this in the context of RFD 357, and in particular [this section](https://github.com/oxidecomputer/rfd/tree/master/rfd/0357#dns-server-and-external-api-networking). Would we use this mechanism for the external DNS servers as well?

uh, not sure. i want to say maybe orthogonal but I need to go re-read that and see how it applies.

luqmana:
Thanks for the review, I've put up some more changes in response to some of the comments left and previous TODOs.

To leverage a lot of the existing functionality we have for guest instance ports, I've added a new built-in organization for resources dedicated to services. During rack population we ensure that org exists and at rack initialization, we create for each service:
1. a project
2. a VPC (+ default router/subnet)
3. & for nexus, a network interface record* + firewall rules

With that, a lot of the hardcoded values previously in sled-agent during nexus zone bringup are reflected in the DB.

(*) I did have to create a new table (`service_network_interface`) as the existing `network_interface` table contains a non-NULL FK to the `instance` table. I spent some time trying to consolidate but that ended up requiring a lot more sprawling changes I don't want to include in this (already large-ish) PR. (If only cockroachdb supported inserting into views)

One other question is IPv6 support. Today I believe you can set the nexus external IP to an IPv6 address and things work fine, unfortunately the OPTE setup today won't work in that case. The comments in the code have a TODO for when opte supports IPv6 but I believe that's been the case for a bit now in upstream OPTE. Now that we're on a more recent commit I believe just wiring it up is all that's needed. I'll try to wire that up, test and push up those changes.

Related to the IPv6 stuff, do we plan to allow nexus to listen on multiple external addresses? That could mean IPv4 + IPv6 or some other combination. On the OPTE side of things in theory we can configure it with one or both but there's some various assumptions in the rest of the control plane right now about each `[service_]network_interface` record containing a single IP. If we do need multiple IPs, we can always just use multiple NICs/ports so not a big deal but something to consider.

nexus/db-queries/src/db/datastore/organization.rs
```diff
@@ -49,11 +49,27 @@ impl DataStore {
             .authn
             .silo_required()
             .internal_context("creating an Organization")?;
+        self.organization_create_in_silo(opctx, None, organization, authz_silo)
+            .await
+    }
+
+    /// Create a organization in the given silo
+    pub(super) async fn organization_create_in_silo(
```
jmpesp, line 57: Organization most likely is going away - see RFD 332

nexus/db-model/src/vni.rs
```diff
@@ -20,6 +20,12 @@ use serde::Serialize;
 #[diesel(sql_type = sql_types::Int4)]
 pub struct Vni(pub external::Vni);
 
+impl Vni {
+    pub fn random_system() -> Self {
+        Self(external::Vni::random_system())
```
jmpesp, line 25: Collisions are unlikely but still possible using this - you could generate a random VNI with a CRDB query that also validates that it's not in use

nexus/db-model/src/service_network_interface.rs
```diff
@@ -0,0 +1,22 @@
...
+use db_macros::Resource;
+use uuid::Uuid;
+
+#[derive(Selectable, Queryable, Insertable, Clone, Debug, Resource)]
+#[diesel(table_name = service_network_interface)]
+pub struct ServiceNetworkInterface {
+    #[diesel(embed)]
+    pub identity: ServiceNetworkInterfaceIdentity,
+
+    pub service_id: Uuid,
```
jmpesp, line 16: `interface_id` vs `service_id` is the only difference between this and the instance network interface table, right?

jmpesp (changes requested):

nexus/db-queries/src/db/datastore/organization.rs
luqmana: Right still needed it here but will rebase + remove once those changes land.

nexus/db-model/src/service_network_interface.rs
luqmana: Basically. The instance table also has an `is_primary` column.

nexus/db-model/src/vni.rs
luqmana: This function is just used to seed an initial choice. I did reuse `InsertVpcQuery` & `NextVni` for the DB query.

nexus/db-queries/src/db/datastore/organization.rs
luqmana: removed

luqmana:
Subsumed by #2857 and follow up PRs.

PR #2423, opened 2023-02-24
RFD-322: users list (+ filter by group)

david-crespo:
Needed to make `PaginatedById` take a generic `Selector` param so I could give it the optional group selector.


nexus/src/external_api/http_entrypoints.rs
```diff
@@ -7372,7 +7378,7 @@ async fn user_list(
     let handler = async {
         let opctx = OpContext::for_external_api(&rqctx).await?;
         let users = nexus
-            .silo_users_list_current(&opctx, &pagparams)
+            .silo_users_list_current_by_id(&opctx, &pagparams)
```
david-crespo, line 7381: I could make them both use the same method like you did here if the group filter query param is purely additive.
https://github.com/oxidecomputer/omicron/pull/2420/files#diff-7429b402f97c8a901b88c3599e8505de4afae78092b5f27e4542bc5818125f17R5794-R5798

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -7386,13 +7390,49 @@ async fn user_list(
...
 
+/// List users
+#[endpoint {
+    method = GET,
+    path = "/v1/users",
+    tags = ["silos"],
+}]
+async fn user_list_v1(
+    rqctx: RequestContext<Arc<ServerContext>>,
+    query_params: Query<PaginatedById<params::OptionalGroupSelector>>,
```
zephraph, line 7401: Is the group selector actually optional?

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -7386,13 +7390,49 @@ async fn user_list(
...
+) -> Result<HttpResponseOk<ResultsPage<User>>, HttpError> {
+    let apictx = rqctx.context();
+    let nexus = &apictx.nexus;
+    let query = query_params.into_inner();
+
+    let pag_params = data_page_params_for(&rqctx, &query)?;
+    // let scan_params = ScanById::from_query(&query)?;
```
zephraph, line 7408: I minor thing, but move all this into the handler definition. We're doing that because timing data is only recorded for things inside the handler so we want as much as possible to be grouped there.

nexus/src/external_api/http_entrypoints.rs
david-crespo: The way I pictured it working was: if you don't specify a group, you get all users. If you specify a group you get the users in that group.

nexus/src/app/iam.rs
```diff
@@ -63,15 +63,34 @@ impl super::Nexus {
...
+                    &authz_silo_user_list,
+                    pagparams,
+                    &authz_group,
+                )
+                .await
+        } else {
+            self.db_datastore
+                .silo_users_list(opctx, &authz_silo_user_list, pagparams)
+                .await
+        }
```
david-crespo, line 93: I'm splitting the logic here, but really these could be two Nexus functions. In fact, I'll make that change. I'm remembering the wisdom of Sandi Metz: move logic forks as high in the call stack as you can.

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -7372,7 +7376,44 @@ async fn user_list(
...
+    let apictx = rqctx.context();
+    let handler = async {
+        let nexus = &apictx.nexus;
+        let query = query_params.into_inner();
+        let pagparams = data_page_params_for(&rqctx, &query)?;
+        let opctx = OpContext::for_external_api(&rqctx).await?;
+        let scan_params = ScanById::from_query(&query)?;
+        // TODO: a valid UUID gets parsed here and will 404 if it doesn't exist
+        // (as expected) but a non-UUID string just gets let through as None
+        // (i.e., ignored) instead of 400ing
```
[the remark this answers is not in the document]
zephraph: Hmmm, I'll have to dig into the underlying mechanism here. What we could do is add validation in `ScanById::from_query` that returns an error if the UUID is invalid. The flow here is a bit different though so I want to make sure I'm understanding it better.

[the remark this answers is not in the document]
david-crespo: That’s a good point, we could do it semi-manually here. I was thinking it might need some serde attribute thing, but that would prevent the struct from being used in a place where we don’t want it to error out.

nexus/src/app/iam.rs
zephraph: Not sure about the quote but it does make sense to me to break this down.

nexus/src/external_api/http_entrypoints.rs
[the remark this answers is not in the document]
david-crespo: Interestingly, this fixed it. The nested serde flatten was making it more lenient somehow.
https://github.com/oxidecomputer/omicron/pull/2423/commits/5402185764b0c208e15fb7ee11dd5d64b8052264#diff-aee01ef600c8eb547421397e5d89c41a67dd6e6ca6033676cac47b7909e1e344

common/tests/output/pagination-schema.txt
```diff
@@ -32,7 +32,7 @@ schema for pagination parameters: scan parameters, scan by name only
 schema for pagination parameters: scan parameters, scan by id only
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
-  "title": "ScanById",
+  "title": "ScanById_for_Null",
```
david-crespo, line 35: Is this fine? I don't know what this title is for.

nexus/src/external_api/http_entrypoints.rs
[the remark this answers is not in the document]
zephraph: That is interesting. Once we're done with the V1 migration a next big step is wrapping up the selectors in a macro or something and potentially simplifying them. It won't result in an API change but I definitely feel like there's some improvement to be had there.

nexus/src/db/datastore/silo_user.rs
```diff
@@ -192,20 +192,47 @@ impl DataStore {
...
+    pub async fn silo_group_users_list(
+        &self,
+        opctx: &OpContext,
+        authz_silo_user_list: &authz::SiloUserList,
+        pagparams: &DataPageParams<'_, Uuid>,
+        // TODO: this should be an authz::SiloGroup probably
+        authz_silo_group: &authz::SiloGroup,
+    ) -> ListResultVec<SiloUser> {
+        use db::schema::silo_group_membership as sgm;
+        use db::schema::silo_user as su;
```
zephraph, line 225: Minor thing but could we change `sgm` to `group` and `su` to `user`? I think that might make it a bit easier to read the query below.

zephraph (approved):

PR #2424, opened 2023-02-24
[MGS] [wicketd] Report progress from MGS to wicketd of trampoline image delivery

jgallagher:
During gimlet recovery, booting OS is the ultimate source of truth for how far along the boot process is, but we can't communicate (directly) with it while it's booting. We can get progress indirectly: this PR adds endpoints to MGS that get and clear the most-recently-requested host image message from the gimlet's SP. Assuming the host fetches the trampoline image sequentially, this gives us progress-by-proxy.

We start to flesh out one of the TODOs in wicketd where we need to wait for installinator completion: we now first wait for the first installinator progress message (which is still a placeholder for now, and will come in after #2413), and while we're waiting for that first message, we poll MGS periodically to attempt to record the trampoline boot progress.

### DIFF
--- a/wicketd/src/update_tracker.rs
+++ b/wicketd/src/update_tracker.rs
@@ -18,6 +18,8 @@ use anyhow::Context;
 use buf_list::BufList;
 use bytes::Bytes;
 use futures::TryStream;
+use gateway_client::types::HostPhase2Progress;
+use gateway_client::types::HostPhase2RecoveryImageId;
 use gateway_client::types::HostStartupOptions;
 use gateway_client::types::InstallinatorImageId;
 use gateway_client::types::PowerState;
@@ -102,9 +104,9 @@ struct SpUpdateData {
 struct UploadTrampolinePhase2ToMgsStatus {
     id: ArtifactId,
     // The upload task retries forever until it succeeds, so we don't need to
-    // keep a "tried but failed" variant here; we just need to know if the
-    // upload has completed.
-    complete: bool,
+    // keep a "tried but failed" variant here; we just need to know the ID of
+    // the uploaded image once it's done.
+    uploaded_image_id: Option<HostPhase2RecoveryImageId>,
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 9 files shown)


sunshowers (commented):
This looks great!

sunshowers (approved):

wicketd/src/update_tracker.rs
```diff
@@ -380,31 +382,158 @@ impl UpdateDriver {
...
+        // long as it matches the trampoline image we're expecting the SP to be
+        // pulling, but it's possible we could be seeing stale status from a
+        // previous update attempt with the same image.
+        //
+        // To start, _clear out_ the most recent status that MGS may have
+        // cached, so we don't see any stale progress from a previous update
+        // through this SP. If somehow we've lost the race and our SP is already
+        // actively requesting host blocks, this will discard a real progress
+        // message, but that's fine - in that case we expect to see another real
+        // one imminently. It's possible (but hopefully unlikely?) that the SP
```
andrewjstone, line 443: It would be nice to eliminate this possibility entirely, and during MUP we likely will. I'm not too concerned right now though.

I really appreciate the detail in this comment. It's going to help later when I forget what I just reviewed :)

wicketd/src/update_tracker.rs
```diff
@@ -380,31 +382,158 @@ impl UpdateDriver {
...
+                    offset,
+                    total_size,
+                }) => {
+                    // Does this image ID match the one we uploaded? If so,
+                    // record our current progress; if not, this is probably
+                    // stale data from a past update, and we have no progress
+                    // information.
+                    let progress = if image_id == uploaded_trampoline_phase2_id
+                    {
+                        HostPhase2Progress::Available {
```
andrewjstone, line 493: This should be impractical if not impossible on our network, but would it be useful to only update if the most recent update is a later update? Does MGS already do this check?

jgallagher: I'm not sure what you mean by "only update if the most recent update is a later update" - can you clarify? :)

In general I don't think MGS can do much checking of these progress messages; they're driven by the SP (which itself is driven by the host requesting data, in this case), and it's even possible (albeit unlikely, at least for MUP) that the SP is alternating between MGS instances if they both have the necessary image.

andrewjstone: Oh, sorry, I realize that comment was totally unclear. What I meant was, we could check (probably here in wicketd) if `offset` is larger than the prior saved offset, so that progress doesn't go backwards for the same install.

PR #2425, opened 2023-02-24
[wicket] Colour and style updates

benjaminleonard:
Just a first pass until we get it more populated. Some of the colours we could probably remove — I've added the palette for the TUI — the rest of the palette I left for the rack illustration.

Also using the short form labels for the rack components as it means they are still visible at the smallest size.


andrewjstone (approved):
Thanks @benjaminleonard you mind reader!

andrewjstone:
I'm going to merge this in locally, so I don't have to mess with my local branch due to an unnecessary rebase.

PR #2427, opened 2023-02-24
[wicketd] start listening to installinator progress reports

sunshowers:
This gets wicketd to the point where it can start processing
installinator reports, and use that to figure out when the first
installinator message has been received.

Also use the sender side being closed as an indication that the
installinator has sent out its completion message.

Still to do in a future PR is processing the rest of the messages
received from the progress receiver.

TODO in this PR:

* [x] Unit tests


wicketd/src/update_tracker.rs
```diff
@@ -392,22 +411,27 @@ impl UpdateDriver {
...
+                }
+            })?;
 
-        // TODO wait for installinator completion, ideally with progress along
-        // the way.
+        while let Some(report) = ipr_receiver.recv().await {
+            // TODO: process progress reports, not just completion.
+        }
+
+        // The receiver being closed means that the installinator has completed.
```
sunshowers, line 434: This feels so serene compared to everything going on in the background...

wicketd/src/artifacts.rs
```diff
@@ -39,6 +41,73 @@ struct ArtifactsWithPlan {
     plan: Option<UpdatePlan>,
 }
 
+/// The artifact server interface for wicketd.
+#[derive(Debug)]
+pub(crate) struct WicketdArtifactServer {
+    log: Logger,
+    store: WicketdArtifactStore,
+    ipr_artifact: IprArtifactServer,
```
sunshowers, line 49: I didn't want to make this `Clone` unnecessarily, so I split up the server and store into separate components.

sunshowers:
I started to write this using channels/message passing rather than mutexes, because I figured I'd need to use a channel to send the data across anyway. On second thought, this might be a bit simpler if I just used an Arc<Mutex<_>> around the RunningUpdate hash map while keeping the rest of the code as is (including the state transitions within the hash map). The code as it stands does work fine though.

I'll let y'all decide whether it's worth trying to rewrite this right now.

sunshowers:
Redid this with `Arc<Mutex<_>>`; I like this a lot more.

wicketd/src/update_tracker.rs
```diff
@@ -210,24 +224,25 @@ impl UpdateTracker {
                 )),
             };
 
-            let task = tokio::spawn(update_driver.run(plan));
+            let task =
+                tokio::spawn(update_driver.run(plan, ipr_start_receiver));
 
-            SpUpdateData { task, update_log }
+            Ok::<_, StartUpdateError>(SpUpdateData { task, update_log })
```
jgallagher, line 230: Why is this closure returning a result now? It doesn't look like the addition of `ipr_start_receiver` adds any fallible calls - maybe it did at one point?

wicketd/src/installinator_progress.rs
```diff
@@ -0,0 +1,277 @@
...
+        let mut running_updates = self.running_updates.lock().await;
+        if let Some(update) = running_updates.get_mut(&update_id) {
+            slog::debug!(
+                self.log,
+                "progress report seen ({} completion events, {} progress events)",
+                report.completion_events.len(),
+                report.progress_events.len();
+                "update_id" => %update_id
+            );
+            match update.take() {
```
jgallagher, line 60: Might be worth a brief comment that `.take()` leaves `update` in the `Invalid` state, and each arm _must_ set `update` back to a valid state. This is my least favorite part of enum type states; I usually end up with `Option<State>` (which is basically equivalent) and it's always unsatisfying.

wicketd/src/installinator_progress.rs
```diff
@@ -0,0 +1,277 @@
...
+        } else {
+            Self::ReportsReceived(sender)
+        }
+    }
+
+    fn is_completed(report: &ProgressReport) -> bool {
+        report
+            .completion_events
+            .iter()
+            // Scan in reverse order since the completion message is very likely
```
jgallagher, line 174: Do we need to scan, or would it be sufficient to just check the last event?

jgallagher (commented):
Looks great! Just a couple minor comments.

wicketd/src/update_tracker.rs
sunshowers: Ah yeah, it used to -- will fix.

wicketd/src/installinator_progress.rs
sunshowers: Done

wicketd/src/installinator_progress.rs
sunshowers: Done, I don't think there's any legitimate reason it wouldn't be the last event.

wicketd/src/installinator_progress.rs
```diff
@@ -0,0 +1,277 @@
...
+        }
+    }
+}
+
+/// The update tracker's interface to the progress store.
+#[derive(Debug)]
+#[must_use]
+pub(crate) struct IprUpdateTracker {
+    log: slog::Logger,
+    running_updates: Arc<Mutex<HashMap<Uuid, RunningUpdate>>>,
```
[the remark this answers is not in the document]
sunshowers: > Do we ever remove entries from this map?

No, not with the current design -- we move to the `RunningUpdate::Closed` status. Is there a reason removing entries would be beneficial? I figured having the idempotency is valuable.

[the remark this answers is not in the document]
jgallagher: An insert-but-never-remove map in a long-running service leaves us growing in memory indefinitely (albeit extremely slowly in this case). Maybe just a comment that we should eventually figure out a path to garbage collect completed updates out of this map somehow? I'm not worried about it in the immediate term since updating will almost always also involve updating the scrimlets, which will result in a wicketd restart.

jgallagher (approved):

andrewjstone (approved):
Ship it!

wicketd/src/installinator_progress.rs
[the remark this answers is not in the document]
sunshowers: Added a comment.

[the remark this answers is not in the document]
jgallagher: Thanks!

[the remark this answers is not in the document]
sunshowers: I think there are probably other cases like crashed installs where we'd need to clean up entries anyway so if this ever becomes an issue (really doubt it) I think we'd just want a time-based garbage collector I think.

PR #2428, opened 2023-02-24
[nexus] Update schema for Name.

luqmana:
The code actually accepts single character names but the schema regex didn't. Also set a minimum length.

progenitor based clients weren't too happy with my 1-char names :P


david-crespo (approved):
seems good to me

david-crespo:
Would be nice to have a test somewhere with a few name examples. There probably already is one.

PR #2440, opened 2023-02-27
Redesign of Wicket TUI

andrewjstone:
This is a very large change to the look of wicket and its interaction model. The redesign of the look is due to @benjaminleonard, although it doesn't quite fit his figma renderings exactly :) In order to simplify things, mouse support was removed, and the rendering model was changed to allow using stateful widgets in the tui.rs model, where mutating state during drawing is allowed. 

The biggest change here is the fact that each `Control` no longer maintains its location to be drawn on the screen in the form of a `Rect`. This removes the necessity to handle resize events from the terminal. This is in line with the tui.rs model, but eliminates the ability to properly handle mouse events, as we cannot do rectangle intersection on user events. It also prevents what I believe could be somewhat simplified, and faster renderings via minimizing the amount of computation done to that strictly necessary during rendering. For example, if a user scrolls content in a `Paragraph` widget, if we know they are at the bottom of the content already we don't have to redraw. With the current model we have to redraw, because we only know the size of the screen during the draw itself.

Oddly enough, as a result of working in these constraints and writing a lot of this code for the second time, the way events are passed from `Control` to inner controls and the structure of all that is much simpler and easier to understand. I believe we could actually add the rect/resize handling and mouse handling back with a fraction of the code that has been deleted here. So all in all this change is a win-win, despite my complaints about the tui.rs interaction model.

I don't think any of this is really controversial, and I"m also not too concerned about it being looked over with a fine tooth comb. I'd vastly prefer you run and play with it to get an idea how it works. Happy to chat tomorrow!


andrewjstone:
The screen recording below shows wicket talking to wicketd which is talking to MGS which talks to two sp-sims. All of this is running on my local helios box.
All this can be run in 4 terminals from the root of the `omicron` repo

```bash
# start the sp-sims from omicron
cargo run -p sp-sim sp-sim/examples/config.toml

# start MGS
cargo run --bin mgs -- run --id 14c8ad58-c55a-430c-9db5-5db2bfc3780e --address '[::1]:9999' gateway/examples/config.toml

# start wicketd
cargo run --bin wicketd run wicketd/examples/config.toml --address '[::1]:12226' --artifact-address '[::1]:12227' --mgs-address '[::1]:9999'

# start wicket
cargo run -p wicket
```

https://user-images.githubusercontent.com/466937/221504946-a6496713-f061-492c-9e0c-230f5f0aa993.mov

wicket/src/state/mod.rs
```diff
@@ -0,0 +1,41 @@
...
+mod rack;
+mod status;
+
+pub use inventory::{
+    Component, ComponentId, Inventory, PowerState, Sp, ALL_COMPONENT_IDS,
+};
+pub use rack::{KnightRiderMode, RackState};
+pub use status::{ComputedLiveness, LivenessState, ServiceStatus};
+
+/// The data state of the Wizard
```
andrewjstone, line 17: Fix this outdated comment.

wicket/src/state/rack.rs
```diff
@@ -0,0 +1,164 @@
...
+            ComponentId::Sled(i) => ComponentId::Sled((i + 2) % 32),
+            ComponentId::Switch(0) => ComponentId::Psc(0),
+            ComponentId::Switch(1) => {
+                if self.left_column {
+                    ComponentId::Sled(16)
+                } else {
+                    ComponentId::Sled(17)
+                }
+            }
+            ComponentId::Switch(_) => unreachable!(),
```
andrewjstone, line 76: Should really just have one unreachable at the bottom of the match.

wicket/src/ui/main.rs
```diff
@@ -0,0 +1,292 @@
...
+        let mgs_spans = state.service_status.mgs_liveness.compute().to_spans();
+        let mut spans = vec![Span::styled("WICKETD: ", style::service())];
+        spans.extend_from_slice(&wicketd_spans);
+        spans.push(Span::styled(" | ", style::divider()));
+        spans.push(Span::styled("MGS: ", style::service()));
+        spans.extend_from_slice(&mgs_spans);
+        let main = Paragraph::new(Spans::from(spans));
+        frame.render_widget(main, rect);
+
+        let test = Paragraph::new(Spans::from(vec![
```
andrewjstone, line 201: I'm not sure this makes sense for wicket during mupdate. Maybe we can put the currently operated version of the TUF repo there?

wicket/src/ui/panes/overview_pane.rs
```diff
@@ -0,0 +1,325 @@
...
+                    let offset =
+                        self.scroll_offsets.get_mut(&component_id).unwrap();
+                    *offset = offset.saturating_sub(1);
+                    Some(Action::Redraw)
+                }
+                _ => None,
+            },
+            Event::Tick => {
+                // TODO: This only animates when the pane is active. Should we move the
+                // tick into the wizard instead?
```
andrewjstone, line 314: s/wizard/runner

wicket/src/ui/widgets/mod.rs
```diff
@@ -0,0 +1,17 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! Custom tui widgets
+
+mod animated_logo;
+
+#[allow(unused)]
+mod banner;
```
andrewjstone, line 10: Can probably just go ahead and delete this.

jgallagher:
A few comments from running it (haven't looked at the code yet, and only plan to give it a quick skim based on your suggestion):

* A couple screens that used to work now look like placeholder (update, inventory) - expected?
* If I'm in the left pane, I can press Enter or RightArrow to "pick" a screen, but to go back I have to press Esc. Esc seems like a reasonable pair with Enter, but since I can RightArrow over I expected to be able to LeftArrow back and I can't (especially irksome on the rack screen where LeftArrow now does sled navigation). Maybe remove the RightArrow-to-select option from the left pane entirely?
* I like the `<ESC>` header on the left pane that tells you how to get back to it - could we add a `<TAB>` to the top pane of the overview to indicate how to switch between tabs?

andrewjstone:
> A few comments from running it (haven't looked at the code yet, and only plan to give it a quick skim based on your suggestion):
> 
> * A couple screens that used to work now look like placeholder (update, inventory) - expected?

Inventory should work as shown in the linked video. As discussed offline you probably were just not on sled0/1.

> * If I'm in the left pane, I can press Enter or RightArrow to "pick" a screen, but to go back I have to press Esc. Esc seems like a reasonable pair with Enter, but since I can RightArrow over I expected to be able to LeftArrow back and I can't (especially irksome on the rack screen where LeftArrow now does sled navigation). Maybe remove the RightArrow-to-select option from the left pane entirely?
This makes sense. I'll do this. I'm also not sold on the key combinations currently anyway, may play around a bit.

> * I like the `<ESC>` header on the left pane that tells you how to get back to it - could we add a `<TAB>` to the top pane of the overview to indicate how to switch between tabs?

Great idea!

wicket/src/runner.rs
```diff
@@ -0,0 +1,253 @@
...
+                    let action = self.screen.on(
+                        &mut self.state,
+                        Event::Term(TermEvent::Key(key_event)),
+                    );
+                    self.handle_action(action)?;
+                }
+                Event::Term(TermEvent::Resize(_, _)) => {
+                    self.screen.draw(&self.state, &mut self.terminal)?;
+                }
+                Event::Inventory(event) => {
```
jgallagher, line 132: Tiny style nit - this match arm is arguably long enough (and has enough rightward-drift) it warrants its own method.

wicket/src/runner.rs
```diff
@@ -0,0 +1,253 @@
...
+/// Listen for terminal related events
+async fn run_event_listener(log: slog::Logger, events_tx: Sender<Event>) {
+    info!(log, "Starting event listener");
+    tokio::spawn(async move {
+        let mut events = EventStream::new();
+        let mut ticker = interval(Duration::from_millis(30));
+        loop {
+            tokio::select! {
+                _ = ticker.tick() => {
+                        if events_tx.send(Event::Tick).is_err() {
```
jgallagher, line 224: Indentation is off here and in the arm below; I think rustfmt gives up on `select!` bodies some (or all?) of the time.

wicket/src/state/rack.rs
```diff
@@ -0,0 +1,164 @@
...
+        if self.knight_rider_mode.is_none() {
+            self.knight_rider_mode = Some(KnightRiderMode::default());
+        } else {
+            self.knight_rider_mode = None;
+        }
+    }
+
+    pub fn up(&mut self) {
+        self.selected = match self.selected {
+            ComponentId::Sled(i) if i == 14 || i == 15 => {
```
jgallagher, line 64: This might be slightly more readable as
```suggestion
            ComponentId::Sled(14 | 15) => {
```

wicket/src/state/rack.rs
```diff
@@ -0,0 +1,164 @@
...
+            ComponentId::Switch(_) => unreachable!(),
+            ComponentId::Psc(0) => ComponentId::Psc(1),
+            ComponentId::Psc(1) => ComponentId::Switch(1),
+            ComponentId::Psc(_) => unreachable!(),
+        };
+    }
+
+    pub fn down(&mut self) {
+        self.selected = match self.selected {
+            ComponentId::Sled(i) if i == 16 || i == 17 => {
```
jgallagher, line 85:
```suggestion
            ComponentId::Sled(16 | 17) => {
```

wicket/src/state/rack.rs
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    pub fn next(&mut self) {
+        self.selected = match self.selected {
+            ComponentId::Sled(i)
+                if ((0..=14).contains(&i) || (16..=31).contains(&i)) =>
+            {
+                ComponentId::Sled((i + 1) % 32)
+            }
+            ComponentId::Sled(15) => ComponentId::Switch(0),
```
jgallagher, line 126: If you put this one first, I think you could remove the long `if...` on the other `Sled` case entirely. (Same comment on `prev()`)

wicket/src/state/status.rs
```diff
@@ -0,0 +1,141 @@
...
+            self.last_redraw_at = Some(Instant::now());
+            true
+        }
+    }
+}
+
+/// Tracker used by a single instance of liveness.
+#[derive(Debug)]
+pub struct LivenessState {
+    // Some means that the stopwatch hasn't yet been initialized.
```
jgallagher, line 71: Should this be "`None` means ..."?

wicket/src/ui/main.rs
jgallagher: Just double-checking definitions - by "currently operated version of the TUF repo" do you mean "system update version contained in the TUF repo"? I think the repo itself doesn't have a version.

jgallagher (approved):

wicket/src/events.rs
```diff
@@ -0,0 +1,65 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+use crate::state::ComponentId;
+use tokio::time::Instant;
+use wicketd_client::types::RackV1Inventory;
+
+use crossterm::event::Event as TermEvent;
+
+/// An event that will update state in the wizard
```
sunshowers, line 10: are we still calling it a wizard?

wicket/src/events.rs
```diff
@@ -0,0 +1,65 @@
...
+    /// Inventory was received.
+    Inventory {
+        /// This is Some if the inventory changed.
+        changed_inventory: Option<RackV1Inventory>,
+
+        /// The time at which at which information was received from wicketd.
+        wicketd_received: Instant,
+
+        /// The time at which information was received from MGS.
+        mgs_received: libsw::Stopwatch,
```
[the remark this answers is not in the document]
sunshowers: Ah yeah I wrote this code. A stopwatch represents an instant + a duration. For wicketd, we only need to store an instant -- the exact moment at which we got info from wicketd. For MGS, we need to store both:

* an instant: when wicket was informed by wicketd
* a duration: how long ago MGS told wicketd, which gets transmitted over the wire to wicket.

re libsw: I believe we're still on 2.0, but libsw upstream is on 3.0. I worked with the author of the crate to add support for tokio instants, which caused the change from `Stopwatch` to `Sw` (since `TokioStopwatch` was a bit too long).

sunshowers (commented):
Just gave this a cursory glance since it's still a draft, but looks good!

Do you think you can do the file moves and the changes in separate PRs? That would aid in readability and reviewability quite a bit.

wicket/src/ui/main.rs
andrewjstone: yep, that's what I meant.

andrewjstone:
e


> > A few comments from running it (haven't looked at the code yet, and only plan to give it a quick skim based on your suggestion):
> > 
> > * A couple screens that used to work now look like placeholder (update, inventory) - expected?
> 
> Inventory should work as shown in the linked video. As discussed offline you probably were just not on sled0/1.
> 
> > * If I'm in the left pane, I can press Enter or RightArrow to "pick" a screen, but to go back I have to press Esc. Esc seems like a reasonable pair with Enter, but since I can RightArrow over I expected to be able to LeftArrow back and I can't (especially irksome on the rack screen where LeftArrow now does sled navigation). Maybe remove the RightArrow-to-select option from the left pane entirely?
> 
> This makes sense. I'll do this. I'm also not sold on the key combinations currently anyway, may play around a bit.
> 
> > * I like the `<ESC>` header on the left pane that tells you how to get back to it - could we add a `<TAB>` to the top pane of the overview to indicate how to switch between tabs?
> 
> Great idea!

Implemented in [5ba8357](https://github.com/oxidecomputer/omicron/pull/2440/commits/5ba8357c296ca00479a9abfdc9100af666de8473)

Now you can only use enter to get to the pane on the right, and enter to switch between views.

wicket/src/events.rs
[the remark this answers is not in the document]
andrewjstone: Going to leave this as is, as it's just moved code.

wicket/src/state/status.rs
```diff
@@ -0,0 +1,141 @@
...
+impl ServiceStatus {
+    pub fn new() -> Self {
+        Self {
+            // Wicketd is polled every 500ms by wicket. Setting a 1 second
+            // liveness threshold means that under normal operation it should
+            // never flip to delayed.
+            wicketd_liveness: LivenessState::new(Duration::from_secs(1)),
+            // MGS is polled every 10 seconds by wicketd. Set a 11 second
+            // threshold to account for wicket -> wicketd and wicketd -> MGS
+            // delay.
```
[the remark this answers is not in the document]
andrewjstone: Going to hold off on this for now. We can tweak in a follow up.

wicket/src/ui/main.rs
andrewjstone: I'm going to do this in the follow up with update changes.

wicket/src/events.rs
andrewjstone: Fixed in [fd4ea2d](https://github.com/oxidecomputer/omicron/pull/2440/commits/fd4ea2d1c6e5f6a298b987e48eeefb5ee6bbeaf0)

wicket/src/runner.rs
```diff
@@ -0,0 +1,253 @@
...
+        self.start_tokio_runtime();
+        enable_raw_mode()?;
+        execute!(self.terminal.backend_mut(), EnterAlternateScreen,)?;
+        self.mainloop()?;
+        disable_raw_mode()?;
+        execute!(self.terminal.backend_mut(), LeaveAlternateScreen,)?;
+        Ok(())
+    }
+
+    fn mainloop(&mut self) -> anyhow::Result<()> {
```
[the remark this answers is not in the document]
andrewjstone: Fixed in [fd4ea2d](https://github.com/oxidecomputer/omicron/pull/2440/commits/fd4ea2d1c6e5f6a298b987e48eeefb5ee6bbeaf0)

wicket/src/state/status.rs
andrewjstone: I think so. Fixed in [fd4ea2d](https://github.com/oxidecomputer/omicron/pull/2440/commits/fd4ea2d1c6e5f6a298b987e48eeefb5ee6bbeaf0)

wicket/src/state/rack.rs
andrewjstone: Wow, great catch! Fixed in [fd4ea2d](https://github.com/oxidecomputer/omicron/pull/2440/commits/fd4ea2d1c6e5f6a298b987e48eeefb5ee6bbeaf0)

wicket/src/state/rack.rs
andrewjstone: I forgot this syntax worked! Thanks. Fixed in [fd4ea2d](https://github.com/oxidecomputer/omicron/pull/2440/commits/fd4ea2d1c6e5f6a298b987e48eeefb5ee6bbeaf0)

wicket/src/state/rack.rs
andrewjstone: Fixed in [fd4ea2d](https://github.com/oxidecomputer/omicron/pull/2440/commits/fd4ea2d1c6e5f6a298b987e48eeefb5ee6bbeaf0)

wicket/src/runner.rs
andrewjstone: Yeah, rustfmt doesn't work inside macros. Fixed in [fd4ea2d](https://github.com/oxidecomputer/omicron/pull/2440/commits/fd4ea2d1c6e5f6a298b987e48eeefb5ee6bbeaf0)

wicket/src/runner.rs
andrewjstone: Fixed in [fd4ea2d](https://github.com/oxidecomputer/omicron/pull/2440/commits/fd4ea2d1c6e5f6a298b987e48eeefb5ee6bbeaf0)

wicket/src/events.rs
```diff
@@ -0,0 +1,65 @@
...
+pub enum Action {
+    Redraw,
+    Update(ComponentId),
+}
+
+impl Action {
+    /// Should the action result in a redraw
+    ///
+    /// For now, all actions result in a redraw.
+    /// Some downstream operations will not trigger this in the future.
```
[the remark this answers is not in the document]
andrewjstone: Agreed! Fixed in [fd4ea2d](https://github.com/oxidecomputer/omicron/pull/2440/commits/fd4ea2d1c6e5f6a298b987e48eeefb5ee6bbeaf0)

rmustacc:
Looking at the demo reel, one thing I noticed is that this currently has both a pwr0 and pwr1 populated. As we're only selling racks with a single power shelf populated (need to double check location), I'd suggest we assume from a wicket perspective that only one is present and the other is a blank. Otherwise I'd worry about a user asking why isn't this what I expect. We don't really have a reliable way to detect double power shelf presence, which is a bit unfortunate, but that can also be a problem for future us.

wicket/src/events.rs
```diff
@@ -0,0 +1,68 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
andrewjstone, line 1: This code is mostly just copied from the old `wizard.rs` and `status_bar.rs`

wicket/src/runner.rs
```diff
@@ -0,0 +1,254 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
andrewjstone, line 1: This is nearly all code from the old `wizard.rs`. Not necessary to review.

wicket/src/state/mod.rs
```diff
@@ -0,0 +1,42 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
andrewjstone, line 1: This code was just moved out of `wizard.rs` and `StatusBar` was renamed to `ServiceStatus` since it tracks state.

wicket/src/state/rack.rs
```diff
@@ -0,0 +1,150 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
andrewjstone, line 1: This was moved code from `widgets/rack.rs`, but simplified dramatically.

wicket/src/state/status.rs
```diff
@@ -0,0 +1,141 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
andrewjstone, line 1: This code is all just copied from other files. The only thing that changed is the styling in `LivenessState`.

wicket/src/ui/widgets/rack.rs
```diff
@@ -0,0 +1,365 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
andrewjstone, line 1: This is primarily a moved file and doesn't require review. There are two major changes however: 
 * Resizing is no longer an event, and so the resize method of `RackState` is no a function that operates during render.
 *  Knight rider mode was simplified dramatically since we don't have Resize events anymore. This is actually a vast improvement :)

andrewjstone:
> Looking at the demo reel, one thing I noticed is that this currently has both a pwr0 and pwr1 populated. As we're only selling racks with a single power shelf populated (need to double check location), I'd suggest we assume from a wicket perspective that only one is present and the other is a blank. Otherwise I'd worry about a user asking why isn't this what I expect. We don't really have a reliable way to detect double power shelf presence, which is a bit unfortunate, but that can also be a problem for future us.

Thanks @rmustacc. I knew this, but haven't focused on it yet. I'm going to style the rack components that are not present differently in a follow up PR and I can also make this change. If you are saying that there's no way to detect PSC presence, then are we only populating PWR0 or PWR1 on all racks and not some random mixture?

rmustacc:
> > Looking at the demo reel, one thing I noticed is that this currently has both a pwr0 and pwr1 populated. As we're only selling racks with a single power shelf populated (need to double check location), I'd suggest we assume from a wicket perspective that only one is present and the other is a blank. Otherwise I'd worry about a user asking why isn't this what I expect. We don't really have a reliable way to detect double power shelf presence, which is a bit unfortunate, but that can also be a problem for future us.
> 
> Thanks @rmustacc. I knew this, but haven't focused on it yet. I'm going to style the rack components that are not present differently in a follow up PR and I can also make this change. If you are saying that there's no way to detect PSC presence, then are we only populating PWR0 or PWR1 on all racks and not some random mixture?

The plan makes sense. Regarding PSC detection, the challenge here is that we have no way to detect if a power shelf is installed or not. The PSC is installed into the power shelf. A missing PSC could be because there is no shelf or there is no PSC itself. Currently we are only populating `pwr0` for all racks. `pwr1` is a blank in all racks.

PR #2441, opened 2023-02-27
[wicket] More style tweaks

benjaminleonard:
Nothing substantial, just some colour changes and capitalisation.

Using the purple as a item selection colour on the rack since that's what we will use eventually for row selection when we implement update.

One thing I wanted to discuss (can do here or in the main PR) is the UX of what is currently overview and 
inventory. I think we could have what is currently inventory be a child view of the rack view.

You'd have something like this: 
image

Oxide Rack -> Sub component

I know it's an added layer of complexity but it's a really useful pattern that we we'd reuse for software update (and other future panes) too — so you can start an update or view an update in progress — which is a child view of the main update view.

### DIFF
--- a/wicket/src/state/inventory.rs
+++ b/wicket/src/state/inventory.rs
@@ -126,9 +126,9 @@ impl ComponentId {
 impl Display for ComponentId {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
-            ComponentId::Sled(i) => write!(f, "sled {}", i),
-            ComponentId::Switch(i) => write!(f, "switch {}", i),
-            ComponentId::Psc(i) => write!(f, "psc {}", i),
+            ComponentId::Sled(i) => write!(f, "SLED {}", i),
+            ComponentId::Switch(i) => write!(f, "SWITCH {}", i),
+            ComponentId::Psc(i) => write!(f, "PSC {}", i),
         }
     }
 }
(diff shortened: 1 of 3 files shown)


andrewjstone (approved):
Thanks Benji!

andrewjstone:
> > One thing I wanted to discuss (can do here or in the main PR) is the UX of what is currently overview and
> > inventory. I think we could have what is currently inventory be a child view of the rack view.
> 
> I like how this looks, need to think about implications across the UI. I also think it makes sense to do this in that it doesn't really make much sense to show sled inventory when it isn't available. This would go hand in hand with styling those sleds as unavailable in the rack view.

I thought about this a little bit more and it's actually not a big deal to implement. Each `Control` which is what a `Pane` is already has the notion of sub controls. The big change here is that the Main Screen no longer always has a tab bar, which isn't strictly necesssary for all screens anyway. So in essence, we're removing the top-level concept of a `Tab` and just making everything a `Control`. I think I like this better and will work on it tonight before I merge this in.

andrewjstone:
I implemented this change in [5ba8357](https://github.com/oxidecomputer/omicron/pull/2440/commits/5ba8357c296ca00479a9abfdc9100af666de8473)

andrewjstone:
> I implemented this change in [5ba8357](https://github.com/oxidecomputer/omicron/pull/2440/commits/5ba8357c296ca00479a9abfdc9100af666de8473)

Damn, somehow this regressed some of your style changes. Fixing...

andrewjstone:
> > I implemented this change in [5ba8357](https://github.com/oxidecomputer/omicron/pull/2440/commits/5ba8357c296ca00479a9abfdc9100af666de8473)
> 
> Damn, somehow this regressed some of your style changes. Fixing...

Fixed in [d65f8ad](https://github.com/oxidecomputer/omicron/pull/2440/commits/d65f8ad53846d4177ff4b6b47636143c31e4937f)

benjaminleonard:
Fantastic work as always!

PR #2442, opened 2023-02-27
V1 APIs for Silos, Identity Providers, and Users

zephraph:
Part of #2028

Implements the APIs for

- Silos
- Silo users
- Identity providers

I've noted a big departure from the original API around silo users that I've taken here that I'd specifically like feedback on.


nexus/src/external_api/http_entrypoints.rs
```diff
@@ -838,10 +1027,46 @@ async fn silo_policy_update(
 // Silo-specific user endpoints
 
 /// List users in a silo
+#[endpoint {
+    method = GET,
+    path = "/v1/system/users",
+    tags = ["system"],
+}]
```
zephraph, line 1034: cc @davepacheco, @ahl, @jmpesp 

This diverges from the initial implementation as discussed here: https://github.com/oxidecomputer/omicron/pull/1784#discussion_r990515243. 

As per the original implementation this would've been `/v1/system/users/all`. The problem (as I see it) with that approach is that it's divergent from our other endpoint definitions. The stated challenge we were trying to work around is that there are potentially other ways that we may want to look up a user other than ID. While I appreciate that challenge, I think there are other ways we can go about solving it. 

My current suggestion (for looking up a user by `external_id`) is simply to add a filter to the list endpoint. 

```
GET /v1/system/users?external_id={id}`
```

This would yield an array with 0 or 1 results. While not strictly as clean from a result perspective, this approach could be added to other endpoints which already implement ID or name lookups as a subroute of the list endpoint without requiring breaking path changes. 

Bulk lookups were also mentioned. I feel like this could also be handled via filtering the list endpoint. One could envision an array query parameter like `/v1/system/users?ids[]=xxx&ids[]=yyy`.

I acknowledge that there may well be a strong counter argument here that having multiple operating modes in a single endpoint makes that endpoints implementation more complex. While acknowledging that, I'd also argue that we're simply talking about different types of filters here. Generally I think we'll _want_ to give more robust filtering capabilities to users and this'll be something we should work towards adding anyway. 

All that said, feed back is definitely welcome.

nexus/src/app/silo.rs
```diff
@@ -6,30 +6,49 @@
...
+                    LookupPath::new(opctx, &self.db_datastore).silo_id(*id);
+                Ok(silo)
+            }
+            NameOrId::Name(name) => {
+                let silo = LookupPath::new(opctx, &self.db_datastore)
+                    .silo_name(Name::ref_cast(name));
+                Ok(silo)
+            }
+        }
+    }
```
zephraph, line 51: This lookup mechanism differs from the normal lookups in that it doesn't take a selector. Once we've moved over to the V1 API I fully expect we'll simplify the selector implementations anyway so I think it's fine that this one case is slightly different.

david-crespo: I'm doing something similar here:
https://github.com/oxidecomputer/omicron/pull/2455/files#diff-882861bfb859d6d0519f20dfcb9abad82e23d52a6963f921fb144a8d9fc2f273R810

nexus/src/external_api/http_entrypoints.rs
david-crespo: I have the same problem in #2423 with looking up groups by external ID. I agree with Justin that while it's a bit weird, treating it as a list filter is the best way.

david-crespo:
Some endpoints are missing `deprecated = true`:

- `silo_list`
- `silo_create`
- `silo_policy_update`
- `saml_identity_provider_create`
- `local_idp_user_create`
- `silo_user_view`
- `system_policy_update`

This list is comprehensive, I went through them all (last one is not even from this PR).

nexus/src/external_api/http_entrypoints.rs
davepacheco: I think I initially went with separate endpoints out of my general fear of scalably implementing arbitrary filtering.  More specifically, if we choose _not_ to implement arbitrary filtering, and instead implement only a handful of filters that are important (e.g., users in one group, user for an external id, all users, users for a set of ids), will the resulting endpoint restrict users in ways that they'll find surprising and annoying (like: "I can take a set of user ids or a group id, but not both")?  I'm still worried about arbitrary filtering but I don't have a strong feeling either way on which makes a cleaner API.

Tangentially: I think we have generally avoided structured query parameters (i.e., where the backend synthesizes more complicated structures from the key-value pairs), but maybe that's just because we haven't had enough need for them.  The main downside (to me) is that I don't think there's really a standard for it so it winds up being API-specific instead of predictable or conventional.  But we're probably going to need them at some point so we may as well come up with a standard.  In this case that could be `ids[]=a&ids[]=b` (as you suggested) or `id=a&id=b` or `id='%5Ba%2C%20b%5D` (filter field is url-encoded JSON) or even `filter='%22%7B%5C%22id%5C%22%3A%5B%5C%22a%5C%22%2C%5C%22b%5C%22%5D%7D%22` (filter is url-encoded JSON).

I think I wouldn't worry too much about either of these here.

nexus/src/app/silo.rs
```diff
@@ -640,16 +621,46 @@ impl super::Nexus {
...
+                let saml_provider = self
+                    .silo_lookup(opctx, &silo_selector.silo)?
+                    .saml_identity_provider_name(Name::ref_cast(name));
+                Ok(saml_provider)
+            }
+            params::SamlIdentityProviderSelector {
+                saml_identity_provider: NameOrId::Id(_),
+                silo_selector: Some(_),
+            } => Err(Error::invalid_request("when providing provider as an ID, silo should not be specified")),
+            _ => Err(Error::invalid_request("provider should either be a UUID or silo should be specified"))
```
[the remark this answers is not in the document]
zephraph: I'm 👍 to this, but the message is consistent between all the other lookup functions. Maybe we should make this change across all of them at the same time?

Also makes me wonder if the should just be a lookup error that wraps around these cases 🤔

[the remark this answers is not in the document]
david-crespo: Yeah, probably do a followup to change them all.

nexus/src/app/silo.rs
```diff
@@ -661,13 +672,11 @@ impl super::Nexus {
...
-        silo_name: &Name,
+        silo_lookup: &lookup::Silo<'_>,
         params: params::SamlIdentityProviderCreate,
     ) -> CreateResult<db::model::SamlIdentityProvider> {
-        let (authz_silo, db_silo) = LookupPath::new(opctx, &self.db_datastore)
-            .silo_name(silo_name)
-            .fetch()
-            .await?;
+        // TODO-security: This should likely be fetch_for CreateChild on the silo
+        let (authz_silo, db_silo) = silo_lookup.fetch().await?;
```
david-crespo, line 679: wouldn't that auth check get done when you actually do the creating?

zephraph: This is a bit of an imperfect process across the API. We do a lot of the authz calls twice. Once at the entrance of the nexus layer and again at the datastore layer. Sometimes the nexus layer has more authz checks because there are logical permissions that must be validated beyond just what the datastore layer would check. @davepacheco has mentioned potentially caching duplicatively authz queries.

The pattern here is fairly typical though. Do an authz check at the entrance of the nexus function.

nexus/src/app/silo.rs
```diff
@@ -181,13 +170,10 @@ impl super::Nexus {
...
         opctx: &OpContext,
-        silo_name: &Name,
+        silo_lookup: &lookup::Silo<'_>,
         silo_user_id: Uuid,
     ) -> LookupResult<db::model::SiloUser> {
-        let (authz_silo,) = LookupPath::new(opctx, self.datastore())
-            .silo_name(silo_name)
-            .lookup_for(authz::Action::Read)
-            .await?;
+        let (authz_silo,) = silo_lookup.lookup_for(authz::Action::Read).await?;
```
david-crespo, line 176: As discussed in chat with @zephraph, we considered the possibility that this would succeed at lookup up a user by ID as long as the specified silo exists, even if the silo doesn't match the user's ID (because the silo is technically not needed for the lookup if we have a user ID to look up directly). But there is an explicit check that the retrieved user's silo ID matches the one passed in:

https://github.com/oxidecomputer/omicron/blob/7be926a4c4764162f58adaaa16e5565ccecb8b1d/nexus/src/app/silo.rs#L156-L158

david-crespo (approved):
think we got em

PR #2444, opened 2023-02-28
[installinator] send status events for artifact writes

sunshowers:
Also add retries: unlike the fetch operation, it is unlikely that this
will fail, though still possible.


installinator/src/write.rs
```diff
@@ -0,0 +1,173 @@
...
+    let mut written_bytes = 0u64;
+
+    let start = Instant::now();
+
+    while artifact.has_remaining() {
+        // BufWriter shouldn't be necessary because we've downloaded, and are
+        // writing, data in typical buffer-sized chunks. Besides,
+        // tokio::fs::File already uses a buffer internally as of tokio 1.25.0.
+        match file.write(artifact.chunk()).await {
+            Ok(n) => {
```
andrewjstone, line 98: I believe it's possible that `file.write` doesn't write the entire contents of `artifact.chunk()`. We should ensure that the entire chunk is written. 

From https://docs.rs/tokio/latest/tokio/io/trait.AsyncWriteExt.html#method.write

> This function will attempt to write the entire contents of buf, but the entire write may not succeed, or the write may also generate an error.

> It is not considered an error if the entire buffer could not be written to this writer.

We could use `write_all` but that is not cancellation safe. It's unclear if that matters here.

sunshowers: Oh I need to call advance(n), totally forgot haha

sunshowers: I should probably also write a test for this

installinator/src/write.rs
```diff
@@ -0,0 +1,173 @@
...
+            destination,
+            event_sender,
+        )
+        .await
+        {
+            Ok(()) => break,
+            Err(error) => {
+                slog::info!(log, "{error:?}"; "artifact_id" => ?artifact_id);
+                // Give it a short break, then keep trying.
+                tokio::time::sleep(Duration::from_secs(5)).await;
```
jgallagher, line 64: I think this is more of a note for @andrewjstone and me: If installinator is willing to retry forever (which might be reasonable? or maybe we should give up after N attempts?), we should expose a way to cancel a running update from wicket; currently there's no way to cancel or restart an update that's running until it finishes.

jgallagher: > In this case I think I'd suggest a reboot of the host, although I realize that would trigger a pull of at the phase 2 trampoline again, which is quite slow.

This is the (admittedly blunt) path I was thinking. "Cancel the update and start over."

installinator/src/write.rs
sunshowers: Switched this to `file.write_buf` which already does the advance, and wrote some tests for reliability.

installinator/src/write.rs
```diff
@@ -0,0 +1,466 @@
...
+        // merely inspecting the list of operations, but partial-io is a bit
+        // broken:
+        // https://github.com/sunshowers-code/partial-io/issues/32
+        //
+        // For now, always error out on earlier attempts and have one successful
+        // attempt at the end. Revisit this after fixing upstream.
+        let success_strategy = prop::collection::vec(
+            partial_op_strategy(interrupted_would_block_strategy(), 1024),
+            0..16,
+        );
```
sunshowers, line 284: This is pretty annoying, something I want to fix in partial-io upstream after PVT1.

installinator/src/write.rs
```diff
@@ -171,3 +187,140 @@ async fn write_artifact_impl(
...
+                    destination,
+                    written_bytes,
+                    total_bytes,
+                    ..
+                }) => {
+                    assert!(
+                        !seen_completion,
+                        "no more progress events after completion"
+                    );
+                    assert_eq!(attempt, 1);
```
andrewjstone: maybe use prop_assert_eq! so it only displays on shrunken version?

andrewjstone (approved):

andrewjstone:
Thanks for the fixes and tests @sunshowers

installinator/src/write.rs
sunshowers: Sadly if you do that you get:

```
error[E0308]: mismatched types
   --> installinator/src/write.rs:431:9
    |
431 | /         prop_assert_eq!(
432 | |             read_num_bytes,
433 | |             artifact.num_bytes(),
434 | |             "read num_bytes matches"
435 | |         );
    | |         ^
    | |         |
    | |_________expected struct `anyhow::Error`, found enum `TestCaseError`
    |           arguments to this enum variant are incorrect
```

Issue #2422, opened 2023-02-24
[sled agent] Fakes are better than mocks; get rid of mocks


smklein:
Sled agent uses a lot of mocks to represent interactions with underlying resources. It uses a library named `mockall` to help with this interaction. We should use fakes instead.

- [ ] svc interface (sled-agent/src/illumos/svc.rs)
- [ ] dladm interface (sled-agent/src/illumos/dladm.rs)
- [ ] fstyp interface (sled-agent/src/illumos/fstyp.rs)
- [ ] zpool interface (sled-agent/src/illumos/zpool.rs)
- [ ] zfs interface (sled-agent/src/illumos/zfs.rs)
- [ ] some free functions (sled-agent/src/illumos/mod.rs)
- [ ] instances (sled-agent/src/instance.rs)

## Background

When the sled agent is provisioning a zone, it "mocks" access to zones. This enables unit tests to verify "hey, you *would* have created a zone here", but in reality, make no such call.

This is okay-ish for some types of unit tests, where we have a module `foo` which depends on a interface `resource`:

- `foo` can be conditionally compiled with `cfg(test)` to use the `mockResource`
- `foo`'s tests can then use `mockall`'s API to set expectations about calls to the `mockResource`

This is [one such example](https://github.com/oxidecomputer/omicron/blob/e7a10f3e1a1e3e1d6cad6b281feb986485cdffdb/sled-agent/src/instance_manager.rs#L312-L399) of a test written in this style.

## Problems

This testing strategy really breaks down when it gets nested. Suppose we have a module `baz`, which depends on `foo` and `bar`.

- `foo`'s dependency on `mockResource` still needs to be captured, so `baz` *also* needs to set up "expectation" calls in tests
- The same applies to `bar`, and any other modules which contains mocks
- This basically leaks implementation details, and makes tests really difficult to write

This is apparent for modules like the `storage_manager`, where we're interfacing with:
- ZFS provisioning
- Zpool provisioning
- Zone provisioning

and *all* of them would need to be mocked to actually write tests.

## Proposal

We should implement fakes wherever these mocks are being used. This will give us a more flexible interface for testing, and make dependencies on "global-ish resources" much more apparent.

[...]

PR #2451, opened 2023-02-28
Extract `sled-hardware` and `illumos-utils` crates from `sled-agent`

jgallagher:
This PR should make no substantive changes other than changes to visibility, imports, etc., associated with moving:

* `sled_agent::hardware` -> `sled-hardware`
* `sled_agent::illumos` -> `illumos-utils`

The primary motivation of these is to allow `installinator` to also use `sled-hardware` using the same hardware scraping/monitoring logic as sled-agent. `sled_agent::hardware` had dependencies on `sled_agent::illumos`, which led to extracting it too.

A handful of supporting but smaller changes:

* `illumos_utils::running_zone` had a dependency on `sled_agent::opte::Port`, and it didn't seem right to pull all of `opte` out with `illumos_utils`, so I broke this dependency by adding an `OptePort` trait and making `RunningZone` and `InstalledZone` generic over a `Port: OptePort` type.
* I moved `sled_agent::vlan` to `omicron_common::vlan`.
* `sled-agent`'s tests depend on mocks set up in what is now a separate crate (`illumos_utils`). #2422 tracks replacing the mocks altogether; in the meantime, 16c901789a3451aa464f8762c596daf6695f0feb is a workaround that adds a `testing` feature to `illumos_utils` that builds the mocks. `sled-agent` enables that feature when _it_ is being tested, allowing the mocks to exist for its use.


package/Cargo.toml
```diff
@@ -10,6 +10,7 @@ anyhow.workspace = true
 clap.workspace = true
 futures.workspace = true
 hex.workspace = true
+illumos-utils.workspace = true
```
jgallagher, line 13: `omicron-package` is _nearly_ able to remove its dependency on `sled-agent`, but it still uses [`sled_agent::cleanup_networking_resources()`](https://github.com/oxidecomputer/omicron/blob/25f05bd648d42943a84f358a472f31578e88c505/package/src/bin/omicron-package.rs#L12). I think that function (and the handful of functions it calls) could be pulled out pretty easily, but I didn't do it as part of this PR (both because we don't need to for installinator and because it's not clear to me where to put it).

jgallagher:
(switched to draft until I fix the build on illumos, derp)

smklein:
FYI to @luqmana , @internet-diglett  - this change is pretty small, but I'm pretty sure it'll conflict with your in-flight PRs, as a warning

illumos-utils/src/running_zone.rs
```diff
@@ -438,7 +446,10 @@ pub struct InstalledZone {
     link: Option<Link>,
 }
 
-impl InstalledZone {
+impl<Port> InstalledZone<Port>
+where
+    Port: OptePort,
+{
```
smklein, line 452: Are these generic parameters + the `OptePort` trait actually necessary? Previously, a snippet of code in `sled-agent/src/opte/mod.rs` conditionally used either the "Meant for illumos" or "not meant for illumos" variant depending on target platform:

https://github.com/oxidecomputer/omicron/blob/e12441e9fcd8529cc6924d90db9ff963e1b1ae52/sled-agent/src/opte/mod.rs#L5-L13

Did you add these to break the `illumos-utils` -> `sled-agent` dependency? This is making me wonder if we should also move the `opte` module to a separate repo...

Regardless, I don't think this needs to hold up this PR, but I'm curious what you think, and if we should file any follow-up issues.

package/Cargo.toml
smklein: I filed https://github.com/oxidecomputer/omicron/issues/2458to track

smklein (approved):

illumos-utils/src/running_zone.rs
jgallagher: > Did you add these to break the illumos-utils -> sled-agent dependency? This is making me wonder if we should also move the opte module to a separate repo...

Precisely this, yes. I'll take a quick stab at moving opte to a separate repo and see how painful that looks.

jgallagher: Sorry, I skipped over your first question:

> Are these generic parameters + the OptePort trait actually necessary? 

Not really, _except_ to break the dependency. Everywhere in `sled-agent` that uses `InstalledZone` / `RunningZone` uses the same `Port` generic argument (which is either the "illumos" or "not illumos" version of `Port`, depending on `target_os`), so there's still only one monomorphized instance of those types.

smklein: Sounds great, thanks for following up on this!

PR #2455, opened 2023-02-28
`/v1/groups/{group}`

david-crespo:
As mentioned in a comment, groups have an `external_id` instead of a `name` and therefore don't implement `ObjectIdentity`, which makes doing a lookup by external ID a little less cute than lookup by name. `external_id` is unique per silo, so looking up by it with the silo implicit is perfectly possible in theory. _However_, unlike with `Name`, we do not (and probably cannot) enforce that the external ID of a group is _not_ a UUID, which means our usual technique of falling back to name only once the thing fails to parse as a UUID does not work. If an external ID was a UUID, there would be no way to fetch a group by it.

### DIFF
--- a/nexus/src/external_api/http_entrypoints.rs
+++ b/nexus/src/external_api/http_entrypoints.rs
@@ -363,6 +363,7 @@ pub fn external_api() -> NexusApiDescription {
         api.register(silo_user_view)?;
         api.register(group_list)?;
         api.register(group_list_v1)?;
+        api.register(group_view)?;
 
         // Console API operations
         api.register(console_api::login_begin)?;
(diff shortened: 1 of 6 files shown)


nexus/src/external_api/http_entrypoints.rs
```diff
@@ -7562,6 +7563,28 @@ async fn group_list_v1(
     apictx.external_latencies.instrument_dropshot_handler(&rqctx, handler).await
 }
 
+/// Fetch group
+#[endpoint {
+    method = GET,
+    path = "/v1/groups/{group}",
+    tags = ["silos"],
```
zephraph, line 7570: I wonder about the tag here. Isn't anything to be changed in this PR, but should silo scoped resources be listed in the silos tag?

zephraph (approved):
Pretty straight forward. 👍

nexus/src/external_api/http_entrypoints.rs
david-crespo: I agree that it's weird. Perhaps even weirder: it seems to _only_ be silo-scoped resources that get this tag. All the `/system/silos/*` endpoints for working with silos themselves are tagged `system`. We could probably stand to rework all the tags.

```
API operations found with tag "silos"
OPERATION ID                             URL PATH
group_list                               /groups
group_list_v1                            /v1/groups
group_view                               /v1/groups/{group}
policy_update                            /policy
policy_update_v1                         /v1/policy
policy_view                              /policy
policy_view_v1                           /v1/policy
user_list                                /users
user_list_v1                             /v1/user
```

Issue #2454, opened 2023-02-28
sled_agent::updates::test::test_write_artifact_to_filesystem() writes to a global path


jgallagher:
Running the sled-agent tests on atrium after @leftwo has previously done so:

```
failures:

---- updates::test::test_write_artifact_to_filesystem stdout ----
thread 'updates::test::test_write_artifact_to_filesystem' panicked at 'called `Result::unwrap()` on an `Err` value: Io { message: "create temp file", err: Custom { kind: PermissionDenied, error: PathError { path: "/var/tmp/control-plane/.tmpFBTKRJ", err: Os { code: 13, kind: PermissionDenied, message: "Permission denied" } } } }', sled-agent/src/updates.rs:152:58
```

```
$ ls -la /var/tmp/control-plane/
total 11
drwxr-xr-x   2 alan     staff          3 Feb 27 18:00 .
drwxrwxrwt  29 root     sys          129 Feb 28 21:55 ..
-rw-------   1 alan     staff         11 Feb 27 18:00 omicron-test-component
```

davepacheco:
FYI: https://github.com/oxidecomputer/omicron/blob/e12441e9fcd8529cc6924d90db9ff963e1b1ae52/test-utils/src/dev/test_cmds.rs#L92-L116
This uses https://doc.rust-lang.org/std/env/fn.temp_dir.html, which honors things like TMPDIR, which is really useful.  It also puts the test name (well, whatever you give it as the argument) into the file path, along with the pid and a per-process counter (in hopes of making sure the result unique -- although this approach isn't always sufficient either, see #1936, but I think that's pretty uncommon).

smklein:
Ah, normally in cases like these, I use https://docs.rs/tempfile/latest/tempfile/ , which we appear to be using already over omicron. Do you prefer the custom path creation over using tempfile?

davepacheco:
No real preference.  Agreed it's nice to use something that cleans up automatically.  I do think it's pretty important that it honor TMP or TMPDIR environment variable.  I use that a lot.

PR #2456, opened 2023-03-01
[sled-agent] Avoid using /var/tmp in tests

smklein:
Fixes https://github.com/oxidecomputer/omicron/issues/2454


iliana (approved):

smklein:
The tests appear to be failing because this notion of "which path should be used for storing zones" is also baked into Nexus. In 78e11faa5ae72651d369608f52cf09d007544700 , I plumbed more configuration, so this path can be used as a `tempdir` within the Nexus tests too.

iliana (approved):

PR #2459, opened 2023-03-01
[installinator] add some more variants to wire format

sunshowers:
Add variants for:

* formatting the disk failed/succeeded/progress
* some sort of miscellaneous error occurred that isn't described by one
  of the other variants

Even though the installinator isn't formatting disks yet, we need
wicketd to at least try and understand these events.


jgallagher (approved):
LGTM :+1:

PR #2460, opened 2023-03-01
Move `sled_agent::opte` to `illumos_utils::opte`

jgallagher:
This came out of [a comment](https://github.com/oxidecomputer/omicron/pull/2451#discussion_r1122074337) on #2451 (on which this PR is based, and it currently points at that PR's branch). 2451 introduced the `illumos-utils` crate, which had a dependency on `sled_agent::opte::Port`. This would've led to a circular dependency, but we broke it by introducing an `OptePort` trait. This PR breaks that dependency a different way: we move all of `sled_agent::opte` into `illumos_utils`.

Making this move without depending back on `sled_agent` required a couple of changes that are less trivial than moving files / adjusting imports:

* A handful of OPTE-related structs from `sled_agent::params` moved to the new module `illumos_utils::opte::params`. They're `pub`licly re-exported from `sled_agent::params`.
* `opte::PortManager` no longer calls `find_chelsio_links()`; instead, it expects to be given its data link. `sled_agent` now calls `find_chelsio_links()` itself prior to calling `PortManager::new()`.
* Similarly, `initialize_xde_driver()` no longer calls `underlay::find_nics()`; instead, it expects to be given the list of (at least two) `AddrObject`s.


illumos-utils/src/opte/non_illumos/port_manager.rs
```diff
@@ -82,14 +82,10 @@ impl PortManager {
     /// interfaces
     pub fn new(
         log: Logger,
+        data_link: PhysicalLink,
```
smklein, line 85: IMO this is just across-the-board better, for testability too.

smklein (approved):

PR #2462, opened 2023-03-02
[installinator] set and read Content-Length header for artifact fetches

sunshowers:
The main advantage is that the installinator can check, and report, how
far along the download is.

Also do a small rework of how channels work in the installinator:
overall this cleans some of this code up.

This is still a bit hard to test end-to-end until we have a host image
to test against, but I think it's pretty straightforward.

(thanks to @andrewjstone for showing me how to add a Content-Length header!)

### DIFF
--- a/installinator/src/mock_peers.rs
+++ b/installinator/src/mock_peers.rs
@@ -6,7 +6,6 @@ use std::{
     collections::BTreeMap,
     fmt,
     net::{Ipv6Addr, SocketAddrV6},
-    pin::Pin,
     time::Duration,
 };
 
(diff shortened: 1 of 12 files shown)


installinator/src/mock_peers.rs
```diff
@@ -228,18 +230,22 @@ impl PeersImpl for MockPeers {
...
         let peer_data = self
             .get(peer)
             .unwrap_or_else(|| panic!("peer {peer} not found in selection"))
             .clone();
-        Box::pin(async move { peer_data.send_response(sender).await })
+        let artifact_size = peer_data.artifact.len() as u64;
+
+        let (sender, receiver) = mpsc::channel(8);
+        tokio::spawn(async move { peer_data.send_response(sender).await });
+        // TODO: add tests to ensure an invalid artifact size is correctly detected
```
jgallagher, line 247: Just checking - did you want to hit this TODO before merging? (I'm fine either way)

jgallagher (commented):
LGTM! :+1:

installinator/src/mock_peers.rs
sunshowers: No, it's not essential for PVT1 so I want to come back to it after that deadline.

PR #2464, opened 2023-03-02
[wicketd] track installinator progress in update log

sunshowers:
This tracks installinator progress in the update log. Failures that
retry aren't yet tracked but will be implemented in a followup.

Depends on #2462.


wicketd/src/update_tracker.rs
```diff
@@ -346,6 +347,10 @@ impl UpdateDriver {
         self.update_log.lock().unwrap().current = Some(state);
     }
 
+    fn clear_current_update_state(&self) {
+        self.update_log.lock().unwrap().current = None;
```
jgallagher, line 351: Hmm. I think wicket is currently assuming `current = None` means no update is running, but that's not true here since it could observe `current = None` in between installinator completing and the host phase 1 being installed. I think we'd be okay to remove this and the call to it altogether - as soon as the installinator channel is closed, we move into `install_host_phase_1_and_boot` which should set a new state.

wicketd/src/update_tracker.rs
```diff
@@ -919,9 +981,14 @@ impl UpdateDriver {
                 } => {
                     ensure!(id == update_id, "SP processing different update");
                     self.set_current_update_state(
-                        UpdateStateKind::ArtifactUpdateProgress {
-                            bytes_received: bytes_received.into(),
+                        UpdateStateKind::ArtifactWriteProgress {
+                            // We currently don't do any retries, so this is attempt 1.
+                            attempt: 1,
```
jgallagher, line 986: Are we starting attempt counts at 1 or 0? (I would have assumed 0 but can understand 1)

sunshowers: I was starting them at 1. I have a mild preference for 1 since it makes more sense intuitively and we aren't going to be doing modulo operations on them (best argument for starting an index at 0).

jgallagher: Yeah, works for me. Starting at 1 means they can be displayed directly in the UI too.

wicketd/src/update_tracker.rs
sunshowers: Oh interesting, I didn't realize that. My specific concern here is that it's possible there are long-ish periods where the installinator doesn't return a progress update. This would result in step completion being ahead of progress, which feels wonky to me and might show up as an inconsistent state in the wicket UI.

Maybe we should just document this for now.

jgallagher: > Oh interesting, I didn't realize that. My specific concern here is that it's possible there are long-ish periods where the installinator returns completion but not progress update (e.g. if it hasn't reached out to peers). This would result in step completion being ahead of progress, which feels wonky to me and might show up as an inconsistent state in the wicket UI.

I don't think I follow - what does "returns completion but not progress update" mean?

In terms of step completion being ahead of progress, this is what the (admittedly extremely clunky) `push_update_success()` function is for - while the update log lock is held, it both pushes a completion event and sets the new state, so a caller can't observe a completion event associated with a stale progress message. The clunkyness comes from needing to know what the next state is going to be, though; e.g., https://github.com/oxidecomputer/omicron/blob/main/wicketd/src/update_tracker.rs#L395-L400 pushes an "update done" and sets the next state to "I'm going to reset", which is weird because it requires knowledge of both what comes before and what comes after.

sunshowers: > I don't think I follow - what does "returns completion but not progress update" mean?

So let's say that the installinator has just finished downloading the host phase 2 image, and is moving to start downloading the control plane image. In that case, what's going to happen is:

1. The host phase 2 image fetcher sends a completion event to the progress tracker running inside the installinator.
2. As part of receiving the message, the progress tracker will record the completion event and clear progress.
3. Then, when the control plane image fetcher connects to the server, it sends a progress message to the progress tracker, which records the message.

If the installinator sends a report in between steps 2 and 3, it will send an empty progress message list. I haven't measured the delay in practice yet.

Some possible solutions to this are:

1. Measure the delay and don't bother addressing this at all. It's possible the race condition lasts just a few milliseconds and is unlikely to ever be hit in practice. (Though given that reports go out every second, even 10 millseconds a is 1% chance).
2. Change the update tracker data model to handle the case of a cleared progress message.
3. Build in the capability you mentioned in the update tracker: it needs to know what the next step is for every completion step.
4. In the installinator, always set a progress message. In other words, build the capability *in the installinator* to know what the next step is given a completion step.
5. (others that I haven't thought of yet)

I was proposing option 4 in my prior comment.

wicketd/src/update_tracker.rs
sunshowers: Yeah, that's what I was thinking.

wicketd/src/update_tracker.rs
sunshowers: Published an update which doesn't clear progress for now.

wicketd/src/update_events.rs
```diff
@@ -22,6 +27,13 @@ pub struct UpdateState {
 #[derive(Clone, Debug, JsonSchema, Serialize)]
 #[serde(rename_all = "snake_case", tag = "state", content = "data")]
 pub enum UpdateStateKind {
+    // This is a hack to handle cases where a completion has just happened and
+    // there's no progress from the next step yet: see
+    // https://github.com/oxidecomputer/omicron/pull/2464#discussion_r1123308999
+    // for some discussion.
```
jgallagher, line 33: Thanks for working through this. I know where we landed doesn't feel _super_ satisfying but I think it's perfectly reasonable.

jgallagher (approved):

PR #2465, opened 2023-03-02
[Wicket] Add resize handlers to controls

andrewjstone:
This re-adds the capability for individual `Control`s to optionally
receive resize events. Such resize events allow a `Control` to compute
its layout outside of being rendered. This in turn allows the ability
to dynamically size contents based on the current state without having
to do it on every draw. Each `Control` is responsible for ensuring its
children receive resize events with their proper `Rect` computed if
their children require this functionality.

The current code on main does not require this functionality, but it is
needed for Update. Specifically it's needed for dynamic construction  of
[`TreeItem`s](https://docs.rs/tui-tree-widget/0.11.0/tui_tree_widget/struct.TreeItem.html)
based on changing status, since we must know the width of the pane in
order to construct the tree faithfully. We could work around this by
building a fake tree structure of the same shape and then recomputing
the real tree structure on every draw, but this is likely more code
than proper resize events, and is confusing. It also costs us CPU
unnecessarily.

As a side-effect of this change, we will now be able to re-introduce
mouse events if we desire in the future.

### DIFF
--- a/wicket/src/ui/main.rs
+++ b/wicket/src/ui/main.rs
@@ -27,6 +27,9 @@ use tui::widgets::{Block, BorderType, Borders, List, ListItem, Paragraph};
 pub struct MainScreen {
     sidebar: Sidebar,
     panes: BTreeMap<&'static str, Box<dyn Control>>,
+    rect: Rect,
+    sidebar_rect: Rect,
+    pane_rect: Rect,
 }
 
 impl MainScreen {
@@ -41,6 +44,9 @@ impl MainScreen {
         MainScreen {
             sidebar: Sidebar::new(sidebar_keys),
             panes: BTreeMap::from_iter(sidebar_ordered_panes),
+            rect: Rect::default(),
+            sidebar_rect: Rect::default(),
+            pane_rect: Rect::default(),
         }
     }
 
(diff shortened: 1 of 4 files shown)


wicket/src/ui/main.rs
```diff
@@ -113,8 +147,17 @@ impl MainScreen {
                 }
             },
             e => {
+                let current_pane = self.sidebar.selected();
                 if self.sidebar.selected {
```
jgallagher, line 151: Hm, the proximity of `self.sidebar.selected()` (which returns the currently-selected pane?) and `self.sidebar.selected` (which is a bool that means... actually I'm not sure - the sidebar is focused?) is a little confusing here. Longer names for one or both might be warranted.

jgallagher (approved):
LGTM! Only comment is a nitpick on some names that already existed before this PR.

wicket/src/ui/main.rs
andrewjstone: Totally agree. I noticed that last night and would have slapped my forhead, except it was already bruised from banging into a desk :D

andrewjstone: I'll change it in the update PR that I'm working on.

Issue #2215, opened 2023-01-24
The bootstrap agent will also need to expose an API to expose version information of host-level software


smklein:

smklein:
Going to depend on https://github.com/oxidecomputer/omicron/pull/2448 to pull this out

smklein:
I mentioned this in chat with @jgallagher , but I think I need some help here. FYI @sunshowers , @andrewjstone .

Bootstrap agent is currently unwilling to serve requests until the trust quorum has been established:
https://github.com/oxidecomputer/omicron/blob/e12441e9fcd8529cc6924d90db9ff963e1b1ae52/sled-agent/src/bootstrap/server.rs#L173-L210

This means I can't actually expose versioning information before trust quorum - aka, when interacting with Wicket.

My plan is to serve a dropshot server for the bootstrap agent which exposes this versioning information alongside the sockets-based connection, using a different port. Does this seem acceptable to y'all?

andrewjstone:
ooh, this means we can actually get the host OS and control plane version software from wicket pre-mupdate.

I think this is a great idea. Separate HTTP socket on bootstrap network.

PR #2448, opened 2023-02-28
[omicron-package] Add ability to stamp component versions onto packages

smklein:
Fixes https://github.com/oxidecomputer/omicron/issues/2447

Example commands look like the following:

```bash
$ cargo run --release -p omicron-package -- stamp omicron-nexus 1.0.1
Created: out/versioned/omicron-nexus.tar.gz

$ tar -tvf omicron/out/versioned/omicron-nexus.tar.gz
...
-rw-r--r--   0/0        5 Jul 24 01:21 2006 root/opt/oxide/VERSION

$ cargo run --release -p omicron-package -- stamp omicron-sled-agent 1.0.0
Created: out/versioned/omicron-sled-agent.tar

$ tar -tvf omicron/out/versioned/omicron-sled-agent.tar 
...
-rw-r--r--   0/0        5 Jul 24 01:21 2006 VERSION
```

PR #2466, opened 2023-03-02
[bootstrap-agent] Expose HTTP server for component version querying

smklein:
Adds an HTTP server interface to the bootstrap agent (on port 80 of the bootstrap address), which allows for component versioning.

Adds tests that verify the component exported from https://github.com/oxidecomputer/omicron-package/pull/38 and https://github.com/oxidecomputer/omicron/pull/2448 can be parsed and returned as a component.

Fixes https://github.com/oxidecomputer/omicron/issues/2215


sled-agent/src/bootstrap/http_entrypoints.rs
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    let mut api = BootstrapApiDescription::new();
+    if let Err(err) = register_endpoints(&mut api) {
+        panic!("failed to register entrypoints: {}", err);
+    }
+    api
+}
+
+#[endpoint {
```
david-crespo, line 36: I know this is a draft but it would be helpful to mention here the (relatively narrow, as I understand it) set of situations where this endpoint is meant to be called.

smklein: Sure, I can add some notes before this is merged. This endpoint is intended to be used by wicket, as part of the recovery / MUP process, to scan for host software information.

I also plan on exposing a similar endpoint out of the "normal" sled agent interface, which will use the same implementation under-the-hood.

However, I'm pushing this out of the bootstrap agent for now because of the reasons outlined in #2215  - we need to access this interface before trust quorum has been established, which means the sled agent isn't actually listening on an underlay address when we first want wicket to query this interface.

smklein: Doc comment added

sled-agent/src/bootstrap/http_entrypoints.rs
```diff
@@ -0,0 +1,50 @@
...
+    let mut api = BootstrapApiDescription::new();
+    if let Err(err) = register_endpoints(&mut api) {
+        panic!("failed to register entrypoints: {}", err);
+    }
+    api
+}
+
+/// Provides a list of components known to the bootstrap agent.
+///
+/// This API is inteneded to allow early boot services (such as Wicket)
```
david-crespo, line 38:
```suggestion
/// This API is intended to allow early boot services (such as Wicket)
```

smklein: Fixed

andrewjstone (approved):
Looks great. Sorry for the delayed review.

sled-agent/src/bootstrap/agent.rs
```diff
@@ -172,10 +180,8 @@ fn bootstrap_ip(
 fn bootstrap_address(
     link: PhysicalLink,
     interface_id: u64,
-    port: u16,
-) -> Result<SocketAddrV6, dladm::GetMacError> {
-    let ip = bootstrap_ip(link, interface_id)?;
-    Ok(SocketAddrV6::new(ip, port, 0, 0))
+) -> Result<Ipv6Addr, dladm::GetMacError> {
+    bootstrap_ip(link, interface_id)
```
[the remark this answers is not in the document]
smklein: Done

Issue #2458, opened 2023-03-01
omicron-package should not depend on sled-agent


smklein:
The dependency is small, but it forces significant recompilations.

See: https://github.com/oxidecomputer/omicron/pull/2451/files#r1120785171 for some context, we're close to removing this dependency already.

PR #2468, opened 2023-03-02
package no longer depends on sled-agent

smklein:
`package` now depends on `sled-hardware`, which is a smaller interface

- Pulls `cleanup_networking_resources` and friends into `sled-hardware/cleanup`
- Pulls `bootstrap/agent`'s `delete_omicron_vnics` into `sled-hardware/cleanup`
- Pulls `sled-agent/common/underlay` into `sled-hardware/underlay`

Fixes https://github.com/oxidecomputer/omicron/issues/2458


sled-agent/src/instance_manager.rs
```diff
@@ -67,11 +68,8 @@ impl InstanceManager {
         underlay_ip: Ipv6Addr,
         gateway_mac: MacAddr6,
     ) -> InstanceManager {
-        let data_link = crate::common::underlay::find_chelsio_links()
-            .unwrap()
-            .into_iter()
-            .next()
-            .unwrap();
+        let data_link =
+            underlay::find_chelsio_links().unwrap().into_iter().next().unwrap();
```
jgallagher, line 72: I realize this didn't change in this PR (or mine where I moved it too), but looking at it again: I don't think we should be unwrapping here (in either case). On non-gimlets this is fine because `find_chelsio_links` doesn't actually do anything and just returns a fixed list, but on real gimlets `find_chelsio_links` can fail, or it can succeed while returning an empty vec.

If you don't want to fix it as part of this PR I can do it in a followup.

jgallagher (approved):

sled-agent/src/instance_manager.rs
smklein: Done in https://github.com/oxidecomputer/omicron/pull/2469

PR #2470, opened 2023-03-02
RFD-322: v1 instance external IPs endpoint

david-crespo:


nexus/src/external_api/http_entrypoints.rs
```diff
@@ -4606,31 +4607,61 @@ async fn instance_network_interface_update(
...
+    query_params: Query<params::OptionalProjectSelector>,
+    path_params: Path<params::InstancePath>,
 ) -> Result<HttpResponseOk<ResultsPage<views::ExternalIp>>, HttpError> {
     let apictx = rqctx.context();
     let nexus = &apictx.nexus;
     let path = path_params.into_inner();
-    let organization_name = &path.organization_name;
-    let project_name = &path.project_name;
-    let instance_name = &path.instance_name;
+    let query = query_params.into_inner();
```
zephraph, line 4654: This should be dropped down inside the handler.

zephraph (approved):
Awesome, thanks. One minor comment, but otherwise good to merge 👍

PR #2471, opened 2023-03-03
Implement `/v1/images` as per RFD-322; RFD-362

zephraph:
Part of #2028. This implementation is informed by context from RFD-362 (see https://github.com/oxidecomputer/rfd/pull/569#issuecomment-1452859365).

This PR focuses on adding project images. There's an additional notion of being able to _promote_ a project image such that it's globally available to used by other projects. This meets our desire to have images that are shareable across projects without necessitating a unique endpoint like `/v1/global-images` and avoids requiring a duplicative database entry. 

There's more work to do here including some integration test coverage. I wanted to get this up early for feedback on the approach.


common/src/sql/dbinit.sql
```diff
@@ -875,10 +875,12 @@ CREATE TABLE omicron.public.image (
     time_deleted TIMESTAMPTZ,
 
     project_id UUID NOT NULL,
+    is_global BOOL NOT NULL,
```
david-crespo, line 878: Given any boolean column, there is a substantial chance it eventually needs to have more than two values. Therefore in the interest of extensibility it is usually better to make it a two-valued enum instead.

nexus/db-model/src/image.rs
```diff
@@ -43,7 +44,9 @@ impl From<Image> for views::Image {
         Self {
             identity: image.identity(),
             project_id: image.project_id,
+            is_global: image.is_global.into(),
```
david-crespo, line 47: this `.into()` is unnecessary. clippy confirms

nexus/tests/integration_tests/endpoints.rs
```diff
@@ -329,7 +329,9 @@ lazy_static! {
     // Project Images
     pub static ref DEMO_IMAGE_NAME: Name = "demo-image".parse().unwrap();
     pub static ref DEMO_PROJECT_IMAGE_URL: String =
-        format!("{}/{}", *DEMO_PROJECT_URL_IMAGES, *DEMO_IMAGE_NAME);
+        format!("/v1/images/{}?organization={}&project={}", *DEMO_IMAGE_NAME, *DEMO_ORG_NAME, *DEMO_PROJECT_NAME);
+    pub static ref DEMO_PROJECT_IMAGE_PROMOTE_URL: String =
+        format!("/v1/images/{}/promote?organization={}&project={}", *DEMO_IMAGE_NAME, *DEMO_ORG_NAME, *DEMO_PROJECT_NAME);
```
david-crespo, line 334: I like the pattern of using `DEMO_PROJECT_SELECTOR` for these

david-crespo:
You say we don't need a `/v1/global-images`, but how do you fetch global images?

The `os` field feels a little odd to me. Like if it's meant as a free-form label (which it seems to be) then it might make sense to give it a different name. Even if it's not meant to be free-form, if there are no constraints on it, it will surely end up being used that way.

zephraph:
> You say we don't need a /v1/global-images, but how do you fetch global images?

All images are project images. Some project images are available in every project (when `is_global` is `true`). The only mechanism for retrieving images is via `/v1/images`.

Previously we had `/system/global-images` which is a cross silo concept that's going away. 

---

> The os field feels a little odd to me. Like if it's meant as a free-form label (which it seems to be) then it might make sense to give it a different name. Even if it's not meant to be free-form, if there are no constraints on it, it will surely end up being used that way.

See @askfongjojo's [specification](https://github.com/oxidecomputer/rfd/pull/569#issuecomment-1428233263) for details and context. 

Relatedly, RFD-4 initially spec'd this out just as `name` and `version`:

> Images require both a name and a version. This is because there can be multiple images in the family "debian" but various versions of the image. Since, the name debian:sid is named after the version of the image, where sid is the version.

That of course won't work with how we handle names because they're intended to be unique in a collection. We went back and forth on structured vs unstructured. If we made it an enum it'd require a code change anytime we want to support a new OS which seems unfortunate. Freeform means users can do whatever they want which can also be unfortunate.

common/src/sql/dbinit.sql
zephraph: My initial inclination is to disagree given the context and usage but I'll consider it further. 

Enums come at a cost. There's a bit of boilerplate (new db type, new data store type, new view type, conversions between the two previous types) as well as the need to be defensive in their usage to ensure adding a new variant doesn't introduce a breaking change. Not understanding how it would expand could mean that we make the wrong choice in writing the query and end up in a similar awkward situation where we still have to do a migration before we can make an update to how the enum is used.

david-crespo:
I see, you always fetch images for a specific project, and that response will include global images if there are any. That works for now. I hope our error messages make it easy to tell that the selector must determine a project.

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -4096,26 +4172,53 @@ async fn image_create(
     new_image: TypedBody<params::ImageCreate>,
 ) -> Result<HttpResponseCreated<Image>, HttpError> {
     let apictx = rqctx.context();
-    let nexus = &apictx.nexus;
-    let path = path_params.into_inner();
-    let organization_name = &path.organization_name;
-    let project_name = &path.project_name;
-    let new_image_params = &new_image.into_inner();
     let handler = async {
```
david-crespo, line 4175: mark deprecated

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -4137,21 +4240,17 @@ async fn image_view(
     path_params: Path<ImagePathParam>,
 ) -> Result<HttpResponseOk<Image>, HttpError> {
     let apictx = rqctx.context();
```
david-crespo, line 4242: deprecated

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -4168,17 +4267,51 @@ async fn image_view_by_id(
     path_params: Path<ByIdPathParams>,
 ) -> Result<HttpResponseOk<Image>, HttpError> {
     let apictx = rqctx.context();
```
david-crespo, line 4269: deprecated

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -4194,26 +4327,52 @@ async fn image_delete(
     path_params: Path<ImagePathParam>,
```
david-crespo, line 4327: deprecated

david-crespo (changes requested):
missing some deprecations

nexus/src/external_api/http_entrypoints.rs
zephraph: I hate these 😭.

Thanks, I'll update it.

common/src/sql/dbinit.sql
```diff
@@ -890,6 +892,7 @@ CREATE UNIQUE INDEX on omicron.public.image (
 ) WHERE
     time_deleted is NULL;
```
david-crespo, line 893: Do you also need an index on `is_global` for the filter by `project_id=id OR is_global = true`?

david-crespo:
Overall looks great, just needs some integration tests and there's that one question about adding a DB index for `is_global`.

common/src/sql/dbinit.sql
zephraph: Yeah, that's a good point.

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -4087,35 +4140,90 @@ async fn image_list(
...
 #[endpoint {
     method = POST,
-    path = "/organizations/{organization_name}/projects/{project_name}/images",
+    path = "/v1/images",
     tags = ["images"]
 }]
+async fn image_create_v1(
+    rqctx: RequestContext<Arc<ServerContext>>,
+    query_params: Query<params::ProjectSelector>,
+    new_image: TypedBody<params::ImageCreate>,
```
david-crespo, line 4149: Since `ImageCreate` doesn't have an `is_global` param, that means you can't create an image already promoted. Worth considering?

zephraph: It's intentional. Promoting an image is an action that requires different permissions than creating an image in a project. I wanted to avoid conditional permissions based on the presence of a given property.

david-crespo (approved):
Ready to go with the medium-sized caveat that `integration_tests::images::test_unpriv_use_global_image_from_other_project` passes spuriously because I'm just not doing the thing it's supposed to test. We discussed this in chat, but the permissions on the image lookup are working as designed, which is unfortunately not the same as working as _desired_.

The idea is this: an image lives in project 1, and the user in question only has access to project 2. The image is promoted to global, which means it shows up in the list of images for project 2, even for the user that can't access project 1. This works already because the auth check is explicitly done on project 2 only, and in the SQL query we simply also include any images with `is_global: true` regardless of their project.

But when we try to create a disk for an image, if the user doesn't have access to that image directly (which must come through access to the project it belongs to), the [lookup](https://github.com/oxidecomputer/omicron/pull/2471/files#diff-86bbfa40cf138457d51850f2320605a0a9a290f9c294d6cb9c3f427fba77969bR176-R179) for the image fails. We could work around this at a low level by making the permissions on global images more complicated (deviating from the generic pattern, as some resources do) or we can work around it in app code, which is probably easier but much less safe.

I am inclined to merge this PR and put the above into an issue, but I think we're going to talk to some people first.

PR #2472, opened 2023-03-03
RFD-322: v1 subnet nics endpoint

david-crespo:
I considered also changing `/v1/network-interfaces` to `/v1/instances/{instance}/network-interfaces` for consistency. I'm not sure yet what's best. 

You could imagine a single `/v1/network-interfaces` that lets you select by instance or by subnet. That might be confusing to users, since as far as I know it would be the only endpoint that allows filtering by two totally different parents. Perhaps even more importantly, it would be pretty complicated to implement.

I think for now I'll leave things as-is because the actual parent of a network interface is an instance:

https://github.com/oxidecomputer/omicron/blob/b61a56555bdd86865ba79e187ae3ff3d66f1ab3b/nexus/src/db/lookup.rs#L593-L600

so, given that fact, it makes sense for the top-level `/v1/network-interfaces` to use instance as its selector. That's what we do for every other kind of resource: the top-level endpoint always uses that resource's intrinsic parent as the selector. For example `/v1/vpcs` takes a project selector in the query params because VPC's parent is Project.

It may be slightly misleading to have `/v1/vpc-subnets/{subnet}/network-interfaces` because that makes you think NICs hang off a subnet; they don't, and one major indication of that is that you can have two NICs with the same name from different instances come back in the response to the subnet-filtered list (I think). So we probably do have to think more about this.

### DIFF
--- a/nexus/src/external_api/http_entrypoints.rs
+++ b/nexus/src/external_api/http_entrypoints.rs
@@ -229,7 +229,7 @@ pub fn external_api() -> NexusApiDescription {
         api.register(vpc_subnet_create_v1)?;
         api.register(vpc_subnet_delete_v1)?;
         api.register(vpc_subnet_update_v1)?;
-        // api.register(vpc_subnet_list_network_interfaces_v1)?;
+        api.register(vpc_subnet_list_network_interfaces_v1)?;
 
         api.register(instance_network_interface_create)?;
         api.register(instance_network_interface_list)?;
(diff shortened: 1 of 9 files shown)


zephraph:
I don't think we should (or could) move network interfaces under instances. `/v1/instances/{instance_id}/network-interfaces/{interface_id}` is a situation we're trying to avoid. There _are_ other instances slightly similar to this. For instance `/v1/disks` and `/v1/instance/{instance}/disks`. The latter is just a filtered view of the former. This more or less matches that usecase.

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -5640,15 +5640,17 @@ async fn vpc_subnet_update(
...
 #[endpoint {
     method = GET,
     path = "/organizations/{organization_name}/projects/{project_name}/vpcs/{vpc_name}/subnets/{subnet_name}/network-interfaces",
     tags = ["vpcs"],
+    deprecated = true,
 }]
 async fn vpc_subnet_list_network_interfaces(
     rqctx: RequestContext<Arc<ServerContext>>,
-    query_params: Query<PaginatedByName>,
+    query_params: Query<PaginatedByNameOrId>,
```
zephraph, line 5653: This is fine, but you didn't have to change the pagination type for the old endpoint. There are other examples of the pagination style being updated under the hood but keeping the endpoint the same. 

Honestly, it doesn't matter here given we're going to delete this anyway and it's not a breaking change to the existing API.

zephraph (approved):
This looks good! Thanks.

nexus/src/external_api/http_entrypoints.rs
david-crespo: Yeah, I figured this was actually slightly easier.

david-crespo:
See #2476 for a better discussion and plan regarding the issues raised in the description.

PR #2473, opened 2023-03-03
support multiple api addresses for dendrite

Nieuwejaar:
- Let dendrite listen on multiple addresses: i.e., underlay (for nexus) and localhost (or tfportd)
- tfportd no longer needs to be restarted when dendrite is

### DIFF
--- a/sled-agent/src/services.rs
+++ b/sled-agent/src/services.rs
@@ -540,6 +540,9 @@ impl ServiceManager {
         }
 
         for addr in &request.addresses {
+            if *addr == Ipv6Addr::LOCALHOST {
+                continue;
+            }
             info!(
                 self.inner.log,
                 "Ensuring address {} exists",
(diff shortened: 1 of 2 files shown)


sled-agent/src/services.rs
```diff
@@ -1093,24 +1103,21 @@ impl ServiceManager {
...
+                            for address in &request.addresses {
+                                smfh.addpropvalue(
+                                    "config/address",
+                                    &format!("[{}]:{}", address, DENDRITE_PORT),
+                                )?;
+                            }
                             smfh.refresh()?;
                             // TODO: For this restart to be optional, Dendrite must
                             // implement a non-default "refresh" method.
                             smfh.restart()?;
```
smklein, line 1116: I saw that there was ongoing work here - presumably, we'll be able to refresh Dendrite without a restart soon too, right?

smklein (approved):

sled-agent/src/services.rs
Nieuwejaar: Yes.  That's currently in review.

PR #2477, opened 2023-03-03
RFD-322: `/v1/disks/{disk}/metrics/{metric}`

david-crespo:
@zephraph and I talked about `/v1/disk-metrics/{metric}?{disk_selector}` and `/v1/disks/{disk}/metrics?metric={metric}&{disk_selector}`, but decided against them. In general we are avoiding nesting selectors with names or IDs, but the set of disk metric types is a finite enum.

### DIFF
--- a/nexus/src/external_api/http_entrypoints.rs
+++ b/nexus/src/external_api/http_entrypoints.rs
@@ -155,6 +155,7 @@ pub fn external_api() -> NexusApiDescription {
         api.register(disk_create_v1)?;
         api.register(disk_view_v1)?;
         api.register(disk_delete_v1)?;
+        api.register(disk_metrics_list_v1)?;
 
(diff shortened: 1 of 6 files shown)


nexus/src/external_api/http_entrypoints.rs
```diff
@@ -2713,11 +2714,22 @@ pub enum DiskMetricName {
...
 
+/// Path parameters for metrics requests where `/metrics/{metric_name}` is
+/// appended to an existing path parameter type
+#[derive(Deserialize, JsonSchema)]
+struct MetricsPathParam<T, M> {
+    #[serde(flatten)]
+    inner: T,
+    metric_name: M,
+}
+
```
david-crespo, line 2725: didn't change the above, just moved it from the bottom of the file

PR #2481, opened 2023-03-03
Add an SMF manifest for installinator, and add an `image_type` target to package-manifest

jgallagher:
The main intent for this PR was to add an SMF manifest to run `installinator` as a service and to build an `installinator` package we could include in host OS images, but it grew a bit when I went to add `installinator` to `package-manifest.toml`. In particular, we do _not_ want installinator to be installed alongside `sled-agent` in a standard omicron install. It should only be included in the `trampoline` image. Therefore, this PR adds a new target key `image_type` that takes the possible options `standard` and `trampoline`. I marked the majority of the pre-existing packages in `package-manifest` as `only_for_targets.image_type = standard` with the exception of maghemite, which should be included in both standard and trampoline images.

The end result of this is that a normal `omicron-package package` looks almost the same, except for the additional logged key/value for the image_type (which defaults to `standard`); note that `installinator` is not present:

```
$ ./target/release/omicron-package package
Mar 03 21:40:40.251 DEBG target: Target({"image_type": "standard", "switch_variant": "stub"})
[00:00:04] ########################################       6/6       clickhouse: done
[00:00:08] ########################################      12/12      cockroachdb: done
[00:00:14] ########################################       1/1       crucible: done
[00:00:05] ########################################       1/1       crucible-pantry: done
[00:00:29] ######################################## 28471030/28471030 dendrite-stub: done
[00:00:00] ########################################       1/1       dns-server: done
[00:00:00] ########################################       3/3       external-dns-customizations: done
[00:00:00] ########################################       3/3       internal-dns-customizations: done
[00:00:06] ########################################       1/1       maghemite: done
[00:00:00] ########################################       4/4       omicron-gateway: done
[00:00:02] ########################################      65/65      omicron-nexus: done
[00:00:00] ########################################       5/5       omicron-sled-agent: done
[00:00:01] ########################################       4/4       oximeter-collector: done
[00:00:04] ########################################       1/1       propolis-server: done
[00:00:02] ########################################       1/1       wicket: done
[00:00:01] ########################################       4/4       wicketd: done
[00:00:01] ########################################       1/1       external-dns: done
[00:00:02] ########################################       1/1       internal-dns: done
[00:00:03] ########################################       1/1       switch-stub: done
```

Passing `--target image_type=trampoline` builds only maghemite and installinator:

```
$ ./target/release/omicron-package --target image_type=trampoline package
[00:00:00] ########################################       3/3       installinator: done
[00:00:01] ########################################       1/1       maghemite: done
```

### DIFF
--- a/package-manifest.toml
+++ b/package-manifest.toml
@@ -43,11 +43,10 @@ source.rust.release = true
 source.paths = [ { from = "smf/sled-agent", to = "pkg" } ]
 output.type = "tarball"
 
-# Installinator is a special package: we do not want it bundled into the normal
-# omicron install; it duplicates some of sled-agent's work and would conflict
-# with it if run side-by-side, and it's completely useless! We only want to
-# build it for inclusion in the trampoline ("recovery") image, which we indicate
-# here via `only_for_target.image_type`.
+# Installinator is a service which should run in the Global Zone for the
+# explicit purpose of recovery and OS re-installation. It should not be
+# installed concurrently with the sled-agent, and is built separately using the
+# target only_for_target.image_type = "trampoline".
 [package.installinator]
 service_name = "installinator"
 only_for_targets.image_type = "trampoline"
@@ -56,7 +55,6 @@ source.rust.binary_names = ["installinator"]
 source.rust.release = true
 source.paths = [ { from = "smf/installinator", to = "pkg" } ]
 output.type = "tarball"
-output.intermediate_only = true
 
 [package.omicron-nexus]
 service_name = "nexus"
--- a/package-manifest.toml
+++ b/package-manifest.toml
@@ -36,14 +36,21 @@
 
 [package.omicron-sled-agent]
 service_name = "sled-agent"
+only_for_targets.image_type = "standard"
 source.type = "local"
 source.rust.binary_names = ["sled-agent"]
 source.rust.release = true
 source.paths = [ { from = "smf/sled-agent", to = "pkg" } ]
 output.type = "tarball"
 
+# Installinator is a special package: we do not want it bundled into the normal
+# omicron install; it duplicates some of sled-agent's work and would conflict
+# with it if run side-by-side, and it's completely useless! We only want to
+# build it for inclusion in the trampoline ("recovery") image, which we indicate
+# here via `only_for_target.image_type`.
 [package.installinator]
 service_name = "installinator"
+only_for_targets.image_type = "trampoline"
 source.type = "local"
 source.rust.binary_names = ["installinator"]
 source.rust.release = true
@@ -53,6 +60,7 @@ output.intermediate_only = true
 
 [package.omicron-nexus]
 service_name = "nexus"
+only_for_targets.image_type = "standard"
 source.type = "local"
 source.rust.binary_names = ["nexus"]
 source.rust.release = true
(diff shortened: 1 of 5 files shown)


smf/installinator/manifest.xml
```diff
@@ -0,0 +1,36 @@
...
+  <create_default_instance enabled='false' />
+  <single_instance />
+
+  <dependency name='multi_user' grouping='require_all' restart_on='none'
+    type='service'>
+  <service_fmri value='svc:/milestone/multi-user:default' />
+  </dependency>
+
+  <exec_method type='method' name='start'
+    exec='ctrun -l child -o noorphan,regent /opt/oxide/installinator/bin/installinator install --from-ipcc --install-on-gimlet --stay-alive &amp;'
```
sunshowers, line 16: For my own knowledge: what is the ampersand at the end for? Is that shell job control and if so what does it achieve in smf? Naively I'd expect that smf would want to wait for the process to exit rather than running it in the background.

sunshowers (approved):
Looks reasonable to me.

package-manifest.toml
```diff
@@ -36,14 +36,31 @@
...
 source.rust.binary_names = ["sled-agent"]
 source.rust.release = true
 source.paths = [ { from = "smf/sled-agent", to = "pkg" } ]
 output.type = "tarball"
 
+# Installinator is a special package: we do not want it bundled into the normal
+# omicron install; it duplicates some of sled-agent's work and would conflict
+# with it if run side-by-side, and it's completely useless! We only want to
+# build it for inclusion in the trampoline ("recovery") image, which we indicate
+# here via `only_for_target.image_type`.
```
smklein, line 50: I might re-phrase this as:

> Installinator is a service which should run in the Global Zone for the explicit purpose of recovery and OS
> re-installation. It should not be installed concurrently with the sled-agent, and is built separately using
> the target `only_for_target.image_type = "trampoline"`.

package-manifest.toml
```diff
@@ -36,14 +36,31 @@
...
+[package.installinator]
+service_name = "installinator"
+only_for_targets.image_type = "trampoline"
+source.type = "local"
+source.rust.binary_names = ["installinator"]
+source.rust.release = true
+source.paths = [ { from = "smf/installinator", to = "pkg" } ]
+output.type = "tarball"
+output.intermediate_only = true
```
smklein, line 59: I don't think `intermediate_only` should be set here

smklein (approved):

package-manifest.toml
jgallagher: Good catch; this was from an earlier experiment before adding the `image_type`.

smf/installinator/manifest.xml
jgallagher: It is shell job control, yeah. There's a bit of cargo culting here, but experimentally: I believe SMF is expecting these methods (`start` here and `stop` below; others are also allowed like `refresh`) to exit indicating success/failure, and to possibly spawn a child process if necessary. If I remove the `&amp;` everything seems to still work, but I think only because `timeout_seconds='0'` means "infinite" - if I remove the `&amp;` and set `timeout_seconds` to a nonzero value, I see an error like this after that many seconds:

```
[ Mar  6 07:44:38 Method or service exit timed out.  Killing contract 224. ]
[ Mar  6 07:44:38 Method "start" failed due to signal KILL. ]
```

davepacheco: That seems right.  With `timeout_seconds='0'`, things may appear to work at first but I expect SMF would believe the service to be stuck transitioning.  Instead of showing up in `svcs` as `online`, I'd expect it to say `offline*` (which means "offline but in-transition").

PR #2502, opened 2023-03-06
Assemble installinator files into an overlay tarball

smklein:
This is necessary automation to construct recovery host OS images.

As a follow-up, I think it makes sense to have `omicron-package` produce these "overlay" files, rather than the intermediate `.tar` file, but that'll require more broad changes.


.github/buildomat/jobs/package.sh
```diff
@@ -42,39 +49,77 @@ files=(
...
-pkg_dir="$tmp/root/opt/oxide/mg-ddm"
+pkg_dir="$tmp_trampoline/root/opt/oxide/mg-ddm"
 mkdir -p "$pkg_dir"
 cd "$pkg_dir"
 tar -xvfz "$tarball_src_dir/maghemite.tar"
 cd -
 
 mkdir -p /work
-cd "$tmp" && tar cvfz /work/global-zone-packages.tar.gz oxide.json root
+cd "$tmp_trampoline" && tar cvfz /work/trampoline-global-zone-packages.tar.gz oxide.json root
```
smklein, line 116: This tarball can be unpacked, and it'll contain:

```
oxide.json
root/opt/oxide/installinator/...
root/opt/oxide/mg-ddm/...
root/lib/svc/manifest/site/installinator.xml
```

This can be used by the image-building tools (via the `-P` flag) and it should be able to construct a recovery image which includes all these files

jgallagher (approved):
Thanks! I was able to use this + `helios-build -P` successfully (with some unrelated tweaks to installinator itself).

PR #2503, opened 2023-03-06
[wicket] Add 2 new widgets

andrewjstone:
`Fade` is used to dim a rect via `style::faded_background`. `Popup` fades the background via `Fade` and overlays a Popup

Neither of these are used in this commit. Use will come in a follow up.

This code also includes necessary style changes, and some extraneous styles used for upcoming commits.

### DIFF
--- a/wicket/src/ui/widgets/fade.rs
+++ b/wicket/src/ui/widgets/fade.rs
@@ -1,0 +1,23 @@
+use crate::ui::defaults::style;
+use tui::buffer::Buffer;
+use tui::layout::Rect;
+use tui::widgets::Widget;
+
+/// Style every character as `style::faded_background`
+#[derive(Default)]
+pub struct Fade {}
+
+impl Widget for Fade {
+    fn render(self, area: Rect, buf: &mut Buffer) {
+        for x in area.left()..area.right() {
+            for y in area.top()..area.bottom() {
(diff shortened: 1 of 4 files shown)


andrewjstone:
Screen Shot 2023-03-06 at 5 07 00 PM

This is an example of a popup with a faded background. Note that we don't fade the status bar at the bottom because it's not part of the `Rect` passed to `Fade`.

wicket/src/ui/widgets/fade.rs
```diff
@@ -0,0 +1,23 @@
...
+impl Widget for Fade {
+    fn render(self, area: Rect, buf: &mut Buffer) {
+        for x in area.left()..area.right() {
+            for y in area.top()..area.bottom() {
+                buf.set_string(
+                    x,
+                    y,
+                    buf.get(x, y).symbol.clone(),
+                    style::faded_background(),
+                );
```
sunshowers, line 19: Hah, that's a fun way to do this

sunshowers (approved):
Looks good!

wicket/src/ui/widgets/fade.rs
andrewjstone: I stole the idea from @benjaminleonard :)

PR #2505, opened 2023-03-07
`/v1/` ip-pools API

zephraph:
Part of #2028


david-crespo:
Think you gotta update these too.

https://github.com/oxidecomputer/omicron/blob/a23cf3b6b0bbc5bc529f12e84e853df3ef693117/nexus/tests/integration_tests/ip_pools.rs#L34-L38

david-crespo (approved):
👌 once tests pass

PR #2507, opened 2023-03-07
[wicket] Start of Update Pane UI

andrewjstone:
This builds upon #2503 and #2504. It adds the visual components to the relevant state and wicketd interaction changes.

A tree view for update components has been added. It currently only live updates target versions for components when a TUF repo has been uploaded. If a TUF repo has not been added, and the user presses `Enter`, a popup instructing the user how to upload the repo is provided.

Currently initiating updates, and live updating progress is not implemented in the UI, although many of the underlying mechanisms exist already. It's expected for a follow up commit , that if `Enter` is hit in the tree view, and the component requires updating a warning popup will  be launched. If the user selects ok, and the update starts, then further presses of `Enter` will toggle  log views of the ongoing update. If the update has failed, we may also want a way to show this. I need to think a bit about these interactions and likely play with them once implemented.

Actual known versions of running software from inventory also needs to be put into this view. This information for SP and RoT is currently known from MGS, but we will soon have additional info available from sled-agents. It's unknown if we will actually plumb this through immediately.

The system version for an update is currently listed as "UNKNOWN" in the lower right of the screen (in the status bar). This is because we don't currently have that info inside the TUF repo. This will be fixed in a follow up as well.

Additional styling to fade either the sidebar for selecting panes, or the panes themselves was added to make it more apparent the context the user is operating in. It also makes the rack rendering look cool :)

### DIFF
--- a/wicket/src/ui/defaults/dimensions.rs
+++ b/wicket/src/ui/defaults/dimensions.rs
@@ -30,7 +30,7 @@ pub trait RectExt {
 
 impl RectExt for Rect {
     fn center_horizontally(mut self, width: u16) -> Self {
-        let center = (self.width - width) / 2;
+        let center = (self.width.saturating_sub(width)) / 2;
         self.x += center;
         self.width = width;
         self
--- a/wicket/src/ui/panes/update.rs
+++ b/wicket/src/ui/panes/update.rs
@@ -1,0 +1,286 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use std::collections::BTreeMap;
+
+use super::{align_by, help_text, Control};
+use crate::state::{artifact_title, ALL_COMPONENT_IDS};
+use crate::ui::defaults::style;
+use crate::ui::widgets::{BoxConnector, BoxConnectorKind, ButtonText, Popup};
+use crate::{Action, Event, Frame, State};
+use crossterm::event::Event as TermEvent;
+use crossterm::event::KeyCode;
+use omicron_common::api::internal::nexus::KnownArtifactKind;
+use tui::layout::{Constraint, Direction, Layout, Rect};
+use tui::text::{Span, Spans, Text};
+use tui::widgets::{Block, BorderType, Borders, Paragraph};
+use tui_tree_widget::{Tree, TreeItem, TreeState};
+
+/// Overview of update status and ability to install updates
+/// from a single TUF repo uploaded to wicketd via wicket.
+pub struct UpdatePane {
+    tree_state: TreeState,
+    items: Vec<TreeItem<'static>>,
+    help: Vec<(&'static str, &'static str)>,
+    rect: Rect,
+    // TODO: These will likely move into a status view, because there will be
+    // other update views/tabs
+    title_rect: Rect,
+    table_headers_rect: Rect,
+    contents_rect: Rect,
+    help_rect: Rect,
+    popup_open: bool,
+}
+
+impl UpdatePane {
+    pub fn new() -> UpdatePane {
+        UpdatePane {
+            tree_state: Default::default(),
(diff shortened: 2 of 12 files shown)


wicket/src/state/update.rs
```diff
@@ -0,0 +1,193 @@
...
+            FinalInstallArtifact::ControlPlane => write!(f, "CONTROL PLANE"),
+            FinalInstallArtifact::GimletRot => write!(f, "ROOT OF TRUST"),
+            FinalInstallArtifact::GimletSp => write!(f, "SERVICE PROCESSOR"),
+            FinalInstallArtifact::PscSp => write!(f, "ROOT OF TRUST"),
+            FinalInstallArtifact::PscRot => write!(f, "SERVICE PROCESSOR"),
+            FinalInstallArtifact::SwitchSp => write!(f, "ROOT OF TRUST"),
+            FinalInstallArtifact::SwitchRot => write!(f, "SERVICE PROCESSOR"),
+        }
+    }
+}
```
sunshowers, line 82: Hmm, does this need to be rebased?

sunshowers (commented):
Started reviewing this -- will let you rebase before continuing.

wicket/src/ui/defaults/dimensions.rs
```diff
@@ -30,7 +30,7 @@ pub trait RectExt {
 
 impl RectExt for Rect {
     fn center_horizontally(mut self, width: u16) -> Self {
-        let center = (self.width - width) / 2;
+        let center = (self.width.saturating_sub(width)) / 2;
```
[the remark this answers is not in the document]
andrewjstone: When shrinking the terminal during dev. It's not critical, just a robustness thing. We should probably be using saturating arithmetic in lots of cases to prevent crashes.

wicket/src/state/update.rs
andrewjstone: Rebased.

wicket/src/ui/panes/update.rs
```diff
@@ -0,0 +1,286 @@
...
+            Event::Term(TermEvent::Key(e)) => match e.code {
+                KeyCode::Up => {
+                    // Keep the rack selection in sync across panes
+                    state.rack_state.prev();
+                    self.tree_state.key_up(&self.items);
+                    Some(Action::Redraw)
+                }
+                KeyCode::Down => {
+                    // Keep the rack selection in sync across panes
+                    state.rack_state.next();
```
andrewjstone, line 173: This doesn't actually work because the tree has more elements (its subcomponents) than the rack. We'd need to keep track manually to make it work, but I don't think it's worth worrying about right now, so I'm going to remove it.

wicket/src/ui/panes/update.rs
```diff
@@ -0,0 +1,293 @@
...
+                                artifact_title(*artifact),
+                                style::selected(),
+                            ),
+                            Span::styled("UNKNOWN", style::selected_line()),
+                            Span::styled(version, style::selected()),
+                            Span::styled(s.to_string(), s.style()),
+                        ];
+                        TreeItem::new_leaf(align_by(
+                            0,
+                            25,
```
sunshowers, line 118: What does the 25 mean here? Could you add a comment?

There's another 25 below so maybe this should be defined as a constant.

sunshowers (approved):
Looks good, thanks! Will let you address CI failures before landing.

wicket/src/ui/panes/update.rs
andrewjstone: Good catch. It's the max number of columns per `Span`. I'll add a constant.

wicket/src/ui/panes/update.rs
andrewjstone: I fixed this in a follow up, with some judicious usage of the `TreeIdentifierVec` from the tui tree we are using. Huge thanks to @sunshowers for listening to me rant and pointing out something I should have recognized in the code.

wicket/src/ui/panes/update.rs
andrewjstone: fixed.

PR #2508, opened 2023-03-07
[buildomat] Job to create Helios trampoline image

smklein:
Akin to https://github.com/oxidecomputer/omicron/pull/2500 , but for the recovery image


.github/buildomat/jobs/recovery-image.sh
```diff
@@ -0,0 +1,89 @@
...
+
+pfexec mkdir -p /work
+cd /work
+
+# /work/gz: Global Zone artifacts to be placed in the Helios image.
+mkdir gz && cd gz
+ptime -m tar xvzf /input/package/work/trampoline-global-zone-packages.tar.gz
+cd -
+
+# TODO: Consider importing zones here too?
```
jgallagher, line 70: I don't think the trampoline image (currently?) has any zones to import; it's just installinator and maghemite, both in the global zone.

smklein: Comment removed

PR #2511, opened 2023-03-07
[installinator] Configure and auto-enable smf service

jgallagher:
This PR makes a handful of small-ish tweaks to installinator:

* The SMF service is now set to enabled by default
* Adds all the service dependencies that sled-agent also has
* Adds a `--bootstrap-sled` flag; if passed, we act like bootstrap agent and import+start maghemite

It also makes one related tweak to sled-agent: adding an (optional) dependency on the `t6init` service. "Optional" here means if the service is present, it is either running or in a "requires manual intervention to recover" state. From sled-agent's point of view, `t6init` is the service that causes the `cxgbeX` data links to come up, which sled-agent expects to already exist.

Builds on #2502. Using the published `trampoline-global-zone-packages.tar.gz` from this branch, I was able to boot a host OS image that successfully started maghemite and installinator.


installinator/src/dispatch.rs
```diff
@@ -126,6 +126,11 @@ struct InstallOpts {
     #[command(flatten)]
     artifact_ids: ArtifactIdOpts,
 
+    /// If true, perform sled-agent-like bootstrapping operations on startup
+    /// (e.g., configure and enable maghemite).
+    #[clap(long)]
+    bootstrap_sled: bool,
```
smklein, line 132: In a production environment, we expect this to basically always be "true", right?

smf/installinator/manifest.xml
```diff
@@ -4,16 +4,43 @@
...
+    type='service'>
+    <service_fmri value='svc:/system/filesystem/minimal' />
+  </dependency>
+
+  <!-- The bootstrapping portion of the agent uses the network to communicate with
+    other servers on the rack. -->
+  <dependency name='loopback' grouping='require_all' restart_on='none'
+    type='service'>
+  <service_fmri value='svc:/milestone/network' />
+  </dependency>
```
smklein, line 25: We wait on `svc:/milestone/network` here, but `svc:/milestone/network:default` below - is that intentional?

smklein (approved):

installinator/src/dispatch.rs
jgallagher: Yep: the command in `manifest.xml` unconditionally passes this flag. This is only here to allow installiantor to run by hand / in tests.

smf/installinator/manifest.xml
jgallagher: No idea! This is a direct copy-paste from sled-agent. I'm not sure what this is doing.

jgallagher: After chatting offline: cleaned up in 86540d2. Both installinator and sled-agent are cut down to just one `svc:/milestone/network` dependency (named `network` and does not include the `:default` suffix).

PR #2513, opened 2023-03-07
[wicketd] add code to extract host and trampoline phase 1 and 2 images

sunshowers:
This is essentially the inverse of the code to build up the tarball,
introduced in #2506. Put this code in tufaceous-lib so that it lives in
the same place and can be shared with other consumers.

With this, it should be possible to test that the installinator's
download functionality works.

Depends on #2512.


tufaceous-lib/src/artifact.rs
```diff
@@ -201,6 +205,70 @@ impl io::Read for FillerReader {
...
+    let mut buf = Vec::with_capacity(size as usize);
+    entry
+        .read_to_end(&mut buf)
+        .with_context(|| format!("error reading {file_name} from archive"))?;
+    Ok(buf.into())
+}
+
 static FILLER_TEXT: &[u8; 16] = b"tufaceousfaketxt";
 static PHASE_1_FILE_NAME: &str = "image/rom";
 static PHASE_2_FILE_NAME: &str = "image/zfs.img";
```
jgallagher, line 274: cc @smklein - these are the paths wicketd is (currently) expecting to be in the host OS tarball; we should match what you do to build it (or you can match this :man_shrugging:)

jgallagher (approved):

andrewjstone (approved):

PR #2518, opened 2023-03-08
Create distinct configs for gimlet and non-gimlet devices

smklein:
It has historically been a bit of a pain to switch between "real gimlet" and "not-real-gimlet" configuration files.

Some of this variation includes:
- Which datalinks should we be trying to use?
- Should sled agent be automatically executing an RSS plan?
- Should sled agent attempt to use emulated ZFS pools?

Modifying this config is pretty nasty, requiring in-place modification, because we have used a *single* file to represent all sled-agent configuration:
- `smf/sled-agent/config.toml`

This PR splits that configuration across different, mutually exclusive directories:
- `smf/sled-agent/gimlet/`
- `smf/sled-agent/nongimlet/`

By running a command like the following:

```bash
./target/release/omicron-package \
  --target 'image_type=standard machine_type=gimlet switch_variant=asic' \
  package
```

The value of those key-value pairs in `--target` can be in injected into paths (thanks to https://github.com/oxidecomputer/omicron-package/pull/39), so our interpretation in `package-manifest.toml` can dynamically select between different configuration options:

```toml
source.paths = [
  { from = "smf/sled-agent/manifest.xml", to = "pkg/manifest.xml" },
  # Valid options for "machine_type" include:
  # - gimlet: Assumes control of real hardware, does not emulate hardware
  # - nongimlet: Used for testing on development illumos machines
  { from = "smf/sled-agent/{{machine_type}}", to = "pkg" }
]
```


leftwo:
What about the "gimlet on a bench"?

leftwo:
> What about the "gimlet on a bench"?

Or, could the `nongimlet` be installed on an actual gimlet and it would behave as if it was not on a gimlet.
Specifically, what is currently blocking the bench gimlet is the check in `sled-hardware/src/underlay.rs`

```
    if is_gimlet().map_err(Error::SystemDetection)? {
        Dladm::list_physical().map_err(Error::FindLinks).map(|links| {
            links
                .into_iter()
                .filter(|link| link.0.starts_with(CHELSIO_LINK_PREFIX))
                .collect()
        })
    } else {
```
This forces omicron to look for and use the CHELSIO_LINK_PREFIX, and for bench gimlet, we actually want
to take the else path here:
```
    } else {
        Ok(XDE_VNIC_NAMES
            .into_iter()
            .map(|name| PhysicalLink(name.to_string()))
            .collect())
    }
```

smklein:
To be clear, I think there are infinitely many possible configs - we can continue adding as many variants as we'd like here, but I dunno if it's ever going to be possible to capture all variants. The gimlet-on-a-bench case *has* a chelsio nic, it just happens to be connected through a different physical interface (e.g., igb0 / rge1).

Should we cover the case of:
- You have *real* networking, but emulated disks?
- You have *real* disks, but you're using a standalone gimlet?
- You have a standalone gimlet, but you want to test an emulated scrimlet?

I think the variability in the configs gives us an ability to test these things - but the line of "what should we check-in" seems subjective.

This PR mostly tries to address the gap between the following use-cases:
- We want to have a configuration that works in CI - which means, works for non-gimlet test machines, which require hardware emulation and a pre-canned RSS config...
- ... but we'd also like to have a configuration that will work on our real gimlets, without requiring `sed` surgery on an existing config file

jgallagher (approved):
Code changes LGTM. I have no opinion about the names / variants of different gimlet configs that are checked in; I'd lean toward "check in any config that isn't specific to one developer's environment".

smf/sled-agent/gimlet/config.toml
```diff
@@ -0,0 +1,19 @@
+# Sled Agent Configuration
+
+# Identifies the revision of the sidecar that is attached, if one is attached.
+# TODO: This field should be removed once Gimlets have the ability to auto-detect
+# this information.
+sidecar_revision = "rev_a"
```
leftwo, line 6: Do we have any `a`s left?  Maybe this should default to `b`?

.github/buildomat/jobs/package.sh
```diff
@@ -42,17 +43,17 @@ tarball_src_dir="$(pwd)/out"
 files=(
 	out/*.tar
 	package-manifest.toml
-	smf/sled-agent/config.toml
+	smf/sled-agent/nongimlet/config.toml
```
leftwo, line 46: This only includes the `nongimlet` config file now.
Does that mean we don't need `smf/sled-agent/config.toml`, or that this build will only be for nongimlet systems?

leftwo (approved):
I guess what I was looking for here, and maybe it's okay that this does not solve my "gimlet on a bench" issue.

I agree that we don't want to have a bunch of various different configs, but I'm fine with either you have the real hardware, or you have "some other config" and if it's the other, then the config file could specify what networking interfaces should be used.

I think there will always be the desire to run a gimlet on a bench, but that's just my opinion, unnumbered by data.

smf/sled-agent/gimlet/config.toml
smklein: @Nieuwejaar - do you have an opinion on this?

.github/buildomat/jobs/package.sh
smklein: This *particular* package is only being used by the `deploy` buildomat job, which can only run on non-gimlet systems.

Separately, we actually are building the `smf/sled-agent/gimlet/config.toml` into a tarball that will become part of the regular host OS image.

smf/sled-agent/gimlet/config.toml
rmustacc: We only have a single rev A left in use on the bench and it is slated for removal next week according to @arjenroodselaar. I would not propagate Rev A as otherwise it means we have to make mods to work in every other env.

smklein: I'll pull this into https://github.com/oxidecomputer/omicron/pull/2534 to be reviewed separately, but updating this SGTM.

Nieuwejaar: Seems fine to me.

Nieuwejaar (commented):
The big-picture idea makes sense.

The `gimlet`/`nongimlet` naming feels a bit awkward.  I don't have any better ideas to offer, so I'm sure it's fine.

Maybe I'm missing something, but it feels like `gimlet` should imply `switch_variant=asic` and `nongimlet` should imply `switch_variant=stub`.  Is there a config where either of those wouldn't hold true?

Nieuwejaar:
> My opinion here isn't that strong - I think I mostly want to make the "common workflows" as easy as possible, and the "more complicated workflows" possible, at least. I'd be willing to collapse this into "`gimlet` implies you _must_ use the real switch zone, `nongimlet` implies you _must_ use soft NPU" if that seems reasonable once #2089 is merged?

I don't know enough about the usual workflows used by control plane folks to say whether requiring softnpu on non-gimlets is acceptable.  I don't see any reason not to stick with your current approach now, until we know something else would be better.

leftwo:
> > My opinion here isn't that strong - I think I mostly want to make the "common workflows" as easy as possible, and the "more complicated workflows" possible, at least. I'd be willing to collapse this into "`gimlet` implies you _must_ use the real switch zone, `nongimlet` implies you _must_ use soft NPU" if that seems reasonable once #2089 is merged?
> 
> I don't know enough about the usual workflows used by control plane folks to say whether requiring softnpu on non-gimlets is acceptable. I don't see any reason not to stick with your current approach now, until we know something else would be better.

I do want to bring up the "gimlet on a bench" again :)
That being said, I also don't really care if "gimlet" means "switch_variant=asic" and that "non-gimlet" means you get "switch_variant=stub" and you get softnpu.    As long is there is a path forward for making a "gimlet on a bench" work like a standalone system, I'm good :)

smklein:
> > > My opinion here isn't that strong - I think I mostly want to make the "common workflows" as easy as possible, and the "more complicated workflows" possible, at least. I'd be willing to collapse this into "`gimlet` implies you _must_ use the real switch zone, `nongimlet` implies you _must_ use soft NPU" if that seems reasonable once #2089 is merged?
> > 
> > 
> > I don't know enough about the usual workflows used by control plane folks to say whether requiring softnpu on non-gimlets is acceptable. I don't see any reason not to stick with your current approach now, until we know something else would be better.
> 
> I do want to bring up the "gimlet on a bench" again :) That being said, I also don't really care if "gimlet" means "switch_variant=asic" and that "non-gimlet" means you get "switch_variant=stub" and you get softnpu. As long is there is a path forward for making a "gimlet on a bench" work like a standalone system, I'm good :)

See: https://github.com/oxidecomputer/omicron/pull/2550

PR #2519, opened 2023-03-08
[wicketd] Add installinator completion events to update event log

jgallagher:
The update event log types are getting a little hard to follow and are probably ripe for a rework. This PR renames a few types and enum variants:

* `UpdateEventSuccessKind` -> `UpdateNormalEventKind`, which also gained an `::InstallinatorEvent` variant. This rename is because installinator events can represent transient failures that will be retried, which didn't seem to fit under either `Success` or `Failure`
* `UpdateEventKind::Success` -> `UpdateEventKind::NormalEvent` to go with the previous rename
* `UpdateEventKind::Failure` -> `UpdateEventKind::TerminalFailure` to clarify that these result in the update process stopping until manual intervention is applied to restart it / debug it

A couple of minor bumps in the road:

* Reporting age of events from installinator through wicketd is a little weird (see the bit in the diff where we convert to an `Instant`)
* Installinator's `CompletionEventKind` has variants that cover both success and failure cases - maybe the wicketd update log should use a flatter representation like this too?
* wicketd's openapi is now effectively "re-exporting" `CompletionEventKind` from installinator. That seems... okay? but a little weird.

Recommend viewing the diff ignoring whitespace; it counts down the changes fairly substantially.

### DIFF
--- a/wicketd/src/update_events.rs
+++ b/wicketd/src/update_events.rs
@@ -90,8 +90,8 @@ pub struct UpdateEvent {
 #[derive(Clone, Debug, JsonSchema, Serialize)]
 #[serde(rename_all = "snake_case", tag = "kind", content = "data")]
 pub enum UpdateEventKind {
-    NormalEvent(UpdateNormalEventKind),
-    TerminalFailure(UpdateEventTerminalFailureKind),
+    Normal(UpdateNormalEventKind),
+    Terminal(UpdateTerminalEventKind),
 }
 
 #[derive(Clone, Debug, JsonSchema, Serialize)]
@@ -104,7 +104,7 @@ pub enum UpdateNormalEventKind {
 
 #[derive(Clone, Debug, JsonSchema, Serialize)]
 #[serde(rename_all = "snake_case", tag = "kind", content = "data")]
-pub enum UpdateEventTerminalFailureKind {
+pub enum UpdateTerminalEventKind {
     SpResetFailed { reason: String },
     ArtifactUpdateFailed { artifact: ArtifactId, reason: String },
 }
--- a/wicketd/src/update_events.rs
+++ b/wicketd/src/update_events.rs
@@ -8,6 +8,7 @@ use camino::Utf8PathBuf;
 use gateway_client::types::HostPhase2Progress;
 use gateway_client::types::PowerState;
 use gateway_client::types::UpdatePreparationProgress;
+use installinator_common::CompletionEventKind;
 use omicron_common::update::ArtifactId;
 use omicron_common::update::ArtifactKind;
 use schemars::gen::SchemaGenerator;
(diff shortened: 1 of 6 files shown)


wicketd/src/update_tracker.rs
```diff
@@ -591,57 +618,80 @@ impl UpdateDriver {
...
+                    },
+                }
+            } else {
+                UpdateStateKind::WaitingForProgress {
+                    component: "installinator".to_owned(),
+                }
+            };
+
+        if report.completion_events.is_empty() {
+            self.set_current_update_state(progress_kind);
```
sunshowers, line 672: I don't think we need to do this. If the installinator sends both a progress and a completion message, it means that prior steps have completed and the installinator has moved to the next step.

sunshowers (commented):
I think this makes sense -- the client can display non-terminal messages as needed.

wicketd/src/update_events.rs
```diff
@@ -89,20 +90,21 @@ pub struct UpdateEvent {
 #[derive(Clone, Debug, JsonSchema, Serialize)]
 #[serde(rename_all = "snake_case", tag = "kind", content = "data")]
 pub enum UpdateEventKind {
-    Success(UpdateEventSuccessKind),
-    Failure(UpdateEventFailureKind),
+    NormalEvent(UpdateNormalEventKind),
```
andrewjstone, line 93: Nit: I'd probably shorten these to variants to just `Normal` and `Terminal` | `Failure`, since the types inside provide a good amount of info.

wicketd/src/update_events.rs
```diff
@@ -89,20 +90,21 @@ pub struct UpdateEvent {
...
+pub enum UpdateNormalEventKind {
     SpResetComplete,
     ArtifactUpdateComplete { artifact: ArtifactId },
+    InstallinatorEvent(CompletionEventKind),
 }
 
 #[derive(Clone, Debug, JsonSchema, Serialize)]
 #[serde(rename_all = "snake_case", tag = "kind", content = "data")]
-pub enum UpdateEventFailureKind {
+pub enum UpdateEventTerminalFailureKind {
```
andrewjstone, line 107: Nit: For naming symmetry I'd probably name this one `UpdateTerminalFailureEventKind`

wicketd/src/update_tracker.rs
jgallagher: Sorry, not sure I follow. If we get a report with no completion events we still want to set the current progress, right?

andrewjstone (approved):
LGTM

wicketd/src/update_tracker.rs
sunshowers: Yes, but if progress is available we should unconditionally set it.

jgallagher: Oh! I see. :facepalm: Thanks, will fix momentarily.

jgallagher: Actually, I think the code does this already! If the completion set is empty, it is set here; if the completion set is not empty, it is set (potentially repeatedly, but guaranteed at least once) in the iteration over the events.

sunshowers: Ahh, I was looking for a `set_current_update_state` call in the other branch and totally missed that.

wicketd/src/update_tracker.rs
```diff
@@ -347,18 +347,21 @@ impl UpdateDriver {
...
+    //
+    // * Push all the completion events described by `events` (possibly none)
+    // * Set the current state to `new_current` (also possibly `None`, if this
+    //   update is complete).
+    fn push_update_events(
         &self,
-        kind: UpdateEventKind,
-        new_current: Option<UpdateStateKind>,
-        timestamp: Instant,
+        events: &mut impl Iterator<Item = (Instant, UpdateEventKind)>,
```
sunshowers, line 357: Can this just be

```
events: impl Iterator<Item = (Instant, UpdateEventKind)>,
```

?

jgallagher: Yep, fixed. Thanks!

PR #2520, opened 2023-03-08
[wicket] switch to being configured via sshd ForceCommand

sunshowers:
Previously, wicket was expecting to be run as a login shell. However, we
decided in
https://github.com/oxidecomputer/minimum-upgradable-product/issues/20
that we'd instead configure wicket to be run via sshd `ForceCommand`.

The mechanism for doing that is slightly different, using
`SSH_ORIGINAL_COMMAND` rather than `-c`. Support this new mechanism and
remove support for the old login shell mechanism.

Tested through the instructions included in the PR.


wicket/README.md
```diff
@@ -99,20 +99,38 @@ return actions when they need to be redrawn. However, the `Runner` also doesn't
...
+    * To configure ssh keys, you'll need to first log in as the `wicket-test` user. To do so, run `sudo -u wicket-test -i` (Linux) or `pfexec su - wicket-test` (illumos).
+    * If using `authorized_keys`, be sure to set up the correct permissions for `~/.ssh` and its contents. As the `wicket-test` user, run `chmod go-rwx -R ~/.ssh`.
+4. Test that you can log in as the user: run `ssh wicket-test@localhost`.
+    * To debug issues related to logging in, e.g. `~/.ssh` permissions issues, check the sshd authentication log.
+    * On Linux, the authentication log is typically at `/var/log/auth.log`.
+    * On illumos, the authentication log is at `/var/log/authlog`. If it is empty, logging needs to be enabled. (If you're an Oxide employee, see [this issue](https://github.com/oxidecomputer/helios-engvm/issues/18) for how to enable logging.)
+5. Add this to the end of `/etc/ssh/sshd_config`:
+    ```
+    Match User wicket-test
+        ForceCommand /usr/local/bin/wicket
```
andrewjstone, line 122: This is going to be different for illumos most likely, as there's no `/usr/local` on my helios box at least. Maybe just make it more open ended?

andrewjstone (approved):
LGTM.

wicket/src/dispatch.rs
```diff
@@ -13,64 +13,51 @@ use slog::Drain;
...
-    /// Wicket is designed to be a login shell for use over ssh. If no arguments are specified,
-    /// wicket behaves like a TUI. However, if arguments are specified with "-c" (as in other login
-    /// shells e.g. bash -c), wicketd accepts an upload command.
-    ///
-    /// Login shell arguments are provided in a quoted form, so we expect a single String here.
-    /// This string is split using shell quoting logic to get the actual arguments.
-    #[arg(short = 'c', allow_hyphen_values = true)]
-    shell_args: Option<String>,
+pub fn exec() -> Result<()> {
+    // TODO: make this configurable?
```
jgallagher, line 17: wicketd's smf manifest has a property for its address; since wicket lives in the same zone, I believe it should be able to look it up. I can take that as a followup if you want, or you can crib from MGS, which [looks up its own SMF properties](https://github.com/oxidecomputer/omicron/blob/7c2e28c1f6a77ccab3a96deea2661846ee7f267e/gateway/src/bin/mgs.rs#L139).

wicket/README.md
jgallagher: I think step 1 kinda does make it open-ended; you could also create `/usr/local/bin` if you want.

jgallagher (approved):

wicket/README.md
sunshowers: Yeah mostly I wanted to pick a concrete example while leaving it open-ended. I've reworded things a bit to make this clearer.

wicket/src/dispatch.rs
sunshowers: I'll try doing this in a followup.

PR #2522, opened 2023-03-09
[common] switch version field to being a SemverVersion

sunshowers:
We've decided that the version field is always semver. Encode that in
the type system by using our SemverVersion wrappers throughout omicron.


sunshowers:
Do I need to update the sql scripts somehow?

common/src/api/external/mod.rs
```diff
@@ -376,14 +376,33 @@ impl JsonSchema for NameOrId {
...
     pub fn new(major: u64, minor: u64, patch: u64) -> Self {
         Self(semver::Version::new(major, minor, patch))
     }
+
+    /// This is a permissive version of the official validation regex for
+    /// semver:
+    /// https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string
+    /// It accepts some inputs that are not valid semver. See
+    /// test::test_semver_validation for an example.
+    const VALIDATION_REGEX: &str = r"^\d+\.\d+\.\d+([\-\+].+)?$";
```
david-crespo, line 405: I got that regex from [here](https://github.com/GREsau/schemars/pull/195/files#diff-3e6e1ed692353553259403498a33b36ca4920d868b70f8b1511ceb36be44110aR17) and didn't think about it as much as I should have. No reason we have to use this weaker regex instead of the semi-official ones.

david-crespo (approved):
Looks good, feel free to change the regex though

common/src/api/external/mod.rs
sunshowers: I switched over to the official validation regex.

PR #2524, opened 2023-03-09
[wicketd] store and expose system version from artifacts.json

sunshowers:
We'll want to expose the system version for a pending update in the
wicket UI.

Depends on #2523.


andrewjstone (approved):

wicketd/src/artifacts.rs
```diff
@@ -38,6 +40,7 @@ use crate::installinator_progress::IprArtifactServer;
 struct ArtifactsWithPlan {
     // TODO: replace with BufList once it supports Read via a cursor (required
     // for host tarball extraction)
+    system_version: Option<SemverVersion>,
```
jgallagher, line 43: Nit - should this live inside `UpdatePlan`? I think they're always either `Some(_)` or `None` (never one of each), and it seems reasonable to say "the system version of the update plan we'll apply if you tell us to".

jgallagher (approved):

wicketd/src/artifacts.rs
sunshowers: Yeah, makes sense. Done.

PR #2526, opened 2023-03-09
[wicket] Trigger updates and report installed versions for SP and RoT

andrewjstone:
This is a couple of commits for independent features. Future commits will add a warning popup for triggering updates and also provide progress on status and event logs.

### DIFF
--- a/wicket/src/runner.rs
+++ b/wicket/src/runner.rs
@@ -22,6 +22,7 @@ use tui::backend::CrosstermBackend;
 use tui::Terminal;
 
 use crate::ui::Screen;
+use crate::wicketd;
 use crate::wicketd::{WicketdHandle, WicketdManager};
 use crate::{Action, Event, InventoryEvent, State};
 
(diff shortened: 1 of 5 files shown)


wicket/src/runner.rs
```diff
@@ -22,6 +22,7 @@ use tui::backend::CrosstermBackend;
 use tui::Terminal;
 
 use crate::ui::Screen;
+use crate::wicketd;
```
andrewjstone, line 25: This should probably move into the next line as `self`

wicket/src/runner.rs
```diff
@@ -205,7 +206,11 @@ impl Runner {
             Action::Redraw => {
                 self.screen.draw(&self.state, &mut self.terminal)?;
             }
-            Action::Update(_component_id) => todo!(),
+            Action::Update(component_id) => {
+                self.wicketd.tx.blocking_send(
+                    wicketd::Request::StartUpdate(component_id),
```
jgallagher, line 211: Just double-checking terminology - a "component" here identifies a sled/switch/psc and not a component within a sled, right?

jgallagher (approved):
LGTM!

wicket/src/runner.rs
andrewjstone: Correct.

wicket/src/runner.rs
andrewjstone: Fixed in a follow up.

PR #2528, opened 2023-03-09
`/v1/` current user endpoints

zephraph:
Part of #2028

I've taken some artistic privileges here. Previously we had `/session/me/*`. Nothing else is hanging under `/session/` nor have I seen a proposal for anything to go there. In my investigations of other APIs folks have handled retrieving the current user in different ways:

- GitHub uses [`GET /user`](https://docs.github.com/en/rest/users/users?apiVersion=2022-11-28#get-the-authenticated-user).
- AWS uses [`GET /api/v1/me`](https://docs.aws.amazon.com/workdocs/latest/APIReference/API_GetCurrentUser.html)
- Stack Exchange uses [`GET /me`](https://api.stackexchange.com/docs/me)
- Cloudflare uses [`GET /user` (sort of)](https://api.cloudflare.com/#user-properties)

~I've done `/v1/current-user` here which is admittedly a bit verbose. It's somewhat self documenting though. I'm happy to change it back to `/me` or whatever else. The one thing about `/me` is that the operation names are strange. `me_ssh_keys_list` vs `current_user_ssh_keys`~

After polling some folks it seems like `/v1/me` is the favored route, so I just went with that.


nexus/src/external_api/http_entrypoints.rs
```diff
@@ -8884,6 +8890,66 @@ async fn role_view(
...
+}]
+pub async fn current_user_view_v1(
+    rqctx: RequestContext<Arc<ServerContext>>,
+) -> Result<HttpResponseOk<views::User>, HttpError> {
+    let apictx = rqctx.context();
+    let nexus = &apictx.nexus;
+    let handler = async {
+        // We don't care about authentication method, as long as they are authed
+        // as _somebody_. We could restrict this to session auth only, but it's
+        // not clear what the advantage would be.
```
ahl, line 8909: I know you're just copying this comment, but I think we can lose this. Maybe just something along the lines of "we report information about the user irrespective the authentication method"

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -8884,6 +8890,66 @@ async fn role_view(
...
+ }]
+pub async fn current_user_groups_v1(
+    rqctx: RequestContext<Arc<ServerContext>>,
+    query_params: Query<PaginatedById>,
+) -> Result<HttpResponseOk<ResultsPage<views::Group>>, HttpError> {
+    let apictx = rqctx.context();
+    let nexus = &apictx.nexus;
+    let query = query_params.into_inner();
+    let handler = async {
+        // We don't care about authentication method, as long as they are authed
```
ahl, line 8931: ditto

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -8884,6 +8890,66 @@ async fn role_view(
     apictx.external_latencies.instrument_dropshot_handler(&rqctx, handler).await
 }
 
+// Current user
+
+/// Fetch the user associated with the current session
+#[endpoint {
+   method = GET,
+   path = "/v1/current-user",
+   tags = ["hidden"],
```
ahl, line 8899: why hidden? My understanding of the "hidden" tag originally was that we didn't want to actually hide it from the OpenAPI document for ... convenience? I think these interfaces really are first-class and supported, right?

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -8924,6 +8992,45 @@ async fn session_sshkey_list(
     apictx.external_latencies.instrument_dropshot_handler(&rqctx, handler).await
 }
 
+/// List SSH public keys
+///
+/// Lists SSH public keys for the currently authenticated user.
+#[endpoint {
+    method = GET,
+    path = "/v1/current-user/sshkeys",
```
ahl, line 9000: ssh-keys? kabob seems to be our preference for multi-word path components?

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -8952,6 +9059,34 @@ async fn session_sshkey_create(
     apictx.external_latencies.instrument_dropshot_handler(&rqctx, handler).await
 }
 
+/// Create an SSH public key
+///
+/// Create an SSH public key for the currently authenticated user.
+#[endpoint {
+    method = POST,
+    path = "/v1/current-user/sshkeys",
```
ahl, line 9067: ditto

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -8952,6 +9059,34 @@ async fn session_sshkey_create(
     apictx.external_latencies.instrument_dropshot_handler(&rqctx, handler).await
 }
 
+/// Create an SSH public key
```
ahl, line 9062: is it weird that we have this in the API?

ahl (approved):
This is great.
1. I like `current-user`
2. I prefer `ssh-keys` to `sshkeys`
3. we need to figure out this `hidden` tag.

nexus/src/external_api/http_entrypoints.rs
david-crespo: Doesn’t need a comment at all, really. This was early when we didn’t really know how we’d think about authz and session. At this point it goes without saying that whatever operations are authorized will go through, otherwise they’ll fail, and the calling code isn’t responsible for deciding any of that.

david-crespo:
I'm also fine with `current-user` (still prefer `/me` slightly) and also prefer `ssh-keys` over `sshkeys`. Do we have a strict rule about the operation names? If not, you could make the route `/me/ssh-keys` while keeping `current_user_ssh_keys` as the operation.

nexus/src/external_api/http_entrypoints.rs
zephraph: I was just carrying these over. I'll update it.

nexus/src/external_api/http_entrypoints.rs
zephraph: Yep, I think that was just a mistake on my part.

zephraph:
> I'm also fine with current-user (still prefer /me slightly) and also prefer ssh-keys over sshkeys. Do we have a strict rule about the operation names? If not, you could make the route /me/ssh-keys while keeping current_user_ssh_keys as the operation.

I'm fine with that

nexus/src/external_api/http_entrypoints.rs
zephraph: Is it weird? I don't think so. Maybe labeled improperly? It's not actually _creating_ the key so much as creating a record of the key which can be used later to access an instance. Should we be doing something different?

david-crespo (approved):
let’s party

PR #2529, opened 2023-03-10
[package] Add build targets notion into 'omicron-package'

smklein:
Previously, `omicron-package` used a `--targets` flag that acted as a very weakly-typed key-value set.

This worked, but had a few problems:

- Being weakly typed, it was easy to get one of the characters wrong. This could result in a silent error, rather than flagging the issue. Note: this issue also applies to changing the value of any of the key-value pairs.
- For backward compatibility, all the keys were *optional*. For values like "is this a gimlet" vs "is this not a gimlet", we can pick defaults, but it feels a little hidden to never have the user opt-in to the target being built.
- It involved a lot of typing, and required repeating the "target" value at each invocation of `omicron-package`. Since the target value changes many commands, not *just* packaging, it can be irritating to re-supply the full list of key/value pairs.

## TL;DR

- Run `cargo run --release -p omicron-package -- -t default target create` once, and a `default` target will be created for you. Add the `-h` argument to that `target create` command if you want to see which defaults you picked.
- The `-t <TARGET>` command can be used to other `omicron-package` subcommands, and it'll use the configuration you picked. If you omit it, it'll use whatever last target value you created or set explicitly.
- If you want to re-visit your target choice, the `omicron-package target` subcommand lets you view / modify your selection.

## The longer version

In this PR, we introduce the notion of "build targets" explicitly.

Rather than supplying `--target <LIST OF KV PAIRS>`, the target flag is used as a label. Additionally, when parsing targets, we use strong types, which prevents errors, and future-proofs the addition of new build target types.

### The `target` subcommand

The `target` subcommand allows adding, viewing, and deleting targets:

```
Usage: omicron-package target <COMMAND>

Commands:
  create  Creates a new build target, and sets it as "active"
  list    List all existing targets
  set     Activates a build target, with a symlink named "active"
  delete  Delete an existing target
  help    Print this message or the help of the given subcommand(s)
```

For example, to create a new image, intended for utilization on an in-rack gimlet:
```
$ cargo run --release -p omicron-package -- -t deploy target create -i standard -m gimlet -s asic
Created new build target 'deploy' and set it as active
```

As another example, to create a recovery image:
```
 $ cargo run --release -p omicron-package -- -t recovery target create -i trampoline -m gimlet -s asic
Created new build target 'recovery' and set it as active
```

Information about the created targets can be displayed:

```
$ cargo run --release -p omicron-package -- target list
             deploy: image=standard machine=gimlet switch=asic 
SELECTED >>> recovery: image=trampoline machine=gimlet switch=asic
```

The "SELECTED" target is tracked via a symlink in `out/target/active`, and can be changed with:
```
$ omicron-package -t <TARGET> target set
```

### Using targets

All commands which previously operated on a list of keys can now simply reference an existing target name:

```
# Uses the 'recovery' target we made earlier
$ cargo run --release -p omicron-package -- -t recovery package
...
```

As a convenience, the target can be omitted if it has been selected:
```
$ cargo run --release -p omicron-package -- -t recovery target set
Set build target 'recovery' as active
$ cargo run --release -p omicron-package -- package
...
```


package/src/target.rs
```diff
@@ -0,0 +1,117 @@
...
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let target = Target::from_str(s)?;
+
+        let mut parsed = Self::default();
+
+        for (k, v) in target.0.into_iter() {
+            match k.as_str() {
+                "image" => {
+                    parsed.image = <Image as std::str::FromStr>::from_str(&v)?
```
jgallagher, line 93: Nit - `FromStr` powers `str::parse()`, so this could be

```rust
parsed.image = v.parse()?;
```

(and similarly for machine/switch)

jgallagher (approved):
Big :+1: on replacing the arbitrary key/value pairs with strong types. I don't _love_ implicit arguments based on the filesystem / environment (especially in combination with `create` changing the currently-active target), but agree the combination of args is starting to get unwieldy.

Not all combinations are sensible (e.g., trampoline images only make sense for real gimlets and effectively ignore the switch argument, I believe). Spitballing here - would it be reasonable to define more a few built-in combinations, maybe

* `trampoline` (trampoline, gimlet, asic/stub doesn't matter)
* `gimlet` or `prod` (standard, gimlet, asic)
* `dev` (standard, nongimlet, stub)

that could be passed as `-t` _without_ modifying or requiring changes to the filesystem? That doesn't cover all combos but might be nice for CI / folks who don't work in omicron regularly. (I'm assuming other combos are less used and it would be okay for them to be more verbose, but maybe that's wrong.)

package/src/bin/omicron-package.rs
```diff
@@ -170,6 +165,97 @@ async fn do_dot(config: &Config) -> Result<()> {
...
+                machine: machine.clone(),
+                switch: switch.clone(),
+            };
+
+            let path = get_single_target(&target_dir, name).await?;
+            tokio::fs::write(&path, Target::from(target).to_string()).await?;
+
+            replace_link(&name, &target_dir).await?;
+
+            println!("Created new build target '{name}' and set it as active");
```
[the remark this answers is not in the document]
smklein: Done

package/src/target.rs
smklein: Oooo that's much cleaner, thanks! Done

smklein:
> Not all combinations are sensible (e.g., trampoline images only make sense for real gimlets and effectively ignore the switch argument, I believe). Spitballing here - would it be reasonable to define more a few built-in combinations, maybe
> 
> * `trampoline` (trampoline, gimlet, asic/stub doesn't matter)
> * `gimlet` or `prod` (standard, gimlet, asic)
> * `dev` (standard, nongimlet, stub)
> 
> that could be passed as `-t` _without_ modifying or requiring changes to the filesystem? That doesn't cover all combos but might be nice for CI / folks who don't work in omicron regularly. (I'm assuming other combos are less used and it would be okay for them to be more verbose, but maybe that's wrong.)

So I'm not a big fan of creating another layer on top of this - I think the subtlety of "which flavor of target options you pick"  will definitely be up-to-a-developer (e.g., calling the developer build `dev` is IMO super loaded). **However**, I'm totally on-board to perform validation of some properties:

- `trampoline` should be mutually exclusive with the `image` and `machine` inputs
- `asic` should only be enabled on `machine=gimlet`

I've gone ahead and implemented this in e287a26bb274760232214270af9966c9ca69bfe9 - this should be flexible enough to allow additional validation in `KnownTarget::new`, regardless of the origin of the arguments (coming from a saved target, or from CLI arguments)

jgallagher:
> I've gone ahead and implemented this in https://github.com/oxidecomputer/omicron/commit/e287a26bb274760232214270af9966c9ca69bfe9 - this should be flexible enough to allow additional validation in KnownTarget::new, regardless of the origin of the arguments (coming from a saved target, or from CLI arguments)

That LGTM, thanks!

PR #2530, opened 2023-03-10
[nexus] carve out db queries into a crate

sunshowers:
I got a bit annoyed at nexus's build times, so I spent some time
splitting up all the db queries into their own crate. This should reduce
build times for incremental compilation.

This seems to be roughly a third of nexus, compilation-time-wise.


sunshowers:
On illumos, running `touch nexus/build.rs && ptime -m cargo build -p omicron-nexus --lib`.

Before:

```
real       19.571234027
user       12.958014323
sys         7.675880935
trap        0.000907945
tflt        0.000017923
dflt        0.046349449
kflt        0.000000000
lock       56.311348626
slp        38.025048508
lat         0.072596400
stop        0.071250481
```

After:

```
real       16.790674990
user       11.063446155
sys         6.779406520
trap        0.000883728
tflt        0.000576436
dflt        0.059963720
kflt        0.000001833
lock       47.412210898
slp        32.970974722
lat         0.089871323
stop        0.081565825
```

smklein:
Thanks for this PR! I'm taking a look now, but I'm totally on-board with the direction.

IDK if you've had a chance to look at https://buildomat.eng.oxide.computer/wg/0/details/01GV6MX3PJZ164RPV4EZ2J6CAY/tPOgXOC3a7lCcT3dx1DBC2yQa4vejBdkPCOxEchMaFRWnSdy/01GV6MY6V22RH8NPRQJ0T5A5PW#S3941- but:

```
ld.so.1: nexus_db_queries-522748c852f90461: fatal: libpq.so.5: open failed: No such file or directory
```

Looks like either `build.rs` or `Cargo.toml` issue - I recommend looking at how we depend on `omicron-rpaths`.

sunshowers:
@smklein thanks, fixed

nexus/db-queries/src/db/pool.rs
```diff
@@ -53,8 +53,7 @@ impl Pool {
         Pool { pool }
     }
 
-    #[cfg(test)]
-    pub fn new_failfast(db_config: &DbConfig) -> Self {
+    pub fn new_failfast_for_tests(db_config: &DbConfig) -> Self {
```
smklein, line 56: ah, so the `#[cfg(test)]` disappeared because this crosses crate boundaries, right?

smklein (approved):

nexus/db-queries/src/db/pool.rs
sunshowers: Yeah

sunshowers: Yeah, sadly.

Alternatively we could add a `testing` feature and expose this via `#[cfg(feature = "testing")]`, but that would cause build times to get worse.

smklein: Yeah, I figured. I'm okay with limiting feature flags - exposing this API as-is is a con, but a relatively minor one in the grand scheme of things.

PR #2533, opened 2023-03-10
ensure that created zpools support encryption

jmpesp:
ensure that zpools created by the sled agent support encrypted datasets

### DIFF
--- a/illumos-utils/src/zpool.rs
+++ b/illumos-utils/src/zpool.rs
@@ -170,6 +170,17 @@ impl Zpool {
         cmd.arg(&name.to_string());
         cmd.arg(vdev);
         execute(&mut cmd).map_err(Error::from)?;
+
+        // Ensure that this zpool has the encryption feature enabled
+        let mut cmd = std::process::Command::new(PFEXEC);
+        cmd.env_clear();
+        cmd.env("LC_ALL", "C.UTF-8");
+        cmd.arg(ZPOOL)
+            .arg("set")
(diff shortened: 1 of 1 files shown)


illumos-utils/src/zpool.rs
```diff
@@ -170,6 +170,17 @@ impl Zpool {
...
         cmd.arg(vdev);
         execute(&mut cmd).map_err(Error::from)?;
+
+        // Ensure that this zpool has the encryption feature enabled
+        let mut cmd = std::process::Command::new(PFEXEC);
+        cmd.env_clear();
+        cmd.env("LC_ALL", "C.UTF-8");
+        cmd.arg(ZPOOL)
+            .arg("set")
+            .arg("feature@encryption=enabled")
```
smklein, line 180: What does this do?

If I'm reading the docs correctly `zfs create -o encryption=on` is necessary to actually turn encryption on, right?

smklein (approved):

illumos-utils/src/zpool.rs
jmpesp: The `zfs create` is necessary to create a dataset with encryption on, yeah, but the pool feature still has to be enabled (as far as I know). All my illumos machines already have it enabled but this commit makes sure it's on.

smklein: Just confirming that this is essentially flagging down "we can encrypt" rather than anything more material. A reasonable check to add, but just wanted to validate my understanding.

PR #2534, opened 2023-03-11
Bump default sidecar rev to "b"

smklein:


smklein:
See https://github.com/oxidecomputer/omicron/pull/2518#discussion_r1132562979 for original context

leftwo (approved):

Nieuwejaar (approved):

Issue #2290, opened 2023-01-31
Nexus needs to manage OPTE V2P mappings


jmpesp:
In OPTE, a `Virt2Phys` mapping describes the next hop of an encapsulated packet if that inner packet's destination remains in the VPC (otherwise, the destination is boundary services). OPTE currently creates a single `Virt2Phys` mapping when creating an xde device:

https://github.com/oxidecomputer/opte/blob/415bf265fad9af85e2fb5fcb2644c7be8f48b3ca/xde/src/xde.rs#L517-L530

The control plane currently does not set any other `Virt2Phys` mappings. As a result, if VMs in the same VPC are scheduled onto different physical host sleds they cannot communicate with each other - OPTE drops packets when there isn't a mapping for them:

https://github.com/oxidecomputer/opte/blob/415bf265fad9af85e2fb5fcb2644c7be8f48b3ca/oxide-vpc/src/engine/overlay.rs#L250-L304

A `Virt2Phys` mapping requires the following information:

- a VM's NIC's private IP address
- a VM's NIC's mac
- the underlay ipv6 address of the sled hosting the VM
- the VM's VPC VNI

Let's say there are two sleds, both with one VM each, where each VM has one NIC:

```
sled A: fd00:1122:3344:100::1
  VM A:
    NIC A: 172.30.0.5 / A8:40:25:F2:D4:4C

sled B: fd00:1122:3344:200::1
  VM B:
    NIC B: 172.30.0.6 / A8:40:25:FD:6E:AE
```

For these VMs to be able to communicate, two mappings need to be made:

1. on sled B, a mapping to NIC A: `pfexec /opt/oxide/opte/bin/opteadm set-v2p 172.30.0.5 A8:40:25:F2:D4:4C fd00:1122:3344:100::1 <VPC VNI>`

2. on sled A, a mapping to NIC B: `pfexec /opt/oxide/opte/bin/opteadm set-v2p 172.30.0.6 A8:40:25:FD:6E:AE fd00:1122:3344:200::1 <VPC VNI>`

Once this is done, the VMs can communicate.

It should be Nexus' responsibility to create these `Virt2Phys` mappings:

- when an instance is created, OPTE will create a mapping when each xde device is created, but Nexus will be responsible for all other mapping creation
- these mappings need to be updated when an instance is migrated
- these mappings need to be deleted when an instance is destroyed (this requires https://github.com/oxidecomputer/opte/issues/332)
[...]

PR #2536, opened 2023-03-11
Manage OPTE V2P mappings

jmpesp:
During instance creation, create the appropriate OPTE V2P mappings on other sleds so that instances there can connect to the new instance.

During instance deletion, remove the corresponding mappings.

During instance migration, both create and remove the corresponding mappings.

This is one part of #2290. Another commit will be required when https://github.com/oxidecomputer/opte/issues/332 is addressed.


gjcolombo:
I haven't looked at the PR in any detail yet, but I did want to make a comment about one thing before it gets to be Monday and I forget to mention it:

> During instance migration, both create and remove the corresponding mappings.

I don't think Nexus should do this during the migration saga. I think it's better to do it once the saga is finished and Nexus is sure the instance has migrated, since that minimizes the amount of time the instance will be running with the wrong mappings in place.

For background, the general sequence of events in a migration is

1. saga runs, starts the target instance, and tells it to start migrating
2. migration source and target run through the "pre-pause" parts of the migration protocol (negotiating a version, checking compatibility, transferring any state that can be transferred while the source is running)
3. source pauses & finishes sending state to target
4. target resumes
5. Nexus learns migration has succeeded & does any post-migration actions

If Nexus deletes the source's V2P mappings and establish the target's mappings in step 1, then (I assume) connectivity to the source is degraded from step 1 to step 4. If Nexus waits until after the migration is done, connectivity is interrupted in steps 3 through 5. In most cases, I expect step 2 will take much longer than step 5, because step 2 will involve transferring a lot of dirty pages (in the hopes that Propolis can then avoid transferring them while the VM is paused in step 3), whereas step 5 should just involve some (hopefully comparatively speedy) RPCs to the affected sleds. So, to minimize the time where the instance is running with the wrong V2P mappings, Nexus should change mappings after migration finishes.

The other thing to keep in mind here is that successful completion of the migration saga doesn't imply a successful migration--Nexus doesn't know whether the instance will end up running on the source or the target until the saga has succeeded and the involved Propolises take control of its outcome. I suspect we'll need some more general logic in Nexus that watches incoming instance state changes and can say "oops, this instance's location isn't what I thought it was, let me regenerate the V2P mappings for it." I'm happy to try to add that as part of the migration work I have in my queue.

RFD 361 section 6.1 discusses all this in a little more detail.

Sorry for the drive-by comment here--this area's been top-of-mind for me for the last few weeks so I wanted to mention all this. I'd be glad to help take a look at the code changes in more detail next week.

nexus/src/app/sagas/instance_create.rs
```diff
@@ -106,6 +106,9 @@ declare_saga_actions! {
     INSTANCE_ENSURE -> "instance_ensure" {
         + sic_instance_ensure
     }
+    V2P_ENSURE -> "v2p_ensure" {
```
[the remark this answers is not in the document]
gjcolombo: I can't leave comments on an arbitrary part of the file (ugh) so leaving this here because it's closely related: if this ordering changes and `V2P_ENSURE` gets an undo step, we should also update the tests in this module to check that no V2P mappings exist after a saga is torn down (there are a bunch of functions down there that check this sort of thing for other side effects).

gjcolombo (commented):
I took a closer look at the Nexus and sled agent parts of the change and left a couple extra questions about the sagas and synchronization between attempts to program mappings. Some of these are more "file an issue and we'll get to it later" sorts of things than actual merge blockers, but I'd like to get aligned on how we're going to program mappings in the instance sagas.

nexus/src/app/sled.rs
```diff
@@ -231,4 +233,166 @@ impl super::Nexus {
...
+    // OPTE V2P mappings
+
+    /// Ensure that the necessary v2p mappings for an instance are created
+    pub async fn create_instance_v2p_mappings(
+        &self,
+        opctx: &OpContext,
+        instance_id: Uuid,
+    ) -> Result<(), Error> {
+        // For every sled that isn't the sled this instance was allocated to, create
+        // a virtual to physical mapping for each of this instance's NICs.
```
bnaecker, line 246: I'm not sure about this approach. I get the point of a quick solution, but a more appropriate one doesn't seem like significantly more work. We're creating a bunch of irrelevant mappings. And more importantly, we've explicitly decided to defer VPC peering in the MVP. I don't think we _want_ to send mappings for instances in another VPC. You rightly point out that OPTE will drop such traffic, but I don't think we should bake that knowledge in here. I think it makes sense to try to send the mappings we explicitly want to allow.

As an alternative, what about including a function call here, something like `Nexus::sleds_for_vpc_mapping()`. That would currently select all sleds that are in the same VPC as the provided instance. In the future, we can update that to make more fine-grained decisions, taking into account peering for example. But we have an interface for selecting the "right" sleds, encapsulating what "right" means.

I believe the actual SQL query to derive this information would be pretty simple. It should look basically like this:

```sql
SELECT
    active_server_id
FROM
    (
        SELECT
            instance_id
        FROM
            network_interface
        WHERE
            vpc_id = $VPC_ID
            AND time_deleted IS NULL
        LIMIT
            1
    )
        AS inst
    INNER JOIN instance ON inst.instance_id = instance.id
```

That selects the ID of all instances from the NIC table sharing the same VPC. (You have that from the call to `derive_network_interface_info()`.) Then we join with the instance table on the primary key, and select the sled ID. It's not too much additional work to join on the sled table and pull the IP directly, too.

This is definitely no worse than we have now, in terms of atomicity, and I think it allows us to more faithfully represent the intent and evolve as we need finer-grained data.

nexus/src/app/sled.rs
```diff
@@ -231,4 +233,166 @@ impl super::Nexus {
...
+            .await?;
+
+        // Lookup the physical host IP of the sled hosting this instance
+        let instance_sled_id = db_instance.runtime().sled_id;
+        let physical_host_ip: std::net::IpAddr =
+            (*self.sled_lookup(&self.opctx_alloc, &instance_sled_id).await?.ip)
+                .into();
+
+        let mut last_sled_id: Option<Uuid> = None;
+        loop {
```
bnaecker, line 281: Maybe a nit, bu it seems like this loop could be parallelized and then collected, via something like `FuturesUnordered`.

nexus/src/app/sagas/instance_create.rs
```diff
@@ -1009,6 +1014,26 @@ async fn sic_instance_ensure(
...
+async fn sic_v2p_ensure(
+    sagactx: NexusActionContext,
+) -> Result<(), ActionError> {
+    let osagactx = sagactx.user_data();
+    let params = sagactx.saga_params::<Params>()?;
+    let opctx = OpContext::for_saga_action(&sagactx, &params.serialized_authn);
+    let instance_id = sagactx.lookup::<Uuid>("instance_id")?;
+
+    // TODO-idempotent if this action fails half way through, unwind is not
+    // called!
```
bnaecker, line 1027: You make a note about this elsewhere, but I'm wondering about unwinding this saga. The way we've structure this, we fail the entire saga node if the request to update mappings on any one sled fails, but without unwinding the mappings we've already sent.

One option would be to put a no-op forward action before this one, that deletes all mappings for an instance. That way we always unwind them, regardless of whether we succeed partway through the following node. I get that there's no actual way to delete these mappings until we fix OPTE#332, but it would structure the saga in a way to alleviate the fact that the `sic_v2p_ensure` action really takes many individual actions.

nexus/src/app/sled.rs
```diff
@@ -231,4 +233,166 @@ impl super::Nexus {
...
+            let pagparams = DataPageParams {
+                marker: last_sled_id.as_ref(),
+                direction: dropshot::PaginationOrder::Ascending,
+                limit: std::num::NonZeroU32::new(10).unwrap(),
+            };
+
+            let sleds_page =
+                self.sleds_list(&self.opctx_alloc, &pagparams).await?;
+
+            for sled in &sleds_page {
```
bnaecker, line 364: Same note about parallelization.

nexus/src/app/sled.rs
```diff
@@ -231,4 +233,166 @@ impl super::Nexus {
...
+        // For every sled that isn't the sled this instance was allocated to, delete
+        // the virtual to physical mapping for each of this instance's NICs. If
+        // there isn't a V2P mapping, del_v2p should be a no-op.
+        let (.., authz_instance, db_instance) =
+            LookupPath::new(&opctx, &self.db_datastore)
+                .instance_id(instance_id)
+                .fetch_for(authz::Action::Read)
+                .await?;
+
+        let instance_nics = self
```
bnaecker, line 341: So, if we change the endoint in the sled agent to be something like `DELETE /v2p/{instance_id}`, we could remove this call entirely. We only need the instance ID to do that.

nexus/tests/integration_tests/instances.rs
```diff
@@ -3149,6 +3153,83 @@ async fn test_instance_create_in_silo(cptestctx: &ControlPlaneTestContext) {
...
+        .unwrap();
+
+    let mut sled_agents: Vec<&Arc<SledAgent>> =
+        additional_sleds.iter().map(|x| &x.sled_agent).collect();
+    sled_agents.push(&cptestctx.sled_agent.sled_agent);
+
+    for sled_agent in &sled_agents {
+        let v2p_mappings = sled_agent.v2p_mappings.lock().await;
+        if sled_agent.id != db_instance.runtime().sled_id {
+            assert!(!v2p_mappings.is_empty());
```
bnaecker, line 3208: Could we test the contents of the mappings? I don't see any other location that we ensure we've really provided the right data to OPTE.

bnaecker (commented):
Thanks a lot for getting this going! It's really exciting to see more integration with OPTE! I've got a few comments about the general approach. I think we should at least reconsider propagating the mappings to all sleds, regardless of whether they need them. It seems like not a heavy additional lift, and I think it's closer to the intent we're after.

Thanks for all the work @jmpesp!

nexus/src/app/sled.rs
bnaecker: Oh, I should have added a big caveat! Diesel can be an absolute nightmare sometimes. The SQL query here is pretty easy, but it involves a subquery. That can sometimes be very hard to express in Diesel. I _think_ it should be straightforward here, but if we find that it can't be expressed easily in Diesel's DSL, then we should seriously reconsider. Hand-rolled queries are definitely much more work.

nexus/src/app/sled.rs
```diff
@@ -231,4 +233,166 @@ impl super::Nexus {
...
+        // TODO this is a first-pass naive approach. A more optimized approach would
+        // be to see what other instances share the VPC this instance is in (or more
+        // generally, what instances should have connectivity to this one), see what
+        // sleds those are allocated to, and only create a v2p mapping for those
+        // sleds. However, this naive approach has the advantage of not requiring
+        // additional work: v2p mappings have to be bidirectional in order for both
+        // instances's packets to make a round trip. After this function executes, a
+        // v2p mapping to this instance will exist on every sled, and when another
+        // instance is allocated to a random sled, that sled will already contain a
+        // mapping to this instance.
```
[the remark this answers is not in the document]
jmpesp: This was on my mind too. The answer to this is probably a step in the currently non-existent saga that (like you said) fires when a new sled joins the cluster or an existing sled reboots. I opened https://github.com/oxidecomputer/omicron/issues/2545

illumos-utils/src/opte/illumos/port_manager.rs
```diff
@@ -490,6 +493,51 @@ impl PortManager {
...
+
+    pub fn set_virtual_nic_host(
+        &self,
+        mapping: &SetVirtualNetworkInterfaceHost,
+    ) -> Result<(), Error> {
+        let hdl = OpteHdl::open(OpteHdl::XDE_CTL)?;
+
+        hdl.set_v2p(&SetVirt2PhysReq {
+            vip: match mapping.virtual_ip {
+                std::net::IpAddr::V4(ipv4) => oxide_vpc::api::IpAddr::Ip4(
```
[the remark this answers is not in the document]
jmpesp: I do see

```rust
#[cfg(any(feature = "std", test))]
impl From<std::net::IpAddr> for IpAddr {
```

in opte/opte-api/src/ip.rs but when I try this I see

```
error[E0277]: the trait bound `oxide_vpc::api::IpAddr: From<std::net::IpAddr>` is not satisfied
   --> illumos-utils/src/opte/illumos/port_manager.rs:505:18
    |
505 |             vip: mapping.virtual_ip.into(),
    |                  ^^^^^^^^^^^^^^^^^^ ---- required by a bound introduced by this call
    |                  |
    |                  the trait `From<std::net::IpAddr>` is not implemented for `oxide_vpc::api::IpAddr`
    |
    = help: the following other types implement trait `From<T>`:
              <oxide_vpc::api::IpAddr as From<oxide_vpc::api::Ipv4Addr>>
              <oxide_vpc::api::IpAddr as From<oxide_vpc::api::Ipv6Addr>>
    = note: required for `std::net::IpAddr` to implement `Into<oxide_vpc::api::IpAddr>`
```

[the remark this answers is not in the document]
bnaecker: Weird. There might be some feature-inclusion confusion, I will look into it.s

sled-agent/src/http_entrypoints.rs
```diff
@@ -280,3 +283,37 @@ async fn vpc_firewall_rules_put(
 
     Ok(HttpResponseUpdatedNoContent())
 }
+
+/// Create a mapping from a virtual NIC to a physical host
+#[endpoint {
+    method = PUT,
+    path = "/set-v2p",
+}]
```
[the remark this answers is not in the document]
jmpesp: Strictly speaking the mapping we're applying is for a specific _network interface_, not instance. For example, during a `PUT /v1/network-interfaces/{interface}`, this would incur `PUT /v2p/{interface}` and `DELETE /v2p/{interface}` calls, but not affect the instance placement. I've done this in ffb47670

nexus/src/app/sled.rs
jmpesp: I wasn't smart enough to do this with `FuturesUnordered`, so I did it with `tokio::spawn` instead: 117f8b30

nexus/src/app/sled.rs
jmpesp: see 117f8b30

nexus/tests/integration_tests/instances.rs
jmpesp: Good call, done in 210adc9a

nexus/src/app/sled.rs
jmpesp: It's not that it's impossible to do, but I'm anxious about TOCTOU (like Greg mentioned in another comment). Without something always running in the background that can apply corrections we run the risk of these mappings not being correct.

nexus/src/app/sagas/instance_create.rs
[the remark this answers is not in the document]
jmpesp: done in 48754ea2

nexus/src/app/sled.rs
```diff
@@ -231,4 +233,166 @@ impl super::Nexus {
         self.db_datastore.service_upsert(opctx, service).await?;
         Ok(())
     }
+
+    // OPTE V2P mappings
+
+    /// Ensure that the necessary v2p mappings for an instance are created
+    pub async fn create_instance_v2p_mappings(
```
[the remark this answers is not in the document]
jmpesp: comment added in 3d5e677d

nexus/src/app/sagas/instance_create.rs
jmpesp: done in 48754ea295053ca3d2d406d6dcc7bdfbc466e192

nexus/src/app/sled.rs
jmpesp: I don't grok this - won't the OPTE ioctl require all fields (it doesn't know about instance or NIC id) to delete the mapping?

nexus/src/app/sled.rs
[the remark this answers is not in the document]
gjcolombo: New comment LGTM, except that I don't quite follow why this is a problem if the caller targets a subset of sleds but not the complete set of sleds. (The race I had in mind was one where one call starts updating to the correct mappings for time T; that call gets swapped out; the instance moves, generating new mappings for T+1; a second call applies the T+1 mappings; and then the first call resumes and applies the now-obsolete mappings from T. If that's possible, I don't think it matters whether the mappings involve all the sleds or just some of them.)

nexus/src/app/sled.rs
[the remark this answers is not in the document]
gjcolombo: LGTM, thanks!

nexus/src/app/sled.rs
bnaecker: I agree there is a risk of these mappings being incorrect, for example if an instance is moved before they can be applied to all sleds. I don't see how propagating them to sleds that don't need it solves that. As an example, if we propagate a mapping to some sled, but the instance moves, that sled now has the wrong mapping in any case, even if that's only for some period of time.

I had imagined we'd solve this mechanisms put forward in RFD 373 on reliable persistent workflows, such as [generation numbers](https://rfd.shared.oxide.computer/rfd/0373#_generation_numbers). We could compute the set of mappings at some point, along with generation number. If a sled receives a mapping that is prior to its existing generation, it may simply ignore it. It still seems like we want to compute the right set of data at each generation.

sled-agent/src/http_entrypoints.rs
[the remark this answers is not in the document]
bnaecker: Yes, you're right, the mappings tell OPTE where to find a virtual address in the physical world. And all NICs for an instance are on the same host and share the same VNI, so I was only hoping to simplify the interfaces. If you feel it doesn't do that, that's cool too.

I don't quite follow the bit about incurring some additional calls, could you expand on that?

nexus/src/app/sled.rs
bnaecker: Yeah I think you're right. I might have been conflating this with the simulated sled agent, where things are stored that way. It might be possible to send just the VNI + MAC, but I don't see that being a huge win.

nexus/src/app/sled.rs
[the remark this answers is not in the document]
jmpesp: This is the scenario you're describing (I think): with four sleds, where two instances (each with one NIC) that share a VPC are allocated to sleds A and B:

```
|A| |B| |C| |D|
|-| |-| |-| |-|
|1| |2| | | | |
| | | | | | | |
| | | | | | | |
```

The minimum number of OPTE V2P mappings required (in addition to the ones automatically created) for these two instances to communicate is:

```
on sled A: [
  instance 1 -> sled A (automatically created by OPTE)
  instance 2 -> sled B
]
on sled B: [
  instance 1 -> sled A
  instance 2 -> sled B (automatically created by OPTE)
]
on sled C: [
]
on sled D: [
]
```

If an additional instance 3 in the same VPC is allocated to sled C, then the changes required are:

```
- add a mapping on sled A: instance 3 -> sled C
- add a mapping on sled B: instance 3 -> sled C
- add a mapping on sled C: instance 1 -> sled A
- *add a mapping on sled C: instance 2 -> sled B*
```

If instance 2 is migrated from B to D, then the changes required *before instance 3 exists* are:

```
- delete the mapping on sled A: instance 2 -> sled B
- add a mapping on sled A: instance 2 -> sled D
```

If instance 3 exists, then additional changes are required:

```
- delete the mapping on sled C: instance 2 -> sled B
- *add a mapping on sled C: instance 2 -> sled D*
```

The mappings surrounded in asterisks conflict with each other. If the B->D migration occurs in the middle of the instance C creation saga, it's possible the `sled C: instance 2 -> sled B` mapping will be applied last, which is incorrect.

If there are more unnecessary mappings, then I think you're right, it doesn't help. I'm not sure what I was thinking.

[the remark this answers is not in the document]
jmpesp: I was also unsure of OPTE's behaviour: if `instance 2 -> sled B` and `instance 2 -> sled D` are applied to the same sled, what happens? I tried

```
opteadm set-v2p 172.30.0.5 A8:40:25:F2:D4:4C fd00:1122:3344:100::1 912
opteadm set-v2p 172.30.0.5 A8:40:25:F2:D4:4C fd00:1122:3344:101::1 912
```

and dump-v2p showed:

```
IPv4 mappings
VPC IP                   VPC MAC ADDR      UNDERLAY IP
172.30.0.5               A8:40:25:F2:D4:4C fd00:1122:3344:101::1
```

so it appears that the last mapping wins (not that two mappings would exist or that an error would be thrown).

sled-agent/src/http_entrypoints.rs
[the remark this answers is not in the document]
jmpesp: Sorry, I only mean that "moving" a NIC around (if possible) would incur calls to interface specific endpoints, not instance specific endpoints.

nexus/src/app/sled.rs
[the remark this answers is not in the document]
gjcolombo: I thought about this some more and I think your original comment was right. tl;dr: my kingdom for reliable persistent workflows. More analysis below.

---

> If there are more unnecessary mappings, then I think you're right, it doesn't help. I'm not sure what I was thinking.

I might have been mistaken. I think it *might* help, albeit kind of indirectly. Here's the scenario I think you're describing:

- we start with instance 1 on sled A
- we create instance 2 on sled B
- as part of that saga, we read the locations of the other instances in the VPC and see instance 1 is on sled A. So we know we need to create two mappings:
  - tell sled A "instance 2 is on sled B"
  - tell sled B "instance 1 is on sled A"
- then, while this saga is still in progress, we migrate instance 1 to sled C.

We want to end up with the following knowledge on each sled (ignoring the identity mappings):

- Sled A: no longer in use
- Sled B: instance 1 is on sled C
- Sled C: instance 2 is on sled B

And we have two problems:

1. Sled B might get the wrong information about instance 1: instance 1's migration-completion thread and instance 2's creation thread are racing to tell sled B where instance 1 is (instance 2's creator thinks it's sled A, the migration thinks it's sled C).
2. Both sleds A and C will be told where instance 2 is (the end-of-migration thread will tell C and instance 2's creation thread will tell A). This creates an unnecessary mapping, but not an incorrect one.

The trick is that if, when we created instance 1, we told *all* the other sleds about it, then when we create instance 2, there is no need for that creation saga to tell sled B where to find instance 1--we know that that was done already. So the instance creator doesn't have to do anything that will conflict with the migration, which prevents the race.

---

Note that I think this only works because the instance creation saga guarantees that all the V2P mappings exist before the instance starts to run. If we did it in the other order, we could have this race instead:

- Instance 1 is created and starts running on sled A
- Afterward, we start telling other sleds that I1 is on sled A
- But the instance is created, so it can migrate! A migration moves I1 to sled B
- After migration, the end-of-migration thread starts telling other sleds that I1 is on sled B

*This* is the race I think I had in mind--you have two broadcasters of an instance's location who have different ideas about where the instance is. This isn't a problem for creation with the saga ordered as it is today, but I think it *is* a problem for an instance that migrates once and then migrates again before we finish dealing with the consequences of the first migration (at least if we defer programming V2P mappings until after migration is done).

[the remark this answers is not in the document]
jmpesp: > The trick is that if, when we created instance 1, we told all the other sleds about it, then when we create instance 2, there is no need for that creation saga to tell sled B where to find instance 1--we know that that was done already. So the instance creator doesn't have to do anything that will conflict with the migration, which prevents the race.

Right, this is what I was proposing (where any created instance already knows where previously created instances are), though now that I know that calling `set-v2p` overwrites previous mappings with new information, I'm not even sure `del-v2p` is needed!

I think this works out. Say the rule is:

> During instance creation or for the destination of a migration, add a mapping on every other sled: `new/migrated instance -> sled X`.

Let's take the same example as above: two instances, four sleds:

```
|A| |B| |C| |D|
|-| |-| |-| |-|
|1| |2| | | | |
| | | | | | | |
| | | | | | | |
```

Here's what would happen:

```
- OPTE automatically adds to sled A: instance 1 -> sled A
- add a mapping on sled B: instance 1 -> sled A
- add a mapping on sled C: instance 1 -> sled A
- add a mapping on sled D: instance 1 -> sled A

- add a mapping on sled A: instance 2 -> sled B
- OPTE automatically adds to sled B: instance 2 -> sled B
- add a mapping on sled C: instance 2 -> sled B
- add a mapping on sled D: instance 2 -> sled B
```

Now, instance 3 is created and allocated to sled C:

```
- add a mapping on sled A: instance 3 -> sled C
- add a mapping on sled B: instance 3 -> sled C
- OPTE automatically adds to sled C: instance 3 -> sled C
- add a mapping on sled D: instance 3 -> sled C
```

Now, say a concurrent migration is performed of instance 2 from B to D:

```
- add a mapping on sled A: instance 2 -> sled D
- add a mapping on sled B: instance 2 -> sled D
- add a mapping on sled C: instance 2 -> sled D
- OPTE automatically adds to sled D: instance 2 -> sled D
```

There's no mapping conflict here!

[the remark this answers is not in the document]
gjcolombo: We discussed this extensively offline and agreed that broadcasting mappings fixes these races, with one catch: we have to guarantee that we only program V2P mappings when the mapping programmer is certain that the instance's sled assignment is fixed. Fortunately:

- For instance create, this is trivial (the instance was just assigned a sled, and it's not running so it can't migrate).
- For migration it's a little more complicated (the migration-end logic needs to ensure that the "lock" preventing a second migration from starting is not "dropped" until the V2P mappings are re-established). But none of this logic exists yet (I'm working on it now, in fact), so not a problem for this PR.
- Eventually, we're going to want instance start to be a saga that selects a sled for the instance to run on. There we have to be careful not to hit races where e.g. an instance migrates, then *stops* (which Nexus generally can't prevent), then gets restarted on a different sled). But instance start isn't a saga today, and doesn't change any sled assignments, and so this is also not a problem for this PR.

In the fullness of time, it might be simpler just to have a generation number for V2P assignments as @bnaecker suggested [here](https://github.com/oxidecomputer/omicron/pull/2536#discussion_r1134943007) so that we don't have to worry so much about guaranteeing that there's only one assigner who's guaranteed to have the right assignments, but for now it seems like we might just get away with broadcasting mappings.

Homework assignments:

- @jmpesp will update the comment I had originally tagged to mention the "sled assignment can't change while we're doing this" invariant
- @gjcolombo will update RFD 361's discussion of what happens when migration finishes

(My thanks again to @jmpesp for letting me hash all this out here--the migration protocol is going to be better for having discussed all this in detail!)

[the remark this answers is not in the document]
jmpesp: see e9c44118 for the updated comment. Thanks for the discussion Greg :)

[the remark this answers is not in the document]
gjcolombo: Ship it!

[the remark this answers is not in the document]
jmpesp: @gjcolombo I added a comment in the instance migration saga about the unidirectional network blip we discussed in dbafcc12, have a look and let me know if I've missed anything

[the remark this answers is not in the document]
gjcolombo: New comment LGTM. 

If we're going to create the mappings in the saga, I would still probably prefer to put that it in its own saga step that precedes `sim_instance_migrate`, which would of course require the comment to be reworded slightly, but I think its essence is accurate.

nexus/src/app/sled.rs
jmpesp: @bnaecker: @gjcolombo and I discussed the broadcast method, and are convinced at this point that it seems safe. See the comment chain at https://github.com/oxidecomputer/omicron/pull/2536#discussion_r1133143006.

After a re-read, I *think* I grok the generation number approach outlined in RFD 373. At this point though I think I'd like to create a follow on issue for that work given that we're thinking the broadcast method is safe. Thoughts?

nexus/src/app/sled.rs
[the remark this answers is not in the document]
jmpesp: Sounds good, done in 9a731557

illumos-utils/src/opte/illumos/port_manager.rs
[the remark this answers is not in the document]
luqmana: If you've rebased/merged main recently give this a try again? I added those conversion in opte but I don't think they got pulled into the rev omicron was pinned to until this week or so

illumos-utils/src/opte/illumos/port_manager.rs
```diff
@@ -490,6 +493,53 @@ impl PortManager {
...
+                ether: oxide_vpc::api::MacAddr::from(
+                    (*mapping.virtual_mac).into_array(),
+                ),
+                ip: match mapping.physical_host_ip {
+                    std::net::IpAddr::V4(_) => {
+                        return Err(opte_ioctl::Error::InvalidArgument(
+                            String::from("sleds should only have v6 addresses"),
+                        )
+                        .into());
+                    }
```
luqmana, line 523: Any reason not to just leave `physical_host_ip` as an `Ipv6Addr`? That's what we're getting when you look up the sled info to make this request to begin with.

jmpesp: Nope, other than `IpAddr` matches what's in OPTE:

```rust
pub struct SetVirt2PhysReq {
    pub vip: IpAddr,
    pub phys: PhysNet,
}
```

illumos-utils/src/opte/illumos/port_manager.rs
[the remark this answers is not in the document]
jmpesp: Cool, done in 9ba36eb3

illumos-utils/src/opte/illumos/port_manager.rs
jmpesp: Wait! Wrong struct.

jmpesp: changed to Ipv6Addr in f015b252

jmpesp:
@luqmana dealt with the conflicts, let me know if there's further changes you'd like to see

nexus/src/app/sagas/instance_migrate.rs
```diff
@@ -136,6 +141,40 @@ async fn sim_allocate_propolis_ip(
...
+        &sagactx,
+        &params.serialized_authn,
+    );
+    let (instance_id, _) =
+        sagactx.lookup::<(Uuid, InstanceRuntimeState)>("migrate_instance")?;
+
+    // TODO-performance the instance_put in sim_instance_migrate will *start* a
+    // migration, but the source and destination propolis servers will perform
+    // it asynchronously. Updating the mappings here will briefly "disconnect"
+    // the source instance in the sense that it will be able to send packets out
```
luqmana, line 160: Is this true? We haven't yet changed the `sled_id` for the instance which happens in the next stage (`sim_instance_migrate`). I know migration is in flux so doesn't need to be a blocker for this PR

nexus/src/app/sled.rs
```diff
@@ -237,4 +239,227 @@ impl super::Nexus {
...
+        //   necessitate a generation number for V2P mappings)
+        //
+        // - V2P mappings have to be bidirectional in order for both instances's
+        //   packets to make a round trip. This isn't a problem with the
+        //   broadcast approach because one of the sides will exist already, but
+        //   it is something to orchestrate with a more targeted approach.
+        //
+        // TODO-correctness OPTE currently will block instances in different
+        // VPCs from connecting to each other. If it ever stops doing this, the
+        // broadcast approach will create V2P mappings that shouldn't exist.
```
luqmana, line 282: AFAIK this is not OPTE by itself but rather just the way the firewall rules are setup.

nexus/src/app/sled.rs
```diff
@@ -237,4 +239,227 @@ impl super::Nexus {
...
+            let pagparams = DataPageParams {
+                marker: last_sled_id.as_ref(),
+                direction: dropshot::PaginationOrder::Ascending,
+                limit: std::num::NonZeroU32::new(10).unwrap(),
+            };
+
+            let sleds_page =
+                self.sleds_list(&self.opctx_alloc, &pagparams).await?;
+            let mut join_handles =
+                Vec::with_capacity(sleds_page.len() * instance_nics.len());
```
luqmana, line 311: yea, i'd agree with ben here and doing this as as one SQL query. Add follow up issue?

nexus/tests/integration_tests/instances.rs
```diff
@@ -2934,6 +2938,112 @@ async fn test_instance_create_in_silo(cptestctx: &ControlPlaneTestContext) {
...
+                v2p_mappings.get(&nics[0].identity.id).unwrap().len(),
+                1
+            );
+
+            // Confirm that OPTE was passed the correct information
+            let mapping = &v2p_mappings.get(&nics[0].identity.id).unwrap()[0];
+            assert_eq!(mapping.virtual_ip, nics[0].ip);
+            assert_eq!(mapping.virtual_mac, nics[0].mac);
+            assert_eq!(mapping.physical_host_ip, sled_agent.ip);
+            assert_eq!(mapping.vni, guest_nics[0].vni.clone().into());
```
luqmana, line 3017: Woo, thanks for the tests!

luqmana (commented):
Thanks for plumbing this james! Looks mostly good but I did have a few questions

nexus/src/app/sagas/instance_create.rs
```diff
@@ -115,6 +115,13 @@ declare_saga_actions! {
         + sic_add_network_config
         - sic_remove_network_config
     }
+    V2P_ENSURE_UNDO -> "v2p_ensure_undo" {
+        + sic_noop
+        - sic_v2p_ensure_undo
+    }
```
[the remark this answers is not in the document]
jmpesp: If saga node X fails, its undo action is not run. This two node structure separates out the undo action so it will run.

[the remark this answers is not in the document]
luqmana: Ah, that clears it up. Thanks!

illumos-utils/src/opte/illumos/port_manager.rs
```diff
@@ -424,6 +427,36 @@ impl PortManager {
         }
         Ok(())
     }
+
+    pub fn set_virtual_nic_host(
+        &self,
+        _interface_id: Uuid,
```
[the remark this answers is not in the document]
jmpesp: `interface_id` is required for the simulated sled agent, and sticking it in the real sled agent too makes the http entrypoints the same - Nexus can PUT or DELETE the same payload to either one. It has to remain in the HTTP endpoint of the real sled agent, however, otherwise we'll see

```
thread 'main' panicked at 'failed to register entrypoints: path parameters are not consumed (interface_id)', sled-agent/src/http_entrypoints.rs:49:9
```

when generating the openapi JSON. Removed `interface_id` propagation for the real sled agent in 7407c31b

nexus/src/app/sagas/instance_migrate.rs
jmpesp: Yeah, you're right - the comment is still true but `create_instance_v2p_mappings` needs the updated `sled_id`. I moved the v2p ensure step to after the migration start step, and updated the comment a bit in b459ce6e.

nexus/src/app/sled.rs
jmpesp: :+1: updated the comment in 05ee12ba

nexus/src/app/sled.rs
jmpesp: opened https://github.com/oxidecomputer/omicron/issues/2770

luqmana (approved):
Thanks James. This definitely improves the current state of things in allowing cross-sled communication. Still a couple TODOs left but those can be tracked in follow up issues.

PR #2546, opened 2023-03-13
Do not override Dendrite's transceiver interface SMF property

bnaecker:


Nieuwejaar (approved):
It looks good to me.  I'm not sure I have "approval" standing for `omicron`, so you may want to run it by @smklein as well.

smklein (approved):

PR #2548, opened 2023-03-14
[wicket] configure as the captive shell over ssh

sunshowers:
Per https://github.com/oxidecomputer/minimum-upgradable-product/issues/20, configure wicket as the captive shell over ssh in the switch zone.

Also add some further documentation to `how-to-run.adoc`.

TODO:

- [x] Figure out how to add a `wicket` user to the switch zone.


openapi/sled-agent.json
```diff
@@ -1505,6 +1505,20 @@
...
+                "type": "string",
+                "enum": [
+                  "switch_zone_setup"
+                ]
+              }
+            },
+            "required": [
+              "type"
+            ]
+          },
```
sunshowers, line 1521: (I had to add this by hand because sled-agent depends on sled-agent-client. We should break this cycle -- the server shouldn't depend on the client, and instead we should add a sled-agent-common that has the common types shared by the client and server.)

sled-agent/src/params.rs
```diff
@@ -476,22 +479,23 @@ impl ServiceZoneRequest {
     }
 
     // XXX: any reason this can't just be service.to_string()?
-    pub fn srv(&self, service: &ServiceType) -> SRV {
+    pub fn srv(&self, service: &ServiceType) -> Option<SRV> {
```
sunshowers, line 482: The switch zone setup script is just a oneshot script so it shouldn't have a SRV attached to it, I think. Hence the `Option<SRV>`.

sled-agent/src/services.rs
```diff
@@ -991,13 +994,16 @@ impl ServiceManager {
...
                     ServiceType::Wicketd,
+                    ServiceType::SwitchZoneSetup,
                 ]
             }
             None => {
                 vec![
                     ServiceType::ManagementGatewayService,
                     ServiceType::Dendrite { asic: DendriteAsic::TofinoAsic },
                     ServiceType::Tfport { pkt_source: "tfpkt0".to_string() },
+                    ServiceType::Wicketd,
```
sunshowers, line 1005: I added wicketd here, seems like it should always have been here.

sunshowers (commented):
Some comments.

sunshowers:
Added the usual suspects + Josh to look over the smf stuff in case I'm missing anything.

https://utcc.utoronto.ca/~cks/space/blog/solaris/SMFServiceManifestNotes was a pretty useful resource for how to set up smf, and I grepped through the transient services in `/var/svc/manifest` for examples to crib from.

jclulow:
> Added the usual suspects + Josh to look over the smf stuff in case I'm missing anything. Thanks especially to Josh for helping me out!

You are most welcome!

> https://utcc.utoronto.ca/~cks/space/blog/solaris/SMFServiceManifestNotes was a pretty useful resource for how to set up smf, and I grepped through the transient services in `/var/svc/manifest` for examples to crib from.

There are also some manual pages that cover a lot of it:

- [smf(7)](https://illumos.org/man/7/smf): service management framework overview
- [service_bundle(5)](https://illumos.org/man/5/service_bundle): describes _service bundles_, the XML-based service manifest file format
- [smf_method(7)](https://illumos.org/man/7/smf_method): service management framework conventions for methods (what happens on `start` and `stop`)
- [smf_restarter(7)](https://illumos.org/man/7/smf_restarter): service management facility conventions for restarters
- [svc.startd(8)](https://illumos.org/man/8/svc.startd): the primary SMF restarter (describes, for example, the `transient` duration)

The DTD for service bundles is also relatively well-commented.  As noted in **service_bundle(5)** it's on the system at `/usr/share/lib/xml/dtd/service_bundle.dtd.1`.  You can also see it [in the source tree](https://raw.githubusercontent.com/illumos/illumos-gate/master/usr/src/cmd/svc/dtd/service_bundle.dtd.1).

docs/how-to-run.adoc
```diff
@@ -134,17 +134,31 @@ $ pfexec ./target/release/omicron-package install
...
+The first time the bootstrap service runs, it will take a while to initialize
+the Omicron zones:
 
 [source,text]
 ----
 # List all services:
 $ svcs
+
+# View logs for sled-agent:
+$ tail -f /var/svc/log/system-illumos-sled-agent:default.log
```
jclulow, line 152: I would instead make this:

```
$ tail -F $(svcs -L sled-agent)
```

First, the `tail -F` flag should handle truncation/rotation of the log, which can happen from time to time:

```
       -F
              Follow named file. Operates as with -f, except that if the file
              is moved (e.g. if a watched log file is rotated) tail will close
              the original file and begin reading records from the start of
              the file with the specified name if and when that file is
              recreated.
```

The `-L` argument to [svcs(1)](https://illumos.org/man/1/svcs) emits the path to the log file for a service instance:

```
       -L
                           Display the log file of the selected services and
                           service instances, one per-line.
```

docs/how-to-run.adoc
```diff
@@ -134,17 +134,31 @@ $ pfexec ./target/release/omicron-package install
...
+----
+
+Once the zones are initialized, they'll show up in `zoneadm`:
+
+[source,text]
+----
 # View zones managed by Omicron (prefixed with "oxz_"):
 $ zoneadm list -cv
```
jclulow, line 160: I would add the `-n` flag here (so, `zoneadm list -cnv`) as it leaves out the `global` zone line and is probably a bit less confusing to newcomers as a result.

smf/switch_zone_setup/manifest.xml
```diff
@@ -0,0 +1,35 @@
+<?xml version="1.0"?>
+<!DOCTYPE service_bundle SYSTEM "/usr/share/lib/xml/dtd/service_bundle.dtd.1">
+
+<service_bundle type='manifest' name='switch_zone_setup'>
+
+<service name='system/illumos/switch_zone_setup' type='service' version='1'>
+  <create_default_instance enabled='false' />
+  <single_instance />
```
jclulow, line 8: We should omit `<single_instance/>` here.  I know it's common in manifests that ship with the OS, but that's some historical detritus we ought to clean up.  I have also just filed an illumos bug about that: [15477 <single_instance> in service bundles is obsolete](https://www.illumos.org/issues/15477).

smf/switch_zone_setup/manifest.xml
```diff
@@ -0,0 +1,35 @@
+<?xml version="1.0"?>
+<!DOCTYPE service_bundle SYSTEM "/usr/share/lib/xml/dtd/service_bundle.dtd.1">
+
+<service_bundle type='manifest' name='switch_zone_setup'>
+
+<service name='system/illumos/switch_zone_setup' type='service' version='1'>
+  <create_default_instance enabled='false' />
```
jclulow, line 7: Why is this not enabled by default?  I would expect `true` here, so that the manifest just starts when the zone boots up.

smf/switch_zone_setup/manifest.xml
```diff
@@ -0,0 +1,35 @@
...
+  <single_instance />
+
+  <dependency name='multi_user' grouping='require_all' restart_on='none'
+    type='service'>
+    <service_fmri value='svc:/system/filesystem/local:default' />
```
jclulow, line 12: I think you can remove this `filesystem/local` dependency here.  The `multi-user` milestone you're depending on is relatively coarse-grained, and itself always depends on having the local file system available before it is online.

jclulow (changes requested):
Hey, I don't currently understand the omicron/sled-agent bits, but I have looked at the SMF manifest and method parts, etc.

Cheers!

smf/switch_zone_setup/manifest.xml
sunshowers: So the current model is that all the services start disabled and that sled-agent's bootstrap process imports them, provides inputs (e.g. specifies ports), and then enables them. For this script specifically there are no inputs that sled-agent needs to provide, so we could go either way, but I was thinking about uniformity with the other services. What do you think?

sunshowers: Note that even if we set `enabled='true'` the manifest isn't automatically imported, so the script won't run when the zone boots up. Instead, sled-agent imports the manifest explicitly.

jclulow: By placing it in `/var/svc/manifest` it actually will be imported automatically by the `manifest-import` service.

Because it's a short term (touch wood) workaround for a missing feature in the brand I'd be inclined not to make sled agent aware of it if possible.  It's also in line with where we _want_ to be, which is that sled agent doesn't reach into the zone to work around with SMF; i.e., that zone images are as self assembling as possible.

openapi/sled-agent.json
sunshowers: Will do this in a followup.

docs/how-to-run.adoc
sunshowers: Thanks, I keep forgetting about `tail -F`, it's always been super handy.

docs/how-to-run.adoc
sunshowers: Done.

smf/switch_zone_setup/manifest.xml
sunshowers: Done, thanks. I also took the liberty of removing all the `<single_instance/>` from the other manifests.

smf/switch_zone_setup/manifest.xml
sunshowers: OK, so we discussed this in #oxide-q&a on Matrix, and here's what we came to understand:

1. I ran into what seemed like an issue once where the `switch_zone_setup` script appeared not to start. I thought that it meant that manifests had to be imported explicitly.
2. My incorrect belief was reinforced by me finding the bootstrap code which imports all the site-specific manifests. In reality, the bootstrap code doesn't 

I removed all the sled-agent code I'd added in the earlier version and tried again. In 5 separate tries I was unable to reproduce my issue from earlier: each time, the setup script automatically ran in the switch zone and the wicket user was created.

I'm going to chalk this up to me screwing up the manifest.xml somehow on the one try that I did where the manifest didn't get imported, though I guess that there's a small possibility there's a nondeterministic bug somewhere in the system.

smf/switch_zone_setup/manifest.xml
sunshowers: Done, thanks.

sunshowers:
The latest update to this PR adds the support user to the switch zone. We need this user for development and debugging in the very immediate term.

The user is currently unauthenticated, but we will lock it down in the very near future.

andrewjstone:
> The latest update to this PR adds the support user to the switch zone, as discussed in today's update sync meeting. We need this user for development and debugging in the very immediate term.
> 
> The user is currently unauthenticated, but we will lock it down in the very near future. (Adam and I have discussed this plan.)

Thank you @sunshowers !

PR #2552, opened 2023-03-14
[wicket] Consistent key management

andrewjstone:
In order to be more consistent with keys across panes, a `KeyHandler` was added that processes all key presses and returns `Cmd`s. Raw key events are no longer sent to `Control`s directly. Instead `Cmd`s are handled in `Control::on`. This results in a more consistent experience, and also puts all key-handlign in one place. There is one exception to this, which is when the user needs to enter raw data in a text box. For this we allow a `Cmd::Raw` variant, using the vim command `i` to trigger it. We don't currently do this as we don't have any need for it. We are preparing for the future though.

Additionally, this change moves out scroll functionality into its own function in preparation for scrollable popups.


sunshowers (approved):
lgtm!

PR #2553, opened 2023-03-14
Remove deprecated `/system/policy`

zephraph:
Part of #2028, this removes the deprecated `/system/policy` endpoints in favor of `/v1/system/policy`


david-crespo (approved):
burn it down

PR #2555, opened 2023-03-14
[wicket] Add live update status for artifacts

andrewjstone:
* Parse UpdateLogAll to update state
* Start some small styling changes based on discussion Friday. More to come.

### DIFF
--- a/wicket/src/state/update.rs
+++ b/wicket/src/state/update.rs
@@ -248,7 +248,9 @@ fn update_artifact_state(
     new_state: UpdateState,
 ) {
     if let Some(known) = &known {
-        items.get_mut(known).map(|state| *state = new_state);
+        if let Some(state) = items.get_mut(known) {
+            *state = new_state;
+        }
     }
 }
 
--- a/wicket/src/state/update.rs
+++ b/wicket/src/state/update.rs
@@ -6,17 +6,23 @@ use tui::style::Style;
 
 use crate::ui::defaults::style;
 
-use super::{ComponentId, ALL_COMPONENT_IDS};
+use super::{ComponentId, ParsableComponentId, ALL_COMPONENT_IDS};
 use omicron_common::api::internal::nexus::KnownArtifactKind;
+use slog::{o, warn, Logger};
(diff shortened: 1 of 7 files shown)


wicket/src/state/update.rs
```diff
@@ -103,6 +127,162 @@ impl RackUpdateState {
...
+                    }
+                    UpdateNormalEventKind::InstallinatorEvent(
+                        _completion_event_kind,
+                    ) => {
+
+                        // TODO: Do we want to update state
+                        // with intermediate events?
+                        //
+                        // This would be useful for reporting progress
+                        // percentages, but that's about it.
```
sunshowers, line 175: Can we represent "downloaded" and "installed" as separate events here? That would fit nicely here then, I think

wicket/src/state/update.rs
```diff
@@ -103,6 +127,162 @@ impl RackUpdateState {
...
+            }
+            SettingHostPowerState { .. }
+            | InstallinatorFormatProgress { .. }
+            | SettingInstallinatorOptions
+            | SettingHostStartupOptions => {
+                // Should we bother doing something here?
+                None
+            }
+        };
+        update_artifact_state(items, known, UpdateState::Updating);
```
sunshowers, line 239: Per discussion, filed https://github.com/oxidecomputer/omicron/issues/2578 to show a progress percentage here.

sunshowers (approved):

wicket/src/state/update.rs
andrewjstone: Maybe. One problem is that this doesn't really work for the Sp and RoT which write while they are downloading. It's all streaming. I'll merge this through and we can discuss in a follow up. There's also what @jgallagher suggested which is in the status column to indicate for sleds which thing was being updated. I didn't do that just due to the current structure of the code where `UpdateState` is an enum and all states are currently valid for all entities. But that could change and may line up with this here.

wicket/src/state/update.rs
andrewjstone: Thank you!

PR #2564, opened 2023-03-15
Remove deprecated silo user endpoints

zephraph:
Part of #2028


david-crespo (approved):
just needs openapi spec regen

PR #2570, opened 2023-03-15
DNS servers should use generation-based API

davepacheco:
This is a pretty big change.  Sorry.  But it's a little smaller than it appears because a lot of stuff was just moved around.   Here's a high-level summary of changes.  Definitely read this before diving in:

DNS server (@rcgoodfellow I'd love to get your review on these parts):

- Between all the changes below, the DNS server was mostly rewritten.  (It wasn't that big to begin with.) It's probably easiest to review it as all new code.  If you want to know what changed or why, read the rest of these bullets.  Otherwise, feel free to skip these.
- The Dropshot API for configuring the names served over DNS is now generation-oriented as described in RFD 367.  This was the main goal of the PR.  The big change is that instead of having APIs to add/remove individual DNS names, you provide all the DNS names at once, with an associated generation number, and the server atomically moves to that generation as long as it's newer than its current generation.  For more on why this approach, see the comment in http_server.rs or RFD 367.
- As part of this, I also made it so that the DNS server can serve any number of DNS zones, all specified in that same update request body.  (I initially did this because I wanted the DNS zone name to be runtime-configurable, since I didn't think we'd necessarily know the zone name when we initially deploy the external DNS servers.  I think we probably _do_, but it seems worth keeping this.)
- In terms of the library interface (`dns_server` crate): the crate now exposes three pieces: the `storage::Store` for the persistent store (still based on the `sled` crate), the `dns_server` module for the DNS protocol server, and the `http_server` module for the Dropshot server.  Most consumers create their own `Store` and use a top-level helper function called `start_servers()` to start the DNS and Dropshot servers.
- `Store` now encapsulates everything related to the DNS server's persistent state.  Previously, the Dropshot server communicated over a channel to a task that would update the Sled DB, and the DNS part would read directly out of Sled.  Now, this is all encapsulated in the Store.  There's no channel or separate task needed because Sled does its own synchronization.
- As a result, much of the DNS server part was refactored in terms of higher-level operations provided by the `Store`.  On the plus side, I think it's a fair bit clearer, and many (small) code paths are shared that were previously duplicated.
- The `dnsadm` tool still exists.  This was a little tricky because it provides subcommands to add/remove individual DNS records, and these are undoubtedly useful for testing, but not supported on the server any more.  Now, `dnsadm` implements these by fetching the current config, modifying it, and updating it again.  But this made me worry because it would be grossly unsafe on a production system (since CockroachDB must be the sole source of truth about what each generation number means).  So this tool will now only make writes to DNS zones ending in ".oxide.test" so that we can never accidentally make modifications in production with it.

Crate renaming and refactoring (@smklein I thought maybe you'd be a good person to review this?):

- Renamed `internal-dns-names` to `internal-dns` and moved a bunch of `dns-service-client` into `internal-dns`.  This is where the biggest churn in this PR came from.  Recall that we had `dns-service-client` (which is intended to provide both HTTP and DNS clients for our own DNS servers, independent of whether they're for internal or external DNS) and `internal-dns-names` (which had a bunch of well-known names for the _internal_ DNS zone).  But a lot of the stuff in `dns-service-client` was still pretty specific to the internal DNS zone.  So I renamed `internal-dns-names` to `internal-dns` and moved a bunch of the internal DNS zone-specific stuff out of `dns-service-client` and into `internal-dns`.  This included the resolver and the stuff that replaced the updater.
- Having done that, the pieces of `internal-dns` got split into separate modules for clarity.
- Relatedly, the DNS Updater was removed altogether.  It's replaced with `DnsConfigBuilder` plus a direct call to each DNS server's progenitor client.  Before, this object allowed its consumer to make a best-effort to update all DNS servers concurrently, either adding or removing some names.  This essentially comprised two pieces: (1) constructing the new DNS data (now done by `DnsConfigBuilder`), and (2) propagating it to the DNS servers (currently just a direct call to each DNS server).  For propagation, the right answer will eventually involve a reliable persistent workflow in Nexus (see RFD 267), but the only consumer today is RSS (and testing code) so I think it makes sense to have RSS explicitly update each server.

RSS (in the Sled Agent) (also @smklein?):
- Previously, RSS updated DNS entries _ad hoc_ after deploying datasets and services.  Now, it creates the entire generation-1 DNS configuration when it generates the service plan.  This configuration is written to all the internal DNS servers right after we deploy them.

Some minor stuff:
- added logging to the Resolver (needed this to debug some helios-deploy stuff)
- added a lot of documentation, including to the dns-server CLI tools
- added a dns-db-dump tool for dumping the contents of the Sled database

Sorry again that this is so large.  I wasn't sure how to split it better into smaller pieces.

Current status: I think this PR is basically ready for review.  It's missing a bunch of tests I still want to write.  But I thought it'd be worth getting feedback at this point.


dns-server/tests/basic_test.rs
```diff
@@ -206,59 +208,124 @@ async fn lookup_ip_expect_nxdomain(resolver: &TokioAsyncResolver, name: &str) {
...
     assert!(records.is_empty());
 
     // Add an empty DNS record
-    let name = DnsRecordKey { name: "devron.oxide.internal".into() };
-    client
-        .dns_records_create(&vec![DnsKv { key: name.clone(), records: vec![] }])
-        .await?;
-
-    // read back the aaaa record
-    let records = client.dns_records_list().await?;
```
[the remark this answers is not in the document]
rcgoodfellow: The intent behind this test was to ensure that we return NXDOMAIN even in the case where the name exists but the corresponding record set is empty. I wanted to make sure that if for some reason all the records for a name were cleared out, but the name was still present in the KV store we were still returning NXDOMAIN to indicate that the name has no records whatsoever. This is helpful in preventing resolvers for iterating over subsequent nameservers.

In the new design this is much less likely to happen as incremental record updates are not present.

dns-server/src/dns_server.rs
```diff
@@ -22,191 +31,244 @@ use trust_dns_proto::rr::{Name, Record};
...
 
-    let mut response_records: Vec<Record> = vec![];
-    for record in &records {
-        let resp = match record {
+    let header = Header::response_from_request(mr.header());
+    let query = mr.query();
+    let name = query.original().name().clone();
+    let records = store.query(mr)?;
+    let rb = MessageResponseBuilder::from_message_request(mr);
+    let response_records = records
```
rcgoodfellow, line 228: If this result comes back empty, we should probably check if the logical DNS tree rooted at the queried name is also empty, and if so return an NXDOMAIN per RFCs 1034 and 8020. I could see this happening a few different ways.

1. If we stored the keys (names) in reverse, we could use sled's `scan_prefix` under the hood to do a prefix check.
2. We look at the query and see if the name extends beyond the zone prefix, and if so, just assume this is a terminal name and that no names exist below it. I think this may work with the constrained way we are using DNS, but I could be missing something.

dns-server/src/dns_server.rs
```diff
@@ -22,191 +31,244 @@ use trust_dns_proto::rr::{Name, Record};
...
+        .into_iter()
+        .map(|record| match record {
```
rcgoodfellow, line 230: For external DNS, we'll need to handle `A` and `NS` records too.

dns-server/src/storage.rs
```diff
@@ -0,0 +1,899 @@
...
+                attempted_generation: config.generation,
+            });
+        }
+        if old_config.generation == generation {
+            return Ok(());
+        }
+
+        // Prune any trees in the db that are newer than the current
+        // generation.
+        self.prune_newer(&old_config);
```
rcgoodfellow, line 385: Should this be possible given the check at line 373 above?

rcgoodfellow (commented):
Hey @davepacheco I think the changes here make a lot of sense for the role that DNS is playing at a system/infrastructure level of the Oxide platform. The generational atomic updates and more constrained/focused lens we're now viewing DNS through simplifies things, making it easier to reason about overall DNS behavior at a system level.

When we get to DNS for instances or provide a DNS API for users/operators to modify records, I think we'll want to consider an approach that allows for incremental updates. But that's a problem for future us, and I don't think it changes the approach taken here for naming Oxide endpoints.

dns-server/src/dns_server.rs
```diff
@@ -22,191 +31,244 @@ use trust_dns_proto::rr::{Name, Record};
...
     let header = Header::response_from_request(mr.header());
-    let zone = LowerName::from(Name::from_str(&zone).unwrap());
-
-    // Ensure the query is for this zone, otherwise bail with servfail. This
-    // will cause resolvers to look to other DNS servers for this query.
-    let name = mr.query().name();
-    if !zone.zone_of(name) {
-        nack(&log, &mr, &socket, &header, &src).await;
-        return;
+    let rb_nxdomain = MessageResponseBuilder::from_message_request(&mr);
```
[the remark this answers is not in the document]
davepacheco: Done in e86747334326bb001eb6f55f731357a47ee4f96c.

dns-server/src/dns_server.rs
davepacheco: Good call.  I believe this case is handled already, if somewhat implicitly.  The contract of `query()` is that it returns a non-empty list of records or else an error.  This wasn't documented, though.  I've documented it in 2ffdb9e8.

More specifically: if the name were not found in the tree at all, `query()` would return `QueryError::NoName` here:
https://github.com/oxidecomputer/omicron/blob/e86747334326bb001eb6f55f731357a47ee4f96c/dns-server/src/storage.rs#L636

The storage implementation never puts a name in the tree with an empty list of records, but it handles the case where it finds that by returning `QueryError::NoName` instead:
https://github.com/oxidecomputer/omicron/blob/e86747334326bb001eb6f55f731357a47ee4f96c/dns-server/src/storage.rs#L644-L656

So we should always get `QueryError::NoName` for this case, and that gets translated to NXDOMAIN:
https://github.com/oxidecomputer/omicron/blob/2ffdb9e8a222e2a43aeb8a62a958ab6088c96d96/dns-server/src/dns_server.rs#L206

I think this corresponds to your suggestion 2.  Thanks for mentioning it though because this was rather implicit.  (I could make it even more explicit by having `query()` return not a `Result`, but an enum that includes the various errors plus a `NonEmptyRecords(Vec<DnsRecord>)` variant.  But that's more unwieldy -- you can't use `?` to return early and convert it to the DNS error, for example.)

dns-server/src/dns_server.rs
davepacheco: 💯  I'd like to put this in a follow-up PR since this one is already so large.

dns-server/src/storage.rs
davepacheco: This is intended to handle the case that we previously attempted to update to this generation but crashed while doing it, leaving some possibly incomplete trees around (and never having updated the config to reflect the new generation).  This approach wipes whatever we found and starts the update again.  I updated this comment to explain that case, since it's pretty non-obvious.

davepacheco:
Thanks @rcgoodfellow for taking a look!  I'm glad you're on board with the general direction here.

> Hey @davepacheco I think the changes here make a lot of sense for the role that DNS is playing at a system/infrastructure level of the Oxide platform. The generational atomic updates and more constrained/focused lens we're now viewing DNS through simplifies things, making it easier to reason about overall DNS behavior at a system level.
> 
> When we get to DNS for instances or provide a DNS API for users/operators to modify records, I think we'll want to consider an approach that allows for incremental updates. But that's a problem for future us, and I don't think it changes the approach taken here for naming Oxide endpoints.

Agreed with all that, especially that this is a future problem. 

Here are my thoughts on it:

- Even when the DNS data and DNS updates scale up, it will still be important to associate each version of the DNS tree with a generation number for the same reasons as now: to ensure data integrity and for use as a checkpoint in the (never-ending) stream of incremental updates.
- I think this basic approach can support incremental updates without too much work, by having a separate API that updates from generation X to generation Y by providing a list of names added or removed.  RFD 367 (section 4.2.2) talks about how Nexus might construct such an update.  The harder part would be doing that atomically on the storage side, but there's lots of flexibility here since we don't have to worry about on-disk compatibility.  (An idea: similar to what RFD 367 proposes for Nexus/CockroachDB, we could associate with each name an "introduced in" and "removed in" generation number.  That would let us modify the tree while we're still reading from it without affecting what's visible to end users until we update our top-level generation number.)
- All that said, I was pretty compelled by the arguments in [this article about "constant-work systems"](https://aws.amazon.com/builders-library/reliability-and-constant-work/) to just do the full update all the time.  The downside is that at high rates of update, we'd probably batch up the updates, which could increase latency from change to propagation.  Basically, this approach doesn't scale _down_ as well, but it might be more predictable and reliable.

Thanks again for the review!

dns-server/src/dns_server.rs
rcgoodfellow: Sounds good.

dns-server/src/dns_types.rs
```diff
@@ -0,0 +1,58 @@
...
+
+#[derive(Clone, Debug, Serialize, Deserialize, JsonSchema)]
+#[serde(rename = "DnsKv")]
+pub struct DnsKV {
+    pub key: DnsRecordKey,
+    pub records: Vec<DnsRecord>,
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize, JsonSchema, PartialEq)]
+pub struct DnsRecordKey {
```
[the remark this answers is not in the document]
rcgoodfellow: I believe these extra data structures were to make client interactions easier through more structured serialization/deserialization. I would have no objection to squashing them down to simpler representations.

dns-server/src/dns_server.rs
davepacheco: Is that notation in reversed order?  For the external endpoints, I imagine the customer will delegate us something like `oxide.corp.globex.example`, they'll create Silo `prod` (say), and we'll have records for `prod.sys.oxide.corp.globex.example`.  There won't be anything at `sys.oxide.corp.globex.example`.  There will presumably be NS records for `oxide.corp.globex.example`.  I suspect this violates the assumption you said.  I'm not sure what the impact is?  What's the right DNS response for `sys.oxide.corp.globex.example`?  A normal response with zero records?

As an aside: I think the behavior here hasn't changed in this PR.  (Is that wrong?)

rcgoodfellow: Yes, this is in reverse order. Sorry, my brain was still thinking in DNS names as trees, and I was starting from the root.

So if we have no records at `sys.oxide.corp.globex.example` but a record logically below that in the tree `prod.sys.oxide.corp.globex.example` exists, the correct response for `sys.oxide.corp.globex.example` would be a normal response with no records.

I probably got this wrong in my initial implementation too.

sled-agent/src/rack_setup/plan/service.rs
```diff
@@ -319,7 +382,8 @@ impl Plan {
             services.insert(addr, allocation);
         }
 
-        let plan = Self { services };
+        let dns_config = dns_builder.build();
```
smklein, line 385: This new format looks good to me - as mentioned above, should we be adding the sled underlay addresses themselves to these initial DNS records?

smf/external-dns/manifest.xml
```diff
@@ -12,15 +12,13 @@
...
   <exec_method type='method' name='start'
-      exec='ctrun -l child -o noorphan,regent /opt/oxide/dns-server/bin/dns-server --config-file /var/svc/manifest/site/external_dns/config.toml --server-address %{config/server_address} --dns-address %{config/dns_address} --dns-zone %{config/dns_zone} &amp;'
+      exec='ctrun -l child -o noorphan,regent /opt/oxide/dns-server/bin/dns-server --config-file /var/svc/manifest/site/external_dns/config.toml --http-address %{config/http_address} --dns-address %{config/dns_address} &amp;'
     timeout_seconds='0' />
   <exec_method type='method' name='stop' exec=':kill' timeout_seconds='0' />
 
-  <!-- TODO-completeness This all needs to be run-time-configurable -->
   <property_group name='config' type='application'>
-    <propval name='server_address' type='astring' value='unknown' />
+    <propval name='http_address' type='astring' value='unknown' />
```
smklein, line 20: Thanks; this is a much clearer name

internal-dns/src/config.rs
```diff
@@ -0,0 +1,690 @@
...
+    /// name
+    pub(crate) fn dns_name(&self) -> String {
+        match &self {
+            AAAA::Sled(id) => format!("{}.sled", id),
+            AAAA::Zone(id) => format!("{}.host", id),
+        }
+    }
+}
+
+/// Builder for assembling DNS data for the control plane's DNS zone
```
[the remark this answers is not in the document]
smklein: Is the intent that Nexus uses this interface to perform updates when it manages services?

If that's the case, do we need a mechanism for building `DnsConfigParams` with a generation other than `1`?

smklein (commented):
As documented in 367, it seems like this is going to shift the DNS server interface from being an "incremental" (add + remove) interface to a more "absolute" one (PUT everything you want).

Given the correctness constraints we need w.r.t. concurrency and crash safety, this seems reasonable to me.

- The refactors of the `internal-dns` crate LGTM
- The RSS changes also LGTM, with some comments in-line

internal-dns/src/config.rs
```diff
@@ -0,0 +1,690 @@
...
+    }
+
+    /// Specify that service `service` has a backend instance running directly
+    /// on the given (sled) host (generally in the sled's global zone)
+    ///
+    /// # Errors
+    ///
+    /// This function fails only if the given host has already been added as a
+    /// backend for this service.
+    pub fn service_backend_sled(
```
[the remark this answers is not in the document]
davepacheco: I think it's only used in tests today, but it's here because at least in principle, in the spirit of RFD 248, each sled agent should be its own (SRV) service with exactly one backend:

> For something like Sled Agent, each Sled Agent would be its own service with exactly one backend, since the Sled Agents cannot be treated as interchangeable with each other.

(from RFD 248)

Today, we're not defining those SRV services at all.  This may not be a big deal.  In Nexus, this works because Nexus bypasses DNS to locate sled agents.  It uses the IP address and port stored in the database.  I'm not sure who else uses the Sled Agent client and how they look it up.

Anyway, to your question: this is here to eventually support defining an SRV service for each sled agent.

internal-dns/src/config.rs
```diff
@@ -0,0 +1,690 @@
...
+
+/// Builder for assembling DNS data for the control plane's DNS zone
+///
+/// `DnsConfigBuilder` provides a much simpler interface for constructing DNS
+/// zone data than using `DnsConfig` directly.  That's because it makes a number
+/// of assumptions that are true of the control plane DNS zone (all described in
+/// RFD 248), but not true in general about DNS zones:
+///
+/// - We assume that there are only two kinds of hosts: a "sled" (an illumos
+///   global zone) or a "zone (an illumos non-global zone).  (Both of these are
```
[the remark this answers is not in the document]
davepacheco: Nice catch.  Fixed in fcb109ab.

internal-dns/src/config.rs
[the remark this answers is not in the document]
davepacheco: I didn't really address the second part of your question: I believe that we do already add a AAAA record for the sled and that it's the address on the underlay network...but now I'm realizing maybe I'm wrong about which network that one is on.

dns-server/src/dns_server.rs
rcgoodfellow: SGTM

rcgoodfellow (approved):
On the more DNS-y side of things, looks good to me.

PR #2571, opened 2023-03-15
[wicketd] update the RoT, change order of updates

sunshowers:
* Update the RoT before updating the SP.
* Change the order of updates to RoT -> SP -> host.

TODO:

- [x] Figure out the running slot (A or B). From Andrew:
  > What needs to be done is to check the active slot that is in the RotBootState and then send the update to the other one. The other tricky bit about this is that there are different A and B images, since the code is not position independent. They are actually built with absolute flash addresses so they only run out of their given slots. I added a check in the lpc55_update_server that looks at the interrupt table and ensures the addresses are within the right range, and so you wont' break anything if you send an image to the wrong slot, but you will get an error back.

  **Update:** we decided to make MGS do this determination, not wicketd. However, wicketd needs to be able to tell wicket which slot's currently being updated, so we need to add support for that to the messages MGS sends wicket.
- [x] Pull apart the RoT image into its A and B components
  
  **Update:** MGS will do this.

- (Not in this PR) Check version numbers of installed components and don't update if they're the same as what we're looking for. This is still waiting on other folks.


wicketd/src/update_events.rs
```diff
@@ -53,7 +53,7 @@ pub enum UpdateStateKind {
     ArtifactWriteProgress {
         attempt: usize,
         kind: ArtifactKind,
-        #[schemars(schema_with = "path_schema")]
+        #[schemars(schema_with = "path_opt_schema")]
```
sunshowers, line 56: This was a bad bug, whoops!

wicketd/tests/integration_tests/updates.rs
```diff
@@ -72,12 +73,31 @@ async fn test_updates() {
...
+            if let UpdateEventKind::Terminal(event) = event.kind {
+                break 'outer event;
+            }
+        }
+    };
+
+    match terminal_event {
+        UpdateTerminalEventKind::ArtifactUpdateFailed { artifact, .. } => {
+            // TODO: obviously we shouldn't stop here, get past more of the
+            // update process in this test.
```
sunshowers, line 94: I've set this up and it's been a lot of work, but also I don't feel like going further down this path is the highest priority for now.

There's a lot of work that needs to be done here to make this test behave more like the real system. For example, the test currently fails with "updating SP failed: failed to parse SP update as a zip file: invalid Zip archive: Could not find central directory end", because the fake SP artifact doesn't look like a zip file.

Maybe we can make this test use a real artifact downloaded over HTTP or similar, or extend the fake artifact to work for this, but for now I think we should just move on and instead rely on manual tests with real systems. What do you think?

An outcome that might be worth pursuing is creating a binary that runs all of wicketd, MGS and sp-sim within the same process, so the full flow is easier to test.

andrewjstone (approved):
I think it's fine to merge as is and move on. Up to you what you want to do @sunshowers.

wicketd/src/update_tracker.rs
```diff
@@ -949,9 +971,12 @@ impl UpdateDriver {
...
 
-    async fn update_sp(&self, artifact: &ArtifactIdData) -> anyhow::Result<()> {
-        const SP_COMPONENT: &str = SpComponent::SP_ITSELF.const_as_str();
-
+    async fn update_sp_component(
+        &self,
+        artifact: &ArtifactIdData,
+        component: SpComponent,
+    ) -> anyhow::Result<()> {
+        let component_name = component.const_as_str();
```
jgallagher, line 979: This feels a tiny bit dicey, because `const_as_str` can panic if `component` is malformed. Its docs say 

```
    /// This function should only be used in const contexts when the caller
    /// knows the component is valid (e.g., one of this type's associated
    /// constants); for component names parsed or constructed at runtime, prefer
    /// [`SpComponent::as_str()`] which performs runtime validation.
```

In this case, we're being called with only constants defined in `SpComponent` so everything is fine, but it's not obvious locally that that's the case - we have to check all the callers of this function to know. I'd suggest one of two options:

1. Call `.as_str()` instead and just return an error if a component is invalid
2. Change the function to take `component: &str` and call `.const_as_str()` at each call site where it's obviously correct

wicketd/tests/integration_tests/updates.rs
jgallagher: Agree with both of you

wicketd/src/update_tracker.rs
```diff
@@ -394,25 +394,43 @@ impl UpdateDriver {
...
-        // RoT, then reboot the SP/RoT?
-        if self.sp.type_ == SpType::Sled {
-            self.run_sled(&plan, ipr_start_receiver).await?;
-        }
-
-        let sp_artifact = match self.sp.type_ {
-            SpType::Sled => &plan.gimlet_sp,
-            SpType::Power => &plan.psc_sp,
-            SpType::Switch => &plan.sidecar_sp,
+        // TODO: We currently do updates in the order RoT -> SP -> host. This is
```
andrewjstone, line 397: Great comment :)

jgallagher (approved):

wicketd/tests/integration_tests/updates.rs
sunshowers: Thanks both of you -- this was a good learning experience but I'm glad I was able to timebox it and move on.

wicketd/src/update_tracker.rs
sunshowers: Yeah that makes sense. I'll probably just switch to 2.

sunshowers: Done.

PR #2581, opened 2023-03-16
Project integration tests use v1 endpoints

david-crespo:


david-crespo:
Moved commits into #2582

PR #2582, opened 2023-03-16
Update tests to have V1 references

zephraph:
This PR pulls test updates out of several of the deprecation PRs as it's pre-mergable and will reduce overall conflicts.

### DIFF
--- a/nexus/tests/integration_tests/basic.rs
+++ b/nexus/tests/integration_tests/basic.rs
@@ -67,12 +67,9 @@ async fn test_basic_failures(cptestctx: &ControlPlaneTestContext) {
         // appropriate?
         TestCase {
             method: Method::GET,
-            uri: "/organizations/test-org/projects/-invalid-name",
-            // TODO
-            // uri: "/v1/projects/-invalid-name?organization=test-org",
+            uri: "/v1/projects/-invalid-name?organization=test-org",
             expected_code: StatusCode::BAD_REQUEST,
-            expected_error: "bad parameter in URL path: name must begin with \
-            an ASCII lowercase character",
+            expected_error: "bad parameter in URL path: data did not match any variant of untagged enum NameOrId",
             body: None,
         },
         TestCase {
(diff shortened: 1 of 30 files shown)


nexus/tests/integration_tests/basic.rs
```diff
@@ -54,55 +54,53 @@ async fn test_basic_failures(cptestctx: &ControlPlaneTestContext) {
...
 
-        // Error case: GET /organizations/test-org/projects/nonexistent (a
-        // possible value that does not exist inside a collection that does
-        // exist) from an authorized user results in a 404.
+        // Error case: a possible value that does not exist inside a collection
+        // that does exist) from an authorized user results in a 404.
         TestCase {
             method: Method::GET,
-            uri: "/organizations/test-org/projects/nonexistent",
+            uri: "/v1/projects/nonexistent?organization=test-org",
```
david-crespo, line 61: This error message is different now that we take name or ID. Can't say `data did not match any variant of untagged enum NameOrId` is particularly nice, but it does suggest that the thing you passed is not a name or ID. I'm thinking I should change the assert here to the new message and make an issue to improve it. @zephraph?

```
thread 'integration_tests::basic::test_basic_failures' panicked at 'assertion failed: `(left == right)`
  left: `"bad parameter in URL path: name must begin with an ASCII lowercase character"`,
 right: `"bad parameter in URL path: data did not match any variant of untagged enum NameOrId"`', nexus/tests/integration_tests/basic.rs:155:9
```

nexus/tests/integration_tests/basic.rs
```diff
@@ -54,55 +54,53 @@ async fn test_basic_failures(cptestctx: &ControlPlaneTestContext) {
...
             body: None,
         },
-        // Error case: GET /organizations/test-org/projects/-invalid-name
         // TODO-correctness is 400 the right error code here or is 404 more
         // appropriate?
         TestCase {
             method: Method::GET,
             uri: "/organizations/test-org/projects/-invalid-name",
+            // TODO
+            // uri: "/v1/projects/-invalid-name?organization=test-org",
```
david-crespo, line 72: Sorry, made this comment on the wrong line.

This error message is different now that we take name or ID. Can't say `data did not match any variant of untagged enum NameOrId` is particularly nice, but it does suggest that the thing you passed is not a name or ID. I'm thinking I should change the assert here to the new message and make an issue to improve it. @zephraph?

```
thread 'integration_tests::basic::test_basic_failures' panicked at 'assertion failed: `(left == right)`
  left: `"bad parameter in URL path: name must begin with an ASCII lowercase character"`,
 right: `"bad parameter in URL path: data did not match any variant of untagged enum NameOrId"`', nexus/tests/integration_tests/basic.rs:155:9
```

david-crespo:
Got all the ones that start with `/organizations` but there are a few `/system` ones left.

PR #2585, opened 2023-03-16
[wicketd] start adding integration tests with MGS and sp-sim

sunshowers:
Start adding integration tests, starting with a really basic test to
fetch the inventory.


andrewjstone (approved):
Looks good!

PR #2590, opened 2023-03-17
RFD 322: v1 built-in users

david-crespo:
Roles don't have IDs exposed in the responses so it doesn't make sense to make them fetchable by ID.

### DIFF
--- a/nexus/src/external_api/http_entrypoints.rs
+++ b/nexus/src/external_api/http_entrypoints.rs
@@ -331,6 +331,9 @@ pub fn external_api() -> NexusApiDescription {
         api.register(role_list)?;
         api.register(role_view)?;
 
+        api.register(role_list_v1)?;
+        api.register(role_view_v1)?;
+
         api.register(session_sshkey_list)?;
         api.register(session_sshkey_view)?;
         api.register(session_sshkey_create)?;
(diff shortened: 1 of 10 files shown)


nexus/src/external_api/http_entrypoints.rs
```diff
@@ -8890,6 +8953,75 @@ async fn role_view(
     apictx.external_latencies.instrument_dropshot_handler(&rqctx, handler).await
 }
 
+/// List built-in roles
+#[endpoint {
+    method = GET,
+    path = "/v1/roles",
```
zephraph, line 8959: So this was a failure on my part when introducing `/system`. This resource is fleet scoped and should be moved under `/system`. Also, given that both `roles` and `users` apply to builtin we could do `/v1/built-in/users` and `/v1/built-in/roles`. I'd also like to find a term better than `built-in`, but that can be deferred.

zephraph: See the fleet scoping of roles here: https://github.com/oxidecomputer/omicron/blob/369af0ee7ab0bf70dc47422126804b19117dd850/nexus/db-queries/src/db/datastore/role.rs#L43

david-crespo: Definitely on `/system`, that makes perfect sense. Not sure about `/built-in/...`. It is clean, but a built-in is not really a thing, so maybe it's misleading?

david-crespo: Yeah, thinking more about it, I think the fact that they're built in is an implementation detail I'd rather not expose through the API for now. We can revisit that later when we have roles that are not built in. The built in users thing is different. I'm not sure we even need these endpoints.

zephraph: > the fact that the roles are built in is an implementation detail I'd rather not expose through the API for now

That was my initial inclination for both. Users (aside from "built-in" users) don't exist at a system context so `/system/users` made sense to me. I think the same can be said of roles. They're both denoting oxide provided resources.

david-crespo: Problem is we are using `/system/users?silo={silo}`.

PR #2591, opened 2023-03-17
Delete `/timeseries/schema`

### LINKED ISSUE #1331: Remove Nexus timeseries schema endpoint
See the linked comment thread for more context. The TL;DR is that the existing endpoint for listing timeseries schema is probably unhelpful and should be deleted. It was built when we had no consumers and thought we might export raw data by timeseries. At this point, we'll likely move towards exposing data about resources, which will be internally derived from one or more timeseries in the database. But the raw schema will probably not be visible in the public API.

_Originally posted by @ahl in https://github.com/oxidecomputer/omicron/pull/1329#discussion_r911381485_

david-crespo:
~~I don't think anyone is using this, so I'm breaking the rules and converting it straight.~~ See below. Closes #1331.


bnaecker:
I think we may want to nuke this instead. See [this issue](https://github.com/oxidecomputer/omicron/issues/1331) and [this comment thread](https://github.com/oxidecomputer/omicron/pull/1329#discussion_r911381485) for more background. I'm not sure we want to expose all the timeseries schema to customers if they have no way of querying them. @ahl that was your comment initially, do you have any more thoughts on it at this point?

david-crespo:
I am always happy to delete things outright.

zephraph:
I agree, I think we should kill it. We can always reintroduce something similar if we have a product need for it.

bnaecker:
I'm on board. We can fish it out of the history if needed, and the internal methods for grabbing these from ClickHouse can stick around.

zephraph (approved):

PR #2593, opened 2023-03-17
Remove deprecated disks endpoints

david-crespo:
Part of #2028


end-to-end-tests/src/bin/bootstrap.rs
```diff
@@ -41,12 +41,14 @@ async fn main() -> Result<()> {
     eprintln!("ensuring datasets are ready...");
     let ctx = Context::from_client(client).await?;
     let disk_name = generate_name("disk")?;
+    let org_name = NameOrId::Name(ctx.org_name.clone());
+    let project_name = NameOrId::Name(ctx.project_name.clone());
```
david-crespo, line 45: This is all kind of verbose. The async block seems to mess with its ability to infer a simple `.into()`.

Screenshot 2023-03-17 at 2 53 55 PM

david-crespo: Well this solves the entire problem.

https://github.com/oxidecomputer/typify/pull/224

PR #2597, opened 2023-03-17
[tufaceous] update format for host tarballs

sunshowers:
Per https://github.com/oxidecomputer/helios#os-image-archives:

* The host images are gzipped rather than being uncompressed tarballs.
* There's an oxide.json file with some contents in it that we should
  verify.

### DIFF
--- a/tufaceous-lib/src/artifact.rs
+++ b/tufaceous-lib/src/artifact.rs
@@ -11,11 +11,13 @@ use anyhow::{bail, Context, Result};
 use buf_list::BufList;
 use bytes::Bytes;
 use camino::Utf8PathBuf;
+use flate2::Compression;
 use fs_err::File;
 use omicron_common::{
     api::{external::SemverVersion, internal::nexus::KnownArtifactKind},
     update::ArtifactKind,
 };
+use serde::{Deserialize, Serialize};
 
 /// The location a artifact will be obtained from.
 #[derive(Clone, Debug)]
@@ -136,11 +138,28 @@ fn write_host_tarball_fake_artifact<W: Write>(
     size: u64,
     writer: &mut W,
 ) -> Result<()> {
-    let mut builder = tar::Builder::new(BufWriter::new(writer));
+    let mut builder = tar::Builder::new(flate2::write::GzEncoder::new(
+        BufWriter::new(writer),
+        Compression::fast(),
+    ));
 
     let times = (size as usize) / FILLER_TEXT.len();
     let phase_1_times = times / 8;
     let phase_2_times = times - phase_1_times;
+
+    {
+        let oxide_metadata_json =
+            serde_json::to_string(&OxideMetadata::os_metadata())
+                .expect("oxide metadata should serialize correctly");
+        let mut header = tar::Header::new_gnu();
+        header.set_path(OXIDE_JSON_FILE_NAME).unwrap();
+        header.set_size(oxide_metadata_json.len() as u64);
+        header.set_cksum();
+
+        builder
+            .append(&header, oxide_metadata_json.as_bytes())
+            .with_context(|| format!("error writing {OXIDE_JSON_FILE_NAME}"))?;
+    }
 
     {
         let mut header = tar::Header::new_gnu();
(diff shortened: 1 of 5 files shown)


tufaceous-lib/src/artifact.rs
```diff
@@ -244,18 +268,39 @@ impl HostPhaseImages {
...
+                let json_bytes = read_entry(entry, OXIDE_JSON_FILE_NAME)?;
+                let oxide_json: OxideMetadata =
+                    serde_json::from_slice(&json_bytes).with_context(|| {
+                        format!(
+                        "error deserializing JSON from {OXIDE_JSON_FILE_NAME}"
+                    )
+                    })?;
+
+                let expected = OxideMetadata::os_metadata();
+                if oxide_json != expected {
```
jclulow, line 281: It seems like this strict comparison would not allow us to add keys other than the `v` and `t` keys.  Those represent a _minimum_ set of keys, but others are allowed; e.g., we're already adding a dictionary of extra information in a new object-valued `i` property as well.

tufaceous-lib/src/artifact.rs
```diff
@@ -283,5 +328,18 @@ fn read_entry<R: io::Read>(
...
+struct OxideMetadata {
+    v: String,
+    t: String,
+}
+
+impl OxideMetadata {
+    fn os_metadata() -> Self {
+        Self { v: "1".to_owned(), t: "os".to_owned() }
+    }
+}
```
jclulow, line 345: It might be good to use the utils crate I factored out to get this: https://github.com/oxidecomputer/helios-omicron-brand/blob/main/utils/src/metadata.rs

There is both a builder and a parse routine, the latter of which will check that the `v` and `t` fields are understood.  You could then use the `.is_os()` helper or match on `.archive_type()`.

tufaceous-lib/src/artifact.rs
```diff
@@ -136,12 +138,29 @@ fn write_host_tarball_fake_artifact<W: Write>(
...
 
     let times = (size as usize) / FILLER_TEXT.len();
     let phase_1_times = times / 8;
     let phase_2_times = times - phase_1_times;
 
+    {
+        let oxide_metadata_json =
+            serde_json::to_string(&OxideMetadata::os_metadata())
+                .expect("oxide metadata should serialize correctly");
+        let mut header = tar::Header::new_gnu();
```
jclulow, line 154: These should be `new_ustar()` not `new_gnu()` FWIW, and I think you probably need to `set_entry_type(tar::EntryType::Regular)` as well to make sure they're files in the archive.

tufaceous-lib/src/artifact.rs
```diff
@@ -283,5 +328,18 @@ fn read_entry<R: io::Read>(
 }
```
jclulow, line 328: I suspect `read_entry()` should confirm that these are `tar::EntryType::Regular` (i.e., file) entries before trying to extract them.

jclulow (changes requested):

tufaceous-lib/src/artifact.rs
sunshowers: Thanks, updated.

tufaceous-lib/src/artifact.rs
sunshowers: Yeah, totally missed this before, thanks.

tufaceous-lib/src/artifact.rs
sunshowers: This is fine as it currently stands because we're comparing deserialized formats, and the only keys we're deserializing are `v` and `t`. I'm happy to either make this explicit or add a comment about this, though.

sunshowers: Switched to using the metadata impl from helios-omicron-brand which addresses this issue.

tufaceous-lib/src/artifact.rs
sunshowers: This is a private repo -- discussed this with @jclulow, copied over `metadata.rs` for now with the goal of using this when it's opened up.

tufaceous-lib/src/artifact.rs
jclulow: > This is fine as it currently stands because we're comparing deserialized formats, and the only keys we're deserializing are `v` and `t`. I'm happy to either make this explicit or add a comment about this, though.

Ah ha!  Yes, a comment on that might help.

jclulow (approved):

PR #2598, opened 2023-03-17
Run mg-ddm in the switch zone in transit mode

### LINKED ISSUE #2562: Run maghemite in the switch zone in transit mode
Maghemite is current launched in server mode in the global zone and will broadcast over the sled's two attached NICs, but those advertisements currently don't go anywhere. Maghemite needs to run in the switch zone in transit mode to relay those advertisements.

jmpesp:
Install the mg-ddm zone image into the switch zone, configure it to run
in transit mode, and listen on either all the rear tfport links or a
delegated list of physical links (for non-gimlet based testing).

RSS now asks the switch zone mg-ddm for a list of peers before creating
a sled and service plan.

Fixes #2562.

Prereqs before merging:
- XXX there's an open PR changing Nexus to use OPTE that removes this
- XXX rearN soon


illumos-utils/src/link.rs
```diff
@@ -124,7 +124,7 @@ pub struct InvalidLinkKind(String);
 /// Note that the "ownership" of the VNIC is based on convention;
 /// another process in the global zone could also modify / destroy
 /// the VNIC while this object is alive.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
smklein, line 127: It's dangerous to `Clone` this without taking into account `deleted`.

With my upcoming changes in https://github.com/oxidecomputer/omicron/pull/2725 , this will result in us attempting to delete a VNIC that has already been deleted - an operation that will never succeed.

```rust
let allocator = VnicAllocator::new(...);

// Usage before this PR

// Allocates the VNIC
let mut vnic = allocator.new_control(None)?;
// De-allocates the VNIC, sets `deleted` to true.
vnic.delete();

// Usage after this PR

// Allocates the VNIC
let mut vnic = allocator.new_control(None)?;
let vnic_clone = vnic.clone();
// De-allocates the VNIC, sets `deleted` to true... but NOT for `vnic_clone`!
vnic.delete();
// This tries to destroy a VNIC that has already been deleted!
drop(vnic_clone);
```

sled-agent/src/services.rs
```diff
@@ -613,13 +645,21 @@ impl ServiceManager {
             // opte_ports=
             vec![],
             bootstrap_vnic,
-            link,
+            links.clone(),
```
smklein, line 648: Regarding my comment above about `derive(Clone)` - I don't think you need to clone the links here. Cloning their names would be totally reasonable, so you can iterate on 'em when the zone is up, but "ownership" of the links should be given to the "zone" object being created.

sled-agent/src/services.rs
```diff
@@ -613,13 +645,21 @@ impl ServiceManager {
...
         )
         .await?;
 
         let running_zone = RunningZone::boot(installed_zone).await?;
 
+        for link in &links {
+            info!(self.inner.log, "Ensuring {}/ll exists in zone", link.name(),);
+            Zones::ensure_has_link_local_v6_address(
+                Some(running_zone.name()),
+                &AddrObject::new(link.name(), "ll").unwrap(),
```
smklein, line 659: Use `IPV6_LINK_LOCAL_NAME` from `illumos-utils/src/addrobj.rs`?

sled-agent/src/services.rs
```diff
@@ -1002,11 +1042,82 @@ impl ServiceManager {
...
+
+                    let maghemite_interfaces: Vec<AddrObject> =
+                        if is_gimlet().map_err(|e| {
+                            Error::Underlay(underlay::Error::SystemDetection(e))
+                        })? {
+                            (1..32)
+                                .map(|i| {
+                                    AddrObject::new(
+                                        &format!("tfport{}", i), // XXX rearN soon!
+                                        "ll",
```
smklein, line 1069: Use `IPV6_LINK_LOCAL_NAME` from `illumos-utils/src/addrobj.rs`?

sled-agent/src/services.rs
```diff
@@ -1233,6 +1345,8 @@ impl ServiceManager {
             addresses,
             gz_addresses: vec![],
             services,
+            enable_ipv6_forwarding: true,
+            enable_ipv6_routing: true,
```
smklein, line 1349: Is this a configuration option we *want* to set for all zones? Or is it something that the switch zone should just know to do to itself?

sled-hardware/src/underlay.rs
```diff
@@ -70,3 +70,37 @@ pub fn find_chelsio_links() -> Result<Vec<PhysicalLink>, Error> {
...
+pub(crate) fn find_switch_zone_links_for_maghemite(
+) -> Result<Vec<PhysicalLink>, Error> {
+    #[allow(clippy::if_same_then_else)]
+    let nic_names: Vec<&str> = if is_gimlet().map_err(Error::SystemDetection)? {
+        // tfportd in the switch zone will add links when it boots, no
+        // delegation is required.
+        vec![]
+    } else {
+        // For non-gimlet based testing, hard code your own interfaces here.
+        vec![]
```
smklein, line 99: Should this be a configuration option? Seems weird to separate this out in code, but then have it be something a caller would need to manually modify + recompile anyway

sled-hardware/src/underlay.rs
```diff
@@ -70,3 +70,37 @@ pub fn find_chelsio_links() -> Result<Vec<PhysicalLink>, Error> {
             .collect())
     }
 }
+
+/// Return links to delegate to the switch zone for transmit mode maghemite
+pub fn find_switch_zone_nics() -> Result<Vec<AddrObject>, Error> {
```
smklein, line 75: If I'm reading `find_switch_zone_links_for_maghemite` correctly, this will always return an empty `Vec`, unless the code is changed?

The comment about "no delegation required" below doesn't really make this clear IMO

illumos-utils/src/link.rs
jmpesp: Good point, addressed in 8a37b046

sled-agent/src/services.rs
jmpesp: Addressed in 8a37b046

sled-agent/src/services.rs
jmpesp: Done in 85874f9a

sled-agent/src/services.rs
jmpesp: Done in 85874f9a

sled-hardware/src/underlay.rs
```diff
@@ -70,3 +70,37 @@ pub fn find_chelsio_links() -> Result<Vec<PhysicalLink>, Error> {
...
+    let underlay_nics = find_switch_zone_links_for_maghemite()?;
+
+    let mut addr_objs = Vec::with_capacity(underlay_nics.len());
+    for nic in underlay_nics {
+        let addrobj = AddrObject::new(&nic.0, "ll")?;
```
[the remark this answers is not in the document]
jmpesp: Done in e11e08e3

common/src/address.rs
```diff
@@ -43,6 +43,8 @@ pub const WICKETD_PORT: u16 = 12226;
 pub const BOOTSTRAP_ARTIFACT_PORT: u16 = 12227;
 pub const CRUCIBLE_PANTRY_PORT: u16 = 17000;
 
+pub const MAGHEMITE_PORT: u16 = 8000;
+
```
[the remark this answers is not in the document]
jmpesp: Nope, removed in 1f89a791

sled-hardware/src/underlay.rs
jmpesp: Yeah, that makes sense. Done in 80afc01e

sled-hardware/src/underlay.rs
jmpesp: See 80afc01e

sled-agent/src/services.rs
jmpesp: > should just be a part of the switch zone starting up

I'm having trouble parsing this - do you mean something that the switch zone services do themselves, or something the sled-agent does using something like `run_cmd`?

smklein: Summarizing some offline discussion:

- Presumably, we always want these options set for the switch zone
- Therefore, it's static data we can bake into the switch zone image
- and so it doesn't *need* to be a configuration option

(but if it is to unblock dogfood, that's honestly fine)

jmpesp: Done, see eb3e63427cb41b5a4331dea8569a7fc09d57640a

smklein (approved):

PR #2421, opened 2023-02-24
Remove deprecated instance and networking endpoints

zephraph:
Part of #2028 

This is the first of many PRs to remove non-V1 endpoints. We won't merge this until we've ported the CLI, but it serves as a good first test of what the required changes will be.

PR #2599, opened 2023-03-17
Remove deprecated networking endpoints

zephraph:
Part of #2028, deletes non-v1 networking endpoints


zephraph:
Closing in favor of #2421

PR #2600, opened 2023-03-17
Delete system images endpoints and tables

david-crespo:
This isn't quite ready, but I wanted to see what it would involve. We'd need to update the following scripts to use project images, which would creating a project for them with some canonical name, with the assumption that anybody wanting to work with those image would work in that project. That's kind of annoying, so I'm wondering if maybe we should instead hold off on deleting these endpoints until we have project image promotion figured out. We could do that in a week or two, and the continued presence of `/system/images` doesn't block us from deleting organizations, which is the next thing we want to do.

- [populate-alpine.sh](https://github.com/oxidecomputer/omicron/blob/d51a085f/tools/populate/populate-alpine.sh)
- [populate-images.sh](https://github.com/oxidecomputer/omicron/blob/d51a085f/tools/populate/populate-images.sh)

We will also have to update the mention of system images in [how-to-run.adoc](https://github.com/oxidecomputer/omicron/blob/d51a085ffb400ac4c544e5ee978629321d95d614/docs/how-to-run.adoc).


end-to-end-tests/src/instance_launch.rs
```diff
@@ -41,15 +41,15 @@ async fn instance_launch() -> Result<()> {
     eprintln!("create system image");
     let image_id = ctx
         .client
-        .system_image_create()
-        .body(GlobalImageCreate {
+        .image_create_v1()
```
david-crespo, line 44: this becomes `image_create()` after #2596

david-crespo:
Superseded by #2870

PR #2623, opened 2023-03-20
[RSS] Add testing for RSS

smklein:
- Adds fake interfaces for "Nexus", "DNS", "Bootstrap Network" and "Sled Access", which can be used by RSS or stubbed out for tests.
- Adds multiple tests, validating the behavior of RSS

### DIFF
--- a/sled-agent/src/rack_setup/service.rs
+++ b/sled-agent/src/rack_setup/service.rs
@@ -74,6 +74,7 @@ use crate::rack_setup::plan::sled::{
 };
 use crate::rack_setup::sled_interface::{RealSledAccess, SledInterface};
 use dns_service_client::multiclient::DnsError;
+use dns_service_client::multiclient::ResolveError;
 use internal_dns_names::{ServiceName, SRV};
 use nexus_client::types as NexusTypes;
 use omicron_common::address::{get_sled_address, NEXUS_INTERNAL_PORT};
@@ -132,6 +133,9 @@ pub enum SetupServiceError {
 
     #[error("Failed to access DNS servers: {0}")]
     Dns(#[from] DnsError),
+
+    #[error("Failed to resolve IP address: {0}")]
+    DnsResolve(#[from] ResolveError),
 }
 
 // The workload / information allocated to a single sled.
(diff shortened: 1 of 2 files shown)


sled-agent/src/rack_setup/service.rs
```diff
@@ -151,26 +171,63 @@ impl RackSetupService {
...
+        };
+
+        let marker_directory = omicron_common::OMICRON_CONFIG_PATH.into();
+        Self::new_internal(
+            log,
+            interfaces,
+            marker_directory,
+            request,
+            member_device_id_certs,
+        )
```
smklein, line 202: This is basically the meat-and-potatoes of this PR -- everything else falls out of it.

This new interface shares the same structure as the old implementation, but all access to:
- DNS
- Nexus
- Sled agents
- The bootstrap agent
- Local storage

Is controlled explicitly by these input arguments, meaning that tests can define "whatever they want" and observe RSS behaves as expected under test.

sled-agent/src/rack_setup/dns_interface.rs
```diff
@@ -0,0 +1,87 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! Stubs for the DNS server interface.
```
davepacheco, line 5: I assume this will wind up changing as a result of #2570.

You might consider just spinning up a real DNS server.  I think that would be pretty straightforward.

davepacheco (approved):
I love the test coverage that this gets us!

My only comment is about the general approach.  This is definitely subjective opinion, and I don't suggest doing anything in this PR.  Basically: I don't love using traits to abstract functionality like this just for testing.  I find it hurts code understandability rather a lot (because you can't trace code flow through the trait object or impl -- you have to first go figure out which impl you're looking at) and inhibits tooling (e.g., you want to use rust-analyzer's "find definition of this function" but all you get is the trait definition, not the impl that you care about).  In this PR, I think all these interfaces are already at a network boundary.  I've had success with abstracting at the network boundary instead.  So instead of having a `NexusInterfaces` trait with one function, and one impl of the trait in the real code, and one impl of the trait in the test code, the test code would spin up a FakeNexus dropshot server that impls the same API.  This way you avoid the downsides above.  Again, I'm not suggesting redoing any of this for this PR but I thought I'd mention it.  (It would be neat if Dropshot had a way to do this for you.)

Thanks for doing this!

smklein:
Right off the bat, I totally agree with you. I'd prefer to have this functionality tested "exactly the same, up to the network boundary" for all these services, but the only reason I didn't pursue that route is because, candidly, it seemed like way more effort to create usable fake versions of all these services, especially related to bootstrap-specific interfaces.

I may be overestimating that cost. Before merging this, I'll give an honest shot at replacing these "interface" traits with network-interface-based fakes, but that might take slightly longer to implement.

sled-agent/src/rack_setup/dns_interface.rs
jmpesp: Does https://github.com/oxidecomputer/omicron/blob/b6d9317bc914f0cd4ea7c33e94bc0d737d492537/sled-agent/src/sim/server.rs#L35 fit here?

PR #2626, opened 2023-03-21
Delete organizations

david-crespo:
Per RFD 332. Goodbye. Very little interesting going on, just a lot of little things. I'll see if there's anything to comment on.

One mildly interesting takeaway I have is that almost nothing in our integration tests was depending on the existence of orgs in a meaningful way, i.e., besides just using them to specify a project. This is a good confirmation that they don't need to exist.


david-crespo:
It's pretty cool that organizations are effectively gone (only the table is left) and yet 

```sh
~/oxide/omicron $ rg -i organization | wc -l                                                         
     473
```

david-crespo:
Things are looking surprisingly solid. The main thing left to do is go through the grep results for `organization` and try to get rid of as many of them as is reasonable. Virtually all are in docs and comments.

PR #2628, opened 2023-03-21
NTP zones and time synchronisation

citrus-it:
Every sled runs an NTP zone which is started once the underlay
network is available. An NTP zone runs in one of two roles - either as
a boundary server or non-boundary. A boundary server is responsible
for synchronising time with the external network, whereas a
non-boundary server synchronises with the boundary servers.

The basic flow of time synchronisation is, per RFD34,

```
                       +----------+
                       | External |
                       |  Source  |
                       +----------+
                         /      \
                       /          \
                     v              v
               +----------+    +----------+
               | Boundary |    | Boundary |
               |   NTP    |    |   NTP    |
               |  Zone    |    |  Zone    |
               +----------+    +----------+
                   |||             |||
                   vvv             vvv
     +--------+ +--------+    +--------+ +--------+
     | Gimlet | | Gimlet |    | Gimlet | | Gimlet |
     |   NTP  | |   NTP  |....|   NTP  | |   NTP  |
     |  Zone  | |  Zone  |    |  Zone  | |  Zone  |
     +--------+ +--------+    +--------+ +--------+
```

Several services on a sled should be held back until time is
synchronised "enough".

#### Boundary NTP zone

A boundary NTP zone needs a way to communicate with the external
source. It also needs to be able to perform DNS resolution as it is
likely that the external source will be a DNS name resolving to
multiple time servers. This traffic is all outbound, so could be
behind NAT.

#### Non-Boundary NTP zone

A non-boundary NTP zone synchronises time with the boundary NTP zones. It
discovers these via the internal DNS.

### DIFF
--- a/common/src/address.rs
+++ b/common/src/address.rs
@@ -45,6 +45,8 @@ pub const CRUCIBLE_PANTRY_PORT: u16 = 17000;
 
 pub const NEXUS_INTERNAL_PORT: u16 = 12221;
 
+pub const NTP_PORT: u16 = 123;
+
 // Anycast is a mechanism in which a single IP address is shared by multiple
 // devices, and the destination is located based on routing distance.
 //
--- a/sled-agent/src/bootstrap/params.rs
+++ b/sled-agent/src/bootstrap/params.rs
@@ -33,6 +33,12 @@ pub struct RackInitializeRequest {
     /// Internet gateway information.
     pub gateway: Gateway,
 
+    /// The external NTP server addresses.
+    pub ntp_servers: Vec<String>,
+
+    /// The external DNS server addresses.
+    pub dns_servers: Vec<String>,
+
     /// The address on which Nexus should serve an external interface.
     // TODO(https://github.com/oxidecomputer/omicron/issues/1530): Eventually,
     // this should be pulled from a pool of addresses.
@@ -70,6 +76,12 @@ pub struct SledAgentRequest {
     // Longer-term, it probably makes sense to store this in CRDB and transfer
     // it to Sled Agent as part of the request to launch Nexus.
     pub gateway: Gateway,
+
+    /// The external NTP servers to use
+    pub ntp_servers: Vec<String>,
+    //
+    /// The external DNS servers to use
+    pub dns_servers: Vec<String>,
 
     // Note: The order of these fields is load bearing, because we serialize
     // `SledAgentRequest`s as toml. `subnet` serializes as a TOML table, so it
@@ -205,6 +217,8 @@ mod tests {
                         address: None,
                         mac: MacAddr6::nil().into(),
                     },
+                    ntp_servers: vec![String::from("test.pool.example.com")],
+                    dns_servers: vec![String::from("1.1.1.1")],
                     subnet: Ipv6Subnet::new(Ipv6Addr::LOCALHOST),
                 }),
                 Some(
--- a/sled-agent/src/params.rs
+++ b/sled-agent/src/params.rs
@@ -431,6 +431,10 @@ pub struct ServiceEnsureBody {
 #[derive(Clone, Debug, Deserialize, Serialize, JsonSchema, PartialEq)]
 pub struct TimeSync {
     pub sync: bool,
+    // These could both be f32, but there is a problem with progenitor/typify
+    // where, although the f32 correctly becomes "float" (and not "double") in
+    // the API spec, that "float" gets converted back to f64 when generating
+    // the client.
     pub skew: f64,
     pub correction: f64,
 }
--- a/sled-agent/src/params.rs
+++ b/sled-agent/src/params.rs
@@ -427,3 +427,10 @@ impl From<ServiceZoneRequest> for sled_agent_client::types::ServiceZoneRequest {
 pub struct ServiceEnsureBody {
     pub services: Vec<ServiceZoneRequest>,
 }
+
+#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema, PartialEq)]
+pub struct TimeSync {
+    pub sync: bool,
+    pub skew: f64,
+    pub correction: f64,
+}
--- a/sled-agent/src/params.rs
+++ b/sled-agent/src/params.rs
@@ -272,6 +272,8 @@ pub enum ServiceType {
     Dendrite { asic: DendriteAsic },
     Tfport { pkt_source: String },
     CruciblePantry,
+    Ntp { servers: Vec<String>, boundary: bool },
+    DnsClient { servers: Vec<String>, domain: Option<String> },
 }
 
 impl std::fmt::Display for ServiceType {
@@ -285,6 +287,8 @@ impl std::fmt::Display for ServiceType {
             ServiceType::Dendrite { .. } => write!(f, "dendrite"),
             ServiceType::Tfport { .. } => write!(f, "tfport"),
             ServiceType::CruciblePantry => write!(f, "crucible_pantry"),
+            ServiceType::Ntp { .. } => write!(f, "ntp"),
+            ServiceType::DnsClient { .. } => write!(f, "dns_client"),
         }
     }
 }
@@ -318,6 +322,10 @@ impl From<ServiceType> for sled_agent_client::types::ServiceType {
             }
             St::Tfport { pkt_source } => AutoSt::Tfport { pkt_source },
             St::CruciblePantry => AutoSt::CruciblePantry,
+            St::Ntp { servers, boundary } => AutoSt::Ntp { servers, boundary },
+            St::DnsClient { servers, domain } => {
+                AutoSt::DnsClient { servers, domain }
+            }
         }
     }
 }
@@ -337,6 +345,8 @@ pub enum ZoneType {
     Switch,
     #[serde(rename = "crucible_pantry")]
     CruciblePantry,
+    #[serde(rename = "ntp")]
+    NTP,
 }
 
 impl From<ZoneType> for sled_agent_client::types::ZoneType {
@@ -347,6 +357,7 @@ impl From<ZoneType> for sled_agent_client::types::ZoneType {
             ZoneType::Oximeter => Self::Oximeter,
             ZoneType::Switch => Self::Switch,
             ZoneType::CruciblePantry => Self::CruciblePantry,
+            ZoneType::NTP => Self::Ntp,
         }
     }
 }
@@ -360,6 +371,7 @@ impl std::fmt::Display for ZoneType {
             Oximeter => "oximeter",
             Switch => "switch",
             CruciblePantry => "crucible_pantry",
+            NTP => "ntp",
         };
         write!(f, "{name}")
     }
--- a/sled-agent/src/rack_setup/config.rs
+++ b/sled-agent/src/rack_setup/config.rs
@@ -51,6 +51,8 @@ mod test {
                 address: None,
                 mac: macaddr::MacAddr6::nil().into(),
             },
+            ntp_servers: vec![String::from("test.pool.example.com")],
+            dns_servers: vec![String::from("1.1.1.1")],
             nexus_external_address: "192.168.1.20".parse().unwrap(),
         };
 
--- a/sled-agent/src/rack_setup/plan/service.rs
+++ b/sled-agent/src/rack_setup/plan/service.rs
@@ -9,10 +9,10 @@ use crate::params::{
 };
 use crate::rack_setup::config::SetupServiceConfig as Config;
 use dns_service_client::types::DnsConfigParams;
-use internal_dns::ServiceName;
+use internal_dns::{ServiceName, DNS_ZONE};
 use omicron_common::address::{
     get_switch_zone_address, Ipv6Subnet, ReservedRackSubnet, DNS_PORT,
-    DNS_SERVER_PORT, RSS_RESERVED_ADDRESSES, SLED_PREFIX,
+    DNS_SERVER_PORT, NTP_PORT, RSS_RESERVED_ADDRESSES, SLED_PREFIX,
 };
 use omicron_common::backoff::{
     retry_notify, retry_policy_internal_service_aggressive, BackoffError,
@@ -27,6 +27,9 @@ use std::net::{Ipv6Addr, SocketAddrV6};
 use std::path::{Path, PathBuf};
 use thiserror::Error;
 use uuid::Uuid;
+
+// The number of boundary NTP servers to create from RSS.
+const BOUNDARY_NTP_COUNT: usize = 2;
 
 // The number of Nexus instances to create from RSS.
 const NEXUS_COUNT: usize = 1;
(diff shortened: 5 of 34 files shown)


sled-agent/src/services.rs
```diff
@@ -1015,57 +1092,146 @@ impl ServiceManager {
...
     pub async fn deactivate_switch(&self) -> Result<(), Error> {
-        self.ensure_switch_zone(None).await
+        self.ensure_zone(ZoneType::Switch, None).await
     }
 
-    // Forcefully initialize a switch zone.
+    /// Ensures that an NTP zone exists with the provided configuration.
+    pub async fn activate_ntp(
+        &self,
+        ntp_zone_ip: Option<Ipv6Addr>,
```
smklein, line 1106: `ntp_zone_ip` is provided as optional, but it always exists, right? *Should* it be optional?

sled-agent/src/services.rs
```diff
@@ -1015,57 +1092,146 @@ impl ServiceManager {
...
+        //
+        // This should be done from the local running NTP zone found
+        // via self.innner.ntp_zone - SledLocalZone::Running { request, zone }
+        //
+        // /usr/sbin/chronyd -t10 -L3 -q 'server <sled prefix>::3 iburst'
+        // Exit status 1, no luck, status 0 means we synchronised (the time
+        // will jump) and we can add this IP to the set.
+        //
+        // This strategy is blocked by at least
+        //   https://github.com/oxidecomputer/omicron/issues/2562
```
smklein, line 1160: I'll take a look at this issue, but yeah, we should be able to use maghemite for this.

sled-agent/src/sled_agent.rs
```diff
@@ -338,6 +371,13 @@ impl SledAgent {
             }
         }
 
+        // Every sled runs an NTP zone, but in one of two flavours, depending
+        // on whether it is co-located with a switch zone. When co-located, the
+        // NTP zone runs in boundary mode and will synchronise time with
+        // servers outside of the rack as provided via RSS. Otherwise it runs
+        // as an internal server, sychronising with all boundary NTP servers.
+        self.configure_ntp(log, scrimlet).await;
```
smklein, line 379: Just to confirm:

- If we boot a sled, but the Tofino driver is slow to load...
- ... so we `configure_ntp` with `scrimlet=false`
- ... but later on, the Tofino driver loads, and we call `self.configure_ntp(log, scrimlet = true)`...
- ... do we correctly remove the old NTP Zone, and load the Boundary NTP one?

smklein (commented):
This is awesome, looks like a great start. We may want to meet up to coordinate how I can help with the next steps on this, before and after merging this PR.

jmpesp:
I tried running this on my setup, and it looks like sled-agent may need to create the `chrony` user:

```
root@oxz_ntp:~# svcs ntp
STATE          STIME    FMRI
maintenance    11:04:02 svc:/system/illumos/ntp:default
root@oxz_ntp:~# tail -n30 $(svcs -L ntp)
 Servers: ntp.ubuntu.com
   Allow: fd00:1122:3344:100::/56
Boundary: true
Template: /etc/inet/chrony.conf.boundary
  Config: /etc/inet/chrony.conf
* Stopping daemon
[ Mar 22 11:04:01 Method "stop" exited with status 0. ]
[ Mar 22 11:04:01 Executing start method ("/var/svc/method/svc-site-ntp start"). ]
NTP Service Configuration
-------------------------
 Servers: ntp.ubuntu.com
   Allow: fd00:1122:3344:100::/56
Boundary: true
Template: /etc/inet/chrony.conf.boundary
  Config: /etc/inet/chrony.conf
* Starting daemon
Could not get user/group ID of chrony
[ Mar 22 11:04:02 Method "start" exited with status 0. ]
[ Mar 22 11:04:02 Stopping because all processes in service exited. ]
[ Mar 22 11:04:02 Executing stop method ("/var/svc/method/svc-site-ntp stop 3094"). ]
NTP Service Configuration
-------------------------
 Servers: ntp.ubuntu.com
   Allow: fd00:1122:3344:100::/56
Boundary: true
Template: /etc/inet/chrony.conf.boundary
  Config: /etc/inet/chrony.conf
* Stopping daemon
[ Mar 22 11:04:02 Method "stop" exited with status 0. ]
[ Mar 22 11:04:02 Restarting too quickly, changing state to maintenance. ]
root@oxz_ntp:~# id chrony
id: invalid user name: "chrony"
```

common/src/address.rs
```diff
@@ -142,6 +142,7 @@ impl ReservedRackSubnet {
 
 const SLED_AGENT_ADDRESS_INDEX: usize = 1;
 const SWITCH_ZONE_ADDRESS_INDEX: usize = 2;
+const NTP_ZONE_ADDRESS_INDEX: usize = 3;
```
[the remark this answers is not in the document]
citrus-it: Yes, what you have described is as intended - once underlay is up, an NTP zone is started on it. On a scrimlet, a switch zone is also started. Any switch zone is ::2 on the sled subnet, the NTP zone is ::3.

I originally experimented with having the NTP zones on the bootstrap network. With the right chrony configuration it would allow them to synchronise together to achieve a consistent time, but I think that's unnecessary, and there are discovery challenges.

More context in my other replies here.

sled-agent/src/bootstrap/params.rs
```diff
@@ -71,6 +71,13 @@ pub struct SledAgentRequest {
     // it to Sled Agent as part of the request to launch Nexus.
     pub gateway: Gateway,
 
+    /// The external NTP server to use, a hostname, IPv4 or IPv6 address.
+    pub ntp_server: String,
+
+    // XXXNTP - Similarly, we will presumably need DNS servers in order to
+    // configure the boundary NTP zones so that they can resolve a provided
+    // external NTP hostname.
```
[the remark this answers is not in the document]
citrus-it: Yes, the customer can provide the NTP server as an IPv4, IPv6 or host name through wicket/RSS. We could constrain this to an IP address for MPV. We just plug that into the NTP configuration on the boundary servers.

If they provide a host name, those boundary servers need to be able to resolve that, and through the normal OS mechanisms, which is basically just writing out `/etc/resolv.conf` and
doing effectively `cp /etc/nsswitch.{dns,conf}`.

I think that external DNS servers may already be something we'll need to ask for. I don't think we would need to support separate DNS servers for resolving the NTP server's IPs.

sled-agent/src/sled_agent.rs
citrus-it: The way I've done this up to now (and I am happy to talk about alternatives) is that there is only one NTP zone image that is either acting as a boundary or not. To switch it from one to the other does not involve tearing it down and re-creating, it is just a matter of setting the SMF properties (`boundary` and `servers`), and refreshing the service. It reconfigures itself and restarts if necessary. You can test this in the NTP zone with:

```
svccfg -s ntp setprop config/boundary = false
svccfg -s ntp setprop config/server = test.example.com
svcadm refresh ntp
```

for example

```
root@oxz_ntp:~# sed -n 2p /etc/inet/chrony.conf
# Template configuration file for a boundary NTP server - one which

root@oxz_ntp:~# svccfg -s ntp setprop config/boundary = false
root@oxz_ntp:~# svcadm refresh ntp

root@oxz_ntp:~# sed -n 2p /etc/inet/chrony.conf
# Template configuration file for an internal NTP server - one which
```

smklein: Excellent, that's exactly what I was wondering, and looks like it's working-as-intended.

jclulow:
> Interesting, the user exists in the oxz_ntp zone on my test gimlet built from fairly recent stlouis bits.
> 
> ```
> root@oxz_ntp:~# getent passwd chrony
> chrony:x:23:23:chrony user:/var/lib/chrony:
> ```

This is probably related to https://github.com/oxidecomputer/helios-omicron-brand/commit/317afc5aa30195e1b922fd82ae676f9d7be843fa and I will make a fix for that now, with my apologies.

jclulow:
I believe I have fixed this in https://github.com/oxidecomputer/helios-omicron-brand/commit/dfa3d986c1861f4f186f6b6193d49297acf25bd0 and I have published new packages.

@jmpesp can you `pkg update -v omicron1/tools` and make sure you get `1.0.12` of that and the base `omicron1` package, and then `svcadm restart baseline && tail -F $(svcs -L baseline)` to get a new baseline generated in `/var/run/brand/omicron1`?  After that, newly created zones should have the `chrony` user as they do on Gimlets today.

sled-agent/src/params.rs
```diff
@@ -333,6 +333,8 @@ pub enum ServiceType {
     Dendrite { asic: DendriteAsic },
     Tfport { pkt_source: String },
     CruciblePantry,
+    NTP { servers: Vec<String>, boundary: bool },
```
andrewjstone, line 336: Nit: This should probably follow conventions for Enum's by saying `Ntp` rather than `NTP`. Clippy may complain as well.

smf/ntp/method/svc-site-ntp
```diff
@@ -0,0 +1,121 @@
+#!/usr/bin/ksh
+
+# This file and its contents are supplied under the terms of the
```
andrewjstone, line 3: Why are we licensing this script as CDDL? Is it mostly copied from illumos? If it's Oxide specific and new it should likely carry the standard Oxide license of MPL v2.0

andrewjstone (commented):
@citrus-it This looks great. Just a couple nits from me as Sean did a thorough review, and James tested it already!

citrus-it:
I've moved this on a bit, re-basing back on top of main and adding the new `ServiceType::DnsClient` which is responsible for OS DNS configuration in a zone. It's currently only used by a boundary NTP service but might be useful for other zone types later.
I've mapped the new `dns_servers` through from RSS and changed `ntp_server` to be a Vec since I was in there.

I still need to finish testing the case where tofino takes a while to appear.

sled-agent/src/bootstrap/params.rs
```diff
@@ -205,6 +217,8 @@ mod tests {
                         address: None,
                         mac: MacAddr6::nil().into(),
                     },
+                    ntp_servers: vec!["test.pool.example.com".parse().unwrap()],
+                    dns_servers: vec!["1.1.1.1".parse().unwrap()],
```
smklein, line 221: These are just `&'static str` types, so to convert to a `String` type, you can call either:

`String::from("test.pool.example.com")` - https://doc.rust-lang.org/std/string/struct.String.html#impl-From%3C%26str%3E-for-String
or
`"test.pool.example.com".to_string()` - https://doc.rust-lang.org/std/string/trait.ToString.html#tymethod.to_string


`parse` makes sense if the conversion can fail, but since this conversion is infallible, we can use one of these alternatives to avoid the `unwrap`.

sled-agent/src/rack_setup/config.rs
```diff
@@ -51,6 +51,8 @@ mod test {
                 address: None,
                 mac: macaddr::MacAddr6::nil().into(),
             },
+            ntp_servers: vec!["test.pool.example.com".parse().unwrap()],
```
smklein, line 54: Same note here on conversions

sled-agent/src/rack_setup/plan/service.rs
```diff
@@ -28,6 +28,9 @@ use std::path::{Path, PathBuf};
 use thiserror::Error;
 use uuid::Uuid;
 
+// The number of boundary NTP servers to create from RSS.
+const BOUNDARY_NTP_COUNT: usize = 2;
```
smklein, line 32: We *might* want to make this `1` for now, and have it be a configuration option. Totally agree with this disposition on the production build, but it might make development builds painful.

If we require a minimum of `2`, will it be possible to call RSS when running from a single test sled?

citrus-it: As it is now, two is okay here even with a single sled. The first BOUNDARY_NTP_COUNT sleds get their NTP server configured in a boundary role, all others get it set up as an internal server. If there is just one sled, we end up with one boundary server.

In a change from an earlier version of this PR, having now decoupled the boundary server role from the scrimlets, the NTP servers don't start until after RSS in the bootstrap case. The sequence is (based on your earlier comment):

1. Boot to bootstrap network
1. Operator provides underlay address, DNS + NTP configuration through wicket / RSS
1. RSS starts sled agents, tells them to use underlay
1. RSS starts Internal DNS and NTP
1. Wait for all sleds to have synchronised time (that's the second commit in this branch that I've been testing today).
1. RSS starts all other services

Internal servers find the boundary ones via internal DNS, boundary ones use the settings from the operator.

I still have to test the cold reboot case.

jclulow: Does the _operator_ provide the underlay addresses or do we just generate those ourselves?

citrus-it: They're generated by RSS and persisted in the service plan.

```toml
[[services."[fd00:1122:3344:101::1]:12345".service]]
addresses = ["fd00:1122:3344:101::10"]
gz_addresses = []
id = "3db6e379-17e4-48da-b6f5-34b80c114343"
zone_type = "ntp"

[[services."[fd00:1122:3344:101::1]:12345".service.services]]
boundary = true
servers = ["0.pool.ntp.org"]
type = "ntp"

[[services."[fd00:1122:3344:101::1]:12345".service.services]]
servers = ["9.9.9.9", "80.80.80.80"]
type = "dns_client"
```

along with the DNS entries that are used by the non-boundary NTP servers
(this could be switched to just use the IPs directly too, and that might be worthwhile).

```
[[dns_config.zones.records.records]]
data = "fd00:1122:3344:101::10"
type = "AAAA"

[dns_config.zones.records.key]
name = "3db6e379-17e4-48da-b6f5-34b80c114343.host"
```

sled-agent/src/rack_setup/plan/service.rs
```diff
@@ -362,6 +373,61 @@ impl Plan {
...
+                let id = Uuid::new_v4();
+                let address = addr_alloc.next().expect("Not enough addrs");
+                let zone = dns_builder.host_zone(id, address).unwrap();
+
+                let (services, svcname) = if idx < BOUNDARY_NTP_COUNT {
+                    boundary_ntp_servers
+                        .push(format!("{}.host.{}", id, DNS_ZONE));
+                    (
+                        // XXXNTP - these boundary servers need a path to the
+                        // external network via OPTE.
```
smklein, line 389: PSA to @luqmana - we should coordinate this integration with https://github.com/oxidecomputer/omicron/pull/2419

smklein (approved):
Thanks for all this great work - this is a big, cross-cutting integration piece, and getting NTP reliably integrated is a major unblocker for a lot of follow-up development (e.g., multi-node CRDB).

Only a couple questions, but feel free to merge at your discretion!

sled-agent/src/services.rs
```diff
@@ -497,14 +525,64 @@ impl ServiceManager {
...
+        running_zone: &RunningZone,
+        service: &ServiceType,
+    ) -> Result<(), Error> {
+        if let ServiceType::DnsClient { servers, domain } = service {
+            let smfh = SmfHelper::new(&running_zone, service);
+
+            let etc = PathBuf::from(running_zone.root()).join("etc");
+            let resolv_conf = etc.join("resolv.conf");
+            let nsswitch_conf = etc.join("nsswitch.conf");
+            let nsswitch_dns = etc.join("nsswitch.dns");
```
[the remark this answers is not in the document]
smklein: Oh, to be clear, I'm not asking you to use `new_from_ip` - I'm actually asking about the opposite, if we should be plumbing this DNS info into other zones, so we don't need to explicitly reference the DNS server.

PR #2632, opened 2023-03-22
[wicketd] add integration test with installinator, fix some issues

sunshowers:
* Only send the last few failure events rather than all of them, capping
  the size of progress reports on the client side.
* Set a generous upper bound on the size of progress reports on the
  server side, with the aim of ensuring that a report that's too large
  doesn't block completion.
* Add an integration test against the real installinator and wicketd
  instances.

### DIFF
--- a/installinator/src/reporter.rs
+++ b/installinator/src/reporter.rs
@@ -118,7 +118,30 @@ where
 
     async fn on_tick(&mut self) {
         // Assemble a report out of pending completion and progress events.
-        let completion_events = self.completion.clone();
+        // For completion, include all success events and the last 20 failure
+        // events (with the goal being to keep the size of the report down,
+        // since a report that's too large will cause the max payload size to be
+        // exceeded.)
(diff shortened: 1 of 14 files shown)


installinator/src/reporter.rs
```diff
@@ -118,7 +118,30 @@ where
...
+                } else {
+                    failure_events_seen += 1;
+                    failure_events_seen <= 20
+                }
+            })
+            .cloned()
+            .collect();
+        // IMPORTANT: You might be tempted to replace this reverse call with
+        // another `.rev()` above. Don't do that! It will cause the *first* 20
+        // elements to be taken rather than the *last* 20 elements.
```
jgallagher, line 142: Nice comment - I'm surprised by that behavior TBH - I would've guessed a `.rev()` after the filter would be okay.

jgallagher (approved):
LGTM :+1:

andrewjstone (approved):
Looks great!

PR #2635, opened 2023-03-22
[wicket] make tab and shift-tab switch panes

sunshowers:
I kept wanting to type in tab and shift-tab to do this, and it seems
like a more natural fit.


andrewjstone (approved):
Good change!

PR #2637, opened 2023-03-22
[wicket] show PWR1 as empty, skip over it while tabbing over

### LINKED ISSUE #2559: [wicket] - Only populate pwr0
We will only have one powershelf, `pwr0`, running in production. We need to style the other powershelf so that it doesn't show up as present. We also should either remove `ComponentId::Psc1` or just always skip it in iteration. 

See https://github.com/oxidecomputer/omicron/pull/2440#issuecomment-1449135088

cc @rmustacc

sunshowers:
Our current shipping racks just have one power shelf, so show PWR1 as empty.

image

Closes #2559.

### DIFF
--- a/wicket/src/state/rack.rs
+++ b/wicket/src/state/rack.rs
@@ -71,6 +71,8 @@ impl RackState {
                     ComponentId::Sled(17)
                 }
             }
+            // Skip over Psc(1) because it is always empty in currently shipping
+            // racks.
             ComponentId::Psc(0) => ComponentId::Switch(1),
             _ => unreachable!(),
         };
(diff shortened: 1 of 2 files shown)


rmustacc:
> Rename "PWR0" to just PWR.

Can we possibly not rename this? This should ideally match the external labeling that we have on the rack which still has the positional information. It's more accurate to show PWR1 as a blank. See https://rfd.shared.oxide.computer/rfd/0200#component_naming_conventions.

rmustacc:
> Thanks @rmustacc -- I misunderstood what was required. Updated.

Much appreciated, thank you!

andrewjstone (approved):
Great stuff. Thanks!

sunshowers:
Ugh, merged it to the dependent branch by accident. Will redo and land on main later.

PR #2642, opened 2023-03-22
[wicket] Introduce wicket-dbg

andrewjstone:
This builds on top of #2638

This commit introduces the wicket-dbg package which contains two new binaries `wicket-dbg-server` and `wicket-dbg`. `wicket-dbg` is an interactive CLI for interacting with the `wicket-dbg-server`, which is capable of loading and playing recordings from previous wicket dumps as of this commit.

This is just the start of the debugger. Future commits will add single steps and jumps (stepping over ticks), breakpoints, speed up / slow down of playback and pause functionality.

One major limitation is that resize currently does not work. In particular if your terminal running the `wicket-dbg-server` is smaller than the terminal of the recording, the server may crash. Additionally, it's unknown what the effect of resizes during recoring will be during playback. This is somewhat difficult to fix, and not immediately relevant since we anticipate using this debugger mostly during iterative development right now. Note however, that the debugger, modulo signing is currently capable of playing recordings dumped from production wicket systems, and so we will likely want to fix this for that use case.


wicket-dbg/src/server.rs
```diff
@@ -0,0 +1,95 @@
...
+        loop {
+            let cmd: Cmd = ciborium::de::from_reader(&mut stream)?;
+            self.handle(cmd, &mut stream)?;
+        }
+    }
+
+    fn handle(
+        &mut self,
+        cmd: Cmd,
+        stream: &mut TcpStream,
```
sunshowers, line 47: I think the stream here and elsewhere needs some sort of buffering around it.

I'd also express a preference for using async code here, though I'll let you decide whether you want to do that or not. (It's ok to use cbor to convert the data into an in-memory chunk and then write that out over the wire.)

andrewjstone: > I think the stream here and elsewhere needs some sort of buffering around it.

I don't think that's true. We only send one command at a time, and the operations are synchronous. We're not trying to reduce syscalls here, we want to send with as low-latency as possible. This is also why I turned off nagle on the socket, so that the cli responds quickly.

> 
> I'd also express a preference for using async code here, though I'll let you decide whether you want to do that or not. (It's ok to use cbor to convert the data into an in-memory chunk and then write that out over the wire.)

Cargo.toml
```diff
@@ -106,9 +108,9 @@ resolver = "2"
 anyhow = "1.0"
 api_identity = { path = "api_identity" }
 assert_matches = "1.5.0"
-assert_cmd = "2.0.10"
+assert_cmd = "2.0.8"
 async-bb8-diesel = { git = "https://github.com/oxidecomputer/async-bb8-diesel", rev = "7944dafc8a36dc6e20a1405eca59d04662de2bb7" }
-async-trait = "0.1.67"
+async-trait = "0.1.66"
 authz-macros = { path = "nexus/authz-macros" }
```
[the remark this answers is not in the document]
andrewjstone: Good catch. Fixed.

wicket-dbg/src/bin/wicket-dbg-server.rs
```diff
@@ -0,0 +1,13 @@
...
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use anyhow::Result;
+use wicket_dbg::Runner;
+
+fn main() -> Result<()> {
+    let drain = slog::Discard;
+    let root = slog::Logger::root(drain, slog::o!());
+    let runner = Runner::new(root);
+    wicket_dbg::Server::run("::1:9010", runner)
```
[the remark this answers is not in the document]
andrewjstone: Done.

wicket-dbg/src/server.rs
andrewjstone: On second thought, I'm switching over to async for the server in a follow up as I add more functionality. Specifically, the ability to pause playback and get back the current event index paused with a oneshot channel is a leading reason.

sunshowers: > I don't think that's true.

So I very recently fixed an issue with dropshot's test harness which was incredibly slow because `serde_json` was writing one character at a time: https://github.com/oxidecomputer/dropshot/pull/539

I'm not entirely sure what ciborium does, but adding a little bit of buffering before flushing it out immediately would make sense I think without causing any sort of significant delay.

sunshowers (approved):
Fantastic, thanks! Just the one comment about buffering but it isn't blocking, since it seems like you're going to re-do this with tokio anyway.

wicket-dbg/src/server.rs
andrewjstone: > > I don't think that's true.
> 
> So I very recently fixed an issue with dropshot's test harness which was incredibly slow because `serde_json` was writing one character at a time: [oxidecomputer/dropshot#539](https://github.com/oxidecomputer/dropshot/pull/539)
> 

Wow, that's some crazy pathological behavior. You literally can't get worse 😆  😭. And yeah, I don't think a buffer really hurts even if it adds a touch of latency, due to the fact we are dealing with human timescales here. I already have code that buffers and uses tokio, so you I guess I decided you were right!

> I'm not entirely sure what ciborium does, but adding a little bit of buffering before flushing it out immediately would make sense I think without causing any sort of significant delay.

I assumed it would behave sanely and dump as much as it could in a single write, but that is clearly not always a good assumption. Thanks for the pointer. This has been a learning experience lol.

andrewjstone:
Closed in favor of #2682

PR #2643, opened 2023-03-22
[WIP] [nexus-db-macros] allow name to be owned in addition to being b…

sunshowers:
…orrowed


sunshowers:
@zephraph is going to take and adapt this

zephraph:
I've rolled this into #2647

PR #2644, opened 2023-03-22
[nexus-db-macros] add extra-traits to syn features

sunshowers:
Means `cargo build -p db-macros` works.


zephraph (approved):
Thanks!

zephraph:
I think the test failures are environment flakes.

PR #2647, opened 2023-03-23
Simplify resource selectors introduced in V1 API and do some misc cleanup

zephraph:
Prior to this PR selectors were constructed kind of like nesting dolls. Each selector contained whatever parameter was important for that resource and a `<resource>_selector` property with a flattened struct that was essentially the same structure. I largely adopted this approach because it let me build full selectors in the http entrypoints while also limiting the match logic of the `<resource>_lookup` functions to just the property we care about and its error cases (while delegating parent resolution to nested lookups). 

With this PR my aim was to explicitly condense down the selectors to a flat struct with all its dependencies written directly inline. No more serde flatten. Unfortunately, due to how the lookup handler was constructed, this was impossible without making the `<resource>_lookup` functions handle the resolution of every parameters (even parent parameters). The reason for this is that the lookup macro uses `&Name` as an input type for `.<resource>_name()` chained lookup calls. The only way to have `&Name` passed to the lookup was to have a reference of the selector passed in the selector lookup... which means you either have to construct the full selector outside the selector lookup (nesting doll structs) or _don't_ make the parent selectors thus not calling the parent lookups thus handling all the logic in every lookup....

SO. I tagged in @sunshowers and they helped me convert it to an owned version. Essentially this means the lookup macro now takes `Name` instead of `&Name` for `.<resource_name()`. The practical implications of this are some syntactic churn and simpler selector construction. 

---

Other changes include:
1. A small macro rule to dry up path param definitions
2. Removal of some structs that weren't used.


davepacheco:
Sorry in advance that I think I'm missing the purpose/upside here.  So consider all this tentative.  But what I'm thinking is: the lookup structs used to take names by reference, and that now they take owned names, which means a bunch of callers have to clone the names when they otherwise didn't.  All things being equal, it seems nicer if they can take references?  Also: with the nested "Russian doll" approach, the hierarchy wasn't duplicated in each struct, which seems better?  (It seems like lots of opportunities for error to duplicate these fields in a lot of places.)  The previous approach presumably makes it easier to write functions that operate on the selector for some middle node in the hierarchy (like Vpc), whereas it seems like you can't do that today because things further in the hierarchy don't _use_ a VpcSelector any more?  These aren't necessarily big problems but they do seem like downsides and I'm not clear what the advantage is.

david-crespo (commented):
Justin and I talked about this in chat, but basically I agree with @davepacheco that the change is mostly a wash. I was hoping the flat structs would be nice but the change to `Name` required to make it work is a lot of add a clone here, remove a reference there. The line savings pulled out in #2652 are the remaining good part.

sunshowers:
I don't know enough to have strong opinions here, but one approach is to add separate `BorrowedName` (or just stay with `Name` for this to avoid churn) and `OwnedName` variants.

zephraph:
@sunshowers's initial implementation in #2643 had a split implementation as suggested above. That would allow for `instance_name` or `instance_name_owned`. That would reduce a lot of the churn in terms of `&` and `.clone()` changes. I could pull that back in and change the implementation here to reduce the noise, but I'm hesitant to spend the time on it if the general approach is in question.

Ultimately the idea here was that the flattened selectors would be easier to reason about. While yes, it's true there's more opportunities for mistakes without the rigidly encoded structure, this is somewhat mitigated by the fact that lookups themselves are still nested in nature.

PR #2653, opened 2023-03-23
[wicket] show PWR1 as empty, skip over it while tabbing over

### LINKED ISSUE #2559: [wicket] - Only populate pwr0

sunshowers:
Our current shipping racks just have one power shelf, so show PWR1 as empty.

image

(This was originally reviewed in #2637 but I screwed up the land so putting this up again. Will land without reviewing since there are no changes compared to 2637.)

### DIFF
--- a/wicket/src/state/rack.rs
+++ b/wicket/src/state/rack.rs
@@ -71,8 +71,9 @@ impl RackState {
                     ComponentId::Sled(17)
                 }
             }
-            ComponentId::Psc(0) => ComponentId::Psc(1),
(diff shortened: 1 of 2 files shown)


andrewjstone (approved):
Looks great!

PR #2657, opened 2023-03-24
SoftNPU related docs and tweaks

luqmana:
With SoftNPU landed just been going through trying to get it deployed locally and ran into some small things here and there. Probably easiest to review by commit, but some of the highlights are:

1. Separating the notion of runtime sled mode config (auto-detect, force gimlet, force scrimlet) and build/package time switch zone type (asic, stub, softnpu).
2. Updating how-to-run doc with how to setup SoftNPU.
3. Bumping OPTE rev to pickup latest upstream fixes (specifically https://github.com/oxidecomputer/opte/issues/342 which can cause packets on some TCP packets).

cc @smklein @internet-diglett


smklein (approved):
I really like the usage of feature flags here - not to actually compile things differently, but to just validate that the pre-requisites at build-time are enforced at runtime.

Thanks for all this cleanup; I think these changes make it harder to get the configuration combinations "wrong".

PR #2677, opened 2023-03-27
[bootstrap] Add API to undo rack initialization

smklein:
- Adds a couple endpoints to the bootstrap HTTP interface:
  - `/rack-initialize` (DELETE) - resets all sleds within the rack
  - `/sled-initialize` (DELETE) - resets a single sled, removing all configuration RSS might try to apply
- Adds `bootstrap-agent-client`, which includes generated bindings for the bootstrap HTTP interface

How I tested this:
- Deployed to `BRM42220026` in the lab
- Ran `curl -XPOST --data '{"rack_subnet": "fd00:1122:3344:0100::", "rack_secret_threshold": 1, "nexus_external_address": "fd02:1122:3344:0101::3", "gateway": {"address": "10.1.0.50", "mac": "02:43:9b:55:ed:7c"}}' -H "Content-Type: application/json"  http://[fdb0:a840:2504:351::1]:80/rack-initialize`
  - Observed RSS doing what it does, to get Nexus up and running
- Ran `curl -XDELETE http://[fdb0:a840:2504:351::1]:80/rack-initialize`
  - Observed the rack had removed all datasets, zones, IP addresses, local-to-sled configuration


illumos-utils/src/zfs.rs
```diff
@@ -244,3 +244,26 @@ impl Zfs {
         Ok(value.to_string())
     }
 }
+
+/// Returns all datasets managed by Omicron
+pub fn get_all_omicron_datasets() -> anyhow::Result<Vec<String>> {
```
smklein, line 249: This was pulled exactly from `omicron-package` -- but the sled-reset command wants to use it too.

sled-agent/src/bootstrap/agent.rs
```diff
@@ -729,6 +746,163 @@ impl Agent {
...
+    /// Resets this sled, removing:
+    ///
+    /// - All control plane zones (except the switch zone)
+    /// - All sled-local configuration
+    /// - All underlay networking
+    /// - All storage managed by the control plane
+    ///
+    /// This API is intended to put the sled into a state where it can
+    /// subsequently be initialized via RSS.
+    pub async fn sled_reset(&self) -> Result<(), BootstrapError> {
```
smklein, line 856: This function is worth reviewing; it's the implementation of all "local-to-my-sled" resetting

sled-agent/src/bootstrap/agent.rs
```diff
@@ -729,6 +746,163 @@ impl Agent {
         Ok(())
     }
 
+    /// Runs the rack setup service to completion
+    pub async fn rack_reset(&self) -> Result<(), BootstrapError> {
+        RssHandle::run_rss_reset(&self.parent_log, self.ip, None).await?;
+        Ok(())
+    }
```
smklein, line 753: Just to make the callstack a little more obvious:

- `sled-agent/src/bootstrap/http_entrypoints.rs` has a `rack_reset` endpoint, which calls...
- `sled-agent/src/bootstrap/agent.rs` 's `rack_reset`. This is a pass-through function which calls calls...
- `sled-agent/src/bootstrap/rss_handle.rs`'s `RssHandle::run_rss_reset`, which abstracts away some of the "local vs remote" details of accessing the bootstrap agent. This calls...
- `sled-agent/src/rack_setup/service.rs`'s `new_reset_rack`, which spawns a tokio task that calls...
- `sled-agent/src/rack_setup/service.rs`'s `RackSetupService::reset`. This function does the following:
  - Query for all visible bootstrap addresses from the current sled
  - Send a `reset` request to all of them (including our own!). This goes through...
- `sled-agent/src/bootstrap/rss_handle.rs` - `BootstrapAgentHandle::reset_sleds`, which sends reset requests to all sleds (including ourselves) in the same way that all the initialization requests are normally sent out.

sled-agent/src/bootstrap/agent.rs
```diff
@@ -729,6 +750,163 @@ impl Agent {
...
+            SledAgentState::Before(None) => {
+                let hardware_monitor = self.start_hardware_monitor()
+                    .await
+                    .map_err(|err| {
+                        warn!(self.log, "Failed to restart bootstrap agent hardware monitor");
+                        err
+                    })?;
+                *state = SledAgentState::Before(Some(hardware_monitor));
+            }
+            _ => {}
```
jgallagher, line 901: Just double-checking my own understanding: at this point, we cannot be in `SledAgentState::After()` (because of the swap up on line 867). We could be already be in `Before(Some(_))`, in which case we do nothing, and if we were in `Before(None)` we're now `Before(Some(_))` instead thanks to line 899?

sled-agent/src/bootstrap/agent.rs
```diff
@@ -729,6 +750,163 @@ impl Agent {
...
+        };
+
+        // Return any errors encountered resetting the sled.
+        result?;
```
jgallagher, line 905: Trivial nitpick - it looks like these types match and we could just end the function with

```suggestion
        result
```

sled-agent/src/bootstrap/http_entrypoints.rs
```diff
@@ -66,3 +68,29 @@ async fn rack_initialize(
...
+/// Resets the rack to an unconfigured state.
+#[endpoint {
+    method = DELETE,
+    path = "/rack-initialize",
+}]
+async fn rack_reset(
+    rqctx: RequestContext<Arc<Agent>>,
+) -> Result<HttpResponseUpdatedNoContent, HttpError> {
+    let ba = rqctx.context();
+    ba.rack_reset().await.map_err(|e| Error::from(e))?;
```
jgallagher, line 81: Apologies for the vagueness of this question - do we need any kind of locking / queuing to handle this endpoint being hit multiple times before the first request finishes? Similarly for `sled_reset` and/or interactions between reset and initialize.

jgallagher (approved):

sled-agent/src/bootstrap/agent.rs
smklein: Yup! I really wish there was a better way for me to have the `Before` and `After` states without the `Option` in the before state, but unfortunately, starting the hardware monitor *is* a fallible operation, so it's possible to fall-back to that as a defunct state.

In plain English, this function should read as the following:

1. If a Sled Agent server is running, stop it.
2. Try to uninstall everything we can.
3. If the bootstrap agent is not monitoring hardware, for any reason, start the monitor.

Step (3) could trigger because we stopped the sled agent. It could *also* trigger because we previously failed to restart the hardware monitor.

sled-agent/src/bootstrap/agent.rs
smklein: Updated

sled-agent/src/bootstrap/http_entrypoints.rs
smklein: The `sled_reset` function accesses the `sled_state` variable of the bootstrap agent, which is protected by a `Mutex`. This means, at a baseline, all accesses to `sled_reset` are serialized.

I *do* think it's technically possible to POST and DELETE `rack-initialize` concurrently. I can add a mutex to ensure that only one of those methods can be called at once.

This is done in https://github.com/oxidecomputer/omicron/pull/2677/commits/3fc2c7ba43ee0d0a3ffa3d40d45ae27b390d9ae8

sled-agent/src/bootstrap/agent.rs
```diff
@@ -733,6 +738,9 @@ impl Agent {
         &self,
         request: RackInitializeRequest,
     ) -> Result<(), BootstrapError> {
+        // Avoid concurrent initialization and teardown.
+        let _rss_access = self.rss_access.lock().await;
```
jgallagher, line 742: Should these be `try_lock()` and return an error immediately if the lock is held? Otherwise the caller is going to be waiting a while for a response, I think, and perhaps they don't even really want this to attempt to run twice.

smklein: I like that idea; done in bdf912723bf151d2f1f7a06cc938f5990326b3e0

PR #2680, opened 2023-03-27
[wicket] fix popup sizing

sunshowers:
* Reduce size to a maximum of 80% of the screen.
* Wrap text rather than overflowing it.
* Don't make text overflow into the button area.


sunshowers:
Here's what it looks like now.

image

andrewjstone (approved):
LGTM! Thanks.

PR #2684, opened 2023-03-27
[sled agent] Create reservoir for VMs

### LINKED ISSUE #2650: [sled agent] Provision a reservoir for Propolis
See:

- https://github.com/illumos/illumos-gate/blob/79492562b32b5e6bc03e14ad2b51f986335f3709/usr/src/uts/intel/io/vmm/vmm_reservoir.c#L16-L78
- https://github.com/oxidecomputer/illumos-gate/blob/stlouis/usr/src/cmd/rsrvrctl/rsrvrctl.c
  - I found this in the ramdisk image in `/usr/lib/rsrvrctl`

This will probably be a blocker for accurate accounting in https://github.com/oxidecomputer/omicron/issues/2648

smklein:
- Adds a VMM reservoir to the sled agent on setup (50% of physical memory)
- Reports the size used by that VMM reservoir to Nexus, where may be used for future accounting (such as that described in RFD 380)

Fixes #2650


smklein:
FYI @jclulow  - I'm blocked on the `lab-opte-0.21` image not having an up-to-date `rsrvrctl` . AFAICT it *does* have what we need in `helios-latest`, but I'm not 100% sure what to kick to get that `lab-opte` target up-to-date

smklein:
FYI - as of https://github.com/oxidecomputer/omicron/pull/2684/commits/a0b4f324a15bdb7d11444714d565cda1e13120b5 , I'm no longer calling `rsrvrctl` -- I'm using the `bhyve_api` instead, as @pfmooney suggested

illumos-utils/src/lib.rs
```diff
@@ -11,6 +11,7 @@ pub mod dladm;
 pub mod fstyp;
 pub mod link;
 pub mod opte;
+pub mod rsrvrctl;
```
pfmooney, line 14: `rsrvrctl` is just a "cute" name for the Private utility.  I'd probably call this something like `vmm_reservoir`, since it's not bound by that

illumos-utils/src/rsrvrctl.rs
```diff
@@ -0,0 +1,39 @@
...
+//! Wrappers around reservoir controls
+
+use omicron_common::api::external::ByteCount;
+
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error("Reservoir size must be a multiple of MiB, not: {0}")]
+    InvalidSize(ByteCount),
+
+    #[error("Failed to create reservoir: {0}")]
```
pfmooney, line 14: * Failed to set reservoir size

pfmooney (commented):
Thanks for wiring this in

illumos-utils/src/lib.rs
smklein: Done

illumos-utils/src/rsrvrctl.rs
smklein: Updated

Cargo.toml
```diff
@@ -236,6 +236,7 @@ pretty-hex = "0.3.0"
 proc-macro2 = "1.0"
 progenitor = { git = "https://github.com/oxidecomputer/progenitor", branch = "main" }
 progenitor-client = { git = "https://github.com/oxidecomputer/progenitor", branch = "main" }
+bhyve_api = { git = "https://github.com/oxidecomputer/propolis", rev = "db39a40be172109de3ca34af8a1c37dd76e715a9" }
```
luqmana, line 239: Perhaps it's time to split this off into it's own home. Though, we're gonna be packaging together the OS/bhyve, propolis & sled-agent so maybe we should just make sure to keep all the propolis stuff on the same rev. Not really possibly yet since omicron is still pinned on an older rev but in the future. Anyways, not a blocker for this PR.

illumos-utils/src/vmm_reservoir.rs
```diff
@@ -0,0 +1,39 @@
...
+    ReservoirError(#[from] std::io::Error),
+}
+
+/// Controls the size of the memory reservoir
+pub struct ReservoirControl {}
+
+impl ReservoirControl {
+    /// Sets the reservoir to a particular size.
+    ///
+    /// - "size" must be exactly divisible by a MiB
```
luqmana, line 24: Should this just take the size in `MiB` then?

sled-agent/src/sled_agent.rs
```diff
@@ -251,26 +251,22 @@ impl SledAgent {
...
+            .map_err(|e| Error::Hardware(e))?;
+
         let instances = InstanceManager::new(
             parent_log.clone(),
             lazy_nexus_client.clone(),
             etherstub.clone(),
             *sled_address.ip(),
             request.gateway.mac.0,
         )?;
+        instances.set_reservoir_size(&hardware)?;
```
luqmana, line 264: Do we anticipate changing the reservoir size at runtime? If not maybe just set it as part of `InstanceManager::new`. Would also let you get rid of the `Mutex`

sled-agent/src/instance_manager.rs
```diff
@@ -86,6 +96,27 @@ impl InstanceManager {
         })
     }
 
+    /// Stores the reservoir size to 80% of the usable physical RAM
```
luqmana, line 99:
```suggestion
    /// Sets the reservoir size to 80% of the usable physical RAM
```

sled-agent/src/sled_agent.rs
smklein: I actually intentionally kept it separate -- I think that in the future, we may want Nexus to have control over this value, and actually modify it at runtime. It seems like a powerful tool for binpacking purposes to be able to tweak this on a per-sled basis.

(Namely, the "80% of Phys RAM goes to the reservoir" bit is totally arbitrary, but we could make this smaller or larger on sleds dedicated more or less towards guest workloads)

illumos-utils/src/vmm_reservoir.rs
smklein: We *could*, but would probably want a new type to represent that, since IMO it's confusing to have `u64`s with overloaded meaning. This was why we originally used `ByteCount` everywhere, to make it explicit that "it's a count of bytes", but that does have the downside of permitting values that need to be validated.

Do you have a strong opinion on where this new type should live?

sled-agent/src/instance_manager.rs
smklein: Fixed

jordanhendricks:
Closing this as it is now superseded by #3200

PR #2686, opened 2023-03-28
Update pinned OPTE and start installing directly from helios-netdev repo

luqmana:
Update pinned rev to latest main (corresponds to `driver/network/opte@0.21.162`) which includes a fix for a kernel panic (https://github.com/oxidecomputer/opte/pull/352).

Also revives https://github.com/oxidecomputer/omicron/pull/1907 to pull in opte from the helios-netdev package repo rather than just downloading the p5p artifact from a specific commit's CI run. Added a PR workflow to make sure all the necessary places get updated in lockstep as well.


bnaecker (approved):
Small suggestion to make fetching the number of commits a bit more clear, but otherwise LGTM. Thanks!

PR #2691, opened 2023-03-29
Remove docker image builder

karencfv:
As far as I know the only thing using this image was the old CLI? Anyone have any objections to me removing this?


augustuswm:
Nothing that I know of.

PR #2695, opened 2023-03-29
Bump rust toolchain to 1.68.2

iliana:
This is a somewhat selfish PR, as [configuring `registries.crates-io.protocol = sparse` globally](https://github.com/iliana/dotfiles/blob/7ed7d4f33bccc7ee63d74b6312cc6caa27844b63/.cargo/config) doesn't work for the window of compilers where `protocol = sparse` existed as a nightly-only option prior to stabilization.

A follow-up to this PR might shave a couple of minutes off of CI by opting into the sparse registry protocol.


iliana:
oh my god clippy

PR #2697, opened 2023-03-29
[WIP] [wicketd] structured updates

sunshowers:
This is a WIP that shows the general direction I'd like to take
wicketd's update process in. Putting this up here for visibility.

TODOs:

- [ ] Switch wire protocol to using update-engine types
- [ ] Switch wicket to using update-engine types
	- [ ] (using progenitor's `replace` -- search for `replace = ` in omicron -- will make our lives easier I'm pretty sure, so update_events.rs should probably move to its own crate called e.g. `wicket-types`)
- [ ] Handle installinator progress reports (probably easier to do once installinator is switched to the update engine)
- [ ] Do we want to return some metadata? I've just put in `serde_json::Value` as a placeholder for now but it might be reasonable to put in some sort of metadata there.


wicketd/src/update_tracker.rs
```diff
@@ -636,16 +1069,17 @@ impl UpdateDriver {
...
     }
 
-    async fn process_installinator_report(
-        &mut self,
-        mut report: ProgressReport,
-    ) {
+    async fn process_installinator_report(&self, mut report: ProgressReport) {
+        // XXX This needs to be rewritten once the installinator switches to the
+        // update engine. Note that errors returned by the installinator are not
+        // currently bubbled up!
```
sunshowers, line 1075: I think this should say "currently NOT bubbled up".

sunshowers:
This is now included in #2867, closed.

PR #2700, opened 2023-03-29
simplify DNS data types

davepacheco:
Following up on https://github.com/oxidecomputer/omicron/pull/2570#discussion_r1143903324, here's my attempt at simplifying the DNS-related types.  On the plus side, it's a considerable net reduction in code!

In terms of review, the essence of the change is just the change to `dns-server/src/dns_types.rs`:
https://github.com/oxidecomputer/omicron/pull/2700/files#diff-d3365fde6300987efb3dc8dd45ff52e22a8e55e33d4c6e5df1e8dfbfc5e4e721

Everything else just follows from that.  For the most part, instead of:

```rust
DnsKv {
    key: DnsRecordKey { name: name },
    records: records,
}
```

we have:

```rust
(name, records): (String, Vec<DnsRecord>)
```

And where we had a `Vec<DnsKv>`, we now have a `HashMap<String, Vec<DnsRecord>>`.

### DIFF
--- a/dns-server/src/bin/dnsadm.rs
+++ b/dns-server/src/bin/dnsadm.rs
@@ -18,12 +18,12 @@ use anyhow::Result;
 use clap::{Args, Parser, Subcommand};
 use dns_service_client::types::DnsConfig;
 use dns_service_client::{
-    types::{
-        DnsConfigParams, DnsConfigZone, DnsKv, DnsRecord, DnsRecordKey, Srv,
-    },
+    types::{DnsConfigParams, DnsConfigZone, DnsRecord, Srv},
     Client,
 };
 use slog::{Drain, Logger};
+use std::collections::BTreeMap;
+use std::collections::HashMap;
 use std::iter::once;
 use std::net::Ipv6Addr;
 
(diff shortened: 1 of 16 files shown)


dns-server/src/bin/dnsadm.rs
```diff
@@ -115,9 +115,12 @@ async fn main() -> Result<()> {
             for zone_config in &config.zones {
                 println!("\nzone {:?}", zone_config.zone_name);
 
-                for kv in &zone_config.records {
-                    println!("    key {:?}:", kv.key.name);
-                    for record in &kv.records {
+                // Sort the records so that we get consistent ordering.
+                let records: BTreeMap<_, _> =
```
rcgoodfellow, line 119: Should we store them in a `BTreeMap` to begin with so they're always in a predictable order?

rcgoodfellow (commented):
Hey @davepacheco this looks good to me. Just one comment on potentially using `BTreeMap` across the board.

dns-server/src/bin/dnsadm.rs
davepacheco: Yeah, I started with BTreeMap, but the progenitor-generated client uses a HashMap, and this is not configurable.

davepacheco:
Thanks for the review!

PR #2701, opened 2023-03-29
actions: bump everything

iliana:
There are [a few dozen deprecation warnings](https://github.com/oxidecomputer/omicron/actions/runs/4556020859) on any given Actions run, so I re-pinned everything to latest versions (and also pinned toml-cli).

(PR in draft mode while I see if any of this works.)

### DIFF
--- a/.github/workflows/check-opte-ver.yml
+++ b/.github/workflows/check-opte-ver.yml
@@ -6,7 +6,7 @@ on:
       - 'tools/opte_version'
 jobs:
   check-opte-ver:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v2
       - name: Install jq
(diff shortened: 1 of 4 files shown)


david-crespo (approved):
That all seems like a good idea. Only bit I'm not sure about is pinning a patch version of the actions, why not leave them at major?

iliana:
I pinned them to minor for the same reason we pin rust-toolchain.toml to minor: repeatability. Less of an opportunity for CI to change out from under us over time.

(I didn't go as far as [pinning to commit hashes](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions) because I find it more obnoxious and I don't think I am terribly worried about a malicious re-tag in `actions/checkout` or `Swatinem/rust-cache` for our remaining GHA use cases.)

PR #2694, opened 2023-03-29
Add dendrite-stub to test context

internet-diglett:
- [x] Add helpers for pulling correct dendrite stub version
- [x] Start instance during test context creation
- [x] Supply listening port to dendrite client configuration in relevant test cases
- [x] Add support for linux environments (requires publishing Linux `dendrite` binaries)

Closes #2629 
Closes #2649

PR #2703, opened 2023-03-29
`omicron-dev` could run all the things

davepacheco:
This adds a `run-all` command to `omicron-dev` that automates setting up a bunch of stuff: CockroachDB, Clickhouse, Nexus, simulated sled agent, etc.  It actually just creates a ControlPlaneTestContext so you're getting exactly the same setup that you'd have in a test.

As part of this I pulled `omicron-dev` out from the common package it was in so that it could depend on other stuff that it now needs to run.


david-crespo:
Great, would love to retire [this](https://github.com/oxidecomputer/console/blob/main/tools/start_api.sh) or cut it down to 10 lines.

ahl:
WHEN ARE WE GOING TO GET TO THE FIREWORKS FACTORY!

davepacheco:
This is ready for review!

david-crespo (approved):
This is great. Definitely replaces most of my [tmux script](https://github.com/oxidecomputer/console/blob/1b1cc264a7f5ac0a553cf53af69e54f15ec56228/tools/start_api.sh). I think I'll still use tmux to run the thing in one pane and open the logs I like to see in another pane or two.

I had to remember to add `SKIP_ASIC_CONFIG=1` to `cargo run --bin omicron-dev -- run-all` — always annoying because I don't find out until instance create 500s. Luckily #2694 gets rid of that. After adding the env var, my usual [populate DB script](https://github.com/oxidecomputer/console/blob/1b1cc264a7f5ac0a553cf53af69e54f15ec56228/tools/populate_omicron_data.sh) ran fine.

My main question: what is the best way to view the logs? It's kind of neat that everything gets through in the same log (I think) and then all you have to do is filter for what you want. I got tolerable results with `tail -f blah.log | rg nexus | npx bunyan` and could definitely improve it more by using a better grep patter or with bunyan's filtering (this gets me too much authz logs), but I wonder if there are some other good tricks. It might be worth documenting. In projects at previous jobs, the most generally relevant log commands were built right into the dev runner.

davepacheco:
> My main question: what is the best way to view the logs? It's kind of neat that everything gets through in the same log (I think) and then all you have to do is filter for what you want. I got tolerable results with `tail -f blah.log | rg nexus | npx bunyan` and could definitely improve it (this gets me too much authz logs) by using a better grep pattern or with bunyan's filtering, but I wonder if there are some other good tricks. It might be worth documenting. In projects at previous jobs, the most generally relevant log commands were built right into the dev runner.

Yeah, this is a good question.  When I first did the work to set up logging for the test suite (like, probably three years ago?) I thought it would be really valuable to have the logs go to one place.  The idea is that we'd use either the `name` or `component` fields of the logger/log messages to filter by subcomponent, and at the same time would be easy to view, say, client log messages interpersed with the corresponding server ones.   Unfortunately I don't think we've been diligent enough about assigning the name/component properties in a coherent way for this to work that well.  So I think `grep`/`rg` wind up being the best bet.  (Actually, I wind up using `grep -v` a lot to filter _out_ noisy stuff.)  Also, bunyan supports merging multiple files on the CLI, ordered by timestamp, so it doesn't _really_ matter if they're in one file or multiple.  But [looker doesn't support that yet](https://github.com/oxidecomputer/looker/issues/3).

I think your options are:

- I use `bunyan -l` a lot (e.g., `bunyan -linfo` to filter info-level and higher).  I feel like I've run into some issue I haven't debugged where using `bunyan` directly winds up skipping either some messages or the tail end of the log.
- `grep`/`rg` -- this is what I do most of the time, and it's okay but not great
- `bunyan -c` for programmatic filtering, like `bunyan -c 'this.name == "nexus"`.  I think this is a better basis for canned commands that show you, say, the Nexus log entries vs. the simulated sled agent ones, but we'd have to rationalize the names we use.  This is still useful for cases where you want to filter for a really specific kind of message.
- `json`/`jq` to filter the raw JSON and then pass it to [bunyan](https://github.com/trentm/node-bunyan) or [looker](https://github.com/oxidecomputer/looker).

david-crespo:
> we'd have to rationalize the names we use

...yeah. This line two `name` and 3 `component` 😂 

```json
{
  "msg": "authorize result",
  "v": 0,
  "name": "omicron-dev",
  "level": 20,
  "time": "2023-04-04T11:08:18.973144-05:00",
  "hostname": "Davids-M1-MBP.local",
  "pid": 84687,
  "component": "ExternalAuthn",
  "component": "nexus",
  "component": "ServerContext",
  "name": "e6bff1ff-24fb-49dc-a54e-c6a350cd4d6c",
  "result": "Ok(())",
  "resource": "SiloUser { parent: Silo { parent: Fleet, key: 001de000-5110-4000-8000-000000000000, lookup_type: ById(001de000-5110-4000-8000-000000000000) }, key: 001de000-05e4-4000-8000-000000004007, lookup_type: ById(001de000-05e4-4000-8000-000000004007) }",
  "action": "Read",
  "actor": "Some(Actor::UserBuiltin { user_builtin_id: 001de000-05e4-4000-8000-000000000003, .. })"
}
```

davepacheco:
I think part of what's unfortunate here is that in node-bunyan, when a child logger overrides `component`, you get a hierarchical component like `nexus/ServerContext/ExternalAuthn`.  I think this would be a good approach.  I'm not sure it's possible with slog without baking it into slog, because I don't think we can get the key-value pairs back out of the logger.

PR #2704, opened 2023-03-29
[nexus] Add silo name to /v1/me response

david-crespo:
There are places in the web console where we will want to use the silo name as a kind of root org name. We know from [RFD 357](https://rfd.shared.oxide.computer/rfd/0357#_external_dns_names) that in production we will have the silo name in the domain, but if we can instead pull that from the API we get a solution that works regardless of how the API and console are being served.

The one real Design Choice I'm making here is to have `views::CurrentUser` include `views::User` with `#[serde(flatten)]` rather than inlining the same fields. Initially I thought this was a neat way to make it clear that `CurrentUser`'s fields are a superset of `User`'s, but if you look at the tests you can see there are a couple of spots where we actually want to check a `User` against a `CurrentUser`, and having the `user` field makes that easy. If we wanted to inline the fields instead, we could of course compare the fields one by one.

I think I had attempted this change before and run into issues with permissions. I vaguely remember something about not being able to access the silo. But every authenticated user can read their silo:

https://github.com/oxidecomputer/omicron/blob/18177ba83b0c80e933b004c1263b11ca76fca7d6/nexus/db-queries/src/authz/omicron.polar#L152-L154

So I'm thinking I must have messed up by trying to do the silo fetch lower down with an authz context that had already been narrowed too far, if that makes sense.

### DIFF
--- a/nexus/src/app/iam.rs
+++ b/nexus/src/app/iam.rs
@@ -131,7 +131,7 @@ impl super::Nexus {
         let authz_silo = opctx
             .authn
             .silo_required()
-            .internal_context("loading current silo")?;
+            .internal_context("loading current user's silo")?;
         let silo_id = authz_silo.id().into();
         let (.., db_silo) = self.silo_lookup(&opctx, &silo_id)?.fetch().await?;
         Ok(db_silo)
--- a/nexus/src/app/iam.rs
+++ b/nexus/src/app/iam.rs
@@ -123,6 +123,20 @@ impl super::Nexus {
         Ok(db_silo_user)
     }
 
+    /// Fetch the currently-authenticated Silo user's Silo
+    pub async fn silo_user_fetch_silo(
+        &self,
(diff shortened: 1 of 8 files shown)


nexus/src/app/iam.rs
```diff
@@ -123,6 +123,20 @@ impl super::Nexus {
...
 
+    /// Fetch the currently-authenticated Silo user's Silo
+    pub async fn silo_user_fetch_silo(
+        &self,
+        opctx: &OpContext,
+    ) -> LookupResult<db::model::Silo> {
+        let authz_silo = opctx
+            .authn
+            .silo_required()
+            .internal_context("loading current silo")?;
```
zephraph, line 134: Given that (I believe) this is used for debugging it might be better to say something like `fetching silo for current user`

zephraph (approved):
Ship it

PR #2705, opened 2023-03-30
clarify DNS-related type names

davepacheco:
This change brings the type names used for describing internal DNS into better alignment with RFD 248 (and specifically [section 4.1](https://github.com/oxidecomputer/rfd/tree/master/rfd/0248#terminology)).  Quoting it here since it's not public:

> Consider the case where Nexus depends on CockroachDB. We say that Nexus is a **service**. It may have one or more **instances**, each being a different Unix process, typically running on a different host.
> 
> CockroachDB is also a service. It too may have many instances.
> 
> In the context of a service, we refer to instances of a dependent service as **backends**. Different backends of a service are interchangeable. So for Nexus, there’s only one service containing all the backends.
> 
> In the end, we might have:
> 
>     Service Nexus instance N1 at 192.168.0.10 port 12220
> 
>     Service Nexus instance N2 at 192.168.0.11 port 12220
> 
>     Service CockroachDB instance D1 at 192.168.0.6 port 26257
> 
>     Service CockroachDB instance D2 at 192.168.0.7 port 26257
> 
>     Service CockroachDB instance D3 at 192.168.0.8 port 26257
> 
> Nexus thus has two backends. CockroachDB has three backends.
> 
> For something like Sled Agent, each Sled Agent would be its own service with exactly one backend, since the Sled Agents cannot be treated as interchangeable with each other.

It's that last sentence where I found the code confusing.  _Each_ Sled Agent is its own service, but it's not part of ServiceName.  (Ditto each Crucible.)  And all the _other_ services have backends, but they're not in BackendName.

After this PR, there's only `ServiceName`.  This is the name of a service (as in the "n" part of DNS): it's the identifier you use to look up a single logical service, no matter how many backends it might have.  That means for stuff like CockroachDB, the variant has no data (it's just `ServiceName::Cockroach`).  For sled agents, the variant has a datum which is the id of the specific Sled Agent _service_ you want (`ServiceName::SledAgent(Uuid)`).

I also got rid of `SRV` -- this layer of indirection is no longer necessary (and it was confusing because it doesn't actually describe an SRV record -- it has no port information or anything like that.  It was just a wrapper that you'd use in a context where you were doing a DNS lookup of a service, but that's what `ServiceName` now is.)

### DIFF
--- a/internal-dns/src/config.rs
+++ b/internal-dns/src/config.rs
@@ -67,15 +67,13 @@ use std::collections::BTreeMap;
 use std::net::Ipv6Addr;
 use uuid::Uuid;
 
-/// Describes the DNS name used for a control plane host
-///
-/// This does not describe the AAA record itself.
+/// Used to construct the DNS name for a control plane host
 #[derive(Clone, Debug, PartialEq, PartialOrd)]
 enum Host {
-    /// Identifies an AAAA record for a sled.
+    /// Used to construct an AAAA record for a sled.
     Sled(Uuid),
 
-    /// Identifies an AAAA record for a zone within a sled.
+    /// Used to construct an AAAA record for a zone on a sled.
     Zone(Uuid),
 }
 
--- a/internal-dns/src/config.rs
+++ b/internal-dns/src/config.rs
@@ -60,46 +60,18 @@
 //!
 //! This module provides types used to assemble that configuration.
 
-use crate::names::{BackendName, ServiceName, DNS_ZONE};
+use crate::names::{ServiceName, DNS_ZONE};
 use anyhow::{anyhow, ensure};
 use dns_service_client::types::{DnsConfigParams, DnsConfigZone, DnsRecord};
 use std::collections::BTreeMap;
 use std::net::Ipv6Addr;
 use uuid::Uuid;
 
-/// Describes the DNS name that will be used for a control plane service
-///
-/// This does not describe the SRV record itself.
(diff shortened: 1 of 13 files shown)


internal-dns/src/config.rs
```diff
@@ -60,60 +60,32 @@
...
 ///
 /// This does not describe the AAA record itself.
 #[derive(Clone, Debug, PartialEq, PartialOrd)]
-enum AAAA {
+enum Host {
     /// Identifies an AAAA record for a sled.
     Sled(Uuid),
 
     /// Identifies an AAAA record for a zone within a sled.
     Zone(Uuid),
```
smklein, line 79: Should we update the commentary here? "Identifies an AAAA record" seems a little odd now.

internal-dns/src/names.rs
```diff
@@ -4,12 +4,12 @@
...
-use std::fmt;
+use uuid::Uuid;
 
 /// Name for the control plane DNS zone
 pub const DNS_ZONE: &str = "control-plane.oxide.internal";
 
-/// Names for services where backends are interchangeable.
+/// Names of services within the control plane
 #[derive(Clone, Debug, Hash, Eq, Ord, PartialEq, PartialOrd)]
 pub enum ServiceName {
```
smklein, line 14: Rad, cool to see this unified

smklein (approved):

PR #2709, opened 2023-03-30
[wicket-dbg] Fix hang in server after client disconnect

### LINKED ISSUE #2683: wicket-dbg connection hang
This is a bug currently affecting the not yet merged wicket-dbg in #2682. 

The wicket-dbg-server only allows 1 connection at a time. Closing the `wicket-dbg` shell and connecting again causes the new session to hang. It's possible that the server doesn't recognize the disconnection and a is hung in a `read` or something else. Should be straighforward to debug and fix. Just logging this so I don't forget.

andrewjstone:
The wicket-dbg-server wasn't checking for EOF on reads. It now does. Additionally, don't always reparse the size header on every short read over 4 bytes. This makes the code slightly more efficient, but also a bit clearer and more linear.

Lastly, add some logging in case of errors.

Fixes #2683


sunshowers:
Thanks!

sunshowers (approved):

PR #2710, opened 2023-03-30
[mgs] Send keepalives while attached to SP serial console

jgallagher:
As of https://github.com/oxidecomputer/hubris/pull/1214, the SP expects periodic keepalive packets if an MGS instance is attached to its host's serial console but not actively sending data. `faux-mgs` (which is what is used today in the lab to do serial console things) already has this change; this catches MGS proper up.


andrewjstone (approved):
Ship it!

PR #2712, opened 2023-03-30
Remove `verification_url_complete` from device auth response

david-crespo:
`verification_url_complete`, a URI that includes the auth code, is [optional in the spec ](https://www.rfc-editor.org/rfc/rfc8628#section-3.2)and we are no longer using it. The CLI [now](https://github.com/oxidecomputer/oxide-sdk-and-cli/pull/59) opens a console page with a form and the user has to type in the code they see printed in the terminal.

image


plotnick (approved):
LGTM, thanks for continuing to push on this!

PR #2713, opened 2023-03-30
Fixes omicron#2702: switch zone not getting an underlay address

Nieuwejaar:
When passing control from rss to sled-agent, we assign newly created zones addesses on the underlay network.  The switch zone is already running, so it doesn't get a new address.  We update the SMF properties of the services in the zone to be aware of the underlay subnet, but the underlay nic itself isn't updated.

With this change, applying the new underlay address becomes part of the "re-enable the switch zone" task.


smklein (approved):
Thanks for this fix - that totally explains the "works on reboot" behavior.

I need to boost the priority of the "self-assembling zones", so all this surgery becomes less mandatory after it's booted.

PR #2716, opened 2023-03-30
Flatten selectors; allow owned names to be looked up

zephraph:
This is a resurrection of #2647 with a simplified approach as originally suggested by @sunshowers. Instead of the approach I took last time, this introduces `OwnedName` in the lookup macro such that you can lookup resources via `_name` or `_name_owned`. This added flexibility enables the simplified selector approach (which I'll discuss more below) and allows us to pass owned names in a given scope when it may otherwise be impossible.

I've endeavored to change as little as possible and reduce the overall noise. 

---

The selector approach that I implemented in the V1 work is nice in that it's compositional. The niceness of it soft of ends there though. It has an over-reliance on serde's flatten. It's not the easiest to see all the properties that a selector contains. On top of that it's not possible to individually document parameters that are inherited from a parent selector. This point is important because our hierarchy is only explicit through error handling. I think it's important to be able to communicate through our docs that a certain parameter depends on another. Given that each parameter is _contextual_ to the selector in which it lives it's not possible to give unique docs to each given their implementations. 

@davepacheco noted in the last PR
> ...with the nested "Russian doll" approach, the hierarchy wasn't duplicated in each struct, which seems better? (It seems like lots of opportunities for error to duplicate these fields in a lot of places.) The previous approach presumably makes it easier to write functions that operate on the selector for some middle node in the hierarchy (like Vpc), whereas it seems like you can't do that today because things further in the hierarchy don't use a VpcSelector any more?

That feedback is still relevant but I think it's important to keep in mind the role that selectors play in our system. A selector is constructed at the API layer and _should not_ be passed to the app layer (beyond performing lookups). We specifically want lookups to be the primitives that we're passing around at the app layer. Given their intentional limited exposure I feel like the flattened selectors don't pose any outsized risk in producing error prone code. The makeup of the selector itself is enforced by its lookup (which every selector has) and the lookups themselves are recursive (such that any selector only cares about its principle parameter and delegates the remaining parameters to the next lookup handler in the chain).

### DIFF
--- a/nexus/src/app/sagas/instance_create.rs
+++ b/nexus/src/app/sagas/instance_create.rs
@@ -1677,15 +1677,12 @@ pub mod test {
         let nexus = &cptestctx.server.apictx().nexus;
         let opctx = test_opctx(&cptestctx);
 
-        let project_selector = params::ProjectSelector {
-            project: PROJECT_NAME.to_string().try_into().unwrap(),
-        };
         let instance_selector = params::InstanceSelector {
-            project_selector: Some(project_selector),
+            project: Some(PROJECT_NAME.to_string().try_into().unwrap()),
             instance: INSTANCE_NAME.to_string().try_into().unwrap(),
         };
         let instance_lookup =
-            nexus.instance_lookup(&opctx, &instance_selector).unwrap();
+            nexus.instance_lookup(&opctx, instance_selector).unwrap();
         nexus.project_destroy_instance(&opctx, &instance_lookup).await.unwrap();
     }
 
--- a/nexus/src/app/vpc_router.rs
+++ b/nexus/src/app/vpc_router.rs
@@ -8,7 +8,6 @@ use crate::authz;
 use crate::db;
 use crate::db::lookup;
 use crate::db::lookup::LookupPath;
-use crate::db::model::Name;
 use crate::db::model::RouterRoute;
 use crate::db::model::VpcRouter;
 use crate::db::model::VpcRouterKind;
@@ -23,7 +22,6 @@ use omicron_common::api::external::LookupResult;
 use omicron_common::api::external::NameOrId;
 use omicron_common::api::external::RouterRouteKind;
 use omicron_common::api::external::UpdateResult;
-use ref_cast::RefCast;
 use uuid::Uuid;
 
 impl super::Nexus {
@@ -31,34 +29,36 @@ impl super::Nexus {
     pub fn vpc_router_lookup<'a>(
         &'a self,
         opctx: &'a OpContext,
-        router_selector: &'a params::RouterSelector,
+        router_selector: params::RouterSelector,
     ) -> LookupResult<lookup::VpcRouter<'a>> {
         match router_selector {
             params::RouterSelector {
                 router: NameOrId::Id(id),
-                vpc_selector: None,
+                vpc: None,
+                project: None
             } => {
                 let router = LookupPath::new(opctx, &self.db_datastore)
-                    .vpc_router_id(*id);
+                    .vpc_router_id(id);
                 Ok(router)
             }
             params::RouterSelector {
                 router: NameOrId::Name(name),
-                vpc_selector: Some(vpc_selector),
+                vpc: Some(vpc),
+                project
             } => {
                 let router = self
-                    .vpc_lookup(opctx, vpc_selector)?
-                    .vpc_router_name(Name::ref_cast(name));
+                    .vpc_lookup(opctx, params::VpcSelector { project, vpc })?
+                    .vpc_router_name_owned(name.into());
                 Ok(router)
             }
             params::RouterSelector {
                 router: NameOrId::Id(_),
-                vpc_selector: Some(_),
+                ..
             } => Err(Error::invalid_request(
-                "when providing vpc_router as an ID, vpc should not be specified",
+                "when providing router as an ID vpc and project should not be specified",
             )),
             _ => Err(Error::invalid_request(
-                "vpc_router should either be an ID or vpc should be specified",
+                "router should either be an ID or vpc should be specified",
             )),
         }
     }
@@ -141,34 +141,41 @@ impl super::Nexus {
     pub fn vpc_router_route_lookup<'a>(
         &'a self,
         opctx: &'a OpContext,
-        route_selector: &'a params::RouteSelector,
+        route_selector: params::RouteSelector,
     ) -> LookupResult<lookup::RouterRoute<'a>> {
         match route_selector {
             params::RouteSelector {
                 route: NameOrId::Id(id),
-                router_selector: None,
+                router: None,
+                vpc: None,
+                project: None,
             } => {
                 let route = LookupPath::new(opctx, &self.db_datastore)
-                    .router_route_id(*id);
+                    .router_route_id(id);
                 Ok(route)
             }
             params::RouteSelector {
                 route: NameOrId::Name(name),
-                router_selector: Some(selector),
+                router: Some(router),
+                vpc,
+                project,
             } => {
                 let route = self
-                    .vpc_router_lookup(opctx, selector)?
-                    .router_route_name(Name::ref_cast(name));
+                    .vpc_router_lookup(
+                        opctx,
+                        params::RouterSelector { project, vpc, router },
+                    )?
+                    .router_route_name_owned(name.into());
                 Ok(route)
             }
             params::RouteSelector {
                 route: NameOrId::Id(_),
-                router_selector: Some(_),
+                ..
             } => Err(Error::invalid_request(
-                "when providing subnet as an ID, vpc should not be specified",
+                "when providing route as an ID router, subnet, vpc, and project should not be specified",
             )),
             _ => Err(Error::invalid_request(
-                "router should either be an ID or vpc should be specified",
+                "route should either be an ID or router should be specified",
             )),
         }
     }
--- a/nexus/src/external_api/http_entrypoints.rs
+++ b/nexus/src/external_api/http_entrypoints.rs
@@ -921,7 +921,7 @@ async fn project_view(
         let project_selector =
             params::ProjectSelector { project: path.project };
         let (.., project) =
-            nexus.project_lookup(&opctx, &project_selector)?.fetch().await?;
+            nexus.project_lookup(&opctx, project_selector)?.fetch().await?;
         Ok(HttpResponseOk(project.into()))
     };
     apictx.external_latencies.instrument_dropshot_handler(&rqctx, handler).await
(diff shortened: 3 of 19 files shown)


nexus/src/app/sagas/instance_create.rs
```diff
@@ -1677,15 +1677,12 @@ pub mod test {
...
         let opctx = test_opctx(&cptestctx);
 
-        let project_selector = params::ProjectSelector {
-            project: PROJECT_NAME.to_string().try_into().unwrap(),
-        };
         let instance_selector = params::InstanceSelector {
-            project_selector: Some(project_selector),
+            project: Some(PROJECT_NAME.to_string().try_into().unwrap()),
             instance: INSTANCE_NAME.to_string().try_into().unwrap(),
         };
```
zephraph, line 1683: A clear benefit of this approach (in my opinion) is that you don't need build child selectors to fill out the parent given that it's just a flat struct.

nexus/src/app/vpc_router.rs
```diff
@@ -23,42 +22,43 @@ use omicron_common::api::external::LookupResult;
...
                 router: NameOrId::Id(_),
-                vpc_selector: Some(_),
+                ..
             } => Err(Error::invalid_request(
-                "when providing vpc_router as an ID, vpc should not be specified",
+                "when providing router as an ID vpc and project should not be specified",
             )),
             _ => Err(Error::invalid_request(
-                "vpc_router should either be an ID or vpc should be specified",
+                "router should either be an ID or vpc should be specified",
```
zephraph, line 61: These error messages were wrong before

nexus/types/src/external_api/params.rs
```diff
@@ -95,100 +95,114 @@ pub struct ProjectSelector {
...
+    // Name or ID of the project
+    pub project: Option<NameOrId>,
 }
 
 #[derive(Deserialize, JsonSchema)]
 pub struct DiskSelector {
-    #[serde(flatten)]
-    pub project_selector: Option<ProjectSelector>,
+    /// Name or ID of the project, only required if `disk` is provided as a `Name`
+    pub project: Option<NameOrId>,
```
zephraph, line 105: This level of documentation isn't possible with the former selector composition approach. I mentioned this in the description but now that our hierarchy isn't explicit via the hierarchy having correct per parameter documentation feels important.

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -1689,17 +1675,21 @@ async fn disk_import_blocks_from_url(
...
         let query = query_params.into_inner();
+        let params = finalize_params.into_inner();
+        let disk_selector =
+            params::DiskSelector { disk: path.disk, project: query.project };
+        let disk_lookup = nexus.disk_lookup(&opctx, disk_selector)?;
 
-        let opctx = crate::context::op_context_for_external_api(&rqctx).await?;
-
-        nexus.disk_finalize_import(&opctx, path.disk, query).await?;
+        nexus.disk_finalize_import(&opctx, &disk_lookup, &params).await?;
```
zephraph, line 1692: While I was going through and cleaning up selectors I noticed this from #2418. There was a few things about this implementation that deviated from some of the V1 patterns. Of primary concern is passing selectors directly into the app layer. In general we want to resolve all selectors in the handlers and only pass lookups down to the app layer. Also inside of `FinalizeDisk` there was a query parameter that had runtime implications. Aside from selectors, we try to avoid query parameters that have side effects in post bodies. I separated out the selectors and moved the side effect (`snapshot_name`) into the post body.

nexus/types/src/external_api/params.rs
```diff
@@ -1117,11 +1131,11 @@ pub struct ImportBlocksBulkWrite {
...
-    #[serde(flatten)]
-    pub project_selector: Option<ProjectSelector>,
-
-    /// an optional snapshot name
-    pub snapshot_name: Option<String>,
+    /// If specified a snapshot of the disk will be created with the given
+    /// name at the end of finalization. If not specified, a snapshot for
+    /// the disk will _not_ be created. A snapshot can be manually created
+    /// once the disk transitions into the `Detached` state.
+    pub snapshot_name: Option<Name>,
```
zephraph, line 1138: @jmpesp could you read over this and let me know if I've represented it accurately?

david-crespo (commented):
> A selector is constructed at the API layer and should not be passed to the app layer (beyond performing lookups). We specifically want lookups to be the primitives that we're passing around at the app layer.

I agree with this very much. And I buy that the flattened selectors are better on net despite (or maybe because of) losing the hierarchical structure. I like that we can give the fields contextually specific descriptions. However, the explicitly named `*_owned` methods on lookups are rather horrible — it feels like a big leak of Rusty implementation detail into a fairly high-level interface — so I want to see the argument for why that is the only way to accomplish the above. Are there any non-`owned` methods still there? Do they need to be? Could these just be called `network_interface_name`, etc?

zephraph:
> However, the explicitly named *_owned methods on lookups are rather horrible — it feels like a big leak of Rusty implementation detail into a fairly high-level interface

Yeah, I agree with this in part. That's why in #2647 I changed all the named lookups to be owned. The consequence of that was that it ended up necessitating a _lot_ of changes across the codebase that aren't entirely relevant. @davepacheco rightly mentioned that for most cases that's kind of annoying because you need to clone names much more often. As @ahl likes to say, that was sort of the tail wagging the dog. The _only_ reason we need owned names is because we can't otherwise construct a selector inside a lookup function which is required to ensure the lookups are compositional. It's a very narrow case but an important aspect of ensuring the flattened selectors are still hierarchically related. So the `_name_owned` lookup functions are _only_ ever used in lookup functions. They're available elsewhere if needed, but we current have no other usages of them. If we generated the lookup from a macro you'd never even see an instance of it in the codebase outside the macro.

Long winded way to say, yes, it's weird, but it's isolated and accomplishes the goal with the least amount of workflow churn.

nexus/types/src/external_api/params.rs
jmpesp: I would reword it from `at the end of finalization` to `during finalization`, but otherwise this is accurate.

It also makes **way** more sense for this to be Name instead of String! Thanks for catching that.

davepacheco:
@zephraph wrote:

> @davepacheco noted in the last PR
> 
> > ...with the nested "Russian doll" approach, the hierarchy wasn't duplicated in each struct, which seems better? (It seems like lots of opportunities for error to duplicate these fields in a lot of places.) The previous approach presumably makes it easier to write functions that operate on the selector for some middle node in the hierarchy (like Vpc), whereas it seems like you can't do that today because things further in the hierarchy don't use a VpcSelector any more?
> 
> That feedback is still relevant but I think it's important to keep in mind the role that selectors play in our system. A selector is constructed at the API layer and _should not_ be passed to the app layer (beyond performing lookups). We specifically want lookups to be the primitives that we're passing around at the app layer. Given their intentional limited exposure I feel like the flattened selectors don't pose any outsized risk in producing error prone code. The makeup of the selector itself is enforced by its lookup (which every selector has) and the lookups themselves are recursive (such that any selector only cares about its principle parameter and delegates the remaining parameters to the next lookup handler in the chain).

Thanks for articulating that.  It makes a lot of sense.  But it looks like we _are_ still passing the selectors into the app layer in quite a lot of places.  I'm not sure the exposure is really that limited?

If I'm understanding right, this is the crux of your motivation:

> The selector approach that I implemented in the V1 work is nice in that it's compositional. The niceness of it sort of ends there though. It has an over-reliance on serde's flatten. It's not the easiest to see all the properties that a selector contains. On top of that it's not possible to individually document parameters that are inherited from a parent selector. This point is important because our hierarchy is only explicit through error handling. I think it's important to be able to communicate through our docs that a certain parameter depends on another. Given that each parameter is _contextual_ to the selector in which it lives it's not possible to give unique docs to each given their implementations.

That's very helpful.  Candidly, I still prefer the nested approach to this one mainly because of the duplication here (in definitions, in match blocks, etc.).  But I think this is mostly an aesthetic preference -- or, put differently, both options have tradeoffs and I can see going either way.  I do appreciate it's nice to be able to give more specific docs.  At the same time, I can see this being annoying if we want to change the hierarchy around, as in the case of "Project Groups" that have been discussed, or add new stuff (which will involve copying more code).

I'm interested to hear what others prefer here but I'm happy with however we want to proceed!

zephraph:
> it looks like we are still passing the selectors into the app layer in quite a lot of places. I'm not sure the exposure is really that limited?

The only place we should be passing the selectors to are to lookup functions. It's possible that the lookup functions themselves could be placed in a different module to isolate them from the rest of the app layer (though they'd still need to be attached to the nexus impl). The ideal case here is that every http entrypoint is responsible for constructing and resolving selectors to lookups which are the medium of communication with the app layer. Selectors are really just an implementation detail of the API.

>  Candidly, I still prefer the nested approach to this one mainly because of the duplication here (in definitions, in match blocks, etc.)

I can appreciate the duplication point, for sure. And I agree that it's a tradeoff we'd be making. I think it's still slightly mitigated by the fact that lookups enforce the hierarchy (which would guide the addition of a new layer of hierarchy via type checking failures add different layers of lookup). 

I don't really think either approach is ideal, but I still would like to give this way a shot.

david-crespo (approved):
Thanks for figuring all this out. I was ambivalent about this change because when we add have to new concepts I prefer them to be domain concepts that add features. When we have to add an inward-focused Rusty concept like owned name, I want to make sure it has limited and clear context of use. What made me nervous was imagining a dev trying to use a name and not know whether to use an owned name or a borrowed name. That would be pointless mental overhead. But I now think the use of owned names is limited enough that that won't happen.

Here is what persuades me that it’s worth moving forward:

1. The documentation advantage is very strong. Without flattening the selectors, we cannot put a context-specific doc comment on each query param. And I am persuaded that the only way to make flattened selectors work is to add this concept of owned names.
2. I do like being able see immediately what’s on the query params for an endpoint without clicking around, but as has been pointed out, this is a very minor advantage.
3. We only explicitly refer to owned names in the `*_lookup()` constructor functions like `instance_lookup`. The lookup macros include some logic around this too, but it does not leak out to the developer. To me, that eliminates most of the problem — someone adding a new endpoint doesn’t run into this distinction unless they are adding a new lookup constructor, and in that case they’re likely copying one that’s already there. And if, as Justin suggested, we follow up by moving those functions out of the `nexus` app layer (the only reason they’re in there is they depend on `nexus.db_datastore`), then we can achieve the goals of a) never passing selectors into the app layer, and b) limiting the use of `*_owned_name()` functions to a very narrow and clear spot. We already sort of achieve (a) if you squint because those functions aren’t _really_ part of the app layer.

Maybe in a followup we can eliminate the owned/not-owned name distinction by making all names the owned variety. We can try it and see if it feels worse or better on net. Names are limited to 63 characters so I am not too worried about the memory impact of copying them a bit more.

I appreciate the point that getting rid of nesting makes it harder to, e.g., change the project selector and have that change cascade to the other selectors, but after #2626 I know that most of the work involved in such a change is a lot of manual munging in the tests, so nesting doesn't save as much work as you might think.

zephraph:
I've created an [issue](https://github.com/oxidecomputer/omicron/issues/2822) about externalizing the lookup functions. I think as a part of that work we could take a look at alternative methods for specifying owned names without requiring the `_name_owned` methods.

PR #2719, opened 2023-03-31
[installinator] reuse types from common in client

sunshowers:
Saves us having to hand-write a bunch of conversions.


andrewjstone (approved):
NICE!

ahl:
with the caveat that I have no familiarity with this code, I want to mention that I've found that it's often beneficial for the client to just use generated types and not try to share types with the server. Of these types, only `CompletionEventKind` has any kind of associated `impl`.

I'd try to restructure the code to eliminate this installinator-common crate ... at least reduce its contents, but perhaps you've already tried that. FWIW that approach led to (what I think is) simpler code in omicron at least back in the day.

sunshowers:
@ahl I hear you, and I think if the current code was the extent of it I wouldn't be proposing the change. However the scope of installinator-common is going to grow significantly in some work I've been doing; writing 1000+ lines of boilerplate conversion code is really not worth it imho.

ahl:
To be clear: what I'm saying is that often one can avoid writing **any** boiler plate by using the generated types on the client-side. Also to be clear: I don't know the details of this code and don't claim to know if that approach is viable.

PR #2721, opened 2023-03-31
[wicket] Plumb SP and (not yet completed) RoT caboose information through MGS -> wicketd -> wicket

jgallagher:
The endpoint for fetching caboose information already existed in MGS; this PR adds calling it to wicket'd MGS poller and forwarding it to wicket. Wicket now populates the version strings it displays to the user from the caboose version string instead of the old `SpState` version integer.


wicketd/src/mgs.rs
```diff
@@ -200,7 +204,11 @@ async fn update_inventory(
         inventory
             .entry(id)
             .and_modify(|curr| {
-                if curr.state != state || curr.components.is_none() {
+                if curr.state != state
+                    || curr.components.is_none()
+                    || curr.caboose.is_none()
+                    || curr.rot.caboose.is_none()
```
jgallagher, line 210: I think this logic is wrong but wanted to get your input on fixing it. This conditional is used to determine whether we fetch all 3 of the components + SP caboose + RoT caboose. I think what this should do instead is:

1. If the state has changed, fetch all 3, because the SP might have rebooted / been updated / etc, and all our old data is potentially invalid. In this case, we should probably set components/caboose/rot.caboose to `None` first, so if fetching fails we don't have stale data that we keep around?
2. If the state is unchanged, only fetch {components,caboose,rot.caboose} if we're missing it (i.e., this step should trigger 0, 1, 2, or 3 MGS requests instead of the existing code which triggers 0 or 3).

jgallagher: Both of these are implemented in faa907c. If `state` has changed we now fetch all 3, and if state has not changed, we only fetch fields that are still `None`.

Issue #2656, opened 2023-03-24
Dropping VNICs: Make a futureunordered to eat the vnics


smklein:
As seen during dogfood setup, the sled agent tried to make a zone (oxz_nexus).

For reasons that do not really matter, Nexus died. 

In the logs, we saw this:

```
Failed to delete Link: Failed to delete vnic oxControlService89: Command [/usr/sbin/dladm delete-vnic oxControlService89] executed and failed with status: exit status: 1  stdout:   stderr: dladm: vnic deletion failed: link busy
Failed to delete Link: Failed to delete vnic oxControlPublic39: Command [/usr/sbin/dladm delete-vnic oxControlPublic39] executed and failed with status: exit status: 1  stdout:   stderr: dladm: vnic deletion failed: link busy
```

TL;DR: When the zone is getting torn down, the VNICs are still in use *somewhere*.

Chatting with @bnaecker , here are our conclusions:
- We don't know what interface we could use to ask the OS "when is this vnic no longer in use" -- such an interface may not exist.
- We *do* still want the sled agent to delete this vnic.

To resolve:
- We should make the `drop` functionality for VNICs hand off to an out-of-band queue, where they can be deleted asynchronously w.r.t. the drop function. This should basically just be a task that is re-trying to delete the vnics whenever they can be deleted.

jclulow:
FWIW, even if we had an interface that could tell you _some_ of the things that might be holding it open, I think that there are always going to be places where it's just a reference count and all we know is it hasn't gone to 0 yet.  As with all reference county things, it might be stuck above zero because of a bug where we forgot to drop the reference count, as well.

I think that means we'll want to consider:

- this might be a condition that _never_ terminates for a particular VNIC; we should keep retrying forever, but not let one stuck VNIC at the head of the queue block the destroy queue for other VNICs that are probably not stuck
- if a VNIC remains invulnerable for some length of time, that's a fact we want to hear about as some kind of reported soft failure that is likely the result of a software defect

This is also a property of other system resources we will be trying to tear down in future, like zones or file systems or even in some cases processes that we'd like to kill but seemingly cannot.

luqmana:
So there's also https://github.com/illumos/ipd/blob/master/ipd/0003/README.adoc which @citrus-it recently updated and has a [CR](https://code.illumos.org/c/illumos-gate/+/2662) out for. The relevant bit specifically is:

[...]

PR #2725, opened 2023-03-31
[sled-agent] Destroy VNICs in a background task

smklein:
- Implements a `Destructor` struct in `illumos-utils`, which helps asynchronously destroy objects. This should be a general tool in cases where...
  - You have an object that you want to destroy when it is dropped...
  - ... but destroying the object may be (1) asynchronous, (2) fallible, or (3) both.
- This `Destructor` is then added to the `VnicAllocator`, where it is referenced by created VNICs.
  - As a result: when VNICs are dropped, they are handled by a background task, until they are successfully destroyed.

Fixes https://github.com/oxidecomputer/omicron/issues/2656


illumos-utils/src/link.rs
```diff
@@ -178,29 +207,42 @@ impl Link {
...
-            eprintln!("Failed to delete Link: {}", e);
+        if let Some(destructor) = self.destructor.take() {
+            destructor
+                .enqueue_destroy(LinkDestruction { name: self.name.clone() });
         }
     }
 }
 
+// Represents the request to destroy a VNIC
+struct LinkDestruction {
```
bnaecker, line 218: Not sure if we'll delete other kinds of links, but if so, consider renaming to `VnicDestruction`

illumos-utils/src/link.rs
```diff
@@ -124,28 +155,24 @@ pub struct InvalidLinkKind(String);
...
-            Some(kind) => Ok(Self {
-                name: name.as_ref().to_owned(),
-                deleted: false,
-                kind,
-            }),
-            None => Err(InvalidLinkKind(name.as_ref().to_owned())),
-        }
+impl std::fmt::Debug for Link {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(
```
bnaecker, line 167: Nit: I tend to use [`Formatter::debug_struct`](https://doc.rust-lang.org/std/fmt/struct.Formatter.html#method.debug_struct) when I want to just limit the debug output to a subset of fields.

illumos-utils/src/destructor.rs
```diff
@@ -0,0 +1,389 @@
...
+            // Unwrap safety:
+            // - The DestructorWorker can only be stopped by "try_close"
+            // - "try_close" can only succeed if it's called on the last strong reference to the
+            // Destructor.
+            // - Since "enqueue_destroy" was called on a "Destructor", one more reference must
+            // exist.
+            .unwrap();
+    }
+
+    /// Closes the destructor if this is the only reference to it.
```
bnaecker, line 159: Maybe make a note that `Self` is returned in the `Result::Err` if there is more than one reference outstanding.

illumos-utils/src/destructor.rs
```diff
@@ -0,0 +1,389 @@
...
+        let mut exit = false;
+        loop {
+            tokio::select! {
+                _ = self.futs.next() => {
+                    if exit && self.futs.is_empty() {
+                        return;
+                    }
+                }
+                msg = self.rx.recv() => {
+                    let msg = match msg {
```
bnaecker, line 64: Maybe `let msg = msg.unwrap_or(Message::Exit)`?

illumos-utils/src/destructor.rs
```diff
@@ -0,0 +1,389 @@
...
+        // Note that the strong reference inside "obj"
+        // also counts as a strong reference.
+        let destructor = Destructor::<ObjectDestruction>::new();
+        let obj = Object { ctx: Some(()), destructor: destructor.clone() };
+        let destructor =
+            destructor.try_close().await.expect_err("Should have failed");
+        drop(obj);
+        assert!(destructor.try_close().await.is_ok());
+
+        // Dropping the detructor before the object is fine, if weird.
```
bnaecker, line 376: Spelling nit: "destructor"

bnaecker (approved):
Looks good, and great tests. I've a few small nits, but otherwise LGTM.

illumos-utils/src/link.rs
smklein: Done

illumos-utils/src/link.rs
smklein: Oooo, TIL - thanks! much better than manually making a string through `write`.

illumos-utils/src/destructor.rs
smklein: Good call, updated

illumos-utils/src/destructor.rs
smklein: Done

PR #2730, opened 2023-04-01
sled agent: defer to Propolis's instance state machine

gjcolombo:
This is the first in a series of PRs that will add basic live migration support. To see the whole commit sequence, see [this branch](https://github.com/gjcolombo/omicron/tree/gjcolombo/lets-migrate/8-migration-end).

---

Today, sled agent and Propolis each implement an instance state machine that defines an instance's state and the set of allowed state transitions. Having a two-headed state machine is perilous because it requires us to reason about (and correctly implement) what happens when one state machine rejects a transition that another accepts.

Simplify matters by removing most of the sled agent state machine and delegating almost completely to Propolis's state machine. Delegating to Propolis is appropriate here because (a) at the end of the day, any state transition Propolis rejects is de facto inapplicable no matter what its caller thinks, and (b) Propolis's state machine can "lead" sled agent's, because instance state changes requested by a guest OS are known to and applied in Propolis before sled agent finds out about them.

The most immediate consequence of this is that simulated instances, which derive their states from the `InstanceStates` state machine, are left adrift. To fix them, reimplement `SimInstance` more explicitly as a test double that implements a fake Propolis state machine and state queue, then fix up the tests that relied on the old state machine's properties. For most tests, the problem is that a single state change request can now produce multiple state transitions; these are fixed by changing the "instance poke" operation to drive an instance to its next resting state.

Tested in part by deploying a non-simulated Omicron and starting/stopping a few instances.

### DIFF
--- a/nexus/tests/integration_tests/instances.rs
+++ b/nexus/tests/integration_tests/instances.rs
@@ -298,15 +298,6 @@ async fn test_instances_create_reboot_halt(
     let instance = instance_next;
     instance_simulate(nexus, &instance.identity.id).await;
     let instance_next = instance_get(&client, &instance_url).await;
-    assert_eq!(instance_next.runtime.run_state, InstanceState::Starting);
-    assert!(
-        instance_next.runtime.time_run_state_updated
-            > instance.runtime.time_run_state_updated
-    );
-
-    let instance = instance_next;
-    instance_simulate(nexus, &instance.identity.id).await;
-    let instance_next = instance_get(&client, &instance_url).await;
     assert_eq!(instance_next.runtime.run_state, InstanceState::Running);
     assert!(
         instance_next.runtime.time_run_state_updated
@@ -379,15 +370,6 @@ async fn test_instances_create_reboot_halt(
     let instance_next =
         instance_post(&client, instance_name, InstanceOp::Reboot).await;
     assert_eq!(instance_next.runtime.run_state, InstanceState::Rebooting);
-    assert!(
-        instance_next.runtime.time_run_state_updated
-            > instance.runtime.time_run_state_updated
-    );
-
-    let instance = instance_next;
-    instance_simulate(nexus, &instance.identity.id).await;
-    let instance_next = instance_get(&client, &instance_url).await;
-    assert_eq!(instance_next.runtime.run_state, InstanceState::Starting);
     assert!(
         instance_next.runtime.time_run_state_updated
             > instance.runtime.time_run_state_updated
--- a/sled-agent/src/instance.rs
+++ b/sled-agent/src/instance.rs
@@ -12,6 +12,7 @@ use crate::params::SourceNatConfig;
 use crate::params::VpcFirewallRule;
 use crate::params::{
     InstanceHardware, InstanceMigrateParams, InstanceRuntimeStateRequested,
+    InstanceStateRequested,
 };
 use anyhow::anyhow;
 use futures::lock::{Mutex, MutexGuard};
@@ -504,10 +505,7 @@ impl Instance {
         inner: &mut MutexGuard<'_, InstanceInner>,
     ) -> Result<PropolisSetup, Error> {
         // Update nexus with an in-progress state while we set up the instance.
-        let desired = inner.state.desired().clone();
-        inner
-            .state
-            .transition(InstanceState::Starting, desired.map(|d| d.run_state));
+        inner.state.transition(InstanceState::Starting);
         inner
             .lazy_nexus_client
             .get()
@@ -779,10 +777,16 @@ impl Instance {
         }
     }
 
-    /// Transitions an instance object to a new state, taking any actions
-    /// necessary to perform state transitions.
+    /// Attempts to change the current state of the instance by issuing an
+    /// appropriate state change command to its Propolis.
     ///
-    /// Returns the new state after starting the transition.
+    /// Returns the instance's state after applying any changes required by this
+    /// call. Note that if the instance's Propolis is in the middle of its own
+    /// state transition, it may publish states that supersede the state
+    /// published by this routine in perhaps-surprising ways. For example, if an
+    /// instance begins to stop when Propolis has just begun to handle a prior
+    /// request to reboot, the instance's state may proceed from Stopping to
+    /// Rebooting to Running to Stopping to Stopped.
     ///
     /// # Panics
     ///
@@ -792,17 +796,36 @@ impl Instance {
         target: InstanceRuntimeStateRequested,
     ) -> Result<InstanceRuntimeState, Error> {
         let mut inner = self.inner.lock().await;
-        if let Some(action) = inner
-            .state
-            .request_transition(&target)
-            .map_err(|e| Error::Transition(e))?
-        {
-            info!(
-                &inner.log,
-                "transition to {:?}; action: {:#?}", target, action
-            );
-            inner.take_action(action).await?;
+        match target.run_state {
+            InstanceStateRequested::Running => {
+                inner
+                    .propolis_state_put(
+                        propolis_client::api::InstanceStateRequested::Run,
+                    )
+                    .await?;
+            }
+            InstanceStateRequested::Stopped
+            | InstanceStateRequested::Destroyed => {
+                inner
+                    .propolis_state_put(
+                        propolis_client::api::InstanceStateRequested::Stop,
+                    )
+                    .await?;
+                inner.state.transition(InstanceState::Stopping);
+            }
+            InstanceStateRequested::Reboot => {
+                inner
+                    .propolis_state_put(
+                        propolis_client::api::InstanceStateRequested::Reboot,
+                    )
+                    .await?;
+                inner.state.transition(InstanceState::Rebooting);
+            }
+            InstanceStateRequested::Migrating => {
+                unimplemented!("Live migration not yet implemented");
+            }
         }
+
         Ok(inner.state.current().clone())
     }
 
--- a/sled-agent/src/instance.rs
+++ b/sled-agent/src/instance.rs
@@ -796,34 +796,28 @@ impl Instance {
         target: InstanceRuntimeStateRequested,
     ) -> Result<InstanceRuntimeState, Error> {
         let mut inner = self.inner.lock().await;
-        match target.run_state {
+
+        let (propolis_state, next_published) = match target.run_state {
             InstanceStateRequested::Running => {
-                inner
-                    .propolis_state_put(
-                        propolis_client::api::InstanceStateRequested::Run,
-                    )
-                    .await?;
+                (propolis_client::api::InstanceStateRequested::Run, None)
             }
             InstanceStateRequested::Stopped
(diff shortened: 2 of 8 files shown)


sled-agent/src/sim/collection.rs
```diff
@@ -160,6 +160,16 @@ impl<S: Simulatable> SimObject<S> {
     }
 }
 
+/// The mdoes of operation for the "poke" operation.
```
smklein, line 163:
```suggestion
/// The modes of operation for the "poke" operation.
```

sled-agent/src/sim/instance.rs
```diff
@@ -47,33 +76,109 @@ impl Simulatable for SimInstance {
...
-                InstanceStateRequested::Running => {
-                    PropolisInstanceState::Running
+        match target {
+            InstanceStateRequested::Running => match self.terminal_state() {
+                ApiInstanceState::Creating => {
+                    // The non-simulated sled agent explicitly and synchronously
+                    // publishes the "Starting" state when cold-booting a new VM
+                    // (so that the VM appears to be starting while its Propolis
+                    // process is being launched).
+                    self.state.transition(ApiInstanceState::Starting);
```
smklein, line 88: Yeah, see my earlier comment - shouldn't starting still be visible to users of the simulated sled agent?

sled-agent/src/sim/instance.rs
```diff
@@ -47,33 +76,109 @@ impl Simulatable for SimInstance {
...
-                InstanceStateRequested::Stopped => {
-                    PropolisInstanceState::Stopped
+                ApiInstanceState::Starting
+                | ApiInstanceState::Running
+                | ApiInstanceState::Rebooting
+                | ApiInstanceState::Migrating => {}
+                ApiInstanceState::Stopping
+                | ApiInstanceState::Stopped
+                | ApiInstanceState::Failed
+                | ApiInstanceState::Destroyed => {
```
smklein, line 99: Should we allow Failed / Destroyed instances to be "restarted" this way? Doesn't this imply behavior that could never happen in the non-simulated case?

sled-agent/src/sim/instance.rs
```diff
@@ -47,33 +76,109 @@ impl Simulatable for SimInstance {
...
+                    )))
+                }
+            },
+            InstanceStateRequested::Reboot => match self.terminal_state() {
+                ApiInstanceState::Running => {
+                    self.state.transition(ApiInstanceState::Rebooting);
+                    self.propolis_queue
+                        .push_back(PropolisInstanceState::Rebooting);
+                    self.propolis_queue
+                        .push_back(PropolisInstanceState::Running);
```
smklein, line 154: Is this idempotent? If we reboot while rebooting, will this enqueue another reboot?

sled-agent/src/sim/instance.rs
```diff
@@ -87,12 +192,28 @@ impl Simulatable for SimInstance {
...
+    fn desired(&self) -> Option<Self::RequestedState> {
+        self.propolis_queue.back().map(|terminal| match terminal {
+            PropolisInstanceState::Creating
+            | PropolisInstanceState::Starting
+            | PropolisInstanceState::Migrating
+            | PropolisInstanceState::Repairing
+            | PropolisInstanceState::Failed => panic!(
+                "terminal state {:?} doesn't map to a requested state",
+                terminal
+            ),
```
smklein, line 204: The safety on this is just that "whenever we `push_back` to the `propolis_queue`, we don't *end* with these states, right?

smklein (approved):

sled-agent/src/sim/instance.rs
gjcolombo: That's right--a state change request might queue one of these states as a transitional state, but it should never be the very last state on the queue. This is supposed to be an internal invariant in this module, hence the panic. I'll add a comment clarifying this.

sled-agent/src/sim/instance.rs
gjcolombo: Currently we will ([relevant Propolis code](https://github.com/oxidecomputer/propolis/blob/master/bin/propolis-server/src/lib/vm/request_queue.rs#L352)), but I could buy that that's not quite the right behavior; it certainly differs from how we handle other state transitions, and I can see why it's not intuitive for things to behave this way. I filed [propolis#363](https://github.com/oxidecomputer/propolis/issues/363) to track over there. I'll fix the simulated behavior up in this PR as well (just to avoid having to remember to do it later).

gjcolombo (commented):
Thanks @smklein! I'll have revisions up shortly.

nexus/tests/integration_tests/instances.rs
```diff
@@ -307,15 +307,6 @@ async fn test_instances_create_reboot_halt(
             > instance.runtime.time_run_state_updated
     );
 
-    let instance = instance_next;
-    instance_simulate(nexus, &instance.identity.id).await;
-    let instance_next = instance_get(&client, &instance_url).await;
-    assert_eq!(instance_next.runtime.run_state, InstanceState::Starting);
```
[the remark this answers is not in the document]
smklein: Got it - no need to insert artificial states, just verifying that the simulated environment reflects reality.

sled-agent/src/sim/instance.rs
smklein: Gotcha. If this matches what we're doing in propolis, I'm fine with it, and having that supplementary issue to follow-up seems like the right next step.

sled-agent/src/sim/instance.rs
```diff
@@ -47,33 +84,113 @@ impl Simulatable for SimInstance {
...
+                    )))
                 }
-                InstanceStateRequested::Stopped => {
-                    PropolisInstanceState::Stopped
+            },
+            InstanceStateRequested::Reboot => match self.terminal_state() {
+                ApiInstanceState::Running => {
+                    // Further requests to reboot are ignored if the instance
+                    // is currently rebooting or about to reboot (though in this
+                    // simulation reboot
```
gjcolombo, line 159: Didn't :w at the right time. Will fix presently.

sled-agent/src/instance.rs
```diff
@@ -792,17 +796,36 @@ impl Instance {
...
-            .state
-            .request_transition(&target)
-            .map_err(|e| Error::Transition(e))?
-        {
-            info!(
-                &inner.log,
-                "transition to {:?}; action: {:#?}", target, action
-            );
-            inner.take_action(action).await?;
+        match target.run_state {
```
luqmana, line 799: it'd be nice to have just a single `propolis_state_put` command and just match to get the propolis `InstanceStateRequested` variant. Helps a bit with rightward drift.

That would possibly change the order of calls to `propolis_state_put` and `transition` for some arms I guess. But we do have the `inner` lock here so it's not like any observed propolis state can prempt us.

sled-agent/src/sim/instance.rs
```diff
@@ -47,33 +84,113 @@ impl Simulatable for SimInstance {
...
-            // These operations would typically be triggered via responses from
-            // Propolis, but for a simulated sled agent, this does not exist.
-            //
-            // Instead, we make transitions to new states based entirely on the
-            // value of "desired".
-            let observed = match desired.run_state {
-                InstanceStateRequested::Running => {
-                    PropolisInstanceState::Running
+        match target {
+            InstanceStateRequested::Running => match self.terminal_state() {
```
luqmana, line 90: Hmm, could we change `terminal_state`'s return type to be more explicit about what's the actual terminal state we're currently headed to vs the current state. Otherwise reading this match almost makes it seem like `Creating`/`Starting`/etc are valid terminal state when they aren't.

luqmana: Oh, though I suppose this is using terminal to mean "last" here rather than a steady state.

sled-agent/src/sim/collection.rs
```diff
@@ -536,91 +524,34 @@ mod test {
         assert!(!rnext.run_state.is_stopped());
         rprev = rnext;
 
+        // Propolis publishes its own transition to Stopping before it publishes
+        // Stopped.
```
luqmana, line 528: If we're going to just delegate to the propolis state machine, should we bother having sled agent publish a `Stopping` (and `Rebooting`) transition as well?

luqmana (commented):
Woo, thanks for tackling this Greg!

sled-agent/src/instance.rs
gjcolombo: Cleaned this up a bit in commit 4e7373c2. I think it's useful to keep the Propolis request and the internal state transition in order so that we don't have to undo state transitions when our Propolis calls fail (it's safe to do that from a synchronization point of view but still adds a bit of friction).

FWIW, this path will also get reworked somewhat in a [subsequent commit](https://github.com/oxidecomputer/omicron/commit/504fa56f1c33e62ba2786f1f7386c9fdffd8a5dd#diff-e58fdced2e69108a490916cca62b391c44fdde81d564de6f86b8d4e1febc5998R597).

sled-agent/src/sim/instance.rs
gjcolombo: I renamed this to "next resting state" in 892f0b56 to try to clarify this.

sled-agent/src/sim/collection.rs
gjcolombo: This tore me up a little bit while I was writing this. I ended up keeping this behavior to (try to) match what the old code did, but I could be persuaded to change it.

---

As background for anyone who chances across this discussion, on a stop/reboot request issued to Nexus, what happens is that

1. Nexus asks sled agent to stop the instance
2. Sled agent asks Propolis to stop the VM (or reboot it, but for now we'll focus on stopping)
3. Propolis accepts that request and then (internally) queues a "stop" action on the VM
4. Control returns to sled agent, which *immediately* changes the instance state to "Stopping" without waiting for Propolis to start processing the request
5. Sled agent returns this updated instance state to Nexus
6. Nexus writes the state returned in (4) back to CRDB
7. Nexus refetches the instance record and returns it from the stop API

The nice thing about step 4 is that it guarantees that someone has published "state = Stopping" to the database before the instance is refetched. If we don't (try to) update the record synchronously in this path, and Propolis doesn't pick up the stop request and publish a state change quickly enough, then step 7 will return that the instance is running. That seems... odd? I called `oxide instance stop`, and the command succeeded, but the instance is still running? Hmm.

That said, this can still behave strangely: if Propolis *does* publish a state change between step 6 and step 7, then the refetched instance will have whatever state Propolis published, which might not have anything at all to do with what the user requested (e.g. if Propolis handles a queued reboot request in that window). And in that sense it's most *consistent* just to cut out step 4 and say "whatever Propolis publishes is the current state, no ifs/ands/buts" instead of having this kind of edge case.

---

All told: my preference is to keep publishing the "present tense" states from sled agent at this point, just to try to match the old sled agent behavior as much as possible. But there's an interesting API design question here that's probably worth discussing further. I can file an issue for it.

luqmana (approved):

Issue #2720, opened 2023-03-31
omicron-nexus test flake: `thread panicked while panicking. aborting.` in ubuntu-20.04


iliana:
```
     Running `/work/oxidecomputer/omicron/target/debug/deps/omicron_nexus-1f0fdf2f9fdc6d9e`

running 37 tests
test app::instance::tests::test_serial_console_stream_proxying ... ok
thread panicked while panicking. aborting.
error: test failed, to rerun pass `-p omicron-nexus --lib`

Caused by:
  process didn't exit successfully: `/work/oxidecomputer/omicron/target/debug/deps/omicron_nexus-1f0fdf2f9fdc6d9e` (signal: 6, SIGABRT: process abort signal)
```

Seen at least five times on ubuntu-20.04:
- https://buildomat.eng.oxide.computer/wg/0/details/01GWRA0YPXNR3TZ8CSN0E4VEB1/qJZCu7Uc2yakmOaapzS6QOiUw9Ocjo3oY885qYt4uvm4FKu6/01GWRA1WAS8BKQ5XASDKSB0XRC#S4986
- https://buildomat.eng.oxide.computer/wg/0/details/01GWR6YWEJ94WAZQS1C3EQ8TFP/tkdwNFMN1dj03FdMbrqHQs1ZrEE2yVkVYavYfbmpuZwlXuDr/01GWR70E48M9W68GPYVSNEGPBW#S4932
- https://buildomat.eng.oxide.computer/wg/0/details/01GWN705N1SM725V913K4RN0R1/zkY3yLDP53MJjJSHjau99qUCGVoUVpKXpr7veeaeIaojDNxS/01GWN70VSQ46H1DDHAPEJMH7N4#S4925
- https://buildomat.eng.oxide.computer/wg/0/details/01GWVYY585KA1B7GSTZPNS51SY/6aEmc0miFHAKpLbsnqXA1eBymJbu7CVFAWu023ZvZhbnCEVW/01GWVYYGJV035QDE63Y5Q1V2NC#S4985
- https://buildomat.eng.oxide.computer/wg/0/details/01GWT5XKHSF0E29YNR1MZ00SSK/e9Eqw5XQ9ypZPNQjHItx1JGXqgejTcdlwJtIto7yEqPYVBfe/01GWT5YD4C7RNG46BTYPTM5FCX#S4986

smklein:
[...]

PR #2746, opened 2023-04-03
[dns-server] Convert DNS panic to logged error

smklein:
In the context of https://github.com/oxidecomputer/omicron/issues/2720 , this makes errors easier to parse, as it prevents "panic-while-panicking" during an unwind.


davepacheco:
Sorry I didn't see this until after I did #2747.  I think that obviates this one.  I'm also working on something better in #2748.

PR #2748, opened 2023-04-03
panic on UpdateGuard misusage, but not if we're panicking

davepacheco:
This is a fix for #2720 that attempts to preserve the debuggability of UpdateGuard misusage by panicking like we were before, but not if we're currently panicking.  There's a detailed comment explaining what we're trying to do and why we're doing it this way.

### DIFF
--- a/dns-server/src/storage.rs
+++ b/dns-server/src/storage.rs
@@ -122,7 +122,7 @@ pub struct Store {
     log: slog::Logger,
     db: Arc<sled::Db>,
     keep: usize,
-    updating: Arc<Mutex<Option<UpdateInfo>>>,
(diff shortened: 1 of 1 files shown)


dns-server/src/storage.rs
```diff
@@ -123,6 +125,7 @@ pub struct Store {
     db: Arc<sled::Db>,
     keep: usize,
     updating: Arc<Mutex<Option<UpdateInfo>>>,
+    poisoned: AtomicBool,
```
jgallagher: I think this needs to be `Arc<AtomicBool>`; otherwise different `Store` clones will have distinct instances and won't see each others' stores.

jgallagher (approved):

PR #2749, opened 2023-04-03
configure dendrite for oximeter support

Nieuwejaar:
To produce statistics for oximeter, dendrite needs to be given a unique ID to identify itself and it needs to be able to find a nexus daemon to register with.  The change passes a sidecar uuid and list of DNS server addresses to dendrite via `smf` properties,

### DIFF
--- a/sled-agent/src/services.rs
+++ b/sled-agent/src/services.rs
@@ -2152,7 +2152,7 @@ mod test {
             EtherstubVnic(UNDERLAY_ETHERSTUB_VNIC_NAME.to_string()),
             Etherstub(BOOTSTRAP_ETHERSTUB_NAME.to_string()),
             SledMode::Auto,
-            None, // skip_timesync
+            Some(true),
             "rev-test".to_string(),
             SWITCH_ZONE_BOOTSTRAP_IP,
             vec![],
@@ -2182,7 +2182,7 @@ mod test {
             EtherstubVnic(UNDERLAY_ETHERSTUB_VNIC_NAME.to_string()),
             Etherstub(BOOTSTRAP_ETHERSTUB_NAME.to_string()),
             SledMode::Auto,
-            None, // skip_timesync
+            Some(true),
             "rev-test".to_string(),
             SWITCH_ZONE_BOOTSTRAP_IP,
             vec![],
@@ -2220,7 +2220,7 @@ mod test {
             EtherstubVnic(UNDERLAY_ETHERSTUB_VNIC_NAME.to_string()),
             Etherstub(BOOTSTRAP_ETHERSTUB_NAME.to_string()),
             SledMode::Auto,
-            None, // skip_timesync
+            Some(true),
             "rev-test".to_string(),
             SWITCH_ZONE_BOOTSTRAP_IP,
             vec![],
@@ -2252,7 +2252,7 @@ mod test {
             EtherstubVnic(UNDERLAY_ETHERSTUB_VNIC_NAME.to_string()),
             Etherstub(BOOTSTRAP_ETHERSTUB_NAME.to_string()),
             SledMode::Auto,
-            None, // skip_timesync
+            Some(true),
             "rev-test".to_string(),
             SWITCH_ZONE_BOOTSTRAP_IP,
             vec![],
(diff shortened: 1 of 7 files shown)


smf/sled-agent/gimlet-standalone/config.toml
```diff
@@ -14,6 +14,14 @@ sled_mode = "scrimlet"
 # this information.
 sidecar_revision = "b"
 
+# Assigns a unique name to the sidecar, which is used to tag its data in
+# oximeter.
+# TODO: we need a way to generate a UUID for a sidecar that is actually unique
+# and persists across reboots.  Presumably we want this UUID to be unique to
+# this switch and rack combination, so a new UUID will to be generated when the
+# switch is replaced or moved to a new rack.
+sidecar_name = "8f280386-1e33-482e-bf30-948a4e36790d"
```
smklein, line 23: These are all called `sidecar_name`, but isn't the value called `sidecar_uuid`? Are these getting parsed correctly?

sled-agent/src/services.rs
```diff
@@ -928,6 +950,13 @@ impl ServiceManager {
                                 "config/board_rev",
                                 &self.inner.sidecar_revision,
                             )?;
+                            smfh.setprop(
+                                "config/sidecar_uuid",
+                                &self
+                                    .inner
+                                    .sidecar_uuid
+                                    .unwrap_or_else(Uuid::new_v4),
```
smklein, line 958: Is this right? If we don't have a UUID for the sidecar when we're launching the service, and we just make one up here, it won't be stable at all. Isn't that a problem for associating data?

smklein (commented):
So I'm on-board with the general direction here, but it seems like we want a way for this "switch identifier" to be stable across reboots. I had a similar discussion with @bnaecker here: https://github.com/oxidecomputer/omicron/pull/2759#discussion_r1157798490

If this PR lands as-is, it seems like the UUIDs for the sidecars either:
1. Need to be bundled with the configs at build-time (which is bad)
or
2. Are generated on each reboot (also bad)

smf/sled-agent/gimlet-standalone/config.toml
Nieuwejaar: They are not at the moment.  I made a late change in one place but not the other.  Thanks.

Nieuwejaar:
> So I'm on-board with the general direction here, but it seems like we want a way for this "switch identifier" to be stable across reboots. I had a similar discussion with @bnaecker here: [#2759 (comment)](https://github.com/oxidecomputer/omicron/pull/2759#discussion_r1157798490)
> 
> If this PR lands as-is, it seems like the UUIDs for the sidecars either:
> 
> 1. Need to be bundled with the configs at build-time (which is bad)
>    or
> 2. Are generated on each reboot (also bad)

All correct.  In the config file I left that as a TBD, but I can certainly make it part of the initial push.

I think we would want to generate the UUID using the sled/scrimlet location and rack uuid, both of which are mentioned in the other issue you linked to.  This gives us persistence across reboots, but allows us to distinguish between the two sidecars in the same rack, as well as data collected from different racks.

It looks like the `rack_id` is available in `SledAgentInfo`,  The sled location doesn't seem to be available, but there is a sled ID available in `SledAgentInner`.  Would using those two IDs to generate one for the sidecar make sense?

smklein:
Does it really matter so much that the `rack_id` be incorporated into the switch tracking? If we had a stable identifier for the switch itself, wouldn't we able to track it, even as it (potentially) moves between racks?

Nieuwejaar:
> Does it really matter so much that the `rack_id` be incorporated into the switch tracking? If we had a stable identifier for the switch itself, wouldn't we able to track it, even as it (potentially) moves between racks?

If/when we have a serial number or somebody in the sidecar VPD, it's my understanding that it would only be available once `dpd` is running.  That is, `dendrite` would be responsible for getting its own UUID.  If I understand correctly how `oximeter` works, we would define a `Target` inside of `dendrite` that includes the rack and sled ID from `omicron`, along with the switch ID from `dendrite`.  This would let us collate traffic information based on location, error stats based on the switch ID, or any other combination.

smklein (approved):

Nieuwejaar:
Thanks for the review.  This will have to be queued up behind https://github.com/oxidecomputer/omicron/pull/2645, which I believe fixes the CI test failures caused by `softnpu` being out of sync.

PR #2753, opened 2023-04-03
add background tasks for propagating DNS

davepacheco:
This is still very much a work-in-progress.

This change does a few things:

- makes some changes to the way we keep track of services
    - The InternalDNS "service kind" is split into two separate services: "InternalDNS" (the DNS-protocol service) and "InternalDNSConfig" (the http service that configures InternalDNS).  To be clear, these run in the same process in the same illumos zone.  It's just that in terms of describing the things in the system, we treat them as separate services.
    - I have changed the internal representation of a "service" (which is to say: the database schema, database model, and the request made from Sled Agent to Nexus to report these services) to include not just the IP address, but the TCP/UDP port number.  In general, we need this to actually _use_ any network service and to advertise it in DNS.
- adds DNS propagation to Nexus, by adding:
    - a small subsystem in Nexus for running background tasks, which are activated on-demand and periodically
    - a background task for monitoring the DNS configuration in CockroachDB
    - a background task for monitoring the list of DNS servers in CockroachDB
    - a background task for propagating the DNS configuration to the list of DNS servers

I could split this up into a few different pieces if it's helpful but I wasn't sure it was worth it.


davepacheco:
@smklein This PR is not really ready for review, but I am interested in your feedback on:

> I have changed the internal representation of a "service" (which is to say: the database schema, database model, and the request made from Sled Agent to Nexus to report these services) to include not just the IP address, but the TCP/UDP port number.  In general, we need this to actually _use_ any network service and to advertise it in DNS.

Here's my thinking: I _think_ these "services" correspond to [RFD 248 "services"](https://github.com/oxidecomputer/rfd/tree/master/rfd/0248#terminology), which means they're network services provided to the rest of the control plane.  The expectation is that clients would find these services using a DNS query for SRV records.  They'd expect to get back a bunch of (IP, port) pairs and connect to one or more of the returned socket addresses, either over TCP or UDP.

Prior to this change, we only stored the IP of the service.  This isn't enough to connect to it.  I assume clients were assuming what the port was based on the default ports that we've hardcoded into omicron_common.  This doesn't work in a lot of contexts though, like the test suite where you want things to be able to bind to whatever ports are available and you want to communicate those ports (via DNS, eventually) to clients that want to talk to the services.

Obviously _somebody_ has to decide what port these things are listening on.  In my opinion, that should be the deployment code (so, RSS right now in a real system, and the test setup code in the test suite).  From that point on, the system should record the port used and pass it along (via DNS, eventually) to anybody that needs to use the service.  The point is nobody else should have to bake in the assumption of the standard port number.

I feel pretty good about this _if_ my assumption is right that these are supposed to represent RFD 248-style network services.  If they don't for some reason, then this might not make sense.  But they don't, then I don't really know what "service" is supposed to mean here.  The main thing that gives me pause is the recent addition of the dns_client "service".  I haven't figured out yet what this is used for but I don't _think_ there's a corresponding network service?

smklein:
I'm totally on-board with including the ports in the DB. Only possible exception would be "what if a service listens on multiple ports", but worst-case, we're back to inferring the port, which we already do today.

The `dns_client` was added by @citrus-it in https://github.com/oxidecomputer/omicron/pull/2628 - it's used to setup a DNS client within the NTP zone:

https://github.com/oxidecomputer/omicron/blob/3511c5f75df0fb8242c9c457c3e8a2169c3f39f4/sled-agent/src/services.rs#L544-L584

Since this is a client, I actually don't think it's going to have *any* IP address. We may want to refactor this information to not actually be a `service_kind`, but be accessed from Nexus some other way.

If you file a bug, I can help make that change?

nexus/src/app/background/common.rs
```diff
@@ -0,0 +1,257 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! Common facilities for background tasks
```
smklein, line 5: I'm especially interested in this module - I'd like to leverage this for service rebalancing, updates, periodic system monitoring, etc.

Lemme know when it's ready for me to take a look -- if we're searching for subcomponents of this PR to split out, I'd happily review this first.

davepacheco:
> I'm totally on-board with including the ports in the DB. Only possible exception would be "what if a service listens on multiple ports", but worst-case, we're back to inferring the port, which we already do today.
> 
> The `dns_client` was added by @citrus-it in #2628 - it's used to setup a DNS client within the NTP zone:
> 
> https://github.com/oxidecomputer/omicron/blob/3511c5f75df0fb8242c9c457c3e8a2169c3f39f4/sled-agent/src/services.rs#L544-L584
> 
> Since this is a client, I actually don't think it's going to have _any_ IP address. We may want to refactor this information to not actually be a `service_kind`, but be accessed from Nexus some other way.
> 
> If you file a bug, I can help make that change?

Great!  Filed https://github.com/oxidecomputer/omicron/issues/2761.  I wasn't sure if you were volunteering to drive it or just help but let me know what I can do.

nexus/src/app/background/common.rs
davepacheco: Will do.

The approach I've taken here is quite lightweight: there's a trait with one function called `activate()` and you impl that.  Then the subsystem creates and registers all of the impls of this trait (see `init()`).  Then the thing can be activated on some time interval or explicitly by other parts of Nexus.  There's some simple support for causing one of these to activate another one.  The `DnsPropagator` one uses this.  This is the `GenericWatcher` stuff.

There's a lot this _doesn't_ do, but I really wanted to avoid overengineering it ahead of time.

david-crespo:
Is 

> a small subsystem in Nexus for running background tasks, which are activated on-demand and periodically

meant to be a start on RFD 373?

davepacheco:
> Is
> 
> > a small subsystem in Nexus for running background tasks, which are activated on-demand and periodically
> 
> meant to be a start on RFD 373?

Basically yes, but emphasis on the "start" part!  These background tasks don't know anything about RPWs or reconcilers, intended state vs. actual state, targets, or anything like that.  I'm not yet clear how useful it will be to have framework for these parts.  I figured it'll be a lot easier to see what's common once we've implemented one or a few.

PR #2755, opened 2023-04-04
sled agent: remove unused migration-related types

gjcolombo:
Tidy up the migration-related types in sled agent's Nexus API. This is just a bit of pre-work to reduce the size of subsequent commits that change this API.

In `sled-agent/src/params.rs`:

- Rename migration-related types so that they indicate whether they're used with migration sources or migration targets (without having to look at their fields' names).
- Remove types and variants that were only used for migration sources.
  - Remove `InstanceRuntimeStateRequested` and replace it with `InstanceStateRequested`.
  - Remove the `Migrating` variant from `InstanceStateRequested`.

Sled agent does need an API that allows Nexus to set an instance's migration ID, but this doesn't (won't) actually change any Propolis state or user-visible instance state, so subsequent changes will implement this in a separate API.

Clean up the call sites that are affected by this change. For now, just add `todo!()`s and comment out the Nexus call sites that use migration types that were removed here. These will be replaced in subsequent commits that create sled agent's new migration APIs.

### DIFF
--- a/nexus/src/app/sagas/instance_create.rs
+++ b/nexus/src/app/sagas/instance_create.rs
@@ -27,7 +27,6 @@ use omicron_common::api::external::Name;
 use omicron_common::api::internal::nexus::InstanceRuntimeState;
 use serde::Deserialize;
 use serde::Serialize;
-use sled_agent_client::types::InstanceRuntimeStateRequested;
 use sled_agent_client::types::InstanceStateRequested;
 use slog::warn;
 use std::convert::TryFrom;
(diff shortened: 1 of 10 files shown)


nexus/src/app/sagas/instance_create.rs
```diff
@@ -1297,7 +1292,7 @@ async fn sic_instance_ensure(
                 &opctx,
                 &authz_instance,
                 &db_instance,
-                runtime_params,
+                InstanceStateRequested::Running,
```
luqmana, line 1295: Love to see the simplification!

gjcolombo:
Thanks for reviewing, @luqmana!

@smklein @jordanhendricks - since this PR is one of the smaller, less impactful ones in this sequence, I'm going to go ahead and merge it to unblock its considerably larger successor. The bits of code that this PR removed will come back in a different form in future PRs, so there should still be plenty of room for us to discuss the final shape of the sled agent/Nexus API.

Issue #2724, opened 2023-03-31
Do not hard code the switch zone IP


jmpesp:
The switch zone IP is currently hard coded in smf/nexus/config-partial.toml:

https://github.com/oxidecomputer/omicron/blob/743c159e079bc2a97e10a8485cb0e1658e1b5256/smf/nexus/config-partial.toml#L29

This is correct for single node deployments, but for multi node deployments where the switch zone is not on the sled with index 0, it will get a different IP:

```
root@oxz_switch:~# ipadm
ADDROBJ           TYPE     STATE        ADDR
lo0/v4            static   ok           127.0.0.1/8
lo0/v6            static   ok           ::1/128
ixgbe0/ll         addrconf ok           fe80::8261:5fff:fe07:41d8/10
ixgbe1/ll         addrconf ok           fe80::8261:5fff:fe07:41d9/10
ixgbe2/ll         addrconf ok           fe80::8261:5fff:fe11:ab30/10
ixgbe4/ll         addrconf ok           fe80::8edc:d4ff:feaf:d2a8/10
oxBootstrap0/ll   addrconf ok           fe80::8:20ff:feeb:da5e/10
oxBootstrap0/bootstrap6 static ok       fdb0:8061:5f11:ab31::2/64
oxControlService0/ll addrconf ok        fe80::8:20ff:fe25:2e15/10
oxControlService0/omicron6 static ok    fd00:1122:3344:104::2/64
```

This is causing instance creation to fail:

```
{
  "msg": "recording saga event",
  "v": 0,
  "name": "0234cca9-728d-479a-a3ef-a757ac52c39e",
  "level": 20,
  "time": "2023-03-31T14:37:40.40312969-07:00",
  "hostname": "oxz_nexus",
  "pid": 4132,
  "component": "ServerContext",
  "event_type": "Failed(ActionFailed { source_error: Object {\"InternalError\": Object {\"internal_message\": String(\"failed to query dpd: Communication Error: error sending request for url (http://[fd00:1122:3344:101::2]:12224/nat/ipv4/10.1.0.100/0): operation timed out\")}} })",
  "node_id": "50",
  "saga_id": "4d3e2ea1-feef-4db5-91c2-fdab379a3ddc"
}
```

smklein:
FYI @internet-diglett

smklein:
[...]

PR #2759, opened 2023-04-04
Don't hard-code underlay addresses in Nexus configs

smklein:
- Removes underlay addresses from `smf/nexus/config-partial.toml`. We should not be referencing underlay addresses in any shipped-in-production configurations!
- Updates Nexus to pull this address from the internal DNS server, rather than the config file
- Adds DNS records for the switch zone in the rack setup service (this relies on a simple "get sled role" API added to the sled agent)

Fixes https://github.com/oxidecomputer/omicron/issues/2724


nexus/src/app/mod.rs
```diff
@@ -210,9 +211,19 @@ impl Nexus {
...
+            &config.pkg.dendrite.address
+        {
+            (dpd_address.ip().to_string(), dpd_address.port())
+        } else {
+            let addr = resolver
+                .lock()
+                .await
+                .lookup_socket_v6(ServiceName::Dendrite)
+                .await
+                .map_err(|e| format!("Cannot access Dendrite address: {e}"))?;
```
smklein, line 224: So, while building this, I figured I'd mention - it's a little weird that we only have *one* `dpd_client` -- there are multiple scrimlets, so should we have a `Vec` of `dpd_client`s?

bnaecker: Probably, yes. Which one(s) need to be used depends on the operation, but yeah, we'll need to be able to talk to both. I don't know how to distinguish them at the level of DNS, though.

common/src/nexus_config.rs
```diff
@@ -285,6 +286,7 @@ pub struct PackageConfig {
     #[serde(default)]
     pub tunables: Tunables,
     /// `Dendrite` dataplane daemon configuration
+    #[serde(default)]
     pub dendrite: DpdConfig,
```
jmpesp, line 290: With this PR, is this still required?

sled-agent/src/rack_setup/plan/service.rs
```diff
@@ -198,17 +221,34 @@ impl Plan {
...
-            let mut addr_alloc = AddressBumpAllocator::new(subnet);
+            let is_scrimlet = Self::is_sled_scrimlet(log, sled_address).await?;
 
+            let mut addr_alloc = AddressBumpAllocator::new(subnet);
             let mut request = SledRequest::default();
 
+            // Scrimlets get DNS records for running dendrite
+            if is_scrimlet {
+                let id = Uuid::new_v4();
+                let address = get_switch_zone_address(subnet);
```
jmpesp, line 240: I'm pretty sure that when creating the service plan with multiple scrimlets, this block of code will add multiple dendrite addresses to the DNS :+1:

jmpesp (approved):
I deployed this to the Canada region, and hard coding the dendrite address is no longer required. :rocket:

nexus/src/app/mod.rs
jmpesp: Pasting this here, I screwed up where I typed this: I'm pretty sure that when creating the service plan with multiple scrimlets, after this PR there will be multiple dendrite addresses for the Dendrite DNS record. This does mean that:

- RSS will have to wait for two scrimlets to appear as peers
- we should be creating dpd clients dynamically based on the DNS instead

jmpesp: To be clear - instead of once and storing it in the Nexus object

smklein: I filed https://github.com/oxidecomputer/omicron/issues/2760 - this should be possible by iterating over the results returned from the resolver: https://github.com/oxidecomputer/omicron/blob/e031c290bd09e1a5114427b5172256b73d697726/internal-dns/src/resolver.rs#L111-L114

common/src/nexus_config.rs
internet-diglett: Hmmm, I'm going to lean towards yes, it will be okay to resolve dendrite via internal DNS going forward, and yes, it will _probably_ break the integration tests if we remove `DpdConfig` right now. Hard for me to say for sure since I'm not up to speed on how DNS records will be created in the testing context. In #2694 we are creating a `dpd` instance per test like we do with cockroachdb and clickhouse.

nexus/src/app/mod.rs
bnaecker: I might be missing something, but we really need to treat every `dpd` client as distinct. If I remember the DNS parlance correctly, they are each different services, with exactly one backend. Is that what's happening here?

smklein: @bnaecker this PR should be generating separate AAAA records for each dendrite zone, but you're right, I believe this generates **one** SRV record to represent both dendrite services, even though they're distinct.

Do we have any way of distinguishing "switch 1" from "switch 2" currently? I could label them with SRV records that distinguish the dendrite instances from each other explicitly, but I want some reliable way of distinguishing them

jgallagher: > Do we have any way of distinguishing "switch 1" from "switch 2" currently? I could label them with SRV records that distinguish the dendrite instances from each other explicitly, but I want some reliable way of distinguishing them

You could (probably?) do this via MGS the same way we've talked about identifying gimlets: if there was a way for the scrimlet to self-report its switch's identifying info (serial number/board/revision), you could compare that against the identifying info MGS reports for "switch 1" vs "switch 2".

bnaecker: I agree the best answer is something like what's laid out in RFD 360 like you're saying: model / revision / serial as a unique ID. Unfortunately, I'm not sure how we get that data currently from host software on the Scrimlet. We've implemented a way to fetch the MAC address range programmed into the FRUID EEPROM, but not those other fields.

bnaecker: So, if this is definitely the data we want to use for identifying the DNS records, then we'll probably need to add a message to the `transceiver-controller` crate for fetching it. See [this issue](https://github.com/oxidecomputer/transceiver-control/issues/73) for tracking.

From talking more with Robert, it sounds like we probably want two different schemes in DNS for services like `dpd`:

### `<serial>.<type>.<az>`

- Serial number (from this hypothetical new message above),
- The type of machine the program runs on: `sled`, `switch` or `psc` at this point,
- The AZ, either a number or UUID, not sure yet.

### `<loc>.<rack>.<az>`

- A rack location, e.g., `sled23`, `switch0` or `psc0`,
- A rack number or UUID, not sure yet.
- An AZ number or UUID, not sure yet.


It seems like populating the serial-number-based DNS records would require a request to `dpd` (at this point listening on `::1`) and out to the SP over the management network. From John's comment, it sounds like the location-based name could come from MGS, but I'm definitely guessing.

jmpesp:
Deployed this again, still looks good :rocket:

nexus/src/app/mod.rs
smklein: I still think this is worth fixing, but I filed https://github.com/oxidecomputer/omicron/issues/2760 as a separate issue, where we can follow-up.

Issue #2315, opened 2023-02-02
Instances' sled assignments won't change if an instance is stopped and restarted


gjcolombo:
The Nexus external API's "instance start" command passes through to `Nexus::instance_start_runtime`:

https://github.com/oxidecomputer/omicron/blob/9d1bd5539225b6a225ba3655ceab852c71baa04a/nexus/src/external_api/http_entrypoints.rs#L3088

https://github.com/oxidecomputer/omicron/blob/9d1bd5539225b6a225ba3655ceab852c71baa04a/nexus/src/app/instance.rs#L358-L377

If I'm reading things right, this function selects the sled to which to send the runtime state update by looking at the instance record in CRDB without regard for the instance's current state:

https://github.com/oxidecomputer/omicron/blob/9d1bd5539225b6a225ba3655ceab852c71baa04a/nexus/src/app/instance.rs#L608-L619

This seems like the right thing to do if the instance is already incarnated on a sled somewhere. But if the instance is stopped and doesn't exist on any sled, this will try to create the instance on the sled on which it most recently ran, which might not have capacity for it (even though some other sled might). This function should distinguish the "instance already incarnated" and "instance stopped" cases and select a new sled in the latter case.

gjcolombo:
I suspect fixing this problem will more or less require instance start to become a saga, because once it's done, starting an instance will require a lot of attendant work (reserving space on a sled, setting up V2P mappings) that we need to be able to retry if interrupted and that may need to be undone if the entire attempt to start the instance fails.

PR #2765, opened 2023-04-04
sled agent: split ensure into "register" and "ensure state" APIs

### LINKED ISSUE #2209: Race in instance creation saga can cause an instance to be running when it's supposed to be stopped
Repro steps: none yet; found this bug by inspection.

Consider the following steps:

1. Instance creation saga begins and reaches the `sic_instance_ensure` phase
2. Nexus calls sled agent to ensure the instance exists and to drive it to the Running state
3. Saga executor crashes
4. Sled agent creates and starts the instance
5. Guest OS in the instance shuts down (e.g. because a user logged into it and shut it down, or because of a problem the guest detects)
6. Propolis stops and destroys the instance, reporting this to sled agent; sled agent observes these transitions and tears down and forgets about the instance
7. Saga executor restarts from the start of the `sic_instance_ensure` step; this recreates and starts the instance when it really should be stopped

There's much more discussion in the comments on [this gist](https://gist.github.com/gjcolombo/506c2016584e96f6adb20e47b3c51894). Some important notes from that discussion that influence how we can fix this:

- Sled agent can crash/restart at any time, so we can't rely on a sled agent having any particular memory of the commands Nexus has sent it
- Once the VM is running, guest software can request a shutdown at any time and for any reason, with no user intervention required, so we can't prevent this by e.g. trying to prevent the user from accessing the instance until we know the creation saga is done

### LINKED ISSUE #2517: Stopping a stopped instance creates a Propolis VM for it
Repro steps:

1. Create an instance and make sure it's stopped
2. `oxide instance stop -o org -p project instance`

Expected: operation succeeds immediately without any major side effects
Observed: sled agent creates a Propolis for the instance and prepares it to run, then immediately tears the Propolis down

Since the created Propolis gets an `instance_ensure` call, it creates all its objects and allocates resources for the VM (including pages to back the guest's memory) even though we don't intend to run it.

I have some changes drafted that I believe will address this problem (by allowing sled agent to return early if the instance being stopped isn't present on the sled).

gjcolombo:
This is the third PR in the [live migration series](https://github.com/oxidecomputer/omicron/compare/main...gjcolombo:omicron:gjcolombo/lets-migrate/8-migration-end). Not a net negative diff this time, I'm afraid :( But at least it fixes some outstanding issues!

---

Split the sled agent's `/instances/{id}` PUT endpoint into two endpoints:

- A PUT to `/instances/{id}` "registers" an instance with a sled. This creates a record for the instance in the manager, but does not start its Propolis and does not try to drive the instance to any particular state.
- A PUT to `/instances/{id}/state` attempts to change the state of a previously- registered instance's VM by starting it, stopping it, rebooting it, initializing by live migration, or unceremoniously destroying it. (This last case is meant to provide a safety valve that lets Nexus get an unresponsive Propolis off a sled.)

This allows the instance create saga to avoid a class of problems in which an instance starts, stops (due to user input to the VM), and then is errantly restarted by a replayed saga step: because sled agent will only accept requests to run a registered instance, and stopping an instance unregisters it, a replayed "run this VM" saga node won't restart the VM. The migration saga is vulnerable to a similar class of problem, so this groundwork is necessary to write that saga correctly.

A secondary benefit of this change is that operations on running instances (like "stop" and "reboot") no longer need to construct an (unused) `InstanceHardware` to pass to the sled agent's ensure endpoint.

Update the simulated sled agent to support these APIs, update callers in Nexus to use them, and split the instance create saga's "instance ensure" step into two steps as described above. This requires some extra affordances in simulated collections to support simulated disks, since instance state changes no longer go through a path where an instance's hardware manifest is available.

Finally, add some Nexus logging to record information about CRDB updates that Nexus applies when a call to sled agent produces a new `InstanceRuntimeState`, since these are handy for debugging.

Tested: cargo test; installed Omicron locally and played around with some instances.

Fixes #2209. Fixes #2517.

### DIFF
--- a/nexus/src/app/instance.rs
+++ b/nexus/src/app/instance.rs
@@ -35,6 +35,7 @@ use omicron_common::api::external::NameOrId;
 use omicron_common::api::external::UpdateResult;
 use omicron_common::api::external::Vni;
 use omicron_common::api::internal::nexus;
+use sled_agent_client::types::InstancePutStateBody;
 use sled_agent_client::types::InstanceStateRequested;
 use sled_agent_client::types::SourceNatConfig;
 use sled_agent_client::Client as SledAgentClient;
@@ -327,19 +328,8 @@ impl super::Nexus {
         opctx: &OpContext,
         instance_lookup: &lookup::Instance<'_>,
     ) -> UpdateResult<db::model::Instance> {
-        // To implement reboot, we issue a call to the sled agent to set a
-        // runtime state of "reboot". We cannot simply stop the Instance and
-        // start it again here because if we crash in the meantime, we might
-        // leave it stopped.
-        //
-        // When an instance is rebooted, the "rebooting" flag remains set on
-        // the runtime state as it transitions to "Stopping" and "Stopped".
-        // This flag is cleared when the state goes to "Starting".  This way,
-        // even if the whole rack powered off while this was going on, we would
-        // never lose track of the fact that this Instance was supposed to be
-        // running.
         let (.., authz_instance, db_instance) = instance_lookup.fetch().await?;
-        self.instance_set_runtime(
+        self.instance_request_state(
             opctx,
             &authz_instance,
             &db_instance,
@@ -355,8 +345,29 @@ impl super::Nexus {
         opctx: &OpContext,
         instance_lookup: &lookup::Instance<'_>,
     ) -> UpdateResult<db::model::Instance> {
-        let (.., authz_instance, db_instance) = instance_lookup.fetch().await?;
-        self.instance_set_runtime(
+        // TODO(#2824): This needs to be a saga for crash resiliency
+        // purposes (otherwise the instance can be leaked if Nexus crashes
+        // between registration and instance start).
+        let (.., authz_instance, mut db_instance) =
+            instance_lookup.fetch().await?;
+
+        // The instance is not really being "created" (it already exists from
+        // the caller's perspective), but if it does not exist on its sled, the
+        // target sled agent will populate its instance manager with the
+        // contents of this modified record, and that record needs to allow a
+        // transition to the Starting state.
+        //
+        // If the instance does exist on this sled, this initial runtime state
+        // is ignored.
+        let initial_runtime = nexus_db_model::InstanceRuntimeState {
+            state: nexus_db_model::InstanceState(InstanceState::Creating),
+            ..db_instance.runtime_state
+        };
+        db_instance.runtime_state = initial_runtime;
+        self.instance_ensure_registered(opctx, &authz_instance, &db_instance)
+            .await?;
+
+        self.instance_request_state(
             opctx,
             &authz_instance,
             &db_instance,
@@ -373,7 +384,7 @@ impl super::Nexus {
         instance_lookup: &lookup::Instance<'_>,
     ) -> UpdateResult<db::model::Instance> {
         let (.., authz_instance, db_instance) = instance_lookup.fetch().await?;
-        self.instance_set_runtime(
+        self.instance_request_state(
             opctx,
             &authz_instance,
             &db_instance,
@@ -395,13 +406,20 @@ impl super::Nexus {
     fn check_runtime_change_allowed(
         &self,
         runtime: &nexus::InstanceRuntimeState,
+        requested: &InstanceStateRequested,
     ) -> Result<(), Error> {
         // Users are allowed to request a start or stop even if the instance is
         // already in the desired state (or moving to it), and we will issue a
         // request to the SA to make the state change in these cases in case the
-        // runtime state we saw here was stale.  However, users are not allowed
-        // to change the state of an instance that's migrating, failed or
-        // destroyed.
+        // runtime state we saw here was stale.
+        //
+        // Users cannot change the state of a failed or destroyed instance.
+        // TODO(#2825): Failed instances should be allowed to stop.
+        //
+        // Migrating instances can't change state until they're done migrating,
+        // but for idempotency, a request to make an incarnation of an instance
+        // into a migration target is allowed if the incarnation is already a
+        // migration target.
         let allowed = match runtime.run_state {
             InstanceState::Creating => true,
             InstanceState::Starting => true,
@@ -409,7 +427,9 @@ impl super::Nexus {
             InstanceState::Stopping => true,
             InstanceState::Stopped => true,
             InstanceState::Rebooting => true,
-            InstanceState::Migrating => false,
+            InstanceState::Migrating => {
+                matches!(requested, InstanceStateRequested::MigrationTarget(_))
+            }
             InstanceState::Repairing => false,
             InstanceState::Failed => false,
             InstanceState::Destroyed => false,
@@ -427,9 +447,7 @@ impl super::Nexus {
         }
     }
 
-    /// Modifies the runtime state of the Instance as requested.  This generally
-    /// means booting or halting the Instance.
-    pub(crate) async fn instance_set_runtime(
+    pub(crate) async fn instance_request_state(
         &self,
         opctx: &OpContext,
         authz_instance: &authz::Instance,
@@ -437,10 +455,34 @@ impl super::Nexus {
         requested: InstanceStateRequested,
     ) -> Result<(), Error> {
         opctx.authorize(authz::Action::Modify, authz_instance).await?;
-
         self.check_runtime_change_allowed(
             &db_instance.runtime().clone().into(),
+            &requested,
         )?;
+
+        let sa = self.instance_sled(&db_instance).await?;
+        let instance_put_result = sa
+            .instance_put_state(
+                &db_instance.id(),
+                &InstancePutStateBody { state: requested },
+            )
+            .await
+            .map(|res| res.into_inner().updated_runtime);
+
+        self.handle_instance_put_result(db_instance, instance_put_result)
+            .await
+            .map(|_| ())
+    }
+
+    /// Modifies the runtime state of the Instance as requested.  This generally
+    /// means booting or halting the Instance.
+    pub(crate) async fn instance_ensure_registered(
+        &self,
+        opctx: &OpContext,
+        authz_instance: &authz::Instance,
+        db_instance: &db::model::Instance,
+    ) -> Result<(), Error> {
+        opctx.authorize(authz::Action::Modify, authz_instance).await?;
 
         // Gather disk information and turn that into DiskRequests
         let disks = self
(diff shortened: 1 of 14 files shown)


sled-agent/src/instance.rs
```diff
@@ -492,32 +525,147 @@ impl Instance {
...
+            // Nexus to reflect that the instance is starting (so that the
+            // external API will display this state while the zone is being
+            // started).
+            //
+            // Migration targets don't do this because the instance is still
+            // logically running (on the source) while the target Propolis is
+            // being launched.
+            if migration_params.is_none() {
+                inner.state.transition(InstanceState::Starting);
+                inner.publish_state_to_nexus().await?;
```
smklein, line 568: If this fails, are we relying on the saga to unwind our state?

If we exit with an error here, the `InstanceState` will be `Starting`, even though no propolis has actually been started. Perhaps we should handle an error here by transitioning to `Failed`?

smklein: I see we're doing something similar with `setup_result` below - maybe these transitions should be lumped into that block?

sled-agent/src/instance.rs
```diff
@@ -492,32 +525,147 @@ impl Instance {
...
+            }
+
+            let setup_result: Result<(), Error> = {
```
smklein, line 571: Is this supposed to be wrapped up in a lambda?

I believe this code is equivalent to the following: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=5e000e905c95ffe9541c0d62f66f834a

In which the errors from setting up propolis will never be assigned to `setup_result`, but which will instead cause the `propolis_ensure` function to exit.

In contrast, the following actually does let the result get a non-`Ok(())` value:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=50f83c62b40e5ebf27a22876828b0542

sled-agent/src/instance_manager.rs
```diff
@@ -87,97 +87,121 @@ impl InstanceManager {
...
+                    // path) to handle the case where a stop request arrived,
+                    // Propolis handled it, sled agent unregistered the
+                    // instance, and only then did a second stop request
+                    // arrive.
+                    InstanceStateRequested::Stopped
+                    | InstanceStateRequested::Destroyed => {
+                        return Ok(InstancePutStateResponse {
+                            updated_runtime: None,
+                        });
+                    }
```
smklein, line 195: (Not a request, more just thinking out loud)

This special casing makes me wonder if the sled agent should expose a separate API for fully "destroying the instance" instead of relying on particular state transitions to pull that off for us -- similar to how we divided "creation" away from "starting".

smklein (commented):
Great overhaul - really appreciate this work that not only moves us closer to migration, but also helps fix our idempotency issues.

One error-handling concern aside, this LGTM.

gjcolombo:
Force-pushed to pick up #2807; there are no changes in this push. Will work on feedback this afternoon PDT.

nexus/src/app/instance.rs
```diff
@@ -357,8 +347,26 @@ impl super::Nexus {
         opctx: &OpContext,
         instance_lookup: &lookup::Instance<'_>,
     ) -> UpdateResult<db::model::Instance> {
-        let (.., authz_instance, db_instance) = instance_lookup.fetch().await?;
-        self.instance_set_runtime(
+        // TODO-correctness: This needs to be a saga for crash resiliency
```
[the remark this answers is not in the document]
gjcolombo: Filed #2824 and updated the TODO in 46e7d379b.

Yeah, this should ultimately be a sub-saga within instance creation. Depending on when other things land (looking at you, #2315) we might also need to reorder things in the creation saga a bit to make sure that all of the start sub-saga steps can happen sequentially.

nexus/src/app/instance.rs
```diff
@@ -357,8 +347,26 @@ impl super::Nexus {
...
+        // target sled agent will populate its instance manager with the
+        // contents of this modified record, and that record needs to allow a
+        // transition to the Starting state.
+        let new_runtime = nexus_db_model::InstanceRuntimeState {
+            state: nexus_db_model::InstanceState(InstanceState::Creating),
+            ..db_instance.runtime_state
+        };
+        db_instance.runtime_state = new_runtime;
+        self.instance_ensure_registered(opctx, &authz_instance, &db_instance)
+            .await?;
```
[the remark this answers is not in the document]
gjcolombo: It's a no-op. I clarified this a bit in cefcf42e1.

I could probably pick this apart a little bit more as well. The reason this exists at all is that the call to register an instance needs an `InstanceHardware` that specifies an initial `InstanceRuntimeState`. Registration doesn't actually have to have that, I think, so long as sled agent is prepared to deal with the existence of instances that have no current runtime state. LMK if you'd like me to give this a shot in this change.

nexus/src/app/instance.rs
```diff
@@ -397,21 +405,25 @@ impl super::Nexus {
...
+        requested: &InstanceStateRequested,
     ) -> Result<(), Error> {
         // Users are allowed to request a start or stop even if the instance is
         // already in the desired state (or moving to it), and we will issue a
         // request to the SA to make the state change in these cases in case the
         // runtime state we saw here was stale.  However, users are not allowed
         // to change the state of an instance that's migrating, failed or
-        // destroyed.
+        // destroyed, except that multiple requests to migrate in are allowed
+        // for idempotency.
```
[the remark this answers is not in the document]
gjcolombo: Fixed in 8977ba432.

nexus/src/app/instance.rs
```diff
@@ -397,21 +405,25 @@ impl super::Nexus {
...
         let allowed = match runtime.run_state {
             InstanceState::Creating => true,
             InstanceState::Starting => true,
             InstanceState::Running => true,
             InstanceState::Stopping => true,
             InstanceState::Stopped => true,
             InstanceState::Rebooting => true,
-            InstanceState::Migrating => false,
+            InstanceState::Migrating => {
+                matches!(requested, InstanceStateRequested::MigrationTarget(_))
```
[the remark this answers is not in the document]
gjcolombo: That's tricky to do here because the body of `MigrationTarget` contains the source's Propolis address and Propolis ID, neither of which we'll have on this path.

The reworked migration saga registers the instance on the destination sled and passes it a specially-crafted `InstanceRuntimeState` that (more or less) reflects the state CRDB *will get* if migration actually succeeds. That means, in part, that its Propolis ID and address will be the destination Propolis's ID and address. That `InstanceRuntimeState` is the one that will be passed to `runtime` when the saga moves to the `MigrationTarget` state (and not what's currently in CRDB), so there's no way (without refactoring this routine to take extra information or having it do another query) to verify that the migration target params point to the instance's current source.

nexus/src/app/instance.rs
```diff
@@ -594,26 +628,67 @@ impl super::Nexus {
...
+    /// - `result`: The result of the relevant sled agent operation. If this is
+    ///   `Ok`, the payload is the updated instance runtime state returned from
+    ///   sled agent, if there was one.
+    ///
+    /// # Return value
+    ///
+    /// - `Ok(true)` if the caller supplied an updated instance record and this
+    ///   routine successfully wrote it to CRDB.
+    /// - `Ok(false)` if the sled agent call succeeded, but this routine did not
+    ///   update CRDB. This can happen either because sled agent didn't return
```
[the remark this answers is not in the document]
gjcolombo: Fixed in c1089f3b4.

sled-agent/src/http_entrypoints.rs
```diff
@@ -33,6 +33,7 @@ pub fn api() -> SledApiDescription {
         api.register(zpools_get)?;
         api.register(filesystem_put)?;
         api.register(instance_put)?;
+        api.register(instance_put_state)?;
```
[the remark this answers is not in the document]
gjcolombo: Touched this up in 09a68fbb5.

sled-agent/src/instance.rs
gjcolombo: Good catch. Fixed in 7672fcfa5.

sled-agent/src/instance.rs
gjcolombo: You're right, this is incorrect. Unless I'm missing something (very possible here!), turning it into a closure is nontrivial because [async closures are still unstable](https://github.com/rust-lang/rust/issues/62290). From what I've read, it might be possible to use a closure anyway if I write out a sufficiently gnarly `Future` return type for it, but in 7672fcf I went with a label-based approach that I think will do the job.

sled-agent/src/instance_manager.rs
gjcolombo: Acknowledged. I think that makes sense, but will leave things as they are in this particular PR.

smklein (approved):
Thanks for incorporating the feedback -- this looks good to merge!

gjcolombo:
Thanks for taking the time to review! The merge conflicts are mostly in `use` declarations; will clean those up and merge once main's build is fixed.

nexus/src/app/sagas/instance_create.rs
```diff
@@ -1310,58 +1365,72 @@ async fn sic_instance_ensure(
...
+        .map_err(ActionError::action_failed)?;
 
     osagactx
         .nexus()
-        .create_instance_v2p_mappings(&opctx, instance_id)
+        .instance_request_state(
+            &opctx,
+            &authz_instance,
+            &db_instance,
+            InstanceStateRequested::Destroyed,
```
luqmana, line 1400: Seems a little odd the undo for `ensure_registered` is via `instance_request_state`.

sled-agent/src/instance_manager.rs
```diff
@@ -87,97 +87,121 @@ impl InstanceManager {
...
-                        // instance from the instance manager. It does this based off the
-                        // instance's ID. Given that we just replaced said instance using
-                        // the same ID, that would inadvertantly remove our newly created
-                        // instance instead.
-                        // TODO: cleanup source instance properly
-                        std::mem::forget(old_instance);
-                    }
+            } else {
+                info!(&self.inner.log, "new instance");
+                let instance_log = self.inner.log.new(o!());
```
luqmana, line 144: Mind adding `instance_id` to the log context here

nexus/src/app/sagas/instance_create.rs
luqmana: Ah, I see sean's similar comment in `instance_manager`.

luqmana (approved):
Woo, thanks for getting this together Greg! One small nit but looks good to me. Also left some comments for some possible follow up

nexus/src/app/sagas/instance_create.rs
```diff
@@ -1281,15 +1331,20 @@ async fn sic_instance_ensure(
     if !params.create_params.start {
         let instance_id = db_instance.id();
         // If we don't need to start the instance, we can skip the ensure
-        // and just update the instance runtime state to `Stopped`
+        // and just update the instance runtime state to `Stopped`.
+        //
+        // TODO-correctness: This is dangerous if this step is replayed, since
+        // a user can discover this instance and ask to start it in between
+        // attempts to run this step.
```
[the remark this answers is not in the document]
gjcolombo: Another approach might be to make the "Creating" DB record the output of an earlier saga step instead of refetching on line 1324. That way, if the initial record has generation 1, all the attempts to run this step will attempt to transition to generation 2, but only the first will succeed. I added a comment describing this in 7e4c45b6f.

nexus/src/app/sagas/instance_create.rs
```diff
@@ -1310,58 +1365,72 @@ async fn sic_instance_ensure(
...
-async fn sic_noop(_sagactx: NexusActionContext) -> Result<(), ActionError> {
-    Ok(())
-}
-
-/// Ensure that the necessary v2p mappings exist for this instance
-async fn sic_v2p_ensure(
+async fn sic_instance_ensure_registered_undo(
     sagactx: NexusActionContext,
-) -> Result<(), ActionError> {
+) -> Result<(), anyhow::Error> {
```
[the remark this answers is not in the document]
gjcolombo: Our hand's forced here because Steno (currently) defines undo actions to return `anyhow::Error`: https://docs.rs/steno/latest/steno/type.UndoResult.html

sled-agent/src/instance_manager.rs
gjcolombo: Done in 2067fc1a3.

gjcolombo:
@smklein @luqmana I've added an explicit "unregister this instance" API to sled agent in commit a960d29. Could you PTAL when you have a free moment? It's not hugely complicated, I think, but it's a big enough change that I'd like to get at least one set of eyes on it before merging.

luqmana (approved):

gjcolombo:
Thanks as always for the reviews!

PR #2767, opened 2023-04-05
Generalize network_interface table to allow for new kinds of NICs.

luqmana:
Since I wanted to split up #2419 a bit, took the opportunity to get rid of the
duplicate table I added to get going initially.

Previously, all NIC records in the DB were tied to a guest instance. In
enabling OPTE usage for services, it'd be nice to be able to reuse a lot
of the same NetworkInterface machinery we already have without
duplicating it completely. This commit adds a new `kind` column to the
`network_interface` table which at the moment will be either 'instance'
(NICs attached to a guest VM and exposed in the external API) or
'service' (NIC associated with an internal control plane service). The
previous `instance_id` column is renamed to `parent_id` and the table to
which it refers to as a FK is now dependant on the kind (either
`instance` or `service`).

Since a lot of the db and authz lookup macros end up relying on the
specific column name, this also introduces database views for each kind
which can be queried as if they were their own tables. This also allows
differentiating the different kinds of NICs as necessary.

CockroachDB, unlike Postgres, does not allow inserting or updating into
simple view and so we also model the base table itself to execute
queries that modify it.

From an external perspective, this doesn't change anything (modulo some
renaming from `NetworkInterface*` -> `InstanceNetworkInterface*`) in
that all external APIs still only deal with instance NICs.

There are some basic definitions for service NICs included but those
will be more fleshed in subsequent commits.

### DIFF
--- a/nexus/db-model/src/network_interface.rs
+++ b/nexus/db-model/src/network_interface.rs
@@ -45,7 +45,7 @@ pub struct NetworkInterface {
 
     pub mac: MacAddr,
     // TODO-correctness: We need to split this into an optional V4 and optional V6 address, at
-    // least one of which will lways be specified.
+    // least one of which will always be specified.
     //
     // If user requests an address of either kind, give exactly that and not the other.
     // If neither is specified, auto-assign one of each?
--- a/nexus/db-model/src/network_interface.rs
+++ b/nexus/db-model/src/network_interface.rs
@@ -3,7 +3,10 @@
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 use super::{MacAddr, VpcSubnet};
+use crate::impl_enum_type;
+use crate::schema::instance_network_interface;
 use crate::schema::network_interface;
+use crate::schema::service_network_interface;
 use crate::Name;
 use chrono::DateTime;
 use chrono::Utc;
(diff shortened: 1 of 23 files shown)


nexus/db-model/src/network_interface.rs
```diff
@@ -14,36 +17,184 @@ use nexus_types::identity::Resource;
...
+    pub kind: NetworkInterfaceKind,
+    pub parent_id: Uuid,
+
     pub vpc_id: Uuid,
     pub subnet_id: Uuid,
+
     pub mac: MacAddr,
     // TODO-correctness: We need to split this into an optional V4 and optional V6 address, at
-    // least one of which will always be specified.
+    // least one of which will lways be specified.
```
zephraph, line 48:
```suggestion
    // least one of which will always be specified.
```

zephraph:
This is great. I was working on implementing silo level images and doing it in a way that was much harder than what you've done here. I'm definitely interested in following this approach.

nexus/db-model/src/network_interface.rs
```diff
@@ -14,36 +17,184 @@ use nexus_types::identity::Resource;
...
+    #[diesel(column_name = is_primary)]
+    pub primary: bool,
+}
+
+impl NetworkInterface {
+    /// Treat this `NetworkInterface` as an `InstanceNetworkInterface`.
+    ///
+    /// # Panics
+    /// Panics if this isn't an 'instance' kind network interface.
+    pub fn as_instance(self) -> InstanceNetworkInterface {
```
zephraph, line 110: Is there a reason not to do a `TryFrom` instead?

luqmana: Hmm, mostly just cause the only places I used it expect the conversion to be infallible and so using `TryFrom`/`TryInto` would just mean shuffling the panic/unwrap.

jmpesp (approved):
Looks good, one comment needs fixing

Issue #2761, opened 2023-04-04
`DNSClient` need not be a `ServiceKind`


davepacheco:
I gather that ["services" in Omicron correspond with the traditional idea of network services -- things available over the network via TCP or UDP and typically discovered by DNS](https://github.com/oxidecomputer/omicron/pull/2753#issuecomment-1495142734).  It sounds like [DNSClient](https://github.com/oxidecomputer/omicron/pull/2753#issuecomment-1496288040) need not be one of these.  It'd be helpful to make it _not_ so that we can build other infrastructure that assumes these really are network services (e.g., storing IPs and network ports in the database for these things).

PR #2768, opened 2023-04-05
Remove DNSClient from ServiceKind

smklein:
This PR aims to launch the `svc:/network/dns/client` identically to the way we did before, but by coupling it with the "NTP service". The DNSClient is an SMF service that we **do** want to launch, but it's not a "network service" which would receive an internal IP address.

Fixes https://github.com/oxidecomputer/omicron/issues/2761

### DIFF
--- a/sled-agent/src/params.rs
+++ b/sled-agent/src/params.rs
@@ -249,16 +249,30 @@ impl From<DatasetEnsureBody> for sled_agent_client::types::DatasetEnsureBody {
 )]
 #[serde(tag = "type", rename_all = "snake_case")]
 pub enum ServiceType {
-    Nexus { internal_ip: Ipv6Addr, external_ip: IpAddr },
-    InternalDns { server_address: SocketAddrV6, dns_address: SocketAddrV6 },
+    Nexus {
+        internal_ip: Ipv6Addr,
+        external_ip: IpAddr,
+    },
+    InternalDns {
+        server_address: SocketAddrV6,
+        dns_address: SocketAddrV6,
+    },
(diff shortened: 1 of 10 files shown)


sled-agent/src/params.rs
```diff
@@ -249,16 +249,30 @@ impl From<DatasetEnsureBody> for sled_agent_client::types::DatasetEnsureBody {
...
-    Tfport { pkt_source: String },
+    Dendrite {
+        asic: DendriteAsic,
+    },
+    Tfport {
+        pkt_source: String,
+    },
     CruciblePantry,
-    Ntp { servers: Vec<String>, boundary: bool },
-    DnsClient { servers: Vec<String>, domain: Option<String> },
```
smklein, line 261: This is the only non-formatting change -- the arguments to launch the `DnsClient` are passed through the NTP server

sled-agent/src/rack_setup/plan/service.rs
```diff
@@ -391,30 +391,22 @@ impl Plan {
...
-                        vec![
-                            ServiceType::Ntp {
-                                servers: boundary_ntp_servers.clone(),
-                                boundary: false,
-                            },
-                            ServiceType::DnsClient {
-                                servers: rack_dns_servers.clone(),
-                                domain: None,
-                            },
-                        ],
```
smklein, line 417: This is a reasonable "summary" of this PR -- whenever we ask for the NTP service, the "DNS client" service is now implied.

davepacheco (approved):
Thank you for doing this!

PR #2769, opened 2023-04-05
Add method column to the nexus tags file

david-crespo:
I noticed while looking at #2645 that sometimes multiple operations have the same path and the operation names don't always make it trivial to guess what the method is on each.

### DIFF
--- a/nexus/tests/integration_tests/commands.rs
+++ b/nexus/tests/integration_tests/commands.rs
@@ -124,8 +124,9 @@ fn test_nexus_openapi() {
 
     // Construct a string that helps us identify the organization of tags and
     // operations.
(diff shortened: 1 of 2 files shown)


nexus/tests/integration_tests/commands.rs
```diff
@@ -151,6 +152,7 @@ fn test_nexus_openapi() {
             .or_default()
             .push((
                 op.operation_id.as_ref().unwrap().to_string(),
+                method.to_string(),
```
zephraph, line 155:
```suggestion
                method.to_string().to_uppercase(),
```

Minor nit here, but I think uppercase versions would be more consistent w/ how method names are generally displayed.

zephraph (approved):
Solid improvement, thanks @david-crespo

PR #2772, opened 2023-04-05
Implement silo level images

zephraph:
Implements silo level images as a replacement for global images. 

This implementation preserves a single API endpoint `/v1/images` and a single database table to store all images. It uses database views like #2767 to map the notion of `ProjectImage` and `SiloImage` to the images table. From an authz perspective there are actually _three_ resources represented:`SiloImage`, `ProjectImage`, and `Image`.  As the names suggest, `SiloImage` is the child of a silo and `ProjectImage` is the child of a project. `Image` occupies an odd space where it's currently considered the child of a Silo though it technically straddles both positions in the hierarchy.


nexus/types/src/external_api/views.rs
```diff
@@ -128,22 +128,14 @@ pub struct GlobalImage {
...
 #[derive(ObjectIdentity, Clone, Debug, Deserialize, Serialize, JsonSchema)]
-#[serde(deny_unknown_fields)] // CC @ahl Is this fine?
 pub struct Image {
     #[serde(flatten)]
     pub identity: IdentityMetadata,
 
-    #[serde(flatten)]
-    pub parent_id: SiloOrProjectId,
+    /// ID of the parent project if the image is a project image
+    pub project_id: Option<Uuid>,
```
david-crespo, line 138: Probably worth going all the way and saying ", otherwise null", or even "On silo-scoped images this will be null.`

nexus/src/app/image.rs
```diff
@@ -24,54 +25,77 @@ use omicron_common::api::external::LookupResult;
...
+                let (authz_silo, authz_project) =
+                    project.lookup_for(authz::Action::CreateChild).await?;
+                (authz_silo, Some(authz_project))
+            }
+            ImageParentLookup::Silo(silo) => {
+                let (.., authz_silo) =
+                    silo.lookup_for(authz::Action::CreateChild).await?;
+                (authz_silo, None)
+            }
+        };
```
david-crespo, line 98: cute

common/src/sql/dbinit.sql
```diff
@@ -916,7 +918,51 @@ CREATE TABLE omicron.public.image (
...
+CREATE VIEW omicron.public.project_image AS
+SELECT
+    id,
+    name,
+    description,
+    time_created,
+    time_modified,
+    time_deleted,
+    silo_id,
+    project_id,
```
zephraph, line 930: Note that project images also include `silo_id`. This is necessary to convert them back to the base image type.

nexus/db-queries/src/authz/api_resources.rs
```diff
@@ -867,6 +867,23 @@ authz_resource! {
...
+}
+
+// This resource is a collection of _all_ images in a silo, including project images.
+authz_resource! {
+    name = "Image",
+    parent = "Silo",
+    primary_key = Uuid,
+    roles_allowed = false,
+    polar_snippet = InSilo,
+}
```
zephraph, line 885: It's notable here that I've used the `InSilo` polar snippet to define auth for the base image when that's not _strictly_ true. The only place `Image` is referred to in the auth system is when we're looking up an image based on ID. In that case, we don't know if it's a project or silo image so we must use the generic form of the image lookup. For the current implementation that means a user _must_ have read permissions on the Silo to even be able to look up a project image by ID. Subsequent authz checks validates the appropriate permissions on the project (if it's a project image) so we get the right checks on the resource eventually.

nexus/db-queries/src/db/datastore/image.rs
```diff
@@ -17,50 +18,149 @@ use crate::db::error::public_error_from_diesel_pool;
...
+                    &pagparams.map_name(|n| Name::ref_cast(n)),
+                ),
+            }
+            .filter(project_dsl::time_deleted.is_null())
+            .filter(project_dsl::project_id.eq(authz_project.id()))
+            .select(ProjectImage::as_select())
+            .load_async::<ProjectImage>(self.pool_authorized(opctx).await?)
+            .await
+            .map_err(|e| public_error_from_diesel_pool(e, ErrorHandler::Server))
+            .map(|v| v.into_iter().map(|v| v.into()).collect()),
```
zephraph, line 83: Note that we're still returning the generic `Image` type here so we have to map over the results and convert them in this case.

nexus/db-queries/src/db/datastore/image.rs
```diff
@@ -17,50 +18,149 @@ use crate::db::error::public_error_from_diesel_pool;
...
+                    dsl::name,
+                    &pagparams.map_name(|n| Name::ref_cast(n)),
+                ),
+            }
+            .filter(dsl::time_deleted.is_null())
+            .filter(
+                dsl::project_id
+                    .is_null()
+                    .or(dsl::project_id.eq(authz_project.id())),
+            )
```
zephraph, line 58: This is the special sauce of this logic branch. If an image has no `project_id` then it's a silo image.... and I'm just realizing I should probably be filtering it by silo id 😅

nexus/db-queries/src/db/datastore/image.rs
```diff
@@ -78,7 +178,48 @@ impl DataStore {
...
+        &self,
+        opctx: &OpContext,
+        authz_silo: &authz::Silo,
+        project_image: ProjectImage,
+    ) -> UpdateResult<Image> {
+        let silo_id = authz_silo.id();
+        let image: Image = project_image.into();
+        let name = image.name().clone();
+
+        opctx.authorize(authz::Action::CreateChild, authz_silo).await?;
```
zephraph, line 201: The process of promoting an image is essentially the same as creating a silo image.

zephraph: I'm realizing this is also missing permissions for the project image itself.

nexus/db-queries/src/db/datastore/image.rs
```diff
@@ -78,7 +178,48 @@ impl DataStore {
...
+
+        use db::schema::image::dsl;
+        let image: Image = Silo::insert_resource(
```
zephraph, line 204: Q: should this be `update_resource` or something instead?

nexus/db-queries/src/db/datastore/image.rs
```diff
@@ -78,7 +178,48 @@ impl DataStore {
...
+    }
+
+    pub async fn project_image_promote(
+        &self,
+        opctx: &OpContext,
+        authz_silo: &authz::Silo,
+        project_image: ProjectImage,
+    ) -> UpdateResult<Image> {
+        let silo_id = authz_silo.id();
+        let image: Image = project_image.into();
```
zephraph, line 198: Something I missed: this should be converted to a silo image first.

nexus/db-queries/src/db/lookup.rs
```diff
@@ -453,7 +483,7 @@ impl<'a> Root<'a> {
 lookup_resource! {
     name = "Silo",
     ancestors = [],
-    children = [ "IdentityProvider", "SamlIdentityProvider", "Project" ],
+    children = [ "IdentityProvider", "SamlIdentityProvider", "Project", "SiloImage" ],
```
zephraph, line 486: Should `Image` be added here? cc @davepacheco

nexus/db-queries/src/db/lookup.rs
```diff
@@ -540,6 +579,15 @@ lookup_resource! {
 
 lookup_resource! {
     name = "Image",
+    ancestors = ["Silo"],
+    children = [],
+    lookup_by_name = false,
```
zephraph, line 584: It's notable that this resource is _only_ used for ID lookups. If there's a name it's always either `SiloImage` or `ProjectImage` instead.

nexus/db-queries/src/db/lookup.rs
```diff
@@ -742,6 +790,18 @@ lookup_resource! {
...
+pub enum ImageLookup<'a> {
+    ProjectImage(ProjectImage<'a>),
+    SiloImage(SiloImage<'a>),
+}
+
+pub enum ImageParentLookup<'a> {
+    Project(Project<'a>),
+    Silo(Silo<'a>),
+}
+
```
zephraph, line 804: This might not really be the best place for these little helper enums, but I wasn't sure where else to put them.

nexus/src/app/image.rs
```diff
@@ -24,54 +25,77 @@ use omicron_common::api::external::LookupResult;
...
 use uuid::Uuid;
 
 impl super::Nexus {
-    pub fn image_lookup<'a>(
+    pub async fn image_lookup<'a>(
         &'a self,
         opctx: &'a OpContext,
         image_selector: params::ImageSelector,
-    ) -> LookupResult<lookup::Image<'a>> {
+    ) -> LookupResult<ImageLookup<'a>> {
```
zephraph, line 38: The return type here is an enum that wraps either a `ProjectImage` or a `SiloImage`. It's important that we returned the appropriate type for authz purposes.

nexus/src/app/image.rs
```diff
@@ -24,54 +25,77 @@ use omicron_common::api::external::LookupResult;
...
+                let (.., db_image) = LookupPath::new(opctx, &self.db_datastore)
+                    .image_id(id).fetch().await?;
+                let lookup = match db_image.project_id {
+                    Some(_) => ImageLookup::ProjectImage(LookupPath::new(opctx, &self.db_datastore)
+                        .project_image_id(id)),
+                    None => {
+                        ImageLookup::SiloImage(LookupPath::new(opctx, &self.db_datastore)
+                            .silo_image_id(db_image.silo_id))
+                    },
+                };
```
zephraph, line 53: This is a bit unique. Because there's only a single endpoint for lookup images by id (`/v1/images/{image_id}`) we don't actually know if it's a silo or project image when we get the ID. We first have to lookup the image and see if `project_id` is populated. If it is, we know it's a project id. Otherwise it's a silo id.

nexus/src/app/image.rs
zephraph: It helps avoid repeating the rather complex logic contained below.

nexus/db-model/src/image.rs
```diff
@@ -3,14 +3,15 @@
...
+use crate::schema::{image, project_image, silo_image};
 use db_macros::Resource;
 use nexus_types::external_api::views;
 use nexus_types::identity::Resource;
+use omicron_common::api::external::Error;
 use serde::{Deserialize, Serialize};
 use uuid::Uuid;
 
-// Project images
+// Shared image definition
```
david-crespo, line 14: It's not that hard to see this from the `TryFrom` impls and stuff, but I think it would be helpful to add a comment to the effect of:

```
// Image, ProjectImage, and SiloImage are almost identical. Image is the model
// for the underlying table and ProjectImage and SiloImage are backed by views
// of that table. They have all the same fields except ProjectImage has both a
// silo_id and a project_id, while SiloImage only has a silo_id. Image has a
// silo_id and an optional project_id to cover both possibilities
```

nexus/types/src/external_api/params.rs
```diff
@@ -115,6 +115,18 @@ pub struct SnapshotSelector {
     pub snapshot: NameOrId,
 }
 
+/// A specialized selector for image list, it contains an extra feield to indicate
```
david-crespo, line 118:
```suggestion
/// A specialized selector for image list, it contains an extra field to indicate
```

david-crespo (approved):
Looks great. We went over this in detail on a call. This has been a long journey.

nexus/db-queries/src/db/lookup.rs
davepacheco: I believe the only purpose of putting things here is to generate selectors on this lookup type that will look up child resources.  e.g., if you wanted to do `LookupPath(...).silo_id(...).image_name(...)`, then you'd want to put that here.  I haven't looked at the rest of this PR yet to know if you'd want that here but it seems like not since you should already have `SiloImage`.  But I think it's safe to leave it out if you're not sure.

zephraph: This base `Image` type will only ever be used to access an image on ID so it seems fine.

nexus/db-queries/src/db/datastore/image.rs
```diff
@@ -17,50 +19,152 @@ use crate::db::error::public_error_from_diesel_pool;
...
+
+        let name = image.name().clone();
+        let silo_id = image.silo_id;
+
+        use db::schema::image::dsl;
+        let image: Image = Silo::insert_resource(
+            silo_id,
+            diesel::insert_into(dsl::image)
+                .values(image)
+                .on_conflict(dsl::id)
```
davepacheco, line 136: This looks more like an "upsert" than "create".  I'm not sure if that's necessary here?

zephraph: The project images look like they're in the same boat

PR #2773, opened 2023-04-05
Add a layer of caching to omicron-package

steveklabnik:
This commit adds a very simple caching system to omicron-package. Upon running for the first time, omicron-package will write out a JSON file containing file paths and hashes of content. Upon subsequent builds, if the file exists with the same contents, it will skip the step of building the package.

On atrium, repeated builds take 3 seconds for me, compared to roughly ten seconds currently.

I am happy with any and all feedback here; I am trying to make basic progress rather than build a build system from scratch, ha! Also, I decided to go with a global here, because then I wouldn't need to pass configuration through a bunch of layers of functions. If you'd prefer a style where this is stored inside the configuration, or sitting on the stack of main, or something else, happy to make changes like that as well.


package/src/bin/omicron-package.rs
```diff
@@ -359,23 +382,58 @@ async fn get_package(
...
-                })?;
+
+            let output_file = match package.output {
+                PackageOutput::Zone { .. } => {
+                    output_directory.join(format!("{}.tar.gz", package_name))
+                }
+                PackageOutput::Tarball => {
+                    package.get_output_path(package_name, output_directory)
+                }
+            };
```
smklein, line 393: Because this is already implemented as:

https://github.com/oxidecomputer/omicron-package/blob/7caf0ba7c2b36e3308b0efe52df8d94c04bf38fc/src/package.rs#L328-L347

I think we could do the following:

```suggestion
            let output_file = package.get_output_path(package_name, output_directory);
```

package/src/bin/omicron-package.rs
```diff
@@ -359,23 +382,58 @@ async fn get_package(
...
+
+            let cache = BUILD_CACHE.lock().unwrap();
+            let hash = cache.get(&output_file).map(ToOwned::to_owned);
+            mem::drop(cache);
+
+            if hash.is_none()
+                || !does_hash_match(&output_file, &hash.unwrap()).await
```
smklein, line 400: So I'm theoretically on-board with an approach that uses hashes this way to avoid rebuilding - but don't we need to hash the *inputs* instead of the *outputs*?

For example:

- Suppose we build a **Nexus** package from some input files: (1) `db.rs`, (2) `http.rs`, and (3) `sagas.rs`
- Building this package produces `nexus.tar.gz`, which is then hashed and stored in `build-cache.json`
- If we rebuild without changing anything, the hash of `nexus.tar.gz` matches the hash stored in `build-cache.json`, which is great.
- However, if we update `http.rs` to, say, add a new endpoint...
  - `build-cache.json` will still store the old hash
  - `nexus.tar.gz` still exists, built using the **old** files
  - The hash will still appear to match, right?

Doesn't this basically mean that *all* re-builds where inputs change would be "skipped" incorrectly?

package/src/bin/omicron-package.rs
steveklabnik: Oh awesome! I got here in a ... roundabout way, but glad to see this is done in a better way.

package/src/bin/omicron-package.rs
steveklabnik: So, hashing *inputs* is 10000% the Right Way to do this; but it requires a lot more work for various reasons, and not that I am opposed to doing said work, I was just tired of not shipping *any* wins and wanted to get something going. Incremental progress is better than no progress.

My hope was to get something that would re-build spuriously more often than it should in a world where the inputs are hashed, but still usefully cache some of the time. That said, I think you're 100% right that this doesn't actually do this. Sigh. Time to go back to the drawing board.

smklein:
I've gone ahead and implemented a still-hopefully-relatively-simple cache in https://github.com/oxidecomputer/omicron-package/pull/60

PR #2776, opened 2023-04-06
[wicket/wicketd] Add ignition control

jgallagher:
This PR adds a few related-but-not-entirely-the-same things; if I need to break this up into separate bits (or only keep some of them), that's fine. In order from "most sure it's fine to merge" to "lease sure":

1. Adds an ignition command endpoint to `wicketd`; this just acts as a proxy for the same endpoint in MGS.
2. Adds a `force_refresh` query parameter to `wicketd`'s `get_inventory` endpoint. If the caller passes `true`, wicketd will immediately trigger a poll to MGS and only return a response to the caller after the next MGS poll comes back instead of returning the latest cached data.
3. Adds some mechanics in `wicket` to use `force_refresh` after a user interaction that we know should change the inventory state. Currently this is only used by 4.
4. Adds an ignition control popup to `wicket`. This is reachable by pressing `<i>` on either the inventory or update screens:

ignition-popup

I'm _very much_ not sold on the UI here, but wanted to put something in for now. One major flaw currently is that there is no place for feedback (either success or failure); pressing enter sends the command to `wicketd`, but results can only be observed indirectly (e.g., by seeing the ignition power state in the inventory screen change).

### DIFF
--- a/wicket/src/ui/panes/update.rs
+++ b/wicket/src/ui/panes/update.rs
@@ -7,7 +7,9 @@ use std::collections::BTreeMap;
 use super::{align_by, help_text, Control};
 use crate::state::{artifact_title, ComponentId, Inventory, ALL_COMPONENT_IDS};
 use crate::ui::defaults::style;
-use crate::ui::widgets::{BoxConnector, BoxConnectorKind, ButtonText, Popup, IgnitionPopup};
+use crate::ui::widgets::{
+    BoxConnector, BoxConnectorKind, ButtonText, IgnitionPopup, Popup,
+};
 use crate::{Action, Cmd, Frame, State};
 use omicron_common::api::internal::nexus::KnownArtifactKind;
 use slog::{info, o, Logger};
--- a/wicket/src/ui/panes/update.rs
+++ b/wicket/src/ui/panes/update.rs
@@ -7,7 +7,7 @@ use std::collections::BTreeMap;
 use super::{align_by, help_text, Control};
 use crate::state::{artifact_title, ComponentId, Inventory, ALL_COMPONENT_IDS};
 use crate::ui::defaults::style;
-use crate::ui::widgets::{BoxConnector, BoxConnectorKind, ButtonText, Popup};
+use crate::ui::widgets::{BoxConnector, BoxConnectorKind, ButtonText, Popup, IgnitionPopup};
 use crate::{Action, Cmd, Frame, State};
 use omicron_common::api::internal::nexus::KnownArtifactKind;
 use slog::{info, o, Logger};
@@ -23,6 +23,7 @@ enum PopupKind {
     MissingRepo,
     StartUpdate,
     Logs,
+    Ignition,
 }
 
 /// Overview of update status and ability to install updates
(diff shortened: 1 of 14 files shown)


wicket/src/ui/panes/update.rs
```diff
@@ -218,6 +240,24 @@ impl UpdatePane {
...
+                    self.ignition.key_down();
+                    Some(Action::Redraw)
+                }
+                Cmd::Enter => {
+                    let command = self.ignition.selected_command();
+                    let selected = state.rack_state.selected;
+                    info!(self.log, "Sending {command:?} to {selected}");
+                    self.popup = None;
+                    Some(Action::Ignition(selected, command))
+                }
```
sunshowers, line 258: Could you add a comment here saying to keep this in sync with the impl in `overview.rs`? (I guess there might be some differences over time which mean that the code can't directly be shared.)

wicketd/src/http_entrypoints.rs
```diff
@@ -48,6 +53,12 @@ pub fn api() -> WicketdApiDescription {
     api
 }
 
+#[derive(Clone, Copy, Debug, Serialize, Deserialize, JsonSchema, PartialEq)]
+pub struct ForceRefresh {
```
sunshowers, line 57: I think this should be called something like `GetInventoryParams`.

wicket/src/wicketd.rs
```diff
@@ -116,6 +135,35 @@ impl WicketdManager {
...
+                client.post_ignition_command(sp.type_, sp.slot, command).await;
+            // We don't return errors or success values, as there's nobody to
+            // return them to. How do we relay this result to the user?
+            slog::info!(
+                log,
+                "Ignition response for {} ({:?}): {:?}",
+                component_id,
+                command,
+                res
+            );
```
sunshowers, line 159: It seems like if we had the three-pane setup we talked about we could probably show the status in the third pane. Otherwise, the other option I can think of is us putting a spinner in the popup, keeping it open until the operation succeeds. If the operation fails, we make the popup bigger and show a message in it.

wicketd/src/http_entrypoints.rs
```diff
@@ -48,6 +53,12 @@ pub fn api() -> WicketdApiDescription {
...
+    /// If true, refresh inventory from MGS instead of returning cached data.
+    pub force_refresh: bool,
```
sunshowers, line 59: Realized that this should probably be a list of SPs that we want to force refresh.

wicket/src/ui/panes/update.rs
jgallagher: Certainly - added comments to both modules in 8610a7ea

wicketd/src/http_entrypoints.rs
jgallagher: Good call; renamed in b3976eac

wicketd/src/http_entrypoints.rs
jgallagher: Good idea; changed in b3976eac. Since `wicketd` is currently using the bulk "get all SPs" endpoint it doesn't actually use the list (other than "empty or not"), but we can start using it in a future PR.

PR #2783, opened 2023-04-07
[mgs] Add /local/switch-id endpoint

jgallagher:
This adds a `/local/switch-id` endpoint, which will return either

```json
{"type":"switch","slot":0}
```

or

```json
{"type":"switch","slot":1}
```

identifying the physical location of the switch this MGS instance is connected to. Additional details about the switch's SP (serial number, etc.) can then be fetched using the existing endpoints (e.g., `/sp/switch/0` or `/sp/switch/1`).

Note that like all other MGS endpoints, this requires MGS's location discovery process (RFD 250) to have completed; on "human scale" this is usually quite fast, but if dendrite is using this early in switch zone bringup it may be able to ask MGS before MGS knows an answer (in which case it will get back an error response and will need to retry).

### DIFF
--- a/gateway/src/http_entrypoints.rs
+++ b/gateway/src/http_entrypoints.rs
@@ -1382,12 +1382,14 @@ async fn recovery_host_phase2_upload(
 ///
 /// Note that most MGS endpoints behave identically regardless of which scrimlet
 /// the MGS instance is running on; this one, however, is intentionally
-/// different.
+/// different. This endpoint is _probably_ only useful for clients communicating
+/// with MGS over localhost (i.e., other services in the switch zone) who need
+/// to know which sidecar they are connected to.
 #[endpoint {
     method = GET,
     path = "/local/switch-id",
 }]
-async fn sp_local(
+async fn sp_local_switch_id(
     rqctx: RequestContext<Arc<ServerContext>>,
 ) -> Result<HttpResponseOk<SpIdentifier>, HttpError> {
     let apictx = rqctx.context();
@@ -1441,7 +1443,7 @@ pub fn api() -> GatewayApiDescription {
         api.register(ignition_get)?;
         api.register(ignition_command)?;
         api.register(recovery_host_phase2_upload)?;
-        api.register(sp_local)?;
+        api.register(sp_local_switch_id)?;
         Ok(())
     }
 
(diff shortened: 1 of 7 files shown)


gateway/src/http_entrypoints.rs
```diff
@@ -1378,6 +1378,25 @@ async fn recovery_host_phase2_upload(
     Ok(HttpResponseOk(HostPhase2RecoveryImageId { sha256_hash }))
 }
 
+/// Get the identifier for the switch this MGS instance is connected to.
+///
+/// Note that most MGS endpoints behave identically regardless of which scrimlet
+/// the MGS instance is running on; this one, however, is intentionally
+/// different.
```
bnaecker, line 1385: Is it worth it to be even more explicit in this docstring? I.e., that this should probably only be used when requesting to `localhost`, or maybe if one is _really sure_ one can map the IP to a physical location.

gateway/src/http_entrypoints.rs
```diff
@@ -1378,6 +1378,25 @@ async fn recovery_host_phase2_upload(
...
+#[endpoint {
+    method = GET,
+    path = "/local/switch-id",
+}]
+async fn sp_local(
```
bnaecker, line 1390: Maybe include `switch` in the method name? I don't know MGS too well, but it also reports data about the local Gimlet's SP right? If that's true, including `switch` would help disambiguate.

bnaecker (commented):
Couple of naming suggestions, but this looks great! I'm going to run MGS manually in `london`, where I've already got a Dendrite setup that brings up the management link for MGS to use. I'll report what I find!

gateway/src/http_entrypoints.rs
jgallagher: Yeah, that's a good idea. Expanded in d4267b38601022b8d0c2ae660c239e045dea7a47

gateway/src/http_entrypoints.rs
jgallagher: Yes, good call. Renamed in d4267b38601022b8d0c2ae660c239e045dea7a47

bnaecker:
Looks good to me. Feel free to merge whenever you want. Low priority but also low risk given that you've already tested the implementation. I'll add testing notes when I get a moment to try it out.

PR #2796, opened 2023-04-10
Nexus services ought to include ports

davepacheco:
After [discussing with @smklein])(https://github.com/oxidecomputer/omicron/pull/2753#issuecomment-1495142734), "services" in Omicron correspond with the traditional idea of network services -- things available over the network via TCP or UDP and typically discovered by DNS.  This change updates the Nexus internal API to expect that and updates Sled Agent to provide it.

This also means that _each_ of the current DNS services (InternalDNS and ExternalDNS) are really two different services: one DNS-protocol service on UDP port 53, and one HTTP-protocol service for configuring the server.  So I split up InternalDNS and ExternalDNS into two separate services (each).  This is still kind of an intermediate state because we're not reporting all of these to Nexus yet.  But I wanted to split this PR out from #2753 and I think this is pretty logically consistent as-is.

### DIFF
--- a/common/src/sql/dbinit.sql
+++ b/common/src/sql/dbinit.sql
@@ -108,13 +108,16 @@ CREATE INDEX ON omicron.public.sled (
  */
 
 CREATE TYPE omicron.public.service_kind AS ENUM (
+  'crucible_pantry',
+  'dendrite',
+  'external_dns_config',
+  'external_dns',
+  'internal_dns_config',
   'internal_dns',
   'nexus',
+  'ntp',
   'oximeter',
-  'dendrite',
-  'tfport',
-  'crucible_pantry',
-  'ntp'
+  'tfport'
 );
 
 CREATE TABLE omicron.public.service (
@@ -127,13 +130,21 @@ CREATE TABLE omicron.public.service (
     sled_id UUID NOT NULL,
     /* The IP address of the service. */
     ip INET NOT NULL,
+    /* The UDP or TCP port on which the service listens. */
+    port INT4 CHECK (port BETWEEN 0 AND 65535) NOT NULL,
     /* Indicates the type of service. */
     kind omicron.public.service_kind NOT NULL
 );
 
 /* Add an index which lets us look up the services on a sled */
 CREATE INDEX ON omicron.public.service (
-    sled_id
+    sled_id,
+    id
+);
+
+CREATE INDEX ON omicron.public.service (
+    kind,
+    id
 );
 
 -- Extended information for services where "service.kind = nexus"
--- a/sled-agent/src/rack_setup/service.rs
+++ b/sled-agent/src/rack_setup/service.rs
@@ -74,7 +74,10 @@ use internal_dns::ServiceName;
 use nexus_client::{
     types as NexusTypes, Client as NexusClient, Error as NexusError,
 };
-use omicron_common::address::{get_sled_address, NEXUS_INTERNAL_PORT};
+use omicron_common::address::{
+    get_sled_address, CRUCIBLE_PANTRY_PORT, NEXUS_INTERNAL_PORT, NTP_PORT,
+    OXIMETER_PORT,
+};
 use omicron_common::backoff::{
     retry_notify, retry_policy_internal_service_aggressive, BackoffError,
 };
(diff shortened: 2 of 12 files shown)


common/src/sql/dbinit.sql
```diff
@@ -108,13 +108,16 @@ CREATE INDEX ON omicron.public.sled (
  */
 
 CREATE TYPE omicron.public.service_kind AS ENUM (
+  'crucible_pantry',
+  'dendrite',
+  'external_dns_config',
+  'external_dns',
+  'internal_dns_config',
   'internal_dns',
```
smklein, line 116: Ah, so the expansion here is because these have the same IP but different ports?

common/src/sql/dbinit.sql
```diff
@@ -127,13 +130,21 @@ CREATE TABLE omicron.public.service (
     sled_id UUID NOT NULL,
     /* The IP address of the service. */
     ip INET NOT NULL,
+    /* The UDP or TCP port on which the service listens. */
+    port INT4 CHECK (port BETWEEN 0 AND 65535) NOT NULL,
```
smklein, line 134: This is *probably* low priority, but should we also store the protocol? We're now storing services that aren't just TCP.

(Totally okay to defer)

smklein (approved):

common/src/sql/dbinit.sql
davepacheco: Basically yes, though it's not just an artifact of only supporting one port in the database.  Logically they're two different things -- different consumers, different protocol, different APIs, etc.

common/src/sql/dbinit.sql
davepacheco: That's a fair question.  SRV records don't have that (I think).  I don't think it's necessary in the same way that the IP and port are.  I think it's assumed clients know what kind of thing they're going to talk to, just not how to reach it.

davepacheco:
Thanks for the review!

common/src/sql/dbinit.sql
luqmana: The protocol would be part of the record if I'm reading [this](https://en.wikipedia.org/wiki/SRV_record#Record_format) right:

>A SRV record has the form:
>`_service._proto.name. ttl IN SRV priority weight port target.`

(`_proto` here being the transport protocol, e.g. `_tcp`, `_udp`)

But looks like it's all just hard-coded to TCP right now:

https://github.com/oxidecomputer/omicron/blob/5d39727039cca86a7fef48d35f3675187386c333/internal-dns/src/names.rs#L53-L76

sled-agent/src/rack_setup/service.rs
```diff
@@ -641,35 +660,88 @@ impl ServiceInner {
...
+                                address: server_address.to_string(),
+                                kind:
+                                    NexusTypes::ServiceKind::InternalDNSConfig,
+                            });
+                            services.push(NexusTypes::ServicePutRequest {
+                                service_id: zone.id,
+                                sled_id,
+                                address: dns_address.to_string(),
+                                kind: NexusTypes::ServiceKind::InternalDNS,
+                            });
```
luqmana, line 694: This was in the back of my mind and I had forgotten but I don't think we should be using `zone.id` as the `service_id` in `ServicePutRequest`. Any subsequent entries with the same id will just overwrite the previous ones:

https://github.com/oxidecomputer/omicron/blob/5d39727039cca86a7fef48d35f3675187386c333/nexus/db-queries/src/db/datastore/rack.rs#L181-L193

This wasn't previously an issue since we only ever had 1 service per zone up til this point AFAIK.

And sorry for the post-merge comments here, I had meant to look but earlier :/

common/src/sql/dbinit.sql
davepacheco: Thank you -- I'd forgotten about that!

sled-agent/src/rack_setup/service.rs
davepacheco: Thank you for catching this!

I wonder if we should keep calling this `service_id` (or `zone_id`) but make the primary key (`service_id`, `kind`).  My thinking here is that It'd be annoying to have some extra uuid that's not traceable to anything in particular, and also to have no way to get to the corresponding zone id.

luqmana: A tuple primary key w/ `kind` would fix the current setup but not if we want to ever run two instances of the same service in the same zone. (Not saying we'd ever need/want to but there's no reason we couldn't). Maybe we just add a ` zone_id` column? A little verbose but makes things super clear

davepacheco: Filed as #2813.

PR #2799, opened 2023-04-10
[wicketd] switch to StreamingBody

sunshowers:
Dropshot now has support for streaming request bodies, which we can use
here to support large request body uploads.

Also switch some code to using a BufList.


andrewjstone (approved):
Ship it!

Issue #1530, opened 2022-08-01
RSS should consume "pool of addresses for Nexus"


smklein:
In a handful of spots, this is being hardcoded as a single address. We will want to support multiple external IPs for Nexus, and should instead consume a pool of IPs.

rmustacc:
In addition to allowing us to specify a specific IP pool that isn't being used for customers, we should assume that operators will want to explicitly set and control these external IPs and select addresses from the pool.

smklein:
> In addition to allowing us to specify a specific IP pool that isn't being used for customers, we should assume that operators will want to explicitly set and control these external IPs and select addresses from the pool.

This tracks - I figured there would need to be:

- A setup / transfer mechanism to get a minimal set of this info from RSS -> Nexus, for bootstrapping
- A general purpose API within Nexus for operators to use - ideally, [similar to the customer-visible IP pool APIs](https://github.com/oxidecomputer/omicron/blob/b270033cd460f4ae429d0f9643b0a92548b39719/nexus/src/external_api/http_entrypoints.rs#L104-L112) - for managing these "used-for-Oxide-service" IPs.

bnaecker:
Just clarifying another point from the networking sync today. We will need to add a way to refer to one or more IP Pools when configuring each individual service. As an example, customers may create more than one IP Pool during RSS, on totally separate networks, and want to have Nexus listen on an address from each pool. So the service configuration data needs to include a list of IP Pools.

bnaecker:
I'm closing this one. We've had a "pool of addresses" here for a while, and the work needed to specify actual IP pool details is better tracked by https://github.com/oxidecomputer/omicron/issues/8946.

PR #2801, opened 2023-04-11
Explicitly configure services IP pool as part of RSS

luqmana:
We've been populating the services IP pool with a single address (the external IP configured for Nexus) up til now. This flips that around by instead requiring a list of IP ranges be given as part of rack initialization. We can then pull addresses from there for Nexus (and any other services as needed).

Further work is still needed to support multiple IP pools and allowing services to utilize addresses across all (or some) of them.

Part of https://github.com/oxidecomputer/omicron/issues/1530.

### DIFF
--- a/common/src/address.rs
+++ b/common/src/address.rs
@@ -7,11 +7,11 @@
 //! This addressing functionality is shared by both initialization services
 //! and Nexus, who need to agree upon addressing schemes.
 
-use crate::api::external::Ipv6Net;
+use crate::api::external::{self, Error, Ipv6Net};
 use ipnetwork::Ipv6Network;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
-use std::net::{Ipv6Addr, SocketAddrV6};
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddrV6};
 
 pub const AZ_PREFIX: u8 = 48;
 pub const RACK_PREFIX: u8 = 56;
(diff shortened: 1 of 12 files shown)


common/src/address.rs
```diff
@@ -177,6 +177,256 @@ pub fn get_64_subnet(
...
+    fn next(&mut self) -> Option<Self::Item> {
+        let next = self.next?;
+        if next < self.last {
+            self.next = Some(next + 1);
+        } else {
+            self.next = None;
+        }
+        Some(next.into())
+    }
+}
```
smklein, line 393: This could probably be a generic over u32 vs u128 but also frankly if you wanna keep 'em separate I get it

smf/sled-agent/gimlet-standalone/config-rss.toml
```diff
@@ -11,12 +11,17 @@ rack_subnet = "fd00:1122:3344:0100::"
...
-# NOTE: In the lab, use "172.20.15.226"
-nexus_external_address = "192.168.1.20"
-
 ntp_servers = [ "ntp.eng.oxide.computer" ]
 dns_servers = [ "1.1.1.1", "9.9.9.9" ]
 
+# The IP ranges configured as part of the services IP Pool.
+# e.g., Nexus will be configured to use an address from this
+# pool as its external IP.
+# NOTE: In the lab, use "172.20.15.226"
```
smklein, line 20: TBH, I think we should delete this comment - there are multiple lab configs, and I've found this to be a little misleading

smf/sled-agent/non-gimlet/config-rss.toml
```diff
@@ -11,12 +11,17 @@ rack_subnet = "fd00:1122:3344:0100::"
...
-# NOTE: In the lab, use "172.20.15.226"
-nexus_external_address = "192.168.1.20"
-
 ntp_servers = [ "ntp.eng.oxide.computer" ]
 dns_servers = [ "1.1.1.1", "9.9.9.9" ]
 
+# The IP ranges configured as part of the services IP Pool.
+# e.g., Nexus will be configured to use an address from this
+# pool as its external IP.
+# NOTE: In the lab, use "172.20.15.226"
```
smklein, line 20: Same as above

sled-agent/src/rack_setup/service.rs
```diff
@@ -641,25 +639,6 @@ impl ServiceInner {
                     // (which is useful in dev/test situations).
                     match svc {
                         ServiceType::Nexus { external_ip, internal_ip: _ } => {
-                            // NOTE: Eventually, this IP pool will be entirely
```
smklein, line 644: Thank you for this cleanup; I'm happy to see this hack removed!

smklein (approved):
Looks great, thanks for patching this up!

Issue #2805, opened 2023-04-11
Track provisioned CPU and RAM usage for instances


smklein:
Part of https://github.com/oxidecomputer/omicron/issues/2804

PR #2807, opened 2023-04-11
Add per-sled CPU and RAM capacities for instance usage

smklein:
Fixes https://github.com/oxidecomputer/omicron/issues/2805


nexus/tests/integration_tests/instances.rs
```diff
@@ -1927,77 +1927,6 @@ async fn test_attach_one_disk_to_instance(cptestctx: &ControlPlaneTestContext) {
     assert_eq!(disks[0].state, DiskState::Attached(instance.identity.id));
 }
 
-#[nexus_test]
-async fn test_instance_fails_to_boot_with_disk(
```
smklein, line 1931: FYI to @jmpesp , this was removed, but I believe we have coverage with the other idempotency tests.

nexus/src/app/sagas/instance_create.rs
```diff
@@ -609,17 +607,40 @@ async fn sic_alloc_server(
     //   schedule instances that belong to a cluster on different failure
     //   domains. See https://github.com/oxidecomputer/omicron/issues/1705.
 
-    osagactx
+    // TODO: Fix these values. They're wrong now, but they let us move
+    // forward with plumbing.
+    let params = sagactx.saga_params::<Params>()?;
+    let hardware_threads = params.create_params.ncpus.0;
+    let rss_ram = params.create_params.memory;
+    let reservoir_ram = omicron_common::api::external::ByteCount::from(0);
```
smklein, line 615: *right now*, our policy for "how many resources are consumed by instances" is an exact 1:1 mapping of:

- Guest CPUS -> Hardware threads
- Guest RAM -> RSS RAM

Arguably, propolis will consume some overhead, and we will need to make considerations for the reservoir cost, in addition to the "rss ram" cost.

However, I wanted to put this PR up as-is, since it'll mechanically be very similar - just a difference of accounting.

nexus/src/app/sagas/instance_create.rs
```diff
@@ -609,17 +607,40 @@ async fn sic_alloc_server(
...
+    Ok(resource.sled_id)
+}
+
+async fn sic_alloc_server_undo(
+    sagactx: NexusActionContext,
+) -> Result<(), anyhow::Error> {
+    let osagactx = sagactx.user_data();
+    let instance_id = sagactx.lookup::<Uuid>("instance_id")?;
+
+    osagactx.nexus().delete_sled_reservation(instance_id).await?;
```
smklein, line 642: FYI to @gjcolombo - "picking a sled" for an instance is now selected via `reserve_on_random_sled`, and removed via `delete_sled_reservation`.

It seems highly likely this will impact how instance migration will work -- we need to reserve on the destination, and eventually delete the reservation on the source sled.

gjcolombo: Thanks, this makes sense.

> we need to reserve on the destination

Is it possible to rearrange the parts of `sled_reservation_create` so that it's straightforward to add constraints to the search for sleds with available resources? In the migration case, where the external API currently allows a caller to select a destination sled explicitly, this would be "the sled ID is exactly X." More generally, I think we'll eventually want to be able to say, "this VM requires a sled with at least this software version" or "this instance must be in this particular rack" or even just "pick any compatible sled *except* this one." In a perfect world these would be easy to compose ("add these WHERE clauses to this query") but I'm not good enough with Diesel to assess how challenging that is.

> and eventually delete the reservation on the source sled

ACK. I already need to add a commit to the LM series that will deal with this sort of post-migration task, so this can be another piece of work that task has to do.

gjcolombo: @smklein and I talked this over offline and concluded that the easiest way to make progress for now is to do the following:

- Add a parameter to `sled_reservation_create` (either in this PR or in the live migration series, I'm not sure it matters which) that allows the caller to specify a sled, or set of sleds, to exclude from consideration. In the migration case, this is the instance's current sled.
- In my pending migration changes, change the migration saga to use `sled_reservation_create` instead of an explicitly-selected sled. (This would change the external API from "migrate this instance to sled X" to "migrate this instance to any eligible sled. cc @askfongjojo about this part--is this going to leave us with an unworkable migration API? Do we need to think about this more?)

If/when we need more constraints in the future, we can come up with a suitably elegant way to add them (e.g. a `SledSelectionConstraints` type with a fancy builder interface).

gjcolombo: I can add it to the migration series.

sled-agent/src/bin/sled-agent-sim.rs
```diff
@@ -76,6 +77,7 @@ async fn do_run() -> Result<(), CmdError> {
             ip: (*args.sled_agent_addr.ip()).into(),
         },
         updates: ConfigUpdates { zone_artifact_path: tmp.path().to_path_buf() },
+        hardware: ConfigHardware { hardware_threads: 4, physical_ram: 1 << 30 },
```
gjcolombo, line 80: I think these should be more generous (e.g. matching what's in test-utils). With this config, a sled-agent-sim sled can have exactly one instance with 1 GiB of RAM, since we require guest memory to be specified in 1 GiB increments.

nexus/src/app/sagas/instance_delete.rs
```diff
@@ -289,6 +293,24 @@ async fn sid_account_resources(
...
+    let osagactx = sagactx.user_data();
+    let params = sagactx.saga_params::<Params>()?;
+    let opctx = crate::context::op_context_for_saga_action(
+        &sagactx,
+        &params.serialized_authn,
+    );
+
+    osagactx
+        .datastore()
+        .sled_reservation_delete(&opctx, params.instance.id())
```
gjcolombo, line 308: Is this the right ID? `sled_reservation_delete` seems to want the resource ID, unless I'm missing something about how the tables are arranged.

gjcolombo (commented):
Thanks for putting this together! LGTM except for one small reservation (aha) about the change to the delete saga.

sled-agent/src/bin/sled-agent-sim.rs
smklein: Done in ec2d3f82bd1bbd9999e2512140d68e9036ba775f

nexus/src/app/sagas/instance_delete.rs
smklein: Ah, I'm exploiting the fact that the instance ID and resource ID are tightly coupled! The instance *is* the resource being provisioned, so it's using the same UUID.

I tried to document this in the schema:

```sql
-- Accounting for programs using resources on a sled
CREATE TABLE omicron.public.sled_resource (
    -- Should match the UUID of the corresponding service
    id UUID PRIMARY KEY,
```

But I'm happy to make this more clear, where possible?

gjcolombo: Ah, that makes sense. I just whistled past that part of the schema definition.

I might add a short comment here for the benefit of future readers who might not have read the schema first (since reading `reservation_delete .. instance.id` might seem as odd to them as it did to me), but I don't think we need to massively rework or reorganize the rest of the PR.

gjcolombo (approved):

gjcolombo:
:rocket:

PR #2809, opened 2023-04-11
[mgs] Clean up /sp and /sp/{type}/{slot} based on wicketd usage

jgallagher:
This makes a pass over a couple problematic MGS endpoints based on experience using it from `wicketd`. The two major changes are:

1. We remove the `/sp` endpoint entirely. This endpoint fetched the ignition state and the current SP state from all SPs in the rack, but was problematic for (at least) two reasons: (a) deciding when to time out SPs (particularly on an underpopulated rack) is hard; the current implementation performed quite poorly if any SPs were missing or unresponsive, and (b) it had to return a mix of "success" and "failure" within one HTTP response. It was also the most complex endpoint (by far), and we'll take any opportunity to keep MGS as simple as possible.
2. We change the `/sp/{type}/{slot}` endpoint to only return `SpState`, not a combination of `SpState` and ignition state. This means MGS only needs to send one management network request (to the target SP) instead of two (to the target SP and the sidecar SP). The response is also much cleaner: we previously had to return any combination of "ignition success/fail + state success/fail", and now we either return an HTTP OK with the state or an HTTP failure if talking to the target SP fails. We therefore get to remove the error variants of both `SpState` (which is now a struct instead of an enum!) and `SpIgnition`.

Builds on #2808, which changes `wicketd` to not use `/sp`.


andrewjstone (approved):
Hell yeah!

sunshowers (approved):
This is awesome!

PR #2815, opened 2023-04-11
Update user accounting database once time is synced

citrus-it:
(PST within the zones notwithstanding, see
https://github.com/oxidecomputer/helios-omicron-brand/issues/4)

This change rewrites the user accounting databases for all running zones
(including the GZ) once time is synchronised. It uses the
`/usr/platform/oxide/bin/tmpx` command which is available in very recent
stlouis. If that's missing, it will just carry on regardless, leaving the
system uptime showing 13k days as before.

```
{"msg":"Timesync for [fd00:1122:3344:101::1]:12345 TimeSync { sync: false, skew: 0.0, correction: 0.0 }","v":0,"name":"SledAgent","level":30,"time":"1986-12-28T00:15:45.669141687Z","hostname":"gimlet-sn06","pid":101256,"component":"RSS"}
{"msg":"Time is not yet synchronized","v":0,"name":"SledAgent","level":40,"time":"1986-12-28T00:15:45.669154267Z","hostname":"gimlet-sn06","pid":101256,"component":"RSS","error":"\"Time is synchronized on 0/1 sleds\""}
...
gimlet-sn06 # date; uptime
Sun Dec 28 00:26:49 UTC 1986
00:26:49    up 1 min(s),  2 users,  load average: 1.31, 1.54, 0.77
gimlet-sn06 # zlogin oxz_internal_dns 'date; uptime'
Sat Dec 27 16:27:07 PST 1986
16:27:07    up 1 min(s),  0 users,  load average: 0.00, 0.00, 0.00
...
{"msg":"Timesync for [fd00:1122:3344:101::1]:12345 TimeSync { sync: true, skew: 117.186, correction: 2.207e-5 }","v":0,"name":"SledAgent","level":30,"time":"2023-04-11T22:34:20.543455275Z","hostname":"gimlet-sn06","pid":101256,"component":"RSS"}
...
gimlet-sn06 # date; uptime
Tue Apr 11 23:28:19 UTC 2023
23:28:19    up 1 min(s),  2 users,  load average: 4.75, 2.14, 1.10
gimlet-sn06 # zlogin oxz_internal_dns 'date; uptime'
Tue Apr 11 16:28:23 PDT 2023
16:28:23    up 1 min(s),  0 users,  load average: 0.00, 0.00, 0.00
```

### DIFF
--- a/sled-agent/src/services.rs
+++ b/sled-agent/src/services.rs
@@ -37,6 +37,7 @@ use illumos_utils::running_zone::{InstalledZone, RunningZone};
 use illumos_utils::zfs::ZONE_ZFS_DATASET_MOUNTPOINT;
 use illumos_utils::zone::AddressRequest;
 use illumos_utils::zone::Zones;
+use illumos_utils::{execute, PFEXEC};
 use omicron_common::address::Ipv6Subnet;
 use omicron_common::address::BOOTSTRAP_ARTIFACT_PORT;
 use omicron_common::address::CRUCIBLE_PANTRY_PORT;
@@ -56,11 +57,14 @@ use sled_hardware::underlay;
 use sled_hardware::SledMode;
 use slog::Logger;
 use std::collections::HashSet;
+use std::iter;
 use std::iter::FromIterator;
 use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr};
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
+use std::time::{SystemTime, UNIX_EPOCH};
 use tokio::io::AsyncWriteExt;
 use tokio::sync::oneshot;
 use tokio::sync::Mutex;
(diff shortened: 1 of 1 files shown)


sled-agent/src/services.rs
```diff
@@ -1204,14 +1211,59 @@ impl ServiceManager {
...
+        let now = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .expect("SystemTime before UNIX EPOCH");
+
+        info!(self.inner.log, "Setting boot time to {:?}", now);
+
+        let files: Vec<String> = zones
+            .iter()
+            .map(|z| z.root())
+            .chain(iter::once("".to_string()))
```
smklein, line 1234: This is for the database of the global zone, right?

When we create zones in the future, zones will derive values from the GZ's notion of time, right?

smklein (approved):

sled-agent/src/services.rs
citrus-it: Yes, and yes. There's only one system clock and once it's right then new zones will be fine, and will create their own `wtmpx` and `utmpx` files with correct timestamps. We just have to take a pass through the databases in the GZ and in any running zones, once time is synchronised for the first time.

smklein: Makes sense, good catch to fix-up the old zones in place.

PR #2826, opened 2023-04-12
(Fix build breakage) Dendrite address in configuration is now optional

smklein:
I believe this happened as a racy merge of #2579 and #2694 .

At the end of the day, this configuration value is now optional, so treat it as such, but require it in the tests.


jmpesp (approved):
Just hit this, was going to open a PR but saw that I was too slow!

PR #2827, opened 2023-04-12
Add retries to downloading scripts

smklein:
This seems to be a common source of CI failure -- some request to an external server fails when installing prerequisites, and that whole job needs to be restarted.

To mitigate, build a concept of "automatic retries" into the downloader scripts.

### DIFF
--- a/tools/install_builder_prerequisites.sh
+++ b/tools/install_builder_prerequisites.sh
@@ -19,17 +19,26 @@ function on_exit
 
 trap on_exit ERR
 
-# Parse command line options:
-#
-# -y  Assume "yes" intead of showing confirmation prompts.
-# -p  Skip checking paths
+function usage
+{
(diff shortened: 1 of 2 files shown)


tools/install_builder_prerequisites.sh
```diff
@@ -71,57 +102,62 @@ function confirm
...
-    'postgresql'
-    'pkg-config'
-    'libxmlsec1'
-  )
-  confirm "Install (or update) [${packages[*]}]?" && brew install ${packages[@]}
-else
-  echo "Unsupported OS: ${HOST_OS}"
-  exit -1
-fi
+install_packages
```
jgallagher, line 160: Do we want to `retry` this too? I know we've seen failures in CI when `apt`-installing things (although maybe that's fixed now with the change of apt source?).

smklein: Sure, I'll do the same for the `install_runner_prerequisites.sh` script too.

PR #2828, opened 2023-04-13
Flesh out some authz infra around customer network integration 

zephraph:
This is in no way complete, but a start at add in some of the infrastructure required by #2645.

### DIFF
--- a/common/src/api/external/mod.rs
+++ b/common/src/api/external/mod.rs
@@ -681,6 +681,7 @@ impl TryFrom<i64> for Generation {
 #[display(style = "kebab-case")]
 pub enum ResourceType {
     AddressLot,
+    AddressLotBlock,
     Fleet,
     Silo,
     SiloUser,
--- a/nexus/db-queries/src/db/datastore/address_lot.rs
+++ b/nexus/db-queries/src/db/datastore/address_lot.rs
@@ -3,6 +3,7 @@
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 use super::DataStore;
+use crate::authz;
 use crate::context::OpContext;
 use crate::db;
 use crate::db::datastore::PgConnection;
@@ -25,7 +26,7 @@ use nexus_types::identity::Resource;
 use omicron_common::api::external::http_pagination::PaginatedBy;
 use omicron_common::api::external::{
     CreateResult, DataPageParams, DeleteResult, Error, ListResultVec,
-    LookupResult, LookupType, NameOrId, ResourceType,
+    LookupResult, ResourceType,
 };
 use ref_cast::RefCast;
 use serde::{Deserialize, Serialize};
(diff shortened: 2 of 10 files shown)


nexus/db-queries/src/db/datastore/address_lot.rs
```diff
@@ -210,50 +183,15 @@ impl DataStore {
...
-        use db::schema::address_lot::dsl as lot_dsl;
-        use db::schema::address_lot_block::dsl as block_dsl;
+        use db::schema::address_lot_block::dsl;
 
         let pool = self.pool_authorized(opctx).await?;
 
-        let id = match &address_block_id {
-            NameOrId::Id(id) => *id,
-            NameOrId::Name(name) => {
-                let name = name.to_string();
```
zephraph, line 225: We don't want this decision happening at the datastore layer. In the http entrypoint we use the query and path parameters to retrieve a lookup object. This lookup is usually an authz resource that has a reference to the ID. Once we get to this layer we should already have the ID to work with. 

The caveat to the above is for list endpoints that do name or ID pagination.

nexus/db-queries/src/db/lookup.rs
```diff
@@ -742,6 +750,15 @@ lookup_resource! {
...
 
+lookup_resource! {
+    name = "AddressLot",
+    ancestors = [],
+    children = [], // TODO: Should this include AddressLotBlock?
+    lookup_by_name = true,
+    soft_deletes = true,
+    primary_key_columns = [ { column_name = "id", rust_type = Uuid } ]
+}
+
```
zephraph, line 761: We're 100% going to need more lookups than this. I was just starting it off here.

nexus/src/app/address_lot.rs
```diff
@@ -19,23 +22,43 @@ use std::sync::Arc;
...
+                    .address_lot_id(id);
+                Ok(lot)
+            }
+            NameOrId::Name(name) => {
+                let lot = LookupPath::new(opctx, &self.db_datastore)
+                    .address_lot_name_owned(name.into());
+                Ok(lot)
+            }
+        }
+    }
```
zephraph, line 42: I need to document this, but a typical pattern you'll see for every resource in the app layer is a lookup function that looks similar to this. It takes either the name or id (for top level resources) or a selector (for resources nested deeper in the hierarchy) and returns a lookup. The lookup that's returned from here is all that's intended to be passed to other layers of the API.

nexus/src/app/address_lot.rs
```diff
@@ -19,23 +22,43 @@ use std::sync::Arc;
...
         validate_blocks(&params)?;
-        opctx.authorize(authz::Action::Modify, &authz::FLEET).await?;
         self.db_datastore.address_lot_create(opctx, &params).await
     }
 
     pub async fn address_lot_delete(
         self: &Arc<Self>,
         opctx: &OpContext,
-        name_or_id: &Option<NameOrId>,
+        address_lot_lookup: &lookup::AddressLot<'_>,
```
zephraph, line 57: As mentioned above, I've swapped out the `name_or_id` reference for a lookup. That lookup makes it easy for us to do authz checks on it like you'll see below.

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -2286,19 +2286,21 @@ async fn networking_address_lot_create(
...
     method = DELETE,
-    path = "/v1/system/networking/address-lot",
+    path = "/v1/system/networking/address-lot/{address_lot}",
     tags = ["external-networking"],
 }]
 async fn networking_address_lot_delete(
     rqctx: RequestContext<Arc<ServerContext>>,
-    query_params: Query<params::AddressLotSelector>,
+    path_params: Path<params::AddressLotPath>,
 ) -> Result<HttpResponseDeleted, HttpError> {
```
zephraph, line 2295: The API only uses selectors to help narrow down resources that may have a parent container. For example, if you're looking up an instance you specify the instance in the path (`/v1/instances/{instance}`). If you pass a `Name` for `instance`, we don't have enough information to do a lookup (because instance names are scoped to projects) so you need to narrow it with a selector. Like `/v1/instances/{instance_name}?project={project}`. If `instance` is provided by an ID no selector is needed so the call is just `/v1/instances/{instance_id}`.

nexus/tests/output/nexus_tags.txt
```diff
@@ -13,9 +13,9 @@ disk_view                                GET      /v1/disks/{disk}
...
 API operations found with tag "external-networking"
 OPERATION ID                             METHOD   URL PATH
-networking_address_lot_block_list        GET      /v1/system/networking/address-lot/blocks
+networking_address_lot_block_list        GET      /v1/system/networking/address-lot/{address_lot}/blocks
 networking_address_lot_create            POST     /v1/system/networking/address-lot
-networking_address_lot_delete            DELETE   /v1/system/networking/address-lot
+networking_address_lot_delete            DELETE   /v1/system/networking/address-lot/{address_lot}
 networking_address_lot_list              GET      /v1/system/networking/address-lot
 networking_loopback_address_create       POST     /v1/system/networking/loopback-address
 networking_loopback_address_delete       DELETE   /v1/system/networking/loopback-address
```
zephraph, line 21: There are definitely some more API updates that need to be made. Every delete should have a path param for instance.

nexus/types/src/external_api/params.rs
```diff
@@ -1143,14 +1144,7 @@ pub struct FinalizeDisk {
...
-    pub address_lot: Option<NameOrId>,
-}
-
-/// Select an address lot block by name or id.
-#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema, PartialEq)]
-pub struct AddressLotBlockSelector {
-    /// The address lot name or id to select.
+    /// Name or id of the address lot to select
     pub address_lot: NameOrId,
 }
```
zephraph, line 1149: I actually don't know that this selector is needed at all.

common/src/api/external/mod.rs
```diff
@@ -681,6 +681,7 @@ impl TryFrom<i64> for Generation {
 #[display(style = "kebab-case")]
 pub enum ResourceType {
     AddressLot,
+    AddressLotBlock,
```
zephraph, line 684: I added `AddressLotBlock` as a dedicated resource because it's represented distinctly in the API but I do wonder if it really should be a resource. `AddressLot` by itself (as I [mentioned in the parent PR](https://github.com/oxidecomputer/omicron/pull/2645/files#r1164898672)) feels incomplete without a `blocks` property. I get that adding that wouldn't work if we wanted it to be a paginated resource (which I assume is the case).

rcgoodfellow: Logically I agree that `AddressLot` and `AddressLotBlock` are incomplete without each other. The blocks are broken out because there is the potential for the number to be quite large.

rcgoodfellow:
Sorry, I just rebased the `external-networking` branch without considering this PR. I went ahead and rebased this branch on the rebase to line things up again.

PR #2829, opened 2023-04-13
[update-engine] add the ability to create nested engines

sunshowers:
For the installinator, we need to be able to break a step into a series
of sub-steps. That is because while writing to disk, we need to handle
multiple different kinds of progress events:

* once for each M2 slot
* once for each artifact
* with retries

To handle these cases, allow for an individual step to create update
engines of its own. Parent steps can create as many child engines as
necessary, with as many steps as necessary.

This is not quite fully dynamic/monadic -- it would be so if we
flattened all events at the top level, but we maintain a hierarchy which
is really nice in practice, since it allows recording elapsed times in a
hierarchical fashion.

To handle nested engines, we also need to introduce the ability to
represent a generic spec that can represent any arbitrary value. I've
added that via a new GenericSpec spec and a NestedSpec type alias.
NestedError specifically represents a set of serialized errors sent over
the wire -- I haven't done the work to plug this in so the exact details
might change. (Most of this PR is boilerplate to support generic specs.)

I've managed to use nested engines in the installinator (PR #2830),
and they work super well! I've also updated the example to
use nested engines, though I didn't have the time to update the UI to
display nested events. It is totally possible to do this, just a bit of
additional work that I want to come back to in a future PR.


update-engine/src/spec.rs
```diff
@@ -85,10 +85,93 @@ pub trait StepSpec: JsonSchema {
...
 
+/// Represents a fully generic step specification, as can be serialized over
+/// JSON.
+///
+/// Since errors aren't directly serialized, they can be any type that
+/// implements [`AsError`].
+///
+/// Use this if you don't care about assigning types to any of the metadata
+/// components.
+pub struct GenericSpec<E> {
```
sunshowers, line 96: This is a bit warty! I'm not happy about the fact that the E parameter exists here. The error absolutely does not matter for serialized data, just for running the engine itself.

I think I'm going to revisit this later. I think we may have to break this up into two separate types, one for serialized data and one for the error.

update-engine/src/engine.rs
```diff
@@ -782,7 +820,7 @@ mod tests {
                         if failed_step.info.component == "bar"
                         && message == "example failed"
                     ),
-                    "event didn't match: {last_event:?}"
+                    "event di't match: {last_event:?}"
```
andrewjstone, line 823: Looks like a finger slip here

update-engine/src/context.rs
```diff
@@ -24,30 +28,112 @@ use crate::{events::StepProgress, StepSpec};
...
+                            result = Some(Err(error));
+                        }
+                    }
+                }
+                event = receiver.recv(), if !events_done => {
+                    match event {
+                        Some(event) => {
+                            match event.into_generic() {
+                                Ok(event) => {
+                                    self.payload_sender.send(
```
andrewjstone, line 98: Is the `payload_sender` a channel shared among all nested engines, such that events come out in one receiver?

andrewjstone (approved):
Looks good to me. There's a lot here, but it looks straightforward. Nice job @sunshowers!

update-engine/src/errors.rs
```diff
@@ -51,3 +51,68 @@ impl<S: StepSpec> From<mpsc::error::SendError<Event<S>>> for ExecutionError<S> {
...
+        self.path
+            .push_front(ConvertGenericPathElement::ArrayIndex(elem, index));
+        self
+    }
+}
+
+impl fmt::Display for ConvertGenericError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str("error converting path: ")?;
+        // VecDeque doesn't have join()...
```
jgallagher, line 90: I don't think this comment is necessary - `join()` would create a temp string that this loop doesn't, right? Either way, the loop seems perfectly reasonable to me.

update-engine/src/context.rs
jgallagher: I think "yes and no" - the nested engine created in this function has its own payload sender (our local `sender` variable) that is distinct from `self.payload_sender`; however, any events sent to the nested `sender` are here forwarded out through `self.payload_sender` as `StepContextPayload::Nested(_)`. That would apply recursively if this nested engine itself had steps that created a nested engine: that innermost engine would have its own `sender`, which would get forwarded out via the middle `sender`, and then forwarded here through `self.payload_sender` (at this point as the moral equivalent of a `StepContextPayload::Nested(StepContextPayload::Nested(event))`, although I think it's not actually that exact type.

update-engine/src/context.rs
```diff
@@ -24,30 +28,112 @@ use crate::{events::StepProgress, StepSpec};
...
+        // Now run the engine.
+        let engine = engine.execute();
+        tokio::pin!(engine);
+
+        let mut result = None;
+        let mut events_done = false;
+
+        loop {
+            tokio::select! {
+                ret = &mut engine, if result.is_none() => {
```
jgallagher, line 80: Just double-checking my understanding here: when we enter this `select!` we start polling both `engine` and `receiver.recv()`. If `engine` becomes ready, the future returned by `receiver.recv()` is cancelled (which is fine, `recv()` returns a cancel-safe future) and we handle the result here. If the future returned by `receiver.recv()` becomes ready, `engine` is _not_ cancelled, because we're polling `&mut engine`, right?

jgallagher (approved):

update-engine/examples/update-engine-basic/main.rs
```diff
@@ -239,3 +283,67 @@ impl ExampleContext {
...
+
+                    while buf_list.has_remaining() {
+                        tokio::time::sleep(Duration::from_millis(200)).await;
+                        let written_bytes = file
+                            .write_buf(&mut buf_list)
+                            .await
+                            .context("error writing data")?;
+
+                        total_written += written_bytes;
+                        cx2.send_progress(
```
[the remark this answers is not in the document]
sunshowers: Yes, exactly.

update-engine/src/context.rs
sunshowers: > If engine becomes ready, the future returned by receiver.recv() is cancelled (which is fine, recv() returns a cancel-safe future) and we handle the result here.

Right, the other part of this is that polling the engine, even if it doesn't complete, drives it forward.

> engine is not cancelled, because we're polling &mut engine, right?

Correct. In general, we can always use the `&mut` pattern to turn any non-cancel-safe future into a cancel-safe one.

update-engine/src/engine.rs
sunshowers: Thanks, fixed.

update-engine/src/events.rs
```diff
@@ -36,6 +69,37 @@ pub struct StepEvent<S: StepSpec> {
     pub kind: StepEventKind<S>,
 }
 
+impl<S: StepSpec> StepEvent<S> {
+    /// Converts a generic version into self.
+    ///
+    /// This version can be used to convert a generic type into a more concrete
+    /// form.
+    pub fn from_generic<E: AsError>(
```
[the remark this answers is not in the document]
sunshowers: Additionally, event receivers don't necessarily have to use this -- depends on the specifics of what they want to do.

PR #2830, opened 2023-04-13
[installinator] switch to using update engine

sunshowers:
This switches the installinator over to using the new update engine. The installinator uses several steps to:

1. Download the host artifact.
2. Download the control plane artifact.
3. Format the host and control plane.
4. Write the host and control plane artifacts.
    * This breaks down into a loop with several steps, done by using the nested engine.

Also integrate the installinator with wicketd.

Depends on #2829 and #2890.

Closes https://github.com/oxidecomputer/minimum-upgradable-product/issues/76.


installinator-common/src/progress.rs
```diff
@@ -2,329 +2,261 @@
...
-        kind: ArtifactKind,
-
-        /// The peer the artifact was being downloaded from, if any.
+pub enum InstallinatorProgressMetadata {
+    Download {
+        /// The peer being downloaded from.
+        ///
+        /// Available with downlad events.
         peer: SocketAddrV6,
-
```
andrewjstone, line 69: 👏 So awesome to collapse these all into `update_engine::ProgressEventKind`

installinator/src/dispatch.rs
```diff
@@ -184,27 +179,62 @@ impl InstallOpts {
...
-            &host_phase_2_id,
-            &self.discover_opts.mechanism,
-            &log,
-            &event_sender,
-        )
-        .await?;
+        let engine = UpdateEngine::new(&log, event_sender);
+
+        let host_phase_2_id = ArtifactHashId {
+            // TODO: currently we're assuming that wicket will unpack the host
```
andrewjstone, line 187: s/wicket/wicketd

installinator/src/dispatch.rs
```diff
@@ -215,23 +245,45 @@ impl InstallOpts {
...
+                        &host_2_phase_id_2,
+                        &host_phase_2_artifact.artifact,
+                        &control_plane_id_2,
+                        &control_plane_artifact.artifact,
+                        destination,
+                    );
+
+                    let slots_written = writer.write(&cx, &log).await;
+
+                    // TODO: warning message in case one of the slots wasn't
```
andrewjstone, line 271: Were you planning to add these in this PR or a follow up? Just flagging in case you forgot.

installinator/src/peers.rs
```diff
@@ -135,6 +134,10 @@ impl FetchedArtifact {
                         "(attempt {attempt}) failed to discover peers, retrying: {}",
                         DisplayErrorChain::new(AsRef::<dyn std::error::Error>::as_ref(&error)),
                     );
+                    cx.send_progress(StepProgress::retry(format!(
```
andrewjstone, line 137: ❤️

andrewjstone (approved):
This is huge improvement by using the update engine types!

installinator/src/dispatch.rs
sunshowers: Thanks for catching this! I did meant to add these in this PR, will fix.

installinator-common/src/progress.rs
sunshowers: Yeah it's so much nicer now.

installinator/src/dispatch.rs
sunshowers: Done.

sunshowers:
Going to rebase and land this on top of #2867.

installinator/src/dispatch.rs
```diff
@@ -184,54 +179,121 @@ impl InstallOpts {
...
-            &log,
-            &event_sender,
-        )
-        .await?;
+        let engine = UpdateEngine::new(&log, event_sender);
+
+        let host_phase_2_id = ArtifactHashId {
+            // TODO: currently we're assuming that wicketd will unpack the host
+            // phase 2 image. We may instead have the installinator do it.
+            // kind: KnownArtifactKind::Host.into(),
```
jgallagher, line 189: This just moved from above, but I think this line of the comment can be deleted.

installinator/src/peers.rs
```diff
@@ -135,6 +134,10 @@ impl FetchedArtifact {
...
+                        "failed to discover peers: {error}"
+                    )))
+                    .await;
```
jgallagher, line 140: If we haven't found any peers, this retry message ultimately gets dropped, right?

installinator/src/reporter.rs
```diff
@@ -137,15 +121,15 @@ where
             })
             .cloned()
             .collect();
-        // IMPORTANT: You might be tempted to replace this reverse call with
+        // STEP: You might be tempted to replace this reverse call with
```
jgallagher, line 124: I think this change might've been an accident? :)

installinator/src/write.rs
```diff
@@ -213,92 +192,69 @@ impl<'a> ArtifactWriter<'a> {
...
+                        Ok(())
+                    })
+                    .await;
 
-                        new_progress
+                match res {
+                    Ok(_) => {
+                        // This drive succeeded in this iteration.
+                        *progress = DriveWriteProgress::Done;
+                        done_drives.insert(*drive);
```
jgallagher, line 241: Can this push the same drive onto `done_drives` twice? The sequence I'm worried about:

1. First iteration
  a. Writing drive A succeeds; push A on to `done_drives`
  b. Writing drive B fails
  c. End of iteration; `success_this_iter` is 1, so we loop
2. Second iteration
  a. Drive A's progress is `DriveWriteProgress::Done`, so `write_cx.register_steps()` doesn't register any steps - I think this means the engine result will be `Ok(_)`, right? Then we hit this branch again, and push A into `done_drives` a second time (and also increment `success_this_iter`, which doesn't seem right).
  b. Result on drive B is irrelevant
  c. `success_prev_iter` is 1, so we break out of the loop

Maybe we should check for `DriveWriteProgress::Done` at the top of the for loop and `continue` if the drive is done?

installinator/src/dispatch.rs
```diff
@@ -184,54 +179,121 @@ impl InstallOpts {
...
-
-        // TODO: verify artifact was correctly written out to disk.
-
-        // Drop the event sender: this signals completion.
-        _ = event_sender
-            .send(ReportEvent::Completion(CompletionEventKind::Completed))
-            .await;
-        std::mem::drop(event_sender);
+        // XXX: look at breaking this up into separate steps. This is tricky
+        // because of the combined logic around handling retries.
```
jgallagher, line 249: Is this comment still relevant given `ArtifactWriter` itself uses a nested engine with separate steps?

jgallagher (commented):
This all looks really great. Mostly nits, with one minor issue on reporting drives (maybe!).

installinator/src/dispatch.rs
sunshowers: Done, thanks for catching this.

installinator/src/peers.rs
sunshowers: Yes, it will be buffered up, and then dropped if we run out of capacity.

installinator/src/reporter.rs
sunshowers: Yes. I got rid of this code and comment with the more principled approach in #2890.

installinator/src/dispatch.rs
sunshowers: No, not any longer. Thanks!

installinator/src/write.rs
sunshowers: Hmm, so `done_drives` is a `BTreeSet` and not a `Vec` which means that inserting into it is idempotent.

`success_this_iter` just means that this drive was successful, whether in a previous iteration or in this one.

I added some comments explaining this.

jgallagher: Huh, when I wrote my comment I was definitely thinking `done_drives` was a `Vec`. My bad!

installinator/src/reporter.rs
```diff
@@ -117,37 +89,7 @@ where
...
-        // elements to be taken rather than the *last* 20 elements.
-        completion_events.reverse();
-
-        let progress_events = self.last_progress.clone().into_iter().collect();
-        let report = ProgressReport {
-            total_elapsed: self.start.elapsed(),
-            completion_events,
-            progress_events,
-        };
+        let report = self.buffer.generate_report();
```
jgallagher, line 92: Hah, yes, this is much nicer indeed :+1:

wicketd/src/update_tracker.rs
```diff
@@ -383,12 +374,6 @@ impl UpdateDriver {
         }
 
         // Spawn a task to accept all events from the executing engine.
-        //
-        // TODO see note in `process_installinator_report()` below: we're racing
-        // it for access to `event_buffer`. We should be the only one with it,
-        // but for now we share it to allow installinator reports to go into the
-        // log directly.
-        let event_buffer = Arc::clone(&update_cx.event_buffer);
```
jgallagher, line 391: Great to see this disappear :+1:

wicketd/src/update_tracker.rs
```diff
@@ -1021,283 +1001,6 @@ impl UpdateContext {
...
-        // TODO: This races with our task that receives events from the update
-        // engine in `run()` above. Once we integrate installinator sub-steps we
-        // should remove `event_buffer` from `UpdateContext` altogether, and let
-        // just the event-receiving task hold it.
-        let mut event_buffer = self.event_buffer.lock().unwrap();
-        for event in events {
-            event_buffer.add_step_event(event);
-        }
-        event_buffer.add_progress_event(progress_event);
-    }
```
jgallagher, line 1299: Even better to see _this_ disappear :heart_eyes:

jgallagher (approved):

PR #2833, opened 2023-04-13
[wicketd] sled-agent configures wicketd to serve artifacts on the bootstrap network

jgallagher:
Addresses a lingering TODO. Tested on the `london` scrimlet:

```
support@oxz_switch:~$ ipadm | grep boot
oxBootstrap0/bootstrap6 static ok       fdb0:a840:2504:392::2/64
support@oxz_switch:~$ netstat -aun | grep fdb0:a840:2504:392::2
fdb0:a840:2504:392::2.12227             *.*                                                             0      0 128000      0 LISTEN
```

12227 is the expected artifact server port, and I confirmed it's the `wicketd` process listening. I can't yet test connectivity from outside the sled, but I think this is now working.

### DIFF
--- a/sled-agent/src/services.rs
+++ b/sled-agent/src/services.rs
@@ -502,33 +502,20 @@ impl ServiceManager {
     //
     // No other zone besides the switch and global zones should have a
     // bootstrap address.
-    fn bootstrap_vnic_needed(
-        &self,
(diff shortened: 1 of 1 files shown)


andrewjstone (approved):
Looks great. Thanks!

Issue #2341, opened 2023-02-09
Tracking: System Update API (phase 2)


david-crespo:
See #2107 for phase 1 tracking issue and #2100 for most of the code. That work was about setting up the endpoints and the underlying data model without much reference to the actual TUF repo and update artifacts or the update planner. The updates we display in the console is mock data added in #2283 and #2294. 

This issue tracks work around replacing that mock data with real data by populating the update tables from the TUF repo and the component version from the actual installed software present in the rack. See #250 for the work on the underlying systems and planner.

```[tasklist]
- [ ] https://github.com/oxidecomputer/omicron/issues/2344
- [ ] https://github.com/oxidecomputer/omicron/issues/2040
- [ ] https://github.com/oxidecomputer/omicron/pull/2834
- [ ] https://github.com/oxidecomputer/omicron/issues/2838
- [ ] Combine UpdateArtifact and ComponentUpdate, they are the same thing
- [ ] Sort update deployments by `time_created` desc (requires #2343)
- [ ] https://github.com/oxidecomputer/omicron/issues/766
```

davepacheco:
I believe this needs we ripped most of this out already and need to add new APIs for this.  I expect we'll file new issues for that.  (Please feel free to reopen if I'm wrong!)

PR #2834, opened 2023-04-13
[update] Renames to prep for real update work

david-crespo:
I have been working on #2341, and I have some renames that would just clutter up the real PR, so here they are. The only real change here is that `create_system_update` is now an upsert and the name change reflects that.


nexus/db-queries/src/db/datastore/update.rs
```diff
@@ -77,6 +78,9 @@ impl DataStore {
 
         diesel::insert_into(system_update)
             .values(update.clone())
+            .on_conflict(version)
+            .do_update()
+            .set(time_modified.eq(Utc::now()))
```
david-crespo, line 83: the only non-rename change here

PR #2836, opened 2023-04-14
Introduce Propolis ID generations

gjcolombo:
This is the fourth of eight (soon to be nine) commits in the [live migration sequence](https://github.com/oxidecomputer/omicron/compare/main...gjcolombo:omicron:gjcolombo/lets-migrate/8-migration-end).

Live migration introduces a way for an instance's Propolis ID, sled ID, server IP, and migration information to change on the fly. To update this state safely, introduce to the instance runtime state the concept of a "Propolis generation" that governs how an instance's current "location" is allowed to change.

An instance's runtime state can update from A to B if

- A and B have the same Propolis generation and Propolis ID, and B has a later runtime state generation than A, or
- B has a later Propolis generation than A.

This creates an orderly way to change an instance's "Propolis pointer" during a migration without forcing the newer Propolis's sled to worry about the state generations the old sled was supplying. See RFD 361 for more details.

This change just defines the new column. It will come into regular use in future changes.


nexus/db-queries/src/db/datastore/instance.rs
```diff
@@ -180,11 +180,18 @@ impl DataStore {
         let updated = diesel::update(dsl::instance)
             .filter(dsl::time_deleted.is_null())
             .filter(dsl::id.eq(*instance_id))
-            .filter(dsl::state_generation.lt(new_runtime.gen))
+            // Runtime state updates are allowed if either:
+            // - the active Propolis ID will not change, the state generation
+            //   increased, and the Propolis generation will not change, or
```
smklein, line 185: Nice, it's impressive you pulled this off while keeping a single "runtime update" function. It's nice this logic is centralized.

nexus/db-model/src/instance.rs
```diff
@@ -100,32 +100,74 @@ impl DatastoreAttachTargetConfig<Disk> for Instance {
 )]
 #[diesel(table_name = instance)]
 pub struct InstanceRuntimeState {
-    /// runtime state of the Instance
+    /// The instance's current user-visible instance state.
```
smklein, line 103: Thank you for fully refreshing these docs!

smklein (approved):

PR #2856, opened 2023-04-17
OPTE port cleanup and refactor.

luqmana:
Some refactoring and cleanup primarily to support OPTE ports for services. Each commit is relatively self-contained but the higlights are:
1. Updating `PortManager` to handle service vs guest ports.
2. Enabling IPv6 support (opte has supported this for a while now, just needed to plumb things through).
3. Leaving port cleanup to `RunningZone`.


illumos-utils/src/running_zone.rs
```diff
@@ -368,8 +368,21 @@ impl RunningZone {
...
     /// Return references to the OPTE ports for this zone.
-    pub fn opte_ports(&self) -> &[Port] {
-        &self.inner.opte_ports
+    pub fn opte_ports(&self) -> impl Iterator<Item = &Port> {
+        self.inner.opte_ports.iter().map(|(port, _)| port)
+    }
+
+    /// Remove the OPTE ports on this zone from the port manager.
+    ///
+    /// Returns the last error, if any.
```
leftwo, line 377: This is how it was before, so I don't think a change here is in scope for this PR, but why only return the last error?  Won't an error in the middle be dropped if we just return the last result?

illumos-utils/src/opte/mod.rs
```diff
@@ -66,7 +74,7 @@ impl Gateway {
             ip: subnet
                 .iter()
                 .nth(1)
-                .expect("IP subnet must have at least 1 address"),
+                .expect("IP subnet must have at least 2 addresses"),
```
leftwo, line 77: Was this check wrong before?  It seems like it was already trying to return the 2nd element (if I'm understanding `nth()` correctly)

illumos-utils/src/opte/port.rs
```diff
@@ -54,9 +36,10 @@ struct PortInner {
     vnic: String,
 }
 
+#[cfg(target_os = "illumos")]
```
leftwo, line 39: Has the door finally closed on non-illumos builds?  (I know I don't, but I was not sure if others do)

illumos-utils/src/opte/mod.rs
luqmana: This is just updating the message here to be more accurate. You're right in that we're grabbing the second address, hence why the given subnet needs at least 2 addresses to begin with, not 1.

illumos-utils/src/opte/port.rs
luqmana: It's still at the same level of support for non-illumos as previously, i.e. replacing illumos specific calls with stubs. But this reduces the duplication we had and also keeps a lot of the parameter validation and stuff consistent without having to hop between 2 files every time you want to change something in PortManager/Port.

illumos-utils/src/running_zone.rs
luqmana: Yes, it's not ideal but I think the original idea was we also don't want to release some of the ports and not do the rest if we encounter a failure. Maybe worth returning all the errors encountered instead?

leftwo: Yeah, I agree that we should try to delete all the ports, even if some fail.  I was just worried that a failure in
the middle would be ignored and not reported.  Now, is anyone listening to that failure on the other side and
what possibly could they do if a port release failed, I don't know, but I would think it nicer to at least give
the caller the option of doing something about it.

illumos-utils/src/opte/mod.rs
leftwo: Ah, I thought we were adding 2 address support, but you were just being a good citizen and fixing the message.

illumos-utils/src/running_zone.rs
luqmana: Ok looking at the code again I realized we could just make this infallible by having `PortTicket::release` take `self` by-value. There's always a single `PortTicket` per-port and the error condition as it was written before was if you somehow tried to release and already release `Port`. By just consuming the `PortTicket` instead we shouldn't be able to run into that scenario altogether.

leftwo (approved):
No further comments

Issue #2858, opened 2023-04-17
Sled Agent should be capable of utilizing swap space on the M.2 dump device


smklein:
Parent bug: https://github.com/oxidecomputer/omicron/issues/2688

jclulow:
NB: I have some issues at a design level I've outlined in a command on the PR: https://github.com/oxidecomputer/omicron/pull/2859#issuecomment-1511770397

jordanhendricks:
I've begun this work, but am not going to have a PR ready before I go on PTO for ~1.5 weeks.

The design we've reached consensus on is that the swap device will be an encrypted zvol (with an ephemeral key) on the zpool of the M.2 the sled booted from.

We will create the device as such (though the actual zfs commands will be run by sled-agent):
```
# starting with 256GB until testing proves otherwise, but will make this configurable via sled-agent manifest
size=<from sled-agent manifest>
blocksize=`getconf PAGESIZE`

/sbin/zfs create \
        # creates a sparse volume with no reservation
        -s \

        # volume of a given size; exported as a block device in /dev/zvol/{dsk,rdsk}/<path>
        -V $size \

        # equivalent to volblocksize=blocksize
        -b $blocksize \

        # -o property=value
        #
        # don't use log device
        -o logbias=throughput \

        # only cache metadata in ARC
        -o primarycache=metadata \

        # don't use secondary cache
        -o secondarycache=none \

        # encryption format
        -o encryption=aes-256-gcm \

        # raw key (32 bytes)
        -o keyformat=raw \

        # path to key
        -o keylocation=file://path/to/key \

        # name of device
        oxi_<M.2 zpool uuid>/swap
```

The device will be added via `swapctl(2)`.

Because the size the device should be is still an open question, we may need to resize it if it already exists. Thus, on startup, sled-agent will do the following:
- check if there is a swap device, and if so, whether it is a zvol of the right size
- if the zvol exists, but is the wrong size, resize it; then give the existing zvol a new key, and add it as a swap device if it's not already
- if there is no swap device, then create the zvol and add it via swapctl

PR #2859, opened 2023-04-17
[sled-agent] Add DumpDevice on M.2 as swap device

smklein:
Fixes https://github.com/oxidecomputer/omicron/issues/2858

### DIFF
--- a/sled-hardware/src/lib.rs
+++ b/sled-hardware/src/lib.rs
@@ -3,6 +3,7 @@
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 use illumos_utils::fstyp::Fstyp;
+use illumos_utils::swap::Swap;
 use illumos_utils::zpool::Zpool;
 use illumos_utils::zpool::ZpoolName;
 use schemars::JsonSchema;
@@ -138,6 +139,8 @@ pub enum DiskError {
     ZpoolCreate(#[from] illumos_utils::zpool::CreateError),
     #[error("Cannot import zpool: {0}")]
     ZpoolImport(illumos_utils::zpool::Error),
+    #[error("Failed to access swap: {0}")]
+    SwapError(#[from] illumos_utils::swap::Error),
     #[error("Cannot format {path}: missing a '/dev' path")]
     CannotFormatMissingDevPath { path: PathBuf },
     #[error("Formatting M.2 devices is not yet implemented")]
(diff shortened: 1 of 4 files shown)


sled-hardware/src/lib.rs
```diff
@@ -286,17 +289,69 @@ pub struct Disk {
...
+        let variant = unparsed_disk.variant;
         // First, ensure the GPT has the right format. This does not necessarily
         // mean that the partitions are populated with the data we need.
-        let partitions =
-            ensure_partition_layout(&log, &paths, unparsed_disk.variant)?;
+        let partitions = ensure_partition_layout(&log, &paths, variant)?;
 
+        Self::initialize_swap(&log, &paths, &partitions)?;
+        let zpool_name =
+            Self::initialize_zpool(&log, variant, &paths, &partitions)?;
```
smklein, line 306: This code is unchanged from before this PR, just refactored

sled-hardware/src/lib.rs
```diff
@@ -286,17 +289,69 @@ pub struct Disk {
...
     ) -> Result<Self, DiskError> {
         let paths = &unparsed_disk.paths;
+        let variant = unparsed_disk.variant;
         // First, ensure the GPT has the right format. This does not necessarily
         // mean that the partitions are populated with the data we need.
-        let partitions =
-            ensure_partition_layout(&log, &paths, unparsed_disk.variant)?;
+        let partitions = ensure_partition_layout(&log, &paths, variant)?;
 
+        Self::initialize_swap(&log, &paths, &partitions)?;
```
smklein, line 304: This call is new

sled-hardware/src/lib.rs
```diff
@@ -286,17 +289,69 @@ pub struct Disk {
...
+    // Initize a swap device within the DumpDevice partition, if one exists.
+    fn initialize_swap(
+        log: &Logger,
+        paths: &DiskPaths,
+        partitions: &Vec<Partition>,
+    ) -> Result<(), DiskError> {
+        match paths.partition_device_path(
+            &partitions,
+            Partition::DumpDevice,
+            false,
```
jordanhendricks, line 327: Part of this is probably my own unfamiliarity with the sled-hardware code, but it might be good to explain why this is going in the DumpDevice partition?

sled-hardware/src/lib.rs
```diff
@@ -286,17 +289,69 @@ pub struct Disk {
...
+            }
+            // The disk should not have a dump device: ignore it.
+            Err(DiskError::NotFound { .. }) => {}
+            // The disk maybe has a dump device, but we can't read it: bail.
+            Err(e) => return Err(e),
+        };
+        Ok(())
+    }
+
+    // Initialize a swap within the ZfsPool partition.
```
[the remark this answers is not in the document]
jordanhendricks: Also, I don't see where we do that here. Shouldn't there be a call to `Swap::set_swap` if that were true?

sled-hardware/src/lib.rs
smklein: I've struggled to find a great summary for this in the illumos docs, but I'm really just trying to follow the convention for this which has been set there.

Reading a combination of 
- https://illumos.org/man/8/dumpadm
- https://www.illumos.org/man/8/swap
- https://github.com/illumos/illumos-gate/blob/master/usr/src/cmd/swap/swap.c

I was able to infer:

- Apparently, it's common to use swap space as a dump device
- This is even the default mode of `swap -a` -- if [a swap device is added, and no dump devices exist, that swap device is used as a dump device](https://github.com/illumos/illumos-gate/blob/cd6904adfb4d5d78d93238b8062f8dd3ef70a103/usr/src/cmd/swap/swap.c#L571-L576).

I added a summary in 4039c7a3fffdba392e576e55493433f68ecbe774

sled-hardware/src/lib.rs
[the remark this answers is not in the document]
smklein: good catch, this comment is totally wrong - fixed in 2e3533b41d8f0f3415d4e50d98465621b6fc74cd

sled-hardware/src/lib.rs
jordanhendricks: thanks, this was helpful.

jordanhendricks (approved):

PR #2861, opened 2023-04-17
sim sled agent: allow direct access to simulated instances

gjcolombo:
No. 5 in the [live migration series](https://github.com/oxidecomputer/omicron/compare/main...gjcolombo:omicron:gjcolombo/lets-migrate/8-migration-end). I suggest reading individual commits for this one to avoid the Clippy debris that's scattered about the overall PR.

---

Wrap `SimInstance`'s current state in an `Arc`/`Mutex` so that `SimInstance`s can be `Clone`. Then add a `SimCollection` impl that returns a clone of one of the objects in the collection.

This allows simulated sled agent code sitting above `SimCollection` to access a `SimInstance` in its collection and call methods on directly (with side effects that are visible to other users of the collection). Previously, the sled agent was more or less restricted to calling generic `Simulatable` methods for which `SimCollection` provided an appropriate proxy.

This allows the `InstanceStates` struct to implement interesting methods (e.g. setting and clearing migration IDs) that regular and simulated sled agents can share (previously such methods had to map to generic functionality in the `Simulatable` trait). Subsequent changes will use this heavily. For now, use it to clean up a bit of other simulated instance code.

### DIFF
--- a/sled-agent/src/sim/collection.rs
+++ b/sled-agent/src/sim/collection.rs
@@ -393,6 +393,19 @@ impl<S: Simulatable + 'static> SimCollection<S> {
     }
 }
 
+impl<S: Simulatable + Clone + 'static> SimCollection<S> {
+    pub async fn sim_get_cloned_object(
+        self: &Arc<Self>,
+        id: &Uuid,
+    ) -> Result<S, Error> {
+        let objects = self.objects.lock().await;
+        let instance = objects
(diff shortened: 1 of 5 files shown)


smklein (approved):
Makes sense, it tracks that we'd want to let tests manipulate these simulated instances as much as possible to test out / validate intrusive operations.

PR #2862, opened 2023-04-17
sled agent: add facility to set/clear instances' migration IDs

gjcolombo:
This is PR no. 6 in the [live migration series](https://github.com/oxidecomputer/omicron/compare/main...gjcolombo:omicron:gjcolombo/lets-migrate/8-migration-end). The work in this PR will be used in the next PR in this sequence.

---

Add a sled agent API that allows Nexus to set or clear the migration ID and destination Propolis ID in an instance's runtime state. This advances the instance's Propolis generation number.

Sled agent requires callers to furnish the instance runtime state they intend to transition from. This lets Nexus use this API and an instance's migration ID field to arbitrate between multiple concurrent migration attempts: if an instance is running with Propolis generation X, and two (or more) migration sagas observe that generation, then only one will successfully set migration IDs and be allowed to continue migration.

Subsequent commits will update the migration saga to use this API.

### DIFF
--- a/sled-agent/src/instance.rs
+++ b/sled-agent/src/instance.rs
@@ -7,11 +7,10 @@
 use crate::common::instance::{Action as InstanceAction, InstanceStates};
 use crate::instance_manager::InstanceTicket;
 use crate::nexus::LazyNexusClient;
-use crate::params::NetworkInterface;
-use crate::params::SourceNatConfig;
-use crate::params::VpcFirewallRule;
 use crate::params::{
-    InstanceHardware, InstanceMigrationTargetParams, InstanceStateRequested,
+    InstanceHardware, InstanceMigrationSourceParams,
+    InstanceMigrationTargetParams, InstanceStateRequested, NetworkInterface,
+    SourceNatConfig, VpcFirewallRule,
 };
 use anyhow::anyhow;
 use futures::lock::{Mutex, MutexGuard};
@@ -513,6 +512,11 @@ mockall::mock! {
             &self,
             state: InstanceStateRequested
         ) -> Result<InstanceRuntimeState, Error>;
+        pub async fn put_migration_ids(
+            &self,
+            old_runtime: &InstanceRuntimeState,
+            migration_ids: &Option<InstanceMigrationSourceParams>
+        ) -> Result<InstanceRuntimeState, Error>;
         pub async fn issue_snapshot_request(
             &self,
             disk_id: Uuid,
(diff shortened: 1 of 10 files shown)


sled-agent/src/instance.rs
```diff
@@ -715,6 +719,40 @@ impl Instance {
...
+        old_runtime: &InstanceRuntimeState,
+        migration_ids: &Option<InstanceMigrationSourceParams>,
+    ) -> Result<InstanceRuntimeState, Error> {
+        let mut inner = self.inner.lock().await;
+
+        // Check that the instance's current generation matches the one the
+        // caller expects to transition from. This helps Nexus ensure that if
+        // multiple migration sagas launch at Propolis generation N, then only
+        // one of them will successfully set the instance's migration IDs.
+        if inner.state.current().propolis_gen != old_runtime.propolis_gen {
```
smklein, line 733: I'm not necessarily suggesting this, but why are we not doing:

```suggestion
        if inner.state.current().propolis_gen < old_runtime.propolis_gen {
```

Is it possible for the "state on the sled" to lag behind Nexus' idea of the propolis_gen?

(To affirm my mental model - sled agent "owns" this generation number, so the value here *should* be the most recent?)

smklein (approved):

sled-agent/src/instance.rs
gjcolombo: My mental model of this sort of thing always goes toward "interlocked compare and exchange," which is an equality test, so I reached for `!=` by default.

If we wanted this to be directional, I think it would have to be `>` and not `<` (perhaps another argument for just making it `!=` and not having to scruple over it). For reference, the scenario I had in mind here is the following:

- Instance is running, let's say with Propolis generation 5; sled agent and CRDB agree on this value.
- Two migration sagas start in parallel. They generate different migration IDs.
- Both sagas read generation 5 from CRDB and ask to set the instance's migration ID to their chosen ID.

One of the two callers will take the instance lock first (line 727), set the migration IDs, and bump the generation to 6. It will (hopefully) get an `Ok` result and then proceed with its migration.

When the other call gets the lock, this condition will fail (6 != 5), and the check on line 736 will also fail because the ID from the second call doesn't match the ID from the first one. That error will cause the second saga to abort.

(The check on line 736 is there for idempotency. If the "winning" saga replays this step, the routine will still succeed: the idea is to say that the generation numbers don't match, but if we mutate the Nexus-supplied runtime state in the way the caller is asking, it *would produce* sled agent's record, so it's OK to proceed.)

smklein: Sounds good, I see the idempotency check, and that makes sense. I just wanted to *triple check* that ownership of the generation number, since bumping it on both sled agent + nexus would be hairy, and I wanted to confirm my understanding of "what are the limits on how out-of-sync this can get".

gjcolombo: Yeah, the way I'd summarize it is that if an instance is on a sled, then (as with the runtime state) the relevant sled agent synchronizes changes to the generation numbers, which is convenient because sled agents can use, y'know, *locks*. If an instance isn't on any sled, Nexus owns the information, but has to update it with caution and must use all the right transactional semantics in CRDB to update fields atomically.

PR #2863, opened 2023-04-17
[update-engine] improve jsonschema names for generated types

sunshowers:
As suggested in https://github.com/oxidecomputer/openapi-lint/issues/14,
use `schemars(rename = ...)` to generate names that pass openapi-lint
specifications.


ahl:
looks good; I'm surprised that this doesn't change any materialized openapi documents checked into the repo

andrewjstone (approved):
Ship it!

sunshowers:
> looks good; I'm surprised that this doesn't change any materialized openapi documents checked into the repo

We aren't using it yet, but soon will be in #2830 and #2697.

PR #2867, opened 2023-04-18
[wicketd] Integrate update-engine

jgallagher:
This will have some (hopefully relatively minor) conflicts with #2830, and the areas in conflict have TODOs in both PRs related to integrating them together, I believe.


wicketd-client/src/lib.rs
```diff
@@ -38,9 +38,14 @@ progenitor::generate_api!(
         RotInventory = { derives = [ PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize]},
         RotSlot = { derives = [ PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize]},
         ImageVersion = { derives = [ PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize]},
+        UpdateComponent = { derives = [ Copy, PartialEq, Eq, PartialOrd, Ord ] },
     }
 );
 
+pub type ProgressEventKind = types::ProgressEventKindForWicketdEngineSpec;
+pub type StepEventKind = types::StepEventKindForWicketdEngineSpec;
+pub type StepInfoWithMetadata = types::StepInfoWithMetadataForWicketdEngineSpec;
```
sunshowers, line 47: Ah interesting, the approach I took in #2830 was to instead depend on `update-engine` and use the `replace` directive. I think that approach works out slightly better overall because you can use methods defined on the various types more easily.

If we go down this path, we'd need to split some of the code out into a separate `wicketd-types` or `wicketd-common` crate.

wicketd/src/update_events.rs
```diff
@@ -248,21 +145,20 @@ impl fmt::Display for SpComponentUpdateStage {
     }
 }
 
+// On `allow(large_num_variant)`: "large" here is ~200 bytes, and we usually
```
sunshowers, line 148: Should be `large_enum_variant`

wicketd/src/update_tracker.rs
```diff
@@ -1070,90 +1042,266 @@ impl UpdateContext {
...
-                .unwrap_or(now);
-
-            (timestamp, UpdateEventKind::Normal(event_kind))
+            StepEvent { total_elapsed: event.total_elapsed, kind }
         });
+
+        // TODO: This races with our task that receives events from the update
+        // engine in `run()` above. Once we integrate installinator sub-steps we
+        // should remove `update_log` from `UpdateContext` altogether, and let
+        // just the event-receiving task hold it.
```
sunshowers, line 1298: Really looking forward to this haha

wicketd/tests/integration_tests/commands.rs
```diff
@@ -33,7 +33,13 @@ fn test_wicketd_openapi() {
...
 
     // Check for lint errors.
-    let errors = openapi_lint::validate(&spec);
+    let mut errors = openapi_lint::validate(&spec);
+    // XXX: For types like `StepEvent<WicketdEngineSpec>`, schemars generates
+    // names like `StepEvent_for_WicketdEngineSpec`. Filter them out since we
+    // can't control those names.
+    errors.retain(|s| {
+        !(s.contains("that is not PascalCase") && s.contains("_for_"))
+    });
```
sunshowers, line 42: This is no longer required as of #2863.

sunshowers (approved):
Looks fantastic!

wicketd-client/src/lib.rs
jgallagher: TBH these were just to shorten all the type names - I didn't need any methods defined on the types in `update-engine`, but since there's (currently?) only one spec involved in wicketd it didn't seem necessary to have `..ForWicketdEngineSpec` attached to all the type names.

I think this falls into the "do we even need `replace`" which was some feedback from Adam at one point, I think? On this PR I don't think we do.

wicketd/tests/integration_tests/commands.rs
jgallagher: I'm pointed at #2697 which I think (?) doesn't have that yet.

sunshowers: Yeah, #2697 doesn't have that yet. Feel free to combine this with #2697 and rebase :)

wicketd-client/src/lib.rs
sunshowers: Yeah I agree that this isn't immediately useful. But I'm wondering if reconciling this with #2830 isn't going to involve some of that. I think we can tackle that as needed though (I plan to do the integration work between the two.)

wicketd/tests/integration_tests/commands.rs
jgallagher: Rebased and removed this in 3fc0ac021d742a811dfc7feb56beec211e109a26

wicketd-client/src/lib.rs
```diff
@@ -37,6 +37,8 @@ progenitor::generate_api!(
     },
     replace = {
         Duration = std::time::Duration,
+        ProgressEventKindForGenericSpec = update_engine::events::ProgressEventKind<update_engine::GenericSpec<update_engine::NestedError>>,
```
sunshowers: This can be `ProgressEventForGenericSpec`.

jgallagher: Thanks, fixed in [3d88905](https://github.com/oxidecomputer/omicron/pull/2867/commits/3d889056d828b0b86983e56c8572d3b68aaea4e2)

PR #2868, opened 2023-04-18
fix case of DNS, NTP enums

davepacheco:
We have several enums whose variants include `InternalDNS`, `ExternalDNS`, `NTP`, etc.  [Rust's naming conventions](https://rust-lang.github.io/api-guidelines/naming.html) suggest these should be `InternalDns`, `ExternalDns`, `Ntp`, etc:

> In UpperCamelCase, acronyms and contractions of compound words count as one word: use Uuid rather than UUID, Usize rather than USize or Stdin rather than StdIn.

I don't really care, but what we're doing right now confuses tooling like serde/json-schema:
https://github.com/oxidecomputer/omicron/blob/9d6aeab9d14c6ca7ef32cfca9bc11f147b7bfff6/openapi/nexus-internal.json#L2267

It only affects the internal API right now.  But it seems like we should standardize on one convention here and it may as well be the Rust convention that existing tooling expects.

### DIFF
--- a/internal-dns/src/config.rs
+++ b/internal-dns/src/config.rs
@@ -426,7 +426,7 @@ mod test {
     fn display_srv_service() {
         assert_eq!(ServiceName::Clickhouse.dns_name(), "_clickhouse._tcp",);
         assert_eq!(ServiceName::Cockroach.dns_name(), "_cockroach._tcp",);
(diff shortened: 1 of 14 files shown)


david-crespo (approved):
👍

PR #2871, opened 2023-04-19
Follow migrated propolis instance in nexus' serial console proxy

lifning:
Draft to get CI looking things over for me - Still testing the full stack locally; requires https://github.com/oxidecomputer/propolis/pull/345

Implements the seamless following of migrated instances' serial console streams within nexus such that nexus' own clients just see a continuous stream of bytes.


david-crespo:
I thought you demoed serial console staying connected through a live migration — I'm curious what it did before. Did it stay connected but with some kind of hiccup? Or was that with you directly connected to propolis?

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -2006,11 +2011,34 @@ async fn instance_serial_console_stream(
...
+            Ok(())
+        }
+        Err(e) => {
+            let _ = client_stream
+                .close(Some(CloseFrame {
+                    code: CloseCode::Error,
+                    reason: e.to_string().into(),
+                }))
+                .await
+                .is_ok();
```
gjcolombo, line 2038: turbo nit/question for my understanding: since we're throwing away the result here anyway, is the `is_ok` needed?

nexus/src/app/instance.rs
```diff
@@ -1394,33 +1394,61 @@ impl super::Nexus {
...
-
-        Self::proxy_instance_serial_ws(conn.into_inner(), propolis_upgraded)
-            .await
-            .map_err(|e| Error::internal_error(&format!("{}", e)))
+                .await
+                .map_err(|e| Error::internal_error(&format!("{}", e)))
+            }
+            Err(e) => {
+                let message = format!(
+                    "Failed to connect to instance's propolis server: {}",
```
gjcolombo, line 1439: This predates this change and is present elsewhere, and so is very nitpicky (and could also be deferred to an issue), but I wonder if we should also take this opportunity to remove references to "Propolis" from the error strings in this path, especially if we're being very explicit about sending these strings back as reason strings in a websocket close message. (If I didn't know what Propolis was I might be confused by the message string.) I like the "websocket connection to client closed unexpectedly" verbiage below as a substitute; "websocket connection to instance's serial port" also has a nice ring to it, I think.

Again, totally optional for this PR, but might be a bit of fit-and-finish we want to file an issue for and come back to.

gjcolombo (approved):
LGTM, just a couple of cosmetic bits. Thanks for hooking this up!

nexus/src/external_api/http_entrypoints.rs
lifning: I think I wrote this as a way of being more explicit that the `must_use` thing we're ignoring is a Result rather than, say, a Future... after [cargo clippy caught a `let _ = ` that hid that from a refactor a couple weeks ago](https://github.com/oxidecomputer/propolis/pull/393#discussion_r1191912693).

nexus/src/app/instance.rs
lifning: agreed, just tossed it in here before merge: https://github.com/oxidecomputer/omicron/pull/2871/commits/db1e8f9ba9362731e3f6e89ec579126fd1ddcfae

PR #2879, opened 2023-04-19
Configure and launch Maghemite in the switch zone on a real scrimlet

Nieuwejaar:


rcgoodfellow:
I think this might not work as expected on a real scrimlet (or atop softnpu for that matter).

In order for `ddmd` to use `dpd` as the underlying data plane, it needs to be [told to do so](https://github.com/oxidecomputer/maghemite/blob/main/smf/ddm/manifest.xml#L22-L24). In looking at the [way things are currently set up](https://github.com/oxidecomputer/omicron/blob/f3828f31b134e35d7938a5f2474fbbd4e38db32a/sled-agent/src/services.rs#L1201-L1261), I don think that is happening.

jmpesp:
Something like this works:

```patch
diff --git a/sled-agent/src/services.rs b/sled-agent/src/services.rs
index f8f4633a..e832cf62 100644
--- a/sled-agent/src/services.rs
+++ b/sled-agent/src/services.rs
@@ -1344,10 +1344,12 @@ impl ServiceManager {
                     smfh.setprop("config/mode", &mode)?;
                     smfh.setprop("config/admin_host", "::")?;
 
-                    let maghemite_interfaces: Vec<AddrObject> =
-                        if is_gimlet().map_err(|e| {
+                    let is_gimlet = is_gimlet().map_err(|e| {
                             Error::Underlay(underlay::Error::SystemDetection(e))
-                        })? {
+                        })?;
+
+                    let maghemite_interfaces: Vec<AddrObject> =
+                        if is_gimlet {
                             (0..31)
                                 .map(|i| {
                                     // See the `tfport_name` function for how
@@ -1392,6 +1394,14 @@ impl ServiceManager {
                         ),
                     )?;
 
+                    if is_gimlet {
+                        // Maghemite for a scrimlet needs to be configured to
+                        // talk to dendrite
+                        smfh.setprop("config/dendrite", "true")?;
+                        smfh.setprop("config/dpd_host", "[::1]")?;
+                        smfh.setprop("config/dpd_port", DENDRITE_PORT)?;
+                    }
+
                     smfh.refresh()?;
                 }
             }
```

luqmana (approved):

PR #1902, opened 2022-11-01
[sled agent] Create and use self-assembling zones for storage zones

smklein:
Part of https://github.com/oxidecomputer/omicron/issues/1898
Relies on https://github.com/oxidecomputer/crucible/pull/498

Converts Crucible, Cockroach, and Clickhouse to be (mostly) self-assembling.

I'm happy to proceed and convert the rest of the zones we're launching using a similar format, if we like how this looks.

Fixes https://github.com/oxidecomputer/omicron/issues/2886

PR #2889, opened 2023-04-20
Update crucible

smklein:
I *think* this is blocking https://github.com/oxidecomputer/omicron/pull/1902 , because it's causing that PR to fail with a similar failure mode.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -436,7 +436,17 @@ dependencies = [
 [[package]]
 name = "bhyve_api"
 version = "0.0.0"
-source = "git+https://github.com/oxidecomputer/propolis?rev=40e8d8e280c312e44047ff9d6b3b261e4537977a#40e8d8e280c312e44047ff9d6b3b261e4537977a"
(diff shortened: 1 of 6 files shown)


leftwo (commented):
Do we also need propolis to be updated as well?

leftwo:
> Do we? Has there been an API-breaking change in the past few months?

There have been a few changes to the message protocol between upstairs and downstairs (and
one that is currently in review) so depending on where propolis is, it could be incompatible with
latest.  I'll go check and report back here.

leftwo:
So, we do need propolis to move forward.
Propolis is at: 894d44988c56713105bbbd07900d628781ab4dda
```
commit 894d44988c56713105bbbd07900d628781ab4dda
Author: liffy <629075+lifning@users.noreply.github.com>
Date:   Wed Feb 8 11:18:54 2023 -0800
```

And, a change in message protocol happened here:
```
commit 75519cb2dc91be78ee5792718575e5348ddbdc1c
Author: Alan Hanson <alan@oxidecomputer.com>
Date:   Mon Feb 27 05:50:15 2023 -0800 
```

So, at a minimum propolis will need that.

leftwo (commented):
When Gregs update to propolis lands, we will want that to be included here as well.

jmpesp (approved):

PR #2890, opened 2023-04-20
[update-engine] add an event buffer and reports

sunshowers:
The event buffer records step and progress events, and allows for periodic
generation of reports.

Also switch wicketd to using the event buffer and reports.

I use petgraph for the event tree, which seems like overkill at first but dramatically simplifies the algorithms compared to manually creating the tree.

TODO:

- [x] Tests
- [x] Move "waiting" state into a separate PR
- [x] Generate "waiting" events when the corresponding step event is seen
- [x] Add documentation


wicketd/src/update_tracker.rs
```diff
@@ -251,31 +250,29 @@ impl UpdateTracker {
...
         for (sp, update_data) in &*sp_update_data {
-            let update_log = update_data.update_log.lock().unwrap().clone();
+            let event_report =
+                update_data.event_buffer.lock().unwrap().generate_report();
             let inner: &mut BTreeMap<_, _> =
                 converted_logs.entry(sp.type_).or_default();
-            inner.insert(sp.slot, update_log);
+            inner.insert(sp.slot, event_report);
         }
         converted_logs
```
sunshowers, line 277: A thing these methods do not do is reset the event report once it's delivered. It looks like the old update log implementation didn't do that either. We either need to find a way to run some code after a successful request, or provide a way for wicket to tell wicketd/the event buffer which events it has already seen, and an event buffer to diff against that.

update-engine/src/buffer.rs
```diff
@@ -0,0 +1,913 @@
...
+        if let Some(key) = self.recurse_for_step_event(&event, None) {
+            match self.map.entry(key) {
+                hash_map::Entry::Occupied(mut entry) => {
+                    let value = entry.get_mut();
+                    let progress_event = event.progress_event();
+                    if event.kind.priority() == StepEventPriority::High {
+                        // Handle duplicate events.
+                        if self.high_priority.last() != Some(&event) {
+                            self.high_priority.push(event);
+                        }
```
sunshowers, line 171: This isn't correct -- if we want to be resilient to duplicate events, we need to search through all earlier events (and ensure `high_priority` is in sorted order).

wicketd/src/update_tracker.rs
```diff
@@ -388,32 +385,17 @@ impl UpdateDriver {
...
-        // directly.
-        let update_log = Arc::clone(&update_cx.update_log);
+        // it for access to `event_buffer`. We should be the only one with it,
+        // but for now we share it to allow installinator reports to go into the
+        // log directly.
+        let event_buffer = Arc::clone(&update_cx.event_buffer);
         let event_receiving_task = tokio::spawn(async move {
             // TODO Is this event receiving handler right? Can we end up seeing
             // `current` set to a progress report for a step that is also
             // present in the completed `.events` list?
```
jgallagher, line 395: I think this comment can be removed now that `event_buffer` is handling it.

wicketd/src/update_tracker.rs
jgallagher: That was intentional: we don't assume a single `wicket` invocation will be alive the entire time an update is running, so if a new `wicket` starts it could always fetch the full update log at any point. I think with your option 2 that would still work: a `wicket` that has seen some state before could request just new events, and a new `wicket` could request the full report.

I _think_ this is an optimization we could delay for now, though, right? The old update log had a pretty small max size - the `EventReport` is bigger but maybe not too big?

sunshowers: Yeah, this is not a relevant optimization for now. However, for other reasons I implemented option 2 so we can plug this in at any time.

update-engine/src/buffer.rs
sunshowers: Done. This simplifies the algorithm quite a bit!

wicketd/src/update_tracker.rs
sunshowers: Yep, good catch!

update-engine/src/buffer.rs
```diff
@@ -0,0 +1,911 @@
...
+    fn add_node(
+        &mut self,
+        execution_id: ExecutionId,
+        index: usize,
+    ) -> Option<EventKey> {
+        if Some(&CompletedIndex::LastSeen(index))
+            <= self.last_completed_index.get(&execution_id)
+        {
+            // We've already seen this step be completed. Ignore this
+            // progress event.
```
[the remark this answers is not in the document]
sunshowers: Yeah, it could. I rewrote this to clean it up so this comment is obsolete and removed.

update-engine/src/buffer.rs
```diff
@@ -0,0 +1,1011 @@
...
+/// * dropped *low-priority* step events
+///
+/// The buffer is currently *not* resilient against:
+/// * dropped *high-priority* step events
+/// * reordered progress or step events
+///
+/// These cases can be handled on a best-effort basis in the future at some cost
+/// to complexity, if required.
+#[derive_where(Clone, Debug)]
+pub struct EventBuffer<S: StepSpec> {
```
sunshowers, line 52: BTW I didn't really have a better name for this. I originally called it `EventCoalescer` but I felt it was too wordy. The main goal of this is to track the state of an update engine running elsewhere.

jgallagher: I like `EventBuffer` more than `EventCoalescer` too. Maaaaybe something like `EventCombiner` or `EventReportCombiner`? I'm fine with `EventBuffer` though.

jgallagher (approved):

PR #2891, opened 2023-04-20
Add some scripts to make updating packages easier

smklein:
## What's New

This PR adds some scripts:
- `./tools/update_propolis.sh`
- `./tools/update_crucible.sh`
- `./tools/update_maghemite.sh`
- `./tools/update_dendrite.sh`

## How to Use Them

```
usage: ./tools/update_propolis.sh [-c COMMIT] [-n]

  -c COMMIT   Ask to update Propolis to a specific commit.
              If this is unset, Github is queried.
  -n          Dry-run
```

These scripts can be used to automate the following:
- They parse commits and SHAs from `Cargo.toml` and `package-manifest.toml`
- They automatically update the corresponding SHA256 hashes for all artifacts

Examples:

Update the propolis images and crates to the latest revision in Github
```
./tools/update_propolis.sh
UPDATING PACKAGES ---
  target: e7ce7a1dbe2d410b787df2e52478229b008b4750
  dry_run: 
  repo: oxidecomputer/propolis
  packages: propolis-server
+ for IDX in "${!PACKAGES[@]}"
+ echo 'UPDATING: [propolis-server]'
UPDATING: [propolis-server]
+ sed -i -e s/e7ce7a1dbe2d410b787df2e52478229b008b4750/e7ce7a1dbe2d410b787df2e52478229b008b4750/g /home/smklein/repos/oxide/omicron/tools/../package-manifest.toml
+ sed -i -e s/d4782227629464288a6d1cfdd7bdef135c179ddd2b0d573e785e758b7d063285/d4782227629464288a6d1cfdd7bdef135c179ddd2b0d573e785e758b7d063285/g /home/smklein/repos/oxide/omicron/tools/../package-manifest.toml
+ set +o xtrace
OK: Update complete
UPDATING CRATES ---
  target: e7ce7a1dbe2d410b787df2e52478229b008b4750
  dry_run: 
  repo: oxidecomputer/propolis
  crates: propolis-client propolis-server
+ for IDX in "${!CRATES[@]}"
+ echo 'UPDATING: [propolis-client]'
UPDATING: [propolis-client]
+ sed -i -e s/e7ce7a1dbe2d410b787df2e52478229b008b4750/e7ce7a1dbe2d410b787df2e52478229b008b4750/g /home/smklein/repos/oxide/omicron/tools/../Cargo.toml
+ for IDX in "${!CRATES[@]}"
+ echo 'UPDATING: [propolis-server]'
UPDATING: [propolis-server]
+ sed -i -e s/e7ce7a1dbe2d410b787df2e52478229b008b4750/e7ce7a1dbe2d410b787df2e52478229b008b4750/g /home/smklein/repos/oxide/omicron/tools/../Cargo.toml
+ set +o xtrace
OK: Update complete
```

## Implementation

- Basically, these scripts just send curl requests to github / buildomat, and do a find/replace of these exact commit strings within the specified files.
- This obviously has some downsides! If these strings appear in comments / unrelated spots, we'd replace them.
- Ideally, we'd use a more structured system for parsing and updating our toml files -- perhaps like https://docs.rs/toml_edit/latest/toml_edit/ - but I leave that as an exercise for the reader.


package-manifest.toml
```diff
@@ -246,10 +246,10 @@ service_name = "propolis-server"
 only_for_targets.image = "standard"
 source.type = "prebuilt"
 source.repo = "propolis"
-source.commit = "880a031fbdd7417cd3a4643f8bf419d8d5ea8224"
+source.commit = "40e8d8e280c312e44047ff9d6b3b261e4537977a"
 # The SHA256 digest is automatically posted to:
 # https://buildomat.eng.oxide.computer/public/file/oxidecomputer/propolis/image/<commit>/propolis-server.sha256.txt
-source.sha256 = "93265033a62a998ebc2d64ed3c0f0d4a97c16d34132d3d59660f5430a24990e9"
+source.sha256 = "41ebe7b2cda0fbf955ac72090760e33230dfa65ed2f71e31a996922d31d2304b"
```
smklein, line 252: These were out-of-sync between `Cargo.toml` and `package-manifest.toml`, so I just updated to the "latest available" one to let the scripts function.

jmpesp:
> Ideally, we'd use a more structured system for parsing and updating our toml files -- perhaps like https://docs.rs/toml_edit/latest/toml_edit/ - but I leave that as an exercise for the reader.

I was trying to do something like this with https://github.com/oxidecomputer/lockstep, which would compare checked out directories to get around the problem of private repos. lockstep only prints out the instructions though, requiring manual editing. I have a branch locally that tries to automatically edit the files too but haven't had the time to test it.

what do you think, is that tool worth fixing up or should we scrap it?

smklein:
> > Ideally, we'd use a more structured system for parsing and updating our toml files -- perhaps like https://docs.rs/toml_edit/latest/toml_edit/ - but I leave that as an exercise for the reader.
> 
> I was trying to do something like this with https://github.com/oxidecomputer/lockstep, which would compare checked out directories to get around the problem of private repos. lockstep only prints out the instructions though, requiring manual editing. I have a branch locally that tries to automatically edit the files too but haven't had the time to test it.
> 
> what do you think, is that tool worth fixing up or should we scrap it?

I think lockstep has utility for someone who is actively working across all repos, but IMO the requirement to check out and point to the whole git repo is an unnecessary extra lift that *probably* isn't needed? 

These scripts just require "any git commit", which makes jumping forward / backward in time a bit easier

tools/crucible_version
```diff
@@ -0,0 +1,5 @@
+# This should match the commit on package-manifest.toml
```
jmpesp, line 1: Do we need to duplicate the version here?

smklein: I went ahead and beefed up these scripts:

- They now parse all the files they modify, rather than having a duplicated version
- I moved a lot of this logic into a `./tools/update_helpers.sh`

smklein:
> FWIW, I actually want to add something similar for Maghemite and Dendrite -- and perhaps could? But the feature of "ask github for the latest commit" doesn't really work because they're private repos.

Thanks to a pro-tip from @jclulow , this is done - I'm pulling the latest commit hash from *buildomat* which also hosts this info.

jmpesp (approved):

PR #2896, opened 2023-04-20
[update-engine] add a "waiting for progress" state

sunshowers:
This new state cleans up a lot of code that would otherwise have to
manually emulate progress events.

For certain step events, we can synthesize this state. This lets consumers
use a "waiting for progress" state internally without having to resort
to workarounds like their own enum.


sunshowers:
Need to:

- [ ] Move progress to being an item within the corresponding step events, to ensure structural correctness.
- [ ] Generate progress for nested events.

sunshowers:
So we have two alternatives:

1. Make `progress` a field within the step event, do the generation of events in the update engine.
2. Generate a synthetic progress event out of the step event on the fly.

1 feels better in principle, but leads to some bloat in serialized data (especially when nested events are involved). Thoughts on 1 vs 2?

sunshowers:
Discussed with John, going to choose 2 for this.

sunshowers:
Discussed with John, going to choose option 2.

update-engine/src/events.rs
```diff
@@ -75,6 +75,94 @@ pub struct StepEvent<S: StepSpec> {
 }
 
 impl<S: StepSpec> StepEvent<S> {
+    /// Returns a progress event associated with this step event, if any.
+    ///
+    /// Some step events have an implicit progress event of kind
+    /// [`ProgressEventKind::WaitingForProgress`] associated with them. This
+    /// causes those step events to generate progress events.
+    pub fn progress_event(&self) -> Option<ProgressEvent<S>> {
```
jgallagher, line 83: This looks great; glad we went with option 2. :+1:

jgallagher (approved):

PR #2898, opened 2023-04-20
Remove params::ImageSource::Url

### LINKED ISSUE #2893: Remove the ability to create images from URLs
Remove the Url variant here:

https://github.com/oxidecomputer/omicron/blob/dee9d45f3381a70d6947b8b00743b7582efaf7d0/nexus/types/src/external_api/params.rs#L1188-L1202

This is superceded by importing via uploaded files or from a url to a disk.

jmpesp:
Do not allow customers to create an image from a URL, that was for development purposes only. Now that Nexus supports importing blocks via the Pantry this is no longer required.

Closes #2893


smklein (approved):

jmpesp:
One bit of fallout is that all the tests that previously used Url (via httptest) now use `YouCanBootAnythingAsLongAsItsAlpine`. That also seems like a target to remove before shipping, but it can't be removed without requiring each integration test to use snapshots. Maybe that isn't all that bad?

It's a shame that `params::ImageSource` can't have a `#[cfg(test)]` variant. I also thought for a second that `#[serde(skip)]` might work:

```patch
diff --git a/nexus/types/src/external_api/params.rs b/nexus/types/src/external_api/params.rs
index 2fc64585..8be56885 100644
--- a/nexus/types/src/external_api/params.rs
+++ b/nexus/types/src/external_api/params.rs
@@ -1194,7 +1194,9 @@ pub enum ImageSource {
     },
 
     /// Boot the Alpine ISO that ships with the Propolis zone. Intended for
-    /// development purposes only.
+    /// development purposes only, so skip it when serializing and
+    /// deserializing.
+    #[serde(skip)]
     YouCanBootAnythingAsLongAsItsAlpine,
 }
 ```

but this fails during the integration tests because they do

```rust
    // Create an image in the project
    let image_create_params = get_image_create();

    NexusRequest::objects_post(client, &images_url, &image_create_params)
        .authn_as(AuthnMode::PrivilegedUser)
        .execute_and_parse_unwrap::<views::Image>()
        .await;
```

which will serialize the struct.

jmpesp:
Another wrinkle! The end to end tests also use this:

```rust
    let image_id = ctx
        .client
        .system_image_create()
        .body(GlobalImageCreate {
            name: generate_name("debian")?,
            description: String::new(),
            block_size: 512.try_into().map_err(anyhow::Error::msg)?,
            distribution: Distribution {
                name: "debian".try_into().map_err(anyhow::Error::msg)?,
                version: "propolis-blob".into(),
            },
            source: ImageSource::Url {
                url:
                    "http://[fd00:1122:3344:101::1]:54321/debian-11-genericcloud-amd64.raw"
                        .into(),
            },
        })
        .send()
        .await?
        .id;
```

karencfv (commented):
Should the `url` field in the `Image` type be removed as well? https://github.com/oxidecomputer/omicron/blob/2a0595c70e16126d88a2b0140422cb052d8ce0ce/nexus/types/src/external_api/views.rs#L136-L137

leftwo:
Once this change is in, I think we can remove the `Url` `VolumeConstructionRequest` type as well.

jmpesp:
> Should the `url` field in the `Image` type be removed as well?
> 
> https://github.com/oxidecomputer/omicron/blob/2a0595c70e16126d88a2b0140422cb052d8ce0ce/nexus/types/src/external_api/views.rs#L136-L137

Yep! done in 2936101751c98806e682b7bda6c08937e6e7e556

karencfv (approved):
> Yep! done in https://github.com/oxidecomputer/omicron/commit/2936101751c98806e682b7bda6c08937e6e7e556

Thanks!!! Looks good to me :)

Issue #1215, opened 2022-06-15
Sled Agent startup logging is confusing


davepacheco:
While setting up the stack a few weeks ago, I found the Sled Agent logging pretty confusing about the current state of things.

The first thing to know is that Sled Agent startup makes calls out to Nexus, but the first time it starts up, Nexus isn't running yet.  Eventually, it will deploy the first Nexus instance.

I'll include the full log here for reference:

<details>

```
dap@sock ~/omicron $ BUNYAN_NO_PAGER=1 bunyan  $(svcs -L sled-agent) 
[ Jun  1 23:07:29 Enabled. ]
[ Jun  1 23:07:29 Executing start method ("ctrun -l child -o noorphan,regent /opt/oxide/sled-agent/sled-agent run /opt/oxide/sled-agent/pkg/config.toml &"). ]
[ Jun  1 23:07:29 Method "start" exited with status 0. ]
[ Jun  1 23:07:29 Rereading configuration. ]
[ Jun  1 23:07:29 No 'refresh' method defined.  Treating as :true. ]
note: configured to log to "/dev/stdout"
[2022-06-01T23:07:29.498928583Z]  INFO: SledAgent/2909 on sock: setting up bootstrap agent server
[2022-06-01T23:07:29.49930742Z]  INFO: SledAgent/2909 on sock: Ensuring config directory exists (path=/var/oxide)
[2022-06-01T23:07:29.525905774Z]  INFO: SledAgent/dropshot (Bootstrap)/2909 on sock: listening (local_addr=[fdb0:18c0:4d0d:a02a::1]:12346)
[2022-06-01T23:07:29.526285081Z]  INFO: SledAgent/BootstrapAgent/2909 on sock: bootstrap service initializing (server=fb0f7546-4d46-40ca-9d56-cbb810684ca7)
[2022-06-01T23:07:29.527375441Z]  INFO: SledAgent/RSS/2909 on sock:
    Injecting RSS configuration: SetupServiceConfig {
        rack_subnet: fd00:1122:3344:100::,
        requests: [
            SledRequest {
                datasets: [
                    DatasetEnsureBody {
                        zpool_uuid: d462a7f7b62840fe80ff4e4189e2d62b,
                        dataset_kind: Crucible,
                        address: [fd00:1122:3344:101::6]:32345,
                    },
                    DatasetEnsureBody {
                        zpool_uuid: e4b4dc87ab4649fba4b4d361ae214c03,
[...]

PR #2899, opened 2023-04-21
[sled-agent] Less noisy warning messages to logfile

smklein:
I noticed myself glossing over many `WARN` lines in the sled-agent, so I decided to make them less spammy.

Hopefully we can get to the point where each `WARN` is actually a notable event worthy of attention, without losing useful information about retries.

Part of https://github.com/oxidecomputer/omicron/issues/1215


common/src/backoff.rs
```diff
@@ -11,12 +11,69 @@
...
+    B: Backoff,
+    Fn: FnMut() -> Fut,
+    Fut: Future<Output = Result<I, BackoffError<E>>>,
+    N: NotifyTracking<E>,
+{
+    let mut count = 0;
+    let start = Instant::now();
+
+    let backoff_notify = |error, _duration| {
+        notify.notify_ext(error, count, Instant::now().duration_since(start));
```
jgallagher, line 70: Nit: I think this could be

```suggestion
        notify.notify_ext(error, count, start.elapsed());
```

common/src/backoff.rs
```diff
@@ -11,12 +11,69 @@
...
+pub async fn retry_notify_ext<I, E, Fn, Fut, B, N>(
+    backoff: B,
+    operation: Fn,
+    mut notify: N,
+) -> Result<I, E>
+where
+    B: Backoff,
+    Fn: FnMut() -> Fut,
+    Fut: Future<Output = Result<I, BackoffError<E>>>,
+    N: NotifyTracking<E>,
```
jgallagher, line 64: Out of curiosity more than a request for a change: Why do we have a special `NotifyTracking` trait (implemented for all `FnMut` of a certain shape) instead of taking an `FnMut` directly?

smklein: The backoff crate defines a `Notify` trait that's part of the interface for retry_notify: https://docs.rs/backoff/0.4.0/backoff/trait.Notify.html

This trait is also implemented generically for all `FnMut` that take "error, duration since last call" as an argument.

With `NotifyTracking`, I was trying to create something similar, just with different parameters ("error, call count, total duration").

I probably *could* skip the trait and just take a `FnMut`? But, as a nitpick:
- It's technically more restrictive than the generic type here
- It's divergent from what the `backoff` crate is normally doing, which IMO makes it slightly less clear.

common/src/backoff.rs
smklein: Done: d545fdbc100609473b882d2fca814f9b71e8d9f2

smklein:
I went ahead and added a "total duration" k/v pair to all warning cases, to make it more obvious "why this is a warning now"

jgallagher:
> I do suspect these timeouts are things we'll need to tweak to make this useful, but part of me thinks that this problem is unavoidable given that we're using a fallible retry scheme in the first place.

Yeah, I get this and agree.

> I went ahead and added a "total duration" k/v pair to all warning cases, to make it more obvious "why this is a warning now"

:+1: I like that.

jgallagher (approved):

common/src/backoff.rs
davepacheco: I also found this a little confusing.  What about adding some of what you wrote there in a comment on the new trait?

davepacheco (approved):
I could go either way on this and it's made me think about the point of severity in log files.  I tend to conflate logging severity with "is there a problem that an operator (or support) needs to know about".  I think that's not right though -- this is what metrics and alarms are for.  Eventually I imagine we will want metrics for, say, failed requests to Nexus, and then raise alarms on them, and those should definitely have thresholds like "if it's only been a few seconds, don't worry yet".  But severity in log files seems more aimed at someone debugging a problem and wanting to quickly find potentially problematic events.  In that sense, a message like "we failed to notify Nexus about this" _is_ noteworthy even if it's expected because Nexus isn't up yet.  (At least, while doing the DNS work, I debugged some issues that were in some sense expected behavior (DNS propagation failed because we couldn't read the config because Nexus had not yet populated the role assignments that would allow it to read the config) but where the resulting behavior was still surprising and so it was still helpful that these were called out as warning-level problems.)

Obviously I'm overthinking this though and if this improves quality of life, let's do it.

common/src/backoff.rs
jgallagher: After thinking about this more, I think the `Notify` trait in backoff is a poor style choice, and this function would be clearer if we did not emulate it.

Just looking at this function signature, neither (a) the fact that I can pass a closure directly nor (b) the args / return type of that closure are visible: I have to go look at `NotifyTracking` and then see the `impl<F: FnMut...>` defined for it. On the other hand, if we took an `FnMut` directly, both of those become obvious, and if I had a value / concrete type that I wanted to pass here, it's trivial to pass a 1-line closure that calls into the relevant handling method.

smklein: Fair enough, I hear the confusion, I'll switch it back. I guess the goal of aligning with the `backoff` implementation is not as important as clarity within our own codebase, and I buy that.

Updated in https://github.com/oxidecomputer/omicron/pull/2899/commits/cc8f37c7ba4ca91434dc1fb5098dec67bdd6da42 to use `FnMut` directly, plus impl traits to make it even more clear.

PR #2901, opened 2023-04-21
[sled-hardware] Minor refactor to improve clarity around naming

smklein:
- Rename "disk" to "partition"
- Move "disk" into its own module


sled-hardware/src/disk.rs
```diff
@@ -0,0 +1,397 @@
...
+    #[test]
+    fn test_partition_device_paths() {
+        const DEVFS_PATH: &'static str = "/devices/my/disk";
+        let paths =
+            DiskPaths { devfs_path: PathBuf::from(DEVFS_PATH), dev_path: None };
+
+        assert_eq!(
+            paths
+                .partition_device_path(
+                    &[Partition::ZfsPool,],
```
jgallagher, line 366: Trivial nit that existed before this PR:

```suggestion
                    &[Partition::ZfsPool],
```

jgallagher (approved):

sled-hardware/src/disk.rs
smklein: Done

PR #2903, opened 2023-04-21
Do not create a link local for tfpkt links

jmpesp:
Change `links_needed` to return whether or not a link needs a link local address in the zone. This was required for non-gimlet testing but broke gimlets - see: oxidecomputer/stlouis#391.

This also changes `links_needed` to check all service requests for required links, not just stopping at the first one found.

### DIFF
--- a/sled-agent/src/services.rs
+++ b/sled-agent/src/services.rs
@@ -549,14 +549,16 @@ impl ServiceManager {
     }
 
     // Check the services intended to run in the zone to determine whether any
-    // physical links or vnics need to be mapped into the zone when it is created.
-    //
-    // NOTE: This function is implemented to return the first link found, under
-    // the assumption that "at most one" would be necessary.
+    // physical links or vnics need to be mapped into the zone when it is
+    // created. Returns a list of links, plus whether or not they need link
+    // local addresses in the zone.
     fn links_needed(
         &self,
         req: &ServiceZoneRequest,
-    ) -> Result<Vec<Link>, Error> {
+    ) -> Result<Vec<(Link, bool)>, Error> {
+        let mut links: Vec<(Link, bool)> =
+            Vec::with_capacity(req.services.len());
+
         for svc in &req.services {
             match svc {
                 ServiceType::Nexus { .. }
@@ -572,7 +574,7 @@ impl ServiceManager {
                         .new_control(None)
                     {
                         Ok(n) => {
-                            return Ok(vec![n]);
+                            links.push((n, false));
                         }
                         Err(e) => {
                             return Err(Error::NexusVnicCreation(e));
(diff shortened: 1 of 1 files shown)


sled-agent/src/services.rs
```diff
@@ -549,14 +549,16 @@ impl ServiceManager {
...
+    // physical links or vnics need to be mapped into the zone when it is
+    // created. Returns a list of links, plus whether or not they need link
+    // local addresses in the zone.
     fn links_needed(
         &self,
         req: &ServiceZoneRequest,
-    ) -> Result<Vec<Link>, Error> {
+    ) -> Result<Vec<(Link, bool)>, Error> {
+        let mut links: Vec<(Link, bool)> =
+            Vec::with_capacity(req.services.len());
```
jgallagher, line 560: Tiny nit - I'd probably just make this `Vec::new()` because (a) it looks like `Nexus`/`Ntp`/`ExternalDns` may not need a link at all in the future and (b) `Maghemite` looks like it needs multiple links.

sled-agent/src/services.rs
```diff
@@ -779,17 +787,22 @@ impl ServiceManager {
...
-                "Ensuring {}/{} exists in zone",
-                link.name(),
-                IPV6_LINK_LOCAL_NAME
-            );
-            Zones::ensure_has_link_local_v6_address(
-                Some(running_zone.name()),
-                &AddrObject::new(link.name(), IPV6_LINK_LOCAL_NAME).unwrap(),
-            )?;
+        for (link, needs_link_local) in
+            running_zone.links().iter().zip(links_need_link_local)
```
jgallagher, line 791: On a lark I tried using `.filter_map()` here (since we're already doing `.iter().zip(..)`):

```diff
@@ -787,22 +787,21 @@ impl ServiceManager {

         let running_zone = RunningZone::boot(installed_zone).await?;

-        for (link, needs_link_local) in
-            running_zone.links().iter().zip(links_need_link_local)
+        for link in
+            running_zone.links().iter().zip(links_need_link_local).filter_map(
+                |(link, needs_link_local)| needs_link_local.then_some(link),
+            )
         {
-            if needs_link_local {
-                info!(
-                    self.inner.log,
-                    "Ensuring {}/{} exists in zone",
-                    link.name(),
-                    IPV6_LINK_LOCAL_NAME
-                );
-                Zones::ensure_has_link_local_v6_address(
-                    Some(running_zone.name()),
-                    &AddrObject::new(link.name(), IPV6_LINK_LOCAL_NAME)
-                        .unwrap(),
-                )?;
-            }
+            info!(
+                self.inner.log,
+                "Ensuring {}/{} exists in zone",
+                link.name(),
+                IPV6_LINK_LOCAL_NAME
+            );
+            Zones::ensure_has_link_local_v6_address(
+                Some(running_zone.name()),
+                &AddrObject::new(link.name(), IPV6_LINK_LOCAL_NAME).unwrap(),
+            )?;
         }
```

I don't feel like either version is clearly better; take your pick.

jgallagher (commented):
Changes look good to me. Building a ramdisk with this change now to test on `madrid`; will update here once that's done.

jgallagher (approved):
`madrid` is back to working with this change, thank you!

```
root@oxz_switch:~# ping -i gimlet14 fe80::aa40:25ff:fe04:345
fe80::aa40:25ff:fe04:345 is alive
```

sled-agent/src/services.rs
jmpesp: done in b8a8fd84

PR #2904, opened 2023-04-21
Update dendrite to remove need for `P4_DIR`

### LINKED ISSUE #2844: `cargo run --bin=omicron-dev -- run-all` fails from the root of the workspace
From a fresh clone:

```
dap@ivanova omicron-service-ports $ cargo run --bin=omicron-dev -- run-all
    Finished dev [unoptimized + debuginfo] target(s) in 0.68s
log file: /dangerzone/omicron_tmp/omicron-dev-omicron-dev.21201.0.log
note: configured to log to "/dangerzone/omicron_tmp/omicron-dev-omicron-dev.21201.0.log"
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: time out while discovering dendrite port number

Caused by:
    deadline has elapsed', /home/dap/omicron-service-ports/nexus/test-utils/src/lib.rs:146:68
Abort (core dumped)
```

It works from a subdirectory:

```
$ cargo run --bin=omicron-dev -- run-all
    Finished dev [unoptimized + debuginfo] target(s) in 0.69s
     Running `/home/dap/omicron-service-ports/target/debug/omicron-dev run-all`
log file: /dangerzone/omicron_tmp/omicron-dev-omicron-dev.21220.0.log
note: configured to log to "/dangerzone/omicron_tmp/omicron-dev-omicron-dev.21220.0.log"
omicron-dev: nexus internal API:    127.0.0.1:12221
omicron-dev: cockroachdb pid:       21224
omicron-dev: cockroachdb URL:       postgresql://root@127.0.0.1:54734/omicron?sslmode=disable
omicron-dev: cockroachdb directory: /dangerzone/omicron_tmp/.tmpQuUv4n
```

I think what's going on here is that the code that starts up Dendrite looks for data in a path relative to the current working directory:
https://github.com/oxidecomputer/omicron/blob/2882a12a69375a604b5210569eef6d7e1a78dc64/test-utils/src/dev/dendrite.rs#L44-L80

The relative path is correct if you're current working directory is one level below the workspace root.  That's true for integration tests that use this code.  It's not necessarily true for people run `omicron-dev run-all`.  (Side note: there's a test that runs`omicron-dev run-all`, and it doesn't hit this because it, like all tests, it runs from a subdirectory of the root of the workspace.)

I'm not exactly sure what the right answer is here.  I think it's a little brittle for code to depend on a path relative to the current working directory.  If this were a shipping tool, I'd suggest we ship a directory tree containing the executable and data and use the relative path from the executable to the data.  But this isn't something we ship.  Maybe the answer is to use the relative path from the executable to the data _in the workspace_?  But this is a little tricky because this is run from executables in different places relative to the workspace root.

Also, it would be nice if this code printed out a more useful message if `dpd` bails out early like this.  This was a little hard to debug because the stdout file that it reads appears to get cleaned up on error, so you don't have any of the output from `dpd`.

internet-diglett:
https://github.com/oxidecomputer/dendrite/pull/244 removes the need for setting the `P4_DIR` environment variable, which eliminates the relative path bug users have been experiencing when running `dpd` from various contexts.

Closes #2844


.github/workflows/rust.yml
```diff
@@ -32,6 +32,9 @@ jobs:
       if: ${{ github.ref != 'refs/heads/main' }}
     - name: Report cargo version
       run: cargo --version
+    - name: Install coreutils
+      run: brew install coreutils
+      if: matrix.os == 'macos-12'
```
davepacheco, line 37: I think this should probably go into install_builder_prerequisites.sh?  Part of the purpose of that is so that we're testing in CI exactly what developers following the usual build/test instructions would do.

internet-diglett: Hmm. I guess my primary hesitation is that on macos, this isn't the only way to ensure that the `sha256sum` command is installed, and running this command in `install_builder_prerequisites` could cause some real problems in some user environments because of how much homebrew does behind the scenes. What if a user prefers to manage their dependencies with `nix`, `asdf-vm`, or the like? How can we be sure they even have `brew` installed?

internet-diglett:
> If anyone currently exports P4_DIR, it looks like dpd will still pick that up. This PR then is for removing the spot in Omicron where P4_DIR is set, and letting dpd do infer_p4_dir?

Yes, that is correct.

.github/workflows/rust.yml
internet-diglett: Ah, I just saw that we are expecting users to have `brew` installed on macos environments:

```
levon@workstation-1:~/workspace/omicron$ rg brew
tools/install_builder_prerequisites.sh
153:    confirm "Install (or update) [${packages[*]}]?" && brew install "${packages[@]}"
```

Since that's the case, I can try moving this there.

davepacheco (approved):
I don't know about the dendrite parts but the changes here look good and work for me (to run `cargo run --bin=omicron-dev -- run-all` from the root)

bnaecker:
This seems to have moved Dendrite itself backwards in time:

```diff
 output.type = "zone"
 output.intermediate_only = true

@@ -345,8 +345,8 @@ only_for_targets.image = "standard"
bnaecker@feldspar : ~/omicron $ git show 00221b4221026e3a2e2c4c84bf1e6325ab992cca
commit 00221b4221026e3a2e2c4c84bf1e6325ab992cca
Author: Levon Tarver <11586085+internet-diglett@users.noreply.github.com>
Date:   Tue Apr 25 15:16:57 2023 -0500

    Update dendrite to remove need for `P4_DIR` (#2904)

    https://github.com/oxidecomputer/dendrite/pull/244 removes the need for
    setting the `P4_DIR` environment variable, which eliminates the relative
    path bug users have been experiencing when running `dpd` from various
    contexts.

    Closes #2844

diff --git a/package-manifest.toml b/package-manifest.toml
index 773977b30..cacfd17c1 100644
--- a/package-manifest.toml
+++ b/package-manifest.toml
@@ -300,8 +300,8 @@ only_for_targets.image = "standard"
 # 3. Use source.type = "manual" instead of "prebuilt"
 source.type = "prebuilt"
 source.repo = "dendrite"
-source.commit = "7d4ab83dcfe1c95cd01e091f92eb9664c96930bb"
-source.sha256 = "fe88ec54f426a8726deccbf9fd9aa50b733520284757110d36a1ad3879e25170"
+source.commit = "963bcd784e4385cbb45ecfba548ac4b1abb6edc1"
+source.sha256 = "be6b06a71e1c60fa60411787f5090d5f32a3b26d0f36a9bd2256f369d9621351"
 output.type = "zone"
 output.intermediate_only = true

@@ -326,8 +326,8 @@ only_for_targets.image = "standard"
 # 3. Use source.type = "manual" instead of "prebuilt"
 source.type = "prebuilt"
 source.repo = "dendrite"
-source.commit = "7d4ab83dcfe1c95cd01e091f92eb9664c96930bb"
-source.sha256 = "165c6dbf679db22e6f1c7cbff82c4f0a2c5df93469012283d6ecc944306eb790"
+source.commit = "963bcd784e4385cbb45ecfba548ac4b1abb6edc1"
+source.sha256 = "043c7ef19d315a43ecc50dc5456cadaeb1d2b9e99c0fe7f703ebc48ca888511c"
 output.type = "zone"
 output.intermediate_only = true

@@ -345,8 +345,8 @@ only_for_targets.image = "standard"
 # 3. Use source.type = "manual" instead of "prebuilt"
 source.type = "prebuilt"
 source.repo = "dendrite"
-source.commit = "7d4ab83dcfe1c95cd01e091f92eb9664c96930bb"
-source.sha256 = "294f22046a46f27fc1027090e980d8c7a9cb9ff48f63f92ac0a0554a9f46ef5b"
+source.commit = "963bcd784e4385cbb45ecfba548ac4b1abb6edc1"
+source.sha256 = "8ff8660f3ff0a59c19253a10b9725afd8c316ec55f6769811bfdb8d9a0bc549e"
```

Those commits are:

```
bnaecker@feldspar : ~/dendrite $ git log 7d4ab83 -1
commit 7d4ab83dcfe1c95cd01e091f92eb9664c96930bb
Author: bnaecker <ben@oxide.computer>
Date:   Tue Apr 18 13:31:41 2023 -0700

    Actually set the link MAC on creation (#246)
bnaecker@feldspar : ~/dendrite $ git log 963bcd7 -1
commit 963bcd784e4385cbb45ecfba548ac4b1abb6edc1
Author: Nils Nieuwejaar <nils@oxide.computer>
Date:   Tue Apr 18 13:14:01 2023 -0400

    special handling for non-CI p4 artifacts
```

I.e., we've moved to one that's ~15 minutes older. Part of the confusion here is that `963bcd` is a _PR_ commit, and so not in `main`'s history. I think we should enforce a policy of only bringing in commits here from `main`. If we need to test Omicron and Dendrite together with PR on a branch in Dendrite, then we can do that, but that commit should be rewritten to the one on `main` once it merges. I.e., test with PR commits; merge Dendrite into main; push a new Omicron commit with the updated SHA pointing to Dendrite main; wait for CI again; then merge.

bnaecker:
So the main issue here is not that the commit itself is old. It's that the base commit on `main` that the PR works from is old:

```
bnaecker@feldspar : ~/dendrite $ git log --graph --all --oneline 963bcd
<SNIP>
| * 963bcd7 special handling for non-CI p4 artifacts
| * 035341d cleanup
| * 52e9a89 Reduce the need for a P4_DIR environment variable.  This is still available as an override, but by default dpd will use the location of the dpd binary to locate the p4 artifacts.  In the dendrite_stub environment, we can now run the daemon directly - with no command-line arguments.
|/
* c4bc148 dependabot updates (#238)
* <SNIP>
bnaecker@feldspar : ~/dendrite $ git log -1 c4bc148
commit c4bc1481cf91eb26c08a843c54b1254fdcb09fcd
Author: Nils Nieuwejaar <nils@oxidecomputer.com>
Date:   Mon Apr 17 15:42:34 2023 -0400

    dependabot updates (#238)
```

PR #2905, opened 2023-04-22
Set failmode=continue on zpools in sled-agent.

### LINKED ISSUE #2766: sled-agent probably needs to set failmode on zpools
A lengthy debugging session on some unrelated bugs revealed the following essential behaviour with respect to an ordinary mirrored pool created by hand on a mirrored pair of SSDs:
```
19:26 wesolows | So to summarise: in the process of exporting the pool, which ::spa thinks completed just
               | dandy, we encountered a checksum error on a read.  There is no conceivable way that
               | retrying or waiting for anything is ever going to make that succeed, but nevertheless we're
               | sitting here waiting on some oracular operator to restore the proper contents of that
               | block, which is not readily identifiable nor are the contents present anywhere else we'd
               | have done this automatically already, so we come to rest with a mutex held.
19:30 wesolows | The punch line is that the checksum errors appeared spontaneously during a previous boot in
               | which all I was doing was importing and exporting the pool alternating between each of the
               | mirrored vdevs being unavailable.  After that wedged, I rebooted and now this.
 19:30      jmc | so we could at least confirm that, say, setting the mode to continue might be better
 19:31      jmc |
 19:31          |        failmode=wait|continue|panic
 19:31          |                Controls the system behavior in the event of catastrophic pool
 19:31          |                failure.  This condition is typically a result of a loss of
 19:31          |                connectivity to the underlying storage device(s) or a failure of
 19:31          |                all devices within the pool.  The behavior of such an event is
 19:31          |                determined as follows:
 19:31          |
 19:31          |                wait      Blocks all I/O access until the device connectivity is
 19:31          |                          recovered and the errors are cleared.  This is the
 19:31          |                          default behavior.
 19:31          |
 19:31          |                continue  Returns EIO to any new write I/O requests but allows
 19:31          |                          reads to any of the remaining healthy devices.  Any
 19:31          |                          write requests that have yet to be committed to disk
 19:31          |                          would be blocked.
 19:31          |
 19:31          |                panic     Prints out a message to the console and generates a
 19:31          |                          system crash dump.
```
In this particular case, it might at least sometimes be possible for this to succeed, but perhaps it has failed because we're in the export path.  Regardless, the pools Crucible uses, and presumably any others sled-agent is going to create, have only a single vdev.  They will never, ever be able to recovery from any kind of I/O failure by waiting for something to happen.  If retries didn't work, we're done.  Therefore, we should consider setting failmode to either continue or panic.

There is some possible complexity here: the intended use case for `failmode` is not at all how we got here in the case under investigation: the SSDs are present and electrically functional.  One thing we *don't* want to do is allow a situation in which the single vdev backing a pool has metadata corruption and we end up in a panic loop.  Ideally, this kind of condition will result in a faulted pool, and that diagnosis can be persisted across a reboot so that sled-agent will know that recovery should not be attempted (i.e., it needs to reinitialise the underlying device and have Crucible recover/reconstruct).  In the absence of that, I'm unsure whether `continue` is preferable to `panic`.  We will want to consider this also in light of *actual* loss of connectivity to the underlying device during surprise hotplug.

This is only for pools managed by sled-agent; the ramdisk-backed root pool is unrelated and has different considerations associated with it.

cc: @leftwo @ahl @jclulow @askfongjojo

faithanalog:
fixes #2766

Each pool is only backed by one vdev. There is no recovery if one starts breaking, so if connectivity to one dies it's actively harmful to try to wait for it to come back; we'll be waiting forever and get stuck. We'd rather get the errors so we can deal with them ourselves.

This change unconditionally sets failmode=continue on zpools in sled-hardware's Disk::new(). That's the codepath all zpools go through when they're created or imported. This won't apply to virtual hardware because those go through the SyntheticDisk codepath, so if anyone cares about failmode=continue on virtual hardware we should modify `create_virtual_hardware.sh`.

I've tested this on dogfood cubby 20.

Before:
```
root@BRM42220014:/tmp/artemis# zpool get failmode                                                             
NAME                                      PROPERTY  VALUE     SOURCE                                                                                           
oxi_55d7e113-05f1-4af7-acc7-2205d016aab2  failmode  wait      default                                                                                          
oxi_9058a1f2-99cb-40ad-8fcd-55adf8cfb127  failmode  wait      default                                                                                          
oxp_1b125d25-1e86-4d9e-8d6a-8157b4775cf6  failmode  wait      default                                                                                          
oxp_2f00ff04-dfee-4c5d-bc66-91266a0fc7a6  failmode  wait      default                                                                                          
oxp_30b34c2a-901a-4582-8e8a-a2b0368bd868  failmode  wait      default                                                                                          
oxp_4d750b62-760b-43b4-9d6e-a4b02c893673  failmode  wait      default                                                                                          
oxp_6a03c94a-48bc-45e7-8f4c-bed46e897027  failmode  wait      default                                                                                          
oxp_70c4b238-cc1b-478f-a818-d15ec09dcb5d  failmode  wait      default                                                                                          
oxp_7a7d6fef-27f1-47de-b05f-e48f1555c81d  failmode  wait      default
oxp_7c3e8a93-6660-4632-9db4-8a09bb6bffff  failmode  wait      default
oxp_d8b6a101-bd4f-4921-9723-fd4707fde84f  failmode  wait      default          
oxp_fc53d384-6eb5-4754-b4e9-7bce5474e58c  failmode  wait      default
rpool                                     failmode  wait      default
```

After:
```
root@BRM42220014:/tmp/artemis# zpool get failmode                                                                                                             
NAME                                      PROPERTY  VALUE     SOURCE
oxi_55d7e113-05f1-4af7-acc7-2205d016aab2  failmode  continue  local
oxi_9058a1f2-99cb-40ad-8fcd-55adf8cfb127  failmode  continue  local
oxp_1b125d25-1e86-4d9e-8d6a-8157b4775cf6  failmode  continue  local
oxp_2f00ff04-dfee-4c5d-bc66-91266a0fc7a6  failmode  continue  local
oxp_30b34c2a-901a-4582-8e8a-a2b0368bd868  failmode  continue  local
oxp_4d750b62-760b-43b4-9d6e-a4b02c893673  failmode  continue  local
oxp_6a03c94a-48bc-45e7-8f4c-bed46e897027  failmode  continue  local
oxp_70c4b238-cc1b-478f-a818-d15ec09dcb5d  failmode  continue  local
oxp_7a7d6fef-27f1-47de-b05f-e48f1555c81d  failmode  continue  local
oxp_7c3e8a93-6660-4632-9db4-8a09bb6bffff  failmode  continue  local
oxp_d8b6a101-bd4f-4921-9723-fd4707fde84f  failmode  continue  local
oxp_fc53d384-6eb5-4754-b4e9-7bce5474e58c  failmode  continue  local
rpool                                     failmode  wait      default
```

### DIFF
--- a/illumos-utils/src/zpool.rs
+++ b/illumos-utils/src/zpool.rs
@@ -213,6 +213,19 @@ impl Zpool {
         }
     }
 
+    /// zpool set failmode=continue <name>
+    pub fn set_failmode_continue(name: &ZpoolName) -> Result<(), Error> {
+        let mut cmd = std::process::Command::new(PFEXEC);
+        cmd.env_clear();
+        cmd.env("LC_ALL", "C.UTF-8");
+        cmd.arg(ZPOOL)
+            .arg("set")
+            .arg("failmode=continue")
+            .arg(&name.to_string());
+        execute(&mut cmd).map_err(Error::from)?;
+        Ok(())
+    }
+
     pub fn list() -> Result<Vec<ZpoolName>, ListError> {
         let mut command = std::process::Command::new(ZPOOL);
         let cmd = command.args(&["list", "-Hpo", "name"]);
--- a/sled-hardware/src/disk.rs
+++ b/sled-hardware/src/disk.rs
@@ -259,6 +259,7 @@ impl Disk {
         zpool_name: &ZpoolName,
     ) -> Result<(), DiskError> {
         Self::ensure_zpool_imported(log, &zpool_name)?;
+        Self::ensure_zpool_failmode_is_continue(log, &zpool_name)?;
         Self::ensure_zpool_has_datasets(&zpool_name)?;
         Ok(())
     }
@@ -310,6 +311,27 @@ impl Disk {
     ) -> Result<(), DiskError> {
         Zpool::import(zpool_name.clone()).map_err(|e| {
             warn!(log, "Failed to import zpool {zpool_name}: {e}");
+            DiskError::ZpoolImport(e)
+        })?;
+        Ok(())
+    }
+
+    fn ensure_zpool_failmode_is_continue(
+        log: &Logger,
+        zpool_name: &ZpoolName,
+    ) -> Result<(), DiskError> {
+        // Ensure failmode is set to `continue`. See
+        // https://github.com/oxidecomputer/omicron/issues/2766 for details. The
+        // short version is, each pool is only backed by one vdev. There is no
+        // recovery if one starts breaking, so if connectivity to one dies it's
+        // actively harmful to try to wait for it to come back; we'll be waiting
+        // forever and get stuck. We'd rather get the errors so we can deal with
+        // them ourselves.
+        Zpool::set_failmode_continue(&zpool_name).map_err(|e| {
+            warn!(
+                log,
+                "Failed to set failmode=continue on zpool {zpool_name}: {e}"
+            );
             DiskError::ZpoolImport(e)
         })?;
         Ok(())


smklein (approved):
Looks great, thanks for fixing this!

Testing things out manually, this also appears idempotent, so the "just set the value", rather than attempting to "read-modify-write" seems like the right call.

faithanalog:
oops i deleted... the change lol. love to rebase

smklein (approved):
Still LGTM!

PR #2916, opened 2023-04-24
remove internal references from external API

ahl:
This PR intends to clean up a bunch of the external API in these regards:
- removing rustdoc-link references that don't make sense in the external docs
- removing internally focused docs (@bnaecker can you take a looksee?)
- moving saga-related endpoints to the nexus-internal API?

I need to make another pass over this for consistency and will move it out of draft once I've done that.


common/src/api/external/mod.rs
```diff
@@ -490,21 +488,59 @@ impl JsonSchema for RoleName {
...
+/// inconvenient to define constructors: a u32 constructor can be infallible,
+/// but an i64 constructor can fail (if the value is negative) and a u64
+/// constructor can fail (if the value is larger than i64::MAX).  We provide
+/// all of these for consumers' convenience.
+///
+/// The maximum byte count of i64::MAX comes from the fact that this is stored
+/// in the database as an i64.  Constraining it here ensures that we can't fail
+/// to serialize the value.
+///
+/// TODO-cleanup This could benefit from a more complete implementation.
```
smklein, line 503: This documentation seems inaccurate next to the `JsonSchema` implementation below?

common/src/api/internal/mod.rs
```diff
@@ -5,3 +5,93 @@
...
+        .map(|maybe_object| maybe_object.unwrap().into())
+        .collect::<Vec<U>>()
+        .await
+}
+
+// Sagas
+//
+// These are currently only intended for observability by developers.  We will
+// eventually want to flesh this out into something more observable for end
+// users.
```
smklein, line 32: Nit: This could be a rustdoc comment - we're probably okay documenting this for internal usage, since it's not part of the external API anymore.

openapi/nexus.json
```diff
@@ -7306,13 +7211,14 @@
...
       },
       "ByteCount": {
-        "description": "A count of bytes, typically used either for memory or storage capacity\n\nThe maximum supported byte count is [`i64::MAX`].  This makes it somewhat inconvenient to define constructors: a u32 constructor can be infallible, but an i64 constructor can fail (if the value is negative) and a u64 constructor can fail (if the value is larger than i64::MAX).  We provide all of these for consumers' convenience.",
+        "title": "Count of bytes",
+        "description": "Used to measure memory or disk capacity",
         "type": "integer",
-        "format": "uint64",
-        "minimum": 0
+        "minimum": 0,
+        "maximum": 9223372036854774784
```
smklein, line 7218: Should we test this?

smklein (approved):
Looks good, though you might want to re-run the openapi tests with EXPECTORATE

sled-agent/src/sim/server.rs
```diff
@@ -100,7 +100,7 @@ impl Server {
                             .hardware
                             .hardware_threads,
                         usable_physical_ram: NexusTypes::ByteCount::try_from(
-                            config.hardware.physical_ram,
+                            config.hardware.physical_ram as i64,
```
davepacheco, line 103: The `as i64` around ByteCount seems unfortunate.  (not saying it's a blocker.)  Part of the point of that type was to make it hard to accidentally confuse bytes for mebibytes or megabytes or whatever by having explicit constructors and getters.  If we're just throwing an i64 into a constructor, it seems like we're losing that (and making it easy to introduce that kind of problem).

I don't know what the answer is.  Could/should this be one of those types that we actually share across client/server, as a sort of primitive?

common/src/api/external/mod.rs
```diff
@@ -919,7 +944,7 @@ impl From<crate::api::internal::nexus::InstanceRuntimeState>
     }
 }
 
-/// Client view of an [`Instance`]
+/// View of an Instance
```
davepacheco, line 947: Would this be a good place for us to (eventually) put the user-facing overview of each resource?  e.g., https://stripe.com/docs/api/products (though I was imagining a fair bit more detail).

GCP has this sort of thing at https://cloud.google.com/compute/docs/instances, though worth noting that that the REST resource itself is really more of a stub that _points_ to that guide: https://cloud.google.com/compute/docs/reference/rest/v1/instances

benjaminleonard: We were discussing that earlier this week. It's something I'd like to have on the docs site.
We could do something with the API spec, with AsciiDoc, or both.

I'm sure I could grab the resource object from its schema ref, e.g. "#/components/schemas/Instance".

ahl: Yes. That makes a ton of sense.

common/src/api/external/mod.rs
ahl: I'm going to pull this out and do it in another PR--too many snags.

oximeter/oximeter/src/histogram.rs
```diff
@@ -191,51 +191,51 @@ pub struct Bin<T> {
...
-///
-/// The short of this is that, most of the time, it shouldn't matter. If one specifies the extremes
-/// of the support as their bins, be aware that the left-most may be converted from a
-/// `BinRange::RangeTo` into a `BinRange::Range`. In other words, the first bin of a histogram is
-/// _always_ a `Bin::Range` or a `Bin::RangeFrom` after construction. In fact, every bin is one of
-/// those variants, the `BinRange::RangeTo` is only provided as a convenience during construction.
+//
+// Example
+// -------
+// ```rust
```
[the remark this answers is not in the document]
ahl: Yes... but a subset of the unit tests at the bottom

Issue #2888, opened 2023-04-20
Sled Agent must manage durable storage for configs, zones, explicitly


smklein:
See also: RFD 118

Sled Agent currently configures a few pieces of data outside datasets explicitly allocated in pools:

- `/var/oxide` contains a variety of configuration information, including...
  - RSS setup information
  - The "Sled Request" used to launch the sled-agent (including underlay information)
  - A list of "All services which should be launched on this sled"
  - A list of "All services which should be launched on u.2 zpools"
- `/zone` contains the "filesystem for all zones"
- `/opt/oxide` contains all the "latest installed system images", and is used to update control plane software which exists outside the ramdisk.

Q: So, why is this bad?
A: All those paths are currently backed by a ramdisk -- specifically `rpool` -- on gimlets.

This means that when we reboot, a significant portion of the necessary configuration information to launch the sled will be lost. Furthermore, for the `zonepath` filesystems, a significant portion of user RAM will be dedicated to zone-based filesystems, which we'd prefer to distribute to disk-backed file storage.

Here's a list of some of the work we need to accomplish to mitigate this in a production environment:

- [ ] Sled agent should be responsible for formatting U.2s and M.2s with necessary datasets
  - [ ] See: https://github.com/oxidecomputer/omicron/issues/2688
- [x] #2998
- [x] #2970
- [x] #2969
  - [x] This would be made easier with https://github.com/oxidecomputer/omicron/issues/1898 , as we'd only need to store the zone information, and not the corresponding config for "how to launch the zone".
- [ ] Sled agent should allocate `zonepath`s from U.2s
  - [x] This means sled agent needs to handle the allocation / de-allocation of zonepaths much more explicitly.
  - [ ] Related: https://github.com/oxidecomputer/omicron/issues/1119 , https://github.com/oxidecomputer/omicron/issues/1313
- [x] #2971

andrewjstone:
I just remembered that https://github.com/oxidecomputer/omicron/pull/3007 adds some support for persisting zone filesystems under crypt/zone.

This code has been ready to go for a few weeks now and has been tested numerous times. I've done cold boot testing on paris with it, but since there is no persistent install of software on the M.2s like dogfood, I have to re-run omicron-package install after reboot to rediscover and mount the U.2 encrypted datasets. The PR has the details about this. While I'd like to test with a proper install on dogfood I"m at about the point where I'd like to just merge this and see what happens.

PR #2919, opened 2023-04-24
[sled-agent] Conform datasets to RFD 118 format, but still use ramdisks

smklein:
This PR is split off of https://github.com/oxidecomputer/omicron/pull/2902 , which also tries to *use* those datasets,
and attempts to make a subset of that change with a smaller diff.

This PR:
- Formats U.2 and M.2s with expected datasets
- Adds `/pool/ext/<UUID>/zone`, and places storage-based zone filesystems there
- Adds "oxi_" internal zpools to the "virtual hardware" scripts, emulating M.2s.

Part of https://github.com/oxidecomputer/omicron/issues/2888


illumos-utils/src/running_zone.rs
```diff
@@ -440,6 +440,9 @@ pub enum InstallZoneError {
 pub struct InstalledZone {
     log: Logger,
 
+    // Filesystem path of the zone
+    zonepath: PathBuf,
+
```
smklein, line 445: Within this PR, this is always `/zone`, which resides in the ramdisk. However, this will change in the future, as some zones move to:

```
/pool/ext/<UUID>/zone
```

illumos-utils/src/zfs.rs
```diff
@@ -8,8 +8,8 @@ use crate::{execute, PFEXEC};
 use std::fmt;
 use std::path::PathBuf;
 
-pub const ZONE_ZFS_DATASET_MOUNTPOINT: &str = "/zone";
-pub const ZONE_ZFS_DATASET: &str = "rpool/zone";
+pub const ZONE_ZFS_RAMDISK_DATASET_MOUNTPOINT: &str = "/zone";
+pub const ZONE_ZFS_RAMDISK_DATASET: &str = "rpool/zone";
```
smklein, line 12: I went ahead and renamed these constants to make it a little more clear that "anything you put here is on a ramdisk, and will not last across reboots!"

illumos-utils/src/zfs.rs
```diff
@@ -137,15 +137,15 @@ impl Zfs {
     }
 
     /// Creates a new ZFS filesystem named `name`, unless one already exists.
-    pub fn ensure_zoned_filesystem(
+    pub fn ensure_filesystem(
```
smklein, line 140: When provisioning filesystems in `sled-hardware/src/disk.rs` this doesn't want `zoned`, so it's configurable.

This function was also renamed to make the intent a bit more clear.

illumos-utils/src/zfs.rs
```diff
@@ -254,19 +249,21 @@ pub fn get_all_omicron_datasets_for_delete() -> anyhow::Result<Vec<String>> {
     // This includes cockroachdb, clickhouse, and crucible datasets.
     let zpools = crate::zpool::Zpool::list()?;
     for pool in &zpools {
-        // For now, avoid erasing any datasets which exist on internal zpools.
-        if pool.kind() == crate::zpool::ZpoolKind::Internal {
-            continue;
-        }
+        let internal = pool.kind() == crate::zpool::ZpoolKind::Internal;
```
smklein, line 252: Although this PR doesn't actually migrate any config data out of the ramdisk, eventually we'll move things from `/var/oxide` into `/pool/int/<UUID>/config` on the M.2s.

As a result, we *will* want to destroy internal datasets to "uninstall".

sled-agent/src/bootstrap/hardware.rs
```diff
@@ -73,6 +78,12 @@ impl HardwareMonitorWorker {
                                     warn!(self.log, "Failed to deactivate switch: {e}");
                                 }
                             }
+                            sled_hardware::HardwareUpdate::DiskAdded(disk) => {
+                                self.storage.upsert_disk(disk).await;
+                            }
+                            sled_hardware::HardwareUpdate::DiskRemoved(disk) => {
+                                self.storage.delete_disk(disk).await;
+                            }
```
smklein, line 86: The bootstrap agent now uses a `StorageManager`, so it can create the necessary datasets on the M.2s before the sled agents launch. This will be *especially* important as we move the sled agent's configuration data out of `/var/oxide` -- these datasets must exist before the sled agent is created by RSS.

sled-agent/src/storage_manager.rs
```diff
@@ -800,7 +863,12 @@ impl StorageWorker {
         disk: UnparsedDisk,
     ) -> Result<(), Error> {
         info!(self.log, "Deleting disk: {disk:?}");
-
+        // TODO: Don't we need to do some accounting, e.g. for all the information
+        // that's no longer accessible? Or is that up to Nexus to figure out at
+        // a later point-in-time?
+        //
+        // If we're storing zone images on the M.2s for internal services, how
+        // do we reconcile them?
```
smklein, line 871: We definitely needed to do this before this PR, but I'm labeling it. This will be more and more relevant as we move configs from the ramdisks onto U.2s.

sled-agent/src/storage_manager.rs
```diff
@@ -1106,15 +1179,15 @@ pub struct StorageManager {
     task: JoinHandle<Result<(), Error>>,
 }
 
+/// A sled-local view of all attached storage.
+#[derive(Clone)]
+pub struct StorageManager {
+    inner: Arc<StorageManagerInner>,
```
smklein, line 1185: This is now `Arc` wrapped so that the `ServiceManager` can also reference it.

In the future, the `ServiceManager` will be asking the `StorageManager` questions like:
- "I have a zone I want to install - what U.2s are available for me to use for the backing filesystem?"

sled-hardware/src/disk.rs
```diff
@@ -179,18 +183,93 @@ pub struct Disk {
...
+];
+
+const M2_EXPECTED_DATASET_COUNT: usize = 5;
+static M2_EXPECTED_DATASETS: [&'static str; M2_EXPECTED_DATASET_COUNT] = [
+    // Stores a "factory install" set of software
+    FACTORY_DATASET,
+    // Stores software images.
+    //
+    // Should be duplicated to both M.2s.
+    INSTALL_DATASET,
```
smklein, line 206: This gives us a landing spot for the update process - we should be placing non-ramdisk control plane zones within

```
/pool/int/<UUID>/install
```

On each of the M.2s

sled-agent/src/services.rs
```diff
@@ -212,16 +206,31 @@ impl Config {
...
-                PathBuf::from(ZONE_ZFS_DATASET_MOUNTPOINT)
-                    .join(PathBuf::from(zone_name))
-                    .join("root")
-                    .join(format!("var/svc/manifest/site/{}", svc_name))
-            }),
         }
     }
 }
 
+// A wrapper around `ZoneRequest`, which allows it to be serialized
```
andrewjstone, line 213: Is this configuration that is only used during development? Is it something that goes away with self-assembling zones or is this actually part of making self-assembling zones work?

sled-agent/src/storage_manager.rs
andrewjstone: I guess this is part of your thoughts about merging the two above. As it stands I also find it unfortunate to share an inner resource of one manager with that of another. I'd love to collaborate on how we can separate the abstractions entirely or merge them, although I admit a preference for separation at this point.

andrewjstone (approved):
I have a few comments about abstraction, but nothing you don't already know. Other than that this is a good step forward. Let's merge it.

smklein:
Thanks for the review! One last thing that I need to figure out before merging, and which (I believe?) is breaking the `deploy` job -- this PR is perhaps a bit too excited about using the "dataset mountpoint" for zones (e.g., `/pool/ext/<UUID>/zone`). For non-gimlet systems using synthetic disks, this mountpoint is not actually being created, which is a problem.

andrewjstone:
> Thanks for the review! One last thing that I need to figure out before merging, and which (I believe?) is breaking the `deploy` job -- this PR is perhaps a bit too excited about using the "dataset mountpoint" for zones (e.g., `/pool/ext/<UUID>/zone`). For non-gimlet systems using synthetic disks, this mountpoint is not actually being created, which is a problem.

That makes sense. You are welcome!

sled-agent/src/storage_manager.rs
smklein: If it helps, the dependency is "real", in a sense -- provisioning services does (and will increasingly) depend on access to storage as we migrate the location of zone filesystems. Sharing this reference makes the sled agent coherent.

I think the unfortunate part is that both the `StorageManager` and `ServiceManager` are both in the business of provisioning services. If anything, the service management side of `StorageManager` -- which currently handles the CRDB, Clickhouse, and Crucible zones -- could be migrated out.

sled-agent/src/services.rs
smklein: This is unfortunately not a development-only piece of metadata.

Here's my primer on "old vs new" zone management:

## Old

- All zone filesystems are provisioned in `rpool/zone` (the ramdisk)
- Zones are **not** self-assembling, meaning that they need tweaking from the sled agent after booting
- The list of zones under the purview of the sled are stored in a config file in `/var/oxide` (... unfortunately, this is also in the ramdisk)

## New (not totally done in this PR, but we're heading this direction)

- Zone filesystems are provisioned in `/pool/ext/<UUID>/zone` on particular U.2s
- Zones are self-assembling, meaning that once they are created, they can be booted
  - NOTE: The sled agent may still need to supply some metadata prior to running the zone, such as creating VNICs!
- The list of zones under the purview of the sled agent are stored in `/pool/int/<UUID>/config` on both M.2s


## So this metadata...

Yeah, so this piece of metadata is here because we're breaking the previous standard that "every zone filesystem lives in `/zone`". By shuffling that around, the sled agent actually needs to know where to find zones to launch them.

I admittedly **could** store this information alongside the zone filesystem on the U.2, but by storing it within the `config` directory of the M.2, I figured we'd have a better ability for sleds to eventually notice / report "hey, I **should** launch a particular zone, but it hasn't come up yet because that disk is gone".

smklein:
@faithanalog - I'm adding you as an FYI here - if you're busy, you can skim through this, but figured you might be interested since you went through some of this code recently!

PR #2922, opened 2023-04-24
Omicron services need more unique ids

### LINKED ISSUE #2813: service inserts clobber other services for the same zone
#2796 introduced a case where we have multiple services in the same zone (InternalDNS and InternalDNSConfig).  However, we use the zone uuid as the service_id, that's a unique column, and we ignore conflicts on it when inserting services in at least some cases.  The result is that we clobber one of these records with the other one and we can't store both of them.

I could see:

- making a new uuid instead of using the zone id here (and maybe adding zone_id as a different column?)
- continuing to use the zone id here but using `(service_id, kind)` as the primary key

Reference:
https://github.com/oxidecomputer/omicron/blob/5d39727039cca86a7fef48d35f3675187386c333/nexus/db-queries/src/db/datastore/rack.rs#L181-L193

(Thanks to @luqmana for [spotting this](https://github.com/oxidecomputer/omicron/pull/2796#discussion_r1163343988).) 

This definitely needs to be fixed, but I believe it's not urgent, even though it's sort of a regression, because the only thing that can get clobbered like this today is the `InternalDNS` record, and we don't currently use it.

davepacheco:
See #2813 for more context.  Today, services have one id, the `service_id`, which is unique.  We generally use the zone id for this.  That means that if we wind up registering two logical services in one zone (as we do for things like Internal DNS), we cannot store both records.  In this change I'm trying to balance:

- it's very useful for debugging if the zone id matches the service id (it's one less mapping to write down and keep in your head)
- we might want more than one service in some zones

What I did was:

- There's a new `zone_id` column in the `service` table.  It's nullable because we might have some services not in standard control plane zones (e.g., in the GZ).  This column is really only intended for debugging.
- The service id is now expected to be unique.  If we have two services in a zone, they can (and must, and do, in this PR) have separate ids.
- For most zones, I kept the behavior that the `service_id` matches the `zone_id`, again for debugging.

I could see this last part being the worst of both worlds.  It'd be a problem if we accidentally depended on the fact that service_id == zone_id, which is true in most cases.  Thoughts?

Fixes #2813.


sled-agent/src/rack_setup/service.rs
```diff
@@ -689,14 +693,16 @@ impl ServiceInner {
                             dns_address,
                         } => {
                             services.push(NexusTypes::ServicePutRequest {
-                                service_id: zone.id,
+                                service_id: Uuid::new_v4(),
```
luqmana, line 696: I feel like this Uuid generation shouldn't happen here (`handoff_to_nexus`) but somewhere earlier as part of creating the allocation plan. Otherwise, next boot we'll see the previously completed plan and still end up here but end up passing a slightly different `ServicePutRequest` to Nexus.

davepacheco: Good catch!  I've fixed this in 901c04a.  Let me know what you think.  I basically replaced the list of `ServiceKind` with a list of a new struct containing both the `ServiceKind` and a unique id.  On the one hand, this feels overly complex.  But I don't see a way around it without more substantial refactoring.

One surprising consequence of this is that there are places _inside_ the Sled Agent where we pass around a `ServicePutRequest` that was manufactured inside the Sled Agent (e.g., `activate_switch()`).  These structs are changed in this PR and now get these extra ids.  That's even though as far as I can tell, these never wind up registered in Nexus.  I think this is an artifact of using the same struct to represent "a request to provision a zone with one or more services in it" for two different purposes: one is to report it to Nexus as a `service`, and the other is to actually cause that zone to be provisioned.  The former requires these extra ids.  The latter doesn't, but winds up getting them.

sled-agent/src/params.rs
```diff
@@ -485,6 +485,20 @@ impl From<ServiceZoneRequest> for sled_agent_client::types::ServiceZoneRequest {
     }
 }
 
+#[derive(
+    Clone, Debug, Deserialize, Serialize, JsonSchema, PartialEq, Eq, Hash,
+)]
+pub struct ServiceZoneService {
```
smklein, line 491: I can probably do another pass on the naming here to collapse "Service" out of this in a separate PR, but the addition of this structure makes sense to me!

davepacheco: Thanks for taking a look!

luqmana (approved):
Thanks for getting that fixed!

PR #2926, opened 2023-04-25
[installinator] Improving networking situation

jgallagher:
This PR includes several small changes that aren't particularly closely related, other than they all affect `installinator`'s ability to talk to `wicketd`. Even after this PR that isn't completely working (currently bumping into https://github.com/oxidecomputer/maghemite/issues/57), but I believe all of these are required even if there ends up being more to do.

It may make sense to review the commits individually instead of the PR as a whole:

* 4fcbf9c76efb659b2e304cdf2efde9cd4a66fd5d - tiny, makes `installinator` less spammy when it fails to find `wicketd`
* aa2809235a75b4b77dc8ba0a83460450187ec53a - refactor to move some bootstrap network setup things out of `sled-agent` and into `sled-hardware`
* 9e09f897955e9d16dbfc4f782249393770e95eb6 - installinator sets up a bootstrap network interface
* cdeec3894e591a7adbe2712e72f8ea05d0ba4234 - tiny TODO fix for wicketd's listening port
* c08181b583965a226a8e938453387c50da5a6aad - off-by-one error fix in configuring `mg-ddm` within the switch zone


installinator/src/peers.rs
```diff
@@ -124,6 +122,10 @@ impl FetchedArtifact {
         F: FnMut() -> Fut,
         Fut: Future<Output = Result<Peers, DiscoverPeersError>>,
     {
+        // How long to sleep between retries if we fail to find a peer or fail
+        // to fetch an artifact from a found peer.
+        const RETRY_DELAY: Duration = Duration::from_secs(1);
```
sunshowers, line 127: Let's make this even larger -- maybe 2-5 seconds I think?

andrewjstone:
Looks good besides Rain's comments.

installinator/src/peers.rs
jgallagher: Whoops, we chatted about this and I forgot to change it. Fixed in af124505

installinator/src/bootstrap.rs
```diff
@@ -10,7 +10,15 @@ use anyhow::ensure;
 use anyhow::Context;
 use anyhow::Result;
 use illumos_utils::addrobj::AddrObject;
+use illumos_utils::dladm;
+use illumos_utils::dladm::Dladm;
+use illumos_utils::zone::Zones;
+use omicron_common::address::Ipv6Subnet;
+// TODO Maybe `DdmAdminClient` should be in a separate crate so we don't have to
+// depend on sled-agent?
+use omicron_sled_agent::bootstrap::ddm_admin_client::DdmAdminClient;
```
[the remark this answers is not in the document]
jgallagher: At that point sled-agent and installinator would essentially have a copy/paste `DdmAdminClient` wrapper, so in cdb375cda13787a59ce630df5b59c4755cba5867 I moved that wrapper out to the `ddm-admin-client` crate proper, which both sled-agent and installinator can now use directly.

smklein (approved):
I'm always on-board with refactors to make the sled agent core slightly smaller

PR #2936, opened 2023-04-27
[wicketd] move engine spec to a separate wicket-common crate

sunshowers:
We'd like to use the `EventBuffer` defined in #2890 to track the update
status for display. To acheive that, we need to use the update-engine
types rather than the progenitor-generated ones.

Depends on #2830.


wicketd-client/src/lib.rs
```diff
@@ -33,20 +33,16 @@ progenitor::generate_api!(
...
         RotSlot = { derives = [ PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize]},
         ImageVersion = { derives = [ PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize]},
-        UpdateComponent = { derives = [ Copy, PartialEq, Eq, PartialOrd, Ord ] },
     },
     replace = {
         Duration = std::time::Duration,
-        ProgressEventForGenericSpec = update_engine::events::ProgressEvent<update_engine::NestedSpec>,
+        EventReportForWicketdEngineSpec = wicket_common::update_events::EventReport,
+        StepEventForWicketdEngineSpec = wicket_common::update_events::StepEvent,
+        ProgressEventForWicketdEngineSpec = wicket_common::update_events::ProgressEvent,
```
jgallagher, line 41: I'll chalk this up to general unease about `replace` from the conversations you and Adam had previously, but: what are we gaining from these replacements? I don't see any method calls on the replaced types, for example, but maybe I'm missing it?

sunshowers: Not at the moment, but I have work in progress to use `EventBuffer` within wicket, which requires that these are the wicket types.

I'll ask for a re-review after ensuring that that direction holds.

sunshowers:
The code in #2980 (which I'll be splitting up into several PRs, e.g. #3004, #3017 etc) uses the update engine types extensively, so moving this out of draft status.

jgallagher (approved):

PR #2937, opened 2023-04-27
Small `nexus.json` copy tweaks

benjaminleonard:
Small copy tweaks on copy for the docs site.
I'll update `nexus.json` as soon as I get omicron compiling.

### DIFF
--- a/nexus/src/external_api/http_entrypoints.rs
+++ b/nexus/src/external_api/http_entrypoints.rs
@@ -657,7 +657,7 @@ async fn silo_user_view(
 
 // Silo identity providers
 
-/// List a silo's IDPs_name
+/// List a silo's IdP's name
 #[endpoint {
     method = GET,
     path = "/v1/system/identity-providers",
@@ -694,7 +694,7 @@ async fn silo_identity_provider_list(
 
 // Silo SAML identity providers
 
-/// Create a SAML IDP
+/// Create a SAML IdP
 #[endpoint {
     method = POST,
     path = "/v1/system/identity-providers/saml",
@@ -723,7 +723,7 @@ async fn saml_identity_provider_create(
     apictx.external_latencies.instrument_dropshot_handler(&rqctx, handler).await
 }
 
-/// Fetch a SAML IDP
+/// Fetch a SAML IdP
 #[endpoint {
     method = GET,
     path = "/v1/system/identity-providers/saml/{provider}",
@@ -1170,7 +1170,7 @@ async fn ip_pool_update(
     apictx.external_latencies.instrument_dropshot_handler(&rqctx, handler).await
 }
 
-/// Fetch the IP pool used for Oxide services.
+/// Fetch the IP pool used for Oxide services
 #[endpoint {
     method = GET,
     path = "/v1/system/ip-pools-service",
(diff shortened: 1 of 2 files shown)


nexus/src/external_api/http_entrypoints.rs
```diff
@@ -694,7 +694,7 @@ async fn silo_identity_provider_list(
 
 // Silo SAML identity providers
 
-/// Create a SAML IDP
+/// Create a SAML IdP
```
benjaminleonard, line 697: This is the version of the acronym I see most often

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -2550,6 +2550,8 @@ async fn image_delete(
     apictx.external_latencies.instrument_dropshot_handler(&rqctx, handler).await
 }
 
+/// Promote a project image
+///
 /// Promote a project image to be visible to all projects in the silo
```
david-crespo, line 2555: It looks weird when the paragraph copy starts exactly the same as the title. Maybe we could change this to `Make a project image visible to all projects in the silo`

david-crespo (approved):
LGTM, one suggestion

ahl:
Can you hold off on this until I've landed #2916 please?

nexus/src/external_api/http_entrypoints.rs
benjaminleonard: I suppose so — but there's a few like that, if we change this we should do the others.

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -1188,7 +1188,7 @@ type IpPoolRangePaginationParams = PaginationParams<EmptyScanParams, IpNetwork>;
 
 /// List ranges for an IP pool
 ///
-/// Ranges are ordered by their first address.
+/// List ranges for an IP pool. Ranges are ordered by their first address.
```
zephraph, line 1191: Is this first part needed? It's already captured above.

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -1280,9 +1280,10 @@ async fn ip_pool_range_remove(
     apictx.external_latencies.instrument_dropshot_handler(&rqctx, handler).await
 }
 
-/// List ranges for the IP pool used for Oxide services.
+/// List ranges for the IP pool used for Oxide services
 ///
-/// Ranges are ordered by their first address.
+/// List ranges for the IP pool used for Oxide services. Ranges are ordered by
+/// their first address.
```
zephraph, line 1286: This is another one where the title is duplicated

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -1542,6 +1543,8 @@ async fn disk_metrics_list(
     apictx.external_latencies.instrument_dropshot_handler(&rqctx, handler).await
 }
 
+/// Start importing blocks into a disk
+///
 /// Start the process of importing blocks into a disk
```
zephraph, line 1548: The title and body here say the same thing in a slightly different way. Do we need them both?

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -1602,6 +1605,8 @@ async fn disk_bulk_write_import(
     apictx.external_latencies.instrument_dropshot_handler(&rqctx, handler).await
 }
 
+/// Stop importing blocks into a disk
+///
 /// Stop the process of importing blocks into a disk
```
zephraph, line 1610: Same as above

nexus/src/external_api/http_entrypoints.rs
benjaminleonard: The summary and description are not combined in the docs. At least not currently.

PR #2939, opened 2023-04-27
[installinator] Functional bootstrap networking 

jgallagher:
This PR contains mostly changes to `sled-agent` to support `installinator` connecting to `wicketd` over the bootstrap network. Highlighting the most significant commits:

1. Move enabling of ipv6-forwarding earlier (from sled-agent to bootstrap-agent): 59fe9c16f1503e3b0c83c6fd4743aecd1c32ba1b
2. Add the proper route in the switch zone to enable its bootstrap address: 625011346e73f0ee5575baa07fbebe621622d4f4, which builds on the commits immediately prior to it that do some work in `illumos-utils` to let us get the link-local address we need
3. Bumps to the latest maghemite to pick up the changes made in https://github.com/oxidecomputer/maghemite/pull/58
4. Replaces calls to `bootstrap_ip(..., 1)` and `bootstrap_ip(..., 2)` with `BootstrapInterface::GlobalZone.ip(...)` and `BootstrapInterface::SwitchZone.ip(...)`: ac752c71eaaa8945ff6998833355ab1482fb4348
5. Extends the ddm client `get_peers()` helper function to take a list of `BootstrapInterface`s the caller wants: `installinator` wants to find both gz and switch zone peers; `sled-agent` only wants to find gz peers

Edit: Two additional bugfixes based on chat and @andrewjstone's experience:

1. RSS should talk to its localhost mg-ddm, not its switch zone mg-ddm (since only scrimlets will have a switch zone: [b845621](https://github.com/oxidecomputer/omicron/pull/2939/commits/b8456218eb167e06a7328f8a94eaeb012f49ea97)
2. Rename `DdmAdminClient::get_peers()` to `DdmAdminClient::derive_bootstrap_addrs_from_prefixes`: [ea19f9c](https://github.com/oxidecomputer/omicron/pull/2939/commits/ea19f9c658939480088f97b4c7edc6a8b3840d99)

PR #2940, opened 2023-04-27
[wicketd] Refuse to update the sled where wicketd is running

jgallagher:
The point of the PR is in 49af4230ba1f8927ca85ece8c1200d6397aab8ba: wicketd now refuses to update the sled on which it is running, by comparing the baseboard info it retries from the SP of the to-be-updated sled against the baseboard info of its own host. However, the _bulk_ of the PR is in 917996fa10fa9659f28040ffbbdf51f2b82766a6, which plumbs that baseboard info through sled-agent down into wicketd. A nontrivial byproduct of this is the addition of `#[serde(skip)]` on several `ServiceType` variants that should not be passed in from external clients (i.e., services that sled agent itself is responsible for starting autonomously, either in the gz (maghemite) or in the switch zone (the rest)).

Builds on #2939

### DIFF
--- a/sled-agent/src/params.rs
+++ b/sled-agent/src/params.rs
@@ -420,11 +420,11 @@ impl crate::smf_helper::Service for ServiceType {
 /// started autonomously by sled-agent) into a
 /// `sled_agent_client::types::ServiceType` to be sent to a remote sled-agent.
 #[derive(Debug, Clone, Copy, Error)]
-#[error("This service may not be requested externally")]
-pub struct InternalServiceOnlyError;
+#[error("This service may only be started autonomously by sled-agent")]
+pub struct AutonomousServiceOnlyError;
 
 impl TryFrom<ServiceType> for sled_agent_client::types::ServiceType {
-    type Error = InternalServiceOnlyError;
+    type Error = AutonomousServiceOnlyError;
 
     fn try_from(s: ServiceType) -> Result<Self, Self::Error> {
         use sled_agent_client::types::ServiceType as AutoSt;
@@ -472,7 +472,7 @@ impl TryFrom<ServiceType> for sled_agent_client::types::ServiceType {
             | St::Wicketd { .. }
             | St::Dendrite { .. }
             | St::Tfport { .. }
-            | St::Maghemite { .. } => Err(InternalServiceOnlyError),
+            | St::Maghemite { .. } => Err(AutonomousServiceOnlyError),
         }
     }
 }
@@ -576,7 +576,7 @@ impl ServiceZoneRequest {
 impl TryFrom<ServiceZoneRequest>
     for sled_agent_client::types::ServiceZoneRequest
 {
-    type Error = InternalServiceOnlyError;
+    type Error = AutonomousServiceOnlyError;
 
     fn try_from(s: ServiceZoneRequest) -> Result<Self, Self::Error> {
         let mut services = Vec::with_capacity(s.services.len());
@@ -607,7 +607,7 @@ pub struct ServiceZoneService {
 impl TryFrom<ServiceZoneService>
     for sled_agent_client::types::ServiceZoneService
 {
-    type Error = InternalServiceOnlyError;
+    type Error = AutonomousServiceOnlyError;
 
     fn try_from(s: ServiceZoneService) -> Result<Self, Self::Error> {
         let details = s.details.try_into()?;
--- a/sled-agent/src/params.rs
+++ b/sled-agent/src/params.rs
@@ -10,8 +10,10 @@ use omicron_common::api::internal::shared::{
 };
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
+use sled_hardware::Baseboard;
 use std::fmt::{Debug, Display, Formatter, Result as FormatResult};
 use std::net::{IpAddr, Ipv6Addr, SocketAddr, SocketAddrV6};
+use thiserror::Error;
 use uuid::Uuid;
 
 pub use illumos_utils::opte::params::VpcFirewallRule;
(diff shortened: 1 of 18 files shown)


sled-agent/src/params.rs
```diff
@@ -374,59 +389,60 @@ impl crate::smf_helper::Service for ServiceType {
...
 }
 
-impl From<ServiceType> for sled_agent_client::types::ServiceType {
-    fn from(s: ServiceType) -> Self {
+/// Error returned by attempting to convert an internal service (i.e., a service
+/// started autonomously by sled-agent) into a
+/// `sled_agent_client::types::ServiceType` to be sent to a remote sled-agent.
+#[derive(Debug, Clone, Copy, Error)]
+#[error("This service may not be requested externally")]
+pub struct InternalServiceOnlyError;
```
jgallagher, line 397: Looking over this again, this may be a confusing name given "internal / external" in other contexts. This is supposed to mean "internal to sled agent". This is the error type returned if we try to serialize one of the `serde::skip`'d services into a `sled_agent_client::types::ServiceType`; I'm very open to better naming ideas. Maybe `AutonomousServiceOnlyError`, or even `SwitchZoneServiceError` (although that's technically a lie because we also autonomously start Maghemite in the global zone).

smklein: I like `AutonomousServiceOnlyError` - the distinction really is "who controls the deployment of this service", so that's intuitive.

smf/wicketd/manifest.xml
```diff
@@ -12,7 +12,7 @@
   </dependency>
 
   <exec_method type='method' name='start'
-    exec='ctrun -l child -o noorphan,regent /opt/oxide/wicketd/bin/wicketd run /var/svc/manifest/site/wicketd/config.toml --address %{config/address} --artifact-address %{config/artifact-address} --mgs-address %{config/mgs-address}  &amp;'
+      exec='ctrun -l child -o noorphan,regent /opt/oxide/wicketd/bin/wicketd run /var/svc/manifest/site/wicketd/config.toml --address %{config/address} --artifact-address %{config/artifact-address} --mgs-address %{config/mgs-address} --baseboard-identifier %{config/baseboard-identifier} --baseboard-model %{config/baseboard-model} --baseboard-revision %{config/baseboard-revision} &amp;'
```
smklein, line 15: I'm forgetting where we landed, but just to confirm: Did we determine that this information is not accessible from within a non-global zone?

If possible, it seems massively preferable (from a safety and plumbing point-of-view) for wicketd to ask "what is my baseboard info" rather than relying on the sled agent to plumb it through.

wicketd/src/http_entrypoints.rs
```diff
@@ -165,13 +166,65 @@ async fn post_start_update(
...
+        GetInventoryResponse::Unavailable => None,
+    };
+
+    match sp_state {
+        Some(sp_state) => {
+            // If we have the state of the SP, are we allowed to update it?
+            if !rqctx.can_update_target_sp(target, &sp_state) {
+                return Err(HttpError::for_bad_request(
+                    None,
+                    "sleds cannot update themselves".into(),
```
smklein, line 209: Nit: This error could be a little confusing in the case of 

```rust
        if self.baseboard == Baseboard::unknown() {
            !is_scrimlet(id)
        } ...
```

We don't *know* we're on the scrimlet, we just wouldn't want to risk it here, right?

wicketd/src/context.rs
```diff
@@ -7,11 +7,45 @@
...
+    pub(crate) fn can_update_target_sp(
+        &self,
+        id: SpIdentifier,
+        state: &SpState,
+    ) -> bool {
+        // If we failed don't know our own baseboard, fall back to just using
+        // the id, from which we know whether or not the target sled is one
+        // of the scrimlets. If it is, refuse to update, because it might be
+        // where we're running now.
+        if self.baseboard == Baseboard::unknown() {
```
smklein, line 34: This state does seem kinda like a pain to deal with -- could we just make the baseboard information non-optional, and have wicketd refuse to start if the correct baseboard info is not identified?

sled-hardware/src/lib.rs
```diff
@@ -73,7 +73,7 @@ pub enum SledMode {
 }
 
 /// Describes properties that should uniquely identify a Gimlet.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
```
smklein, line 76: I can't comment on line 91 because you didn't actually change it, but FYI, I'd have some caution depending on `Baseboard::unknown` -- that's kinda on my "gross hacks to clean up" list, rather than my "public API we should be really using" list.

smklein (approved):

smf/wicketd/manifest.xml
jgallagher: This information is indeed not accessible from a non-global zone. I chatted with @citrus-it briefly about making the `devinfo` device accessible in a zone, and it sounded like there were enough questions around permissions/security that it would be more straightforward to have sled-agent plumb this through.

wicketd/src/context.rs
jgallagher: It would be very easy from wicketd's point of view to remove this check and just assume sled-agent gave it something valid. Do you know what you expect sled-agent to do if/when you remove `Baseboard::unknown()` and it has problems reading the baseboard info?

smklein: To clarify, I think that it's the responsibility of the SMF `start` method + the CLI argument parsing to catch the case where `unknown` was provided, and opt to throw an error.

My only suggestion here is just to bail out *earlier* in the process, so that the `wicketd` implementation doesn't need to deal with something that is functionally `Option<Baseboard>`.

wicketd/src/http_entrypoints.rs
jgallagher: That's right; from the other comments we may ditch that case entirely, but if we decide to keep it, I could have `can_update_target_sp` return an enum instead to cover all three cases ("definitely yes", "definitely no", "no because we had problems and don't want to risk it") and give different error responses for the latter two.

wicketd/src/context.rs
smklein: > Hmm, that seems a little weird - either wicket's CLI argument parsing now has special checks for the `Baseboard::unknown()` values that are themselves placeholders

This is kinda what I was thinking - we're doing the same thing here, but it's  further away from the argument parsing.

jgallagher: Ok, I reworked this in [bc7aed0](https://github.com/oxidecomputer/omicron/pull/2940/commits/bc7aed05b88f5cba709d97200da0a3d93995b97e). There's still one call to `Baseboard::unknown()`, but it's in wicketd's main and has a TODO that it will need to be removed at some point. The rest of wicketd traffics solely in `Option<Baseboard>`.

wicketd/src/http_entrypoints.rs
jgallagher: As part of the rework in [bc7aed0](https://github.com/oxidecomputer/omicron/pull/2940/commits/bc7aed05b88f5cba709d97200da0a3d93995b97e) we now return different error messages for these two cases.

PR #2941, opened 2023-04-27
Update Dendrite and flesh out update script

bnaecker:
- Bumps pinned commit of Dendrite
- Adds functions to `./toos/update_dendrite.sh` for pulling the other artifacts that we need that are not specified in the package manifest TOML file. This is really the SHAs for the OpenAPI document and the stub binaries used to run `dpd` in CI.


bnaecker:
This is to resolve this issue @leftwo found, discussed [here](https://github.com/oxidecomputer/omicron/pull/2904#issuecomment-1526155913). We went back in time a bit in Dendrite, by pinning to a PR branch commit. We should stick to those on `main`, to make sure we're up to date.

smklein (approved):

leftwo (approved):
Yeah, it fixed my issue.

luqmana:
Now that `dendrite` gets MAC addresses from the sidecare FRUID, I think you need to configure an explicit `mac_base` for the stub/softnpu impls here:

https://github.com/oxidecomputer/omicron/blob/86826161f0a03d367755332949c6883008603de9/sled-agent/src/services.rs#L1302-L1312

bnaecker:
It’s sillier than that unfortunately. Dendrite will create a random base MAC if one isn’t provided, which we do, as can be seen in the log files. We also clear it each time we do and SMF refresh, which is less cool. There’s an issue in Dendrite, I’ll fix it in the morning. On 27 Apr 2023, at 17:53, Luqman Aden ***@***.***> wrote:﻿
Now that dendrite gets MAC addresses from the sidecare FRUID, I think you need to configure an explicit mac_base for the stub/softnpu impls here:
https://github.com/oxidecomputer/omicron/blob/86826161f0a03d367755332949c6883008603de9/sled-agent/src/services.rs#L1302-L1312

bnaecker:
Fix is up in https://github.com/oxidecomputer/dendrite/pull/265. I'm going to wait for that to publish CI artifacts, and the push a WIP commit here that points to it. I'll overwrite that with the CI artifacts on Dendrite `main` once the PR merges.

bnaecker:
Moved the checksums to those from the Dendrite `main` CI runs. Will automerge.

Issue #1681, opened 2022-09-07
better flesh out operator privileges in other Silos


davepacheco:
#1340 proposed that users with fleet-level privileges would have no privileges to access siloed resources in Silos other than their own.  (This phrasing wasn't really fleshed out until RFD 297, but I believe that's essentially what #1340 meant.)  The principle was essentially: we shouldn't be deciding who's allowed to cross Silo lines -- if an operator wants to access another Silo, they can do that, but they do it by having an account in that Silo's IdP, which makes it noisy and auditable.

RFD 309 raises a number of user stories that cast some doubt on this approach.  To be clear, I'm not sure yet what the right answer is, but there are enough questions that I think we want to revisit this before committing too far one way or the other.

CC @plotnick @rmustacc @kc8apf

zephraph:
I don't think it's quite clear to me how 309 casts doubt on the separation. We've always acknowledged that operators are going to need some lens into silo resources to be able to debug escalated issues, but my understand is that we still want to preserve the separation so that operators don't have unnecessary access to what might be confidential details.

#3092 is an example of where we're bending the boundaries a bit by making project/instance names visible to the operator. While this gives operators more insight into silo specific resources it does so to aid in establishing shared communication between the operator and the developer. I chose very specifically to only display the minimal context that the operator needs in this case.

davepacheco:
I understood #1340 (and my summary in [this comment](https://github.com/oxidecomputer/omicron/issues/1340#issuecomment-1171787290)) to be proposing that Fleet Administrators would have _no_ privileges to see just about anything inside other Silos.  Not the list of projects, instances, or anything like that.  I agree that's too rigid and RFD 309 explains why.  This ticket is about figuring out what these boundaries really are and ideally what principles guide them: e.g., provide access when there's particular value provided by a holistic, cross-Silo view; or where it's not even possible to correlate things without it: e.g., if a Fleet Admin can see sleds but not instances, and the Silo Admin can see instances but not sleds, then literally nobody can tell you what instances are on what sleds.

davepacheco:
When we do this, we should re-evaluate:

- whether it makes sense to allow Silo and Project policies to refer to identities outside the Silo.  See #1580.
- disallow fleet administrators from reading or modifying the policy on the Silo.  See #1580 for details.  Since fleet admins need to be able to modify which group is the "admin" group for a Silo, they may wind up changing the policy, but would not have carte blanche to change it however they like.

[...]

PR #1580, opened 2022-08-11
Fleet privileges should not cascade into Silos

davepacheco:
See #1340.  This change makes it so that having privileges at the Fleet level does _not_ mean you can see anything at all inside any Silo.  You can read, modify, and delete Silos, but not list children or create children.  One motivation for this change is that it means a security team wanting to audit a Silo can limit the search to users in the Silo's IdP, knowing that nobody else is able to do anything within the Silo.

Right now, Fleet administrators can still read and update the Silo's IAM policy.  I'm not sure if that's right, since in principle they could update the policy to include themselves (although they wouldn't have a way to access any siloed resources if they were in a different Silo, per RFD 297).  Tangentially, with this change in place, there's no reason any Silo's role assignments should ever refer to a silo user or group from another Silo.  We could enforce this.  That would satisfy the goal above (because it would still be true that only users in the Silo's IdP could actually do anything).  However, it would still let a Fleet admin grant a role to someone inside a Silo, which seems not what we want.  I think to prevent this, we'd have to actually separate the Read/ReadPolicy actions (which are currently identical) and the Modify/ModifyPolicy actions (same) so that we can grant "read" and "modify" here but not "ReadPolicy" or "ModifyPolicy".

If folks think this is a good idea, I can do this in a follow-on change.

CC @kc8apf

PR #2943, opened 2023-04-27
set up Recovery Silo and user during rack initialization

### LINKED ISSUE #2304: create Recovery Silo and initial user during RSS handoff
See RFD 234 section 1.4 ("Initial setup, special Silos") and RFD 278 determinations.  The plan (as I understand it) is that during initial setup on the technician port, the customer will provide (or get from the system) a password.  We will use this to create the initial Recovery Silo and an initial user in that Silo using that password.  The customer can then log into the Recovery Silo with that user and password and set up another Silo that's hooked up to their identity provider.

I think this work involves:

- [ ] The transfer-of-control request needs to accept whatever information is needed to create the initial Silo and Silo User.  At minimum I think this would be the new user's password.  Nexus can probably figure out the rest.  I'm not sure if there's value in allowing the Silo name or user name to be provided in this request.  (Even if we decide the customer can't change this, RSS could.  Is that useful?)
- [ ] All callers of the transfer-of-control request need to be updated:
    - [ ] I think the non-simulated Sled Agent currently acts as RSS, using input from [this config file](https://github.com/oxidecomputer/omicron/blob/main/smf/sled-agent/config-rss.toml)?  In that case this config file could grow a field for specifying the initial password?
    - [ ] The simulated Sled Agent makes this request.  It should probably get the password from the user (e.g., a command-line argument or interactive prompt) rather than hardcoding one.
    - [ ] The test suite is deliberately as close to the simulated Sled Agent as possible.  We should make sure its simulated Sled Agent makes the transfer-of-control request too.  It should probably use a hardcoded password that's only in the test suite code (i.e., not compiled into any shipping binaries).

I believe most of that is necessary for MVP anyway.  With that done, I think we should be able to remove the built-in Silo and privileged/test-privileged users.  I'll file a separate ticket for that.

davepacheco:
This change plumbs configuration for the initial (Recovery) Silo (see [RFD 234](https://github.com/oxidecomputer/rfd/tree/master/rfd/0234#initial-setup-special-silos)), the initial (privileged) user in that Silo, and the delegated DNS name for external DNS all the way from RSS to Nexus.  Nexus now creates that SIlo, the user, grants the user Fleet Admin privileges, etc.

Fixes #2304.


nexus/db-queries/src/db/datastore/silo.rs
```diff
@@ -91,6 +92,37 @@ impl DataStore {
         new_silo_params: params::SiloCreate,
         dns_update: DnsVersionUpdateBuilder,
     ) -> CreateResult<Silo> {
+        let conn = self.pool_authorized(opctx).await?;
+        self.silo_create_conn(
+            conn,
+            opctx,
```
smklein, line 98: I know this isn't totally new in this PR, but what's the justification for separately passing an `opctx` and a `group_opctx`?

I've always considered `OpContext` a representation of "who called this function, and with what authn/authz", so seeing two of them in a single function is a little jarring. After all, isn't only one "caller" asking for the silo to be created?

(I think this has something to do with "automatically giving admin access to a particular group within a newly created silo", I guess I'm just a little surprised to see that take the form of an `OpContext`, since I would have suspected that this group name would just be a "name" parameter. **I think passing an OpContext may be totally fine here**, but I think it's probably worth clarifying the distinction between two parameters to a function that happen to have the same type)

nexus/db-queries/src/db/datastore/rack.rs
```diff
@@ -456,68 +596,89 @@ mod test {
...
-    fn external_dns_empty() -> InitialDnsGroup {
-        InitialDnsGroup::new(
-            DnsGroup::External,
-            "testing.oxide.example",
-            "test suite",
-            "test suite",
-            HashMap::new(),
-        )
-    }
+    /// returns a
```
smklein, line 661: ?

nexus/db-queries/src/db/datastore/rack.rs
```diff
@@ -531,24 +692,69 @@ mod test {
...
+        // It should *not* show up in the list because it's not discoverable.
+        assert_eq!(silos.len(), 0);
+        let (authz_silo, db_silo) = LookupPath::new(&opctx, &datastore)
+            .silo_name(&nexus_db_model::Name(
+                rack_init.recovery_silo.identity.name.clone(),
+            ))
+            .fetch()
+            .await
+            .expect("Failed to lookup Silo");
+        assert!(!db_silo.discoverable);
```
smklein, line 721: Do we confirm that it's not possible for an operator to create a new silo with the same name as the recovery silo?

nexus/passwords/src/lib.rs
```diff
@@ -416,4 +497,74 @@ mod test {
                 .unwrap();
         assert_eq!(alt_hash, password_hash_str.to_string());
     }
+
+    #[test]
+    fn test_weak_hashes() {
```
smklein, line 502: Thanks for these tests!

nexus/test-utils/src/lib.rs
```diff
@@ -260,11 +264,26 @@ pub async fn test_setup_with_config<N: NexusServer>(
     };
     let external_dns_zone_name =
         internal_dns::names::DNS_ZONE_EXTERNAL_TESTING.to_string();
+    let silo_name: Name = "test-suite-silo".parse().unwrap();
+    let user_name = UserId::try_from("test-privileged".to_string()).unwrap();
+    let recovery_silo = RecoverySiloConfig {
+        silo_name: silo_name.clone(),
+        user_name: user_name.clone(),
+        // The test suite's password is "oxide".  This password is only used by
```
smklein, line 272: Suppose we need to change this value, maybe because we're using a new hash function. How would I go about updating this value?

I noticed the string "oxide" hard-coded in some tests, so this seems a little subtle to change.

smklein (approved):
Looks great, thanks for adding this!

nexus/test-utils/src/lib.rs
davepacheco: Good question.  What belongs here is a password hash in PHC format using algorithm argon2id and parameters at least as strong as we currently use to generate password hashes for new users.  The API documentation for this type explains this.  It's a little vague about what "secure enough parameters" means because (1) I didn't want to encode that information in multiple places and have it go out of date, (2) it's an internal API anyway, and (3) if you get it wrong, the failure mode is very explicit (e.g., https://github.com/oxidecomputer/omicron/blob/2317079d26e22c78c67054e4ec7fa2a2ad5aeb79/nexus/passwords/src/lib.rs#L548-L549).

As you pointed out, it _also_ needs to be consistent with any places that _use_ the password.  There shouldn't be many of these.  Presumably tests will fail (somewhat explicitly -- with a login failure) if that happens.  There are a few different places that use the same password (e.g., `omicron-dev run-all`), but that's only for convenience.  If you changed this one and "forgot" to update that one, that's fine.

Given these constraints, you can construct this hash with any tool that can generate Argon2 hashes.  I use [argon2.online](https://argon2.online/) for test stuff like this.

Background: I started from the position that it's safer to transfer the password hash in the APIs (from Wicket -> RSS and from RSS -> Nexus) rather than the password itself.  That said, for the test suite and `omicron-dev run-all`, we could define a Rust constant which is the password itself, and then have us hash that and put _that_ into the API calls.  That might require more things outside of Nexus to depend on `nexus-passwords`, but maybe that's okay, and probably has to happen eventually if Wicket is to hash whatever password the user provides.

---

I'll think a bit more about whether to commonize some of the password and hash uses in the test suite, `run-all`, and sim sled agent.  (I would probably _not_ commonize these with the two config-rss.toml files, even though they use the same password, because (1) nothing depends on these being the same as each other or anything else, and (2) these are supposed to be exactly what would come in via RSS, so I don't want to put a cleartext password there, and (3) I don't think there's a way to refer to a Rust constant in those.

nexus/db-queries/src/db/datastore/rack.rs
davepacheco: Only by way of existing tests that verify that you cannot create a Silo with a name that conflicts with an existing Silo.  I could add an explicit test here.

nexus/db-queries/src/db/datastore/rack.rs
davepacheco: Oops!

nexus/db-queries/src/db/datastore/silo.rs
davepacheco: Yeah, the problem here is that you have two different actors performing this operation together: (1) the end user creating the Silo, and (2) Nexus itself, acting on behalf of that same user.  The reason is that while the person is probably a Fleet Admin, they do not generally have the privilege to create a group inside the Silo (see #1580, #1681), and they certainly don't have permission to read and modify the fleet-wide DNS configuration.  But we _do_ want them to be able to do create a group and read/modify the DNS configuration _in this specific case_ that is highly controlled.  This is similar to the behavior that during provisioning, Nexus needs access to the list of Sleds in the system, but the _user_ doing the provisioning doesn't have permission to see that.  In both cases, Nexus assumes an identity _as itself_ to do those operations.  This one's a little more complicated because we need _both_ OpContexts.  We can't _just_ use Nexus's because we want to use the user's OpContext to decide whether they're allowed to do other stuff as part of this operation, like create the Silo itself.

There's a comment about this in the caller:
https://github.com/oxidecomputer/omicron/blob/2317079d26e22c78c67054e4ec7fa2a2ad5aeb79/nexus/src/app/silo.rs#L85-L88

I will update that comment to talk about DNS and add some comments here to explain this.

nexus/db-queries/src/db/datastore/rack.rs
davepacheco: I filed #2985 for part of this.  I'm also adding a test that we cannot re-use the name while it's in use.  (Turns out there was a bug here where we would fail on a 500 because we were not handling the database error correctly.  This is not specific to the recovery Silo.)

nexus/db-queries/src/db/datastore/silo.rs
davepacheco: Fixed in the latest commit.

nexus/db-queries/src/db/datastore/rack.rs
davepacheco: Fixed in the latest commit.

nexus/db-queries/src/db/datastore/rack.rs
davepacheco: Test added in the latest commit.

nexus/test-utils/src/lib.rs
davepacheco: Updated in the latest commit.  With each hardcoded hash I added a comment explaining what depends on it and referring people to the relevant docs.  I also had the test suite itself encode the password rather than the hash so that it can generate its own hash and we don't need this hardcoded there.

davepacheco:
I think I've addressed all of the review feedback (and it was approved anyway), so I'm going to enable auto-merge.  Feel free to disable if I've missed something, and of course I'm happy to iterate post-integration as needed.

PR #2946, opened 2023-04-28
[sled-agent] Refactor service management out of `StorageManager`

smklein:
## History

The Sled Agent has historically had two different "managers" responsible for Zones:

1. `ServiceManager`, which resided over zones that do not operate on Datasets
2. `StorageManager`, which manages disks, but also manages zones which operate on those disks

This separation is even reflected in the sled agent API exposed to Nexus - the Sled Agent exposes:

- `PUT /services`
- `PUT /filesystem`

For "add a service (within a zone) to this sled" vs "add a dataset (and corresponding zone) to this sled within a particular zpool".

This has been kinda handy for Nexus, since "provision CRDB on this dataset" and "start the CRDB service on that dataset" don't need to be separate operations. Within the Sled Agent, however, it has been a pain-in-the-butt from a perspective of diverging implementations. The `StorageManager` and `ServiceManager` have evolved their own mechanisms for storing configs, identifying filesystems on which to place zpools, etc, even though their responsibilities (managing running zones) overlap quite a lot.

## This PR

This PR migrates the responsibility for "service management" entirely into the `ServiceManager`, leaving the `StorageManager` responsible for monitoring disks.

In detail, this means:

- The responsibility for launching Clickhouse, CRDB, and Crucible zones has moved from `storage_manager.rs` into `services.rs`
  - Unfortunately, this also means we're taking a somewhat hacky approach to formatting CRDB. This is fixed in https://github.com/oxidecomputer/omicron/pull/2954.
- The `StorageManager` no longer requires an Etherstub device during construction
- The `ServiceZoneRequest` can operate on an optional `dataset` argument
- The "config management" for datastore-based zones is now much more aligned with non-dataset zones. Each sled stores `/var/oxide/services.toml` and `/var/oxide/storage-services.toml` for each group.
  - These still need to be fixed with https://github.com/oxidecomputer/omicron/issues/2888 , but it should be simpler now.
- `filesystem_ensure` - which previously asked the `StorageManager` to format a dataset and also launch a zone - now asks the `StorageManager` to format a dataset, and separately asks the `ServiceManager` to launch a zone.
  - In the future, this may become vectorized ("ensure the sled has *all* the datasets we want...") to have parity with the service management, but this would require a more invasive change in Nexus.


sled-agent/src/bootstrap/hardware.rs
```diff
@@ -178,10 +178,7 @@ impl HardwareMonitor {
         let hardware = HardwareManager::new(log, sled_mode)
             .map_err(|e| Error::Hardware(e))?;
 
-        // TODO: The coupling between the storage and service manager is growing
-        // pretty tight; we should consider merging them together.
-        let storage_manager =
-            StorageManager::new(&log, underlay_etherstub.clone()).await;
+        let storage_manager = StorageManager::new(&log).await;
```
andrewjstone, line 181: Woohoo!

sled-agent/src/services.rs
```diff
@@ -226,11 +215,42 @@ impl Config {
...
+        .join(SERVICES_CONFIG_FILENAME)
+}
+
+// The default path to storage service configuration
+fn default_storage_services_ledger_path() -> PathBuf {
+    Path::new(omicron_common::OMICRON_CONFIG_PATH)
+        .join(STORAGE_SERVICES_CONFIG_FILENAME)
+}
+
+// TODO(ideas):
```
andrewjstone, line 240: Looks like this first part is implemented.

sled-agent/src/services.rs
```diff
@@ -226,11 +215,42 @@ impl Config {
...
+// - "ServiceLedger"
+// - Manages the serializable "AllZoneRequests" object
+// - Constructor which reads from config location (kinda like
+// "read_from")
+// - ... Writer which *knows the type* to be serialized, so can direct it to the
+// appropriate output path.
+//
+// - TODO: later: Can also make the path writing safer, by...
```
andrewjstone, line 248: Maybe move this technique out into an issue?

sled-agent/src/services.rs
```diff
@@ -423,88 +576,14 @@ impl ServiceManager {
...
-                .await?;
-        } else {
-            info!(
-                &self.inner.log,
-                "No sled services found at {}",
-                config_path.to_string_lossy()
-            );
-        }
+        self.load_non_storage_services().await?;
+        // TODO: These will fail if the disks aren't attached.
```
andrewjstone, line 580: I'm having a hard time reasoning about when we'd want to retry, since it's unclear to me when a disk would become attached within a retry period. My admittedly, somewhat uninformed, take at this moment is that we shouldn't retry. 

I think the distinction between these calls not retrying and NTP retrying is twofold:
 1. NTP retries even on successful startup, since we are waiting for time sync
 2. NTP is a barrier to starting other services. If it fails later services will not work.

sled-agent/src/services.rs
```diff
@@ -211,9 +199,10 @@ pub struct Config {
     /// An optional internet gateway address for external services.
     pub gateway_address: Option<Ipv4Addr>,
 
-    /// The path for the ServiceManager to store information about
-    /// all running services.
-    pub all_svcs_config_path: PathBuf,
+    // The path for the ServiceManager to store information about
+    // all running services.
+    all_svcs_ledger_path: PathBuf,
```
andrewjstone, line 204: Why the change from `config` to `ledger` here? 

Also, if we are going to change the names, we may also want to change the constants to refer to `LEDGER` instead of `CONFIG`.

andrewjstone (commented):
This is a nice cleanup! I'm happy to get the `StorageManager` out of the business of launching zones.

sled-agent/src/services.rs
smklein: I actually implemented this out fully within https://github.com/oxidecomputer/omicron/pull/2972

sled-agent/src/services.rs
smklein: See: https://github.com/oxidecomputer/omicron/pull/2972 - I just finished implementing this.

andrewjstone: Oh. Excellent!

sled-agent/src/services.rs
smklein: I think I got confused between the following overloaded use of the term "config":

- We use "config" to describe parameters for a variety of modules within the sled agent, to describe tweaks on how they should be executed. For example, the sled agent has a "config", the bootstrap agent has a "config", the storage manager also has a "config".
- **totally separately**, we're storing the list of services that the sled manages. I had previously called this "config", but I think the name was not-totally-accurate, so I'm updating it to ledger. After all, it's just a ledger of "what am I responsible for running".

RE: the constants, will do!

sled-agent/src/services.rs
smklein: This definitely deserves a follow-up bug - it's only relevant in the "cold boot" case, but that does matter!

Here's the deal:

- When we boot the sled agent, it's possible that not all disks have been parsed (e.g., suppose there's a U.2 that's slow to bind a driver).
- When we call this function, we'll read the service ledger from the M.2s, see what services *should* be started, and try starting them.
- If any of those services...
  - ... have datasets in the U.2s
  - ... have zone filesystems in the U.2s
- ... then we'd fail to start them.

But that doesn't mean we should *never* launch the service - if it's in the M.2 service ledger, either RSS or Nexus provisioned that service, so we should keep giving it a shot. "The driver did bind, but it just took a while"  and "The U.2 was unplugged, but someone put it back" are both cases where we should be able to launch these services, even if we might initially fail.

If, after a long enough period of time, we determine that we cannot launch that service (if the U.2 is detached, fails, etc), then we'd have an opportunity to do some notification through the fault tolerance system (we either tell Nexus that the service isn't booting, or Nexus notices on its own somehow).

smklein: Filed https://github.com/oxidecomputer/omicron/issues/2973 , will mention it in this PR

sled-agent/src/services.rs
smklein: Done in https://github.com/oxidecomputer/omicron/pull/2946/commits/5d599516d4e4504bbae00347dcddbeea8f17146a

sled-agent/src/services.rs
smklein: Comment removed in https://github.com/oxidecomputer/omicron/pull/2946/commits/541f68dee6cdf9be61f2a0c71475ccbdccd65def

sled-agent/src/services.rs
smklein: Mentioned in https://github.com/oxidecomputer/omicron/pull/2946/commits/ed20fffa842f9f048373d59cff5fe157794b1e6f

andrewjstone: Thanks for the explanation. Those justifications make sense to me.

andrewjstone (approved):
Thanks for the explanations and cleanup. Ship it!

PR #2949, opened 2023-04-28
[installinator] Run diskinfo before scanning for M.2s

jgallagher:
Works around https://github.com/oxidecomputer/stlouis/issues/395.

Minor change: moves scanning for disks inside the relevant update-engine step to avoid installinator exiting without reporting an error to wicketd if something goes wrong while scanning.

Builds on #2940


jclulow:
Have you been able to confirm that this is an adequate workaround?

jgallagher:
It has worked every time I've tried it (which is not very many; < 5).

andrewjstone (approved):
Hooray for workarounds.

PR #2950, opened 2023-04-28
[installinator] Write host OS to M.2 block device in appropriately-sized chunks

jgallagher:
This adds a couple utility modules:

* `illumos_utils::dkio` contains a wrapper for the `dkio` ioctl we need to call to ask a device what its block size is.
* `installinator::block_size_writer` implements an `AsyncWrite` adapter that buffers data and always passes exactly block-size bytes down to its inner writer (except when explicitly flushed).

and then uses them in `installinator` to ensure that when we've found the host OS destination by scanning hardware for M.2s, we configure our writing transport to use a block-size writer configured for the target device.

Builds on #2949.

### DIFF
--- a/installinator/src/block_size_writer.rs
+++ b/installinator/src/block_size_writer.rs
@@ -1,0 +1,223 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use pin_project_lite::pin_project;
+use std::io;
+use std::pin::Pin;
+use std::task::ready;
+use std::task::Context;
+use std::task::Poll;
+use tokio::io::AsyncWrite;
+
+pin_project! {
+    /// `BlockSizeWriter` is analogous to a tokio's `BufWriter`, except it
+    /// guarantees that writes made to the underlying writer are always
+    /// _exactly_ the requested block size, with two exceptions: explicitly
+    /// calling (1) `flush()` or (2) `shutdown()` will write any
+    /// buffered-but-not-yet-written data to the underlying buffer regardless of
+    /// its length.
+    pub(crate) struct BlockSizeWriter<W> {
+        #[pin]
+        inner: W,
+        buf: Vec<u8>,
+        block_size: usize,
+    }
+}
+
+impl<W: AsyncWrite> BlockSizeWriter<W> {
+    pub(crate) fn with_block_size(block_size: usize, inner: W) -> Self {
+        Self { inner, buf: Vec::with_capacity(block_size), block_size }
+    }
+
+    #[cfg(test)]
+    fn into_inner(self) -> W {
+        self.inner
+    }
+
+    fn get_pin_mut(self: Pin<&mut Self>) -> Pin<&mut W> {
+        self.project().inner
+    }
+
+    fn flush_buf(
+        self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+    ) -> Poll<io::Result<()>> {
+        let mut me = self.project();
+        let mut written = 0;
+        let mut ret = Ok(());
+        while written < me.buf.len() {
+            match ready!(me.inner.as_mut().poll_write(cx, &me.buf[written..])) {
+                Ok(0) => {
+                    ret = Err(io::Error::new(
+                        io::ErrorKind::WriteZero,
+                        "failed to write the buffered data",
+                    ));
+                    break;
+                }
+                Ok(n) => written += n,
+                Err(e) => {
+                    ret = Err(e);
+                    break;
+                }
+            }
+        }
+        if written > 0 {
+            me.buf.drain(..written);
+        }
+        Poll::Ready(ret)
+    }
+}
+
+impl<W: AsyncWrite> AsyncWrite for BlockSizeWriter<W> {
+    fn poll_write(
+        mut self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+        buf: &[u8],
+    ) -> Poll<Result<usize, std::io::Error>> {
+        // We should never buffer up more than `block_size` data.
+        assert!(self.buf.len() <= self.block_size);
+
+        // If we already have exactly `block_size` bytes, begin by flushing.
+        if self.buf.len() == self.block_size {
+            ready!(self.as_mut().flush_buf(cx))?;
+        }
+
+        let me = self.project();
+        if me.buf.is_empty() {
+            // If our buffer is empty, either directly write `block_size` bytes
+            // from `buf` to `inner` (if there's enough data in `buf`) or just
+            // copy it into our buffer.
+            if buf.len() >= *me.block_size {
+                me.inner.poll_write(cx, &buf[..*me.block_size])
+            } else {
+                // `me.buf` is empty and `buf` is strictly less than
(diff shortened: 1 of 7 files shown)


installinator/src/block_size_writer.rs
```diff
@@ -0,0 +1,223 @@
...
+        if self.buf.len() == self.block_size {
+            ready!(self.as_mut().flush_buf(cx))?;
+        }
+
+        let me = self.project();
+        if me.buf.is_empty() {
+            // If our buffer is empty, either directly write `block_size` bytes
+            // from `buf` to `inner` (if there's enough data in `buf`) or just
+            // copy it into our buffer.
+            if buf.len() >= *me.block_size {
```
andrewjstone, line 91: I think it may be clearer to use `==` here, since we explicitly want `buf.len() <= self.block_size` and we assert that above.

installinator/src/block_size_writer.rs
```diff
@@ -0,0 +1,223 @@
...
+        fn poll_shutdown(
+            self: Pin<&mut Self>,
+            _cx: &mut Context<'_>,
+        ) -> Poll<Result<(), io::Error>> {
+            Poll::Ready(Ok(()))
+        }
+    }
+
+    #[proptest]
+    fn proptest_block_writer(
```
andrewjstone, line 168: Nice test :)

andrewjstone (approved):
Ship it!

installinator/src/block_size_writer.rs
jgallagher: Thanks! I was happy when I replaced `BlockSizeWriter` with tokio's `BufWriter` and saw the test fail immediately (even when constructed with capacity == block_size).

installinator/src/block_size_writer.rs
andrewjstone: Instead of renaming I suggest I do a better job reviewing :) Let's keep it as is. The naming is idiomatic.

installinator/src/block_size_writer.rs
```diff
@@ -0,0 +1,223 @@
...
+                // `me.buf` is empty and `buf` is strictly less than
+                // `self.block_size`, so just copy it.
+                me.buf.extend_from_slice(buf);
+                Poll::Ready(Ok(buf.len()))
+            }
+        } else {
+            // Our buffer already has data - just copy as much of `buf` as we
+            // can onto the end of it.
+            let n = usize::min(*me.block_size - me.buf.len(), buf.len());
+            me.buf.extend_from_slice(&buf[..n]);
```
andrewjstone, line 103: After looking at this again, I see that we may fill up a block size of data, but if so, we don't necessarily write it to the underlying writer via a call to `me.inner.poll_write`. If this is the last chunk of data though, are we guaranteed to get another call to `self.poll_write` or a call to `poll_flush`?

jgallagher: We're not; the onus is on the caller to call `.flush()` when they're done. tokio's `BufWriter` and `std::io::BufWriter` have the same requirements; tokio's docs include

> When the BufWriter is dropped, the contents of its buffer will be discarded... If you need to write out the contents of its buffer, you must manually call flush before the writer is dropped.

I could add something similar to the block comment on `BlockSizeWriter`?

andrewjstone: This makes sense. Thank you! I think a block comment would be warranted.

jgallagher: :+1: Added in [ad4f011](https://github.com/oxidecomputer/omicron/pull/2950/commits/ad4f01147dcd0966799fac039346b8c3ffd3420f)

installinator/src/write.rs
```diff
@@ -300,18 +330,23 @@ impl<'a> SlotWriteContext<'a> {
...
-        transport: &'b mut impl WriteTransport,
+        host_phase_2_transport: &'b mut impl WriteTransport,
+        control_plane_transport: &'b mut impl WriteTransport,
     ) {
         match self.progress {
             DriveWriteProgress::Unstarted
             | DriveWriteProgress::HostPhase2Failed => {
-                let transport_handle =
-                    self.register_host_phase_2_step(engine, transport);
-                self.register_control_plane_step(engine, transport_handle);
```
sunshowers, line 310: Nice simplification!

installinator/src/block_size_writer.rs
```diff
@@ -0,0 +1,227 @@
...
+    /// guarantees that writes made to the underlying writer are always
+    /// _exactly_ the requested block size, with two exceptions: explicitly
+    /// calling (1) `flush()` or (2) `shutdown()` will write any
+    /// buffered-but-not-yet-written data to the underlying buffer regardless of
+    /// its length.
+    ///
+    /// When `BlockSizeWriter` is dropped, any buffered data it's holding will
+    /// be discarded. It is critical to manually call `BlockSizeWriter:flush()`
+    /// or `BlockSizeWriter::shutdown()` prior to dropping to avoid data loss.
+    pub(crate) struct BlockSizeWriter<W> {
```
sunshowers, line 24: What do you think of calling this `BlockSizeBufWriter`?

sunshowers (approved):
Looks good! `BlockSizeWriter` could probably live in its own crate/repo even, seems generally useful.

installinator/src/block_size_writer.rs
```diff
@@ -0,0 +1,227 @@
...
+    /// `BlockSizeWriter` is analogous to a tokio's `BufWriter`, except it
+    /// guarantees that writes made to the underlying writer are always
+    /// _exactly_ the requested block size, with two exceptions: explicitly
+    /// calling (1) `flush()` or (2) `shutdown()` will write any
+    /// buffered-but-not-yet-written data to the underlying buffer regardless of
+    /// its length.
+    ///
+    /// When `BlockSizeWriter` is dropped, any buffered data it's holding will
+    /// be discarded. It is critical to manually call `BlockSizeWriter:flush()`
+    /// or `BlockSizeWriter::shutdown()` prior to dropping to avoid data loss.
```
sunshowers, line 23: I don't know how to feel about this, seems like a big footgun. I would prefer either writing the data in the `Drop` or just panicking if there's any data to be written.

installinator/src/block_size_writer.rs
jgallagher: Sure, seems good to hint at the other related types.

installinator/src/block_size_writer.rs
jgallagher: I don't think we can write the data without async drop, right?

I agree that this is a pretty gnarly footgun, but it's the exact same one tokio's `BufWriter` has. `std`'s BufWriter notes that it tries to write the data in drop but you have no feedback if it fails, so even if we could write the data we'd still want a "but you should really use flush when you're done" doc comment just like `std`'s BufWriter does.

Every time I've tried to add a panic in drop to catch client misbehavior it's come back to bite me. I'd be perfectly fine putting a `debug_assert` that the buffer is empty in `Drop`, but I'm not sure that would actually help us since the block writer only runs on real hardware (and is built in release mode basically always AFAIK).

sunshowers: Aha, forgot about async drop. Yeah, matching tokio's `BufWriter` is ok.

installinator/src/block_size_writer.rs
jgallagher: Renamed in a5329619

PR #2951, opened 2023-04-29
end-to-end tests could use external DNS and real authn

davepacheco:
Building on #2943, this updates the end-to-end tests to use the external DNS server to locate Nexus.  They also now use password-based authn using the user created during rack setup, just as a customer will do.  (Previously the tests used "spoof" authn with the built-in "test-privileged" user.)

Depends on #2943.


end-to-end-tests/src/helpers/ctx.rs
```diff
@@ -54,24 +64,16 @@ impl Context {
...
-        .timeout(Duration::from_secs(60))
-        .build()?;
-    Ok(Client::new_with_client(&get_base_url(), client))
+fn rss_config() -> Result<SetupServiceConfig> {
+    toml::from_str(RSS_CONFIG_STR)
+        .with_context(|| format!("parsing {:?} as TOML", RSS_CONFIG_PATH))
 }
 
-pub fn nexus_addr() -> SocketAddr {
+pub async fn nexus_addr() -> SocketAddr {
```
luqmana, line 76: Could make this an `anyhow::Result<SocketAddr>` to replace all the `unwrap`'s w/ ?

end-to-end-tests/src/helpers/ctx.rs
```diff
@@ -84,32 +86,168 @@ pub fn nexus_addr() -> SocketAddr {
...
+    .await
+    .unwrap_or_else(|e| panic!("failed to get Nexus addr: {}", e))
 }
 
-fn get_base_url() -> String {
+async fn get_base_url() -> String {
     // Check $OXIDE_HOST first.
     if let Ok(host) = std::env::var("OXIDE_HOST") {
         return host;
     }
```
luqmana, line 165: Remove this since `nexus_addr` seems to be checking it already?

end-to-end-tests/src/helpers/ctx.rs
```diff
@@ -84,32 +86,168 @@ pub fn nexus_addr() -> SocketAddr {
...
+    .await
+    .context("logging in")?;
+
+    let session_cookie = response
+        .headers()
+        .get(http::header::SET_COOKIE)
+        .ok_or_else(|| anyhow!("expected session cookie after login"))?
+        .to_str()
+        .context("expected session cookie token to be a string")?;
+    let (session_token, rest) = session_cookie.split_once("; ").unwrap();
```
luqmana, line 237: Can be a follow up but would be nice if `oxide_client` wrapped some of this up.

end-to-end-tests/src/helpers/ctx.rs
```diff
@@ -84,32 +86,168 @@ pub fn nexus_addr() -> SocketAddr {
...
+        .redirect(reqwest::redirect::Policy::none())
+        .timeout(Duration::from_secs(60))
+        .build()?;
+    let login_url = format!("{}/login/{}/local", base_url, silo_name);
+
+    // By the time we get here, we generally would have successfully resolved
+    // Nexus's external IP address from the external DNS server.  So we'd
+    // expect Nexus to be up.  But that's not necessarily true: external DNS can
+    // be set up during rack initialization, before Nexus has opened its
+    // external listening socket.  This is arguably a bug, advertising a service
```
luqmana, line 201: Should we create an issue to track that? I know the rack/service initialization process could still use some polish.

end-to-end-tests/src/helpers/ctx.rs
```diff
@@ -1,13 +1,23 @@
...
-use std::path::Path;
 use std::time::Duration;
+use trust_dns_resolver::config::{
+    NameServerConfig, Protocol, ResolverConfig, ResolverOpts,
+};
+use trust_dns_resolver::error::ResolveErrorKind;
+use trust_dns_resolver::TokioAsyncResolver;
+
+const RSS_CONFIG_PATH: &str =
+    "../../../smf/sled-agent/non-gimlet/config-rss.toml";
```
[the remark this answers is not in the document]
davepacheco: Done in the latest commits.  (There still needs to be a ".." because CARGO_MANIFEST_DIR is at "end-to-end-tests" relative to the workspace root.)

end-to-end-tests/src/helpers/ctx.rs
davepacheco: Done in the latest commits.

end-to-end-tests/src/helpers/ctx.rs
davepacheco: I think this was true before as well.  I went ahead and made this change though.

end-to-end-tests/src/helpers/ctx.rs
davepacheco: Yeah, fair question.  At first I thought you meant the progenitor client, where the plan is sort of uncertain (see oxidecomputer/progenitor#451), but I think you meant the oxide_client workspace crate, where we could definitely put something like this.

I'm not clear how reasonable an idea it is to use the login endpoint in this way (and that's part of the uncertainty around having progenitor support it).  In a real system we really expect people to use the console to log in and use a device authn token for programmatic access.  I think it's just automated rack setup where we can't do that.  So I think I'll hold off for now.

end-to-end-tests/src/helpers/ctx.rs
davepacheco: You mean to track the behavior that Nexus publishes its addresses to DNS before it's opened its listening socket?  I'm  not clear that's a bug but I'm happy to file an issue to document it and/or have that discussion.

luqmana (approved):
Thanks!

end-to-end-tests/src/helpers/ctx.rs
luqmana: If the expectation is that we should only be adding the DNS records once the service is up then seems we should have an issue tracking that

Issue #725, opened 2022-03-07
[sled-agent] Re-construct Zone management state within sled agent after reboot


smklein:
Note: This issue tracks a follow-up from https://github.com/oxidecomputer/omicron/pull/686

### Problem

Sled Agent should re-initialize in-memory state for managing Zones on boot, including:
- Identifying Omicron-owned zones
- Identifying their VNICs and other attached devices
- Creating in-memory structures to accurately represent their state (e.g. "RunningZone") such that they can be handled within the context of other Zones, and cleaned up accordingly.

### Current workaround

As the Sled Agent "deletes all zones on initialization", we can punt on this issue slightly.

### When this'll be a bigger problem

If we want to cope with Sled Agent itself crashing (without taking out all other software on the sled), this issue will become blocking.

bnaecker:
This issue also needs to expand to include recovering the data links used for guest instances. These are currently the `xde` devices, plus VNICs over them (required by the Viona virtio-net driver).

We are currently deleting all of the guest Propolis zones, and also deleting these links. We'll need to recover and track them similar to the control VNICs used for communicating with Oxide services running in zones.

askfongjojo:
Naive question: If sled-agent stops reaping propolis zones upon restart, would it be able to rely on info from the config files `/etc/zones` to re-construct the necessary zone info?
```
BRM42220081 # cat /etc/zones/oxz_propolis-server_8e00d49b-29e2-4b4f-8ab4-aefab6043c5b.xml 
<?xml version="1.0"?>
<!DOCTYPE zone PUBLIC "-//Sun Microsystems Inc//DTD Zones//EN" "file:///usr/share/lib/xml/dtd/zonecfg.dtd.1">
<!--
    DO NOT EDIT THIS FILE.  Use zonecfg(8) instead.
-->
<zone name="oxz_propolis-server_8e00d49b-29e2-4b4f-8ab4-aefab6043c5b" zonepath="/zone/oxz_propolis-server_8e00d49b-29e2-4b4f-8ab4-aefab6043c5b" autoboot="false" brand="omicron1" ip-type="exclusive">
  <device match="/dev/vmm/*"/>
  <device match="/dev/vmmctl"/>
  <device match="/dev/viona"/>
  <network physical="vopte1"/>
  <network physical="oxControlInstance0"/>
</zone>
```
The zone running state isn't there but putting a user VM at a stopped state would still be better than blowing it away as an initial solution to the problem.

[...]

PR #2953, opened 2023-04-30
Stop deleting chelsio addresses during uninstall

smklein:
## Before this PR

Running on rack2 and calling `omicron-package uninstall` would involve a fatal termination of the connection, as it would delete the `cxgbe0/ll` and `cxgbe1/ll` IP addresses necessary for contacting the sled.

## After this PR

Those addresses are left alone. This is pretty useful for development, as it allows us to run `uninstall` to cleanly wipe a Gimlet, preparing it for future "clean installs".

### DIFF
--- a/sled-hardware/src/cleanup.rs
+++ b/sled-hardware/src/cleanup.rs
@@ -6,11 +6,11 @@
 
 use anyhow::Error;
 use futures::stream::{self, StreamExt, TryStreamExt};
+use illumos_utils::dladm::Dladm;
 use illumos_utils::dladm::BOOTSTRAP_ETHERSTUB_NAME;
 use illumos_utils::dladm::BOOTSTRAP_ETHERSTUB_VNIC_NAME;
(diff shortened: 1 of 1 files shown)


smklein:
I previously managed my "write-compile-test" loop on Gimlets with the following:

1. Before installing new control plane software, delete `/var/oxide` to remove all Sled configs
2. Rely heavily on the sled agent functionality to "delete all old zones and vnics" on boot

However, this is getting less reasonable, because:

1. https://github.com/oxidecomputer/omicron/issues/2888 is coming, which means sled configs will be stored on zpool datasets, rather than `/var/oxide`.
2. https://github.com/oxidecomputer/omicron/issues/725 should also be done, which means that sled agent should *reconstruct* state on reboot, rather than destroy it.

Ergo, I need to have a better way to get Gimlets to a "clean slate" - by fixing the uninstall command, this should be a universal command to wipe (almost all) signs of control plane usage from prior installations.

luqmana (approved):
Thanks! beats having to get onto the console to fix

Issue #727, opened 2022-03-07
[sled-agent] Initialize multi-node CockroachDB, rather than single-node


smklein:
Note: This issue tracks a follow-up from https://github.com/oxidecomputer/omicron/pull/686

CockroachDB is initialized using "start-single-node" instead of "start".

This works for testing, but is obviously not representative of an actual production-ready rack. We should ensure that CRDB instances can be initialized / booted jointly, across multiple nodes.

smklein:
As an FYI for whoever picks this up - maybe future me - CRDB is picky about the clocks of all nodes being in sync. This will require clock synchronization between nodes, likely via NTP or some equivalent.

davepacheco:
Yeah, I recommend we use chrony for this.  When I was testing CockroachDB in late 2020, I found that the stock illumos ntpd was not even close to good enough to keep clocks in sync.  CockroachDB expects something like 250ms of drift.  After a few days or a few VM restarts in AWS, it was common for the clocks to drift further than this.  After setting up crony (which was pretty straightforward), the clocks were generally within 1ms (as measured by CockroachDB).

There are open questions about what we'd point any NTP server at.  RFD 34 remains open.

smklein:
Blocked on https://github.com/oxidecomputer/omicron/issues/2219

Issue #2122, opened 2023-01-09
Populate `/etc` for zones


smklein:
- [ ] [/etc/hosts](https://github.com/oxidecomputer/omicron/issues/1388)
- [ ] /etc/resolv.conf - transfer information about nameservers to zones, rather than having zones infer information from the /48 subnet

luqmana:
What's the split for things hardcoded into `/etc/hosts` vs resolved using internal DNS listed in `/etc/resolv.conf`?

smklein:
Good question, I dunno.

luqmana:
Ok, [RFD 248](https://rfd.shared.oxide.computer/rfd/248) was the one I was thinking of and based on that it seems for the most part we'd want things to resolve via DNS.

jclulow:
It is generally good UNIX hygiene to map whatever the system's name is (as seen via `hostname` or `uname -n`) to some IP interface on the local system via `/etc/inet/hosts`.  This could just be `::1` and `127.0.0.1` though, at which point the file could be essentially static.

I would do everything else through DNS.

leftwo:
From https://github.com/oxidecomputer/rfd/tree/0206/rfd/0206/#41-how-do-you-find-the-dns-servers
The DNS server I.P. can be determined/calculated using the address of the system/zone that needs it.

Do we want to put that into `/etc/resolv.conf` for a zone when it is created?

smklein:
> From https://github.com/oxidecomputer/rfd/tree/0206/rfd/0206/#41-how-do-you-find-the-dns-servers The DNS server I.P. can be determined/calculated using the address of the system/zone that needs it.
> 
> Do we want to put that into `/etc/resolv.conf` for a zone when it is created?

This seems like the right call - as far as I understand it, adding the `resolv.conf` server basically just short-cuts some of the manual address calculation that a zone would need to do, by providing the address of the nameserver up-front.

jclulow:
I'm not sure of the ordering of startup here.  It seems like it would be possible that at least some zones (and certainly the global zone) to start prior to our knowing what the local `/48` actually is.  We might need a service on each system (or, at least, on a subset of systems) that waits for the prefix to appear and then updates the file?

citrus-it:
[...]

PR #2954, opened 2023-04-30
[sled-agent] Separate CockroachDB "start" from CockroachDB "init"

smklein:
This is a necessary part of https://github.com/oxidecomputer/omicron/issues/727

## History

- Previously, CockroachDB was initialized via https://www.cockroachlabs.com/docs/stable/cockroach-start-single-node.html
- Additionally, the CRDB tables were explicitly initialized via a request to run `sql` using `dbinit.sql`
- This was historically done on CRDB datasets immediately when they were initialized - see: https://github.com/oxidecomputer/omicron/blob/9f31e37445f6d94c7277550e8048f09f5db61179/sled-agent/src/storage_manager.rs#L372-L394 
- In https://github.com/oxidecomputer/omicron/pull/2946 this was changed to just "unconditionally format during zone initialization", which is wrong, but was a holdover for...

## This PR

- Makes CRDB initialization an explicit endpoint. After booting, the CRDB zone only runs https://www.cockroachlabs.com/docs/stable/cockroach-start.html 
- ... but the all CRDB nodes can be started using https://www.cockroachlabs.com/docs/stable/cockroach-init.html , which is invoked by RSS.
- To pull this off, RSS invokes a new `POST /cockroachdb_init` endpoint exposed by the sled agent.


davepacheco:
I've taken on this PR.  The biggest change I've made so far here is that previously, when provisioning a CockroachDB zone, Sled Agent would look up the Cockroach service in DNS, take the addresses it found, and write those into SMF properties.  Then the start method would pull those out and use those as the "join" addresses so that Cockroach can find the other nodes in its cluster.  The problem with this is that there's nothing to keep those addresses in sync with what's in DNS (which is the source of truth about who's in the cluster).  Instead, I've added some code to configure DNS in the CockroachDB zones (using the dns/install SMF service provided by illumos, which I didn't previously know about!) and then use that in the start method to locate the CockroachDB nodes.

It's worth mentioning explicitly that we've designed internal DNS to be able to start without any external runtime dependencies, including CockroachDB (see RFD 248), which means that using DNS here should not cause a problem for cold boot.  As I write this, though, I realize that using `dig` from the start method like I'm currently doing probably isn't going to cut it.  We need something that will keep retrying until it succeeds in case it takes a little while for internal-DNS to come up.

There's another bit of this I'm a little worried about, which is exposing an API in sled agent to initialize the CockroachDB cluster.  That seems a little dangerous and also overkill since we only ever intend to do this once, and only before the control plane is initialized.  I considered changing this to instead have RSS pass configuration to an SMF service which would do this.  The problem is that there's not a great way to propagate success/failure information back to RSS so that it can decide whether to proceed (or, I guess, burn down the world and try again).  I'm going to defer fixing this for now because we really need to start playing with multi-node CockroachDB.

sled-agent/src/services.rs
```diff
@@ -1074,6 +1085,40 @@ impl ServiceManager {
...
+                // zone with the list of DNS nameservers.  This will cause
+                // /etc/resolv.conf to be populated inside the zone.  To do
+                // this, we need the full list of nameservers.  Fortunately, the
+                // nameservers provide a DNS name for the full list of
+                // nameservers.
+                //
+                // Note that when we configure the dns/install service, we're
+                // supplying values for an existing property group on the SMF
+                // *service*.  We're not creating a new property group, nor are
+                // we configuring a property group on the instance.t by default.
```
smklein, line 1099:
```suggestion
                // we configuring a property group on the instance by default.
```

smf/cockroachdb/method_script.sh
```diff
@@ -21,10 +21,24 @@ ipadm show-addr "$DATALINK/ll" || ipadm create-addr -t -T addrconf "$DATALINK/ll
 ipadm show-addr "$DATALINK/omicron6"  || ipadm create-addr -t -T static -a "$LISTEN_ADDR" "$DATALINK/omicron6"
 route get -inet6 default -inet6 "$GATEWAY" || route add -inet6 default -inet6 "$GATEWAY"
 
+# We need to tell CockroachDB the DNS names or IP addresses of the other nodes
+# in the cluster.  Look these up in internal DNS.  Per the recommendations in
+# the CockroachDB docs, we choose at most five addresses.  Providing more
+# addresses just increases the time for the cluster to stabilize.
+JOIN_ADDRS="$(/opt/oxide/internal-dns-cli/bin/dnswait cockroach \
+    | head -n 5 \
+    | tr '\n' ,)"
```
smklein, line 30: Nice, I like this solution. Thanks for fitting everything into the "self-assembling zone" approach -- it's neat that we can now refresh all our service info by simply restarting the CRDB service.

sled-agent/src/http_entrypoints.rs
```diff
@@ -226,6 +227,19 @@ async fn filesystem_put(
...
+    path = "/cockroachdb",
+}]
+async fn cockroachdb_init(
+    rqctx: RequestContext<SledAgent>,
+) -> Result<HttpResponseUpdatedNoContent, HttpError> {
+    let sa = rqctx.context();
+    sa.cockroachdb_initialize().await?;
+    Ok(HttpResponseUpdatedNoContent())
+}
+
```
smklein, line 242: Agreed with your comment that "this is sketchy, and it would be cool to not have it".

It seems like it could also be possible to have the method script auto-initialize the DB if it's empty, but that seems like it could raise other problems.

All that said, this approach is probably good enough to last a while.

sled-agent/src/services.rs
```diff
@@ -1074,6 +1085,40 @@ impl ServiceManager {
                 let Some(info) = self.inner.sled_info.get() else {
                     return Err(Error::SledAgentNotReady);
                 };
+
+                // We want to configure the dns/install SMF service inside the
```
smklein, line 1089: Should we pull this out into a separate method? I'd be shocked if this was the only service for which we wanted to populate `/etc/resolv.conf`.

(I mean, arguably, shouldn't we populate basically all our services with `/etc/resolv.conf`?)

internal-dns-cli/src/bin/dnswait.rs
```diff
@@ -0,0 +1,98 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
smklein, line 1: Just to make sure I'm clear - the main reason for using this instead of `dig +short <name>` is the timeout?

smklein (commented):
I can't approve this because technically it's my PR, but I do approve! This looks great!

sled-agent/src/services.rs
davepacheco: Thanks.  Fixed.

sled-agent/src/services.rs
davepacheco: I do think we should do this, but it's more complicated to use it:
https://github.com/oxidecomputer/omicron/issues/2122#issuecomment-1611870486

so I'd like to defer it for now.

internal-dns-cli/src/bin/dnswait.rs
davepacheco: Yeah, I had `dig` at first, but I realized it will only try a handful of times and only wait a few seconds, _and_ you can't tell from its exit status that it's failed.  I needed something that's going to retry an arbitrary number of times, waits until it finds some results, too, and if it's going to fail, fails clearly.

davepacheco:
I've gone ahead and enabled auto-merge: however one accounts this, @smklein both worked on this and both reviewed it.

internal-dns-cli/src/bin/dnswait.rs
smklein: Not for this PR, but maybe it makes sense to move this utility into the omicron-brand zone alongside some of the other network setup that might be more general? Seems a bit like a pain-in-the-butt to manually package it into each zone that could need it

davepacheco (approved):

Issue #732, opened 2022-03-07
[rss][nexus][sled-agent] Responsibility for deploying services should (where possible) migrate into Nexus


smklein:
At the time of writing, https://github.com/oxidecomputer/omicron/pull/686 introduces the first usage of the "RSS", which makes requests to sled agent to create datasets/services.

Aside from the datasets necessary for initializing Nexus (that is, Nexus itself and CRDB), these service requests should be handled by Nexus "as much as possible" instead of the RSS.

Many operations can trigger a need to request these services partitions:
- Initializing a new sled
- Adjusting service location to maintain availability while turning down an old sled
- Adjusting capacity to serve request load
- Etc

All these conditions are ongoing, and best handled by Nexus, which maintains a "global" view of the rack and exists beyond initialization.

Fortunately, the APIs defined by the Sled Agent should (more-or-less) remain the same - this issue just addresses the matter of "who calls them".

- [ ] Datasets
  - [x] Nexus should have RPW for provisioning CockroachDB 
  - [ ] Nexus should have RPW for provisioning Clickhouse 
  - [ ] Nexus should have RPW for provisioning Crucible 
- [ ] Services
  - [x] Nexus should have RPW for provisioning Nexus 
  - [ ] Nexus should have RPW for provisioning Internal DNS 
  - [ ] Nexus should have RPW for provisioning External DNS 
  - [ ] Nexus should have RPW for provisioning Oximeter
  - [ ] Nexus should have RPW for provisioning Crucible Pantry
- [x] Housekeeping
  - [x] #2977
    - [x] This generation number should be stored in the on-M.2 ledger for services
  - [x] #2978
    - [x] This generation number should be stored in the on-M.2 ledger for datasets

davepacheco:
> Aside from the datasets necessary for initializing Nexus (that is, Nexus itself and CRDB), these service requests should be handled by Nexus "as much as possible" instead of the RSS.

> Many operations can trigger a need to request these services partitions:

There are two things that might be meant by this issue:

- "Nexus should be able to invoke these various requests for the use cases that it's responsible for (e.g., adding a new sled)."  This covers all the examples listed in the description above.  This part has been done for a while now.
- "RSS should _not_ invoke these requests for anything that Nexus _could_ do (i.e., we should have RSS just provision NTP, DNS, Cockroach, and Nexus and let Nexus provision the rest)."  This is _not_ done but I think we should file a separate issue for this if/when we get serious about doing that.

So I'm going to close this one.

PR #2956, opened 2023-04-30
Provision multiple CRDB nodes, if multiple sleds exist

smklein:
This PR bumps the count of "attempted CRDB provisions" across sleds from one to three. Note that if only one sled is requested through RSS, we continue to initialize with a single CRDB node.

https://github.com/oxidecomputer/omicron/issues/732 will need to provide an auxiliary solution to ensure that this number of CRDB instances continues to be deployed, but this is a start.

Fixes https://github.com/oxidecomputer/omicron/issues/727


smklein:
Closing in favor of https://github.com/oxidecomputer/omicron/pull/3450

PR #2976, opened 2023-05-01
[sled-agent] Allow synthetic disks to be used as M.2s and U.2s.

smklein:
This fixes a bug where the internal zpools created by `./tools/create_virtual_hardware.sh` were ignored by the sled agent - this utility is used in other locations to access "m.2 datasets".


smklein:
Found this while debugging https://github.com/oxidecomputer/omicron/pull/2972 , figured I'd split it off to its own PR

andrewjstone (approved):
Thanks!

PR #2979, opened 2023-05-01
[sql] Add a constraint that sled baseboard info is unique

smklein:


zephraph (approved):

zephraph:
Looks great. I've also added https://github.com/oxidecomputer/omicron/issues/2984 as a follow up to the conversation we had.

PR #2981, opened 2023-05-02
Remove remaining bits of OPTE External IP hack.

luqmana:
Pull in updated OPTE and with that get rid of remaining bits of this hack!


tools/create_virtual_hardware.sh
```diff
@@ -138,21 +138,7 @@ function enable_softnpu {
...
-
-    grep "ext_ip_hack = 0;" /kernel/drv/xde.conf || {
-        echo "failed to disable ext_ip_hack"
-        exit 1
-    }
-    success "ext_ip_hack disabled"
-}
-
 ensure_run_as_root
-ensure_ext_ip_hack_disabled
```
internet-diglett, line 155: This function ensures that on any machine that is being upgraded from an older version of Omicron (pre-OPTE) that the hack gets disabled. I don't know if there are still machines / environments out there with the hack, but maybe we want to leave this here until we know for sure that there aren't any opte-hacked users still around?

internet-diglett (approved):
Left one comment, otherwise LGTM!

tools/create_virtual_hardware.sh
luqmana: Well the `ext_ip_hack` config should no longer exist once they upgrade to this newer xde build.

internet-diglett: Ah, so it won't even look for the value anymore. Perfect!

PR #2988, opened 2023-05-02
[sled-agent and friends] std::path to camino

smklein:
Convert `sled-agent`, `sled-hardware`, and `illumos-utils` to use `camino` instead of `std::path` wherever possible.


illumos-utils/src/running_zone.rs
```diff
@@ -144,7 +147,7 @@ impl RunningZone {
 
     /// Returns the filesystem path to the zone's root
     pub fn root(&self) -> String {
-        format!("{}/root", self.inner.zonepath.display())
+        format!("{}/root", self.inner.zonepath)
     }
```
sunshowers, line 151: This can be a `Utf8PathBuf`:

```rust
    pub fn root(&self) -> Utf8PathBuf {
        self.inner.zonepath.join("root")
    }
```

illumos-utils/src/running_zone.rs
```diff
@@ -638,7 +641,7 @@ impl InstalledZone {
 
         let full_zone_name = Self::get_zone_name(zone_type, unique_name);
         let zone_image_path =
-            PathBuf::from(&format!("/opt/oxide/{}.tar.gz", zone_type));
+            Utf8PathBuf::from(&format!("/opt/oxide/{}.tar.gz", zone_type));
```
sunshowers, line 644: No need to pass a `&String` here:

```suggestion
            Utf8PathBuf::from(format!("/opt/oxide/{}.tar.gz", zone_type));
```

sled-agent/src/config.rs
```diff
@@ -124,7 +124,7 @@ mod test {
     fn test_smf_configs() {
         let manifest = std::env::var("CARGO_MANIFEST_DIR")
             .expect("Cannot access manifest directory");
-        let smf = PathBuf::from(manifest).join("../smf/sled-agent");
+        let smf = Utf8PathBuf::from(manifest).join("../smf/sled-agent");
 
         let mut configs_seen = 0;
         for variant in std::fs::read_dir(smf).unwrap() {
```
sunshowers, line 130: One option here is to use [`read_dir_utf8`](https://camino-rs.github.io/camino/rustdoc/camino/struct.Utf8Path.html#method.read_dir_utf8). However, that will fail if any of the paths are non-UTF-8.

* You can keep iterating over the even if one of the entries fails to parse as valid UTF-8.
* This code uses unwraps on I/O errors already, so maybe these additional kinds of errors are not a huge deal.

sled-agent/src/updates.rs
```diff
@@ -220,6 +223,7 @@ impl UpdateManager {
             let entry = entry.map_err(|err| io_err(dir, err))?;
             let file_type =
                 entry.file_type().map_err(|err| io_err(dir, err))?;
+            let path: Utf8PathBuf = entry.path().try_into()?;
```
sunshowers, line 226: Definitely worth using `read_dir_utf8` here.

sunshowers (approved):
Looks fantastic, thanks for doing this! Just a few comments.

illumos-utils/src/running_zone.rs
smklein: Done in https://github.com/oxidecomputer/omicron/pull/2988/commits/ced5dc92aa41c74da6a846aaa9cd59d20d9c7843

illumos-utils/src/running_zone.rs
smklein: Done in https://github.com/oxidecomputer/omicron/pull/2988/commits/ced5dc92aa41c74da6a846aaa9cd59d20d9c7843

sled-agent/src/config.rs
smklein: Done in https://github.com/oxidecomputer/omicron/pull/2988/commits/ced5dc92aa41c74da6a846aaa9cd59d20d9c7843

Given that these are tests, I'm totally fine bailing if we hit something non-UTF-8

sled-agent/src/updates.rs
smklein: Done in https://github.com/oxidecomputer/omicron/pull/2988/commits/ced5dc92aa41c74da6a846aaa9cd59d20d9c7843

PR #2990, opened 2023-05-02
Introduce a `KeyManager`

andrewjstone:
The `KeyManager` will be used by sled-agent to retrieve rack secrets via an implementation of a `SecretRetriever` and generate derived keys.

Currently the only keys produced are those necessary for U.2 drive encryption, and follow the derivation scheme in section 4 of RFD 301.

For initail implementation, input key material (IKM) will be retrieved from a `SecretRetriever` backed by a sled local mechanism. Future implementations of `SecretRetriever` will use the trust quorum rack secret retrieved from the bootstore.

### DIFF
--- a/key-manager/src/lib.rs
+++ b/key-manager/src/lib.rs
@@ -5,6 +5,7 @@
 //! A crate used to derive keys useful for the Oxide control plane
 
 use std::collections::BTreeMap;
+use std::fmt::Debug;
 
 use async_trait::async_trait;
 use hkdf::Hkdf;
@@ -20,6 +21,14 @@ use zeroize::{Zeroize, ZeroizeOnDrop};
 /// `Zeroizable` data, and a `Box<[u8; 32]>` is not zeroizable on its own.
 #[derive(Zeroize, ZeroizeOnDrop)]
 pub struct Ikm(pub Box<[u8; 32]>);
+
+/// We impl Debug here only to guarantee nobody re-implements this to actually
+/// expose the secret.
+impl Debug for Ikm {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "Ikm(secret)")
+    }
+}
 
 /// Secret Input Key Material for a given rack reconfiguration epoch
 pub struct VersionedIkm {
--- a/key-manager/src/lib.rs
+++ b/key-manager/src/lib.rs
@@ -1,0 +1,344 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! A crate used to derive keys useful for the Oxide control plane
+
+use std::collections::BTreeMap;
+
+use async_trait::async_trait;
+use hkdf::Hkdf;
+use secrecy::{ExposeSecret, Secret};
+use sha3::Sha3_256;
+use sled_hardware::DiskIdentity;
+use zeroize::{Zeroize, ZeroizeOnDrop};
+
+/// Input Key Material
+///
+/// This should never be used directly, and always wrapped in a `Secret<Ikm>`
+/// upon construction. We sparate the two types, because a `Secret` must contain
+/// `Zeroizable` data, and a `Box<[u8; 32]>` is not zeroizable on its own.
+#[derive(Zeroize, ZeroizeOnDrop)]
+pub struct Ikm(pub Box<[u8; 32]>);
+
+/// Secret Input Key Material for a given rack reconfiguration epoch
+pub struct VersionedIkm {
+    pub epoch: u64,
+    pub salt: [u8; 32],
+    pub ikm: Secret<Ikm>,
+}
+
+impl VersionedIkm {
+    pub fn new(epoch: u64, salt: [u8; 32], data: &[u8; 32]) -> VersionedIkm {
+        let ikm = Secret::new(Ikm(Box::new(*data)));
+        VersionedIkm { epoch, salt, ikm }
+    }
+}
+
+/// An error returned by the [`KeyManager`]
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error("Secret not loaded for {epoch}")]
+    SecretNotLoaded { epoch: u64 },
+
+    #[error("Failed to retreive secret: {0}")]
+    SecretRetreival(#[from] SecretRetrieverError),
+}
+
+/// Derived Disk Encryption key
+#[derive(Zeroize, ZeroizeOnDrop, Default)]
+pub struct Aes256GcmDiskEncryptionKey(pub Box<[u8; 32]>);
+
+pub struct VersionedAes256GcmDiskEncryptionKey {
+    pub epoch: u64,
+    pub key: Secret<Aes256GcmDiskEncryptionKey>,
+}
+
+/// The main mechanism used to derive keys from a shared secret for the Oxide
+/// control plane.
+pub struct KeyManager<S: SecretRetriever> {
+    // A mechanism for retrieving input key material
+    secret_retriever: S,
+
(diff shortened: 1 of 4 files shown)


key-manager/src/lib.rs
```diff
@@ -0,0 +1,344 @@
...
+use sled_hardware::DiskIdentity;
+use zeroize::{Zeroize, ZeroizeOnDrop};
+
+/// Input Key Material
+///
+/// This should never be used directly, and always wrapped in a `Secret<Ikm>`
+/// upon construction. We sparate the two types, because a `Secret` must contain
+/// `Zeroizable` data, and a `Box<[u8; 32]>` is not zeroizable on its own.
+#[derive(Zeroize, ZeroizeOnDrop)]
+pub struct Ikm(pub Box<[u8; 32]>);
```
jgallagher, line 22: I assume the omission of `Debug` impls is intentional. Would it be useful to implement `Debug` manually in a way that does not reveal any data? I can think of a couple benefits but am not sure it's completely worth it:

1. types containing any of these values can derive `Debug`
2. it's very obvious that the `Debug` impl is special to avoid leaking information

key-manager/src/lib.rs
```diff
@@ -0,0 +1,344 @@
...
+/// The main mechanism used to derive keys from a shared secret for the Oxide
+/// control plane.
+pub struct KeyManager<S: SecretRetriever> {
+    // A mechanism for retrieving input key material
+    secret_retriever: S,
+
+    /// Pseudo-Random-Keys (PRKs) wrapped in an `Hkdf` structure so that we can
+    /// create keys from them with `HKDF-Expand` (`prk.expand_multi_info`).
+    ///
+    /// In the common case, we will only have a single PRK for everything.
```
jgallagher, line 66: Maybe note that the keys of this map are epochs?

key-manager/src/lib.rs
```diff
@@ -0,0 +1,344 @@
...
+            ],
+            key.0.as_mut(),
+        )
+        .unwrap();
+
+        Ok(VersionedAes256GcmDiskEncryptionKey { epoch, key: Secret::new(key) })
+    }
+
+    /// Return the epochs for all secrets which are loaded
+    pub fn loaded_secrets(&self) -> Vec<u64> {
```
jgallagher, line 131: Nit - this method name sounds like it returns the secrets themselves. `epochs_of_loaded_secrets()` is a mouthful - maybe just `epochs()`?

key-manager/src/lib.rs
andrewjstone: It's definitely intentional. I'm also not sure why you'd want a `Debug` impl here. This is a secret and we don't want to log it by accident. The secret is further protected in the `Secret` type below.

key-manager/src/lib.rs
andrewjstone: Will do

key-manager/src/lib.rs
andrewjstone: Good point. I'll rename. `loaded_epochs` is probably a good middle ground.

key-manager/src/lib.rs
andrewjstone: Comment updated in [75d92cf](https://github.com/oxidecomputer/omicron/pull/2990/commits/75d92cf2095a16ce622234f5c06c16d232e4a712)

key-manager/src/lib.rs
andrewjstone: Fixed in [75d92cf](https://github.com/oxidecomputer/omicron/pull/2990/commits/75d92cf2095a16ce622234f5c06c16d232e4a712)

key-manager/src/lib.rs
andrewjstone: So one other thing here. The secrets themselves are meant to be transient, and shouldn't be stored in any longer lived data structure. The IKM gets created once and put in a secret, where it is fed into Hkdf and discarded.

andrewjstone: It's a good point. I'll just derive it to avoid the comment and problems :)

andrewjstone: > It's a good point. I'll just derive it to avoid the comment and problems :)

I'll **impl** it, not derive. Sigh.

key-manager/src/lib.rs
```diff
@@ -0,0 +1,346 @@
...
+use hkdf::Hkdf;
+use secrecy::{ExposeSecret, Secret};
+use sha3::Sha3_256;
+use sled_hardware::DiskIdentity;
+use zeroize::{Zeroize, ZeroizeOnDrop};
+
+/// Input Key Material
+///
+/// This should never be used directly, and always wrapped in a `Secret<Ikm>`
+/// upon construction. We sparate the two types, because a `Secret` must contain
```
smklein, line 20: If this is true, why expose this type as `pub`? Why not just define a pub type alias to the secret, and keep this private?

key-manager/src/lib.rs
andrewjstone: Fixed in 6abfb00

key-manager/src/lib.rs
```diff
@@ -0,0 +1,355 @@
...
+pub enum Error {
+    #[error("Secret not loaded for {epoch}")]
+    SecretNotLoaded { epoch: u64 },
+
+    #[error("Failed to retreive secret: {0}")]
+    SecretRetreival(#[from] SecretRetrieverError),
+}
+
+/// Derived Disk Encryption key
+#[derive(Zeroize, ZeroizeOnDrop, Default)]
```
smklein, line 58: Dumb question, but why should this derive default? Seems potentially dangerous to make it "easy" to hand out invalid keys

key-manager/src/lib.rs
```diff
@@ -0,0 +1,355 @@
...
+/// A mechanism for retrieving a secrets to use as input key material to HKDF-
+/// Extract.
+#[async_trait]
+pub trait SecretRetriever {
+    /// Return the latest secret
+    ////
+    /// This is useful when a new entity is being encrypted and there is no need
+    /// for a reconfiguration. When an entity is already encrypted, and needs to
+    /// be decrypted, the user should instead call the [`SecretRetriever::get`].
+    async fn get_latest(&self) -> Result<VersionedIkm, SecretRetrieverError>;
```
smklein, line 186: Out of curiosity, why isn't this returning an epoch? (... which could potentially be passed to `Self::get`?)

What happens if this is called while a reconfiguration is happening?

key-manager/src/lib.rs
andrewjstone: Great call. This should not be `pub` at all. Neither should `Aes256GcmDiskEncryptionKey`. I made both of them private and added accessors to their versioned counterparts which are the types a user sees.

andrewjstone: Fixed in [79cf577](https://github.com/oxidecomputer/omicron/pull/2990/commits/79cf577eda5264af8622c8e48299c706b2a89a8e)

key-manager/src/lib.rs
andrewjstone: I actually removed the `Debug` impl again and instead made this type private in [79cf577](https://github.com/oxidecomputer/omicron/pull/2990/commits/79cf577eda5264af8622c8e48299c706b2a89a8e)

key-manager/src/lib.rs
andrewjstone: `VersionedIkm` contains the epoch.

What happens if this is called when reconfiguration is happening is the same as when it isn't. The latest secret will be returned as that's what the caller asked for. In general this should only be used for configuring a new device and so the latest is what the user actually wants. They have no use for older secrets.

In general, we can't deal with reconfiguration entirely within this part of the code. Even the call to `get` is only optimistic. There's an inherent race condition where a reconfiguration can be started after the caller gets the values of secrets back. It's the responsibility of the system supplying the input key material to ensure that that is fine and keys can be rotated one after another for any still available epochs (up to a policy limit).  In trust quorum, where we expect to use this, the sled-agent will get informed of a new configuration and trigger a rekey again.

jgallagher (approved):

PR #2993, opened 2023-05-03
[sled-agent] Improve error message on ledger write failure

smklein:
Namely, identify the paths which we tried to access.

These would typically be "both mountpoints within M.2s", though it may be informative if we only try one (or none!) of them.

### DIFF
--- a/sled-agent/src/ledger.rs
+++ b/sled-agent/src/ledger.rs
@@ -25,7 +25,7 @@ pub enum Error {
     },
 
     #[error("Failed to write the ledger to storage (tried to access: {failed_paths:?})")]
-    FailedToAccessStorage { failed_paths: Vec<PathBuf> },
+    FailedToAccessStorage { failed_paths: Vec<(PathBuf, Error)> },
 }
 
 impl Error {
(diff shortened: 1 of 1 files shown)


davepacheco:
This looks good; can we also put the error message (or one of them) into the error that we return?

smklein:
> This looks good; can we also put the error message (or one of them) into the error that we return?

Sure, it's a tuple of (PathBuf, Error) now.

davepacheco (approved):
Thanks for adding that!  What's it look like now?

smklein:
> Thanks for adding that! What's it look like now?

This is the error I'm seeing in the test that removes the directories for both configs:

```
Failed to write the ledger to storage (tried to access: [("/tmp/.tmpJW4l0D/ledger.toml", Io { message: "Error accessing /tmp/.tmpJW4l0D/.ledger.toml.tmp", err: Os { code: 2, kind: NotFound, message: "No such file or directory" } }), ("/tmp/.tmp91lA2G/ledger.toml", Io { message: "Error accessing /tmp/.tmp91lA2G/.ledger.toml.tmp", err: Os { code: 2, kind: NotFound, message: "No such file or directory" } })])
```

Issue #2984, opened 2023-05-02
Hardware id fields should be deterministic


zephraph:
## Problem

As @rmustacc outlined in [RFD-360](https://rfd.shared.oxide.computer/rfd/0360#_hardware_uuids) there's an issue w/ using Nexus generated UUIDs to refer to hardware components in that we may need to look a component up.

>  At no point can a request be initiated with a Nexus-generated UUID without having previously gotten it. Which doesn’t work as part of how the sled starts up and fundamentally makes it hard to use for any kind of fault management as these errors can begin to happen before we know this.

## Proposed solution

To solve this within the current constraints of our system I propose that we use UUIDv5 as specified in [RFC 4122](https://www.rfc-editor.org/rfc/rfc4122#section-4.3) (and supported by the UUID crate) to deterministically generate a UUID from the set of hardware identifiers relevant to the device in question. 

## Justification

We've had a stated desire to have richer identifiers like `disk:nvme:<vendor id>:<part name>:<serial>`. Holistically a richer identification system that communicates semantic (and/or deterministic) context to the user is highly desirable. Unfortunately that still constitutes a broad effort that'll take time to implement. I have hope that we can deliver on that functionality in a non-breaking manner but I don't see it as a viable short term path.

There's also the case to be made that consistency provides a valuable intuition when working with the product. All of our first class API resources have IDs as a viable method of address. Having a set of resources that _don't_ expose ID would be a potential UX stumbling block that seems worthwhile to avoid.

zephraph:
This issue won't be completed as described. I do still _strongly_ believe that having reliable ways to reference hardware is incredibly important despite the plethora of issues. See #3008 for more details of what I tried for the initial implementation.

On May 4th, 2023 we had a meeting titled _The Serial Number Paradoxes_ in which @rmustacc, @jclulow, @jmpesp, @davepacheco, and I all got together to discuss the broader topic of having reliable hardware identifiers. For folks who stumble on this issue later, I recommend you catch up on that meeting.

- [Transcript](https://docs.google.com/document/d/161NHG7JLaZjWRbmrT077aMtW_w8oeLA_MJOpZUHZ9Tw/edit#heading=h.cz6t8mzhkjpk) 🔒 
- [Video](https://drive.google.com/file/d/1LPygwvjaBAVeDo-1XitPKpw6lOAg_1s5/view) 🔒
[...]

PR #2996, opened 2023-05-03
Outline basic switch api

zephraph:
As noted in the review of #2645, we need a top level switch API that we'll use for inventory and for associating sub resources like switch ports to. 

This PR outlines the most basic, incomplete form of that API. There's likely a lot more to be done just from a completeness perspective, but once I get this green I'd like to merge to be able to iteratively add functionality in conjunction with #2645


zephraph:
Required followup work:
- Switches should be registered during rack initialization
- We should have some mechanism to update their state when they're removed and reinserted. 
- Should be associated with switch-ports once #2645 is merged

rcgoodfellow:
Thanks for putting this together @zephraph. I have a few comments/concerns about the switch UUID direction.

While I understand having switches referenced by a UUID is likely a good underlying data model, I do think we'll need a way to refer to racks and switches that relates to their physical points of presence.

For example, updating the peering configuration on port 7, switch 1, rack la4-1701-301 - where la4 indicates a COLO facility in Los Angeles, 1701 is a cage number in that facility, and 301 is a rack position within that cage - clearly indicates what's going on, whether in communications between operators, or in scripts/configuration files for deploying Oxide systems. Contrast that with updating the peering configuration on port 7, switch 93F473A3-C3AA-443D-A07C-66258C14D6C8, rack E221C0E2-E63E-4302-AC05-24E1B377A7DF.

On a separate note I think switch UUIDs should be stable relative to the containing rack id and switch location. The primary distinguishing characteristic of a switch is what it's connected to. If a rack switch is disconnected, replaced and then reconnected in the same configuration as before the replacement - it's functionally the same switch and having it's id change would be a) confusing, b) require transitive of updates referring objects for no other purpose than id shifting and, c) require any customer scripts / automation / supervisory systems to be updated when no actual functional changes took place.

zephraph:
TL;DR: Yes I agree but we can't implement most of that right now and this PR is mostly a bookkeeping stub for the switch representation.

---

Yeah, there's a lot to unpack here @rcgoodfellow.

First off, we have no current way of mapping a physical rack to a logical rack. E.g. the piece of metal that has a serial number on it has no deterministic mapping to the rack database entry. There's no electrical identification on the rack it self to provide persistent identity and we have no rack identity quorum at the moment. The only reason this isn't a major problem today is because we're only doing single rack deployments so there will only be one entry in the rack table. We'll need to figure this out.

I was of a similar mindset that all hardware UUIDs should be stable (though I was thinking about it more based off of its serial/part number rather than rack and position). I opened #2984 and subsequently #3008 to try to get there. Folks had some concerns about that approach so we had a meeting on _the serial number paradox_ on May 4th. Here's the video and transcript:

https://docs.google.com/document/d/161NHG7JLaZjWRbmrT077aMtW_w8oeLA_MJOpZUHZ9Tw/edit?usp=drive_web
https://drive.google.com/file/d/1LPygwvjaBAVeDo-1XitPKpw6lOAg_1s5/view?usp=drive_web

I don't have a ton of clarity on the other side of the meeting except to establish a few facts:
- We can't rely on hardware identifiers being unique so we can't employ hard constraints
- There probably will end up being multiple db entries for the same piece of hardware over time. We'll need to figure out how to reconcile that. 

I hear you on wanting stable IDs. I want them too. I don't have the knowledge or context to get us there from here. @jclulow and @davepacheco have some ideas that area.

rcgoodfellow (approved):

PR #2999, opened 2023-05-03
[sled-agent] Stop using `/var/oxide` (on ramdisk), start using `config` dataset (on M.2)

smklein:
This PR finds any spot where we use `/var/oxide` -- a configuration directory which exists on the ramdisk -- and replaces it with storage in the M.2 datasets.

Fixes https://github.com/oxidecomputer/omicron/issues/2970
Fixes https://github.com/oxidecomputer/omicron/issues/2998


sled-agent/src/bootstrap/agent.rs
```diff
@@ -965,40 +976,34 @@ impl Agent {
...
-            trust_quorum_share: Option<SerializableShareDistribution>,
-        }
-
-        let mut out = String::with_capacity(128);
-        let serializer = toml::Serializer::new(&mut out);
-        PersistentSledAgentRequestDef::serialize(self, serializer)?;
-        Ok(out)
+impl<'a> Ledgerable for PersistentSledAgentRequest<'a> {
+    fn is_newer_than(&self, _other: &Self) -> bool {
+        true
```
smklein, line 989: @jgallagher Would appreciate your input here.

This is a type which we really need to serialize and deserialize to/from TOML to put it in persistent storage.

I'm using a utility called `Ledger` to help out with the "write back to both M.2s, cope if only one is actually on-line" logic".

This utility requires that `Ledgerable` types implement `Serialize` and `Deserialize`, which makes this type a pain-in-the-butt, since it wasn't previously Serializable.

jgallagher: I don't think we want to store trust quorum shares in the ledger, but since we don't really have trust quorum shares anyway... How painful would it be to remove the `trust_quorum_share` field from `PersistentSledAgentRequest` altogether?

smklein: I thought this was where we *should* store trust quorum shares when they're ready though? I can rip this out, but I'd also probably ripping out a lot more of the foundation of where trust quorum shares are stored. Can give this a shot in a follow-up commit.

smklein: From chat:

If we pull on this thread and don't store tq shares, then a major portion of the bootstrap server also needs to be re-worked. This may or may not be what we wanna do, but it should probably be decoupled from this PR either way.

andrewjstone: I mentioned this to Sean in chat, but I think it's fine to move everything including keyshares over to the ledger right now. The protocol for TQ is going to change a bunch and will require a bunch of work in the bootstrap server as mentioned. It makes sense to decouple to a separate PR.

common/src/lib.rs
```diff
@@ -48,12 +48,3 @@ macro_rules! generate_logging_api {
...
 }
-
-/// Location on internal storage where sled-specific information is stored.
-///
-/// This is mostly private to the `omicron-sled-agent` crate, but exists in
-/// common so it may be cleared by the installation tools.
-///
-/// NOTE: Be careful when modifying this path - the installation tools will
-/// **remove the entire directory** to re-install/uninstall the system.
-pub const OMICRON_CONFIG_PATH: &'static str = "/var/oxide";
```
smklein, line 59: I also grepped for it, and didn't find any references to `/var/oxide` that exist within the global zone.

sled-agent/src/bootstrap/agent.rs
```diff
@@ -374,41 +365,61 @@ impl Agent {
...
+                    info!(agent.log, "Found boot disk M.2: {disk:?}");
+                    break;
+                }
+                None => {
+                    info!(agent.log, "Waiting for boot disk M.2...");
+                    tokio::time::sleep(core::time::Duration::from_millis(250))
+                        .await;
+                }
+            }
+        }
```
smklein, line 411: This is new. Without this, it's possible that we simply don't catch our own bootdisk, due to the storage manager lagging behind.

I could possibly update this to less of a "retry-based" interface, but it seems to work.

sled-agent/src/bootstrap/agent.rs
```diff
@@ -827,28 +845,39 @@ impl Agent {
         &self,
         _state: &tokio::sync::MutexGuard<'_, SledAgentState>,
     ) -> Result<(), BootstrapError> {
-        tokio::fs::remove_dir_all(omicron_common::OMICRON_CONFIG_PATH)
+        let config_dirs = self
+            .storage_resources
+            .all_m2_mountpoints(sled_hardware::disk::CONFIG_DATASET)
```
smklein, line 850: This is *basically* the same code it was before, but we're removing everything within the dataset, rather than deleting the dataset itself.

sled-agent/src/bootstrap/hardware.rs
```diff
@@ -186,6 +191,18 @@ impl HardwareMonitor {
 
         let storage_manager = StorageManager::new(&log).await;
 
+        // If our configuration asks for synthetic zpools, insert them now.
```
smklein, line 194: Now that the bootstrap agent has a dependency on M.2-based storage, we rely on these synthetic zpools being inserted earlier.

sled-agent/src/ledger.rs
```diff
@@ -58,15 +61,26 @@ pub struct Ledger<T> {
 }
 
 impl<T: Ledgerable> Ledger<T> {
+    /// Creates a ledger with a new initial value, ready to be written to
```
smklein, line 64: I sorta needed to change these APIs for the RSS markers to work okay. Basically, if there was no file, I wanted to *know* about it, rather than getting a "default" value back. 

This PR removes the constraint that `Ledgerable` also implement `Default`, and updates these constructors to cope.

smklein:
This is ready for review!

sled-agent/src/bootstrap/agent.rs
```diff
@@ -965,40 +994,33 @@ impl Agent {
     }
 }
 
-// We intentionally DO NOT derive `Debug` or `Serialize`; both provide avenues
+// We intentionally DO NOT derive `Debug``; it provides avenues
```
jgallagher, line 997: I'd maybe replace this with something like

```suggestion
// TODO-correctness: remove / rework trust quorum share storage
```

since we're not really caring about storage of these shares (which are effectively always `None` anyway).

jgallagher (approved):

jgallagher (approved):

sled-agent/src/bootstrap/agent.rs
andrewjstone: This makes sense to me. Seems fine to retry, since we need M.2s to boot.

andrewjstone (approved):

sled-agent/src/bootstrap/agent.rs
smklein: Done

PR #3004, opened 2023-05-03
[wicketd] move artifact store to within update tracker

sunshowers:
* Don't allow uploading a new repository to the update tracker while an
  update is in progress. This simplifies the UI and fixes a bug where
  multiple trampoline images can be uploaded at the same time and race
  against each other.
* Reset the update tracker when a new repository is uploaded.

Test this by adding some functionality to UpdateTracker to fake an
update, so we can test some of the possible interleavings.


wicketd/src/update_tracker.rs
```diff
@@ -268,8 +368,47 @@ impl UpdateTracker {
...
+    sp_update_data: BTreeMap<SpIdentifier, SpUpdateData>,
+}
+
+impl UpdateTrackerData {
+    fn new(artifact_store: WicketdArtifactStore) -> Self {
+        Self { artifact_store, sp_update_data: BTreeMap::new() }
+    }
+
+    fn put_repository(&mut self, bytes: BufList) -> Result<(), HttpError> {
+        // Are there any updates currently running? If so, then
```
jgallagher, line 383: > If so, then

then what?! :joy:

jgallagher (approved):

wicketd/src/update_tracker.rs
sunshowers: Fixed, thanks haha

PR #3005, opened 2023-05-03
fix spurious metrics test failures

davepacheco:
Fixes #2994.  See https://github.com/oxidecomputer/omicron/issues/2994#issuecomment-1533786110 for details.

I'm going to see if the instances test failure looks similar.

PR #3006, opened 2023-05-04
more first-class environment setup

davepacheco:
I like the new `tools/setup_env.sh`.  Here I'm taking a few steps to first-class it a bit more:

- it should now work to add the various binaries to your PATH regardless of where you run it from (i.e., you don't have to run it from the root of the workspace)
- the CI scripts use it rather than hardcoding the same paths, which hopefully will help make sure this stays up to date
- I moved it to the root and called it `env.sh`, only because I've seen this convention before.  If others have other preferences I'm happy to leave it as it is or call it something else.


davepacheco:
One test run seems to have hit #3005: https://github.com/oxidecomputer/omicron/pull/3006/checks?check_run_id=13220424276
I'm going to rerun it.

docs/how-to-run-simulated.adoc
```diff
@@ -30,6 +30,13 @@ Prerequisite software may be installed with the following script:
 $ ./tools/install_builder_prerequisites.sh
 ----
 
+Now put these on your PATH with:
+
+[source,text]
+----
+$ source env.sh
+----
+
```
internet-diglett, line 39: We might want to advise running this _before_ installing prerequisites, as I think the script will bail out early if the path isn't already set up?

internet-diglett (approved):
LGTM, just one comment

PR #3007, opened 2023-05-04
[sled-agent] Encrypt U.2 zpool top-level datasets

andrewjstone:
Encryption of U.2 devices relies on key generated by the `KeyManager`
introduced in https://github.com/oxidecomputer/omicron/pull/2990. The `KeyManager` is parameterized by a
`SecretRetriever`, which provides input key material (IKM) that the
`KeyManager` can use to derive keys. In this commit we introduce a
`LocalSecretRetriever` which provides secrets local to a sled. Right
now the secret is hardcoded, but we intend to either store random
secrets on the M.2s or retrieve them from VPD data in the short
term. Longer term, we will replace the `LocalSecretRetriever` with a
`TrustQuorumSecretRetriever` which will provide a rack secret recomputed
from trust quorum key shares.

The `KeyManger` has been  extended to run in an tokio task and process
requests for from different "requesters". Each requester only has the
ability to request keys that it requires to do its job. Currently there
is a single requester, `StorageKeyRequester`, that is owned by the
`StorageManager` which loans it out to `sled_hardware::Disk` to request
keys on demand.

`sled_hardware::Disk` requests keys for both creation of encrypted
datasets and mounting of existing encrypted datasets. In order to
actually allow ZFS to use the generated keys, it creates keyfiles inside
a RAMDisk. Attempts are made to limit the lifetime of these files as
well as accidental leakage of secrets.

Changes have also been made to `illumos_utils::Zfs` to support
encrytpion. To support future secret reconfiguration with trust quorum
we store an additional `epoch` property on each encrypted dataset, which
allows the `KeyManager` to retrieve the right vesion of the secret from
a `SecretRetriever`.


andrewjstone:
This is mostly ready for review. We can create encrypted filesystems and mount them after a reboot. There are a few things that still need doing:
 - [ ] Rebase on master
 - [ ] Add some extra logging
 - [ ] Move all zone filesystems under `<uuid>/crypt`

sled-hardware/src/disk.rs
```diff
@@ -317,23 +381,82 @@ impl Disk {
...
+            let epoch =
+                if let Ok(epoch_str) = Zfs::get_oxide_value(dataset, "epoch") {
+                    if let Ok(epoch) = epoch_str.parse::<u64>() {
+                        epoch
+                    } else {
+                        return Err(DiskError::MissingEpochProperty(
+                            dataset.to_string(),
+                        ));
+                    }
+                } else {
```
andrewjstone, line 418: Add a comment that the filesystem doesn't exist at this point, so we want to use the secret from the latest epoch to create it

sled-hardware/src/disk.rs
```diff
@@ -387,6 +510,52 @@ impl From<ZpoolKind> for DiskVariant {
...
+        Ok(KeyFile { path, file, log: log.clone() })
+    }
+
+    // It'd be nice to `impl Drop for `KeyFile` and then call `zero`
+    // from within the drop handler, but async `Drop` isn't supported.
+    pub async fn zero_and_unlink(&mut self) -> std::io::Result<()> {
+        let zeroes = [0u8; 32];
+        let _ = self.file.seek(SeekFrom::Start(0)).await?;
+        self.file.write_all(&zeroes).await?;
+        info!(self.log, "Zeroed keyfile {}", self.path);
```
andrewjstone, line 549: Change this log message to indicate deletion?

illumos-utils/src/zfs.rs
```diff
@@ -183,6 +211,45 @@ impl Zfs {
...
+        let cmd = command.args(&[ZFS, "mount", "-l", name]);
+        execute(cmd).map_err(|err| EnsureFilesystemError {
+            name: name.to_string(),
+            mountpoint: mountpoint.clone(),
+            err: EnsureFilesystemErrorRaw::MountEncryptedFsFailed(err),
+        })?;
+        Ok(())
+    }
+
+    // Return true if the dataset exists, with an optional epoch if there is one.
```
andrewjstone, line 228: This comment is out of date. We don't return an epoch here anymore.

key-manager/src/lib.rs
```diff
@@ -96,22 +162,65 @@ pub struct KeyManager<S: SecretRetriever> {
...
+
+        (key_manager, storage_key_requester)
+    }
+
+    /// Run the main receive loop of the `KeyManager`
+    ///
+    /// This should be spawned into a tokio task
+    pub async fn run(&mut self) {
+        loop {
+            tokio::select! {
```
andrewjstone, line 194: This `tokio::select` is here because we will have some other types of key requesters other than `StorageKeyRequester` in the future. We also may want a sled-agent requester that is allowed to clear keys rather than use a timer. Having the select doesn't really hurt if we don't use it, but I'm open to removing it if its confusing and we don't add anything else to select on in this PR.

key-manager/src/lib.rs
```diff
@@ -162,7 +272,11 @@ impl<S: SecretRetriever> KeyManager<S> {
     /// TODO(AJS): From what I can tell, the internal PRKs inside
     /// Hkdf instances are not zeroized, and so will remain in memory
     /// after drop. We should fix this one way or another.
-    pub fn clear(&mut self) {
+    ///
+    /// TODO(AJS): We should have a timeout mechanism so that we can clear
```
andrewjstone, line 276: Alternatively we can have sled-agent perform the clear with it's own `KeyManager` requester.

sled-hardware/src/disk.rs
```diff
@@ -341,23 +404,91 @@ impl Disk {
...
+            let keypath: Keypath = disk_identity.into();
+
+            let epoch =
+                if let Ok(epoch_str) = Zfs::get_oxide_value(dataset, "epoch") {
+                    if let Ok(epoch) = epoch_str.parse::<u64>() {
+                        epoch
+                    } else {
+                        return Err(DiskError::MissingEpochProperty(
+                            dataset.to_string(),
+                        ));
```
smklein, line 439: This error seems a little misleading - the property exists, it just can't be parsed.

sled-hardware/src/disk.rs
```diff
@@ -341,23 +404,91 @@ impl Disk {
...
+                    }
+                } else {
+                    // We assume the error indicates that the dataset doesn't
+                    // exist here. Use the latest secret to create it. If the
+                    // error did not actually indicate that the dataset didn't
+                    // exist the creation will fail, so we don't need to do any
+                    // special handling.
+                    key_requester.load_latest_secret().await?
```
smklein, line 447: What happens if this assumption is wrong? For example, if we have a dataset where the epoch property has been deleted?\

sled-hardware/src/disk.rs
```diff
@@ -415,6 +546,53 @@ impl From<ZpoolKind> for DiskVariant {
...
+            .write(true)
+            .open(&path.0)
+            .await?;
+        file.write_all(key).await?;
+        info!(log, "Created keyfile {}", path);
+        Ok(KeyFile { path, file, log: log.clone() })
+    }
+
+    // It'd be nice to `impl Drop for `KeyFile` and then call `zero`
+    // from within the drop handler, but async `Drop` isn't supported.
```
smklein, line 581: I don't actually think this does exactly what you want on most systems.

- If the filesystem is copy-on-write, this would just write new pages out, rather than actually removing old ones
- When we have swap, I think this file could be swapped back out to disk, unless `MAP_NORESERVE` was explicitly set on the region of memory used to hold the file

All this is to say: What do you think is the effect of writing zeroes out to the file before unlinking, as opposed to just unlinking the keyfile?

sled-hardware/src/disk.rs
```diff
@@ -4,15 +4,22 @@
...
 use illumos_utils::zpool::ZpoolKind;
 use illumos_utils::zpool::ZpoolName;
+use key_manager::StorageKeyRequester;
 use slog::Logger;
 use slog::{info, warn};
+use tokio::fs::{remove_file, File};
+use tokio::io::{AsyncSeekExt, AsyncWriteExt, SeekFrom};
 use uuid::Uuid;
 
+pub const KEYPATH_ROOT: &str = "/var/run/oxide/";
```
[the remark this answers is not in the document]
smklein: Reading below: Seems like this would be a significant problem if this was lost between boots?

smklein (commented):
Looks like a great start, some comments below (especially re: cold-boot usage)

illumos-utils/src/zfs.rs
```diff
@@ -40,6 +40,9 @@ enum EnsureFilesystemErrorRaw {
 
     #[error("Unexpected output from ZFS commands: {0}")]
     Output(String),
+
+    #[error("Failed to mount encrypted filesystem")]
```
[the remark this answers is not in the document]
andrewjstone: Good catch! Will fix.

illumos-utils/src/zfs.rs
```diff
@@ -142,22 +161,16 @@ impl Zfs {
...
-                    mountpoint,
-                    err: EnsureFilesystemErrorRaw::Output(stdout.to_string()),
-                });
+        if Self::dataset_exists(name, &mountpoint)? {
+            if encryption_details.is_none() {
+                // If the dataset exists, we're done.
+                return Ok(());
+            } else {
+                // We need to load the encryption key and mount the filesystem
+                return Self::mount_encrypted_dataset(name, &mountpoint);
```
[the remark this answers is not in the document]
andrewjstone: The keypath is stored as a property on the dataset. Same with encryption type, and a few other related parameters.

key-manager/src/lib.rs
```diff
@@ -77,6 +79,70 @@ impl VersionedAes256GcmDiskEncryptionKey {
...
+///
+/// We provide `From<sled_hardware::DiskIdentity> for <key_manager::DiskIdentity>`
+/// in the `sled-hardware` crate, which is what uses `StorageKeyRequester` in the first
+/// place.
+#[derive(Debug, Clone)]
+pub struct DiskIdentity {
+    pub vendor: String,
+    pub serial: String,
+    pub model: String,
+}
```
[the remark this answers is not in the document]
andrewjstone: Sounds good to me! Will do.

sled-agent/src/bootstrap/hardware.rs
```diff
@@ -184,7 +186,22 @@ impl HardwareMonitor {
...
+        // TODO: Eventually other things are going to want access to "key
+        // requesters" returned by the key manager. We should probably move the
+        // creation and initialization upwards in the call stack along with the
+        // call to `StorageManager::new`.
+        let (mut key_manager, storage_key_requester) =
+            KeyManager::new(LocalSecretRetriever {});
+
+        // TODO: Should we keep a handle to the key_manager? It shouldn't ever
+        // actually be stopped.
+        tokio::spawn(async move { key_manager.run().await });
```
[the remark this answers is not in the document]
andrewjstone: Will do.

sled-hardware/src/disk.rs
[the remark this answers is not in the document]
andrewjstone: This is very much intentional. We want to limit the existence of the keyfile to the minimum amount of time necessary to either create the encrypted filesystem or to load the keys so we can mount it. After that we delete it. This key is actually just a wrapper around the internal ZFS key and we want to limit any extra mechanisms to access that.

The reason this works is that the `KeyManager` will re-derive the key from the secret for the given epoch on demand, and we rewrite the key file before we need it. I've tested this on cold-boot a number of times and it works. Since up until recently we didn't have persistent M.2 data, I had to redeploy after rebooting a sled and run RSS again. But since the zpools and datasets already exist, they just get mounted after the key is recreated. I verified they are not re-created by writing out a file to the encrypted dataset that I then read-back after reboot.

sled-hardware/src/disk.rs
andrewjstone: Perhaps I was not really clear in this comment. I'll rewrite it. The third sentence states that if the dataset does actually exist, the attempt to create it immediately below will fail.

key-manager/src/lib.rs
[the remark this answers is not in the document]
andrewjstone: Fixed in [83c825d](https://github.com/oxidecomputer/omicron/pull/3007/commits/83c825d37a57272cb551cf0b45fd354713fc1d78)

illumos-utils/src/zfs.rs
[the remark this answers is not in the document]
andrewjstone: Fixed in [83c825d](https://github.com/oxidecomputer/omicron/pull/3007/commits/83c825d37a57272cb551cf0b45fd354713fc1d78)

sled-agent/src/bootstrap/hardware.rs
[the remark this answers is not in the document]
andrewjstone: Fixed in [c618baa](https://github.com/oxidecomputer/omicron/pull/3007/commits/c618baaa143cf6e5975963bafdc0347eb81c6536)

sled-hardware/src/disk.rs
andrewjstone: For some reason I missed the second part of your sentence. If the epoch value was deleted, the call to `Zfs::get_oxide_value` will succeed. The epoch will just be represented by a `-` and won't be parsable.

I updated the comment in [580259d](https://github.com/oxidecomputer/omicron/pull/3007/commits/580259d853c2f1fa4668e040e074e6c6c8f0a0a7) to clarify.

sled-hardware/src/disk.rs
rmustacc: The suggestions I gave are specific to the use of a tmpfs and the fact that `/var/run` is a tmpfs (which means `/var/run/oxide` is as well unless something really weird is going on).

* There is no copy-on-write going on in tmpfs. We're just copying data into pages and overwriting them.
* It is strictly true that this could get paged out against swap. `MAP_NORESERVE` however doesn't prevent swap. It just says that make this without a reservation and if I can't then you are allowed to kill me. (It's also an mmap concept and not really applicable to a file in a file system). However, if it does, ultimately this is what encryption of the swap device is for. This problem isn't really specific to this, in my opinion. The same series of things could happen with sled agent or zfs(8)'s memory. To prevent swapping you need to use mlock or similar. However, while that prevents paging, it again is specific to a processes's memory not a file that we're trying to read from.
* The zeroing concept specifically on tmpfs means that when this page is released on a tmpfs file system back to system memory pool, it won't be lingering somewhere with the data. Getting at that page would imply having access to physical memory, so it's maybe not that important, but just a kind of belt and suspenders kind of thing I would do as a moral equivalent to an `explicit_bzero()` or similar going on. This is definitely specific to the implementation of tmpfs. If the page was reused for another process it'd definitely be zeroed.

andrewjstone: I added some comments regarding this in [b9fef27](https://github.com/oxidecomputer/omicron/pull/3007/commits/b9fef273b22da301e87dbe5e1e353e7316d1ee04)

andrewjstone:
@smklein I resolved all your actionable comments and replied to the rest. I also retested on paris, both initial boot and cold boot. The cold boot test was performed by writing a file named `hello.txt` with the contents `hello` to each crypto filesystem, then rebooting and running `OMICRON_NO_UNINSTALL=true ./omicron-package -t dogfood install` followed by:

```
BRM42220066 # for i in `zfs list -Ho mountpoint | grep crypt`; do ls $i; done
hello.txt  zone
hello.txt  zone
hello.txt  zone
hello.txt  zone
hello.txt  zone
hello.txt  zone
hello.txt  zone
hello.txt  zone
hello.txt  zone
BRM42220066 # for i in `zfs list -Ho mountpoint | grep -e crypt$`; do cat $i/hello.txt; done
hello
hello
hello
hello
hello
hello
hello
hello
hello
```
You can also see that the crypt filesystems are actually encrypted by doing the following, which I did after reboot.

```
BRM42220066 # zfs get all oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt
NAME                                            PROPERTY              VALUE                                                                    SOURCE
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  type                  filesystem                                                               -
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  creation              Mon Dec 29 22:51 1986                                                    -
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  used                  232K                                                                     -
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  available             2.82T                                                                    -
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  referenced            132K                                                                     -
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  compressratio         1.00x                                                                    -
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  mounted               yes                                                                      -
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  quota                 none                                                                     default
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  reservation           none                                                                     default
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  recordsize            128K                                                                     default
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  mountpoint            /pool/ext/016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt                     local
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  sharenfs              off                                                                      default
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  checksum              on                                                                       default
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  compression           off                                                                      default
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  atime                 on                                                                       default
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  devices               on                                                                       default
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  exec                  on                                                                       default
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  setuid                on                                                                       default
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  readonly              off                                                                      default
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  zoned                 off                                                                      default
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  snapdir               hidden                                                                   default
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  aclmode               discard                                                                  default
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  aclinherit            restricted                                                               default
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  createtxg             83600                                                                    -
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  canmount              on                                                                       default
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  xattr                 on                                                                       default
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  copies                1                                                                        default
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  version               5                                                                        -
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  utf8only              off                                                                      -
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  normalization         none                                                                     -
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  casesensitivity       sensitive                                                                -
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  vscan                 off                                                                      default
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  nbmand                off                                                                      default
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  sharesmb              off                                                                      default
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  refquota              none                                                                     default
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  refreservation        none                                                                     default
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  guid                  10976894679408763812                                                     -
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  primarycache          all                                                                      default
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  secondarycache        all                                                                      default
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  usedbysnapshots       0                                                                        -
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  usedbydataset         132K                                                                     -
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  usedbychildren        99.5K                                                                    -
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  usedbyrefreservation  0                                                                        -
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  logbias               latency                                                                  default
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  dedup                 off                                                                      default
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  mlslabel              none                                                                     default
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  sync                  standard                                                                 default
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  dnodesize             legacy                                                                   default
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  refcompressratio      1.00x                                                                    -
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  written               132K                                                                     -
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  logicalused           116K                                                                     -
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  logicalreferenced     66.5K                                                                    -
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  filesystem_limit      none                                                                     default
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  snapshot_limit        none                                                                     default
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  filesystem_count      none                                                                     default
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  snapshot_count        none                                                                     default
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  redundant_metadata    all                                                                      default
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  special_small_blocks  0                                                                        default
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  encryption            aes-256-gcm                                                              -
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  keylocation           file:///var/run/oxide/1b96-A079E7A6-WUS4C6432DSP3X3-zfs-aes-256-gcm.key  local
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  keyformat             raw                                                                      -
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  pbkdf2iters           0                                                                        default
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  encryptionroot        oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt                           -
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  keystatus             available                                                                -
oxp_016dfed7-8c6b-43f7-882c-f11dbaac4e9e/crypt  oxide:epoch           0                                                                        local
```

andrewjstone:
I just remembered that https://github.com/oxidecomputer/omicron/pull/3007 adds some support for persisting zone filesystems under `crypt/zone`. 

This code has been ready to go for a few weeks now and has been tested numerous times. I've done cold boot testing on paris with it, but since there is no persistent install of software on the M.2s like dogfood, I have to re-run `omicron-package install` after reboot to rediscover and mount the U.2 encrypted datasets. The PR has the details about this. While I'd like to test with a proper install on dogfood I"m at about the point where I'd like to just merge this and see what happens.

PR #3009, opened 2023-05-04
Support upserting multiple non-gimlet sled agents

### LINKED ISSUE #3003: Multiple non-gimlet sled agents cannot register themselves to Nexus
non-gimlets have the following [baseboard information](https://github.com/oxidecomputer/omicron/blob/adcc588c92bf422af38981f39f2d6907949e1578/sled-hardware/src/illumos/mod.rs#L634):

```
"baseboard":"Baseboard { identifier: \"Unknown\", model: \"Unknown\", revision: 0 }"
```

Which due to the `sled` table constraint:

```
    -- This constraint should be upheld, even for deleted disks
    -- in the fleet.
    CONSTRAINT serial_part_revision_unique UNIQUE (
      serial_number, part_number, revision
    )
```

means that multiple non-gimlet sleds cannot register themselves, erroneously returning that the sled was already registered:

```
root@[fd00:1122:3344:101::6]:32221/omicron> \x
root@[fd00:1122:3344:101::6]:32221/omicron> select * from sled;
-[ RECORD 1 ]
id                      | 8e67787c-e844-4740-a5a2-f28d11df0b9e
time_created            | 2023-05-03 21:20:10.40225+00
time_modified           | 2023-05-03 21:20:10.409566+00
time_deleted            | NULL
rcgen                   | 5
rack_id                 | 17c3d49b-b0e3-4364-b3b5-4fc0d2b24008
is_scrimlet             | false
serial_number           | Unknown
part_number             | Unknown
revision                | 0
usable_hardware_threads | 32
usable_physical_ram     | 137346322432
ip                      | fd00:1122:3344:101::1
port                    | 12345
last_used_address       | fd00:1122:3344:101::21


Time: 1ms total (execution 0ms / network 0ms)
```

Sled `f4a81217-3d17-4bff-8a1e-541b2fe3220e` cannot register:

```json
{
  "msg": "failed to notify nexus about sled agent",
  "v": 0,
  "name": "SledAgent",
  "level": 40,
  "time": "2023-05-03T21:32:55.469497786Z",
  "hostname": "kibblesnbits",
  "pid": 21385,
  "sled_id": "f4a81217-3d17-4bff-8a1e-541b2fe3220e",
  "component": "SledAgent",
  "total duration": "859.231836618s",
  "error": "Error Response: status: 400 Bad Request; headers: {\"content-type\": \"application/json\", \"x-request-id\": \"9ddfe6ef-d7f2-4341-8298-6ea730edbac7\", \"content-length\": \"175\", \"date\": \"Wed, 03 May 2023 21:32:55 GMT\"}; value: Error { error_code: Some(\"ObjectAlreadyExists\"), message: \"already exists: sled \\\"f4a81217-3d17-4bff-8a1e-541b2fe3220e\\\"\", request_id: \"9ddfe6ef-d7f2-4341-8298-6ea730edbac7\" }"
}
```

Until `Baseboard::unknown` is removed (there's an XXX saying as much?), it should generate a random serial number so that non-gimlet testing can be performed.

jmpesp:
Sled baseboard is used as a unique identifier for sleds, but for i86pc this was set to ("unknown", "unknown", 0), and this caused sled upsert to fail due to "object already exists".

Change Baseboard from a struct to an enum, preserving the type of the baseboard along with its identifying information. Serialize this to a file for wicketd to consume instead of supplying it via command line arguments.

Generate a unique serial number for i86pc baseboards during hardware view snapshot.

Fixes #3003


sled-hardware/src/lib.rs
```diff
@@ -73,48 +73,76 @@ pub enum SledMode {
...
+#[derive(
+    Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize, JsonSchema,
+)]
+pub enum Baseboard {
+    Gimlet { identifier: String, model: String, revision: i64 },
+
+    Unknown,
+
+    Pc { identifier: String, model: String, revision: i64 },
 }
```
zephraph, line 85: It's not clear to me why it's worth diverging from other implementations of baseboard when both arms of the enum have the same shape?

jmpesp: Really, only to preserve the type information for later pattern matching. I could have hard coded `Pc`'s return values in the model and revision impl functions for Baseboard but thought it would make more sense as configurable fields.

jgallagher (approved):
Sorry for the delay reviewing this; I wanted to test it on `madrid`. Have done so and everything LGTM.

PR #3015, opened 2023-05-04
[illumos-utils] Better dataset deletion handling

smklein:
- if `rpool/zone` doesn't exist, ignore it
- if any datasets cannot be queried, throw an error more explicitly


andrewjstone (approved):

leftwo (approved):
dogfood approved

PR #3016, opened 2023-05-04
Remove the factory dataset from m.2s

smklein:
As discussed on 5/4, this shouldn't be necessary. We're just going to store zone filesystems within (1) the ramdisk, and (2) the `install` dataset


smklein:
I'll update RFD 118 too after this merges

andrewjstone (approved):

smklein:
> I'll update RFD 118 too after this merges

This is done

jgallagher (approved):

PR #3019, opened 2023-05-04
[wicketd] Update the RoT

jgallagher:
The major changes in this PR are:

* Teach `tufaceous-lib` how to unpack an RoT artifact into its A/B images (I chose the names in this artifact completely arbitrarily; hubris currently doesn't produce a combo artifact like this)
* Change `wicketd`'s update plan to track separate A/B RoT images for each component type
* Flesh out the steps in `wicketd`'s update spec to actually update the RoT

with some minor supporting changes in MGS.


tufaceous-lib/src/artifact.rs
```diff
@@ -332,7 +391,84 @@ fn read_entry<R: io::Read>(
     Ok(buf.into())
 }
 
+/// Represents RoT A/B hubris archives.
+///
+/// RoT artifacts are actually tarballs, with both A and B hubris archives
+/// inside the. This code extracts those archives out of the tarballs.
```
sunshowers, line 397:
```suggestion
/// inside them. This code extracts those archives out of the tarballs.
```

sunshowers (approved):
Looks fantastic, thank you so much for doing this!

wicketd/src/artifacts.rs
```diff
@@ -806,22 +868,61 @@ mod tests {
         RandomHostOsImage { phase1, phase2, tarball }
     }
 
+    struct RandomRotImage {
+        archive_a: Bytes,
+        archive_b: Bytes,
+        tarball: Bytes,
+    }
+
+    fn make_random_rot_image() -> RandomRotImage {
```
sunshowers, line 877: I feel like this can potentially be unified with the rot image generator in tufaceous-lib, though I definitely don't want to block on this. (Might have to refactor the code in tufaceous-lib a little.)

jgallagher: Yeah, definitely agreed. I have a tentative plan to make `tufaceous-lib` a little more knowledgeable about RoT and host composite artifacts, which would let us clean up both this and `make_random_host_image()` on which this is based.

PR #3020, opened 2023-05-04
validate recovery user password hash when parsing RSS config

### LINKED ISSUE #3000: Can get all the way to attempting handoff to Nexus with a password that will be rejected
If a user has set `recovery_silo.user_password_hash` in their `config-rss.toml` with a password that will be rejected by Nexus, it won't be rejected until the end of RSS during the handoff to Nexus:

```json
{
  "msg": "Failed to handoff to nexus: Error Response: status: 400 Bad Request; headers: {\"content-type\": \"application/json\", \"x-request-id\": \"44075f34-7836-44d0-98a0-e20aea9c51eb\", \"content-length\": \"216\", \"date\": \"Wed, 03 May 2023 20:10:46 GMT\"}; value: Error { error_code: None, message: \"unable to parse JSON body: recovery_silo.user_password_hash: password hash: algorithm: expected argon2id, found argon2i at line 1 column 5205\", request_id: \"44075f34-7836-44d0-98a0-e20aea9c51eb\" }",
  "v": 0,
  "name": "SledAgent",
  "level": 30,
  "time": "2023-05-03T20:10:46.279576554Z",
  "hostname": "frostypaws",
  "pid": 7736,
  "component": "RSS"
}
```

Sled-agent will loop here until cancelled, having already performed all the other necessary steps for RSS.

I think we should be rejecting this earlier, for example when deserializing config-rss.toml, or when receiving the POST /rack-initialize. We shouldn't be able to start if we will eventually fail to complete.

davepacheco:
Fixes #3000.

Background: the Nexus "rack-initialize" request accepts a password hash for the initial recovery user.  We validate (during serde deserialization of the type that we use to represent the password hash) that the hash meets some basic requirements (e.g., it's Argon2id with strong parameters).  In practice, that value is received by RSS and passed through to Nexus.  But RSS does not do any of this validation of the hash.  This leads to an annoying user experience where you don't find out until late in the game (and only by scanning logs) that the hash you provided won't work.

This change causes RSS to validate the password hash when it receives the request.  This is basically just moving some code around.  There are three main pieces to it:

1. Move the `nexus-passwords` crate (from "nexus/passwords") to `omicron-passwords` ("passwords" at the root).  This is needed so that RSS can access the validation logic in that crate.  Note that in the real product, presumably this hash will come from Wicket or Wicketd hashing a user-entered password, so we were already going to need to make this crate available to more than just Nexus.
2. Move the type that validates a password hash string from nexus/types into omicron-passwords.  This is just a newtype around `password_hash::PasswordHashString` that provides serde Serialize/Deserialize + JsonSchema, with the Deserialize impl validating the properties I mentioned above.  (The validation logic was already in omicron-passwords.  It's just the newtype that moved here.)
3. In nexus-client (the progenitor-generated client for the Nexus internal API, which RSS uses to make this request), we override the generated type for `NewPasswordHash` to use the type directly from `omicron-passwords`.  The reason: the definition of the type that RSS uses for _its_ rack-initialize request directly uses the one from nexus-client.  So the net result is that the type used in RSS to accept a password hash is the same type that Nexus uses to accept the password hash, both of them coming from omicron-passwords and validating the password hash.

I split this into two commits: the first commit moves/renames the crate (and is mechanical, brought to you by sed(1)); the second commit does the rest.

I also extended the test that parses our in-repo config-rss.toml files to test this case.  So in fact if you update config-rss.toml with a bad hash, this fails `cargo test`:

```
---- bootstrap::params::tests::parse_rack_initialization stdout ----
thread 'bootstrap::params::tests::parse_rack_initialization' panicked at 'failed to parse "/home/dap/omicron-dns/sled-agent/../smf/sled-agent/non-gimlet/config-rss.toml": TOML parse error at line 42, column 22
   |
42 | user_password_hash = "$argon2i$v=19$m=16,t=2,p=1$aGZ4V042T1RDRjhQaWM3NQ$Zwi6nqjvFgzy83qWrf+gMA"
   |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
password hash: algorithm: expected argon2id, found argon2i
', sled-agent/src/bootstrap/params.rs:180:33
```

Finally, it's worth noting that there's a sense in which this approach works by accident: Nexus and RSS are not guaranteed to be built from the same repo version, and it's conceivable that the Nexus present at runtime enforces different constraints than RSS does.  In practice this seems very unlikely, and the improvement in user experience seems well worth the risk here.


jmpesp (approved):
Thanks!

PR #3021, opened 2023-05-05
[sled-agent] Look for zone images in the boot M.2's install directory

smklein:
For most services: Look in the M.2's install directory after looking in "/opt/oxide" when searching for zone images.

Note that propolis images are an exception - we only look for them in the ramdisk.

Fixes https://github.com/oxidecomputer/omicron/issues/2971


illumos-utils/src/running_zone.rs
```diff
@@ -564,6 +564,9 @@ pub enum InstallZoneError {
         #[source]
         err: crate::zone::AdmError,
     },
+
+    #[error("Failed to find zone image '{image}' from `{paths:?}'")]
```
jgallagher, line 568: Nit - mismatched quotes; maybe
```suggestion
    #[error("Failed to find zone image '{image}' from '{paths:?}'")]
```

or even

```suggestion
    #[error("Failed to find zone image '{image}' from {paths:?}")]
```

since the debug impl will enclose it in brackets?

jgallagher (approved):

illumos-utils/src/running_zone.rs
smklein: Done

Issue #3018, opened 2023-05-04
DNS zones should provision storage in a dataset explicitly


smklein:
## How Omicron manages storage

Zones generally store data in one of two ways:

- "Anywhere within their zone filesystem" - this is the case for data such as logs, tmp directories within zones, etc
- "A durable dataset mounted at a well-known location" - this is the case for zones like Crucible, CockroachDB, and Clickhouse, which explicitly manage storage that should be durable across reboots

## How DNS manages storage

The internal and external DNS services are configured to use the "anywhere within their zone filesystem" storage.

https://github.com/oxidecomputer/omicron/blob/7e8273e754c4deb83df90a89204d67fc392431f4/smf/internal-dns/config.toml#L14
https://github.com/oxidecomputer/omicron/blob/7e8273e754c4deb83df90a89204d67fc392431f4/smf/external-dns/config.toml#L14

**In general**, this data acting as a cache from Nexus is fine - after all, Nexus periodically updates these values, bumps a generation number, and should ensure that a (redundant!) number of DNS servers are storing this information.

**However** in the case of cold boot, it's pretty important that **at least one** internal DNS server boots, to provide the necessary machinery for:
- The CockroachDB instances to boot up and find each other, and
- Also for Nexus to be able to find CockroachDB

## Why would this be a problem

@jclulow has identified that the "service model" for zones means that we should be able to delete/recreate zone filesystems arbitrarily, and recreate them with persistent dataset, to recreate a stable state after cold boot.

Personally, I'm on board with this, as it means there's less "reconstruction" logic the Sled Agent needs to perform after reboot (e.g., parsing of vnics, IP interfaces, etc). We can just set everything as temporary, and recreate what we need, relying only on datasets to be durable.

smklein:
FYI @davepacheco - I can take a stab at this, but I wanted to give you a heads-up

davepacheco:
Thanks for looking at this.  I'd been wondering about this.

The intent was always that the DNS data was on persistent storage so that it could come up in a functional state without any other dependencies.  (This is mentioned in RFD 248.)  However, another important assumption is that the persistent storage is tied to the image -- i.e., the server would never find stuff in /var/oxide/dns that was written by a different version of the server.  I know we've discussed this in the context of update and I don't recall where we landed.  (We obviously _could_ support that, but it's a lot more work, both up front and ongoing, and I think it's more brittle, riskier upgrade process.)

jclulow:
[...]

PR #3022, opened 2023-05-05
[nexus][sled-agent] DNS services are backed by explicit datasets

smklein:
This PR ensures that DNS services are explicitly backed by datasets. After this PR, it should be possible
to "delete and re-initialize" arbitrary zones, as long as they are re-initialized with their associated datasets and requested configuration.

NOTE: Within the Sled Agent, "datasets" and "services" are treated differently. Requesting datasets additionally can cause a dataset to be formatted within a zpool, in addition to launching a zone-based service - since the logic here was initially different, the sled agent exposed different endpoints for each case. This PR crosses that line a bit, by converting two entities (InternalDns, ExternalDns) that were initially "services" to become a "datasets". In the future, these concepts should be merged (or at least, the "services managing dataset" should be pulled into the "services" table properly). As this PR crosses that line, it contributes to that "merging" slightly.

Fixes https://github.com/oxidecomputer/omicron/issues/3018


bnaecker:
I'm planning to take this one on starting today. First step is merging in `main`...will it blend?

smklein:
Merged in https://github.com/oxidecomputer/omicron/pull/3390

Issue #2978, opened 2023-05-01
Nexus should pass request for "all datasets on a sled" as a single endpoint with a generation number


smklein:
https://github.com/oxidecomputer/omicron/pull/2972#issuecomment-1530453752 for some context, but it basically would let the "local ledger of services" on a sled be more cleanly a cache of data from CRDB.

davepacheco:
I think this is obviated now that datasets and services are the same API and we already have #2977.  Is that right @smklein?

smklein:
Unfortunately still missing a generation number

davepacheco:
But is it now covered by #2977 since datasets and services are the same endpoint now?

davepacheco:
I think this is obviated by #4466.

PR #3025, opened 2023-05-05
[sled-agent] Vectorize the request for dataset zones

smklein:
This is necessary to unify the interface between "storage" and "non-storage" zones.

Part of https://github.com/oxidecomputer/omicron/issues/2978


smklein:
Merged in https://github.com/oxidecomputer/omicron/pull/3390

Issue #2973, opened 2023-05-01
Sled Agent service launching should include automatic retries


smklein:
- RSS / Nexus both tell sled agent to launch services
- If we succeed when starting a service, the sled agent stores "here are the services I should launch on cold boot" in the M.2 configuration dataset
- On reboot, we read this list of services from the M.2 and try to start 'em.

**RIGHT NOW** if we fail to start those services, we don't try again.

This is a bummer! Especially with the dependencies on U.2s - which might take a sec to load - we should add a retry-with-backoff to service launching, kinda like we already do for the NTP and Switch zones.

davepacheco:
I was trying to figure out why this could fail in a retryable way because that seemed surprising.  Thanks for the link to the comment in #2946 that provides more context.  I wonder if it would make sense to have Sled Agent track the fact that a zone depends on a ZFS pool and avoid trying to start the zone until that pool becomes available.  My worry with this approach is the same as always with retry+backoff: that it retries when it's not helpful to (because the pool still isn't available), doesn't retry when it _would_ be helpful (because the pool has just been imported but we're in a backoff), and potentially papers over some other kind of failure (e.g., an actual _bug_ that will cause a zone never to be able to start up will not be visible until we exhaust the retry policy, after which point we may have clobbered lots of useful state by retrying a bunch of times).

(Sorry if I'm missing more relevant context here.)

smklein:
Agreed, and I'll rename the name of this bug to be slightly less prescriptive about the solution.

I think that it's viable to set up a system where we only attempt the service launch once the appropriate datasets from U.2s are loaded, but it might require more coordination between the ServiceManager and StorageManager. As you say, however, that mitigates a lot of the downsides that would be encountered with an arbitrary retry.

Issue #2614, opened 2023-03-18
nexus crashes when it can't reach internal DNS on startup


davepacheco:
Found this in CI (originally [this job](https://github.com/oxidecomputer/omicron/pull/2570/checks?check_run_id=12095990773)):

```
{"msg":"Setting up resolver on subnet: Ipv6Subnet { net: Ipv6Net(Ipv6Network { addr: fd00:1122:3344::, prefix: 48 }) }","v":0,"name":"nexus","level":30,"time":"2023-03-17T20:19:23.752265804-07:00","hostname":"oxz_nexus","pid":7362,"component":"ServerContext","name":"f836cbe7-f69c-4642-b314-6f6c026276e7"}
{"msg":"Accessing DB url from DNS","v":0,"name":"nexus","level":30,"time":"2023-03-17T20:19:23.752407736-07:00","hostname":"oxz_nexus","pid":7362,"component":"ServerContext","name":"f836cbe7-f69c-4642-b314-6f6c026276e7"}
nexus: Failed to lookup IP: request timed out
[ Mar 17 20:19:38 Stopping because all processes in service exited. ]
[ Mar 17 20:19:38 Executing stop method (:kill). ]
[ Mar 17 20:19:38 Restarting too quickly, changing state to maintenance. ]
```

[rendered](https://buildomat.eng.oxide.computer/wg/0/artefact/01GVS99JKN9TBC2RQ6RAYC9YME/NiQKHUrSKQSHYjZ7KskdFFCN3Dyuzf04V85QDBlcUziSibSA/01GVS9A8ERXB2F7PN1TKRS1KP1/01GVSBR2X28PT0GFVH4ZT00NDJ/system-illumos-nexus:default.log?format=x-bunyan#L204):

```
[2023-03-17T20:19:23.752265804-07:00]  INFO: f836cbe7-f69c-4642-b314-6f6c026276e7/ServerContext/7362 on oxz_nexus: Setting up resolver on subnet: Ipv6Subnet { net: Ipv6Net(Ipv6Network { addr: fd00:1122:3344::, prefix: 48 }) }
[2023-03-17T20:19:23.752407736-07:00]  INFO: f836cbe7-f69c-4642-b314-6f6c026276e7/ServerContext/7362 on oxz_nexus: Accessing DB url from DNS
nexus: Failed to lookup IP: request timed out
[ Mar 17 20:19:38 Stopping because all processes in service exited. ]
[...]

PR #3026, opened 2023-05-05
[sled-agent] Unify requests for services and services backed by datasets

smklein:
Previously, the sled agent had to request for "services using datasets" and "services not using datasets" with different APIs - one acting on `/filesystem`, and one acting on `/services`. These APIs evolved independently, even though they have significant overlap in utility.

By merging the two of them together, several aspects of service management are simpler:
- Storing "config for what services should run" can be stored in a single config file on the M.2s
- "The set of every service that should run on a sled" can be tagged with a generation number by Nexus
- A smaller API surface is exposed, with less overlap

TODO: I think this will be blocked on:
- [x] https://github.com/oxidecomputer/omicron/issues/2614
- [ ] and possibly https://github.com/oxidecomputer/omicron/issues/2973


smklein:
Merged in #3390

PR #3028, opened 2023-05-05
Add swadm & ddmadm to default PATH in switch zone.

luqmana:
Bringing up dogfood rack still requires some manual config in the switch zone so let's make it a bit easier.

### DIFF
--- a/package-manifest.toml
+++ b/package-manifest.toml
@@ -273,6 +273,7 @@ source.paths = [
   { from = "smf/switch_zone_setup/manifest.xml", to = "/var/svc/manifest/site/switch_zone_setup/manifest.xml" },
   { from = "smf/switch_zone_setup/switch_zone_setup", to = "/opt/oxide/bin/switch_zone_setup" },
   { from = "smf/switch_zone_setup/support_authorized_keys", to = "/opt/oxide/support/authorized_keys" },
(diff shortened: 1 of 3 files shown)


smf/switch_zone_setup/switch_zone_setup
```diff
@@ -21,4 +21,7 @@ done
 pfexec routeadm -e ipv6-forwarding -u
 pfexec routeadm -e ipv6-routing -u
 
+# Add tools like swadm & ddmadm to the PATH by default
+echo 'export PATH=$PATH:/opt/oxide/dendrite/bin:/opt/oxide/mg-ddm/bin' >> ~root/.profile
```
jclulow, line 25: I suspect we should just ship a `/root/.profile` file in the zone image that meets your expectations, rather than edit it at runtime.

If you're going to edit it like this, we should at least make it idempotent: if this service runs more than once, it will add this line more than once to the file, and `PATH` will get longer and longer.

luqmana: Added it as part of the switch zone image in [58487c3](https://github.com/oxidecomputer/omicron/pull/3028/commits/58487c33f15841e2750eb2ca3130072eb8c392a5).

citrus-it (approved):
Seems good to me. The default `.profile` is all comments so replacing it wholesale like this isn't a problem.

PR #3037, opened 2023-05-07
[update-engine] add the name of the spec to each event

sunshowers:
This makes it easy to look at a generic event and know which spec it
should be converted to.


update-engine/src/events.rs
```diff
@@ -807,6 +823,15 @@ impl<S: StepSpec> StepOutcome<S> {
 #[serde(bound = "", rename_all = "snake_case")]
 #[schemars(rename = "ProgressEventFor{S}")]
 pub struct ProgressEvent<S: StepSpec> {
+    /// The specification that this event belongs to.
+    ///
+    /// This is typically the name of the type `S` for which `StepSpec` is
+    /// implemented.
+    ///
+    /// This can be used with `Self::from_generic` to deserialize generic
+    /// metadata.
```
jgallagher, line 832: >    /// This can be used with `Self::from_generic` to deserialize generic
>    /// metadata.

Could you expand on this a bit?

jgallagher (approved):

update-engine/src/events.rs
sunshowers: Done, added a basic example.

PR #3051, opened 2023-05-08
[tufaceous-lib] More structured support for composite artifacts

jgallagher:
We previously had some support for two kinds of composite artifacts:

* Host/Trampoline OS images (phase1/phase2)
* RoT archives (A/B images)

but it had grown a bit organically. This PR attempts to add more structured
support for defining composite artifacts in the manifest used by
`tufaceous assemble`, and adds support for the control plane composite
artifacts (which holds an arbitrary number of control plane zone
tarballs).

This is partly followup from #3019 (specifically https://github.com/oxidecomputer/omicron/pull/3019#discussion_r1186772275) and partly to move toward installinator understanding how to write real control plane artifacts.

### DIFF
--- a/tufaceous-lib/src/artifact.rs
+++ b/tufaceous-lib/src/artifact.rs
@@ -3,7 +3,7 @@
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 use std::{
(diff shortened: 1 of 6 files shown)


sunshowers (approved):
Looks great!

PR #3055, opened 2023-05-09
[wicket] a new way to display update logs

sunshowers:
This PR bundles up several changes:

1. A number of improvements to `update-engine`'s `EventBuffer` which
   make it easier to interrogate the status of individual steps or
   executions.
2. Improvements to the wicket frontend.

There is still work to be done on the wicket frontend (see
https://github.com/oxidecomputer/minimum-upgradable-product/issues/78):

* Provide the ability to reset the state for a failed update.
* Do we need a way to hard cancel running updates? That would leave the
  update in an undefined state if not done carefully, but we might need
  to provide a way to do it to resolve a stuck update. (Though maybe we
  do it via the support account instead.)

## Screenshots

Running update:

image

Failed update:

image

Progress for running step:

image

Status for completed step:

image

Status for failed step (**note this still has cut-off text**, this is tracked in https://github.com/oxidecomputer/minimum-upgradable-product/issues/79):

image

Depends on #3054.


wicket/src/ui/panes/update.rs
```diff
@@ -42,14 +49,22 @@ pub struct UpdatePane {
     /// capture all state.
     tree_state: TreeState,
     items: Vec<TreeItem<'static>>,
+
+    // Per-component update state that isn't serializable.
```
andrewjstone, line 53: This is going to prevent recording snapshots from ever working if we exceed our number of max events.  This is why @jgallagher had to increase the buffer size so much the other day, which may not be feasible for production workloads. We don't really know how long wicket is going to be open or how many events to expect and we don't have unlimited memory.

In effect adding more state here prevents us from being able to use this in production. If you notice above, there's a comment about moving `tree_state` into global `State`. I'd really prefer we don't add more state here. Is there a way that you can make this serializable and add it to the global `State`? 

Really, there shouldn't be any type of state maintained here. This likely also means moving the `Rect`s out into `State`. We have to do this if we ever support resize events in recordings because we will lose the resize events when we take a snapshot.

sunshowers: In a meeting we decided to not block on this. @andrewjstone filed https://github.com/oxidecomputer/omicron/issues/3070 to track this.

update-engine/src/buffer.rs
```diff
@@ -253,153 +277,409 @@ impl<S: StepSpec> EventStore<S> {
...
+        self.event_tree.add_node(EventTreeNode::Step(key))
+    }
+
+    fn mark_step_key_completed(
+        &mut self,
+        root_key: StepKey,
+        info: CompletionInfo,
+    ) {
+        if let Some(value) = self.map.get_mut(&root_key) {
+            // Completion status only applies to the root key.
```
jgallagher, line 482: I found this comment misleading at first - my misinterpretation was "why does completion only apply to the root key if we're marking anything reachable from it complete too?", missing the importance of `status`. Maybe expand with something like this?

```rust
// Completion status only applies to the root key. Nodes
// reachable from this node are still marked as complete but
// without status.
```

jgallagher (approved):
Excited to see this land!

update-engine/src/buffer.rs
```diff
@@ -165,65 +217,42 @@ impl<S: StepSpec> EventStore<S> {
...
-                            // of this function.
-                            debug_assert!(
-                                false,
-                                "This branch cannot be reached"
-                            );
-                        }
-                    }
+        if let Some(key) = actions.step_key {
+            if let Some(value) = self.map.get_mut(&key) {
+                if event.kind.priority() == StepEventPriority::High {
```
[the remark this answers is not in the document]
sunshowers: Good idea, done.

update-engine/src/buffer.rs
sunshowers: Good catch, thanks.

update-engine/src/buffer.rs
```diff
@@ -142,18 +167,45 @@ impl<S: StepSpec> EventBuffer<S> {
...
 #[derive_where(Clone, Debug, Default)]
 struct EventStore<S: StepSpec> {
     // A tree where edges are from parent event keys to child nested event keys.
     //
     // While petgraph seems like overkill at first, it results in really
     // straightforward algorithms below compared to alternatives like storing
     // trees using Box pointers.
-    event_tree: DiGraphMap<StepKey, ()>,
-    map: HashMap<StepKey, EventMapValue<S>>,
+    event_tree: DiGraphMap<EventTreeNode, ()>,
```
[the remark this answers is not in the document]
sunshowers: Done, added a little ASCII art graph.

PR #3058, opened 2023-05-09
[console] cache assets in browser for a year, index.html not at all

david-crespo:
The assets have content hashes in the filenames, so whenever there's a new version, `index.html` (which is not cached at all) will point to the new filename, making the cache irrelevant.

Decided to hard code the cache time on the assets. I don't see a point in plumbing it through as a configurable value.

```
dist
├── assets
│  ├── app-2b12dbb1.js
│  ├── app-2b12dbb1.js.map
│  ├── favicon-f85aa1d6.svg
│  ├── GT-America-Mono-Medium-3af01ada.woff
│  ├── GT-America-Mono-Medium-14eb7e1f.woff2
│  ├── GT-America-Mono-Regular-OCC-4a8bb2aa.woff2
│  ├── GT-America-Mono-Regular-OCC-9b5fa807.woff
│  ├── index-1b4af6c1.css
│  ├── oxide-hero-rack-fa9186cf.webp
│  ├── SuisseIntl-Book-WebS-a34a1356.woff
│  ├── SuisseIntl-Book-WebS-fdc357d4.woff2
│  ├── SuisseIntl-Light-WebS-02ceb8ff.woff
│  ├── SuisseIntl-Light-WebS-d4cb666c.woff2
│  ├── SuisseIntl-Regular-WebS-9b09c5ee.woff
│  ├── SuisseIntl-Regular-WebS-bfa8da72.woff2
│  ├── SuisseIntl-RegularItalic-WebS-3bd20b80.woff2
│  ├── SuisseIntl-RegularItalic-WebS-d96d0ab7.woff
│  ├── Terminal-93061908.js
│  ├── Terminal-93061908.js.map
│  ├── TimeSeriesChart-0ea03c61.js
│  └── TimeSeriesChart-0ea03c61.js.map
└── index.html
```

### DIFF
--- a/nexus/src/external_api/console_api.rs
+++ b/nexus/src/external_api/console_api.rs
@@ -723,7 +723,8 @@ fn with_gz_ext(path: &PathBuf) -> PathBuf {
 /// Fetch a static asset from `<static_dir>/assets`. 404 on virtually all
 /// errors. No auth. NO SENSITIVE FILES. Will serve a gzipped version if the
 /// `.gz` file is present in the directory and `Accept-Encoding: gzip` is
-/// present on the request.
+/// present on the request. Cache in browser for a year because assets have
+/// content hash in filename.
 #[endpoint {
    method = GET,
    path = "/assets/{path:.*}",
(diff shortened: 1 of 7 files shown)


nexus/src/external_api/console_api.rs
```diff
@@ -809,7 +805,8 @@ pub async fn serve_console_index(
     Ok(Response::builder()
         .status(StatusCode::OK)
         .header(http::header::CONTENT_TYPE, "text/html; charset=UTF-8")
-        .header(http::header::CACHE_CONTROL, cache_control_value(apictx))
+        // do not cache this response in browser
+        .header(http::header::CACHE_CONTROL, "no-store")
         .body(file_contents.into())?)
 }
```
zephraph, line 811: Eventually it would be nice to have etag support for this at least so we could 304 if it hasn't changed.

zephraph (approved):
Assuming we have a strategy to resolve the upgrade conflicts, I'm good with this change 👍

david-crespo:
Definitely a problem, but I don't think the append-only thing solves it. As long as there are Nexus instances that haven't been updated, there is always a chance of being routed to a Nexus instance that doesn't have your file. I also don't think this PR really changes the situation around this problem, either.

PR #3077, opened 2023-05-10
[installinator] Write control plane zone files to M.2s

jgallagher:
Also adds some support code in `tufaceous-lib` to unpack the composite control plane artifact.

### DIFF
--- a/installinator-common/src/progress.rs
+++ b/installinator-common/src/progress.rs
@@ -66,6 +66,8 @@ pub enum InstallinatorComponent {
 pub enum InstallinatorStepId {
     Download,
     Format,
+    Scan,
+    UnpackControlPlaneArtifact,
     Write,
 }
 
@@ -103,6 +105,16 @@ pub enum InstallinatorProgressMetadata {
 #[derive(Clone, Debug, Eq, PartialEq, Deserialize, Serialize, JsonSchema)]
 #[serde(rename_all = "snake_case", tag = "reason")]
 pub enum InstallinatorCompletionMetadata {
+    HardwareScan {
+        /// Number of disks found.
+        disks_found: usize,
+    },
+
+    ControlPlaneZones {
+        /// Number of zone images that will be installed.
+        zones_to_install: usize,
+    },
+
     Download {
         /// The address the artifact was downloaded from.
         address: SocketAddrV6,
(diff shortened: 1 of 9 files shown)


installinator/src/write.rs
```diff
@@ -32,12 +35,13 @@ struct ArtifactDestination {
...
+    // Path to write the host image; either a raw device corresponding to an M.2
+    // partition (real gimlet) or a file (test).
     host_phase_2: Utf8PathBuf,
 
-    // TODO-completeness This SHOULD NOT be optional, but at the time of this
-    // writing we don't know how to write the control plane artifacts on a real
-    // gimlet, so we leave it optional for now. This should be fixed very soon!
-    control_plane: Option<Utf8PathBuf>,
+    // Directory in which we unpack the control plane zone images.
+    control_plane_dir: Utf8PathBuf,
```
sunshowers, line 44: 🥳

installinator/src/write.rs
```diff
@@ -448,32 +452,80 @@ impl ArtifactsToWrite<'_> {
...
+impl ControlPlaneZoneWriteContext<'_> {
+    fn register_steps<'b>(
+        &'b self,
+        engine: &UpdateEngine<'b, ControlPlaneZonesSpec>,
+        transport: &'b mut impl WriteTransport,
+    ) {
+        use update_engine::StepHandle;
+
+        let slot = self.slot;
+        let mut transport = StepHandle::ready(transport);
```
sunshowers, line 499: I don't think you need `StepHandle::ready` in this case (it's mainly useful for when you have conditional steps -- a step returns a value in some cases and not in others).

sunshowers (approved):
Looks great!

installinator-common/src/progress.rs
```diff
@@ -66,6 +66,8 @@ pub enum InstallinatorComponent {
 pub enum InstallinatorStepId {
     Download,
     Format,
+    Scan,
+    UnpackControlPlaneArtifact,
```
sunshowers, line 70: As discussed in our chat, it would be really nice to have a comment here explaining why we unpack control plane zones here rather than in wicketd (when we unpack other artifacts in wicketd).

installinator/src/write.rs
jgallagher: Ah, I should put a comment here. I need to do this because I can't share the `&mut impl WriteTransport` we get as an argument with every step, so instead I stash it into a `StepHandle` and reassign it on every loop iteration: each step passes it on to the next.

installinator-common/src/progress.rs
jgallagher: Added in 18175672

installinator/src/write.rs
jgallagher: Added an explanation in 18175672

sunshowers: Ah understood!

PR #3080, opened 2023-05-10
[nexus] Add `image_demote`

### LINKED ISSUE #2897: Want ability to demote a silo-level image
It may be worthwhile to have a way to roll back a promote action, in case user accidentally promotes the wrong image or finds issues after promotion and wants to pull it back into a project without having to delete and re-create the image.

benjaminleonard:
This is my first time adding something to nexus, so I've likely made some egregious error.

I haven't tested with image name since I assumed it wouldn't work without a project ID — if you could confirm @zephraph. Added tests to confirm an image is remove from silo images when demoted and added to the passed project.

Perhaps worthwhile adding a test to check that a user without sufficient permissions can demote an image.

Fixes #2897


david-crespo:
Am I right in reading this as letting you move the image into any project you have write access on? Does that sound a little weird? It means that in the UI we have to give the user a projects combobox to choose from at demote time.

I see that after promotion we do not keep track of the original project the image was created in — promotion sets the `project_id` to null — so given that, this has to work this way. My question is: do we want it to work this way?

zephraph:
Yeah @david-crespo, that's right. If there's any weirdness, it may just be the implications of the name. I don't think there's much value in restricting the process to be tied to a single project. Once we have an audit log you'll be able to track which project it was moved from. Locking it to a project wouldn't be great though, because that project could've been deleted since the promotion happened.

PR #3081, opened 2023-05-10
[wicket] correctly wrap popup text within step logs

sunshowers:
Previously, we were relying on tui's paragraph wrap functionality. This
had several limitations:

1. There's no way to obtain the actual height of the text as computed.
   This meant that wrapped text was previously cut off.
3. There's no sensible way to implement scrolling.

To address these issues, we implement wrapping ourselves, via the
textwrap library. The library exposes support for wrapping strings, but
we're working with tui Spans (which have styles associated with them) so
we have to do some work ourselves.  Thankfully, it isn't too much work:
textwrap exposes an abstract interface called `Fragment`, so it's mostly
a matter of copy-pasting some code from textwrap, and annotating each
word with its corresponding tui style.

This PR does not implement scrolling, but now that we're working with
pre-wrapped text in the popup, scrolling should be pretty simple to do
in a followup.

Screenshot of wrapped text:

image

### DIFF
--- a/wicket/src/ui/wrap.rs
+++ b/wicket/src/ui/wrap.rs
@@ -1,0 +1,357 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! This module is an adaptation of textwrap's `Word` library to work with
+//! `tui`'s `Span`s and lines.
+//!
+//! The code is mostly copy-pasted, with the following changes:
+//!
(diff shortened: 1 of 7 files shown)


wicket/src/ui/wrap.rs
```diff
@@ -0,0 +1,357 @@
...
+}
+
+impl<'a> Fragment for StyledWord<'a> {
+    fn width(&self) -> f64 {
+        // self.width is the display width, which is what we care about here.
+        self.width as f64
+    }
+
+    fn whitespace_width(&self) -> f64 {
+        // Since whitespace is always ASCII spaces, this is
```
jgallagher, line 298: this is... "equal to the number of whitespace characters", maybe?

jgallagher (approved):

PR #3088, opened 2023-05-11
allow locale and shell preference environment to pass through SSH

jclulow:
SSH is not enthusiastic in general about allowing random parts of the user's environment to leak in to a remote session.  We can relax that a little, though, to make sure that:

- the user's locale may be passed through correctly, which is important for correct output processing (e.g., it is the only way to know if the remote system is UTF-8 capable or not)
- we can pass in the shell editing mode preference of the user in `OXIDE_PREF_SHELL_MODE`, so that we can solve the problem of a mixed `set -o vi` and `set -o emacs` user population


andrewjstone (approved):
Thanks!

andrewjstone (approved):
Thanks!

PR #3089, opened 2023-05-11
ci: enable the crates.io sparse protocol

iliana:
This exports [`CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse`](https://blog.rust-lang.org/inside-rust/2023/01/30/cargo-sparse-protocol.html) across all of our CI jobs that touch cargo. As I'm about to add yet another job onto the very long chain of Buildomat jobs, I wanted to get an easy win in time spent waiting for CI.

| job | `main` sample, n=10 | this commit, n=1 |
| - | - | - |
| `helios / deploy`[^1] | 92s | [22s](https://buildomat.eng.oxide.computer/wg/0/details/01H05Y3TPX4PNE2YYHW1Y5THEW/82ZeGXmpOYTzxFzT63IOvzQdygjD5lCMlE5rJA2dhQwNKstf/01H05Y4366GZFN7RC3TPD9V09R#S490) |

The amount of time spent in this step depends on how recently [the git index](https://github.com/rust-lang/crates.io-index) was compacted; it was most recently compacted in April, so this is on the lower end of how long updating the crates.io index takes.

(I'm unsure yet if this is impacting build times for the OS image tasks.)

[^1]: Measured from first `+ ptime -m cargo run` to `Downloading crates ...`

### DIFF
--- a/.github/buildomat/jobs/build-and-test-linux.sh
+++ b/.github/buildomat/jobs/build-and-test-linux.sh
@@ -16,6 +16,7 @@ set -o xtrace
 
 cargo --version
 rustc --version
+export CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse
 
 #
 # Set up a custom temporary directory within whatever one we were given so that
--- a/.github/buildomat/jobs/build-and-test.sh
+++ b/.github/buildomat/jobs/build-and-test.sh
@@ -16,6 +16,7 @@ set -o xtrace
 
 cargo --version
 rustc --version
+export CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse
 
 #
 # Set up a custom temporary directory within whatever one we were given so that
--- a/.github/buildomat/jobs/build-end-to-end-tests.sh
+++ b/.github/buildomat/jobs/build-end-to-end-tests.sh
@@ -14,6 +14,7 @@ set -o xtrace
 
 cargo --version
 rustc --version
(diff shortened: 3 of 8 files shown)


jclulow:
Would it make sense to export this for all buildomat jobs?  Are there circumstances when one would not want this behaviour?

iliana:
I'm not aware of any circumstances where it would be the incorrect behavior, and inverting it is simple enough (`CARGO_REGISTRIES_CRATES_IO_PROTOCOL=git`).

I personally think this should be explicitly listed in our jobs up until we move to a toolchain that changes the default in case it causes problems and someone is trying to debug what's going on, but a ~1m-or-more improvement across all Rust jobs on Buildomat is probably better reason to export it globally.

iliana:
> I'm not aware of any circumstances where it would be the incorrect behavior

Actually now that I've thought about it again: specific stable/beta versions of cargo that _know about_ the `sparse` protocol but have it marked as unstable will no longer work if this is set globally by Buildomat. (This is the very situation that led me to bump the toolchain to 1.68 a while back, because I have it set in ~/.cargo/config lol)

That unfortunate state makes me think this should probably be a repo-level decision.

iliana:
Rust 1.70.0 is out, which enables sparse by default. I will probably bump rust-toolchain.toml instead.

iliana:
Obsoleted by #3274.

PR #3092, opened 2023-05-11
Sled/instance association for inventory and insights

zephraph:
This PR creates a view of instances that's exposed via the sleds endpoint to provide operators an insight into what's running on what hardware. This will aid in both understanding the service impact of maintenance as well as providing debugging aids.

We've made some explicit decisions to expose the project and instance names (but as little other details as possible).


common/src/sql/dbinit.sql
```diff
@@ -862,10 +863,43 @@ CREATE UNIQUE INDEX ON omicron.public.instance (
...
+JOIN
+    omicron.public.project as project
+ON  
+    instance.project_id = project.id
+JOIN
+    omicron.public.silo as silo
+ON  
+    project.silo_id = silo.id
+WHERE
+    instance.active_sled_id IS NOT NULL
```
luqmana, line 899: I'm a little inclined to say just remove this condition. I don't think this is ever `NULL` today once the instance has been created but there's an issue (#2315) to actually make it `Nullable/Option`. But even at then, `DataStore::sled_instance_list` already adds a filter `active_sled_id` matches the given `authz_sled`.

nexus/db-queries/src/db/lookup.rs
```diff
@@ -760,6 +760,15 @@ lookup_resource! {
     primary_key_columns = [ { column_name = "id", rust_type = Uuid } ]
 }
 
+lookup_resource! {
+    name = "SledInstance",
+    ancestors = [],
+    children = [],
+    lookup_by_name = true,
```
luqmana, line 767: Don't think I saw anywhere actually doing a lookup by name?

common/src/sql/dbinit.sql
```diff
@@ -862,10 +863,43 @@ CREATE UNIQUE INDEX ON omicron.public.instance (
...
+AND instance.time_deleted IS NULL;
```
david-crespo, line 900: Nitpick but everything after `FROM` is hard to read with all the newlines. I ran it through [`sqlfmt`](https://sqlfum.pt/) (which is by a former CRDB employee) and it came out with this:

```sql
FROM
    omicron.public.instance AS instance
    JOIN omicron.public.project AS project ON
            instance.project_id = project.id
    JOIN omicron.public.silo AS silo ON
            project.silo_id = silo.id
WHERE
    instance.active_sled_id IS NOT NULL
    AND instance.time_deleted IS NULL;
```

Do with that what you will.

nexus/db-queries/src/db/lookup.rs
zephraph: We're not, that's true! Good catch.

david-crespo (approved):
Looks great! You made it look easy and obvious, which I know it wasn't.

luqmana (approved):

PR #3095, opened 2023-05-12
TLS certificates should be Silo-scoped

### LINKED ISSUE #2367: tls: need per-DNS-name (per-Silo) configuration of certificates
Currently Nexus has external APIs for CRUD of "system-wide" x.509 TLS certificates.  Nexus uses one of these configured TLS certificates for the external API's HTTP server.

The problem: TLS certificates are generally associated with a specific DNS Name, and each Silo will have its own DNS name.  See RFD 357 for more on this.

I think the right approach here will be:

- These certificate management APIs should probably move under the Silo level and manage certificates _for a particular Silo_'s public endpoint.  There will be no system-wide certificates because there is no system-wide HTTPS endpoint.
- We need to use TLS SNI to select the right certificate for a particular incoming TCP connection.  I've confirmed that rustls supports this by providing a [`ResolvesServerCertUsingSni`](https://docs.rs/rustls/latest/rustls/server/struct.ResolvesServerCertUsingSni.html) when [building the `rustls::Config`](https://docs.rs/rustls/0.20.4/rustls/struct.ConfigBuilder.html#method.with_cert_resolver).
- Today, in `omicron_nexus::app::certificate::Nexus::get_nexus_tls_config()`, we load all the system-wide certificates, pick one, use it to make a new dropshot::ConfigTls with that certificate.  I think instead we're going to want to load every Silo's certificate (picking one for each Silo).  Then we're going to want a new `dropshot::ConfigTls` variant that lets Omicron specify the certificate resolver (or else provides the information needed for Dropshot to do that, which would be something like a map of server names to certificates).  I think much of the code around this won't need to change beyond that.

davepacheco:
Fixes #2367 (see that issue for details).

### DIFF
--- a/nexus/db-queries/src/authz/api_resources.rs
+++ b/nexus/db-queries/src/authz/api_resources.rs
@@ -498,6 +498,66 @@ impl AuthorizedResource for DeviceAuthRequestList {
     }
 }
 
+/// Synthetic resource describing the list of Certificates associated with a
+/// Silo
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct SiloCertificateList(Silo);
+
+impl SiloCertificateList {
+    pub fn new(silo: Silo) -> SiloCertificateList {
+        SiloCertificateList(silo)
+    }
+
+    pub fn silo(&self) -> &Silo {
+        &self.0
+    }
+}
+
+impl oso::PolarClass for SiloCertificateList {
+    fn get_polar_class_builder() -> oso::ClassBuilder<Self> {
+        oso::Class::builder()
+            .with_equality_check()
+            .add_attribute_getter("silo", |list: &SiloCertificateList| {
+                list.0.clone()
+            })
+    }
+}
+
+impl AuthorizedResource for SiloCertificateList {
+    fn load_roles<'a, 'b, 'c, 'd, 'e, 'f>(
+        &'a self,
+        opctx: &'b OpContext,
+        datastore: &'c DataStore,
+        authn: &'d authn::Context,
+        roleset: &'e mut RoleSet,
+    ) -> futures::future::BoxFuture<'f, Result<(), Error>>
+    where
+        'a: 'f,
+        'b: 'f,
+        'c: 'f,
+        'd: 'f,
+        'e: 'f,
+    {
+        // There are no roles on this resource, but we still need to load the
+        // Silo-related roles.
+        self.silo().load_roles(opctx, datastore, authn, roleset)
+    }
+
+    fn on_unauthorized(
+        &self,
+        _: &Authz,
+        error: Error,
+        _: AnyActor,
+        _: Action,
+    ) -> Error {
+        error
+    }
+
+    fn polar_class(&self) -> oso::Class {
+        Self::get_polar_class()
+    }
+}
+
 /// Synthetic resource describing the list of Identity Providers associated with
 /// a Silo
 #[derive(Clone, Debug, Eq, PartialEq)]
@@ -940,10 +1000,10 @@ authz_resource! {
 
 authz_resource! {
     name = "Certificate",
-    parent = "Fleet",
+    parent = "Silo",
     primary_key = Uuid,
     roles_allowed = false,
-    polar_snippet = FleetChild,
+    polar_snippet = Custom,
 }
 
 authz_resource! {
--- a/nexus/db-queries/src/authz/omicron.polar
+++ b/nexus/db-queries/src/authz/omicron.polar
@@ -204,6 +204,30 @@ has_relation(silo: Silo, "parent_silo", project: Project)
 # macro.  Some resources require custom Polar code.  Those appear here.
 #
 
+resource Certificate {
+	permissions = [
+	    "read",
+	    "modify",
+	    "create_child",
+	    "list_children",
+	];
+	relations = { parent_silo: Silo, parent_fleet: Fleet };
+
+	# Fleet-level and silo-level roles both grant privileges on certificates.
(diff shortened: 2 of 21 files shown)


nexus/db-queries/src/authz/api_resources.rs
```diff
@@ -498,6 +498,66 @@ impl AuthorizedResource for DeviceAuthRequestList {
     }
 }
 
+/// Synthetic resource describing the list of Certificates associated with a
+/// Silo
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct SiloCertificateList(Silo);
```
zephraph, line 504: I know we've talked about this a little bit, but I just want to clarify... is the reason this needs to have a synthetic resource because it's permissions differ in some meaningful way from other silo resources or is it because you think all silo resources (or every resource) should have a synthetic resource?

nexus/db-queries/src/authz/omicron.polar
```diff
@@ -387,6 +411,25 @@ resource DeviceAuthRequestList {
...
+# Describes the policy for creating and managing Silo certificates
+resource SiloCertificateList {
+	permissions = [ "list_children", "create_child" ];
+
+	relations = { parent_silo: Silo, parent_fleet: Fleet };
+
+	# Both Fleet and Silo administrators can see and modify the Silo's
+	# certificates.
+	"list_children" if "admin" on "parent_silo";
+	"list_children" if "admin" on "parent_silo";
```
zephraph, line 423: duplicate?

nexus/db-queries/src/authz/api_resources.rs
davepacheco: The obvious alternative would be directly checking "list children" on the Silo.  Compared to that, a synthetic resource gives us two things: first what you said that the policy is different for certificates than for, say, projects (both of which are Silo children); and second: the policy ("a person can list the certificates in the Silo if they are an administrator of the Silo's parent fleet") can be separated from the code that checks it (which just asks "can this person list the certificates in the Silo").  This makes the code easier to write and verify (since you're not making any judgments about the policy there) and the policy is easier to write and verify too (since you're not worrying about implementation details there).

nexus/db-queries/src/db/datastore/rack.rs
```diff
@@ -321,6 +312,31 @@ impl DataStore {
...
+                    })
+                    .collect::<Result<Vec<_>, _>>()
+                    .map_err(RackInitError::Certificate)
+                    .map_err(TxnError::CustomError)?;
+                {
+                    use db::schema::certificate::dsl;
+                    diesel::insert_into(dsl::certificate)
+                        .values(certificates)
+                        .on_conflict(dsl::id)
+                        .do_nothing()
```
zephraph, line 334: Does this mean there would be no error?

nexus/db-queries/src/db/lookup.rs
```diff
@@ -122,7 +122,7 @@ impl<'a> LookupPath<'a> {
             .opctx
             .authn
             .silo_required()
-            .internal_context("looking up Organization by name")
+            .internal_context("looking up Project by name")
```
zephraph, line 125: good catch!

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -2118,14 +2118,14 @@ async fn instance_disk_detach(
...
-/// Returns a list of all the system-wide certificates. System-wide certificates
-/// are returned sorted by creation date, with the most recent certificates
-/// appearing first.
+/// Returns a list of TLS certificates used for the external API (for the
+/// current Silo).  These are sorted by creation date, with the most recent
+/// certificates appearing first.
 #[endpoint {
     method = GET,
-    path = "/v1/system/certificates",
+    path = "/certificates",
```
zephraph: The `/v1/` must be retained.

zephraph (changes requested):
Overall it looks good. I did have a few questions about authz and noticed one duplicated line in the polar file. 

The `/v1/` prefix was accidentally removed when `/system` was taken out so you'll need to re-add that.

zephraph (changes requested):
Overall it looks good. I did have a few questions about authz and noticed one duplicated line in the polar file. 

The `/v1/` prefix was accidentally removed when `/system` was taken out so you'll need to re-add that.

davepacheco:
Thanks for the review.  It looks like I was fixing the /v1 thing as you were reviewing -- that should be fixed now.  I'll address the rest in a bit.

zephraph (approved):

nexus/db-queries/src/authz/omicron.polar
```diff
@@ -204,6 +204,30 @@ has_relation(silo: Silo, "parent_silo", project: Project)
...
+	    "create_child",
+	    "list_children",
+	];
+	relations = { parent_silo: Silo, parent_fleet: Fleet };
+
+	# Fleet-level and silo-level roles both grant privileges on certificates.
+	"read" if "admin" on "parent_silo";
+	"modify" if "admin" on "parent_silo";
+	"create_child" if "admin" on "parent_silo";
+	"list_children" if "admin" on "parent_silo";
```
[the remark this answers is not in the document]
davepacheco: It's probably not necessary.  Removed in af6d6d5decbd3ac3ab0e1315414b82dc6004e65a.

nexus/db-queries/src/db/datastore/rack.rs
davepacheco: Yeah.  The uuid was just generated, though, so it would be rather surprising if this happened.  But I'll go ahead and remove that.

Issue #3099, opened 2023-05-12
Would like to make an `oximeter` producer server with an existing logger


bnaecker:
We currently build `oximeter` producer servers using the [`Server::start()`](https://github.com/oxidecomputer/omicron/blob/f5c38cde1ed67d60c3931a8e92e590ee52fc712e/oximeter/producer/src/lib.rs#L54) method. That accepts configuration from which we build a logger, but not an actual logger. The latter is pretty useful, since you can chain key-value pairs from a parent logger or send logs to the same stream. This tracks adding such a constructor.

PR #3101, opened 2023-05-12
Allow building an oximeter producer using a concrete logger

bnaecker:
- Add `LogConfig` enum which supports either configuration for building a new logger, or an actual logger from which to create a child for the producer.
- Update consuming code.


bnaecker:
Resolves #3101

leftwo (approved):

zephraph:
@bnaecker the issue reference is recursive. I assume it should be #3099?

bnaecker:
Sure did, thanks!

PR #3110, opened 2023-05-13
Bump web console (skip empty chunks during image upload)

david-crespo:
Changes: https://github.com/oxidecomputer/console/compare/b3ddd888...ab407766


nexus/src/app/sagas/disk_create.rs
```diff
@@ -253,7 +253,7 @@ async fn sdc_alloc_regions(
         )
         .await
         .map_err(ActionError::action_failed)?;
-    Ok(datasets_and_regions)
+    Ok(dbg!(datasets_and_regions))
```
iliana, line 256: (this wasn't meant to be included, right?)

david-crespo: Aaaaaaa no. I use a script to make these PRs and I should make it fail if there are any unrelated changes hanging around. Or just don’t `git add --all` 🤦‍♂️

PR #3126, opened 2023-05-15
[update-engine] dedup remote nested events

### LINKED ISSUE #3102: wicket logs / EventReports after running multiple mupdates are enormous
Today when trying to update 6 gimlets in the dogfood rack simultaneously, I noticed a few oddities:

* After it had been running for a while, `wicket` became sluggish
* Within the switch zone, `/tmp/wicket.log` was several gigabytes, and growing quickly (~ 1GiB/minute, eyeballing it)
* From skimming the logs, the vast majority was debug prints of `EventReports`, many of which were extremely large (75,000+ lines), repeated very frequently

I think there are at least two things to address here:

1. Stop debug logging the full event reports
2. Why are the event reports so large?

For 2, the issue might be [where we generate the report](https://github.com/oxidecomputer/omicron/blob/857a9b95a83e7c833822fc29e3b89c4d1be41386/installinator/src/reporter.rs#L92); should that be `generate_report_since(self.last_reported)`?

sunshowers:
Deduplicate nested events coming from remote machines in a couple of spots:

1. When we're generating nested events, such that duplicate events don't make their way over the event channel. Maintain event buffers within `StepContext`s to handle that.
2. In `EventBuffer` itself, dedup events based on the leaf index rather than the root index. If two events have the same leaf event index and leaf execution ID, they can always be deduped.

While testing this, I also realized it would be really useful to add a `root_execution_id` to every event report: do so, and depend on it (note that you'll have to rebuild installinator with this PR to pick up support for that).

Finally, address the review comment in https://github.com/oxidecomputer/omicron/pull/3081 -- do it here to avoid having to create a whole separate PR for this.

Fixes #3102.

### DIFF
--- a/update-engine/src/buffer.rs
+++ b/update-engine/src/buffer.rs
@@ -131,9 +131,10 @@ impl<S: StepSpec> EventBuffer<S> {
         // Gather step events across all keys.
         let mut step_events = Vec::new();
         let mut progress_events = Vec::new();
-        for value in self.event_store.map.values() {
-            step_events.extend(value.step_events_since(last_seen).cloned());
-            progress_events.extend(value.step_status.progress_event().cloned());
+        for (_, step_data) in self.steps().as_slice() {
+            step_events.extend(step_data.step_events_since(last_seen).cloned());
+            progress_events
+                .extend(step_data.step_status.progress_event().cloned());
         }
 
         // Sort events.
@@ -145,7 +146,12 @@ impl<S: StepSpec> EventBuffer<S> {
             last_seen = Some(last.event_index);
         }
 
-        EventReport { step_events, progress_events, last_seen }
+        EventReport {
+            step_events,
+            progress_events,
+            root_execution_id: self.root_execution_id(),
+            last_seen,
+        }
     }
 
     /// Returns true if any further step events are pending since `last_seen`.
@@ -154,8 +160,8 @@ impl<S: StepSpec> EventBuffer<S> {
     /// step events. A typical use for this is to check that all step events
     /// have been reported before a sender shuts down.
     pub fn has_pending_events_since(&self, last_seen: Option<usize>) -> bool {
-        for value in self.event_store.map.values() {
-            if value.step_events_since(last_seen).next().is_some() {
+        for (_, step_data) in self.steps().as_slice() {
+            if step_data.step_events_since(last_seen).next().is_some() {
                 return true;
             }
         }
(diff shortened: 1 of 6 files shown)


jgallagher:
I believe this fixes #3102 :tada:

update-engine/src/buffer.rs
```diff
@@ -131,9 +131,10 @@ impl<S: StepSpec> EventBuffer<S> {
         // Gather step events across all keys.
         let mut step_events = Vec::new();
         let mut progress_events = Vec::new();
-        for value in self.event_store.map.values() {
-            step_events.extend(value.step_events_since(last_seen).cloned());
-            progress_events.extend(value.step_status.progress_event().cloned());
+        for (_, step_data) in self.steps().as_slice() {
```
jgallagher, line 134: Just for my own understanding: what's the difference between

```rust
for value in self.event_store.map.values() {
```

and

```rust
for (_, step_data) in self.steps().as_slice() {
```

? It looks like the latter sorts the events first, but I would've guessed that was happening already?

update-engine/src/context.rs
```diff
@@ -60,28 +71,45 @@ impl<S: StepSpec> StepContext<S> {
...
-                &event.kind
-            {
-                res = Err(NestedError::new(message.clone(), causes.clone()));
+        let delta_report = if let Some(id) = report.root_execution_id {
+            let mut nested_buffers = self.nested_buffers.lock().unwrap();
+            Some(nested_buffers.entry(id).or_default().add_event_report(report))
+        } else {
+            // If there's no root execution ID set, report is expected to be
+            // empty. However, report is untrusted data so we can't assert on
+            // it. Instead, log this.
```
jgallagher, line 80: This is not a case that we need to handle, but again just checking my understanding: if we got reports from an out-of-date installinator we'd hit this case, right? It would send events but wouldn't know about the new `root_execution_id` field?

jgallagher (approved):
I tested this on my local reproducing-what-we-saw-on-dogfood setup and confirm the reports are now correctly deduped - thanks! My only comments are for my own understanding.

One followup question: last week I suspected that https://github.com/oxidecomputer/omicron/blob/df4e9fa4a4e50115830810d8397cd336937b8b6c/installinator/src/reporter.rs#L92 ought to use `generate_report_since(self.last_seen)` - did we already talk about that and discard it out of concerns about wicketd possibly losing state and forgetting old events?

update-engine/src/buffer.rs
sunshowers: The first just iterates over all the values in the key while the second performs a DFS from the root execution node. In the simple case of a single event buffer they should be equivalent. However, if a buffer is *reused*, things can get confusing.

I think as a matter of principle it's important to search from the root node. I've been transitioning over to `self.steps()` everywhere 

Separately, at some point in the future I'd also like to make it so that reused buffers get rejected. However that's a bit of extra work that isn't necessary to do at the moment.

update-engine/src/context.rs
sunshowers: Yes, that would be a problem, and I'd do this differently if this came up after we shipped.

Issue #2727, opened 2023-03-31
Transient failures in instance state monitor cause instances to be lost at sea


gjcolombo:
Sled agent spawns a state monitor task per instance that calls Propolis's `instance_state_monitor` endpoint, watches for VM state transitions, processes them, and relays any resulting instance state changes to Nexus. Any error in `Instance::monitor_state_task` or any of its callees bubbles up through the task and causes it to exit. After this, nothing monitors Propolis for subsequent state changes or notifies Nexus if they occur.

The DNS resolution error in #2726 brought this to my immediate attention, but any failure in the monitoring task will do the trick; another not-too-esoteric one is a Propolis panic that causes the call to `instance_state_monitor` to fail.

askfongjojo:
> However, this raises the question: if we fail to alert nexus for long enough, or if we fail to monitor the instance, is that a valid condition for "setting the instance to failed"?

It makes sense to give up after reaching a certain threshold. I mentioned in an earlier discussion today (in the context of sled fault tolerance) that we probably shouldn't automatically reap the failed instances because they could still be working perfectly from the end-user's POV. What may be a good compromise is some amount of human intervention, e.g. the failed instance status handling described [here](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/monitoring-system-instance-status-check.html#types-of-instance-status-checks).

Issue #3107, opened 2023-05-12
Instance V2P management functions should include the instance's current sled


gjcolombo:
`create_instance_v2p_mappings` (and its `delete` counterpart) take an instance and propagate V2P mappings for that instance's NICs to every sled *except* the instance's current sled. Nexus ignores the current sled because incarnating an instance on a sled (or stopping an incarnation of an instance on a sled) creates (or destroys) XDE ports for that instance's NICs, and the XDE driver creates/destroys the relevant mappings on the local sled when it creates/destroys the port.

This behavior creates some tension for live migration, because Nexus has to worry about being undercut by XDE while applying mappings after a migration completes. The plan to resolve this tension is to have XDE stop manipulating mappings on port creation/deletion (https://github.com/oxidecomputer/opte/issues/368) and let Nexus manage them all instead. Once that change lands and Nexus ingests it, Nexus's V2P mapping routines will need to stop skipping their subject instances' current sleds.

internet-diglett:
Marked this for resolution in #5568

PR #3127, opened 2023-05-15
nexus: update instance networking config after live migration

gjcolombo:
Whenever Nexus gets a new instance runtime state from a sled agent, compare the state to the existing runtime state to see if applying the new state will update the instance's Propolis generation. If it will, use the sled ID in the new record to create updated OPTE V2P mappings and Dendrite NAT entries for the instance.

Retry with backoff when sled agent fails to publish a state update to Nexus. This was required for correctness anyway (see #2727) but is especially important now that there are many more ways for Nexus to fail to apply a state update. See the comments in the new code for more details.

In the future, it might be better to update this configuration using a reliable persistent workflow that's triggered by Propolis location changes. This approach will require at least some additional work in OPTE to assign generation numbers to V2P mappings (Dendrite might have a similar problem but I'm not as familiar with the tables Nexus is trying to maintain in this change).

### DIFF
--- a/nexus/src/app/instance.rs
+++ b/nexus/src/app/instance.rs
@@ -1025,7 +1025,13 @@ impl super::Nexus {
         //   race with this one.
         // - This work is not done in a saga. The presumption is instead that
         //   if any of these operations fail, the entire update will fail, and
-        //   sled agent will retry the update.
+        //   sled agent will retry the update. Unwinding on failure isn't needed
+        //   because (a) any partially-applied configuration is correct
+        //   configuration, (b) if the instance is migrating, it can't migrate
+        //   again until this routine successfully updates configuration and
+        //   writes an update back to CRDB, and (c) sled agent won't process any
+        //   new instance state changes (e.g. a change that stops an instance)
+        //   until this state change is successfully committed.
         let (.., db_instance) = LookupPath::new(&opctx, &self.db_datastore)
             .instance_id(*id)
             .fetch_for(authz::Action::Read)
--- a/nexus/src/app/instance.rs
+++ b/nexus/src/app/instance.rs
@@ -1343,7 +1343,7 @@ impl super::Nexus {
                 ))
             })?;
 
-            debug!(log, "creation of nat entry successful for: {target_ip:#?}");
+            info!(log, "creation of nat entry successful for: {target_ip:#?}");
         }
 
         Ok(())
--- a/nexus/src/app/instance.rs
+++ b/nexus/src/app/instance.rs
@@ -42,6 +42,7 @@ use sled_agent_client::types::InstanceStateRequested;
 use sled_agent_client::types::SourceNatConfig;
 use sled_agent_client::Client as SledAgentClient;
 use std::net::SocketAddr;
+use std::str::FromStr;
 use std::sync::Arc;
 use tokio::io::{AsyncRead, AsyncWrite};
 use tokio_tungstenite::tungstenite::protocol::frame::coding::CloseCode;
@@ -993,6 +994,7 @@ impl super::Nexus {
     /// Instance.
     pub async fn notify_instance_updated(
         &self,
+        opctx: &OpContext,
         id: &Uuid,
         new_runtime_state: &nexus::InstanceRuntimeState,
     ) -> Result<(), Error> {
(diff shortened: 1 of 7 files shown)


nexus/src/app/instance.rs
```diff
@@ -1052,6 +1091,264 @@ impl super::Nexus {
...
+        let network_interface = match self
+            .db_datastore
+            .derive_guest_network_interface_info(&opctx, &authz_instance)
+            .await?
+            .into_iter()
+            .find(|interface| interface.primary)
+        {
+            Some(interface) => interface,
+            // Return early if instance does not have a primary network
+            // interface
```
jmpesp, line 1188: why?

nexus/src/app/instance.rs
```diff
@@ -1052,6 +1091,264 @@ impl super::Nexus {
...
+                        .await
+                }
+            }
+            .map_err(|e| {
+                Error::internal_error(&format!(
+                    "failed to create nat entry: {e}"
+                ))
+            })?;
+
+            debug!(log, "creation of nat entry successful for: {target_ip:#?}");
```
jmpesp, line 1346: why is this debug when all the others in this function are info? Personally I'd be interested in the success message not being filtered out depending on log level :)

nexus/src/app/sagas/instance_create.rs
```diff
@@ -373,131 +372,36 @@ async fn sic_add_network_config(
...
-
-    let (.., authz_instance, db_instance) = LookupPath::new(&opctx, &datastore)
+    let datastore = osagactx.datastore();
+    let (.., db_instance) = LookupPath::new(&opctx, &datastore)
         .instance_id(instance_id)
         .fetch()
         .await
-        .map_err(ActionError::action_failed)?;
+        .map_err(ActionError::action_failed)
+        .unwrap();
```
jmpesp, line 381: why not `?` ?

nexus/tests/integration_tests/instances.rs
```diff
@@ -569,6 +570,149 @@ async fn test_instance_migrate(cptestctx: &ControlPlaneTestContext) {
...
+        .derive_guest_network_interface_info(&opctx, &authz_instance)
+        .await
+        .unwrap();
+    let original_sled_id = nexus.instance_sled_id(&instance_id).await.unwrap();
+
+    let mut sled_agents = vec![cptestctx.sled_agent.sled_agent.clone()];
+    sled_agents.extend(other_sleds.iter().map(|tup| tup.1.sled_agent.clone()));
+    for sled_agent in &sled_agents {
+        // Starting the instance should have programmed V2P mappings to all the
+        // sleds except the one where the instance is running.
```
jmpesp, line 648: `TODO(#3107)` here too?

nexus/tests/integration_tests/instances.rs
```diff
@@ -569,6 +570,149 @@ async fn test_instance_migrate(cptestctx: &ControlPlaneTestContext) {
...
+    instance_simulate_on_sled(cptestctx, nexus, dst_sled_id, instance_id).await;
+    let instance = instance_get(&client, &instance_url).await;
+    assert_eq!(instance.runtime.run_state, InstanceState::Running);
+    let current_sled = nexus.instance_sled_id(&instance_id).await.unwrap();
+    assert_eq!(current_sled, dst_sled_id);
+
+    for sled_agent in &sled_agents {
+        // Completing migration published a sled ID update that should have
+        // programmed new V2P mappings to all sleds other than the destination
+        // sled.
```
jmpesp, line 700: `TODO(#3107)` here too?

nexus/tests/integration_tests/instances.rs
```diff
@@ -3189,22 +3333,15 @@ async fn test_instance_v2p_mappings(cptestctx: &ControlPlaneTestContext) {
     sled_agents.push(&cptestctx.sled_agent.sled_agent);
 
     for sled_agent in &sled_agents {
-        let v2p_mappings = sled_agent.v2p_mappings.lock().await;
         if sled_agent.id != db_instance.runtime().sled_id {
```
jmpesp, line 3336: `TODO(#3107)`

nexus/src/app/instance.rs
```diff
@@ -1002,6 +1004,43 @@ impl super::Nexus {
...
+        //   migration is in progress, the instance's `migration_id` field will
+        //   be non-NULL, and a new migration cannot start until it is cleared.
+        //   This routine must notify recipients before writing new records
+        //   back to CRDB so that these "locks" remain held until all
+        //   notifications have been sent. Otherwise, Nexus might allow new
+        //   operations to proceed that will produce system updates that might
+        //   race with this one.
+        // - This work is not done in a saga. The presumption is instead that
+        //   if any of these operations fail, the entire update will fail, and
+        //   sled agent will retry the update.
```
jmpesp, line 1028: There's no need to unwind these changes, or will the unwind == another update comes along with a higher gen to overwrite?

jmpesp (approved):
Looks good to me, some questions:

nexus/src/app/instance.rs
gjcolombo: Touched up the comment in 1c36e9ce. The general idea is that any partially-applied updates are correct, and they'll remain correct because there's no way to produce a newer configuration until this one is written back. That's the theory, anyway; thinking about this stuff is making me want to get the TLA+ specs out again.

nexus/src/app/sagas/instance_create.rs
gjcolombo: Good catch, fixed in 024360f7. The instance record must exist at this point (it was created in the Creating state and can't move to a state that allows deletion until the saga finishes), but that doesn't mean the saga should assume the fetch will succeed.

nexus/src/app/instance.rs
gjcolombo: Fixed in df655b34.

nexus/src/app/instance.rs
gjcolombo: I'm not sure--most of this part of the change is copied from `sic_instance_create`. My guess is that it's possible in principle to have external IPs assigned to an instance without actually having any NICs attached, and this is this routine's way of handling that case.

I tweaked the comment a bit in aa982b2d to reflect this understanding.

nexus/tests/integration_tests/instances.rs
gjcolombo: Fixed this and the others in 862a2b43.

nexus/src/app/instance.rs
jmpesp: :+1:

PR #3128, opened 2023-05-15
[wicket / wicketd] Minor cleanup from Mupdate testing

jgallagher:
This is a PR that rolls up several small cleanups/bug fixes from testing MUpdate last week:

1. wicketd: Log the hash of the extracted host phase 2 artifact (we already log the hashes of the other artifacts, and at one point I wanted to find this - cracking open the tuf repo and then untarring the composite host artifact was annoying)
2. wicket: Only log the full EventReport we receive about one per minute. We were hitting #3102 so these were huge, but even after that's fixed I don't think we need to log these every second.
3. wicket: Remove the ESC/ENTER titles on the main panes. Switching between panes now happens via TAB.
4. wicket: Fix ESC closing the ignition popup - previously it always sent us back to the rack view, and there was no way to close the ignition popup other than running one of the commands.

### DIFF
--- a/wicket/src/runner.rs
+++ b/wicket/src/runner.rs
@@ -14,6 +14,7 @@ use futures::StreamExt;
 use slog::{debug, error, info};
 use std::io::{stdout, Stdout};
 use std::net::SocketAddrV6;
+use std::time::Instant;
 use tokio::sync::mpsc::{
     unbounded_channel, UnboundedReceiver, UnboundedSender,
 };
@@ -21,6 +22,7 @@ use tokio::time::{interval, Duration};
 use tui::backend::CrosstermBackend;
 use tui::Terminal;
 
+use crate::events::EventReportMap;
 use crate::ui::Screen;
 use crate::wicketd::{self, WicketdHandle, WicketdManager};
 use crate::{Action, Cmd, Event, KeyHandler, Recorder, State, TICK_INTERVAL};
@@ -49,6 +51,10 @@ pub struct RunnerCore {
 
     // Our friendly neighborhood logger
     pub log: slog::Logger,
+
+    // Helper to limit our logging of event reports (which can be quite large)
+    // to a slower cadence than their arrival.
+    log_throttler: EventReportLogThrottler,
 }
 
 impl RunnerCore {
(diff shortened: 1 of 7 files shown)


wicket/src/runner.rs
```diff
@@ -345,3 +356,52 @@ async fn run_event_listener(
...
+
+struct EventReportLogThrottler {
+    last_log: Option<Instant>,
+    min_time_between_logs: Duration,
+}
+
+impl Default for EventReportLogThrottler {
+    fn default() -> Self {
+        const DEFAULT_TIME_BETWEEN_LOGS: Duration = Duration::from_secs(60);
+        Self::new(DEFAULT_TIME_BETWEEN_LOGS)
```
sunshowers, line 368: I suspect 60s is a long time here. Can this be brought down to eg 15s?

wicketd/src/update_tracker.rs
```diff
@@ -151,19 +151,20 @@ impl UpdateTracker {
...
+                        // task (which might still be trying to upload) and
+                        // start a new one with our current image.
                         //
-                        // TODO-correctness If we still have updates running that
-                        // expect the old image, they're probably going to fail.
-                        // Should we handle that more cleanly or just let them fail?
+                        // TODO-correctness If we still have updates running
+                        // that expect the old image, they're probably going to
+                        // fail. Should we handle that more cleanly or just let
+                        // them fail?
```
sunshowers, line 162: Now that we've made it so that a new repository can't be uploaded while updates are running, I think we can just make old running updates (which definitely can't be spun up by this copy of wicket) fail.

sunshowers (approved):
Thanks!

wicket/src/runner.rs
jgallagher: Sure, I wasn't sure what was reasonable here. Even with the report size correct these still end up pretty large - a few hundred lines, multiplied by however many sleds have relevant reports (from running or completed updates). We can try 15s and see how it is though!

wicketd/src/update_tracker.rs
jgallagher: Ah good point. This was just a formatting change, but I'll remove this TODO entirely. Nice catch.

Issue #2148, opened 2023-01-10
Need API endpoint for listing IP Pools scoped to a project


bnaecker:
Need a way to list the IP pools available to resources within a project. That means any IP Pool not reserved to another project, or reserved for a specific project.

smklein:
Related: https://github.com/oxidecomputer/omicron/issues/2055

I thought there were not "IP Pools scoped to a project", but rather, "IP Pools which are available to a project". I think that's a subtle distinction, but also a matter of ownership.

bnaecker:
Yeah, I was rushing here. I meant that the _endpoint_ should be project-scoped, and that should return the IP Pools _available_ to a project.

So `../projects/{project_name}/ip-pools` returns all IP Pools that can be used by Project `project_name`.

bnaecker:
I'm catching back up to a bunch of work I was only vaguely aware of in the last few months. This issue is possibly still relevant, but much simpler after https://github.com/oxidecomputer/omicron/issues/2056. There is no project ID at all on IP Pools anymore, so this just needs to return all non-internal pools. That logic may change in the future as we flesh out the scoping of pools, but for now, it's just anything that any developer can use. I.e., all projects will return the same list now.

lifning:
just to clarify -- for the time being, this would literally be a duplicate of `/v1/system/ip-pools`'s current behavior (which already does `.filter(dsl::internal.eq(false))`), with the purpose being that `/v1/projects/{}/ip-pools` would be subject to potentially change in the future?

rmustacc:
Yes, I believe that this is correct. Let me try to provide some surrounding context here.

One of the main things that we know that we're going to need, both from prior experience and from customer feedback, is that they want a way to have a group of IPs that can be used for provisioning that is eventually restricted to a subset of users, whether that ACL is eventually associated with a project, a silo, or some other more complex thing (but critically not based on user id). So the idea is that really if you're using the CLI or any other thing, the question of what images, IP pools, or other resources can I use is ultimately specific to the project and therefore by using a project-scoped URL you'll get what you can use.

So today the only ACL that really exists is the internal-only bit. I think we want to look at what it means to update all the user and terraform tooling to use this endpoint. The fact that the system IP pools interface filters out internal things raises a question for me because it makes me ask how does an operator change that.

Hopefully this helps a bit. Thanks for picking this one up @lifning.

zephraph:
Sounds good.

I think the shape of the API is likely going to need to be different though. 

[...]

PR #3132, opened 2023-05-16
Add API endpoints for retrieving IP pools available to projects (#2148)

lifning:
Per discussion in #2148, this is a future-looking API that's functionally equivalent to the `/v1/system/`-prefixed IP-pool-viewing routes we already have, but with room to grow into project-level scoping when it's implemented, so they can be used optatively wherever appropriate.

(This PR doesn't yet add a `GET /v1/ip-pools/{pool}/ranges` -- should it?)

### DIFF
--- a/nexus/src/external_api/http_entrypoints.rs
+++ b/nexus/src/external_api/http_entrypoints.rs
@@ -39,8 +39,10 @@ use dropshot::{
     channel, endpoint, WebsocketChannelResult, WebsocketConnection,
 };
 use ipnetwork::IpNetwork;
+use nexus_db_queries::authz::ApiResource;
 use nexus_db_queries::db::lookup::ImageLookup;
 use nexus_db_queries::db::lookup::ImageParentLookup;
+use nexus_types::external_api::params::ProjectSelector;
 use nexus_types::{
     external_api::views::{SledInstance, Switch},
     identity::AssetIdentityMetadata,
@@ -107,6 +109,8 @@ pub fn external_api() -> NexusApiDescription {
         api.register(project_update)?;
         api.register(project_policy_view)?;
         api.register(project_policy_update)?;
+        api.register(project_ip_pool_list)?;
+        api.register(project_ip_pool_view)?;
 
         // Operator-Accessible IP Pools API
         api.register(ip_pool_list)?;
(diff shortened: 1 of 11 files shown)


nexus/src/external_api/http_entrypoints.rs
```diff
@@ -1041,6 +1043,78 @@ async fn project_policy_update(
 
 // IP Pools
 
+/// List all IP Pools that can be used by a given project.
+#[endpoint {
+    method = GET,
+    path = "/v1/ip-pools",
+    tags = ["projects"],
+}]
+async fn ip_pool_list_project_scoped(
```
zephraph, line 1052:
```suggestion
async fn project_ip_pool_list(
```

To be consistent with our current naming scheme.

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -1041,6 +1043,78 @@ async fn project_policy_update(
...
+    let apictx = rqctx.context();
+    let handler = async {
+        let nexus = &apictx.nexus;
+        let query = query_params.into_inner();
+        let pag_params = data_page_params_for(&rqctx, &query)?;
+        let scan_params = ScanByNameOrId::from_query(&query)?;
+        let paginated_by = name_or_id_pagination(&pag_params, scan_params)?;
+        let opctx = crate::context::op_context_for_external_api(&rqctx).await?;
+        let pools = nexus
+            .ip_pools_list(&opctx, &paginated_by)
```
zephraph, line 1071: I'd expect this to fail. The permissions of a user calling this endpoint inside of a silo and a user calling it at the system level should be different. If it doesn't, that's a bug.

I would recommend creating an app layer function called `project_ip_pool_list` in the `projects` app layer file. The resource that's being acted on is a synthetic resource called `IP_POOL_LIST` as seen below. 

https://github.com/oxidecomputer/omicron/blob/891acc7db7d798a914007df70fd1ae412de2141f/nexus/db-queries/src/db/datastore/ip_pool.rs#L56

We'll also need to update the policy for that synthetic resource to allow folks with project access to list the endpoint. You may want to go ahead and resolve the project to a lookup and pass that down to the app layer even if it's not used yet. (The code below is how to resolve the query to a lookup). 

https://github.com/oxidecomputer/omicron/blob/891acc7db7d798a914007df70fd1ae412de2141f/nexus/src/external_api/http_entrypoints.rs#L2869

I know there's a lot here, but I'm happy to help pair on it if you need.

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -1041,6 +1043,78 @@ async fn project_policy_update(
...
+    apictx.external_latencies.instrument_dropshot_handler(&rqctx, handler).await
+}
+
+/// Fetch an IP pool
+#[endpoint {
+    method = GET,
+    path = "/v1/ip-pools/{pool}",
+    tags = ["projects"],
+}]
+async fn ip_pool_view_project_scoped(
```
zephraph, line 1091:
```suggestion
async fn project_ip_pool_view(
```

to match our current naming conventions.

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -1076,6 +1079,87 @@ async fn project_policy_update(
...
+            .await?;
+        if pool.internal {
+            return Err(HttpError::for_bad_request(
+                None,
+                format!(
+                    "Requested pool {} is reserved for internal use.",
+                    pool_selector
+                ),
+            ));
+        }
```
zephraph, line 1157: This should just be a 404. There's no real reason we should inform the user that exists or not. Maybe also a TODO that once the project_ip_pool_lookup is actually implemented to query only the project we can remove this part.

zephraph (approved):
Looks good. The only thing I'd ask you update is changing the 400 to a 404 when the user fetches an internal pool.

PR #3136, opened 2023-05-17
[installinator] Sync I/O before reporting completion to wicketd

jgallagher:
Today when mupdating `madrid` we observed data loss: `installinator` reported it wrote the control plane zones to both M.2s, but after a reboot they were only present on one. Under the hypothesis that we could have lost data because wicketd reboots the sled as soon as installinator reports success and installinator had no filesystem sync operations after finishing writing, this PR contains several sync operations, most of which came from a conversation with @rmustacc and @jclulow:

1. When opening the M.2 raw device, pass `O_SYNC`.
2. After writing the OS to the M.2 raw device, call `ioctl(fd, DKIOFLUSHWRITECACHE, NULL)`.
3. Before writing the control plane zones, remove any files present in the directory. (Expectation: these are old control plane zones. There should be nothing else here.)
4. After writing each control plane zone, `fsync()` the file descriptor.
5. When writing each control plane zone, write to a temp file in that directory first, then rename it after writing is finished.
6. After writing all control plane zones, open and `fsync()` the directory in which we wrote the files.
7. As a final step after all control plane zones have been written and all the above `fsync()`s are done, `zpool export` the M.2 datasets we wrote to.

~~I want to test this several times on `madrid` before merging, but I think it's sufficiently close to done for review.~~ Pushed a couple small fixes after testing on `madrid`:

* Ignore `ENOTSUP` when attempting to `DKIOFLUSHWRITECACHE`; our current devices down't support this, but we can still leave it in case future drives do.
* Drop the open control plane directory handle before attempting to `zpool export` (holding it open causes the export to fail because the zpool is still busy).

I'll continue testing on `madrid` to see if anything else shakes out after more tries, but this has now worked a couple times.

### DIFF
--- a/illumos-utils/src/zpool.rs
+++ b/illumos-utils/src/zpool.rs
@@ -218,7 +218,7 @@ impl Zpool {
         cmd.env_clear();
         cmd.env("LC_ALL", "C.UTF-8");
         cmd.arg(ZPOOL).arg("export").arg(&name.to_string());
-        execute(&mut cmd).map_err(Error::from)?;
+        execute(&mut cmd)?;
 
         Ok(())
     }
@@ -232,7 +232,7 @@ impl Zpool {
             .arg("set")
             .arg("failmode=continue")
             .arg(&name.to_string());
-        execute(&mut cmd).map_err(Error::from)?;
+        execute(&mut cmd)?;
         Ok(())
     }
 
--- a/illumos-utils/src/zpool.rs
+++ b/illumos-utils/src/zpool.rs
@@ -217,9 +217,7 @@ impl Zpool {
         let mut cmd = std::process::Command::new(PFEXEC);
         cmd.env_clear();
         cmd.env("LC_ALL", "C.UTF-8");
-        cmd.arg(ZPOOL)
-            .arg("export")
-            .arg(&name.to_string());
+        cmd.arg(ZPOOL).arg("export").arg(&name.to_string());
         execute(&mut cmd).map_err(Error::from)?;
 
         Ok(())
--- a/illumos-utils/src/zpool.rs
+++ b/illumos-utils/src/zpool.rs
@@ -213,6 +213,18 @@ impl Zpool {
         }
     }
 
+    pub fn export(name: &ZpoolName) -> Result<(), Error> {
+        let mut cmd = std::process::Command::new(PFEXEC);
+        cmd.env_clear();
+        cmd.env("LC_ALL", "C.UTF-8");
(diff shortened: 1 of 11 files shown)


illumos-utils/src/zpool.rs
```diff
@@ -213,6 +213,16 @@ impl Zpool {
         }
     }
 
+    pub fn export(name: &ZpoolName) -> Result<(), Error> {
+        let mut cmd = std::process::Command::new(PFEXEC);
+        cmd.env_clear();
+        cmd.env("LC_ALL", "C.UTF-8");
+        cmd.arg(ZPOOL).arg("export").arg(&name.to_string());
+        execute(&mut cmd).map_err(Error::from)?;
```
sunshowers, line 221: `.map_err(Error::from)` shouldn't be required, `?` does that already I think.

installinator-common/src/progress.rs
```diff
@@ -262,17 +263,28 @@ pub enum WriteStepId {
 
 /// The error that occurred.
 #[derive(Debug, Error)]
-#[error(
-    "writing {component} to slot {slot} failed \
+pub enum WriteError {
+    #[error(
+        "writing {component} to slot {slot} failed \
      after {written_bytes}/{total_bytes} bytes"
```
sunshowers, line 269: Indent needs to be matched with previous line here.

sunshowers (approved):

installinator/src/write.rs
```diff
@@ -492,11 +522,40 @@ impl ControlPlaneZoneWriteContext<'_> {
...
     ) {
         use update_engine::StepHandle;
 
         let slot = self.slot;
 
+        // If we're on a gimlet, remove any files in the control plane
+        // destination directory.
+        if self.clean_output_directory {
+            let output_directory = self.output_directory.to_path_buf();
+            engine
```
andrewjstone, line 535: I like how these are explicit steps. Really digging the update engine.

andrewjstone (approved):
Looks good! The number of steps involved making sure data gets persisted safely has me worried about third party and other control-plane software though.

illumos-utils/src/zpool.rs
jgallagher: Nice catch, copy-pasted from a method where it was required. Fixed in 5623fe1d

installinator-common/src/progress.rs
jgallagher: :+1: fixed in 2a518741

PR #3138, opened 2023-05-17
silo creation should accept TLS certificates

davepacheco:
This is necessary because in a real system:

1. You may only be able to access a Silo from its own console/API endpoint (this is not yet implemented, but I assume it's coming)
2. Nexus will not be listening on HTTP (only HTTPS).  (This isn't true today, but needs to be soon outside of development.)
3. There is currently no way to administer the certificates of another Silo.  (We generally don't want to allow this sort of thing.  See #1681.  The list of TLS certificates is probably an exception similar to identity providers -- if it gets messed up, you can't log into the Silo directly to fix it.  But anyway, we haven't implemented it yet.)

Thus, without providing certificates when you create the Silo, you'd have no way to log into it.  We get away without this today because (1) and (2) aren't true yet.  Nexus _does_ listen on HTTP and you can currently use any Silo's endpoint to log into any other Silo.  Also, until recently (with #3095), certificates were fleet-wide.

### DIFF
--- a/nexus/db-queries/src/db/datastore/rack.rs
+++ b/nexus/db-queries/src/db/datastore/rack.rs
@@ -18,7 +18,6 @@ use crate::db::error::public_error_from_diesel_pool;
 use crate::db::error::ErrorHandler;
 use crate::db::error::TransactionError;
 use crate::db::identity::Asset;
-use crate::db::model::Certificate;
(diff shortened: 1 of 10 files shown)


luqmana (approved):
Thanks for digging into and fleshing out this space! LGTM

PR #3140, opened 2023-05-17
Spawn a task to run services_ensure

### LINKED ISSUE #3098: crucible pantry zone got the wrong control vnic
The profile states that oxControlService9 should be the datalink:

```
root@oxz_crucible_pantry:~# cat /var/svc/profile/site.xml
<!DOCTYPE service_bundle SYSTEM "/usr/share/lib/xml/dtd/service_bundle.dtd.1">
<service_bundle type="profile" name="omicron">
  <service version="1" type="service" name="oxide/crucible/pantry">
    <instance enabled="true" name="default">
      <property_group type="application" name="config">
        <propval type="astring" name="datalink" value="oxControlService9"/>
        <propval type="astring" name="gateway" value="fd00:1122:3344:101::1"/>
        <propval type="astring" name="listen_addr" value="fd00:1122:3344:101::9"/>
        <propval type="astring" name="listen_port" value="17000"/>
      </property_group>
    </instance>
  </service>
```

but the zone has the wrong one!

```
root@oxz_crucible_pantry:~# dladm
oxControlService8 vnic 9000  up       --         ?
```

The zone cfg:

```
james@dinnerbone:~$ pfexec zonecfg -z oxz_crucible_pantry info
zonename: oxz_crucible_pantry
zonepath: /zone/oxz_crucible_pantry
brand: omicron1
autoboot: false
bootargs: 
pool: 
limitpriv: 
scheduling-class: 
ip-type: exclusive
hostid: 
fs-allowed: 
net:
	address not specified
	allowed-address not specified
	physical: oxControlService8
	defrouter not specified
```

jmpesp:
dropshot/hyper will cancel an endpoint's task if the call times out or is otherwise cancelled, and this leads to a specific issue where booting all the service zones takes longer than the progenitor client default timeout. The request times out, `services_ensure` gets cancelled, and this leaves zones partially configured but not added to the list of `existing_zones`. When the next `PUT /services` call is made, `services_ensure` will eventually try to bring up the same zone it was interrupted at, leading to configuration issues.

Fixes #3098

### DIFF
--- a/sled-agent/src/http_entrypoints.rs
+++ b/sled-agent/src/http_entrypoints.rs
@@ -67,22 +67,19 @@ async fn services_put(
     let sa = rqctx.context().clone();
     let body_args = body.into_inner();
 
-    // Spawn a task to run `services_ensure`: dropshot/hyper will cancel an
-    // endpoint's task if the call times out or is otherwise cancelled, and this
-    // leads to a specific issue where booting all the service zones takes
-    // longer than the progenitor client default timeout. The request times out,
-    // `services_ensure` gets cancelled, and this leaves zones partially
-    // configured but not added to the list of `existing_zones`. When the next
-    // `PUT /services` call is made, `services_ensure` will eventually try to
-    // bring up the same zone it was interrupted at, leading to configuration
-    // issues. See: oxidecomputer/omicron#3098.
-
+    // Spawn a separate task to run `services_ensure`: cancellation of this
+    // endpoint's future (as might happen if the client abandons the request or
+    // times out) could result in leaving zones partially configured and the
+    // in-memory state of the service manager invalid. See:
+    // oxidecomputer/omicron#3098.
     match tokio::spawn(async move { sa.services_ensure(body_args).await }).await
     {
         Ok(result) => result.map_err(|e| Error::from(e))?,
 
         Err(e) => {
-            return Err(HttpError::for_internal_error(e.to_string()));
+            return Err(HttpError::for_internal_error(
+                format!("unexpected failure awaiting \"services_ensure\": {:#}", e)
+            ));
         }
     }
 
--- a/sled-agent/src/http_entrypoints.rs
+++ b/sled-agent/src/http_entrypoints.rs
@@ -64,9 +64,28 @@ async fn services_put(
     rqctx: RequestContext<SledAgent>,
     body: TypedBody<ServiceEnsureBody>,
 ) -> Result<HttpResponseUpdatedNoContent, HttpError> {
-    let sa = rqctx.context();
+    let sa = rqctx.context().clone();
     let body_args = body.into_inner();
-    sa.services_ensure(body_args).await.map_err(|e| Error::from(e))?;
+
+    // Spawn a task to run `services_ensure`: dropshot/hyper will cancel an
+    // endpoint's task if the call times out or is otherwise cancelled, and this
+    // leads to a specific issue where booting all the service zones takes
+    // longer than the progenitor client default timeout. The request times out,
+    // `services_ensure` gets cancelled, and this leaves zones partially
+    // configured but not added to the list of `existing_zones`. When the next
+    // `PUT /services` call is made, `services_ensure` will eventually try to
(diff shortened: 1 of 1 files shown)


sled-agent/src/http_entrypoints.rs
```diff
@@ -64,9 +64,28 @@ async fn services_put(
...
+
+    // Spawn a task to run `services_ensure`: dropshot/hyper will cancel an
+    // endpoint's task if the call times out or is otherwise cancelled, and this
+    // leads to a specific issue where booting all the service zones takes
+    // longer than the progenitor client default timeout. The request times out,
+    // `services_ensure` gets cancelled, and this leaves zones partially
+    // configured but not added to the list of `existing_zones`. When the next
+    // `PUT /services` call is made, `services_ensure` will eventually try to
+    // bring up the same zone it was interrupted at, leading to configuration
+    // issues. See: oxidecomputer/omicron#3098.
```
davepacheco, line 78: I'd suggest summarizing this, maybe something like:

```suggestion
    // Spawn a task to run `services_ensure` so that it runs to completion even if
    // this future is cancelled (as might happen if the client abandons the request).
    // Otherwise, unexpected cancellation of this future could result in leaving the in-
    // memory state of this zone invalid.
```

There's a lot of other detail there that's pretty specific to the bug you saw.

Tangentially: what I understood happened (which could well be wrong!) is that the _client_ timed out its request, the socket got closed, the corresponding Future got dropped, and so it became cancelled.  As far as I know, neither hyper nor dropshot has a policy that would cause a request to be timed out.

sled-agent/src/http_entrypoints.rs
```diff
@@ -64,9 +64,28 @@ async fn services_put(
...
+
+    match tokio::spawn(async move { sa.services_ensure(body_args).await }).await
```
davepacheco, line 80: Ideally it would seem good to have a limit to how many of these tasks can be outstanding.  I'm not sure this is worth prioritizing now though.

sled-agent/src/http_entrypoints.rs
```diff
@@ -64,9 +64,28 @@ async fn services_put(
...
+    {
+        Ok(result) => result.map_err(|e| Error::from(e))?,
+
+        Err(e) => {
+            return Err(HttpError::for_internal_error(e.to_string()));
```
davepacheco, line 85:
```suggestion
            return Err(HttpError::for_internal_error(&format!("failed to spawn tokio task for services_ensure: {:#}",e));
```

jgallagher: I don't think the suggested message is right. I don't think `tokio::spawn()` itself is fallible: if it returns an error, it means the task that was spawned either panicked or was cancelled. I usually `.unwrap()` `tokio::spawn(..).await` calls that cannot be canceled (other than by the future calling them itself being canceled, which means you wouldn't get to the unwrap), since it can only panic if another panic has already happened.

davepacheco: Good call.  I didn't read that closely enough.  In that case I'd suggest:

```rust
            return Err(HttpError::for_internal_error(&format!("unexpected failure awaiting \"services_ensure\" task: {:#}",e));
```

I'm not attached to the message.  I just don't like assuming that the message provided by the callee is going to be sufficient for someone looking through the logs to figure out what Nexus was trying to do when it ran into this problem.

sled-agent/src/http_entrypoints.rs
jmpesp: I thought about changing where `services_ensure` grabs the lock from a `lock` to a `try_lock`, bubbling up the `TryLockError` as maybe a 429. This means that each client will now have to catch this error and retry on timeout and this new 429. That's not a terrible amount of work but it does mean that all current and future callers will have to do this.

sled-agent/src/http_entrypoints.rs
jmpesp: I trimmed it in fb7b9f7d, let me know what you think

davepacheco (approved):

Issue #3139, opened 2023-05-17
Want all migration destination information in CRDB's instance table


gjcolombo:
The CRDB instance table records a migrating instance's destination Propolis ID, but not its destination sled or the destination Propolis's server IP. This is a minor hassle in a couple of places:

- When a migration succeeds, the source and target both know that migration has succeeded, but only the target can push a correct instance state update to Nexus (because the source doesn't have the right sled ID/Propolis IP to insert into the updated state).
- When a migration fails, Nexus can determine from the updates it receives from the relevant sleds that the destination Propolis is no longer needed and that its resources can be released. Ideally, before freeing resources, Nexus would explicitly verify with the destination sled that the destination Propolis really is gone, but Nexus doesn't know what the target sled was when migration fails (because there's no field for it).

Issue #2824, opened 2023-04-12
Instance start needs to be a saga


gjcolombo:
Once #2765 lands, instance start will need to be a saga, since start requires two separate actions (registering an instance + starting an instance) that must either both complete or both be undone.

This will also be needed to fix #2315, since that will add a "assign this instance to a sled & allocate resources for it" step to this saga.

gjcolombo:
I'm actively working on this.

PR #3141, opened 2023-05-17
nexus: release unused Propolis reservations after migration

gjcolombo:
This is the last "big rock" PR in the live migration series. There will be some follow-up work to do to polish things up once this lands (#3107, #3139), but with this in place it should be possible to exercise live migration in the dogfood environment without progressively exhausting sleds' resources.

---

When an instance's Propolis generation changes, examine how its Propolis ID and destination Propolis ID are changing and unregister and remove resource reservations from any Propolises that are no longer active.

Update the migration integration test to check this behavior by checking to make sure that a simulated cluster that can fit exactly two instances can start one instance, migrate it, and then start the second instance.

### DIFF
--- a/nexus/src/app/instance.rs
+++ b/nexus/src/app/instance.rs
@@ -1097,6 +1097,28 @@ impl super::Nexus {
         }
     }
 
+    /// Applies any system configuration changes that result from changes to an
+    /// instance's active sled and/or active Propolis ID.
+    ///
+    /// This routine is meant to be called in response to a call from sled agent
+    /// to update an instance's runtime state. This routine expects that sled
+    /// agent will retry these updates if they fail. Accordingly, this routine
+    /// is idempotent and will not panic on unexpected input (since a panic on
+    /// unexpected input could lead to a single sled taking down Nexuses one by
+    /// one).
+    ///
+    /// # Parameters
+    ///
+    /// - `opctx`: The internal operation context to use for database
+    ///   operations.
+    /// - `new_runtime`: The runtime state for the instance of interest that the
+    ///   caller will attempt to write back to CRDB.
+    /// - `db_instance`: The current instance runtime state the caller fetched
+    ///   from CRDB.
+    ///
+    /// # Return value
+    ///
+    /// `Ok` if all required updates were applied, `Err` otherwise.
     async fn handle_instance_propolis_gen_change(
         &self,
         opctx: &OpContext,
@@ -1142,6 +1164,11 @@ impl super::Nexus {
             None,
         )
         .await?;
+
+        // Clean up any resource reservations for Propolises that no longer
+        // appear in the instance record.
+        self.reclaim_unused_propolis_resources(opctx, new_runtime, db_instance)
+            .await?;
 
         Ok(())
     }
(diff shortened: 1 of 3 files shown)


nexus/src/app/instance.rs
```diff
@@ -1357,6 +1384,127 @@ impl super::Nexus {
...
+        let instance_id = db_instance.id();
+
+        // Handle changes to the instance's active Propolis ID. The only
+        // expected change is from the current ID to a migration destination ID.
+        let old_propolis_id = db_instance.runtime().propolis_id;
+        if new_runtime.propolis_id != old_propolis_id {
+            match db_instance.runtime().dst_propolis_id {
+                Some(dst_id) if dst_id != new_runtime.propolis_id => {
+                    warn!(log,
+                          "active Propolis did not change to migration target";
```
luqmana, line 1407: Should this actually fail instead of just warn? An instance suddenly being served by a new separate propolis in the midst of a migration seems like a bug.

nexus/src/app/instance.rs
```diff
@@ -1357,6 +1384,127 @@ impl super::Nexus {
...
+                          "instance_id" => %instance_id,
+                          "new_propolis_id" => %new_runtime.propolis_id,
+                          "old_propolis_id" => %old_propolis_id,
+                          "dst_propolis_id" => %dst_id);
+                }
+                None => {
+                    warn!(log,
+                          "active Propolis changed without an active migration";
```
luqmana, line 1415: Not a blocker here but reminds me that maybe we should make `active_propolis_id` nullable on the instance record and allocate a new one whenever starting an instance. Anyways, that'd probably be part of making instance start a saga itself.

nexus/src/app/instance.rs
```diff
@@ -1357,6 +1384,127 @@ impl super::Nexus {
...
+                  "instance_id" => %instance_id,
+                  "propolis_id" => %old_propolis_id);
+
+            let old_sa =
+                self.sled_client(&db_instance.runtime().sled_id).await?;
+
+            // The caller is going to replace the instance runtime state with
+            // the newly-advanced state it supplied, so there's no need to write
+            // the returned runtime state back to CRDB here.
+            let _ = old_sa.instance_unregister(&instance_id).await?;
```
luqmana, line 1439: Just to confirm, it's always the new sled-agent that would've triggered this nexus call right? i'd have to double check but if we got an update from the old sled-agent and tried to call back to it we might dead lock if it's holding the instance lock in sled-agent.

nexus/src/app/instance.rs
gjcolombo: Agreed--the active sled should also be nullable and should be allowed to change in that saga. We have #2315 tracking nullable sled IDs and #2824 for making instance start a saga; I think having a nullable Propolis ID will fall out pretty quickly when we break ground on #2824.

nexus/src/app/instance.rs
gjcolombo: Ooh, good catch.

Yes, right now it's always the new sled agent that triggers this call, though that's somewhat accidental--the update procedure is supposed to allow both sides to send the "transfer" state change, but the source doesn't do it because (today) the `InstanceRuntimeState` doesn't contain enough information about the destination for the source to be able to create the correct record. If it *did* have that information, though, we'd run right into the problem you're describing, because the monitor thread indeed holds the instance lock while publishing to Nexus.

In general, I'd prefer not to have the sled agent hold locks while calling out to other services. I think that's relatively doable in the monitor task's context. It might be more difficult in the `propolis_ensure` path, but that's less pressing for this PR because those calls to `publish_state_to_nexus` are explicitly only used when starting an instance without migrating it. I'll see if I can tighten up the monitor loop.

nexus/src/app/instance.rs
gjcolombo: I'll need to think about this one a little more. I agree that if we get to this path, we've definitely violated an invariant somewhere. I'm unsure what to do about it from this path, though. In theory, we can talk to any of the sleds involved from this path:

- this routine can return a response to this sled agent that tells it to do something specifically
- it can also use the `sled_id` and `dst_sled_id` from the DB to contact either of the other involved sled agents

We could maybe use that to try to get all the instances to stop and then report the instance is Failed...? I'll need to mull it over (and I'm very much open to suggestions here :)).

(The main things I want to avoid are (a) silently going on like nothing's wrong, and (b) panicking--because this might mean that something unexpected is in CRDB, I want to avoid having the sled that's making this call go to each Nexus in turn and perform an operation that will deterministically cause it to panic.)

gjcolombo:
I'm going to approach this in a different way in an upcoming PR, so closing this one as obsolete.

PR #3142, opened 2023-05-17
Set MTU of Chelsio links

### LINKED ISSUE #3086: Determine default MTU values and set them on sled startup.
We need to:

1. Determine what the default MTU value should be for compute sled physical interfaces.
2. Have the sled agent ensure that value is set when a sled starts up.

Later on we may want to discuss MTU as a tunable knob, but we need to get a sensible default behavior going first.

The MTU needs to be at least large enough to encapsulate a guest 1500 MTU packet in a Geneve outer packet. In testing we have been using an ad-hoc value of 1600 for this purpose. The default sled image is currently setting physical interface MTU values to 9000.

There are a number of tradeoffs in determining a default MTU value. VM instance interfaces are not the only consumers, we also need to consider Crucible traffic.

Another thing to note is that the MTU cannot currently be changed once there are consumers (e.g. IP interfaces) atop the physical data links. It may be possible to overcome that limitation later, but we should assume this is the case for the foreseeable future.

jmpesp:
Set MTU of Chelsio links to 1600 before adding link local v6 addresses, as that can't be done afterwards :)

Fixes #3086.


sled-hardware/src/underlay.rs
```diff
@@ -44,13 +46,38 @@ pub enum Error {
...
+    SetLinkprop(#[from] SetLinkpropError),
+
+    #[error("Could not get linkprop")]
+    GetLinkprop(#[from] GetLinkpropError),
 }
 
 /// Convenience function that calls
 /// `ensure_links_have_global_zone_link_local_v6_addresses()` with the links
 /// returned by `find_chelsio_links()`.
 pub fn find_nics() -> Result<Vec<AddrObject>, Error> {
```
jgallagher, line 60: Installinator won't get this change because it doesn't call `find_nics()` directly; it calls `find_chelsio_links()` and `ensure_links_have_global_zone_link_local_v6_addresses()` directly: https://github.com/oxidecomputer/omicron/blob/051f524a63ae5a000b915f187c99e880536616f7/installinator/src/bootstrap.rs#L30-L40

This is because it wants to grab the first `link` to construct a bootstrap address: https://github.com/oxidecomputer/omicron/blob/051f524a63ae5a000b915f187c99e880536616f7/installinator/src/bootstrap.rs#L61-L65

I think (?) I wrote it that way because it didn't look like I could get from an `AddrObject` (returned by `find_nics()`) back to a `PhysicalLink` (required by `BootstrapInterface::GlobalZone.ip()`), but maybe that was wrong? Given an AddrObject, is `PhysicalLink(addrobj.interface().to_string())` its physical link? If so, I can follow up this PR with a change to installinator to use `find_nics()`.

sled-hardware/src/underlay.rs
```diff
@@ -44,13 +46,38 @@ pub enum Error {
...
     let underlay_nics = find_chelsio_links()?;
+
+    // Before these links have any consumers (eg. IP interfaces), set the MTU.
+    // If we have previously set the MTU, do not attempt to re-set.
+    const MTU: &str = "1600";
```
rcgoodfellow, line 65: Based on @rmustacc's comment [here](https://github.com/oxidecomputer/omicron/issues/3086#issuecomment-1544437094), I think we should use 9k as a L2 (data link) MTU. And then set the L3 (vnic) MTU on a case by case basis. Bootstrap vnic MTUs should be 1500 and others can be decided based on data path needs, i.e crucible vnics can be 9k etc. We should be careful to avoid fragmenting where possible. For example, right now the crucible zones have vnics with a 9k mtu over a 9k etherstub that egress through a 1600 physical MTU. This is likely causing performance issues.

If we can do something like modify the vnic create functions in omicron to require an MTU argument so we're not getting unfortunate implicit values, that would be good.

sled-hardware/src/underlay.rs
jmpesp: > Given an AddrObject, is PhysicalLink(addrobj.interface().to_string()) its physical link?

I believe this is true, yeah. My read of things is that `PhysicalLink` and the word interface are interchangeable, but that `Link` is different because it implements Drop and therefore has ownership consequences.

sled-hardware/src/underlay.rs
jmpesp: done in 4fcd69da261804f4408fd2e6c1bb345f501a70f4 and 5bb49d2b75693b7642d18593818414f5f8d93331

rcgoodfellow: In the latter we're passing the MTU in as an option. We should be more strict and require explicitly setting an actual MTU value. I think the option path will lead to situations of just specifying `None` and hoping for the best.

jmpesp: 96d2729c635e429b4cbe1b744af33b09fab8ec8b changes this to not optional

illumos-utils/src/link.rs
```diff
@@ -63,7 +63,7 @@ impl<DL: VnicSource + Clone> VnicAllocator<DL> {
         let name = allocator.next();
         debug_assert!(name.starts_with(VNIC_PREFIX));
         debug_assert!(name.starts_with(VNIC_PREFIX_CONTROL));
-        Dladm::create_vnic(&self.data_link, &name, mac, None)?;
+        Dladm::create_vnic(&self.data_link, &name, mac, None, 9000)?;
```
rcgoodfellow, line 66: This is not a request for change, just for clarity for my future self or other folks wondering about this 9000 value. The `oxControl<Thing><N>` interfaces are the ones that get handed to things like crucible (`oxControlStorage<N>`) and propolis (`oxControlInstance<N>`). For propolis, we are not talking about the customer ports that go through OPTE, but rather the control interfaces for propolis over which traffic such as API commands and live migration go over. In both of these cases we want to match the physical MTU of 9000.

rcgoodfellow (approved):

jmpesp:
The sled-agent that was deployed as part of `helios / deploy` failed with:

```
sled-agent: Failed to find address objects for maghemite: Could not set linkprop
```

There's no more error context telling us what's going on (fixed in e9be7a8802e1cc88fbf29e345a3c6b82cae337ed), but I believe this is due to this PR attempting to apply an MTU of 9000 to net0/1 but `tools/create_virtual_hardware.sh` does:

```
dladm set-linkprop -p mtu=1600 "net$I" # encap headroom
```

If I try and replicate this on the command line, I see an invalid argument error:

```
james@dinnerbone:~$ pfexec dladm create-simnet -t testsimnet0
james@dinnerbone:~$ pfexec dladm set-linkprop -p mtu=1600 testsimnet0
james@dinnerbone:~$ pfexec dladm set-linkprop -p mtu=9000 testsimnet0
dladm: warning: cannot set link property 'mtu' on 'testsimnet0': invalid argument
```

Checking the possible MTU values also shows that only 1600 is possible:

```
james@dinnerbone:~$ pfexec dladm show-linkprop -p mtu testsimnet0
LINK         PROPERTY        PERM VALUE          DEFAULT        POSSIBLE
testsimnet0  mtu             rw   1600           1500           1600
```

If I destroy and recreate the simnet, it shows that 9000 is possible:

```
james@dinnerbone:~$ pfexec dladm delete-simnet testsimnet0
james@dinnerbone:~$ pfexec dladm create-simnet -t testsimnet0
james@dinnerbone:~$ pfexec dladm show-linkprop -p mtu testsimnet0
LINK         PROPERTY        PERM VALUE          DEFAULT        POSSIBLE
testsimnet0  mtu             rw   9000           1500           9000
```

And indeed it is:

```
james@dinnerbone:~$ pfexec dladm set-linkprop -p mtu=9000 testsimnet0
james@dinnerbone:~$ 
```

ded47689cab22a91cc24b50e46aeaaa4ec40df69 removes the setting of MTU in create_virtual_hardware, because this PR is now taking care of it.

PR #3145, opened 2023-05-18
tufaceous-lib: `load` => `load_untrusted`

iliana:
This renames the `load` family of functions in `tufaceous_lib::OmicronRepo` to `load_untrusted`, because these functions (out of necessity) trust any TUF repository regardless of the provenance of its keys.

This change makes it a little more obvious that these methods should never be used outside of rack bootstrapping or development tooling; namely, they should never show up in Nexus.

For the tufaceous binary, we may want to require the user pass a root.json or a `--untrusted` flag, as I can imagine us wanting to use that tool in automation in such a way that we trust a repository we're reading from. Other uses of these methods are appropriate.

### DIFF
--- a/tufaceous-lib/src/archive.rs
+++ b/tufaceous-lib/src/archive.rs
@@ -170,8 +170,8 @@ where
     /// Extracts this archive into the specified directory.
     ///
     /// Once this is completed, use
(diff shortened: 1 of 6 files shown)


sunshowers (approved):
Missed this, sorry. Looks great!

iliana:
(Checking clippy/doc on a local merge, then will merge here.)

PR #3147, opened 2023-05-18
Remove "from_start" parameter from serial stream.

lifning:
Removal of an API footgun, as this parameter really only makes sense for the history-retrieval GET endpoint, not the continuous websocket connection.

Eventually over an instance's lifespan, it makes less sense for a client connecting to try asking for bytes starting from the beginning of time, as we don't buffer everything forever and this eventually becomes impossible to satisfy.

(further explanation in this comment https://github.com/oxidecomputer/propolis/issues/409#issuecomment-1552351154 - a bug encountered while investigating what is likely an unrelated issue.)

### DIFF
--- a/nexus/types/src/external_api/params.rs
+++ b/nexus/types/src/external_api/params.rs
@@ -973,9 +973,17 @@ pub struct InstanceSerialConsoleRequest {
     /// Maximum number of bytes of buffered serial console contents to return. If the requested
     /// range runs to the end of the available buffer, the data returned will be shorter than
     /// `max_bytes`.
-    /// This parameter is only useful for the non-streaming GET request for serial console data,
-    /// and *ignored* by the streaming websocket endpoint.
     pub max_bytes: Option<u64>,
+}
+
+/// Forwarded to a propolis server to request the contents of an Instance's serial console.
+#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema, PartialEq)]
+pub struct InstanceSerialConsoleStreamRequest {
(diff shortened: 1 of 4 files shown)


nexus/types/src/external_api/params.rs
```diff
@@ -973,11 +973,19 @@ pub struct InstanceSerialConsoleRequest {
...
 }
 
+/// Forwarded to a propolis server to request the contents of an Instance's serial console.
+#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema, PartialEq)]
+pub struct InstanceSerialConsoleStreamRequest {
+    /// Name or ID of the project, only required if `instance` is provided as a `Name`
+    pub project: Option<NameOrId>,
+    /// Character index in the serial buffer from which to read, counting *backward* from the most
+    /// recently buffered data retrieved from the instance.
+    pub most_recent: Option<u64>,
```
david-crespo, line 986: Do we validate somewhere that this is below a certain maximum and 400 otherwise? Or do we accept any value but only return at most the whole buffer?

david-crespo (approved):

zephraph (approved):
Yeah, this looks a lot more intuitive. Thanks @lifning!

nexus/types/src/external_api/params.rs
lifning: propolis is the source of truth of how big the buffer is, and I think it's probably the right place from which to send an error-type close frame, which i'm working on now.

PR #3150, opened 2023-05-18
[wicket] Allow user to force RoT/SP upgrade

jgallagher:
When showing the start update pane, if the current RoT and SP versions are both different from the version in the uploaded TUF repo, the screen looks basically the same (but now left-justified, for reasons that will become clearer below):

force-none

If one of the two (the SP in this example) has a version that exactly matches the TUF repo, we now display a toggle-able line for the user to force an upgrade (also note the `Toggle` help):

force-sp

If both the RoT and SP have a version that exactly mathes the TUF repo, we display two lines with extra help for up/down (the RoT version in this screenshot doesn't actually match; I hacked up the check to display the line anyway to get this screenshot):

force-both

A couple notes:

* There is only one `ForceInstallState`, not one per SP. I think (?) this is right because if one is updating multiple sleds, one will presumably want the same setting for all of them?
* The wicketd side of this accepts the new bools, but doesn't do anything with them. That work will come in a followup PR.

Builds on #3143

### DIFF
--- a/wicket/src/ui/panes/update.rs
+++ b/wicket/src/ui/panes/update.rs
@@ -544,20 +544,39 @@ impl UpdatePane {
         );
         match cmd {
             Cmd::Up => {
-                let id_state = self
-                    .component_state
-                    .get_mut(&state.rack_state.selected)
-                    .unwrap();
-                id_state.prev_item();
+                if self.is_force_update_visible(state) {
+                    ForceUpdateSelectionState::from(&*state)
+                        .prev_component(state);
+                } else {
+                    let id_state = self
+                        .component_state
+                        .get_mut(&state.rack_state.selected)
+                        .unwrap();
+                    id_state.prev_item();
+                }
                 Some(Action::Redraw)
             }
             Cmd::Down => {
-                let id_state = self
-                    .component_state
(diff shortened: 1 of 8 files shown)


andrewjstone (approved):
Ship it!

wicket/src/ui/panes/update.rs
```diff
@@ -1018,6 +1055,160 @@ impl UpdatePane {
...
+        if self.num_spans() > 1 {
+            state.force_update_state.prev_component();
+        }
+    }
+
+    fn help_text(&self) -> Vec<(&'static str, &'static str)> {
+        match self.num_spans() {
+            0 => vec![],
+            1 => vec![("Toggle", "Space")],
+            _ => vec![("Toggle", "Space"), ("Up", "Up"), ("Down", "Down")],
```
sunshowers, line 1169: I think this needs to be `<Space>` etc for consistency.

jgallagher: So it does! Thanks.

jgallagher: :+1: fixed in c38e13e

PR #3151, opened 2023-05-18
Add ensure command to dpd-client

### LINKED ISSUE #2982: Switch nat mapping conflict should not prevent service zone starting
In a test run today, the NTP zone was failing to come up because the NAT configuration for this zone was already resident on the switch. This should not be a blocking issue. The saga should detect that the NAT setup it needs is already resident on the switch and return successfully.

jmpesp:
Move logic that ensures a NAT entry exists instance.rs into new ensure function for dpd-client, and call it from both instance.rs (when setting an instance's nat entry) and from services.rs (when setting a service's nat entry).

Fixes #2982


dpd-client/src/lib.rs
```diff
@@ -28,3 +29,106 @@ pub struct ClientState {
...
+                            self.nat_ipv6_delete(
+                                &network.ip(),
+                                target_first_port,
+                            )
+                            .await
+                        }
+                    }?;
+                } else {
+                    info!(log,
+                      "nat entry with expected internal ip exists";
```
luqmana, line 86: Looks like this was basically existing code, but why is comparing just the internal ip enough? @internet-diglett might have some more context

jmpesp: Note it checks the response of `nat_ipv[46]_get(&network.ip(), target_first_port)` which includes the first port of the range in its query

gjcolombo: This one's on me. This code used to live in the instance creation saga and checked only for the existence of a NAT entry before continuing. In #3127, I extracted that out to a common routine to use both here and when a migration completes. In the latter case we need to update incorrect existing entries, which led to the creation of this code. When @jmpesp and I were testing that change we discovered that the code couldn't just call `nat_ipv?_create` unconditionally if an entry already existed. I think I added this check to try to avoid needlessly deleting and recreating identical entries, but it's entirely possible that we need to refine it.

@internet-diglett would love to get your feedback here--mea culpa for not checking with you on this before merging #3127.

jmpesp:
I've deployed this to the Canada region and it looks ok.

gjcolombo (approved):

dpd-client/src/lib.rs
```diff
@@ -28,3 +29,106 @@ pub struct ClientState {
...
+        match existing_nat {
+            Ok(existing) => {
+                let existing = existing.into_inner();
+                if existing.internal_ip != *sled_ip_address {
+                    info!(log, "deleting old nat entry";
+                      "target_ip" => ?target_ip);
+
+                    match target_ip {
+                        ipnetwork::IpNetwork::V4(network) => {
+                            self.nat_ipv4_delete(
```
rcgoodfellow, line 70: There is an opportunity for a race condition between these delete calls, and the calls to `nat_ipv[4,6]_create` below. Perhaps this is ok if we assume Nexus will never be operating on a single NAT entry concurrently (I don't know enough about Nexus to know if that is a reasonable assumption). I'm currently working on a [similar function](https://github.com/oxidecomputer/dendrite/pull/333) inside Dendrite itself to combine the delete/create stages of the ensure operation in a region guarded by a mutex.

gjcolombo: > Perhaps this is ok if we assume Nexus will never be operating on a single NAT entry concurrently (I don't know enough about Nexus to know if that is a reasonable assumption).

Nexus shouldn't be doing this, at least for iinstance NAT entries. During instance creation (IIUC) only one Nexus should be running the instance's create saga. If an instance migrates, Nexus won't allow a new migration to begin (which could kick off another parallel round of updates) until it's updated NAT entries to reflect the migration that just finished.

dpd-client/src/lib.rs
internet-diglett: Apologies for the delayed response. For some reason Github didn't send me an email, or I overlooked it? If I recall correctly, the internal ip address and the first port are the keys used to lookup nat entries in dpd:

https://github.com/oxidecomputer/dendrite/blob/60a7c97d08d9fc26cda1e133f7f25e65895fe167/dpd/src/nat.rs#L371-L383

As for feedback on the currently implemented process of updating NAT entries, admittedly I'm still learning more about how Nexus works under the hood, but as long as we're not at risk of a race condition with regards to asic updates we should be ok (your answer to @rcgoodfellow basically satisfies any concerns I had).

PR #3154, opened 2023-05-18
[MGS] update hubtools dep

jgallagher:


jmpesp (approved):

gateway/src/http_entrypoints/conversions.rs
```diff
@@ -128,6 +128,30 @@ impl From<gateway_messages::SpState> for SpState {
...
 
+impl From<gateway_messages::SpStateV2> for SpState {
+    fn from(state: gateway_messages::SpStateV2) -> Self {
+        Self {
+            serial_number: stringify_byte_string(&state.serial_number),
+            model: stringify_byte_string(&state.model),
+            revision: state.revision,
+            hubris_archive_id: hex::encode(state.hubris_archive_id),
+            base_mac_address: state.base_mac_address,
+            version: ImageVersion { epoch: 0, version: 0 }, // TODO FIXME
```
jgallagher, line 139: This is Definitely Wrong:tm:, as are a few other quick fixes in this file. They shouldn't hurt anything for now, and I'll fix them in the next couple of days as part of the MUPdate work that wants to use these fields. Will merge with the gross hacks in place to unbreak building.

PR #3159, opened 2023-05-19
[installinator] Report progress on a background task

jgallagher:
Both last week and this week when trying to mupdate dogfood, we would see the first sled or two work, but then future sled mupdates would hang with installinator timing out trying to download artifacts from wicketd (even though it was able to send progress reports). With a lot of help from @rcgoodfellow I think I tracked this down to a channel block between tasks: our downloading task gets blocked [trying to send progress reports](https://github.com/oxidecomputer/omicron/blob/5280a10fcb529b5e54ae8b7fd37737d53d66b57b/installinator/src/peers.rs#L319-L324) because the UpdateEngine is blocked [trying to send events to the progress reporter](https://github.com/oxidecomputer/omicron/blob/5280a10fcb529b5e54ae8b7fd37737d53d66b57b/update-engine/src/engine.rs#L655-L765) because the progress reporter is [waiting to send reports to all peers](https://github.com/oxidecomputer/omicron/blob/5280a10fcb529b5e54ae8b7fd37737d53d66b57b/installinator/src/reporter.rs#L116-L117).

This PR changes the progress reporter to send reports to all peers on a background task, allowing it to still receive events from the UpdateEngine promptly. I tested this on `BRM42220057` in dogfood, and it does appear to fix the problem. I am not sure this is the best fix though - open to suggestions!

### DIFF
--- a/installinator/src/reporter.rs
+++ b/installinator/src/reporter.rs
@@ -9,7 +9,7 @@ use std::time::Duration;
 use display_error_chain::DisplayErrorChain;
 use futures::{Future, StreamExt};
 use installinator_common::{Event, EventBuffer};
-use tokio::{sync::mpsc, time};
+use tokio::{sync::mpsc, task::JoinHandle, time};
 use uuid::Uuid;
 
 use crate::{errors::DiscoverPeersError, peers::Peers};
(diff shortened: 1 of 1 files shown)


sunshowers (approved):
Looks great, thank you so much for catching this!

Issue #3158, opened 2023-05-19
Use ed25519 /ristretto with vssrs


andrewjstone:
We use ed25519 as our preferred curve when we can. However ed25519 is not a prime order curve and so it can't safely be used with shamir secret sharing. However we can use ristretto to alleviate the issue and maintain the benefits of ed25519. vsss_rs has support for this.

https://ristretto.group/what_is_ristretto.html
https://soatok.blog/2022/05/19/guidance-for-choosing-an-elliptic-curve-signature-algorithm-in-2022/

andrewjstone:
Fixed in #3167

PR #3167, opened 2023-05-22
[trust-quorum] Start implementing version 0

### LINKED ISSUE #1513: Use Pedersen Scheme instead of Feldman scheme for rack secret verification
The Pedersen scheme doesn't leak info about the rack secret and so it has information theoretic security. In other words, an unbounded adversary cannot retrieve any information about the rack secret from a Pedersen commitment. This is not true for Feldman and so we prefer Pedersen. It should, however, be noted that Feldman is still quite secure, and only if ECDLP can be broken is it vulnerable.

andrewjstone:
This is the start of a stripped down version of the trust quorum bootstore for use in our MVP racks. We define "schemes", such that each version of trust quorum can get stronger security wise over time, and we can upgrade into new schemes. This change starts to implement `bootstore scheme version 0`, where all important parameters exist in a self describing zero-sized struct, `V0Scheme`, that can be debug printed for informational purposes, such as logging in sled-agent startup.

This commit implements the first part of the V0 scheme, which is the necessary datastructures around trust quorum. As these structures are all security sensitive, none of them are going to be leaving the bootstore logic itself. The sole exception to this is that the rack secret will be handed off to the `KeyManager` in sled-agent in order to derive disk encryption keys. This will come in a follow up commit.

There are two primary datastructures introduced here:
 * `RackSecret` - A thin wrapper around shamir secret sharing using Curve25519
 * `SharePkg` - A container of information distributed to each of the initial sleds that are part of the rack during rack initialization

`RackSecret` itself will never be stored on disk, and `SharePkg` and its analog `LearnedSharePkg` for sleds that were not part of the original rack membership, will be stored on the M.2s on each sled in a rack. We plan to extract the `Ledger` from sled-agent into `omicron-common` and use that to synchronize writing across both M.2s.  We include a `SharePkgHeader` that can be used to distinguish what type of share header is stored on the M.2s so that we can properly deserialize them. We expect this structure to be remain backwards compatible indefinitely.

For each sled that is part of the initial group, we include a set of encrypted shares in its `SharePkg`. This is to allow other sleds to get added to the rack and get their own key-shares over time. These shares are encrypted so that an attacker stealing fewer than threshold sleds (or their M.2s) cannot recover the rack secret. When a new sled is added it will ask for a share from one of the existing sleds which will return it a share and mark it as used on its M.2s. In order to return a share, an existing sled will have to retrieve enough shares to reconstruct the rack secret and decrypt the shares it has in its `SharePkg`. If it does not have any shares remaining, the asking sled will have to go and ask another sled. This is a best-effort mechanism with known security and reliability issues. However, it allows us to ship a version of trust quorum that will be able to survive sled additions long enough for us to implement trust quorum V1.

In scheme version 0, shares and packages will be distributed between sleds over unencrypted tcp links rather than sprockets sessions. We also will distribute them to anyone on the bootstrap network who asks. This is a strategy designed to allow our primary goal of preventing offline data recover from stolen U.2 devices at MVP. It aligns well with our current control-plane limitations around unencrypted transport that we plan to fix over time.

There will be a few follow up commits in order to complete scheme v0 and integrate it into our system:

 1. Augment the bootstore with the V0 protocol over TCP channels that allows `RackSecret` reconstruction and `SharePkg` distribution
 2. Integrate this protocol into the bootstrap agent with calls into the bootstore and remove the ad-hoc code that currently exists.
 3. Add a `StorageKeyRequester` for the the v0 trust quorum so that the `KeyManager` can generate disk encryption keys using the `RackSecret` as input key material (IKM)

As part of this commit, almost all the previous bootstore code was deleted. It is obsolete, as we are planning to use a quorum based, rather than 2 phase commit based strategy for reconfiguration. We also plan to separate the persistence layer from the stored data and not use sqlite transactions as the implementation of invariants for our protocol. This should make testing easier, and allow us a choice of using async or sync connections to sqlite, or even abandoning sqlite in favor of flat files or another persistence strategy. The new protocol still needs to be written as an update to RFD 355.

Fixes #1513 and #3158


bootstore/src/trust_quorum/rack_secret.rs
```diff
@@ -54,31 +52,12 @@ impl PartialEq for RackSecret {
...
-    pub fn verify(&self, share: &Share) -> bool {
-        self.verifier.verify(share)
-    }
-}
-
-// Temporary until the using code is written
 impl RackSecret {
     /// Create a secret based on the NIST P-256 curve
     pub fn new() -> RackSecret {
+        // For some reason curve25519-dalek requires rand_core here
```
andrewjstone, line 58: I fixed whatever was causing that. Need to remove the obsolete comment.

bootstore/src/trust_quorum/share_pkg.rs
```diff
@@ -0,0 +1,319 @@
...
+use secrecy::Secret;
+use serde::{Deserialize, Serialize};
+use sha3::{Digest, Sha3_256};
+use std::collections::BTreeMap;
+use std::fmt;
+use uuid::Uuid;
+use zeroize::Zeroizing;
+use zeroize::{Zeroize, ZeroizeOnDrop};
+
+// Each share is 33 bytes (one identifier (x-index) byte, and one 32-byte
```
andrewjstone, line 20: should say (x-point) and y-point

bootstore/src/trust_quorum/share_pkg.rs
```diff
@@ -0,0 +1,319 @@
...
+
+    pub fn new_v0_learned() -> SharePkgHeader {
+        SharePkgHeader {
+            version: 0,
+            metadata: BTreeMap::from([("learned".into(), "true".into())]),
+        }
+    }
+}
+
+/// A container used distributed among trust quorum participants for
```
andrewjstone, line 49: typo: remove `used`

bootstore/src/trust_quorum/share_pkg.rs
```diff
@@ -0,0 +1,319 @@
...
+/// trust quorum version 0 Scheme: [`crate::SchemeV0`].
+///
+/// This scheme does not verify membership, and will hand out shares
+/// to whoever asks.
+#[derive(Clone, PartialEq, Serialize, Deserialize, Zeroize, ZeroizeOnDrop)]
+pub struct SharePkgV0 {
+    #[zeroize(skip)]
+    pub rack_uuid: Uuid,
+    // We aren't planning on doing any reconfigurations with this version of
+    // the protocol
```
andrewjstone, line 59: Ugh, never finished this comment. It should read "epoch is there just in case"

bootstore/src/trust_quorum/share_pkg.rs
```diff
@@ -0,0 +1,319 @@
...
+
+    /// Nonce used for encryption.
+    ///
+    /// We generate all pkgs during rack init and ensure uniqueness of all nonces
+    pub nonce: [u8; 12],
+
+    /// We include a distinct subset of unused shares for each sled in the
+    /// initial  group. This allows the sled to hand out unique shares when a
+    /// new sled joins the cluster. We keep these encrypted so that a single
+    /// sled does not have enough unencrrypted shares to unlock the rack
```
andrewjstone, line 83: typo: `unencrypted`

bootstore/src/trust_quorum/share_pkg.rs
```diff
@@ -0,0 +1,319 @@
...
+    /// without participating in trust quorum.
+    ///
+    /// Each sled should keep local track of which shares it has already handed
+    /// out.
+    ///
+    /// Note, each sled derives the same encryption key, but uses the unique
```
andrewjstone, line 89: "uses the unique nonce for encrypting its unique shares"

bootstore/src/trust_quorum/share_pkg.rs
```diff
@@ -0,0 +1,319 @@
...
+
+/// Create a package for each sled
+pub fn create_pkgs(
+    rack_uuid: Uuid,
+    n: usize,
+) -> Result<Secret<Vec<SharePkgV0>>, TrustQuorumError> {
+    let rack_secret = RackSecret::new();
+    let threshold = n / 2 + 1;
+    let epoch = 0;
+    // We always generate 256 shares to allow new sleds to come online
```
andrewjstone, line 125: s/256/255

bootstore/src/trust_quorum/share_pkg.rs
```diff
@@ -0,0 +1,319 @@
...
+        pkgs.push(pkg);
+    }
+    Ok(Secret::new(pkgs))
+}
+
+// This is a fairly standard nonce construction consisting of a random part and
+// a counter.
+//
+// We know we only have up to 32 packages (1 for each sled), so we fill the
+// first 11 bytes of the nonce with random byes, and the last byte with a
```
andrewjstone, line 167: s/byes/bytes

andrewjstone:
I opened an issue to remind myself to remove the vsss-rs duplicate exclusion in ci: https://github.com/oxidecomputer/omicron/issues/3183

bootstore/src/trust_quorum/error.rs
```diff
@@ -4,17 +4,20 @@
...
     #[error("Rack secret construction failed: {0:?}")]
-    RackSecretConstructionFailed(vsss_rs::Error),
+    Vsssrs(vsss_rs::Error),
+
+    // Failed to encrypt something
+    #[error("Failed to encrypt")]
+    FailedToEncrypt,
+
+    // Failed to decrypt something
+    #[error("Failed to encrypt")]
```
internet-diglett, line 21:
```suggestion
    #[error("Failed to decrypt")]
```

bootstore/src/trust_quorum/share_pkg.rs
```diff
@@ -0,0 +1,320 @@
...
+    let mut salt = [0u8; 32];
+    OsRng.fill_bytes(&mut salt);
+    let cipher = derive_encryption_key(&rack_uuid, &rack_secret, &salt);
+    let mut pkgs = Vec::with_capacity(n);
+    for i in 0..n {
+        // Each pkg gets a distinct subset of shares
+        let mut iter =
+            shares.iter().skip(i * shares_per_sled).take(shares_per_sled);
+        let share = iter.next().unwrap();
+        let plaintext: Vec<u8> = iter.fold(Vec::new(), |mut acc, x| {
```
inickles, line 140:
```suggestion
        let plaintext: Vec<u8> = iter.fold(Vec::with_capacity(shares_per_sled - 1), |mut acc, x| {
```

maybe save some an alloc or two.

bootstore/src/trust_quorum/share_pkg.rs
```diff
@@ -0,0 +1,320 @@
...
+    let shares = rack_secret.split(threshold, total_shares)?;
+    let share_digests = share_digests(&shares);
+    let mut salt = [0u8; 32];
+    OsRng.fill_bytes(&mut salt);
+    let cipher = derive_encryption_key(&rack_uuid, &rack_secret, &salt);
+    let mut pkgs = Vec::with_capacity(n);
+    for i in 0..n {
+        // Each pkg gets a distinct subset of shares
+        let mut iter =
+            shares.iter().skip(i * shares_per_sled).take(shares_per_sled);
```
inickles, line 138: I'm thinking this loop could be marginally easier to follow by iterating over `shares.chunks(shares_per_sled)`:

```
for chunk in shares.chunks(shares_per_sled) {
        let iter = chunk.iter();
...
```

bootstore/src/trust_quorum/share_pkg.rs
```diff
@@ -0,0 +1,320 @@
...
+// This is a fairly standard nonce construction consisting of a random part and
+// a counter.
+//
+// We know we only have up to 32 packages (1 for each sled), so we fill the
+// first 11 bytes of the nonce with random bytes, and the last byte with
+// a counter.
+fn new_nonce(i: usize) -> [u8; 12] {
+    let mut nonce = [0u8; 12];
+    OsRng.fill_bytes(&mut nonce[..11]);
+    nonce[11] = u8::try_from(i).unwrap();
```
inickles, line 173: why take `i` as `usize` instead of `u8` in the first place?

bootstore/src/trust_quorum/share_pkg.rs
```diff
@@ -0,0 +1,320 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("SharePkgV0")
+            .field("rack_uuid", &self.rack_uuid)
+            .field("epoch", &self.epoch)
+            .field("threshold", &self.threshold)
+            .field("share", &"Share")
+            .field("share_digests", &self.share_digests)
+            .finish()
+    }
+}
```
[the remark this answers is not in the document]
andrewjstone: I started to do this and realized that I won't be able to derive Serialize/Deserialize if I do this. I already wrap the returned packages in a `Secret` which makes this all safe enough. By doing that, I also don't end up debug printing a lot of encrypted bytes, which while mostly harmless, takes up space in output and makes things hard to read. The manual `Debug` impl doesn't have this problem.

bootstore/src/trust_quorum/share_pkg.rs
andrewjstone: Maybe, but I'd still need the index for the nonce, so I'd have to do something like:

`for (i, chunk) in shares.chunks(shares_per_sled).enumerate()`

I'm also not sure that I'd actually end up with the exact right number of chunks, since leftovers may be present in smaller chunks, which is why I went down the explicit count route.

bootstore/src/trust_quorum/share_pkg.rs
andrewjstone: All the values that go into generating i are usizes, because all the APIs were working with take them. So there has to be a conversion somewhere. I did it in the function to not clutter `create_pkgs` anymore.

bootstore/src/trust_quorum/share_pkg.rs
andrewjstone: This isn't quite right, as it doesn't include the length of each share. This is a vec of all the plaintexts of the shares concatenated together. 

However, this concatenated plaintext should be wrapped in a secret. 

I fixed both of these in [671d4b1](https://github.com/oxidecomputer/omicron/pull/3167/commits/671d4b1c7b8cf395db8f25beeadc013668f9e594)

bootstore/src/trust_quorum/share_pkg.rs
```diff
@@ -0,0 +1,320 @@
...
+pub fn create_pkgs(
+    rack_uuid: Uuid,
+    n: usize,
+) -> Result<Secret<Vec<SharePkgV0>>, TrustQuorumError> {
+    let rack_secret = RackSecret::new();
+    let threshold = n / 2 + 1;
+    let epoch = 0;
+    // We always generate 255 shares to allow new sleds to come online
+    let total_shares = 255;
+    let shares_per_sled = 255 / n;
```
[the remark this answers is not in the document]
andrewjstone: Fixed in [671d4b1](https://github.com/oxidecomputer/omicron/pull/3167/commits/671d4b1c7b8cf395db8f25beeadc013668f9e594)

bootstore/src/trust_quorum/share_pkg.rs
```diff
@@ -0,0 +1,320 @@
...
+/// trust quorum scheme version 0: [`crate::SchemeV0`].
+///
+/// This scheme does not verify membership, and will hand out shares
+/// to whomever asks.
+#[derive(Clone, PartialEq, Serialize, Deserialize, Zeroize, ZeroizeOnDrop)]
+pub struct SharePkgV0 {
+    #[zeroize(skip)]
+    pub rack_uuid: Uuid,
+    // We aren't planning on doing any reconfigurations with this version of
+    // the protocol. This is here in case we decide otherwise, and becaus an
```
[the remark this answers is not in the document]
andrewjstone: Fixed in [671d4b1](https://github.com/oxidecomputer/omicron/pull/3167/commits/671d4b1c7b8cf395db8f25beeadc013668f9e594)

bootstore/src/trust_quorum/rack_secret.rs
```diff
@@ -87,58 +65,52 @@ impl RackSecret {
         &self,
         threshold: usize,
         total_shares: usize,
-    ) -> Result<(Vec<Share>, Verifier), vsss_rs::Error> {
+    ) -> Result<Vec<Vec<u8>>, vsss_rs::Error> {
```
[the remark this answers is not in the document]
andrewjstone: Fixed in [671d4b1](https://github.com/oxidecomputer/omicron/pull/3167/commits/671d4b1c7b8cf395db8f25beeadc013668f9e594)

bootstore/src/trust_quorum/rack_secret.rs
```diff
@@ -43,7 +41,7 @@ use zeroize::{Zeroize, ZeroizeOnDrop};
 /// type and orthogonal to its implementation.
 #[derive(Zeroize, ZeroizeOnDrop)]
 pub struct RackSecret {
-    secret: NonZeroScalar,
+    secret: Scalar,
```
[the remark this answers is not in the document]
andrewjstone: Fixed in [671d4b1](https://github.com/oxidecomputer/omicron/pull/3167/commits/671d4b1c7b8cf395db8f25beeadc013668f9e594)

bootstore/src/trust_quorum/error.rs
andrewjstone: Fixed in [671d4b1](https://github.com/oxidecomputer/omicron/pull/3167/commits/671d4b1c7b8cf395db8f25beeadc013668f9e594)

bootstore/src/trust_quorum/share_pkg.rs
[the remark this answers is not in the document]
inickles: > I started to do this and realized that I won't be able to derive Serialize/Deserialize if I do this. 

Though you would be able to by enabling the optional `serde` feature on the Secrecy crate? Looks like that's not set in Omicron atm, and it's not a default feature.

Not that I think you should, I see your points for skipping fields in Debug, just testing my understanding.

bootstore/src/trust_quorum/share_pkg.rs
inickles: My preference would be to avoid the possibility of panic caused by the` unwrap()`, even though may currently be used in such a way where that won't happen*, so either do the conversion here and make the return type be a Result or do the conversion in `create-pkgs`.*

*Looking again at `create_pkgs` it looks like it can only work with values of `n` that fit within a `u8` in the first place, because of:
```
    let total_shares = 255;
    let shares_per_sled = total_shares / n;
```
and so its type might as well be `u8` too then? That way there's no more clutter there, and one wouldn't be left with unexpected results trying to use this code with larger values.

andrewjstone: > *Looking again at `create_pkgs` it looks like it can only work with values of `n` that fit within a `u8` in the first place, because of:
> 
> ```
>     let total_shares = 255;
>     let shares_per_sled = total_shares / n;
> ```
> 
> and so its type might as well be `u8` too then? That way there's no more clutter there, and one wouldn't be left with unexpected results trying to use this code with larger values.

Ha, I went to do exactly that but then realized that the calls to `RackSecret::split` take a usize because that's what's taken underneath in the `vsss_rs::shamir::split_secret` call. However there's no reason I can't constrain the type more in `RackSecret::split` then expand it later via `as usize`. The internals of `vsss_rs` limit the number of shares anyway, so limiting to a `u8` is no loss, and makes things return fewer errors.

bootstore/src/trust_quorum/share_pkg.rs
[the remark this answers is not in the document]
andrewjstone: I took a look and the `serde` feature on `Secrecy` only implements `DeserializeOwned`. You can implement `Serialize` with a manual marker trait impl of [SerializableSecret](https://docs.rs/secrecy/latest/secrecy/trait.SerializableSecret.html), but that only works on a specific type, so I would need to create a wrapper type for `Vec<u8>` to get around rust coherence rules. Given that shares never leave the bootstore crate, I'm loath to make this any more complicated than it already is.

bootstore/src/trust_quorum/share_pkg.rs
andrewjstone: Fixed in [2ec7776](https://github.com/oxidecomputer/omicron/pull/3167/commits/2ec7776cb7e7fc8455ecd1ba0bdeb82e73519d8f)

PR #3181, opened 2023-05-22
Add propolis and switch zone images to OS ramdisk

### LINKED ISSUE #3166: Fetch propolis image from boot disk or install it to /opt/oxide on ramdisk
After mupdating and installinating a rack, we wound up with sled agents attempting to run propolis instance zones from a zone image in `/opt/oxide`.

https://github.com/oxidecomputer/omicron/blob/86676b9dafc67596f1ccd50f921f2288fcb14684/sled-agent/src/instance.rs#L791-L813

However from the install, the propolis zone image was on the M.2 boot disk and not in `/opt/oxide`. This resulted in all instances failing to launch. So we either need to update the code above to handle the propolis zone image being on the boot disk (similar to service zones), or update the ramdisk to contain the propolis zone in `/opt/oxide`.

jgallagher:
The primary change in this PR is to add `propolis-server.tar.gz` and `switch-asic.tar.gz` (renamed to `switch.tar.gz` within the ramdisk) to the OS image produced by CI. This may affect anyone using the same scripts CI uses to build OS images, so a bit of detail:

* `propolis-server.tar.gz` is unconditionally added to `global-zone-packages.tar.gz` produced by the CI packaging script. `tools/build-host-image.sh` was already using that file to populate the gz, so this gets in "for free".
* We can't add the switch zone to `global-zone-packages.tar.gz` because we don't know which variant we want at that point. Instead, `tools/build-host-image.sh` takes a `-S path-to-switch-zone-tarball` argument; if present, it will copy that file into `/opt/oxide/switch.tar.gz` in the gz.

I made a few small related cleanups while I was here:

* `mg-ddm.tar.gz` is now marked as an intermediate (it goes into the switch zone during packaging, but isn't needed outside of that)
* The TUF repo built by CI no longer includes `mg-ddm.tar.gz` (by virtue of the previous point) or `propolis-server` (by explicitly grepping it out).
* Bumped our helios pin to the latest.

I installed the OS produced on this branch on `madrid`, and see the following under `/opt/oxide`:

```
BRM44220001 # find /opt/oxide -type f | sort
/opt/oxide/mg-ddm/ddmadm
/opt/oxide/mg-ddm/ddmd
/opt/oxide/mg-ddm/pkg/ddm/manifest.xml
/opt/oxide/mg-ddm/pkg/ddm_method_script.sh
/opt/oxide/opte/bin/opteadm
/opt/oxide/propolis-server.tar.gz
/opt/oxide/sled-agent/VERSION
/opt/oxide/sled-agent/pkg/README.md
/opt/oxide/sled-agent/pkg/config.toml
/opt/oxide/sled-agent/sled-agent
/opt/oxide/switch.tar.gz
```

I believe this is correct; the remaining control plane zone tarballs should be installed under `/pool/int/*/install` during mupdate.

Fixes #3166.


iliana (approved):
lgtm!

Issue #3103, opened 2023-05-12
wicketd requests to MGS fail with HTTP 400 and a 0-length response body


jgallagher:
Today when trying to update 6 gimlets in the dogfood rack simultaneously, three of the updates failed partway through because wicketd got an HTTP 400 with no body response from MGS. There was no corresponding entry in the MGS logs, which makes it very likely `hyper` itself was sending the 400, which it does if it receives a non-HTTP request. https://github.com/hyperium/hyper/issues/3225 is a request for better after-the-fact debugging support from hyper, but in the meantime, the next time we try to mupdate the dogfood rack we should snoop the localhost traffic between wicketd and MGS in hopes of catching details on what's causing these 400s:

```
snoop -o mgs-traffic.snoop -x0 -d lo0 port 12225
```

We should (probably?) also add some kind of retries around at least some of the requests wicketd makes of MGS.

davepacheco:
Could this have been caused by oxidecomputer/stlouis#454?

jgallagher:
> Could this have been caused by [oxidecomputer/stlouis#454](https://github.com/oxidecomputer/stlouis/issues/454)?

It's certainly possible. This occurred when installinator was incorrectly sending extremely large progress reports to wicketd, and we never saw it after we fixed that issue - I could believe that those reports pushed wicketd's heap up into the bad VA range and cause it to send malformed a few HTTP requests, and fixing the large reports kept us out of the bad VA range since then.

PR #3188, opened 2023-05-22
[wicket] add a way to clear update state

sunshowers:
Part of the state machine flow for updates. Clearing update state can
only be done once the update completes or fails.

I originally wanted the shortcut to be Ctrl-Shift-R, but Ctrl-Shift
doesn't work as a modifier in many terminals it turns out.

Screenshot of failure:

image


wicketd/src/http_entrypoints.rs
```diff
@@ -328,6 +336,43 @@ async fn get_update_sp(
...
+async fn post_clear_update_state(
+    rqctx: RequestContext<ServerContext>,
+    target: Path<SpIdentifier>,
+    opts: TypedBody<ClearUpdateStateOptions>,
+) -> Result<HttpResponseUpdatedNoContent, HttpError> {
+    let log = &rqctx.log;
+    let target = target.into_inner();
+
+    let opts = opts.into_inner();
+    match opts.test_error {
```
jgallagher, line 355: Could this be a method on `UpdateTestError`? I'm imagining both this and `post_start_update` could do something like

```rust
if let Some(test_error) = opts.test_error.take() {
    return Err(test_error.into_http_error())
}
```

instead of both handling all variants.

wicketd/src/http_entrypoints.rs
```diff
@@ -286,13 +294,13 @@ async fn post_start_update(
...
             return Err(HttpError::for_bad_request(
                 None,
                 "Simulated failure while starting update".into(),
             ));
         }
-        Some(UpdateTestError::StartTimeout { secs }) => {
+        Some(UpdateTestError::Timeout { secs }) => {
             slog::info!(log, "Simulating timeout while starting update");
             // 15 seconds should be enough to cause a timeout.
             tokio::time::sleep(Duration::from_secs(secs)).await;
```
jgallagher, line 306: I realize this was here before, but I think we probably want to either return an http error at this point ("we expected you to time out by now"?) or rename `Timeout` to something like `DelayResponse` to indicate that we will still respond, just after the specified number of seconds of delay.

wicketd/src/update_tracker.rs
```diff
@@ -380,6 +388,26 @@ impl UpdateTrackerData {
...
     }
 
+    fn clear_update_state(
+        &mut self,
+        sp: SpIdentifier,
+    ) -> Result<(), ClearUpdateStateError> {
+        // Is an update currently running? If so, then reject the request.
+        //
+        // TODO: maybe add a way to force state to be cleared (this is
+        // dangerous!)
```
jgallagher, line 398: Is "force state to be cleared" a different operation from "cancel a running update"? If we had cancelling a running update (which we should have eventually), would that remove the need to force a state clear?

wicket/src/runner.rs
```diff
@@ -254,6 +229,44 @@ impl RunnerCore {
     }
 }
 
+fn get_update_test_error(
+    env_var: &str,
+) -> Result<Option<UpdateTestError>, anyhow::Error> {
+    // 15 seconds should be enough to cause a timeout.
+    const DEFAULT_TEST_TIMEOUT_SECS: u64 = 15;
```
jgallagher, line 236: The default progenitor client timeout is currently [15 seconds](https://github.com/oxidecomputer/progenitor/blob/6af969081a475d85e686960b917f2ee6e24251e9/progenitor-impl/src/lib.rs#L351), so maybe bump this up to 16 with a comment to that effect?

wicket/src/wicketd.rs
```diff
@@ -153,6 +163,41 @@ impl WicketdManager {
...
+        &self,
+        component_id: ComponentId,
+        options: ClearUpdateStateOptions,
+    ) {
+        let log = self.log.clone();
+        let addr = self.wicketd_addr;
+        let events_tx = self.events_tx.clone();
+        tokio::spawn(async move {
+            let update_client =
+                create_wicketd_client(&log, addr, WICKETD_TIMEOUT);
```
jgallagher, line 176: This wasn't changed as part of this PR, but I bumped into it while testing it. I suspended wicketd and hit the "clear update state" to see what would happen with an unresponsive wicketd, and I got a timeout error after just 1 second, which tracks back to `WICKETD_TIMEOUT` here. Some random thoughts:

* 1 second is probably too aggressive even for wicket -> wicket being on localhost. Maybe 5-10 seconds?
* I'm a little surprised we haven't seen problems due to this.
* Is it possible we actually have, particularly #3103? Could we trigger that "400 no body" response from MGS's hyper if we timed out while we were still sending a request somehow?

wicket/src/ui/panes/update.rs
```diff
@@ -460,6 +469,64 @@ impl UpdatePane {
         frame.render_widget(popup, full_screen);
     }
 
+    fn draw_clear_update_state_waiting_popup(
+        &self,
+        state: &State,
+        frame: &mut Frame<'_>,
+    ) {
+        // What we show here depends on the current status.
```
jgallagher, line 477: Is this comment still accurate? It looks like this only ever shows the "waiting" status, and a different method is responsible for showing errors.

wicket/src/ui/panes/update.rs
```diff
@@ -757,6 +858,50 @@ impl UpdatePane {
...
+                            }
+                            (false, _) => {
+                                // This message isn't meant for this component.
+                                // It's a bit of a weird situation (we should
+                                // only be making one start-update request at a
+                                // time, and shouldn't let
+                                // state.rack_state.selected be changed in the
+                                // meantime) so log this.
+                                slog::warn!(
+                            self.log,
```
jgallagher, line 893: Nit - indentation is off here

jgallagher (commented):
Thank you for this. It's been a real "I wish we had this" during the last couple dogfood mupdates; looking forward to having it for future ones!

wicket/src/ui/panes/update.rs
jgallagher: Ah, this was removed in #3194

wicketd/src/http_entrypoints.rs
sunshowers: Done, thanks.

wicketd/src/http_entrypoints.rs
sunshowers: I added an error with a notice that this shouldn't be hit.

wicketd/src/update_tracker.rs
sunshowers: Ah yeah, it is the same thing. Thanks.

wicket/src/runner.rs
sunshowers: Made it 30 seconds as a round number.

wicket/src/wicketd.rs
sunshowers: That's a good hypothesis. I've updated the wicket timeout to 5 seconds and added a comment about that.

wicket/src/ui/panes/update.rs
sunshowers: I removed them from here as well.

wicket/src/ui/panes/update.rs
sunshowers: Fixed, thanks.

jgallagher (approved):

PR #3194, opened 2023-05-23
[wicket] wrap all popup text, add padding

sunshowers:
Previously, we had a regime where we were inconsistently adding padding
to popups in wicket. However, padding does look really good!

With this change:

1. Now the popup is responsible for doing the padding, not users of it.
2. We always wrap text, both header and body (and use the prefix support in our textwrap
   impl to add padding on the left).

Depends on #3188 

**Screenshots:**

Before:

image

Before, with narrow display:

image

After:

image

After, with narrow display:

image


wicket/src/ui/panes/update.rs
```diff
@@ -532,13 +507,15 @@ impl UpdatePane {
...
-        let popup = self.ignition.popup(state.rack_state.selected);
         let full_screen = Rect {
             width: state.screen_width,
             height: state.screen_height,
             x: 0,
             y: 0,
         };
+        let (header, body, buttons) =
+            self.ignition.popup_components(state.rack_state.selected);
+        let popup = Popup::new(full_screen, &&header, &body, buttons);
```
jgallagher, line 518: Nit, typo here?

```suggestion
        let popup = Popup::new(full_screen, &header, &body, buttons);
```

wicket/src/ui/widgets/ignition.rs
```diff
@@ -48,36 +47,41 @@ impl IgnitionPopup {
...
-            buttons: vec![ButtonText { instruction: "CLOSE", key: "ESC" }],
-        }
+    /// Return the components to create a popup -- the header, body and button
+    /// text.
+    ///
+    /// Can't return a `Popup` here due to lifetime issues.
+    pub fn popup_components(
+        &self,
+        component: ComponentId,
+    ) -> (Spans<'static>, Text<'static>, Vec<ButtonText>) {
```
jgallagher, line 57: This seems fine since all three elements are different types, but would it be cleaner to have a `PopupComponents` type?

jgallagher (approved):
Looks great :+1:

wicket/src/ui/widgets/ignition.rs
sunshowers: I think this is a good idea. What I'm going to do is rename the old `Popup` into `PopupComponents` and make `Popup` derive from that. That should also clean up the old PR a bit.

sunshowers: Done, this is a lot cleaner now. Thank you!

PR #3203, opened 2023-05-23
[console] Local (username/password) login page

david-crespo:
- [x] Delete `/spoof_login` and redirects to it
  - Note this was misleadingly named and has nothing to do with the header-based spoof login scheme (which is still in place) other than relying on the same fake privileged and unprivileged users
- [ ] When we don't know the current silo from the subdomain (because silo name in subdomain isn't done yet) `/login` should redirect to `/login/{silo_name}/local` with the first silo it finds
- [ ] TBD


david-crespo:
Closing in favor of #3217. Will get to this stuff incrementally.

PR #3205, opened 2023-05-23
[update-engine] add an optional message for success operations

sunshowers:
Also make metadata optional since in the majority of cases we aren't
returning any metadata.

With this, the new states and their requirements are:

| **kind** | **message** |     **metadata**     |
|:--------:|:-----------:|:--------------------:|
| success  | optional    | optional, completion |
| warning  | required    | optional, completion |
| skipped  | required    | optional, skipped    |

To express all of these separate cases, we have to introduce new
`StepSuccess`, `StepWarning` and `StepSkipped` builders.

Screenshot of a message attached to success:

image


jgallagher (approved):
LGTM :+1:

sunshowers:
Ah looks like my cherry-pick didn't go well. It's based on #3194 so I'll just wait until that is landed before looking at this.

Issue #3208, opened 2023-05-24
instance_serial_console_data should verify instance state


jordanhendricks:
An instance on rack2 panicked, and propolis-server was restarted. Sled-agent did not notice the state change (#3206), so the instance was still marked as running.

After propolis-server was restarted, requests for the serial console state continued:
```
[2023-05-23T15:28:28.400864013-07:00] ERROR: propolis-server/12186 on oxz_propolis-server_7ffa6f90-cbd3-407a-9c9a-aac81e975a40: Error returned from handler: HttpError { status_code: 500, error_code: Some("Internal"), external_message: "Internal Server Error", internal_message: "Server not initialized (no instance)" } (req_id=01e04ecf-9cc9-4583-b847-87728604da76, upgrade=websocket, uri=/instance/serial?most_recent=262144, method=GET, remote_addr=[fd00:1122:3344:106::4]:42599, local_addr=[fd00:1122:3344:101::23]:12400)
```

(The 500 here is https://github.com/oxidecomputer/propolis/issues/348.)

This led us to notice that `instance_serial_console_data` [does not check the instance state](https://github.com/oxidecomputer/omicron/blob/046e1e378e26916687d40fb1b71c18412f385213/nexus/src/app/instance.rs#L1367-L1370) before requesting the console history.

In the case of the stuck instance, a check of the state wouldn't have prevented these requests, as the instance was still marked running. But we should check instance state here.

PR #3210, opened 2023-05-24
Check instance state before constructing a propolis-server client

lifning:
(#3208)

In `instance_serial_console_data` (and `instance_serial_console_stream`[^1]), we must connect to propolis-server directly, but we hadn't been verifying that it makes sense to try (i.e. the instance is in a state in which a propolis-server is expected to exist)

(I'm assuming here that it may be useful to allow attempts to retrieve serial output from an instance in state `InstanceState::Repairing`)

[^1]: See #2871 for surfacing the error message to clients of `instance_serial_console_stream` when it occurs -- and for making connecting during `InstanceState::Migrating` make more sense

### DIFF
--- a/nexus/src/app/instance.rs
+++ b/nexus/src/app/instance.rs
@@ -19,6 +19,7 @@ use crate::external_api::params;
 use futures::future::Fuse;
 use futures::{FutureExt, SinkExt, StreamExt};
 use nexus_db_model::IpKind;
+use nexus_db_queries::authz::ApiResource;
 use nexus_db_queries::context::OpContext;
 use omicron_common::address::PROPOLIS_PORT;
 use omicron_common::api::external::http_pagination::PaginatedBy;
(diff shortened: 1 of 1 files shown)


nexus/src/app/instance.rs
```diff
@@ -1361,17 +1361,38 @@ impl super::Nexus {
...
+                    "http://{}",
+                    socket_addr
+                )))
+            }
+            InstanceState::Creating
+            | InstanceState::Starting
+            | InstanceState::Stopping
+            | InstanceState::Stopped
+            | InstanceState::Failed
+            | InstanceState::Destroyed => Err(Error::ServiceUnavailable {
```
luqmana, line 1389: Should `Destroyed` at least return a 404 instead of a 503 here?

luqmana (approved):
Minor q about status code aside, this seems like an improvement to the status quo. Thanks!

PR #3211, opened 2023-05-24
update steno dep to handle undo actions failing

davepacheco:
Depends on oxidecomputer/steno#138.

This updates Omicron for the small breaking change associated with having Steno not panic when undo actions fail.  The breakage is basically that there's a new kind of error that can be produced by running a saga and we want to handle that specially (include that in the error message, log loudly, and report the saga as "stuck" over the debug API).


nexus/src/app/saga.rs
```diff
@@ -149,14 +150,27 @@ impl super::Nexus {
...
+                ));
+            if let Some((undo_node, undo_error)) = saga_error.undo_failure {
+                error = error.internal_context(format!(
+                    "UNDO ACTION failed (node {:?}, error {:#}) after",
+                    undo_node, undo_error
+                ));
+
+                error!(log, "saga stuck";
+                    "saga_id" => runnable_saga.id.to_string(),
+                    "error" => format!("{:#}", e),
```
bnaecker: Nit, but I believe you can format the error directly, without converting to a string. Something like `"error" => #%e`, I think. (From [here](https://docs.rs/slog/latest/slog/macro.log.html#fmtdisplay-and-fmtdebug-values).)

nexus/src/app/sagas/test_saga.rs
```diff
@@ -0,0 +1,107 @@
...
+        .expect_err("expected saga to finish stuck");
+
+    match result {
+        Error::InternalError { internal_message } => {
+            assert_eq!(
+                internal_message,
+                "UNDO ACTION failed (node \"n1\", error undo action failed \
+                permanently: {\n  \"message\": \"undo action attempt 1: \
+                injected error\"\n}) after: saga ACTION error at node \
+                \"n2\": error injected"
```
bnaecker, line 102: Unrelated to this PR, but I see that the injected error message is swapped in these two: `"injected error"` vs `"error injected"`.

nexus/types/src/internal_api/views.rs
```diff
@@ -42,7 +44,16 @@ impl From<steno::SagaView> for Saga {
 pub enum SagaState {
     Running,
     Succeeded,
-    Failed { error_node_name: steno::NodeName, error_info: SagaErrorInfo },
+    Failed {
+        error_node_name: steno::NodeName,
+        error_info: SagaErrorInfo,
+    },
+    Stuck {
```
bnaecker, line 51: I'm not too clear on the distinction between these variants. I gather failed implies a forward action failed, and stuck means an undo failed. Is the intent to convey that a `Failed` saga could be retried, but a `Stuck` saga can't? I guess I'm wondering the degree to which the variant names convey what happened (action failed, undo action failed, etc) as opposed to what can be done for a saga now, given that it's in either of these states.

Does that make sense?

bnaecker (commented):
I think this is a good path forward! It'll definitely be nice to have more insight into when undo actions fail, rather than a crash-loop :)

nexus/types/src/internal_api/views.rs
davepacheco: First, just to be clear about scope, this is a debug-only API exposed only on the internal network for operators to inspect saga state.

I've been a little on the fence about how to describe sagas whose undo actions fail.  Normally, the invariant provided by sagas is that either you wind up at the end state (all the actions having completed successfully) or you wind up back at the initial state (all the undo actions having been applied).  In this case, we've gotten stuck somewhere in the middle.  Some of the actions have been applied and not undone.  It's pretty different from a "failed" saga (that was successfully unwound) because the system has come to rest broken.  I wanted to elevate that here to its own state.  In the limit, I imagine we'll have queries that report the Stuck sagas and we'll want to open support cases or something.

At first I took this same approach with steno#138, having "Stuck" be a new terminal state alongside "Done".  But I found that it complicated a lot of stuff without buying much.  So Steno _does_ expose these sagas as "'done', but with a slightly different error variant" rather than "come to rest 'stuck' rather than 'done'".  So there's a bit of a difference between Steno and Omicron's debug API here.  I decided to go with it and see how we actually wind up using this.

nexus/src/app/sagas/test_saga.rs
davepacheco: Will fix.

nexus/src/app/saga.rs
davepacheco: Cool.  Will try that.

PR #3214, opened 2023-05-24
[wicketd] Respect installinator's report of which disks were written

jgallagher:
We use this in two (new) ways:

1. We write the corresponding host boot flash slots (previously we only and always wrote slot 0)
2. Set the host to boot off of "the first disk successfully written" (previously we didn't set this at all, and left the host booting off of slot 0 as a side effect of trampoline delivery)

A big chunk of the diff is shifting engine steps to deliver an update to a component via the SP into a new nested engine, making it easier to loop over the disks returned by installinator (which are only available to us inside a running step).

### DIFF
--- a/wicketd/src/update_tracker.rs
+++ b/wicketd/src/update_tracker.rs
@@ -56,8 +56,6 @@ use tokio::sync::oneshot;
 use tokio::sync::watch;
 use tokio::sync::Mutex;
 use tokio::task::JoinHandle;
-use update_engine::events::StepEventKind;
-use update_engine::events::StepOutcome;
 use update_engine::StepSpec;
 use uuid::Uuid;
 use wicket_common::update_events::ComponentRegistrar;
--- a/wicketd/src/update_tracker.rs
+++ b/wicketd/src/update_tracker.rs
@@ -1102,27 +1102,14 @@ impl UpdateContext {
             if write_output.is_none() {
                 for event in &report.step_events {
                     // We only care about the outcome of completion events.
-                    let outcome = match &event.kind {
-                        StepEventKind::StepCompleted { outcome, .. }
-                        | StepEventKind::ExecutionCompleted {
-                            last_outcome: outcome,
-                            ..
-                        } => outcome,
-                        StepEventKind::NoStepsDefined
-                        | StepEventKind::ExecutionStarted { .. }
-                        | StepEventKind::ProgressReset { .. }
-                        | StepEventKind::AttemptRetry { .. }
-                        | StepEventKind::ExecutionFailed { .. }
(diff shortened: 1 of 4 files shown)


wicketd/src/update_tracker.rs
```diff
@@ -1134,14 +1089,61 @@ impl UpdateContext {
...
+                    };
+
+                    match metadata {
+                        InstallinatorCompletionMetadata::Write { output } => {
+                            write_output = Some(output.clone());
+                        }
+                        InstallinatorCompletionMetadata::HardwareScan { .. }
+                        | InstallinatorCompletionMetadata::ControlPlaneZones { .. }
+                        | InstallinatorCompletionMetadata::Download { .. }
+                        | InstallinatorCompletionMetadata::Unknown => (),
```
sunshowers, line 1135: Do we need to express all the use cases here?

wicketd/src/update_tracker.rs
```diff
@@ -1134,14 +1089,61 @@ impl UpdateContext {
...
+                            ..
+                        } => outcome,
+                        StepEventKind::NoStepsDefined
+                        | StepEventKind::ExecutionStarted { .. }
+                        | StepEventKind::ProgressReset { .. }
+                        | StepEventKind::AttemptRetry { .. }
+                        | StepEventKind::ExecutionFailed { .. }
+                        | StepEventKind::Nested { .. }
+                        | StepEventKind::Unknown => continue,
+                    };
```
sunshowers, line 1118: Wondering how you would feel about making this a method on `StepEventKind`.

wicketd/src/update_tracker.rs
```diff
@@ -1134,14 +1089,61 @@ impl UpdateContext {
...
+
+                    // We only care about successful (including "success with
+                    // warning") outcomes.
+                    let metadata = match outcome {
+                        StepOutcome::Success { metadata }
+                        | StepOutcome::Warning { metadata, .. } => metadata,
+                        StepOutcome::Skipped { .. } => continue,
+                    };
```
sunshowers, line 1126: And this a method on `StepOutcome` (`completion_metadata`). That will avoid a lot of this boilerplate.

sunshowers (approved):
Looks good, just a couple minor comments that I'm fine resolving either way.

wicketd/src/update_tracker.rs
jgallagher: Yeah, seems perfectly reasonable. Moved in 7c6dceef

wicketd/src/update_tracker.rs
jgallagher: Yeah, love that. Done in 7c6dceef

PR #3217, opened 2023-05-24
Serve unauthed console page at `/login/{silo_name}/local`

david-crespo:
A version of #3203 where I keep it simple instead.

- Serve console at `/login/{silo_name}/local` without auth check
- Do not get rid of spoof login yet. People are using it on dogfood and we don't have a great way to replace that
- Add `/v1` prefix to `POST /login/{silo_name}/local` and `POST /logout`
  - A few more endpoints are missing it but I'll get those later
- Change `POST /login/{silo_name}/local` success response from a 303 redirect to a 204
  - This is definitely debatable, but because it's a JSON API endpoint, it's hard for me to see why anyone would be in a situation where they want to follow that redirect. The problem is that because it's made in the background, the browser doesn't navigate to the redirect URL, the background fetch just follows it and downloads the HTML. The TS client errors out because it doesn't know what to do with HTML.
  - Alternatives:
    - Use `redirect: manual` in `fetch()` call to prevent it following the redirect. This does seem to work, but based on the [spec](https://fetch.spec.whatwg.org/#concept-request-redirect-mode) it really seems like you're not supposed to do this. The resulting response is called an [opaque-filtered redirect response](https://fetch.spec.whatwg.org/#concept-filtered-response-opaque-redirect) and is not supposed to be good for anything but following the same redirect _later_
    - Fix `oxide.ts` to not error out on HTML. That would be [here](https://github.com/oxidecomputer/oxide.ts/blob/8a376956b03422ade24471207ceb0ab8c8961c4e/client/http-client.ts#L103-L104). It's feasible but it would complicate things and it doesn't seem worthwhile.


nexus/tests/integration_tests/console_api.rs
```diff
@@ -177,39 +198,20 @@ async fn test_console_pages(cptestctx: &ControlPlaneTestContext) {
...
-            .expect_status(Some(StatusCode::OK))
-            .expect_response_header(
-                http::header::CONTENT_TYPE,
-                "text/html; charset=UTF-8",
-            )
-            .execute()
-            .await
-            .expect("failed to get login form");
+    let unauthed_console_paths =
+        &["/spoof_login", "/login/irrelevant-silo/local"];
```
david-crespo, line 210: The changes in this file are mostly cleanup. This line is the only substantive change: check the new path in addition to `/spoof_login`.

nexus/src/external_api/console_api.rs
```diff
@@ -387,17 +401,34 @@ pub struct LoginPathParam {
...
+    rqctx: RequestContext<Arc<ServerContext>>,
+    _path_params: Path<LoginPathParam>,
+) -> Result<Response<Body>, HttpError> {
+    // TODO: figure out why instrumenting doesn't work
+    // let apictx = rqctx.context();
+    // let handler = async { serve_console_index(rqctx.context()).await };
+    // apictx.external_latencies.instrument_dropshot_handler(&rqctx, handler).await
+    serve_console_index(rqctx.context()).await
+}
+
```
david-crespo, line 420: I'm not sure this endpoint will need to exist after we have silo in the domain. I was thinking a single `GET /login` endpoint could show either username/password login or a link to the IdP as appropriate depending on the silo. Because this is all done through the console, the generic console page be served as usual, and then the console would have to make an API call to determine which thing to show.

zephraph: It'd be nice if we could avoid having to make an extra API call by including the silo in the header or something. Not for this PR, but in whatever solution we ultimately end up with.

nexus/src/external_api/console_api.rs
```diff
@@ -387,17 +401,34 @@ pub struct LoginPathParam {
...
 /// Authenticate a user via username and password
 #[endpoint {
    method = POST,
-   path = "/login/{silo_name}/local",
+   path = "/v1/login/{silo_name}/local",
```
zephraph, line 424: I intentionally left the `/v1/` off the console endpoints which the thought being that they'd need to be relatively stable. I can see wanting to gate the POST endpoint though given that it'll mostly be treated like any other API endpoint.

nexus/src/external_api/console_api.rs
david-crespo: Yeah, it really would. I think we might soon get to the point where doing that is less ridiculous than the stuff we're doing to get around it. On the other hand, the total decoupling of the frontend from the API has been very beneficial to my sanity so far.

nexus/src/external_api/console_api.rs
david-crespo: Yeah, I went back and forth. It is a JSON API endpoint, so I figured it should probably follow the convention. I'm thinking we should get to a point where only the GETs don't have v1 and they're all unpublished. It feels weird to have prefixed and non-prefixed endpoints right next to each other, but that's where I landed as least unreasonable.

david-crespo: I'm not sure. We could have `POST /v1/login` just always be username/password, and that would make this endpoint unnecessary. Or maybe we could have `POST /v1/login/local` and we'd change `POST /login/{silo_name}/saml/{provider_name}` to `POST /v1/login/saml/{provider_name}`?

zephraph: The latter part feels better than me. Given that username/password log ins should be relatively rare I don't know that I'd want it to be the root endpoint (which makes it feel like the default).

zephraph (approved):
Nice! Let's jam.

nexus/src/external_api/console_api.rs
david-crespo: Yeah, I think that's right.

david-crespo:
😅

PR #3218, opened 2023-05-24
[wicket] add scrolling to popups

sunshowers:
Now that we're displaying error messages in popups, we can have a pretty
large amount of text here. Add the ability to scroll down in a popup,
building on #3194 which wraps all popup text.

This is a bit tricky because we want to cap the offset to prevent
scrolling down once we've reached the bottom of the text, but a lot of
values are computed at layout time.  To handle this, introduce the
notion of an "actual scroll offset", which is the scroll offset as
computed after text is laid out.  Also compute the layout parameters in
the constructor for `Popup` so we can return the actual scroll offset.

Display scroll status in the button bar (I didn't figure out a better
place to show it; we could potentially show an actual scroll bar to the
right but that's a lot more work than this.)

Recording:

[Screencast from 05-24-2023 04:30:21 PM.webm](https://github.com/oxidecomputer/omicron/assets/180618/b6875528-95cc-4c8a-b340-49aa5ee2c899)

This also works if there are no buttons to begin with:

[Screencast from 05-24-2023 04:35:58 PM.webm](https://github.com/oxidecomputer/omicron/assets/180618/2fad3259-cd4b-4c0a-9f5f-c9b97ce45c40)


andrewjstone (approved):
Looks great!

sunshowers:
I'm going to bundle up some changes to make the command bar at the bottom have its style match up with popup buttons.

sunshowers:
New style, slightly different (note help bar + no longer uppercase to match help bar)

image

image

andrewjstone:
This is great @sunshowers. Thanks for paying attention to the small details as always!

PR #3220, opened 2023-05-25
[wicketd] Skip updating the SP if it already has the version we're about to update to

jgallagher:
Running a mupdate where the version specified in the TUF repo matches the version reported by the SP now skips updating the SP:

sp-skip

The user can override this via the functionality added in #3150, in which case we do the update and mark the step as a "Warning":

sp-force


andrewjstone (approved):
Looks great!

Issue #3207, opened 2023-05-24
Instance delete left behind CRDB records after propolis zone is already removed


askfongjojo:
This happened when I used a script to invoke `oxide instance stop` for 10 instances successively in a loop, and then delete them with cli in another loop. The requests came back with success response (ignore the two destroyed instances - they will be covered in another ticket):
```
pisces-2:tests angela$ for i in {1..12}; do oxide instance delete --project try --instance sysbench-mysql-$i; done
success
()
success
()
success
()
success
()
success
()
success
()
success
()
success
()
success
()
error
Error Response: status: 400 Bad Request; headers: {"content-type": "application/json", "x-request-id": "730032c9-f928-4a5b-9244-bc36c14096ef", "content-length": "158", "date": "Tue, 23 May 2023 22:41:07 GMT"}; value: Error { error_code: Some("InvalidRequest"), message: "instance cannot be deleted in state \"destroyed\"", request_id: "730032c9-f928-4a5b-9244-bc36c14096ef" }
error
Error Response: status: 404 Not Found; headers: {"content-type": "application/json", "x-request-id": "dd8a7e48-c584-4fa8-afc8-f4fa7c0c9c7f", "content-length": "160", "date": "Tue, 23 May 2023 22:41:07 GMT"}; value: Error { error_code: Some("ObjectNotFound"), message: "not found: instance with name \"sysbench-mysql-11\"", request_id: "dd8a7e48-c584-4fa8-afc8-f4fa7c0c9c7f" }
success
()
```

However, these instances are now "shell" records in the instance table - the propolis zones are already gone from the sled:
```
root@[fd00:1122:3344:106::6]:32221/omicron> select name, state, id, active_sled_id, active_propolis_id from instance where name like 'sysbench%' and state = 'running';
        name        |   state   |                  id                  |            active_sled_id            |          active_propolis_id
--------------------+-----------+--------------------------------------+--------------------------------------+---------------------------------------
  sysbench-mysql-5  | running   | 0dcd991a-36de-4a15-a999-a6fda4efcc92 | aa7c82d9-6e59-406e-b1e3-3648890b4bec | 21078e19-c1e0-4aeb-98aa-5c337352353d
  sysbench-mysql-2  | running   | 2ab5e713-5d25-486a-9083-d5b7cd1e870d | 4f4d89de-b2ba-4a96-b585-1ef2b7a31942 | 06995cdb-54cc-4923-aea8-0fc6f617277c
[...]

PR #3232, opened 2023-05-26
use bb8 error sink to log more errors

davepacheco:
I found while debugging #3207 that bb8 provides a way to receive database connection errors that aren't associated with a checked-out connection.  We should use this to report those errors.


davepacheco:
I tested this by running `omicron-dev db-run`, then starting Nexus, then killing the `db-run` command.  Nexus emitted eight lines of this:

```
03:00:37.734Z ERRO e6bff1ff-24fb-49dc-a54e-c6a350cd4d6c (ServerContext): database connection error
    database_url = postgresql://root@127.0.0.1:32221/omicron?sslmode=disable
    error_message = Failed to checkout a connection: server is shutting down
```

The error message is a little misleading because we weren't trying to checkout a connection.  I will take a look at fixing this up in async-bb8-diesel.  But that's unrelated to this change.

sunshowers (approved):
Awesome, thanks!

PR #3234, opened 2023-05-26
[sled-agent] Delay nexus hw notifications

andrewjstone:
Before underlay access has been allocated to the `StorageWorker`, there is no chance of successfully talking to nexus. Delay any notifications until after underlay access is available. This occurs when sled-agent starts, after disks and zpools may have already been allocated/recorded locally.

By waiting we reduce unnecessary backoff timer increases that are not due to any actual faults, and delay normal sled startup. This is especially important during rack setup since those backoff timers wait indefinitely until RSS is run.

This is a partial fix to dependency order, but just a small one. We  should take a more systematic approach ala https://github.com/ oxidecomputer/omicron/issues/1917 in the future.

### DIFF
--- a/sled-agent/src/storage_manager.rs
+++ b/sled-agent/src/storage_manager.rs
@@ -333,7 +333,20 @@ impl StorageWorker {
 
     // Adds a "notification to nexus" to `nexus_notifications`,
     // informing it about the addition of `pool_id` to this sled.
-    fn add_zpool_notify(&mut self, pool: &Pool, size: ByteCount) {
+    async fn add_zpool_notify(&mut self, pool: &Pool, size: ByteCount) {
+        // The underlay network is setup once at sled-agent startup. Before
+        // there is an underlay we want to avoid sending notifications to nexus for
+        // two reasons:
+        //  1. They can't possibly succeed
+        //  2. They increase the backoff time exponentially, so that once
+        //   sled-agent does start  it may take mucyh longer to notify nexus
+        //   than it would if we avoid this. This goes especially so for rack
+        //   setup, when bootstrap agent is waiting an aribtrary time for RSS
+        //   initialization.
+        if self.underlay.lock().await.is_none() {
+            return;
+        }
+
         let pool_id = pool.name.id();
         let DiskIdentity { vendor, serial, model } = pool.parent.clone();
         let underlay = self.underlay.clone();
@@ -540,7 +553,8 @@ impl StorageWorker {
         self.physical_disk_notify(NotifyDiskRequest::Add {
             identity: disk.identity(),
             variant: disk.variant(),
-        });
+        })
+        .await;
         self.upsert_zpool(&resources, disk.identity(), disk.zpool_name())
             .await?;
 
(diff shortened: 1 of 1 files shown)


sled-agent/src/storage_manager.rs
```diff
@@ -333,7 +333,20 @@ impl StorageWorker {
...
+        // there is an underlay we want to avoid sending notifications to nexus for
+        // two reasons:
+        //  1. They can't possibly succeed
+        //  2. They increase the backoff time exponentially, so that once
+        //   sled-agent does start it may take much longer to notify nexus
+        //   than it would if we avoid this. This goes especially so for rack
+        //   setup, when bootstrap agent is waiting an aribtrary time for RSS
+        //   initialization.
+        if self.underlay.lock().await.is_none() {
+            return;
```
jgallagher, line 347: I think there's a simpler way to accomplish this, but it might be too hacky and/or interact with `backoff` in undesirable ways.  A few lines below this, the closure we give to the nexus notifications queue returns a future that begins with this:

```rust
async move {
    let underlay_guard = underlay.lock().await;
    let Some(underlay) = underlay_guard.as_ref() else {
        return Err(backoff::BackoffError::transient(Error::UnderlayNotInitialized.to_string()));
};
```

What if instead of returning a transient error there, it did its own looping to wait for the underlay to exist? This doesn't need to be backoff'd because it's just a local mutex check, so sleeping for order(seconds) is probably fine. Something like (untested, maybe typos):

```rust
async move {
    let underlay_guard = loop {
        let guard = underlay.lock().await;
        if guard.is_some() {
            break guard; // underlay exists; proceed with notifying nexus
        }
        mem::drop(guard); // unlock before sleeping

        info!(..., "underlay doesn't exist yet...");
        tokio::time::sleep(Duration::from_secs(...)).await;
    };

    // We only break out of the loop when `underlay_guard.is_some()`, so
    // we can safely unwrap here.
    let underlay = underlay_guard.as_ref().unwrap();
```

I'm very unsure this is actually better (e.g., this might make moving to bulk notifications as you noted in a comment below harder), but I do think it would have the same effect with fewer surrounding changes (e.g., you would no longer need `notify_nexus_about_existing_resources()` at all, because once the underlay exists all the pending notifications would be released from this loop to actually start talking to nexus).

david-crespo: If the fix is tolerable as-is, I suggest merging and tweaking later because this CI failure is blocking every Omicron PR.

andrewjstone: I like @jgallagher's suggestion better, but rebuilding/testing will take a little bit. I"ll merge this as is and see if it fixes the problem, then go back and add John's suggested change later today.

andrewjstone: Uh, I implemented @jgallagher's version and tested it on paris. It does indeed prevent hardware notifications before RSS, but once I run RSS the flurry of notifications and frequency is fierce. I think it does not play nicely with backoff since the check itself is embedded in the future. I'm going to keep my version as merged and stop playing with this in the interest of time.

andrewjstone: For the record, John's version is [here](https://github.com/oxidecomputer/omicron/compare/sled-hw-notification-simplification?expand=1)

PR #3235, opened 2023-05-26
Support for nexus sub tags

benjaminleonard:
Using the format `{tag}/{subtag}` we can nest operations. This solves the issue of tags that contain so many operations that it gets [difficult to navigate](https://github.com/oxidecomputer/docs/pull/96#issuecomment-1563226809).

I've updated the URLs, these don't yet work but I'll add an issue on the docs site to get these matched with the tags.
The docs site will also need updating to render the nested operations, but I'll address that after this is merged.

I'd initially went with `networking/{subtag}` — but figured `vpcs/{subtag}` was closer to the existing API.

This will break the traits: `ClientSystemExt` and `ClientVpcsExt`. I've updated usage of those in this repo and done a check across the org but there's a good chance I might have missed something.

I've tried to work on the descriptions, will need someone to vet those to make sure they are accurate.


end-to-end-tests/src/helpers/ctx.rs
```diff
@@ -5,7 +5,7 @@ use omicron_sled_agent::rack_setup::config::SetupServiceConfig;
 use omicron_test_utils::dev::poll::{wait_for_condition, CondCheckError};
 use oxide_client::types::{Name, ProjectCreate};
 use oxide_client::CustomDnsResolver;
-use oxide_client::{Client, ClientProjectsExt, ClientVpcsExt};
+use oxide_client::{Client, ClientProjectsExt, ClientVpcsSubnetsExt};
```
david-crespo, line 8: this isn’t supposed to be here, is it?

david-crespo: oh I see, this actually does come from the tags

benjaminleonard: yes sir

nexus/tests/output/nexus_tags.txt
```diff
@@ -104,7 +104,31 @@ snapshot_delete                          DELETE   /v1/snapshots/{snapshot}
 snapshot_list                            GET      /v1/snapshots
 snapshot_view                            GET      /v1/snapshots/{snapshot}
 
-API operations found with tag "system"
+API operations found with tag "system/hardware"
+OPERATION ID                             METHOD   URL PATH
+networking_switch_port_apply_settings    POST     /v1/system/hardware/switch-port/{port}/settings
+networking_switch_port_clear_settings    DELETE   /v1/system/hardware/switch-port/{port}/settings
+networking_switch_port_list              GET      /v1/system/hardware/switch-port
```
ahl, line 111: I know that these are under the `/v1/system/hardware...` path, but looking at the peers, I think this might make more sense under the "system/networking" tag (and possibly changing the path)

@internet-diglett and @rcgoodfellow what do you think?

rcgoodfellow: I could see this going either way. When we're applying/clearing switch port settings, we're doing so on a physical port which is a part of the system's hardware. These endpoints take selectors that specify the switch, but maybe the endpoint would be better as something like `/v1/system/hardware/switches/{switch}/switch-port/{port}/settings` now that #2996 has made switches UUID'd objects?

zephraph: @rcgoodfellow I'm working on wrapping that change up right now. So @ahl, no, the switch-ports _do_ need to be in the hardware section. I'm thinking about removing the `networking` prefix though.

nexus/tests/output/nexus_tags.txt
```diff
@@ -166,32 +182,49 @@ system_update_view                       GET      /v1/system/update/updates/{ver
...
-vpc_router_view                          GET      /v1/vpc-routers/{router}
+
+API operations found with tag "vpcs/subnets"
+OPERATION ID                             METHOD   URL PATH
 vpc_subnet_create                        POST     /v1/vpc-subnets
 vpc_subnet_delete                        DELETE   /v1/vpc-subnets/{subnet}
 vpc_subnet_list                          GET      /v1/vpc-subnets
 vpc_subnet_list_network_interfaces       GET      /v1/vpc-subnets/{subnet}/network-interfaces
 vpc_subnet_update                        PUT      /v1/vpc-subnets/{subnet}
 vpc_subnet_view                          GET      /v1/vpc-subnets/{subnet}
```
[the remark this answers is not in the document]
zephraph: Well, for the docs at least it's probably going to make more sense to do subgroupings. With that said, perhaps tags + operation partial matches could be used to derive that logic. What do you think @benjaminleonard?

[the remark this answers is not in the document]
benjaminleonard: Partial matches are probably a reasonable way of adding tertiary groupings — happy to move these into a single `vpcs` tag.

benjaminleonard:
@ahl Should these be under `/system/roles/`:
https://github.com/oxidecomputer/omicron/blob/8bd2fd44a9812f37cfb2664bf96ed58b2f870866/nexus/tests/output/nexus_tags.txt#L74-L77

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -1160,8 +1160,8 @@ async fn ip_pool_view(
 /// Delete an IP Pool
 #[endpoint {
     method = DELETE,
-    path = "/v1/system/ip-pools/{pool}",
-    tags = ["system"],
+    path = "/v1/system/networking/ip-pools/{pool}",
```
zephraph, line 1163:
```suggestion
    path = "/v1/system/ip-pools/{pool}",
```

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -1182,8 +1182,8 @@ async fn ip_pool_delete(
 /// Update an IP Pool
 #[endpoint {
     method = PUT,
-    path = "/v1/system/ip-pools/{pool}",
-    tags = ["system"],
+    path = "/v1/system/networking/ip-pools/{pool}",
```
zephraph, line 1185:
```suggestion
    path = "/v1/system/ip-pools/{pool}",
```

nexus/src/external_api/tag-config.json
```diff
@@ -5,145 +5,139 @@
...
     },
-    "metrics": {
-      "description": "Metrics provide insight into the operation of the Oxide deployment. These include telemetry on hardware and software components that can be used to understand the current state as well as to diagnose issues.",
+    "vpcs/firewall-rules": {
+      "description": "Firewall rules are used to control incoming and outgoing network traffic.",
       "external_docs": {
-        "url": "http://oxide.computer/docs/#xxx"
+        "url": "http://docs.oxide.computer/api/firewall-rules"
       }
     },
```
zephraph, line 46: If we're consolidating vpcs we can drop this tag

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -2605,8 +2605,8 @@ async fn networking_switch_port_settings_list(
 /// Get information about a switch port
 #[endpoint {
     method = GET,
-    path = "/v1/system/networking/switch-port-settings/{port}",
-    tags = ["system"],
+    path = "/v1/system/networking/switch-port-settings/{port}/info",
```
zephraph, line 2608: @benjaminleonard was this intentional?

nexus/src/external_api/tag-config.json
```diff
@@ -5,145 +5,133 @@
...
+        "url": "http://docs.oxide.computer/api/disks"
       }
     },
     "hidden": {
       "description": "TODO operations that will not ship to customers",
       "external_docs": {
-        "url": "http://oxide.computer/docs/#xxx"
+        "url": "http://docs.oxide.computer/api/"
       }
     },
```
ahl, line 16: we have to deal with these things ASAP...

nexus/src/external_api/tag-config.json
```diff
@@ -5,145 +5,169 @@
...
+        "url": "http://docs.oxide.computer/api/system/metrics"
+      }
+    },
+    "system/networking": {
+      "description": "This provides rack-level network configuration.",
+      "external_docs": {
+        "url": "http://docs.oxide.computer/api/system/networking"
+      }
+    },
+    "system/silos": {
```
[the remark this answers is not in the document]
ahl: back to this: I think squish them all into "system/silos"

ahl (approved):
fine I guess

PR #3236, opened 2023-05-26
Use meaningful (and shorter) image names from CI

citrus-it:
The host OS image names that come out of CI at present include the build user (always `build`) and the AWS IP address that runs the job, they are also a bit long and make the output of `pilot host ls` overflow 80 columns:

```
CUBBY IP                        SERIAL      IMAGE
8     fe80::aa40:25ff:fe04:3d5  BRM44220011 build@ip-10-150-1-170.us-west-2.compute.internal: 2023-05-25 04:43:28
9     fe80::aa40:25ff:fe04:357  BRM44220005 build@ip-10-150-1-170.us-west-2.compute.internal: 2023-05-25 04:43:28
...
```

After this change, they should look like

```
CUBBY IP                        SERIAL      IMAGE
8     fe80::aa40:25ff:fe04:3d5  BRM44220011 omicron-ci: 2023-05-25 04:43:28
9     fe80::aa40:25ff:fe04:357  BRM44220005 omicron-ci: 2023-05-25 04:43:28
...
```


tools/build-host-image.sh
```diff
@@ -117,10 +119,13 @@ function main
 
     pfexec zfs create -p rpool/images/$USER
 
+    IMAGE_NAME="$IMAGE_PREFIX: $(date +'%Y-%m-%d %H:%M:%S')"
```
jgallagher, line 122: Would it be worth including the (short) git commit too?

jgallagher (approved):

tools/build-host-image.sh
citrus-it: Yes, probably - out of omicron, helios or stlouis, which do you think would be most useful here?
I don't think helios tells you much (and it's presumably whatever is pinned in the omicron revision)
I'd prefer stlouis but it's a bit tricky to extract.

citrus-it: I'm going to see how hard it is to include both omicron and host OS hashes in the image name. I think that will be most useful when looking at a list of sleds.

tools/build-host-image.sh
```diff
@@ -117,10 +119,41 @@ function main
...
+                    )
+                    if [[ -n "$BUILDHASH" && -f "$PKGDIR/$BUILDHASH" ]]; then
+                        OS_HASH="$(awk -F- '{print $NF}' "$PKGDIR/$BUILDHASH")"
+                        break
+                    fi
+                fi
+            done
+        fi
+        rm -rf "$tmpd"
+    fi
```
jclulow, line 146: I feel like this could potentially be misleading if new packages arrive between when we do this generation, and when we then do the work below to actually install things; i.e., the hash might be wrong depending on things that are occurring asynchronously outside this process.

Could we instead perhaps add support to `helios-build` for macro expansion in the `-N` argument?  You can borrow the expansion stuff from https://github.com/illumos/image-builder/blob/main/src/expand.rs as I think it would likely work well there also.  (I wrote that while at Oxide)

Then we could say something like:

```
-N "ci-${GITHUB_SHA:0:7}-\${os_short_commit}"
```

and have `${os_short_commit}` expanded to basically what you're getting from `${OS_HASH:1:8}` there by `helios-build` itself, which can inspect the ramdisk root to find whatever it needs.

jclulow (commented):
I tried to leave this comment earlier and somehow left it on ... the commit?  But not the PR?!  I don't even with this website.  Trying again!

tools/build-host-image.sh
citrus-it: That's a good idea to solve the chicken-and-egg problem, thanks, I will look into that.

citrus-it:
I'm going to merge this as-is to get improved names for this week's work, then come back and close the race window once the tools are updated following jclulow's suggestion.

Tested by downloading each of the generated archives from this PR's CI run.

```
atrium% lookimage os/image/zfs.img| head -1
image name = ci bb2da82/177d123 2023-05-30 11:46
atrium% lookimage tramp/image/zfs.img| head -1
image name = recovery bb2da82/177d123 2023-05-30 11:46
```

PR #3237, opened 2023-05-26
minor cleanup of the networking API

ahl:


david-crespo (approved):
#nice

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -2565,7 +2565,7 @@ async fn networking_switch_port_settings_delete(
     apictx.external_latencies.instrument_dropshot_handler(&rqctx, handler).await
 }
 
-/// List port settings.
+/// List port settings
```
rcgoodfellow, line 2568: port -> switch port here too?

rcgoodfellow (commented):
Thanks @ahl! Just one small nit.

nexus/src/external_api/http_entrypoints.rs
ahl: yes; thanks

rcgoodfellow (approved):

PR #3239, opened 2023-05-26
[mgs/wicketd/wicket] Plumb RoT and SP caboose information up through wicket

jgallagher:
This addresses some lingering TODOs from the quick fixes in #3185: we now relay the modern RoT state information, and wicket now displays versions for both RoT slots and both SP slots, if available; e.g.,

pane

This is a precursor to supporting "skip updating the RoT if it already has the version we want to update to".

### DIFF
--- a/gateway/src/http_entrypoints.rs
+++ b/gateway/src/http_entrypoints.rs
@@ -27,7 +27,6 @@ use dropshot::UntypedBody;
 use dropshot::WebsocketEndpointResult;
 use dropshot::WebsocketUpgrade;
 use futures::TryFutureExt;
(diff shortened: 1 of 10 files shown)


gateway/src/http_entrypoints.rs
```diff
@@ -58,7 +57,6 @@ pub struct SpState {
     pub revision: u32,
     pub hubris_archive_id: String,
     pub base_mac_address: [u8; 6],
-    pub version: ImageVersion,
```
andrewjstone, line 61: 👯

andrewjstone (approved):
Ship it!

PR #3240, opened 2023-05-26
[wicketd] Skip updating the RoT if it already has the target version

jgallagher:
This skips updating the RoT by default (can be overridden by the user from wicket) if the _currently active_ RoT slot's version matches the artifact version in the TUF repository.

Builds on #3239.


sunshowers (approved):
Looks good!

PR #3254, opened 2023-05-30
Fix unique constraints for images

### LINKED ISSUE #2895: Image promote does not check for duplicate names
I was able to promote an image with the same name as another image that exists at the silo-level. It should probably complain about name duplication.

zephraph:
Fixes #2895 

As @davepacheco pointed out in #2895, CRDB doesn't treat `null` as unique which ultimately means duplicate entries can be present. To prevent that I've broken down the constraint down into two indices to ensure the unique constraint is enforced.


david-crespo:
Great news if that's all it takes.

nexus/tests/integration_tests/images.rs
```diff
@@ -588,11 +588,47 @@ async fn test_image_access(cptestctx: &ControlPlaneTestContext) {
...
+
     let silo_image_url = format!("/v1/images/{}", image_id);
     let silo_image = NexusRequest::object_get(client, &silo_image_url)
         .authn_as(AuthnMode::PrivilegedUser)
         .execute_and_parse_unwrap::<views::Image>()
         .await;
 
     assert_eq!(silo_image.identity.id, image_id);
+
+    // Create another project image
```
david-crespo, line 608:
```suggestion
    // Create another project image with the same name
```

david-crespo (approved):
Looks great assuming you confirmed the test fails without the index change.

zephraph:
Yeah, that was my conclusion. Unfortunately there's a _slight_ hiccup in that `Conflict` is mostly used in cases where we have the name in question. In this particular case we don't actually have the name on hand. We're promoting an image by ID and the uniqueness constraint is failing on name. That may mean we need to do a secondary lookup for the name of the resource before doing the insert?

zephraph:
As an update on this

I paid with @davepacheco and he helped me with some query execution analysis. The query to list a project's images combined with silo images (which is used for situations like instance create) ends up doing a full table scan. We were able to avoid the full table scan by breaking it down into two select queries with a UNION clause combining the results. Diesel exposes capacities for performing a union but I've been struggling to figure out how to slot that in with our current paginated approach.

zephraph:
I've decided to sidestep the whole issue with the query and just remove the `include_silo_images` flag when listing project images. It reduces the option space for the UI, but we can always look to add it again later if it's a customer requested feature.

david-crespo:
So how do you get a list of silo images?

zephraph:
The images list endpoint already had to modes. `/v1/images` gives you the silo images. `/v1/images?project=:project` gives you a project's images. The flag that I removed was `/v1/images?project=:project&include_silo_images` which allowed for listing both a single projects images alongside all the silo images. The use case there was so that we could have a single image selection interface.

david-crespo:
Oh right. I forgot about selecting with no param. So the console will just make two requests. No problem.

PR #3257, opened 2023-05-30
[benchmarks] update criterion version

sunshowers:
Criterion.rs 0.5.0+ supports nextest, which I've been doing some tests
with.


sunshowers:
Adam said "ship it" in DMs, so shipping it :)

PR #3258, opened 2023-05-30
[sled-agent] Add /baseboard bootstrap server endpoint

jgallagher:
This is a precursor to wicketd being able to initiate RSS: we need to be able to pair baseboards (i.e., the identifying triple wicketd gets from all the SPs in the rack) to bootstrap addresses. An easy way to do that is hitting a `/baseboard` endpoint (added by this PR) on every peer mg-ddm finds on the bootstrap network.

Running this on the madrid scrimlet:

```
BRM44220001 # curl http://[fdb0:a840:2504:355::1]/baseboard
{"type":"gimlet","identifier":"BRM44220001","model":"913-0000019","revision":6}
```

### DIFF
--- a/openapi/bootstrap-agent.json
+++ b/openapi/bootstrap-agent.json
@@ -10,6 +10,30 @@
     "version": "0.0.1"
   },
   "paths": {
+    "/baseboard": {
+      "get": {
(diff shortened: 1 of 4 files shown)


andrewjstone (approved):
Looks good.

PR #3266, opened 2023-05-31
Extract omicron-certificates from nexus-db-model

jgallagher:
This will allow wicketd to share certificate validation for user-provided certs during rack setup.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,6 +3,7 @@ members = [
     "api_identity",
     "bootstore",
     "bootstrap-agent-client",
+    "certificates",
     "common",
     "ddm-admin-client",
     "dpd-client",
@@ -201,6 +202,7 @@ nexus-client = { path = "nexus-client" }
 nexus-db-model = { path = "nexus/db-model" }
 nexus-db-queries = { path = "nexus/db-queries" }
 nexus-defaults = { path = "nexus/defaults" }
+omicron-certificates = { path = "certificates" }
 omicron-passwords = { path = "passwords" }
 nexus-test-interface = { path = "nexus/test-interface" }
 nexus-test-utils-macros = { path = "nexus/test-utils-macros" }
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -3,6 +3,7 @@ members = [
     "api_identity",
     "bootstore",
     "bootstrap-agent-client",
+    "certificates",
```
davepacheco, line 6: Shouldn't this go into default-members too?

certificates/src/lib.rs
```diff
@@ -0,0 +1,122 @@
...
+    pub fn danger_disable_expiration_validation(&mut self) {
+        self.validate_expiration = false;
+    }
+
+    /// Validate that we can parse the cert chain, that the key matches, and
+    /// that the certs in the chain are not expired (unless we have disabled
+    /// expiration validation).
+    //
+    // TODO-completeness: Can we take an optional hostname and validate that
+    // this cert will work for it? (This might be a paramter to this method, or
```
davepacheco, line 78:
```suggestion
    // this cert will work for it? (This might be a parameter to this method, or
```

certificates/src/lib.rs
```diff
@@ -0,0 +1,122 @@
...
+    // could be an additional property on `Self`.)
+    pub fn validate(
+        &self,
+        certs: &[u8],
+        key: &[u8],
```
davepacheco, line 83: I think it'd be worth a note in the doc comment about what format these are supposed to be in.

Cargo.toml
jgallagher: Hah, yes, thanks. I think I stopped searching for "passwords" and started searching for "omicron-passwords", which led me to miss default-members. Fixed in 3e16589

certificates/src/lib.rs
jgallagher: Added in ec5db70

davepacheco (approved):

PR #3267, opened 2023-05-31
[update-engine] add support for progress units

sunshowers:
It's a bit weird to just show numbers without their corresponding units.
Add support for progress units which are mandatory to specify (we can do
that because we haven't shipped the update engine yet.)

Depends on #3243.


andrewjstone (approved):

wicketd/src/update_tracker.rs
```diff
@@ -1522,6 +1525,7 @@ impl UpdateContext {
                                 StepProgress::with_current_and_total(
                                     progress.current as u64,
                                     progress.total as u64,
+                                    ProgressUnits::BYTES,
```
jgallagher, line 1528: This one is not bytes; the units depend on the component being updated and are kinda weird to specify even then. For the host boot flash the units are something like "sectors erased"; for the sidecar SP it's "aux flash slots scanned for a match". I'm tempted to propose making the units optional?

wicketd/src/update_tracker.rs
```diff
@@ -1570,14 +1573,24 @@ impl UpdateContext {
...
     where
         S::ProgressMetadata: Default,
     {
         // How often we poll MGS for the progress of an update once it starts.
         const STATUS_POLL_FREQ: Duration = Duration::from_millis(300);
 
+        let progress_units = match component {
+            UpdateComponent::Rot => {
+                // XXX what's the correct unit for this?
+                ProgressUnits::new("flash slots scanned")
```
sunshowers, line 1588: @jgallagher what's the correct unit for this?

jgallagher: There isn't one - the Rot (currently?) [never](https://github.com/oxidecomputer/hubris/blob/6c0c483c47f5050696032b7166f4b650ca66c993/task/control-plane-agent/src/update/rot.rs#L103-L105) returns a "preparing" status.

wicketd/src/update_tracker.rs
```diff
@@ -1560,14 +1573,24 @@ impl UpdateContext {
...
+            }
+            UpdateComponent::Sp => ProgressUnits::new("flash slots scanned"),
```
jgallagher, line 1590: Ug, with apologies, this isn't quite right, and I'm not sure how to suggest making this right. There are two issues at play:

1. The only SP updates that deliver "preparing" status are sidecar SP updates - neither gimlet nor psc SPs should ever return a preparing status, and if they do at some point in the future there's no guarantee this would be the correct unit.
2. For sidecar, preparation is divided into two stages: first, the SP has to scan the flash slots, then if it doesn't find a match, it has to erase one of the slots. When the SP [sends its (unitless) progress](https://github.com/oxidecomputer/hubris/blob/master/task/control-plane-agent/src/update/sp/auxflash.rs#L88-L103), the total count is "number of slots + number of sectors to erase", where a `current` value less than "number of slots" means it's still scanning and a `current` value >= the number of slots means it's erasing (and if erasing is unnecessary, it will skip from "number of slots" directly to "preparation complete").

Would it be unreasonable to lean into the unitless idea here and make both this and the RoT something like `"preparation steps"`? It's not particularly useful, but I'm not sure there _is_ much use to anyone other than a hubris developer in the specific units here.

sunshowers: Yeah, that makes sense. I'll change this to "preparation steps".

wicketd/src/update_tracker.rs
sunshowers: Made moot by the update to using "preparation steps" unconditionally.

wicketd/src/update_tracker.rs
sunshowers: Done.

PR #3268, opened 2023-05-31
remove "spoof" authentication from most configurations

### LINKED ISSUE #372: Remove "spoof" authn scheme from configuration files
To be able to continue accessing Nexus' endpoints, configuration files must use the `spoof` authn scheme.

This issue tracks removal of that scheme (we don't want to ship with it!) when we have a reasonable replacement.

davepacheco:
There are actually two things called "spoof" in Omicron:

1. When authenticating any external API request, Nexus looks for the "oxide-spoof" flavor of bearer token.  If specified, it trusts that the client is who they say they are for that request.  You don't need to log in.  You don't get a session token or anything.  The request is just authenticated.  This authentication scheme was the first one we supported and was always intended to be temporary until we'd built out better schemes.  I believe we finally have all the pieces needed to remove this.
2. Unrelated to that, there's a `POST /spoof_login` endpoint that requires _no_ authentication that creates a session and returns a session token.  When you use this, you're using this unauthenticated API to get a _real_ session token, and then after that you're using normal session token authentication.

This PR:

- Disables the "spoof" authentication scheme (number 1 above) outside of the test suite.  It will take more work to excise "spoof" from the test suite.
- Removes the spoof_login unauthenticated endpoint that returns a real session token (number 2 above).  Also removes the related routes (e.g., `GET /spoof_login`).  This required changing `GET /login` to produce some other kind of useful redirect.  It now redirects you to an "appropriate" login URL for the Silo it thinks you're coming into, if possible.
- Adds an end-to-end test that ensures that spoof login doesn't work (i.e., that a more typical Nexus configuration does not support spoof login)

Fixes #372.

Still to-do here:

- [x] write some automated tests for the new functions
- [x] verify that all the key workflows work
    - [x] the test suite (mostly unchanged, since this still has spoof authn enabled)
    - [x] `omicron-dev run-all`
    - [x] the simulated stack run by hand
    - [x] end-to-end tests (tested by CI).  This is as close to a real deployment as I can easily test.


end-to-end-tests/src/helpers/ctx.rs
```diff
@@ -179,15 +216,11 @@ pub async fn build_client() -> Result<oxide_client::Client> {
             // codes.  See progenitor#451.
             eprintln!("{}: attempting to log into API", Utc::now());
 
-            let mut builder = reqwest::ClientBuilder::new()
+            let builder = client_params
+                .reqwest_builder()
```
david-crespo, line 220: btw, I got rid of the 303 redirect on local login in #3217, so you might not need the raw reqwest builder anymore. The SAML endpoint still 303s on success, but that one is more plausibly necessary and it's not being hit by the API client.

nexus/src/app/external_endpoints.rs
```diff
@@ -212,16 +256,25 @@ impl ExternalEndpoints {
...
-    /// the Silo associated with this endpoint
+#[derive(Debug, PartialEq, Eq, Serialize)]
+pub struct ExternalEndpoint {
+    /// the id of the Silo associated with this endpoint
+    // This is redundant with `db_silo`, but it's convenient to put it here and
+    // it shows up in the serialized form this way.
     silo_id: Uuid,
+    /// the silo associated with this endpoint
+    #[serde(skip)]
+    db_silo: Arc<nexus_db_model::Silo>,
```
david-crespo, line 267: why does this need to be Arc?

nexus/src/app/external_endpoints.rs
```diff
@@ -595,6 +648,125 @@ impl rustls::server::ResolvesServerCert for NexusCertResolver {
...
+            return Ok(endpoint.clone());
+        }
+
+        // There was no endpoint for the requested name.  This should generally
+        // not happen in deployed systems where we expect people to have set up
+        // DNS to find the external endpoints.  But in development, we don't
+        // always have DNS set up.  People may use an IP address to get here.
+        // As a best-effort, we'll pick a default endpoint for this case.  (We
+        // may want to put this behind a policy knob that differs in
+        // production.)
```
david-crespo, line 751: The obvious question is what is the worst-case scenario where this does comes up in a production system? Seems fine because you either have creds for that silo or you don't, and if you don't, you can't do anything. Might be good to say that?

nexus/src/external_api/device_auth.rs
```diff
@@ -71,37 +71,15 @@ pub async fn device_auth_request(
...
-                            "error": "invalid_request",
-                            "error_description": format!("could not decode Host header: {}", e)
-                        }),
-                    )
-                }
-            }
-        };
-        let host = match host {
-            Some(host) => host,
-            None => {
```
david-crespo, line 99: nice to centralize this logic!

nexus/src/external_api/console_api.rs
```diff
@@ -546,40 +490,73 @@ pub struct LoginUrlQuery {
...
+                "silo_id" => ?silo.id()
+            );
+        }
+
+        format!(
+            "/login/{}/saml/{}",
+            silo.name(),
+            idps.into_iter().next().unwrap().name()
+        )
+    };
```
david-crespo, line 545: 👌

nexus/src/external_api/console_api.rs
```diff
@@ -546,40 +490,73 @@ pub struct LoginUrlQuery {
...
 
-    match serde_urlencoded::to_string(query_data) {
+    Ok(match serde_urlencoded::to_string(query_data) {
         // only put the ? in front if there's something there
         Ok(encoded) if !encoded.is_empty() => format!("{login_uri}?{encoded}"),
         // redirect_url is either None or not url-encodable for some reason
-        _ => login_uri.to_string(),
-    }
+        _ => login_uri,
+    })
```
david-crespo, line 556: Thinking out loud here, so bear with me. The `state` param will not work for SAML login, at least not directly. It was designed for local login: when the login POST comes back successful, the [console pulls the param from the URL](https://github.com/oxidecomputer/console/blob/1eb450f0a382cf5308b627c3597d7459cd3eb5cb/app/pages/LoginPage.tsx#L34) and navigates to it if present. For SAML login we have to send the state to the IdP so they can send it back to us on success. We do that with `RelayState`, which is a free-form thing — you can put anything in there. We have a referer param (analogous to `state`) and we only base64 it right now, but we will probably eventually want to sign it too, which means the console cannot construct it.

https://github.com/oxidecomputer/omicron/blob/8d32437da0fe9a56c63929c5a70083ebfeff7982/nexus/src/external_api/console_api.rs#L210-L236

For the SAML login page, I figured there would be a link button that goes to the IdP. Because we need the server to encode the redirect URL into `RelayState` (assuming we want to plan for a future where the thing is signed), this button would have to go to Nexus, including the `state` param in the URL again, and Nexus would return a redirect to the SAML IdP. Alternatively (and more or less equivalent) instead of hitting Nexus to redirect to the IdP, when rendering the page client-side we could make an API call that gives us the direct IdP URL and we stick that in the link `href`.

Here is a way it _could_ work, which is a bit silly and roundabout, but it's a straightforward consequence of the fact that we are not doing any console templating server-side.

1. User hits a 401 and gets send to `/login?state=/path/I/wanted`
2. `/login` determines we should use SAML and returns a redirect to `/login/my-silo/saml/my-idp?state=/path/I/wanted`
3. This is a "console page" so we render the console and show a page with a button that says "Go to IdP" or whatever
4. That button is a link to `/login/my-silo/saml/my-idp/direct?state=/path/I/wanted`, and when clicked, nexus responds with a 302 redirect to `my-idp.com/login/whatever?RelayState=kdfjhsdjdks54d565`

Everything after this point is already working, I think — after successful login, read the `referer` out of `RelayState` and redirect there.

Like I said, this feels silly because it requires a whole other endpoint in Nexus to turn `state` into `RelayState`. If we were trying to be Normal and render the login page server-side, we would be able to do that in the handler for `/login/my-silo/saml/my-idp`. I do often wonder when we're going to hit the point where it becomes worth it to do some server-side HTML templating instead of doing things like this, but having the console frontend continue to be a hermetically sealed box still seems safest to me. We could also punt on that and let console construct `RelayState` itself, but again that will stop working once we want to sign it.

david-crespo (approved):
Everything looks good pending that `state` param question. Even that can be probably be punted to a followup because nothing breaks if we leave it in.

nexus/src/external_api/console_api.rs
david-crespo: See draft console PR for SAML login page here:
https://github.com/oxidecomputer/console/pull/1549

david-crespo: While writing that PR, I thought about what words to put on the page to tell the user what is going on, and it occurred to me that that's an advantage of the approach where console makes an API call to get the complete IdP URL — it would let us include some other info like a name or description of the IdP.

image

end-to-end-tests/src/helpers/ctx.rs
davepacheco: That's cool.  Unwinding this will be a bit more work (there are two callers, each setting up reqwest clients for this) so I'd like to defer it from this PR.

nexus/src/app/external_endpoints.rs
davepacheco: I think it doesn't strictly need to be right now.  The intent is that we will support multiple endpoints for a Silo (e.g., a DNS alias).  Most of the code here I think is written to assume that might be the case and I didn't want to clone the whole record.

nexus/src/app/external_endpoints.rs
davepacheco: Updated this in 6991974e.  It's a little wishy-washy because it's describing what the current callers do in a case that should never happen.  i.e., it's not the impact of this code per se, and it's subject to change in the near future.  But agreed someone may want to better understand this.

nexus/src/external_api/console_api.rs
davepacheco: I think that's all reasonable, and also stuff I'd like to defer for now.  I think most of this PR is a prereq for that change anyway.  That okay?

davepacheco:
There's just one more automated test I'd like to write and I'll plan to land this. (though this is dragging on and I may land this sooner and add the test later)

nexus/src/external_api/console_api.rs
david-crespo: Yeah, I think it’s good to merge.

PR #3274, opened 2023-06-01
[meta] update to Rust 1.70

sunshowers:
There's a fix to Rust that I'm interested in picking up (can't find it right now but it makes FnMut and `&mut` pointers work in more cases).


iliana:
You will also need to change the Rust toolchain version in `.github/buildomat/jobs/*.sh`.

sunshowers:
Ah thanks!

sunshowers:
@iliana fixed!

davepacheco (approved):

PR #3276, opened 2023-06-01
Perform rack network initialization earlier in rack setup

### LINKED ISSUE #3271: Advertise boundary services tunnel endpoint address via ddm
In addition to setting up the boundary services tunnel endpoint address on the switch:

https://github.com/oxidecomputer/omicron/blob/83b37bd543675cc401da4b21ca87a420ca6653da/nexus/src/app/rack.rs#L366-L388

We also need to advertise this address through the `ddmd` instance running in the switch zone. This way a route to boundary services is propagated to all the sleds in the rack.

internet-diglett:
Perform rack network initialization before NTP sync is required, since NTP needs to communicate with the outside world.

* This splits the network configuration logic from the record creation logic, since CRDB is not online when the network is created (NTP must be sync'd before starting CRDB )
* Logic added for advertising boundary services loopback address via `ddm`
* Added PortSpeed setting to RackNetworkConfig

Closes  #3271

Awaiting https://github.com/oxidecomputer/dendrite/pull/356


sled-agent/src/rack_setup/service.rs
```diff
@@ -988,6 +1006,119 @@ impl ServiceInner {
...
+                ))
+                })?;
+
+            info!(self.log, "advertising boundary services loopback address");
+            let mut ddmd_addr = dpd_addr;
+            ddmd_addr.set_port(8000);
+            let ddmd_client = DdmAdminClient::new(&self.log, ddmd_addr)?;
+            ddmd_client.advertise_prefix(Ipv6Subnet::new(
+                BOUNDARY_SERVICES_ADDR.parse().unwrap(),
+            ));
```
internet-diglett, line 1119: Are these calls all idempotent?

sled-agent/src/rack_setup/service.rs
```diff
@@ -988,6 +1006,119 @@ impl ServiceInner {
...
+        // services will not be available and NTP will fail to sync
+        info!(self.log, "Checking for Rack Network Configuration");
+        if let Some(rack_network_config) = &config.rack_network_config {
+            info!(self.log, "Initializing Rack Network");
+            info!(self.log, "Looking up address for Dendrite");
+            let resolver = DnsResolver::new_from_subnet(
+                self.log.new(o!("component" => "DnsResolver")),
+                config.az_subnet(),
+            )
+            .expect("Failed to create DNS resolver");
```
rcgoodfellow, line 1019: Initially wasn't sure about reaching for `expect` here. But then I traced the `new_from_subnet` call down to a [`new_with_conn`](https://docs.rs/trust-dns-resolver/latest/src/trust_dns_resolver/async_resolver.rs.html#207) call, and it seems like this call cannot return an error? But I still wonder if we should return an error here instead of `expect`, so if the underlying implementation changes, we won't panic here.

sled-agent/src/rack_setup/service.rs
```diff
@@ -988,6 +1006,119 @@ impl ServiceInner {
...
+                    tag: "sled-agent".to_string(),
+                    log: self.log.new(o!(
+                        "component" => "DpdClient"
+                    )),
+                },
+            );
+
+            info!(self.log, "Building Rack Network Configuration");
+            let body = dpd_client::types::Ipv6Entry {
+                addr: BOUNDARY_SERVICES_ADDR.parse().map_err(|e| {
```
rcgoodfellow, line 1038: Need a follow-up issue for dynamically assigning boundary services addresses as this will not work for multiple switches.

sled-agent/src/rack_setup/service.rs
```diff
@@ -988,6 +1006,119 @@ impl ServiceInner {
...
+                    "unable to parse rack_network_config.uplink_up as IpAddr: {e}"))
+                })?;
+
+            let link_settings = LinkSettings {
+                // TODO Allow user to configure link properties
+                // https://github.com/oxidecomputer/omicron/issues/3061
+                params: LinkCreate {
+                    autoneg: false,
+                    kr: false,
+                    fec: PortFec::None,
```
rcgoodfellow, line 1068: Can we make FEC a parameter too?

sled-agent/src/rack_setup/service.rs
```diff
@@ -988,6 +1006,119 @@ impl ServiceInner {
...
+                        format!("unable to parse rack_network_config.gateway_ip as Ipv4Addr: {e}")
+                ))?);
+
+            dpd_port_settings.v4_routes.insert(
+                Ipv4Cidr { prefix: "0.0.0.0".parse().unwrap(), prefix_len: 0 }
+                    .to_string(),
+                RouteSettingsV4 { link_id: link_id.0, nexthop },
+            );
+
+            info!(self.log, "Waiting for dendrite to come online");
```
rcgoodfellow, line 1092: Maybe we should print this in the while loop so things are less mysterious if we get stuck here. Printing the error that comes back from `dpd_uptime` would be helpful too.

sled-agent/src/rack_setup/service.rs
rcgoodfellow: This ddm call for advertising prefixes is idempotent.

rcgoodfellow (commented):
Hey Levon - thanks for the quick turn on this. A few comments.

sled-agent/src/rack_setup/service.rs
internet-diglett: I yoinked this code from somewhere else and the `expect` came along for the ride 😂 
I'll handle the error appropriately, good catch.

nexus/src/app/rack.rs
```diff
@@ -455,59 +427,7 @@ impl super::Nexus {
...
-                }
-                .to_string(),
-                RouteSettingsV4 { link_id: link_id.0, nexthop },
-            );
-
-            self.dpd_client
-                .port_settings_apply(&port_id, &dpd_port_settings)
-                .await
-                .map_err(|e| Error::internal_error(&format!("unable to apply initial uplink port configuration: {e}")))?;
+            // TODO - record port speed
```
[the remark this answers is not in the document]
internet-diglett: https://github.com/oxidecomputer/omicron/issues/3277

sled-agent/src/rack_setup/service.rs
internet-diglett: https://github.com/oxidecomputer/omicron/issues/3278

sled-agent/src/rack_setup/service.rs
internet-diglett: `sled-agent` will now show the following while waiting for `dendrite`:
```
[2023-06-02T02:55:46.031Z]  INFO: SledAgent/19965 on workstation-1: Checking dendrite uptime (component=RSS)
...
[2023-06-02T02:56:01.032Z]  INFO: SledAgent/19965 on workstation-1: Unable to check Dendrite uptime (component=RSS)
    reason: Communication Error: error sending request for url (http://[fd00:1122:3344:101::2]:12224/dpd-uptime): operation timed out
[2023-06-02T02:56:01.032Z]  INFO: SledAgent/19965 on workstation-1: Waiting for dendrite to come online (component=RSS)
```

rcgoodfellow (approved):

PR #3281, opened 2023-06-02
Do not error if encrypted fileystem is mounted

andrewjstone:


andrewjstone:
This fixes a bug on dogood rack that will block RSS. Merging.

PR #3314, opened 2023-06-06
Use native helios-build image name tokens

citrus-it:
Rather than retrieving the package repository ourselves and parsing the OS hash out of it, we can make use of the new token feature in helios-build and have it populate the OS hash during build. This also closes the small window that could result in the image name having an older hash than the image.


citrus-it:
I extracted the two OS archives from CI on this PR and checked the image names were as expected:

```
atrium% lookimage core/image/zfs.img | head -1
image name = ci e049cd5/fe6b465 2023-06-06 21:39
atrium% lookimage tramp/image/zfs.img | head -1
image name = recovery e049cd5/fe6b465 2023-06-06 21:38
```

jgallagher (approved):

PR #3315, opened 2023-06-06
Do not accept IDP metadata documents without keys

### LINKED ISSUE #3155: Require `signing_keypair` on IdP create
Current these two fields are optional. I talked to @jmpesp and we decided they shouldn’t be.

jmpesp:
Do not initialize a SAML identity provider if the provided IDP metadata document does not have a signing key. The control plane must require and only accept signed assertions.

Fixes #3155

### DIFF
--- a/nexus/src/app/silo.rs
+++ b/nexus/src/app/silo.rs
@@ -828,6 +828,59 @@ impl super::Nexus {
             }
         };
 
+        // Once the IDP metadata document is available, parse it into an
+        // EntityDescriptor
+        use samael::metadata::EntityDescriptor;
+        let idp_metadata: EntityDescriptor =
+            idp_metadata_document_string.parse().map_err(|e| {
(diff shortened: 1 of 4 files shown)


luqmana:
This doesn't actually seem to change `signing_keypair` to not being an `Option`? what is actually the relationship between that field and the signing certificate within the IdP metadata?

jmpesp:
> This doesn't actually seem to change `signing_keypair` to not being an `Option`? what is actually the relationship between that field and the signing certificate within the IdP metadata?

My mistake, I should have updated the issue: when @david-crespo and I were talking, I was referring to the IDP metadata signing key, not the SP (aka us) signing key. Our signing key can remain optional.

jmpesp:
(note I pushed up a test that was missing from our coverage)

david-crespo (approved):

PR #3316, opened 2023-06-07
Use `zone_enter` instead of `zlogin` to avoid shell expansion.

### LINKED ISSUE #3216: Commands sled-agent runs through zlogin are subject to shell expansion
While investigating why dendrite didn't see a SIGUSR1 when it should've in #3199, @bnaecker noticed this oddity in sled-agent's logs:

```
02:15:10.006Z WARN SledAgent: Failed to activate switch: Failed to issue SMF command: Failed to do 'del config/address smf property value' by running command in zone: Error running command in zone 'oxz_switch': Command [/usr/sbin/zlogin oxz_switch /usr/sbin/svccfg -s svc:/system/illumos/dendrite delpropvalue config/address *] executed and failed with status: exit status: 1  stdout:   stderr: svccfg: Syntax error.
    sled_id = 96c409e3-a150-46de-9915-25b839ba2508
```

On its surface this doesn't make any sense:

```
/usr/sbin/svccfg -s svc:/system/illumos/dendrite delpropvalue config/address *
```

is perfectly valid syntax, and we use the exact same wrapper to delete properties when configuring MGS. [Our SMF wrapper uses std::process::Command](https://github.com/oxidecomputer/omicron/blob/ea9f70a3bd9300bc1adfecac6db1e159613039a9/illumos-utils/src/running_zone.rs#L162-L170), so we should not (ominous sounds) be subject to shell expansion.

However, we _are_ getting shell expansion by virtue of the `zlogin`. Reproducing is trivial outside of sled-agent:

```
BRM44220001 # /usr/sbin/zlogin oxz_switch ls -l
total 2
-rw-r--r--   1 root     root           0 Dec 27 18:00 bar
-rw-r--r--   1 root     root           0 Dec 27 18:00 foo
BRM44220001 # echo '*'
*
BRM44220001 # /usr/sbin/zlogin oxz_switch echo '*'
bar foo
```

This also explains the syntax error we see from svccfg, if there are any files in root's home directory:

```
BRM44220001 # /usr/sbin/zlogin oxz_switch /usr/sbin/svccfg -s svc:/system/illumos/dendrite delpropvalue config/address '*'
svccfg: Syntax error.
```

bnaecker:
- Fixes #3216
- For running zones, use the `Command::pre_exec` hook to call `zone_enter()` between `fork()` and `exec()`. This lets us run the command directly in the zone, without using `zlogin`. The latter runs all its arguments through a shell in the form of `sh -c`, which means they're subject to expansion. Using `zone_enter()` instead avoids any expansion.
- Add `Zones::id()` method, which mostly exists for testing. `mockall` requires us to specify the return value for the mock methods, but the `Zone` type we're using has private fields, so it can't be easily faked.

### DIFF
--- a/illumos-utils/src/running_zone.rs
+++ b/illumos-utils/src/running_zone.rs
@@ -173,10 +173,9 @@ mod zenter {
     impl Drop for Template {
         fn drop(&mut self) {
             self.clear();
-            if unsafe { libc::close(self.fd) } != 0 {
-                let e = std::io::Error::last_os_error();
-                eprintln!("failed to close ctfs template file: {e}");
-            }
+            // Ignore any error, since printing may interfere with `slog`'s
+            // structured output.
+            unsafe { libc::close(self.fd) };
         }
     }
 
--- a/illumos-utils/src/running_zone.rs
+++ b/illumos-utils/src/running_zone.rs
@@ -308,7 +308,8 @@ impl RunningZone {
         // it must actually call `crate::execute()` for the testing purposes.
         // That's mocked by `mockall` to return known data, and so the command
         // that's actually run is irrelevant.
-        let command = std::process::Command::new("echo").args(args);
+        let command = std::process::Command::new("echo");
+        let command = command.args(args);
         crate::execute(command)
             .map_err(|err| RunCommandError {
                 zone: self.name().to_string(),
--- a/illumos-utils/src/running_zone.rs
+++ b/illumos-utils/src/running_zone.rs
@@ -257,7 +257,12 @@ impl RunningZone {
         S: AsRef<std::ffi::OsStr>,
     {
         use std::os::unix::process::CommandExt;
-        let id = self.id.expect("Must have a zone ID for a running zone");
+        let Some(id) = self.id else {
+            return Err(RunCommandError {
+                zone: self.name().to_string(),
+                err: crate::ExecutionError::NotRunning,
+            });
+        };
         let template =
             std::sync::Arc::new(zenter::Template::new().map_err(|err| {
                 RunCommandError { zone: self.name().to_string(), err }
--- a/illumos-utils/src/running_zone.rs
+++ b/illumos-utils/src/running_zone.rs
@@ -172,6 +172,7 @@ mod zenter {
 
     impl Drop for Template {
         fn drop(&mut self) {
+            self.clear();
             if unsafe { libc::close(self.fd) } != 0 {
                 let e = std::io::Error::last_os_error();
                 eprintln!("failed to close ctfs template file: {e}");
@@ -250,6 +251,22 @@ impl RunningZone {
     }
 
     /// Runs a command within the Zone, return the output.
+    //
+    // NOTE: It's important that this function is synchronous.
+    //
+    // Internally, we're setting the (thread-local) contract template before
+    // forking a child to exec the command inside the target zone. In order for
+    // that to all work correctly, that template must be set and then later
+    // cleared in the _same_ OS thread. An async method here would open the
+    // possibility that the template is set in some thread, and then cleared in
+    // another, if the task is swapped out at an await point. That would leave
+    // the first thread's template in a modified state.
+    //
+    // If we do need to make this method synchronous, we will need to change the
+    // internals to avoid changing the thread's contract. One possible approach
+    // here would be to use `libscf` directly, rather than `exec`-ing `svccfg`
+    // directly in a forked childe. That would obviate the need to work on the
+    // contract at all.
     #[cfg(target_os = "illumos")]
     pub fn run_cmd<I, S>(&self, args: I) -> Result<String, RunCommandError>
     where
--- a/illumos-utils/src/running_zone.rs
+++ b/illumos-utils/src/running_zone.rs
@@ -42,6 +42,9 @@ pub enum BootError {
     #[error("Zone booted, but timed out waiting for {service} in {zone}")]
     Timeout { service: String, zone: String },
 
+    #[error("Zone booted, but failed to find zone ID for zone {zone}")]
+    NoZoneId { zone: String },
+
     #[error("Zone booted, but running a command experienced an error: {0}")]
     RunCommandError(#[from] RunCommandError),
 }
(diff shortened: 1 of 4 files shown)


illumos-utils/src/running_zone.rs
```diff
@@ -137,9 +140,102 @@ pub enum GetZoneError {
...
+                } != 0
+                || unsafe { ct_tmpl_activate(fd) } != 0
+            {
+                let err = std::io::Error::last_os_error();
+                return Err(crate::ExecutionError::ZoneEnter { err });
+            }
+            Ok(Self { fd })
+        }
+
+        pub fn clear(&self) {
```
jgallagher, line 229: Should this return an error on failure?

illumos-utils/src/running_zone.rs
```diff
@@ -586,8 +727,7 @@ impl RunningZone {
     ///
     /// Allows callers to synchronously stop a zone, and inspect an error.
     pub async fn stop(&mut self) -> Result<(), String> {
```
jgallagher, line 729: Could this take `mut self` instead of `&mut self`? With the change from `self.running` to `self.id.take()`, if a caller incorrectly calls `run_cmd()` after `stop()` it will now panic on the `self.id.expect(...)` in `run_cmd()`. Taking `mut self` here would make that impossible, although it would also mean a failure here would leave the zone in a zombie state.

Maybe the zombie state issue is already present regardless of `&mut self` though - if `halt_and_remove_logged()` fails, we've already stolen `self.id`, which means neither another `stop()` nor `drop()` will try again to shut down the zone. It's unclear to me what a caller would expect if `stop()` fails, though.

illumos-utils/src/running_zone.rs
```diff
@@ -154,26 +250,66 @@ impl RunningZone {
...
+    #[cfg(target_os = "illumos")]
     pub fn run_cmd<I, S>(&self, args: I) -> Result<String, RunCommandError>
     where
         I: IntoIterator<Item = S>,
         S: AsRef<std::ffi::OsStr>,
     {
+        use std::os::unix::process::CommandExt;
+        let id = self.id.expect("Must have a zone ID for a running zone");
+        let template =
+            std::sync::Arc::new(zenter::Template::new().map_err(|err| {
```
jgallagher, line 262: I don't really understand the template things going on, but want to double check: all of the raw calls in `zenter::Template::new()` affect our _current_ process, right? Do we have to worry about thread safety or panic safety here? The thread safety concern specifically being - can what we're doing in `Template::new()` mess up some other thread that might be doing arbitrary things, including potentially calling `run_cmd` on this or some other zone? And panic safety: if we panic between this line and `template.clear()`, `template`'s drop will close the fd but won't clear the template - is that okay?

illumos-utils/src/running_zone.rs
```diff
@@ -137,9 +140,102 @@ pub enum GetZoneError {
...
+    #[derive(Debug)]
+    pub struct Template {
+        fd: c_int,
+    }
+
+    impl Drop for Template {
+        fn drop(&mut self) {
+            if unsafe { libc::close(self.fd) } != 0 {
+                let e = std::io::Error::last_os_error();
+                eprintln!("failed to close ctfs template file: {e}");
```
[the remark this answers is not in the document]
bnaecker: It probably will, yes. I'll look into getting a logger down here, though it might take a bit more work since I don't believe the `RunningZone` object has one either.

illumos-utils/src/running_zone.rs
bnaecker: Maybe? I'm not sure what to do with it though. In particular, suppose that we successfully execute the subprocess in the zone, but calling `clear()` fails. Should that return an error, even though the actual guts of `run_cmd()` worked?

illumos-utils/src/running_zone.rs
bnaecker: I agree, this isn't great. At least we should check for `self.id.is_some()` in `run_cmd()`, to avoid the panic. It's probably a programming error to call `stop()` more than once, but we should still not panic I think.

I think we can't take self by value here, at least not without more significant changes at the call sites. The zombie issue was already there, even with the boolean. Calling `stop()` only works at most once, since it previously set `self.running = false;`. IIRC, the point of `stop()` was just to be able to inspect / log the failure.

illumos-utils/src/running_zone.rs
bnaecker: > all of the raw calls in zenter::Template::new() affect our current process, right?

My understanding is that they affect the current _thread_. From `man ct_tmpl_clear`:

> The `ct_tmpl_clear()` function clears calling thread's active template.

So I think if another thread is calling `run_cmd` on another zone, that should not interfere. The calls will set that thread's template, independently from this one. If it's the same zone, I think the same thing applies. All this is synchronous, so this shouldn't be swapped around onto different threads, as might happen if it were a `tokio` task.

But now I'm worried about `tokio`...That should be OK, though, right? What I mean is, there is no await point inside `run_cmd`, which means that the task will run that method to completion without yielding. It'll not be swapped out part-way through, where another call to `run_cmd()` might activate / clear a different template on the same OS thread. Does that sound right? It makes me nervous nonetheless.

> And panic safety: if we panic between this line and template.clear(), template's drop will close the fd but won't clear the template - is that okay?

I think so, again because the template is activated for a thread, not the process. (We also abort rather than unwind, but I'm not sure that's relevant to your question.)

illumos-utils/src/running_zone.rs
```diff
@@ -137,9 +140,102 @@ pub enum GetZoneError {
...
+        // Process experienced an uncorrectable error.
+        const CT_PR_EV_HWERR: c_uint = 0x20;
+        // Only kill process group on fatal errors.
+        const CT_PR_PGRPONLY: c_uint = 0x04;
+        // Automatically detach inherited contracts.
+        const CT_PR_REGENT: c_uint = 0x08;
+
+        pub fn new() -> Result<Self, crate::ExecutionError> {
+            let path = CStr::from_bytes_with_nul(Self::TEMPLATE_PATH).unwrap();
+            let fd = unsafe { libc::open(path.as_ptr(), libc::O_RDWR) };
```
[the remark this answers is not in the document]
bnaecker: So, there _is_ something special about that path. Using `File::open()` throws back `EINVAL`.

illumos-utils/src/running_zone.rs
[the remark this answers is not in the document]
bnaecker: There's no straightforward way to get a logger here, since we have to go quite a ways up the object tree to get to one that has a logger. Bummer, but we'll live with it for now.

illumos-utils/src/running_zone.rs
jgallagher: > But now I'm worried about tokio...That should be OK, though, right? What I mean is, there is no await point inside run_cmd, which means that the task will run that method to completion without yielding. It'll not be swapped out part-way through, where another call to run_cmd() might activate / clear a different template on the same OS thread. Does that sound right? It makes me nervous nonetheless.

I think this is right; in particular, `run_cmd` is not even an async function, so there's no opportunity for tokio to move it around. It might be worth adding a comment noting that it's important that this function continues to be non-async because it does illumos-specific thread-local things that would not play nicely with tokio.

> I think so, again because the template is activated for a thread, not the process. (We also abort rather than unwind, but I'm not sure that's relevant to your question.)

Abort instead of unwind is definitely relevant, because if we abort we don't have to worry about panic safety at all. If we want this to be panic-safe from an unwind point of view, I'm not sure it is: if I'm on a thread that calls this function, but inside (possibly deeply nested inside) a [catch_unwind](https://doc.rust-lang.org/std/panic/fn.catch_unwind.html), it's possible a panic here could leave my thread around with this template activated. That's a lot of "if"s that don't apply in practice, so maybe that is also worth just commenting? The real fix would be to have the `clear` called in a `Drop` method, guaranteeing it would happen even if we unwound.

bnaecker: Done in bbdffe8f1

illumos-utils/src/running_zone.rs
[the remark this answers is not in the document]
jgallagher: I'm very tempted to propose removing the `eprintln!()` altogether then. :-/ I worry this will get intermingled with slog output is a weird way. Not logging errors on close is not great but does have precedence: normal `File`s do the same thing.

illumos-utils/src/running_zone.rs
```diff
@@ -154,26 +251,88 @@ impl RunningZone {
...
+    //
+    // Internally, we're setting the (thread-local) contract template before
+    // forking a child to exec the command inside the target zone. In order for
+    // that to all work correctly, that template must be set and then later
+    // cleared in the _same_ OS thread. An async method here would open the
+    // possibility that the template is set in some thread, and then cleared in
+    // another, if the task is swapped out at an await point. That would leave
+    // the first thread's template in a modified state.
+    //
+    // If we do need to make this method synchronous, we will need to change the
```
jgallagher, line 265:
```suggestion
    // If we do need to make this method asynchronous, we will need to change the
```

illumos-utils/src/running_zone.rs
```diff
@@ -154,26 +251,88 @@ impl RunningZone {
...
+    // internals to avoid changing the thread's contract. One possible approach
+    // here would be to use `libscf` directly, rather than `exec`-ing `svccfg`
+    // directly in a forked childe. That would obviate the need to work on the
```
jgallagher, line 268:
```suggestion
    // directly in a forked child. That would obviate the need to work on the
```

jgallagher (approved):
(Just in case github is still hiding it, I'd lean toward removing the `eprintln!()`.)

Thanks for doing this - removes a huge footgun. :+1:

illumos-utils/src/running_zone.rs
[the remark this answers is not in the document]
bnaecker: Sure that seems OK to me too. I'm pretty agnostic. We also mix an `eprintln!()` call with `slog` output when we fail to clean up other resources, like VNICs. So there's even precedent within sled-agent! But it's still not the best. Will remove for now.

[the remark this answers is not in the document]
bnaecker: Removed in 58c78606c

illumos-utils/src/running_zone.rs
bnaecker: Thanks, fixed!

PR #3318, opened 2023-06-07
Fix SAML login redirect

david-crespo:
When you get logged out in the console or get linked to a page while not logged in (like `/device/verify` for authenticating in the CLI), we want to send you somewhere you can log in and, after you succeed, we want to send you back where you were trying to go. In order to do that, we need to keep track of the original target page and hold onto it through the login process. 

With local login, this is easy: in the case of a server-side 401, we take the current path, and redirect to `/login/{silo}/local?redirect_uri={path}`, which loads a console page that handles username/password login through a background API call. When the login POST comes back successful, we pull the target path from the `state` query param and nav to it client-side ([console source](https://github.com/oxidecomputer/console/blob/d52e87d69877e9dd63254dfa44ef8c284a2d3518/app/pages/LoginPage.tsx#L29-L36)). (We need to make sure you can't link someone to `oxide-rack.com/login/{silo}/local?redirect_uri=https://evil.com/virus` — see to-do list below.) A client-side 401 is almost the same, except there's one more step at the beginning where we nav to `/login?redirect_uri={current_path}`, and then that handler sends you to local or SAML login as appropriate, making sure to pass along the `state` param (very important!).

With SAML login it's slightly more complicated because we need to send the target URL over to the IdP using SAML's `RelayState` concept and they will send it back with the login POST. Most of that is already in place, so the main thing this change does is pull the target URL from the `state` query param server side instead of the `referer` header and stick it in the `RelayState` we get back from the IdP on login.

### What's in here

- [x] Move existing redirect to IdP logic from `/login/{silo}/saml/{provider}` to `/login/{silo}/saml/{provider}/redirect`, `/login/{silo}/saml/{provider}` now serves console landing page
- [x] Get post-login redirect URL from `state` query param instead of `referer` header
- [x] Rename `state` param to `redirect_uri` and use that in both the query params and the RelayState object
- [x] Validate that `redirect_uri` is a relative path to prevent someone being given a login link with an arbitrary URL 45e2fbe93bef09a1c04ecc8ccee79b221ef3d5ed
- [x] Better tests, probably
-  ~~Change state param from path to full URL per this comment~~ Relative URIs are valid in a 302 response Location header ([RFC 7231](https://www.rfc-editor.org/rfc/rfc7231#section-7.1.2)). I wrote that comment because I thought the IdP would be returning a redirect to Nexus, but it seems they post to Nexus from the browser, and browsers resolve a relative Location against the request URI
    https://github.com/oxidecomputer/omicron/blob/8a1fbdfb8244d232274ee03d82a47b3186b89cf4/nexus/src/external_api/console_api.rs#L593-L599

### Why not the `referer` header

`referer` certainly does sound like the natural choice, but browser behavior around it is inconsistent, especially in the case of multiple redirects. When you run into a 401 client-side (like if your session expired), the console uses `window.location.assign` to send the user to `/login` on Nexus, which will return a redirect to either `/login/{silo}/local` or `/login/{silo}/saml/{idp}` depending on the silo.

https://github.com/oxidecomputer/omicron/blob/8a1fbdfb8244d232274ee03d82a47b3186b89cf4/nexus/src/external_api/console_api.rs#L559-L577

In theory the original `referer` should be passed through those redirects and accessible on the final page, but I don't see it when I test it manually. I might do more testing, but overall I don't trust it. It's browser-specific, who knows what could happen. For now I think it's more reliable to use a query param we fully control.

### DIFF
--- a/nexus/src/external_api/console_api.rs
+++ b/nexus/src/external_api/console_api.rs
@@ -179,16 +179,33 @@ impl RelayState {
     }
 }
 
-/// Prompt user login
-///
-/// Either display a page asking a user for their credentials, or redirect them
-/// to their identity provider.
+/// SAML login console page (just a link to the IdP)
 #[endpoint {
    method = GET,
    path = "/login/{silo_name}/saml/{provider_name}",
    tags = ["login"],
+   unpublished = true,
 }]
 pub async fn login_saml_begin(
+    rqctx: RequestContext<Arc<ServerContext>>,
+    _path_params: Path<LoginToProviderPathParam>,
+    _query_params: Query<LoginUrlQuery>,
+) -> Result<Response<Body>, HttpError> {
+    serve_console_index(rqctx.context()).await
+}
+
+/// Get a redirect straight to the IdP
+///
+/// Console uses this to avoid having to ask the API anything about the IdP. It
+/// already knows the IdP name from the path, so it can just link to this path
+/// and rely on Nexus to redirect to the actual IdP.
+#[endpoint {
+   method = GET,
+   path = "/login/{silo_name}/saml/{provider_name}/redirect",
+   tags = ["login"],
+   unpublished = true,
+}]
+pub async fn login_saml_redirect(
     rqctx: RequestContext<Arc<ServerContext>>,
     path_params: Path<LoginToProviderPathParam>,
     query_params: Query<LoginUrlQuery>,
--- a/nexus/src/external_api/console_api.rs
+++ b/nexus/src/external_api/console_api.rs
@@ -93,7 +93,7 @@ use std::{collections::HashSet, ffi::OsString, path::PathBuf, sync::Arc};
 //   https://some.idp.test/auth/saml?SAMLRequest=...&RelayState=...&SigAlg=...&Signature=...
 //
 // SAMLRequest is base64 encoded zlib compressed XML, and RelayState can be
-// anything - Nexus currently encodes the referer header so that when SAML login
+// anything - Nexus currently encodes a redirect_uri so that when SAML login
 // is successful the user can be sent back to where they were originally.
 //
 // The user will then authenticate with that IdP, and if successful will be
@@ -114,8 +114,8 @@ use std::{collections::HashSet, ffi::OsString, path::PathBuf, sync::Arc};
 //
 // The IdP's SAMLResponse will authenticate a subject, and from this external
 // subject a silo user has to be created or retrieved (depending on the Silo's
-// user provision type). After that, users will be redirected to the referer in
-// the relay state, or to `/organizations`.
+// user provision type). After that, users will be redirected to the `redirect_uri`
+// in the relay state, or to `/organizations`.
 //
 // SAML logout flow
 // ----------------
@@ -153,7 +153,7 @@ pub struct LoginToProviderPathParam {
 
 #[derive(Clone, Debug, Deserialize, Serialize, JsonSchema)]
 pub struct RelayState {
-    pub referer: Option<String>,
+    pub redirect_uri: Option<String>,
 }
 
 impl RelayState {
@@ -191,7 +191,7 @@ impl RelayState {
 pub async fn login_saml_begin(
     rqctx: RequestContext<Arc<ServerContext>>,
     path_params: Path<LoginToProviderPathParam>,
-    query_params: Query<StateParam>,
+    query_params: Query<LoginUrlQuery>,
 ) -> Result<HttpResponseFound, HttpError> {
     let apictx = rqctx.context();
     let handler = async {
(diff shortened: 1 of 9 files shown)


jmpesp:
> Think about validation to prevent someone being given a login link with an arbitrary URL

How about requiring the state to be a relative path, not any old URL?

david-crespo:
Yep, that's a good idea and consistent with what I have so far. I could even try to validate that at the Dropshot serde level. The other safeguard could/should be that we sign `RelayState`.

jmpesp:
> The other safeguard could/should be that we sign RelayState

Nexus does sign the redirect if it was handed signing keys: https://github.com/oxidecomputer/omicron/blob/c726e2f03c94b474a017ada24e7d58460b685605/nexus/db-queries/src/authn/silos.rs#L165-L172

david-crespo:
I can't figure out where we check the signature. Where's that?

jmpesp:
> Cool, missed that. I can't figure out where we check the signature. Where's that?

That happens in the samael library: https://github.com/njaremko/samael/blob/4974e40ac036462baaf4eff1c0c33b4a6802b6d0/src/crypto.rs#L550

nexus/src/external_api/console_api.rs
```diff
@@ -506,17 +510,54 @@ pub struct RestPathParam {
...
+    type Error = String;
+
+    fn try_from(uri: Uri) -> Result<Self, Self::Error> {
+        if uri.host().is_none() && uri.scheme().is_none() {
+            Ok(Self(uri.to_string()))
+        } else {
+            Err(format!("\"{}\" is not a relative URI", uri))
+        }
+    }
+}
```
david-crespo, line 538: Not strictly necessary to have a separate `TryFrom<Uri>` (I thought I was going to need it, but didn't) but it's a reasonable way to factor it out anyway.

nexus/tests/integration_tests/console_api.rs
```diff
@@ -467,21 +467,46 @@ async fn test_login_redirect_simple(cptestctx: &ControlPlaneTestContext) {
...
 
-    // empty state param gets dropped
-    expect_redirect(testctx, "/login?state=", &expected).await;
+    let paths = [
+        "/login",
+        "/login/my-silo/local",
+        "/login/my-silo/saml/my-idp",
+        "/login/my-silo/saml/my-idp/redirect",
+    ];
+    let bad_uris = ["foo", "", "http://example.com"];
```
david-crespo, line 495: Erroring out on `?redirect_uri=` feels a little un-generous when we could treat it as `None` instead, but I'm not sure how to accomplish this. There's probably a serde attr one-liner I could put on the param that would make it handle empty string specially.

nexus/src/external_api/console_api.rs
```diff
@@ -36,12 +36,48 @@ use nexus_types::external_api::params;
...
 use omicron_common::api::external::{DataPageParams, Error, NameOrId};
+use parse_display::Display;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 use serde_urlencoded;
 use std::num::NonZeroU32;
+use std::str::FromStr;
 use std::{collections::HashSet, ffi::OsString, path::PathBuf, sync::Arc};
 
+// -----------------------------------------------------
```
davepacheco, line 47: Thanks for adding this.

nexus/src/external_api/console_api.rs
```diff
@@ -354,6 +386,7 @@ pub struct LoginPathParam {
 pub async fn login_local_begin(
     rqctx: RequestContext<Arc<ServerContext>>,
     _path_params: Path<LoginPathParam>,
+    _query_params: Query<LoginUrlQuery>,
```
davepacheco, line 389: Why is this here?  I see the endpoint is unpublished so it's not to appear in the spec, and it doesn't seem used.

davepacheco (approved):
Thanks for writing up the detailed summary above..  That made this a lot easier to review.

jmpesp (approved):
Looks good to me, :ship:

nexus/src/external_api/console_api.rs
```diff
@@ -36,12 +36,48 @@ use nexus_types::external_api::params;
...
+//
+// Local login is very simple. We show a login form, the username and password
+// are POSTed to the API, and on success, the console pulls `redirect_uri` out
+// of the query params and navigates to that route client-side.
+//
+// SAML is more complicated. /login/{silo}/saml/{provider}?redirect_uri={} shows
+// a page with a link to /login/{silo}/saml/{provider}/redirect?redirect_uri={}
+// (note the /redirect), which redirects to the IdP  with `redirect_uri` encoded
+// in a RelayState query param). On successful login in the IdP, the IdP will
+// POST /login/{silo}/saml/{} with a body including that redirect_uri, so that
```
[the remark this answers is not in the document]
david-crespo: Yep

nexus/src/external_api/console_api.rs
david-crespo: The idea is that the redirect_uri is still being validated as a relative path server-side. There’s a new test exercising this I believe.

david-crespo:
The [Console side of this](https://github.com/oxidecomputer/console/pull/1549) is ready to go as well. I will merge that as soon as this is in so I can bump console to the latest API version.

PR #3320, opened 2023-06-08
Update crucible and propolis

jmpesp:
Bump crucible rev to pick up:

- Accept a length for pantry validation requests
- use workspace dependencies
- Allow some jobs to remain behind
- Update the test_mem.sh test to show region sizes too.
- Test live repair using dummy downstairs
- test-live-repair is not finishing, disable for now
- New dtrace counters for upstairs state of downstairs connections
- Allow dsc more than three downstairs.
- Fix LR hang, polish dtrace, make CI test more
- Update all Omicron packages in Cargo.lock
- Fix double counting of repair jobs
- remove some &mut self for Downstairs
- Update Rust crate slog-term to 2.9
- Update Rust crate proptest to 1.2.0
- Update Rust crate uuid to 1.3.3
- Update Rust crate reedline to 0.19.1
- Update Rust crate libc to 0.2.144
- Update Rust crate base64 to 0.21.2

Bump propolis rev to pick up:

- server: improve HTTP error codes
- Bump crucible rev to latest

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -452,7 +452,7 @@ dependencies = [
 [[package]]
 name = "bhyve_api"
 version = "0.0.0"
-source = "git+https://github.com/oxidecomputer/propolis?rev=78314ae5f74f951c03b0cb97b06f9bc28446bdb5#78314ae5f74f951c03b0cb97b06f9bc28446bdb5"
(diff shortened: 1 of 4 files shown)


gjcolombo (approved):
LGTM. We need some follow-up changes in sled agent to get Propolis `96ba33d2` (the server error code change) to produce its desired effect, but I think it's safe to ingest that change as-is (i.e. I don't think it'll make anything worse, it just won't make anything better until I merge the requisite follow-up changes).

leftwo (approved):

PR #3321, opened 2023-06-08
Remove `silo` path param from login endpoints

david-crespo:
We don't need silo in the paths now that we have it from DNS. This is on top of #3318, but it'll reset to being based on main once that's merged. The tests are doing surprisingly well on the first try, but there might need to be some tweaks to tests that work on multiple silos.

### DIFF
--- a/end-to-end-tests/src/helpers/ctx.rs
+++ b/end-to-end-tests/src/helpers/ctx.rs
@@ -196,8 +196,7 @@ pub async fn build_client() -> Result<oxide_client::Client> {
     let client_params = ClientParams::new()?;
     let config = &client_params.rss_config;
     let base_url = client_params.base_url();
-    let silo_name = config.recovery_silo.silo_name.as_str();
-    let login_url = format!("{}/v1/login/{}/local", base_url, silo_name);
+    let login_url = format!("{}/v1/login/local", base_url);
     let username: oxide_client::types::UserId =
         config.recovery_silo.user_name.as_str().parse().map_err(|s| {
             anyhow!("parsing configured recovery user name: {:?}", s)
--- a/nexus/src/external_api/console_api.rs
+++ b/nexus/src/external_api/console_api.rs
@@ -54,10 +54,9 @@ use std::{collections::HashSet, ffi::OsString, path::PathBuf, sync::Arc};
 // 1. Navigate an auth-gated console page (e.g., /projects) directly while
 //    logged out. This goes through `console_index_or_login_redirect` and
 //    therefore results in a redirect straight to either
-//    /login/{silo}/local?redirect_uri={} or
-//    /login/{silo}/saml/{provider}?redirect_uri={} depending on the silo's
-//    authn mode. Nexus takes the path the user was trying to hit and sticks it
-//    in a `redirect_uri` query param.
+//    /login/local?redirect_uri={} or /login/saml/{provider}?redirect_uri={}
+//    depending on the silo's authn mode. Nexus takes the path the user was
+//    trying to hit and sticks it in a `redirect_uri` query param.
 // 2. Hit a 401 on a background API call while already in the console (for
 //    example, if session expires while in use). In that case, the console will
 //    navigate to `/login?redirect_uri={current_path}`, which will respond with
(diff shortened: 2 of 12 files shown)


davepacheco:
It seems like it'd still be useful to have the silo name there for the development case where DNS is not set up, or else you'll have no way to pick which silo you want to access?  I'm not sure if that's worth keeping it here but it seems like a pretty big downside.

david-crespo:
Definitely a downside. Now that I think about it, having the silo in the path on these endpoints was a sneaky hack around the lack of silo info in all the other endpoints. As long as you can _sign in_ to a specific silo, the rest of the endpoints just use your session to figure out which one you're in.

My main motivation here is that `my-silo.sys.rack.whatever/login/my-silo/local` looks kind of stupid, but I have to concede it's not a big deal. One option would be to support an optional silo query parameter instead (that is only checked when DNS has nothing, so you can't override the DNS-dervied silo), so the redundancy is eliminated in production but those in the know can use it in dev to get to a particular silo. Then again, when I write out that the aim is to eliminate silly-looking redundancy, doing all this stuff doesn't seem worth it.

---

I see that there's an external DNS server in test setup — does that mean that it should not be that bad to fix these tests?

```
FAIL [  22.585s] omicron-nexus::test_all integration_tests::saml::test_create_a_hidden_silo_saml_idp
FAIL [  22.884s] omicron-nexus::test_all integration_tests::saml::test_create_a_saml_idp
FAIL [  25.007s] omicron-nexus::test_all integration_tests::saml::test_post_saml_response
FAIL [  23.542s] omicron-nexus::test_all integration_tests::saml::test_post_saml_response_with_relay_state
FAIL [  24.479s] omicron-nexus::test_all integration_tests::silos::test_saml_idp_metadata_data_valid
```

davepacheco:
> Definitely a downside. Now that I think about it, having the silo in the path on these endpoints was a sneaky hack around the lack of silo info in all the other endpoints. As long as you can _sign in_ to a specific silo, the rest of the endpoints just use your session to figure out which one you're in.
> 
> My main motivation here is that `my-silo.sys.rack.whatever/login/my-silo/local` looks kind of stupid, but I have to concede it's not a big deal. One option would be to support an optional silo query parameter instead (that is only checked when DNS has nothing, so you can't override the DNS-dervied silo), so the redundancy is eliminated in production but those in the know can use it in dev to get to a particular silo. Then again, when I write out that the aim is to eliminate silly-looking redundancy, doing all this stuff doesn't seem worth it.

Agreed on all parts.  A query param would make it look cleaner in practice but still let folks do what they need in development, and also might not be worth the time 🤷 


> I see that there's an external DNS server in test setup — does that mean that it should not be that bad to fix these tests?
> 
> ```
> FAIL [  22.585s] omicron-nexus::test_all integration_tests::saml::test_create_a_hidden_silo_saml_idp
> FAIL [  22.884s] omicron-nexus::test_all integration_tests::saml::test_create_a_saml_idp
> FAIL [  25.007s] omicron-nexus::test_all integration_tests::saml::test_post_saml_response
> FAIL [  23.542s] omicron-nexus::test_all integration_tests::saml::test_post_saml_response_with_relay_state
> FAIL [  24.479s] omicron-nexus::test_all integration_tests::silos::test_saml_idp_metadata_data_valid
> ```

Sadly, I think the external DNS server doesn't help because of the way the low-level `NexusRequest`/`RequestContext` things work (which are what we use almost everywhere in the test suite, I think).  I believe those things use a HTTP 1.1 `hyper::Client` that's pointed at the IP address of Nexus.  I think if you specified the "Host" header as the DNS name of whatever Silo you wanted to hit, that would probably work?  We might want to first-class that (e.g., `NexusRequest::silo_name()`).

You can also set up a reqwest_client / oxide_client to use the external DNS server.  We do that in a few places (look for uses of `CustomDnsResolver`, e.g., in the end-to-end tests and the certificates tests).  In some of those places we're also setting up TLS so it looks worse than it really is.  But it is a bit of boilerplate regardless.  At this point I wonder if it's worth it to do this once and put an oxide_client into ControlPlaneTestContext so that we could write more tests with that.  I'm not sure how authn would work though.

The oxide_client won't work with stuff that expects a redirect.  But a reqwest_client will as long as you set the redirection policy to not follow them.

zephraph:
I'm on board with this approach. I think if absolutely necessary, the query param solution for specifying the silo when none is presented on the DNS sounds like a good plan. If the silo is read via the subdomain _and_ it's specified from the query param we should probably throw a 400 error though.

david-crespo:
I think I’m going to hold off for now and try to work on more important things. I’m satisfied just seeing what the change would have looked like.

PR #3327, opened 2023-06-09
[wicketd] Add periodic check for sled agent bootstrap addrs

jgallagher:
RSS wants to know the bootstrap addresses of the sleds to include; this PR teaches wicketd how to find them. It also updates the wicket TOML config introduced in #3326 to include the sled's bootstrap address (if known) in a comment in the sled list; I'm unsure if this is something we'll want to keep once customers are running RSS, but it seems handy for debugging for now.

Builds on #3326.

### DIFF
--- a/wicketd/src/bootstrap_addrs.rs
+++ b/wicketd/src/bootstrap_addrs.rs
@@ -1,0 +1,175 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use ddm_admin_client::Client as DdmAdminClient;
+use futures::stream::FuturesUnordered;
+use sled_hardware::underlay::BootstrapInterface;
+use sled_hardware::Baseboard;
+use slog::warn;
+use slog::Logger;
+use std::collections::BTreeMap;
+use std::net::Ipv6Addr;
+use std::sync::Arc;
+use std::sync::Mutex;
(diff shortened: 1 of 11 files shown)


wicketd/src/bootstrap_addrs.rs
```diff
@@ -0,0 +1,175 @@
...
+) {
+    // How frequently do we attempt to refresh the set of peers? This does not
+    // count the time it takes to do each refresh, which is potentially
+    // significant if any prefixes reported by DDM are not running responsive
+    // sled-agents, since we'll wait for them to time out.
+    const SLEEP_BETWEEN_REFRESH: Duration = Duration::from_secs(30);
+
+    let ddm_client = make_ddm_admin_client(&log).await;
+
+    // We only share `sleds` with the `BootstrapPeers` that created us, and it
```
andrewjstone, line 59: Nice!

wicketd/src/bootstrap_addrs.rs
```diff
@@ -0,0 +1,175 @@
...
+
+        tokio::time::sleep(SLEEP_BETWEEN_REFRESH).await;
+    }
+}
+
+async fn possible_sled_agent_addrs(
+    ddm_client: &DdmAdminClient,
+    log: &Logger,
+) -> impl Iterator<Item = Ipv6Addr> {
+    // TODO: Should we use `backoff` here instead of a loop/sleep? We're talking
```
andrewjstone, line 133: I don't think we should use a backoff here.

andrewjstone (approved):
Looks great!

PR #3330, opened 2023-06-09
Bump web console

david-crespo:
Changes: https://github.com/oxidecomputer/console/compare/ce4f8f04...baef7af7


david-crespo:
This was for a demo, we don't actually need to do this now.

PR #3331, opened 2023-06-09
document repo structure and goals a bit better

davepacheco:
Recent discussions made me realize that there are a lot of goals we've adopted but never wrote down, nor the rationale for why we've done things this way.  Here's a first cut.

Edit: I'm also updating the README to link to this doc.  I also removed the "Configuration Reference".  This doesn't really make sense here.  There are a bunch of config files in Omicron now.  This one is not particularly special.  In general these are probably best documented in either the rustdoc for the config structs or else the example files.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -51,7 +51,7 @@ You can **run the https://github.com/rust-lang/rust-clippy[Clippy linter]** usin
 
 == Working in Omicron
 
-Omicron is a pretty large repo containing a bunch of related components.  If you just build the whole thing with `cargo build` or `cargo test`, it can take a while, even for incremental builds.  Since most people are only working on a few of these components at a time, it's helpful to be know about Cargo's tools for working with individual packages in a workspace.
+Omicron is a pretty large repo containing a bunch of related components.  (Why?  See xref:docs/repo.adoc[].)  If you just build the whole thing with `cargo build` or `cargo test`, it can take a while, even for incremental builds.  Since most people are only working on a few of these components at a time, it's helpful to be know about Cargo's tools for working with individual packages in a workspace.
 
 NOTE: This section assumes you're already familiar with the prerequisites and environment setup needed to do _any_ work on Omicron.  See xref:docs/how-to-run-simulated.adoc[] or xref:docs/how-to-run.adoc[] for more on that.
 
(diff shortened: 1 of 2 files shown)


steveklabnik (commented):
This is really helpful context, thank you!

jordanhendricks:
Thanks, Dave. It is great to see a lot of this context written down.

Another pain point of building Omicron that might be worth calling out is that even beyond build times, there may be a larger demand for disk and memory resources on a development machine than one might be used to. Like with some other things you listed, this is more related to Rust, but made worse by the fact that the repository is larger.

Another thing I have observed, similar to the myth that one does not need to consider interface versioning in monorepos, is that it can be harder to grok separation of concerns for different packages in the repo. This one is harder for me to define precisely, but I can talk about a couple of examples. Understanding what something like sled-agent is, for example, is straightforward, but for something like  [illlumos-utils](https://github.com/oxidecomputer/omicron/tree/main/illumos-utils), I find less clear what should go there, and why it needs to be in the omicron repository if it's effectively a library for illumos-specific tools. (Again, not to say this is unique to Omicron -- and that specific case sort of reminds me of libmanta at Joyent.) There are examples I've seen in Omicron where code that lives in the same package maybe shouldn't be, but when everything is in the same spot, maybe it's easier to overlook where there should be clear separation of concerns. This class of problem may not be worth calling out in this doc, but is something I have thought about more as I have done more work in Omicron.

davepacheco:
> Another pain point of building Omicron that might be worth calling out is that even beyond build times, there may be a larger demand for disk and memory resources on a development machine than one might be used to. Like with some other things you listed, this is more related to Rust, but made worse by the fact that the repository is larger.

I have added a note about this.  While here, I reread the dependabot bullet point and realized it _is_ affected by having one big repo so I moved that item.

> Another thing I have observed, similar to the myth that one does not need to consider interface versioning in monorepos, is that it can be harder to grok separation of concerns for different packages in the repo. This one is harder for me to define precisely, but I can talk about a couple of examples. Understanding what something like sled-agent is, for example, is straightforward, but for something like [illlumos-utils](https://github.com/oxidecomputer/omicron/tree/main/illumos-utils), I find less clear what should go there, and why it needs to be in the omicron repository if it's effectively a library for illumos-specific tools. (Again, not to say this is unique to Omicron -- and that specific case sort of reminds me of libmanta at Joyent.) There are examples I've seen in Omicron where code that lives in the same package maybe shouldn't be, but when everything is in the same spot, maybe it's easier to overlook where there should be clear separation of concerns. This class of problem may not be worth calling out in this doc, but is something I have thought about more as I have done more work in Omicron.

Yeah... I see what you're saying.  I'm not sure if that's Omicron-specific, or big-repo-specific, or what.  The libmanta case is interesting since that one comes from the everything-is-a-repo world.  I think it's probably true that having a big repo makes it easier to put code in the wrong spot.  But I've also observed quite often that it's much easier to move code to the right spot.  In considering the everything-is-its-own-repo path, one of my bigger fears is that we'll wind up duplicating lots of code because it's so much more annoying to move it where it belongs.  Anyway, I think this is nebulous enough that I'll leave it out of the doc (which it sounds like you're on board with).

david-crespo:
Helpful to me too. I didn’t know you could spin up a DB with the state from a failed test!

PR #3334, opened 2023-06-10
sled agent: propagate Propolis error statuses to Nexus

gjcolombo:
Fix instance-related HTTP entrypoints so that they invoke the correct `From` impl to convert Propolis errors to Dropshot `HttpError`s. Update sled agent's conversion routine to work around dropshot#693 (can't use `for_status` to re- wrap server errors).

Without this fix, all Propolis errors get converted into 500 Internal Server Error on egress from sled agent, which can cause Nexus to conclude incorrectly that instances have failed.

Tested with ad hoc Omicron testing and stress testing.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4751,7 +4751,7 @@ dependencies = [
  "chrono",
  "clap 4.3.3",
  "crucible-agent-client",
- "crucible-client-types",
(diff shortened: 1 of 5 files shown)


smklein (approved):

gjcolombo:
Thanks for reviewing! I'll rebase this and do another test lap before merging.

PR #3346, opened 2023-06-12
[nexus, minor] Use helper for disk create params validation in one more spot

david-crespo:
I noticed we were doing some manual checks in the create from image case that are covered by `validate_disk_create_params`.

### DIFF
--- a/nexus/src/app/disk.rs
+++ b/nexus/src/app/disk.rs
@@ -149,19 +149,10 @@ impl super::Nexus {
                     .fetch()
                     .await?;
 
-                // Reject disks where the block size doesn't evenly divide the
-                // total size
-                if (params.size.to_bytes()
-                    % db_image.block_size.to_bytes() as u64)
-                    != 0
-                {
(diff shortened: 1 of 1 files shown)


nexus/src/app/disk.rs
```diff
@@ -149,19 +149,10 @@ impl super::Nexus {
...
-                // Reject disks where the block size doesn't evenly divide the
-                // total size
-                if (params.size.to_bytes()
-                    % db_image.block_size.to_bytes() as u64)
-                    != 0
-                {
-                    return Err(Error::InvalidValue {
-                        label: String::from("size and block_size"),
-                        message: String::from(
-                            "total size must be a multiple of global image's block size",
```
david-crespo, line 161: "global image" is out of date language

zephraph (approved):
Looks good to me. I did a quick check just to ensure the removed cases were the same as in the validate function. 👍

david-crespo:
Perfect, thanks for checking.

PR #3348, opened 2023-06-12
[nexus] Make Nexus resilient to arbitrary boot other of other services

smklein:
This PR makes Nexus more resilient to arbitrary boot order. Specifically, it allows Nexus to boot successfully even in cases where CRDB, Dendrite, or the Internal DNS services are not yet running at Nexus initialization.

This change is actually relatively small:
- When setting up the DNS resolver and accessing CockroachDB / Dendrite, add a retry (in `nexus/src/context.rs`) during initialization.
- Test this behavior in `nexus/tests/integration_tests/initialization.rs`.

However, adding these tests required a major refactor to the process of constructing a `ControlPlaneTestContext`.

- To test the behavior we want, it must be possible to launch Nexus **before** launching CockroachDB/Dendrite, to validate that "nothing bad happens, Nexus just waits".
- Furthermore, we must be able to run a test where Nexus looks up CockroachDB's address from the internal DNS system (it hasn't launched beforehand, so the address is not known). This has a couple implications...
  - First, the CockroachDB address must be parsed after it is launched, and passed into the internal DNS server as a record.
  - Second, Nexus-under-test must be able to access an internal DNS server that actually runs, and it must be able to parse records from that server. This means, unlike Nexus-in-production, we shouldn't "infer" the address of the DNS server from a subnet - we need to actually launch and manage per-test DNS servers.

These test changes are quite large, but they are worthwhile:

- Customization of boot order is a useful tool generally for testing
- Accessing services through the DNS subsystem allows the system-under-test to more accurately align with the system-under-production. Setting up a DNS server for each test requires some additional metadata tracking, but avoids having distinct "access the address under test" vs "access the resolver under prod" pathways.

### Previously

`test_setup_with_config`, from `nexus/test-utils/src/lib.rs` , was used as the "one-stop-shop" for setting up all state before running tests. This includes launching CockroachDB, **then** launching Nexus. For most tests, this is useful, but for our explicitly-testing-weird-order initialization test, this function is too opinionated to call directly.

### Now 

- We introduce a new structure called `ControlPlaneTestContextBuilder`. This adds a "builder pattern" to the construction of `ControlPlaneTestContext`, using a bunch of `Option`s to store intermediate state.
- We add an `InMemoryServer` to the `dns-server` crate, which the tests can use for a dynamic DNS server.
- We allow Nexus to refer to the DNS servers "by subnet" (previous default) or "by a specific address", which is useful for systems-under-test where the DNS server addresses may be dependent on test-specific ports.
- We populate an `InMemoryServer` DNS server with values created by the `ControlPlaneTestContextBuilder`.
- We cleanup the simulated sled agent a bit, by removing the DNS server from the "sim::server::Server".

Fixes https://github.com/oxidecomputer/omicron/issues/2614


david-crespo:
😵😵😵

smklein:
I'm going to split this into a couple different PRs to hopefully make it a little more readable.

smklein:
This has been basically pulled into:
- https://github.com/oxidecomputer/omicron/pull/3364
- https://github.com/oxidecomputer/omicron/pull/3365
- and I might have a follow-up for some more minor stuff

PR #3351, opened 2023-06-13
[bootstrap-agent] Split long-running rack-{init,reset} operations

jgallagher:
POST'ing or DELETE'ing to `/rack-initialize` now spawns a task to do the work, and a client can `GET /rack-initialize` to check on its status. This also severs the long-running endpoints from cancellation if the client disconnects.

### DIFF
--- a/sled-agent/src/bootstrap/agent/rack_ops.rs
+++ b/sled-agent/src/bootstrap/agent/rack_ops.rs
@@ -10,6 +10,9 @@ use crate::bootstrap::params::RackInitializeRequest;
 use crate::bootstrap::rss_handle::RssHandle;
 use crate::config::SidecarRevision;
 use crate::rack_setup::service::SetupServiceError;
+use schemars::JsonSchema;
+use serde::Deserialize;
+use serde::Serialize;
 use std::mem;
 use std::sync::Arc;
 use std::sync::Mutex;
@@ -17,10 +20,34 @@ use tokio::sync::oneshot;
 use tokio::sync::oneshot::error::TryRecvError;
 use uuid::Uuid;
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+#[derive(
+    Debug,
+    Clone,
+    Copy,
+    PartialEq,
+    Eq,
+    Hash,
+    PartialOrd,
+    Ord,
+    Serialize,
+    Deserialize,
+    JsonSchema,
+)]
 pub struct RackInitId(pub Uuid);
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+#[derive(
+    Debug,
+    Clone,
+    Copy,
+    PartialEq,
+    Eq,
+    Hash,
+    PartialOrd,
+    Ord,
+    Serialize,
+    Deserialize,
+    JsonSchema,
+)]
 pub struct RackResetId(pub Uuid);
 
 #[derive(Debug, Clone, thiserror::Error)]
@@ -66,6 +93,7 @@ impl RssAccess {
 
         match &mut *status {
             RssStatus::Initializing { id, completion } => {
+                let id = *id;
                 // This is our only chance to notice the initialization task has
                 // panicked: if it dropped the sending half of `completion`
                 // without reporting in.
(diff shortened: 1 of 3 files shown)


jgallagher:
This PR does not touch the sled-level init / reset operations; if this looks reasonable I can give them the same treatment in a followup PR.

sled-agent/src/bootstrap/agent/rack_ops.rs
```diff
@@ -0,0 +1,300 @@
...
+use uuid::Uuid;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub struct RackInitId(pub Uuid);
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub struct RackResetId(pub Uuid);
+
+#[derive(Debug, Clone, thiserror::Error)]
+pub enum ConcurrentRssAccess {
```
smklein, line 27: Could we call this something like `RssAccessError`? The `Error` part is more salient than the `Concurrency` part - it's possible a single client making sequential requests triggers one of these.

sled-agent/src/bootstrap/agent/rack_ops.rs
```diff
@@ -0,0 +1,300 @@
...
+    // operations: fill it in or read its current value.
+    status: Arc<Mutex<RssStatus>>,
+}
+
+impl RssAccess {
+    pub(super) fn new(initialized: bool) -> Self {
+        let status = if initialized {
+            RssStatus::InitializedOnStartup
+        } else {
+            RssStatus::ResetOnStartup
```
smklein, line 59: Just to make sure I understand the semantics here:

- If we can't read necessary metadata from the M.2, we pass `initialized = false`
- This sets the state to `ResetOnStartup`, but does not trigger any work
- **We implicitly treat this as if reset has executed correctly**, right?

So it's functionally more like "ResetOnStartup = We think the state is 'Reset' after reboot" rather than "ResetOnStartup = We need to trigger a reset operation after reboot"?

sled-agent/src/bootstrap/http_entrypoints.rs
```diff
@@ -40,6 +44,52 @@ pub(crate) fn api() -> BootstrapApiDescription {
...
+    },
+    ResetPanicked {
+        id: Uuid,
+    },
+}
+
+#[derive(
+    Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize, JsonSchema,
+)]
+pub struct OperationId {
```
smklein, line 89: Nitpick, there's a really small amount of asymmetry here.

It seems like `OperationId` is passed in the result types, but raw `Uuid`s are exposed from `RackOperationStatus`? Should we be doing one of the other everywhere?

smklein (approved):
Thank you for doing this! Agreed that something similar for sled-level init/reset would be useful.

sled-agent/src/bootstrap/agent/rack_ops.rs
jgallagher: Yes, that's right, this `RssStatus` is only setting the initial state of the internal state machine, which only allows initialization to proceed if we're currently in a reset state (either `ResetOnStartup` or `Reset` from an actual reset request).

`ResetOnStartup` does sound too much like an action that is going to happen. A couple ideas:

* We could squish these states into one and keep the boolean: `RssStatus::Startup { initialized: bool }`
* Keep the two statuses, but rename `ResetOnStartup` to... something better. Maybe `NotInitializedOnStartup`?

jgallagher: Another option for renaming `ResetOnStartup`: `RssStatus::Uninitialized`. That is maybe my favorite? We're not initialized but haven't been reset, either.

sled-agent/src/bootstrap/http_entrypoints.rs
jgallagher: Do newtypes place nicely with openapi? For some reason I thought they didn't and so squished the two newtypes down to raw `Uuid`s here for the openapi interface, but I didn't actually try it.

I'll give it a shot. If newtypes _don't_ work nicely with openapi, do you have a preference between:

* Keep it as-is
* Remove the newtypes and use raw `Uuid` everywhere
* Change the newtypes to be openapi-friendly (something like `RackInitId { id: Uuid }`)

sled-agent/src/bootstrap/agent/rack_ops.rs
smklein: I also think a state named `Startup` would convey this information - we just booted, we don't know what operations occurred, but we haven't finished initialization.

My only hesitation with the term `Reset` in the name comes from me incorrectly assuming it describes "this state means a reset operation **finished**"

smklein: Oh, nevermind, I see you used the `InitializedOnStartup` state name, and just calling it `Startup` is a bit ambiguous in comparison

smklein: Calling it `Uninitialized` seems great.

sled-agent/src/bootstrap/http_entrypoints.rs
smklein: I'm fine with either of:

- Keep it as-is (after all, this was a nitpick. I wanted to prod in case there was a clean fix - but it isn't a big deal).
- Remove the newtypes and use raw Uuid everywhere

Excessive wrapping of the UUID type under an explicit variable name seems like more pain than it's worth.

sled-agent/src/bootstrap/agent/rack_ops.rs
jgallagher: Renamed in [5f71e54](https://github.com/oxidecomputer/omicron/pull/3351/commits/5f71e543309f98da4a2bbc72ab83a8dce7823f6b)

sled-agent/src/bootstrap/http_entrypoints.rs
jgallagher: Newtypes are fine! Removed the raw UUIDs in [cfd2ce7](https://github.com/oxidecomputer/omicron/pull/3351/commits/cfd2ce754f38947c56e49d87a60e4963953c0c42)

sled-agent/src/bootstrap/agent/rack_ops.rs
jgallagher: Renamed (and slightly reworked to collapse the two initialized and two uninitialized states down into one) in [cfd2ce7](https://github.com/oxidecomputer/omicron/pull/3351/commits/cfd2ce754f38947c56e49d87a60e4963953c0c42)

PR #3352, opened 2023-06-13
Only update relevant ports when updating firewall rules on sled.

### LINKED ISSUE #3336: VPC firewall rule changes can still wipe out nexus access on external IP sometimes
The symptom is the same as #3046 but would happen after a few consecutive firewall rule updates. The last few sled-agent log lines show the following:
```
..., FirewallRule { direction: In, filters: Filters { hosts: Vni(Vni { inner: 4072733 }), protocol: Proto(TCP), ports: PortList([5201]) }, action: Deny, priority: 10 }, FirewallRule { direction: In, filters: Filters { hosts: Vni(Vni { inner: 4072733 }), protocol: Proto(TCP), ports: PortList([5201]) }, action: Deny, priority: 10 }, FirewallRule { direction: In, filters: Filters { hosts: Vni(Vni { inner: 4072733 }), protocol: Proto(TCP), ports: PortList([5201]) }, action: Deny, priority: 10 }, FirewallRule { direction: In, filters: Filters { hosts: Vni(Vni { inner: 4072733 }), protocol: Proto(TCP), ports: PortList([5201]) }, action: Deny, priority: 10 }, FirewallRule { direction: In, filters: Filters { hosts: Vni(Vni { inner: 4072733 }), protocol: Proto(TCP), ports: PortList([5201]) }, action: Deny, priority: 10 }, FirewallRule { direction: In, filters: Filters { hosts: Vni(Vni { inner: 4072733 }), protocol: Proto(TCP), ports: PortList([5201]) }, action: Deny, priority: 10 }]
23:49:35.503Z INFO SledAgent (PortManager): Setting OPTE firewall rules
    port = "opte0"
    rules = []
23:49:35.504Z INFO SledAgent (PortManager): Setting OPTE firewall rules
    port = "opte2"
    rules = []
23:49:35.504Z INFO SledAgent (PortManager): Setting OPTE firewall rules
    port = "opte6"
    rules = []
23:49:35.504Z INFO SledAgent (PortManager): Setting OPTE firewall rules
    port = "opte1"
    rules = []
23:49:35.504Z INFO SledAgent (dropshot (SledAgent)): request completed
    local_addr = [fd00:1122:3344:107::1]:12345
    method = PUT
    remote_addr = [fd00:1122:3344:107::4]:35052
    req_id = 498dedff-9f9b-4d8e-a072-bffe39adf31d
    response_code = 204
    uri = /vpc/3db8d3fd-5fb7-45d9-a84c-59b05e2fe2ca/firewall/rules
```
The complete sled-agent log of the sled that hosts Nexus can be found at `catacomb.eng.oxide.computer:/data/staff/dogfood/jun-10/oxide-sled-agent_default_gc9.log`

I used the workaround suggested in #3046 to restore access
```
/opt/oxide/opte/bin/opteadm add-fw-rule --action allow -p opte2 --protocol tcp --ports 443  --dir in --priority 1 --hosts any
```

luqmana:
Firewall rules are always pushed per-VPC but we were previously attempting to apply said rules to every OPTE port on a sled. This would cause any ports not tied to the specific VPC to have their firewall rules cleared out. Filtering to just the relevant ports before attempting to apply the rules addresses this.

Fixes #3336

### DIFF
--- a/sled-agent/src/http_entrypoints.rs
+++ b/sled-agent/src/http_entrypoints.rs
@@ -315,10 +315,10 @@ async fn vpc_firewall_rules_put(
     body: TypedBody<VpcFirewallRulesEnsureBody>,
 ) -> Result<HttpResponseUpdatedNoContent, HttpError> {
     let sa = rqctx.context();
-    let vpc_id = path_params.into_inner().vpc_id;
+    let _vpc_id = path_params.into_inner().vpc_id;
     let body_args = body.into_inner();
 
-    sa.firewall_rules_ensure(vpc_id, &body_args.rules[..])
+    sa.firewall_rules_ensure(body_args.vni, &body_args.rules[..])
         .await
         .map_err(Error::from)?;
 
--- a/illumos-utils/src/opte/params.rs
+++ b/illumos-utils/src/opte/params.rs
@@ -13,6 +13,7 @@ use std::net::Ipv6Addr;
 /// Update firewall rules for a VPC
 #[derive(Clone, Debug, Serialize, Deserialize, JsonSchema)]
 pub struct VpcFirewallRulesEnsureBody {
+    pub vni: external::Vni,
     pub rules: Vec<VpcFirewallRule>,
 }
 
--- a/illumos-utils/src/opte/port_manager.rs
+++ b/illumos-utils/src/opte/port_manager.rs
@@ -13,6 +13,7 @@ use crate::opte::Gateway;
 use crate::opte::Port;
 use crate::opte::Vni;
 use ipnetwork::IpNetwork;
+use omicron_common::api::external;
 use omicron_common::api::internal::shared::NetworkInterface;
 use omicron_common::api::internal::shared::NetworkInterfaceKind;
 use omicron_common::api::internal::shared::SourceNatConfig;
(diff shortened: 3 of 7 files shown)


sled-agent/src/http_entrypoints.rs
```diff
@@ -315,10 +315,10 @@ async fn vpc_firewall_rules_put(
     body: TypedBody<VpcFirewallRulesEnsureBody>,
 ) -> Result<HttpResponseUpdatedNoContent, HttpError> {
     let sa = rqctx.context();
-    let vpc_id = path_params.into_inner().vpc_id;
+    let _vpc_id = path_params.into_inner().vpc_id;
```
plotnick, line 318: Just ignoring this seems a little suspicious. Can we maybe eliminate the path parameter or check it against something?

plotnick (approved):
I'm a bit out of the networking loop these days (and don't have a test setup to verify the fix), but I read the linked PRs & issues, and plumbing the VNI through with the rules makes sense and seems like the right fix to me. Moving the required logic into the port manager instead of relaying through the instance manager also seems like a pretty unambiguous win. I think probably @rcgoodfellow or @bnaecker would be in a better position to give it a thorough review, but it's got my :+1: fwiw.

sled-agent/src/http_entrypoints.rs
luqmana: Yea, it wasn't used before either. I hesitated to change as part o this because it does change the shape of the API a bit (though this is just sled-agent's API which AFAIK only nexus calls). I'll open a follow-up issue for possibly changing that.

luqmana: https://github.com/oxidecomputer/omicron/issues/3355

Issue #3090, opened 2023-05-11
Switch API follow up work


zephraph:
Required followup work:
- Switches should be registered during rack initialization
- Switch should include location information
- We should have some mechanism to update their state when they're removed and reinserted. 
- Should be associated with switch-ports once #2645 is merged

_Originally posted at https://github.com/oxidecomputer/omicron/issues/2996#issuecomment-1540765949_

zephraph:
I don't know if work still needs to be done here or if the switch API just needs to be changed altogether. @internet-diglett if you have a chance, it'd be good to review the API for switches that we have currently and figure out if it needs to be changed/removed/updated.

andrewjstone:
There are partially completed APIs (Get /switches) that we should either finish or remove. It's not really a multirack issue, it's an everything issue.

I'm leaning in the remove the API and empty DB table right now, then we can add back the right thing when we need it.

PR #3359, opened 2023-06-15
Associate switch port with switches

zephraph:
Part of the work required for #3090. This PR aims to make switch ports and official child of switches. 

There's a still lot to be done here. The biggest chunk of work is initializing switch ports when switches are created. I've went around on a few different ways to approach this, but it may be easiest to do after (or alongside) #3262.


zephraph:
Dropping this. I suspect some of the switch APIs should just be followed up on (if they haven't been already).

PR #3364, opened 2023-06-16
Test setup as builder, pull DNS server out of simulated sled agent

smklein:
This change is necessary for https://github.com/oxidecomputer/omicron/pull/3365 , which tests arbitrary boot order for Nexus.

### What changes do we need?

- To test the behavior we want, it must be possible to launch Nexus before launching CockroachDB/Dendrite, to validate that "nothing bad happens, Nexus just waits".
- Furthermore, we must be able to run a test where Nexus looks up CockroachDB's address from the internal DNS system (it hasn't launched beforehand, so the address is not known). This has a couple implications...
  - First, the CockroachDB address must be parsed after it is launched, and passed into the internal DNS server as a record.
  - Second, Nexus-under-test must be able to access an internal DNS server that actually runs, and it must be able to parse records from that server. This means, unlike Nexus-in-production, we shouldn't "infer" the address of the DNS server from a subnet - we need to actually launch and manage per-test DNS servers.

### Why are these changes useful?

- Customization of boot order is a useful tool generally for testing
- Accessing services through the DNS subsystem allows the system-under-test to more accurately align with the system-under-production. Setting up a DNS server for each test requires some additional metadata tracking, but avoids having distinct "access the address under test" vs "access the resolver under prod" pathways.

### Previously

- `test_setup_with_config`, from `nexus/test-utils/src/lib.rs`, was used as the "one-stop-shop" for setting up all state before running tests. This includes launching CockroachDB, then launching Nexus. For most tests, this is useful, but for our explicitly-testing-weird-order initialization test, this function is too opinionated to call directly.

### Now

- We introduce a new structure called `ControlPlaneTestContextBuilder`. This adds a "builder pattern" to the construction of ControlPlaneTestContext, using a bunch of Options to store intermediate state.
- We add an `InMemoryServer` to the dns-server crate, which the tests can use for a dynamic DNS server.
- We allow Nexus to refer to the DNS servers "by subnet" (previous default) or "by a specific address", which is useful for systems-under-test where the DNS server addresses may be dependent on test-specific ports.
- We populate an InMemoryServer DNS server with values created by the ControlPlaneTestContextBuilder.
- We cleanup the simulated sled agent a bit, by removing the DNS server from the "sim::server::Server".


common/src/nexus_config.rs
```diff
@@ -115,8 +130,9 @@ pub struct DeploymentConfig {
     pub dropshot_external: ConfigDropshotWithTls,
     /// Dropshot configuration for internal API server.
     pub dropshot_internal: ConfigDropshot,
-    /// Portion of the IP space to be managed by the Rack.
-    pub subnet: Ipv6Subnet<RACK_PREFIX>,
```
smklein, line 119: This was only used for finding DNS servers. We change it to be more appropriately named, and to provide a variant for access-by-address, for tests.

dns-server/src/lib.rs
```diff
@@ -86,3 +91,79 @@ pub async fn start_servers(
...
+///
+/// Intended to be used for testing only.
+pub struct InMemoryServer {
+    /// Server storage dir
+    pub storage_dir: tempfile::TempDir,
+    /// DNS server
+    pub server: dns_server::ServerHandle,
+    /// Dropshot server
+    pub dropshot_server: dropshot::HttpServer<http_server::Context>,
+}
```
smklein, line 105: This was "rolled-by-hand" in the Sled Agent before this PR.

For test / simulated workloads, it seems pretty reasonable to provide this low-touch implementation.

nexus/src/context.rs
```diff
@@ -134,14 +134,24 @@ impl ServerContext {
...
+                )
+                .map_err(|e| format!("Failed to create DNS resolver: {}", e))?
+            }
+            nexus_config::InternalDns::FromAddress { address } => {
+                internal_dns::resolver::Resolver::new_from_addrs(
+                    log.new(o!("component" => "DnsResolver")),
+                    vec![address],
+                )
+                .map_err(|e| format!("Failed to create DNS resolver: {}", e))?
+            }
```
smklein, line 153: This is the chunk that's new - tests can use internal DNS without squatting on the whole subnet.

nexus/test-utils/src/lib.rs
```diff
@@ -138,247 +140,626 @@ pub async fn test_setup<N: NexusServer>(
...
-    initial_cert: Option<Certificate>,
-) -> ControlPlaneTestContext<N> {
-    let start_time = chrono::Utc::now();
-    let logctx = LogContext::new(test_name, &config.pkg.log);
-    let log = &logctx.log;
+struct RackInitRequestBuilder {
+    services: Vec<nexus_types::internal_api::params::ServicePutRequest>,
+    datasets: Vec<nexus_types::internal_api::params::DatasetCreateRequest>,
+    internal_dns_config: internal_dns::DnsConfigBuilder,
+}
```
smklein, line 147: This is basically a builder to support accurate information to DNS and to the RSS handoff request.

sled-agent/src/sim/server.rs
```diff
@@ -32,36 +32,30 @@ use uuid::Uuid;
...
     pub http_server: dropshot::HttpServer<Arc<SledAgent>>,
     /// simulated pantry server
-    pub pantry_server: PantryServer,
-    /// real internal dns server storage dir
-    pub dns_server_storage_dir: tempfile::TempDir,
-    /// real internal dns server
-    pub dns_server: dns_server::dns_server::ServerHandle,
-    /// real internal dns dropshot server
-    pub dns_dropshot_server:
-        dropshot::HttpServer<dns_server::http_server::Context>,
```
smklein, line 47: This is kinda a heavy thing to force on every single Sled Agent server. It should be possible to run simulated sled agents without DNS servers, so this has been decoupled.

sled-agent/src/sim/server.rs
```diff
@@ -177,221 +171,43 @@ impl Server {
...
-                service_id: Uuid::new_v4(),
-                sled_id: config.id,
-                zone_id: Some(Uuid::new_v4()),
-            });
-
-            internal_services_ip_pool_ranges
-                .push(IpRange::V6(Ipv6Range { first: ip, last: ip }));
-        }
-
-        let recovery_silo = NexusTypes::RecoverySiloConfig {
```
smklein, line 332: All this stuff - silo setup, internal DNS, certs, IP ranges, etc, **still exists**, it has just been moved into the `run_standalone_server` function.

This is still useful for running the standalone sled agent binary, but tests likely want finer-grained control over services.

test-utils/src/dev/db.rs
```diff
@@ -1542,7 +1542,7 @@ mod test {
 
         // unexpected contents in initial listen URL (extra param)
         let error = make_pg_config(
-            "postgresql://root@127.0.0.1:45913/foobar?application_name=foo",
+            "postgresql://root@[::1]:45913/foobar?application_name=foo",
```
david-crespo, line 1545: I confirmed that these IP changes do not affect my ability to use `omicron-dev run-all` — why did they change to IPv6, though?

nexus/test-utils/src/lib.rs
```diff
@@ -139,247 +141,626 @@ pub async fn test_setup<N: NexusServer>(
...
+        ControlPlaneTestContextBuilder::<N>::new(test_name, config);
+
+    builder.start_crdb().await;
+    builder.start_clickhouse().await;
+    builder.start_dendrite().await;
+    builder.start_internal_dns().await;
+    builder.start_external_dns().await;
+    builder.start_nexus_internal().await;
+    builder.start_sled(sim_mode).await;
+    builder.start_crucible_pantry().await;
```
david-crespo, line 749: nice for readability to wrap these steps up, even though this is pretty much doing what it was doing before

test-utils/src/dev/db.rs
smklein: - Basically everything on the underlay network runs on IPv6
- Localhost, decidedly, it "not really" part of the underlay network, but just a thing we use under test
- However, a good portion of the wrappers around internal DNS access **assume** we're using IPv6. This is not a hard blocker, it just happens to be a common call where we expect underlay addresses: https://github.com/oxidecomputer/omicron/blob/d70f7ff0c75bdeb9651ebcb4c3a71250c07c75a3/internal-dns/src/resolver.rs#L126
- Switching to using IPv6 under tests means we can use the same DNS calls without modification

The alternative would be:
- Keep the test code IPv4
- Modify the non-test code to cope with both IPv4 and IPv6 addresses on the underlay

smklein: FWIW I **can** go do the alternative, but I figured "make test and prod more aligned on addressing scheme" seemed like a worthwhile goal

david-crespo: Yeah, sounds good

sled-agent/src/sim/server.rs
```diff
@@ -177,221 +171,43 @@ impl Server {
...
+        Ok(Server {
+            config: config.clone(),
+            log: log.clone(),
+            sled_agent,
+            http_server,
+            pantry_server: None,
+        })
+    }
+
+    /// Starts the pantry service and add it to the DNS config builder
```
davepacheco, line 183: Is this comment still accurate?  I don't see where we add it to the DNS config builder.

nexus/test-utils/src/lib.rs
```diff
@@ -139,247 +141,626 @@ pub async fn test_setup<N: NexusServer>(
...
+            .internal_dns_config
+            .host_zone(dataset_id, *address.ip())
+            .expect("Failed to set up DNS for {kind}");
+        self.internal_dns_config
+            .service_backend_zone(service_name, &zone, address.port())
+            .expect("Failed to set up DNS for {kind}");
+    }
+}
+
+pub struct ControlPlaneTestContextBuilder<'a, N: NexusServer> {
```
davepacheco, line 212: This cleanup is long overdue.  Thanks for taking it on!

The only real note I have here is that it's not that clear to me what use cases are supposed to be supported here and I'm afraid people it will become a footgun.  We obviously have the use case via `build()` that builds a full `ControlPlaneTestContext` out of what's here.  I gather the Builder itself exists to support more flexible constructions, like doing things in a different order or not starting some components.  Are you supposed to do this by using the `pub` fields of the Builder and never calling `build()`?  That seems a little confusing and error-prone.   But that may be practical if we really only expect this to be used in a few places!  Relatedly, it seems like you could misuse this in a lot of ways -- like calling `start_internal_nexus()` more than once.

I'm not really sure what to suggest instead.  Thinking out loud: hopefully, each of these components (DNS server, sled agent, nexus, etc.) already has a function that makes it easy to instantiate it in isolation.  The test setup is really providing the glue that (1) actually does it all for you in the common case; (2) handles the "glue", like DNS and rack init request setup, that is otherwise tedious and error-prone; and (3) makes sure that each thing is pointed at the copies of its dependencies that have been set up already.  That suggests maybe a `ControlPlaneBuilder` (which isn't testing-specific, really) with functions like `start_internal_dns_server()` and `start_nexus()` and `nexus_handoff()` -- probably very similar to what's here.  But it wouldn't have to use an explicit Builder that suggests (wrongly) that you're supposed to call `build()` or that you can't use the stuff you've instantiated without calling `build()`.  We could use typestates to ensure that you can't, say, do a handoff if you haven't started Nexus, and you also can't start more stuff after that.  On the other hand, using typestates this way might get in the way of some of the stuff we actually want to test.  So I'm less sure about that.

Take that as food for thought...the only real issue I have is the footgun fear, that people might mistake this for "the way to test anything other than what ControlPlaneTestContext gives you", and then winding up with spaghetti code dealing with a mess of `Option`s.  That's not a big deal if we really only expect this to be used in, like, the one place where you're using it in #3365.

dns-server/src/lib.rs
smklein: I intended for this to be a really thin wrapper around the production DNS server - the only difference is the `dns_server::storage::Store` argument, which is created from this `tempfile::TempDir`.

If you compare what I'm doing here with tests, I intended for it to be the same. For example, see:

https://github.com/oxidecomputer/omicron/blob/9b85b6190cf58e7860a04471f506fbaecabd219f/nexus/src/app/background/init.rs#L227-L256

- This creates a `TempDir::new()`
- It uses it as a `dns_server::storage::Store`
- It calls `start_servers`, and creates a new dropshot server

`InMemoryServer` is supposed to just be a convenience wrapper around this same thing.

nexus/src/context.rs
smklein: Before this PR, we passed an argument as a part of Nexus' `DeploymentConfig`:

https://github.com/oxidecomputer/omicron/blob/9b85b6190cf58e7860a04471f506fbaecabd219f/common/src/nexus_config.rs#L118-L119

This "subnet" argument was then used to determine the address of DNS server(s):

https://github.com/oxidecomputer/omicron/blob/9b85b6190cf58e7860a04471f506fbaecabd219f/nexus/src/context.rs#L136-L144

This is how I expect we'll at least bootstrap the discovery of internal DNS servers on the underlay.

However, this means that:
- Given an input subnet...
- ... Nexus assumed it could infer the DNS server(s) from that subnet, by knowing that they'd reliably be at particular reserved addresses / ports.
- ... But in tests, this means we'd need full control over everything running in a subnet. This breaks down if a test wants to try placing a internal DNS server on "localhost" (for example, letting the host OS decide the port) instead of a reserved address within a subnet.

nexus/examples/config.toml
```diff
@@ -51,8 +51,9 @@ request_body_max_bytes = 1048576
 bind_address = "[::1]:12221"
 request_body_max_bytes = 1048576
 
-[deployment.subnet]
-net = "fd00:1122:3344:0100::/56"
+[deployment.internal_dns]
+type = "from_subnet"
+subnet.net = "fd00:1122:3344:0100::/56"
```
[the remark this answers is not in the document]
smklein: I'll update the documentation - I believe if the how-to-run-simulated instructions are followed, and internal DNS server is already being spun up, and this value is overwritten: 65e047323ce4388e30ee87e9972ae6fa86acdc32

nexus/src/context.rs
```diff
@@ -134,14 +134,24 @@ impl ServerContext {
...
-        info!(log, "Setting up resolver on subnet: {:?}", az_subnet);
-        let resolver = internal_dns::resolver::Resolver::new_from_subnet(
-            log.new(o!("component" => "DnsResolver")),
-            az_subnet,
-        )
-        .map_err(|e| format!("Failed to create DNS resolver: {}", e))?;
+        let resolver = match config.deployment.internal_dns {
+            nexus_config::InternalDns::FromSubnet { subnet } => {
+                let az_subnet = Ipv6Subnet::<AZ_PREFIX>::new(subnet.net().ip());
+                info!(log, "Setting up resolver on subnet: {:?}", az_subnet);
```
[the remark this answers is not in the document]
smklein: Done: 4046281baa2e950c98a1b1fa7101a04439fb33bb

nexus/src/context.rs
```diff
@@ -134,14 +134,24 @@ impl ServerContext {
...
+                internal_dns::resolver::Resolver::new_from_subnet(
+                    log.new(o!("component" => "DnsResolver")),
+                    az_subnet,
+                )
+                .map_err(|e| format!("Failed to create DNS resolver: {}", e))?
+            }
+            nexus_config::InternalDns::FromAddress { address } => {
+                internal_dns::resolver::Resolver::new_from_addrs(
```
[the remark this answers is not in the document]
smklein: Done: 93bcca6ee1495c1eaae3feedb0c7fc7ae8cf4884

sled-agent/src/sim/server.rs
smklein: Good catch, the DNS config was separated from this invocation. Updated in e340a0439

nexus/test-utils/src/lib.rs
smklein: Given that this is test code, I significantly prefer to rely on assertions and calls to `.expect`, rather than typestates. My main reason for this disposition is: I struggle to predict all the interleavings someone might want to use when setting up a test.

Maybe we want to spin up Nexus without a sled agent? Maybe we want to spin up Nexus with ten sled agents? Maybe we want to spin up *almost* everything, but take a particular dropshot server offline right before "nexus external" is created?

Regardless of what we're trying to test, my goals were:
- de-duplicate the (massive!) amount of code necessary for setup
- make it easy to do the right thing
- make it *possible* to test weirder things

That's part of why I used so many `Option` types - the builder lets the caller "construct whatever state they want", and, if really desired, access those types via `pub` fields.

I agree with you that this is an API which "could let a caller do the wrong thing", but I also struggle with that, because we also **want** to test weird behaviors during tests, whenever possible.

If I'm overlooking an alternative, I'm totally open to suggestions! But providing a "single-use, hard-to-get-wrong API for test setup" that sacrifices flexibility is basically what we had before.

dns-server/src/lib.rs
davepacheco: Sorry about that!  I didn't look at the implementation closely enough and was confused by the name.  That sounds perfect, but can we not call it "in-memory"?  If you're trying to convey that it goes away, maybe "Transient"?

nexus/examples/config.toml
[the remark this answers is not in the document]
davepacheco: I think you have updated the "quick start" pieces in that commit (thanks!) but if somebody follows the "Running the pieces by hand" section, won't they be missing the internal DNS server?

dns-server/src/lib.rs
```diff
@@ -86,3 +91,80 @@ pub async fn start_servers(
...
 }
+
+/// An in-memory DNS server running on localhost.
+///
+/// Intended to be used for testing only.
+pub struct InMemoryServer {
+    /// Server storage dir
+    pub storage_dir: tempfile::TempDir,
+    /// DNS server
+    pub server: dns_server::ServerHandle,
```
davepacheco, line 102: suggestion: `dns_server`?  There are so many "server"s in these structs I find it helpful to be explicit that this is the DNS [protocol] server.

nexus/examples/config.toml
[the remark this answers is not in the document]
smklein: Starting the simulated sled agent still starts a crucible pantry and an internal DNS server, as it did before this PR. It certainly isn't getting *populated* with all the records we want, but I think that's the same as before this PR -- it only adds a record for the Crucible Pantry.

I think it would certainly be an improvement to start the Internal DNS server **separately** from the simulated sled agent, and have more of the configs point to it (rather than using hard-coded addresses, as they currently do!), but I don't think that would be a small change.

(Regardless, thanks for pointing this out - it prompted me to run through all the "simulated sled agent" manually, where I patched an unrelated bug.)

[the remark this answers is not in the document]
davepacheco: But this configuration file is also used directly in the how-to-run-simulated instructions.  It still points Nexus at a specific IP address which is not right for the user's configuration.  Right?  It seems like there's a missing step in these instructions which is to customize this file somehow?  (and the next question is whether there's any way we could avoid that -- e.g., maybe running the simulated sled agent by hand should use a fixed IP and port for the DNS server so that we can put that directly in this file?  That's what we do for the other components there.)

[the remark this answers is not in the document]
smklein: I tried to tackle this in https://github.com/oxidecomputer/omicron/pull/3364/commits/e745c76b7387bd904fd2788ca3ec8060150f9fb7 - that lets callers customize the internal DNS, DNS (not http) address which can then be propagated to Nexus.

This means that the value in the Nexus config file will be respected, as long as the same address is passed to the simulated sled agent to start a DNS server listening on it too.

davepacheco (approved):
Thanks!  It'd be good to know that the updated "how-to-run" instructions work as-is.

smklein:
> Thanks! It'd be good to know that the updated "how-to-run" instructions work as-is.

Just manually ran it, seems to work: Nexus comes up (and can respond to requests) using the provided docs.

PR #3365, opened 2023-06-16
[nexus] Resilience to arbitrary boot order

smklein:
- Ensures that Nexus can boot safely, even if CockroachDB, Dendrite, or Internal DNS boot lazily
- Adds tests for this behavior (relying on https://github.com/oxidecomputer/omicron/pull/3364)

Fixes https://github.com/oxidecomputer/omicron/issues/2614


nexus/src/context.rs
```diff
@@ -158,14 +158,27 @@ impl ServerContext {
...
+                                log,
+                                "Failed to lookup cockroach address: {e}"
+                            );
+                            tokio::time::sleep(std::time::Duration::from_secs(
+                                1,
+                            ))
+                            .await;
+                        }
+                    }
+                };
```
smklein, line 178: I'd kinda like to land somewhere that:

- Whenever we want to access a service, we *could* look up the address again through DNS
- Whenever we want to access DNS, we *could* access any of the DNS server
- If we time-out or error accessing anything, we can customize a back-off policy

... which sorta seems like a Rust version of cueball. Maybe that's a natural follow-up from here?

david-crespo (approved):
LGTM

nexus/src/app/mod.rs
```diff
@@ -178,13 +178,22 @@ impl Nexus {
...
+                    .lookup_socket_v6(ServiceName::Dendrite)
+                    .await
+                    .map_err(|e| format!("Cannot access Dendrite address: {e}"))
+                {
+                    Ok(addr) => break (addr.ip().to_string(), addr.port()),
+                    Err(e) => {
+                        warn!(log, "Failed to access Dendrite address: {e}");
+                        tokio::time::sleep(std::time::Duration::from_secs(1))
+                            .await;
+                    }
```
davepacheco, line 194: This seems like something we probably want to revisit.  (Nexus can't come up _at all_ without Dendrite?  It seems like it'd be better to just construct a lazily-resolving DPD client and let the rest of Nexus come up.)

But this PR is a clear improvement from what's there now.

nexus/src/context.rs
davepacheco: Yes, I think so.  Related to that: we want to proactively resolve stuff _before_ we need it so that DNS resolution latency/availability does not affect latency/availability of the stuff that needs it so much.

Maybe file an issue and add a reference both here and in the Dendrite loop (so that people don't, like, keep copying and pasting this pattern)?

PR #3366, opened 2023-06-16
Bump web console

david-crespo:
Changes: https://github.com/oxidecomputer/console/compare/ce4f8f04...fbace84e

- Confirmation popup on delete actions, handle errors instead of failing silently
- Promote image UI on silo images page
- Validate boot disk size against image size on instance create
- Fix `created` column header on subnets table


david-crespo:
Whoops, bumped to wrong commit.

PR #3363, opened 2023-06-16
ci: generate a full TUF repo with hubris artifacts

iliana:
This, hopefully, updates the tuf-repo Buildomat job to pull the following artifacts:

**dvt-dock** (via git clone):
- staging/build-gimlet-image-a-cert-dev.zip
- staging/build-gimlet-image-b-cert-dev.zip
- staging/build-psc-image-a-cert-dev.zip
- staging/build-psc-image-b-cert-dev.zip
- staging/build-sidecar-image-a-cert-dev.zip
- staging/build-sidecar-image-b-cert-dev.zip

**hubris** (via GHA artifacts):
- dist-ubuntu-latest-gimlet-c.zip
- dist-ubuntu-latest-psc-b.zip
- dist-ubuntu-latest-sidecar-b.zip

and adds them to the now-renamed repo-dogfood.zip. In theory we would generate additional repo zips for different hardware revisions, but also we're going to someday have to figure out how to have multiple revisions of Hubris images in the same TUF repo...

(This does what the relish repo was originally intended to do, but perhaps in a more obvious way.)

PR #3369, opened 2023-06-16
ci: --netrc + [[publish]] in tuf-repo.sh

iliana:
This, hopefully, fixes two issues:

- e7b9526: I didn't update the `[[publish]]` keys to refer to the new paths.
- e87169a: The curl binary does not enable ~/.netrc support by default; [Buildomat drops in a ~/.netrc file with a GitHub token](https://github.com/oxidecomputer/buildomat/blob/f7683e113f0d5b1745ded9b9ba83ca02a3431f64/github/server/src/variety/basic.rs#L739-L777) so jobs can clone private repos. Most EC2 public IPs are constantly exceeding anonymous rate limits, so it was sheer luck that it worked in the job for #3363.

### DIFF
--- a/.github/buildomat/jobs/tuf-repo.sh
+++ b/.github/buildomat/jobs/tuf-repo.sh
@@ -25,18 +25,18 @@
 #:
 #: [[publish]]
 #: series = "tuf-repo"
-#: name = "repo.zip.parta"
-#: from_output = "/work/repo.zip.parta"
+#: name = "repo-dogfood.zip.parta"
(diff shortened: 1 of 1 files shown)


luqmana (approved):
Ooof yea, I'd even asked about the `[publish]`! Clearly i should've expanded more of the file. Thanks!

jclulow:
(NB: silent failure of publish directives is oxidecomputer/buildomat#10)

PR #3370, opened 2023-06-16
[internal DNS] Implement reqwest::dns::Resolve for internal DNS resolver, use it

smklein:
## Meta

Original title of this PR: "[sled agent] Clarify LazyNexusClient constructor arguments, cache address"
After feedback, I'm opting to use the internal DNS resolver in the client which is passed to Progenitor, which changes the scope of this PR a bit.

# Internal DNS

- Implements [reqwest::dns::Resolve](https://docs.rs/reqwest/0.11.18/reqwest/dns/trait.Resolve.html) for the internal DNS server. This enables [usage within the reqwest client builder](https://docs.rs/reqwest/0.11.18/reqwest/struct.ClientBuilder.html#method.dns_resolver).
- Add a couple tests verifying that the DNS resolver can successfully be integrated with Progenitor bindings.

# Sled Agent

- Use the Internal DNS resolver during the construction of the Nexus client. This *mostly* removes the need for the `LazyNexusClient` structure which was previously used to lazily perform DNS resolution.
- Replace `LazyNexusClient` with `NexusClientWithResolver`. I would *prefer* to just have it be a progenitor-created `NexusClient`, but there are some spots where we're still using the resolver to look up and pass IP addresses into services which are created by the Sled Agent. This is worth fixing, but I'm considering it out-of-scope for this PR, because it'll involve some cross-repo changes.


sled-agent/src/nexus.rs
```diff
@@ -39,27 +39,57 @@ pub struct LazyNexusClient {
 }
 
 impl LazyNexusClient {
-    pub fn new(log: Logger, addr: Ipv6Addr) -> Result<Self, ResolveError> {
+    pub fn new_from_subnet(
```
davepacheco, line 42: Can we document what this does?  (I find it very non-obvious what it would mean to construct a Nexus client from an "ip" that we refer to as a subnet.)  I gather what we're really saying is: given this IP address, which could be _any_ IP in the AZ, use a hardcoded policy to find the DNS servers ("the AZ subnet is a /56 and the first N addresses are DNS servers") and use those to find Nexus.

sled-agent/src/nexus.rs
```diff
@@ -39,27 +39,57 @@ pub struct LazyNexusClient {
...
+    }
+
+    pub async fn get_addr(&self) -> Result<SocketAddrV6, ResolveError> {
+        if let Some(addr) = self.inner.address.lock().unwrap().as_ref() {
+            return Ok(*addr);
+        }
+        let addr =
+            self.inner.resolver.lookup_socket_v6(ServiceName::Nexus).await?;
+        *self.inner.address.lock().unwrap() = Some(addr);
+        Ok(addr)
```
davepacheco, line 85: If I'm reading this right, this has the behavior that if several clients come into `get_addr()` concurrently, they will all try to look up the socket, and any that succeed will wind up storing whatever they find in here.  The last one will "win" forever -- we'll never look it up again.  That doesn't seem great.  We'll wind up hitting the DNS servers harder than we need to with a bunch of concurrent requests (which isn't a big deal right now) and we won't ever learn about any change in DNS.

I haven't taken a look at the surrounding code yet.  But are we not using a Progenitor-generated Oxide client?  I know we can give reqwest a DNS Resolver.  Is that not sufficient here?

sled-agent/src/nexus.rs
```diff
@@ -39,27 +39,57 @@ pub struct LazyNexusClient {
...
     }
 
     pub async fn get(&self) -> Result<NexusClient, ResolveError> {
-        let address = self.get_ip().await?;
+        let address = self.get_addr().await?;
 
         Ok(NexusClient::new(
-            &format!("http://[{}]:{}", address, NEXUS_INTERNAL_PORT),
+            &format!("http://{address}"),
```
davepacheco, line 92: This is not a problem right now, but if we ever do internal TLS then this will need to be a DNS name here.  This would be another reason to (eventually) go with the approach of handing reqwest a DNS resolver rather than resolving it ourselves here.

sled-agent/src/nexus.rs
```diff
@@ -39,27 +39,57 @@ pub struct LazyNexusClient {
...
                 )?,
+                address: Mutex::new(None),
+            }),
+        })
+    }
+
+    pub fn new_from_dns(
+        log: Logger,
+        dns_addrs: Vec<SocketAddr>,
+    ) -> Result<Self, ResolveError> {
```
davepacheco, line 61: I wonder if it would better separate the concerns to have the caller provide a `Resolver` directly here.

davepacheco (commented):
I like the renaming of `new` to `new_from_subnet` for clarity.  I'm not sure about `LazyNexusClient` overall and I'm not sure the caching is an improvement (see the separate comment).

davepacheco:
Sorry, I hit "comment" too soon.  I wonder whether `LazyNexusClient::new()` should just take a `Resolver` and let the caller determine how to construct that Resolver.  The caller is already making the decisio and we already have constructors on `Resolver` to do this.

sled-agent/src/nexus.rs
smklein: > I haven't taken a look at the surrounding code yet. But are we not using a Progenitor-generated Oxide client? I know we can give reqwest a DNS Resolver. Is that not sufficient here?

I've spent a while digging into this, implementing https://docs.rs/reqwest/latest/reqwest/dns/trait.Resolve.html for our `internal_dns` Resolver type, and I've come to the following conclusions:

- **Using the DNS resolver with progenitor / reqwest ignores the supplied port**. For reference, see: https://docs.rs/reqwest/0.11.18/reqwest/struct.ClientBuilder.html#method.resolve , but I also tested this manually. This sucks for testing, since we will need to know / commit to the port when the client is being constructed, and manually stick it in the URL. Granted, the ports **will** be stable for services in prod, so maybe we figure out a different workaround here when the system's under test.
- **Using the DNS resolver with progenitor / reqwest appears to query the DNS server on every request**. I tested this manually, but observed that there is no caching happening, and each individual client -> server request also hits the resolver. Perhaps this is fine, but I think it means that the responsibility for caching will need to live in the internal DNS resolver itself.

However, that being said, it *does* seem like using the [dns_resolver](https://docs.rs/reqwest/0.11.18/reqwest/struct.ClientBuilder.html#method.dns_resolver) method when constructing a Client, before it's passed to a progenitor client, does allow us to cope with the IP address of services shifting over time (and also, the DNS servers going up and down).

sled-agent/src/nexus.rs
smklein: I'm deprecating `LazyNexusClient`, per your suggestion in a different comment, but I can clean up the documentation in `internal-dns/src/resolver.rs`, where this is also referenced.

smklein: Done in fb74e5c9e

sled-agent/src/nexus.rs
smklein: To wrap this up: I've changed the name of this PR, and shifted the focus to "deprecate LazyNexusClient in favor of supplying the DNS resolver to the progenitor client (through the reqwest::Client)".

sled-agent/src/nexus.rs
smklein: Yeah, shifted to supplying reqwest a DNS resolver directly, so this *should* be resolved now. I do still need to specify the port though.

sled-agent/src/nexus.rs
smklein: Done (indirectly), see other comments about how I'm passing the resolver.

smklein:
Thanks for the feedback - I didn't realize I even *could* pass the DNS resolver directly through to the reqwest client, but I can clearly see the advantages. I made that change here, and it seems like we should be using the resolver like this generally throughout Omicron.

internal-dns/src/resolver.rs
```diff
@@ -515,4 +588,219 @@ mod test {
...
+        let mut dns_config = DnsConfigBuilder::new();
+        let zone = dns_config.host_zone(Uuid::new_v4(), ip).unwrap();
+        dns_config
+            .service_backend_zone(ServiceName::Nexus, &zone, port)
+            .unwrap();
+        let dns_config = dns_config.build();
+        dns_server2.update(&dns_config).await.unwrap();
+
+        // Confirm that the progenitor client can access this record,
+        // even though the first DNS server doesn't know about it.
```
davepacheco, line 797: I'm not sure that this should necessarily work.  I think all DNS servers are equally authoritative.  If this didn't work, I don't think anything would be broken.  (A related test would be if you _shut down_ the first server -- then I think it should work.)

internal-dns/src/names.rs
```diff
@@ -84,4 +84,9 @@ impl ServiceName {
             }
         }
     }
+
+    /// Returns the full DNS name of this service
+    pub fn srv_name(&self) -> String {
```
davepacheco, line 89:
```suggestion
    pub(crate) fn srv_name(&self) -> String {
```

suggesting this just to make it harder for people to accidentally use the DNS name for something.

sled-agent/src/nexus.rs
```diff
@@ -18,50 +18,54 @@ use std::sync::Arc;
...
-    inner: Arc<Inner>,
+pub struct NexusClientWithResolver {
+    client: NexusClient,
+    resolver: Arc<Resolver>,
 }
 
-impl LazyNexusClient {
-    pub fn new(log: Logger, addr: Ipv6Addr) -> Result<Self, ResolveError> {
+impl NexusClientWithResolver {
+    pub fn new(
```
davepacheco, line 31: I wonder if we could name this and/or document it more clearly.  It's weird that this is "a thin wrapper" over the progenitor-generated NexusClient but its constructor arguments are so different, and it's surprising too that it would take a sled agent address.

Why not have the client be responsible for creating the Resolver so that we don't leak the policy choice ("the DNS resolvers are these fixed addresses relative to the sled IP") into this abstraction?

internal-dns/src/resolver.rs
```diff
@@ -92,6 +112,14 @@ impl Resolver {
             .collect()
     }
 
+    /// Create a DNS resolver using the implied DNS servers within this subnet.
+    ///
+    /// The addresses of the DNS servers are inferred within an Availability
+    /// Zone's subnet: normally, each rack within an AZ (/48) gets a unique
+    /// subnet (/56), but the FIRST /56 is reserved for internal DNS servers.
```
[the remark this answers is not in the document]
smklein: They are global to the AZ, not the rack.

Recall:
- The AZ subnet is a /48
- The rack subnet is a /56
- So if we wanted to linearly allocate rack subnets from within an AZ, we could assign "the first /56 in the /48" to the first rack, the second one to the second rack, etc.

This is just saying: "the **first** rack subnet within a /48 **should not** be provisioned to a rack -- it's dedicated to internal DNS servers.

This means that any rack within an arbitrary /48 subnet can:
- Use it's own address to find the /56 of the rack
- Look up the first (reserved) /48 within the /56
- Find the internal DNS servers for the AZ within that reserved /48 subnet

internal-dns/src/resolver.rs
```diff
@@ -21,22 +22,42 @@ pub enum ResolveError {
     #[error(transparent)]
     Resolve(#[from] trust_dns_resolver::error::ResolveError),
 
-    #[error("Record not found for SRV key: {}", .0.dns_name())]
-    NotFound(crate::ServiceName),
+    #[error("Record not found for SRV key: {0}")]
+    NotFound(String),
 
     #[error("Record not found for {0}")]
     NotFoundByString(String),
```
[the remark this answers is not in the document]
smklein: Yeah, I'll change this back.

internal-dns/src/names.rs
smklein: I was planning on letting callers construct progenitor clients using this name as part of the URL during construction -- would that be wrong?

For example, see: `sled-agent/src/nexus.rs` in this PR -- I use this SRV name when making the progenitor client.

internal-dns/src/resolver.rs
```diff
@@ -127,20 +155,21 @@ impl Resolver {
...
+        let name = srv.srv_name();
+        debug!(self.inner.log, "lookup_socket_v6 srv"; "dns_name" => &name);
+        let response =
+            self.inner.resolver.lookup(&name, RecordType::SRV).await?;
 
         let rdata = response
             .iter()
             .next()
-            .ok_or_else(|| ResolveError::NotFound(srv))?;
+            .ok_or_else(|| ResolveError::NotFound(name.to_string()))?;
```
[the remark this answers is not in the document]
smklein: I'll fix this to be the more strongly-typed `ServiceName` -- agreed that we should have a preference for that over strings, when it's easy.

internal-dns/src/resolver.rs
smklein: I'll shut down the first server first -- the test still passes with that tweak.

sled-agent/src/nexus.rs
smklein: I'd be happy with that choice -- updated to pass the `Resolver` as an argument. Agreed that it's better to not smear that policy decision across API boundaries.

PR #3372, opened 2023-06-17
plumbing for techport router advertisements via tfportd

### LINKED ISSUE #3229: Configure router advertisements on technician port.
The original plan for accessing the technician port was to use IPv6 link local addresses. However, there is a question of how the link local address for the technician port is discovered. If the NDP implementations involved are go-getters, then maybe the technician port link-local address will be discoverable using `ndp -na` (or the OS equivalent of whatever is plugged into the tech port). Experience has shown that not all NDP implementations will resolve this automagically and so the operator is left to send some coaxing packets to get NDP to kick in. Not great.

What we can do is take matters into our own hands in the switch zone through NDP router advertisements. Consider the following view into a live switch zone.

```
root@oxz_switch:~# ipadm | grep tech
techport0/ll      addrconf ok           fe80::aa40:25ff:fe05:402%techport0/10
techport1/ll      addrconf ok           fe80::aa40:25ff:fe05:402%techport1/10
```

Ok so we've got two technician ports each with an IPv6 link local address.

```
root@oxz_switch:~# /opt/oxide/dendrite/bin/swadm addr list | grep 402
int0/0                  fe80::aa40:25ff:fe05:402
```
The technician port IPv6 addresses are plumbed onto the ASIC, this is necessary for packets destined to this address to reach the switch zone through the ASIC.

```
root@oxz_switch:~# ipadm | grep bootstrap
oxBootstrap0/bootstrap6 static ok       fdb0:a840:2504:352::2/64
```

We also have a bootstrap address assigned to the switch zone. This is sitting on a gigantic /64 address space. So we can grab another 2 addresses from that prefix, assign them to the technician ports, and advertise them with NDP. This would look like the following 

```
ipadm create-addr -t -T static fdb0:a840:2504:352::3/64 techport0/v6
ipadm create-addr -t -T static fdb0:a840:2504:352::4/64 techport1/v6
swadm addr add int0/0 fdb0:a840:2504:352::3
swadm addr add int0/0 fdb0:a840:2504:352::4
```

And in `/etc/inet/ndpd.conf` in the switch zone

```
ifdefault StatelessAddrConf off
prefixdefault AdvOnLinkFlag on AdvAutonomousFlag on

if techport0 AdvSendAdvertisements on
techport0 fdb0:a840:2504:352::/64 techport0

if techport1 AdvSendAdvertisements on
techport1 fdb0:a840:2504:352::/64 techport1
```

Now, when the user plugs their laptop/server/workstation/toaster into the technician port and enables IPv6 on the interface they're plugged in on - they'll begin receiving router advertisements with the autonomous autoconfiguration flag set. This will queue their operating system to assign a unique SLAAC address on the bootstrap prefix of the scrimlet. I've mocked this up in a falcon environment and it looks like this from the technician laptop side.

```
root@violin:~# ipadm
vioif0/ll         addrconf ok           fe80::8:20ff:fe65:54ab/10
vioif0/ll         addrconf ok           fdb0:a840:2504:352:8:20ff:fe65:54ab/64
```
```
root@violin:~# netstat -nr -f inet6

Routing Table: IPv6
  Destination/Mask            Gateway                   Flags Ref   Use    If
--------------------------- --------------------------- ----- --- ------- -----
::1                         ::1                         UH      2     442 lo0
fdb0:a840:2504:352::/64     fdb0:a840:2504:352:8:20ff:fe65:54ab U       2       0 vioif0
fe80::/10                   fe80::8:20ff:fe65:54ab      U       2       0 vioif0
default                     fe80::8:20ff:fe8a:d94       UG      1       0 vioif0
```
```
root@violin:~# ping fdb0:a840:2504:352::3
fdb0:a840:2504:352::3 is alive
```

So if we reserve some address suffix on the bootstrap prefix of the scrimlet (in this case I used `::3`), this will provide a network path for the technician to set things up.

rcgoodfellow:
- Plumbing for oxidecomputer/dendrite#401.
- Fixes #3229
- Fixes oxidecomputer/minimum-upgradable-product#82


rcgoodfellow:
Update:

The original idea of this PR was to advertise the bootstrap prefix of the scrimlet through the technician port. However, there are some concerns with just extending the bootstrap network out the technician port. This PR is now a slight modification on the original intent that still gives us a unique prefix per scrimlet and puts technician ports on their own prefixes.

We start from the bootstrap prefix and change the leading 16 bits from `fdb0` to `fdb1`. We keep the subsequent 48 bits the same to maintain a unique prefix per scrimlet. Then each scrimlet is assigned the address `fdb1:<48-bit-id>::1` in its loopback interface, which can be used to access `wicketd` et al. through the technician port.

rcgoodfellow:
When this lands, I think It would be good to plug a Linux-based system into a technician port to ensure things work as expected.

jgallagher (approved):
Confirmed this works on the `madrid` environment as well:

```
john@jeeves ~ $ ipadm | grep madrid_tp
madrid_tp0/v6     addrconf ok           fe80::3617:ebff:fed4:92a6/10
madrid_tp0/v6     addrconf ok           fdb2:a840:2504:355:3617:ebff:fed4:92a6/64
john@jeeves ~ $ ssh wicket@fdb2:a840:2504:355::1
... success ...
```

Thanks!

rcgoodfellow:
Awesome. Thank you so much for taking this for a spin end to end!

PR #3374, opened 2023-06-18
Clarify the CTE around region allocation

smklein:
Some minor docs improvements, as I dug through this codebase recently:

- I wanted to clarify, "Does this CTE only provision Crucible datasets to U.2s, and not M.2s?". The answer to this question is: "yes, but implicitly decided when we provision the datasets themselves". Added a comment to clarify this during region selection.
- Additionally, when referencing a virtual table from an earlier `WITH candidate_zpools AS ...` arm of the CTE, make that reference explicit. This is a no-op change, but it clarifies the dependencies within the CTE.

### DIFF
--- a/nexus/db-queries/src/db/queries/region_allocation.rs
+++ b/nexus/db-queries/src/db/queries/region_allocation.rs
@@ -74,6 +74,11 @@ impl OldRegions {
 }
 
 /// A subquery to find datasets which could be used for provisioning regions.
+///
+/// We only consider datasets which are already allocated as "Crucible".
+/// This implicitly distinguishes between "M.2s" and "U.2s" -- Nexus need
+/// to determine during dataset provisioning which devices should be considered
+/// for usage as Crucible storage.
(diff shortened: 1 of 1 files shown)


nexus/db-queries/src/db/queries/region_allocation.rs
```diff
@@ -74,6 +74,11 @@ impl OldRegions {
 }
 
 /// A subquery to find datasets which could be used for provisioning regions.
+///
+/// We only consider datasets which are already allocated as "Crucible".
+/// This implicitly distinguishes between "M.2s" and "U.2s" -- Nexus need
```
bnaecker, line 79: Spelling nit: "Nexus needs"

bnaecker (approved):

nexus/db-queries/src/db/queries/region_allocation.rs
smklein: Done

PR #3382, opened 2023-06-19
Remove tools/helios_version

jmpesp:
This is motivated by a commit in stlouis that requires the latest helios commit. https://pkg.oxide.computer/helios-netdev is updated whenever stlouis is, so images that are built here before helios_version is updated will be wrong. Unpin the helios version so that this does not occur.

### DIFF
--- a/.github/buildomat/jobs/host-image.sh
+++ b/.github/buildomat/jobs/host-image.sh
@@ -42,13 +42,11 @@ rustc --version
 
 TOP=$PWD
 
-source "$TOP/tools/helios_version"
 source "$TOP/tools/include/force-git-over-https.sh"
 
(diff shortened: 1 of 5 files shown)


iliana (approved):
I think I would still prefer to lock to a specific Helios commit in this repo, but I agree that if we can't lock to a particular revision of the package repositories too, then it's better to have a correct-but-unlocked build than an incorrect build.

jclulow:
We can eventually, it just doesn't do that right at this moment.  @citrus-it is working on updating some of the OS bits, and when we get that together we'll also switch to using OS bits built from **stlouis** with a proper version number that we'll be able to lock properly.

PR #3383, opened 2023-06-20
plumb switch route VLAN identifiers

rcgoodfellow:
Plumbing for
- https://github.com/oxidecomputer/dendrite/pull/381


common/src/api/external/mod.rs
```diff
@@ -2459,6 +2459,10 @@ pub struct SwitchPortRouteConfig {
 
     /// The route's gateway address.
     pub gw: IpNet,
+
+    /// The VLAN identifier for the route. Use this if the gateway is reachable
+    /// over an 802.1Q tagged L2 segment.
+    pub vid: Option<u16>,
```
internet-diglett, line 2465: _tiny_ nit, `vlan_id` would probably be more recognizable at a glance.

internet-diglett (approved):
One comment, LGTM!

common/src/api/external/mod.rs
rcgoodfellow: done

PR #3387, opened 2023-06-20
[wicket/wicketd] Run RSS from Wicket

jgallagher:
The bulk of code in this PR is the new `rack_setup` wicket pane, which includes a bunch of popup-related things (which are all fairly verbose for now; I followed what the update pane does structurally, and am very thankful it exists!). On a fresh (uninitialized and unconfigured) system, the pane should look roughly like this, although hopefully not with `UNKNOWN` bootstrap addresses (those are because these screenshots are from a simulated dev environment that doesn't have a bootstrap network):

pr0

If we attempt to start an update, we'll get a confirm popup; pressing "Yes" will fail with the error from wicketd (in this case, "you need to configure things!"):

pr1

After filling in all the configuration options via the CLI added in #3326, we see the options reflected in the UI:

pr2

at which point starting RSS should succeed, and the `Current rack status: ` line will change to `Initializing`. This will eventually progress to a new state once RSS completes (whether successful or not).


wicket/src/ui/panes/rack_setup.rs
```diff
@@ -0,0 +1,816 @@
...
+pub struct RackSetupPane {
+    help: Vec<(&'static str, &'static str)>,
+    rack_uninitialized_help: Vec<(&'static str, &'static str)>,
+    rack_initialized_help: Vec<(&'static str, &'static str)>,
+    scroll_offset: usize,
+
+    popup: Option<Popup>,
+}
+
+impl Default for RackSetupPane {
```
andrewjstone, line 115: Contextual help is so hot these days!

wicket/src/ui/panes/rack_setup.rs
```diff
@@ -0,0 +1,816 @@
...
+            ) => {
+                self.popup = None;
+                Some(Action::Redraw)
+            }
+            (Popup::RackSetup(kind @ PopupKind::Prompting), Cmd::Yes) => {
+                *kind = PopupKind::Waiting;
+                Some(Action::StartRackSetup)
+            }
+            (Popup::RackReset(kind @ PopupKind::Prompting), Cmd::Yes) => {
+                // TODO: actually trigger rack reset
```
andrewjstone, line 181: Seems like this is a left over comment

andrewjstone (approved):
Looks awesome. Ship it!

jmpesp (approved):
I deployed this in the Canada region and it worked (after setting up the appropriate simulated stuff) - some suggestions though:

- values that are input into `setup set-config` should be rejected earlier. I made it through RSS with values that would not parse as `IpAddr` and init failed
- a similar comment for the uploaded certificates: I got all the way through RSS with certificates that couldn't be used without TLS related errors. the certificates should be validated somehow. for posterity:

```
jwm@fancyfeast:~$ curl --cacert ~/wicket.ca.pem https://recovery.sys.oxide.test/v1/me
curl: (60) SSL certificate problem: unsupported certificate purpose
More details here: https://curl.se/docs/sslcerts.html

curl failed to verify the legitimacy of the server and therefore could not
establish a secure connection to it. To learn more about this situation and
how to fix it, please visit the web page mentioned above.
```

PR #3390, opened 2023-06-21
Unify services and datasets, external and internal DNS are datasets too

### LINKED ISSUE #3018: DNS zones should provision storage in a dataset explicitly

smklein:
## Prelude 

This combines:

- https://github.com/oxidecomputer/omicron/pull/3022
- https://github.com/oxidecomputer/omicron/pull/3025
- https://github.com/oxidecomputer/omicron/pull/3026

Into one PR. Trying to split them out resulted in intermediate states that were not worthwhile debugging - I'd rather invest the energy into fixing this "more-desirable end-state".

## What does this PR do

- (Internal and External) DNS services are explicitly backed by datasets, making it possible to "delete and re-initialize" arbitrary zones without data loss. This is necessary for correct behavior in cold boot.
- Merges the concept of "datasets" into "services".
  - Previously, "datasets" were considered services managing data, and "services" were considered services *without datasets* - the two were totally disjoint. This was a bit arbitrary, and made it more complicated to conform sleds during service management.
  - As one example of an inconsistency, "requesting services" was vectorized ("ask for all services that should be on the sled"), "requesting datasets" was not (it was an ask-for-one-at-a-time API).
  - Now, "services" are universal, and they may optionally manage datasets. This unifies pathways a bit, and simplifies the API.

Fixes #3018
Part of #2978


illumos-utils/src/zpool.rs
```diff
@@ -284,12 +284,39 @@ pub enum ZpoolKind {
...
 /// when reading the structure, and validate that the UUID can be utilized.
-#[derive(Clone, Debug, Hash, PartialEq, Eq, JsonSchema)]
+#[derive(Clone, Debug, Hash, PartialEq, Eq)]
 pub struct ZpoolName {
     id: Uuid,
     kind: ZpoolKind,
 }
 
+/// Custom JsonSchema implementation to encode the constraints on Name.
+impl JsonSchema for ZpoolName {
```
smklein, line 294: Why am I doing this?

This type has a custom serialization / deserialization method, to encode it as "ox{i,p}_UUID". That's how it's currently represented in e.g. TOML config files.

IMO this format is kinda nice, because it exactly matches the name a user can see if they run `zpool list`. At the same time, the in-memory structure is nice, because it lets us do handy things like "match on `kind`".

Well, this works when serializing/deserializing to TOML, but sorta broke when using `JsonSchema` and progenitor -- in that case, it was serialized as `{"id":..., "kind":....}`, and couldn't be deserialized from that format.

By using this custom `JsonSchema` impl, this type is always just serialized "as a single string".

sled-agent/src/bootstrap/agent.rs
```diff
@@ -425,7 +425,7 @@ impl Agent {
         let paths = sled_config_paths(&storage_resources).await;
         let maybe_ledger =
             Ledger::<PersistentSledAgentRequest>::new(&ba_log, paths).await;
-        let make_agent = move |initialized| Agent {
+        let make_bootstrap_agent = move |initialized| Agent {
```
smklein, line 428: This was just a rename for clarity - I had to scan this code when debugging, and wanted to make it more clear.

sled-agent/src/http_entrypoints.rs
```diff
@@ -31,7 +30,6 @@ type SledApiDescription = ApiDescription<SledAgent>;
 pub fn api() -> SledApiDescription {
     fn register_endpoints(api: &mut SledApiDescription) -> Result<(), String> {
         api.register(disk_put)?;
-        api.register(filesystem_put)?;
```
smklein, line 34: Initialization of datasets is done implicitly when requesting services that have attached datasets.

sled-agent/src/params.rs
```diff
@@ -623,10 +595,6 @@ impl TryFrom<ServiceZoneService>
 }
 
 /// Used to request that the Sled initialize multiple services.
-///
-/// This may be used to record that certain sleds are responsible for
-/// launching services which may not be associated with a dataset, such
-/// as Nexus.
```
smklein, line 629: This comment has been removed, as now these requests handle both kinds of services (with and without datasets).

sled-agent/src/sled_agent.rs
```diff
@@ -528,6 +528,25 @@ impl SledAgent {
...
+            .services
+            .iter()
+            .filter_map(|service| service.dataset.clone())
+            .collect();
+
+        // TODO:
+        // - If these are the set of filesystems, we should also consider
+        // removing the ones which are not listed here.
+        // - It's probably worth sending a bulk request to the storage system,
+        // rather than requesting individual datasets.
```
smklein, line 541: These are still worthwhile follow-ups, but not regressions

smklein:
Rebased around the service bundle changes - this is still ready for review!

sled-agent/src/rack_setup/plan/service.rs
```diff
@@ -388,99 +390,127 @@ impl Plan {
...
+                        name: DatasetName::new(
+                            u2_zpools[0].clone(),
+                            crate::params::DatasetKind::Clickhouse,
+                        ),
+                    }),
+                    gz_addresses: vec![],
+                    services: vec![ServiceZoneService {
+                        id,
+                        details: ServiceType::Clickhouse,
+                    }],
```
karencfv, line 442: When scaling up ClickHouse, the coordinator nodes (keepers) will need to be installed in separate hosts than the databases (main and replica). We'll need to have a separate service for this from what I can tell.

smf/external-dns/config.toml
```diff
@@ -11,5 +11,5 @@ path = "/dev/stdout"
 if_exists = "append"
 
 [storage]
-storage_path = "/var/oxide/dns"
+storage_path = "/data"
```
davepacheco, line 14: `/data/dns` (in case we ever want to put anything else there)?

smf/internal-dns/config.toml
```diff
@@ -11,5 +11,5 @@ path = "/dev/stdout"
 if_exists = "append"
 
 [storage]
-storage_path = "/var/oxide/dns"
+storage_path = "/data"
```
davepacheco, line 14: Same as the other file

davepacheco (approved):

illumos-utils/src/zpool.rs
bnaecker: It might be useful to have a regular expression in this. That can help progenitor create better error messages, for example.

sled-agent/src/rack_setup/service.rs
```diff
@@ -371,36 +330,52 @@ impl ServiceInner {
...
+    //
+    // This is useful in a rack-setup context, where initial boot ordering
+    // can matter for first-time-setup.
+    //
+    // Note that after first-time setup, the initialization order of
+    // services should not matter.
+    async fn ensure_all_services_of_type(
         &self,
         service_plan: &ServicePlan,
+        zone_types: &HashSet<std::mem::Discriminant<ZoneType>>,
```
bnaecker, line 343: I'm a bit confused by the use of `Discriminant` here. Does `ZoneType` carry data? If not, I think it'd be less surprising to store the enum variants directly in the set.

sled-agent/src/sled_agent.rs
```diff
@@ -229,7 +228,7 @@ impl SledAgent {
             "component" => "SledAgent",
             "sled_id" => request.id.to_string(),
         ));
-        info!(&log, "created sled agent");
+        info!(&log, "SledAgent::new(..) staring");
```
bnaecker, line 231: Spelling nit: "starting". Unless you mean to say the sled agent is staring at the other services, waiting for them to come up :)

bnaecker (approved):
Looks good to me, a few nits, but nice work overall!

sled-agent/src/rack_setup/service.rs
smklein: I wish I could! But [enum variants are not types](https://github.com/rust-lang/rfcs/pull/1450).

I really just wanted a way to say "accept some subset of the enum" -- `std::mem::discriminant` uniquely identifies enum variants, so this seemed to fit.

sled-agent/src/sled_agent.rs
smklein: 👀 fixed

illumos-utils/src/zpool.rs
smklein: Added in ae93a3ece49a63da7a352edbfd7d111dd086953f , with tests!

common/src/sql/dbinit.sql
```diff
@@ -186,6 +186,9 @@ CREATE INDEX ON omicron.public.switch (
  */
 
 CREATE TYPE omicron.public.service_kind AS ENUM (
+  'crucible',
+  'cockroach',
+  'clickhouse',
```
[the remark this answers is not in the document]
smklein: Done in 0d4e9dca6adf2b1b3927b3d9c78863c78b527174

smf/external-dns/config.toml
smklein: Done in 02e52b4741c5a499690b29c77b77e7614a9536a8

smf/internal-dns/config.toml
smklein: Done in 02e52b4741c5a499690b29c77b77e7614a9536a8

sled-agent/src/rack_setup/service.rs
davepacheco: I'm confused, too.  Can't you have a `HashSet<ZoneType>`?  That would allow any variant to appear in the hashset, but so does `HashSet<Discriminant<ZoneType>>`, doesn't it?

sled-agent/src/rack_setup/plan/service.rs
smklein: That tracks - if we want this service to be provisioned as a part of rack setup, this is where we could make that allocation.

sled-agent/src/rack_setup/service.rs
bnaecker: Yeah, that makes sense, and is why I was asking about the variants containing data. I guess there are a few things. First, given that they _don't_ currently carry data, it seems preferable to me to use the enum directly as the value for the map. Second, even if we _do_ add data, do we really not want to distinguish between them in this code?

smklein: I went ahead and swapped it back to a non-discriminant version in 96f3afd40 - maybe we can just revisit that question if we ever need to add data to this enum in the future?

PR #3391, opened 2023-06-21
Update Omicron to latest propolis with InstanceSerialConsoleHelper

leftwo:
Update Omicron to get newer Crucible and Propolis, which also includes the changes to support
the new Propolis `InstanceSerialConsoleHelper`.


nexus/src/app/instance.rs
```diff
@@ -1362,6 +1361,43 @@ impl super::Nexus {
...
+                let ip_addr = instance
+                    .runtime_state
+                    .propolis_ip
+                    .ok_or_else(|| {
+                        Error::internal_error(
+                            "instance's hypervisor IP address not found",
+                        )
+                    })?
+                    .ip();
+                Ok(SocketAddr::new(ip_addr, PROPOLIS_PORT))
```
leftwo, line 1385: I think this might not be the proper port?

nexus/src/app/instance.rs
```diff
@@ -1362,6 +1361,43 @@ impl super::Nexus {
         }
     }
 
+    async fn propolis_addr_for_instance(
```
sunshowers, line 1364: Could `propolis_client_for_instance` below use `propolis_addr_for_instance` to avoid duplicating code?

sunshowers (commented):
Looks good overall, though I wrote half of it so maybe someone else should review it ;)

leftwo:
> Looks good overall, though I wrote half of it so maybe someone else should review it ;)

And, thank you for your half (or really more than half).  :)

nexus/src/app/instance.rs
leftwo: For sure this is ripe for a refactor.
I want to be sure that we are actually using the correct address:port for the serial console and once I verify
that I'll refactor this to something better.

Cargo.lock
```diff
@@ -6853,8 +6854,7 @@ dependencies = [
  "schemars_derive",
  "serde",
  "serde_json",
- "uuid 0.8.2",
```
ahl, line 6856: woot!

nexus/src/app/instance.rs
leftwo: Ah good, thanks.  I was not sure if we required a different port here for the serial console stuff.

nexus/src/app/instance.rs
leftwo: Refactor done.

leftwo:
This should be good to go now.
I did test that with these changes I can still access the serial console on an instance.

lifning (approved):

PR #3396, opened 2023-06-22
CertificateCreate cert/key Vec<u8> -> String

ahl:
This is intended to address: https://github.com/oxidecomputer/oxide.rs/issues/161

In short: the API current calls for the key and cert to be specified as a `Vec<u8>`. While progenitor **could**, say, infer that a `Vec<u8>` should be some file that we load from disk, in fact the inputs here are PEM-formatted strings. I think therefore that a `String` is a better way to model these parameters.

This might not be the right approach; input welcome.

### DIFF
--- a/.github/buildomat/jobs/deploy.sh
+++ b/.github/buildomat/jobs/deploy.sh
@@ -39,7 +39,7 @@ _exit_trap() {
 	pfexec netstat -rncva
 	pfexec netstat -anu
 	pfexec arp -an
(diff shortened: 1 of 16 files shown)


smklein (approved):

davepacheco:
Nice!  Should we store in the database as a string too?

ahl:
> Nice! Should we store in the database as a string too?

I defer to your judgement on that. What do you think?

PR #3399, opened 2023-06-22
Use Ipv4Addr instead of String in RackNetworkConfig IP address fields

### LINKED ISSUE #3397: `RackNetworkConfig`: Use stronger types than `String`?
When @jmpesp was testing #3387 in the Canada region, we tried running with a `RackNetworkConfig` of:

```toml
[rack_network_config]
gateway_ip = ""
infra_ip_first = ""
infra_ip_last = ""
uplink_port = ""
uplink_port_speed = "speed_10g"
uplink_port_fec = "none"
uplink_ip = ""
```

This is "valid" as far as type type system is concerned: the two enum values (`uplink_port_{speed,fec}`) have valid values, and the remaining values are all `String`s; however, attempting to run RSS with these values eventually failed with

```
Bad configuration for setting up rack: unable to parse rack_network_config.uplink_up as IpAddr: invalid IP address syntax
```

If we're trying to parse values as `IpAddr`, I think we should be able to make the fields `IpAddr` from the top.

jgallagher:
I believe based on current usage that all of these are required to be `Ipv4Addr`s specifically, and not `IpAddr`/`Ipv6Addr`:

* `infra_ip_first` and `infra_ip_last` are stored in a variable named `ipv4_block`, and a later `ipv6_block` (that does not use those config values) is also created
* `uplink_ip`'s doc comment says it should be in the `infra_ip_first..last` range, which means it would also have to be ipv4
* `gateway_ip` was being parsed into `nexthop: Option<Ipv4Addr>`.

but I don't have a lot of context here so would appreciate confirmation.

Fixes #3397.

### DIFF
--- a/common/src/api/internal/shared.rs
+++ b/common/src/api/internal/shared.rs
@@ -7,7 +7,7 @@
 use crate::api::external;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
-use std::net::IpAddr;
+use std::net::{IpAddr, Ipv4Addr};
 use uuid::Uuid;
 
 /// The type of network interface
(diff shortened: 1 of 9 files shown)


internet-diglett (approved):
LGTM!

PR #3401, opened 2023-06-22
Split system metrics endpoint into system and silo metrics

### LINKED ISSUE #3030: API 403s on silo utilization despite being silo admin
On dogfood, I see `/v1/system/metrics/{metric}` 403s on `/utilization` when I am silo admin. (Note that because of #3029 you have to go to the silo and navigate to Utilization client-side rather than hitting the route directly.) This endpoint requires fleet read and we presumably don't have that as a non-operator user.

https://github.com/oxidecomputer/omicron/blob/8984c3bc1c7d5d2ad17c7ad8003d2195b80d16a8/nexus/src/app/metrics.rs#L27

This endpoint is pretty free-form and can pull metrics from different levels depending on the query. Off the top of my head, a relatively easy (though probably temporary) fix would be to parse the query and use different permission checks depending on the contents.

david-crespo:
Closes #3030

Instead a single endpoint that requires fleet viewer and allows a `resource_id` for any type of resource, we now have

- System metrics endpoint, which takes silo ID or nothing (for fleet)
  - Requires fleet read no matter what, and then if silo is specified also requires read on that silo
- Silo metrics endpoint scoped to the current silo, which takes project ID or nothing (for whole silo)
  - Requires silo read no matter what ([which all users have on their own silo](https://github.com/oxidecomputer/omicron/blob/dc64fc04b93160e194b291780efa05c09f467068/nexus/db-queries/src/authz/omicron.polar#L157-L176)) and then project read if project is specified


nexus/tests/output/nexus_tags.txt
```diff
@@ -54,6 +54,10 @@ OPERATION ID                             METHOD   URL PATH
 login_local                              POST     /v1/login/{silo_name}/local
 login_saml                               POST     /login/{silo_name}/saml/{provider_name}
 
+API operations found with tag "metrics"
+OPERATION ID                             METHOD   URL PATH
+silo_metric                              GET      /v1/metrics/{metric_name}
+
```
david-crespo, line 60: @benjaminleonard @ahl what am I doing here

benjaminleonard: You didn't do this manually right? It's generated 🤖

david-crespo: Yeah, but I manually put the tag `metrics` on the endpoint.

nexus/tests/integration_tests/instances.rs
```diff
@@ -717,14 +717,64 @@ async fn test_instance_migrate_v2p(cptestctx: &ControlPlaneTestContext) {
...
+            get_latest_system_metric(cptestctx, "cpus_provisioned", *id).await,
+            cpus
+        );
+        assert_eq!(
+            get_latest_system_metric(cptestctx, "ram_provisioned", *id).await,
+            ram
+        );
+    }
+}
+
```
david-crespo, line 770: now asserting about both metrics endpoints. extracted helper because the test would be ridiculously long otherwise

nexus/src/app/metrics.rs
```diff
@@ -39,4 +55,45 @@ impl super::Nexus {
...
+            }
+        };
+        self.select_timeseries(
+            &timeseries,
+            &[&format!("id=={}", resource_id)],
+            pagination,
+            limit,
+        )
+        .await
+    }
```
david-crespo, line 98: @davepacheco any thoughts on the authz checks here, especially how to test them? I don't think the verify endpoints test is sufficient.

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -269,6 +269,7 @@ pub fn external_api() -> NexusApiDescription {
         api.register(certificate_delete)?;
 
         api.register(system_metric)?;
+        api.register(silo_metric)?;
```
zephraph, line 272: @ahl what do you think about the operation names here? It's a little awkward to match our current naming scheme. Technically the metrics endpoint does return a list of metrics so we could potentially call it `system_metric_list`... but that makes it sound like it's listing out different _kinds_ of metrics. It would almost be more accurate to use `system_metric_view` in that you're viewing the system metrics... but it's a list.

zephraph (approved):
This is a huge improvement so I'm 👍. There are some follow ups we should tackle:

- Given that the authz tests won't allow us to retest the same endpoint twice (with different query params) we don't have complete coverage of all the authz cases. We should update the tests to treat different query params as a unique URL in terms of test coverage
- I'm still unsure about the operation names for metrics. We want to get those right because it'll impact the SDKs.

david-crespo:
I'll merge and make an issue for those followups.

PR #3403, opened 2023-06-23
ci_download_softnpu_machinery: don't skip just because file exists locally

luqmana:
Ran into this fun issue again where `ci_download_softnpu_machinery` would skip updating the softnpu binaries if it found them in `out/softnpu` already. That's all well and good until the pinned commit actually gets updated and you've run `install_runner_prerequisites.sh` thinking that's enough.

To anyone who's softnpu deploy doesn't seem to be working and `/softnpu.log` in the `softnpu` zone is spamming:
```
Dec 28 10:34:48.723 WARN got sidecar egress port of 0
Dec 28 10:34:48.723 WARN this is probably a p4 program bug
Dec 28 10:34:48.723 WARN dropping packet
```
You're running an old copy of `softnpu` is why.

Fix that by using the published sha256 hashes to verify any local copy that exists.

### DIFF
--- a/tools/ensure_buildomat_artifact.sh
+++ b/tools/ensure_buildomat_artifact.sh
@@ -1,0 +1,130 @@
+#!/bin/bash
+#
+# Ensure a buildomat artifact is downloaded and available locally.
+
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+# Published buildomat artifacts are available at a predictable URL:
+# https://buildomat.eng.oxide.computer/public/file/ORG/REPO/SERIES/HASH/ARTIFACT
+BUILDOMAT_BASE_URL="https://buildomat.eng.oxide.computer/public/file"
+
+# Default value for optional flags
+ORG="oxidecomputer"
+SERIES="image"
+CHECK_HASH=true
+OUTDIR="$PWD"
+
+function usage() {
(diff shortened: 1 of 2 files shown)


tools/ensure_buildomat_artifact.sh
```diff
@@ -0,0 +1,130 @@
...
+Usage: $0 [OPTIONS] <ARTIFACT> <REPO> <COMMIT>
+
+    REPO:     The repository that published the artifact.
+    ARTIFACT: The name of the artifact.
+    COMMIT:   The commit the artifact was published from.
+
+Options:
+    -o <ORG>    Org containing the repository. [default: $ORG]
+    -s <SERIES> The series artifact was published to. [default: $SERIES]
+    -n          Disable artifact validation.
```
leftwo, line 31: small nit: I would use `-f` for "force" instead of `-n`.  `-n` is typically used for "dry run", where a tool
would tel you what it would do, but stop short of actually doing it.

leftwo (approved):
This got me, so thanks for fixing it!

tools/ensure_buildomat_artifact.sh
luqmana: a dry run flag might also be nice to add later

PR #3404, opened 2023-06-23
Fix tag `external_docs` links (and cleanup unused)

benjaminleonard:
I'm updating the docs site so these automatically redirect to the associated section. Changing the system ones so they can be easily handled. Removing the tags that are not used.


david-crespo (approved):
Oh good. I just noticed this file and was wondering what those links were about.

PR #3408, opened 2023-06-23
nexus: provide stable PCI slot assignments for disks

### LINKED ISSUE #3224: Want stable PCI slot assignments for disks (to avoid invalidating boot order on disk attach/detach)
I couldn't replicate this issue consistently but this is the instance which I ran into such an issue.

Order of events:
1. Created instance with two disks - disk0: Ubuntu focal, disk1: blank disk
2. After starting the vm, formatted and mounted disk1
```
ubuntu@focal3:~$ lsblk
NAME         MAJ:MIN RM  SIZE RO TYPE MOUNTPOINT
loop0          7:0    0 91.9M  1 loop /snap/lxd/24061
loop1          7:1    0 63.3M  1 loop /snap/core20/1822
loop2          7:2    0 49.9M  1 loop /snap/snapd/18357
vda          252:0    0   21K  1 disk 
nvme0n1      259:0    0    3G  0 disk 
├─nvme0n1p1  259:2    0  2.9G  0 part /
├─nvme0n1p14 259:3    0    4M  0 part 
└─nvme0n1p15 259:4    0  106M  0 part /boot/efi
nvme1n1      259:1    0   30G  0 disk /data
```
3. Stopped vm. Added 2 more blank disks (disk2 and disk3) via the console which invoked the disk attach API. Started vm again:
```
ubuntu@focal3:~$ lsblk
NAME         MAJ:MIN RM  SIZE RO TYPE MOUNTPOINT
loop0          7:0    0 63.3M  1 loop /snap/core20/1822
loop1          7:1    0 63.3M  1 loop /snap/core20/1879
loop2          7:2    0 91.9M  1 loop /snap/lxd/24061
loop3          7:3    0 49.9M  1 loop /snap/snapd/18357
vda          252:0    0   21K  1 disk 
nvme0n1      259:0    0   30G  0 disk 
nvme1n1      259:1    0   30G  0 disk 
nvme2n1      259:2    0    3G  0 disk 
├─nvme2n1p1  259:3    0  2.9G  0 part /
├─nvme2n1p14 259:4    0    4M  0 part 
└─nvme2n1p15 259:5    0  106M  0 part /boot/efi
nvme3n1      259:6    0   30G  0 disk /data4096
```
I noticed the boot disk changed from `nvme0n1` to `nvme2n1`. Assuming that `nvme3n1` was one of the newly attached disks, I tried to format it but then found that it was actually disk1 (i.e. the device previously named `nvme1n1`)
```
ubuntu@focal3:~$ sudo mkfs -t ext4 /dev/nvme3n1
mke2fs 1.45.5 (07-Jan-2020)
/dev/nvme3n1 contains a ext4 file system
	last mounted on /data on Sat May  6 04:57:54 2023
Proceed anyway? (y,N) N
```
4. Stopped vm again. Removed disk1 via console. Started vm again and it stayed in `starting` state forever, presumably because it used the incorrect disk to boot?

gjcolombo:
Add a `slot` column to the disk table and modify `instance_attach_disk` to set it when a disk is attached to an instance.

Slots are chosen using the `NextItem` query generator that NICs use for their slot assignments. The bulk of this change is new code to combine the `NextItem` fragment with the other SQL statements that mark a disk as attached (changing its state and setting its `attach_instance_id`) in a manner that allows the resulting query fragment to be passed as the body of the SET clause of the UPDATE statement that `instance_attach_disk` passes to the `attach_resource` generic function.

Slot assignments remain stable for as long as a disk is attached, but they aren't preserved if a disk is detached and reattached (the algorithm will choose the first available slot, even if the disk's previously assigned slot is available).

Tests:

- New integration test
- On a dev cluster, created an instance with a single boot disk, then attached disks with alphabetically earlier and later names. Verified that the boot disk remained in slot 0 and that booting the instance succeeded (i.e. guest firmware was able to find the boot OS successfully). Also played around with detaching and reattaching disks and making sure via CRDB shell that their slot assignments were as expected. 

Fixes #3224.

### DIFF
--- a/nexus/db-model/src/disk.rs
+++ b/nexus/db-model/src/disk.rs
@@ -44,6 +44,16 @@ pub struct Disk {
     #[diesel(embed)]
     pub runtime_state: DiskRuntimeState,
 
+    /// The PCI slot (within the bank of slots reserved to disks) to which this
+    /// disk should be attached if its attached instance is started, or None
+    /// if there is no such assignment.
+    ///
+    /// Slot assignments are managed entirely in Nexus and aren't modified by
+    /// runtime state changes in the sled agent, so this field is part of the
+    /// "main" disk struct and not the runtime state (even though the attachment
+    /// state and slot assignment will often change together).
+    pub slot: Option<i16>,
+
     /// size of the Disk
     #[diesel(column_name = size_bytes)]
     pub size: ByteCount,
@@ -95,6 +105,7 @@ impl Disk {
             project_id,
             volume_id,
             runtime_state: runtime_initial,
+            slot: None,
             size: params.size.into(),
             block_size,
             create_snapshot_id,
--- a/nexus/db-model/src/disk.rs
+++ b/nexus/db-model/src/disk.rs
@@ -3,7 +3,7 @@
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 use super::{BlockSize, ByteCount, DiskState, Generation};
-use crate::schema::disk;
+use crate::{schema::disk, unsigned::SqlU8};
 use chrono::{DateTime, Utc};
 use db_macros::Resource;
 use nexus_types::external_api::params;
@@ -52,7 +52,7 @@ pub struct Disk {
     /// runtime state changes in the sled agent, so this field is part of the
     /// "main" disk struct and not the runtime state (even though the attachment
     /// state and slot assignment will often change together).
-    pub slot: Option<i16>,
+    pub slot: Option<SqlU8>,
 
     /// size of the Disk
     #[diesel(column_name = size_bytes)]
--- a/nexus/db-queries/src/db/queries/disk.rs
+++ b/nexus/db-queries/src/db/queries/disk.rs
@@ -1,0 +1,168 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! Helper queries for working with disks.
+
+use crate::db;
+use crate::db::queries::next_item::{DefaultShiftGenerator, NextItem};
+use diesel::{
+    pg::Pg,
+    query_builder::{AstPass, QueryFragment, QueryId},
+    sql_types, Column, QueryResult,
+};
+use uuid::Uuid;
+
+/// The maximum number of disks that can be attached to an instance.
+//
+// This is defined here for layering reasons: the main Nexus crate depends on
+// the db-queries crate, so the disk-per-instance limit lives here and Nexus
+// proper re-exports it.
+pub const MAX_DISKS_PER_INSTANCE: u32 = 8;
+
+/// A wrapper for the query that selects a PCI slot for a newly-attached disk.
+///
+/// The general idea is to produce a query that left joins a single-column table
+/// containing the sequence [0, 1, 2, ..., MAX_DISKS] against the disk table,
+/// filtering for disks that are attached to the instance of interest and whose
+/// currently-assigned slots match the slots in the sequence. Filtering this
+/// query to just those rows where a slot has no disk attached yields the first
+/// available slot number. If no slots are available, the returned value is
+/// MAX_DISKS, which is not a valid slot number and which will be rejected when
+/// attempting to update the row (the "slot" column has a CHECK clause that
+/// enforces this at the database level).
+///
+/// See the `NextItem` documentation for more details.
+#[derive(Debug, Clone, Copy)]
+struct NextDiskSlot {
+    inner: NextItem<
+        db::schema::disk::table,
+        i16,
+        db::schema::disk::dsl::slot,
+        Uuid,
+        db::schema::disk::dsl::attach_instance_id,
+    >,
+}
+
+impl NextDiskSlot {
+    fn new(instance_id: Uuid) -> Self {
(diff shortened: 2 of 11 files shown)


jmpesp (approved):
This looks good to me, but I'd want some explicit documentation stating (mea culpa if this is wrong):

- A Disk is assigned the first available slot when attached.
- A Disk's slot is erased when it is detached from an instance.
- Slot assignments are not configurable by the user, other than by following those rules.

leftwo:
> matically.

If the above is true, lets at least make an issue [here I believe](https://github.com/oxidecomputer/docs/issues)

gjcolombo:
> explicit documentation

Your summary is correct. Do you have strong feelings about whether to document this in a freestanding "guide to disks" in the docs repo vs. in the `instance_disk_attach` API documentation? I think I'd pick the former if I were forced to choose but see benefits to both approaches ("keep this level of detail out of the API docs, it's too much inside baseball" vs. "document API behaviors in the API docs and not in some other doc").

jmpesp:
> Do you have strong feelings about whether to document this in a freestanding "guide to disks" in the docs repo vs. in the instance_disk_attach API documentation?

Both!

nexus/db-model/src/disk.rs
```diff
@@ -44,6 +44,16 @@ pub struct Disk {
...
 
+    /// The PCI slot (within the bank of slots reserved to disks) to which this
+    /// disk should be attached if its attached instance is started, or None
+    /// if there is no such assignment.
+    ///
+    /// Slot assignments are managed entirely in Nexus and aren't modified by
+    /// runtime state changes in the sled agent, so this field is part of the
+    /// "main" disk struct and not the runtime state (even though the attachment
+    /// state and slot assignment will often change together).
+    pub slot: Option<i16>,
```
bnaecker, line 55: IIRC, there are a few helper types for creating unsigned integers out of SQL signed types. `SqlU32` I think is a good example. That might be useful here, since this is basically a `u8` it seems.

nexus/db-queries/src/db/queries/disk.rs
```diff
@@ -0,0 +1,168 @@
...
+        // query and parenthesizes it so that it can be used as the argument to
+        // a SET statement. The sticky bit is the trait bounds: SingleValueDsl
+        // is implemented for SelectQuery + LimitDsl, and while it's easy enough
+        // to impl SelectQuery for NextDiskSlot, implementing LimitDsl is harder
+        // (and seems to be discouraged by the trait docs)--it is natively
+        // implemented for diesel::Table, but here there's no table type, in
+        // part because the left side of the join is synthesized from thin air
+        // by the NextItem query fragment.
+        //
+        // Rather than spar extensively with Diesel over these details, just
```
bnaecker, line 141: This is a good call. I generally have only come out bruised and bloodied from such battles.

nexus/tests/integration_tests/disks.rs
```diff
@@ -351,6 +352,115 @@ async fn test_disk_create_disk_that_already_exists_fails(
...
+    let listed_disks = disks_list(&client, &url_instance_disks).await;
+    assert_eq!(listed_disks.len(), 0);
+
+    let url_instance_attach_disk =
+        get_disk_attach_url(&instance.identity.id.into());
+
+    async fn get_disk_slot(
+        ctx: &ControlPlaneTestContext,
+        disk_id: Uuid,
+    ) -> i16 {
```
bnaecker, line 388: So I think one consequence of the change I suggested earlier around the model type is that this would be a `u8`. I _think_ that's what we want, since we cast it to that internally in at least one place. It should be unsigned, for sure, and that seems an improvement in terms of the public API.

nexus/tests/integration_tests/disks.rs
```diff
@@ -351,6 +352,115 @@ async fn test_disk_create_disk_that_already_exists_fails(
     disks_eq(&disks[0], &disk);
 }
 
+#[nexus_test]
+async fn test_disk_slot_assignment(cptestctx: &ControlPlaneTestContext) {
```
bnaecker, line 356: This is a great test.

bnaecker (approved):
Looks great to me, glad the `NextItem` query helper was useful! I have one suggestion around making the slot an unsigned value everywhere outside the DB, but otherwise all LGTM.

gjcolombo:
> > Do you have strong feelings about whether to document this in a freestanding "guide to disks" in the docs repo vs. in the instance_disk_attach API documentation?
> 
> Both!

That works too :) Having slept on it, I think I'd be inclined to see if we can have a guide that's cross-referenced from the API docs (theory: you don't need to know all the inside baseball about PCI slots to use the attach API succesfully, especially because it has no "slot" parameters, but it's important to cross-reference the two so that the details are discoverable). I'm not sure how best to do that, so I'll file an issue in the docs repo to figure this out so that we can start getting some miles on the code changes.

nexus/db-model/src/disk.rs
```diff
@@ -3,7 +3,7 @@
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 use super::{BlockSize, ByteCount, DiskState, Generation};
-use crate::schema::disk;
+use crate::{schema::disk, unsigned::SqlU8};
```
bnaecker, line 6: 🤘

nexus/db-model/src/disk.rs
gjcolombo: Great suggestion, thanks! Implemented (as you've seen) in f002264.

Issue #3356, opened 2023-06-14
proxy_instance_serial_ws() uses `select!` with non-cancel-safe future(s)


jgallagher:
In `proxy_instance_serial_ws()`, we [construct four futures](https://github.com/oxidecomputer/omicron/blob/73545a9e601c33f347c80c1c4aaa2a107599660d/nexus/src/app/instance.rs#L1423-L1458) over which we will `tokio::select!`. In any given loop iteration, at least two of the futures are instances of `Fuse::terminated()` (and therefore ignored by `select!`). The possible combinations of non-`Fuse::terminated()` futures are:

1. `nexus_read` (a [`StreamExt::next()`](https://docs.rs/futures-util/latest/futures_util/stream/trait.StreamExt.html#method.next) future) and `nexus_write`(a [`SinkExt::send()`](https://docs.rs/futures-util/latest/futures_util/sink/trait.SinkExt.html#method.send) future)
2. `nexus_read` and `propolis_read` (a future from [`InstanceSerialConsoleHelper::recv()`](https://github.com/oxidecomputer/propolis/blob/54a1fc672ae38b77b7f9beeb0dad8afcb3095d5d/lib/propolis-client/src/lib.rs#L124) in propolis)
3. `propolis_write` (a future from [`support::send`](https://github.com/oxidecomputer/propolis/blob/54a1fc672ae38b77b7f9beeb0dad8afcb3095d5d/lib/propolis-client/src/lib.rs#L112-L114) in propolis which is a wrapper around `SinkExt::send()`) alone
4. `propolis_write` and `nexus_write`

Case 3 is fine because there is only one possible future for `select!` to choose, so there is no possibility for cancellation. But in the other three cases, `select!` will cancel one of the two futures once it chooses one. In terms of cancel safety, none of these futures is explicitly documented; however:

* I believe (hope?) that `StreamExt::next()` is cancel safe, because the equivalent method in [tokio-stream's StreamExt](https://docs.rs/tokio-stream/latest/tokio_stream/trait.StreamExt.html#cancel-safety) is explicitly documented as cancel safe.
[...]

PR #3411, opened 2023-06-24
[nexus] make serial console select loop cancel-safe

sunshowers:
Use the cancel-safe adapters defined in
https://github.com/oxidecomputer/cancel-safe-futures to ensure that the
branches of the select loop are all cancel-safe.

See #3356 for a detailed description of what's going on here.

I'd still like to test this out with a real serial console.


nexus/src/app/instance.rs
```diff
@@ -1488,8 +1484,12 @@ impl super::Nexus {
...
                     }
                 }
-                result = nexus_write => {
-                    result?;
+                result = nexus_reserve => {
+                    let permit = result?;
+                    let message = buffered_output
+                        .take()
+                        .expect("nexus_reserve is only active when buffered_output is Some");
+                    permit.send(message)?.await?;
```
sunshowers, line 1492: Note a change in behavior here -- the flush involved here now blocks the whole task. I *think* this is OK since we're on the local network. We could do this differently but if we decide to do that, I think we should do a bit of an overall -- spawn 2-3 different tasks and communicate across them using channels.

lifning: Yeah, it should be acceptable to let the task wait here; I imagine the performance impact being negligible (unless you've observed otherwise). If we end up wanting to further optimize this endpoint's throughput/latency in the future, we can certainly revisit it.

lifning (approved):

leftwo (approved):
Changes look good, And I actually installed them on gimlet-sn21 and verified I
could still get to the serial console in the web UI.

You have to update the `package-manifest.toml` also, with the new propolis commit
id and the sha256.
```
alan@atrium:omicron$ git diff package-manifest.toml 
diff --git a/package-manifest.toml b/package-manifest.toml
index 196b64bd2..f63406546 100644
--- a/package-manifest.toml
+++ b/package-manifest.toml
@@ -248,10 +248,10 @@ service_name = "propolis-server"
 only_for_targets.image = "standard"
 source.type = "prebuilt"
 source.repo = "propolis"
-source.commit = "54dc09da004b46b17819fd80dff8a295c15eb294"
+source.commit = "04a275736e9f3316de6bf2a4077d03acfa4e2cdf"
 # The SHA256 digest is automatically posted to:
 # https://buildomat.eng.oxide.computer/public/file/oxidecomputer/propolis/image/<commit>/propolis-server.sha256.txt
-source.sha256 = "42986a43e4ac69f5eb09e3aed4af01e0d6117254950e81b5afeddd46c127259f"
+source.sha256 = "3b52f303d7f0a687fd895e7129c93ca1d683d2cb676bff9c6c9d15bd3d2b3b5e"
 output.type = "zone"
```

If you want to try it, go to:
`http://172.20.24.12/`
You can log in with:
recovery
oxide

You can also just use the regular CLI and connect that way.

leftwo:
We should also figure out why those CI tests are failing.  Probably unrelated, but better to verify that now before merging.

nexus/src/app/instance.rs
```diff
@@ -1418,41 +1419,36 @@ impl super::Nexus {
...
-                            nexus_write = nexus_sink.send(msg).fuse();
-                            propolis_read = Fuse::terminated();
-                        }
-                    }
+            } else {
+                nexus_read = nexus_stream.next().fuse();
+                propolis_reserve = Fuse::terminated();
+                if buffered_output.is_some() {
+                    nexus_reserve = nexus_sink.reserve().fuse();
+                    propolis_read = propolis_conn.recv().fuse();
```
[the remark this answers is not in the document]
sunshowers: Also added a test for this.

sunshowers:
> We should also figure out why those CI tests are failing. Probably unrelated, but better to verify that now before merging.

The issue I saw looks like it was the propolis-server hash not being updated.

PR #3433, opened 2023-06-27
Improve failure mode in omicron-dev tests that fail

bnaecker:
This fixes a difficult failure mode I ran into recently. Some of the tests in `omicron-dev` spawn CLI programs, such as `cockroach`. These are put into a progress group so that, on failure, the entire group can be signaled and reaped. However, these tests were previously structured so that the group is only killed on a successful test run; an unwrap or other panic early would exit before the shutdown code was run. This leaves the group running and the test hangs.

This PR fixes a couple of those tests, by running the bulk of test code we care about in a separate fallible future. We run it; store the result; kill and reap the child processes; and then assert that the body ran successfully.

### DIFF
--- a/dev-tools/tests/test_omicron_dev.rs
+++ b/dev-tools/tests/test_omicron_dev.rs
@@ -289,66 +289,94 @@ async fn test_db_run() {
     let exec =
         Exec::cmd("bash").arg("-c").arg(cmdstr).stderr(Redirection::Merge);
     let dbrun = run_db_run(exec, true);
-    let (client, connection) = dbrun
-        .listen_config
-        .connect(tokio_postgres::NoTls)
-        .await
-        .expect("failed to connect to newly setup database");
-    let conn_task = tokio::spawn(connection);
+    let test_task = async {
+        let (client, connection) = dbrun
+            .listen_config
+            .connect(tokio_postgres::NoTls)
+            .await
+            .context("failed to connect to newly setup database")?;
+        let conn_task = tokio::spawn(connection);
 
-    assert!(has_omicron_schema(&client).await);
+        anyhow::ensure!(
+            has_omicron_schema(&client).await,
+            "Missing Omicron schema"
+        );
 
-    // Now run db-populate.  It should fail because the database is already
-    // populated.
-    eprintln!("running db-populate");
-    let populate_result = Exec::cmd(&cmd_path)
-        .arg("db-populate")
-        .arg("--database-url")
-        .arg(&dbrun.listen_config_url)
-        .stdout(Redirection::Pipe)
-        .stderr(Redirection::Pipe)
-        .capture()
-        .expect("failed to run db-populate");
-    eprintln!("exit status: {:?}", populate_result.exit_status);
-    eprintln!("stdout: {:?}", populate_result.stdout_str());
-    eprintln!("stdout: {:?}", populate_result.stderr_str());
-    assert!(matches!(populate_result.exit_status, ExitStatus::Exited(1)));
-    assert!(populate_result
-        .stderr_str()
-        .contains("database \"omicron\" already exists"));
-    assert!(has_omicron_schema(&client).await);
+        // Now run db-populate.  It should fail because the database is already
+        // populated.
+        eprintln!("running db-populate");
+        let populate_result = Exec::cmd(&cmd_path)
+            .arg("db-populate")
+            .arg("--database-url")
+            .arg(&dbrun.listen_config_url)
+            .stdout(Redirection::Pipe)
+            .stderr(Redirection::Pipe)
+            .capture()
+            .context("failed to run db-populate")?;
+        eprintln!("exit status: {:?}", populate_result.exit_status);
+        eprintln!("stdout: {:?}", populate_result.stdout_str());
+        eprintln!("stdout: {:?}", populate_result.stderr_str());
+        anyhow::ensure!(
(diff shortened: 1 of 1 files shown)


bnaecker:
To test this, I ran with this diff:

```diff
diff --git a/nexus/examples/config.toml b/nexus/examples/config.toml
index a116b31a5..5e514c497 100644
--- a/nexus/examples/config.toml
+++ b/nexus/examples/config.toml
@@ -44,7 +44,7 @@ request_body_max_bytes = 1048576
 # initialization.  If you're using this config file, you're probably running a
 # simulated system.  In that case, the initial certificate is provided to the
 # simulated sled agent (acting as RSS) via command-line arguments.
-#tls = true
+tls = true

 [deployment.dropshot_internal]
 # IP Address and TCP port on which to listen for the internal API
```

This is the same diff that I had that actually made me notice this in the first place. Without this fix, this test will hang until someone manually interrupts it (leaving the child processes running) or kills the child processes manually. With this fix, we see:

```
bnaecker@shale : ~/omicron $ cargo t -- test_run_all --nocapture
   Compiling omicron-dev-tools v0.1.0 (/home/bnaecker/omicron/dev-tools)
    Finished test [unoptimized + debuginfo] target(s) in 26.03s
     Running unittests src/lib.rs (target/debug/deps/authz_macros-ef2d1b9a1f28b83f)
     Running tests/test_omicron_dev.rs (target/debug/deps/test_omicron_dev-d4d924e565519942)

<SNIP>

running 1 test
will run: bash -c '( set -o monitor; /home/bnaecker/omicron/target/debug/omicron-dev run-all --nexus-listen-port 0)'
waiting for stdout from child process
subproc stdout: omicron-dev: setting up all services ...
subproc stdout: log file: /tmp/omicron-dev-omicron-dev.25038.0.log
subproc stdout: note: configured to log to "/tmp/omicron-dev-omicron-dev.25038.0.log"
subproc stdout: omicron-dev: services are running.
subproc stdout: omicron-dev: nexus external API:    127.0.0.1:48072
found Nexus external API:    127.0.0.1:48072
subproc stdout: omicron-dev: nexus internal API:    [::1]:12221
subproc stdout: omicron-dev: cockroachdb pid:       25042
found cockroachdb pid:       25042
subproc stdout: omicron-dev: cockroachdb URL:       postgresql://root@127.0.0.1:58694/omicron?sslmode=disable
found postgres listen URL:       postgresql://root@127.0.0.1:58694/omicron?sslmode=disable
subproc stdout: omicron-dev: cockroachdb directory: /tmp/.tmp2qbr5O
found cockroachdb directory: /tmp/.tmp2qbr5O
cleaned up connection
sending SIGINT to process group 25038
wait result: Exited(0)
thread 'test_run_all' panicked at 'failed to run test: Nexus failed to start

Caused by:
    0: Communication Error: error sending request for url (http://127.0.0.1:48072/v1/logout): invalid HTTP version parsed
    1: error sending request for url (http://127.0.0.1:48072/v1/logout): invalid HTTP version parsed
    2: invalid HTTP version parsed', dev-tools/tests/test_omicron_dev.rs:471:9
test test_run_all ... FAILED

failures:

failures:
    test_run_all

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 7 filtered out; finished in 6.44s

error: test failed, to rerun pass `-p omicron-dev-tools --test test_omicron_dev`
```

dev-tools/tests/test_omicron_dev.rs
```diff
@@ -289,66 +289,94 @@ async fn test_db_run() {
...
+            .arg(&dbrun.listen_config_url)
+            .capture()
+            .context("failed to run db-wipe")?;
+        anyhow::ensure!(
+            matches!(wipe_result.exit_status, ExitStatus::Exited(0)),
+            "db-wipe did not exit cleanly",
+        );
+        anyhow::ensure!(
+            !has_omicron_schema(&client).await,
+            "Missing Omicron schema"
```
davepacheco, line 366: On all these `anyhow::ensure` calls, you don't have to provide a string.  If you don't, it will report the exact condition it was checking.  I personally prefer that.  If we want to keep the strings, then this one should be:

```suggestion
            "Unexpectedly found Omicron schema"
```

dev-tools/tests/test_omicron_dev.rs
```diff
@@ -388,25 +417,35 @@ async fn test_run_all() {
...
+            .context("connection failed with an error")?;
+        eprintln!("cleaned up connection");
+
+        // Make sure we can connect to Nexus.
+        let client = oxide_client::Client::new(&format!(
+            "http://{}",
+            runall.external_url
+        ));
+        let _ =
+            client.logout().send().await.context("Nexus failed to start")?;
```
davepacheco, line 445:
```suggestion
            client.logout().send().await.context("unexpectedly failed to reach Nexus logout endpoint")?;
```

(we don't know why this failed -- Nexus may well have started)

davepacheco (approved):
Thanks for debugging and fixing this!  I'm not clear on why the test hung, but the change looks like an improvement anyway.

dev-tools/tests/test_omicron_dev.rs
bnaecker: Good idea, I'll leave the strings out.

dev-tools/tests/test_omicron_dev.rs
bnaecker: True, fixed.

PR #3436, opened 2023-06-27
[wicket/nexus] Additional up-front SSL certificate validation

jgallagher:
This PR adds a couple extra checks on certificates:

1. (Optional) Is the cert suitable for a given hostname?
2. (Required) Is the cert key usage suitable for a web server?

Wicket uses both; nexus only uses 2 (for now, but presumably could be extended to pass a hostname to enable that check too.)

I am a little nervous about a couple things here:

1. We don't want to reject valid certs; I tried to be careful to only reject certs that are definitely invalid or malformed, but would appreciate another set of eyes on this topic.
2. The openssl Rust crate does not expose the extensions necessary to check key usage (https://github.com/sfackler/rust-openssl/issues/373), so I'm pulling in `x509-parser` just for that check.

On 2, I'm a little tempted to replace openssl-rs entirely with x509-parser, but am not sure how others feel about that vs the combo I have in place now.


nexus/db-model/src/certificate.rs
```diff
@@ -47,7 +47,13 @@ impl Certificate {
         params: params::CertificateCreate,
     ) -> Result<Self, CertificateError> {
         let validator = CertificateValidator::default();
-        validator.validate(params.cert.as_bytes(), params.key.as_bytes())?;
+
+        // TODO-security: Could we have a hostname here to pass to `validate()`?
```
jmpesp, line 51: follow up issue?

certificates/src/lib.rs
```diff
@@ -21,30 +23,60 @@ pub enum CertificateError {
...
-    #[error("Unexpected error: {0}")]
-    Unexpected(openssl::error::ErrorStack),
+    #[error("Failed to convert common name to UTF8")]
+    NonUtf8CommonName(#[source] openssl::error::ErrorStack),
+
+    #[error("Certificate subject does not contain a common name")]
+    NoCommonName,
+
+    #[error("Certificate contains no subject alternate name dNSName entries and multiple common names")]
+    NoDnsNameSansMultipleCommonNames,
```
davepacheco, line 38:
```suggestion
    AmbiguousName,
```

(feel free to ignore.  I misread `Sans` to mean "without" and was looking for something less of a mouthful to put here.)

certificates/src/lib.rs
```diff
@@ -108,6 +142,33 @@ impl CertificateValidator {
         // to use with verifying the private key.
         let cert = certs.swap_remove(0);
 
+        if let Some(hostname) = hostname {
+            let dns_names = cert_dns_names(&cert)?;
+            let hostname = DnsName::from_utf8(hostname)
+                .map_err(CertificateError::InvalidDnsName)?;
```
davepacheco, line 148: This seems like kind of a weird one because the invalid DNS name is not in the certificate, but was provided by the caller, right?

certificates/src/lib.rs
```diff
@@ -108,6 +142,33 @@ impl CertificateValidator {
         // to use with verifying the private key.
         let cert = certs.swap_remove(0);
 
+        if let Some(hostname) = hostname {
```
davepacheco, line 145: This logic seems reasonable based on how I expect this would work.  Is there some standard that specifies how it's supposed to work?  I'm a little surprised there's no library to do this.

certificates/src/lib.rs
```diff
@@ -125,3 +186,363 @@ impl CertificateValidator {
...
+    use x509_parser::certificate::X509CertificateParser;
+    use x509_parser::nom::Parser;
+
+    let mut parser = X509CertificateParser::new();
+    let Ok((_remaining, cert)) = parser.parse(der) else {
+        return Ok(());
+    };
+
+    if let Ok(Some(key_usage)) = cert.key_usage() {
+        dbg!(&key_usage);
```
davepacheco, line 208: To be removed?

certificates/src/lib.rs
```diff
@@ -125,3 +186,363 @@ impl CertificateValidator {
...
+        // digitalSignature / keyEncipherment bits set.
+        if !key_usage.value.digital_signature()
+            && !key_usage.value.key_encipherment()
+        {
+            return Err(CertificateError::UnsupportedPurpose);
+        }
+    }
+
+    if let Ok(Some(ext_key_usage)) = cert.extended_key_usage() {
+        dbg!(&ext_key_usage);
```
davepacheco, line 221: Same

certificates/src/lib.rs
```diff
@@ -108,6 +142,33 @@ impl CertificateValidator {
...
+        // Unfortunately, the openssl crate does not expose these
+        // (https://github.com/sfackler/rust-openssl/issues/373), so we'll do
+        // something pretty gross: convert this cert to DER, then re-parse it
+        // via the pure Rust x509-parser crate, which does. We want to err on
+        // the side of "allow the cert" here; we're using a non-openssl crate
+        // and manually checking some SSL extensions, so we will only fail if we
+        // _know_ the cert is not valid.
+        if let Ok(cert_der) = cert.to_der() {
+            validate_cert_der_extended_key_usage(&cert_der)?;
+        }
```
davepacheco, line 170: I am not sure that this check is worth it (but I don't feel strongly we should take it out, either).  I don't think I've ever run into such a certificate.

certificates/src/lib.rs
```diff
@@ -125,3 +186,363 @@ impl CertificateValidator {
...
+//   (including a leading wildcard).
+// * If neither `hostname` nor `dns_name` begin with a wildcard, they must match
+//   exactly.
+// * If `hostname` does not begin with a wildcard but `dns_name` does, the first
+//   portion of each will be removed, and the remainders must match exactly.
+fn dns_name_matches_hostname(dns_name: &DnsName, hostname: &DnsName) -> bool {
+    // If `hostname` starts with a wildcard _or_ `dns_name` does not, the only
+    // way for them to match is if they match exactly. (Note that `DnsName`
+    // provides _case insensitive_ equality, so "match exactly" means "match
+    // exactly but case insentively".)
```
davepacheco, line 308:
```suggestion
    // exactly but case insensitively".)
```

certificates/src/lib.rs
```diff
@@ -125,3 +186,363 @@ impl CertificateValidator {
...
+        0 => Err(CertificateError::NoCommonName),
+        1 => Ok(dns_names),
+        _ => Err(CertificateError::NoDnsNameSansMultipleCommonNames),
+    }
+}
+
+// Returns `Ok(true)` if `dns_name` (from a cert) matches `hostname`, including
+// handling a leading `*` wildcard:
+//
+// * If `hostname` begins with a wildcard, `dns_name` must be an exact match
```
davepacheco, line 298: I read `hostname` here like "an actual DNS name, as might be provided when someone makes an HTTP request to Nexus".  I wouldn't expect it to be possible for that to have wildcards.  Is it really "a hostname pattern", sort of of our own invention, so that we can validate `*.sys.$delegated_domain`?

wicketd/src/rss_config.rs
```diff
@@ -121,6 +122,32 @@ impl CurrentRssConfig {
         if self.external_certificates.is_empty() {
             bail!("at least one certificate/key pair is required");
         }
+
+        // When certs were uploaded we might not have known our external DNS
+        // name, so we skipped validating hostnames. We now _do_ know our
+        // external DNS name, so repeat validation with that hostname.
+        let mut cert_validator = CertificateValidator::default();
+        cert_validator.danger_disable_expiration_validation();
```
davepacheco, line 130: Mention why we're disabling this?

wicketd/src/rss_config.rs
```diff
@@ -121,6 +122,32 @@ impl CurrentRssConfig {
...
+
+        // This _requires_ all certs have wildcard entries; see
+        // https://github.com/oxidecomputer/omicron/issues/3163. We could check
+        // for only `recover.sys.{}` instead if we only want these certs to be
```
davepacheco, line 134:
```suggestion
        // for only `recovery.sys.{}` instead if we only want these certs to be
```

wicketd/src/rss_config.rs
```diff
@@ -121,6 +122,32 @@ impl CurrentRssConfig {
...
             bail!("at least one certificate/key pair is required");
         }
+
+        // When certs were uploaded we might not have known our external DNS
+        // name, so we skipped validating hostnames. We now _do_ know our
+        // external DNS name, so repeat validation with that hostname.
+        let mut cert_validator = CertificateValidator::default();
+        cert_validator.danger_disable_expiration_validation();
+
+        // This _requires_ all certs have wildcard entries; see
```
davepacheco, line 132: I don't think we want to validate that here because pre-3163 this _is_ only used for the Recovery Silo.  I think we just want to validate that it matches the recovery Silo's DNS name.  If we add fleet-wide certificates later, that will still be a valid check (though not the best check we could make).

(sorry for the confusion on this)

certificates/src/lib.rs
```diff
@@ -21,30 +23,60 @@ pub enum CertificateError {
...
     #[error("Certificate and private key do not match")]
     Mismatch,
 
-    #[error("Unexpected error: {0}")]
-    Unexpected(openssl::error::ErrorStack),
+    #[error("Failed to convert common name to UTF8")]
+    NonUtf8CommonName(#[source] openssl::error::ErrorStack),
+
+    #[error("Certificate subject does not contain a common name")]
+    NoCommonName,
```
davepacheco, line 35:
```suggestion
    #[error("Certificate subject contains neither alternate names nor a common name")]
    NoName,
```

(feel free to ignore)

certificates/src/lib.rs
jgallagher: I added this check specifically because @jmpesp generated a cert that fails this check but not the others. :sweat_smile:

certificates/src/lib.rs
jgallagher: Me too! I looked and didn't find one, but maybe I wasn't looking for the right thing. OpenSSL has [`X509_check_host`](https://www.openssl.org/docs/manmaster/man3/X509_check_host.html), which is (I think?) what we want, but AFAICT it isn't exposed by the rust openssl bindings, and I can't find an equivalent in `rustls` or `x509-parser`.

certificates/src/lib.rs
jgallagher: Ah, yes, the error message we'd return here would be incorrect. I'll make this its own error case.

certificates/src/lib.rs
jgallagher: That is how I used it from wicket (to check that the cert uploaded would work for `*.sys.$delegated_domain`, exactly). If wicketd should only be checking for `recovery.sys.$delegated_domain`, I could change this to only match actual hostnames?

certificates/src/lib.rs
luqmana: seems like `rustls` may use `webpki` for this https://github.com/rustls/rustls/blob/c21eca793b90c74b8c6710416a66a56266abae6f/rustls/src/webpki/verify.rs#L102-L120

jgallagher: I looked at `webpki` a bit, and I think it would have similar downsides: we have to re-parse the cert, and the other checks we want to do aren't really supported by `webpki` because we don't have a root cert to pass it for validation.

I realized we can also call `X509_check_host` directly; I'm going to close this PR and rework it to only use OpenSSL and will reopen.

jgallagher:
This is both stale and needs rework; closing this PR in favor of a fresh (and fairly different) implementation.

PR #3437, opened 2023-06-27
ci/tuf-repo: Consistent artifact versions

### LINKED ISSUE #3430: the CI-generated TUF repo should have consistent version stamping
What I am presently working on is to make the artifact version as stored in a TUF repo's `artifacts.json` match the version that is stamped in that artifact.

For things that have an oxide.json at the root of a tarball (zone images and OS images), this means making sure the artifact version (and ideally commit hash) is stamped in oxide.json.

For Hubris SP/ROT images, this means extracting the `VERS` field out of the caboose and using that in artifacts.json.

I am tentatively putting this on the MVP milestone because I don't think "TUF and the artifacts agree on version numbers" stops us from doing MUPdate?

cc @smklein @jgallagher

iliana:
Fixes #3430. Changes:
- Changes the prerelease field in our version from `alpha` to `0.ci`, which comes before alpha, so that we can use `alpha` or other words found early in the dictionary and have them count as later than a CI build.
- Stamps a non-placeholder version into the global zone packages installed in the host and trampoline OS images.
- Adds `caboose-util`, loosely based on the small part of [https://github.com/oxidecomputer/hubtools](https://github.com/oxidecomputer/omicron/pull/hubedit) I needed, tweaked for usefulness to a shell script.
- Reads the version out of a Hubris archive's caboose and uses that for the TUF repository, instead of the control plane version we generate here.

### DIFF
--- a/.github/buildomat/jobs/tuf-repo.sh
+++ b/.github/buildomat/jobs/tuf-repo.sh
@@ -44,14 +44,12 @@ set -o pipefail
 set -o xtrace
 
 TOP=$PWD
+VERSION=$(< /input/package/work/version.txt)
 
 source "$TOP/tools/dvt_dock_version"
 DVT_DOCK_COMMIT=$COMMIT
 source "$TOP/tools/hubris_version"
 HUBRIS_COMMIT=$COMMIT
-
-COMMIT=$(git rev-parse HEAD)
-VERSION="1.0.0-alpha+git${COMMIT:0:11}"
 
 ptime -m gunzip < /input/ci-tools/work/tufaceous.gz > /work/tufaceous
 chmod a+x /work/tufaceous
--- a/.github/buildomat/jobs/tuf-repo.sh
+++ b/.github/buildomat/jobs/tuf-repo.sh
@@ -46,13 +46,10 @@ set -o xtrace
 TOP=$PWD
 VERSION=$(< /input/package/work/version.txt)
 
-source "$TOP/tools/dvt_dock_version"
-DVT_DOCK_COMMIT=$COMMIT
-source "$TOP/tools/hubris_version"
-HUBRIS_COMMIT=$COMMIT
-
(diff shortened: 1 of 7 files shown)


caboose-util/src/main.rs
```diff
@@ -0,0 +1,38 @@
...
+use tlvc_text::{Piece, Tag};
+
+fn main() -> Result<()> {
+    let mut args = std::env::args().skip(1);
+    match args.next().context("subcommand required")?.as_str() {
+        "read-version" => {
+            let archive = RawHubrisArchive::load(
+                &args.next().context("path to hubris archive required")?,
+            )?;
+            let caboose = archive.read_caboose()?;
```
jgallagher, line 18: I think you could use [`caboose.version()`](https://github.com/oxidecomputer/hubtools/blob/50de6c304df47dfaa19d15c170f4d7980341d2af/hubtools/src/caboose.rs#L43-L45) at this point instead of manually decoding the TLVC content.

.github/buildomat/jobs/tuf-repo.sh
```diff
@@ -113,31 +108,41 @@ EOF
...
+(cd /work/dvt-dock; git checkout "$COMMIT")
 
 for noun in gimlet psc sidecar; do
     tufaceous_kind=${noun//sidecar/switch}_rot
     hubris_kind=${noun}-rot
+    path_a="/work/dvt-dock/staging/build-$hubris_kind-image-a-cert-dev.zip"
+    path_b="/work/dvt-dock/staging/build-$hubris_kind-image-b-cert-dev.zip"
+    version_a=$(/work/caboose-util read-version "$path_a")
+    version_b=$(/work/caboose-util read-version "$path_b")
+    if [[ "$version_a" != "$version_b" ]]; then
```
jgallagher, line 122: Good call :+1:

jgallagher (approved):

caboose-util/src/main.rs
iliana: how did i miss this!

PR #3446, opened 2023-06-28
RSS support for deploying multiple instances of a component on a Sled

davepacheco:
Based on #3390 because that changes a bunch of the same code and I expect it to land sooner than this.

This PR moves us closer to being able to provision multiple CockroachDB zones, even on the same host (e.g., in development and CI) by reworking how RSS assigns components to Sleds to be component-oriented instead of sled-oriented.

I had initially thought I'd do some more work in this PR but I think I want to separate that into follow-on PRs.


davepacheco:
Sorry for the force push.  The change is about the same, but it was much easier to rebase this one than resolve a bunch of the merge conflicts.

sled-agent/src/rack_setup/plan/service.rs
```diff
@@ -230,224 +233,322 @@ impl Plan {
...
+                            subnet,
+                            sled_address,
+                            u2_zpools,
+                            is_scrimlet,
+                        ))
+                    },
+                ))
+                .await;
+            result?
+        };
```
smklein, line 267: Should we just make this `Vec<SledInfo>` a required input to this function?

If we do so, I don't *think* this function tries to contact sled agents directly, and it might be easier to write tests using a "TransientDnsServer"

smklein (approved):
Looks great, thanks for improving this.

I think with your addition of `SledInfo`, we could make a really minor refactor to make the plan-provisioning testable!

sled-agent/src/rack_setup/plan/service.rs
davepacheco: Yes, I think that's a great idea.

PR #3450, opened 2023-06-29
deploy 5 CockroachDB nodes from RSS

### LINKED ISSUE #727: [sled-agent] Initialize multi-node CockroachDB, rather than single-node

davepacheco:
Depends on #3446.  Replaces #2956.  Fixes #727.

This changes RSS to deploy 5 CockroachDB nodes and sets the default replication factor to 5.  Quoting the [CockroachDB production checklist](https://www.cockroachlabs.com/docs/v23.1/recommended-production-settings#topology):

> When deploying in a single availability zone:...To be able to tolerate 2 simultaneous node failures, use at least 5 nodes and [increase the default replication factor for user data](https://www.cockroachlabs.com/docs/v23.1/configure-replication-zones#edit-the-default-replication-zone) to 5. The replication factor for [important internal data](https://www.cockroachlabs.com/docs/v23.1/configure-replication-zones#create-a-replication-zone-for-a-system-range) is 5 by default, so no adjustments are needed for internal data. In this case, if 2 nodes fail at the same time, each range retains 3 of its 5 replicas, a majority.


davepacheco:
This needs a bit more work.  Sled Agent makes some assumptions that either need to be relaxed or worked around (e.g., that there will only be one CockroachDB dataset in a given zpool).

smklein:
> This needs a bit more work. Sled Agent makes some assumptions that either need to be relaxed or worked around (e.g., that there will only be one CockroachDB dataset in a given zpool).

Is this still an issue? I saw you bumped the number of zpools in the simulated environment from three to five, which looks like the right call to me

davepacheco:
> > This needs a bit more work. Sled Agent makes some assumptions that either need to be relaxed or worked around (e.g., that there will only be one CockroachDB dataset in a given zpool).
> 
> Is this still an issue? I saw you bumped the number of zpools in the simulated environment from three to five, which looks like the right call to me

No, this is fixed with a change to the dependent PR (#3446) to allocate from separate zpools when possible, plus a change in this PR to set up more virtual zpools.

tools/create_virtual_hardware.sh
```diff
@@ -47,7 +47,7 @@ function ensure_zpools {
             echo "Zpool: [$ZPOOL]"
             VDEV_PATH="$OMICRON_TOP/$ZPOOL.vdev"
             if ! [[ -f "$VDEV_PATH" ]]; then
-                dd if=/dev/zero of="$VDEV_PATH" bs=1 count=0 seek=10G
+                dd if=/dev/zero of="$VDEV_PATH" bs=1 count=0 seek=6G
```
davepacheco, line 50: I tried to keep the total space used by the zpools the same size.  It used to be 3x10G.  Now it's 5x6G.

smf/sled-agent/non-gimlet/config.toml
```diff
@@ -18,18 +18,23 @@ sidecar_revision.soft = { front_port_count = 1, rear_port_count = 1 }
...
+# different vdev for each.  The `create_virtual_hardware.sh` script does this
+# for you.
 zpools = [
   "oxi_a462a7f7-b628-40fe-80ff-4e4189e2d62b",
   "oxi_b462a7f7-b628-40fe-80ff-4e4189e2d62b",
   "oxp_d462a7f7-b628-40fe-80ff-4e4189e2d62b",
   "oxp_e4b4dc87-ab46-49fb-a4b4-d361ae214c03",
   "oxp_f4b4dc87-ab46-49fb-a4b4-d361ae214c03",
+  "oxp_14b4dc87-ab46-49fb-a4b4-d361ae214c03",
+  "oxp_24b4dc87-ab46-49fb-a4b4-d361ae214c03",
```
davepacheco, line 37: We need two more pools in the system in order to deploy a 5-node CockroachDB cluster.

end-to-end-tests/src/helpers/ctx.rs
```diff
@@ -240,7 +240,7 @@ pub async fn build_client() -> Result<oxide_client::Client> {
             })
         },
         &Duration::from_secs(1),
-        &Duration::from_secs(300),
+        &Duration::from_secs(600),
```
davepacheco, line 243: We're deploying two extra CockroachDB nodes and it can take a little longer for the system to set up now.

smklein: I can look into parallelizing service bringup on the sled-agent side, if that would help mitigate?

common/src/sql/dbinit.sql
```diff
@@ -34,6 +34,12 @@ CREATE DATABASE omicron;
 CREATE USER omicron;
 ALTER DEFAULT PRIVILEGES GRANT INSERT, SELECT, UPDATE, DELETE ON TABLES to omicron;
 
+/*
+ * Configure a replication factor of 5 to ensure that the system can maintain
+ * availability in the face of any two node failures.
+ */
+ALTER RANGE default CONFIGURE ZONE USING num_replicas = 5;
```
smklein, line 41: For local testing CRDB -- e.g. with the omicron-dev tool -- what does this do?

smklein (approved):
Looks great!

end-to-end-tests/src/helpers/ctx.rs
davepacheco: I expect it would.  No pressure on my part in terms of urgency.

smklein: Sounds good, lets not block this PR, but I can do this as a follow-up.

common/src/sql/dbinit.sql
davepacheco: Good question.  I expect it will have no practical impact.  The default replication factor is 3, which is already more nodes than we ever have today.  The impact is that CockroachDB considers all of its data under-replicated.  It reports it as such (e.g., [here](https://www.cockroachlabs.com/docs/stable/query-replication-reports.html#find-out-which-databases-and-tables-have-under-replicated-ranges)) and if new nodes show up, it will try to replicate data until nothing is under-replicated.

One could imagine that this would instead disallow you from writing any data to the cluster when it can't satisfy this constraint.  That doesn't seem to be how it works.  If it did, the tests wouldn't run at all, I think.

Issue #3441, opened 2023-06-28
MUPdate: Seeing corrupted host OS images after update


smklein:
After a mupdate, I'm seeing all sleds back in A0 (as expected) but the ones in cubby 9 and 11 are not showing up in `pilot host ls`.

From the serial console of cubby 9:

```
           WARNING: checksum mismatch
NOTICE:     want: df81ccd1e6e9d4169d56636e50d897116c756c23cf012dffdeaf88742b3d015d
NOTICE:     have: cd1fa165d8beb73a240fe79e387b2ab72670a4ffe364b867b5a95e4bdd1de5af
WARNING: boot image integrity failure want df81ccd1e6e9d4169d56636e50d897116c756c23cf012dffdeaf88742b3d015d got cd1fa165d8beb73a240fe79e387b2ab72670a4ffe364b867b5a95e4bdd1de5af

-----------> Sending IPCC command 0x6, attempt 1/10
Additional data length: 0xa9
Received empty frame
Additional data length: 0x0

panic[cpu0]/thread=fffffffffbe9f1e0: boot image integrity failure want df81ccd1e6e9d4169d56636e50d897116c756c23cf012dffdeaf88742b3d015d got cd1fa165d8beb73a240fe79e387b2ab72670a4ffe364b867b5a95e4bdd1de5af

Warning - stack not written to the dump buffer
fffffffffbea5060 boot_image:oxide_boot_locate+380 ()
fffffffffbea5090 genunix:boot_image_locate+a1 ()
fffffffffbea50d0 genunix:main+137 ()
fffffffffbea50e0 unix:_locore_start+8
```

Seeing something similar in cubby 11:
```
NOTICE: TRYING: boot disk (slot 17, slice 0)
NOTICE: found M.2 device (slot 17, slice 0), @ /pci@0,0/pci1de,fff9@1,3/pci1344,3100@0/blkdev@w00A075013280E36C,0:a
NOTICE:     in image: df81ccd1e6e9d4169d56636e50d897116c756c23cf012dffdeaf88742b3d015d
NOTICE: attempting boot from image name 'ci 6d5da99/46ea833 2023-06-27 17:23'
WARNING: checksum mismatch
NOTICE:     want: df81ccd1e6e9d4169d56636e50d897116c756c23cf012dffdeaf88742b3d015d
NOTICE:     have: 6f427a8268101361097a314a9e3544a880f257355661a28641b9ac15fe038e5a
[...]

PR #3455, opened 2023-06-29
[installinator] Add checks on sha256 of artifacts

jgallagher:
This PR adds two checksum checks:

1. After downloading artifacts from wicketd, confirm the hash of the data we received matches the hash we requested. (This applies to both host OS and control plane artifacts.)
2. After writing the host OS to disk, reread the file and hash it, ensuring it matches the host phase 2 hash we downloaded (and checked in step 1).

We could do a similar check on each control plane file, but we'd have to hash them all individually to know what to expect.

If/when we see #3441 again, this should provide an extra data point that at least when installinator was alive, it believed the host OS had been successfully written to disk.

On `madrid`, this step takes adds about 11 seconds (per slot) to the time to mupdate:

validate-host


installinator/src/dispatch.rs
```diff
@@ -203,6 +204,13 @@ impl InstallOpts {
                         fetch_artifact(&cx, &host_phase_2_id, discovery, log)
                             .await?;
 
+                    check_downloaded_artifact_hash(
+                        "host phase 2",
+                        host_phase_2_artifact.artifact.clone(),
+                        host_phase_2_id.hash,
+                    )
+                    .await?;
```
sunshowers, line 212: Do we want to retry the operation if it fails? (If the answer is "not for now, maybe in the future", could you add a comment about it here?)

sunshowers (approved):

installinator/src/dispatch.rs
jgallagher: I think not, maybe ever? We're fetching this data from wicketd (in memory) over TCP to installinator (in memory), so I think the by-far-most-likely reason for this to fail is if wicketd sends the wrong artifact, which I wouldn't expect to be resolved by a retry (and we would definitely want to know immediately if it happened). I'll add a comment though :+1:.

jgallagher: Added explanatory comments in [8ea3902](https://github.com/oxidecomputer/omicron/pull/3455/commits/8ea3902187af7918fd2e469becca7e7dcb9d604f)

installinator/src/write.rs
```diff
@@ -385,24 +387,101 @@ impl<'a> SlotWriteContext<'a> {
...
+        })
+        .await
+        .unwrap()
+        .map_err(WriteError::ChecksumValidationError)?;
+
+        if computed_hash == self.artifacts.host_phase_2_id.hash {
+            StepSuccess::new(())
+                .with_message(format!(
+                    "validated hash {computed_hash} \
+                     for host phase 2 written to {slot:?}"
```
leftwo, line 475: Thank you for this!

Issue #3454, opened 2023-06-29
Sled Agent crashed in response to new instance request


smklein:
The downstairs were wiped out when the sled-agent crashed and restarted:
```
09:00:01.266Z INFO SledAgent (PortManager): Mapping virtual NIC to physical host
    mapping = SetVirtualNetworkInterfaceHost { virtual_ip: 172.30.0.5, virtual_mac: MacAddr(MacAddr6([168, 64, 37, 243, 152, 219])), physical_host_ip: fd00:1122
:3344:10a::1, vni: Vni(10225803) }
09:00:01.267Z INFO SledAgent (dropshot (SledAgent)): request completed
    local_addr = [fd00:1122:3344:107::1]:12345
    method = PUT
    remote_addr = [fd00:1122:3344:102::4]:43840
    req_id = bcc4fef9-789c-4e73-a1f7-2f249ecde50c
    response_code = 204
    uri = /v2p/b7e955a4-36e3-4d74-ae3b-ad053ae8097b
09:00:01.816Z INFO SledAgent (InstanceManager): Adding address: Static(V6(Ipv6Network { addr: fd00:1122:3344:107::2a, prefix: 64 }))
    instance_id = f1e6ed32-cb42-4b71-a7ac-893ac46467f1
    zone = oxz_propolis-server_d00f74ec-80ea-4419-80e9-ec9b3bbf83f4
09:00:02.006Z ERRO SledAgent (InstanceManager): instance setup failed: Err(ZoneEnsureAddress(EnsureAddressError(EnsureAddressError { zone: "oxz_propolis-server_
d00f74ec-80ea-4419-80e9-ec9b3bbf83f4", request: Static(V6(Ipv6Network { addr: fd00:1122:3344:107::2a, prefix: 64 })), name: AddrObject { interface: "oxControlIn
stance8", name: "omicron6" }, err: Zone execution error: Command [/usr/sbin/zlogin oxz_propolis-server_d00f74ec-80ea-4419-80e9-ec9b3bbf83f4 /usr/sbin/ipadm crea
te-addr -t -T addrconf oxControlInstance8/ll] executed and failed with status: exit status: 1  stdout: 
      stderr: ipadm: Could not create address: Addrconf already in progress
    
    Caused by:
[...]

PR #3456, opened 2023-06-29
[sled-agent] Integrate self-assembling Propolis

smklein:
Integrates https://github.com/oxidecomputer/propolis/pull/454 

This avoid using `zlogin` to modify the propolis zone after launch, instead preferring to write a profile file to the zone ahead-of-time. This matches the pattern for other self-assembling zones, where configuration information is supplied ahead-of-time through the `ProfileBuilder`, which places an auxiliary SMF manifest into `/var/svc/profile/site.sml` to be imported when the zone starts.

Relies on: https://github.com/oxidecomputer/propolis/pull/454
Part of the fix for https://github.com/oxidecomputer/omicron/issues/3454


davepacheco (approved):
Love the SMF profile approach over zlogin!

PR #3460, opened 2023-06-30
Use `IpRange` for address lot blocks

zephraph:
This removes the `AddressLotBlockCreate` struct and to reuses `IpRange`. Address lot blocks and ip pool ranges are both a contiguous ordered range of IP addresses. Given that, it makes sense to reuse the same data structure and validation logic when creating both.


nexus/tests/integration_tests/address_lots.rs
```diff
@@ -129,66 +132,71 @@ async fn test_address_lot_basic_crud(ctx: &ControlPlaneTestContext) {
...
+//     params.push(AddressLotCreate {
+//         identity: IdentityMetadataCreateParams {
+//             name: "family".parse().unwrap(),
+//             description: "an address parking lot".into(),
+//         },
+//         kind: AddressLotKind::Infra,
+//         blocks: vec![IpRange::try_from((
+//             "203.0.113.10".parse().unwrap(),
+//             "fd00:1701::d".parse().unwrap(),
+//         ))
```
ahl, line 152: it's nice that this is structurally invalid

nexus/tests/integration_tests/address_lots.rs
```diff
@@ -129,66 +132,71 @@ async fn test_address_lot_basic_crud(ctx: &ControlPlaneTestContext) {
...
-        )
-        .authn_as(AuthnMode::PrivilegedUser)
-        .execute()
-        .await
-        .unwrap_or_else(|_| panic!("unexpected success for: {:#?}", params))
-        .parsed_body::<dropshot::HttpErrorResponseBody>()
-        .unwrap();
-    }
-}
+// TODO: Figure out a different way to test this
```
ahl, line 135: is it just the first part of the test that doesn't work?

nexus/types/src/external_api/params.rs
```diff
@@ -1242,17 +1245,7 @@ pub struct AddressLotCreate {
...
-
-/// Parameters for creating an address lot block. Fist and last addresses are
-/// inclusive.
-#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema)]
-pub struct AddressLotBlockCreate {
-    /// The first address in the lot (inclusive).
-    pub first_address: IpAddr,
-    /// The last address in the lot (inclusive).
-    pub last_address: IpAddr,
+    pub blocks: Vec<IpRange>,
```
ahl, line 1248: better but still problematic, right? I think we need to change this API to mimic what we do for ip-pools

ahl (approved):
this is great; it would be great to land this ASAP

zephraph:
I dropped the ball on this one, sorry about that. @david-crespo this might be viable as some cleanup, but given how long it's been it might be pretty painful. I'll defer to you if we should just close it.

PR #3410, opened 2023-06-24
Initial docs for gimlet-on-a-bench

leftwo:
Starter doc for how to run a gimlet on a bench.

Lets make this better!

PR #3465, opened 2023-07-01
overhaul how-to-run docs

davepacheco:
While going through the how-to-run instructions to set up Omicron on a lab machine, I found a lot of cases where the docs were incomplete, self-inconsistent, outdated, or otherwise misleading.  Mostly I felt it would benefit from a lot more context (especially around external networking) and a concrete example that's carried the whole way through.  So:

- I basically overhauled the whole structure of the how-to-run doc.
- The boundary-services-a-to-z doc was mostly (but not entirely) redundant with what was/is now in how-to-run, so I trimmed all that out but left some detail that seemed useful.
- I added a lot more comments to the config-rss.toml files, which I also notice are identical, yet diverged in ways that seem likely wrong.  I suspect nobody has used `gimlet-standalone` in quite a while and I wonder if we should remove it altogether.  Even the proposed instructions in #3410 seem not to use it?

I hope this is at least a good direction.  (If not, that'd be useful feedback.)  There's a lot more to be improved here and a lot that's still out of date.  For example, these instructions make direct calls the API where I imagine we could do these things with the CLI at this point.  I'd welcome any contributions to this PR -- feel free to push directly to it!

### DIFF
--- a/docs/boundary-services-a-to-z.adoc
+++ b/docs/boundary-services-a-to-z.adoc
@@ -1,49 +1,12 @@
 = Boundary Services A-Z
 
-This document describes how to run an environment with boundary services.
-It's a quick rundown and assumes knowledge of the basic setup described in the
-Running Omicron (Non-Simulated) document.
+NOTE: The instructions for _deploying_ SoftNPU with Omicron have been folded into xref:how-to-run.adoc[the main how-to-run docs].
 
-== 0. Download SoftNPU ASIC emulator machinery
-
-Grab the SoftNPU (ASIC simulator, scadm, P4 program, etc.) from CI:
-
-----
-./tools/ci_download_softnpu_machinery
-----
-
-(By default `install_runner_prerequisites.sh` will take care of running
-`ci_download_softnpu_machinery` for you.)
-
-== 1. Setup virtual hardware
-
-----
-PHYSICAL_LINK=<wan interface> pfexec ./tools/create_virtual_hardware.sh
-----
-Note that the `PHYSICAL_LINK` environment variable is optional. If not supplied,
-the first link in `dladm show-phys` will be used.
-
-The virtual hardware is a bit different than what has previously been used. What
-we now have looks like this.
+The virtual hardware making up SoftNPU is a bit different than what was previously used. What we now have looks like this.
 
 image::plumbing.png[]
 
-The `softnpu` zone will be configured and launched during the `create_virtual_hardware.sh`
-script.
-
-== 2. Build and install the control plane.
-
-----
-./tools/create_self_signed_cert.sh
-
-cargo build --release --bin omicron-package
-./target/release/omicron-package -t <target_name> target create -i standard -m non-gimlet -s softnpu
-./target/release/omicron-package package
-pfexec ./target/release/omicron-package install
-----
-
-The control plane is now starting, reference the Running Omicron (Non-Simulated)
-doc for more details on determining when things are ready to go.
+The `softnpu` zone will be configured and launched during the `create_virtual_hardware.sh` script.
 
 Once the control plane is running, `softnpu` can be configured via `dendrite`
 using the `swadm` binary located in the `oxz_switch` zone.
@@ -105,12 +68,10 @@ host                            mac                age
 fe80::aae1:deff:fe00:1          a8:e1:de:00:00:01  0s
 ----
 
-== 4. Populating the system
-
-Follow the
+While following
 https://github.com/oxidecomputer/omicron/blob/main/docs/how-to-run.adoc[how-to-run.adoc]
-to set up IPs, images, disks, instances etc. Things to pay particular attention
-to here are the following.
+to set up IPs, images, disks, instances etc, pay particular attention to the
+following.
 
 - The address range in the IP pool should be on a subnet in your local network that
   can NAT out to the Internet.
@@ -129,9 +90,7 @@ pfexec /opt/oxide/softnpu/stuff/scadm \
   $softnpu_mac
 ----
 
-== 5. Configuring scrimlet/sidecar
-
-A this point we have an instance up and running with external connectivity
+By the end, we have an instance up and running with external connectivity
 configured via boundary services:
 ----
 ry@korgano:~/omicron$ ~/propolis/target/release/propolis-cli --server fd00:1122:3344:101::c serial
--- a/docs/how-to-run.adoc
+++ b/docs/how-to-run.adoc
@@ -98,14 +98,14 @@ As an example, in this doc:
 * You can carve off 192.168.1.20 - 192.168.1.29 for externally-facing services provided by the system.
 * You can carve off 192.168.1.30 as the address used by SoftNPU on the external network.
 * You can carve off 192.168.1.31 - 192.168.1.40 as the addresses used for Instances that need some public IPs (even if just for NAT to the internet).
-2. Alternatively, you'll set up an "external" network that only exists on your test machine.  If you go this route, we'll choose 192.168.1.0/24 and all the same other details as in the case above, just for convenience.  In this mode, you'll need create your made-up network and give the global zone an IP address on it.  We'll use 192.168.1.19:
+2. Alternatively, you'll set up an "external" network that only exists on your test machine.  If you go this route, we'll choose 192.168.1.0/24 and all the same other details as in the case above, just for convenience.  In this mode, you'll need create your made-up network and give the global zone an IP address on it.  We'll use 192.168.1.199:
 +
 [source,text]
 ----
 # pfexec dladm create-etherstub -t fake_external_stub0
 $ pfexec dladm create-vnic -t -l fake_external_stub0 fake_external0
 $ pfexec ipadm create-if -t fake_external0
-$ pfexec ipadm create-addr -t -T static --address 192.168.1.19 fake_external0/external
+$ pfexec ipadm create-addr -t -T static --address 192.168.1.199 fake_external0/external
 ----
 
 Other network configurations are possible but beyond the scope of this doc.
@@ -355,7 +355,7 @@ If your external network is one you made up above, you'll need to override the g
 
 [source,console]
 ----
-$ GATEWAY_IP=192.168.1.19 ./tools/scrimlet/softnpu-init.sh
+$ GATEWAY_IP=192.168.1.199 ./tools/scrimlet/softnpu-init.sh
 ----
 
 You can also override GATEWAY_MAC if needed, but that shouldn't be necessary for the configurations described here.
--- a/docs/how-to-run.adoc
+++ b/docs/how-to-run.adoc
@@ -33,7 +33,7 @@ Here's a summary of the tradeoffs:
 
 |CockroachDB
 |Real implementation, 1-node cluster
-|Real implementation, multi-node cluster (coming soon)
+|Real implementation, multi-node cluster
 
 |Internal/External DNS
 |Real implementation
@@ -83,7 +83,7 @@ There are other configurations besides the ones mentioned here:
 
 * A real Oxide rack: real hardware with real implementations of all components.
 * A Gimlet in a real Oxide rack, deployed as a single system in isolation from the rest of the rack.
-* A Gimlet on a bench (i.e., not in a rack -- requires separate hardware to connect power and networking) connected a Sidecar (the Oxide switch, also on a bench)
+* A Gimlet on a bench (i.e., not in a rack -- requires separate hardware to connect power and networking) connected a Sidecar (the Oxide switch), also on a bench.
 * A Gimlet on a bench with no Sidecar.
 
 This document doesn't describe how to deploy these cases.
--- a/docs/how-to-run.adoc
+++ b/docs/how-to-run.adoc
@@ -2,24 +2,131 @@
 :toc: left
 :icons: font
 
-= Running Omicron (Non-Simulated)
+= Running Omicron
 
-Omicron is the control plane for an Oxide rack. It expects to execute
-on Helios systems, and Sleds use Helios-specific interfaces to manage
-resources.
+Omicron is the control plane for the Oxide system.  There are a few different ways to run Omicron depending on what resources you have available and how much of the stack you want to run using real components vs. simulated ones.  Generally speaking, it's easier to get things going and automate testing with simulated components, but of course much of the system's functionality is missing or faked up when using simulated components.
 
-If you're interested in running the control plane on other platforms, including
-Linux and Mac, refer to the guide on xref:how-to-run-simulated.adoc[running
-simulated Omicron].
+The most common configurations are:
 
-== Installing Prerequisite Software
+1. Using one or more simulated Sled Agents.  There are no real VMs.  All components listen on localhost and talk to each other directly.  The automated tests in this repo generally use this kind of deployment.  This is the only mode that's supported on non-illumos systems (i.e., Linux and MacOS).
+2. Using one or more real Sled Agents (on separate physical machines) using SoftNPU-based Boundary Services.  You can provision real VMs this way.  Intra-"rack" networking and boundary Services (i.e., external connectivity) are fully functional using SoftNPU, a software-based simulation of the Tofino device that's normally responsible for these functions.
++
+Within this mode, you can run on real Oxide hardware ("Gimlets") or ordinary server PCs.
 
-A major prerequisite is to have a machine already running Helios. An easy way to
-do this is by using a https://github.com/oxidecomputer/helios-engvm[Helios VM].
-ISOs are also available for download https://pkg.oxide.computer/install[here].
+These are historically called "simulated" vs. "non-simulated", though they both involve simulation of some parts of a real Oxide system.
 
-The steps below will install several executables that will need to be in your
-`PATH`.  You can set that up first using:
+Here's a summary of the tradeoffs:
+
+[cols="1h,1,1", options="header"]
+|===
+|
+|"Simulated" deployment
+|"Non-simulated" deployment
+
+|Works on
+|illumos, Linux, MacOS
+|illumos only
+
+|Nexus
+|Real implementation
+|Real implementation
+
+|CockroachDB
+|Real implementation, 1-node cluster
+|Real implementation, multi-node cluster (coming soon)
+
+|Internal/External DNS
+|Real implementation
+|Real implementation
+
+|Crucible Pantry
+|Real implementation
+|Real implementation
+
+|Sled Agent
+|Simulated, multiple okay
+|Real implementation, requires separate machines for each
+
+|Propolis
+|Missing (VMs are faked-up by simulated Sled Agent)
+|Real implementation
+
+|Dendrite
+|Stub implementation
+|Simulated implementation
+
+|Management Gateway Service (MGS)
+|Missing
+|Real implementation
+
+|Internal ("rack") networking
+|Localhost
+|Real implementation on-sled, simulated implementation across sleds (SoftNPU)
+
+|Boundary services (external connectivity)
+|Localhost
+|Simulated implementation (SoftNPU)
+
+|Wicket
+|Missing
+|Not used
+
+|Management network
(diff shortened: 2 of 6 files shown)


docs/boundary-services-a-to-z.adoc
```diff
@@ -1,49 +1,12 @@
 = Boundary Services A-Z
 
-This document describes how to run an environment with boundary services.
-It's a quick rundown and assumes knowledge of the basic setup described in the
-Running Omicron (Non-Simulated) document.
+NOTE: The instructions for _deploying_ SoftNPU with Omicron have been folded into xref:how-to-run.adoc[the main how-to-run docs].
```
davepacheco, line 3: @rcgoodfellow I mentioned this in the PR description but I think most (but not all) of this doc is now covered by how-to-run so I trimmed this a lot but left the stuff that went a little deeper than what's in how-to-run.  What do you think?

smf/sled-agent/non-gimlet/config-rss.toml
```diff
@@ -22,25 +23,82 @@ ntp_servers = [ "ntp.eng.oxide.computer" ]
...
+# With softnpu, only one address is used.
+infra_ip_first = "192.168.1.30"
+infra_ip_last = "192.168.1.30"
+# Name of the uplink port.  This should always be "qsfp0" when using softnpu.
+uplink_port = "qsfp0"
+uplink_port_speed = "40G"
+uplink_port_fec="none"
+# For softnpu, an address within the "infra" block above that will be used for
+# the softnpu uplink port.  You can just pick the first address in that pool.
+uplink_ip = "192.168.1.30"
```
davepacheco, line 77: @rcgoodfellow, @luqmana could you take a look at the comments I've added here?  Other changes here include moving some sections around so that the stuff people are likely to need to change are earlier in the file and changing the ranges in the internal service pool, infra pool, and the uplink_ip based on various suggestions in chat today.

smf/sled-agent/gimlet-standalone/config-rss.toml
```diff
@@ -22,25 +23,82 @@ ntp_servers = [ "ntp.eng.oxide.computer" ]
...
+# With softnpu, only one address is used.
+infra_ip_first = "192.168.1.30"
+infra_ip_last = "192.168.1.30"
+# Name of the uplink port.  This should always be "qsfp0" when using softnpu.
+uplink_port = "qsfp0"
+uplink_port_speed = "40G"
+uplink_port_fec="none"
+# For softnpu, an address within the "infra" block above that will be used for
+# the softnpu uplink port.  You can just pick the first address in that pool.
+uplink_ip = "192.168.1.30"
```
davepacheco, line 77: @rcgoodfellow this section was missing from gimlet-standalone.  I assumed it was an oversight, but let me know if not.

docs/how-to-run.adoc
```diff
@@ -51,126 +156,92 @@ $ ./tools/install_runner_prerequisites.sh
...
+uplink_port_speed = "40G"
+uplink_port_fec="none"
+# For softnpu, an address within the "infra" block above that will be used for
+# the softnpu uplink port.  You can just pick the first address in that pool.
+uplink_ip = "192.168.1.30"
 ----
 
 Update your `smf/sled-agent/$MACHINE/config.toml` if necessary:
+
+TODO how does one know if this is necessary and if so, what to put here?
```
davepacheco, line 243: @rcgoodfellow could you help flesh this out?  If it's easier, I'm happy to get a brain dump from you and then write up a summary here.

docs/how-to-run.adoc
```diff
@@ -2,24 +2,131 @@
...
 
-The steps below will install several executables that will need to be in your
-`PATH`.  You can set that up first using:
+== Prerequisites
+
+=== Bare-metal operating system
+
+This guide assumes that you have a system running https://github.com/oxidecomputer/helios[Helios] on a bare-metal system.footnote:[You can in principle use a VM, but you wouldn't be able to provision Instances because nested virtualization is not supported.]
+
+=== External networking
```
davepacheco, line 88: @rcgoodfellow @luqmana @internet-diglett (who did I miss?) I'd be interested in your feedback on this section in particular.  This sets up the example that we use for the rest of this doc (and in config-rss.toml).  Hopefully I got the details right!

davepacheco:
I only realized most of the way through writing this that it overlaps somewhat with #3410.  @leftwo maybe we could chat about how best to combine this work (if at all)?

docs/boundary-services-a-to-z.adoc
rcgoodfellow: Yeah, this looks good to me. No need to keep the redundant prose around.

docs/how-to-run.adoc
rcgoodfellow: This looks good to me! This is a great primer for what the address configuration needs to look like. It makes me think that we should figure out a way to just put all this in one config file upfront to reduce the configuration complexity spread out over several different files.

docs/how-to-run.adoc
rcgoodfellow: I don't think any modifications to this file are necessary for basic Omicron development environments.

The `switch_zone_maghemite_links` field is necessary if you are running a Canada-region style setup where you plumb physical interfaces into the switch zone, and `ddmd` in that zone needs to actually peer with other physical machines on your lab network. @jmpesp will likely have more insight here.

smf/sled-agent/gimlet-standalone/config-rss.toml
rcgoodfellow: This looks right to me. @internet-diglett thoughts?

docs/how-to-run.adoc
```diff
@@ -2,24 +2,131 @@
...
+|Management Gateway Service (MGS)
+|Missing
+|Real implementation
+
+|Internal ("rack") networking
+|Localhost
+|Real implementation on-sled, simulated implementation across sleds (SoftNPU)
+
+|Boundary services (external connectivity)
+|Localhost
```
rcgoodfellow, line 67: I would say missing here. Boundary services functions like Geneve encap/decap and NAT are missing in the simulated deployment.

docs/how-to-run.adoc
```diff
@@ -254,95 +342,87 @@ $ zoneadm list -cnv
...
 [source,console]
 ----
-$ cargo build --release --bin omicron-package
-$ pfexec ./target/release/omicron-package uninstall
+$ ./tools/scrimlet/softnpu-init.sh
 ----
 
-Once all the omicron services are uninstalled, you can also remove the
-previously created virtual hardware as mentioned above:
+If your external network is one you made up above, you'll need to override the gateway IP to be the address of the global zone on your made-up network.  In our example:
```
rcgoodfellow, line 354: I would probably replace this with something like the following.

The `softnpu-init.sh` script will attempt to determine a network gateway to use by looking at your operating system's default gateway. If your default gateway is not on the network you configured above (`192.168.1.0/24` in this example), then you'll need to explicitly specify what gateway to use by setting the `GATEWAY_IP` environment variable.

smf/sled-agent/gimlet-standalone/config-rss.toml
```diff
@@ -4,8 +4,9 @@
 # Agent API.  See the `RackInitializeRequest` type in bootstrap-agent or its
 # OpenAPI spec (in openapi/bootstrap-agent.json in the root of this workspace).
 
-# The /56 subnet for the rack.
-# Also implies the /48 AZ subnet.
+# The /56 subnet for this rack.  This subnet is internal to the rack and fully
+# managed by Omicron, so you can pick anything you want.  The rack-specific /56
```
rcgoodfellow, line 8: nit: pick anything you want -> pick anything you want within the IPv6 ULA range. As a heads up, this configuration option will be going away entirely soon, as the control plane will decide this for itself without user input.

smf/sled-agent/non-gimlet/config-rss.toml
```diff
@@ -4,8 +4,9 @@
 # Agent API.  See the `RackInitializeRequest` type in bootstrap-agent or its
 # OpenAPI spec (in openapi/bootstrap-agent.json in the root of this workspace).
 
-# The /56 subnet for the rack.
-# Also implies the /48 AZ subnet.
+# The /56 subnet for this rack.  This subnet is internal to the rack and fully
+# managed by Omicron, so you can pick anything you want.  The rack-specific /56
```
rcgoodfellow, line 8: Same comment as above, any -> any ULA

rcgoodfellow (commented):
Thanks for the docs overhaul @davepacheco! This is a huge improvement.

smf/sled-agent/non-gimlet/config-rss.toml
rcgoodfellow: This looks good to me.

leftwo (commented):
Thanks for improving these notes, things have continued to move forward and they needed some love.

> I suspect nobody has used gimlet-standalone in quite a while and I wonder if we should remove it altogether.

I believe we hand intended for this `gimlet-standalone` to be a gimlet in a rack but without being part of the overall cluster in the rack.
It may be worth considering our use cases are and how best how to divide, delete, or combine documents and packaging targets in a way that can satisfy our requirements with the minimum amount of overlap.

Here is what I think are the current ways we currently run Omicron, if there are others, let me know.

1. The official use case, gimlet in a rack talking to the switch and other gimlets.
2. A Gimlet in a rack, but acting in isolation and not part of a cluster.
3. A Gimlet on a bench but connected to a sidecar (also on a bench).  I believe sn05 is a system with this setup, but there may be others.
4. A Gimlet on a bench, but with no sidecar and one or two K.2 cards providing network interfaces.
5. A "generic" x86 System (not a gimlet) running helios.  This would be a desktop or rack mounted PC, either in the Oxide lab, or at home.
6. A "generic" x86 system running a simulated Omicron sled agent.

These all have some level of overlapping config, and keeping them all working does require some amount of engineering effort, and perhaps some (#3 above) may not exist for much longer.

docs/how-to-run.adoc
```diff
@@ -2,24 +2,131 @@
...
-resources.
+Omicron is the control plane for the Oxide system.  There are a few different ways to run Omicron depending on what resources you have available and how much of the stack you want to run using real components vs. simulated ones.  Generally speaking, it's easier to get things going and automate testing with simulated components, but of course much of the system's functionality is missing or faked up when using simulated components.
 
-If you're interested in running the control plane on other platforms, including
-Linux and Mac, refer to the guide on xref:how-to-run-simulated.adoc[running
-simulated Omicron].
+The most common configurations are:
 
-== Installing Prerequisite Software
+1. Using one or more simulated Sled Agents.  There are no real VMs.  All components listen on localhost and talk to each other directly.  The automated tests in this repo generally use this kind of deployment.  This is the only mode that's supported on non-illumos systems (i.e., Linux and MacOS).
```
jordanhendricks, line 11: I realize this section is highlighting the most common configurations, but I think it's worth mentioning what the product version of Omicron looks like -- namely, how it differs from (2) in this list.

docs/how-to-run.adoc
```diff
@@ -2,24 +2,131 @@
...
+2. Using one or more real Sled Agents (on separate physical machines) using SoftNPU-based Boundary Services.  You can provision real VMs this way.  Intra-"rack" networking and boundary Services (i.e., external connectivity) are fully functional using SoftNPU, a software-based simulation of the Tofino device that's normally responsible for these functions.
++
+Within this mode, you can run on real Oxide hardware ("Gimlets") or ordinary server PCs.
+
+These are historically called "simulated" vs. "non-simulated", though they both involve simulation of some parts of a real Oxide system.
+
+Here's a summary of the tradeoffs:
```
jordanhendricks, line 18: This is precisely the table I've always wanted; thanks for adding this.

docs/how-to-run.adoc
```diff
@@ -2,24 +2,131 @@
...
+
+=== Picking a "machine" type
+
+Omicron packages (discussed in more detail below) are associated with a particular _machine_ type, which is one of:
+
+* `gimlet` (real Oxide hardware deployed in a real Oxide rack)
+* `gimlet-standalone` (real Oxide server hardware running as a single-machine "rack")
+* `non-gimlet` (some kind of PC running as a single-machine "rack")
+
+The main difference are the configuration files used for the Sled Agent and Rak Setup Service (RSS).
```
jordanhendricks, line 123: "Rack"

smf/sled-agent/gimlet-standalone/config-rss.toml
```diff
@@ -22,25 +23,82 @@ ntp_servers = [ "ntp.eng.oxide.computer" ]
...
 # For the structure of these certificates, see `Certificate` in the Nexus
-# internal API.  In practice, it can be unwieldy to put them here.  You can also
-# specify a certificate by including the certificate chain and private key in
-# PEM-format files called "initial-tls-cert.pem" and "initial-tls-key.pem",
-# respectively, in the same place as this configuration file.
+# internal API.  In practice, it's unwieldy to put them in this file.  Instead,
+# you can specify a certificate by including the certificate chain and private
+# key in PEM-format files called "initial-tls-cert.pem" and
+# "initial-tls-key.pem", respectively, in the same place as this configuration
+# file.
```
jordanhendricks, line 61: worth adding an example in the comment here, of what it looks like when this field is populated?

jordanhendricks (commented):
Thanks for fleshing this out so much more.

smf/sled-agent/gimlet-standalone/config-rss.toml
davepacheco: I don't think so because I cannot imagine anybody ever doing this instead of using the separate files.  I've updated the text to make this clearer in 2fe16aa9848d7bed601255980936a17bd4e911c3.

smf/sled-agent/gimlet-standalone/config-rss.toml
davepacheco: Done in 2fe16aa9848d7bed601255980936a17bd4e911c3.

smf/sled-agent/non-gimlet/config-rss.toml
davepacheco: Done in 2fe16aa9848d7bed601255980936a17bd4e911c3.

davepacheco:
> Thanks for improving these notes, things have continued to move forward and they needed some love.
> 
> > I suspect nobody has used gimlet-standalone in quite a while and I wonder if we should remove it altogether.
> 
> I believe we hand intended for this `gimlet-standalone` to be a gimlet in a rack but without being part of the overall cluster in the rack. It may be worth considering our use cases are and how best how to divide, delete, or combine documents and packaging targets in a way that can satisfy our requirements with the minimum amount of overlap.

The reason I thought `gimlet-standalone` may not be used is that `rack_network_config` is missing in its config-rss.toml file, and I concluded (wrongly, it turns out) that this file couldn't be successfully used because of that.  But I see that that field is optional.  So maybe this does work, and _is_ being used?


> Here is what I think are the current ways we currently run Omicron, if there are others, let me know.
> 
>     1. The official use case, gimlet in a rack talking to the switch and other gimlets.
> 
>     2. A Gimlet in a rack, but acting in isolation and not part of a cluster.
> 
>     3. A Gimlet on a bench but connected to a sidecar (also on a bench).  I believe sn05 is a system with this setup, but there may be others.
> 
>     4. A Gimlet on a bench, but with no sidecar and one or two K.2 cards providing network interfaces.
> 
>     5. A "generic" x86 System (not a gimlet) running helios.  This would be a desktop or rack mounted PC, either in the Oxide lab, or at home.
> 
>     6. A "generic" x86 system running a simulated Omicron sled agent.
> 
> 
> These all have some level of overlapping config, and keeping them all working does require some amount of engineering effort, and perhaps some (#3 above) may not exist for much longer.

This is really helpful context.  I thought that the existing instructions were really aimed at case (5), that the Gimlet cases were nearly the same as that one, and that the other cases were one-offs with their own instructions (e.g., dogfood, which has a separate guide).  I wonder now if I've overshot in trying simplify these instructions for only 1-2 of those cases.  The trouble is that I don't know how some of these other cases get set up (really, any of (2), (3), or (4)).  I don't think the existing docs really explained that in the same level of detail that I'm going for here.  But the existing docs may have had steps that I trimmed because I thought they were not relevant.

I'm not sure how best to document all this since these cases all have so many overlapping steps.  I'd welcome any suggestions here.  Otherwise, I'll take another pass over this with the goals of (1) making it clear to people who aren't already very familiar with Omicron how to deploy on a single system [my main goal with this PR] while (2) at least not _removing_ anything that was useful for these other cases, and ideally trying to incorporate these cases into this doc.

docs/how-to-run.adoc
davepacheco: Part of what I found hard to follow about the existing docs is that there are a bunch of instructions like "if <situation I can't tell if it applies to me>, then fill in <some value I'm not sure how to fill in>".  What you wrote here is definitely better because it's more explicit about the condition.  But I think it's useful to be even more explicit: "if you chose option 2 above with this network, this is exactly what you should run".

I'm inferring that maybe this is _too_ specific though?  I reworded this in da57b53515c38a699fd5e2960e24d03160fd0fcc.  Let me know what you think.

davepacheco:
I know everyone's pretty busy with other urgent stuff.  Also, I think folks are continuing to get confused by the current docs.  If folks at least feel this is a step in the right direction, I'd like to land it.  If I don't hear from anyone else soon, I'll plan to land it (but please do speak up if that feels wrong!).

smf/sled-agent/gimlet-standalone/config-rss.toml
internet-diglett: `rack_network_config` triggers a section of code in RSS and in the nexus handoff that isn't strictly necessary for exercising functionality that doesn't require networking (for example, a lot of our unit tests and integration tests omit this part of the config). If we're trying to bring up the control plane with external networking, this config will need to be present, or the underlying APIs (i.e. Dendrite) will need to be called to manually set up the networking. If someone is testing on a bench and the gimlet is not connected to any systems running Dendrite, and the gimlet isn't running a virtual switch via Dendrite Stub or Softnpu + Dendrite, it will fail to startup if this config is present.

davepacheco: I believe this file is never used for the existing unit and integration tests.  My assumption (someone tell me if wrong!) is that when someone is setting up Omicron outside those tests, we're always setting up some kind of external networking, whether SoftNPU- or Tofino-based.  If someone is following the instructions here, I'd expect them to get SoftNPU.  So I'll leave this in.

docs/how-to-run.adoc
```diff
@@ -2,24 +2,138 @@
...
+|illumos, Linux, MacOS
+|illumos only
+
+|Nexus
+|Real implementation
+|Real implementation
+
+|CockroachDB
+|Real implementation, 1-node cluster
+|Real implementation, multi-node cluster (coming soon)
```
smklein, line 36: This is done now, right?

docs/how-to-run.adoc
```diff
@@ -425,84 +450,202 @@ $ pfexec /opt/oxide/softnpu/stuff/scadm \
...
+----
+
+Once Sled Agent has been configured to run as a Scrimlet (whether explicitly or implicitly), it will attempt to create and start the switch zone. This will depend on the switch type that was specified in the build target:
+
+1. `asic` implies we're running on a real Gimlet and are directly attached to the
+Tofino ASIC.
+2. `stub` provides a stubbed out switch implementation that doesn't
+require any hardware.
+3. `softnpu` provides a simulated switch implementation that
+runs the same P4 program as the ASIC, but in software.
```
luqmana, line 618: more like same-ish p4 😅️ (softnpu uses the `sidecar-lite` p4 currently)

luqmana (approved):
Thanks for updating this @davepacheco!

rcgoodfellow (approved):
LGTM. Thanks!

Issue #3461, opened 2023-06-30
Reboot: Switch Zone unhappy with route - missing underlay address


smklein:
After a reboot, the Sled Agent attempts to:

- Delete all zones (fine for now, whether or not we should reconcile some of 'em is a separate discussion)
- Loads local configs (e.g., underlay info, "what services should I launch", etc)
- (If it's a scrimlet) Launch the switch zone

As a part of setting up the switch zone, however, we run into some problems:

```
17:19:03.713Z  INFO SledAgent/BootstrapAgent: Configuring new Omicron zone: oxz_switch                                 
17:19:03.735Z  INFO SledAgent/BootstrapAgent: Installing Omicron zone: oxz_switch                                      
17:19:04.469Z  INFO SledAgent/BootstrapAgent: creating NAT entry for service                                           
    service: ServiceZoneService { id: 75203abd-3d8b-4928-9e29-5235f4ebcdcf, details: BoundaryNtp { ntp_servers: ["ntp.eng.oxide.computer"], dns_servers: ["1.1.1.1", "9.9.9.9"], domain: None, nic: NetworkInterface { id: 50e49355-0dfe-4b4c-9fd4-efd68c5894bc, kind: Service { id: 75203abd-3d8b-4928-9e29-5235f4ebcdcf }, name: Name("ntp-75203abd-3d8b-4928-9e29-5235f4ebcdcf"), ip: 172.30.3.5, mac: MacAddr(MacAddr6([168, 64, 37, 255, 139, 60])), subnet: V4(Ipv4Net(Ipv4Network { addr: 172.30.3.0, prefix: 24 })), vni: Vni(100), primary: true, slot: 0 }, snat_cfg: SourceNatConfig { ip: 192.168.0.222, first_port: 0, last_port: 16383 } } }
17:19:07.793Z  INFO SledAgent/BootstrapAgent: Zone booting (zone=oxz_switch)                                           
17:19:12.865Z  INFO SledAgent/BootstrapAgent: Ensuring bootstrap address fdb0:18c0:4d0c:f4e5::2 exists in switch zone  
17:19:12.865Z  INFO SledAgent/BootstrapAgent: Adding bootstrap address (zone=oxz_switch)                               
17:19:23.445Z  INFO SledAgent/BootstrapAgent: Forwarding bootstrap traffic via oxBootstrap38 to fe80::8:20ff:fe5b:f28  
[...]

Issue #3106, opened 2023-05-12
if the sled agent reboots, it takes out the internal dns zone, leading to a panic


jmpesp:
In debugging another issue, the sled agent panicked. It then tore down the internal dns zone on startup:

```
[ May 12 20:52:56 Stopping because all processes in service exited. ]
[ May 12 20:52:56 Executing stop method (:kill). ]
[ May 12 20:52:56 Executing start method ("ctrun -l child -o noorphan,regent /opt/oxide/sled-agent/sled-agent run /opt/oxide/sled-agent/pkg/config.toml &"). ]
[ May 12 20:52:56 Method "start" exited with status 0. ]
note: configured to log to "/dev/stdout"
20:52:56.888Z INFO SledAgent: Starting mg-ddm service
20:52:56.889Z INFO SledAgent: Importing mg-ddm service
    path = /opt/oxide/mg-ddm/pkg/ddm/manifest.xml
20:53:03.204Z INFO SledAgent: Setting mg-ddm interfaces
    interfaces = ("ixgbe0/ll" "net1/ll")
20:53:03.259Z INFO SledAgent: Enabling mg-ddm service
20:53:03.270Z INFO SledAgent: detecting (real or simulated) SP
20:53:03.270Z INFO SledAgent: setting up bootstrap agent server
20:53:03.317Z INFO SledAgent: Sending prefix to ddmd for advertisement
    DdmAdminClient = [::1]:8000
    prefix = Ipv6Prefix { addr: fdb0:1b:21c1:fd24::, len: 64 }
20:53:03.337Z WARN SledAgent: Deleting existing zone
    zone_name = oxz_internal_dns
```

Later on, it panicked because it couldn't resolve an internal address:

```
20:53:08.955Z WARN SledAgent (HardwareManager): Device is not a Gimlet (i86pc), proceeding with null hardware view
20:53:08.955Z INFO SledAgent (BootstrapAgent): new DNS resolver
    addresses = [[fd00:1122:3344:1::1]:53]
20:53:08.956Z INFO SledAgent (BootstrapAgent): Ensuring service zone is initialized: InternalDns
20:53:08.964Z INFO SledAgent (BootstrapAgent): Service zone oxz_internal_dns does not yet exist
20:53:08.976Z INFO SledAgent (BootstrapAgent): Configuring new Omicron zone: oxz_internal_dns
20:53:08.997Z INFO SledAgent (BootstrapAgent): Installing Omicron zone: oxz_internal_dns
[...]

PR #3466, opened 2023-07-02
[sled agent] Fixes to enable the reboot after RSS initialization

smklein:
This PR consists of changes to the Sled Agent to allow internal services to re-launch successfully after a reboot.

This was tested with the following script, on an initialized system running with SoftNPU:

```bash
# Emulate a "sled shutdown"
svcadm disable sled-agent

# Kill and restart softnpu, to "empty out switch state"
zlogin softnpu pkill softnpu
zlogin softnpu pgrep softnpu || {
    zlogin softnpu 'RUST_LOG=debug /stuff/softnpu /stuff/softnpu.toml &> /softnpu.log &'
}

# Restart the sled agent. Observe that it still tears down all running zones on boot,
# but it successfully re-initializes all sleds which are described by:
# /pool/int/<UUID>/config/services.toml
svcadm enable sled-agent

```

## Before this PR

- Sled Agent was capable of booting and initializing via RSS
- However, after a reboot, the Sled Agent did not successfully re-initialize all zones that had been running before reboot. It attempted to bring up the switch zone, failed, and would not be able to continue initializing other services

## This PR

- Fixes the routing issue described by #3461: Zone initialization only attempts to add a route to the "sled agent underlay address" as a gateway iff (1) the underlay is up, and (2) the zone-being-launched has an address on the same subnet. 
- Delays launching all "Sled Agent services" until after the sled agent has an underlay address. This makes it more likely for services dependent on external connectivity (e.g., NTP, external DNS, Nexus) to initialize correctly.
- The bootstrap agent, which may send a `start_rack_initialize` request if and RSS configuration file is included, no longer throws an error if RSS has already been initialized.

Fixes https://github.com/oxidecomputer/omicron/issues/3461
Fixes https://github.com/oxidecomputer/omicron/issues/3106
Part of https://github.com/oxidecomputer/omicron/issues/725

### DIFF
--- a/sled-agent/src/services.rs
+++ b/sled-agent/src/services.rs
@@ -590,8 +590,9 @@ impl ServiceManager {
         Ok(())
     }
 
-    /// Loads services from the services manager, and returns once all requested
-    /// services have been started.
+    /// Sets up "Sled Agent" information, including underlay info.
+    ///
+    /// Any subsequent calls after the first invocation return an error.
     pub async fn sled_agent_started(
         &self,
         config: Config,
--- a/sled-agent/src/services.rs
+++ b/sled-agent/src/services.rs
@@ -506,6 +506,10 @@ impl ServiceManager {
             .collect()
     }
 
+    // TODO(https://github.com/oxidecomputer/omicron/issues/2973):
+    // These will fail if the disks aren't attached.
+    // Should we have a retry loop here? Kinda like we have with the switch
+    // / NTP zone?
     pub async fn load_services(&self) -> Result<(), Error> {
         let log = &self.inner.log;
         let ledger_paths = self.all_service_ledgers().await;
@@ -610,15 +614,6 @@ impl ServiceManager {
             })
             .map_err(|_| "already set".to_string())
             .expect("Sled Agent should only start once");
-
-        // TODO(https://github.com/oxidecomputer/omicron/issues/2973):
-        // These will fail if the disks aren't attached.
-        // Should we have a retry loop here? Kinda like we have with the switch
-        // / NTP zone?
-        self.load_services().await.map_err(|e| {
-            error!(self.inner.log, "failed to launch services"; "error" => e.to_string());
-            e
-        })?;
 
         Ok(())
     }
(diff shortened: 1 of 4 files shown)


sled-agent/src/services.rs
```diff
@@ -1305,13 +1301,39 @@ impl ServiceManager {
...
+                // to anything.
+                info!(
+                    self.inner.log,
+                    "Zone not using gateway (even though underlay is up)"
+                );
+                None
+            }
         } else {
             // If the underlay doesn't exist, no routing occurs.
+            info!(
```
andrewjstone, line 1333: Will the switch zone eventually get an underlay address and add a route?

andrewjstone (approved):
Sean this is awesome. I'm very excited to see this go in. Just one question for my understanding purposes.

sled-agent/src/services.rs
smklein: Good question, and the answer is: Yes, it will.

When we first boot up, the `BootstrapAgent` is responsible for "hardware monitoring", which starts the switch zone iff the tofino is detected:

https://github.com/oxidecomputer/omicron/blob/3e020c907a400e2005874d57fd9e94d8f369e761/sled-agent/src/bootstrap/agent.rs#L397-L404

This calls `HardwareMonitor::new`, which itself has a reference to the `ServiceManager` structure: https://github.com/oxidecomputer/omicron/blob/92258b4d88b292effaf26f361dddb0580b96c1bf/sled-agent/src/bootstrap/hardware.rs#L211 

(Note, this is the **same** `ServiceManager` structure that the `SledAgent` will later use to manage services when we've fully booted)

The bootstrap's `HardwareManager` spawns a tokio task which monitors hardware for the Tofino, and in response, launches the switch zone:

https://github.com/oxidecomputer/omicron/blob/92258b4d88b292effaf26f361dddb0580b96c1bf/sled-agent/src/bootstrap/hardware.rs#L72-L79

**It's important to note -- the `let switch_zone_ip = None` line is critical. If the bootstrap agent is launching the switch zone, it might be doing so before the underlay is up.**

Later on, when the `BootstrapAgent` tries to launch the `SledAgent`. It stops monitoring hardware at this point.

https://github.com/oxidecomputer/omicron/blob/92258b4d88b292effaf26f361dddb0580b96c1bf/sled-agent/src/bootstrap/agent.rs#L526-L527

And when the `SledAgent` starts, it does hardware monitoring for itself:

https://github.com/oxidecomputer/omicron/blob/92258b4d88b292effaf26f361dddb0580b96c1bf/sled-agent/src/sled_agent.rs#L336-L340

If the switch zone already launched, the `SledAgent` will call `activate_switch` with the underlay IP address:

https://github.com/oxidecomputer/omicron/blob/92258b4d88b292effaf26f361dddb0580b96c1bf/sled-agent/src/sled_agent.rs#L353-L365

(this is poking at the same `ServiceManager` used by the bootstrap agent earlier!)

This case is handled explicitly: If the switch zone is running, and needs to be refreshed to add an underlay address and route. It's the same path we use during cold boot / RSS!

https://github.com/oxidecomputer/omicron/blob/2d46f8cc0681c7c25bde57f0fc48ba592d7cd385/sled-agent/src/services.rs#L2742-L2748

andrewjstone: Aha! It was those last two steps I was missing. Thanks for pointing this out!

PR #3482, opened 2023-07-03
Add documentation to physical disk type

zephraph:
The meaning of `internal` /  `external` isn't clear from a customer perspective. I've added some documentation to be a bit more explicit. I _do_ wonder if this is the right terminology though.

### DIFF
--- a/nexus/types/src/external_api/views.rs
+++ b/nexus/types/src/external_api/views.rs
@@ -325,10 +325,14 @@ pub struct Switch {
 
 // PHYSICAL DISKS
 
+/// Classifies a physical disk as either external to the sled (e.g. a U.2 SSD) or internal
+/// to the sled (e.g. an M.2)
 #[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize, JsonSchema)]
 #[serde(rename_all = "snake_case")]
 pub enum PhysicalDiskType {
+    /// A drive located inside the sled
     Internal,
+    /// A dive located on the front side of the sled
     External,
 }
 
--- a/nexus/types/src/external_api/views.rs
+++ b/nexus/types/src/external_api/views.rs
@@ -332,7 +332,7 @@ pub struct Switch {
 pub enum PhysicalDiskType {
     /// A drive located inside the sled
     Internal,
-    /// A dive located on the front side of the sled
+    /// A drive located on the front side of the sled
     External,
 }
 
--- a/openapi/nexus.json
+++ b/openapi/nexus.json
@@ -11103,10 +11103,22 @@
         ]
       },
       "PhysicalDiskType": {
-        "type": "string",
-        "enum": [
-          "internal",
-          "external"
+        "description": "Classifies a physical disk as either external to the sled (e.g. a U.2 SSD) or internal to the sled (e.g. an M.2)",
+        "oneOf": [
+          {
+            "description": "A drive located inside the sled",
+            "type": "string",
+            "enum": [
+              "internal"
+            ]
+          },
+          {
+            "description": "A dive located on the front side of the sled",
+            "type": "string",
+            "enum": [
+              "external"
+            ]
+          }
         ]
       },
       "Project": {


nexus/types/src/external_api/views.rs
```diff
@@ -325,10 +325,14 @@ pub struct Switch {
...
 // PHYSICAL DISKS
 
+/// Classifies a physical disk as either external to the sled (e.g. a U.2 SSD) or internal
+/// to the sled (e.g. an M.2)
 #[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize, JsonSchema)]
 #[serde(rename_all = "snake_case")]
 pub enum PhysicalDiskType {
+    /// A drive located inside the sled
     Internal,
+    /// A dive located on the front side of the sled
```
smklein, line 335:
```suggestion
    /// A drive located on the front side of the sled
```

nexus/types/src/external_api/views.rs
```diff
@@ -325,10 +325,14 @@ pub struct Switch {
 
 // PHYSICAL DISKS
 
+/// Classifies a physical disk as either external to the sled (e.g. a U.2 SSD) or internal
+/// to the sled (e.g. an M.2)
```
smklein, line 329: I believe I didn't reference "U.2" vs "M.2" in the public-facing API, since that's sorta an implementation detail of our rack -- like, theoretically, we could use a U.2 for internal storage too.

smklein (approved):

nexus/types/src/external_api/views.rs
david-crespo: Does external vs. internal correspond to used by VMs vs. used by internal systems? Is there some more producty level to the explanation we can give?

zephraph: Hmmm, I see. I interpreted it more literally (which goes to show that the docs are important). Is this supposed to be just available for customer usage vs not available for customer usage?

smklein: Not precisely -- we put internal systems on external disks too. 

Generally: Internal storage is tightly coupled to the sled, and stores information that's necessary for the sled to operate at all. Data on the internal storage is treated as an exact replica between the two disks for redundancy.

External storage, on the other hand, stores data for our internal services which are "fleet-wide". For example: We store CockroachDB datasets here, Crucible datasets (which back customer disks), DNS records, etc. It's expected that, for redundancy, Oxide services have multiple copies of data, spread across different disks.

Framed another way: Internal storage stores information that's **specific to this sled**. External storage stores information that's **fleet-wide**, and is redundant (most of the time; debug / log info might end up living on external storage too).

There's also a distinction in serviceability, according to RFD 132 - external storage should be a CRU (customers can replace these), internal storage is a FRU (Oxide technicians can replace these, but customers probably shouldn't).

zephraph: There's a lot of nuance here that seems like it would be more confusing to customers than helpful. I'm leaning towards closing this and never exposing internal disks at all. Is there a reason I _shouldn't_ do that?

smklein: That seems okay with me. I think we may want to end up in a world where we present a more unified picture of hardware, to help identify "what's working, what's not working, what needs customer / oxide support to replace", etc, but that doesn't need to be done through this specific interface.

zephraph:
Closing this in favor of just filtering out the results to only include external disks.

nexus/types/src/external_api/views.rs
rmustacc: I would be careful not to conflate location in the chassis with how it's being used. Realistically M.2s are likely to be CRUs. That's why we put a lot of effort into the toolless design.

We may have future products where the storage medium is such that you have to take out the device to service it (consider a dense storage node), but it's still meant for stuff that's fleet-wide.

Given the feedback from @ericaasen about things being misclassified and confusing, I think it's worth asking if this particular piece of info should be there at all. If we have an endpoint that's meant to list all disks, it seems like it should be listing the M.2s as they are FRUs. It's true that we're not listing other critical storage that is on the board through this API at this time and maybe wouldn't want to because SPI NOR doesn't really have firmware or other things we can change or replace. But I think I may be missing some of the original motivating context for how this was meant to be consumed.

PR #3483, opened 2023-07-03
[nexus] Unpublish system update endpoints

david-crespo:
These endpoints are hooked up to CRDB, but CRDB is not hooked up to the underlying update mechanism. So these endpoints don't do anything right now, and none of the clients are using them.


nexus/src/external_api/tag-config.json
```diff
@@ -103,12 +103,6 @@
       "external_docs": {
         "url": "http://docs.oxide.computer/api/system-silos"
       }
-    },
-    "system/update": {
-      "description": "This includes operations related to system updates.",
-      "external_docs": {
-        "url": "http://docs.oxide.computer/api/system-update"
-      }
```
david-crespo, line 111: Thought about leaving this in, but if not I would have to update the docs site to handle empty tags. The tags test will remind us to put it back in when we re-publish these endpoints.

ahl (approved):
lgtm -- "hidden" tags in a different PR? makes sense.

zephraph (approved):

david-crespo:
> lgtm -- "hidden" tags in a different PR? makes sense.

After looking at the 4 hidden endpoints, I kinda want to just leave them hidden. Each is used by a client, so we need them in the spec, but they don't really need to be in the docs, do they? I guess the argument on the other side would be: why give ourselves the mental overhead of having this one weird category?

| endpoint | what it is |
|---|---|
| `device_access_token` | polled by CLI until it says the user is logged in |
| `device_auth_confirm` | used by console to POST the token |
| `device_auth_request` | used by CLI to get the token |
| `logout` | delete session, only used by console |

Either way I should probably put the `/v1` prefix on the token ones that don't have it.

https://github.com/oxidecomputer/omicron/blob/0325e96f3c4fab1bce868ca2c2d4189b7669262b/nexus/tests/output/nexus_tags.txt#L14-L19

PR #3492, opened 2023-07-05
[sled-agent][nexus] Make the Nexus Zone self-assembling

smklein:
Fixes https://github.com/oxidecomputer/omicron/issues/2882

### DIFF
--- a/sled-agent/src/services.rs
+++ b/sled-agent/src/services.rs
@@ -88,7 +88,7 @@ use std::collections::HashSet;
 use std::io::Cursor;
 use std::iter;
 use std::iter::FromIterator;
-use std::net::{IpAddr, Ipv6Addr, SocketAddr};
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr};
 use std::str::FromStr;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
@@ -122,6 +122,9 @@ pub enum Error {
 
     #[error("Failed to find device {device}")]
     MissingDevice { device: String },
+
+    #[error("Invalid service request: {0}")]
+    BadRequest(String),
 
     #[error("Failed to access ledger: {0}")]
     Ledger(#[from] crate::ledger::Error),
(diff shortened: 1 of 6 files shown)


sled-agent/src/services.rs
```diff
@@ -1194,6 +1197,155 @@ impl ServiceManager {
...
+                // atop an OPTE port which operates on a VPC private IP. OPTE will
+                // map the private IP to the external IP automatically.
+                let port =
+                    installed_zone.opte_ports().nth(0).ok_or_else(|| {
+                        Error::BadRequest(
+                            "Nexus zone missing OPTE port".to_string(),
+                        )
+                    })?;
+                let opte_interface = port.vnic_name();
+                let opte_gateway = &port.gateway().ip().to_string();
```
smklein, line 1239: This bit is new, we're pulling out these values instead of calling `ensure_address_for_port` (which was previously performed with a bunch of `zlogin` operations). Instead, these addresses should be set up in the method scripts.

sled-agent/src/services.rs
```diff
@@ -1194,6 +1197,155 @@ impl ServiceManager {
...
+                        tls: external_tls,
+                        dropshot: dropshot::ConfigDropshot {
+                            // This is a bit silly, but:
+                            // - We do not know the private IP address of Nexus
+                            // until the zone is launched.
+                            // - To remediate, we pass a placeholder value here.
+                            // - We replace that value in the Nexus method
+                            // script. See the "external_ip_override" argument
+                            // to the Nexus binary.
+                            bind_address: SocketAddr::new(
```
smklein, line 1258: I don't love this, and it's probably worth a critical look.

Basically:
- Previously, we waited until the zone was running to create this IP address
- We did so with `ipadm create-addr -t -T dhcp "$OPTE_INTERFACE/public"`
- This created a new address which we plugged into the zone config

However, now that we're trying to plumb this config into the zone **before** it launches, the IP address doesn't exist, and we don't know what it is.

smf/nexus/method_script.sh
```diff
@@ -0,0 +1,43 @@
...
+if [[ "$OPTE_GATEWAY" =~ .*:.* ]]; then
+    # IPv6 gateway
+    echo "IPv6 OPTE gateways are not yet supported"
+    exit 1
+else
+    # IPv4 gateway
+    ipadm show-addr "$OPTE_INTERFACE/public"  || ipadm create-addr -t -T dhcp "$OPTE_INTERFACE/public"
+    OPTE_IP=$(ipadm show-addr -p -o ADDR "$OPTE_INTERFACE/public" | cut -d'/' -f 1)
+    route add -host "$OPTE_GATEWAY" "$OPTE_IP" -interface -ifp "$OPTE_INTERFACE"
+    route add -inet default "$OPTE_GATEWAY"
```
smklein, line 35: I still need to make this idempotent

smf/nexus/method_script.sh
```diff
@@ -0,0 +1,43 @@
...
+# Set up underlay
+ipadm show-addr "$DATALINK/ll" || ipadm create-addr -t -T addrconf "$DATALINK/ll"
+ipadm show-addr "$DATALINK/omicron6"  || ipadm create-addr -t -T static -a "$LISTEN_ADDR" "$DATALINK/omicron6"
+route get -inet6 default -inet6 "$GATEWAY" || route add -inet6 default -inet6 "$GATEWAY"
+
+# Set up OPTE interface
+if [[ "$OPTE_GATEWAY" =~ .*:.* ]]; then
+    # IPv6 gateway
+    echo "IPv6 OPTE gateways are not yet supported"
+    exit 1
```
smklein, line 29: This still needs to be patched; I don't understand how this is working today.

smf/nexus/method_script.sh
smklein: Done

sled-agent/src/services.rs
luqmana: We do know the private, you might have to plumb it up a little but the `Port` type has it: https://github.com/oxidecomputer/omicron/blob/15e441f9ac24f4b3d8221eca07e25fe261297b1a/illumos-utils/src/opte/port.rs#L14-L18

The "dhcp" server is just gonna return the ip configured at port creation. We could've easily just did the `ipadm` calls for a static ip but dhcp also plumbs the correct routes (except we add those manually because illumos' dhcp client doesn't support the classes static routes option).

sled-agent/src/services.rs
```diff
@@ -1194,6 +1197,155 @@ impl ServiceManager {
...
+                let profile = ProfileBuilder::new("omicron").add_service(
+                    ServiceBuilder::new("oxide/nexus").add_instance(
+                        ServiceInstanceBuilder::new("default")
+                            .add_property_group(config),
+                    ),
+                );
+                profile
+                    .add_to_zone(&self.inner.log, &installed_zone)
+                    .await
+                    .map_err(|err| Error::io("crucible pantry profile", err))?;
```
luqmana, line 1345: copy-paste?

smf/nexus/method_script.sh
luqmana: Yea, IPv6 support for OPTE ports hasn't been fully fleshed out. Means only IPv4 addresses supported for services as well rn.

luqmana (commented):
the `external_ip_override` bits aside, looks good.

Issue #2158, opened 2023-01-12
Initialize multi-node ClickHouse rather than single-node


bnaecker:
The analog to https://github.com/oxidecomputer/omicron/issues/727 for ClickHouse. We need to start ClickHouse in a multi-node cluster, rather than single-node. There are a few pieces to this:

- The existing merge-tree-based tables need to be moved to `ReplicatedMergeTree` or a similar `Replicated*` variant. See [this page](https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/replication) for how to set up replication. Briefly, you name the table with a few well-known variables, like `/clickhouse/tables/{shard}'. That key `{shard}` is interpolated from the XML configuration file used to launch the DB itself. So we need a few different XML files (or a way to generate them), and the corresponding names in the SQL for creating the tables.
- We need to actually create [distributed tables](https://clickhouse.com/docs/en/engines/table-engines/special/distributed) on top of the replicated ones. This is a bit confusing, but basically this allows making queries of a single distributed table, which then actually distributes the queries to all replicas and aggregates results. I'm not sure how that part works on insertion, I need to read that again.

bnaecker:
There is also [this general page](https://clickhouse.com/docs/en/manage/replication-and-sharding) on replication and sharding. Note that the version of ClickHouse that we use currently uses ZK, while in later versions CH has actually written their own wire-compatible implementation. (We should update to one of those versions too.)

PR #3494, opened 2023-07-05
Scale out Clickhouse to a multinode cluster

karencfv:
## Replication

This PR implements an initial 2 replica 3 coordinator ClickHouse set up.

I've settled on this initial lean architecture as I want to avoid cluttering with what may be unnecessary additional nodes and using up our customers resources. As we gauge the system alongside our first customers we can decide if we really do need more replicas or not. Inserting an additional replica is very straightforward, as we only need to make a few changes to the templates/service count and restart the ClickHouse services.

## Sharding

Sharding can prove to be very resource intensive, and we have yet to fully understand our customer's needs. I'd like to avoid a situation where we are prematurely optimising when we have so many unknowns. We also have not had time to perform long running testing. See official ClickHouse [recommendations](https://clickhouse.com/blog/common-getting-started-issues-with-clickhouse#2-going-horizontal-too-early). 

Like additional replicas, we can have additional shards if we find them to be necessary down the track.

## Testing

I have left most tests as a single node set up. It feels unnecessary to spin up so many things constantly. If people disagree, I can modify this.

I have run many many manual tests, starting and stopping services and so far the set up has held up.

An Omicron build with these changes is currently running on sn21 please feel free to poke it and prod it :) 
_**UPDATE:** The build is borked because some things changed recently and I can't get Omicron running on the gimlet the way I used to :(_

_NB: I wasn't able to get nexus running on the gimlet :P sorry. I tried and failed miserably_

Using a ClickHouse client:

```console
root@oxz_clickhouse_af08dce0-41ce-4922-8d51-0f546f23ff3e:~# ifconfig
<redacted>
oxControlService13:1: flags=21002000841<UP,RUNNING,MULTICAST,IPv6,FIXEDMTU> mtu 9000 index 2
	inet6 fd00:1122:3344:101::f/64 
root@oxz_clickhouse_af08dce0-41ce-4922-8d51-0f546f23ff3e:~# cd /opt/oxide/clickhouse/
root@oxz_clickhouse_af08dce0-41ce-4922-8d51-0f546f23ff3e:/opt/oxide/clickhouse# ./clickhouse client --host fd00:1122:3344:101::f
ClickHouse client version 22.8.9.1.
Connecting to fd00:1122:3344:101::f:9000 as user default.
Connected to ClickHouse server version 22.8.9 revision 54460.

oximeter_cluster node 2 :) SELECT * FROM oximeter.fields_i64

SELECT *
FROM oximeter.fields_i64

Query id: dedbfbba-d949-49bd-9f9c-0f81a1240798

┌─timeseries_name───┬───────timeseries_key─┬─field_name─┬─field_value─┐
│ data_link:enabled │  9572423277405807617 │ link_id    │           0 │
│ data_link:enabled │ 12564290087547100823 │ link_id    │           0 │
│ data_link:enabled │ 16314114164963669893 │ link_id    │           0 │
│ data_link:link_up │  9572423277405807617 │ link_id    │           0 │
│ data_link:link_up │ 12564290087547100823 │ link_id    │           0 │
│ data_link:link_up │ 16314114164963669893 │ link_id    │           0 │
└───────────────────┴──────────────────────┴────────────┴─────────────┘

6 rows in set. Elapsed: 0.003 sec. 

```

To retrieve information about the keepers you can use the provided [commands](https://clickhouse.com/docs/en/guides/sre/keeper/clickhouse-keeper#four-letter-word-commands) within each of the keeper zones.

Example:

```console
root@oxz_clickhouse_keeper_9b70b23c-a7c4-4102-a7a1-525537dcf463:~# ifconfig
<redacted>
oxControlService17:1: flags=21002000841<UP,RUNNING,MULTICAST,IPv6,FIXEDMTU> mtu 9000 index 2
	inet6 fd00:1122:3344:101::11/64 
root@oxz_clickhouse_keeper_9b70b23c-a7c4-4102-a7a1-525537dcf463:~# echo mntr | nc fd00:1122:3344:101::11 9181
zk_version	v22.8.9.1-lts-ac5a6cababc9153320b1892eece62e1468058c26
zk_avg_latency	0
zk_max_latency	1
zk_min_latency	0
zk_packets_received	60
zk_packets_sent	60
zk_num_alive_connections	1
zk_outstanding_requests	0
zk_server_state	leader
zk_znode_count	6
zk_watch_count	1
zk_ephemerals_count	0
zk_approximate_data_size	1271
zk_key_arena_size	4096
zk_latest_snapshot_size	0
zk_followers	2
zk_synced_followers	2

```

## Follow up work

- Write up a troubleshooting document
- Implement zone resource quotas? 
- We'll have to have some sort of "ops" APIs for maintenance tasks at some point.
- Generate config templates dynamically.

Closes: https://github.com/oxidecomputer/omicron/issues/2158

## Update

As mentioned previously, we came to an agreement at the last control plane meeting that software installed on the racks should not diverge due to replicated ClickHouse. This means that while ClickHouse [replication is functional](https://github.com/oxidecomputer/omicron/runs/16327261230) in this PR, it has been disabled in the last commit in the following manner:

- The `method_script.sh` for the `clickhouse` service is set to run single node mode by default, but can be switched to run on replicated mode by [swapping a variable to false](https://github.com/oxidecomputer/omicron/pull/3494/files#diff-5475f31ccc4d46ea5ed682a38970067eacc337f0c6cd52581b1609f4ecce6071R31-R49). When we migrate all racks to a replicated ClickHouse setup, all logic related to running on single node will be removed from that file.
- The number of zones defined through RSS will stay the same. Instructions on how to tweak them to launch in replicated mode have been [left in the form of comments](https://github.com/oxidecomputer/omicron/pull/3494/files#diff-9ea2b79544fdd0a21914ea354fba0b3670258746b1350d900285445d399861e1R59-R64).

### Testing

I ran the full CI testing suite on both replicated and single node mode. You can find the replicated test results [here](https://github.com/oxidecomputer/omicron/runs/16327261230), and the single node with disabled replication [here](https://github.com/oxidecomputer/omicron/runs/16360694678)

Additionally, I have added tests that validate the replicated db_init file [here](https://github.com/oxidecomputer/omicron/pull/3494/files#diff-6f5af870905bd92fc2c62db62d674d1e033edee57adbe7ea70d929c79cd03ba1R671), and incorporated [checks](https://github.com/oxidecomputer/omicron/pull/3494/files#diff-6f5af870905bd92fc2c62db62d674d1e033edee57adbe7ea70d929c79cd03ba1R664) in tests that validate whether a CH instance is part of a cluster or not.

### Next steps

To keep this PR compact (if you can call 2000 lines compact), I have created several issues to tackle after this PR is merged from the review comments. In prioritised order, these are:

- https://github.com/oxidecomputer/omicron/issues/3982 from this [comment](https://github.com/oxidecomputer/omicron/pull/3494/#issuecomment-1696096624) 
- https://github.com/oxidecomputer/omicron/issues/3824 from this [comment](https://github.com/oxidecomputer/omicron/pull/3494/#discussion_r1283695665)
- https://github.com/oxidecomputer/omicron/issues/3823 from this [comment](https://github.com/oxidecomputer/omicron/pull/3494/#discussion_r1284673917)


karencfv:
```console
{"msg":"request completed","v":0,"name":"SledAgent","level":30,"time":"2023-07-11T00:28:39.875750454Z","hostname":"EVT22200021","pid":6768,"uri":"/services","method":"PUT","req_id":"52a3f515-115f-48ab-8e98-068cdcf2eda7","remote_addr":"[fd00:1122:3344:101::1]:49950","local_addr":"[fd00:1122:3344:101::1]:12345","component":"dropshot (SledAgent)","error_message_external":"Internal Server Error","error_message_internal":"Failed to find zone image 'clickhouse_keeper.tar.gz' from [\"/opt/oxide\", \"/pool/int/cd01242b-301e-436e-97cb-d76fd1062824/install\"]","response_code":"500"}
```

For some reason it's looking for `clickhouse_keeper.tar.gz` instead of `clickhouse-keeper.tar.gz`. Will look into this when I return

karencfv:
Found a bug :'( https://github.com/oxidecomputer/garbage-compactor/issues/12

smf/clickhouse_keeper/method_script.sh
```diff
@@ -0,0 +1,101 @@
...
+
+ipadm show-addr "$DATALINK/ll" || ipadm create-addr -t -T addrconf "$DATALINK/ll"
+ipadm show-addr "$DATALINK/omicron6"  || ipadm create-addr -t -T static -a "$LISTEN_ADDR" "$DATALINK/omicron6"
+route get -inet6 default -inet6 "$GATEWAY" || route add -inet6 default -inet6 "$GATEWAY"
+
+# Retrieve addresses of the other clickhouse nodes, order them and assign them to be a replica or keeper node.
+# In a follow up PR, keepers will be their own service.
+#
+# This should probably be 5 keepers?
+CH_ADDRS="$(/opt/oxide/internal-dns-cli/bin/dnswait clickhouse-keeper -H \
```
smklein, line 35: I'm not sure we need to fix this now, but worth considering:

Some day, in the near future, we'll be trying to cope with a system where sleds (and disks) are added and removed from the rack. In this scenario, the following could happen:

- We start with this set of 3 keepers
- One of the three keepers dies
- Nexus decides to provision a new keeper somewhere else

For the "original" two keepers, how do they coordinate with the newly provisioned third keeper? Would this require restarting **all** keeper services with the new IPs?

smf/clickhouse_keeper/keeper_config.xml
```diff
@@ -0,0 +1,43 @@
+<clickhouse>
```
smklein, line 1: We can punt this for later, but this does seem like the sorta file we may want to generate dynamically in the future.

smf/clickhouse/method_script.sh
```diff
@@ -28,13 +28,78 @@ ipadm show-addr "$DATALINK/ll" || ipadm create-addr -t -T addrconf "$DATALINK/ll
...
+
+readarray -td, nodes <<<"$CH_ADDRS,"; declare -p nodes
+
+# Assign addresses to replicas
+REPLICA_HOST_01="${nodes[0]}"
+REPLICA_HOST_02="${nodes[1]}"
+
+# Retrieve addresses of the other clickhouse nodes, order them and assign them to be a keeper node.
+#
+# This should probably be 5 keepers?
```
smklein, line 51: Clarifying a little bit -- this isn't really doing assignment, it's basically trying to observe existing DNS records for clickhouse-keepers and use them, right?

Does this operate correctly if not all three can be found?

smf/clickhouse/method_script.sh
```diff
@@ -28,13 +28,78 @@ ipadm show-addr "$DATALINK/ll" || ipadm create-addr -t -T addrconf "$DATALINK/ll
...
-  "--log-file" "/var/tmp/clickhouse-server.log"
-  "--errorlog-file" "/var/tmp/clickhouse-server.errlog"
-  "--"
-  "--path" "${DATASTORE}"
-  "--listen_host" "$LISTEN_ADDR"
-  "--http_port" "$LISTEN_PORT"
-)
-
-exec /opt/oxide/clickhouse/clickhouse server "${args[@]}" &
+# Retrieve addresses of the other clickhouse nodes, order them and assign them to be a replica node.
```
smklein, line 31: Something important to consider:

We already have a rack, in the field, which we are not able to update via RSS. We can **update** it, but we cannot delete all data from scratch.

This rack already has a clickhouse dataset provisioned to it, using the "old mechanism".

If we update that rack in-place, there will be:
- No keepers
- Still the single provisioned clickhouse node
- ... which will try to use this new method script

I believe if we do that, the old clickhouse zone will fail to boot.

sled-agent/src/services.rs
```diff
@@ -1145,6 +1146,39 @@ impl ServiceManager {
...
+                    );
+                }
+                let dns_install = ServiceBuilder::new("network/dns/install")
+                    .add_property_group(dns_config_builder)
+                    // We do need to enable the default instance of the
+                    // dns/install service.  It's enough to just mention it
+                    // here, as the ServiceInstanceBuilder always enables the
+                    // instance being added.
+                    .add_instance(ServiceInstanceBuilder::new("default"));
+
```
smklein, line 1181: This has been fully copy-pasted from the CockroachDB initialization, right? Can we split this into a helper function to avoid duplication?

sled-agent/src/services.rs
```diff
@@ -1172,6 +1208,74 @@ impl ServiceManager {
...
+                        &ns_addr.to_string(),
+                    );
+                }
+                let dns_install = ServiceBuilder::new("network/dns/install")
+                    .add_property_group(dns_config_builder)
+                    // We do need to enable the default instance of the
+                    // dns/install service.  It's enough to just mention it
+                    // here, as the ServiceInstanceBuilder always enables the
+                    // instance being added.
+                    .add_instance(ServiceInstanceBuilder::new("default"));
```
smklein, line 1245: See prior comment about copy-pasting

smklein (commented):
Great work, Karen! I know this has crossed quite a few layers to be integrated, but it'll be awesome to move to a replicated world.

I left a comment about this below -- and we should prioritize confirming this works on a "de novo" deployment -- but we'll need to consider how to migrate from our existing deployments to this replicated scheme without re-running RSS and incurring data loss.

smf/clickhouse/method_script.sh
smklein: One possible path forward, that I *think* is straightforward, but may require some mechanical work:

## Before-this-PR: One service type

- The service `clickhouse` implies "single-node-clickhouse"
- No keepers exist

## This PR, as is: Two service types

- We have a "service type" for `clickhouse` and for `clickhouse-keeper`
- It is implied that `clickhouse` is a replica, acting within a multi-replica system, managed by keepers
- This may prove problematic on a system where "single node clickhouse" already exists

## Proposal: Three different service types

- `clickhouse` can, as it did before this PR, imply "single-node-clickhouse"
- `clickhouse-keeper` stays as-is -- it implies the keeper side of a multi-node cluster
- `clickhouse-replica` is a new service type, which represents a replica acting in a multi-node system

With this proposal, we can provision keepers + replicas on new systems and test them, but we'll keep all the code to launch a "single-node-clickhouse" as it was before this PR, so the use-case for our existing customers can stay the same.

**Later on**, when Nexus provisions services, it can ensure there are sufficient keeper / replica nodes, and tear down the single-node clickhouses.

smf/clickhouse_keeper/method_script.sh
karencfv: Yes, we will sadly need to restart all keepers. I'm assuming they would have a new hostname as well, which means we'd have to restart the replicas as well :/

smf/clickhouse/method_script.sh
karencfv: It cannot. I've added an extra check to make sure it has all addresses

smf/clickhouse_keeper/method_script.sh
```diff
@@ -0,0 +1,101 @@
...
+
+ipadm set-ifprop -t -p mtu=9000 -m ipv4 "$DATALINK"
+ipadm set-ifprop -t -p mtu=9000 -m ipv6 "$DATALINK"
+
+ipadm show-addr "$DATALINK/ll" || ipadm create-addr -t -T addrconf "$DATALINK/ll"
+ipadm show-addr "$DATALINK/omicron6"  || ipadm create-addr -t -T static -a "$LISTEN_ADDR" "$DATALINK/omicron6"
+route get -inet6 default -inet6 "$GATEWAY" || route add -inet6 default -inet6 "$GATEWAY"
+
+# Retrieve addresses of the other clickhouse nodes, order them and assign them to be a replica or keeper node.
+# In a follow up PR, keepers will be their own service.
```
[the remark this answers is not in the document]
karencfv: Fixed the comments to reflect what's actually happening

smf/clickhouse/method_script.sh
karencfv: oh no! I thought we still had one more go at deleting data :(

hmmm, let me think about this one. Having three services can be a way forward, but it'd be really nice to be able update the current use-case to have a replicated set up.

Are we allowed to run one off jobs? That way we could run a migration just the once :thinking:

karencfv (commented):
Thanks for taking the time for reviewing this PR! I've addressed your comments, and will think of a way forward for the current use-case.

karencfv:
helios/deploy is failing with:

```console
5410 | 2023-08-03T05:39:38.098Z | + pfexec zlogin sidecar_softnpu cat /var/log/softnpu.log
-- | -- | --
5411 | 2023-08-03T05:39:38.122Z | *.546                                                         root      28735 dhcpagent      Idle
5412 | 2023-08-03T05:39:38.152Z | + exit 1
```

Feels unrelated to my PR? Will look into this and find out what's up

smf/clickhouse/method_script.sh
```diff
@@ -28,13 +28,90 @@ ipadm show-addr "$DATALINK/ll" || ipadm create-addr -t -T addrconf "$DATALINK/ll
...
+
+for i in {0..1}
+do
+  if ! grep -q "host.control-plane.oxide.internal" <<< "${nodes["${i}"]}"; then
+    printf 'ERROR: retrieved ClickHouse address does not match the expected format\n' >&2
+    exit "$SMF_EXIT_ERR_CONFIG"
+  fi
+done
+
+# Assign addresses to replicas
```
bnaecker, line 51: I think it would help if we made clear these are _hostnames_, specifically SRV records in internal DNS.

smf/clickhouse/method_script.sh
```diff
@@ -28,13 +28,90 @@ ipadm show-addr "$DATALINK/ll" || ipadm create-addr -t -T addrconf "$DATALINK/ll
...
-  "--errorlog-file" "/var/tmp/clickhouse-server.errlog"
-  "--"
-  "--path" "${DATASTORE}"
-  "--listen_host" "$LISTEN_ADDR"
-  "--http_port" "$LISTEN_PORT"
-)
-
-exec /opt/oxide/clickhouse/clickhouse server "${args[@]}" &
+# Retrieve addresses of the clickhouse nodes.
+CH_ADDRS="$(/opt/oxide/internal-dns-cli/bin/dnswait clickhouse -H \
```
bnaecker, line 32: Does `dnswait` actually wait until there are _two_ addresses? Or until there are _any_ addresses? I'm wondering how we can avoid a cluster with 1 node, because `dnswait` returned only one record.

smf/clickhouse_keeper/method_script.sh
```diff
@@ -0,0 +1,106 @@
...
+    | tr '\n' ,)"
+
+if [[ -z "$K_ADDRS" ]]; then
+    printf 'ERROR: found no addresses for other ClickHouse Keeper nodes\n' >&2
+    exit "$SMF_EXIT_ERR_CONFIG"
+fi
+
+readarray -td, nodes <<<"$K_ADDRS,"; declare -p nodes
+
+for i in {0..2}
```
bnaecker, line 43: Same note here, can we iterate over the array directly?

smf/clickhouse_keeper/method_script.sh
```diff
@@ -0,0 +1,106 @@
...
+
+ipadm set-ifprop -t -p mtu=9000 -m ipv4 "$DATALINK"
+ipadm set-ifprop -t -p mtu=9000 -m ipv6 "$DATALINK"
+
+ipadm show-addr "$DATALINK/ll" || ipadm create-addr -t -T addrconf "$DATALINK/ll"
+ipadm show-addr "$DATALINK/omicron6"  || ipadm create-addr -t -T static -a "$LISTEN_ADDR" "$DATALINK/omicron6"
+route get -inet6 default -inet6 "$GATEWAY" || route add -inet6 default -inet6 "$GATEWAY"
+
+# Retrieve addresses of all keeper nodes.
+K_ADDRS="$(/opt/oxide/internal-dns-cli/bin/dnswait clickhouse-keeper -H \
```
bnaecker, line 32: Same question, how do we know we're returning enough SRV records here?

smf/clickhouse_keeper/method_script.sh
```diff
@@ -0,0 +1,106 @@
...
+then
+    KEEPER_ID_CURRENT=$KEEPER_ID_03
+else
+    printf 'ERROR: listen address does not match any of the identified ClickHouse Keeper addresses\n' >&2
+    exit "$SMF_EXIT_ERR_CONFIG"
+fi
+
+# Setting environment variables this way is best practice, but has the downside of
+# obscuring the field values to anyone ssh=ing into the zone.
+export CH_LOG="${DATASTORE}/clickhouse-keeper.log"
```
bnaecker, line 87: Same question, can we dump this into a file so that we can see the configuration?

smf/clickhouse_keeper/method_script.sh
```diff
@@ -0,0 +1,106 @@
...
+elif [[ $KEEPER_ID_02 == $KEEPER_SVC ]]
+then
+    KEEPER_ID_CURRENT=$KEEPER_ID_02
+elif [[ $KEEPER_ID_03 == $KEEPER_SVC ]]
+then
+    KEEPER_ID_CURRENT=$KEEPER_ID_03
+else
+    printf 'ERROR: listen address does not match any of the identified ClickHouse Keeper addresses\n' >&2
+    exit "$SMF_EXIT_ERR_CONFIG"
+fi
```
bnaecker, line 83: Are the SRV records returned by `dnswait` in a guaranteed order? If not, then it seems like we could end up with a different ID for this node if we restart the service.

smf/clickhouse/config_replica.xml
```diff
@@ -0,0 +1,82 @@
...
+    <!-- replication settings -->
+    <macros>
+        <shard>01</shard>
+        <replica from_env="CH_REPLICA_NUMBER"/>
+        <cluster>oximeter_cluster</cluster>
+    </macros>
+
+    <remote_servers replace="true">
+        <oximeter_cluster>
+            <!-- TODO: secret handling TBD -->
```
bnaecker, line 40: Just checking there's an issue tracking this?

bnaecker (commented):
Nice work, this is really great! I've a few questions about DNS and error handling in the method scripts, but overall it looks really good.

One thing I would like to see is at least one CI-run test on the clustered setup. It could be something very simple, such as running a query to verify that there actually are multipe nodes in the cluster. E.g, `SELECT cluster, host_address FROM system.clusters`, and then check there are N with the expected addresses. Is that feasible?

smf/clickhouse/method_script.sh
smklein: If you want to run "one-off jobs", you'd need to define what that means.

## "The right way"

I think, in the limit, the "right way" to re-provision would be using a reliable persistent workflow in Nexus to ensure that necessary services are provisioned -- but this code would need to deal with the fact that "there are many instances of Nexus running at once, that need to be able to converge on the same result".

This is more complication than just "do an allocation once on Nexus boot", because:
- There many Nexus instances running, and they may be running concurrently
- These Nexus instances may crash arbitrarily at any point, and should still converge on the same result
  - This implies that service provisioning must be idempotent

This is doable (see: https://github.com/oxidecomputer/omicron/issues/732 -- I need to plumb generation number info through the sled agent API first, to make them idempotent, before making these calls)

## ... is that doable in a few days?

Honestly, I don't think so. It's a big chunk of work. Worth doing! But not trivial.

smf/clickhouse/config_replica.xml
karencfv: There is now :smile: ! https://github.com/oxidecomputer/omicron/issues/3823

smf/clickhouse/method_script.sh
```diff
@@ -28,13 +28,90 @@ ipadm show-addr "$DATALINK/ll" || ipadm create-addr -t -T addrconf "$DATALINK/ll
...
+then
+    REPLICA_DISPLAY_NAME="oximeter_cluster node 2"
+    REPLICA_NUMBER="02"
+else
+    printf 'ERROR: listen address does not match any of the identified ClickHouse addresses\n' >&2
+    exit "$SMF_EXIT_ERR_CONFIG"
+fi
+
+# Setting environment variables this way is best practice, but has the downside of
+# obscuring the field values to anyone ssh-ing into the zone.
```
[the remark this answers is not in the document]
karencfv: lol. Yes, that would be the obvious thing to do hahaha :woman_facepalming:

smf/clickhouse_keeper/keeper_config.xml
karencfv: https://github.com/oxidecomputer/omicron/issues/3824

smf/clickhouse/method_script.sh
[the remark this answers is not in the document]
karencfv: huh :thinking: that sounds good! let me check that out

karencfv: Added a few checks to make sure we have the correct amount of hostnames with the correct format. Worth to note that the generation of these configs won't always be done via scripts https://github.com/oxidecomputer/omicron/issues/3824 .

smf/clickhouse_keeper/method_script.sh
karencfv: added an extra check here too

smf/clickhouse_keeper/method_script.sh
karencfv: The order shouldn't matter here, both `$KEEPER_ID_*` and `$KEEPER_SVC` are the first 7 numbers of the service's UUID, so the ID will always be `zonename | tr -dc [:digit:] | cut -c1-7` (or `$KEEPER_SVC`).

What this part of the script is doing, is making sure this zone is part of the list of hostnames returned by dnswait. Maybe I'm being overly cautious? What do you think?

smf/clickhouse/method_script.sh
karencfv: oof, yeah, ok I get it now. I was hoping we could run a migration to replicated databases _before_ the update, but it seems like this would be much more hacky/error prone/unprofessional that I thought.

Right, I'll follow the 3 service approach, I'll look into it and may ping you if I get stuck :smile:

karencfv (commented):
Thanks for the review and catching some bugs @bnaecker ! I've addressed some of your comments and will work on the rest tomorrow. 

About the tests, I'll think about how to add at least one :) For now the only tests that spin up replicated clusters are part of helios/deploy.

karencfv (commented):
Thanks for taking the time to review! I think I've addressed all comments here. Please let me know if there is something remaining you disagree with or could be improved in time for the next shipment :)

@smklein, you mentioned about testing an update on dogfood. Can we coordinate this? It's probably best to test this branch before merging it though, what do you think?

karencfv:
By the way, I cannot figure out why the [helios/deploy test](https://buildomat.eng.oxide.computer/wg/0/details/01H7SJT4G7Y9W354311VZ9V7F1/UypAkfRos6RFMPGpbzTjlEDUUzsqnGDaQtQsn2ne3FRl5R0v/01H7SJTDDCCSS9A0FBJQ6FQD0G) is failing with:

```console
  | + pfexec zlogin sidecar_softnpu cat /var/log/softnpu.log
-- | --
5410 | 2023-08-14T09:40:09.109Z |  
5411 | 2023-08-14T09:40:09.133Z | + exit 1
```

Is there some setting I need to tweak or something like that?

karencfv:
Update: Tested update on dogfood and it did not detect that it was a single node installation. On the bright side, I have a pretty good idea how to fix this now that I have a better understanding on how a clean install vs an update work.

bnaecker:
So, is it required that we have different SQL and DB organization between the clustered and single node cases? It seems tricky to me to know when we’re operating against one or the other, and the queries require different syntax AFAIK. Or at least, there exists different syntax for querying the entire cluster, though I’m not sure it’s required in the queries we run. 

Would it make our lives simpler to have the same database organization and query logic, and only have the XML files describe different nodes?

karencfv:
Thanks @bnaecker !

> So, is it required that we have different SQL and DB organization between the clustered and single node cases? It seems tricky to me to know when we’re operating against one or the other, and the queries require different syntax AFAIK. Or at least, there exists different syntax for querying the entire cluster, though I’m not sure it’s required in the queries we run.

From what I have gathered, the only place that needs different syntax is when performing actions on databases and tables (e.g. [create](https://github.com/oxidecomputer/omicron/pull/3494/files#diff-d7e0c19781820dffa1f107bb7e5b9b78bc9f79d85255c3457010f989987d83d3)/[delete](https://github.com/oxidecomputer/omicron/pull/3494/files#diff-5a50af54d5a68272a9417fa3c9eb037045c78ef7d7fda59e7fb5108e6953e0a0)  ). Any other queries can remain the same (e.g. `SELECT * FROM oximeter.fields_i64`). This should simplify things a bit!

> Would it make our lives simpler to have the same database organization and query logic, and only have the XML files describe different nodes?

I wish it was possible :'( You can't have `ReplicatedMergeTree` on a single node set up. I tried. 

I'm planning on implementing a simple mechanism to detect whether it's a single node vs replicated setup in the manifest.xml files. Nothing fancy, since we really will be needing this only until we can get that rack updated to a multi node set up in the near future. I'll update this PR with that change today or tomorrow!

karencfv:
@smklein @bnaecker ready for a new round of reviews :)

In order to detect whether we are dealing with a single node or replicated set up, [ClickHouse](https://github.com/oxidecomputer/omicron/pull/3494/files#diff-5475f31ccc4d46ea5ed682a38970067eacc337f0c6cd52581b1609f4ecce6071R49-R52) and [Oximeter](https://github.com/oxidecomputer/omicron/pull/3494/files#diff-2b0f8701c6aa11cf1ec7ab8aa9e746239790c4c9cdd1e145fa41b64f50bd9f70R324-R329) will no longer be told explicitly, but rather do a series of checks. ClickHouse uses dnswait, and if it only retrieves a single hostname it will launch a single node. Oximeter on the other hand, queries ClickHouse to see if it is part of  `oximeter_cluster`, and will create the different types of tables based on this.

I tested both of these on sn21 by tweaking [`CLICKHOUSE_COUNT`](https://github.com/oxidecomputer/omicron/pull/3494/files#diff-9ea2b79544fdd0a21914ea354fba0b3670258746b1350d900285445d399861e1R59) to `1` and `2`, and works as expected. Hopefully this means that we can go ahead with the non-destructive update seamlessly.

One thing to note: I'm not sure what commit made changes to the oximeter producer, that it seems to have problems with the replicated setup now :( At first I thought it might have been #3878, but no. I tested my branch after that commit and everything seems to be working. I tested my branch on [commit 1ed6f4](https://github.com/oxidecomputer/omicron/commit/1ed6f4b1288f7fcb76f1fc078159e3f735b7cba2) which was the last commit before I made any new changes and that's where t starts failing. Logs from oximeter producer:
```console
06:56:58.410Z INFO oximeter (oximeter-agent): collecting from producer
    collector_id = 31aa9f37-3737-4eee-b054-1736a5cada5f
    file = oximeter/collector/src/lib.rs:75
    producer_id = 8b4ace29-370b-4929-847c-66a5a9fe8431
06:56:58.411Z WARN oximeter (oximeter-agent): failed to collect results from producer: error decoding response body: invalid type: sequence, expected a map at line 1 column 199
    collector_id = 31aa9f37-3737-4eee-b054-1736a5cada5f
    file = oximeter/collector/src/lib.rs:97
    producer_id = 8b4ace29-370b-4929-847c-66a5a9fe8431
```

I'm a bit lost on what could have caused this, as it seems it happened before I made any changes (also, I haven't touched the producer)

@bnaecker do you have some idea on what could have caused this and/or some idea on where/how to fix?

bnaecker:
@karencfv That warning is emitted by the `oximeter-collector`, when it cannot correctly deserialize samples from a producer. 1ed6f4b1288f7fcb76f1fc078159e3f735b7cba2 doesn't seem to have introduced any relevant changes. It appears that the collector, or really `serde_json`, is expecting a map and finds a sequence. We'll need to look at which producer is generating that data, and if it's changed or invalid in some way. Do you have access to the CockroachDB instance backing all this?

karencfv:
> We'll need to look at which producer is generating that data, and if it's changed or invalid in some way.

No data is being collected at all :(

bnaecker:
I don't know why there are _no_ metrics at all, but this particular error is a mismatch between `dendrite` and `oximeter`. We changed the internal interface used to represent the list of fields for a sample from a [list to a map](https://github.com/oxidecomputer/omicron/pull/3878/files/8a84a77e8449f13d1a3e45a256f17907953315dd#r1293823432). Dendrite needs to be updated with this change, since it's currently producing the fields as a list. That was my fault, I didn't think about the fact that Dendrite would need to move beyond this as well, or really anything producing metrics. Propolis / Crucible will also have this issue until they are also updated past this.

oximeter/db/src/db-replicated-init.sql
```diff
@@ -0,0 +1,190 @@
...
+--
+CREATE TABLE IF NOT EXISTS oximeter.measurements_bool ON CLUSTER oximeter_cluster
+(
+    timeseries_name String,
+    timeseries_key UInt64,
+    timestamp DateTime64(9, 'UTC'),
+    datum UInt8
+)
+
+ENGINE = ReplicatedMergeTree()
```
bnaecker, line 11: Here are a few links I found from my search history, which I read through when initially making the DB organization. Not sure if they're useful, just wanted to make sure they're on your radar for this change:

- https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/replication#converting-from-mergetree-to-replicatedmergetree
- https://clickhouse.com/docs/en/engines/table-engines/special/distributed
- https://disk.yandex.com/i/iLA5ssAv3NdYGy
- https://github.com/ClickHouse/ClickHouse/issues/2161

karencfv:
> I don't know why there are no metrics at all, but this particular error is a mismatch between dendrite and oximeter. We changed the internal interface used to represent the list of fields for a sample from a [list to a map](https://github.com/oxidecomputer/omicron/pull/3878/files/8a84a77e8449f13d1a3e45a256f17907953315dd#r1293823432). Dendrite needs to be updated with this change, since it's currently producing the fields as a list. That was my fault, I didn't think about the fact that Dendrite would need to move beyond this as well, or really anything producing metrics. Propolis / Crucible will also have this issue until they are also updated past this.

Ahh gotcha! good to know where that's coming from!

> Can you explain more what this means? How did you test this? What is the plan for rolling this out to existing installations? I'm still a bit concerned about having multiple copies of SQL floating around, where which one is relevant depends on the number of nodes.

The [CLICKHOUSE_COUNT](https://github.com/oxidecomputer/omicron/pull/3494/files#diff-9ea2b79544fdd0a21914ea354fba0b3670258746b1350d900285445d399861e1R59) constant is for RSS. With the replicated set up I set it to 2 (and CLICKHOUSE_KEEPER_COUNT to 3). Since I can't run a real update through wicket for my bench gimlet set up, I changed that constant to 1 to simulate a system that has a single node set up. This means I ran a fresh install changing that constant and making sure it ran on replicated mode if it was on 2 and ran on single node mode if the constant was set to 1.

As I observed when I broke dogfood, during a non-destructive update only [RSS](https://github.com/oxidecomputer/omicron/blob/main/sled-agent/src/rack_setup/service.rs) doesn't run, but each individual service's SMF manifest files do change. This meant that I had to make sure that's where the discovery of the type of installation is detected. 

In the case of the SQL copies, in this proposal, the decision to use one or the other is defined by the response of `[SHOW CLUSTERS](https://github.com/oxidecomputer/omicron/pull/3494/files#diff-6f5af870905bd92fc2c62db62d674d1e033edee57adbe7ea70d929c79cd03ba1R251-R257)`. If the server is part of `oximeter_cluster` it runs on replicated mode, else it will run on single node mode. But yes, I wholly agree with you that we DO NOT want multiple copies of the SQL. I think it would be good to have a module or something that creates the database and tables, I'll write a follow up issue.

On the ClickHouse SMF manifest, as soon as we have all racks on a replicated set up, I will remove the temporary part that runs a different ClickHouse config file depending on the number of hosts. This part is very very ephemeral, just a very short lived workaround, since we'll never have racks running a single node ClickHouse installation again.

Happy to go over other bits of the code or give more clarifications if you'd like!

oximeter/db/src/db-replicated-init.sql
karencfv: Oh nice, thanks!! I'd seen some of these but not all :bowing_woman:

karencfv:
@bnaecker forgot to add, I also added a new test that verifies the replicated SQL is successful on a replicated cluster here -> https://github.com/oxidecomputer/omicron/pull/3494/files#diff-6f5af870905bd92fc2c62db62d674d1e033edee57adbe7ea70d929c79cd03ba1R670

karencfv:
Missing piece to get the helios/deploy test to pass was adding `clickhouse_keeper` to the service enums. 

Next, I will disable the replicated clickhouse set up, but leave all the building blocks ready to migrate all racks to replicated mode at the same time. This was discussed at today's control plane meeting.

karencfv:
# Update

As mentioned previously, we came to an agreement at the last control plane meeting that software installed on the racks should not diverge due to replicated ClickHouse. This means that while ClickHouse [replication is functional](https://github.com/oxidecomputer/omicron/runs/16327261230) in this PR, it has been disabled in the last commit in the following manner:

- The `method_script.sh` for the `clickhouse` service is set to run single node mode by default, but can be switched to run on replicated mode by [swapping a variable to false](https://github.com/oxidecomputer/omicron/pull/3494/files#diff-5475f31ccc4d46ea5ed682a38970067eacc337f0c6cd52581b1609f4ecce6071R31-R49). When we migrate all racks to a replicated ClickHouse setup, all logic related to running on single node will be removed from that file.
- The number of zones defined through RSS will stay the same. Instructions on how to tweak them to launch in replicated mode have been [left in the form of comments](https://github.com/oxidecomputer/omicron/pull/3494/files#diff-9ea2b79544fdd0a21914ea354fba0b3670258746b1350d900285445d399861e1R59-R64).

# Next steps

To keep this PR compact (if you can call 2000 lines compact), I have created several issues to tackle after this PR is merged from the review comments. In prioritised order, these are:

- https://github.com/oxidecomputer/omicron/issues/3982 from this [comment](https://github.com/oxidecomputer/omicron/pull/3494/#issuecomment-1696096624) 
- https://github.com/oxidecomputer/omicron/issues/3824 from this [comment](https://github.com/oxidecomputer/omicron/pull/3494/#discussion_r1283695665)
- https://github.com/oxidecomputer/omicron/issues/3823 from this [comment](https://github.com/oxidecomputer/omicron/pull/3494/#discussion_r1284673917)

end-to-end-tests/src/helpers/ctx.rs
```diff
@@ -240,7 +240,7 @@ pub async fn build_client() -> Result<oxide_client::Client> {
             })
         },
         &Duration::from_secs(1),
-        &Duration::from_secs(600),
+        &Duration::from_secs(1200),
```
smklein, line 243: Can we revert the timeout here? It isn't necessary, right?

karencfv: ah right! will do

schema/crdb/4.0.0/up.sql
```diff
@@ -0,0 +1,25 @@
...
+        ),
+        'true',
+        'Invalid starting version for schema change'
+    ) AS BOOL
+);
+
+ALTER TYPE omicron.public.service_kind ADD VALUE IF NOT EXISTS 'clickhouse_keeper';
+ALTER TYPE omicron.public.dataset_kind ADD VALUE IF NOT EXISTS 'clickhouse_keeper';
+
+COMMIT;
```
smklein, line 25: I know this might seem like a nitpick, but IMO there is an important property described in the schema changing README should respect here: 

https://github.com/oxidecomputer/omicron/blob/01e730acf3efeb6ff2181d558cf373ded383a365/schema/crdb/README.adoc?plain=1#L50

There's a lot more background in https://www.cockroachlabs.com/docs/stable/online-schema-changes#schema-changes-within-transactions , but basically: CRDB docs really strongly recommend against this "multi-statement alteration" within a transaction. In this code, both `ALTER TYPE` calls are statements, so we have "two per transaction".

In this **particular** case, I doubt it would bite us, but I think we should be cautious as a precedent, while we can.

One really easy workaround: we can just issue two transactions from this file, since each one should be idempotent and independent:

```sql
BEGIN;
SELECT CAST ( # Check version...
ALTER TYPE omicron.public.service_kind ADD VALUE IF NOT EXISTS 'clickhouse_keeper';
COMMIT;

BEGIN;
SELECT CAST ( # Check version, again
ALTER TYPE omicron.public.dataset_kind ADD VALUE IF NOT EXISTS 'clickhouse_keeper';
COMMIT;
```

smf/clickhouse/method_script.sh
```diff
@@ -28,13 +28,135 @@ ipadm show-addr "$DATALINK/ll" || ipadm create-addr -t -T addrconf "$DATALINK/ll
...
 route get -inet6 default -inet6 "$GATEWAY" || route add -inet6 default -inet6 "$GATEWAY"
 
-args=(
+# TEMPORARY: Racks will be set up with single node ClickHouse until
+# Nexus provisions services so there is no divergence between racks
+# https://github.com/oxidecomputer/omicron/issues/732
+single_node=true
+
+command=()
+# TODO: Remove single node mode once all racks are running in replicated mode
```
smklein, line 37: Nitpick, for here and elsewhere in this PR -- if you do:

```
$ git diff main | grep "TODO:"
```

Could we embed an appropriate issue number for these TODOs? IMO `TODO(#123)` gives a lot of context, by pointing to an issue number

smklein (approved):
I'm stoked we could get tests for this, and that we got the deployment job to pass.

This looks solid - I have a couple comments below, but the structure of this makes sense to me. Thanks for pushing through this work -- it's hard enough by itself, but harder on a moving target where we're supporting existing platforms.

I'd check-in with @bnaecker to see if he has additional feedback before merging, but modulo my last set of comments, this LGTM!

schema/crdb/4.0.0/up.sql
karencfv: Sorry! tbh I missed that bit. I can do it the correct way, I've added an additional file :)

schema/crdb/4.0.1/up.sql
```diff
@@ -0,0 +1,24 @@
+-- CRDB documentation recommends the following:
```
smklein, line 1: FWIW in my original recommendation, I figured it would have been the following, within a *single* file named `4.0.0/up.sql`, but what you have done works too.

```sql
BEGIN;
SELECT CAST(
    IF(
        (
            SELECT version = '3.0.3' and target_version = '4.0.0'
            FROM omicron.public.db_metadata WHERE singleton = true
        ),
        'true',
        'Invalid starting version for schema change'
    ) AS BOOL
);

ALTER TYPE omicron.public.service_kind ADD VALUE IF NOT EXISTS 'clickhouse_keeper';
COMMIT;

BEGIN;
SELECT CAST(
    IF(
        (
            SELECT version = '3.0.3' and target_version = '4.0.0'
            FROM omicron.public.db_metadata WHERE singleton = true
        ),
        'true',
        'Invalid starting version for schema change'
    ) AS BOOL
);

ALTER TYPE omicron.public.dataset_kind ADD VALUE IF NOT EXISTS 'clickhouse_keeper';
COMMIT;
```

karencfv: thnx!

smf/clickhouse/method_script.sh
```diff
@@ -28,13 +28,136 @@ ipadm show-addr "$DATALINK/ll" || ipadm create-addr -t -T addrconf "$DATALINK/ll
 ipadm show-addr "$DATALINK/omicron6"  || ipadm create-addr -t -T static -a "$LISTEN_ADDR" "$DATALINK/omicron6"
 route get -inet6 default -inet6 "$GATEWAY" || route add -inet6 default -inet6 "$GATEWAY"
 
-args=(
+# TEMPORARY: Racks will be set up with single node ClickHouse until
+# Nexus provisions services so there is no divergence between racks
+# https://github.com/oxidecomputer/omicron/issues/732
+single_node=true
```
bnaecker, line 34: How will this be changed? We'll make a commit which changes it?

At that point, when we boot this new zone...what happens? There already is a database `oximeter`. Now, we can delete that ourselves manually, by just removing the files. But if we don't do that, we'll then run `CREATE DATABASE oximeter ON CLUSTER oximeter_cluster`. Does that conflict? When we create those tables, do those conflict?

test-utils/src/dev/clickhouse.rs
```diff
@@ -141,6 +106,153 @@ impl ClickHouseInstance {
...
+            .env("CH_INTERSERVER_PORT", interserver_port)
+            .env("CH_DATASTORE", data_dir.path())
+            .env("CH_TMP_PATH", tmp_path)
+            .env("CH_USER_FILES_PATH", user_files_path)
+            .env("CH_USER_LOCAL_DIR", access_path)
+            .env("CH_FORMAT_SCHEMA_PATH", format_schemas_path)
+            .env("CH_REPLICA_NUMBER", r_number)
+            // There seems to be a bug using ipv6 with a replicated set up
+            // when installing all servers and coordinator nodes on the same
+            // server. For this reason we will be using ipv4 for testing.
```
bnaecker, line 153: What's the nature of this bug? Will it affect us if we happen to deploy two replicas on the same host?

test-utils/src/dev/clickhouse.rs
```diff
@@ -141,6 +106,153 @@ impl ClickHouseInstance {
...
+                args,
+                child: Some(child),
+            }),
+            Err(e) => Err(e),
+        }
+    }
+
+    /// Start a new ClickHouse keeper on the given IPv6 port.
+    pub async fn new_keeper(
+        port: String,
```
bnaecker, line 182: Kind of a nit, but why are these strings rather than strong types? We can convert them to strings inside when writing out to the environment.

test-utils/src/dev/clickhouse.rs
```diff
@@ -141,6 +106,153 @@ impl ClickHouseInstance {
...
+            "--config-file".to_string(),
+            format!("{}", config_path.display()),
+        ];
+
+        let child = tokio::process::Command::new("clickhouse")
+            .args(&args)
+            .stdin(Stdio::null())
+            .stdout(Stdio::null())
+            .stderr(Stdio::null())
+            .env("CLICKHOUSE_WATCHDOG_ENABLE", "0")
```
bnaecker, line 214: We should probably do `.env_clear()` here to be safe.

test-utils/src/dev/clickhouse.rs
```diff
@@ -141,6 +106,153 @@ impl ClickHouseInstance {
...
```
bnaecker, line 137: We should probably do `.env_clear()` here to be safe.

bnaecker (commented):
Thanks for all the work here @karencfv, this is really a monster PR. I've a few questions about how we're actually going to roll this out, and what can go wrong when we do. Those are mostly not that big a deal, given that it seems like there is some time before we'll actually deploy this.

One bigger question is around whether we should use distributed tables now. I think it would be a good idea, but it's also possible we can defer it a little bit. There's a lot of work here that is independent of the exact DB organization we end up deploying. But I think we want to revisit that question prior to actually pulling the trigger, to minimize the chances of making another breaking change in the future.

oximeter/db/src/db-replicated-init.sql
```diff
@@ -0,0 +1,190 @@
...
+    timeseries_name String,
+    timeseries_key UInt64,
+    timestamp DateTime64(9, 'UTC'),
+    datum Int64
+)
+ENGINE = ReplicatedMergeTree()
+ORDER BY (timeseries_name, timeseries_key, timestamp)
+TTL toDateTime(timestamp) + INTERVAL 30 DAY;
+--
+CREATE TABLE IF NOT EXISTS oximeter.measurements_f64 ON CLUSTER oximeter_cluster
```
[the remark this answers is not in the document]
karencfv: That's a great idea! I've made the change

smf/clickhouse/method_script.sh
karencfv: I've been mulling it over, but tbh I can't say I have a clear answer as to which steps we'll take for the migration (or perhaps remove everything and start from scratch?). I was planning on having that discussion here -> https://github.com/oxidecomputer/omicron/issues/4000 I think a lot of it also depends on "how" Nexus will provision services.

If we are able to perform a one off job for the migration, we may be able to use [`remote()`](https://clickhouse.com/docs/en/sql-reference/table-functions/remote) after renaming the old tables or something like that.

test-utils/src/dev/clickhouse.rs
karencfv: It only appears when using localhost in the same server. It shouldn't be a problem for any rack installation.

test-utils/src/dev/clickhouse.rs
karencfv: Unfortunately that removes the environment variables just before the config files are populated with them so I had to leave it as is.

test-utils/src/dev/clickhouse.rs
karencfv: same

karencfv:
Thanks for taking the time to review @smklein @bnaecker ! I know this isn't a very easy PR to review.

@bnaecker I've addressed your comments. On the matter of implementing distributed tables, I think it's a great idea! I've changed the SQL file with your suggestions, and we can refine things if needed on https://github.com/oxidecomputer/omicron/issues/3982 which is what I'm planning on tackling next.

karencfv:
_sigh_
 
I am really at a loss as to why the build-and-test (ubuntu-20.04)  test is failing with:

```
--- STDERR:              oximeter-db client::tests::test_build_replicated ---
--
5282 | 2023-09-02T00:36:53.298Z | thread 'client::tests::test_build_replicated' panicked at 'Failed to initialize timeseries
 database: Database("Code: 62. DB::Exception: There was an error on [127.0.0.1:9000]: Code: 62. DB::Exception: No macro
 'share' in config while processing substitutions in '/clickhouse/tables/{share}/measurements_i64_local' at '20' or macro is
 not supported here. (SYNTAX_ERROR) (version 22.8.9.24 (official build)). (SYNTAX_ERROR) (version 22.8.9.24 (official 
build))\n")', oximeter/db/src/client.rs:761:14
```

On atrium the test runs just fine

```
karen@atrium ~/src/omicron $ cargo nextest run oximeter-db client::tests::test_build_replicated --nocapture
    Finished test [unoptimized + debuginfo] target(s) in 2.56s
    Starting 1 test across 94 binaries (824 skipped)
       START             oximeter-db client::tests::test_build_replicated

running 1 test
test client::tests::test_build_replicated ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 48 filtered out; finished in 15.33s

        PASS [  15.367s] oximeter-db client::tests::test_build_replicated
------------
     Summary [  15.370s] 1 test run: 1 passed, 824 skipped

```

karencfv:
gargh, ran this in Linux and it passes as well!!! I don't know what's going on here
```
coatlicue@pop-os:~/src/omicron$ cargo nextest run oximeter-db client::tests::test_build_replicated --nocapture
   Compiling serde_json v1.0.105
<...>
    Finished test [unoptimized + debuginfo] target(s) in 8m 40s
    Starting 1 test across 94 binaries (811 skipped)
       START             oximeter-db client::tests::test_build_replicated

running 1 test
test client::tests::test_build_replicated ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 48 filtered out; finished in 10.76s

        PASS [  10.768s] oximeter-db client::tests::test_build_replicated
------------
     Summary [  10.769s] 1 test run: 1 passed, 811 skipped
```

Anybody have any ideas on what I should do about this?

bnaecker:
```
--- STDERR:              oximeter-db client::tests::test_build_replicated ---
--
5282 | 2023-09-02T00:36:53.298Z | thread 'client::tests::test_build_replicated' panicked at 'Failed to initialize timeseries
 database: Database("Code: 62. DB::Exception: There was an error on [127.0.0.1:9000]: Code: 62. DB::Exception: No macro
 'share' in config while processing substitutions in '/clickhouse/tables/{share}/measurements_i64_local' at '20' or macro is
 not supported here. (SYNTAX_ERROR) (version 22.8.9.24 (official build)). (SYNTAX_ERROR) (version 22.8.9.24 (official 
build))\n")', oximeter/db/src/client.rs:761:14
```

This says there is no macro named `share`. Indeed, if you look at [the configuration file](https://github.com/oxidecomputer/omicron/blob/5933f51101dc91ec02cf2402a131c425a83c9977/oximeter/db/src/configs/replica_config.xml#L320), the macro is named `shard`. You'll need to change the interpolated variable [here](https://github.com/oxidecomputer/omicron/blob/5933f51101dc91ec02cf2402a131c425a83c9977/oximeter/db/src/db-replicated-init.sql#L10) (and the other similar table engine definitions) to refer to the `shard` macro.

I think it's worth understanding why the tests pass on `atrium` in this case. Presumably the files are the same, so let's track down why that's not failing. Is the macro `share` defined somewhere in that test case? It also seems plausible that the files (either XML or SQL) are slightly different when run on a native host (`atrium` or a Linux machine). Perhaps those environments have some macro `share` defined in them somehow, or the SQL file is different?

karencfv:
Thanks @bnaecker! It was really strange because I [couldn't replicate that error](https://github.com/oxidecomputer/omicron/pull/3494#issuecomment-1703615320) on a Linux machine either :woman_shrugging: . Made the change, keeping my fingers crossed!

karencfv:
Woohoo! I'm still not sure why the previous run passed in both of my local environments (Helios/Linux) but not in Ubuntu CI. I'm going to look deeper into this and improve this test as part of https://github.com/oxidecomputer/omicron/issues/4001

PR #3496, opened 2023-07-05
Remove internal/external disk type from `PhysicalDisk` view

zephraph:
This PR follows up on https://github.com/oxidecomputer/console/issues/1628. I started #3482 to clarify the field but it was decided there it's best just to take it out.


nexus/db-queries/src/db/datastore/physical_disk.rs
```diff
@@ -338,7 +341,7 @@ mod test {
             String::from("Noxide"),
             String::from("456"),
             String::from("UnrealDisk"),
-            PhysicalDiskKind::M2,
+            PhysicalDiskKind::U2,
```
zephraph, line 344: Changed this to a `U2` b/c the `M2` is filtered out in the results

nexus/tests/integration_tests/sleds.rs
```diff
@@ -119,15 +117,24 @@ async fn test_physical_disk_create_list_delete(
...
+        &internal_client,
+        "v2",
+        "s2",
+        "m2",
+        internal_params::PhysicalDiskKind::M2,
+        sled_id,
+    )
+    .await;
     let disks = physical_disks_list(&external_client, &disks_url).await;
     assert_eq!(disks.len(), 1);
```
zephraph, line 130: This is the part that verifies the filtering works

david-crespo (approved):
Looks good. Only part that worries me is the potential for confusion based on the expectation that we would be listing everything in the DB rather than filtering for U2s. Short of changing the name of the struct, maybe a comment on the list endpoint would help? Or maybe changing the name of the Nexus function? I'm kind of flailing, but I feel like some name somewhere should give a hint that that filter is present in the query.

https://github.com/oxidecomputer/omicron/blob/9eac70a362b951aa333244e37caa110983b67528/nexus/src/external_api/http_entrypoints.rs#L4248-L4254

nexus/db-queries/src/db/datastore/physical_disk.rs
smklein: Personally I find this a little confusing, an internal function called `sled_list_physical_disks` doesn't indicate that it would filter out M.2s.

If we're keeping that name: we should continue inserting this M.2 in the test, and keep returning it.
Otherwise, we should rename `sled_list_physical_disks` to `sled_list_external_disks` or something.

zephraph: I'm fine w/ changing the name if that helps. We can't leave the M.2s in because `sled_list_physical_disk` pipes up to the app layer `sled_list_physical_disk` which maps to `/v1/system/hardware/sleds/:sled/disks`.

rmustacc: Why wouldn't we just list both the U.2s and M.2s? The original reason we had confusion was that I think @ericaasen saw the wrong types here. This endpoint feels like it should list everything. We talked in #3482 about terminology challenges, but I don't see why wouldn't list all disks here.

zephraph:
Chatted offline with @rmustacc about the filtering. I'd slightly misunderstood the original issue so I've removed all the filtering. U.2s and M.2s will be listed but the `type` field is still out.

david-crespo:
Solves my problem!

smklein:
> Chatted offline with @rmustacc about the filtering. I'd slightly misunderstood the original issue so I've removed all the filtering. U.2s and M.2s will be listed but the `type` field is still out.

I'm unclear on this -- how is it less confusing to expose both U.2s and M.2s, but make them totally indistinguishable from one another?

smklein:
Follow-up from some out-of-band discussions: 
- Exposing the info "external" vs "internal" is too opinionated, and not a sufficient replacement of true topology information
- However, exposing the form factor of the device -- u.2 vs m.2 -- is a "fact". We'd like to expose more topology info when we can, but providing this info will be helpful when operating on disks in the short-term, when we have a single gimlet layout.

Therefore: We should probably still report whether or not disks are u.2s or m.2s, and we can provide more topology information later too.

PR #3504, opened 2023-07-05
Update create image API to only take `block_size` for url uploads

### LINKED ISSUE #3497: API to create image from snapshot shouldn't include blocksize
## Task

Block size should be specified only when needed as a part of `ImageSource` instead of always being required. 

## Background

While working on implementing https://github.com/oxidecomputer/console/issues/1456 I ran across something odd. When you create an image from a snapshot it still requires you to enter a block size. Block size is just always required for image create. That made me wonder... what happens if you put in the wrong block size? I chatted with @jmpesp about this and he said crucible would error out if you tried creating an image from a snapshot with a different block size. Looking at the implementation in omicron we actually _silently ignore_ the block size given to us via the API and just use whatever the database said the snapshot had to begin with.

zephraph:
Fixes #3497 

We were only using the block size provided via the API when doing a URL upload. This updates the API so that `block_size` is now part of `ImageSource` and used only in the arm where it's needed. See the linked issue for more details. 

Note that this is a breaking change.

### DIFF
--- a/end-to-end-tests/src/instance_launch.rs
+++ b/end-to-end-tests/src/instance_launch.rs
@@ -45,13 +45,13 @@ async fn instance_launch() -> Result<()> {
         .body(ImageCreate {
             name: generate_name("debian")?,
             description: String::new(),
-            block_size: 512.try_into().map_err(anyhow::Error::msg)?,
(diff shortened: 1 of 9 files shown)


jmpesp (approved):
:+1:

david-crespo (approved):

PR #3508, opened 2023-07-06
Enable timesync for the non-gimlet config

### LINKED ISSUE #3501: following how-to-run, Omicron does not sync clocks when using non-gimlet config
Debugging #3498 I was surprised that Omicron did not use NTP to sync the clocks across the two sleds.  I see that this is deliberately configured:
https://github.com/oxidecomputer/omicron/blob/c3048a1b43b046c284432eba34d0bc1933de4d56/smf/sled-agent/non-gimlet/config.toml#L17-L19

If that's unavoidable, then we should update how-to-run.adoc to include a step that is to configure NTP on all sleds and make sure they're synchronized.  (Would it be possible to just do the same thing we do on real systems here? CC @citrus-it

citrus-it:
Fixes #3501

### DIFF
--- a/.github/buildomat/jobs/deploy.sh
+++ b/.github/buildomat/jobs/deploy.sh
@@ -71,9 +71,18 @@ _exit_trap() {
 		pfexec zlogin "$z" netstat -rncva
 		pfexec zlogin "$z" netstat -anu
 		pfexec zlogin "$z" arp -an
+		pfexec zlogin "$z" cat /etc/inet/hosts
+		pfexec zlogin "$z" cat /etc/resolv.conf
+		pfexec zlogin "$z" cat /etc/nsswitch.conf
+
 	done
 
 	pfexec zlogin softnpu cat /softnpu.log
+
+	pfexec zlogin oxz_ntp chronyc tracking
+	pfexec zlogin oxz_ntp chronyc sources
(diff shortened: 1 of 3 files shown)


luqmana:
This is going to need a little more setup to get the deploy test passing. The "external network" configuration it uses does not actually provide any external connectivity. We just assign a fixed address (`192.168.1.199/24`) to one of the phys links (`igb0`) and that lets us talk to nexus/dns using their "external" `192.168.1.x` addresses.

In order to provide outbound internet connectivity, we can either:
1) Provide a different external networking configuration that'll work the buildomat lab runners. afaik they use a subnet like `10.151.x.y/24` so perhaps carving out some of that. I imagine this could need some config on `centrum`.
2) Enable IP forwarding w/ masquerading (i.e, src-nat)

Alternatively, if all we care about for external connectivity in this test is NTP we could just have something in the global zone listening on the `192.168.1.199` address that would proxy it.

citrus-it:
> This is going to need a little more setup to get the deploy test passing.

Thanks for the explanation. I'll look at options (one of which is to update the documentation for running a non-gimlet configuration with notes about changing the timesync setting).

davepacheco:
Did we end up creating a separate issue for actually doing this?  Even if we don't prioritize it, it would be useful (it would have caught #3589 in CI instead of in dogfood).

Issue #3373, opened 2023-06-15
Firewall rule validation and default rules should treat protocol and host filter as required fields


askfongjojo:
I created a new VPC in my project and some instances with external IP on the `default` subnet that came with it. After that, I did the following:
- disabled the default inbound icmp rule that came with the vpc -> couldn't ping the instance external IP any more from my workstation (this is as expected)
- enabled the inbound icmp rule -> still couldn't ping my instances (this is incorrect)
- updated the priority to 1 -> still couldn't ping
- disabled the default inbound ssh rule -> couldn't ssh the instance any more
- enabled the inbound ssh rule -> similar to the icmp rule, re-enabling the rule still didn't allow ssh (this is incorrect)

I checked the sled-agent log for the sled that hosts the instance. It is getting the fw rule update:
```
[...]

PR #3511, opened 2023-07-06
[firewall rules] Fall back to any host or any protocol when filter list is empty

david-crespo:
The diff is only a couple of lines with [whitespace changes hidden](https://github.com/oxidecomputer/omicron/pull/3511/files?w=1).

This is one possible approach to addressing #3373. Ports already worked this way: `None` or empty vec cause the filter to match any port. But for hosts and and protocols, only `None` falls back to any, while an empty vec will get passed through and presumably never match anything.

I think even if we end up fixing the API types so that an empty `hosts` array is rejected with a 400, we still probably want this logic to be fixed here.

### DIFF
--- a/illumos-utils/src/opte/firewall_rules.rs
+++ b/illumos-utils/src/opte/firewall_rules.rs
@@ -61,41 +61,39 @@ impl FromVpcFirewallRule for VpcFirewallRule {
     }
 
     fn hosts(&self) -> Vec<Address> {
-        self.filter_hosts.as_ref().map_or_else(
-            || vec![Address::Any],
-            |hosts| {
-                hosts
-                    .iter()
-                    .map(|host| match host {
-                        HostIdentifier::Ip(IpNet::V4(net))
-                            if net.prefix() == 32 =>
-                        {
-                            Address::Ip(IpAddr::Ip4(net.ip().into()))
-                        }
-                        HostIdentifier::Ip(IpNet::V4(net)) => {
-                            Address::Subnet(IpCidr::Ip4(Ipv4Cidr::new(
-                                net.ip().into(),
-                                Ipv4PrefixLen::new(net.prefix()).unwrap(),
-                            )))
-                        }
-                        HostIdentifier::Ip(IpNet::V6(net))
-                            if net.prefix() == 128 =>
-                        {
-                            Address::Ip(IpAddr::Ip6(net.ip().into()))
-                        }
-                        HostIdentifier::Ip(IpNet::V6(net)) => {
-                            Address::Subnet(IpCidr::Ip6(Ipv6Cidr::new(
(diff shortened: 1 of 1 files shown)


david-crespo:
Hey @plotnick (because you wrote this) and @internet-diglett (because you’re working on networking APIs now): do you think it makes sense to make ports, hosts, and protocols all treat empty vec the same as null, i.e., matching anything? Without this change, we only do that for ports. I am probably going to tweak the external API to disallow empty vec anyway, but consistency here seems reasonable.

illumos-utils/src/opte/firewall_rules.rs
```diff
@@ -61,41 +61,39 @@ impl FromVpcFirewallRule for VpcFirewallRule {
...
-                        }
-                        HostIdentifier::Vpc(vni) => {
-                            Address::Vni(Vni::new(u32::from(*vni)).unwrap())
-                        }
-                    })
-                    .collect::<Vec<Address>>()
-            },
-        )
+        match self.filter_hosts {
+            Some(ref hosts) if hosts.len() > 0 => hosts
```
plotnick, line 65: The copy of clippy running in my brain saw this and suggested `!hosts.is_empty()`.

plotnick (commented):
So, I think interpreting the empty vec as `Any` for all of hosts, ports, and protocols makes sense, because a rule that never matches anything is pretty useless. It's also how OPTE interprets its filters.

That said, it might not be obvious from the user's perspective, so we should be sure that the behavior is documented and probably indicated somehow in the console.

From a programming perspective, I think if we're going to go this route, we should always use `Vec` instead of `Option<Vec>` in all of the relevant interfaces, but that definitely doesn't have to be done now.

luqmana:
Thanks for taking a look at this @david-crespo! For an immediate fix this works and is fairly small. I feel like I've gone back and forth a bit on whether `Any` should just be an explicit option in the API itself.

luqmana:
@askfongjojo if we're fine with leaving things a bit implicit I think we can mark #3373 as R1. Making things more explicit in the API would be a bit more substantial and require more changes (in omicron and console + sdk/cli too I imagine).

david-crespo:
Cool. I'll merge this, then. 

I'm working on something for the API side that doesn't seem too invasive — I don't think it would require much of a CLI change, if any. The console change would be a couple of lines and I already know which lines. In any case, if I can get it working, I'll push it up and we can decide whether it's worth it.

askfongjojo:
> @askfongjojo if we're fine with leaving things a bit implicit I think we can mark #3373 as R1. Making things more explicit in the API would be a bit more substantial and require more changes (in omicron and console + sdk/cli too I imagine).

Yes, I am fine with the solution. Ship it.

PR #3513, opened 2023-07-06
Disallow creating instances/disks with snapshots/images from another project

### LINKED ISSUE #3335: API: User should be unable to create images and disks with snapshots/images from another project
The use of image id or snapshot id in API request body does not appear to have any validation against the parent project.

I am able to create a disk in Project A from an image or snapshot owned by Project B if I know its uuid. Similarly, I am able to create a snapshot in Project A from a disk owned by Project B by specifying the disk uuid. Even though I have the silo-admin role, it doesn't seem right that I can go across the project boundary to essentially clone someone else's instance (via snapshot/image) into my project.

Here is an example:
```
# list snapshots from anther project
$ oxide snapshot list --project ni
Snapshot {
    description: "snapshot of finalized disk debian-11",
    disk_id: 12ecb2b5-6ad6-4dd9-9e5b-fa3c6642a4b4,
    id: 8e7a81f3-0acd-45ef-a3e0-b1ad980432e0,
    name: Name(
        "debian-cloud",
    ),
    project_id: 991ab788-2af5-46b9-987b-fed15755fc12,
    size: ByteCount(
        10737418240,
    ),
    state: Ready,
    time_created: 2023-06-09T20:43:37.763479Z,
    time_modified: 2023-06-09T20:43:37.763479Z,
}
Snapshot {
    description: "snapshot of finalized disk omniosr47-1686343365",
    disk_id: 9ed58981-4916-4f3e-b1f1-17ea90e2f688,
    id: 7fc43498-903c-4ca6-9a13-af009d6de629,
    name: Name(
        "omnios-cloud-1686343365",
    ),
    project_id: 991ab788-2af5-46b9-987b-fed15755fc12,
    size: ByteCount(
        8589934592,
    ),
    state: Ready,
    time_created: 2023-06-09T20:43:16.398442Z,
    time_modified: 2023-06-09T20:43:16.398442Z,
}

# now use one of their snapshots in my create image request
$ cat cli-image.json 
{
 "name": "omnios-cloud",
 "description": "omnios cloud", "os": "omnios", "version": "latest",
 "block_size": 512, "source": {"type":"snapshot", "id": "8e7a81f3-0acd-45ef-a3e0-b1ad980432e0"}
}

$ oxide image create --project try --json-body cli-image.json 
success
Image {
    block_size: ByteCount(
        512,
    ),
    description: "omnios cloud",
    digest: None,
    id: b53f29dd-4123-4aca-b311-812c83f33f90,
    name: Name(
        "omnios-cloud",
    ),
    os: "omnios",
    project_id: Some(
        fc39a947-1282-495e-86fc-2f9e57b91e7f,
    ),
    size: ByteCount(
        10737418240,
    ),
    time_created: 2023-06-10T04:37:49.554646Z,
    time_modified: 2023-06-10T04:37:49.554646Z,
    url: None,
    version: "latest",
}
```

iliana:
Fixes #3335 (@askfongjojo).

This adds these checks:
- in `project_create_disk`: ensure the snapshot belongs to the same project, or ensure the image belongs to the silo or the same project
- in `image_create`: ensure the snapshot belongs to the same project
- in `project_create_instance`: for created disks, implement the same checks as `project_create_disk`

This is a simple check to close off this kind of cross-project resource usage; we may want to enable this kind of thing in the future (perhaps with authz), but I think we will want to be more careful about the design, and it's easier to restrict it now and add functionality to permit this later.

(From when this was a check in the `create_disk` saga: ~~I don't know that where I added these checks is ideal, especially in a [asynchronous control plane operations](https://rfd.shared.oxide.computer/rfd/402) world. It does not seem common to put these checks in at the saga level, possibly for that reason, but I think at minimum having the checks here in addition to possibly adding them before the saga is created is a good idea, to prevent creating new API calls that start a `disk_create` saga without checking these preconditions.~~)

### DIFF
--- a/nexus/src/app/disk.rs
+++ b/nexus/src/app/disk.rs
@@ -28,49 +28,6 @@ use sled_agent_client::Client as SledAgentClient;
 use std::sync::Arc;
 use uuid::Uuid;
 
-fn validate_disk_create_params(
-    params: &params::DiskCreate,
-    block_size: u64,
-) -> Result<(), Error> {
-    // Reject disks where the block size doesn't evenly divide the
-    // total size
-    if (params.size.to_bytes() % block_size) != 0 {
-        return Err(Error::InvalidValue {
-            label: String::from("size and block_size"),
-            message: format!(
-                "total size must be a multiple of block size {}",
-                block_size,
-            ),
-        });
-    }
-
-    // Reject disks where the size isn't at least
-    // MIN_DISK_SIZE_BYTES
-    if params.size.to_bytes() < params::MIN_DISK_SIZE_BYTES as u64 {
-        return Err(Error::InvalidValue {
-            label: String::from("size"),
-            message: format!(
-                "total size must be at least {}",
-                ByteCount::from(params::MIN_DISK_SIZE_BYTES)
-            ),
-        });
-    }
-
-    // Reject disks where the MIN_DISK_SIZE_BYTES doesn't evenly
-    // divide the size
-    if (params.size.to_bytes() % params::MIN_DISK_SIZE_BYTES as u64) != 0 {
-        return Err(Error::InvalidValue {
-            label: String::from("size"),
-            message: format!(
-                "total size must be a multiple of {}",
-                ByteCount::from(params::MIN_DISK_SIZE_BYTES)
-            ),
-        });
-    }
-
-    Ok(())
-}
-
 impl super::Nexus {
     // Disks
     pub fn disk_lookup<'a>(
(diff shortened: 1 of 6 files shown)


iliana:
Realizing now that I missed that you can create an image with a snapshot from another project, adding that too.

zephraph:
Woof. Good catch.

zephraph:
Hmmm, I see. I'd definitely ask for @davepacheco to weigh in on this as he may have more complex context. My initial reaction is that authz checks should be happening well before a saga initialization. Typically we do all the authz checks at the nexus app layer (and then again at the datastore layer if applicable). Now, if there's code that's doing this that's completely bypassing that code path (e.g. by just directly invoking the saga) then it _would_ bypass the authz checks. Based on the context of the original issue, it looks like this was just some missing checks at the app layer though, so I'm not aware of a need to be extra defensive at the saga level. Again, I could absolutely be missing the bigger picture here so I'd lean on @davepacheco as he as more context on sagas, authz, and their overlap than I do.

iliana:
Fortunately the code I added snaps in just fine into the app layer (that's where I originally wrote it before I moved it); means I would need to figure out how to add it to instance create, but at least the test case is written now.

zephraph:
I think the place to do the image create check is after this:
https://github.com/oxidecomputer/omicron/blob/c2de4805acfd0c20bef9d87afacac9f8e821ad87/nexus/src/app/image.rs#L216-L220

The `..` portion there will have the silo and project authz objects which we can check against and the parent

zephraph:
Hmmm, I'm a bit confused, sorry about this. Looking at @askfongjojo's code example she shows `oxide image create` as doing the wrong thing. Looking at image create I believe that we indeed _don't_ have sufficient checks there. Looks like `disk_create` has that same problem. Yikes.

zephraph:
So yeah, we need more checks after this https://github.com/oxidecomputer/omicron/blob/c2de4805acfd0c20bef9d87afacac9f8e821ad87/nexus/src/app/disk.rs#L122-L127

and after this
https://github.com/oxidecomputer/omicron/blob/c2de4805acfd0c20bef9d87afacac9f8e821ad87/nexus/src/app/disk.rs#L146-L150

These are mostly my fault! Sorry about that.

iliana:
Yeah that's exactly where I wrote the initial checks. I'll just move 'em and write up the right ones for instance/image creation as well.

iliana:
I think in `project_create_instance` we are also not checking everything else that `project_create_disk` checks? e.g. size is a multiple of block size, disk size > snapshot size

iliana:
Updated:

- in `project_create_disk`: ensure the snapshot belongs to the same project, or ensure the image belongs to the silo or the same project
- in `image_create`: ensure the snapshot belongs to the same project
- in `project_create_instance`: for created disks, implement the same checks as `project_create_disk`

davepacheco (approved):

PR #3517, opened 2023-07-07
[installinator/update-engine] some reporting improvements

sunshowers:
* Add some more logging to the update engine.
* Add a way to dump downloaded artifacts, whether they succeed or fail.

Using this to debug https://github.com/oxidecomputer/stlouis/issues/454.


sunshowers:
I also want to add a stress test mode to installinator which downloads the artifacts many times and reports the number of times there was a hash mismatch.

sunshowers:
Don't really need this any more.

PR #3522, opened 2023-07-07
improve local user password docs

### LINKED ISSUE #3516: API: Field label and doc issues with setting password in `local_idp_user_create` and `local_idp_user_set_password` requests
Currently the password value in the request payload is presented as follows in the API docs:
image
The top-level comment "password used to log in" should perhaps be changed to: "set or invalidate user password".

The `user_password_value` field is actually the operation: set password vs invalidate password, not the password value.

The `details` field is where the password value should be set. Apparently this and the field above had their labels swapped during implementation (?).

Finally, it's unclear if the "Passwords may be subject to additional constraints" remark is helpful at all because we don't provide a way to set/apply password policies.

davepacheco:
Fixes #3516.

### DIFF
--- a/nexus/types/src/external_api/params.rs
+++ b/nexus/types/src/external_api/params.rs
@@ -276,7 +276,7 @@ pub struct SiloCreate {
 pub struct UserCreate {
     /// username used to log in
     pub external_id: UserId,
-    /// password used to log in
+    /// describes how to set the user's login password
     pub password: UserPassword,
 }
 
@@ -406,7 +406,7 @@ impl AsRef<omicron_passwords::Password> for Password {
 /// Parameters for setting a user's password
 #[derive(Clone, Deserialize, JsonSchema, Serialize)]
 #[serde(rename_all = "snake_case")]
-#[serde(tag = "user_password_value", content = "details")]
+#[serde(tag = "mode", content = "value")]
 pub enum UserPassword {
     /// Sets the user's password to the provided value
     Password(Password),
--- a/nexus/types/src/external_api/params.rs
+++ b/nexus/types/src/external_api/params.rs
@@ -276,7 +276,7 @@ pub struct SiloCreate {
 pub struct UserCreate {
     /// username used to log in
     pub external_id: UserId,
-    /// describes how to set the user's login password
+    /// how to set the user's login password
     pub password: UserPassword,
 }
 
@@ -411,7 +411,7 @@ pub enum UserPassword {
     /// Sets the user's password to the provided value
     Password(Password),
     /// Invalidates any current password (disabling password authentication)
-    InvalidPassword,
+    LoginDisallowed,
 }
 
 /// Credentials for local user login
--- a/openapi/nexus.json
+++ b/openapi/nexus.json
@@ -12740,7 +12740,7 @@
             ]
           },
           "password": {
-            "description": "describes how to set the user's login password",
+            "description": "how to set the user's login password",
             "allOf": [
               {
                 "$ref": "#/components/schemas/UserPassword"
@@ -12790,7 +12790,7 @@
               "mode": {
                 "type": "string",
                 "enum": [
-                  "invalid_password"
+                  "login_disallowed"
                 ]
               }
             },
--- a/openapi/nexus.json
+++ b/openapi/nexus.json
@@ -12740,7 +12740,7 @@
             ]
           },
           "password": {
-            "description": "password used to log in",
+            "description": "describes how to set the user's login password",
             "allOf": [
               {
                 "$ref": "#/components/schemas/UserPassword"
@@ -12768,26 +12768,26 @@
             "description": "Sets the user's password to the provided value",
             "type": "object",
             "properties": {
-              "details": {
-                "$ref": "#/components/schemas/Password"
-              },
-              "user_password_value": {
+              "mode": {
                 "type": "string",
                 "enum": [
                   "password"
                 ]
+              },
+              "value": {
+                "$ref": "#/components/schemas/Password"
               }
             },
             "required": [
-              "details",
-              "user_password_value"
+              "mode",
+              "value"
             ]
           },
           {
             "description": "Invalidates any current password (disabling password authentication)",
             "type": "object",
             "properties": {
-              "user_password_value": {
+              "mode": {
                 "type": "string",
                 "enum": [
                   "invalid_password"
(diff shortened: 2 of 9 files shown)


davepacheco:
@askfongjojo what do you think of this version?  It looks like this:

Screenshot 2023-07-07 at 11 12 55 AM
Screenshot 2023-07-07 at 11 12 59 AM

openapi/nexus.json
```diff
@@ -12768,34 +12768,34 @@
...
           },
           {
             "description": "Invalidates any current password (disabling password authentication)",
             "type": "object",
             "properties": {
-              "user_password_value": {
+              "mode": {
                 "type": "string",
                 "enum": [
                   "invalid_password"
```
zephraph, line 12793: Perhaps this should be `invalidate_password` instead? `invalid_password` implies a password that's not valid, not something that will remove passwords.

zephraph:
so the input is

```json
{
  "mode": "password",
  "value": "adsfasdl;fajsdf;lajds"
}
```

or 

```json
{
  "mode": "invalid_password"
}
```

---

I would almost prefer an input of

```json
{
  "password": "asdflak;jsdf;"
}
```

or 

```json
{
  "invalidate_password": true
}
```

davepacheco:
> so the input is
> 
> ```json
> {
>   "mode": "password",
>   "value": "adsfasdl;fajsdf;lajds"
> }
> ```
> 
> or
> 
> ```json
> {
>   "mode": "invalid_password"
> }
> ```
> 
> I would almost prefer an input of
> 
> ```json
> {
>   "password": "asdflak;jsdf;"
> }
> ```
> 
> or
> 
> ```json
> {
>   "invalidate_password": true
> }
> ```

I think this is not allowed by openapi-lint.  I initially tried removing the annotation and got:

```
---- integration_tests::commands::test_nexus_openapi stdout ----
writing temp config: /dangerzone/omicron_tmp/test_commands_config.1968.0
thread 'integration_tests::commands::test_nexus_openapi' panicked at 'Problem with type UserCreate: Mismatched types between subschemas; this is often due to enums with different data payloads and can be resolved using serde adjacent tagging.
this schema's type
Object(
    ObjectType {
        properties: {
            "password": Reference {
                reference: "#/components/schemas/Password",
            },
        },
        required: [
            "password",
        ],
        additional_properties: Some(
            Any(
                false,
            ),
        ),
        min_properties: None,
        max_properties: None,
    },
)
differs from this
String(
    StringType {
        format: Empty,
        pattern: None,
        enumeration: [
            Some(
                "invalid_password",
            ),
        ],
        min_length: None,
        max_length: None,
    },
)

For more info, see https://github.com/oxidecomputer/openapi-lint#type-mismatch

Problem with type UserPassword: Mismatched types between subschemas; this is often due to enums with different data payloads and can be resolved using serde adjacent tagging.
this schema's type
Object(
    ObjectType {
        properties: {
            "password": Reference {
                reference: "#/components/schemas/Password",
            },
        },
        required: [
            "password",
        ],
        additional_properties: Some(
            Any(
                false,
            ),
        ),
        min_properties: None,
        max_properties: None,
    },
)
differs from this
String(
    StringType {
        format: Empty,
        pattern: None,
        enumeration: [
            Some(
                "invalid_password",
            ),
        ],
        min_length: None,
        max_length: None,
    },
)

For more info, see https://github.com/oxidecomputer/openapi-lint#type-mismatch', nexus/tests/integration_tests/commands.rs:123:5
```

nexus/types/src/external_api/params.rs
```diff
@@ -276,7 +276,7 @@ pub struct SiloCreate {
 pub struct UserCreate {
     /// username used to log in
     pub external_id: UserId,
-    /// password used to log in
+    /// describes how to set the user's login password
```
ahl, line 279:
```suggestion
    /// how to set the user's login password
```

maybe? in the spirit of
- less is more
- 80 columns in the CLI help menu

nexus/types/src/external_api/params.rs
```diff
@@ -276,7 +276,7 @@ pub struct SiloCreate {
 pub struct UserCreate {
```
ahl, line 276: What do we want the JSON to look like? Today it's something like this:
```json
{
  "external_id": "xyz",
  "password": {
    "mode": "password",
    "value": "l1admin"
  }
}

{
  "external_id": "xyz",
  "password": {
    "mode": "invalid_password"
  }
}
```

That seems good. Is this any better?

```json
{
  "external_id": "xyz",
  "password": "l1admin"
}

{
  "external_id": "xyz",
}
```

The absence of `"password"` or a value of `null` would indicate that the password should be unset? It's the best alternative I can think of, and I hate it.

Or maybe:
```json
{
  "external_id": "xyz",
  "mode": "password",
  "value": "l1admin"
}

{
  "external_id": "xyz",
  "mode": "invalid_password"
}
```

... but that means we'd have a Rust type like this:

```rust
#[serde(tag = "mode")]
pub enum UserCreate {
    Password {
        pub external_id: UserId,
        pub value: String,
    }
    InvalidPassword {
        pub external_id: UserId,
    }
}
```

Enumeration options to say: I like what you've chosen and can't think of something better

openapi/nexus.json
ahl: Or maybe even "no_login" or "login_disallowed"?

ahl (approved):
thanks for pointing me at this

ahl:
> > I would almost prefer an input of
> > ```json
> > {
> >   "password": "asdflak;jsdf;"
> > }
> > ```
> > or
> > ```json
> > {
> >   "invalidate_password": true
> > }
> > ```
> 
> I think this is not allowed by openapi-lint. I initially tried removing the annotation and got:

Removing the `#[serde(...)]`? Yeah, that would make it an "externally" tagged enum which openapi-lint squawks about when you mix a variant with a type and a simple variant. However, I think @zephraph is actually saying something like:

```rust
enum UserPassword {
    Password(Password),
    InvalidatePassword(AlwaysTrue)`
}
/// Serializes as just the value true 
struct AlwaysTrue;
// needs Serialize / Deserialize impls
impl JsonSchema for AlwaysTrue {
    fn json_schema(..) -> Schema {
        serde_json::from_value::<Schema>(json! {
            "enum": [ true ]
        })
    }
    fn is_referenceable() -> bool {
       false
    }
}
```

PR #3524, opened 2023-07-07
Remove hardware constraints given possible manufacture defects

zephraph:
Back in May several of us got together to talk about uniquely identifying hardware. You can find the recording [here](https://drive.google.com/file/d/1LPygwvjaBAVeDo-1XitPKpw6lOAg_1s5/view). The TL;DR is that part, serial, and revision numbers aren't inherently reliable and don't carry the guarantee of always being unique. @smklein had done some work to add db constraints around serial, part, and revision numbers in a few places. Given our limited window for making db changes I think it's a good time to go ahead and figure out if we need to remove these constraints.

@jclulow, @smklein I'd like both of your input on this. I'm fine with closing the PR if that's the consensus, just given the timing I wanted to make sure we had an explicit decision about it.

### DIFF
--- a/common/src/sql/dbinit.sql
+++ b/common/src/sql/dbinit.sql
@@ -98,12 +98,6 @@ CREATE TABLE omicron.public.sled (
 
     /* The last address allocated to an Oxide service on this sled. */
     last_used_address INET NOT NULL,
-
-    -- This constraint should be upheld, even for deleted disks
-    -- in the fleet.
-    CONSTRAINT serial_part_revision_unique UNIQUE (
-      serial_number, part_number, revision
-    )
 );
 
 /* Add an index which lets us look up sleds on a rack */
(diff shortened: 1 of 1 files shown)


smklein (commented):
I'm not a fan of this. There are many spots within the DB where we rely on these values to be unique:

- If a sled notifies Nexus that a new disk exists, we use these values. **Right now**, if these values are duplicated between two different physical disks, the second disk would "fail" - as it **should**, without some other pathway to distinguish the disks.
- With this PR, that notification from sled agent to Nexus would not fail. It would **silently succeed**. Nexus would think the disk moved from one sled to another, without any error or notification. This would put us in an inconsistent state.

smklein:
Put another way: If we choose to remove these values, we need a different mechanism to distinguish pieces of physical hardware. Without that mechanism, I believe this PR is more dangerous than useful.

zephraph:
That's a great point. Disks and sleds are affected here and the practical reality of disks may preclude removing their constraint at this point. To be clear, I would greatly _prefer_ the unique constraints on these identifiers exist because without them we have a much harder problem to solve. It was just my understanding that we _couldn't_ do that. I'm going though dbinit and following up on issues we've discussed but taken no action on. This is one of those items.

jclulow:
I agree in general that merely removing the constraints is insufficient, and potentially less reliable, without also doing the follow-up work we talked about in the past that is more substantial.

zephraph:
Great! My thanks to both of you for the feedback.

Issue #3480, opened 2023-07-03
Want ability to take a crucible dataset out of provisioning pool


askfongjojo:
Similar to #2483 (the ability to take a sled out of provisioning pool), there are multiple circumstatnces (e.g. physical disk failure, crucible-agent failing to come up #3416) that causes a crucible dataset to be non-functional. Regardless of whether the logic for picking crucible is fully randomized, we need a support tool to mark a crucible as "bad".

The tool (probably an API) will be used by Oxide support initially. Eventually, we'd want a heart-beat mechanism in place to make the decision of when to take a crucible out of or back into the provisioning pool.

askfongjojo:
This API can also be triggered optionally by the "take sled out of provisioning pool" feature since talking a sled out will mean not provisioning to its disks in most cases.

askfongjojo:
Volume/region background clean-up also needs to skip crucible zones that are already removed - https://github.com/oxidecomputer/omicron/issues/4331.

sunshowers:
Related: https://github.com/oxidecomputer/omicron/issues/3480

davepacheco:
@sunshowers I think this issue was resolved as of #5032.  Is that right?

davepacheco:
Answering my own question: we discussed this briefly in today's update call.  This ticket is different from #5032 in two ways: it covers not just an underlying database facility, but an API and tool that can be used by support (at least) to control this behavior; and it covers dataset-level granularity, whereas #5032 only covers sled-level.

PR #3526, opened 2023-07-07
Add support for 'disabling' physical disks to prevent future allocations

smklein:
Part of https://github.com/oxidecomputer/omicron/issues/3480 

Fixes https://github.com/oxidecomputer/omicron/issues/5153


openapi/nexus.json
```diff
@@ -3436,6 +3436,69 @@
         }
       }
     },
+    "/v1/system/hardware/disks/{vendor}/{serial}/{model}": {
```
david-crespo, line 3439: A series of open identifiers can create issues, e.g., you can't add `/vendor/specific-page`. We generally try to avoid it.

zephraph: Is there a reason why we can't (or shouldn't) use the UUID associated with the DB entry?

smklein: I would like to, but I need a variant of the `LookupPath` that can lookup by **either** UUID or (vendor, serial, model). We currently look up the disk by (vendor, serial, model) from the Sled Agent's calls to the internal API.

smklein: Gave this a shot -- it requires an intermediate DB operation to translate from UUID -> (vendor, serial, model), which could probably be avoided if we could use the string tuple as a name.

david-crespo: I think that's much better from an API POV, thanks.

schema/crdb/dbinit.sql
```diff
@@ -312,6 +326,9 @@ CREATE TABLE IF NOT EXISTS omicron.public.physical_disk (
     -- FK into the Sled table
     sled_id UUID NOT NULL,
 
+    -- Describes how the control plane manages this disk
+    state omicron.public.physical_disk_state NOT NULL,
+
```
smklein, line 331: TODO(me): add an index to look up all "draining" disks". 

Also maybe update the terminology here to match https://github.com/oxidecomputer/omicron/issues/4719

smklein: Nit: I should use the term "deactivating" here, rather than "draining", because "draining" can refer to cooperative removal (e.g. in the context of draining instances from a sled)

smklein:
During the update sync, we had an interesting discussion where we floated the following idea: Instead of storing the "State" in the disk itself, what if state was **implied** by presence in a blueprint?

In this world:
- A disk that appears in a blueprint is implicitly active
- A disk that is not in the blueprint is not active (distinguishing between `deactivating` or `inactive` is a little more subtle?)

I'm going to explore this space a bit more, since I think it would fundamentally change how the "disk state" is stored in the DB.

smklein:
Closing in favor of https://github.com/oxidecomputer/omicron/pull/5335

Issue #2124, opened 2023-01-10
Some named objects in DB lacking UNIQUE name index


smklein:
Some objects in the DB have a name, and a corresponding index to ensure the name remains unique.

For example:

```sql
CREATE TABLE omicron.public.silo (
    /* Identity metadata */
    id UUID PRIMARY KEY,
    name STRING(63) NOT NULL,
    description STRING(512) NOT NULL,
    time_created TIMESTAMPTZ NOT NULL,
    time_modified TIMESTAMPTZ NOT NULL,
    time_deleted TIMESTAMPTZ,
    ...
);

CREATE UNIQUE INDEX ON omicron.public.silo (
    name
) WHERE
    time_deleted IS NULL;
```

The following objects within the database have a `name` field, but no `UNIQUE` index on the name:

- [ ] [omicron.public.saml_identity_provider](https://github.com/oxidecomputer/omicron/blob/1b995f6d14fd95b3e8a5af1e049cc2a27eae3e54/common/src/sql/dbinit.sql#L437-L463): Has an index on `id` + `silo_id`, but no name-based index. This *might* be fine, if the assumption exists that "only one can ever exist within the scope of a silo anyway".
- [ ] [omicron.public.external_ip](https://github.com/oxidecomputer/omicron/blob/1b995f6d14fd95b3e8a5af1e049cc2a27eae3e54/common/src/sql/dbinit.sql#L1195-L1251): Has no name-based index, but has the constraint that names are NOT NULL for `floating` ips. I believe this means that multiple floating IPs could exist with the same name.
- [ ] [omicron.public.saga](https://github.com/oxidecomputer/omicron/blob/1b995f6d14fd95b3e8a5af1e049cc2a27eae3e54/common/src/sql/dbinit.sql#L1298-L1323): I'm not sure if name uniqueness matters; these names are more like labels than unique identifiers.

davepacheco:
Good call!  For saml_identity_provider, I think the index needs to be there.  We do expect multiple IdPs in a Silo and they should not have overlapping names.  For saga, I think it's fine as is.  As you said, the name is not a unique identifier for these.

davepacheco:
See also #1534

davepacheco:
#3530 addresses the `saml_identity_provider` case and I think `saga` is fine as-is.

That leaves `external_ip`.  I'm not sure in what scope names are supposed to be unique for these.

luqmana:
[...]

PR #3530, opened 2023-07-07
clean up various non-unique indexes

davepacheco:
This addresses part (but not all) of #1497 and #2124.

This could use some careful review.

A common pattern here is that we had a non-unique index on some table T property P, like table `instance` property `active_sled_id`.  I think people usually added these because they wanted to be able to find the rows having a certain value for that property (e.g., Instances on a given Sled).  I believe this only works if you don't plan to paginate that query.  If you do, you need a pagination key, and that needs to be in the index, too, or else Cockroach can't use the index to serve the paginated query.  That's why the index on `instance` is now on `(active_sled_id, id)`.  With the id there, this can be a `UNIQUE` index.

### DIFF
--- a/common/src/sql/dbinit.sql
+++ b/common/src/sql/dbinit.sql
@@ -717,8 +717,6 @@ CREATE UNIQUE INDEX ON omicron.public.saml_identity_provider (
 ) WHERE
     time_deleted IS NULL;
 
-
-
 /*
  * Users' public SSH keys, per RFD 44
  */
--- a/common/src/sql/dbinit.sql
+++ b/common/src/sql/dbinit.sql
@@ -107,14 +107,10 @@ CREATE TABLE omicron.public.sled (
 );
 
 /* Add an index which lets us look up sleds on a rack */
-CREATE INDEX ON omicron.public.sled (
-    rack_id
-) WHERE time_deleted IS NULL;
-
-CREATE INDEX ON omicron.public.sled (
+CREATE UNIQUE INDEX ON omicron.public.sled (
+    rack_id,
     id
-) WHERE
-    time_deleted IS NULL;
+) WHERE time_deleted IS NULL;
 
 CREATE TYPE omicron.public.sled_resource_kind AS ENUM (
     -- omicron.public.Dataset
@@ -152,8 +148,9 @@ CREATE TABLE omicron.public.sled_resource (
 );
 
 -- Allow looking up all resources which reside on a sled
-CREATE INDEX ON omicron.public.sled_resource (
-    sled_id
+CREATE UNIQUE INDEX ON omicron.public.sled_resource (
+    sled_id,
+    id
 );
 
 /*
@@ -178,14 +175,10 @@ CREATE TABLE omicron.public.switch (
 );
 
 /* Add an index which lets us look up switches on a rack */
-CREATE INDEX ON omicron.public.switch (
-    rack_id
-) WHERE time_deleted IS NULL;
-
-CREATE INDEX ON omicron.public.switch (
+CREATE UNIQUE INDEX ON omicron.public.switch (
+    rack_id,
     id
-) WHERE
-    time_deleted IS NULL;
+) WHERE time_deleted IS NULL;
 
 /*
  * Services
@@ -224,12 +217,13 @@ CREATE TABLE omicron.public.service (
 );
 
 /* Add an index which lets us look up the services on a sled */
-CREATE INDEX ON omicron.public.service (
+CREATE UNIQUE INDEX ON omicron.public.service (
     sled_id,
     id
 );
 
-CREATE INDEX ON omicron.public.service (
+/* Look up (and paginate) services of a given kind. */
+CREATE UNIQUE INDEX ON omicron.public.service (
     kind,
     id
 );
@@ -263,13 +257,13 @@ CREATE TABLE omicron.public.physical_disk (
     )
 );
 
-CREATE INDEX ON omicron.public.physical_disk (
+CREATE UNIQUE INDEX ON omicron.public.physical_disk (
     variant,
     id
 ) WHERE time_deleted IS NULL;
 
 -- Make it efficient to look up physical disks by Sled.
-CREATE INDEX ON omicron.public.physical_disk (
+CREATE UNIQUE INDEX ON omicron.public.physical_disk (
     sled_id,
     id
 ) WHERE time_deleted IS NULL;
@@ -299,7 +293,7 @@ CREATE TABLE omicron.public.certificate (
 
 -- Add an index which lets us look up certificates for a particular service
 -- class.
-CREATE INDEX ON omicron.public.certificate (
+CREATE UNIQUE INDEX ON omicron.public.certificate (
     service,
     id
 ) WHERE
@@ -442,7 +436,7 @@ CREATE INDEX on omicron.public.Dataset (
 /*
  * A region of space allocated to Crucible Downstairs, within a dataset.
  */
-CREATE TABLE omicron.public.Region (
+CREATE TABLE omicron.public.region (
     /* Identity metadata (asset) */
     id UUID PRIMARY KEY,
     time_created TIMESTAMPTZ NOT NULL,
@@ -463,15 +457,17 @@ CREATE TABLE omicron.public.Region (
 /*
  * Allow all regions belonging to a disk to be accessed quickly.
  */
-CREATE INDEX on omicron.public.Region (
-    volume_id
+CREATE UNIQUE INDEX on omicron.public.region (
+    volume_id,
+    id
 );
 
 /*
  * Allow all regions belonging to a dataset to be accessed quickly.
  */
-CREATE INDEX on omicron.public.Region (
-    dataset_id
+CREATE UNIQUE INDEX on omicron.public.region (
+    dataset_id,
+    id
 );
 
 /*
@@ -669,15 +665,15 @@ CREATE TABLE omicron.public.identity_provider (
     provider_type omicron.public.provider_type NOT NULL
 );
 
-CREATE INDEX ON omicron.public.identity_provider (
-    id,
-    silo_id
+CREATE UNIQUE INDEX ON omicron.public.identity_provider (
+    silo_id,
+    id
 ) WHERE
     time_deleted IS NULL;
 
-CREATE INDEX ON omicron.public.identity_provider (
-    name,
-    silo_id
+CREATE UNIQUE INDEX ON omicron.public.identity_provider (
+    silo_id,
+    name
 ) WHERE
     time_deleted IS NULL;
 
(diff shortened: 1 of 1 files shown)


common/src/sql/dbinit.sql
```diff
@@ -107,14 +107,10 @@ CREATE TABLE omicron.public.sled (
...
-    rack_id
-) WHERE time_deleted IS NULL;
-
-CREATE INDEX ON omicron.public.sled (
+CREATE UNIQUE INDEX ON omicron.public.sled (
+    rack_id,
     id
-) WHERE
-    time_deleted IS NULL;
+) WHERE time_deleted IS NULL;
```
davepacheco, line 113: This is one of the cases mentioned in the PR description.  If we want to paginate the sleds in a rack, we need the sled id in this index, and then it can also be unique.

common/src/sql/dbinit.sql
```diff
@@ -107,14 +107,10 @@ CREATE TABLE omicron.public.sled (
 );
 
 /* Add an index which lets us look up sleds on a rack */
-CREATE INDEX ON omicron.public.sled (
-    rack_id
-) WHERE time_deleted IS NULL;
-
-CREATE INDEX ON omicron.public.sled (
```
davepacheco, line 114: This index is superfluous because [CockroachDB always has an index on the primary key](https://www.cockroachlabs.com/docs/stable/primary-key.html#details).  (Note that having an index on "rack_id" and one on "id" is not the same as having one on `(rack_id, id)`.)

common/src/sql/dbinit.sql
```diff
@@ -152,8 +148,9 @@ CREATE TABLE omicron.public.sled_resource (
 );
 
 -- Allow looking up all resources which reside on a sled
-CREATE INDEX ON omicron.public.sled_resource (
-    sled_id
+CREATE UNIQUE INDEX ON omicron.public.sled_resource (
```
davepacheco, line 151: This is one of the cases mentioned in the PR description.  If we want tp aginate the sled_resources on a sled, we need the "id" here, too.

common/src/sql/dbinit.sql
```diff
@@ -178,14 +175,10 @@ CREATE TABLE omicron.public.switch (
 );
 
 /* Add an index which lets us look up switches on a rack */
-CREATE INDEX ON omicron.public.switch (
-    rack_id
-) WHERE time_deleted IS NULL;
-
-CREATE INDEX ON omicron.public.switch (
+CREATE UNIQUE INDEX ON omicron.public.switch (
```
davepacheco, line 178: Another of the cases mentioned in the PR description.  This one's arguably unnecessary since we don't expect to have many switches in a rack.

common/src/sql/dbinit.sql
```diff
@@ -178,14 +175,10 @@ CREATE TABLE omicron.public.switch (
 );
 
 /* Add an index which lets us look up switches on a rack */
-CREATE INDEX ON omicron.public.switch (
-    rack_id
-) WHERE time_deleted IS NULL;
-
-CREATE INDEX ON omicron.public.switch (
```
davepacheco, line 185: Another superfluous index because the primary key always has an index.

common/src/sql/dbinit.sql
```diff
@@ -224,12 +217,13 @@ CREATE TABLE omicron.public.service (
 );
 
 /* Add an index which lets us look up the services on a sled */
-CREATE INDEX ON omicron.public.service (
+CREATE UNIQUE INDEX ON omicron.public.service (
```
davepacheco, line 220: This should already have been unique since it includes the primary key.

common/src/sql/dbinit.sql
```diff
@@ -224,12 +217,13 @@ CREATE TABLE omicron.public.service (
...
     sled_id,
     id
 );
 
-CREATE INDEX ON omicron.public.service (
+/* Look up (and paginate) services of a given kind. */
+CREATE UNIQUE INDEX ON omicron.public.service (
```
davepacheco, line 226: This should already have been unique since it includes the primary key.

common/src/sql/dbinit.sql
```diff
@@ -263,13 +257,13 @@ CREATE TABLE omicron.public.physical_disk (
     )
 );
 
-CREATE INDEX ON omicron.public.physical_disk (
+CREATE UNIQUE INDEX ON omicron.public.physical_disk (
```
davepacheco, line 260: This should already have been unique since it includes the primary key.

common/src/sql/dbinit.sql
```diff
@@ -263,13 +257,13 @@ CREATE TABLE omicron.public.physical_disk (
...
     variant,
     id
 ) WHERE time_deleted IS NULL;
 
 -- Make it efficient to look up physical disks by Sled.
-CREATE INDEX ON omicron.public.physical_disk (
+CREATE UNIQUE INDEX ON omicron.public.physical_disk (
```
davepacheco, line 266: This should already have been unique since it includes the primary key.

common/src/sql/dbinit.sql
```diff
@@ -299,7 +293,7 @@ CREATE TABLE omicron.public.certificate (
 
 -- Add an index which lets us look up certificates for a particular service
 -- class.
-CREATE INDEX ON omicron.public.certificate (
+CREATE UNIQUE INDEX ON omicron.public.certificate (
```
davepacheco, line 296: This should already have been unique since it includes the primary key.

common/src/sql/dbinit.sql
```diff
@@ -442,7 +436,7 @@ CREATE INDEX on omicron.public.Dataset (
 /*
  * A region of space allocated to Crucible Downstairs, within a dataset.
  */
-CREATE TABLE omicron.public.Region (
+CREATE TABLE omicron.public.region (
```
davepacheco, line 439: This was one of three tables that were capitalized.

common/src/sql/dbinit.sql
```diff
@@ -374,7 +368,7 @@ CREATE TABLE omicron.public.virtual_provisioning_resource (
  * ZPools of Storage, attached to Sleds.
  * These are backed by a single physical disk.
  */
-CREATE TABLE omicron.public.Zpool (
+CREATE TABLE omicron.public.zpool (
```
davepacheco, line 371: This was one of three tables that were capitalized.

common/src/sql/dbinit.sql
```diff
@@ -402,7 +396,7 @@ CREATE TYPE omicron.public.dataset_kind AS ENUM (
 /*
  * A dataset of allocated space within a zpool.
  */
-CREATE TABLE omicron.public.Dataset (
+CREATE TABLE omicron.public.dataset (
```
davepacheco, line 399: This was one of three tables that were capitalized.

common/src/sql/dbinit.sql
```diff
@@ -463,15 +457,17 @@ CREATE TABLE omicron.public.Region (
 /*
  * Allow all regions belonging to a disk to be accessed quickly.
  */
-CREATE INDEX on omicron.public.Region (
-    volume_id
+CREATE UNIQUE INDEX on omicron.public.region (
+    volume_id,
+    id
```
davepacheco, line 462: This is one of the cases mentioned in the PR description.  To paginate the regions on a disk, the id needs to be here, and then this can be a `UNIQUE` index.

common/src/sql/dbinit.sql
```diff
@@ -463,15 +457,17 @@ CREATE TABLE omicron.public.Region (
...
 );
 
 /*
  * Allow all regions belonging to a dataset to be accessed quickly.
  */
-CREATE INDEX on omicron.public.Region (
-    dataset_id
+CREATE UNIQUE INDEX on omicron.public.region (
```
davepacheco, line 468: Same as above.

common/src/sql/dbinit.sql
```diff
@@ -669,15 +665,15 @@ CREATE TABLE omicron.public.identity_provider (
     provider_type omicron.public.provider_type NOT NULL
 );
 
-CREATE INDEX ON omicron.public.identity_provider (
-    id,
-    silo_id
+CREATE UNIQUE INDEX ON omicron.public.identity_provider (
```
davepacheco, line 668: This one should already be unique because it contains the primary key.

I changed the order of the fields.  As it was, I don't think this index could be used to paginate through the identity providers in a Silo (which I think was the point of this index).

common/src/sql/dbinit.sql
```diff
@@ -669,15 +665,15 @@ CREATE TABLE omicron.public.identity_provider (
...
+    id
 ) WHERE
     time_deleted IS NULL;
 
-CREATE INDEX ON omicron.public.identity_provider (
-    name,
-    silo_id
+CREATE UNIQUE INDEX ON omicron.public.identity_provider (
+    silo_id,
+    name
```
davepacheco, line 676: **This is one of the few functional changes in this PR.**  This index was _not_ unique and I think that was a bug.  See [this comment](https://github.com/oxidecomputer/omicron/issues/1497#issuecomment-1464072388).

I also flipped the order of fields in order to use this index to paginate the IdPs in a Silo by name.

common/src/sql/dbinit.sql
```diff
@@ -709,12 +705,20 @@ CREATE TABLE omicron.public.saml_identity_provider (
...
 );
 
-CREATE INDEX ON omicron.public.saml_identity_provider (
-    id,
-    silo_id
+CREATE UNIQUE INDEX ON omicron.public.saml_identity_provider (
+    silo_id,
+    id
+) WHERE
+    time_deleted IS NULL;
```
davepacheco, line 712: Same as above -- needed to flip the order of these fields.

common/src/sql/dbinit.sql
```diff
@@ -709,12 +705,20 @@ CREATE TABLE omicron.public.saml_identity_provider (
...
+
+CREATE UNIQUE INDEX ON omicron.public.saml_identity_provider (
+    silo_id,
+    name
```
davepacheco, line 716: I added this to ensure that SamlIdentityProvider names are unique within a Silo.

common/src/sql/dbinit.sql
```diff
@@ -861,8 +865,9 @@ CREATE UNIQUE INDEX ON omicron.public.instance (
 
 -- Allow looking up instances by server. This is particularly
 -- useful for resource accounting within a sled.
-CREATE INDEX ON omicron.public.instance (
-    active_sled_id
+CREATE UNIQUE INDEX ON omicron.public.instance (
```
davepacheco, line 868: This is one of the cases mentioned in the PR description. To paginate the instances on a sled, the id needs to be here, and then this can be a UNIQUE index.

common/src/sql/dbinit.sql
```diff
@@ -976,8 +981,9 @@ CREATE UNIQUE INDEX ON omicron.public.disk (
 ) WHERE
     time_deleted IS NULL;
 
-CREATE INDEX ON omicron.public.disk (
-    attach_instance_id
+CREATE UNIQUE INDEX ON omicron.public.disk (
```
davepacheco, line 984: This is one of the cases mentioned in the PR description. To paginate the disks attached to an instance, the id needs to be here, and then this can be a UNIQUE index.

common/src/sql/dbinit.sql
luqmana: This index also includes a filter predicate `time_deleted IS NULL` which makes it a partial index no? Perhaps not super relevant for this particular table but in the limit there could be more old/deleted records that a partial index could speed up access for only live records.

davepacheco: Thanks for pointing that out.  I had missed that.  Are you saying that even if CockroachDB already indexes on "id", it might be beneficial to have an index on "id" limited only to live rows because querying it could be faster?  Yeah, tThat's definitely possible.  I think we probably want to wait and see if that's true, though, since it's somewhat speculative and has downsides (i.e., the extra disk space and memory usage and cost to update it on writes).

luqmana (approved):

Issue #3515, opened 2023-07-07
Clickhouse set data retention time limits


karencfv:
Currently we have no data retention time limits for ClickHouse. For now we will set an arbitrary TTL (time to live) to 30 days with data rollup.

Once we have feedback from customers we may want to change this limit. Further down the track we'll want users to be able to set their own TTL in accordance to any data retention policies they may have.

PR #3553, opened 2023-07-10
Implement a TTL of 1 month to each metric

karencfv:
Closes: https://github.com/oxidecomputer/omicron/issues/3515

This 30 day long lifetime is arbitrary. It is only meant to keep the database from growing infinitely. If we receive feedback from our first customers that they have a specific data retention policy, we can modify.

In the future we will most likely allow our users to set their own metric TTL in accordance with their company data retention policies.

We will also implement [data rollup](https://clickhouse.com/docs/en/guides/developer/ttl#implementing-a-rollup) when deleting records. I initially meant to implement them as part of this PR, but we still don't know what type of aggregations will be most useful, so I decided to hold off on this. We need a TTL for FCS but TTL data rollup aggregations can wait until we have a better understanding of our customer's needs.

### DIFF
--- a/oximeter/db/src/db-init.sql
+++ b/oximeter/db/src/db-init.sql
@@ -8,7 +8,8 @@ CREATE TABLE IF NOT EXISTS oximeter.measurements_bool
     datum UInt8
 )
 ENGINE = MergeTree()
-ORDER BY (timeseries_name, timeseries_key, timestamp);
+ORDER BY (timeseries_name, timeseries_key, timestamp)
+TTL timestamp + INTERVAL 1 MONTH;
 --
 CREATE TABLE IF NOT EXISTS oximeter.measurements_i64
 (
@@ -18,7 +19,8 @@ CREATE TABLE IF NOT EXISTS oximeter.measurements_i64
     datum Int64
 )
 ENGINE = MergeTree()
-ORDER BY (timeseries_name, timeseries_key, timestamp);
+ORDER BY (timeseries_name, timeseries_key, timestamp)
+TTL timestamp + INTERVAL 1 MONTH;
 --
 CREATE TABLE IF NOT EXISTS oximeter.measurements_f64
 (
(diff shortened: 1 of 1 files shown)


karencfv:
Looks like TTL does not support `DateTime64(9, 'UTC')`. Will look into this when I'm back

Related -> https://github.com/ClickHouse/ClickHouse/issues/33493

smklein (approved):
If we're closing #3515 , we should make another issue for the work involved with rollup.

I'm also reading https://clickhouse.com/docs/en/guides/developer/ttl#triggering-ttl-events - 

> The deleting or aggregating of expired rows is not immediate - it only occurs during table merges. 

Does this mean we need to perform additional work to get these TTLs to trigger? Related: have you been able to test this manually with smaller TTLs (or by adjusting your system clock)?

karencfv:
> Does this mean we need to perform additional work to get these TTLs to trigger? 

The merges get triggered by default every 4 hours. They mention this a few sentences down in that link:

> So by default, your TTL rules will be applied to your table at least once every 4 hours. Just modify the settings above if you need your TTL rules applied more frequently.

Given that we're giving each row 30 days to live, I don't think it's necessary to tweak these default TTL settings. What do you think?

I tested these manually with a shorter TTL, it all worked nicely but I didn't make any screenshots (oops!). I can test them again.

smklein:
> Given that we're giving each row 30 days to live, I don't think it's necessary to tweak these default TTL settings. What do you think?

Yeah, sounds good, we can start using these defaults.

> I tested these manually with a shorter TTL, it all worked nicely but I didn't make any screenshots (oops!). I can test them again.

Your call! Might be cool to someday pull out the TTL as a customizable variable so we could make an automated end-to-end test for this behavior, but I also know that time is not cheap these days. I defer to your judgement, as long as you had a chance to manually test in some form!

karencfv:
Tested with a 30 second TTL on a simulated omicron:

Screenshot from 2023-07-18 18-40-21

> Your call! Might be cool to someday pull out the TTL as a customizable variable so we could make an automated end-to-end test for this behavior, but I also know that time is not cheap these days. I defer to your judgement, as long as you had a chance to manually test in some form!

We'll definitely have to make these TTL settings customizable. At some point every customer is going to come along with their own data retention policies :sweat_smile: I'll make an issue for this

PR #3556, opened 2023-07-10
Expose the physical disk form factor

zephraph:
As a follow up to #3496 @smklein, @rmustacc, and I chatted in a few context about still needing to be able to differentiate the disk types. It was agreed that the raw form factor is at least a fact that's not context dependent that we can use as a discriminate for now. 

I've done the plumbing to pipe that through the view layer. One note was that `PhysicalDiskKind` was part of the internal api so I moved it to the external API.


smklein (approved):
Looks good, thanks for cleaning this up!

david-crespo (approved):

Issue #3533, opened 2023-07-08
zone logs should not be in rpool


bcantrill:
Currently, our zones are created from the RAM disk zpool (`rpool`).  A challenge with that in particular is that zone logs are now both volatile (a challenge for debugging if a sled is bounced) and -- more pressingly -- very tightly capped.  In particular we had an instance where a pantry log was 21.6G (a feat only made possible by the ~10X compression factor we were getting there).  Once things right out of space, the world gets grim in a hurry.  We want to trim logs and log less loquaciously and so on -- but we also want/need a bit more breathing room by getting zones either onto one of the `oxi_` (M.2) pools or one of the `oxp_` (U.2) pools.  Because we saw this in dogfood, marking this MVP.

davepacheco:
Related: #2478

PR #3557, opened 2023-07-11
[sled agent] Store zone filesystems on U.2s, not the ramdisk

smklein:
- Removes all references to `rpool/zones`
- Exclusively stores zone filesystems on the U.2, under an encrypted dataset
- Wipes these datasets from the U.2s when parsing them on sled agent boot
- Tangentially related: Removes a handful of low-quality Sled Agent tests, reliant heavily on mock interfaces (see #2422 , though we still have work to improve this test coverage).

Fixes https://github.com/oxidecomputer/omicron/issues/3533


illumos-utils/src/zfs.rs
```diff
@@ -370,14 +383,5 @@ pub fn get_all_omicron_datasets_for_delete() -> anyhow::Result<Vec<String>> {
             datasets.push(format!("{pool}/{dataset}"));
         }
     }
-
-    // Collect all datasets for ramdisk-based Oxide zones,
-    // if any exist.
```
smklein, line 375: We aren't using these, so stop trying to delete 'em

illumos-utils/src/zfs.rs
```diff
@@ -23,13 +21,21 @@ pub struct ListDatasetsError {
...
 }
 
+#[derive(thiserror::Error, Debug)]
+pub enum DestroyDatasetErrorVariant {
+    #[error("Dataset not found")]
+    NotFound,
+    #[error(transparent)]
+    Other(crate::ExecutionError),
+}
+
```
smklein, line 31: We have a more specific error here so that the caller can get a better idea of if we tried to delete a dataset that doesn't exist (I mean, go figure, but it's relevant now because we're doing this on boot for the zone datasets).

sled-agent/src/bootstrap/agent.rs
```diff
@@ -341,27 +338,6 @@ impl Agent {
         let ddmd_client = DdmAdminClient::localhost(&log)?;
         ddmd_client.advertise_prefix(Ipv6Subnet::new(ip));
 
-        // Before we start creating zones, we need to ensure that the
-        // necessary ZFS and Zone resources are ready.
```
smklein, line 345: Removing this because we should be off the ramdisk.

sled-agent/src/instance.rs
```diff
@@ -856,7 +827,14 @@ impl Instance {
...
         let zname = propolis_zone_name(inner.propolis_id());
-        let root = camino::Utf8Path::new(ZONE_ZFS_RAMDISK_DATASET_MOUNTPOINT);
+        let mut rng = rand::rngs::StdRng::from_entropy();
+        let root = inner
+            .storage
+            .all_u2_mountpoints(ZONE_DATASET)
+            .await
+            .choose(&mut rng)
+            .ok_or_else(|| Error::U2NotFound)?
+            .clone();
```
smklein, line 837: We do this in a couple spots, but our policy is roughly the same:

- List all the U.2 "ZONE_DATASET" mountpoints
- Pick one randomly

We aren't accounting for space here, but we also delete these on reboot

sled-agent/src/instance_manager.rs
```diff
@@ -20,11 +22,6 @@ use std::collections::BTreeMap;
 use std::sync::{Arc, Mutex};
 use uuid::Uuid;
 
-#[cfg(not(test))]
-use crate::instance::Instance;
-#[cfg(test)]
-use crate::instance::MockInstance as Instance;
-
```
smklein, line 27: I'm ripping off this band-aid a little early, but these mock-based tests are terrible (see: https://github.com/oxidecomputer/omicron/issues/2422 ) and rather than making them work with this U.2 integration, I got rid of some of them.

sled-agent/src/services.rs
```diff
@@ -3201,7 +3212,7 @@ mod test {
             SidecarRevision::Physical("rev-test".to_string()),
             SWITCH_ZONE_BOOTSTRAP_IP,
             vec![],
-            StorageManager::new(&log, storage_key_requester).await,
+            StorageResources::new_for_test(),
```
smklein, line 3215: We don't need access to the full `StorageManager` here -- we aren't going to be adding new disks here -- we just want to know about "what U.2s exist on the system", which is a smaller interface.

sled-agent/src/services.rs
```diff
@@ -368,7 +373,7 @@ pub struct ServiceManagerInner {
     // TODO(https://github.com/oxidecomputer/omicron/issues/2888): We will
```
andrewjstone, line 373: Can this comment be removed now?

andrewjstone (approved):
So nice to have this change go in!

sled-agent/src/services.rs
smklein: Good call, removed

PR #3560, opened 2023-07-11
[installinator] Temporary workaround for #3507

jgallagher:


sunshowers (approved):
seems legit

sunshowers:
(now that we've narrowed 3507 down to a small section of virtual address space that's busted, this is quite reasonable)

PR #3561, opened 2023-07-11
[installinator-artifactd] accept messages with generic type for metadata

sunshowers:
Currently, a scenario is possible where the installinator has a wire
protocol mismatch with wicketd, causing messages to be rejected by
wicketd even though they have enough semantic information to go through
fine.

Instead, accept arbitrary messages and only deserialize to
installinator's types when necessary.


sunshowers:
This would be good to have still but should be redone.

PR #3564, opened 2023-07-11
deploy multiple Nexus zones

davepacheco:
This required two changes:

- In RSS, deploying 3 zones instead of 1.  I chose three because it's the minimum required to survive a double failure.  More zones should be possible, but take more resources for testing and are also more annoying to work with.
- In Sled Agent, putting the service id into the zone name.  Otherwise, the zonename for the Nexus zone is the same for every zone on the box, and that results in Sled Agent only provisioning one of them.  (When it goes to provision the other two, it thinks it's already done it.)

### DIFF
--- a/sled-agent/src/params.rs
+++ b/sled-agent/src/params.rs
@@ -550,7 +550,27 @@ impl ServiceZoneRequest {
 
     // The name of a unique identifier for the zone, if one is necessary.
     pub fn zone_name_unique_identifier(&self) -> Option<String> {
-        self.dataset.as_ref().map(|d| d.name.pool().to_string())
+        self.dataset.as_ref().map(|d| d.name.pool().to_string()).or_else(|| {
+            match &self.zone_type {
+                // The switch zone is necessarily a singleton.
+                ZoneType::Switch => None,
+
+                // We should never get here for the following zones because they
+                // have a dataset.
+                ZoneType::Clickhouse
+                | ZoneType::CockroachDb
+                | ZoneType::Crucible
+                | ZoneType::ExternalDns
+                | ZoneType::InternalDns => None,
+
+                // Other zones should be able to have multiple instances on the
+                // same Sled.  Not all have been tested yet.
+                ZoneType::Nexus => Some(self.id.to_string()),
+                ZoneType::CruciblePantry
+                | ZoneType::Ntp
+                | ZoneType::Oximeter => None,
+            }
+        })
     }
 }
 
--- a/sled-agent/src/rack_setup/plan/service.rs
+++ b/sled-agent/src/rack_setup/plan/service.rs
@@ -44,7 +44,7 @@ use uuid::Uuid;
 const BOUNDARY_NTP_COUNT: usize = 2;
 
 // The number of Nexus instances to create from RSS.
-const NEXUS_COUNT: usize = 1;
+const NEXUS_COUNT: usize = 3;
 
 // The number of CRDB instances to create from RSS.
 const CRDB_COUNT: usize = 5;


sled-agent/src/params.rs
```diff
@@ -550,7 +550,27 @@ impl ServiceZoneRequest {
...
+                | ZoneType::InternalDns => None,
+
+                // Other zones should be able to have multiple instances on the
+                // same Sled.  Not all have been tested yet.
+                ZoneType::Nexus => Some(self.id.to_string()),
+                ZoneType::CruciblePantry
+                | ZoneType::Ntp
+                | ZoneType::Oximeter => None,
+            }
+        })
```
davepacheco, line 573: I chickened out a little bit here.

- There's a bunch of stuff that assumes `oxz_switch`.  They could be looking for zones of the right type instead, but I didn't want to unpack all that here.
- Similar with `oxz_oximeter`, though that was just in the tests.
- Zones that have a dataset already have their *pool name* in the zone name.  I was reluctant to change this because I don't know what, if anything, depends on it.

davepacheco:
On maxwell, Omicron sets up with my changes.  And I seem to be able to use all three Nexus instances.  They're all in external DNS:

```
$ dig recovery.sys.oxide.test @192.168.1.20 +short
192.168.1.21
192.168.1.22
192.168.1.23
```

I can hit each one:

```
$ curl --cacert $PWD/smf/sled-agent/non-gimlet/initial-tls-cert.pem --resolve recovery.sys.oxide.test:443:192.168.1.21 -i https://recovery.sys.oxide.test
HTTP/2 302 
location: /login/recovery/local?redirect_uri=%2F
x-request-id: 5ef235c6-a01f-49b0-a679-011084a87146
date: Tue, 11 Jul 2023 22:49:42 GMT

$ curl --cacert $PWD/smf/sled-agent/non-gimlet/initial-tls-cert.pem --resolve recovery.sys.oxide.test:443:192.168.1.22 -i https://recovery.sys.oxide.test
HTTP/2 302 
location: /login/recovery/local?redirect_uri=%2F
x-request-id: 377faf2e-2bd0-487c-b053-d6a7ac24ab02
date: Tue, 11 Jul 2023 22:49:47 GMT

$ curl --cacert $PWD/smf/sled-agent/non-gimlet/initial-tls-cert.pem --resolve recovery.sys.oxide.test:443:192.168.1.23 -i https://recovery.sys.oxide.test
HTTP/2 302 
location: /login/recovery/local?redirect_uri=%2F
x-request-id: e8c524ff-9aea-4a77-8687-e821acceed4a
date: Tue, 11 Jul 2023 22:49:53 GMT

```

and each request went to a different zone:

```
$ pfexec bash -c 'grep -l 5ef235c6-a01f-49b0-a679-011084a87146 /zone/oxz_nexus_*/root/var/svc/log/oxide-nexus*'
/zone/oxz_nexus_28187db2-33cd-4500-9c42-ff364b4189a5/root/var/svc/log/oxide-nexus:default.log
$ pfexec bash -c 'grep -l 377faf2e-2bd0-487c-b053-d6a7ac24ab02 /zone/oxz_nexus_*/root/var/svc/log/oxide-nexus*'
/zone/oxz_nexus_40fcf1da-1bdf-4c0b-a84e-ebc872b1562f/root/var/svc/log/oxide-nexus:default.log
$ pfexec bash -c 'grep -l e8c524ff-9aea-4a77-8687-e821acceed4a /zone/oxz_nexus_*/root/var/svc/log/oxide-nexus*'
/zone/oxz_nexus_23d13be3-3d74-4b79-8717-f3bb35db5e5b/root/var/svc/log/oxide-nexus:default.log
```

sled-agent/src/params.rs
smklein: I can try making this change as a follow-up -- I think consistency and predictability wins here, rather than the convention to match zone names with related resources.

smklein (approved):

sled-agent/src/params.rs
smklein: Done here: https://github.com/oxidecomputer/omicron/pull/3566

PR #3565, opened 2023-07-11
Implement lrtq network layer

andrewjstone:
LRTQ uses persistent TCP connections between peers. These connections have the
 following properties:

 1. Each peer learns of other peer addresses via DDM in the real system. These
 addresses are injected to the peer via `PeerHandle::load_peer_addresses`.
 2. Peers with "higher" IP addresses connect to peers with "lower" ip addresses - a
 "downhill" strategy
 3. Peers maintain persistence via application level `Ping` messages sent every
 second. Peers will timeout after not receiving *any* traffic for 10 seconds.
 4. Peers will constantly try to reconnect to any peers they have learned about with a
 "lower" IP address.

 The purpose of these TCP connections is to route LRTQ protocol traffic. This is
 done in the following manner:

 1. When a `TcpStream` is created via a connection from a higher to a lower peer,
 or accepted at a lower peer from a higher peer, the first thing that happens
 is a handshake.
 2. LRTQ messages may not be sent until the handshake completes.
 3. A handshake consists of two messages sent by each side of the connection:
  a. A `Hello` message used for scheme/version negotiation
  b. An `Identify` message used to inform each side of the connection to its
 peer's identity `Baseboard`. For accepting sides to map to the listening address
 of the peer, and not worry about ephemeral ports in testing on a single machine,
 we also send the canonical IP:Port in the `Identify` message.
 4. Once the handshake completes, LRTQ messages may be sent over the
 "established" connections.
 5. All messages except the `Hello` message, which is an exact size, are framed
 with a 4-byte big-endian size header.
 6. LRTQ message sending is triggered by `PeerHandle` method calls which send
 `PeerApiRequest`s to the underlying "main" peer task.
 7. Which messages need to get sent to other peers is determined by the underlying
 `Fsm`, which is managed by the main peer task.
 9. Many API requests, such as `PeerHandle::load_rack_secret`, cannot be answered
 immediately as they require sending messages to other peers and waiting for
 responses. A `responder` is set up for these requests, and when the `Fsm` returns
 a matching output or error, the response message will be sent over the responder. 
 The `PeerHandle` api will block until a response is received.
 8. Each connection has its own task and communicates with the main task over channels.
 10. Connection tasks serialize/deserialize and send/receive messages over the
 established `TcpStream`s.
 11. When a message is received on a connection, it is forwarded to the main peer
 task which calls `Fsm::handle_msg`.
 12. When a message needs to be sent out a connection it is sent from the main
 task to that connection managing task, which will serialize and send it over the network.

### DIFF
--- a/bootstore/src/schemes/v0/peer.rs
+++ b/bootstore/src/schemes/v0/peer.rs
@@ -20,7 +20,6 @@ use std::time::Duration;
 use tokio::net::{TcpListener, TcpStream};
 use tokio::sync::{mpsc, oneshot};
 use tokio::time::{interval, Instant, MissedTickBehavior};
-use uuid::Uuid;
 
 #[derive(Debug, Clone)]
 pub struct Config {
@@ -101,7 +100,7 @@ impl NodeHandle {
         let (tx, rx) = oneshot::channel();
         self.tx
             .send(NodeApiRequest::InitRack {
-                rack_uuid: RackUuid(Uuid::new_v4()),
+                rack_uuid,
                 initial_membership,
                 responder: tx,
             })
@@ -159,11 +158,11 @@ impl NodeHandle {
 
 #[derive(Debug, Clone)]
 pub struct Status {
-    fsm_state: &'static str,
-    peers: BTreeSet<SocketAddrV6>,
-    connections: BTreeMap<Baseboard, SocketAddrV6>,
-    accepted_connections: BTreeSet<SocketAddrV6>,
-    negotiating_connections: BTreeSet<SocketAddrV6>,
+    pub fsm_state: &'static str,
+    pub peers: BTreeSet<SocketAddrV6>,
+    pub connections: BTreeMap<Baseboard, SocketAddrV6>,
+    pub accepted_connections: BTreeSet<SocketAddrV6>,
+    pub negotiating_connections: BTreeSet<SocketAddrV6>,
 }
 
 /// A node in the bootstore protocol
@@ -730,7 +729,7 @@ impl Node {
 mod tests {
     use super::*;
     use slog::Drain;
-    use tokio::time::sleep;
+    use uuid::Uuid;
 
     fn initial_members() -> BTreeSet<Baseboard> {
         [("a", "1"), ("b", "1"), ("c", "1")]
@@ -804,9 +803,10 @@ mod tests {
         }
 
         let rack_uuid = RackUuid(Uuid::new_v4());
-        let output = handle0.init_rack(rack_uuid, initial_members()).await;
+        handle0.init_rack(rack_uuid, initial_members()).await.unwrap();
 
         let status = handle0.get_status().await;
+        println!("Status = {status:?}");
 
         // Ensure we can load the rack secret at all nodes
         handle0.load_rack_secret().await.unwrap();
@@ -818,8 +818,8 @@ mod tests {
 
         // Shutdown the node2 and make sure we can still load the rack
         // secret (threshold=2) at node0 and node1
-        handle2.shutdown().await;
-        jh2.await;
+        handle2.shutdown().await.unwrap();
+        jh2.await.unwrap();
         handle0.load_rack_secret().await.unwrap();
         handle1.load_rack_secret().await.unwrap();
 
@@ -840,14 +840,14 @@ mod tests {
 
         // Shutdown node1 and show that we can still load the rack secret at
         // node0 and the learner, because threshold=2 and it never changes.
-        handle1.shutdown().await;
-        jh1.await;
+        handle1.shutdown().await.unwrap();
+        jh1.await.unwrap();
         handle0.load_rack_secret().await.unwrap();
         learner_handle.load_rack_secret().await.unwrap();
 
         // Now shutdown the learner and show that node0 cannot load the rack secret
-        learner_handle.shutdown().await;
-        learner_jh.await;
+        learner_handle.shutdown().await.unwrap();
+        learner_jh.await.unwrap();
         handle0.load_rack_secret().await.unwrap_err();
 
         // TODO: Once we have persistence, we can bring an old node back from the dead
@@ -855,6 +855,6 @@ mod tests {
 
         // Shutdown node0
         handle0.shutdown().await.unwrap();
-        let _ = jh0.await;
+        jh0.await.unwrap();
     }
 }
--- a/bootstore/src/schemes/v0/peer.rs
+++ b/bootstore/src/schemes/v0/peer.rs
@@ -566,9 +566,6 @@ impl Node {
                 }
             }
             ConnToMainMsgInner::ConnectedInitiator { addr, peer_id } => {
-                // If we don't have a handle, it just means that the connection re-occurred
-                // inside the loop in `spawn_connection_initiator_task`, and we already
-                // have it in `established_connections`.
                 if let Some(handle) = self.initiating_connections.remove(&addr)
                 {
                     // Ignore the stale message if the unique_id doesn't match what
--- a/bootstore/src/schemes/v0/peer.rs
+++ b/bootstore/src/schemes/v0/peer.rs
@@ -462,6 +462,11 @@ impl Node {
                 // need to get the `request_id` from destructuring above
                 if let Some(responder) = self.rack_secret_responder.take() {
                     let _ = responder.send(Ok(secret));
+                } else {
+                    warn!(
+                        self.log,
+                        "Rack secret loaded, but no pending responder"
+                    );
                 }
             }
             ApiOutput::ShareDistributedToLearner => {
@@ -470,6 +475,11 @@ impl Node {
             ApiOutput::LearningCompleted => {
                 if let Some(responder) = self.init_responder.take() {
                     let _ = responder.send(Ok(()));
+                } else {
+                    warn!(
+                        self.log,
+                        "Learning completed, but no pending responder"
+                    );
                 }
                 // TODO: Persistence
             }
@@ -511,7 +521,7 @@ impl Node {
     async fn on_conn_msg(&mut self, msg: ConnToMainMsg) {
         let unique_id = msg.handle_unique_id;
         match msg.msg {
-            ConnToMainMsgInner::ConnectedServer {
+            ConnToMainMsgInner::ConnectedAcceptor {
                 accepted_addr,
                 addr,
                 peer_id,
@@ -555,16 +565,30 @@ impl Node {
                     error!(self.log, "Error on connection: {e}");
                 }
             }
-            ConnToMainMsgInner::ConnectedClient { addr, peer_id } => {
-                let handle = self.initiating_connections.remove(&addr).unwrap();
+            ConnToMainMsgInner::ConnectedInitiator { addr, peer_id } => {
+                // If we don't have a handle, it just means that the connection re-occurred
+                // inside the loop in `spawn_connection_initiator_task`, and we already
+                // have it in `established_connections`.
+                if let Some(handle) = self.initiating_connections.remove(&addr)
+                {
+                    // Ignore the stale message if the unique_id doesn't match what
+                    // we have stored.
+                    if unique_id != handle.unique_id {
+                        return;
+                    }
 
-                // Ignore the stale message if the unique_id doesn't match what
-                // we have stored.
-                if unique_id != handle.unique_id {
-                    return;
+                    self.established_connections
+                        .insert(peer_id.clone(), handle);
+                } else {
+                    error!(
+                        self.log,
+                        "Missing PeerConnHandle";
+                        "addr" => addr.to_string(),
+                        "remote_peer_id" => peer_id.to_string()
+                    );
+                    panic!("Missing PeerConnHandle");
                 }
 
-                self.established_connections.insert(peer_id.clone(), handle);
                 if let Err(e) =
                     self.fsm.on_connected(Instant::now().into(), peer_id)
                 {
@@ -583,9 +607,21 @@ impl Node {
                     if unique_id != handle.unique_id {
                         return;
                     }
+                } else {
+                    error!(
+                        self.log,
+                        "Missing PeerConnHandle";
+                        "remote_peer_id" => peer_id.to_string()
+                    );
+                    panic!("Missing PeerConnHandle");
                 }
                 warn!(self.log, "peer disconnected {peer_id}");
-                self.established_connections.remove(&peer_id);
+                let handle =
+                    self.established_connections.remove(&peer_id).unwrap();
+                if peer_id < self.config.id {
+                    // Put the connection handle back in initiating state
+                    self.initiating_connections.insert(handle.addr, handle);
+                }
                 self.fsm.on_disconnected(&peer_id);
             }
             ConnToMainMsgInner::Received { from, msg } => {
--- a/bootstore/src/schemes/v0/peer.rs
+++ b/bootstore/src/schemes/v0/peer.rs
@@ -630,9 +630,16 @@ impl Node {
                     Err(err) => self.handle_api_error(err).await,
                 }
             }
-            ConnToMainMsgInner::FailedServerHandshake {
+            ConnToMainMsgInner::FailedAcceptorHandshake {
                 addr: SocketAddrV6,
-            } => {}
+            } => {
+                if let Some(handle) = self.accepted_connections.get(&addr) {
+                    if handle.unique_id != unique_id {
+                        return;
+                    }
+                }
+                self.accepted_connections.remove(&addr);
+            }
         }
     }
 
--- a/bootstore/src/schemes/v0/peer.rs
+++ b/bootstore/src/schemes/v0/peer.rs
@@ -1,0 +1,777 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! The entrypoint of the v0 scheme for use by bootstrap agent
+
+use super::peer_networking::{
+    spawn_client, spawn_server, AcceptedConnHandle, ConnToMainMsg,
+    ConnToMainMsgInner, MainToConnMsg, Msg, PeerConnHandle,
+};
+use super::{ApiError, ApiOutput, Config as FsmConfig, Fsm, RackUuid};
+use crate::trust_quorum::RackSecret;
+use derive_more::From;
+use sled_hardware::Baseboard;
+use slog::{debug, error, info, o, warn, Logger};
+use std::collections::{BTreeMap, BTreeSet};
+use std::net::{SocketAddr, SocketAddrV6};
+use std::time::Duration;
+use tokio::net::{TcpListener, TcpStream};
+use tokio::sync::{mpsc, oneshot};
+use tokio::time::{interval, Instant, MissedTickBehavior};
+use uuid::Uuid;
+
+#[derive(Debug, Clone)]
+pub struct Config {
+    id: Baseboard,
+    addr: SocketAddrV6,
+    time_per_tick: Duration,
+    learn_timeout: Duration,
+    rack_init_timeout: Duration,
+    rack_secret_request_timeout: Duration,
+}
+
+// An error response from a `PeerRequest`
+#[derive(Debug, From)]
+pub enum PeerRequestError {
+    // An `Init_` or `LoadRackSecret` request is already outstanding
+    // We only allow one at a time.
+    RequestAlreadyPending,
+
+    // An error returned by the Fsm API
+    Fsm(ApiError),
+
+    // The peer task shutdown
+    Recv(oneshot::error::RecvError),
+
+    // Failed to send to a connection management task
+    Send(mpsc::error::SendError<PeerApiRequest>),
+}
+
+/// A request sent to the `Peer` task from the `PeerHandle`
+pub enum PeerApiRequest {
+    /// Initialize a rack at the behest of RSS running on the same scrimlet as this Peer
+    InitRack {
+        rack_uuid: RackUuid,
+        initial_membership: BTreeSet<Baseboard>,
+        responder: oneshot::Sender<Result<(), PeerRequestError>>,
+    },
+
+    /// Initialize this peer as a learner.
+    ///
+    /// Return `()` from the responder when the learner has learned its share
+    InitLearner { responder: oneshot::Sender<Result<(), PeerRequestError>> },
+
+    /// Load the rack secret.
+    ///
+    /// This can only be successfully called when a peer has been initialized,
+    /// either as initial member or learner who has learned its share.
+    LoadRackSecret {
+        responder: oneshot::Sender<Result<RackSecret, PeerRequestError>>,
+    },
+
+    /// Inform the peer of currently known IP addresses on the bootstrap network
+    ///
+    /// These are generated from DDM prefixes learned by the bootstrap agent.
+    PeerAddresses(BTreeSet<SocketAddrV6>),
+
+    /// Get the status of this peer
+    GetStatus { responder: oneshot::Sender<Status> },
+
+    /// Shutdown the peer
+    Shutdown,
+}
+
+// A handle for interacting with a `Peer` task
+pub struct PeerHandle {
+    tx: mpsc::Sender<PeerApiRequest>,
+}
+
+impl PeerHandle {
+    /// Initialize a rack at the behest of RSS running on the same scrimlet as
+    /// this Peer
+    pub async fn init_rack(
+        &self,
+        rack_uuid: RackUuid,
+        initial_membership: BTreeSet<Baseboard>,
+    ) -> Result<(), PeerRequestError> {
+        let (tx, rx) = oneshot::channel();
+        self.tx
+            .send(PeerApiRequest::InitRack {
+                rack_uuid: RackUuid(Uuid::new_v4()),
+                initial_membership,
+                responder: tx,
+            })
+            .await?;
+        let res = rx.await?;
+        res
+    }
+
+    /// Initialize this peer as a learner
+    pub async fn init_learner(&self) -> Result<(), PeerRequestError> {
+        let (tx, rx) = oneshot::channel();
+        self.tx.send(PeerApiRequest::InitLearner { responder: tx }).await?;
+        let res = rx.await?;
+        res
+    }
+
+    /// Load the rack secret.
+    ///
+    /// This can only be successfully called when a peer has been initialized,
+    /// either as initial member or learner who has learned its share.
+    pub async fn load_rack_secret(
+        &self,
+    ) -> Result<RackSecret, PeerRequestError> {
+        let (tx, rx) = oneshot::channel();
+        self.tx.send(PeerApiRequest::LoadRackSecret { responder: tx }).await?;
+        let res = rx.await?;
+        res
+    }
+
+    /// Inform the peer of currently known IP addresses on the bootstrap network
+    ///
+    /// These are generated from DDM prefixes learned by the bootstrap agent.
+    pub async fn load_peer_addresses(
+        &self,
+        addrs: BTreeSet<SocketAddrV6>,
+    ) -> Result<(), PeerRequestError> {
+        self.tx.send(PeerApiRequest::PeerAddresses(addrs)).await?;
+        Ok(())
+    }
+
+    /// Get the status of this peer
+    pub async fn get_status(&self) -> Result<Status, PeerRequestError> {
+        let (tx, rx) = oneshot::channel();
+        self.tx.send(PeerApiRequest::GetStatus { responder: tx }).await?;
+        let res = rx.await?;
+        Ok(res)
+    }
+
+    /// Shutdown the peer
+    pub async fn shutdown(&self) -> Result<(), PeerRequestError> {
+        self.tx.send(PeerApiRequest::Shutdown).await?;
+        Ok(())
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct Status {
+    fsm_state: &'static str,
+    peers: BTreeSet<SocketAddrV6>,
+    connections: BTreeMap<Baseboard, SocketAddrV6>,
+    accepted_connections: BTreeSet<SocketAddrV6>,
+    negotiating_connections: BTreeSet<SocketAddrV6>,
+}
+
+/// A peer in the bootstore protocol
+pub struct Peer {
+    config: Config,
+    fsm: Fsm,
+    peers: BTreeSet<SocketAddrV6>,
+    handle_unique_id_counter: u64,
+
+    // boolean set when a `PeerApiRequest::shutdown` is received
+    shutdown: bool,
+
+    // Connections that have been accepted, but where negotiation has not
+    // finished At this point, we only know the client port of the connection,
+    // and so cannot identify  it as a `Peer`.
+    accepted_connections: BTreeMap<SocketAddrV6, AcceptedConnHandle>,
+
+    // Connections that have not yet completed handshakes via `Hello` and
+    // `Identify` messages.
+    //
+    // We only store the client (connecting) side, not server (accepting) side here.
+    negotiating_connections: BTreeMap<SocketAddrV6, PeerConnHandle>,
+
+    // Active connections participating in scheme v0
+    //
+    // This consists of both client and server connections
+    connections: BTreeMap<Baseboard, PeerConnHandle>,
+
+    // Handle requests received from `PeerHandle`
+    rx: mpsc::Receiver<PeerApiRequest>,
+
+    // Used to respond to `InitRack` or `InitLearner` requests
+    init_responder: Option<oneshot::Sender<Result<(), PeerRequestError>>>,
+
+    // Used to respond to `LoadRackSecret` requests
+    rack_secret_responder:
+        Option<oneshot::Sender<Result<RackSecret, PeerRequestError>>>,
+
+    log: Logger,
+
+    // Handle messages received from connection tasks
+    conn_rx: mpsc::Receiver<ConnToMainMsg>,
+
+    // Clone for use by connection tasks to send to the main peer task
+    conn_tx: mpsc::Sender<ConnToMainMsg>,
+}
+
+impl From<Config> for FsmConfig {
+    fn from(value: Config) -> Self {
+        FsmConfig {
+            learn_timeout: value.learn_timeout,
+            rack_init_timeout: value.rack_init_timeout,
+            rack_secret_request_timeout: value.rack_secret_request_timeout,
+        }
+    }
+}
+
+impl Peer {
+    pub fn new(config: Config, log: &Logger) -> (Peer, PeerHandle) {
+        // We only expect one outstanding request at a time for `Init_` or
+        // `LoadRackSecret` requests, We can have one of those requests in
+        // flight while allowing `PeerAddresses` updates.
+        let (tx, rx) = mpsc::channel(3);
+
+        // Up to 31 sleds sending messages with some extra room. These are mostly one at a time
+        // for each sled, but we leave some extra room.
+        let (conn_tx, conn_rx) = mpsc::channel(128);
+        let fsm =
+            Fsm::new_uninitialized(config.id.clone(), config.clone().into());
+        let id_str = config.id.to_string();
+        (
+            Peer {
+                config,
+                fsm,
+                peers: BTreeSet::new(),
+                handle_unique_id_counter: 0,
+                shutdown: false,
+                accepted_connections: BTreeMap::new(),
+                negotiating_connections: BTreeMap::new(),
+                connections: BTreeMap::new(),
+                rx,
+                init_responder: None,
+                rack_secret_responder: None,
+                log: log
+                    .new(o!("component" => "bootstore", "peer_id" => id_str)),
+                conn_rx,
+                conn_tx,
+            },
+            PeerHandle { tx },
+        )
+    }
+
+    /// Run the main loop of the peer
+    ///
+    /// This should be spawned into its own tokio task
+    pub async fn run(&mut self) {
+        // select among timer tick/received messages
+        let mut interval = interval(self.config.time_per_tick);
+        interval.set_missed_tick_behavior(MissedTickBehavior::Delay);
+        let listener = TcpListener::bind(&self.config.addr).await.unwrap();
+        while !self.shutdown {
+            tokio::select! {
+                res = listener.accept() => self.on_accept(res).await,
+                Some(request) = self.rx.recv() => {
+                    self.on_api_request(request).await;
+                }
+                Some(msg) = self.conn_rx.recv() => self.on_conn_msg(msg).await,
+                _ = interval.tick() => {
+                    if let Err(errors) = self.fsm.tick(Instant::now().into()) {
+                        for (_, err) in errors {
+                            self.handle_api_error(err).await;
+                        }
+                    }
+                    // Even with errors there may be messages that need sending
+                    self.deliver_envelopes().await;
+                }
+            }
+        }
+    }
+
+    // Handle an accepted connection
+    async fn on_accept(
+        &mut self,
+        res: Result<(TcpStream, SocketAddr), std::io::Error>,
+    ) {
+        match res {
+            Ok((sock, addr)) => {
+                let SocketAddr::V6(addr) = addr else {
+                    warn!(self.log, "Got connection from IPv4 address {}", addr);
+                    return;
+                };
+                // TODO: Log if a peer with a lower address connects?
+                // Remove any existing connection
+                self.remove_accepted_connection(&addr).await;
+                info!(self.log, "Accepted connection from {}", addr);
+                self.handle_unique_id_counter += 1;
+                let handle = spawn_server(
+                    self.handle_unique_id_counter,
+                    self.config.id.clone(),
+                    self.config.addr.clone(),
+                    addr.clone(),
+                    sock,
+                    self.conn_tx.clone(),
+                    &self.log,
+                )
+                .await;
+                self.accepted_connections.insert(addr, handle);
+            }
+            Err(err) => {}
+        }
+    }
+
+    // Handle API requests from the `PeerHandle`
+    async fn on_api_request(&mut self, request: PeerApiRequest) {
+        match request {
+            PeerApiRequest::InitRack {
+                rack_uuid,
+                initial_membership,
+                responder,
+            } => {
+                info!(self.log,
+                    "Rack init started";
+                    "rack_uuid" => rack_uuid.to_string()
+                );
+                if self.init_responder.is_some() {
+                    let _ = responder
+                        .send(Err(PeerRequestError::RequestAlreadyPending));
+                    return;
+                }
+                if let Err(err) = self.fsm.init_rack(
+                    Instant::now().into(),
+                    rack_uuid,
+                    initial_membership,
+                ) {
+                    let _ = responder.send(Err(err.into()));
+                } else {
+                    self.init_responder = Some(responder);
+                    self.deliver_envelopes().await;
+                }
+            }
+            PeerApiRequest::InitLearner { responder } => {
+                if self.init_responder.is_some() {
+                    let _ = responder
+                        .send(Err(PeerRequestError::RequestAlreadyPending));
+                    return;
+                }
+                if let Err(err) = self.fsm.init_learner(Instant::now().into()) {
+                    let _ = responder.send(Err(err.into()));
+                } else {
+                    self.init_responder = Some(responder);
+                    self.deliver_envelopes().await;
+                }
+            }
+            PeerApiRequest::LoadRackSecret { responder } => {
+                if self.rack_secret_responder.is_some() {
+                    let _ = responder
+                        .send(Err(PeerRequestError::RequestAlreadyPending));
+                    return;
+                }
+                if let Err(err) =
+                    self.fsm.load_rack_secret(Instant::now().into())
+                {
+                    let _ = responder.send(Err(err.into()));
+                } else {
+                    self.rack_secret_responder = Some(responder);
+                    self.deliver_envelopes().await;
+                }
+            }
+            PeerApiRequest::PeerAddresses(peers) => {
+                info!(self.log, "Updated Peer Addresses: {:?}", peers);
+                self.manage_connections(peers).await;
+            }
+            PeerApiRequest::GetStatus { responder } => {
+                let status = Status {
+                    fsm_state: self.fsm.state().name(),
+                    peers: self.peers.clone(),
+                    connections: self
+                        .connections
+                        .iter()
+                        .map(|(id, handle)| (id.clone(), handle.addr))
+                        .collect(),
+                    accepted_connections: self
+                        .accepted_connections
+                        .keys()
+                        .cloned()
+                        .collect(),
+                    negotiating_connections: self
+                        .negotiating_connections
+                        .keys()
+                        .cloned()
+                        .collect(),
+                };
+                let _ = responder.send(status);
+            }
+            PeerApiRequest::Shutdown => {
+                self.shutdown = true;
+                // Shutdown all connection processing tasks
+                for (_, handle) in &self.accepted_connections {
+                    let _ = handle.tx.send(MainToConnMsg::Close).await;
+                }
+                for (_, handle) in &self.negotiating_connections {
+                    let _ = handle.tx.send(MainToConnMsg::Close).await;
+                }
+                for (_, handle) in &self.connections {
+                    let _ = handle.tx.send(MainToConnMsg::Close).await;
+                }
+            }
+        }
+    }
+
+    // Route messages to their destination connections
+    async fn deliver_envelopes(&mut self) {
+        for envelope in self.fsm.drain_envelopes() {
+            if let Some(conn_handle) = self.connections.get(&envelope.to) {
+                debug!(
+                    self.log,
+                    "Sending {:?} to {}", envelope.msg, envelope.to
+                );
+                if let Err(e) = conn_handle
+                    .tx
+                    .send(MainToConnMsg::Msg(Msg::Fsm(envelope.msg)))
+                    .await
+                {
+                    warn!(self.log, "Failed to send {:?}", e);
+                }
+            } else {
+                warn!(self.log, "Missing connection to {}", envelope.to);
+            }
+        }
+    }
+
+    // Perform any operations required by a given `ApiOutput`, such as
+    // persisting state, and then inform any callers (via outstanding responders)
+    // of the result.
+    async fn handle_api_output(&mut self, output: ApiOutput) {
+        info!(self.log, "Fsm output = {:?}", output);
+        match output {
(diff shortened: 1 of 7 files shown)


bootstore/src/schemes/v0/peer.rs
```diff
@@ -0,0 +1,797 @@
...
+pub struct Config {
+    id: Baseboard,
+    addr: SocketAddrV6,
+    time_per_tick: Duration,
+    learn_timeout: Duration,
+    rack_init_timeout: Duration,
+    rack_secret_request_timeout: Duration,
+}
+
+// An error response from a `PeerRequest`
```
mkeeter, line 34: Nit: use docstring comments (`///`) here and above the variants, since some editors use different colors for docstrings vs regular comments.

bootstore/src/schemes/v0/peer.rs
```diff
@@ -0,0 +1,797 @@
...
+
+    /// Inform the peer of currently known IP addresses on the bootstrap network
+    ///
+    /// These are generated from DDM prefixes learned by the bootstrap agent.
+    PeerAddresses(BTreeSet<SocketAddrV6>),
+
+    /// Get the status of this peer
+    GetStatus { responder: oneshot::Sender<Status> },
+
+    /// Shutdown the peer
```
mkeeter, line 81: Specify what's being shut down here – I assume this doesn't mean "send the sled to A2"!

bootstore/src/schemes/v0/peer.rs
```diff
@@ -0,0 +1,797 @@
...
+    InitRack {
+        rack_uuid: RackUuid,
+        initial_membership: BTreeSet<Baseboard>,
+        responder: oneshot::Sender<Result<(), PeerRequestError>>,
+    },
+
+    /// Initialize this peer as a learner.
+    ///
+    /// Return `()` from the responder when the learner has learned its share
+    InitLearner { responder: oneshot::Sender<Result<(), PeerRequestError>> },
```
mkeeter, line 63: Nit: this looks non-standard, did you run `cargo fmt` already?

bootstore/src/schemes/v0/peer.rs
```diff
@@ -0,0 +1,797 @@
...
+    /// Get the status of this peer
+    GetStatus { responder: oneshot::Sender<Status> },
+
+    /// Shutdown the peer
+    Shutdown,
+}
+
+// A handle for interacting with a `Peer` task
+pub struct PeerHandle {
+    tx: mpsc::Sender<PeerApiRequest>,
```
mkeeter, line 87: I was going to complain that we don't actually need `mpsc`, but it looks like Tokio doesn't have a `spsc`, so... as you were.

bootstore/src/schemes/v0/peer.rs
```diff
@@ -0,0 +1,797 @@
...
+    PeerAddresses(BTreeSet<SocketAddrV6>),
+
+    /// Get the status of this peer
+    GetStatus { responder: oneshot::Sender<Status> },
+
+    /// Shutdown the peer
+    Shutdown,
+}
+
+// A handle for interacting with a `Peer` task
```
mkeeter, line 85:
```suggestion
/// A handle for interacting with a `Peer` task
```

bootstore/src/schemes/v0/peer.rs
```diff
@@ -0,0 +1,797 @@
...
+#[derive(Debug, Clone)]
+pub struct Status {
+    fsm_state: &'static str,
+    peers: BTreeSet<SocketAddrV6>,
+    connections: BTreeMap<Baseboard, SocketAddrV6>,
+    accepted_connections: BTreeSet<SocketAddrV6>,
+    negotiating_connections: BTreeSet<SocketAddrV6>,
+}
+
+/// A peer in the bootstore protocol
```
mkeeter, line 166: A comment about how this struct is used would be helpful.

Do we have one `Peer` per known other sled doing bootstore?  Or do we only have one `Peer`, representing the sled that this code is running on?

bootstore/src/schemes/v0/peer.rs
```diff
@@ -0,0 +1,797 @@
...
+    fsm: Fsm,
+    peers: BTreeSet<SocketAddrV6>,
+    handle_unique_id_counter: u64,
+
+    // boolean set when a `PeerApiRequest::shutdown` is received
+    shutdown: bool,
+
+    // Connections that have been accepted, but where negotiation has not
+    // finished At this point, we only know the client port of the connection,
+    // and so cannot identify  it as a `Peer`.
```
mkeeter, line 178:
```suggestion
    // finished. At this point, we only know the client port of the connection,
    // and so cannot identify it as a `Peer`.
```

bootstore/src/schemes/v0/peer.rs
```diff
@@ -0,0 +1,797 @@
...
+
+    // Connections that have not yet completed handshakes via `Hello` and
+    // `Identify` messages.
+    //
+    // We only store the client (connecting) side, not server (accepting) side here.
+    negotiating_connections: BTreeMap<SocketAddrV6, PeerConnHandle>,
+
+    // Active connections participating in scheme v0
+    //
+    // This consists of both client and server connections
```
mkeeter, line 189: The "client and server" terminology confuses me here.  It seems like connections flow through `accepted_connections` -> `negotiating_connections` -> `connections`, but the first two only store client connections and the latter somehow also has server connections?

I'll keep reading...

bootstore/src/schemes/v0/peer.rs
```diff
@@ -0,0 +1,797 @@
...
+
+    // Handle requests received from `PeerHandle`
+    rx: mpsc::Receiver<PeerApiRequest>,
+
+    // Used to respond to `InitRack` or `InitLearner` requests
+    init_responder: Option<oneshot::Sender<Result<(), PeerRequestError>>>,
+
+    // Used to respond to `LoadRackSecret` requests
+    rack_secret_responder:
+        Option<oneshot::Sender<Result<RackSecret, PeerRequestError>>>,
```
mkeeter, line 200: Is it possible for both of these to be active at once?

If not, I'd suggest something like
```rust
enum Responder {
    Init(oneshot::Sender<Result<(), PeerRequestError>>),
    RackSecret(oneshot::Sender<Result<RackSecret, PeerRequestError>>),
}

pub struct Peer {
    // ...elided

    // Used to respond to `InitRack`, `InitLearner`, or `LoadRackSecret` requests
    responder: Option<Responder>,
```

bootstore/src/schemes/v0/peer_networking.rs
```diff
@@ -0,0 +1,635 @@
...
+    unique_id: u64,
+    write_sock: OwnedWriteHalf,
+    read_sock: OwnedReadHalf,
+    main_tx: mpsc::Sender<ConnToMainMsg>,
+    rx: mpsc::Receiver<MainToConnMsg>,
+    log: Logger,
+    read_buf: Vec<u8>,
+    total_read: usize,
+
+    // Used for managing inactivity timeouts for the coonnection
```
mkeeter, line 130:
```suggestion
    // Used for managing inactivity timeouts for the connection
```

bootstore/src/schemes/v0/peer_networking.rs
```diff
@@ -0,0 +1,635 @@
...
+    last_received_msg: Instant,
+
+    // Keep a queue to write serialized messages into We limit the queue
```
mkeeter, line 133:
```suggestion
    // Keep a queue to write serialized messages into. We limit the queue
```

bootstore/src/schemes/v0/peer_networking.rs
```diff
@@ -0,0 +1,635 @@
...
+    ) -> EstablishedConn {
+        EstablishedConn {
+            peer_id,
+            unique_id,
+            write_sock,
+            read_sock,
+            main_tx,
+            rx,
+            log,
+            read_buf: vec![0u8; CONN_BUF_SIZE],
```
mkeeter, line 163: Should this be a array instead of a `Vec`, since it's always of known size?

(or a [`Box<[u8]>`](https://doc.rust-lang.org/std/boxed/struct.Box.html#method.from-6) if you want it to be heap-allocated)

bootstore/src/schemes/v0/peer_networking.rs
```diff
@@ -0,0 +1,635 @@
...
+                Some(msg) = self.rx.recv() => {
+                    self.on_msg_from_main(msg).await
+                }
+                res = self.read_sock.read(&mut self.read_buf[self.total_read..]) => {
+                    self.read(res).await
+                }
+                res = self.write_sock.write_buf(&mut self.current_write),
+                   if self.current_write.has_remaining() => {
+                        self.check_write_result(res).await
+                }
```
mkeeter, line 199: Indentation is a little funky here, maybe
```suggestion
                    }
```

bootstore/src/schemes/v0/peer_networking.rs
```diff
@@ -0,0 +1,635 @@
...
+            {
+                self.current_write =
+                    Cursor::new(self.write_queue.pop_front().unwrap());
+            }
+
+            let res = tokio::select! {
+                _ = interval.tick() => {
+                    self.ping().await
+                }
+                Some(msg) = self.rx.recv() => {
```
mkeeter, line 190: Should we also handle the `None` case, which indicates that the main task closed the queue?

bootstore/src/schemes/v0/peer_networking.rs
```diff
@@ -0,0 +1,635 @@
...
+    }
+
+    async fn check_write_result(
+        &mut self,
+        res: Result<usize, std::io::Error>,
+    ) -> Result<(), ConnErr> {
+        match res {
+            Ok(_) => {
+                if !self.current_write.has_remaining() {
+                    self.current_write = Cursor::new(Vec::new());
```
mkeeter, line 249:
```suggestion
                    self.current_write = Cursor::default();
```

mkeeter: (also, this suggests that `current_write` should be an `Option<Cursor<Vec<u8>>>`, instead of using an empty cursor to indicate absence)

bootstore/src/schemes/v0/peer_networking.rs
```diff
@@ -0,0 +1,635 @@
...
+        loop {
+            if self.total_read < FRAME_HEADER_SIZE {
+                return Ok(());
+            }
+            // Read frame size
+            let size = read_frame_size(
+                self.read_buf[..FRAME_HEADER_SIZE].try_into().unwrap(),
+            );
+            let end = size + FRAME_HEADER_SIZE;
+            if end > self.total_read {
```
mkeeter, line 284:
```suggestion
            // If we haven't read the whole message yet, then return
            if end > self.total_read {
```

bootstore/src/schemes/v0/peer_networking.rs
```diff
@@ -0,0 +1,635 @@
...
+                    warn!(
+                        self.log,
+                        "Closing connection: failed to deserialize: {}", e
+                    );
+                    return self.close().await;
+                }
+            };
+            self.last_received_msg = Instant::now();
+            debug!(self.log, "Received {:?}", msg);
+            if let Msg::Fsm(msg) = msg {
```
mkeeter, line 306: This begs the question of what other things `mgs` could be, and why we don't need to handle them.  I'd suggest an explicit `match`, e.g.
```rust
match msg {
    Msg::Fsm(mgs) => {
        // [same as before]
    }
    Msg::Ping => {
        // Nothing to do here, since Ping is just to keep us alive and
        // we updated self.last_received_msg above.
    }
}
```

bootstore/src/schemes/v0/peer_networking.rs
```diff
@@ -0,0 +1,635 @@
...
+                handle_unique_id: self.unique_id,
+                msg: ConnToMainMsgInner::Disconnected {
+                    peer_id: self.peer_id.clone(),
+                },
+            })
+            .await
+        {
+            warn!(self.log, "Failed to send to main task: {:?}", e);
+        }
+        let _ = self.write_sock.shutdown().await;
```
mkeeter, line 367: This makes me a little nervous, because it seems possible to leave this task in a broken state.  On the other hand, this function can _only_ return an error, so maybe it's fine (as long as no one ignores the error)?

bootstore/src/schemes/v0/peer_networking.rs
```diff
@@ -0,0 +1,635 @@
...
+        }
+    }
+
+    // Close and drop the connection.
+    async fn close(&mut self) -> Result<(), ConnErr> {
+        if let Err(e) = self
+            .main_tx
+            .send(ConnToMainMsg {
+                handle_unique_id: self.unique_id,
+                msg: ConnToMainMsgInner::Disconnected {
```
mkeeter, line 359: Is it weird that if the main task sends us `MainToConnMsg::Close`, we both reply along the `main_tx` channel with `Disconnected` **and** return `Err(ConnErr::Close)`?

bootstore/src/schemes/v0/peer_networking.rs
```diff
@@ -0,0 +1,635 @@
...
+    let log = log.clone();
+
+    let handle = tokio::spawn(async move {
+        loop {
+            let sock = match TcpStream::connect(addr).await {
+                Ok(sock) => sock,
+                Err(err) => {
+                    // TODO: Throttle this?
+                    warn!(log, "Failed to connect"; "addr" => addr.to_string());
+                    // TODO: Sleep
```
mkeeter, line 391: This TODO seems done?

bootstore/src/schemes/v0/peer_networking.rs
```diff
@@ -0,0 +1,635 @@
...
+}
+/// Spawn a task that maintains a client connection to a peer
+pub async fn spawn_client(
+    unique_id: u64,
+    my_peer_id: Baseboard,
+    my_addr: SocketAddrV6,
+    addr: SocketAddrV6,
+    log: &Logger,
+    main_tx: mpsc::Sender<ConnToMainMsg>,
+) -> PeerConnHandle {
```
mkeeter, line 379: Vague feelings:

I'm having a lot of trouble with the naming of `fn spawn_client(...) -> PeerConnHandle` and `fn spawn_server(...) -> AcceptedConnHandle`, which _both_ use an `EstablishedConn` internally in their task.

Maybe `ClientConnHandle` and `ServerConnHandle`?  Even with that name, it's hard to tell if it's a connection **to** a client or a connection in which **we** are taking the client role...

Similarly, do we want to split `EstablishedConn` into client / server flavors, since they handle different subsets of messages?  It seems like that would require different types for the `MainToConnMsg` / `ConnToMainMsg`, e.g. the client flavor would only send `Request` FSM messages and not `Response` (?).

bootstore/src/schemes/v0/peer_networking.rs
```diff
@@ -0,0 +1,635 @@
...
+            }
+        }
+    });
+    PeerConnHandle { handle, tx, addr, unique_id }
+}
+
+/// Spawn a task that handles accepted connections from a peer
+pub async fn spawn_server(
+    unique_id: u64,
+    my_peer_id: Baseboard,
```
mkeeter, line 457: Nit: this naming is weird to me, because if it's my ID, it can't belong to a peer (defined as someone _else_ that we're connecting to).  Maybe just `my_id`?

bootstore/src/schemes/v0/peer_networking.rs
```diff
@@ -0,0 +1,635 @@
...
+        if identify.is_some() && !identify_cursor.has_remaining() {
+            return Ok((read_sock, write_sock, identify.unwrap()));
+        }
+
+        tokio::select! {
+            _ = sleep(timeout) => {
+                return Err(HandshakeError::Timeout);
+            }
+            _ = write_sock.writable(), if identify_cursor.has_remaining() => {
+                if hello_cursor.has_remaining() {
```
mkeeter, line 599:
```suggestion
                if !hello_written {
```

bootstore/src/schemes/v0/peer_networking.rs
```diff
@@ -0,0 +1,635 @@
...
+    let identify = write_framed(&Identify {
+        id: local_peer_id.clone(),
+        addr: local_addr,
+    })?;
+    let mut identify_cursor = Cursor::new(&identify);
+
+    let handshake_start = Instant::now();
+
+    // Read `Hello` and the frame size of `Identify`
+    let initial_read = Hello::serialized_size() + FRAME_HEADER_SIZE;
```
mkeeter, line 576:
```suggestion
    const INITIAL_READ: usize = Hello::serialized_size() + FRAME_HEADER_SIZE;
```

(requires making `serialized_size` a `const fn`, but that's fine)

bootstore/src/schemes/v0/peer_networking.rs
```diff
@@ -0,0 +1,635 @@
...
+    let mut read_buf = [0u8; 128];
+    let (mut read_sock, mut write_sock) = sock.into_split();
+
+    // Serialize and write the handshake messages into `write_buf`
+    let mut hello_cursor = Cursor::new(Hello::default().serialize());
+    let identify = write_framed(&Identify {
+        id: local_peer_id.clone(),
+        addr: local_addr,
+    })?;
+    let mut identify_cursor = Cursor::new(&identify);
```
mkeeter, line 571: Having a separate buffer / cursor for `Hello` and `identity` seems awkward and requires extra tracking; could you serialize them back-to-back in a single buffer instead?

```rust
    // Serialize and write the handshake + identity messages to a local buffer
    let out: Vec<u8> = Hello::default()
        .serialize()
        .into_iter()
        .chain(write_framed(&Identify {
            id: local_peer_id.clone(),
            addr: local_addr,
        })?)
        .collect();
    let mut out_cursor = Cursor::new(&out);
```

this requires a few changes later on:
```diff
@@ -585,9 +588,7 @@ async fn perform_handshake(

         let end = initial_read + identify_len;

-        let hello_written = !hello_cursor.has_remaining();
-
-        if identify.is_some() && !identify_cursor.has_remaining() {
+        if identify.is_some() && !out_cursor.has_remaining() {
             return Ok((read_sock, write_sock, identify.unwrap()));
         }

@@ -595,12 +596,8 @@ async fn perform_handshake(
             _ = sleep(timeout) => {
                 return Err(HandshakeError::Timeout);
             }
-            _ = write_sock.writable(), if identify_cursor.has_remaining() => {
-                if hello_cursor.has_remaining() {
-                    write_sock.write_buf(&mut hello_cursor).await?;
-                } else {
-                    write_sock.write_buf(&mut identify_cursor).await?;
-                }
+            _ = write_sock.writable(), if out_cursor.has_remaining() => {
+                write_sock.write_buf(&mut out_cursor).await?;
             }
             res = read_sock.read(&mut read_buf[total_read..end]), if identify.is_none() => {
                 let n = res?;
```

bootstore/src/schemes/v0/peer_networking.rs
mkeeter: (but see my comment above about combining hello + identity into a single buffer)

bootstore/src/schemes/v0/peer.rs
andrewjstone: Honestly, me too. What I'm really trying to get at is that one side is the connection initiating side, and one side accepts connections. We have to manage them separately at the beginning, but once the handshake is complete, they are treated equivalently. I've written this type of code in multiple projects in the past and always struggled naming these things. 

Maybe `initiator` and `acceptor` ?

bootstore/src/schemes/v0/peer.rs
andrewjstone: Excellent idea.

bootstore/src/schemes/v0/peer_networking.rs
andrewjstone: Good call.

bootstore/src/schemes/v0/peer_networking.rs
andrewjstone: Yes. I actually had an `Option<Cursor<Vec<u8>>>` at first, but while iterating I changed it due to a limitation in one version of the code. I don't recall what that was, but I also don't think its relevant any more. I will change it back.

bootstore/src/schemes/v0/peer_networking.rs
```diff
@@ -0,0 +1,635 @@
...
+                        msg: ConnToMainMsgInner::Received {
+                            from: self.peer_id.clone(),
+                            msg,
+                        },
+                    })
+                    .await
+                {
+                    warn!(
+                        self.log,
+                        "Failed to send received msg to main task: {:?}", e
```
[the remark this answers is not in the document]
andrewjstone: works for me!

bootstore/src/schemes/v0/peer_networking.rs
andrewjstone: I don't think anything else can be done here. This is also somewhat of a "courtesy call" for the other end of the socket. The socket itself is going to be dropped which should cleanup resources.

bootstore/src/schemes/v0/peer_networking.rs
```diff
@@ -0,0 +1,635 @@
...
+                    write_sock.write_buf(&mut identify_cursor).await?;
+                }
+            }
+            res = read_sock.read(&mut read_buf[total_read..end]), if identify.is_none() => {
+                let n = res?;
+                total_read += n;
+                if total_read < initial_read {
+                    continue;
+                }
+                if total_read == initial_read {
```
mkeeter, line 611: This section is a little too sneaky for my taste – it requires noticing that `read_sock.read(&mut read_buf[total_read..end])` will only read up to `end` (so we can do an exact comparison `total_read == initial_read`), **and** `end` changes once we've read the `initial_len` of the data and decoded `identify_len`.

What do you think about splitting this function into separate sections? I mocked it up locally (didn't test it, but it compiles fine), along with other recommendations:
```rust
    // Serialize and write the handshake + identity messages to a local buffer
    let out: Vec<u8> = Hello::default()
        .serialize()
        .into_iter()
        .chain(write_framed(&Identify {
            id: local_peer_id.clone(),
            addr: local_addr,
        })?)
        .collect();
    let mut out_cursor = Cursor::new(&out);

    let handshake_start = Instant::now();
    let get_timeout =
        || KEEPALIVE_TIMEOUT.saturating_sub(Instant::now() - handshake_start);

    // Read `Hello` and the frame size of `Identify`
    const INITIAL_READ: usize = Hello::serialized_size() + FRAME_HEADER_SIZE;

    // Read the initial Hello message and the length of the following Identify
    let mut read_pos = 0;
    let identify_len = loop {
        let timeout = get_timeout();
        tokio::select! {
            _ = sleep(timeout) => {
                return Err(HandshakeError::Timeout);
            }
            _ = write_sock.writable(), if out_cursor.has_remaining() => {
                write_sock.write_buf(&mut out_cursor).await?;
            }
            res = read_sock.read(&mut read_buf[read_pos..INITIAL_READ]) => {
                read_pos += res?;
                if read_pos == INITIAL_READ {
                    let hello =
                        Hello::from_bytes(&read_buf[..Hello::serialized_size()]).unwrap();
                    if hello.scheme != 0 {
                        return Err(HandshakeError::UnsupportedScheme);
                    }
                    if hello.version != 0 {
                        return Err(HandshakeError::UnsupportedVersion);
                    }
                    break read_frame_size(
                        read_buf[Hello::serialized_size()..INITIAL_READ]
                            .try_into()
                            .unwrap(),
                    );
                }
            }
        }
    };

    // Read and decode the Identify message, using the length found above
    let mut read_pos = 0;
    let identify: Identify = loop {
        let timeout = get_timeout();
        tokio::select! {
            _ = sleep(timeout) => {
                return Err(HandshakeError::Timeout);
            }
            _ = write_sock.writable(), if out_cursor.has_remaining() => {
                write_sock.write_buf(&mut out_cursor).await?;
            }
            res = read_sock.read(&mut read_buf[read_pos..identify_len]) => {
                read_pos += res?;
                if read_pos == identify_len {
                    break ciborium::from_reader(&read_buf[..identify_len])?;
                }
            }
        }
    };

    // Flush the remaining output from out_cursor
    while !out_cursor.has_remaining() {
        let timeout = get_timeout();
        tokio::select! {
            _ = sleep(timeout) => {
                return Err(HandshakeError::Timeout);
            }
            _ = write_sock.writable() => {
                write_sock.write_buf(&mut out_cursor).await?;
            }
        }
    }
    Ok((read_sock, write_sock, identify))
```

bootstore/src/schemes/v0/peer_networking.rs
andrewjstone: Somewhat. But that's not the only reason this method is called. The connection can close itself and we must inform the main task so it can clean up.

bootstore/src/schemes/v0/peer_networking.rs
andrewjstone: I think the names suck as discussed above. However, I don't think we should split the `EstablishedConn`. Once established both sides are equivalent. Either can send `Request` or `Response` messages.

bootstore/src/schemes/v0/peer_networking.rs
andrewjstone: I totally agree. I definitely had them in a single buffer at one point, but via iteration ended up with this weird mess. I'll do as you suggest.

bootstore/src/schemes/v0/peer.rs
andrewjstone: Yeah. Maybe it's the result of `use_small_heuristics = "max"` in `rustfmt.toml`

bootstore/src/schemes/v0/peer_networking.rs
andrewjstone: I don't think that's correct. The `if` is a precondition on the tokio expression, not standalone. I actually got confused by this yesterday also when re-reading. It's unfortunate.

bootstore/src/schemes/v0/peer.rs
andrewjstone: done

bootstore/src/schemes/v0/peer.rs
andrewjstone: done

bootstore/src/schemes/v0/peer_networking.rs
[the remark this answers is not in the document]
andrewjstone: Done everywhere

bootstore/src/schemes/v0/peer_networking.rs
mkeeter: Yeah, and reading further, it looks like the return code from `run` is either discarded (for a server task) or used to decide whether to restart (for a client task), so it doesn't make it back to the main task either way.

bootstore/src/schemes/v0/peer.rs
```diff
@@ -0,0 +1,802 @@
...
+                    self.config.addr.clone(),
+                    addr.clone(),
+                    sock,
+                    self.conn_tx.clone(),
+                    &self.log,
+                )
+                .await;
+                self.accepted_connections.insert(addr, handle);
+            }
+            Err(err) => {}
```
mkeeter, line 317: This seems suspicious to me, worth adding a comment explaining why it's okay?

bootstore/src/schemes/v0/peer.rs
```diff
@@ -0,0 +1,802 @@
...
+                }
+                Some(msg) = self.conn_rx.recv() => self.on_conn_msg(msg).await,
+                _ = interval.tick() => {
+                    if let Err(errors) = self.fsm.tick(Instant::now().into()) {
+                        for (_, err) in errors {
+                            self.handle_api_error(err).await;
+                        }
+                    }
+                    // Even with errors there may be messages that need sending
+                    self.deliver_envelopes().await;
```
mkeeter, line 283: Would it make sense to move this into the main body of the loop?  i.e.
```rust
while !self.shutdown {
    tokio::select! {
        // ...
    }
    self.deliver_envelopes().await
}
```
Then, we don't need to remember it in every branch of `on_conn_msg` (and it's presumably safe / cheap to call if no envelopes are queued up)

bootstore/src/schemes/v0/peer.rs
```diff
@@ -0,0 +1,802 @@
...
+                // need to get the `request_id` from destructuring above
+                if let Some(responder) = self.rack_secret_responder.take() {
+                    let _ = responder.send(Ok(secret));
+                }
+            }
+            ApiOutput::ShareDistributedToLearner => {
+                // TODO: Persistence
+            }
+            ApiOutput::LearningCompleted => {
+                if let Some(responder) = self.init_responder.take() {
```
mkeeter, line 464: Same here!

bootstore/src/schemes/v0/peer_networking.rs
```diff
@@ -0,0 +1,635 @@
...
+            write_sock,
+            read_sock,
+            main_tx,
+            rx,
+            log,
+            read_buf: vec![0u8; CONN_BUF_SIZE],
+            total_read: 0,
+            last_received_msg: Instant::now(),
+            write_queue: VecDeque::with_capacity(MSG_WRITE_QUEUE_CAPACITY),
+            current_write: Cursor::new(Vec::new()),
```
sunshowers, line 167: Could probably use `Bytes` (or `Option<Bytes>` as below) here since that implements `bytes::Buf` as well.

bootstore/src/schemes/v0/peer.rs
mkeeter: (and below in `handle_api_error`)

bootstore/src/schemes/v0/peer.rs
andrewjstone: fixed in c6735245f

bootstore/src/schemes/v0/peer_networking.rs
andrewjstone: Fixed the names in c6735245f

bootstore/src/schemes/v0/peer_networking.rs
sunshowers: I think `Option<T>` might be a bit more annoying to use -- `has_remaining` is an easier check.

bootstore/src/schemes/v0/peer.rs
```diff
@@ -0,0 +1,802 @@
...
+                    // This can only be a failure to init the rack, so we
+                    // log it as an error and not a warning. It is unrecoverable
+                    // without a rack reset.
+                    error!(self.log, "Error on connection: {e}");
+                } else {
+                    self.deliver_envelopes().await;
+                }
+            }
+            ConnToMainMsgInner::ConnectedClient { addr, peer_id } => {
+                let handle =
```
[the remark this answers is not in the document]
mkeeter: Also, I'm wondering if a misbehaving client could cause us to panic here – if it tries to connect to the same server repeatedly, we'll have removed it from `self.negotiating_connections` after the first connection...

bootstore/src/schemes/v0/peer_networking.rs
```diff
@@ -0,0 +1,635 @@
...
+    });
+    PeerConnHandle { handle, tx, addr, unique_id }
+}
+
+/// Spawn a task that handles accepted connections from a peer
+pub async fn spawn_server(
+    unique_id: u64,
+    my_peer_id: Baseboard,
+    my_addr: SocketAddrV6,
+    addr: SocketAddrV6,
```
sunshowers, line 459: Maybe `client_addr`?

bootstore/src/schemes/v0/peer.rs
```diff
@@ -0,0 +1,802 @@
...
+                    Ok(None) => self.deliver_envelopes().await,
+                    Ok(Some(api_output)) => {
+                        self.handle_api_output(api_output).await
+                    }
+                    Err(err) => self.handle_api_error(err).await,
+                }
+            }
+            ConnToMainMsgInner::FailedServerHandshake {
+                addr: SocketAddrV6,
+            } => {}
```
mkeeter, line 606: In `peer_networking.rs`, it mentions returning this error so the main task can handle "cleaning up any metadata".  Is there nothing to clean up?

bootstore/src/schemes/v0/peer_networking.rs
```diff
@@ -0,0 +1,635 @@
...
+// Serialize and write `msg` into `buf`, prefixed by a 4-byte big-endian size header
+//
+// Return the total amount of data written into `buf` including the 4-byte header
+fn write_framed<T: Serialize + ?Sized>(
+    msg: &T,
+) -> Result<Vec<u8>, ciborium::ser::Error<std::io::Error>> {
+    let mut cursor = Cursor::new(vec![]);
+    // Write a size placeholder
+    std::io::Write::write(&mut cursor, &[0u8; FRAME_HEADER_SIZE]);
+    cursor.set_position(FRAME_HEADER_SIZE as u64);
```
sunshowers, line 530: Is this required? I think the cursor would be moved forward to `FRAME_HEADER_SIZE` by the `write` call above.

bootstore/src/schemes/v0/peer_networking.rs
```diff
@@ -0,0 +1,635 @@
...
+        (cursor.position() - FRAME_HEADER_SIZE as u64).try_into().unwrap();
+    let mut buf = cursor.into_inner();
+    buf[0..FRAME_HEADER_SIZE].copy_from_slice(&size.to_be_bytes());
+    Ok(buf)
+}
+
+// Decode the 4-byte big-endian frame size header
+fn read_frame_size(buf: [u8; FRAME_HEADER_SIZE]) -> usize {
+    u32::from_be_bytes(buf) as usize
+}
```
sunshowers, line 542: Do we also want a magic string in the beginning indicating that this is a lrtq message? Maybe just `"LRTQ"`.

bootstore/src/schemes/v0/peer_networking.rs
```diff
@@ -0,0 +1,635 @@
...
+        let end = initial_read + identify_len;
+
+        let hello_written = !hello_cursor.has_remaining();
+
+        if identify.is_some() && !identify_cursor.has_remaining() {
+            return Ok((read_sock, write_sock, identify.unwrap()));
+        }
+
+        tokio::select! {
+            _ = sleep(timeout) => {
```
sunshowers, line 595: To confirm, the timeout applies to every loop iteration, not the entire operation. Is this desired? (In many cases it is, just confirming.)

bootstore/src/schemes/v0/peer.rs
```diff
@@ -0,0 +1,810 @@
...
+            }
+        }
+    }
+
+    async fn remove_established_connection(&mut self, peer_id: &Baseboard) {
+        if let Some(handle) = self.established_connections.remove(peer_id) {
+            // Gracefully stop the task
+            let _ = handle.tx.send(MainToConnMsg::Close).await;
+        }
+    }
```
sunshowers, line 676: Worth adding a warning here if a connection isn't found in the `established_connections` map.

bootstore/src/schemes/v0/peer.rs
```diff
@@ -0,0 +1,810 @@
...
+            self.remove_peer(addr).await;
+        }
+    }
+
+    async fn remove_accepted_connection(&mut self, addr: &SocketAddrV6) {
+        if let Some(handle) = self.accepted_connections.remove(&addr) {
+            // The connection has not yet completed its handshake
+            let _ = handle.tx.send(MainToConnMsg::Close).await;
+        }
+    }
```
sunshowers, line 650: Worth adding a warning here if a connection isn't found in the `accepted_connections` map.

bootstore/src/schemes/v0/peer.rs
```diff
@@ -0,0 +1,810 @@
...
+    // handshakes via `Hello` and `Identify` messages.
+    //
+    // We only store the initiating (connecting) side, not accepting side here.
+    initiating_connections: BTreeMap<SocketAddrV6, PeerConnHandle>,
+
+    // Active connections participating in scheme v0
+    //
+    // This consists of both initiating and accepting connections that have
+    // completed handshake and are now logically equivalent.
+    established_connections: BTreeMap<Baseboard, PeerConnHandle>,
```
sunshowers, line 199: Why is `established_connections` keyed on `Baseboard` while the other two are keyed on `SocketAddrV6`? Maybe add a comment about this.

Also can `initiating` and `accepted` connections be combined into a single map?

bootstore/src/schemes/v0/peer.rs
```diff
@@ -0,0 +1,810 @@
...
+                    initial_membership,
+                ) {
+                    let _ = responder.send(Err(err.into()));
+                } else {
+                    self.init_responder = Some(responder);
+                    self.deliver_envelopes().await;
+                }
+            }
+            NodeApiRequest::InitLearner { responder } => {
+                if self.init_responder.is_some() {
```
sunshowers, line 354: Add a log here similar to "Rack init started" above?

bootstore/src/schemes/v0/peer.rs
```diff
@@ -0,0 +1,810 @@
...
+                }
+                if let Err(err) = self.fsm.init_learner(Instant::now().into()) {
+                    let _ = responder.send(Err(err.into()));
+                } else {
+                    self.init_responder = Some(responder);
+                    self.deliver_envelopes().await;
+                }
+            }
+            NodeApiRequest::LoadRackSecret { responder } => {
+                if self.rack_secret_responder.is_some() {
```
sunshowers, line 367: Log here as well.

bootstore/src/schemes/v0/peer.rs
```diff
@@ -0,0 +1,810 @@
...
+                    self.rack_secret_responder = Some(responder);
+                    self.deliver_envelopes().await;
+                }
+            }
+            NodeApiRequest::PeerAddresses(peers) => {
+                info!(self.log, "Updated Peer Addresses: {peers:?}");
+                self.manage_connections(peers).await;
+            }
+            NodeApiRequest::GetStatus { responder } => {
+                let status = Status {
```
sunshowers, line 386: Log here as well.

bootstore/src/schemes/v0/peer.rs
```diff
@@ -0,0 +1,810 @@
...
+                    negotiating_connections: self
+                        .initiating_connections
+                        .keys()
+                        .cloned()
+                        .collect(),
+                };
+                let _ = responder.send(status);
+            }
+            NodeApiRequest::Shutdown => {
+                self.shutdown = true;
```
sunshowers, line 408: Log here as well.

bootstore/src/schemes/v0/peer.rs
```diff
@@ -0,0 +1,810 @@
...
+                // Shutdown all connection processing tasks
+                for (_, handle) in &self.accepted_connections {
+                    let _ = handle.tx.send(MainToConnMsg::Close).await;
+                }
+                for (_, handle) in &self.initiating_connections {
+                    let _ = handle.tx.send(MainToConnMsg::Close).await;
+                }
+                for (_, handle) in &self.established_connections {
+                    let _ = handle.tx.send(MainToConnMsg::Close).await;
+                }
```
sunshowers, line 418: Should this be parallelized via `buffer_unordered` or similar?

bootstore/src/schemes/v0/peer.rs
```diff
@@ -0,0 +1,810 @@
...
+    // of the result.
+    async fn handle_api_output(&mut self, output: ApiOutput) {
+        info!(self.log, "Fsm output = {output:?}");
+        match output {
+            // Initialization is mutually exclusive
+            ApiOutput::PeerInitialized | ApiOutput::RackInitComplete => {
+                if let Some(responder) = self.init_responder.take() {
+                    let _ = responder.send(Ok(()));
+                }
+                // TODO: Persistence
```
sunshowers, line 457: What does "persistence" mean here?

bootstore/src/schemes/v0/peer.rs
```diff
@@ -0,0 +1,810 @@
...
+                // Ignore the stale message if the unique_id doesn't match what
+                // we have stored.
+                if unique_id != accepted_handle.unique_id {
+                    return;
+                }
+
+                // Gracefully close any old tasks for this peer if they exist
+                self.remove_established_connection(&peer_id).await;
+
+                // Move from `accepted_connections` to `connections`
```
sunshowers, line 541:
```suggestion
                // Move from `accepted_connections` to `established_connections`
```

bootstore/src/schemes/v0/peer_networking.rs
andrewjstone: fixed in fd92a4bc5

bootstore/src/schemes/v0/peer_networking.rs
andrewjstone: I don't think we need to do that, because the main task will only exit on an explicit shutdown, which will cancel the connnection task first.

bootstore/src/schemes/v0/peer_networking.rs
andrewjstone: In the interest of urgency, I'm also gonna skip this for now.

bootstore/src/schemes/v0/peer.rs
andrewjstone: Actually, this is not ok. I'm going to log an error for lack of a better thing to do.

bootstore/src/schemes/v0/peer.rs
andrewjstone: Good idea. Fixed in 851cc853f

bootstore/src/schemes/v0/peer.rs
andrewjstone: Fixed in 851cc853f

bootstore/src/schemes/v0/peer_networking.rs
andrewjstone: Done in 851cc853f

bootstore/src/schemes/v0/peer_networking.rs
```diff
@@ -0,0 +1,635 @@
...
+                        Err(e) => {
+                            warn!(
+                                self.log,
+                                "Closing connection: Failed to serialize msg: {}",
+                                e
+                            );
+                            self.close().await
+                        }
+                    }
+                }
```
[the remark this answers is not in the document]
andrewjstone: Done in 851cc853f

bootstore/src/schemes/v0/peer_networking.rs
andrewjstone: This doesn't pass tests. The handshake doesn't complete. While I agree this structure is a touch clearer, I don't feel like debugging right now. I took all your other suggestions for the handshake though.

bootstore/src/schemes/v0/peer_networking.rs
andrewjstone: Fixed in 7a9167133

bootstore/src/schemes/v0/peer_networking.rs
andrewjstone: Done in 7a9167133

bootstore/src/schemes/v0/peer_networking.rs
andrewjstone: Used your other fixes, obviating the need for this change.

bootstore/src/schemes/v0/peer.rs
```diff
@@ -0,0 +1,802 @@
...
+            ApiOutput::PeerInitialized | ApiOutput::RackInitComplete => {
+                if let Some(responder) = self.init_responder.take() {
+                    let _ = responder.send(Ok(()));
+                }
+                // TODO: Persistence
+            }
+            ApiOutput::RackSecret { secret, .. } => {
+                // We only allow one outstanding request currently, so no
+                // need to get the `request_id` from destructuring above
+                if let Some(responder) = self.rack_secret_responder.take() {
```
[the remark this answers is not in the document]
andrewjstone: Good call! The original FSM code actually returned `ApiOutput::RackSecret` on learn requests completing, but that doesn't happen anymore, as there was no pending api call asking for the rack secret. The rack secret was just loaded as a side effect so that the `LearnedSharePkg` to distribute could be be created with a decrypted share. The decryption key is derived from the rack secret.

bootstore/src/schemes/v0/peer.rs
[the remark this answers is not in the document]
andrewjstone: Great find! This is definitely a bug. On a connection failure, the connection will be retried inside the loop in `spawn_connection_initiator_task`. But we won't actually have any handle in `self.initiating_connections`, since we only load that once and remove it the first time this message is handled.

I believe we would have found this once I add state persistence and restart tests, as then a reconnection would occur and we would have panicked. But I'm thankful your eagle eye saved me that debugging!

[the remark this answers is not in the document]
andrewjstone: Also, the reason the error is in the `ConnectedServer` case is that there is no loop in that task, as the server side task gets recreated when the remote peer connects to us, versus us connecting to a remote peer in a loop.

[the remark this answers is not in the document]
andrewjstone: Damnit, this reveals a larger issue for status queries. We never go back to the negotiatitng/initiating_connection state, so it will always report we are established. I'm going to have to take this loop out I think.

bootstore/src/schemes/v0/peer_networking.rs
mkeeter: ah, the issue is probably `while !out_cursor.has_remaining()` (should be `while out_cursor.has_remaining()`); up to you if you want to re-test.

bootstore/src/schemes/v0/peer.rs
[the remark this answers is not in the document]
andrewjstone: > Damnit, this reveals a larger issue for status queries. We never go back to the negotiatitng/initiating_connection state, so it will always report we are established. I'm going to have to take this loop out I think.

Actually, no I already can handle this with the `Disconnect` message. I just have to move the handle back into  `self.initiating_connections`

bootstore/src/schemes/v0/peer.rs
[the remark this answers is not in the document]
andrewjstone: Done in f354ee9c5

bootstore/src/schemes/v0/peer.rs
andrewjstone: Done in f354ee9c5

bootstore/src/schemes/v0/peer.rs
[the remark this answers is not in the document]
andrewjstone: Removed an incorrect comment related to the fix: 4291e47ca

bootstore/src/schemes/v0/peer.rs
andrewjstone: Good catch! fixed in dcd5dac0d

bootstore/src/schemes/v0/peer_networking.rs
andrewjstone: I thought so as well, but that definitely was not the case.

bootstore/src/schemes/v0/peer_networking.rs
andrewjstone: No, this is unnecessary.

bootstore/src/schemes/v0/peer_networking.rs
```diff
@@ -0,0 +1,635 @@
...
+                Some(msg) = self.rx.recv() => {
+                    self.on_msg_from_main(msg).await
+                }
+                res = self.read_sock.read(&mut self.read_buf[self.total_read..]) => {
+                    self.read(res).await
+                }
+                res = self.write_sock.write_buf(&mut self.current_write),
+                   if self.current_write.has_remaining() => {
+                        self.check_write_result(res).await
+                }
```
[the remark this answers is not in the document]
andrewjstone: I like this better. Thanks! 

Done in e1856b2c0

bootstore/src/schemes/v0/peer_networking.rs
andrewjstone: Done in e1856b2c0

bootstore/src/schemes/v0/peer_networking.rs
andrewjstone: Agreed, but gonna skip this in the interest of time

bootstore/src/schemes/v0/peer_networking.rs
andrewjstone: Yeah, its per iteration. I adjust it at the top of the loop to account for time moving on.

bootstore/src/schemes/v0/peer.rs
andrewjstone: In the common case the connections shouldn't exist, as it means a connecting client connected twice. The only way they would exist is if there is a race, and the remote client detects shutdown first. In that case, I think no warning is preferred.

bootstore/src/schemes/v0/peer.rs
andrewjstone: Similar rationale to `established_connections` above

bootstore/src/schemes/v0/peer.rs
andrewjstone: I'm not sure it's useful to log every time a status request is made. Seems like that could make logs more noisy than necessary. This is especially true because status requests are answered immediately unlike the other requests.

bootstore/src/schemes/v0/peer.rs
andrewjstone: That seems like overkill for local channel sends. I"m not sure it will actually speed things up either, as I don't expect any blocking here. 

A larger question I've thought about is just going ahead and making all these internal channels unbounded, but that makes me worry about memory blowup due to a bug.

bootstore/src/schemes/v0/peer.rs
andrewjstone: It means saving the state of the Fsm to disk so the peer can be restarted. It's coming in a follow up PR.

bootstore/src/schemes/v0/peer.rs
andrewjstone: Done in 94cf22c86

bootstore/src/schemes/v0/peer.rs
andrewjstone: Done in 94cf22c86

bootstore/src/schemes/v0/peer.rs
andrewjstone: Done in 94cf22c86

bootstore/src/schemes/v0/peer.rs
andrewjstone: Done in 94cf22c86

bootstore/src/schemes/v0/peer.rs
andrewjstone: `initiating` and `accepted` track addresses with different ports so they can't be merged. The port for `accepted` is the ephemeral port, whereas for `initiating` it is the server port. This is also why the types are different for the values. We learn the server port of the accepted side during handshake.

andrewjstone:
@mkeeter @sunshowers I think this is good to go now!

sunshowers (approved):

PR #3567, opened 2023-07-12
[sled-agent] Make the internal DNS servers redundant

smklein:
Provision three copies from RSS, rather than merely one.

Additionally, update `DNS_REDUNDANCY` and `MAX_DNS_REDUNDANCY` usage to be more accurate.

- `DNS_REDUNDANCY` describes how many internal DNS servers are provisioned by RSS
- `MAX_DNS_REDUNDANCY` describes the maximum number of internal DNS servers a client should attempt to query

This change required some changes to "how global zone IP addresses are allocated for internal DNS servers", namely, finding a way to each address object a unique name. Rather than hard-code something specific to the internal DNS service but require it for all services (with the existing `gz_addresses` parameter), I opted to move this logic into the `ServiceType::InternalDNS` parameters.

Fixes https://github.com/oxidecomputer/omicron/issues/3570


sled-agent/src/rack_setup/plan/service.rs
```diff
@@ -295,13 +295,18 @@ impl Plan {
 
         // Provision internal DNS zones, striping across Sleds.
         let reserved_rack_subnet = ReservedRackSubnet::new(config.az_subnet());
-        let dns_subnets = reserved_rack_subnet.get_dns_subnets();
+        assert!(
+            DNS_REDUNDANCY <= MAX_DNS_REDUNDANCY,
```
jgallagher, line 299: Tiny nit - since these are both constants, we could use [static_assertions::const_assert](https://docs.rs/static_assertions/latest/static_assertions/macro.const_assert.html) to check this at compile time instead of runtime.

jgallagher (approved):

Issue #3212, opened 2023-05-24
nexus saga unwrap failure


citrus-it:
I tried to create an image from the CLI with:

```
oxide instance create \
        --project ni \
        --description 'Ptang Zoo Boing!' \
        --hostname ekke2 \
        --memory 64G \
        --ncpus 64 \
        --name ekke2
```

which reported a timeout fairly quickly (around 5 seconds):

```
error
Communication Error: error sending request for url (http://venus.oxide-preview.c
om/v1/instances?project=ni): operation timed out
```

and nexus crashed.

I'll put the relevant log on shared storage somewhere, but here are the final few events that I extracted:

```
2023-05-24 10:33:10.006Z DEBG 496eabae-738d-4497-a183-1b75bf912c7c/29152 (ServerContext) on oxz_nexus: received new runtime state from sled agent
    instance_id = 95f8f4f6-b70c-4207-b1a4-8467ebde3213
    runtime_state = InstanceRuntimeState { run_state: Failed, sled_id: aa7c82d9-6e59-406e-b1e3-3648890b4bec, propolis_id: e0091a98-c23d-4dc0-bcf5-5d60cf4dca9f, dst_propolis_id: None, propolis_addr: Some([fd00:1122:3344:103::28]:12400), migration_id: None, propolis_gen: Generation(1), ncpus: InstanceCpuCount(64), memory: ByteCount(68719476736), hostname: "ekkke", gen: Generation(3), time_updated: 2023-05-24T10:33:09.987050789Z }
2023-05-24 10:33:10.006Z DEBG 496eabae-738d-4497-a183-1b75bf912c7c/29152 (dropshot_internal) on oxz_nexus: roles
    actor_id = 001de000-05e4-4000-8000-000000000002
    authenticated = true
    local_addr = [fd00:1122:3344:106::4]:12221
    method = PUT
    remote_addr = [fd00:1122:3344:103::1]:52079
    req_id = 13334114-d055-47ab-a328-b8d8662908c0
    roles = RoleSet { roles: {} }
    uri = /instances/95f8f4f6-b70c-4207-b1a4-8467ebde3213

2023-05-24 10:33:10.049Z INFO 496eabae-738d-4497-a183-1b75bf912c7c/29152 (ServerContext) on oxz_nexus: instance updated by sled agent
    instance_id = 95f8f4f6-b70c-4207-b1a4-8467ebde3213
    new_state = failed
[...]

PR #3574, opened 2023-07-12
Add API cap to instance vCPUs and memory

zephraph:
Mitigates #3212

This a first stab at adding validation to the API to cap the number of vCPUs to 32 until the bhyve limit can be upped. We also added a 64 GiB cap on instance memory. I'm still working out exactly where the validation checks need to happen. #3212 mentions that it should happen in sled agent but for now I've located it with the other instance creation validation logic.

### DIFF
--- a/nexus/src/app/instance.rs
+++ b/nexus/src/app/instance.rs
@@ -131,7 +131,7 @@ impl super::Nexus {
         if params.ncpus.0 > MAX_VCPU_PER_INSTANCE {
             return Err(Error::invalid_request(&format!(
                 "cannot have more than {} vCPUs per instance",
-                MAX_DISKS_PER_INSTANCE
+                MAX_VCPU_PER_INSTANCE
             )));
         }
         if params.external_ips.len() > MAX_EXTERNAL_IPS_PER_INSTANCE {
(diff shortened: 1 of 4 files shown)


nexus/types/src/external_api/params.rs
```diff
@@ -742,8 +742,6 @@ pub struct IpPoolUpdate {
 
 // INSTANCES
 
-pub const MIN_MEMORY_SIZE_BYTES: u32 = 1 << 30; // 1 GiB
-
```
david-crespo, line 746: This was a weird spot for this, so we moved it.

david-crespo (approved):

nexus/src/app/instance.rs
```diff
@@ -125,6 +128,12 @@ impl super::Nexus {
                     .await?;
             }
         }
+        if params.ncpus.0 > MAX_VCPU_PER_INSTANCE {
+            return Err(Error::invalid_request(&format!(
+                "cannot have more than {} vCPUs per instance",
+                MAX_DISKS_PER_INSTANCE
```
zephraph, line 134: lmao what. woops, let me fix this.

PR #3576, opened 2023-07-12
Zero touch switch config for dev environments

rcgoodfellow:
The primary changes here are for supporting tfportd and other bits of automation for virtual environments that make multi-node setups more automated. You can now launch an omicron development environment without manual switch configuration.

There are also some general robustness fixes in omicron itself that I came across and attempted to fix.

### DIFF
--- a/illumos-utils/src/running_zone.rs
+++ b/illumos-utils/src/running_zone.rs
@@ -1104,6 +1104,7 @@ impl InstalledZone {
         unique_name: Option<Uuid>,
         datasets: &[zone::Dataset],
         filesystems: &[zone::Fs],
+        data_links: &[String],
         devices: &[zone::Device],
         opte_ports: Vec<(Port, PortTicket)>,
         bootstrap_vnic: Option<Link>,
@@ -1140,13 +1141,20 @@ impl InstalledZone {
                     .collect(),
             })?;
 
-        let net_device_names: Vec<String> = opte_ports
+        let mut net_device_names: Vec<String> = opte_ports
             .iter()
             .map(|(port, _)| port.vnic_name().to_string())
             .chain(std::iter::once(control_vnic.name().to_string()))
             .chain(bootstrap_vnic.as_ref().map(|vnic| vnic.name().to_string()))
             .chain(links.iter().map(|nic| nic.name().to_string()))
+            .chain(data_links.iter().map(|x| x.to_string()))
             .collect();
+
+        // There are many sources for device names. In some cases they can
+        // overlap, depending on the contents of user defined config files. This
+        // can cause zones to fail to start if duplicate data links are given.
+        net_device_names.sort();
+        net_device_names.dedup();
 
         Zones::install_omicron_zone(
             log,
--- a/installinator/src/dispatch.rs
+++ b/installinator/src/dispatch.rs
@@ -151,14 +151,16 @@ struct InstallOpts {
     #[clap(long)]
     install_on_gimlet: bool,
 
-    //TODO(ry) how to get this parameter set?
+    //TODO(ry) this probably needs to get plumbed somewhere instead of relying
+    //on a default.
     /// The first gimlet data link to use.
-    #[clap(long)]
+    #[clap(long, default_value = "cxgbe0")]
     data_link0: String,
 
-    //TODO(ry) how to get this parameter set?
+    //TODO(ry) this probably needs to get plumbed somewhere instead of relying
+    //on a default.
     /// The second gimlet data link to use.
-    #[clap(long)]
+    #[clap(long, default_value = "cxgbe1")]
     data_link1: String,
 
     // TODO: checksum?
--- a/installinator/src/dispatch.rs
+++ b/installinator/src/dispatch.rs
@@ -151,6 +151,16 @@ struct InstallOpts {
     #[clap(long)]
     install_on_gimlet: bool,
 
+    //TODO(ry) how to get this parameter set?
+    /// The first gimlet data link to use.
+    #[clap(long)]
+    data_link0: String,
+
+    //TODO(ry) how to get this parameter set?
+    /// The second gimlet data link to use.
+    #[clap(long)]
+    data_link1: String,
+
     // TODO: checksum?
 
     // The destination to write to.
@@ -164,7 +174,8 @@ struct InstallOpts {
 impl InstallOpts {
     async fn exec(self, log: &slog::Logger) -> Result<()> {
         if self.bootstrap_sled {
-            crate::bootstrap::bootstrap_sled(log.clone()).await?;
+            let data_links = [self.data_link0.clone(), self.data_link1.clone()];
+            crate::bootstrap::bootstrap_sled(&data_links, log.clone()).await?;
         }
 
         let image_id = self.artifact_ids.resolve()?;
--- a/sled-agent/src/bin/sled-agent.rs
+++ b/sled-agent/src/bin/sled-agent.rs
@@ -122,9 +122,9 @@ async fn do_run() -> Result<(), CmdError> {
                     // If the rack has already been initialized, we shouldn't
                     // abandon the server.
                     Ok(_)
-                        | Err(
-                            bootstrap_agent::RssAccessError::AlreadyInitialized,
-                        ) => {}
+                    | Err(
+                        bootstrap_agent::RssAccessError::AlreadyInitialized,
+                    ) => {}
                     Err(e) => {
                         return Err(CmdError::Failure(e.to_string()));
                     }
--- a/sled-agent/src/bin/sled-agent.rs
+++ b/sled-agent/src/bin/sled-agent.rs
@@ -105,6 +105,8 @@ async fn do_run() -> Result<(), CmdError> {
                 updates: config.updates.clone(),
             };
 
+            let rss_initiator = config.rss_initiator;
+
             // TODO: It's a little silly to pass the config this way - namely,
             // that we construct the bootstrap config from `config`, but then
             // pass it separately just so the sled agent can ingest it later on.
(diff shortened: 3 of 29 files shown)


rcgoodfellow:
This PR also has a bunch of bespoke config files for my multi-switch/multi-node environment that can be ignored.

sled-agent/src/bin/sled-agent.rs
```diff
@@ -117,16 +119,18 @@ async fn do_run() -> Result<(), CmdError> {
...
-                match server.agent().start_rack_initialize(rss_config) {
-                    // If the rack has already been initialized, we shouldn't
-                    // abandon the server.
-                    Ok(_)
-                    | Err(
-                        bootstrap_agent::RssAccessError::AlreadyInitialized,
-                    ) => {}
-                    Err(e) => {
-                        return Err(CmdError::Failure(e.to_string()));
+            if rss_initiator {
```
rcgoodfellow, line 122: Explicitly calling this change out. For multi-node configs, we now specify in config who the RSS initiator is.

sled-hardware/src/underlay.rs
```diff
@@ -79,7 +74,9 @@ pub fn find_nics() -> Result<Vec<AddrObject>, Error> {
 /// For a real Gimlet, this should return the devices like `cxgbeN`. For a
 /// developer machine, or generally a non-Gimlet, this will return the
 /// VNICs we use to emulate those Chelsio links.
-pub fn find_chelsio_links() -> Result<Vec<PhysicalLink>, Error> {
+pub fn find_chelsio_links(
```
rcgoodfellow, line 77: Moving this from a heuristic over a statically defined set of names to explicit configuration.

In the multi-switch development and testing environments we're building, the data links representing the chelsios may be `ixgbe`, an actual `cxgbe`, or a simnet link.

sled-agent/src/services.rs
```diff
@@ -1904,10 +1928,10 @@ impl ServiceManager {
                     if is_gimlet {
                         // Maghemite for a scrimlet needs to be configured to
                         // talk to dendrite
-                        smfh.setprop("config/dendrite", "true")?;
                         smfh.setprop("config/dpd_host", "[::1]")?;
                         smfh.setprop("config/dpd_port", DENDRITE_PORT)?;
                     }
+                    smfh.setprop("config/dendrite", "true")?;
```
jmpesp, line 1934: I wonder: there isn't a `is_softnpu` that could apply here, right?

sled-agent/src/bin/sled-agent.rs
jmpesp: What about the previous check for if the relevant `config-rss.toml` path? For multi-node configs, this is equivalent to making sure only one node has that file.

sled-hardware/src/underlay.rs
jmpesp: :100:

sled-agent/src/services.rs
rcgoodfellow: My understanding of this `ServiceType::Maghemite { mode } = { ... }` case is that we are exclusively talking about the `mg-ddm` service in the switch zone, and that the maghemite service in the GZ is managed differently somewhere else. With that in mind, I think we always want to be in dendrite mode here. If that's not the case (now I'm paging back into memory that the Canada region does use the kernel routing tables in transit mode), we should have some development-specific configuration for that case, as `is_gimlet` no longer accurately represents that intent.

installinator/src/dispatch.rs
```diff
@@ -150,6 +150,18 @@ struct InstallOpts {
     #[clap(long)]
     install_on_gimlet: bool,
 
+    //TODO(ry) this probably needs to get plumbed somewhere instead of relying
+    //on a default.
+    /// The first gimlet data link to use.
+    #[clap(long, default_value = "cxgbe0")]
```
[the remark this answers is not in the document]
jgallagher: Plumbing things through to installinator before it has network access is pretty tricky - the only path we have right now is via the SP holding a blob of data for us, which is how we send it the sha256 hash of the artifacts it should download (a serialized [InstallinatorImageId](https://github.com/oxidecomputer/omicron/blob/60b6cc205ab3459a825e0a9d2ef69309326a3485/ipcc-key-value/src/lib.rs#L46-L64)). Are we sure defaults aren't okay here? Is it something installinator could infer from inspecting hardware instead of something it needs to be told?

sled-agent/src/bin/sled-agent.rs
rcgoodfellow: I think that could probably work. I'll test it out on my next lap.

installinator/src/dispatch.rs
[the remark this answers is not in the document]
rcgoodfellow: I think the defaults are ok for the product, real gimlets, and development environments that happen to have Chelsio NICs that get numbered `cxgbe0` and `cxgbe1`. If we want to run installinator on lab setups with commodity hardware that have something else like `igb0` and `igb1`, then we'd need to plumb something different in somehow - and I'm not sure how that would work.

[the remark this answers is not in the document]
jgallagher: I'm not sure how installinator in a non-gimlet environment would work at all in the context of a trampoline image, since it fetches the `InstallinatorImageId` over IPCC which only exists on gimlets. We can (and have, for testing) run `installinator` as a standalone binary with a variety of command line args; if these defaults are ok for real gimlets I think have the command line args to override for tests puts this on equal footing with all the other non-gimlet installinator options.

sled-agent/src/bin/sled-agent.rs
rcgoodfellow: I verified this is not needed and have removed this config option. Thanks!

illumos-utils/src/running_zone.rs
```diff
@@ -1041,14 +1042,21 @@ impl InstalledZone {
...
             .chain(bootstrap_vnic.as_ref().map(|vnic| vnic.name().to_string()))
             .chain(links.iter().map(|nic| nic.name().to_string()))
+            .chain(data_links.iter().map(|x| x.to_string()))
             .collect();
 
+        // There are many sources for device names. In some cases they can
+        // overlap, depending on the contents of user defined config files. This
+        // can cause zones to fail to start if duplicate data links are given.
+        net_device_names.sort();
+        net_device_names.dedup();
```
internet-diglett, line 1058: Huh. Didn't know that `Vec` had a `dedup()` method. Nifty!

illumos-utils/src/running_zone.rs
```diff
@@ -1041,14 +1042,21 @@ impl InstalledZone {
...
             })?;
 
-        let net_device_names: Vec<String> = opte_ports
+        let mut net_device_names: Vec<String> = opte_ports
             .iter()
             .map(|(port, _)| port.vnic_name().to_string())
             .chain(std::iter::once(control_vnic.name().to_string()))
             .chain(bootstrap_vnic.as_ref().map(|vnic| vnic.name().to_string()))
             .chain(links.iter().map(|nic| nic.name().to_string()))
+            .chain(data_links.iter().map(|x| x.to_string()))
```
internet-diglett, line 1051: `.chain()` is a a cool pattern I need to add to my repertoire.

sled-hardware/src/underlay.rs
```diff
@@ -24,13 +24,6 @@ pub const BOOTSTRAP_PREFIX: u16 = 0xfdb0;
 /// IPv6 prefix mask for bootstrap addresses.
 pub const BOOTSTRAP_MASK: u8 = 64;
 
-// Names of VNICs used as underlay devices for the xde driver.
-//
-// NOTE: These are only used on non-Gimlet systems. In that case, they are
-// expected to have been created by a run of
-// `./tools/create_virtual_hardware.sh`.
-const XDE_VNIC_NAMES: [&str; 2] = ["net0", "net1"];
```
internet-diglett, line 32: ❤️

tools/install_runner_prerequisites.sh
```diff
@@ -104,6 +104,7 @@ function install_packages {
       'library/postgresql-13'
       'pkg-config'
       'brand/omicron1/tools'
+      'brand/sparse'
```
internet-diglett, line 107: The zone created by the outgoing `create_virtual_hardware.sh` uses an `omicron1` zone (set in the `softnpu-zone.txt`). Do we _need_ to use a `sparse` zone with the new softnpu tooling?

internet-diglett: My memory fails me, but I remember troubleshooting some issue with @jclulow that caused me to switch it from `sparse` to `omicron1`, I think at the time it was because the `sparse` zone relied on `pkg` which [wasn't available in the ramdisk image in CI](https://buildomat.eng.oxide.computer/wg/0/details/01GNTHDT85XRMV9AT5XXYMB3RV/GVMtns2wc4vVF7yOT2BPbH4CrZYoEWYw6K0ktr4PNeskUUQJ/01GNTHE62NHH8DPW2R2MTHVE2V#S332). It seems this may have changed though, because your `deploy` task was successful? 🤔

internet-diglett (approved):
LGTM! Looking forward to kicking the tires on this locally! 😄

tools/install_runner_prerequisites.sh
jclulow: FWIW, the issues that lead me to suggest **omicron1** instead of **sparse** haven't changed, so if it is somehow working in the deploy task ramdisk environment, and it depends on the sparse brand, I don't know how! :D

rcgoodfellow: Yeah this is an artifact from previous attempts, will remove.

PR #3579, opened 2023-07-12
[wicketd] make installinator_progress cancel safe

sunshowers:
This code currently uses a Tokio mutex and is currently not cancel-safe
(`RunningUpdate` can be invalid across an await point). Make it
cancel-safe by using an unbounded receiver.

This means that we can replace Tokio's mutex with `std::sync::Mutex`, and make these methods synchronous.

### DIFF
--- a/wicketd/src/installinator_progress.rs
+++ b/wicketd/src/installinator_progress.rs
@@ -110,7 +110,13 @@ impl IprArtifactServer {
         // out-of-order. (If we were OK with messages being sent out-of-order,
         // this could be a std Mutex rather than a Tokio mutex.)
         if let Some(sender) = sender {
-            _ = sender.send(report).await;
+            // Spawn a task to finish sending the report -- let that run in the
+            // background in case this future gets cancelled.
+            tokio::spawn(async move {
+                _ = sender.send(report).await;
+            })
+            .await
+            .expect("sender.send didn't panic");
         }
 
         status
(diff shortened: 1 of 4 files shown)


wicketd/src/installinator_progress.rs
```diff
@@ -69,34 +78,42 @@ impl IprArtifactServer {
...
             slog::debug!(self.log, "update ID unrecognized"; "update_id" => %update_id);
-            EventReportStatus::UnrecognizedUpdateId
+            (EventReportStatus::UnrecognizedUpdateId, None)
+        };
+
+        // Don't drop the mutex here because we don't want messages to be sent
+        // out-of-order. (If we were OK with messages being sent out-of-order,
+        // this could be a std Mutex rather than a Tokio mutex.)
+        if let Some(sender) = sender {
+            _ = sender.send(report).await;
```
jgallagher, line 113: If we are cancelled here, we have lost `report` but have already updated our `*update` to the next state (possibly `Self::Closed`, if this was a terminal report). Is that okay?

sunshowers: Good catch, fixed by spawning a task to send the report.

sunshowers: Changed this so that it uses an unbounded receiver: that simplifies things substantially.

jgallagher (approved):

PR #3585, opened 2023-07-12
Move Ledger from sled-agent to omicron-common

andrewjstone:


common/src/ledger.rs
```diff
@@ -203,7 +203,29 @@ pub trait Ledgerable: DeserializeOwned + Serialize + Send + Sync {
...
-    use omicron_test_utils::dev::test_setup_log;
+
+    pub use dropshot::test_util::LogContext;
+    use dropshot::ConfigLogging;
+    use dropshot::ConfigLoggingIfExists;
+    use dropshot::ConfigLoggingLevel;
+
+    // Copied from `omicron-test-utils` to avoid a circular dependency where
+    // `omicron-common` depends on `omicron-test-utils` which depends on
+    // `omicron-common`.
```
sunshowers, line 214: Oh omicron-test-utils can't re-export this function because `#[cfg(test)]` doesn't propagate to dependencies, right

sunshowers (approved):

common/src/ledger.rs
andrewjstone: I'm more worried about the fact that `omicron-test-utils` depends on `omicron-common`. `omicron-test-utils` does export this, as it's used all over omicron, unless I"m confused about something.

Issue #2478, opened 2023-03-03
automatic debug data collection without running the system out of space


davepacheco:
This ticket covers minimum target work required to (1) make sure we have basic debug data collected on all systems, while (2) not putting system availability at risk (by filling up important ZFS datasets or the pool itself).  This came out of a recorded discussion on 2023-03-03.

- [x] Make sure that on each sled we create a "debug" ZFS dataset, probably on one U.2 device, that will store crash dumps, core files, log files, and potentially other regularly-collected data.  Determine (as a matter of policy) how large we're willing to let it get and assign that as its quota.  (I assume this will be Sled Agent that does this but I'm not sure.)  (Optional: create separate child datasets for core files, dumps, logs, etc. so that we can control their quotas separately [so that a flurry of crash dumps doesn't starve log files or vice versa].)
- [x] Set up a dump device (Sled Agent?)
- [x] Configure dumpadm to save crash dumps in that dataset.  (Sled Agent?)
- [x] Configure coreadm to save core dumps in that dataset.  (Sled Agent?)
- [x] Configure cron + logadm in all control plane zones to rotate all log files in the zone into some known location.  (Probably part of the image build.)
- [x] Configure cron + logadm in the global zone to rotate all log files into the "debug" dataset. (Sled Agent? Host OS image?)
- [x] Update Sled Agent to manage the storage in the "debug" dataset according to some policy, which will presumably start very simply (e.g., delete the oldest files in the dataset until free space in the dataset reaches a watermark like 20%)

There are some known limitations of this: most notably that files are not replicated across multiple devices. Failure of the wrong U.2 (or removal from the system) means we lose the debug data from that system.  This can be mitigated in future work (e.g., by copying them to a second dataset on another pool, or copying them into some intra-rack storage system, etc.).

It's conceivable that we ship an MVP without this, but it's a pretty big risk.  Either we don't manage the storage (in which case we risk running out of disk space and disrupting service) or we turn off all these sources of data (in which case we'll have a pretty hard time fixing anything).

jclulow:
Note that in addition to configuring the dump directory it will probably be necessary to run `savecore` with appropriate arguments, to check the dump device for an existing dump and save it out.  There is also the question of which M.2 we use for dumping (is it the BSU from which we booted?) and then how and whether to check the _other_ BSU for a prior dump we have not yet extracted.

wesolows:
[...]

Issue #2450, opened 2023-02-28
[sled agent] every gimlet should have a dump device


morlandi7:
Placeholder Per the Host OS meeting 2/28/2023, sled agent should handle dumps

smklein:
See: https://github.com/oxidecomputer/omicron/issues/2478 for much more detail - it's fine for this issue to track the dump device management subcomponent

PR #3586, opened 2023-07-13
sled-agent: Setup dump dev on M.2, savecore to U.2

lifning:
Each time the sled-agent upserts a new disk, it enumerates every disk it knows about. If an M.2 is found, it runs dumpadm(8) to set it as the dump device. If a U.2 is *also* found, it has dumpadm(8) invoke savecore(8) to save the previous dump on the M.2, if any, to an encrypted zvol on the U.2, and mark the dump slice as empty.

This is a bit of a kludge due to savecore(8) not yet exposing a clear way to save-and-clear a dump *partition* other than the one configured system-wide by dumpadm. While redundant, this is for our purposes idempotent - as long as *any* M.2 is configured, dumps have a place to go, and as long as any U.2 is present, we will attempt to save any yet-unsaved kernel core dumps on all known M.2s to it.

(see RFD 118 for details of partition layout) (#2450, #2478)

### DIFF
--- a/illumos-utils/src/dumpadm.rs
+++ b/illumos-utils/src/dumpadm.rs
@@ -197,7 +197,11 @@ fn savecore() -> Result<Option<OsString>, DumpAdmError> {
     cmd.arg("-v");
     let out = cmd.output().map_err(DumpAdmError::ExecSavecore)?;
     if out.status.success() {
-        Ok(Some(OsString::from_vec(out.stdout)))
+        if out.stdout.is_empty() {
+            Ok(None)
+        } else {
+            Ok(Some(OsString::from_vec(out.stdout)))
+        }
     } else {
         Err(DumpAdmError::SavecoreFailure(OsString::from_vec(out.stderr)))
     }
--- a/illumos-utils/src/dumpadm.rs
+++ b/illumos-utils/src/dumpadm.rs
@@ -30,6 +30,12 @@ pub enum DumpHdrError {
     InvalidVersion(u32),
 }
 
+/// Returns Ok(true) if the given block device contains a dump that needs to
+/// be savecore'd, Ok(false) if the given block device contains a dump that's
+/// already been savecore'd, Err(DumpHdrError::InvalidMagic) if there's never
+/// been a core written there at all, Err(DumpHdrError::InvalidVersion) if the
+/// dumphdr isn't the one we know how to handle (10), or other variants of
+/// DumpHdrError if there are I/O failures while reading the block device.
 pub fn dump_flag_is_valid(
     dump_slice: &Utf8PathBuf,
 ) -> Result<bool, DumpHdrError> {
@@ -101,6 +107,12 @@ pub enum DumpAdmError {
     ExecSavecore(std::io::Error),
 }
 
+/// Invokes `dumpadm(8)` to configure the kernel to dump core into the given
+/// `dump_slice` block device in the event of a panic. If a core is already
+/// present in that block device, and a `savecore_dir` is provided, this
+/// function also invokes `savecore(8)` to save it into that directory.
+/// On success, returns Ok(Some(stdout)) if `savecore(8)` was invoked, or
+/// Ok(None) if it wasn't.
 pub fn dumpadm(
     dump_slice: &Utf8PathBuf,
     savecore_dir: Option<&Utf8PathBuf>,
@@ -118,17 +130,20 @@ pub fn dumpadm(
     // Compress crash dumps:
     cmd.arg("-z").arg("on");
 
+    // Do not run savecore(8) automatically on boot (irrelevant anyhow, as the
+    // config file being mutated by dumpadm won't survive reboots on gimlets).
+    // The sled-agent will invoke it manually instead.
+    cmd.arg("-n");
+
     if let Some(savecore_dir) = savecore_dir {
         // Run savecore(8) to place the existing contents of dump_slice (if
         // any) into savecore_dir, and clear the presence flag.
         cmd.arg("-s").arg(savecore_dir);
     } else {
-        // Do not run savecore(8) automatically...
-        cmd.arg("-n");
-
-        // ...but do create and use a tmpfs path (rather than the default
-        // location under /var/crash, which is in the ramdisk pool), because
-        // dumpadm refuses to do what we ask otherwise.
+        // if we don't have a savecore destination yet, still create and use
+        // a tmpfs path (rather than the default location under /var/crash,
+        // which is in the ramdisk pool), because dumpadm refuses to do what
+        // we ask otherwise.
         let tmp_crash = "/tmp/crash";
         std::fs::create_dir_all(tmp_crash).map_err(DumpAdmError::Mkdir)?;
 
@@ -139,7 +154,9 @@ pub fn dumpadm(
 
     match out.status.code() {
         Some(0) => {
+            // do we have a destination for the saved dump
             if savecore_dir.is_some() {
+                // and does the dump slice have one to save off
                 if let Ok(true) = dump_flag_is_valid(dump_slice) {
                     return savecore();
                 }
@@ -166,13 +183,14 @@ pub fn dumpadm(
 
 // invokes savecore(8) according to the system-wide config set by dumpadm.
 // savecore(8) creates a file in the savecore directory called `vmdump.<n>`,
-// where `<n>` is the number in the neighboring plaintext file called `bounds`.
+// where `<n>` is the number in the neighboring plaintext file called `bounds`,
+// or 0 if the file doesn't exist.
 // if savecore(8) successfully copies the data from the dump slice to the
 // vmdump file, it clears the "valid" flag in the dump slice's header and
 // increments the number in `bounds` by 1.
 // in the event that savecore(8) terminates before it finishes copying the
 // dump, the incomplete dump will remain in the target directory, but the next
-// invocation will overwrite it.
+// invocation will overwrite it, because `bounds` wasn't created/incremented.
 fn savecore() -> Result<Option<OsString>, DumpAdmError> {
     let mut cmd = Command::new(SAVECORE);
     cmd.env_clear();
--- a/illumos-utils/src/dumpadm.rs
+++ b/illumos-utils/src/dumpadm.rs
@@ -1,0 +1,186 @@
+use byteorder::{LittleEndian, ReadBytesExt};
+use camino::Utf8PathBuf;
+use std::ffi::OsString;
+use std::fs::File;
+use std::io::{Seek, SeekFrom};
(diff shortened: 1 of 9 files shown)


jclulow:
FYI: I have turned off the `site/compliance/dump` service by default as of https://github.com/oxidecomputer/helios/commit/811f89cdcf2d971dac4e98924cb19b6684364648 which should mean that nothing from the OS image is doing anything with `dumpadm` or `savecore`.  There should be no need to deal with the crash dump directory `pilot` might have created in the past, etc, as that should never end up existing on production systems.

Let me know if you need something else there!

jclulow:
FYI: adjacent relevant follow-up fix I have made to Helios: https://github.com/oxidecomputer/helios/commit/6cc29f68ef7917482ab8a354a209da84ac366079

illumos-utils/src/dumpadm.rs
```diff
@@ -0,0 +1,177 @@
...
+
+    let out = cmd.output().map_err(DumpAdmError::ExecDumpadm)?;
+
+    match out.status.code() {
+        Some(0) => {
+            if savecore_dir.is_some() {
+                if let Ok(true) = dump_flag_is_valid(dump_slice) {
+                    return savecore();
+                }
+            }
```
smklein, line 146: Are these conditionals basically "have we just noticed that a dump already exists on this slice, ready for extraction via savecore?"

sled-agent/src/storage/dump_setup.rs
```diff
@@ -0,0 +1,153 @@
...
+use omicron_common::disk::DiskIdentity;
+use sled_hardware::DiskVariant;
+use slog::Logger;
+use std::collections::HashMap;
+use std::sync::Arc;
+use tokio::sync::MutexGuard;
+
+#[derive(Default)]
+pub struct DumpSetup {
+    savecore_lock: Arc<std::sync::Mutex<()>>,
```
smklein, line 13: Since this lock is particularly opt-in, can we document what it's doing? Looks like "preventing us from running savecore concurrently", right?

sled-agent/src/storage/dump_setup.rs
```diff
@@ -0,0 +1,153 @@
...
+}
+
+impl DumpSetup {
+    pub(crate) async fn poll_dumpdev_setup(
```
smklein, line 17: Nitpick: Maybe `update_dumpev_setup`? Calling this `poll` feels a bit misleading, since it's updating state, not querying for a result.

illumos-utils/src/dumpadm.rs
```diff
@@ -0,0 +1,177 @@
...
+        f.read_u32::<LittleEndian>().map_err(DumpHdrError::ReadMagic)?;
+    if magic != DUMP_MAGIC {
+        return Err(DumpHdrError::InvalidMagic(magic));
+    }
+
+    let version =
+        f.read_u32::<LittleEndian>().map_err(DumpHdrError::ReadVersion)?;
+    if version != DUMP_VERSION {
+        return Err(DumpHdrError::InvalidVersion(version));
+    }
```
iliana, line 65: Is the point of checking the version that the flag semantics might change in the future? I am curious how often `dump_version` changes and whether this check is asking for trouble in the future.

sled-agent/src/storage/dump_setup.rs
```diff
@@ -0,0 +1,153 @@
...
+        &self,
+        disks: &mut MutexGuard<'_, HashMap<DiskIdentity, DiskWrapper>>,
+        log: Logger,
+    ) {
+        let mut dump_slices = Vec::new();
```
smklein, line 22: Not a major issue, but it feels a little wasteful to call this and re-calculate all state for all disks when we're only actually inserting one disk at a time. We do have the `DumpSetup` struct; we could keep a cache of all the information we need?

(Not a blocker for this PR though, the only difference here is a minor efficiency difference)

illumos-utils/src/dumpadm.rs
```diff
@@ -0,0 +1,186 @@
...
+
+// invokes savecore(8) according to the system-wide config set by dumpadm.
+// savecore(8) creates a file in the savecore directory called `vmdump.<n>`,
+// where `<n>` is the number in the neighboring plaintext file called `bounds`.
+// if savecore(8) successfully copies the data from the dump slice to the
+// vmdump file, it clears the "valid" flag in the dump slice's header and
+// increments the number in `bounds` by 1.
+// in the event that savecore(8) terminates before it finishes copying the
+// dump, the incomplete dump will remain in the target directory, but the next
+// invocation will overwrite it.
```
smklein, line 175: Gotcha, so `bounds` is like, "the last successful savecore iteration." seems like this overwrite behavior is kinda what we want then, eh?

sled-agent/src/storage/dump_setup.rs
```diff
@@ -0,0 +1,153 @@
...
+                    ) {
+                        Ok(true) => {
+                            debug!(log, "Dump slice {dump_slice:?} appears to have a valid header; will attempt to savecore to {mountpoint:?}");
+                        }
+                        Ok(false) => {
+                            info!(log, "Dump slice {dump_slice:?} appears to have already been saved");
+                        }
+                        Err(err) => {
+                            debug!(log, "Dump slice {dump_slice:?} appears to be unused: {err:?}");
+                        }
```
smklein, line 119: Should we be breaking out early in these cases? Or is the goal to map "dump slice" to "dump dir" even if no dump exists?

sled-agent/src/storage/dump_setup.rs
```diff
@@ -0,0 +1,153 @@
...
+            }
+        }
+
+        dump_slices.sort();
+        u2_dump_dirs.sort();
+
+        let savecore_lock = self.savecore_lock.clone();
+        tokio::task::spawn_blocking(move || {
+            // TODO: a more reasonable way of deduplicating the effort.
+            let _guard = savecore_lock.lock();
```
smklein, line 60: It looks like this lock exists to prevent us from running savecore concurrently, but we're configuring dumpadm in such as way that it should be running savecore on reboot, if there was a crash.

Do we risk racing between the "system-initiated savecore" and the "sled-agent-initiated savecore"?

illumos-utils/src/dumpadm.rs
```diff
@@ -0,0 +1,186 @@
...
+    ExecDumpadm(std::io::Error),
+
+    #[error("Failed to execute savecore process: {0}")]
+    ExecSavecore(std::io::Error),
+}
+
+pub fn dumpadm(
+    dump_slice: &Utf8PathBuf,
+    savecore_dir: Option<&Utf8PathBuf>,
+) -> Result<Option<OsString>, DumpAdmError> {
```
smklein, line 107: Can we document -- what's the returned string here?

illumos-utils/src/dumpadm.rs
```diff
@@ -0,0 +1,186 @@
...
+            Err(DumpAdmError::InvalidCommand(args, stderr))
+        }
+        Some(n) => Err(DumpAdmError::UnexpectedExitCode(n)),
+        None => Err(DumpAdmError::TerminatedBySignal),
+    }
+}
+
+// invokes savecore(8) according to the system-wide config set by dumpadm.
+// savecore(8) creates a file in the savecore directory called `vmdump.<n>`,
+// where `<n>` is the number in the neighboring plaintext file called `bounds`.
```
iliana, line 169: probably mention the default if bounds is not present (or parseable) is `0`

smklein (approved):
The integration pieces of this make sense to me, and look good!

I have some question that are 90% related to me trying to understand the savecore and dumapdm utilities.

illumos-utils/src/dumpadm.rs
```diff
@@ -0,0 +1,177 @@
...
+pub fn dumpadm(
+    dump_slice: &Utf8PathBuf,
+    savecore_dir: Option<&Utf8PathBuf>,
+) -> Result<Option<OsString>, DumpAdmError> {
+    let mut cmd = Command::new(DUMPADM);
+    cmd.env_clear();
+
+    // Include memory from the current process if there is one for the panic
+    // context, in addition to kernel memory:
+    cmd.arg("-c").arg("curproc");
```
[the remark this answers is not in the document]
iliana: `curproc` dumps more than `kernel` but less than `all`:

```
           kernel

               Kernel memory pages only.

           all

               All memory pages.

           curproc

               Kernel memory pages, and the memory pages of the process whose
               thread was currently executing on the CPU on which the crash
               dump was initiated. If the thread executing on that CPU is a
               kernel thread not associated with any user process, only kernel
               pages will be dumped.
```

sled-agent/src/storage/dump_setup.rs
lifning: there's no system-initiated savecore at boot time, actually! i need to change the flags here to be more clear about that. at boot-time, our system has no configured dump slice until sled-agent sets one, and so won't ever invoke savecore by itself.

illumos-utils/src/dumpadm.rs
```diff
@@ -0,0 +1,177 @@
...
+        cmd.arg("-s").arg(savecore_dir);
+    } else {
+        // Do not run savecore(8) automatically...
+        cmd.arg("-n");
+
+        // ...but do create and use a tmpfs path (rather than the default
+        // location under /var/crash, which is in the ramdisk pool), because
+        // dumpadm refuses to do what we ask otherwise.
+        let tmp_crash = "/tmp/crash";
+        std::fs::create_dir_all(tmp_crash).map_err(DumpAdmError::Mkdir)?;
```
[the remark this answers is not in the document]
lifning: dumpadm doesn't like it if the configured savecore directory doesn't exist, so we're making sure it does (even though we're not using it in this case)

sled-agent/src/storage/dump_setup.rs
lifning: yeah, that's what i'm thinking for a cleaner follow-up, the thought behind what's here now is "get dumps configured at all, these machines can probably handle counting to 10 redundantly a few times at startup for now"

sled-agent/src/storage/dump_setup.rs
lifning: on the contrary, in these cases we definitely want to configure the system to use the dump slice, as there's nothing important there to worry about overwriting

illumos-utils/src/dumpadm.rs
lifning: yeah. i'm operating under the assumption that the position and semantics of the magic number and the version are stable, and that everything else may change if the version changes. ideally (and i've mentioned as much to rm) we wouldn't have any such version-sensitivity and this function would be replaced by shelling out to some kind of `savecore --tell-me-if-there-is-one=/on/this/slice`

PR #3589, opened 2023-07-13
Adjust NTP zone check for new UUID-based zone naming scheme.

luqmana:


davepacheco (approved):

davepacheco:
In case anyone needs this workaround in the future: we ran into this during the dogfood update last night.  All 8 sleds sync'd up their clocks, but rack initialization got stuck on time sync because Sled Agent couldn't verify the timesync.  It was reporting a 500 error.  We worked around this by adding `skip_timesync = true` to the Sled Agent config files, tearing it all down, and starting it again.

smklein:
Oof, sorry about this / thank you for the fix. I didn't have a test setup using NTP, but I should.

PR #3590, opened 2023-07-13
[wicket] Rack overview: display actual sled presence

### LINKED ISSUE #3481: Wicket U/I tweak: Sled overview page does not accurately represent what is populated
All sleds show up as "green" on the visual overview page, regardless of which cubbies are populated.

We should consider updating the colors based on whether or not the cubby is populated? That might be a useful short visual indicator on "what the rack looks like", to guide Oxide support at install time.

jgallagher:
On rack2, before:

wicket-before

and after:

wicket-after

The red styling for switch 0 matches `ComponentPresence::Suspicious` in the PR, which means either "ignition says present but we can't talk to the SP" or "ignition says not present but we _can_ talk to the SP". For this switch we're in the latter case: this switch revision only reports the _other_ switch as an ignition target, not itself. (Note that we do see the _other_ switch as present and drawn as expected.) That has been fixed in future sidecar revisions (https://github.com/oxidecomputer/hardware-sidecar/issues/735). 

Fixes #3481.

### DIFF
--- a/wicket/src/state/inventory.rs
+++ b/wicket/src/state/inventory.rs
@@ -111,6 +111,16 @@ pub struct Sp {
     rot: Option<RotInventory>,
 }
 
+impl Sp {
+    pub fn ignition(&self) -> Option<&SpIgnition> {
+        self.ignition.as_ref()
+    }
+
+    pub fn state(&self) -> Option<&SpState> {
+        self.state.as_ref()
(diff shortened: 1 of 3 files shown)


andrewjstone (approved):

sunshowers:
Thank you for doing this!

PR #3598, opened 2023-07-13
[Bootstore] Persist Fsm::State

andrewjstone:
Persist the lrtq state via a `Ledger` and allow reloading from the Ledger.

This change also creates a `Ledgerable` type for storing replicated network information, but it is not yet used. Support is coming in a follow up.

We also modify `Ledger` to use JSON instead of toml for better generality and commonality with our other tools.


bootstore/src/schemes/v0/peer.rs
```diff
@@ -850,11 +893,82 @@ mod tests {
...
+        // Ensure only one of the peers generation numbers gets bumped
+        assert!(
+            (peer0_gen_new == peer0_gen && peer1_gen_new == peer1_gen + 1)
+                || (peer0_gen_new == peer0_gen + 1
+                    && peer1_gen_new == peer1_gen)
+        );
+
+        // Wipe the learner ledger, restart the learner and instruct it to
+        // relearn its share, and ensure that the neither generation number gets
+        // bumped because persistence doesn't occur.
```
andrewjstone, line 943: This is deterministic because of learner selection of peers to learn from and the fact that the learner timeout is 5 seconds, which is an eternity in terms of computational load here.

bootstore/src/schemes/v0/peer.rs
```diff
@@ -850,11 +893,82 @@ mod tests {
...
+            learner.run().await;
+        });
+        let _ = learner_handle.load_peer_addresses(addrs.clone()).await;
+        learner_handle.init_learner().await.unwrap();
+        let peer0_gen_new_2 =
+            handle0.get_status().await.unwrap().fsm_ledger_generation;
+        let peer1_gen_new_2 =
+            handle1.get_status().await.unwrap().fsm_ledger_generation;
+
+        // Ensure only one of the peers generation numbers gets bumped
```
mkeeter, line 960: Incorrect comment here?

bootstore/src/schemes/v0/peer.rs
mkeeter, line 943: Why doesn't persistence occur?  I believe it's because the learner will ask the same peer for a share, and that peer already knows that it distributed that share to the learner?

bootstore/src/schemes/v0/storage.rs
```diff
@@ -0,0 +1,105 @@
...
+
+/// A persistent version of `Fsm::State`
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct PersistentFsmState {
+    pub generation: u64,
+    pub state: State,
+}
+
+impl Ledgerable for PersistentFsmState {
+    fn is_newer_than(&self, other: &Self) -> bool {
```
mkeeter, line 29: This name is misleading, since it's `>=` instead of `>`.

mkeeter (approved):

bootstore/src/schemes/v0/peer.rs
andrewjstone: It's both. Due to the ordering (via internal BTreeSet) the learner will always ask the same peer first. That peer will also have handed out a share already to that peer. It marks that share as belonging to that peer in [distributed_shares](https://github.com/oxidecomputer/omicron/blob/main/bootstore/src/schemes/v0/fsm.rs#L57) and persisted that fact. The code ensures the same share gets handed out, which doesn't require persistence. We don't want to persist to limit any unnecessary write cycles on the M.2s. 

As a side note, this mechanism is actually intended to solve another problem: If the learner keeps disconnecting /timing out after asking a sled for a share but before receiving it. It will perpetually rotate among sleds asking for a share. By restricting each sled to hand out the same share to a given learner we limit each sled from handing out all there shares by accident and exhausting our pool. So at most we hand out `N-1` extra shares where `N` is the number of initial members in the trust quorum cluster.

bootstore/src/schemes/v0/peer.rs
andrewjstone: Good catch. I fixed it in a follow up, since I'm already on a separate branch.

bootstore/src/schemes/v0/storage.rs
andrewjstone: Good point. I'll change the check, although it doesn't actually matter much for this usage of the code as do checks out of band from the ledger.

Issue #3568, opened 2023-07-12
wicket could degrade better without TrueColor support?


davepacheco:
On the stock Mac Terminal.app, apparently regardless of the color scheme, Wicket looks like this:

![Screenshot 2023-07-11 at 8 08 55 PM](https://github.com/oxidecomputer/omicron/assets/532688/d97e0011-1296-46a4-a16c-333ea38c9985)
![Screenshot 2023-07-11 at 8 06 35 PM](https://github.com/oxidecomputer/omicron/assets/532688/f899ae89-410c-44c5-8dd4-e85d9d004299)

The splash screen is also blank for a few second.

davepacheco:
Mac users are advised to use a terminal with truecolor support.  iTerm2 was easy to install and worked for me.

sunshowers:
For reference: https://github.com/termstandard/colors#not-supporting-truecolor

david-crespo:
> In an interactive program that can read terminal responses, a more reliable method is available, that is transparent to sudo & ssh.
>
> Simply try sending a truecolor value to the terminal, followed by a query to ask what color it currently has. If the response indicates the same color as was just set, then truecolor is supported.

benjaminleonard:
I'm long overdue another look at the wicket UI, we can fold adding colour fallbacks into that work

jclulow:
Interestingly there is a mechanism suggested [for automatic detection](https://github.com/termstandard/colors#querying-the-terminal) in the document you mention, @sunshowers!  But of course it does not appear to work with **Alacritty**, so I've filed an RFE there: https://github.com/alacritty/alacritty/issues/7184

Will be interesting to see if it works with other terminals!

PR #3606, opened 2023-07-13
[wicket] Replace SP inventory debug print with formatted lines

jgallagher:
Replaces our debug print screen:

inventory-old

with formatted output:

inventory-new

This will be more tedious to update with future changes to the information available here, and is already a little tedious in places where related data is split across different elements (e.g., the caboose data is separate from the state, and the rot state is contained within the SP state), but is _so much_ easier to read I think it's definitely worth it.

### DIFF
--- a/wicket/src/ui/panes/overview.rs
+++ b/wicket/src/ui/panes/overview.rs
@@ -582,7 +582,7 @@ fn inventory_description(component: &Component) -> Text {
             power_state,
             revision,
             serial_number,
-            // We git the rot its own section below.
+            // We give the rot its own section below.
             rot: _,
         } = state;
         spans.push(label.into());
--- a/wicket/src/ui/panes/overview.rs
+++ b/wicket/src/ui/panes/overview.rs
@@ -15,6 +15,7 @@ use crate::ui::defaults::style;
 use crate::ui::widgets::IgnitionPopup;
 use crate::ui::widgets::PopupScrollKind;
 use crate::ui::widgets::{BoxConnector, BoxConnectorKind, Rack};
+use crate::ui::wrap::wrap_text;
 use crate::{Action, Cmd, Frame, State};
 use tui::layout::{Constraint, Direction, Layout, Rect};
 use tui::style::Style;
@@ -241,6 +242,17 @@ impl InventoryView {
         }
     }
 
+    /// Returns the wrap options that should be used in most cases for popups.
+    fn default_wrap_options(width: usize) -> crate::ui::wrap::Options<'static> {
+        crate::ui::wrap::Options {
+            width,
+            // The indent here is to add 1 character of padding.
+            initial_indent: Span::raw(" "),
+            subsequent_indent: Span::raw(" "),
+            break_words: true,
+        }
+    }
+
     pub fn draw_ignition_popup(
         &mut self,
         state: &State,
@@ -346,6 +358,13 @@ impl Control for InventoryView {
             Some(inventory) => inventory_description(inventory),
             None => Text::styled("Inventory Unavailable", inventory_style),
         };
+        let text = wrap_text(
+            &text,
+            // -2 each for borders and padding
+            Self::default_wrap_options(
+                chunks[1].width.saturating_sub(4).into(),
+            ),
+        );
 
         let scroll_offset = self.scroll_offsets.get_mut(&component_id).unwrap();
         let y_offset = ComputedScrollOffset::new(
--- a/wicket/src/ui/panes/overview.rs
+++ b/wicket/src/ui/panes/overview.rs
@@ -551,35 +551,6 @@ fn inventory_description(component: &Component) -> Text {
     // blank line separator
     spans.push(Spans::default());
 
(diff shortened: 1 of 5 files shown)


wicket/src/ui/panes/overview.rs
```diff
@@ -439,3 +444,429 @@ impl Control for InventoryView {
...
     }
 }
+
+fn inventory_description(component: &Component) -> Text {
+    let sp = component.sp();
+
+    let label_style = Style::default().fg(colors::OX_OFF_WHITE);
+    let ok_style = Style::default().fg(colors::OX_GREEN_LIGHT);
+    let bad_style = Style::default().fg(colors::OX_RED);
+    let warn_style = Style::default().fg(colors::OX_YELLOW);
```
sunshowers, line 454: Could these be moved to `style.rs`? To address #3568 we'll likely have to switch to a stylesheet approach, and moving this to `style.rs` would be a nice step towards that.

wicket/src/ui/panes/overview.rs
```diff
@@ -439,3 +444,429 @@ impl Control for InventoryView {
...
+        spans.push(label.into());
+    }
+
+    // blank line separator
+    spans.push(Spans::default());
+
+    // Helper closure for appending caboose details (used for both SP and RoT
+    // below).
+    let append_caboose =
+        |spans: &mut Vec<Spans>, caboose: &SpComponentCaboose| {
```
sunshowers, line 558: This can be a `fn append_caboose` I think since it doesn't capture anything in the environment.

wicket/src/ui/panes/overview.rs
```diff
@@ -439,3 +444,429 @@ impl Control for InventoryView {
...
+    // TODO-correctness: component information will change with the IPCC /
+    // topology work that is ongoing; show our current stand-in for now.
+    let mut label = vec![Span::styled("Components: ", label_style)];
+    let components = sp.components();
+    if components.is_empty() {
+        label.push(Span::styled("Not available", bad_style));
+        spans.push(label.into());
+    } else {
+        spans.push(label.into());
+        for component_info in components {
```
sunshowers, line 843: It's worth considering using a destructuring style here.

```rust
let SpComponentInfo { presence, description, device } = component_info;
```

This will make it so that if new fields get added to the component info, then that will be detected as a compile-time issue. That is probably desirable since we'd like to show all component info here.

wicket/src/ui/panes/overview.rs
```diff
@@ -439,3 +444,429 @@ impl Control for InventoryView {
...
+                version.push(Span::styled(v.clone(), ok_style));
+            } else {
+                version.push(Span::styled("Unknown", bad_style));
+            }
+        };
+
+    // Describe the SP.
+    let mut label = vec![Span::styled("Service Processor: ", label_style)];
+    if let Some(state) = sp.state() {
+        spans.push(label.into());
```
sunshowers, line 587: Consider using a destructuring style here (see comment about `SpComponentInfo`).

sunshowers (approved):

wicket/src/ui/panes/overview.rs
```diff
@@ -439,3 +444,429 @@ impl Control for InventoryView {
...
+                        ),
+                        label_style,
+                    ),
+                ]
+                .into(),
+            );
+        }
+    }
+
+    Text::from(spans)
```
sunshowers, line 871: Does this get wrapped today? If not, then it's worth adding wrapping at some point.

wicket/src/ui/panes/overview.rs
sunshowers: Actually I think these are already exposed in `style.rs`, so maybe just use those?

jgallagher: Yep, moved in 88883027

wicket/src/ui/panes/overview.rs
jgallagher: It captured a few small helpers, but those were easy enough to repeat or pass in; moved in 075fb020

wicket/src/ui/panes/overview.rs
jgallagher: Good call; I was already intentionally ignoring a few fields, but making that explicit is a solid improvement. Added in a3e47cdf

wicket/src/ui/panes/overview.rs
jgallagher: Added in 3b1103ae:

it_wraps

wicket/src/ui/panes/overview.rs
```diff
@@ -439,3 +464,461 @@ impl Control for InventoryView {
...
+    let mut label = vec![Span::styled("Service Processor: ", label_style)];
+    if let Some(state) = sp.state() {
+        let SpState {
+            base_mac_address,
+            hubris_archive_id,
+            model,
+            power_state,
+            revision,
+            serial_number,
+            // We git the rot its own section below.
```
sunshowers, line 585:
```suggestion
            // We give the rot its own section below.
```

sunshowers (approved):
Looks great!

PR #3607, opened 2023-07-13
Enable VMM reservoir usage on gimlets

### LINKED ISSUE #3223: sled-agent should use the VMM reservoir by default
With the integration of #3200, sled-agent is capable of configuring the reservoir size, but does not yet do this by default. We should address the blockers here (tracked in #2650) and modify sled-agent to configure a reasonable reservoir size on startup. A ballpack figure is about 80% of DRAM.

jordanhendricks:
With #3200, sled-agent is capable of provisioning the VMM reservoir on startup. It's configured as a percentage of DRAM via sled-agent's config.toml.

This PR toggles this value to 80% for gimlets. I haven't gotten to test this as much as I had hoped due to unrelated issues with my dev setup. But per the hypervisor huddle today, we would like to enable this in the dogfood cluster as soon as we can, and we want this for FCS. So we decided to land this as is.

Fixes #3223

### DIFF
--- a/smf/sled-agent/gimlet/config.toml
+++ b/smf/sled-agent/gimlet/config.toml
@@ -23,7 +23,7 @@ data_link = "cxgbe0"
 
 # Percentage of usable physical DRAM to use for the VMM reservoir, which
 # guest memory is pulled from.
-vmm_reservoir_percentage = 0
+vmm_reservoir_percentage = 80
 
 # Swap device size for the system.
 # We pick 256 GiB somewhat arbitrarily, since the device is sparsely allocated.


jordanhendricks:
I did get a successfully working omicron at least once, and saw that sled agent logged it had provisioned the reservoir. I didn't poke at the system's reservoir live with rsrvrctl, but since I did the testing for the previous PR I'm not worried about that.

smklein (approved):

jordanhendricks:
Landing this without the ubuntu build-and-test job passing (it failed due to a timeout), as we've made a call to disable that job for now.

Issue #3532, opened 2023-07-08
sled-agent: advertise underlay prefix at startup


rcgoodfellow:
We recently had an unexpected sled reboot. When the node came back up, its underlay prefix was not advertised through `ddmd`, and thus the node was not able to reach critical services such as DNS. When the sled agent is starting up, the underlay prefix should always be advertised (this is idempotent, so if it's already advertised, there is no harm).

rmustacc:
Strictly speaking if the sled reboots it will not know what the underlay prefix to use is. It must get this via the LRTQ. So this should be part of the normal transition from being on bootstrap only to unlocking.

rcgoodfellow:
The advertisement should go here.

- https://github.com/oxidecomputer/omicron/blob/5184ab929ce855ba9f92c90c868a7ed570e878c7/sled-agent/src/bootstrap/agent.rs#L510

rcgoodfellow:
I went to go fix this and the issue did not come up on a basic repro attempt.

I brought up a control plane, and from ddm's perspective, things looked like this.

```
ry@masaka:~/src/omicron$ /opt/oxide/mg-ddm/ddmadm get-originated
Prefix
fd00:1122:3344:2::/64
fd00:1122:3344:3::/64
fd00:1122:3344:101::/64
fd00:1122:3344:1::/64
fdb0:a85e:45cd:bd94::/64
```

Then I simulated a reboot.

```
svcadm disable mg-ddm
svcadmvdisable sled-agent
svcadm enable mg-ddm
svcadm enable sled-agent
```

When ddm came back up it had no prefixes (expected). Then sled agent came back up, cleared out all the zones and started rebuilding them one by one. It also started putting prefixes back one by one. By the time sled agent had re-launched all the zones, all the advertisements were also back.

```
ry@masaka:~/src/omicron$ /opt/oxide/mg-ddm/ddmadm get-originated
Prefix
fd00:1122:3344:1::/64
fd00:1122:3344:2::/64
fd00:1122:3344:101::/64
fd00:1122:3344:3::/64
fdb0:a85e:45cd:bd94::/64
```

@smklein is it possible this has been fixed in recent cold boot work since this issue was opened last week?

rcgoodfellow:
This appears to have happened again on Rack 2.

[...]

PR #3611, opened 2023-07-13
Plumb firewall rules for any services on a sled when it comes up post-RSS

### LINKED ISSUE #3519: Nexus zone needs firewall rule plumbing upon re-initialization
The nexus zone was reaped and recreated by sled-agent on rack2 yesterday. The nexus endpoint wasn't accessible outside of the rack even though it was listening on port 443 and didn't report any error. Upon reviewing the sled-agent log, I noticed that the opte port in the re-created instance didn't have the special firewall rule that nexus needs to allow port 443 access:
```
$ grep opte0 /var/svc/log/oxide-sled-agent\:default.log.0 | looker
19:52:31.138Z INFO SledAgent (PortManager): Created OPTE port
    port = Port { inner: PortInner { name: "opte0", _ip: 172.30.2.5, mac: MacAddr6([168, 64, 37, 255, 138, 97]), slot: 0, vni: Vni { inner: 100 }, gateway: Gateway { mac: MacAddr6([168, 64, 37, 255, 119, 119]), ip: 172.30.2.1 }, vnic: "vopte0" } }
20:01:20.873Z INFO SledAgent (PortManager): Setting OPTE firewall rules
    port = "opte0"
    rules = [FirewallRule { direction: In, filters: Filters { hosts: Any, protocol: Proto(TCP), ports: PortList([80, 443]) }, action: Allow, priority: 65534 }]
21:38:37.293Z WARN SledAgent: Deleting existing VNIC
    vnic_kind = GuestVnic
    vnic_name = vopte0
21:38:37.399Z INFO SledAgent: deleting existing OPTE port and xde device
    device_name = opte0
21:42:13.715Z INFO SledAgent (PortManager): Created OPTE port
    port = Port { inner: PortInner { name: "opte0", _ip: 172.30.2.5, mac: MacAddr6([168, 64, 37, 255, 138, 97]), slot: 0, vni: Vni { inner: 100 }, gateway: Gateway { mac: MacAddr6([168, 64, 37, 255, 119, 119]), ip: 172.30.2.1 }, vnic: "vopte0" } }
```
The firewall rules on the opte port only included the zone default rules:
```
/opt/oxide/opte/bin/opteadm dump-layer firewall -p opte0
Layer firewall
Inbound Flows
PROTO  SRC IP           SPORT  DST IP           DPORT  HITS     ACTION                

Outbound Flows
PROTO  SRC IP           SPORT  DST IP           DPORT  HITS     ACTION                

Inbound Rules
ID     PRI    HITS   PREDICATES                             ACTION            
DEF    --     526    --                                     "deny"

Outbound Rules
ID     PRI    HITS   PREDICATES                             ACTION            
DEF    --     1      --                                     "stateful allow"
```
Once I've added the missing rule, API and console became accessible again.
```
BRM42220017 # /opt/oxide/opte/bin/opteadm add-fw-rule --action allow -p opte0 --protocol tcp --ports 443  --dir in --priority 1 --hosts any
         
BRM42220017 # /opt/oxide/opte/bin/opteadm dump-layer firewall -p opte0
Layer firewall
Inbound Flows
PROTO  SRC IP           SPORT  DST IP           DPORT  HITS     ACTION                

Outbound Flows
PROTO  SRC IP           SPORT  DST IP           DPORT  HITS     ACTION                

Inbound Rules
ID     PRI    HITS   PREDICATES                             ACTION            
0      1      0      inner.ip.proto=TCP                     "Stateful Allow"
                     inner.ulp.dst=443                     

DEF    --     526    --                                     "deny"

Outbound Rules
ID     PRI    HITS   PREDICATES                             ACTION            
DEF    --     1      --                                     "stateful allow"
```

luqmana:
Fixes #3519


luqmana:
Haven't tested on multirack yet but with a local single-node softnpu-based deploy, this properly recreated the necessary firewall rules after a `svcadm restart sled-agent`.

sled-agent/src/sled_agent.rs
```diff
@@ -360,15 +360,21 @@ impl SledAgent {
...
 
         // Finally, load services for which we're already responsible.
         //
         // Do this *after* monitoring for harware, to enable the switch zone to
         // establish an underlay address before proceeding.
         sled_agent.inner.services.load_services().await?;
+
+        // Now that we've initialized the sled services, notify nexus again
+        // at which point it'll plumb any necessary firewall rules back to us.
+        sled_agent.notify_nexus_about_self(&log);
```
luqmana, line 376: Note there's already a call to `notify_nexus_about_self` just above but since that's before we've actually initialized any services a Nexus running on a different sled could send us the right rules but we'd have no OPTE ports created yet to apply them to.

nexus/db-queries/src/db/datastore/vpc.rs
```diff
@@ -595,8 +596,14 @@ impl DataStore {
...
             .select(Sled::as_select());
 
-        instance_query
-            .union(service_query)
+        let sleds_filter =
+            sleds_filter.map(|sleds| sleds.to_vec()).unwrap_or(vec![]);
+        sled::table
+            // Grab just the sleds that were asked for (if any)
+            .filter(sled::id.eq_any(sleds_filter))
+            .or_filter(sled::time_deleted.is_null())
```
smklein, line 604: Why use `.or_filter` instead of just `.filter` here? Does this matter in the case when `sleds_filter` is empty?

sled-agent/src/params.rs
```diff
@@ -593,6 +627,158 @@ impl TryFrom<ServiceZoneRequest>
...
+                ServiceType::ManagementGatewayService
+                | ServiceType::Wicketd { .. }
+                | ServiceType::Dendrite { .. }
+                | ServiceType::Maghemite { .. }
+                | ServiceType::Tfport { .. } => {
+                    return Err(AutonomousServiceOnlyError);
+                    // return Err(SetupServiceError::BadConfig(format!(
+                    //     "RSS should not request service of type: {}",
+                    //     svc.details
+                    // )));
```
smklein, line 773: Dead Code?

sled-agent/src/sled_agent.rs
smklein: It makes 100% sense to me *why* we're doing this, but it feels a little odd.

The fact that we store all the services necessary for executing (simply to bootstrap the system) makes sense, so the sled can launch them locally, but all firewall rules are deleted between reboots feels... quirky? Doesn't this mean that in the period of time between "services have started" and "sled agent has notified Nexus, re-received the firewall rules, and re-applied them" we're basically ignoring all user-requested firewall rules? Is that a concern?

sled-agent/src/sled_agent.rs
```diff
@@ -360,15 +360,21 @@ impl SledAgent {
```
smklein, line 360: Do we still need this initial notification? Why just *only* send the latter notification?

(If there's a good reason, we can notify Nexus both times, but it seems simpler to only do it once?)

smklein (approved):

nexus/db-queries/src/db/datastore/vpc.rs
luqmana: Yes, `eq_any` on an empty array would return no rows. From the [diesel docs](https://docs.diesel.rs/master/diesel/expression_methods/trait.ExpressionMethods.html#example-2):
>// Calling `eq_any` with an empty array is the same as doing `WHERE 1=0`

But I think this is still not exactly what I want here. The `or_filter` basically just includes everything regardless. :/ Looks like I'll have to try going back to only adding the filter conditionally. Time to get back to wrestling with diesel...

sled-agent/src/sled_agent.rs
luqmana: These are firewall rules for services running on a sled (nexus & external dns in particular) and not guest instances. So yea, there is a bit of a window between those services coming up and actually being externally accessible.

In theory it's possible an instance could be provisioned on the sled in that window and that would be fine as any firewall rules it needs are plumbed along with it during its initialization on the sled.

sled-agent/src/sled_agent.rs
luqmana: I don't know any reason specifically why we'd _need_ the initial one (and in fact it causes a lot of log noise if the sled in question is the one that's going to be hosting nexus and thus it doesn't yet exists).

nexus/db-queries/src/db/datastore/vpc.rs
smklein: As a cheap fall-back plan: if you box the query, it should be easy to add conditional filters during construction:

```rust
let query = sled::table.filter(sled::time_deleted.is_null()).into_boxed();

// If no filter is provided, we send this to all sleds.
if !sleds_filter.empty() {
  query = query.filter(sled::id.eq_any(sleds_filter));
}
// Keep building query...
```

luqmana: yea, i had something like that earlier but was drowning huge errors about missing trait bounds. But that might've been due to just trying it on the union expr, i think what you have should work.

rcgoodfellow:
I just accidentally took this for a spin working on oxidecomputer/omicron#3532. After a sled-agent restart and a full clobbering and rebuilding of all zones, nexus was instantly reachable from the outside world once things had come back up 🎉 .

PR #3612, opened 2023-07-14
[Bootstore] Network config replication

andrewjstone:
Implement a replication and storage protocol for generic network config blobs inside the bootstore.

Blobs are versioned by a generation number and updated at a single node via `NodeHandle::update_network_config` and broadcast to other connected nodes. When a node comes online it sends its network config to its peer if it has one, and each side updates to the latest version exchanged. Current network config can be retrieved locally via `NodeHandle::get_network_config`.

### DIFF
--- a/bootstore/src/schemes/v0/peer.rs
+++ b/bootstore/src/schemes/v0/peer.rs
@@ -606,24 +606,29 @@ impl Node {
                 "Sending network config with generation {} to {id}",
                 network_config.generation
             );
-            // Send the current network config
-            if let Err(e) = handle
-                .tx
-                .send(MainToConnMsg::Msg(Msg::NetworkConfig(
-                    network_config.clone(),
-                )))
-                .await
-            {
-                warn!(
-                    self.log,
-                    concat!(
-                        "Failed to send network config to connection ",
-                        "management task for {} {:?}"
-                    ),
-                    id,
-                    e
-                );
-            }
+            self.send_network_config(network_config.clone(), id, handle).await;
+        }
+    }
+
+    // Send network config to a peer
+    async fn send_network_config(
+        &self,
+        config: NetworkConfig,
+        peer_id: &Baseboard,
+        handle: &PeerConnHandle,
+    ) {
+        if let Err(e) =
+            handle.tx.send(MainToConnMsg::Msg(Msg::NetworkConfig(config))).await
+        {
+            warn!(
+                self.log,
+                concat!(
+                    "Failed to send network config to connection ",
+                    "management task for {} {:?}"
+                ),
+                peer_id,
+                e
+            );
         }
     }
 
@@ -780,24 +785,12 @@ impl Node {
                     unique_id: accepted_handle.unique_id,
                 };
                 if let Some(network_config) = self.network_config.as_ref() {
-                    // Send the current network config
-                    if let Err(e) = handle
-                        .tx
-                        .send(MainToConnMsg::Msg(Msg::NetworkConfig(
-                            network_config.clone(),
-                        )))
-                        .await
-                    {
-                        warn!(
-                            self.log,
-                            concat!(
-                                "Failed to send network config to connection ",
-                                "management task for {} {:?}"
-                            ),
-                            peer_id,
-                            e
-                        );
-                    }
+                    self.send_network_config(
+                        network_config.clone(),
+                        &peer_id,
+                        &handle,
+                    )
+                    .await;
                 }
 
                 self.established_connections.insert(peer_id.clone(), handle);
@@ -820,24 +813,12 @@ impl Node {
                     }
 
                     if let Some(network_config) = self.network_config.as_ref() {
-                        // Send the current network config
-                        if let Err(e) = handle
-                            .tx
-                            .send(MainToConnMsg::Msg(Msg::NetworkConfig(
-                                network_config.clone(),
-                            )))
-                            .await
-                        {
-                            warn!(
-                                self.log,
-                                concat!(
-                                    "Failed to send network config to ",
-                                    "connection management task for {}: {:?}"
-                                ),
-                                peer_id,
-                                e
-                            );
-                        }
+                        self.send_network_config(
+                            network_config.clone(),
+                            &peer_id,
+                            &handle,
+                        )
+                        .await;
                     }
 
                     self.established_connections
--- a/bootstore/src/schemes/v0/peer.rs
+++ b/bootstore/src/schemes/v0/peer.rs
@@ -999,7 +999,7 @@ mod tests {
             .collect()
     }
 
-    fn initial_config(tempdir: &Utf8TempDir) -> Vec<Config> {
+    fn initial_config(tempdir: &Utf8TempDir, port_start: u16) -> Vec<Config> {
         initial_members()
             .into_iter()
             .enumerate()
@@ -1008,7 +1008,7 @@ mod tests {
                 let network_file = format!("test-{i}-network-config-ledger");
                 Config {
                     id,
-                    addr: format!("[::1]:3333{}", i).parse().unwrap(),
+                    addr: format!("[::1]:{}{}", port_start, i).parse().unwrap(),
                     time_per_tick: Duration::from_millis(20),
                     learn_timeout: Duration::from_secs(5),
                     rack_init_timeout: Duration::from_secs(10),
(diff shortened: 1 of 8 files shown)


bootstore/src/schemes/v0/peer.rs
```diff
@@ -448,6 +522,108 @@ impl Node {
...
+            {
+                warn!(
+                    self.log,
+                    concat!(
+                        "Failed to send network config to connection ",
+                        "management task for {} {:?}"
+                    ),
+                    id,
+                    e
+                );
```
andrewjstone, line 625: Could probably put this in a method since it's identical in 3 places

bootstore/src/schemes/v0/peer.rs
```diff
@@ -448,6 +522,108 @@ impl Node {
...
+                    self.network_config = Some(config.clone());
+                    NetworkConfig::save(
+                        &self.log,
+                        self.config.network_config_ledger_paths.clone(),
+                        config,
+                    )
+                    .await;
+                    // Broadacst the updated config. We only broadcast
+                    // when we successfully update it so we don't trigger an
+                    // endless broadcast storm.
```
mkeeter, line 573: I think this will still trigger a small broadcast storm, since each node in the network will receive a new config, update itself, and then call `broadcast_network_config` itself.

bootstore/src/schemes/v0/peer.rs
```diff
@@ -686,6 +904,31 @@ impl Node {
...
+                if generation > current_gen {
+                    self.network_config = Some(config.clone());
+                    NetworkConfig::save(
+                        &self.log,
+                        self.config.network_config_ledger_paths.clone(),
+                        config,
+                    )
+                    .await;
+                    self.broadcast_network_config(Some(&from)).await;
+                }
```
mkeeter, line 930: If someone sent us an outdated network config (`generation < current_gen`), should we send a message back to them with the current config?

bootstore/src/schemes/v0/peer.rs
```diff
@@ -971,4 +1226,176 @@ mod tests {
...
+                if h.get_network_config().await.unwrap().as_ref()
+                    != Some(&new_config)
+                {
+                    // We need to try again
+                    continue 'outer;
+                }
+            }
+            // Success
+            break;
+        }
```
mkeeter, line 1381:
```suggestion
        loop {
            if Instant::now() - start > POLL_TIMEOUT {
                panic!("network config not replicated");
            }
            if [&handle0, &handle1, &handle2, &learner_handle].into_iter().all(
                |h| {
                    h.get_network_config().await.unwrap().as_ref()
                        == Some(&new_config)
                },
            ) {
                break;
            }
        }
```

bootstore/src/schemes/v0/storage.rs
```diff
@@ -96,10 +97,54 @@ pub struct NetworkConfig {
...
+        mut config: NetworkConfig,
+    ) {
+        // Decrement the generation number of the config, so that when we commit
+        // it the commit bumps it back to the actual generation number. This
+        // keeps the generation number of the network config identical to the
+        // ledger generation.
+        //
+        // Note that generation numbers of valid configs must start at 1, so we
+        // specifically do not use a saturating subtraction.
+        config.generation -= 1;
```
mkeeter, line 124: In that case, maybe `checked_sub(1).unwrap()` as an extra check?

bootstore/src/schemes/v0/peer.rs
```diff
@@ -140,28 +167,64 @@ impl NodeHandle {
...
     pub async fn shutdown(&self) -> Result<(), NodeRequestError> {
-        self.tx.send(NodeApiRequest::Shutdown).await?;
+        self.tx
+            .send(NodeApiRequest::Shutdown)
+            .await
+            .map_err(|_| NodeRequestError::Send)?;
         Ok(())
     }
+
+    /// Update network config needed for bringing up the control plane
```
mkeeter, line 197: The fact that this passes in a `NetworkConfig` (including a `generation`) is a little weird to me.  It means that for an API consumer to update the network config, they have to first call `get_network_config` to extract its generation number, then increment it.  That seems like an internal detail that shouldn't be exposed for an "update network config" endpoint.

bootstore/src/schemes/v0/peer.rs
andrewjstone: Yes, that's intentional although I'm not sure it's strictly necessary inside the rack. The goal is to allow all nodes to learn of the update even if there isn't full connectivity due to a network partition. It's cheap though so I kept it for robustness purposes.

bootstore/src/schemes/v0/peer.rs
andrewjstone: It's not necessary, as peers always send out the latest version when they learn of it or on connection. The only time this is actually likely to occur is on a connection to a peer in which case the node which has the later version has already sent it simultaneously.

bootstore/src/schemes/v0/peer.rs
andrewjstone: I like it!

bootstore/src/schemes/v0/storage.rs
andrewjstone: wait, does this not panic in release mode? Shit. Good call.

bootstore/src/schemes/v0/peer.rs
andrewjstone: This is actually critical to the correctness of the system. The local node itself is never actually updating the `NetworkConfig` directly. It's just informing the bootstore that there is an update. The update is immutable for a given `generation` and it can be written into multiple nodes simultaneously. If we did as you suggest, one node could do the RMW to update its local state simultaneously with another. They'd both bump to generation 2 but with different data. And since we only compare based on generation there would be an unknown conflict with some nodes containing some data and other nodes different data. This is because we don't have consensus or atomic updates across nodes in the bootstore. It's just a dumb replicator of versioned immutable data.

The way this is intended to be used is that update will always happen consistently in Nexus with a CTE that bumps the generation number in CockroachDB. Then a nexus instructs a sled agent in a reliable persistent workflow to inform the bootstore of the modification at the new version, and the bootstore replicates it.  

There is still a race here where the rack can crash before the update is persisted to a bootstore node, but after it is written to CockroachDB. The rack still needs to be able to boot with the old network config so it can come up and the reliable persistent workflow can run to inform the bootstore on a reboot. Even without a crash there cannot every be an atomic cutover from one network config to another because all nodes will not learn of it simultaneously. If, say, every node pointing at router a in generation 1 suddenly stopped working immediately after router b was the new thing to point at in generation 2, then there would be an outage until every node learned of the config and pointed to router b. What we want is to point at both router a and b until every node has cutover and then do another update removing router a.

mkeeter: Okay, having the canonical generation number being in CRDB makes sense to me!

bootstore/src/schemes/v0/storage.rs
mkeeter: Unless the release build it built with [overflow checks](https://rust-lang.github.io/rfcs/1535-stable-overflow-checks.html); we do this in Hubris, but I'm not sure about Omicron.

bootstore/src/schemes/v0/peer.rs
mkeeter: Agreed on all fronts!

bootstore/src/schemes/v0/peer.rs
andrewjstone: Hmm, this actually doesn't work because of the `.await` inside the closure. I get the following compilation error:

```
   Compiling bootstore v0.1.0 (/home/ajs/omicron/bootstore)
error[E0728]: `await` is only allowed inside `async` functions and blocks
    --> bootstore/src/schemes/v0/peer.rs:1354:43
     |
1353 |                 |h| {
     |                 --- this is not `async`
1354 |                     h.get_network_config().await.unwrap().as_ref()
     |                                           ^^^^^^ only allowed inside `async` functions and blocks
```

bootstore/src/schemes/v0/storage.rs
andrewjstone: Done in [cd37295](https://github.com/oxidecomputer/omicron/pull/3612/commits/cd3729569e81ebd495999a73ef5ee06a03af3085)

bootstore/src/schemes/v0/peer.rs
andrewjstone: Done in [cd37295](https://github.com/oxidecomputer/omicron/pull/3612/commits/cd3729569e81ebd495999a73ef5ee06a03af3085)

andrewjstone:
Got approval from @smklein to merge.

smklein (approved):

bootstore/src/schemes/v0/peer.rs
mkeeter: Ugh, I hate it when that happens.  It looks like you may be able to use `join_all`, but I'm fine skipping the cleanup if it's too hard.

andrewjstone: Just bigger fish to fry. Got a marlin and a 55 gallon drum of soybean oil!

PR #3619, opened 2023-07-14
Switch to building on heliosv2

citrus-it:
The pre-requisites for this PR are:

- https://github.com/oxidecomputer/helios/pull/101
- https://github.com/oxidecomputer/propolis/pull/461
- https://github.com/oxidecomputer/crucible/pull/830
- https://github.com/oxidecomputer/dendrite/pull/498
- https://github.com/oxidecomputer/maghemite/pull/73


luqmana:
We also need to update the [`[un]install_opte`](https://github.com/oxidecomputer/omicron/blob/9a09e641137a5d65fbf2e92d23a6e57c34d4b706/tools/install_opte.sh#L29) script at some point as well now that a separate publisher isn't needed for the `driver/network/opte` package.

davepacheco:
One test keeps failing on this PR but we don't want it to block this change.  Filed #3639 to fix it and pushed a commit to this PR to ignore the test for now.

Issue #3580, opened 2023-07-12
`integration_tests::images::test_image_create_bad_image_size` appears to hang?


smklein:
See: https://buildomat.eng.oxide.computer/wg/0/details/01H55KF6XCA5Z1T80V097PNVDA/0iF70b74n9TCKX9VEqlWnoUtcR8liH82CXObGk1Y2Ya8ZvyL/01H55KG8C022ARF2TKBJAXQ7HG

I see this in the logs:

```
5868	2023-07-12T18:37:40.660Z	test integration_tests::images::test_silo_image_create ... ok
5869	2023-07-12T18:37:55.579Z	test integration_tests::images::test_image_create_bad_image_size has been running for over 60 seconds
5870	2023-07-12T18:37:57.650Z	test integration_tests::instances::test_attach_one_disk_to_instance ... ok
```

I audited the log, and all other tests that claim to have run for over 60 seconds have completed, but this particular one seems stuck. I believe the associated PR Is unrelated to virtual disks.

jgallagher:
It looks like on https://buildomat.eng.oxide.computer/wg/0/details/01H55HYZ0D65FE0M0VFERW6DDF/LqRDmRlBcWqvBUJT4kw1nwcJCPBk7gw8Gms4gSqlPJDMsWA4/01H55J01Y32X7YK84GJ47BTQ6D#S5730 it's a slightly different (but close) test? I see

```
test integration_tests::images::test_image_create_bad_content_length has been running for over 60 seconds 
```

as the only stuck test.

jgallagher:
https://buildomat.eng.oxide.computer/wg/0/details/01H55KPK053MMS3A18V5NCEXTD/s1CE0hXujE1uH2Ey8ljt4aFpZOlebYWdZCdfKnbOo6Bv9naK/01H55KRAMWSHJSWE76XTGXRFZP is another hang on another different test:

```
2023-07-12T18:47:09.544Z | test integration_tests::silo_users::test_silo_group_detail_bad_group_id has been running for over 60 seconds

**Comment by davepacheco** [comment]
Excitingly, the CI run for #3622 hit the problem we're trying to catch.  The suspicious thing I see in the log is backoffs for:

```
test_create_a_saml_idp_invalid_descriptor_truncated (oximeter): failed to initialize ClickHouse database, will retry in 219.061027595s
[...]

PR #3622, opened 2023-07-14
time out CI jobs instead of waiting for buildomat timeout

davepacheco:
This is to help debug #3580 and the various test hangs seen lately in the Ubuntu test runner.

### DIFF
--- a/.github/buildomat/jobs/build-and-test-linux.sh
+++ b/.github/buildomat/jobs/build-and-test-linux.sh
@@ -60,9 +60,13 @@ ptime -m cargo build --locked --all-targets --verbose
 # We also don't use `--workspace` here because we're not prepared to run tests
 # from end-to-end-tests.
 #
-
+# Hopefully temporary: we've been seeing hangs in these tests lately.  When this
+# happens, buildomat aborts the job after several hours and we're left with
+# little data with which to debug it (buildomat#8).  For now, implement our own
+# timeout that kills the job first.  On explicit failure like this, buildomat
+# will save the files left around.
 banner test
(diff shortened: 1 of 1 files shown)


smklein (approved):

davepacheco:
Excitingly, the CI run for this PR hit the problem we're trying to catch.  The suspicious thing I see in the log is backoffs for:

```
test_create_a_saml_idp_invalid_descriptor_truncated (oximeter): failed to initialize ClickHouse database, will retry in 219.061027595s
    error = Database(DatabaseUnavailable("error sending request for url (http://[::1]:37603/?output_format_json_quote_64bit_integers=0): error trying to connect: tcp connect error: Connection refused (os error 111)"))
```

That's from Oximeter reporting that Clickhouse isn't running.  I think test setup is stuck because we're waiting for Oximeter to start.  Following along here:
https://github.com/oxidecomputer/omicron/blob/fca839ccfcfba9efad015e79e707f1efaf8258af/nexus/test-utils/src/lib.rs#L759-L776

I think we're inside `start_oximeter()` (because we've seen "Starting Oximeter" in the log and have not seen "Starting test metric Producer Server").

Why isn't Clickhouse running?  We saw "Starting Clickhouse".  One thing that I don't yet understand is that when we create the Clickhouse server, we point it to a log file in a temp directory created by `TempDir`, which claims to use `std::env::temp_dir()`, which claims to use the TMPDIR environment variable.  That appears (from the bash script that runs this job) be set to /var/tmp/omicron_tmp and that means we should have that log file uploaded by buildomat.  But we don't.

PR #3623, opened 2023-07-14
stop logging private key during RSS init

davepacheco:
@jgallagher noticed that during RSS initialization we print the private TLS key.  This stops doing that.

I looked for other instances in the sled agent and Nexus logs by looking for both `BEGIN` and `45,.*45.,*45` (three hyphens, as ASCII in decimal).  I saw both of these in the`Injecting RSS configuration` log message.  I didn't find any others.  (This wasn't the most exhaustive search.)

We could do better here (e.g., use a type or some other metadata to tell Progenitor that this type should impl a dummy Debug?) but urgency is a factor here.

### DIFF
--- a/sled-agent/src/bootstrap/params.rs
+++ b/sled-agent/src/bootstrap/params.rs
@@ -25,7 +25,7 @@ pub enum BootstrapAddressDiscovery {
 
 // "Shadow" copy of `RackInitializeRequest` that does no validation on its
 // fields.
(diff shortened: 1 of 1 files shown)


jgallagher (approved):

davepacheco:
I verified by setting up on maxwell that the "Injecting RSS configuration" message no longer has the private key.  I don't see any other use of the key in the sled agent or Nexus log files.

PR #3624, opened 2023-07-14
Cap disk allocation via the API at 1 TiB

zephraph:
Introduces a 1 TiB limit to disk allocations via the API. As with the instance caps, this is check is only through the entrypoint in the app layer and thus _isn't_ checked deeper in control plane operations like if disks are created by unrelated sagas.

### DIFF
--- a/nexus/src/app/disk.rs
+++ b/nexus/src/app/disk.rs
@@ -28,6 +28,9 @@ use sled_agent_client::Client as SledAgentClient;
 use std::sync::Arc;
 use uuid::Uuid;
 
+use super::MAX_DISK_SIZE_BYTES;
+use super::MIN_DISK_SIZE_BYTES;
+
 impl super::Nexus {
     // Disks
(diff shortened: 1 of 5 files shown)


nexus/src/app/disk.rs
```diff
@@ -148,24 +151,35 @@ impl super::Nexus {
 
         // Reject disks where the size isn't at least
         // MIN_DISK_SIZE_BYTES
-        if params.size.to_bytes() < params::MIN_DISK_SIZE_BYTES as u64 {
+        if params.size.to_bytes() < MIN_DISK_SIZE_BYTES as u64 {
```
david-crespo, line 154: any reason not to make it a `u64` and remove the `as u64` everywhere?

zephraph: Nope, that's probably fine.

david-crespo: sounds like a reason!

PR #3626, opened 2023-07-14
Enable anycast loopback addresseses

internet-diglett:
For our boundary services, we use an anycast ipv6 address. Our existing address allocation and assignment logic did not permit the same address to be assigned multiple times.

* Add an `anycast` column / field for `LoopbackAddress` and reservations
* Update `TestContext` to allow initialization of multiple `Dendrite` instances to simulate multi-switch environments

Another boundary services issue was encountered when launching instances while uplinks are configured on more than one switch. The subsaga nodes would end up generating the same names when generating the subsaga nodes for configuring nat on additional boundary switches.

* Generate a unique name using the index of the NAT entry in combination with the `SwitchLocation`

### DIFF
--- a/nexus/db-queries/src/db/datastore/switch_port.rs
+++ b/nexus/db-queries/src/db/datastore/switch_port.rs
@@ -359,6 +359,9 @@ impl DataStore {
                         crate::db::datastore::address_lot::try_reserve_block(
                             address_lot_id,
                             address.address.ip().into(),
+                            // TODO: Should we allow anycast addresses for switch_ports?
+                            // anycast
+                            false,
                             &conn
                         )
                         .await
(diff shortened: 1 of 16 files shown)


nexus/db-queries/src/db/datastore/switch_port.rs
```diff
@@ -359,6 +359,9 @@ impl DataStore {
                         crate::db::datastore::address_lot::try_reserve_block(
                             address_lot_id,
                             address.address.ip().into(),
+                            // TODO: Should we allow anycast addresses for switch_ports?
```
rcgoodfellow, line 362: Yes, we should. But we should require that this be an explicit aspect of an address that the user specifies, so a follow-up issue is fine here.

rcgoodfellow (approved):
Thanks for the quick turn @internet-diglett!

nexus/db-queries/src/db/datastore/switch_port.rs
internet-diglett: https://github.com/oxidecomputer/omicron/issues/3628

Issue #3620, opened 2023-07-14
Metrics endpoints take several minutes to respond


david-crespo:
Fuller report TBD, just want to track it. It seems to happen on every single request to these endpoints. @davepacheco suggests it's taking all that time to DNS resolution or establish a connection, so I'm looking into that. Didn't happen when 9eac70a362b951aa333244e37caa110983b67528 was deployed, came back today when 7fe972939de3f61b7f115a19004278b2499201cb was deployed. Here are two examples in the network tab. Note the duration on the right. The duration doesn't seem to go up over time. In testing, I saw times around 5 minutes, but also around 1 minute.

<img width="1713" alt="Screenshot 2023-07-14 at 10 28 39 AM" src="https://github.com/oxidecomputer/omicron/assets/3612203/0fa0a1ab-4c5f-41dc-99db-60f1f0807fed">

Metrics endpoints make a call to clickhouse, and I confirmed in the logs that the clickhouse query happens normally at the very **end** of the several minutes. Here's an example bit of Nexus log showing normal request stuff happening at 15:24:53, and then completing at 15:27:15.

```
15:24:53.387Z DEBG 36ed3232-caf0-442b-990e-01d0c0506f1a (dropshot_external): authorize result
    action = Read
    actor = Some(Actor::SiloUser { silo_user_id: a230fbce-8256-467c-b9c9-3e7b7a1e8ae8, silo_id: 6e198f0c-c3a5-4d83-87fb-be945ac1cc16, .. })
    actor_id = a230fbce-8256-467c-b9c9-3e7b7a1e8ae8
    authenticated = true
    local_addr = 172.30.2.5:443
    method = GET
    remote_addr = 172.20.16.70:59581
    req_id = 83c4c6e1-4a47-4d27-892e-347858817b50
    resource = Silo { parent: Fleet, key: 6e198f0c-c3a5-4d83-87fb-be945ac1cc16, lookup_type: ById(6e198f0c-c3a5-4d83-87fb-be945ac1cc16) }
    result = Ok(())
    uri = https://oxide.sys.rack2.eng.oxide.computer/v1/metrics/ram_provisioned?start_time=2023-07-14T14%3A24%3A53.151Z&end_time=2023-07-14T15%3A24%3A53.151Z&limit=3000
15:27:15.904Z INFO 36ed3232-caf0-442b-990e-01d0c0506f1a (dropshot_external): request completed
    local_addr = 172.30.2.5:443
    method = GET
[...]

PR #3634, opened 2023-07-14
[nexus] Avoid locking when accessing DNS resolver

smklein:
This was wrapped in a lock to allow it to be changed dynamically during tests, but test setup no longer requires this customization.

Remove the lock to undo any unnecessary serialization.


david-crespo (approved):
It's beautiful. Related to https://github.com/oxidecomputer/omicron/issues/3620

david-crespo:
It seems to have helped — now they all resolve in small multiples of 5 seconds 😂 But most of them 500.

Screenshot 2023-07-15 at 9 26 43 AM
Screenshot 2023-07-15 at 9 26 12 AM

Issue #3637, opened 2023-07-15
Wiping `crypt/zone` is too aggressive


smklein:
https://github.com/oxidecomputer/omicron/pull/3557 has some flaws, identified by @iliana :

- We intended to "wipe crypt/zone on reboot", to clear out zone filesystems between executions of the sled agent. This, in theory, is fine.
- In practice, the disk may be parsed more than once during the live execution of the Sled Agent, so https://github.com/oxidecomputer/omicron/blob/9a09e641137a5d65fbf2e92d23a6e57c34d4b706/sled-hardware/src/disk.rs#L511-L522 may be invoked, even while there are zones up-and-running, using those filesystems.

PR #3638, opened 2023-07-15
Only wipe crypt/zone if this is the first time we're mounting it

smklein:
Fixes https://github.com/oxidecomputer/omicron/issues/3637


iliana (approved):
Seems like this fixes the issue!

andrewjstone (approved):

PR #3642, opened 2023-07-15
Sled Agent destroys crypt/zone filesystems if it hasn't seen them before

smklein:
This builds on https://github.com/oxidecomputer/omicron/pull/3638 , with a slightly more fine-tuned mechanism.

- `crypt/zone` is created on new U.2s, if it doesn't already exist
- If it **does** exist, then Sled Agent needs to make a call: Should I delete the dataset, possibly containing zones from last boot, or should I leave it alone (for idempotency)?
- In #3638 , we used "was the filesystem previously mounted" as a proxy for this information. However, in cases where the sled agent reboots, this information gets out-of-date -- the filesystem could be mounted, but sled agent should remove the corresponding dataset.

This PR introduces a filesystem property, simply called 'oxide:agent', which is a randomly generated string created statically during the lifetime of the Sled Agent. If the sled agent reboots, this value will be regenerated, which instructs the sled agent to wipe `crypt/zone` and re-create it.


sled-hardware/src/disk.rs
```diff
@@ -503,26 +520,50 @@ impl Disk {
...
-                        Err(err)
+            // Use a value that's alive for the duration of this sled agent
+            // to answer the question: should we wipe this disk, or have
+            // we seen it before?
+            //
+            // If this value comes from a prior iteration of the sled agent,
+            // we opt to remove the corresponding dataset.
+            static AGENT_LOCAL_VALUE: OnceLock<String> = OnceLock::new();
+            let agent_local_value = AGENT_LOCAL_VALUE.get_or_init(|| {
+                Alphanumeric.sample_string(&mut rand::thread_rng(), 12)
```
smklein, line 538: I'm only avoiding a UUID to identify "this is random junk" -- it doesn't matter what it is, as long as it's different each time the sled agent boots.

iliana: I am definitely not certain whether this matters, but if we want the same amount of entropy as a UUIDv4, we want a 21-character long string here (log_2(62^21) >= 122).

sled-hardware/src/disk.rs
```diff
@@ -64,6 +67,20 @@ pub enum DiskError {
...
+        err: Box<zfs::SetValueError>,
+    },
+    #[error(
+        "Encrypted dataset '{dataset}' missing 'oxide:agent' property: {err}"
+    )]
+    CannotParseAgentProperty {
+        dataset: String,
+        #[source]
+        err: zfs::GetValueError,
+    },
```
iliana, line 83: Looks like this variant is unused? (Which is probably a good thing, because otherwise if we wanted to enable wiping a dataset in a later version, it would fail to parse this property because we wouldn't have set it before.)

iliana (approved):

andrewjstone (approved):

sled-hardware/src/disk.rs
smklein: Good catch, I'll remove it. I was trying to handle more of these error cases explicitly, but just decided to lump this into the "auto-destroy" case instead.

sled-hardware/src/disk.rs
smklein: I don't *think* we care about entropy quite as much -- it's not a global space, it's pretty much just a single value that's sled local (so risk of collision is significantly lower) but I can bump the size regardless.

iliana (approved):

PR #3643, opened 2023-07-15
[internal-dns] Access DNS servers concurrently

smklein:
Fixes https://github.com/oxidecomputer/omicron/issues/3620

We should probably consider continuing to tweak these `ResolverOpts` more, but **without** this concurrency, I've been able to consistently reproduce significant timeouts while accessing a mixed set of offline/online DNS servers from Nexus. 

With these setting on my local config, if **any** DNS servers are up, we should be able to resolve the necessary addresses immediately.

### DIFF
--- a/internal-dns/src/resolver.rs
+++ b/internal-dns/src/resolver.rs
@@ -69,7 +69,10 @@ impl Resolver {
                 bind_addr: None,
             });
         }
-        let resolver = TokioAsyncResolver::tokio(rc, ResolverOpts::default())?;
+        let mut opts = ResolverOpts::default();
+        opts.use_hosts_file = false;
+        opts.num_concurrent_reqs = dns_addrs.len();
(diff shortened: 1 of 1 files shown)


internal-dns/src/resolver.rs
```diff
@@ -69,7 +69,10 @@ impl Resolver {
                 bind_addr: None,
             });
         }
-        let resolver = TokioAsyncResolver::tokio(rc, ResolverOpts::default())?;
+        let mut opts = ResolverOpts::default();
+        opts.use_hosts_file = false;
+        opts.num_concurrent_reqs = dns_addrs.len();
```
david-crespo, line 74: Weird that this would make a big difference if it already defaults to 2, though I guess if it's always using up its two on the bad ones it's not enough?

davepacheco (approved):

PR #3647, opened 2023-07-15
WIP: Start to integrate bootstore with sled-agent

andrewjstone:
This is the first part whose goal is saving early network config to the bootstore during RSS so that it can be used during cold boot to restore enough connectivity to bring up the control plane.


andrewjstone:
This is certainly outdated, as this functionality was completed elsewhere.

PR #3652, opened 2023-07-16
LRTQ based disk encryption

andrewjstone:
This PR adds bootstore/lrtq support by launching a tokio task running a `bootstore::Node` in the bootstrap agent and returning a `bootstore::NodeHandle` for interacting with the bootstore. If there are at least 3 sleds in the initial group of sleds sent from wicket to RSS then trust quorum is initialized via `bootstore::NodeHandle::init_rack` and disk encryption keys for the U.2 drives will be derived from the rack secret. If there are fewer than 3 sleds, then the current hard coded secret will be used to derive disk encryption keys.

We don't know whether we are going to be using trust quorum or not before RSS is run, which means that we don't know how to generate disk encryption keys until rack initialization. Therefore we must choose our secret retriever dynamically. And if we end up using trust quorum, we must wait for initialization so we have the key shares needed to recompute the rack secret and derive the disk encryption keys. Because zones are on U.2s we cannot bring up any zones until the rack is initialized. However, we can't initialize the rack without the switch zone, which is where wicket, wicketd, MGS, and RSS run. Therefore this PR also puts the switch zone back on the ramdisk so we can initialize the rack and bring up the other zones.


sled-agent/src/rack_setup/service.rs
```diff
@@ -161,6 +164,9 @@ pub enum SetupServiceError {
 
     #[error("Error during DNS lookup: {0}")]
     DnsResolver(#[from] internal_dns::resolver::ResolveError),
+
+    #[error("Failed to save early network config to bootstore")]
```
andrewjstone, line 168: Need to fix this message

andrewjstone:
Trust quorum rack init implememented. Encryption key derivation coming soon.

andrewjstone:
Encryption key derivation/retrieval implemented. Need to now block the creation of U.2 zpools until the trust quorum is available. I believe that's the last thing needed before I can open this as a PR and test it out.

smklein:
Reviewing now

sled-agent/src/bootstrap/agent.rs
```diff
@@ -265,6 +294,28 @@ async fn sled_config_paths(storage: &StorageResources) -> Vec<Utf8PathBuf> {
...
+    storage: &StorageResources,
+) -> Vec<Utf8PathBuf> {
+    storage
+        .all_m2_mountpoints(sled_hardware::disk::CLUSTER_DATASET)
+        .await
+        .into_iter()
+        .map(|p| p.join(BOOTSTORE_NETWORK_CONFIG_FILE))
+        .collect()
+}
+
```
smklein, line 318: I don't think it's currently an issue because of where you are *currently* calling these functions, but do we want these to return an error if the result is empty? Might be slightly more defensive.

(Admittedly, I did not check if the bootstore validated that the paths are non-empty too)

wicketd/src/rss_config.rs
```diff
@@ -164,6 +164,17 @@ impl CurrentRssConfig {
...
+        // LRTQ requires at least 3 sleds
+        let trust_quorum_peers: Option<
+            Vec<bootstrap_agent_client::types::Baseboard>,
+        > = if known_bootstrap_sleds.len() >= 3 {
+            Some(
+                known_bootstrap_sleds.keys().cloned().map(Into::into).collect(),
+            )
+        } else {
+            None
+        };
```
smklein, line 176: Should we be doing this? Couldn't this just be a configuration option through wicket, that a user could set explicitly?

I have a small amount of fear for "auto-configured options that turn off security features" -- if we set this explicitly, and threw an error in the one/two sled cases ("You enabled trust quorum, but do not have sufficient sleds to have a quorum") that actually seems preferable to quietly shutting it off.

sled-agent/src/bootstrap/agent.rs
```diff
@@ -185,21 +194,41 @@ pub struct Agent {
...
-    // as the KeyManager task should run forever
+    /// We maintain the handle just to show ownership, but don't use it, as the
+    /// KeyManager task should run forever
     #[allow(unused)]
     key_manager_handle: JoinHandle<()>,
 
-    // We maintain a copy of the `StorageKeyRequester` so we can pass it through
-    // from the `HardwareManager` to the `StorageManager` when the `HardwareManger`
-    // gets recreated.
+    /// Used to inofrm key requesters that the KeyManager is ready to be used.
```
smklein, line 202:
```suggestion
    /// Used to inform key requesters that the KeyManager is ready to be used.
```

sled-agent/src/storage_manager.rs
```diff
@@ -574,6 +581,12 @@ impl StorageWorker {
         resources: &StorageResources,
         disk: UnparsedDisk,
     ) -> Result<(), Error> {
+        // We need to encrypt external disks, and so we wait until
+        // the storage key requester is ready.
+        if disk.variant() == DiskVariant::U2 && !self.key_requester.is_ready() {
+            return Ok(());
+        }
```
smklein, line 588: Question for you here, and anywhere else we check `!self.key_request.is_ready()` before maybe returning `Ok(())` -- what causes us to "scan again" once the key requester is ready?

Here's what I vaguely fear:

- Notification to upsert disk A
- Key request is marked as ready
- We never use disk A, because we **think** we upserted it, but in reality we just ignored it

sled-agent/src/bootstrap/agent.rs
```diff
@@ -486,16 +652,26 @@ impl Agent {
     ) -> Result<SledAgentResponse, BootstrapError> {
         info!(&self.log, "Loading Sled Agent: {:?}", request);
 
+        // Initialize the secret retriever used by the `KeyManager`
+        if self.bootstore.get_status().await?.fsm_state == "uninitialized" {
+            // We aren't using LRTQ as this a cluster with fewer than 3 sleds
```
smklein, line 657: This kinda seems like a leap-of-logic here -- it might be right, but it isn't obvious to me.

Why would the bootstore being "uninitialized" imply this?

sled-agent/src/bootstrap/agent.rs
```diff
@@ -434,6 +540,66 @@ impl Agent {
...
+                                current_peers = peers;
+                                if let Err(e) = bootstore_node_handle
+                                    .load_peer_addresses(current_peers.clone())
+                                    .await
+                                {
+                                    warn!(
+                                        log,
+                                        concat!("Bootstore comms error: {}. ",
+                                        "bootstrap agent must be terminating?.",
+                                    ),
```
smklein, line 583: Why is this the case? I see you believe it because of the `return`, but I don't grok why any error from `load_peer_addresses` would imply this.

sled-agent/src/bootstrap/agent.rs
```diff
@@ -434,6 +540,66 @@ impl Agent {
...
 
+    async fn poll_ddmd_for_bootstore_peer_update(
+        log: &Logger,
+        bootstore_node_handle: bootstore::NodeHandle,
+    ) -> JoinHandle<()> {
+        let log = log.clone();
+        tokio::spawn(async move {
+            // Explicit fail fast. We can't do anything without a DdmAdminClient.
+            // Do we need to loop as in wicketd?
+            let ddmd_client = DdmAdminClient::localhost(&log).unwrap();
```
smklein, line 551: This feels like a TODO? Is it possible for this to fail?

sled-agent/src/rack_setup/service.rs
```diff
@@ -866,6 +879,15 @@ impl ServiceInner {
         };
         let config = &plan.config;
 
+        // Initialize the trust quorum if there are peers configured.
+        if let Some(peers) = &config.trust_quorum_peers {
+            // The value doesn't really matter here. It just needs to be unique.
+            let rack_uuid = Uuid::new_v4();
```
smklein, line 885: If this value doesn't matter, why is it exposed from the bootstore API? Why not just have this be an internal value within `init_rack`?

(There is another "rack_uuid" that we generate in RSS and send to Nexus, so this seems kinda confusing to have it re-defined)

illumos-utils/src/zfs.rs
```diff
@@ -9,6 +9,16 @@ use camino::Utf8PathBuf;
...
 
+// These locations in the ramdisk must only be used by the switch zone.
+//
+// We need the switch zone online before we can create the U.2 drives and
+// encrypt the zpools during rack initialization. Without the switch zone we
+// cannot get the rack initialization request from wicketd in RSS which allows
+// us to  initialize the trust quorum and derive the encryption keys needed for
+// the U.2 disks.
+pub const ZONE_ZFS_RAMDISK_DATASET_MOUNTPOINT: &str = "/zone";
+pub const ZONE_ZFS_RAMDISK_DATASET: &str = "rpool/zone";
```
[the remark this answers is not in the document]
andrewjstone: All good, was the easiest bit of code to add back and it clarified again how it all fit together :)

sled-agent/src/bootstrap/agent.rs
andrewjstone: The bootstore doesn't validate that they are non-empty. I wrote it like this copying the `sled_config_paths` above. I'm fine making all of these return an error and `expect` when used if you think it would make things safer.

smklein: I think it would be worthwhile defensive coding, yeah. Would be nice to catch this early and explicitly.

wicketd/src/rss_config.rs
andrewjstone: Good catch! I had a similar thought when writing this code. The only thing that made me confident was the fact that we aren't going to sell a rack to someone with two sleds. Adding a configuration through wicket isn't a bad idea, but its a touch more work right now, that I don't think is strictly necessary. Mind if I open an issue for it and mark it security? I can target it for MVP.

sled-agent/src/storage_manager.rs
andrewjstone: Great question! I was also concerned about this, but I think it's fine from my reading of the code. Please correct me if I'm wrong.

We set the readiness at the [beginning of `request_sled_agent`](https://github.com/oxidecomputer/omicron/blob/main/bootstore/src/schemes/v0/storage.rs#L58-L87). `request_sled_agent` always gets called in the `Before` state on cold boot.  The other important time `request_sled_agent` gets called is via RSS init. During that call we also start in before. The only time I think we start in `After` is when RSS reset gets called.

In the `Before` state we always call [`SledServer::start`](https://github.com/oxidecomputer/omicron/blob/main/bootstore/src/schemes/v0/storage.rs#L58-L87) which calls `SledAgent::new` which ends up [starting a new hardware monitor task](https://github.com/oxidecomputer/omicron/blob/09c6faab7770be71d62d5dd91ccd75a36d87b6a5/sled-agent/src/sled_agent.rs#L365-L367) which does a [`full_hardware_scan`](https://github.com/oxidecomputer/omicron/blob/09c6faab7770be71d62d5dd91ccd75a36d87b6a5/sled-agent/src/sled_agent.rs#L421) which I believe will upsert all the disks.

However, you now have me concerned about the opposite problem. When we do an RSS reset I don't think we unset the readiness. I'll need to look into this.

sled-agent/src/bootstrap/agent.rs
andrewjstone: Another good find. I should definitely add a comment here.

We only call `RequestSledAgent` when a `SledAgentRequest` has been noticed by the bootstrap agent either directly received from RSS or found in the ledger on cold boot. RSS always calls agent`bootstore::init_rack` before any `SledAgentRequest` gets sent and must successfully complete for the `SledAgentRequest`s to be sent. `bootstore::rack_init` can only complete if every single sled has moved to `Initial_member` state. 

Therefore we can be sure that if we are still in the `uninitialized` state when `request_sled_agent` has been called that we had fewer than 3 sleds and the trust quorum wasn't used because we must have skipped over `bootstore::rack_init`.

See https://github.com/oxidecomputer/omicron/blob/09c6faab7770be71d62d5dd91ccd75a36d87b6a5/sled-agent/src/rack_setup/service.rs#L882-L902 for the ordering.

Now, again, the use of this method on rack_reset has me concerned that 'de-initialization' doesn't occur. I think that actually may not be a problem as long as we do a reboot after rss reset, but it still seems racy to me, as U.2. disks can be upserted right after the reset.

sled-agent/src/bootstrap/agent.rs
andrewjstone: You are right. I inferred this from the fact that `load_peer_addresses can only return a `NodeRequestError::Send` which can only result when the receiver side of the tokio MPSC channel hangs up. The receiver side is owned by the `bootstore::Node` which is running in its own task with a handle owned by the bootstrap agent. This task is supposed to run forever but of course it can panic due to programmer error driven invariant violations and possibly other things I'm unaware of. This means that this log message is incorrect.

However, I'm not really sure what to do in this scenario. The bootstore::Node is not supposed to stop. In theory it could be restarted, but it's probably best to panic here and let the OS process restart. What are your thoughts @smklein ?

sled-agent/src/rack_setup/service.rs
andrewjstone: I exposed it specifically so that i could use the `rack_uuid` that gets generated in RSS and sent to Nexus, as that would help inform us which rack a sled came from easily if an initialized sled was place in another rack. I couldn't find where it was created, but I see now it's in the `SledPlan` which actually happens before this code. I will go ahead and use that UUID. Thank you for pointing this out!

sled-agent/src/bootstrap/agent.rs
andrewjstone: Fixed in https://github.com/oxidecomputer/omicron/pull/3652#discussion_r1265970198

sled-agent/src/bootstrap/agent.rs
andrewjstone: Fixed in 8f3d383dc

wicketd/src/rss_config.rs
smklein: I am fine deferring this, but I'd like to remove the comments scattered through the code about "we do this special thing if there are more than three sleds". That seems like a policy decision, and one we should keep in a single spot.

sled-agent/src/bootstrap/agent.rs
smklein: I agree that "de-initialization" seems like it should be a thing? Maybe we should wipe the bootstore in rack reset?

But also, that still seems like a way of justifying "this is what the code happens to do today". Isn't the decision to be "hard-coded" or not very known to the bootstore? Like, don't we have an `is_hardcoded` method somewhere? Why not use **that** to make this call?

sled-agent/src/storage_manager.rs
smklein: I still think this is pretty dependent on "the code is right because the thing a module away from me is implicitly happening to call me in a peculiar way". Silently opting to ignore a disk feels like it's bug-prone behavior, especially because there's no indication to the caller that we **should** re-scan.

As one example of this: I don't think your example works with synthetic disks, which get injected once?

Backing up a sec, what's the **point** of this "key_requester.is_ready" stuff? It's supposed to basically tell us to "back off" on loading U.2s until we've gotten far enough that we can successfully put the `crypt` dataset in place, right?

Take a look at `do_work_internal` in this file -- maybe it makes sense to simply track all added U.2s in some sort of "deferred queue" that gets unloaded once this variable flips to true? (Basically, just push these disks onto a HashSet before returning in all these spots). Then the caller wouldn't need to know to re-scan, **and** we'd handle the synthetic disk case.

sled-agent/src/bootstrap/agent.rs
smklein: From the type-system, that's not super clear -- it's acting on a `NodeRequestError`, I think, which could have a variety of variants.

I'm not sure I'm even suggesting this, but  what would be wrong with "logging an error and retrying later?" If the bootstore is undergoing a transient error -- maybe it gets reset, like you said -- wouldn't we just want to loop again and refresh the set of peers at some point in the future?

(That also seems like it's justification for setting `current_peers = peers` only in the case where we successfully pass this info through the `bootstore_node_handle`).

sled-agent/src/storage_manager.rs
andrewjstone: Ooh. Great idea! I like this much better. This was just the first way to do it that I thought of yesterday, and I agree with you that it is full of peril. Very much +1 to a more straightforward and obviously correct path!

sled-agent/src/bootstrap/agent.rs
andrewjstone: > Maybe we should wipe the bootstore in rack reset?

We *have* to wipe the bootstore on reset. That's the only way it's safe. We either succeed on rack_init and every single initial member is initialized or we have to wipe it. I had assumed prior to this that we were doing a *factory* reset, which would clear the M.2s, as we discussed in the update sync meetings a few months ago. But it looks like we are leaving the sled agent running after we do the wipe, which is a cause for concern to me. If we wiped the M.2s and restarted the process state would then reload from the M.2s and the bootstore would go back to being uninitialized. I'm concerned that we are trying to unwind a bunch of mutable state not just in the bootstore but all throughout the sled agent, and there's no real good way to tell that this is correct.

> But also, that still seems like a way of justifying "this is what the code happens to do today". Isn't the decision to be "hard-coded" or not very known to the bootstore? Like, don't we have an is_hardcoded method somewhere? Why not use that to make this call?

The bootstore doesn't know anything about the key manager or the sled-agent, or disk encryption keys. All it knows about are key shares and how to go about reconstructing a rack secret with those shares. It provides that rack secret to the sled agent which then is able to derive various keys from it using HKDF and other algorithms. Right now the only keys are disk encryption keys, but we also plan to derive the root private keys for our internal CA from the rack secret in the future. 

The `HardcodedSecretRetriever` is just another implementation of the `SecretRetriever` trait that is used to provide input key material (IKM) to the `KeyManager` so that it can derive encryption keys. The bootstore does not know anything about it. This is also why we've been able to be running with ZFS encryption turned on this whole time and none of the code I touched changes how that works. It just changes the raw secret input to the whole thing. 

Now, since we effectively know how to build an arbitrary key-value store on top of the bootstore, we *could* put extra information in a key related to this, but that wouldn't be known to lrtq, it would just be arbitrary data that the sled-agent could read and write. Maybe that's the right decision. But then again, we probably don't even need to do that since we can put that information in the persisted `SledAgentRequest` itself. I can add that information there in this PR, and I think that would be a good idea to clean up this somewhat convoluted logic leap, that only makes sense with a deep understanding of how the bootstore and sled-agent are ordered together.

sled-agent/src/bootstrap/agent.rs
andrewjstone: Here's the thing, I don't think the error here is transient. The `bootstore::Node` should never shutdown. The only reason it would shutdown is if it panic'd. Right now I only know of two explicit panics, and both of those are invariant violations due to programmer error that I want to know about immediately. I suppose we could respawn the task, but that doesn't make me very comfortable.

However, your point about the type system is spot on. I think for these types of permanent errors it would be nice to return a different type. The problem is that a tokio send error can happen on *any* API call, and is really unexpected. Instead of returning an error we could have the tokio task panic, but we never actually wait on the result, because we expect the task to run forever, as we do in other places. Honestly, I'm not really sure what to do here. 

Maybe the best thing is to do an explicit match on the `NodeRequestError` and panic if it's a `Send` variant, which it always will be in the case of this call.

sled-agent/src/bootstrap/agent.rs
andrewjstone: Yeah, I would assume it's possible to fail. I should definitely just add a retry to handle the case of spurious failures.

sled-agent/src/rack_setup/service.rs
andrewjstone: Fixed in [d953813](https://github.com/oxidecomputer/omicron/pull/3652/commits/d9538139aee296fd64b6db1813a2bd3573bff693)

sled-agent/src/storage_manager.rs
andrewjstone: After taking a closer look at your proposal I think we'll have to add another `StorageWorkerRequest` to inform it to go ahead and actually upsert the disks it had queued after the system is ready. We also may want to pass the `HashSet` in mutably from `do_work` so that if there is an error and retry we don't lose the disks.

sled-agent/src/bootstrap/agent.rs
andrewjstone: I added a boolean to `SledAgentRequest` and used that to decide whether to use trust quorum in 1607e8bdaf00a8e22832a77b83bf5cdf10afdb59

The code is *much* more clear now. 

However, I still remain concerned about trying to reset the memory state of the process. I really think we should reconsider that. An RSS reset command should, IMO, wipe the M.2s and take the reboot the sled. Practically speaking this is really what we do, since a mupdate immediately takes the sleds down to A2 anyway before it does anything.

sled-agent/src/bootstrap/agent.rs
smklein: If we want to make a short-term move on this:

- I think it's okay for this task to exit on error, if it wants to, particularly if you're committing to the statement "if we see any error here, we should not retry".
- I would just be cautious about documenting "this is what an error means for the other side." It's this task's choice to exit, which is fine, but I just am scared of making inferences about what the error means from the other side.

wicketd/src/rss_config.rs
andrewjstone: Thinking more about this, I don't think it is correct to give the user the choice to disable trust quorum if they have 3 sleds or more. We can alert them that trust quorum is disabled if they try to configure a rack with only 1 or 2 sleds, but they should not be able to disable trust quorum. Since that's the case, I think the current logic is correct, we just must add a log message and UI warning. I will add a log message and open an issue for a UI warning.

andrewjstone: I created an [issue](https://github.com/oxidecomputer/omicron/issues/3690)

andrewjstone: Logging added in 1ad6da263

sled-agent/src/bootstrap/agent.rs
andrewjstone: Ok, I think that's fair. We can log it and figure out why it crashed later. In the short term this will have practically no effect on the system as the bootstore is really only used during rack initialization. What will happen (until sled-agent is restarted one way or another) is that this sled will be unable to hand out shares to new sleds added to the rack and it will not be able to accept updates for its early networking information once we add that capability.

andrewjstone: Implemented in [d8fcf7e](https://github.com/oxidecomputer/omicron/commit/d8fcf7eecf302be91d647f228708259e2123fc23)

sled-agent/src/bootstrap/agent.rs
andrewjstone: Retries implemented in [d8fcf7e](https://github.com/oxidecomputer/omicron/commit/d8fcf7eecf302be91d647f228708259e2123fc23)

sled-agent/src/storage_manager.rs
andrewjstone: Implemented in [2151d7f](https://github.com/oxidecomputer/omicron/pull/3652/commits/2151d7fba79c5729286c138dbb35edfdc28824fe)

sled-agent/src/bootstrap/agent.rs
```diff
@@ -185,21 +197,37 @@ pub struct Agent {
...
+    /// The join handle of a task that polls DDMD and updates the bootstore with
+    /// known peer addresses. This task runs forever. This handle field shows
+    /// ownership.
+    #[allow(unused)]
+    bootstore_peer_update_handle: JoinHandle<()>,
 }
 
 const SLED_AGENT_REQUEST_FILE: &str = "sled-agent-request.toml";
+const BOOTSTORE_FSM_STATE_FILE: &str = "bootstore-fsm-state.json";
+const BOOTSTORE_NETWORK_CONFIG_FILE: &str = "bootstore-network-config.json";
```
smklein, line 230: FYI I added these to https://github.com/oxidecomputer/omicron/issues/3704

sled-agent/src/storage_manager.rs
```diff
@@ -866,8 +929,11 @@ impl StorageWorker {
         &mut self,
         resources: StorageResources,
     ) -> Result<(), Error> {
+        // We queue U.2 sleds until the StorageKeyRequester is ready to use.
```
smklein, line 932:
```suggestion
        // We queue U.2 sleds until the StorageKeyRequester is ready to be used.
```

sled-agent/src/storage_manager.rs
```diff
@@ -922,6 +1004,30 @@ impl StorageWorker {
                             }
                             let _ = responder.send(Ok(()));
                         }
+                        KeyManagerReady => {
```
smklein, line 1007: This is great, it's a solid implementation of the handwavey idea I had in mind

smklein (approved):

PR #3672, opened 2023-07-17
[internal-dns] IPv6 DNS resolver for underlay

smklein:
Use the https://docs.rs/trust-dns-resolver/0.22.0/trust_dns_resolver/config/enum.LookupIpStrategy.html#variant.Ipv6Only lookup strategy to marginally improve performance of DNS lookups. 

Related to https://github.com/oxidecomputer/omicron/issues/3620 , builds on https://github.com/oxidecomputer/omicron/pull/3643


smklein:
I'm not being particularly scientific here, but I'm seeing client requests to 3-5 DNS servers taking ~0.2-0.3 seconds without this PR, and consistently ~0.1 seconds or less with this PR.

luqmana (approved):

Issue #726, opened 2022-03-07
[sled-agent] Parallelize service initialization


smklein:
Note: This issue tracks a follow-up from https://github.com/oxidecomputer/omicron/pull/686

We iterate using loops when initializing services - if we're confident the services can be initialized independently, we could parallelize this step.

PR #3676, opened 2023-07-17
Improve "services PUT" API: Execute in parallel and allow zone teardown

smklein:
This should reduce the execution time of both RSS and sled boot a fair bit.

Comparing the moment of "omicron-package install" to "Nexus Handoff Complete", indicating that RSS has fully executed, on a 24 core, 64 GiB workstation running Helios:

Without this change: 9m8.171s
With this change: 3m58.622s 

## Before this PR

- The `/services` PUT API was purely additive. Removing anything from the set of zones to-be-run was unsupported
- Zone initialization was performed serially.

## With this PR

- The `/services` PUT API can now both add and remove zones.
- Zone initialization is concurrent, within a `try_for_each_concurrent` invocation.

Fixes https://github.com/oxidecomputer/omicron/issues/726

### DIFF
--- a/sled-agent/src/services.rs
+++ b/sled-agent/src/services.rs
@@ -20,8 +20,8 @@
 //! of what other services Nexus wants to have executing on the sled.
 //!
 //! To accomplish this, the following interfaces are exposed:
-//! - [ServiceManager::ensure_all_services] exposes an API to request a set of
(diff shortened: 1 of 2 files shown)


iliana (approved):
yay parallelism!

PR #3677, opened 2023-07-17
Put process core dumps onto the U.2 debug zvol

lifning:
(Part of #2478, continued in #3713)

This configures coreadm to put all core dumps onto the M.2 'crash'
dataset, and creates a thread that moves them all onto a U.2 'debug'
dataset every 5 minutes.

This also refactors the dumpadm/savecore code to be less redundant and
more flexible, and adds an amount of arbitrary logic for e.g. picking
the U.2 onto which to save cores.

### DIFF
--- a/illumos-utils/src/coreadm.rs
+++ b/illumos-utils/src/coreadm.rs
@@ -1,0 +1,62 @@
+use camino::Utf8PathBuf;
+use std::ffi::OsString;
+use std::os::unix::ffi::OsStringExt;
+use std::process::Command;
+
+#[derive(thiserror::Error, Debug)]
+pub enum CoreAdmError {
+    #[error("Error obtaining or modifying coreadm configuration. core_dir: {core_dir:?}")]
+    Execution { core_dir: Utf8PathBuf },
+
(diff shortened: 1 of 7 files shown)


illumos-utils/src/coreadm.rs
```diff
@@ -0,0 +1,57 @@
...
+pub fn coreadm(core_dir: &Utf8PathBuf) -> Result<(), CoreAdmError> {
+    let mut cmd = Command::new(COREADM);
+    cmd.env_clear();
+
+    // disable per-process core patterns
+    cmd.arg("-d").arg("process");
+
+    // use the global core pattern
+    cmd.arg("-e").arg("global");
+
```
citrus-it, line 35: I'd suggest also explicitly configuring the core file content here, and enabling core dumps from `setid` processes too.

```
-e global-setid
```

```
-G stack+heap+shm+ism+dism+text+data+rodata+anon+shanon+ctf+symtab
```
(and consider adding `debug` to also include any DWARF data from the executable and libraries)

You could also use `-G default` or `-G default+debug` but the default set may not be stable.

citrus-it (approved):

PR #3678, opened 2023-07-17
Update install_opte.sh for helios2

citrus-it:


luqmana (approved):
Sweet, thanks for updating this! There's also `uninstall_opte.sh` but that's only ever run manually compared to this one which is called from the main `install_prerequisites.sh` script so fine to do as a followup.

citrus-it:
> Sweet, thanks for updating this! There's also `uninstall_opte.sh` but that's only ever run manually compared to this one which is called from the main `install_prerequisites.sh` script so fine to do as a followup.

The uninstall script looks fine for the moment as it checks everything before doing it. As it stands, running it on heliosv1 will do the right thing and it's basically a no-op on v2.

I tested this on a v2 VM and it ran through, leaving me with opte installed.

PR #3684, opened 2023-07-17
Bundle xcvradm with switch zone

citrus-it:
This has proven useful during customer installation troubleshooting,
and so far has been manually copied into the rack.

Starting fairly soon, some customers will be less open to arbitrary
binaries being injected from technician's laptops so this includes the
xcvradm binary in the switch zone image under /opt/oxide - for the
illumos asic target only. Other OSs get a non-functional stub.


citrus-it:
```
% gtar tvf out/switch-asic.tar.gz | grep xcvr
-rwxr-xr-x 0/0        20811888 2006-07-24 01:21 root/opt/oxide/bin/xcvradm
```

tools/update_transceiver_control.sh
```diff
@@ -0,0 +1,66 @@
...
+          DRY_RUN="yes"
+          ;;
+        *)
+          usage
+          ;;
+      esac
+    done
+
+    TARGET_COMMIT=$(get_latest_commit_from_gh "$REPO" "$TARGET_COMMIT")
+    install_toml2json
```
luqmana, line 62: Is this needed here?

citrus-it: It is not, thank you.

luqmana (approved):

Issue #3582, opened 2023-07-12
Sporadic omicron-package failures creating sp-sim-softnpu.tar.gz


rcgoodfellow:
When running `omicron-package package` package recently, there are sporadic failures creating `sp-sim-softnpu.tar.gz`

```
Error: failed to create switch-softnpu in "out/"

Caused by:
    Missing gzip header from out/sp-sim-softnpu.tar.gz - cannot add it to zone image
*** Error code 1
```

Retrying (sometimes a few times) usually works.

smklein:
I saw this in CI recently too: https://buildomat.eng.oxide.computer/wg/0/details/01H5E0REHHAHWDZEY7PA163R1N/DUeTWHizDgxYmOuT5sxGFbK0TI3Lo5u92ZWLl4j9ti2PwnOV/01H5E0RR0VHQB2714RSN51SB6G

iliana:
Trying to repro this locally now.

PR #3686, opened 2023-07-18
remove composite dependencies on composites

iliana:
I think this might work around the issue causing #3582. Both sp-sim-softnpu and omicron-gateway-asic are composite packages that depend on other composite packages. It turns out this sort of dependency does not currently work: https://github.com/oxidecomputer/omicron/blob/e2aa6658032d0b23fc487d93e233bdbcbdae817a/package/src/bin/omicron-package.rs#L398-L432

As best as I can tell, the packages omicron-gateway-asic, omicron-gateway-softnpu, omicron-gateway-stub, sp-sim-softnpu, and sp-sim-stub created a layered dependency that we weren't handling. But we can trivially flatten this dependency tree with a little repetition. The two removed packages, sp-sim-{softnpu,stub}, were listed as `intermediate_only = true`.

I am thinking about adding a check to ensure that we don't create nested dependencies we aren't handling in the future, for now, but I want to get eyes on this sooner so that CI can be more reliable for upcoming builds.

### DIFF
--- a/package-manifest.toml
+++ b/package-manifest.toml
@@ -287,24 +287,6 @@ source.paths = [ { from = "smf/sp-sim", to = "/var/svc/manifest/site/sp-sim" } ]
 output.type = "zone"
 output.intermediate_only = true
 
-[package.sp-sim-stub]
-service_name = "sp-sim-customizations"
-only_for_targets.image = "standard"
-only_for_targets.switch = "stub"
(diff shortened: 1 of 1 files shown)


iliana:
(I will probably prefer #3687's solution to this problem instead, but may leave this open for a few days just in case.)

PR #3687, opened 2023-07-18
omicron-package: use topological sort for nested composite package deps

### LINKED ISSUE #3582: Sporadic omicron-package failures creating sp-sim-softnpu.tar.gz

luqmana:
Fixes #3582 hopefully.

### DIFF
--- a/.github/buildomat/jobs/deploy.sh
+++ b/.github/buildomat/jobs/deploy.sh
@@ -135,7 +135,8 @@ for p in /input/ci-tools/work/end-to-end-tests/*.gz; do
 	chmod a+x "tests/$(basename "${p%.gz}")"
 done
 
-ptime -m pfexec ./tools/create_virtual_hardware.sh
+pfexec zpool create -f scratch c1t1d0 c2t1d0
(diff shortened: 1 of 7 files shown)


iliana (approved):

luqmana:
Also realized we were building `sp-sim` for `switch=asic` builds even though it's not actually needed.

smklein (approved):
This is great, and... so much more elegant than what I would have done. Thanks for adding this!

davepacheco:
Does anybody happen to have the text of the error message that you'd get when running into this bug?

luqmana:
> Does anybody happen to have the text of the error message that you'd get when running into this bug?

Something like what [ry ran into](https://github.com/oxidecomputer/omicron/issues/3582):
```
Error: failed to create switch-softnpu in "out/"

Caused by:
    Missing gzip header from out/sp-sim-softnpu.tar.gz - cannot add it to zone image
*** Error code 1
````

PR #3692, opened 2023-07-18
[wicket] add page up/page down/home/end to the keymap

### LINKED ISSUE #3653: [wicket] allow page up/down, home/end for quicker navigation
Scrolling up and down the update status window can be a bit slow if you want to get to the bottom. Vim bindings seem to sort of work, h/j/k/l are as expected. g goes to the top but G doesn't go to the bottom.

sunshowers:
Model this with a new `PendingScroll` action stored on the view. This
action is resolved at draw time.

This ended up bundling several more changes (sorry!):

* Fix an off-by-one issue in the inventory view -- the bottom row of text was getting cut
  off previously, whoops :(
* Fix update pane popup scroll state to be more like the better model
  @jgallagher created while writing the rack setup screen. Previously,
  we were storing a single scroll state at the top. This was actually
  incorrect because when we moved from "waiting" to "failed", we weren't
  setting the scroll state to Some. The new model stores scroll state
  inline, and is correct as a result.
* Put in a compile-time separation between scrollable and non-scrollable
  popups -- I made a couple of mistakes with unwrapping scroll popup
  state which this caught.

I tested all the screens which support scrolling, and it worked great
on all of them.

Closes #3653.


wicket/src/ui/panes/update.rs
```diff
@@ -788,17 +780,8 @@ impl UpdatePane {
         let popup_builder =
             self.ignition.to_popup_builder(state.rack_state.selected);
 
-        // The scroll offset here should always be disabled, but make it a
-        // parameter for uniformity with the other popups.
```
jgallagher, line 792: Nice to be able to delete this :+1:

jgallagher (approved):
Sorry for the long delay on this review :grimacing:

PR #3693, opened 2023-07-18
[wicket] don't use all-caps for key shortcuts

sunshowers:
We're somewhat inconsistent on casing for key shortcuts, e.g. "UP/DOWN"
in some places and "Up/Down" in others. Settle on a consistent case
style for key shortcuts.

I think all-upper-case like "CTRL-R CTRL-R" looks too much like
shouting, so choose title case instead.


jgallagher (approved):

sunshowers:
Thanks @iliana!

PR #3694, opened 2023-07-18
ci/deploy: use lab machine disks for zpools

iliana:
This is a temporary hack to deal with the fact that we are now running /opt/oxide (which we mount as swap) out of space. This script could use some more robustness work, as we're just slightly widening our margin for CI success.


jclulow (approved):
Onwards to victory!

PR #3697, opened 2023-07-18
stop building twice for tests

davepacheco:
This change replaces the separate `cargo build` + `cargo test` steps in CI with just one `cargo test`.  On my test machine this brings total time from 50 minutes to 30 minutes.  What do we lose by doing this?  Well, we're already building all Omicron packages in other jobs.  We're also running `cargo clippy` over everything, which [runs `cargo check`](https://stackoverflow.com/questions/57449356/is-cargo-clippy-a-superset-of-cargo-check).  So I think we're still checking everything, building the binaries we actually need, and testing everything that we were testing before.

(this is not urgent if we need to free up CI slots.)


david-crespo (approved):
yes please!

iliana (approved):

PR #3706, opened 2023-07-18
ci/tuf-repo: fix dvt-dock fetch

iliana:
Adding `--depth 1` here broke clones when tools/dvt_dock_commit is not exactly equal to whatever the HEAD of the dvt-dock repo is, because I didn't ask GitHub for that commit, and also `git clone -b` doesn't work that way anyway. This awful four-liner is equivalent to what I want though, and avoids fetching the full history of the repository.

### DIFF
--- a/.github/buildomat/jobs/tuf-repo.sh
+++ b/.github/buildomat/jobs/tuf-repo.sh
@@ -125,8 +125,13 @@ EOF
 done
 
 source "$TOP/tools/dvt_dock_version"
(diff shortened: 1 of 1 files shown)


iliana:
(Also: we'll be able to tell whether this works because this branch is deliberately based on an older `main` which hasn't updated tools/dvt_dock_commit yet.)

luqmana (approved):

PR #3707, opened 2023-07-18
Avoid task cancellation on service init failure

smklein:
We have seen some behavior where, after failing to initialize services, the sled agent enters "split-brain" behavior, and on subsequent requests to initialize services, the sled agent is not aware of zones which it created itself.

I believe this to be a possible cause: While integrating #3676 , I used `try_for_each_concurrent` to parallelize zone bringup. This is an operation which can potentially return early on cancellation, dropping all other ongoing zone initialization tasks.

This PR mitigates this issue by preventing concurrent drop within service initialization.

### DIFF
--- a/sled-agent/src/services.rs
+++ b/sled-agent/src/services.rs
@@ -39,7 +39,7 @@ use ddm_admin_client::{Client as DdmAdminClient, DdmError};
 use dpd_client::{types as DpdTypes, Client as DpdClient, Error as DpdError};
 use dropshot::HandlerTaskMode;
 use flate2::bufread::GzDecoder;
-use futures::stream::{self, StreamExt, TryStreamExt};
+use futures::stream::{self, StreamExt};
 use illumos_utils::addrobj::AddrObject;
 use illumos_utils::addrobj::IPV6_LINK_LOCAL_NAME;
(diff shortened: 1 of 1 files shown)


sled-agent/src/services.rs
```diff
@@ -1919,30 +1919,49 @@ impl ServiceManager {
...
             return Err(Error::BadServiceRequest {
                 service: name,
-                message: "Should initialize zone twice".to_string(),
+                message: "Should not initialize zone twice".to_string(),
             });
         }
 
+        // We initialize all the zones we can, but only return the first error.
         let local_existing_zones = Arc::new(Mutex::new(existing_zones));
+        let first_err = Arc::new(Mutex::new(None));
```
luqmana, line 1928: Not a blocker, but as written shouldn't this be `last_err`?

luqmana (approved):
Thanks

sunshowers:
The try_ methods seem pretty difficult to use correctly 😬

sled-agent/src/services.rs
smklein: Done in https://github.com/oxidecomputer/omicron/pull/3709 , pulled it out to not block the build for folks MUPdating

PR #3710, opened 2023-07-19
rewrite check-workspace-deps as an xtask

davepacheco:
I've run into some build time issues related to workspace deps and wondered if this CI task would be a lot clearer as an xtask.  Trying that out here.

As part of this I found that we were exempting `omicron-rpaths` from the workspace dep check but I'm not sure there's a good reason for that.


davepacheco:
This isn't very important but I think this version is easier to modify (being Rust and not bash) and run locally (because it doesn't require installing anything not documented in the standard Omicron developer flow).

jordanhendricks (approved):
Looks great! Seems like a good win to me.

PR #3712, opened 2023-07-19
Sled Agent service initialization cleanup

luqmana:
Roughly 3 changes:
1. Skip initializing any zones we already know are running (both from sled-agent's perspective and as reflected via `zoneadm`)
2. Log any errors `service_put` would return on the server-side.
3. Bump file-based zpools to 15G

Lack of space was causing service initialization to fail on single-machine deployments using the `create_virtual_hardware` script. (3) should fix this.

Because of `ENOSPC` errors, we were running into some future cancellation issues which #3707 addressed. But actually determining the underlying cause was a bit difficult as `RSS` was timing out (*) on the `services_put` call and so never got back the error. (2) should hopefully make this easier to catch in the future, e.g.:
```
04:12:31.312Z ERRO SledAgent: failed to init services: Failed to install zone 'oxz_clickhouse_32ce0d6f-38fa-41e7-b699-f0af4f4f1127' from '/opt/oxide/clickhouse.tar.gz': Failed to execute zoneadm command '
Install' for zone 'oxz_clickhouse_32ce0d6f-38fa-41e7-b699-f0af4f4f1127': Failed to parse command output: exit code 1
    stdout:
    A ZFS file system has been created for this zone.
    INFO: omicron: installing zone oxz_clickhouse_32ce0d6f-38fa-41e7-b699-f0af4f4f1127 @ "/pool/ext/24b4dc87-ab46-49fb-a4b4-d361ae214c03/crypt/zone/oxz_clickhouse_32ce0d6f-38fa-41e7-b699-f0af4f4f1127"...
    INFO: omicron: replicating /usr tree...
    INFO: omicron: replicating /lib tree...
    INFO: omicron: replicating /sbin tree...
    INFO: omicron: pruning SMF manifests...
    INFO: omicron: pruning global-only files...
    INFO: omicron: unpacking baseline archive...
    INFO: omicron: unpacking image "/opt/oxide/clickhouse.tar.gz"...
    stderr:
    Error: No space left on device (os error 28)
    sled_id = 8fd66238-6bb3-4f08-89bf-f88fc2320d83

```

(*) speaking of which, do we want to increase that timeout from just 60s? it's not like any subsequent requests will work considering they'll be blocked on a lock from the initial request. and in failure cases like this the subsequent requests will timeout as well, leaving a bunch of tasks in sled-agent all waiting on the same lock to try initializing. might be worth switching to the single processing task model rack/sled initialization use.

Finally, (1) is to address the case where we try to initialize a zone that's already running. That's sometimes fine as we'll eventually realize the zone already exists. But in the case of a zone with an OPTE port, we'll run into errors like so:
```
    error_message_internal = Failed to create OPTE port for service nexus: Failure interacting with the OPTE ioctl(2) interface: command CreateXde failed: MacExists { port: "opte12", vni: Vni { inner: 100 }, mac: MacAddr { inner: A8:40:25:FF:EC:09 } }
```
This happens because the [check](https://github.com/oxidecomputer/omicron/blob/ebd3db27f6bbd08af3194cae84b0efc6b6e7248d/illumos-utils/src/zone.rs#L282C27-L282C27) for "is zone running" comes after we do all the work necessary to create the zone (e.g. creating an opte port). (1) updates `initialize_services_locked` to cross-reference sled-agent and the system's view of what zones are running so that we don't try to do the unnecessary work.

### DIFF
--- a/sled-agent/src/services.rs
+++ b/sled-agent/src/services.rs
@@ -2424,9 +2424,14 @@ impl ServiceManager {
         {
             zone_names.push(String::from(zone.name()))
         }
-        for zone in self.inner.zones.lock().await.values() {
-            zone_names.push(String::from(zone.name()));
-        }
+        zone_names.extend(
+            self.inner
+                .zones
+                .lock()
+                .await
+                .values()
+                .map(|zone| zone.name().to_string()),
+        );
         zone_names.sort();
         Ok(zone_names)
     }
--- a/sled-agent/src/services.rs
+++ b/sled-agent/src/services.rs
@@ -1923,16 +1923,17 @@ impl ServiceManager {
             });
         }
 
-        // We initialize all the zones we can, but only return the first error.
+        // We initialize all the zones we can, but only return one error, if
+        // any.
         let local_existing_zones = Arc::new(Mutex::new(existing_zones));
-        let first_err = Arc::new(Mutex::new(None));
+        let last_err = Arc::new(Mutex::new(None));
         stream::iter(requests)
             // WARNING: Do not use "try_for_each_concurrent" here -- if you do,
             // it's possible that the future will cancel other ongoing requests
             // to "initialize_zone".
             .for_each_concurrent(None, |request| {
                 let local_existing_zones = local_existing_zones.clone();
-                let first_err = first_err.clone();
+                let last_err = last_err.clone();
                 async move {
                     match self
                         .initialize_zone(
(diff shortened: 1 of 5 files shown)


sled-agent/src/services.rs
```diff
@@ -2520,6 +2521,36 @@ impl ServiceManager {
...
+                        if let Err(e) =
+                            existing_zones.remove(&name).unwrap().stop().await
+                        {
+                            error!(
+                                log,
+                                "Failed to stop zone";
+                                "zone" => &name,
+                                "error" => %e,
+                            );
+                        }
```
smklein, line 2550: Good call to do this cleanup here

smklein (approved):
Thanks for this cleanup, I appreciate this a ton. Hopefully this eliminates most developer-local issues, and makes future ones more obvious.

davepacheco:
> (*) speaking of which, do we want to increase that timeout from just 60s? it's not like any subsequent requests will work considering they'll be blocked on a lock from the initial request. and in failure cases like this the subsequent requests will timeout as well, leaving a bunch of tasks in sled-agent all waiting on the same lock to try initializing. might be worth switching to the single processing task model rack/sled initialization use.

I think so.  I'm not sure it ever makes sense to have a request timeout when using TCP keep-alive.  We removed a similar one in #3503.

PR #3715, opened 2023-07-19
validate that on-disk schema's aren't changing in surprising ways

ahl:


smklein (approved):
This is great, exactly what I had in mind. Thank you!

david-crespo (approved):

common/src/nexus_config.rs
```diff
@@ -123,15 +125,17 @@ pub enum InternalDns {
...
 
-#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize, JsonSchema)]
 pub struct DeploymentConfig {
     /// Uuid of the Nexus instance
     pub id: Uuid,
     /// Uuid of the Rack where Nexus is executing.
     pub rack_id: Uuid,
     /// Dropshot configuration for the external API server.
+    #[schemars(skip)] // TODO we're protected against dropshot changes
```
smklein, line 135: Can we expand on this? I thought of this struct as a contract between Sled Agent and Nexus -- if they use different versions of this config (lets suppose we upgrade to dropshot 2.0), wouldn't we have a skew?

smklein (approved):

common/src/nexus_config.rs
luqmana: I had assumed the update boundary was effectively per-sled, i.e., all the software on a given sled gets updated at once. In this case, it's Sled Agent generating a `DeploymentConfig` when it launching Nexus. So they should always agree unless we've somehow generated mismatched artifacts?

davepacheco: I think the medium (short?) term intent is that you'd be able to update various control plane components separately.  Sled Agent _is_ coupled to the OS image, but I don't think we want you to have to bounce the sled to update Nexus (and also update several other components at the same time).

PR #3717, opened 2023-07-19
[wicket] mark skipped components as SKIPPED

sunshowers:
Just check that the "UpdateComponent" step was skipped.

To test this, I had to add two facilities:

1. Detecting and storing the currently active slot in sp-sim.
2. New environment variables for simulating SP and RoT results.

### DIFF
--- a/sp-sim/src/gimlet.rs
+++ b/sp-sim/src/gimlet.rs
@@ -4,6 +4,8 @@
 
 use crate::config::GimletConfig;
 use crate::config::SpComponentConfig;
+use crate::helpers::rot_slot_id_from_u16;
+use crate::helpers::rot_slot_id_to_u16;
 use crate::rot::RotSprocketExt;
 use crate::serial_number_padded;
 use crate::server;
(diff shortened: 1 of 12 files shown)


sp-sim/src/gimlet.rs
```diff
@@ -1107,12 +1115,18 @@ impl SpHandler for Handler {
...
+            &self.log, "asked for component active slot";
             "sender" => %sender,
             "port" => ?port,
             "component" => ?component,
         );
-        Err(SpError::RequestUnsupportedForComponent)
+        if component == SpComponent::ROT {
+            Ok(rot_slot_id_to_u16(self.rot_active_slot))
+        } else if component == SpComponent::SP_ITSELF {
+            Ok(rot_slot_id_to_u16(self.sp_active_slot))
```
jgallagher, line 1126: The real SP returns `RequestUnsupportedForComponent` for anything other than the RoT for both this and `component_set_active_slot` - I think we could drop `self.sp_active_slot` and keep that same behavior here.

sp-sim/src/sidecar.rs
```diff
@@ -297,6 +299,10 @@ struct Handler {
 
     serial_number: String,
     ignition: FakeIgnition,
+    // Use RotSlotId for both RoT and SP slots since they both have two elements
+    // each.
+    rot_active_slot: RotSlotId,
+    sp_active_slot: RotSlotId,
```
jgallagher, line 305: (Same comment as above w.r.t. dropping `sp_active_slot`)

jgallagher (approved):

PR #3729, opened 2023-07-20
Load early network setup from bootstore

andrewjstone:


sled-agent/src/services.rs
```diff
@@ -512,21 +512,35 @@ impl ServiceManager {
...
             info!(log, "Loading services - No services detected");
             return Ok(());
         };
-        let services = ledger.data();
+        let services = ledger.data_mut();
+
+        // Inject boundary_switch_addrs into ServiceZoneRequests
+        // When the opte interface is created for the service,
+        // nat entries will be created using the switches present here
+        //let switch_addrs: Vec<Ipv6Addr> = Vec::from_iter(boundary_switch_addrs);
```
smklein, line 537: Dead code?

sled-agent/src/services.rs
```diff
@@ -512,21 +512,35 @@ impl ServiceManager {
...
+        &self,
+        boundary_switch_addrs: HashSet<Ipv6Addr>,
+    ) -> Result<(), Error> {
         let log = &self.inner.log;
         let ledger_paths = self.all_service_ledgers().await;
         info!(log, "Loading services from: {ledger_paths:?}");
 
         let mut existing_zones = self.inner.zones.lock().await;
-        let Some(ledger) = Ledger::<AllZoneRequests>::new(
+        let Some(mut ledger) = Ledger::<AllZoneRequests>::new(
```
smklein, line 524: I'm kinda confused how this works when used in the `services_put` API.

Suppose Nexus calls that endpoint with a set of services:

https://github.com/oxidecomputer/omicron/blob/fa3b74bee6033261c9f83c1a5cdf7e1e79c4b332/sled-agent/src/sled_agent.rs#L611-L639

When we're in the body of that function:

https://github.com/oxidecomputer/omicron/blob/fa3b74bee6033261c9f83c1a5cdf7e1e79c4b332/sled-agent/src/services.rs#L2445

It seems like we would be missing this critical boundary services information, no?

smklein (commented):
Mostly looks great, with just one small (but critical?) question around "the parameters used when telling the ServiceManager to run some services".

sled-agent/src/services.rs
andrewjstone: Yeah good catch. This definitely does not work as written. I'm still not super familiar with all the pieces here, but it looks like this is the only other place we persist `AllZoneRequests` to a ledger.

andrewjstone: I believe I resolved this in [629065b](https://github.com/oxidecomputer/omicron/pull/3729/commits/629065b70460ed497b4661158de9f8492445c1bb)

sled-agent/src/services.rs
andrewjstone: Vamoose!

andrewjstone:
Closing in favor of #3765

PR #3730, opened 2023-07-20
Use RSS provided DNS servers for resolving external hosts in Nexus

### LINKED ISSUE #3724: Nexus needs to be able to resolve external hosts when configuring an IdP if grabbing metadata via URL.
This regressed probably when services got moved to their own zones instead of just running directly in the global zone 😬. We've just been using the inline base64-encoded variant in dogfood and so hadn't noticed.

cc https://github.com/oxidecomputer/omicron/issues/2122

luqmana:
Fixes #3724

Side-steps the question of what to populate in `/etc/resolv.conf` for Nexus. I think being explicit where we need to do any external name resolution (which only consists of grabbing IdP metadata via URL today AFAIK) seems better than relying on ambient system config.

Most of the changes are just plumbing up the RSS provided external DNS servers up to Nexus and some `String`->`IpAddr` changes. Longer term we'd probably want to persist those to the DB and have some API to update them (https://github.com/oxidecomputer/omicron/issues/3732).

### DIFF
--- a/openapi/sled-agent.json
+++ b/openapi/sled-agent.json
@@ -1917,7 +1917,8 @@
                 "description": "External DNS servers Nexus can use to resolve external hosts.",
                 "type": "array",
                 "items": {
-                  "type": "string"
+                  "type": "string",
+                  "format": "ip"
                 }
               },
               "external_ip": {
@@ -2070,7 +2071,8 @@
               "dns_servers": {
                 "type": "array",
                 "items": {
-                  "type": "string"
+                  "type": "string",
+                  "format": "ip"
                 }
               },
               "domain": {
@@ -2124,7 +2126,8 @@
               "dns_servers": {
                 "type": "array",
                 "items": {
-                  "type": "string"
+                  "type": "string",
+                  "format": "ip"
                 }
               },
               "domain": {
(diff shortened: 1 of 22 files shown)


openapi/sled-agent.json
```diff
@@ -1917,7 +1917,8 @@
                 "description": "External DNS servers Nexus can use to resolve external hosts.",
                 "type": "array",
                 "items": {
-                  "type": "string"
+                  "type": "string",
```
davepacheco, line 1920: Just checking: this is not a backwards-incompatible change, is it?

davepacheco (approved):
I think it's a good call to be clear about where we're using external DNS and not rely on /etc/resolv.conf.

openapi/sled-agent.json
luqmana: You're right, it wouldn't just work on an existing install without updating the corresponding nexus service config on whatever sled `/pool/int/*/config/services.toml`.

davepacheco: Oh, that's worse than I realized.  Is this...okay?

luqmana: I had assumed we're still at the point of requiring a clean slate on mupdate? (especially given in-flight work like trust quorum, early networking, etc)

smklein: Yes, for a few more days, but probably not by the end of next week

PR #3731, opened 2023-07-21
[installinator] use cancel_safe_futures::join_then_try!

sunshowers:
See
https://oxidecomputer.github.io/cancel-safe-futures/rustdoc/cancel_safe_futures/macro.join_then_try.html
for more. This is equivalent to the existing code but cleaner to read.


smklein (approved):
Looks straightforward, I'm eager to use this in more spots!

PR #3733, opened 2023-07-21
add more networking docs

davepacheco:
This doc seeks to give an overview of how networking works in the Oxide rack and especially Omicron.  I'm sure it's incomplete and wrong in places (please correct it if you find errors!) but it's hopefully a useful start.  I'd love feedback of any kind.

Thanks to @bnaecker, @jordanhendricks, and @leftwo.  We met a few weeks ago to go over a lot of this and I'm just now putting this together.


docs/networking.adoc
```diff
@@ -0,0 +1,766 @@
...
+    cxgbe1-- "backplane"  --- Sidecar1
+    cxgbe0_ip["cxgbe0 (IP, GZ)\naddresses:\nLL: fe80::..."]
+    cxgbe1_ip["cxgbe1 (IP, GZ)\naddresses:\nLL: fe80::..."]
+    cxgbe0_ip --- cxgbe0
+    cxgbe1_ip --- cxgbe1
+
+    underlay0["underlay0 (VNIC, GZ)"]
+    underlay0_ip["underlay0 (IP, GZ)\naddresses:\nLL: fe80::..."]
+
+    subgraph cpzone["Every control plane zone\n(Nexus, CockroachDB,\nDNS, etc. and oxz_switch)"]
```
rmustacc, line 439: The bit in parens is cut off in the github rendering, fwiw.

docs/networking.adoc
```diff
@@ -0,0 +1,766 @@
...
+fd00:1122:3344:1::/64       fe80::aa40:25ff:fe05:c      UG      2    2437 cxgbe0
+fd00:1122:3344:2::/64       fe80::aa40:25ff:fe05:40c    UG      2      38 cxgbe1
+fd00:1122:3344:2::/64       fe80::aa40:25ff:fe05:c      UG      2    2437 cxgbe0
+fd00:1122:3344:3::/64       fe80::aa40:25ff:fe05:40c    UG      2   13358 cxgbe1
+fd00:1122:3344:3::/64       fe80::aa40:25ff:fe05:c      UG      2    2437 cxgbe0
+----
+
+Recall that cxgbe0 and cxgbe1 are connected to separate switches in the rack.  So we're seeing the prefixes for the other sleds in this deployment.  We have two routes to reach each sled: one through each switch.  The gateway is the link-local address _of each switch_ on the corresponding link.  One notable exception: the route for this same sled (`fd00:1122:3344:104::/64`) points to `underlay0`, the GZ's VNIC on the sled's underlay network.  In this way, traffic leaving the GZ (whether it originated in this GZ or arrived from one of the switches) is directed to the sled's underlay network etherstub and from there to the right zone VNIC.
+
+(Questions: Why does 107 only have one route?  What are the `fd00:99::` routes?)
```
rmustacc, line 586: I think the fd00:99 is a holdover from how we were reaching boundary services for a bit. This is being overhauled per [RFD 404 Boundary Tunnel Routing](https://rfd.shared.oxide.computer/rfd/0404) so it may have something to do with what's provisioned or otherwise. I'm not sure. @rcgoodfellow may have a better idea here.

docs/networking.adoc
```diff
@@ -0,0 +1,766 @@
...
+* through the etherstub's virtual switch to the Oximeter zone's VNIC
+
+=== Data path on the bootstrap network
+
+Once set up, the bootstrap network works largely the same way as the underlay network:
+
+* each sled has an etherstub that implements the per-sled bootstrap network
+* each sled in the rack has two routes to each other sled: one through each switch
+
+It's simpler than the underlay network because only global zones have addresses on the bootstrap network.  None of the per-zone datalinks, routing tables, and netstack stuff from the underlay network applies to the bootstrap network.
```
rmustacc, line 605: I believe the switch zones are on the bootstrap network. It's also simpler because the bootstrap network will never leave the rack.

docs/networking.adoc
```diff
@@ -0,0 +1,766 @@
...
+opte3                            A8:40:25:F5:B0:49        172.30.0.37      172.20.26.34     None                                     None                                     running
+opte4                            A8:40:25:F2:52:4B        172.30.0.9       172.20.26.35     None                                     None                                     running
+opte5                            A8:40:25:F7:F7:4A        172.30.0.39      172.20.26.37     None                                     None                                     running
+opte6                            A8:40:25:F7:19:71        172.30.0.41      172.20.26.39     None                                     None                                     running
+opte7                            A8:40:25:F2:BB:4B        192.168.0.17     172.20.26.45     None                                     None                                     running
+----
+
+Recall that OPTE is part of boundary services.  This command lists key configuration: it says that external IP 172.20.26.4 maps to an internal IPv4 address 172.30.2.6 on `opte0`.  This corresponds with `vopte0` in the Nexus zone (because it's the same-numbered device name).
```
rmustacc, line 676: This is the first mention of OPTE as a named component. Should we define it earlier or at least spell out that it's the Oxide Packet Transformation Engine?

I'm not sure I'd characterize OPTE as entirely part of boundary services. I think of it as its own component, because it's also here for all VM-VM communication. Dunno.

rmustacc (commented):
Thanks for putting this together. I think this'll be great to have @davepacheco. I've put some initial notes together here.

docs/networking.adoc
```diff
@@ -0,0 +1,766 @@
...
+|Yes
+|Yes (for external API/console)
+
+|External DNS
+|No
+|No
+|Yes
+|Yes (for customer-facing DNS)
+
+|NTP (Boundary)
```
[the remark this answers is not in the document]
davepacheco: Every sled gets an NTP zone.  Only a couple of them are designated as "boundary" NTP zones.  As I understand it, the boundary ones are pointed at NTP servers external to the rack (which is why they need to be able to NAT out).  The "internal" ones are pointed at the boundary NTP servers (and so don't need any external connectivity).

The elephant in the room here is that (I think) there's no overarching documentation on components in the control plane yet.  I've referred to RFD 61 for this, but it doesn't include everything here.

docs/networking.adoc
davepacheco: OPTE does show up earlier, in the bullet list at the start of this section.

I didn't mean to imply that OPTE was entirely within boundary services.  I'll tweak this to hopefully make that clearer.

docs/networking.adoc
```diff
@@ -0,0 +1,766 @@
...
+* `fc00::/7`: https://en.wikipedia.org/wiki/Unique_local_address[Unique local address (ULA)], "somewhat analogous to IPv4 private network addressing" (e.g., `10.0.0.0/8`).
+** `fd00::/8`: the part of the ULA that's _not_ centrally assigned.  The control plane primarily uses addresses in this range.
+
+Because link-local addresses are only valid for the network that the link is attached to, whenever you use a link-local address, you usually have to specify which link's network you mean.  You might see `fe80::aa40:25ff:fe04:3d5%cxgbe0`, where `fe80::aa40:25ff:fe04:3d5` is the link-local address and `cxgbe0` is the **scope_id** (denoting "the network attached to cxgbe0").  We'll look at an example below.
+
+== Inspecting networking configuration on illumos
+
+illumos supports lightweight OS-based virtualization in the form of **zones**.  Each system has a **global zone** (GZ) that contains all other **non-global zones.**  Each zone looks like its own separate illumos system with its own root filesystem and (importantly for us) its own netstack.  It has its own set of datalinks, IP interfaces, addresses, etc.
```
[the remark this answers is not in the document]
davepacheco: Thanks -- clarified the text here.

docs/networking.adoc
```diff
@@ -0,0 +1,766 @@
...
+underlay0   vnic      9000   up       --         underlay_stub0
+----
+
+We can see:
+
+* Three `phys` links that correspond to physical devices.  We're interested in `cxgbe0` and `cxgbe1`, which correspond to the two ports provided by the Chelsio NIC on Gimlets.
+* Two `etherstub` devices, `bootstrap_stub0` and `underlay_stub0`.  Etherstubs function like a synthetic Ethernet network.  We can create VNICs over an etherstub and those VNICs will act as though connected through a (virtual) switch.
+* Two `vnic` devices.  VNICs function like virtualized network links created _over_ some other link.  For example, we could create two VNICs over `cxgbe0`.  Each would have its own MAC address and IP addresses and would be functionally like having another device on the same network that `cxgbe0` is plugged into.  As we'll see, a single system might have a bunch of zones, each with an Omicron component running in it, and each having a VNIC on the underlay network.
+
+We'll discuss these in more detail later.  There are also many other kinds of links.  See the manual page for more details.
```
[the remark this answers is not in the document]
davepacheco: I expanded this a (tiny) bit.

docs/networking.adoc
```diff
@@ -0,0 +1,766 @@
...
+bootstrap_stub0 etherstub 9000 up     --         --
+bootstrap0  vnic      1500   up       --         bootstrap_stub0
+underlay_stub0 etherstub 9000 up      --         --
+underlay0   vnic      9000   up       --         underlay_stub0
+----
+
+We can see:
+
+* Three `phys` links that correspond to physical devices.  We're interested in `cxgbe0` and `cxgbe1`, which correspond to the two ports provided by the Chelsio NIC on Gimlets.
+* Two `etherstub` devices, `bootstrap_stub0` and `underlay_stub0`.  Etherstubs function like a synthetic Ethernet network.  We can create VNICs over an etherstub and those VNICs will act as though connected through a (virtual) switch.
```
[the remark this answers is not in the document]
davepacheco: Makes sense.  I'm not sure a better way to explain this, though.  Feel free to take a stab!

docs/networking.adoc
```diff
@@ -0,0 +1,766 @@
...
+** `cxgbe0`: the IP interface name, which is also the datalink name
+** `ll`: names this specific address on this IP interface.  The software that configures networking can call this whatever it wants.  `ll` is link-local.  Some software configures `v4` or `v6` (as on the loopback interface).  Other software uses the name of the network (as on `bootstrap0`).
+* `ADDR`: `fe80::aa40:25ff:fe04:3d5%cxgbe0/10`
+** `fe80::aa40:25ff:fe04:3d5`: IPv6 address in the link-local space (because it starts with `fe80::` -- see above)
+** `%cxgbe0`: scope_id is `cxgbe0`.  Because this is a link-local address, it needs a scope_id to specify which network it's on.  A scope_id of `cxgbe0` means "the network attached to cxgbe0".  That seems obvious in this context, but it's needed in other contexts where you want to use the address (e.g., `ssh fe80::aa40:25ff:fe04:3d5%cxgbe0`).
+** `/10`: As with IPv4, addresses are often written with the netmask.  Here, the first 10 bits of the address identify the network.
+
+The other big piece of networking configuration is the _routing table_, which you can see using `netstat -rn`:
```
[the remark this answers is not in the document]
davepacheco: Done.

docs/networking.adoc
```diff
@@ -0,0 +1,766 @@
...
+
+* the **underlay network**.  This is the main network used by Omicron components like Nexus, internal/external DNS, CockroachDB, Sled Agents, Dendrite, and Management Gateway Service to talk to each other.  (See <<rfd61>>.)  Customer traffic is also encapsulated and sent between components on this network.
+* the **bootstrap** network.  This network is used by bootstrap agents (part of the sled agent) to reach other bootstrap agents so that sleds can discover each other and form a trust quorum.  Once the trust quorum is established, sleds can unlock encrypted local storage and start the rest of the control plane, which enables the underlay network to be configured.
+* the **management network**.  This is the network that service processors (SPs) are connected to.  It's constrained, in part because the devices on it have limited capabilities (e.g., limited throughput, generally no TCP, etc.) and also because those devices are critical to the safety of the system (like keeping sleds from melting).  The only control plane component on this network is the Management Gateway Service (MGS).  See <<rfd210>> for more on MGS and the management network.
```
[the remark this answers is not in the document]
davepacheco: I added the second bit about the techports.

I don't understand the first bit enough to document it.  At one point I thought they were separate L2 segments, but there was chatter recently about that _not_ being true (and it being a problem).  I also think it's not critical to include here.

docs/networking.adoc
davepacheco: Yeah...I just removed this.

docs/networking.adoc
davepacheco: I realize I didn't even talk about cross-rack connectivity on the underlay network, but I think that's okay given where we are and where this doc is.

docs/networking.adoc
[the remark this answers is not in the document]
luqmana: Also, I only realized this later on, but `etherstub`s are actually just `VNIC`s without an underlying link

docs/networking.adoc
luqmana: While 404 is in flux, I can take a stab at what the current state looks like:

`fd00:99::1/64` is the boundary services address. The small twist there is that there's no single "boundary services" node present on the network but rather, OPTE and the P4 on the switch essentially have a contract in how external traffic gets transited.

If you query the switch, it should appear as a `loopback` address, i.e. it's not associated with any backplane or front panel links but it is an address the switch is aware of:
```
$ pfexec zlogin oxz_switch /opt/oxide/dendrite/bin/swadm addr ls
Link      IPv4         IPv6
qsfp0/0   172.17.16.2  
loopback  fd00:99::1
```

If we look at an OPTE port for an instance or service with external connectivity, we can see all the layers that process incoming/outgoing packets:
```
$ pfexec opteadm list-layers -p opte7
NAME         RULES IN   RULES OUT  DEF IN   DEF OUT  FLOWS     
gateway      1          5          deny     deny     0         
firewall     1          0          deny     stateful allow 0         
router       0          2          allow    deny     0         
nat          1          1          allow    allow    0         
overlay      1          1          deny     deny     0         
```
Outgoing packets are processed by the layers going top-to-bottom and incoming packets go through the reverse. That bottom layer, `overlay`, is where the GENEVE encap/decap happens.

**Egress:**

Outgoing packets are wrapped with outer ETH/IPv6/UDP/GENEVE headers. The outer IPv6 source is the sled's underlay address & the destination is the boundary services address `fd00:99::1`. The UDP destination port is always 6081 (the IANA assigned port for Geneve) and the source port we just always set to 7777 (the Zezeski special 😅️). We also stash a VPC-unique VNI into the Geneve header. The outer ethernet frame addresses get populated based on the entries in the illumos routing table. That's what the entries for `fd00:99::1` on both `cxgeb0` & `cxgeb1` are for. AFAIK, we should be choosing one probabilistically using DDM (Delay Driven Multipath). Either way, we'll set the destination address to a gateway on one of the chelsios and use its MAC as the source address.

Once the packet gets to the switch, it recognizes it as a boundary services destined geneve encapsulated payload and will strip the outer headers. From there, it'll use the uplink configuration like routing and available links to send it out to the customer network as appropriate.

**Ingress:**

For incoming packets from the customer network, they'll hit the switch which will check it's NAT table to determine which sled to forward the packet to:
```
 pfexec zlogin oxz_switch /opt/oxide/dendrite/bin/swadm nat list
External IP   Port low  Port high  Internal IP            Inner MAC          VNI
172.17.16.11  0         65535      fd00:1122:3344:101::1  a8:40:25:ff:ef:ce  100
172.17.16.12  0         65535      fd00:1122:3344:101::1  a8:40:25:ff:db:ff  100
172.17.16.13  0         65535      fd00:1122:3344:101::1  a8:40:25:ff:b2:cc  100
172.17.16.14  0         65535      fd00:1122:3344:101::1  a8:40:25:ff:be:a1  100
172.17.16.15  0         16383      fd00:1122:3344:101::1  a8:40:25:ff:e2:15  100
```
If there's an entry that matches the destination IP and port, it'll similarly add outer ETH/IPv6/UDP/GENEVE headers. The outer IPv6 destination will be the "Internal IP" from the NAT entry which represents a given sled's underlay address (the sled hosting that instance/service). The outer ETH destination will be its corresponding value from the switch's NDP table. Similarly to the egress path, the VNI will also be included in the Geneve header. But that's not enough to identify which specific OPTE port on the sled should receive the packet, so it'll also update the _inner_ ETH destination to the "Inner MAC" from the NAT entry. We can see the MAC on the OPTE ports should match the NAT entries:
```
 pfexec opteadm list-ports                                                                                                                    
LINK                             MAC ADDRESS              IPv4 ADDRESS     EXTERNAL IPv4    IPv6 ADDRESS                             EXTERNAL IPv6                            STATE   
opte0                            A8:40:25:FF:E2:15        172.30.3.5       None             None                                     None                                     running 
opte1                            A8:40:25:FF:EF:CE        172.30.1.5       172.17.16.11     None                                     None                                     running 
opte5                            A8:40:25:FF:BE:A1        172.30.2.7       172.17.16.14     None                                     None                                     running 
opte6                            A8:40:25:FF:B2:CC        172.30.2.6       172.17.16.13     None                                     None                                     running 
opte7                            A8:40:25:FF:DB:FF        172.30.2.5       172.17.16.12     None                                     None                                     running 
```
At this point we'll hit the `overlay` in OPTE as an inbound packet this time where those outer headers will be stripped and further NAT, firewall rules get applied before being delivered to the instance/service's vnic.

docs/networking.adoc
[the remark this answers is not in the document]
rmustacc: Maybe adding something to the VNIC section like:

When you create a vnic on top of a device, it implicitly is creating an internal (virtual) L2 switch. Put another way creating a VNIC, whether on top of a physical device, etherstub, or other types of data links, can be imagined as plugging the VNICs into an unmanaged switch with the other device as the uplink. When traffic is not destined for something within the switch it will always go out the uplink device.

docs/networking.adoc
```diff
@@ -118,6 +116,8 @@ fd00:1122:3344:105::/64     fe80::aa40:25ff:fe05:41c    UG      1       0 cxgbe1
 
 We'll talk about this more later.
 
+Finally, you can use the portable `ifconfig` tool to list much of the same information.  We'll be sticking to the illumos tools.
```
rmustacc, line 119:
```suggestion
Finally, if you're used to using `ifconfig` to look at this information, you can do that on illumos too. However, we prefer using `ipadm` for managing this and stick to it throughout this document.
```

I put this here mostly because ifconfig is only kind of portable and I still think of it as an illumos tool. Parts of it are the same between systems, but not everything. Dunno. Feel free to take or leave.

docs/networking.adoc
```diff
@@ -126,87 +126,55 @@ We talk about a few major networks in a single-rack Oxide system:
...
-|No
-
-|Dendrite
-|No
-|No
-|Yes
-|No
-
-|===
+```mermaid
```
rmustacc, line 132: If it doesn't make it too ugly I might consider putting an explicit box of customer systems to the right of `external network`.

docs/networking.adoc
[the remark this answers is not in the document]
rmustacc: That makes sense now. I'm not sure why I got confused there.

rmustacc (approved):
Feel free to take or merge at your druthers. Thanks for putting this together.

docs/networking.adoc
davepacheco: @luqmana This is very helpful though I didn't follow all of it.  I'm not sure how best to incorporate it.  Do you want to make a PR on this PR, or just push to this PR, or add this in a follow-up PR?

luqmana: > @luqmana This is very helpful though I didn't follow all of it. I'm not sure how best to incorporate it. Do you want to make a PR on this PR, or just push to this PR, or add this in a follow-up PR?

Also, please call out any bits that weren't clear enough!

docs/networking.adoc
davepacheco: Good idea.  Done.

docs/networking.adoc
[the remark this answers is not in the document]
davepacheco: After some discussion offline with @rmustacc I understand this better (along with @luqmana's comment!).  I've expanded this section a bit and added some diagrams based on an ASCII diagram from Robert.  It might be a little distracting given where it is in the flow, but I think it's pretty valuable information.

davepacheco:
I'd love a +1 on the new VNIC/etherstub section from somebody who knows.  But I've got enough confidence (and have taken enough of folks' time on this) that I'll just land it some time tomorrow if I haven't heard back.

rmustacc (approved):
Thanks for folding that all in. Looks good.

PR #3736, opened 2023-07-21
disable RIP for general deployments

internet-diglett:
RIP's method of gratuitous advertisements can accidentally create routing loops or blackhole traffic when used in conjunction with DDM.

### DIFF
--- a/smf/switch_zone_setup/switch_zone_setup
+++ b/smf/switch_zone_setup/switch_zone_setup
@@ -37,8 +37,4 @@ for i in "${!USERS[@]}"; do
     fi
 done
 
-# enable ipv6 forwarding and routing
-pfexec routeadm -e ipv6-forwarding -u
-pfexec routeadm -e ipv6-routing -u
-
 exit $SMF_EXIT_OK


luqmana:
does this impact the transit-mode maghemite use-case for non-gimlet dev? cc @jmpesp

internet-diglett:
@luqmana I actually had to turn this off in order to get transit-mode to work correctly on non-gimlet hardware for a multi-switch topology.

smf/switch_zone_setup/switch_zone_setup
```diff
@@ -37,8 +37,4 @@ for i in "${!USERS[@]}"; do
     fi
 done
 
-# enable ipv6 forwarding and routing
-pfexec routeadm -e ipv6-forwarding -u
-pfexec routeadm -e ipv6-routing -u
```
jgallagher, line 42: Just double checking - when we were checking a system live last week, @rcgoodfellow had us disable `ipv6-routing` but _not_ `ipv6-forwarding`. We do want to disable both, though?

rcgoodfellow: I believe disabling both is the correct thing to do in the switch zone. But we'll definitely want laps on dogfood. However, it is critically important that the GZ have `ipv6-forwarding` enabled.

jmpesp:
> does this impact the transit-mode maghemite use-case for non-gimlet dev? cc @jmpesp

Don't worry about me, I can deal - do the right thing for the real deployment :)

rcgoodfellow (approved):

luqmana (approved):

jclulow:
Do we know why we were turning `ipv6-routing` on in the first place?  I was worried that it was somehow on _by default_, until I saw the diff here.

rcgoodfellow:
IIRC this was done in early dev so the switch zone would advertise prefixes to the GZ - but this has not been required in the implementation for quite some time.

PR #3737, opened 2023-07-21
[sled-agent] improve initialize_services_locked

sunshowers:
I started off using cancel-safe-futures but realized that it would be
better to return errors from all failed init requests.

This cleans up this code nicely, removing a couple of locks.

### DIFF
--- a/sled-agent/src/services.rs
+++ b/sled-agent/src/services.rs
@@ -1946,8 +1946,6 @@ impl ServiceManager {
             });
         }
 
-        // We initialize all the zones we can, but only return one error, if
-        // any.
         let futures = requests.iter().map(|request| {
             async move {
                 self.initialize_zone(
@@ -1960,8 +1958,8 @@ impl ServiceManager {
             }
         });
         let results = futures::future::join_all(futures).await;
+
         let mut errors = Vec::new();
-        // TODO: maybe merge and return errors?
         for result in results {
             match result {
                 Ok(zone) => {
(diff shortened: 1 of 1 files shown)


sled-agent/src/services.rs
```diff
@@ -1931,43 +1948,33 @@ impl ServiceManager {
...
+                    // filesystems=
+                    &[],
+                )
+                .await
+                .map_err(|error| (request.zone.zone_name(), error))
+            }
+        });
+        let results = futures::future::join_all(futures).await;
+        let mut errors = Vec::new();
+        // TODO: maybe merge and return errors?
```
smklein, line 1964: Isn't this basically what you're doing now?

smklein (approved):
Thanks!

sled-agent/src/services.rs
```diff
@@ -1931,43 +1948,33 @@ impl ServiceManager {
 
         // We initialize all the zones we can, but only return one error, if
```
luqmana, line 1949: Can update the comment now that all errors get returned.

luqmana (approved):

sled-agent/src/services.rs
sunshowers: Yes haha, this is what happens when you're working till 3am

sled-agent/src/services.rs
sunshowers: Thanks!

PR #3740, opened 2023-07-21
Update dendrite and maghemite auto-prs

### LINKED ISSUE #3259: do a daily sync of maghemite and dendrite in an automated PR
It would be useful to have automatic updates of dendrite and maghemite along the lines of this sort of PR:

- https://github.com/oxidecomputer/omicron/pull/2941
- https://github.com/oxidecomputer/omicron/pull/2197

I'd suggest we...
1. do this daily
2. do it at a time when CI use is relatively light (3am PT?)

cc @augustuswm who built some similar-ish logic in the https://github.com/oxidecomputer/oxide.rs repo
cc @rcgoodfellow, @Nieuwejaar 

Note would probably be useful for propolis and crucible as well, but perhaps we can start here and figure out the other dependencies.

augustuswm:
Updates the creation / update mechanism that is used for creating dendrite and maghemite update PRs. Previously only GitHub `push` events could be used as a notification source. This caused an issue as the artifacts necessary for updating were still in the process of being built by a buildomat job. To work around this a long (1 hour) debounce delay was used so that by the time PR creation logic ran, the artifacts should have been generated.

Now that `check_run` events are available, they can be used to more accurately specify when PR creation should run. This update changes the dendrite and maghemite listeners to listen for the completion of the `image` buildomat job in their respective repos.

Resolves #3259

### DIFF
--- a/.github/reflector/update-dendrite.toml
+++ b/.github/reflector/update-dendrite.toml
@@ -9,14 +9,10 @@ branch = "main"
 
 requires_token = true
 
-# Debounce values are used to batch together multiple pushes that occur in a relatively short
-# timeframe
+# Run in response to the image buildomat job in dendrite completing
 
 [[on]]
 repository = "oxidecomputer/dendrite"
-paths = [
-  "**",
-]
-
-# Wait one hour for binary builds to complete
-debounce = 3600
+event = "check_run"
+action = "completed"
+check = "image"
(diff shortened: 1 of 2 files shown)


rcgoodfellow (approved):
LGTM. Thanks!

Looks like the CI failure is not related to this change.

augustuswm:
Great thanks. I'll get this merged late tonight when CI is quiet.

PR #3745, opened 2023-07-22
Configure log rotation in all non-global zones

citrus-it:


lifning (approved):
looks good! installed omicron on my machine and verified the new logadm rules are in place in all zones.

PR #3754, opened 2023-07-24
Wait for Crucible agent to do requested work

### LINKED ISSUE #3698: Volume deletion should occur in a background task instead of a saga
It's currently possible for Nexus / CRDB to be overwhelmed with a deluge of disk delete operations, and for those operations to fail. The current `decrease_crucible_resource_count_and_soft_delete_volume` transaction is large and too many invocations causes some transactions to fail. This overload can lead to cases where the unwind of the disk delete saga also fails, and this eventually leaves all regions allocated but unusable.

A better fit would be a mark and sweep pattern: volumes can be marked via a soft-delete by any part of Nexus, and the sweep step would be performed by a background task. The pattern of "constant work" described in the block comment for `nexus/src/app/background/common.rs` would prevent Nexus / CRDB from being overwhelmed.

jmpesp:
Several endpoints of the Crucible agent do not perform work synchronously with that endpoint's handler: changes are instead requested and the actual work proceeds asynchronously. Nexus assumes that the work is synchronous though.

This commit creates three functions in common_storage that properly deal with an agent that does asynchronous work (`delete_crucible_region`, `delete_crucible_snapshot`, and `delete_crucible_running_snapshot`), and calls those.

Part of testing this commit was creating "disk" antagonists in the omicron-stress tool. This uncovered cases where the transactions related to disk creation and deletion failed to complete. One of these cases is in `regions_hard_delete` - this transaction is now retried until it succeeds. The `TransactionError::retry_transaction` function can be used to see if CRDB is telling Nexus to retry the transaction. Another is `decrease_crucible_resource_count_and_soft_delete_volume` - this was broken into two steps, the first of which enumerates the read-only resources to delete (because those don't currently change!).

Another fix that's part of this commit, exposed by the disk antagonist: it should only be ok to delete a disk in state `Creating` if you're in the disk create saga. If this is not true, it's possible for a delete of a disk currently in the disk create saga to cause that saga to fail unwinding and remain stuck.

This commit also bundles idempotency related fixes for the simulated Crucible agent, as these were exposed with the retries that are performed by the new `delete_crucible_*` functions.

What shook out of this is that `Nexus::volume_delete` was removed, as this was not correct to call during the unwind of the disk and snapshot create sagas: it would conflict with what those sagas were doing as part of their unwind. The volume delete saga is still required during disk and snapshot deletion, but there is enough context during the disk and snapshot create sagas to properly unwind a volume, even in the case of read-only parents. `Nexus::volume_delete` ultimately isn't safe, so it was removed: instead, any future sagas should either embed the volume delete saga as a sub saga, or there should be enough context in the outputs said future saga's nodes to properly unwind what was created.

Depends on https://github.com/oxidecomputer/crucible/pull/838, which fixes a few non-idempotency bugs in the actual Crucible agent.

Fixes #3698


nexus/db-queries/src/db/datastore/region.rs
```diff
@@ -159,67 +162,95 @@ impl DataStore {
...
+                        }
+
+                        Ok(())
+                    })
+                    .await
+                    .map_err(|e: TxnError| {
+                        if e.retry_transaction() {
+                            BackoffError::transient(Error::internal_error(
+                                &format!("Retryable transaction error {:?}", e)
+                            ))
```
smklein, line 233: This might be more of a philosophical question, but....

... should we use this pattern for all transactions? I struggle to understand why **this specific transaction** should be treated differently from others.

nexus/db-queries/src/db/datastore/volume.rs
```diff
@@ -45,6 +45,26 @@ impl DataStore {
...
+            let mut crucible_targets = CrucibleTargets::default();
+            read_only_resources_associated_with_volume(
+                &vcr,
+                &mut crucible_targets,
+            );
+            crucible_targets
+        };
+
         self.pool()
             .transaction(move |conn| {
```
smklein, line 69: Not needed for this PR, but this could also likely be converted into a CTE, if we want.

nexus/db-queries/src/db/datastore/volume.rs
```diff
@@ -146,24 +147,44 @@ impl DataStore {
...
+                    .filter(dsl::id.eq(volume_id))
+                    .select(Volume::as_select())
+                    .first::<Volume>(conn)
+                    .optional()?;
+
+                if volume.is_some() {
+                    diesel::delete(dsl::volume)
+                        .filter(dsl::id.eq(volume_id))
+                        .execute(conn)?;
+                }
```
smklein, line 182: Do we need to do the lookup first?

If we just unconditionally performed:

```rust
// We don't care if we deleted this volume, or if it has already been deleted.
let _row_count = diesel::delete(dsl::volume).filter(dsl::id.eq(volume_id)).execute(conn)?;
Ok(())
```

This seems like it would use one less query, and wouldn't require a transaction.

nexus/src/app/sagas/disk_create.rs
```diff
@@ -970,6 +1020,7 @@ pub(crate) mod test {
         let sled_agent = &cptestctx.sled_agent.sled_agent;
         let datastore = cptestctx.server.apictx().nexus.datastore();
 
+        assert!(no_stuck_sagas(&cptestctx.logctx.log, datastore).await);
```
smklein, line 1023: I'm assuming yes, but was this failing before the non-test changes?

smklein (approved):

nexus/db-queries/src/db/datastore/disk.rs
```diff
@@ -568,17 +568,12 @@ impl DataStore {
...
     ) -> Result<db::model::Disk, Error> {
         use db::schema::disk::dsl;
         let pool = self.pool();
         let now = Utc::now();
 
-        let ok_to_delete_states = vec![
-            api::external::DiskState::Detached,
-            api::external::DiskState::Faulted,
-            api::external::DiskState::Creating,
-        ];
```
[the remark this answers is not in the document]
jmpesp: To clarify, it's ok for the `disk_create` saga to delete its disk that is in state `Creating` (during unwind), but not for the `disk-delete` to delete some random disk in state `Creating`: the second case is interfering with another running `disk_create` saga.

nexus/db-queries/src/db/datastore/region.rs
jmpesp: I think we should retry all transactions that CRDB tells us to :) but I think we'll need to convince ourselves that this won't create unbounded work for Nexus.

nexus/db-queries/src/db/datastore/volume.rs
jmpesp: I did this for two reasons:

1) to separate out the errors: in the case where we have an explicit check then delete, any error we receive from the delete is unrelated to the record missing. Although, now that I type this out: what other error could we receive from a delete call?

2) it's more clearly idempotent to my eyes, although I'll admit that's a weaker justification.

nexus/db-queries/src/db/datastore/region.rs
smklein: Would you be okay making a separate issue for this? After this PR merges, I could help with this.

nexus/db-queries/src/db/datastore/volume.rs
smklein: Do we actually do (1) though? If no such volume exists, we don't throw an error, do we? Don't we just set `let volume = None`, and then skip the delete?

The cases of "the volume exists / doesn't exist" seem the same to me, but doing the deletion with these two steps just requires an extra query

nexus/src/app/sagas/disk_create.rs
jmpesp: Not if you apply this patch on `main`, but I did see stuck saga(s) that prompted this check.

nexus/db-queries/src/db/datastore/region.rs
jmpesp: Definitely, opened https://github.com/oxidecomputer/omicron/issues/3814

nexus/db-queries/src/db/datastore/volume.rs
jmpesp: You're right, I've changed to just the delete in 9efe842a4. `test_volume_hard_delete_idempotent` would have caught this if it wasn't true!

Issue #3588, opened 2023-07-13
external-networking: when adding addresses to ASIC, also add them to corresponding tfport


rcgoodfellow:
When we set up an external address, through RSS or otherwise, there is machinery to land the address on the ASIC. But we also need to land the address in the corresponding tfport. Here is why.

When nodes from within the rack try to communicate outbound, boundary services will decap the packet. Then in P4 we try to resolve the MAC for the destination IP. When it we can't, because there is no entry yet, the P4 code will kick a "needs resolving" packet up to the host. If the host does not have an address on the subnet of the address being resolved, the resolution will fail. Therefore we need addresses assigned to the ASIC to also be assigned to the corresponding switch zone tfport.

CC: @internet-diglett

rcgoodfellow:
Maybe we should just do this as a general sync in tfportd like we are doing in this PR

- https://github.com/oxidecomputer/dendrite/pull/497

luqmana:
(Out of curiousity, took a quick look at the illumos side of where the ARP request was failing: [This is where the `ip2mac` request eventually bails out](https://github.com/oxidecomputer/illumos-gate/blob/3b24312d6398cc8e8e513562a52571c416322378/usr/src/uts/common/inet/ip/ip_ndp.c#L4904-L4910).)

jgallagher:
Fixed by #3765

PR #3755, opened 2023-07-24
[RSS] Change `uplink_ip` to `uplink_cidr` in rack network config

jgallagher:
This is necessary (but not sufficient) to address #3588 - tfport needs the netmask/prefix for each uplink, which means we need to collect it at RSS time.

Most of this PR is various openapi plumbing changes. There are a couple spots that warrant a closer look (although they are small); I'll call them out in comments below.

### DIFF
--- a/nexus/src/app/rack.rs
+++ b/nexus/src/app/rack.rs
@@ -34,6 +34,7 @@ use omicron_common::api::external::DataPageParams;
 use omicron_common::api::external::Error;
 use omicron_common::api::external::IdentityMetadataCreateParams;
 use omicron_common::api::external::IpNet;
+use omicron_common::api::external::Ipv4Net;
 use omicron_common::api::external::ListResultVec;
 use omicron_common::api::external::LookupResult;
 use omicron_common::api::external::Name;
@@ -458,11 +459,8 @@ impl super::Nexus {
                     addresses: HashMap::new(),
                 };
 
-                let uplink_address = IpNet::from_str(&format!("{}/32", uplink_config.uplink_ip))
-                    .map_err(|e| Error::internal_error(&format!(
-                    "failed to parse value provided for `rack_network_config.uplink_ip` as `IpNet`: {e}"
-                )))?;
-
+                let uplink_address =
+                    IpNet::V4(Ipv4Net(uplink_config.uplink_cidr));
                 let address = Address {
                     address_lot: NameOrId::Name(address_lot_name.clone()),
                     address: uplink_address,
--- a/sled-agent/src/rack_setup/service.rs
+++ b/sled-agent/src/rack_setup/service.rs
@@ -610,7 +610,7 @@ impl ServiceInner {
                         .map(|config| NexusTypes::UplinkConfig {
                             gateway_ip: config.gateway_ip,
                             switch: config.switch.clone().into(),
-                            uplink_ip: config.uplink_ip,
+                            uplink_cidr: config.uplink_cidr,
                             uplink_port: config.uplink_port.clone(),
                             uplink_port_speed: config
                                 .uplink_port_speed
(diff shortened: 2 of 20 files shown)


nexus/src/app/rack.rs
```diff
@@ -458,11 +459,8 @@ impl super::Nexus {
                     addresses: HashMap::new(),
                 };
 
-                let uplink_address = IpNet::from_str(&format!("{}/32", uplink_config.uplink_ip))
-                    .map_err(|e| Error::internal_error(&format!(
-                    "failed to parse value provided for `rack_network_config.uplink_ip` as `IpNet`: {e}"
-                )))?;
-
+                let uplink_address =
+                    IpNet::V4(Ipv4Net(uplink_config.uplink_cidr));
```
jgallagher, line 463: This is the first place that warrants an extra look: previously we were converting an IP to an `IP/32`; we now take the `IP/prefix` as-is from RSS. Will this still be correct even if the prefix is something other than 32?

sled-agent/src/rack_setup/service.rs
```diff
@@ -1188,7 +1188,10 @@ impl ServiceInner {
             v6_routes: HashMap::new(),
         };
         let link_id = LinkId(0);
-        let addr = IpAddr::V4(uplink_config.uplink_ip);
+        // TODO We're discarding the `uplink_cidr.prefix()` here and only using
+        // the IP address; this needs to be fixed as part of
+        // https://github.com/oxidecomputer/omicron/issues/3588.
+        let addr = IpAddr::V4(uplink_config.uplink_cidr.ip());
```
jgallagher, line 1194: This is the second place that warrants an extra look: we're discarding the prefix here. I think this roughly where we'll need to _not_ do that in order to fully solve #3588?

nexus/src/app/rack.rs
rcgoodfellow: For IPv4, prefix > 32 should be an error. For IPv6, prefix > 128 should be an error. The absence of a prefix is generally interpreted as /32 for IPv4 and /128 for IPv6. As long as these conventions hold, we should be good.

sled-agent/src/rack_setup/service.rs
rcgoodfellow: tl;dr; just the IP is fine here.

Based on my understanding of our conversation today, these addresses are going to two places now.

1. To `dpd` (what's being done here)
2. To the SMF properties of a new service being created right now.

For (1) `dpd` does not care about prefix lengths. It just uses addresses to inspect incoming packets to see if they need to go to the scrimlet e.g., determining if the destination address on a packet is owned by the switch. This has nothing to do with the prefix length. Place (2) is where we'll need the prefix length as that is how the address will get on illumos - and since illumos will be sourcing traffic from that address and resolving other addresses on the same L2 domain, the prefix length becomes important.

smf/sled-agent/gimlet-standalone/config-rss.toml
```diff
@@ -104,7 +104,7 @@ uplink_port_speed = "40G"
 uplink_port_fec="none"
 # For softnpu, an address within the "infra" block above that will be used for
 # the softnpu uplink port.  You can just pick the first address in that pool.
-uplink_ip = "192.168.1.30"
+uplink_cidr = "192.168.1.30/32"
```
rcgoodfellow, line 107: I think these will parse out the same.

```
$ cat src/main.rs
fn main() {
    let v4_a: ipnetwork::Ipv4Network = "10.1.2.3".parse().unwrap();
    let v4_b: ipnetwork::Ipv4Network = "10.1.2.3/32".parse().unwrap();
    println!("{:#?}", v4_a);
    println!("{:#?}", v4_b);
}
```
```
$ cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.00s
     Running `target/debug/ipparse`
Ipv4Network {
    addr: 10.1.2.3,
    prefix: 32,
}
Ipv4Network {
    addr: 10.1.2.3,
    prefix: 32,
}
```

nexus/src/app/rack.rs
jgallagher: :+1: `Ipv4Network` (the type of `uplink_config.uplink_cidr` here) ensures that the prefix is <= 32.

smf/sled-agent/gimlet-standalone/config-rss.toml
jgallagher: You are correct! Given the name change to `..._cidr`, I thought it was slightly clearer to include the explicit `/32`, but if that is surprising or less clear we could take it back out.

jgallagher: I'm going to go ahead and merge; happy to do a small followup to remove the `/32` from the config files if you'd prefer.

PR #3756, opened 2023-07-24
move underlay advertisement before sled-server start

### LINKED ISSUE #3532: sled-agent: advertise underlay prefix at startup

rcgoodfellow:
- Fixes #3532

### DIFF
--- a/sled-agent/src/bootstrap/agent.rs
+++ b/sled-agent/src/bootstrap/agent.rs
@@ -769,6 +769,19 @@ impl Agent {
                     monitor: hardware_monitor,
                 };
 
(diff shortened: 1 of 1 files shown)


luqmana (approved):
Still curious where sled-agent got stuck but seems like this is something we want to do earlier anyways.

Issue #3718, opened 2023-07-19
Add wicket commands to shut down and start up a rack without erasing data


sunshowers:
Add wicket commands to:

- [ ] Shut down a rack. Involves asking sled agent (?) to delete zones but not user data.
- [ ] Start up a rack.

We already have rack init and reset defined in https://github.com/oxidecomputer/omicron/blob/7a4d7a79aef4aa16bb11227015c84d1735745beb/sled-agent/src/bootstrap/http_entrypoints.rs#L127-L157. Shutdown and startup are similar but generally do less.

Some notes:
* This should be exposed via the bootstrap interface.
* https://github.com/oxidecomputer/omicron/blob/fe9d1c2bd18c1a5ed97af8eb83b44ecb47806318/sled-agent/src/bootstrap/agent.rs#L763-L772

sunshowers:
Some thoughts from having stared at this for a few hours:

* Is this really a shutdown operation if we're deleting zones? Do we need to call this something else?
* There's both `sled_reset` and `rack_reset` operations. Do we need `sled_shutdown` and `rack_shutdown`? Or is wicketd going to take care of orchestrating the rack shutdown?
* Where does the UI go? If it's sled-level shutdown then it should go in the sled screen I guess, maybe in the inventory screen?
* But at least right now, the inventory screen does not know if zones are currently alive (I believe), so we need to add some notion of that.
* Is this operation also going to move the sled to A2 (which is what I think "shutdown" entails)?

smklein:
> * Is this really a shutdown operation if we're deleting zones? Do we need to call this something else?

For what it's worth, we expect that zone filesystems will be deleted after a power-cycle, so I think this is "normal" for a shutdown operation?

(there's a subtle detail here -- for zones that want to act on persistent data, like cockroachdb or Crucible downstairs, they act on a dataset mounted at `/data`. These are not deleted during shutdown, nor should they be. But "all other files in the zone" are contained within the zone filesystem, and these **should** be destroyed between boots)

> * There's both `sled_reset` and `rack_reset` operations. Do we need `sled_shutdown` and `rack_shutdown`? Or is wicketd going to take care of orchestrating the rack shutdown?

I think "sled shutdown" probably makes sense, and wicketd could probably be the one doing orchestration. That allows an operator, through wicketd, so confirm what's being powered off.

> * Where does the UI go? If it's sled-level shutdown then it should go in the sled screen I guess, maybe in the inventory screen?
[...]

PR #3760, opened 2023-07-25
[nexus] Implement on-boot DB Schema Migration

smklein:
This PR does the following:

- Adds a `schema/crdb` directory, containing `dbinit.sql`, `dbwipe.sql`, and a collection of versions.
- When new DB versions are added, we can update `dbinit.sql` (as we have been doing) but also add a `up.sql` file to a directory named with the new DB version.
- On boot, nexus acquires a lease for the ability to perform schema changes, and it executes them. Nexus refuses to boot until the on-disk schema version matches the version embedded in the Nexus binary.

Furthermore, we provide an example "schema migration" from version 1.0.0 to version 1.0.1.

Some implementation details:
- This PR adds several tests to confirm this behavior works as expected, across all current and future updates
- This PR makes `db_metadata` more strongly-typed, so make it easier to access the underlying data
- This PR modified `dbinit.sql` to be idempotent, to be in-line with the recommended pattern for all subsequent schema migrations too
- We embed the schema changes into the Nexus binary, so they can be parsed and selectively applied. Along these lines, it might make sense to *stop* embedding a schema into the CRDB zone, and leave the initialization entirely under the control of Nexus.


nexus/db-queries/src/db/datastore/db_metadata.rs
```diff
@@ -8,21 +8,183 @@ use super::DataStore;
...
+///
+/// Should not be constructed unless this Nexus instance successfully modified
+/// the db_metadata table to identify that a schema migration is in progress.
+pub struct SchemaUpdateLease {
+    our_id: Uuid,
+}
 
 impl DataStore {
+    // Ensures that the database schema matches "desired_version"
+    pub async fn ensure_schema(
```
smklein, line 33: There is an assumption here that "all running nexus instances" at a new version are booting together. This is how we can read the schema version, and update to a "new version" without worrying about pulling the rug out from underneath a prior instance.

This isn't true! It may be possible that we run Nexus 1.0.0, and concurrently provision Nexus 2.0.0. This will be necessary follow-up work for online update, but this PR as-is relies on the commands from https://github.com/oxidecomputer/omicron/issues/3718 being executed.

To mitigate: I think Nexus is going to need to rely on tracking "what Nexus instances are provisioned, and where".

schema/crdb/README.adoc
```diff
@@ -0,0 +1,60 @@
...
+
+* We assume that downtime is acceptable to perform an update.
+* We assume that while an update is occuring, all Nexus services
+are running the same version of software.
+* We assume that no (non-upgrade) concurrent database requests will happen for
+the duration of the migration.
+
+This is not an acceptable long-term solution - we must be able to update
+without downtime - but it is a short-term solution, and one which provides a
+fall-back pathway for performing upgrades without data loss.
```
ahl, line 32:
```suggestion
without downtime - but it is an interim solution, and one which provides a
fall-back pathway for performing upgrades.
```

quiet part loud

schema/crdb/README.adoc
```diff
@@ -0,0 +1,60 @@
...
+
+Assumptions:
+
+* The (previously) latest schema version is referred to as `OLD_VERSION`
+* Your new changes will bring the schema to a new version, `NEW_VERSION`
+
+Process:
+
+* Add a file to `schema/crdb/NEW_VERSION/up.sql` with your changes to the schema.
+* Add a file to `schema/crdb/OLD_VERSION/down.sql` with the changes to undo your
```
ahl, line 46: does this imply that to get from A -> B we'd migrate A -> B -> C? If that's the case would it be worth stating explicitly?

schema/crdb/dbinit.sql
```diff
@@ -32,8 +32,8 @@ BEGIN;
  * NOTE: the database and user names MUST be kept in sync with the
```
ahl, line 32: this is a lot of delta, which is fine of course, but if it's NOT all of the form `s/CREATE (.*) (.*)/CREATE \1 IF NOT EXISTS \2/` could you highlight OTHER changes?

schema/crdb/dbinit.sql
```diff
@@ -2513,20 +2514,40 @@ CREATE TABLE omicron.public.switch_port_settings_address_config (
 /*******************************************************************/
 
 /*
- * Metadata for the schema itself.  This version number isn't great, as there's
+ * Metadata for the schema itself. This version number isn't great, as there's
  * nothing to ensure it gets bumped when it should be, but it's a start.
```
ahl, line 2518: What's non-great about the version number?

schema/crdb/dbinit.sql
```diff
@@ -2513,20 +2514,40 @@ CREATE TABLE omicron.public.switch_port_settings_address_config (
...
-CREATE TABLE omicron.public.db_metadata (
-    name  STRING(63) NOT NULL PRIMARY KEY,
-    value STRING(1023) NOT NULL
+CREATE TABLE IF NOT EXISTS omicron.public.db_metadata (
+    -- There should only be one row of this table for the whole DB.
+    -- It's a little goofy, but filter on "singleton = true" before querying
+    -- or applying updates, and you'll access the singleton row.
+    --
+    -- We also add a constraint on this table to ensure it's not possible to
+    -- access the version of this table with "singleton = false".
```
ahl, line 2527: I didn't follow that, but I may once I've looped back through everything.

nexus/db-queries/src/db/datastore/db_metadata.rs
ahl: "Booting together" means "are at the same version and/or are identical"--is that right?

Any reason not to put this note into a comment in the code?

nexus/db-queries/src/db/datastore/db_metadata.rs
```diff
@@ -8,21 +8,183 @@ use super::DataStore;
...
+        {
+            if entry.file_type().await.map_err(|e| e.to_string())?.is_dir() {
+                let name = entry
+                    .file_name()
+                    .into_string()
+                    .map_err(|_| "Non-unicode schema dir".to_string())?;
+                if let Ok(observed_version) = name.parse::<SemverVersion>() {
+                    all_versions.insert(observed_version);
+                } else {
+                    warn!(log, "Failed to parse {name} as a semver version");
```
ahl, line 97: do we want to be more conservative here? I get that we may not want to detonate if someone left behind a temp file or something... but is this surprising enough that we might want to refuse to proceed?

nexus/db-queries/src/db/datastore/db_metadata.rs
```diff
@@ -8,21 +8,183 @@ use super::DataStore;
...
+                }
+            }
+        }
+
+        if all_versions.get(&current_version).is_none() {
```
ahl, line 102:
```suggestion
        if !all_versions.contains(&current_version) {
```

nexus/db-queries/src/db/datastore/db_metadata.rs
```diff
@@ -8,21 +8,183 @@ use super::DataStore;
...
+            return Err(format!(
+                "Current DB version {current_version} is not known"
```
ahl, line 104:
```suggestion
                "Current DB version {current_version} was not found in {}", config.schema_dir
```
maybe?

nexus/db-queries/src/db/datastore/db_metadata.rs
```diff
@@ -8,21 +8,183 @@ use super::DataStore;
...
+            // WITH
+            //   x AS (SHOW JOBS)
+            // SELECT * FROM x WHERE
+            //   job_type = 'SCHEMA CHANGE' AND
+            //   status != 'succeeded';
+            //
+            // This would enable concurrent operations to happen on the database
+            // while we're mid-update. However, there is subtlety here around
+            // the visibility of renamed / deleted fields, unique indices, etc,
+            // so in the short-term we simply block on this job performing the
```
ahl, line 161: "short-term" seems optimistic here, but optimism is a value!

schema/crdb/README.adoc
```diff
@@ -0,0 +1,60 @@
...
+This directory describes the schema(s) used by CockroachDB.
+
+We use the following conventions:
+
+* `schema/crdb/VERSION/up.sql`: The necessary idempotent migrations to transition from the
+  previous version of CockroachDB to this version.
+* `schema/crdb/VERSION/down.sql`: The necessary idempotent migrations to transition from
+  the current version of CockroachDB to the immediately preceding version.
+* `schema/crdb/dbinit.sql`: The necessary operations to create the latest version
+  of the schema. Should be equivalent to running all `up.sql` migrations, in-order.
```
ahl, line 15: do we test that?

ahl (commented):
- do we need to keep migrations around forever?
- how do we feel about the ordering of the names of version directories being critical to the correct execution of migration?
- don't break this up, but if you do, just break out the mechanical / reorg changes
- do we definitely want the application of schema updates to be automated via nexus rather than being a manual step (i.e. for now)?

nexus/db-queries/src/db/datastore/db_metadata.rs
smklein: I will move this comment to the code -- just wanted to highlight these shortcomings and confirm they're "okay enough" for this to land in the short-term.

NOTE TO SELF BEFORE MARKING RESOLVED: File an issue for this

schema/crdb/README.adoc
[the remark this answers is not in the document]
smklein: Fixed

schema/crdb/README.adoc
```diff
@@ -0,0 +1,60 @@
+:showtitle:
+:numbered:
+
+= Database Schemas
+
+This directory describes the schema(s) used by CockroachDB.
+
+We use the following conventions:
+
+* `schema/crdb/VERSION/up.sql`: The necessary idempotent migrations to transition from the
```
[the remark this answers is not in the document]
smklein: We will need to do this for any change to dbinit.sql -- fortunately, there are automated tests which fail if someone changes dbinit.sql without adding the necessary version directory.

"When do we need to add a version vs use an existing version directory" - I think this is entirely dependent on our release schedules? Like, if we wanted to bundle multiple changes together, we could, but also incrementing a minor version for each one doesn't seem atrocious, as long as we can perform the updates quickly.

schema/crdb/README.adoc
smklein: I'm assuming you mean "A -> C" would involve "A -> B -> C", and that's correct. Skipping versions would be disallowed. I'll add a note.

schema/crdb/dbinit.sql
smklein: Almost all these changes are for idempotency and adding explicit names.

The only significant changes to dbinit.sql exist at the bottom, in regard to the `db_metadata` table, which has been re-worked.

schema/crdb/dbinit.sql
smklein: Dunno, this was a comment from before my PR. Do you want me to remove it? It's perhaps not sufficient enough (now) to identify any other information about "how to apply the change".

schema/crdb/dbinit.sql
smklein: This one is straightforward -- we just want a table with one row.

- If we didn't have the `singleton` column: Someone could insert several rows into this DB, which would be confusing (which one would be valid?)
- If we didn't have the `singleton` constraint, we'd be allowed **two** rows (one for "true", one for "false"). This limits it to one.

nexus/db-queries/src/db/datastore/db_metadata.rs
smklein: Sure, I'll throw an error here.

nexus/db-queries/src/db/datastore/db_metadata.rs
smklein: Updated

schema/crdb/README.adoc
smklein: We do, in `nexus/tests/integration_tests/schema.rs`.

- We rely on the `information_schema` virtual table in CockroachDB, and query for the state of tables, indices, columns, views, etc.
- We compare the version of the DB populated via `dbinit.sql` and the version of the DB populated via "all the up.sql files, in-order".
- We try to show a pretty diff if they're divergent, and we fail the test.

nexus/db-queries/src/db/datastore/db_metadata.rs
smklein: On a geological scale...

smklein:
> * do we need to keep migrations around forever?

No, we only need to keep [the oldest version we support, the newest version we support] around. If we want to drop support for older versions, we can replace the earliest `up.sql` file with the `dbinit.sql` file from the time that version was checked-in.

> * how do we feel about the ordering of the names of version directories being critical to the correct execution of migration?

It's certainly a linear history? We rely on the linear sequence of schema changes being present, but we could consider adding an additional step of "referencing the version that must come before this one", if that would help?

E.g., `1.0.1/upgrade_from` could be a file that contains `1.0.0`, to assert that none of the linear history is missed?

> * do we definitely want the application of schema updates to be automated via nexus rather than being a manual step (i.e. for now)?

It's a good question, and I've gone back-and-forth on this. I do like the fact that, as implemented, using this automated mechanism:

1. Lets us test the schema change in the same context that it would exist on a real rack
2. Makes it more difficult to "accidentally mistype" a command
3. Make an operator's life easier, by simply giving them fewer steps

nexus/db-queries/src/db/datastore/db_metadata.rs
```diff
@@ -8,21 +8,183 @@ use super::DataStore;
...
+
+            // Before we attempt to perform an update, grab a lease to allow
+            // this Nexus instance to perform the update.
+            //
+            // If another instance is performing an update, we will fail, and
+            // let them to the work on our behalf.
+            let lease = self
+                .grab_schema_update_lease(&current_version, &target_version)
+                .await
+                .map_err(|e| e.to_string())?;
```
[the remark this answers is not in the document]
andrewjstone: Anytime I hear the word "lease" I get concerned, because of the rough edges involved in making it work. I think you are making the right decision short term on not expiring the lease based on time as that alleviates the bulk of things that can go wrong from a safety POV. You can only lose liveness if the leas holder crashes, as you say. 

If we end up sticking with this, and expiring leases based on time then we must use a fencing token to gate an old lease holder who doesn't realize it has expired from writing over a newer lease holder. It's unclear to me if the `current_version` filter is good enough here. It may be that it is fine because if it matches the stale lease holder is just going to try the same migration and fail. 

This is my go to resource for discussing distributed leases/locks and fencing tokens: https://martin.kleppmann.com/2016/02/08/how-to-do-distributed-locking.html

You can ignore all the redlock stuff. Important sections are "Protecting a resource with a lock" and "Making the lock safe with fencing"

davepacheco (commented):
Thanks for taking this on!

> > * do we definitely want the application of schema updates to be automated via nexus rather than being a manual step (i.e. for now)?
> 
> It's a good question, and I've gone back-and-forth on this. I do like the fact that, as implemented, using this automated mechanism:
> 
>     1. Lets us test the schema change in the same context that it would exist on a real rack
> 
>     2. Makes it more difficult to "accidentally mistype" a command
> 
>     3. Make an operator's life easier, by simply giving them fewer steps

What about putting all the schema update stuff into a command-line tool that could be invoked by an operator as part of the manual upgrade procedure?  So the upgrade procedure would include:

- shut down all (old version) nexus instances
- (update the Nexus images everywhere? or something?  I'm not sure what this step looks like)
- verify that all Nexus instances are offline
- run this tool to do the schema update
- (optionally: inspect the schema / database for correctness)
- start Nexus instances

This feels a lot more predictable and controlled.

The choice to do this automatically on startup feels risky to me.  Admittedly, we got burned by this sort of thing several times at Joyent.  We had a few major outages resulting from the behavior of idempotently, automatically applying schema updates.  These may have largely been PostgreSQL-specific having to do with the exclusive locks taken while doing this.  But I came away from that experience feeling like until we've really ironed out the kinks associated with this complex process, it's dangerous to have software do it automatically, especially when operators and developers alike might not be expecting it.  As a silly example, a developer might be quite surprised that by just deploying a Nexus zone at a newer version than what they're running, they've pulled the rug out from the one that was already running.

On the plus side, I think we know this behavior is not part of the long-term plan.  RFD 319 discusses this under 5.1 -- basically, you _can't_ do online schema updates if new versions always update the schema because you're always pulling the rug out from someone.  So I don't think there's a compelling reason to do this automatically in Nexus?

nexus/db-model/src/db_metadata.rs
```diff
@@ -3,25 +3,35 @@
...
+use uuid::Uuid;
 
 /// Internal database metadata
-#[derive(Queryable, Insertable, Debug, Clone, Selectable)]
+#[derive(
+    Queryable, Insertable, Debug, Clone, Selectable, Serialize, Deserialize,
+)]
 #[diesel(table_name = db_metadata)]
 pub struct DbMetadata {
-    name: String,
```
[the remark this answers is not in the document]
smklein: I did try keeping these values as strings, but there are major downsides:

- Parsing is a pain-in-the-butt
- Comparing the diff of built-in data between schema changes is terrible, in part because we can't identify "date" fields as such, and treat them as strings. As a consequence, `NOW() as TEXT` shows up as two completely different values when created at two different times.
- We could still have schema changes that modify strings! The changes are just implicit, and not codified in SQL itself

So:

I hear you that "if we want to use metadata as a part of schema updates, we need to update first". Totally agreed! But I think that's true of strings **and** of more strongly-typed schemas.

package-manifest.toml
```diff
@@ -94,6 +94,7 @@ source.paths = [
   { from = "/opt/ooce/pgsql-13/lib/amd64", to = "/opt/ooce/pgsql-13/lib/amd64" },
   { from = "smf/nexus", to = "/var/svc/manifest/site/nexus" },
   { from = "out/console-assets", to = "/var/nexus/static" },
+  { from = "schema/crdb", to = "/var/nexus/schema/crdb" },
```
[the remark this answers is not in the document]
smklein: When CRDB was the only database, this made sense, but we also have Clickhouse, which itself may require schema migrations.

Personally I found "crdb" to be clarifying, as it's only two extra letters, but it disambiguates "which database we're using".

Are you okay keeping this as-is?

schema/crdb/1.0.1/down.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE omicron.public.instance
```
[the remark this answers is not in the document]
smklein: I can add comments -- confirmed that all my schema migration tests still pass with comments in-line.

package-manifest.toml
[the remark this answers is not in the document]
davepacheco: 💯 fine with this as-is.  The reason I brought it up is that we have almost nothing named "crdb" today:

```
$ find * -iname '*crdb*'
docs/crdb-debugging.adoc
$
```

whereas we have a bunch of things called "db" already that are CockroachDB-specific (nexus/db-model, nexus/db-queries, test-utils/src/dev/db.rs, omicron-dev db commands, etc.)

smklein:
> What about putting all the schema update stuff into a command-line tool that could be invoked by an operator as part of the manual upgrade procedure? So the upgrade procedure would include:
> 
> * shut down all (old version) nexus instances
> * (update the Nexus images everywhere? or something?  I'm not sure what this step looks like)
> * verify that all Nexus instances are offline
> * run this tool to do the schema update
> * (optionally: inspect the schema / database for correctness)
> * start Nexus instances
> 
> This feels a lot more predictable and controlled.

There are a few caveats here -- namely, that we don't **really** have a great way of having "some-but-not-all" services running, without treading into a lot of the zone management responsibilities of Nexus / Sled Agent, which is hairy.

HOWEVER, Nexus already has the logic to "wait on boot, checking the version", and refusing to boot if it doesn't match.

I could do the following:

- Refactor this migration code into a helper crate
- Keep calling it here! Just don't point to the config file for upgrades, so Nexus won't **use** it. (Notice: The config directory containing schema changes is optional, so we can just omit it for now)
- Write a binary that calls the same code which would be used by Nexus
- Tell an operator to call that

> The choice to do this automatically on startup feels risky to me. Admittedly, we got burned by this sort of thing several times at Joyent. We had a few major outages resulting from the behavior of idempotently, automatically applying schema updates. These may have largely been PostgreSQL-specific having to do with the exclusive locks taken while doing this. But I came away from that experience feeling like until we've really ironed out the kinks associated with this complex process, it's dangerous to have software do it automatically, especially when operators and developers alike might not be expecting it. As a silly example, a developer might be quite surprised that by just deploying a Nexus zone at a newer version than what they're running, they've pulled the rug out from the one that was already running.

I think that's a fair critique, we should get mileage on this process manually. However, it's still possible for an operator to issue these upgrade requests while an old Nexus is running, which would still "pull the rug out".

I do want to move to an automated world when we safely can, but I agree that perhaps the right time to do so would be "when Nexus can make all the safety assertions it needs, for itself".

> On the plus side, I think we know this behavior is not part of the long-term plan. RFD 319 discusses this under 5.1 -- basically, you _can't_ do online schema updates if new versions always update the schema because you're always pulling the rug out from someone. So I don't think there's a compelling reason to do this automatically in Nexus?

Yeah, this is part of why I started with the "offline approach", I think we'll have schema upgrades which are compatible with older Nexuses, and schema upgrades which are **not** compatible with older Nexuses. For example, adding a column with a default value wouldn't break old Nexuses (or at least, it **shouldn't**), but we don't really structured policy and language outside of the semver version for describing this compatibility. Maybe that's enough? But I wanted to be cautious before claiming that we had a generalized solution there.

nexus/db-queries/src/db/datastore/db_metadata.rs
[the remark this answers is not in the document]
smklein: I spent a while working on this, and I have an alternate approach.

https://www.cockroachlabs.com/docs/stable/online-schema-changes#schema-changes-within-transactions

- CockroachDB cautions strongly against performing multiple DDL statements in a transaction
- CockroachDB recommends strongly against performing both DDL and DML statements in a transaction (since one might succeed, and the other might fail).

However, they **don't** disallow running a **query** operation in a transaction which contains a single DDL operation.

So, I have the following proposal for upgrading from N -> N+1:

1. Mark the schema upgrade as in-progress, if we are at the version we expect to be:

```sql
UPDATE db_metadata SET target_version = N + 1 WHERE version = N AND singleton = true;
```

2. Perform the schema upgrade in a transaction, confirming that the version and target versions have not changed:

```sql
BEGIN;

SELECT CAST(
    IF(
        (
            SELECT version = N and target_version = N + 1
            FROM omicron.public.db_metadata WHERE singleton = true
        ),
        'true',
        'Invalid starting version for schema change'
    ) AS BOOL
);

ALTER TABLE omicron.public.instance
    ADD COLUMN IF NOT EXISTS boot_on_fault BOOL NOT NULL DEFAULT false;

COMMIT;
```

3. Finalize the update

```sql
UPDATE db_metadata SET version = N+1, target_version = NULL WHERE version = N AND target_version = N +1 AND singleton = true;
```

These steps are safe to execute from multiple Nexus instances concurrently, and they abandon the DDL changes which might be taken if a different Nexus manages to complete the upgrade early.

I also was able to write a test for this behavior, where I have the following versions:

- v0: Starting version
- v1: Sleep a random amount, Add table
- v2: Delete table

And I do the following:

- Launch 10 Nexus Instances
- Let them all try to upgrade from v0 -> v1 -> v2
- Observe that, as each of them finishes initializing the datastore, they **cannot see the table, because it has been deleted, and no "late upgrades" re-add it**.

(Without the transactional query in step (2), this test fails. With it, this test succeeds)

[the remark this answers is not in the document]
smklein: Why I like this new version:
- There is no "lease" or "lock" mechanism necessary. Nexus instances can concurrently execute.
- If Nexus instances crash before step (3), they should be able to re-do steps (1) and (2) to ensure the update was applied.
- As soon as step (3) is applied, the update is complete, and subsequent calls to step (1) or (2) will fail (as they should!)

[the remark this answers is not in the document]
smklein: Got this feedback from a member of Cockroach Labs, when I asked about the safety the transaction in step (2):

> Hello! The code for a transaction that you showed is safe — it will block the DDL if there’s a concurrent UPDATE happening. Just to note that the hazard is from performing multiple schema changes within a single explicit transaction, so you may want to avoid running other DDL/DML after the ALTER TABLE  transactionally.

schema/crdb/README.adoc
```diff
@@ -0,0 +1,64 @@
...
+  previous version of CockroachDB to this version.
+* `schema/crdb/VERSION/down.sql`: The necessary idempotent migrations to transition from
```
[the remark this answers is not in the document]
smklein: Sure thing, I'll remove `down.sql` and the corresponding docs. This was a "nice-to-have", but it's true, the feature was not being used.

schema/crdb/README.adoc
```diff
@@ -0,0 +1,64 @@
...
+* We assume that no (non-upgrade) concurrent database requests will happen for
+the duration of the migration.
+
+This is not an acceptable long-term solution - we must be able to update
+without downtime - but it is an interim solution, and one which provides a
+fall-back pathway for performing upgrades.
+
+See RFD 319 for more discussion of the online upgrade plans.
+
+=== How to perform an offline upgrade
```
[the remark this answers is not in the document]
smklein: Updated

schema/crdb/README.adoc
```diff
@@ -0,0 +1,64 @@
+:showtitle:
+:numbered:
+
+= Database Schemas
+
+This directory describes the schema(s) used by CockroachDB.
```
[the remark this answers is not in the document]
smklein: If it helps, Nexus has an "EXPECTED_SCHEMA" constant, and will refuse to boot unless this is set. This constant is also set within `dbinit.sql`. So I would expect that, as you say, the merge conflict would appear, and whoever edits the merge conflict would decide "what is the latest version that should be known to Nexus". As a policy, we should only be "moving forward" here.

I can certainly start by bumping the major number only, until we codify what compatibility between multiple versions means more clearly.

smklein:
I've gone ahead and added a tool called `schema-updater`, which I'm bundling with Nexus.

The interface looks like the following:

```bash
root@oxz_nexus_6d1dd98a-3559-4aa6-bf43-22cdfaf59099:~# /opt/oxide/nexus/bin/schema-updater postgresql://root@[fd00:1122:3344:101::6]:32221/omicron?sslmode=disable list
Current Version: 1.0.0
Available Versions:
  1.0.0 (current)
  1.0.1 (expected)
root@oxz_nexus_6d1dd98a-3559-4aa6-bf43-22cdfaf59099:~# /opt/oxide/nexus/bin/schema-updater postgresql://root@[fd00:1122:3344:101::6]:32221/omicron?sslmode=disable up 1.0.1
Upgrading to 1.0.1
Jul 28 15:24:43.380 WARN Incompatible database schema: Saw 1.0.0, expected 1.0.1, unit: schema_updater
Upgrade to 1.0.1 complete
root@oxz_nexus_6d1dd98a-3559-4aa6-bf43-22cdfaf59099:~# /opt/oxide/nexus/bin/schema-updater postgresql://root@[fd00:1122:3344:101::6]:32221/omicron?sslmode=disable list
Current Version: 1.0.1
Available Versions:
  1.0.0
  1.0.1 (current) (expected)
```

I'd like to tackle https://github.com/oxidecomputer/omicron/issues/3763 to make "finding CRDB" slightly easier as a next step, but this does appear to work from my testing:

- Nexus boots at version 1.0.0, and waits to continue until the DB shows version 1.0.1
- An operator (me) runs the `schema-updater up 1.0.1` command, and sees the DB gets updated
- Nexus actually creates the datastore and continues, using DB version 1.0.1

nexus/db-queries/src/db/datastore/db_metadata.rs
smklein: I added a binary `schema-updater`, and removed the 'auto-format-on-boot' configuration from Nexus.

schema/crdb/dbinit.sql
smklein: > Relatedly, there's something missing about tracking what state we're in. If we start applying an update and it fails for any reason (including, say, a complete rack power off), then the schema is wrong, in that it will report version X, when it's in fact at some version between version X and X + 1.

We actually are setting this up correctly. In the DB, we internally track a version, as well as a "target_version" field, which is only non-NULL when we're performing an update.

> I'm not sure if this is a problem right now or not, but certainly it would allow older Nexuses to start back up and think they're at the right version, when in fact they're not compatible with what's currently in the database. I can't see an easy way to address this without adding a second piece of metadata for "the version we're in the process of updating to".

See my comments here: https://github.com/oxidecomputer/omicron/pull/3760#discussion_r1275598857 - I spent a while ensuring that even if Nexus (or `schema-updater`, now) crashes and reboots, we should be able to safely re-apply these updates, and we don't try using structures from some other version until the whole transition completes successfully.

nexus/db-model/src/schema.rs
```diff
@@ -1114,15 +1115,19 @@ table! {
...
+        version -> Text,
+        target_version -> Nullable<Text>,
+        nexus_upgrade_driver -> Nullable<Uuid>,
     }
 }
 
 /// The version of the database schema this particular version of Nexus was
 /// built against.
-pub const SCHEMA_VERSION: SemverVersion = SemverVersion::new(1, 0, 0);
+pub const SCHEMA_VERSION: SemverVersion = SemverVersion::new(1, 0, 1);
```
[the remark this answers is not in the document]
smklein: Sure, updated to 2.0.0 instead of 1.0.1

davepacheco:
> I've gone ahead and added a tool called `schema-updater`, which I'm bundling with Nexus.

Nice!  This flow looks pretty good.

> The interface looks like the following:
> 
> ```shell
> root@oxz_nexus_6d1dd98a-3559-4aa6-bf43-22cdfaf59099:~# /opt/oxide/nexus/bin/schema-updater postgresql://root@[fd00:1122:3344:101::6]:32221/omicron?sslmode=disable list
> Current Version: 1.0.0
> Available Versions:
>   1.0.0 (current)
>   1.0.1 (expected)
> ```

This looks good but I'd tweak these labels.  "current" and "expected" seem ambiguous.  How about:

- "current" -> "reported by the database"
- "expected" -> "deployable by this tool"

I realize that's kind of wordy and I'm not attached to those.  I just think it's important that the language we use to communicate with support people be really clear, since they can't be super familiar with everything and they're often in stressful situations.  Plus, this is typically a risky operation.

> ```shell
> root@oxz_nexus_6d1dd98a-3559-4aa6-bf43-22cdfaf59099:~# /opt/oxide/nexus/bin/schema-updater postgresql://root@[fd00:1122:3344:101::6]:32221/omicron?sslmode=disable up 1.0.1
> Upgrading to 1.0.1
> Jul 28 15:24:43.380 WARN Incompatible database schema: Saw 1.0.0, expected 1.0.1, unit: schema_updater
> Upgrade to 1.0.1 complete
> ```


This warning sounds alarming but if I'm understanding right this is totally expected in this situation, right?


> ```shell
> root@oxz_nexus_6d1dd98a-3559-4aa6-bf43-22cdfaf59099:~# /opt/oxide/nexus/bin/schema-updater postgresql://root@[fd00:1122:3344:101::6]:32221/omicron?sslmode=disable list
> Current Version: 1.0.1
> Available Versions:
>   1.0.0
>   1.0.1 (current) (expected)
> ```

Why is "1.0.0" considered "available" there?  Is that just listing all known versions?  Should it just say "Known versions"?

> I'd like to tackle #3763 to make "finding CRDB" slightly easier as a next step,

Could we make operators' lives easier by automatically using internal DNS to find CockroachDB here?  Then they wouldn't need to find the IP address or construct that URL.  (It should probably still be an argument as an escape hatch and for dev environments.)

> but this does appear to work from my testing:
> 
>     * Nexus boots at version 1.0.0, and waits to continue until the DB shows version 1.0.1
> 
>     * An operator (me) runs the `schema-updater up 1.0.1` command, and sees the DB gets updated
> 
>     * Nexus actually creates the datastore and continues, using DB version 1.0.1

Nice!

nexus/Cargo.toml
```diff
@@ -70,12 +70,14 @@ serde_urlencoded.workspace = true
...
 slog.workspace = true
+slog-async.workspace = true
 slog-dtrace.workspace = true
+slog-term.workspace = true
 steno.workspace = true
 tempfile.workspace = true
 thiserror.workspace = true
 tokio = { workspace = true, features = ["full"] }
-tokio-postgres = { workspace = true, features = ["with-serde_json-1"] }
+tokio-postgres = { workspace = true, features = ["with-serde_json-1", "with-uuid-1"] }
```
davepacheco, line 80: Why did we end up needing this?

nexus/db-model/src/schema.rs
```diff
@@ -1114,15 +1115,18 @@ table! {
...
+        singleton -> Bool,
+        time_created -> Timestamptz,
+        time_modified -> Timestamptz,
+        version -> Text,
+        target_version -> Nullable<Text>,
     }
 }
 
 /// The version of the database schema this particular version of Nexus was
 /// built against.
```
davepacheco, line 1128: I feel like we should flesh this out somewhere so that it's clear to developers (1) under what conditions this needs to be changed, (2) how it should be changed, and (3) what else needs to be changed with it, if anything.    Basically: assume someone comes in and wonders if they need to change this.  How do they answer that?  What do they need to know to change it safely?  If that's covered elsewhere (like the README in the schema file), we can just point people there.

nexus/db-queries/src/db/datastore/db_metadata.rs
```diff
@@ -8,21 +8,206 @@ use super::DataStore;
...
 use omicron_common::api::external::Error;
 use omicron_common::api::external::SemverVersion;
+use omicron_common::nexus_config::SchemaConfig;
+use slog::Logger;
+use std::collections::BTreeSet;
+use std::ops::Bound;
 use std::str::FromStr;
 
 impl DataStore {
+    // Ensures that the database schema matches "desired_version".
```
davepacheco, line 23: I think it'd be worth adding more here about what the invariants are around this.  Like:

- Updating the schema version makes the database incompatible with older Nexus instances.
- This is a one-way operation.  It cannot be undone.
- The caller is responsible for ensuring that this will not cause an outage (or something)

nexus/db-queries/src/db/datastore/mod.rs
```diff
@@ -137,38 +138,51 @@ pub struct DataStore {
...
 // recompilation of that query's module instead of all queries on `DataStore`.
 impl DataStore {
-    /// Constructs a new Datastore object.
+    /// Constructs a new Datastore object, without any version validation.
     ///
-    /// Only returns if the database schema is compatible with Nexus's known
-    /// schema version.
-    pub async fn new(log: &Logger, pool: Arc<Pool>) -> Result<Self, String> {
+    /// Ignores the underlying DB version. Should be used with caution, as usage
+    /// of this method can result in incompatible schemas.
```
davepacheco, line 144: What does "can result in incompatible schemas" mean?  What I'd expect is something like "this function returns a working DataStore whose methods may read and write data in a way that's incompatible with the way other components are reading and writing to the same database.  Database corruption could result."

Is this only used for updating the DataStore's schema version?  It seems like a (future) safer API would be a separate object for that that only lets you update the schema.

schema/crdb/2.0.0/up.sql
```diff
@@ -0,0 +1,25 @@
...
+    IF(
+        (
+            SELECT version = '1.0.0' and target_version = '2.0.0'
+            FROM omicron.public.db_metadata WHERE singleton = true
+        ),
+        'true',
+        'Invalid starting version for schema change'
+    ) AS BOOL
+);
+
```
davepacheco, line 21: Should this be part of the framework instead of having people copy/paste it into each up.sql?

davepacheco (approved):
Thanks.  This was a bit of a quick skim.  None of these notes are blockers.

nexus/Cargo.toml
smklein: We don't, I'll remove it!

nexus/db-model/src/schema.rs
smklein: Sounds good, I'll add some clarification and a pointer to the readme here.

nexus/db-queries/src/db/datastore/db_metadata.rs
smklein: Will update with these comments

schema/crdb/2.0.0/up.sql
smklein: I went back and forth on this. On the one hand: Yes, it would be really nice to cut out this boilerplate. On the other hand, the conditions of "what can someone call from within a transaction" are pretty strict -- I don't want anyone to accidentally write an `up.sql` file that includes multiple DDL statements. I think it would be more correct, in these situations, to have one transaction per DDL statement.

TL;DR: I would like this, but I haven't figured out a good way to safely hide the details of "operating within a single transaction".

nexus/db-queries/src/db/datastore/mod.rs
smklein: I'll update this comment for now. Agreed about the future safe API, but I also dunno how much to optimize for the case, because I think operator-driven updates may not use the CLI tools in the limit.

smklein:
> This looks good but I'd tweak these labels. "current" and "expected" seem ambiguous. How about:
> 
> * "current" -> "reported by the database"
> * "expected" -> "deployable by this tool"

Updated to "reported by database" and "expected by Nexus".
 
> > ```shell
> > root@oxz_nexus_6d1dd98a-3559-4aa6-bf43-22cdfaf59099:~# /opt/oxide/nexus/bin/schema-updater postgresql://root@[fd00:1122:3344:101::6]:32221/omicron?sslmode=disable up 1.0.1
> > Upgrading to 1.0.1
> > Jul 28 15:24:43.380 WARN Incompatible database schema: Saw 1.0.0, expected 1.0.1, unit: schema_updater
> > Upgrade to 1.0.1 complete
> > ```
> 
> This warning sounds alarming but if I'm understanding right this is totally expected in this situation, right?

Yeah, it's expected, re-worded to be a little less scary.

> > ```shell
> > root@oxz_nexus_6d1dd98a-3559-4aa6-bf43-22cdfaf59099:~# /opt/oxide/nexus/bin/schema-updater postgresql://root@[fd00:1122:3344:101::6]:32221/omicron?sslmode=disable list
> > Current Version: 1.0.1
> > Available Versions:
> >   1.0.0
> >   1.0.1 (current) (expected)
> > ```
> 
> Why is "1.0.0" considered "available" there? Is that just listing all known versions? Should it just say "Known versions"?

Updated.

> > I'd like to tackle #3763 to make "finding CRDB" slightly easier as a next step,
> 
> Could we make operators' lives easier by automatically using internal DNS to find CockroachDB here? Then they wouldn't need to find the IP address or construct that URL. (It should probably still be an argument as an escape hatch and for dev environments.)

Chatted about this offline -- this is complicated, so we're punting on it, but it's worth solving.

nexus/db-queries/src/db/datastore/db_metadata.rs
smklein: Filed https://github.com/oxidecomputer/omicron/issues/3794

PR #3761, opened 2023-07-25
Abandon contract after running command in zone

citrus-it:


citrus-it:
I tested this on a bench gimlet, confirming that there were no lingering contracts once the handoff to nexus is complete:

```
gimlet-sn06 # ls /proc/`pgrep sled-agent`/contracts
gimlet-sn06 #
```

During zone boot I watched this directory and saw contracts come and go, as expected.

Prior to this, sled-agent had accumulated a number of contracts, for example:

```
BRM42220017 # ls /proc/`pgrep sled-agent`/contracts
1112  1129  138   149   230   241   250   256   262   268   274   280   292
1113  1130  139   150   231   242   251   257   263   269   275   281
1114  1131  145   208   232   243   252   258   264   270   276   282
1120  1132  146   209   238   245   253   259   265   271   277   283
1121  1133  147   212   239   246   254   260   266   272   278   288
1128  137   148   229   240   249   255   261   267   273   279   289

BRM42220017 # ctstat -vi 149
CTID    ZONEID  TYPE    STATE   HOLDER  EVENTS  QTIME   NTIME
149     0       process owned   655     0       -       -
        cookie:                0
        informative event set: none
        critical event set:    none
        fatal event set:       hwerr
        parameter set:         pgrponly regent
        member processes:      none
        inherited contracts:   none
        service fmri:          svc:/oxide/sled-agent:default
        service fmri ctid:     61
        creator:               sled-agent
        aux:
```

with all of these being `process owned` but having no member processes.

illumos-utils/src/lib.rs
```diff
@@ -92,6 +92,31 @@ mod inner {
...
     // Helper function for starting the process and checking the
     // exit code result.
+    pub fn spawn(
+        command: &mut std::process::Command,
+    ) -> Result<std::process::Child, ExecutionError> {
+        let child = command.spawn().map_err(|err| {
+            ExecutionError::ExecutionStart { command: to_string(command), err }
+        })?;
+
+        Ok(child)
```
bnaecker, line 102: Style nit, but I think this can just be written as:

```rust
command.spawn().map_err(|err| {
    ExecutionError::ExecutionStart { command: to_string(command), err }
})
```

I.e, no need to use the `?`, we can just map the error type.

illumos-utils/src/running_zone.rs
```diff
@@ -169,13 +174,39 @@ mod zenter {
...
 
     #[link(name = "c")]
     extern "C" {
         pub fn zone_enter(zid: zoneid_t) -> c_int;
     }
 
+    pub fn get_contract(pid: pid_t) -> std::io::Result<ctid_t> {
+        // The offset of "id_t pr_contract" in struct psinfo which is an
+        // interface documented in proc(5).
+        const CONTRACT_OFFSET: u64 = 280;
```
bnaecker, line 188: This makes me a bit nervous, but I imagine this is a very stable interface? It's annoying, but we may want to write the struct out in full, with a `repr(C)` attribute, and then parse the full thing from the `psfinfo` file.

bnaecker (approved):
Looks good, just a couple of small thoughts. Thanks for finding and fixing this!

illumos-utils/src/running_zone.rs
citrus-it: Yes, the `/proc` interface for `psinfo` is stable. I thought this approach was less annoying than writing out the entire struct with repr(C) because it's quite large, uses a lot of different types and contains some other embedded structs. At the end of the day this is effectively equivalent.
It's a shame that libproc doesn't have some more ergonomic interfaces for this.

illumos-utils/src/running_zone.rs
```diff
@@ -169,13 +174,39 @@ mod zenter {
...
+    pub fn abandon_contract(ctid: ctid_t) {
+        let path = format!("/proc/{}/contracts/{}/ctl", process::id(), ctid);
+
+        let cpath = CString::new(path).unwrap();
+        let fd = unsafe { libc::open(cpath.as_ptr(), libc::O_WRONLY) };
+        if fd >= 0 {
+            unsafe { ct_ctl_abandon(fd) };
+            unsafe { libc::close(fd) };
+        }
+    }
```
jclulow, line 208: This routine should be fallible.  If `open()` fails, we need to return some detail about why.  We should also check the return values of `ct_ctl_abandon()` and `close()` and fail with error reports if those are not as expected.

Any error here represents a serious misunderstanding of the construction of the system, and could lead to more leaked contracts or leaked file descriptors.  At a minimum we need a noisy error log message describing what went wrong; ideally we'd abort the process and get a core file.

illumos-utils/src/lib.rs
```diff
@@ -90,8 +90,33 @@ mod inner {
...
     // exit code result.
+
+    pub fn spawn(
+        command: &mut std::process::Command,
+    ) -> Result<std::process::Child, ExecutionError> {
+        command.spawn().map_err(|err| ExecutionError::ExecutionStart {
+            command: to_string(command),
+            err,
+        })
+    }
```
jclulow, line 103: I think this routine should actually return a newtype that combines the `Command` and the `Child` so that we can pass _that_ to `run_child()` below.  Otherwise, the `run_child()` interface is pretty easy to misuse: nothing structural prevents passing a `Child` that does not match the `Command`.

illumos-utils/src/running_zone.rs
citrus-it: As both something which should never happen and something which is survivable, I made abandoning the contract best effort. I will adjust so that we get error messages if this occurs but will stop short of panicking.

Issue #3738, opened 2023-07-21
Early networking support for cold boot


andrewjstone:
We need to save some info to bootstore an re-populate on cold boot

andrewjstone:
PR out: https://github.com/oxidecomputer/omicron/pull/3729

askfongjojo:
Closed by https://github.com/oxidecomputer/omicron/pull/3765

PR #3765, opened 2023-07-25
Configure tfport uplinks via `uplink` service

jgallagher:


sled-agent/src/bootstrap/early_networking.rs
```diff
@@ -38,59 +44,185 @@ pub enum EarlyNetworkSetupError {
...
     #[error("Error during DNS lookup: {0}")]
-    DnsResolver(#[from] internal_dns::resolver::ResolveError),
+    DnsResolver(#[from] ResolveError),
 }
 
 /// Code for configuring the necessary network bits to bring up the control
 /// plane
-pub struct EarlyNetworkSetup {
-    log: Logger,
+pub struct EarlyNetworkSetup<'a> {
```
andrewjstone, line 52: Seems a touch overkill to parameterize on a lifetime just for a logger.

sled-agent/src/bootstrap/early_networking.rs
```diff
@@ -38,59 +44,185 @@ pub enum EarlyNetworkSetupError {
...
+        let switch_zone_addrs = self.lookup_switch_zone_addrs(resolver).await;
+        let mut boundary_switch_addrs = HashSet::new();
+
+        for uplink_config in &config.uplinks {
+            if let Some(&zone_addr) =
+                switch_zone_addrs.get(&uplink_config.switch)
+            {
+                boundary_switch_addrs.insert(zone_addr);
+            } else {
+                // TODO-correctness is it okay to just return here if the config
```
andrewjstone, line 75: I think this is the right thing to do in this scenario. It's not a programmer error. It's configuration error, which could even be stale or the switch may not have come online yet, right?

sled-agent/src/rack_setup/service.rs
```diff
@@ -867,6 +859,33 @@ impl ServiceInner {
                 .await?;
         }
 
+        // Save the relevant network config in the bootstore. We want this to
```
andrewjstone, line 862: 🎉

sled-agent/src/rack_setup/service.rs
```diff
@@ -867,6 +859,33 @@ impl ServiceInner {
...
+                generation: 1,
+                rack_subnet: config.rack_subnet,
+                ntp_servers: config.ntp_servers.clone(),
+                rack_network_config: rack_network_config.clone(),
+            };
+            info!(self.log, "Writing Rack Network Configuration to bootstore");
+            bootstore
+                .update_network_config(early_network_config.into())
+                .await?;
+            // TODO: Wait for bootstore to sync to all `plan.sleds`? Every
```
andrewjstone, line 877: One thing I just realized here is that without `config.rack_network_config` the sled-agent loop that waits for the bootstore network config will never complete. To satisfy this condition, we'll need to either wait for bootstore to sync to all plan.sleds or wrap the `EarlyNetworkConfig` in an `Option` so we don't use it.

I'm not really sure why we do this check here either? Do we always expect to have a `rack_network_config` form this point on? Can we just go ahead and remove the conditional? Does this work with single gimlets?

If we remove it, we definitely will not need to do the wait for sync here, and I'd prefer not to add that code to the bootstore right now.

CC @internet-diglett @smklein @rcgoodfellow

sled-agent/src/bootstrap/early_networking.rs
jgallagher: Eh, maybe, but nobody actually stores one of these things; it's really just a way to namespace some functions. I also considered just removing `EarlyNetworkSetup` altogether, and making its methods free functions that take `&Logger`s. Would you prefer that?

andrewjstone: Either way is fine with me. I made it a type originally because I thought it was going to get stored in bootstrap agent as a minimum. Feel free to leave it as is. I should have marked it `nit`

sled-agent/src/bootstrap/early_networking.rs
jgallagher: Hmmm! Now that you mention it, I think this function (more specifically, `map_switch_zone_addrs` which is called by `lookup_switch_zone_addrs`) will hang forever if one of the switches is offline, because it tries to resolve every switch zone IP returned by internal DNS, which should include both of them even if one of them is offline.

I'm not sure what the right thing to do here is to handle that.

sled-agent/src/rack_setup/service.rs
jgallagher: Hm, right, if non-racked gimlet environments run without a `rack_network_config` we should change sled-agent's check for sure. Should we store a `NetworkConfig` with an `Option<RackNetworkConfig>` inside bootstore to tell the difference between "no network config sync'd yet" and "we intentionally sync'd no network config" (e.g., for bench gimlets)?

andrewjstone: Yeah, I think storing an `Option<RackNetworkConfig>` is the right thing to do in this case.

rcgoodfellow: IIUC this code is mostly to support development environments where a configuration file is provided directly to the sled agent.

jgallagher: Switched to storing `Option<RackNetworkConfig>` to continue to support dev envs with no uplinks in [298acb4](https://github.com/oxidecomputer/omicron/pull/3765/commits/298acb4b495540056415a68bea340c223192b10a)

sled-agent/src/rack_setup/service.rs
```diff
@@ -867,6 +859,23 @@ impl ServiceInner {
...
+        // from the bootstore".
+        let early_network_config = EarlyNetworkConfig {
+            generation: 1,
+            rack_subnet: config.rack_subnet,
+            ntp_servers: config.ntp_servers.clone(),
+            rack_network_config: config.rack_network_config.clone(),
+        };
+        info!(self.log, "Writing Rack Network Configuration to bootstore");
+        bootstore.update_network_config(early_network_config.into()).await?;
+        // TODO: Wait for bootstore to sync to all `plan.sleds`? Every
```
andrewjstone, line 874: I'd remove this comment since we poll in sled-agent start *and* we have an option for deployments without `rack_network_config`

sled-agent/src/sled_agent.rs
```diff
@@ -352,41 +353,44 @@ impl SledAgent {
...
-            let early_network_config: EarlyNetworkConfig = serialized_config
-                .try_into()
-                .map_err(|e| Error::EarlyNetworkDeserialize(e))?;
-            info!(
-                log,
-                "Successfully deserialized network config: {:?}",
-                early_network_config
-            );
+            let early_network_config =
+                EarlyNetworkConfig::try_from(serialized_config)
```
andrewjstone, line 371: This is interesting. It will perpetually retry until it can successfully deserialize. That is useful property when the code has been updated, but the new data has not been replicated yet. It would also have the effect of possibly never succeeding, but I guess that's fine also as the error message should be indicative of the problem.

sled-agent/src/rack_setup/service.rs
internet-diglett: > IIUC this code is mostly to support development environments where a configuration file is provided directly to the sled agent.

Yep, that and some of the tests IIRC.

sled-agent/src/bootstrap/early_networking.rs
jgallagher: Attempting to lay out the problem explicitly, working backwards from where we end up using this information (the first few / lowest layers of which are unchanged on this PR, but I think already problematic):

1. In `ServiceManager::initialize_zone()`, we [call ServiceManager::opte_ports_needed()](https://github.com/oxidecomputer/omicron/blob/cold-boot-rss-uplink-config/sled-agent/src/services.rs#L1072).
2. In `ServiceManager::opte_ports_needed()`, we [build a list of DPD clients](https://github.com/oxidecomputer/omicron/blob/cold-boot-rss-uplink-config/sled-agent/src/services.rs#L875-L889), one for each IP addr in `request.boundary_switches`. (I'll come back to how `boundary_switches` gets populated below).
3. Still in `ServiceManager::opte_ports_needed()`, we send an [`ensure_nat_entry()`](https://github.com/oxidecomputer/omicron/blob/cold-boot-rss-uplink-config/sled-agent/src/services.rs#L936) request via each DPD client. This is wrapped in a [`backoff::retry_notify()`](https://github.com/oxidecomputer/omicron/blob/cold-boot-rss-uplink-config/sled-agent/src/services.rs#L959-L964), so `opte_ports_needed()` (and therefore `initialize_zone()`) will not return until it has successfully told DPD at all of the `request.boundary_switches` addresses about the NAT entry it needs.

IIUC, this means that starting any [external DNS, nexus, or NTP zone](https://github.com/oxidecomputer/omicron/blob/cold-boot-rss-uplink-config/sled-agent/src/services.rs#L864-L870) will block until it is able to talk to DPD on every boundary switch.

How do we populate the boundary switches? This is one of the major intended changes from #3729: on #3729, the boundary switch addresses are determined once at RSS time, and then saved in the bootstore. Absent any updates to the bootstore, that would mean any attempt to start an  external DNS/nexus/NTP zone would block until it successfully talks to DPD at the IP addresses of the switch zones that were present at RSS time. On this PR, we no longer store the boundary switch addresses in the bootstore; instead, we look them up whenever we need them:

1. During RSS, we look up the switch zone addresses both on:
  a. [first RSS run](https://github.com/oxidecomputer/omicron/blob/cold-boot-rss-uplink-config/sled-agent/src/rack_setup/service.rs#L922-L925), so we can [set `boundary_switches` on the initial zone requests](https://github.com/oxidecomputer/omicron/blob/cold-boot-rss-uplink-config/sled-agent/src/rack_setup/service.rs#L956-L958) and [send them to nexus](https://github.com/oxidecomputer/omicron/blob/cold-boot-rss-uplink-config/sled-agent/src/rack_setup/service.rs#L1024) (I have not looked into what Nexus does with these at all - it may warrant more investigation)
  b. [subsequent RSS runs that have already been applied](https://github.com/oxidecomputer/omicron/blob/cold-boot-rss-uplink-config/sled-agent/src/rack_setup/service.rs#L783-L785), in this case soley to [send them to nexus](https://github.com/oxidecomputer/omicron/blob/cold-boot-rss-uplink-config/sled-agent/src/rack_setup/service.rs#L794)
2. Any time sled-agent cold boots, it calls [load_services()](https://github.com/oxidecomputer/omicron/blob/cold-boot-rss-uplink-config/sled-agent/src/sled_agent.rs#L432-L440) to restart any services it's responsible for. This starts internal DNS first (if the sled-agent is responsible for internal DNS), then [looks up boundary switches](https://github.com/oxidecomputer/omicron/blob/cold-boot-rss-uplink-config/sled-agent/src/services.rs#L591-L607) so it can set `boundary_switches` for subsequent external DNS/nexus/NTP zones.
3. Any time sled-agent receives a request to start new services, it eventually lands in `ensure_all_services_persistent` which also [looks up boundary switches](https://github.com/oxidecomputer/omicron/blob/cold-boot-rss-uplink-config/sled-agent/src/services.rs#L2531-L2555) so it can set `boundary_switches` for any of the new zones it needs to start as part of the incoming request.

In all of these cases where we need to dynamically look up the boundary switches, we land in this early networking module and do the following:

1. Look up all dendrite services in internal DNS; this happens inside a backoff::retry and therefore cannot fail (but can block indefinitely if internal DNS is unavailable).
2. For every IPv6 addr returned by internal DNS in step 1, we ask MGS at that same address which switch it is. This happens inside an infinite retry loop and therefore cannot fail (but can block indefinitely if MGS is unavilable at any of these IP addresses).
3. In the non-RSS cases, we then filter the switch addresses down to just those that have uplinks present, based on the config provided at RSS time (which is stored in the bootstore). This cannot fail and does not block (we've already read the bootstore on sled-agent startup).

These latter three steps are technically infallible, but there are a bunch of ways they could block indefinitely, including at least:

* Internal DNS has stale addresses for the switch zones (we'd block forever trying to talk to a nonexistent MGS)
* One of the switches is down (we'd block until it comes back and we can talk to its MGS)

If we assume internal DNS is not stale _and_ it returns exactly two IP addresses, we technically only need a response from one: we could assume if the first one we hear back from is (wlog) switch 0, the other IP address must be switch 1, but that seems dicey:

* Assuming internal DNS is not stale without confirming we can actually talk to MGS at that address seems iffy
* If only one of the switch zones is up, even if we make assumptions that let us return, service startup is still going to block trying to talk to DPD at all the addresses we return

rcgoodfellow: > Internal DNS has stale addresses for the switch zones (we'd block forever trying to talk to a nonexistent MGS)

In a loop attempting to contact a switch zone, we should re-resolve each time to make things work in an eventually consistent setting.

> One of the switches is down (we'd block until it comes back and we can talk to its MGS)

This is where reconciliation semantics fit in quite well. We need to continuously try to drive toward the desired state of the world (2 switches with NAT entries in this case) without necessarily blocking a partially desired but still functional state (1 switch with NAT entries) from being realized.

jgallagher: I'm resolving this conversation based on the in-person discussion; will post a separate followup with changes.

jgallagher:
Changes added after the in-person discussion:

* [88a9661](https://github.com/oxidecomputer/omicron/pull/3765/commits/88a9661fed211321e6e405e5ac49208c645daffa) removes `boundary_switches` from `ServiceZoneRequest` altogether, and we only try to query for the boundary switches when we're starting one of the service types that needs it (external DNS, nexus, ntp)
* [3f7c048](https://github.com/oxidecomputer/omicron/pull/3765/commits/3f7c048c1b3e762b3f733083f323c394fc5f8080) continues to block to find all switches, but will give up after 5 minutes and return as soon after that deadline as it finds one
* [7096d17](https://github.com/oxidecomputer/omicron/pull/3765/commits/7096d17a637078824bccad229c7229eed8f2e47f) clears our DNS resolver cache before each query for `Dendrite` services

I still need to update the dendrite link for this to be testable; I think that's waiting on https://github.com/oxidecomputer/dendrite/pull/548 passing CI.

sled-agent/src/bootstrap/early_networking.rs
```diff
@@ -38,59 +47,238 @@ pub enum EarlyNetworkSetupError {
...
+        // `lookup_switch_zone_addrs` does not return unless it finds at least
+        // one switch location, so the only way this could happen is if there
+        // are no uplinks configured at all, or there are only uplinks
+        // configured for one switch location and we found the other one. Both
+        // of those are misconfiguration issues that we cannot overcome here, so
+        // we will return what we have.
+        boundary_switch_addrs
+    }
+
+    pub async fn lookup_switch_zone_addrs(
```
andrewjstone, line 107: It may be clarifying to use `underlay_addrs` for these function name suffixes, rather than just `_addrs`. I realize this is mostly copied, but for the uninitiated it may not be clear that we are not looking up `bootstrap` addresses given this is in `EarlyNetworking`.

sled-agent/src/bootstrap/early_networking.rs
```diff
@@ -38,59 +47,238 @@ pub enum EarlyNetworkSetupError {
...
+        EarlyNetworkSetup { log }
+    }
+
+    pub async fn lookup_boundary_switch_addrs(
+        &self,
+        resolver: &DnsResolver,
+        config: &RackNetworkConfig,
+    ) -> HashSet<Ipv6Addr> {
+        let switch_zone_addrs = self.lookup_switch_zone_addrs(resolver).await;
+        let mut boundary_switch_addrs = HashSet::new();
```
andrewjstone, line 77: This may be just me, but I found myself confused by the name `boundary_switch_addrs`. I had to read the code to figure out that this meant: "the underlay address of a a switch zone for a switch which has an uplink".  I thought a better name would imply all those things, even though it might be a bit long. My best suggestion was `switch_zone_underlay_addr_with_uplink`, even though `switch_zone_underlay_addr_for_switch_with_uplink` is even clearer, but also insanely verbose.

After thinking about this a bit more, I think a block comment at the top of `EarlyNetworkSetup` that explains what it is intended for and defines terms like `boundary_switch` and `uplink` would be most appropriate. Then we can just keep `boundary_switch_addrs` as is. I'd be happy to help write that block comment tomorrow if desired as it's getting late here.
Further thoughts:
Does `boundary_switch` always mean a switch with an uplink? In the common, and preferred case, won't both switches be `boundary_switches` by these definitions?

sled-agent/src/bootstrap/early_networking.rs
```diff
@@ -38,59 +47,238 @@ pub enum EarlyNetworkSetupError {
...
+                // `uplink_config`. Just log a warning and move on.
+                warn!(
+                    self.log,
+                    "No switch zone address found for {:?}",
+                    uplink_config.switch,
+                );
+            }
+        }
+
+        // TODO-correctness: Could `boundary_switch_addrs` be _empty_ here?
```
andrewjstone, line 97: At this point, we also will not be able to sync time and bring up the rest of the control plane. I think it's worth an explicit error being passed upwards or an error log and a panic. Ideally we'd be able to show this in the UI, but for now I think the easiest thing to do is log an error and panic.

sled-agent/src/bootstrap/early_networking.rs
```diff
@@ -38,59 +47,238 @@ pub enum EarlyNetworkSetupError {
...
+                "HashMap keyed on `SwitchLocation` can have at most 2 entries"
+            ),
+        }
     }
 
-    // Initialize the rack network and return the boundary switch addresses to
-    // be injected into zone requests
-    pub async fn init_rack_network(
+    // Initialize a single switch via DPD.
+    pub async fn init_switch_config(
```
andrewjstone, line 254: This is a **much** better name!

sled-agent/src/bootstrap/early_networking.rs
```diff
@@ -38,59 +47,238 @@ pub enum EarlyNetworkSetupError {
...
         &mut self,
         rack_network_config: &RackNetworkConfig,
-        switch_mgmt_addrs: &HashMap<SwitchLocation, Ipv6Addr>,
-    ) -> Result<HashSet<Ipv6Addr>, EarlyNetworkSetupError> {
+        switch_ip: Ipv6Addr,
```
andrewjstone, line 257: `switch_zone_underlay_ip` ?

sled-agent/src/bootstrap/early_networking.rs
jgallagher: This isn't only used during RSS - it's also used any time any sled-agent starts an external DNS / Nexus / NTP service. So for example, we could have this very unlucky sequence:

1. Two switches are present at RSS
2. Only one switch has an uplink
3. We start RSS. NTP services start, and they all call this function - both switches are found, and we filter down to the one switch that has an uplink in this function
4. NTP sync succeeds
5. The one switch zone one connected to the switch with an uplink goes down
6. We try to start nexus / external DNS: we get to this function, only find one switch, and then filter it away because it doesn't have an uplink
7. We return an empty `boundary_switch_addrs`, so nexus and external DNS do not set up NATs (which they _can't_ do anyway, because the one switch with an uplink is down)

I think I have convinced myself the actual check should not be "we found at least one switch", but "we found at least one switch _with an uplink_", and we should block until we do? I'll make that change, and if you disagree I can back it out.

sled-agent/src/bootstrap/early_networking.rs
jgallagher: I did some various renaming to be more explicit (and added more comments) in 56f1e4f2, 592ba19a, and 152ddb86

sled-agent/src/rack_setup/service.rs
jgallagher: removed in 05fb1ded

sled-agent/src/bootstrap/early_networking.rs
rcgoodfellow: > In the common, and preferred case, won't both switches be boundary_switches by these definitions?
 
In a single rack setup, yes. In a multitrack setup, there may be a plethora of switches just functioning as internal transit between racks.

andrewjstone: > > In the common, and preferred case, won't both switches be boundary_switches by these definitions?
> 
> In a single rack setup, yes. In a multitrack setup, there may be a plethora of switches just functioning as internal transit between racks.

Aha. Thanks. Glad I asked.

sled-agent/src/bootstrap/early_networking.rs
andrewjstone: This change looks great! Thank you so much!

andrewjstone:
This supersedes and includes code from #3729. It fixes https://github.com/oxidecomputer/omicron/issues/3738

rcgoodfellow (approved):
This has been tested e2e on madrid and a virtual 2-switch/4-node cluster.

PR #3766, opened 2023-07-26
wicket: fix alignment after uplink_ip -> uplink_cidr rename

luqmana:
Small follow-up to #3755

image


andrewjstone (approved):
Nice. Thanks!

Issue #3767, opened 2023-07-26
rack init failed


leftwo:
During tonights dogfood mupdate, we ran into a problem with the final rack init stage.

The following sleds were all going to be part of the cluster, including both 14 and 16
```
Cubby 9 (BRM44220005, bootstrap address fdb0:a840:2504:357::1) 
Cubby 10 (BRM42220009, bootstrap address fdb0:a840:2504:3d4::1) 
Cubby 11 (BRM42220006, bootstrap address fdb0:a840:2504:191::1)
Cubby 12 (BRM42220057, bootstrap address fdb0:a840:2504:393::1)
Cubby 14 (BRM42220051, bootstrap address fdb0:a840:2504:195::1) 
Cubby 16 (BRM42220014, bootstrap address fdb0:a840:2504:352::1)
Cubby 17 (BRM42220017, bootstrap address fdb0:a840:2504:192::1)
Cubby 21 (BRM42220031, bootstrap address fdb0:a840:2504:353::1)
Cubby 23 (BRM42220016, bootstrap address fdb0:a840:2504:395::1)
Cubby 25 (BRM44220010, bootstrap address fdb0:a840:2504:354::1) 
Cubby 27 (BRM42220026, bootstrap address fdb0:a840:2504:351::1)
```

However, the final RSS setup step using wicketd reported `Initialization Failed`

The failed message from wicketd:
```
Status: Initialization Failed 
 Last initialization operation ID: 58b96a33-90ce-4e6e-8cf9-83a98248b387
 Message: Bootstore error: Fsm error: rack init timeout: unacked_peers: {Gimlet { 
 identifier: "BRM42220051", model: "913-0000019", revision: 6 }} 
```

BRM42220051 is sled 14

leftwo:
I'm starting to suspect a restart of sled-agent is the source of the trouble here.

Sled agent logs from 14 and 16 are on catacomb here: /data/staff/core/rack2/omicron-3767

On 14, at 00:50:41, the sled agent restarts:
```
[ Dec 28 00:50:41 Stopping because service disabled. ]
[ Dec 28 00:50:41 Executing stop method (:kill). ]
[ Dec 28 00:51:58 Enabled. ]
[ Dec 28 00:51:58 Executing start method ("ctrun -l child -o noorphan,regent /opt/oxide/sled-agent/sled-agent run /opt/oxide/sled-agent/pkg/config.toml &"). ]
[...]

PR #3771, opened 2023-07-26
Fix for stranded connection initiators

andrewjstone:
There was a bug where the peer checked to see if it was an initiator via an invalid check where it compared sorted `Baseboard`s, rather than comparing addresses. This resulted in all the `StaleXXX messages` and made it so that the peer fsm would not actually discover that it was connected on the initiating side because of an early return. This would result in retry messages not getting sent from the FSM. This is the most likely cause of the RSS failure on dogfood as error messages occur around rack init time.

While looking at the code I also noticed two other bugs, that I don't believe we've necessarily hit.  These are loss of tracking connections when an *actual* stale message is received.

### DIFF
--- a/bootstore/src/schemes/v0/peer.rs
+++ b/bootstore/src/schemes/v0/peer.rs
@@ -392,7 +392,19 @@ impl Node {
                     return;
                 };
                 // Remove any existing connection
-                self.remove_accepted_connection(&addr).await;
+                if let Some(handle) = self.accepted_connections.remove(&addr) {
+                    info!(
+                        self.log,
+                        concat!(
+                            "Removing acccepted connection from {}: ",
+                            "new connection accepted from same address"
+                        ),
+                        addr
+                    );
+
+                    // The connection has not yet completed its handshake
+                    let _ = handle.tx.send(MainToConnMsg::Close).await;
+                }
                 info!(self.log, "Accepted connection from {addr}");
                 self.handle_unique_id_counter += 1;
                 let handle = spawn_accepted_connection_management_task(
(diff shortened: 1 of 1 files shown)


leftwo:
Possibly fixes: https://github.com/oxidecomputer/omicron/issues/3767

bootstore/src/schemes/v0/peer.rs
```diff
@@ -765,9 +777,12 @@ impl Node {
                     return;
                 };
 
-                // Ignore the stale message if the unique_id doesn't match what
-                // we have stored.
+                // Put back the non-matching connection we removed
+                // The received message is stale, so we return.
                 if unique_id != accepted_handle.unique_id {
+                    self.accepted_connections
```
leftwo, line 783: What does it mean to put this stale connection back on the accepted_connections list?

leftwo (approved):
Dogfood tested, dogfood approved!

bootstore/src/schemes/v0/peer.rs
andrewjstone: TLDR: The incoming message is stale, but the connection we put back is always current. Otherwise it wouldn't be in the map.

We are actually putting back the original connection. `unique_id` here is from the received message on line 757 and essentially says:  "This message come from this tokio task". However, we remove the handle matching the the address in the message on line 765 *before* we do the check. `accepted_handle` holds the current unique_id. When we do the check on line 782 we realize that the message is actually not current, and so we don't want to progress the handle from `self.accepted_connections` into `self.established_connections`. But we don't want to discard our current handle just for a stale message, so we put it back where it was in `self.accepted_connections`. 

I could have instead borrowed to do the check and removed after the borrow as done elsewhere. Maybe that would have been clearer. All in all, I think this code is more confusing than it has to be and I plan to add some more tests and clean it up very soon.

PR #3782, opened 2023-07-27
Add file and line num to log messages

andrewjstone:
This commit adds a `slog::KV` implementation to `omicron_common` that can be plugged into any root slog logger to append file and line number information to all log entries.

An example log message looks like the following:

```
Jul 27 22:19:44.139 WARN Connection error: Close, remote_peer_id: pc-a-1, peer_id: pc-b-1, subcomponent: bootstore, file: bootstore/src/schemes/v0/peer_networking.rs:216
```

This logging support was then added to a bunch of existing servers.

### DIFF
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -50,3 +50,24 @@ macro_rules! generate_logging_api {
         );
     };
 }
+
+/// A type that allows adding file and line numbers to log messages
+/// automatically. It should be instantiated at the root logger of each
+/// executable that desires this functionality, as in the following example.
+/// ```
+///     slog::Logger::root(drain, o!(FileLocationKv))
+/// ```
+pub struct FileKv;
+
+impl slog::KV for FileKv {
+    fn serialize(
+        &self,
+        record: &slog::Record,
+        serializer: &mut dyn slog::Serializer,
+    ) -> slog::Result {
+        serializer.emit_str(
+            "file".into(),
+            &format!("{}:{}", record.file(), record.line()),
+        )
+    }
+}
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -55,7 +55,7 @@ macro_rules! generate_logging_api {
 /// automatically. It should be instantiated at the root logger of each
 /// executable that desires this functionality, as in the following example.
 /// ```ignore
-///     slog::Logger::root(drain, o!(FileLocationKv))
+///     slog::Logger::root(drain, o!(FileKv))
 /// ```
 pub struct FileKv;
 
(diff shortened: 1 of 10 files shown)


smklein (approved):
I'm a little surprised it ended up so simple, but this will be great to have.

common/src/lib.rs
```diff
@@ -50,3 +50,24 @@ macro_rules! generate_logging_api {
...
+impl slog::KV for FileKv {
+    fn serialize(
+        &self,
+        record: &slog::Record,
+        serializer: &mut dyn slog::Serializer,
+    ) -> slog::Result {
+        serializer.emit_str(
+            "file".into(),
+            &format!("{}:{}", record.file(), record.line()),
+        )
```
luqmana, line 71: Can use `emit_arguments` with `format_args!` to avoid extra heap allocaitons:

```suggestion
        serializer.emit_arguments(
            "file".into(),
            &format_args!("{}:{}", record.file(), record.line()),
        )
```

common/src/lib.rs
```diff
@@ -50,3 +50,24 @@ macro_rules! generate_logging_api {
...
+///     slog::Logger::root(drain, o!(FileLocationKv))
+/// ```
+pub struct FileKv;
+
+impl slog::KV for FileKv {
+    fn serialize(
+        &self,
+        record: &slog::Record,
+        serializer: &mut dyn slog::Serializer,
+    ) -> slog::Result {
```
luqmana, line 67: Does it make sense to limit this to `record.level() <= Level::Warning` to limit log bloat? There's a bunch of periodic `INFO`/`DEBUG` logs I think wouldn't really benefit from the file/line being included everytime?

andrewjstone: I think the info ones in particular are helpful to have line numbers, but I wouldn't mind turning this off for debug logs if necessary. Seems somewhat like a premature optimization, but I could totally see us rotating out useful information early just to keep line numbers when not needed, so I think the compromise is useful.

The reason I really wanted these is that a lot of times logs aren't unique, so having the info ones are particularly helpful. Debug logs can always add extra context.

common/src/lib.rs
```diff
@@ -50,3 +50,24 @@ macro_rules! generate_logging_api {
         );
     };
 }
+
+/// A type that allows adding file and line numbers to log messages
+/// automatically. It should be instantiated at the root logger of each
+/// executable that desires this functionality, as in the following example.
+/// ```
+///     slog::Logger::root(drain, o!(FileLocationKv))
```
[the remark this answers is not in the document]
andrewjstone: Good catch! Fixed.

common/src/lib.rs
andrewjstone: Another good one. Thanks!

common/src/lib.rs
andrewjstone: I filtered at info level in [0cf4013](https://github.com/oxidecomputer/omicron/pull/3782/commits/0cf4013d0705b61896c7675598e3b589c79ee3ff)

luqmana (approved):

Issue #3763, opened 2023-07-25
Nexus holds on to specific service backends of crdb and crucible-pantry for too long


askfongjojo:
While testing service failover on rack2, I noticed that nexus held on to the same cockroachdb backend without attempting to use any of the other nodes in the 5-node database cluster, causing requests to fail until the one it favored came back.

The same happened with pantry requests for disk import blocks / bulk writes. I haven't got to the point of seeing the TTL being exhausted. I tried waiting for up to 5 minutes and the request still couldn't succeed until the pantry zone Nexus has been using prior to its outage came up again.

smklein:
I think ignoring the TTL makes sense (not that it's right, just that it makes sense) in the context of this code:

https://github.com/oxidecomputer/omicron/blob/a29f08ba1c0d943101f6811436c6e6715b5ce252/nexus/src/context.rs#L167-L195

Our API to interface with CockroachDB from Nexus involves "constructing a URL", and right now, we embed a single point-in-time address of CockroachDB into this URL.

We should probably avoid doing this, and embed the service name into the URL, so it actually use DNS during later lookups.

askfongjojo:
Understood. The current behavior, in conjunction with #3613, results in an 1 in 15 probability that a user will not be able to use the system at all. I happened to be in this situation when my workstation always used the one nexus that had the first CRDB backend I brought down to test failover. :(

jmpesp:
> The same happened with pantry requests for disk import blocks / bulk writes. I haven't got to the point of seeing the TTL being exhausted. I tried waiting for up to 5 minutes and the request still couldn't succeed until the pantry zone Nexus has been using prior to its outage came up again.

In this case, there's no TTL: a specific pantry from the list of available ones is selected once and used for a disk until that disk is finalized.

One thing that could be done is to check of the selected pantry is still responsive and choose another one if not, but it was important not keep as much code out of the import chunk hot-path as possible, as any checks there will be multiplied by the number of chunks to import and slow down imports. That being said, slow imports are better than non-working imports :) I'll give this some thought.

askfongjojo:
Sorry for not being clear in pantry' case. I understand the need to stay with the same pantry for the same disk snapshot. But the issue I ran into is with new/separate snapshot requests still holding on to the unresponsive one.

luqmana:
[...]

PR #3783, opened 2023-07-27
nexus: use all CockroachDB hosts from DNS to create DB connection URL.

luqmana:
First pass at #3763 for crdb.

Even though we did query internal DNS, we were previously using only a single host as part of connecting to crdb from Nexus. And since the internal DNS server always returns records in the same order, that meant every Nexus instance was always using the same CockroachDB instance even now that we've been provisioning multiple. This also meant if that CRDB instance went down we'd be hosed (as seen in #3763).

To help with that, this PR changes Nexus to use all the CRDB hosts reported via Internal DNS when creating the connection URL. There are some comments in the code, but this still not quite as robust as we could be, but short of something cueball-like it's still an improvement.

To test I disabled the initial crdb nexus connected to and it was able to recover by connecting to the next crdb instance and continue serving requests. From the log we can see a successful query, connection errors once i disabled `fd00:1122:3344:101::5`, and then a successful query with connection reestablished to next crdb instance (`fd00:1122:3344:101::3`):
```
23:43:24.729Z DEBG 7be03b0d-48bf-4f43-a11e-7303236a3c5e (ServerContext): authorize result
    action = Query
    actor = Some(Actor::UserBuiltin { user_builtin_id: 001de000-05e4-4000-8000-000000000003, .. })
    resource = Database
    result = Ok(())
23:43:24.729Z ERRO 7be03b0d-48bf-4f43-a11e-7303236a3c5e (ServerContext): database connection error
    database_url = postgresql://root@[fd00:1122:3344:101::5]:32221,[fd00:1122:3344:101::3]:32221,[fd00:1122:3344:101::6]:32221,[fd00:1122:3344:101::4]:32221,[fd00:1122:3344:101::7]:32221/omicron?sslmode=d
isable
    error_message = Connection error: server is shutting down
23:43:24.729Z ERRO 7be03b0d-48bf-4f43-a11e-7303236a3c5e (ServerContext): database connection error
    database_url = postgresql://root@[fd00:1122:3344:101::5]:32221,[fd00:1122:3344:101::3]:32221,[fd00:1122:3344:101::6]:32221,[fd00:1122:3344:101::4]:32221,[fd00:1122:3344:101::7]:32221/omicron?sslmode=d
isable
    error_message = Connection error: server is shutting down
23:43:24.729Z ERRO 7be03b0d-48bf-4f43-a11e-7303236a3c5e (ServerContext): database connection error
    database_url = postgresql://root@[fd00:1122:3344:101::5]:32221,[fd00:1122:3344:101::3]:32221,[fd00:1122:3344:101::6]:32221,[fd00:1122:3344:101::4]:32221,[fd00:1122:3344:101::7]:32221/omicron?sslmode=d
isable
    error_message = Connection error: server is shutting down
23:43:24.729Z ERRO 7be03b0d-48bf-4f43-a11e-7303236a3c5e (ServerContext): database connection error
    database_url = postgresql://root@[fd00:1122:3344:101::5]:32221,[fd00:1122:3344:101::3]:32221,[fd00:1122:3344:101::6]:32221,[fd00:1122:3344:101::4]:32221,[fd00:1122:3344:101::7]:32221/omicron?sslmode=d
isable
    error_message = Connection error: server is shutting down
23:43:24.729Z ERRO 7be03b0d-48bf-4f43-a11e-7303236a3c5e (ServerContext): database connection error
    database_url = postgresql://root@[fd00:1122:3344:101::5]:32221,[fd00:1122:3344:101::3]:32221,[fd00:1122:3344:101::6]:32221,[fd00:1122:3344:101::4]:32221,[fd00:1122:3344:101::7]:32221/omicron?sslmode=d
isable
    error_message = Connection error: server is shutting down
23:43:24.729Z ERRO 7be03b0d-48bf-4f43-a11e-7303236a3c5e (ServerContext): database connection error
    database_url = postgresql://root@[fd00:1122:3344:101::5]:32221,[fd00:1122:3344:101::3]:32221,[fd00:1122:3344:101::6]:32221,[fd00:1122:3344:101::4]:32221,[fd00:1122:3344:101::7]:32221/omicron?sslmode=d
isable
    error_message = Connection error: server is shutting down
23:43:24.730Z ERRO 7be03b0d-48bf-4f43-a11e-7303236a3c5e (ServerContext): database connection error
    database_url = postgresql://root@[fd00:1122:3344:101::5]:32221,[fd00:1122:3344:101::3]:32221,[fd00:1122:3344:101::6]:32221,[fd00:1122:3344:101::4]:32221,[fd00:1122:3344:101::7]:32221/omicron?sslmode=d
isable
    error_message = Connection error: server is shutting down
23:43:24.730Z ERRO 7be03b0d-48bf-4f43-a11e-7303236a3c5e (ServerContext): database connection error
    database_url = postgresql://root@[fd00:1122:3344:101::5]:32221,[fd00:1122:3344:101::3]:32221,[fd00:1122:3344:101::6]:32221,[fd00:1122:3344:101::4]:32221,[fd00:1122:3344:101::7]:32221/omicron?sslmode=d
isable
    error_message = Connection error: server is shutting down
23:43:30.803Z DEBG 7be03b0d-48bf-4f43-a11e-7303236a3c5e (ServerContext): roles
    roles = RoleSet { roles: {} }
23:43:30.804Z DEBG 7be03b0d-48bf-4f43-a11e-7303236a3c5e (ServerContext): authorize result
    action = Query
    actor = Some(Actor::UserBuiltin { user_builtin_id: 001de000-05e4-4000-8000-000000000003, .. })
    resource = Database
    result = Ok(())
```

### DIFF
--- a/nexus/src/context.rs
+++ b/nexus/src/context.rs
@@ -168,16 +168,21 @@ impl ServerContext {
             nexus_config::Database::FromUrl { url } => url.clone(),
             nexus_config::Database::FromDns => {
                 info!(log, "Accessing DB url from DNS");
-                let address = loop {
+                // It's been requested but unfortunately not supported to directly
+                // connect using SRV based lookup.
+                // TODO-robustness: the set of cockroachdb hosts we'll use will be
+                // fixed to whatever we got back from DNS at Nexus start. This means
+                // a new cockroachdb instance won't picked up until Nexus restarts.
+                let addrs = loop {
                     match resolver
-                        .lookup_socket_v6(ServiceName::Cockroach)
+                        .lookup_all_socket_v6(ServiceName::Cockroach)
                         .await
                     {
-                        Ok(address) => break address,
+                        Ok(addrs) => break addrs,
                         Err(e) => {
                             warn!(
                                 log,
-                                "Failed to lookup cockroach address: {e}"
+                                "Failed to lookup cockroach addresses: {e}"
                             );
                             tokio::time::sleep(std::time::Duration::from_secs(
                                 1,
@@ -186,9 +191,14 @@ impl ServerContext {
                         }
                     }
                 };
-                info!(log, "DB address: {}", address);
+                let addrs_str = addrs
+                    .iter()
+                    .map(ToString::to_string)
+                    .collect::<Vec<_>>()
+                    .join(",");
+                info!(log, "DB addresses: {}", addrs_str);
                 PostgresConfigWithUrl::from_str(&format!(
-                    "postgresql://root@{address}/omicron?sslmode=disable",
+                    "postgresql://root@{addrs_str}/omicron?sslmode=disable",
                 ))
                 .map_err(|e| format!("Cannot parse Postgres URL: {}", e))?
             }


nexus/src/context.rs
```diff
@@ -168,27 +168,34 @@ impl ServerContext {
...
-                let address = loop {
+                // It's been requested but unfortunately not supported to directly
+                // connect using SRV based lookup.
+                // TODO-robustness: the set of cockroachdb hosts we'll use will be
+                // fixed to whatever we got back from DNS at Nexus start. This means
+                // a new cockroachdb instance won't picked up until Nexus restarts.
+                let addrs = loop {
                     match resolver
-                        .lookup_socket_v6(ServiceName::Cockroach)
+                        .lookup_all_socket_v6(ServiceName::Cockroach)
```
luqmana, line 178: The last commit `[be1b1b1](https://github.com/oxidecomputer/omicron/pull/3783/commits/be1b1b14ad49649c463b138b2c5927e56fad9ee8) changed this to doing the lookup in nexus vs using the `SRV` targets and letting libpq or whoever do it at connection time. Going a little back and forth on both options. (The explicit option would make the initial nameserver/resolv.conf stuff in the nexus zone moot too).

smklein: If we had a situation where all the CRDB servers rotated, but Nexus stayed online, this would eventually run out of the ability to contact cockroach services, right?

I have a preference for "whatever solution makes Nexus rely on hardcoded addresses" -- whether or not that's "right away" or "after everyone in the set at a particular point-in-time evacuates".

If we left this up to libpq to do name resolution, wouldn't we be able to continue accessing CRDB nodes, even if they **all* moved to new locations?

dns-server/src/dns_server.rs
```diff
@@ -227,55 +270,70 @@ async fn handle_dns_message(
...
+
+            // DNS allows for the server to return additional records
+            // that weren't explicitly asked for by the client but that
+            // the server expects the client will want. The records
+            // corresponding to a lookup on a SRV target is one such case.
+            // We opportunistically attempt to resolve the target here
+            // and if successful return those additional records in the
+            // response.
+            // NOTE: we only do this one-layer deep.
+            if let Some(RData::SRV(srv)) = record.data() {
```
smklein, line 287: I see you already pulled this into https://github.com/oxidecomputer/omicron/pull/3785/ , but out of curiosity, does it change the behavior of the rest of this PR? Are these additional records actually being read anywhere?

internal-dns/src/resolver.rs
```diff
@@ -287,6 +276,55 @@ impl Resolver {
...
+                log,
+                "lookup_all_socket_v6: looking up SRV target";
+                "name" => ?target,
+            );
+            self.resolver
+                .ipv6_lookup(target.clone())
+                .await
+                .map(|ips| (ips, port))
+        });
+        let results = futures::future::try_join_all(futures).await?;
```
smklein, line 314: Are we sure `self.resolver.ipv6_lookup` is cancel-safe?

Related: Do we actually *want* to exit early here? Or would we want to return all the futures that can return something successful? 

(Just comparing the usage here with https://docs.rs/cancel-safe-futures/0.1.2/cancel_safe_futures/macro.join_then_try.html )

sled-agent/src/services.rs
```diff
@@ -927,11 +927,34 @@ impl ServiceManager {
...
     }
 
+    /// Enable DNS resolution using the internal DNS servers for the given zone.
+    ///
+    /// The Internal DNS servers are allocated on a reserved AZ-wide subnet which we can
+    /// determine from the given Sled underlay address.
+    ///
+    /// For more details on this "reserved" rack subnet, refer to
+    /// [omicron_common::address::ReservedRackSubnet].
+    async fn configure_internal_dns_in_zone(
```
smklein, line 937: This could be a boon for debugging internal services, even if we choose to rely on a non-host resolver at times.

nexus/src/context.rs
luqmana: >If we had a situation where all the CRDB servers rotated, but Nexus stayed online, this would eventually run out of the ability to contact cockroach services, right?

i think both approaches as-is are still susceptible to that. Using either addresses directly or hostnames like `{UUID}.host.control-plane.oxide.internal` that are fixed at Nexus startup will stop working if all those CRDB nodes get cycled out.

Originally was hoping to use the `_cockroach._tcp.control-plane.oxide.internal` SRV record which would work as long as internal-dns was consistent with the set of servers to use but that's not possible as-is. But I wonder if we could do this at the pool-level with a custom `bb8::ManageConnection::connect` impl that resolves the addrs there.

dns-server/src/dns_server.rs
luqmana: No, as-is there are some bugs in `trust-dns` that block that, hence why I just pulled it out of this PR. I need to finish rebasing this one.

internal-dns/src/resolver.rs
luqmana: Good catch! This was basically a counterpart to the existing `lookup_all_ipv6` method and based on the comment we definitely want all the ones that are successful:
```
        // What do we do if some of these queries succeed while others fail?  We
        // may have some addresses, but the list might be incomplete.  That
        // might be okay for some use cases but not others.  For now, we do the
        // simple thing.  In the future, we'll want a more cueball-like resolver
        // interface that better deals with these cases.
```

sled-agent/src/services.rs
luqmana: Yea, but perhaps this specific method wouldn't be as useful when everything has moved to be self-assembling.

nexus/src/context.rs
```diff
@@ -168,27 +168,34 @@ impl ServerContext {
...
-                info!(log, "DB address: {}", address);
+                let addrs_str = addrs
+                    .iter()
+                    .map(ToString::to_string)
+                    .collect::<Vec<_>>()
+                    .join(",");
+                info!(log, "DB addresses: {addrs_str}");
                 PostgresConfigWithUrl::from_str(&format!(
-                    "postgresql://root@{address}/omicron?sslmode=disable",
+                    "postgresql://root@{addrs_str}/omicron?sslmode=disable",
```
davepacheco, line 198: It is wild (but neat) that you can put more than one host here:
https://docs.rs/tokio-postgres/0.7.6/tokio_postgres/config/struct.Config.html#url

sled-agent/src/services.rs
```diff
@@ -927,11 +927,34 @@ impl ServiceManager {
...
+    ) -> Result<(), Error> {
+        let az_services_subnet =
+            ReservedRackSubnet::new(Ipv6Subnet::<AZ_PREFIX>::new(sled_addr));
+        let dns_servers = az_services_subnet
+            .get_dns_subnets()
+            .into_iter()
+            .map(|dns_subnet| dns_subnet.dns_address().ip().into())
+            .collect::<Vec<_>>();
+
+        self.configure_dns_client(running_zone, &dns_servers, None).await
```
davepacheco, line 950: I think it's great to set up /etc/resolv.conf in most zones.  But I think we don't want to use `configure_dns_client()` to do it.  I summarized here:
https://github.com/oxidecomputer/omicron/issues/2122#issuecomment-1611870486

I think what we did for the CockroachDB zone is a more robust approach:
https://github.com/oxidecomputer/omicron/blob/b514ef8b8659fb043b804c88a827bf321e3cb58b/sled-agent/src/services.rs#L1100-L1130

I am not sure if that's going to require the other bits mentioned in https://github.com/oxidecomputer/omicron/issues/2122#issuecomment-1611889979 though.

internal-dns/src/resolver.rs
luqmana: Moved that logic out to a separate helper and fixed in #3785.

nexus/src/context.rs
luqmana: Will do follow-up changes to more intelligently handle this and leave this PR as an incremental improvement by just passing all the crdb hosts instead of just one.

smklein (approved):

PR #3784, opened 2023-07-28
Change ledger file extensions to .json

### LINKED ISSUE #3723: Convert all the ledger file extensions to `.json`
Most of the ledger files are currently named `*.toml` even though the ledger format is now JSON. 

There are also some related comments that need cleaning up.

andrewjstone:
Ledger files are now serialized as json, and not toml. Change the extension to reflect this.

Fixes #3723

### DIFF
--- a/sled-agent/src/rack_setup/plan/service.rs
+++ b/sled-agent/src/rack_setup/plan/service.rs
@@ -114,7 +114,7 @@ impl Ledgerable for Plan {
     }
     fn generation_bump(&mut self) {}
 }
-const RSS_SERVICE_PLAN_FILENAME: &str = "rss-service-plan.toml";
+const RSS_SERVICE_PLAN_FILENAME: &str = "rss-service-plan.json";
 
 impl Plan {
     pub async fn load(
--- a/sled-agent/src/bootstrap/agent.rs
+++ b/sled-agent/src/bootstrap/agent.rs
@@ -227,7 +227,7 @@ pub struct Agent {
     bootstore_peer_update_handle: JoinHandle<()>,
 }
 
-const SLED_AGENT_REQUEST_FILE: &str = "sled-agent-request.toml";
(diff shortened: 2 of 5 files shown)


sled-agent/src/rack_setup/plan/service.rs
```diff
@@ -114,7 +114,7 @@ impl Ledgerable for Plan {
     }
     fn generation_bump(&mut self) {}
 }
-const RSS_SERVICE_PLAN_FILENAME: &str = "rss-service-plan.toml";
+const RSS_SERVICE_PLAN_FILENAME: &str = "rss-service-plan.json";
```
smklein, line 117: Can you grep for these names? I see some remaining references to the old toml versions of them:

- `rss-sled-plan.toml` is referenced on line 19 of `src/rack_setup/service.rs-
- `rss-service-plan.toml` is referenced on line 20 of `src/rack_setup/service.rs`

Also, it looks like `SLED_AGENT_REQUEST_FILE` is now converted to be using JSON with your changes to the ledger, but I think it's also still labelled as a ".toml" file

andrewjstone: Good catch! Thanks. I tried to find ledgers by searching for all `Ledgerable` types. I didn't consider the file names being listed in comments. I also failed to consider the fact that different files could contain the same `Ledgerable` types. A better search, besides looking at every `.toml` which is misleading seems to be `searching for `all_m2_mountpoints`. In any case, I'll clean this up.

andrewjstone: It seems like you found all the outstanding issues. I noticed that zone service bundles don't use the ledger which probably makes sense since we plan to put them on U.2s.

Fixed in 73ab80269c04b3fdeaef646aa25818f080662bc0

smklein (approved):

PR #3787, opened 2023-07-28
[sled-agent] add a sled_shutdown API on the bootstrap interface

sunshowers:
Some notes:

* This API is part of the work on #3718.
* We halt zones rather than doing both a halt and a remove. We
  definitely don't want to delete zone configuration, and the choice
  between halt and halt + remove is pretty arbitrary as I understand it
  (sled-agent will take care of removals at next startup).
* This uses the new `coop_cancel` adapter I added to
  cancel-safe-futures: https://oxidecomputer.github.io/cancel-safe-futures/rustdoc/cancel_safe_futures/coop_cancel.
  It does its job and works great :)
* Still need to test this.


sled-agent/src/bootstrap/http_entrypoints.rs
```diff
@@ -168,3 +169,23 @@ async fn sled_reset(
     ba.sled_reset().await.map_err(|e| Error::from(e))?;
     Ok(HttpResponseUpdatedNoContent())
 }
+
+#[derive(Debug, Deserialize, JsonSchema)]
+pub struct SledShutdownRequest {
+    /// The reason for the shutdown.
```
smklein, line 175: Smart to include this. I know the number of callers here is currently "zero", but we'll want this soon, I bet.

sled-agent/src/bootstrap/http_entrypoints.rs
```diff
@@ -168,3 +169,23 @@ async fn sled_reset(
...
+    pub reason: String,
+}
+
+/// Shuts down the sled, deleting all zones but not user data.
+#[endpoint {
+    method = POST,
+    path = "/sled-shutdown",
+}]
+async fn sled_shutdown(
```
smklein, line 184: Should we be more clear with naming here, and in the methods we call?

The implementation, as you know, doesn't actually shutdown the sled. It puts all control plane state into a "disabled" state. Maybe "sled_agent_shutdown" would be a more apt name?

I just don't want anyone to think this implies anything about the sled as a whole, or about power states.

smklein (approved):
Looks great, and the cooperative canceler makes a lot of sense to me!

Is the next step calling this from wicket?

sunshowers:
Per https://github.com/oxidecomputer/omicron/issues/3718#issuecomment-1656522843, moving this back into draft while we figure out the right approach.

sunshowers:
Way out of date :)

PR #3797, opened 2023-07-29
Switch to nextest

### LINKED ISSUE #3683: Switch omicron to using nextest
- [x] Set up nextest configuration with CI profile (in #3797)
- [x] ~~Move GH Actions jobs to nextest~~ (there are no GH Actions jobs that run tests)
- [x] Move Buildomat jobs to nextest (in #3797)
- [x] Add documentation about nextest

@iliana already has a WIP branch (https://github.com/oxidecomputer/omicron/compare/main...iliana/nextest)

sunshowers:
Nextest's features include:

* Quicker test runs (depends on the number of cores and threads, I measured a 2.2x speedup a while back on my 16 core/32 thread Linux box.)
* Better error reporting.
* Slow and leaky test reporting (eg it looks like nextest already identified a leaky test).
* Test retries (with backoff and jitter) if we ever need them.
* Lots more, check out https://nexte.st
* Most importantly in many ways, leverage to add new features as and when we need them.

I'm the primary author of the project so of course I'm biased here, but I think nextest will be pretty useful to us :)

Closes #3683.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,0 +1,2 @@
+[profile.ci]
+fail-fast = false
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,6 +1,6 @@
 # NOTE: The recommended version should be in sync with the version in
 # .config/nextest.toml. Bump up the required version as necessary.
-nextest-version = { required = "0.9.55", recommended = "0.9.56" }
+nextest-version = { required = "0.9.55", recommended = "0.9.57" }
 
 [profile.ci]
 fail-fast = false
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,4 +1,6 @@
-nextest-version = "0.9.55"
+# NOTE: The recommended version should be in sync with the version in
+# .config/nextest.toml. Bump up the required version as necessary.
+nextest-version = { required = "0.9.55", recommended = "0.9.56" }
 
 [profile.ci]
 fail-fast = false
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,2 +1,4 @@
+nextest-version = "0.9.55"
+
 [profile.ci]
 fail-fast = false
--- a/README.adoc
+++ b/README.adoc
@@ -43,6 +43,21 @@ This mode of operation will be used in production.
 
 To build and run the non-simulated version of Omicron, see: xref:docs/how-to-run.adoc[].
 
+== Run tests with nextest
+
+The supported way to run tests is via https://nexte.st/[cargo-nextest].
+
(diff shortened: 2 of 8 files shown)


davepacheco:
When we're ready to do this, can you update the docs that advise people to use `cargo test`?  It might also be worth calling out explicitly in https://github.com/oxidecomputer/omicron#build-and-run or https://github.com/oxidecomputer/omicron#working-in-omicron.  Basically I just think it's important to communicate expectations to developers about what's expected to work (and what's not).

sunshowers:
@davepacheco yep, updated documentation. I stopped short of saying "don't use `cargo test`, it is no longer supported", though maybe if CI only supports nextest we'll quickly hit that state.

.config/nextest.toml
```diff
@@ -0,0 +1,4 @@
+nextest-version = "0.9.55"
```
sunshowers, line 1: Q, should this stay in sync with the version used in `build-and-test.sh`? Nextest now lets you set both required and recommended versions, maybe this should match the recommended version? https://nexte.st/book/minimum-versions.html

README.adoc
```diff
@@ -43,6 +43,21 @@ This mode of operation will be used in production.
 
 To build and run the non-simulated version of Omicron, see: xref:docs/how-to-run.adoc[].
 
+== Run tests with nextest
+
+The supported way to run tests is via https://nexte.st/[cargo-nextest].
```
david-crespo, line 48: I like the idea of heading off the inevitable cargo test question by saying something like 

> `cargo test` may work, but we can't guarantee that as we do not run it in CI.

README.adoc
```diff
@@ -43,6 +43,21 @@ This mode of operation will be used in production.
...
+
+If you don't already have nextest installed, get started by https://nexte.st/book/pre-built-binaries[downloading a pre-built binary] or installing nextest via your package manager. Nextest has pre-built binaries for Linux, macOS and illumos.
+
+Then, run tests with:
+
+[source,text]
+----
+$ cargo nextest run
+----
```
david-crespo, line 57: It would be redundant with the nextest docs, but I know my first question here would be how do I run the tests for one crate, `cargo nextest run -p omicron-nexus`.

.config/nextest.toml
david-crespo: How would you enforce the sync?

david-crespo: Maybe the run script could actually pull the version from here?

README.adoc
sunshowers: Done, I think that makes sense.

.config/nextest.toml
sunshowers: Just via a comment is what I was thinking.

sunshowers: Put in a comment re this.

sunshowers:
(version 0.9.56 is bad though it doesn't affect us, should use 0.9.57)

PR #3800, opened 2023-07-29
Queue disks for upsert when a transient error occurs

### LINKED ISSUE #3789: sled-agent cold boot: rack secret timeout
While cold boot testing `madrid`, we hit this error on the scrimlet sled-agent:

```
00:01:24.571Z WARN SledAgent (BootstrapAgent): Could not ensure partitions: KeyManager error: Failed to retreive secret: Bootstore error: Fsm error: rack secret load timeout
00:01:24.571Z WARN SledAgent (BootstrapAgent): StorageWorker encountered unexpected error: KeyManager error: Failed to retreive secret: Bootstore error: Fsm error: rack secret load timeout
```

After chatting live, @andrewjstone pointed out that we do not have any retries here, and we need them: not all sleds will boot at the same time, so the early ones will need to retry until they're able to load the rack secret.

The full logs from this are at `/net/catacomb/data/staff/core/madrid/omicron-3789`: note that these logs show _2_ sled agent bounces; I'll open separate issues for each of those.

andrewjstone:
Check every 5 seconds if there are any queued disks and we have a ready
key manager, meaning that this replica has a key share or it is using
the `HardcodedKeyRetriever`. If there are disks that need upserting then
try to do that. This allows not just for cold boot but for hot insert of
new disks when the trust quorum may not be available.

Current the only transient error is a `KeyManager::Error` which
indicates that the trust quorum is unavailable becaue not enough sleds
are online.

Fixes https://github.com/oxidecomputer/omicron/issues/3789

### DIFF
--- a/sled-agent/src/storage_manager.rs
+++ b/sled-agent/src/storage_manager.rs
@@ -618,7 +618,6 @@ impl StorageWorker {
         unparsed_disk: UnparsedDisk,
         queued_u2_drives: &mut Option<HashSet<QueuedDiskCreate>>,
     ) -> Result<Disk, sled_hardware::DiskError> {
-        // Ensure the disk conforms to an expected partition layout.
         match sled_hardware::Disk::new(
             &self.log,
             unparsed_disk.clone(),
--- a/sled-agent/src/storage_manager.rs
+++ b/sled-agent/src/storage_manager.rs
@@ -30,8 +30,11 @@ use std::collections::HashSet;
 use std::convert::TryFrom;
 use std::pin::Pin;
 use std::sync::Arc;
+use std::sync::OnceLock;
+use std::time::Duration;
 use tokio::sync::{mpsc, oneshot, Mutex};
 use tokio::task::JoinHandle;
+use tokio::time::{interval, MissedTickBehavior};
 use uuid::Uuid;
 
 use illumos_utils::dumpadm::DumpHdrError;
@@ -39,6 +42,10 @@ use illumos_utils::dumpadm::DumpHdrError;
 use illumos_utils::{zfs::MockZfs as Zfs, zpool::MockZpool as Zpool};
 #[cfg(not(test))]
 use illumos_utils::{zfs::Zfs, zpool::Zpool};
+
+// A key manager can only become ready once. This occurs during RSS or cold
+// boot when the bootstore has detected it has a key share.
+static KEY_MANAGER_READY: OnceLock<bool> = OnceLock::new();
 
 #[derive(thiserror::Error, Debug)]
 pub enum Error {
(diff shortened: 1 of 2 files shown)


sled-agent/src/storage_manager.rs
```diff
@@ -40,6 +43,10 @@ use illumos_utils::{zfs::MockZfs as Zfs, zpool::MockZpool as Zpool};
 #[cfg(not(test))]
 use illumos_utils::{zfs::Zfs, zpool::Zpool};
 
+// A key manager can only become ready once. This occurs during RSS or cold
+// boot when the bootstore has detected it has a key share.
+static KEY_MANAGER_READY: OnceLock<bool> = OnceLock::new();
```
jgallagher, line 48: Nit - is it possible for this to be set to `false`? If not and we're only using this as a threadsafe "set or not", could we do use `OnceLock<()>` instead, and the set-ness of the lock itself is whether it's ready?

sled-agent/src/storage_manager.rs
```diff
@@ -955,84 +1030,160 @@ impl StorageWorker {
...
-                                self.notify_nexus_about_existing_resources(&resources).await?;
-                            }
-                            let _ = responder.send(Ok(()));
+                    // We want to queue failed requests related to the key manager
+                    match self.handle_storage_worker_request(
+                        resources, queued_u2_drives, request)
+                    .await {
+                        Err(Error::DiskError(_)) => {
+                            // We already handle and log disk errors, no need to
+                            // return here.
```
jgallagher, line 1047: Sorry for a vague complaint, but: this seems hard to verify by reading `handle_storage_worker_request`. Does every function we call in every arm of `handle_store_worker_request` handle every possible `DiskError(_)`? Even if yes today, is there some way to restructure this that makes it easier to ensure it continues to be the case? Maybe we carve `DiskError` out of `Error` entirely? (Not at all sure that's reasonable, just throwing out an idea.)

sled-agent/src/storage_manager.rs
andrewjstone: Oh, I was thinking about this, but I always forget the unit type exists! Thanks for pointing this out.  It's a definite change to make.

jgallagher (approved):

sled-agent/src/storage_manager.rs
andrewjstone: Fixed in [f4d62ad](https://github.com/oxidecomputer/omicron/pull/3800/commits/f4d62ad6e08e39f6fd8f9f192ce8b19916408349)

sled-agent/src/storage_manager.rs
andrewjstone: It's a *very* justifiable complaint. I had the same thought when writing it. However, if you take a close look you'll realize that we don't actually do anything with the `DiskErrors` anyway. We just log them and move on in [`do_work`](https://github.com/oxidecomputer/omicron/blob/main/sled-agent/src/storage_manager.rs#L942-L946) . This code doesn't make things any worse.

As John and I discussed offline there are other problems here in that we block waiting for rack secrets to unlock and so we'll need to add more concurrency. Overall, this is more of a general "we need to make sled-agent" a bit more cohesive and testable. We can tackle that outside this PR.

PR #3804, opened 2023-07-31
NAT RPW

internet-diglett:
* Add db table for tracking nat entries
* Add endpoint for retrieving changesets
* Update instance sagas to update table and trigger RPW
* Periodically cleanup soft-deleted entries that no longer
  need to be sync'd by dendrite.

The other half of the RPW lives in Dendrite. It will periodically
check for a changeset, or check for a changeset when the trigger
endpoint is called by the relevant saga / nexus operation.


common/src/sql/dbinit.sql
```diff
@@ -2354,6 +2354,35 @@ CREATE UNIQUE INDEX ON omicron.public.switch_port_settings_group (
...
+STORING (
+    external_address,
+    first_port,
+    last_port,
+    sled_address,
+    vni,
+    mac,
+    time_created,
+    time_deleted
+);
```
smklein, line 2384: As a PSA:

- This file is now in `schema/crdb/dbinit.sql`. You should be able to update that file directly, and keep using this more-or-less as-is.
- I recommend adding a new "rev" of CRDB (3.0.0? Or whatever the +1 of the major version would be), and including "just the addition of this sequence + table + index" within a file named `up.sql`
- It's possible to make these calls idempotent! The `IF NOT EXISTS` keywords will help.

internet-diglett: Yeah, I need to rebase and use the new mechanism for updating the schema :D

nexus/src/internal_api/http_entrypoints.rs
```diff
@@ -484,3 +488,78 @@ async fn saga_view(
...
+/// Fetch NAT ChangeSet
+///
+/// Caller provides their generation as `from_gen`, along with a query
+/// parameter for the page size (`limit`). Endpoint will return changes
+/// that have occured since the caller's generation number up to the latest
+/// change or until the `limit` is reached. If there are no changes, an
+/// empty vec is returned.
+#[endpoint {
+   method = GET,
+    path = "/rpw/nat/ipv4/changeset/{from_gen}"
```
smklein, line 517: tiny nitpick -- I'd just prefix these paths with "/nat", the fact that they're called from Dendrite is an implementation detail

internet-diglett:
Testing resolution of https://github.com/oxidecomputer/buildomat/issues/38

nexus/db-queries/src/db/datastore/ipv4_nat_entry.rs
```diff
@@ -0,0 +1,495 @@
...
+        // database.
+        let updated_rows = sql_query(
+            "
+            UPDATE omicron.public.ipv4_nat_entry
+            SET
+            version_removed = nextval('omicron.public.ipv4_nat_version'),
+            time_deleted = now()
+            WHERE time_deleted IS NULL AND version_removed IS NULL AND id = $1 AND version_added = $2
+            ",
+        )
```
smklein, line 90: I think we definitely need custom SQL to handle `nextval`, but the rest of this could be vanilla diesel.

If we add the following function:

```rust
fn ipv4_nat_next_version() -> diesel::expression::SqlLiteral<BigInt> {                                                                                                                                                                                                                                                                                            
    diesel::dsl::sql::<BigInt>("nextval('omicron.public.ipv4_nat_version')")                                                                                                                                                                                                                                                                                      
}
```

Then this particular snippet of SQL could be replaced with the following:

```rust
        use db::schema::ipv4_nat_entry::dsl;                                                                                                                                                                                                                                                                                                                      
        let updated_rows = diesel::update(dsl::ipv4_nat_entry)                                                                                                                                                                                                                                                                                                    
            .set((                                                                                                                                                                                                                                                                                                                                                
                dsl::version_removed.eq(ipv4_nat_next_version().nullable()),                                                                                                                                                                                                                                                                                      
                dsl::time_deleted.eq(Utc::now()),                                                                                                                                                                                                                                                                                                                 
            ))                                                                                                                                                                                                                                                                                                                                                    
            .filter(dsl::version_removed.is_null())                                                                                                                                                                                                                                                                                                               
            .filter(dsl::id.eq(nat_entry.id))                                                                                                                                                                                                                                                                                                                     
            .filter(dsl::version_added.eq(nat_entry.version_added))                                                                                                                                                                                                                                                                                               
            .execute_async(&*self.pool_connection_authorized(opctx).await?)                                                                                                                                                                                                                                                                                       
            .await                                                                                                                                                                                                                                                                                                                                                
            .map_err(|e| public_error_from_diesel(e, ErrorHandler::Server))?;  
```

nexus/db-model/src/ipv4_nat_entry.rs
```diff
@@ -0,0 +1,126 @@
...
+pub struct Ipv4NatEntry {
+    pub id: Uuid,
+    pub external_address: ipnetwork::IpNetwork,
+    pub first_port: SqlU16,
+    pub last_port: SqlU16,
+    pub sled_address: ipnetwork::IpNetwork,
+    pub vni: Vni,
+    pub mac: MacAddr,
+    pub version_added: SqlU32,
+    pub version_removed: Option<SqlU32>,
```
smklein, line 48: Should we be storing these versions as BigInts, rather than SqlU32s?

If we overflow a u32 (very possible!) these values would start throwing deserialization errors.

nexus/db-model/src/ipv4_nat_entry.rs
```diff
@@ -0,0 +1,126 @@
...
+    pub fn version_removed(&self) -> Option<u32> {
+        self.version_removed.map(|i| i.into())
+    }
+}
+
+#[derive(Queryable, Debug, Clone, Selectable)]
+#[diesel(table_name = ipv4_nat_version)]
+pub struct Ipv4NatGen {
+    pub last_value: SqlU32,
+    pub log_cnt: SqlU32,
```
smklein, line 75: Same comment about storing this as a u32 vs bigint

nexus/db-model/src/ipv4_nat_entry.rs
```diff
@@ -0,0 +1,126 @@
...
+/// NAT Record
+#[derive(Clone, Debug, Serialize, JsonSchema)]
+pub struct Ipv4NatEntryView {
+    pub external_address: Ipv4Addr,
+    pub first_port: u16,
+    pub last_port: u16,
+    pub sled_address: Ipv6Addr,
+    pub vni: external::Vni,
+    pub mac: external::MacAddr,
+    pub gen: u32,
```
smklein, line 88: Same comment about u32 vs u64 here

nexus/db-model/src/ipv4_nat_entry.rs
```diff
@@ -0,0 +1,126 @@
...
+            gen,
+            deleted,
+        }
+    }
+}
+
+/// NAT Generation
+#[derive(Clone, Debug, Serialize, JsonSchema)]
+pub struct Ipv4NatGenView {
+    pub gen: u32,
```
smklein, line 125: u32 vs u64 again here

nexus/db-queries/src/db/datastore/ipv4_nat_entry.rs
```diff
@@ -0,0 +1,495 @@
...
+            time_deleted = NULL,
+            version_added = nextval('omicron.public.ipv4_nat_version')
+            WHERE version_removed IS NOT NULL AND id = $1 AND version_removed = $2
+            ",
+        )
+            .bind::<Uuid, _>(nat_entry.id)
+            .bind::<BigInt, _>(version_removed)
+            .execute_async(&*self.pool_connection_authorized(opctx).await?)
+            .await
+            .map_err(|e| public_error_from_diesel(e, ErrorHandler::Server))?;
```
smklein, line 133:
```rust
        use db::schema::ipv4_nat_entry::dsl;                                                                                                                                                                                                                                                                                                                      
        let updated_rows = diesel::update(dsl::ipv4_nat_entry)                                                                                                                                                                                                                                                                                                    
            .set((                                                                                                                                                                                                                                                                                                                                                
                dsl::version_removed.eq(Option::<i64>::None),                                                                                                                                                                                                                                                                                                     
                dsl::time_deleted.eq(Option::<DateTime<Utc>>::None),                                                                                                                                                                                                                                                                                              
                dsl::version_added.eq(ipv4_nat_next_version()),                                                                                                                                                                                                                                                                                                   
            ))                                                                                                                                                                                                                                                                                                                                                    
            .filter(dsl::version_removed.is_not_null())                                                                                                                                                                                                                                                                                                           
            .filter(dsl::id.eq(nat_entry.id))                                                                                                                                                                                                                                                                                                                     
            .filter(dsl::version_removed.eq(version_removed))                                                                                                                                                                                                                                                                                                     
            .execute_async(&*self.pool_connection_authorized(opctx).await?)                                                                                                                                                                                                                                                                                       
            .await                                                                                                                                                                                                                                                                                                                                                
            .map_err(|e| public_error_from_diesel(e, ErrorHandler::Server))?
```

nexus/db-queries/src/db/datastore/ipv4_nat_entry.rs
```diff
@@ -0,0 +1,495 @@
...
+        }
+    }
+
+    pub async fn ipv4_nat_delete_by_external_ip(
+        &self,
+        opctx: &OpContext,
+        external_ip: &ExternalIp,
+    ) -> DeleteResult {
+        let updated_rows = sql_query(
+            "
```
smklein, line 173: See the prior two examples -- this can use the `ipv4_nat_next_version()` function and otherwise avoid defining a full manual SQL query

smf/sled-agent/non-gimlet/config-rss.toml
```diff
@@ -41,7 +41,7 @@ external_dns_zone_name = "oxide.test"
 # the DNS domain delegated to the rack by the customer. Each of these addresses
 # must be contained in one of the "internal services" IP Pool ranges listed
 # below.
-external_dns_ips = [ "192.168.1.20", "192.168.1.21" ]
+external_dns_ips = [ "10.85.0.20", "10.85.0.21" ]
```
smklein, line 44: Nit: Is this necessary? Or just a local change that made it into this PR?

schema/crdb/6.1.0/up1.sql
```diff
@@ -0,0 +1 @@
+CREATE SEQUENCE IF NOT EXISTS omicron.public.ipv4_nat_version START 1 INCREMENT 1;
```
smklein, line 1: Couple nitpicks on versioning:

1. We've had other breaking changes, the latest is `8.0.0`, so this should be higher than that
2. This should probably just be `9.0.0` (or whatever that largest major version is at the time of merging)

https://github.com/oxidecomputer/omicron/blob/ebb25ebbb27892cbfe8bb2d00f6cf7b8de563a29/schema/crdb/README.adoc?plain=1#L56

This is semver because we will want to have version changes that aren't breaking in the future, but **right now**, we aren't making guarantees about backward compatibility. To that end, all changes are basically breaking changes.

nexus/src/internal_api/http_entrypoints.rs
```diff
@@ -540,3 +544,79 @@ async fn bgtask_view(
...
+        let path = path_params.into_inner();
+        let query = query_params.into_inner();
+        let mut changeset = nexus
+            .datastore()
+            .ipv4_nat_changeset(&opctx, path.from_gen, query.limit)
+            .await?;
+        changeset.sort_by_key(|e| e.gen);
+        Ok(HttpResponseOk(changeset))
+    };
+    apictx.external_latencies.instrument_dropshot_handler(&rqctx, handler).await
```
smklein, line 593:
```suggestion
    apictx.internal_latencies.instrument_dropshot_handler(&rqctx, handler).await
```

nexus/src/internal_api/http_entrypoints.rs
```diff
@@ -540,3 +544,79 @@ async fn bgtask_view(
...
+) -> Result<HttpResponseOk<Ipv4NatGenView>, HttpError> {
+    let apictx = rqctx.context();
+    let handler = async {
+        let opctx = crate::context::op_context_for_internal_api(&rqctx).await;
+        let nexus = &apictx.nexus;
+        let gen = nexus.datastore().ipv4_nat_current_version(&opctx).await?;
+        let view = Ipv4NatGenView { gen };
+        Ok(HttpResponseOk(view))
+    };
+    apictx.external_latencies.instrument_dropshot_handler(&rqctx, handler).await
```
smklein, line 621:
```suggestion
    apictx.internal_latencies.instrument_dropshot_handler(&rqctx, handler).await
```

nexus/db-queries/src/db/datastore/ipv4_nat_entry.rs
```diff
@@ -0,0 +1,495 @@
...
+                type_name: ResourceType::Ipv4NatEntry,
+                lookup_type: LookupType::ByCompositeId(
+                    "id, version_added".to_string(),
+                ),
+            });
+        }
+        Ok(())
+    }
+
+    pub async fn ipv4_nat_cancel_delete(
```
smklein, line 108: Could this be marked `#[cfg(test)]`?  The act of cancelling a delete seems somewhat unsafe to me outside test contexts.

nexus/src/internal_api/http_entrypoints.rs
```diff
@@ -540,3 +544,79 @@ async fn bgtask_view(
...
 }
+
+// NAT RPW internal APIs
+
+/// Path parameters for NAT ChangeSet
+#[derive(Deserialize, JsonSchema)]
+struct RpwNatPathParam {
+    /// which change number to start generating
+    /// the change set from
+    from_gen: u32,
```
smklein, line 555: Same as other comment, about u32 vs u64 representation of generation numbers

nexus/src/internal_api/http_entrypoints.rs
```diff
@@ -540,3 +544,79 @@ async fn bgtask_view(
...
+}
+
+// TODO: Remove (maybe)
+// This endpoint may be more trouble than it's worth. The reason this endpoint
+// may be a headache is because the underlying SEQUENCE in the DB will advance,
+// even on failed transactions. This is not a big deal, but it will lead to
+// callers trying to catch up to entries that don't exist if they check this
+// endpoint first. As an alternative, we could just page through the
+// changesets until we get nothing back, then revert to periodic polling
+// of the changeset endpoint.
```
smklein, line 603: Where'd we land on this?

smklein (commented):
My first pass here is focused a bit more at the DB than the RPW itself

nexus/src/app/background/nat_cleanup.rs
```diff
@@ -0,0 +1,111 @@
...
+        opctx: &'b OpContext,
+    ) -> BoxFuture<'c, serde_json::Value>
+    where
+        'a: 'c,
+        'b: 'c,
+    {
+        async {
+            let log = &opctx.log;
+
+            let result = self.datastore.ipv4_nat_current_version(opctx).await;
```
smklein, line 46: To verify my understanding:

- This task periodically reads the latest generation number from the DB
- It sends that generation number to Dendrite instances (`client.ipv4_nat_generation()`)
- It cleans old records

So the "dendrite-playing-catch-up" entirely relies on Dendrite reaching back to the internal API within Nexus, correct?

nexus/db-queries/src/db/datastore/ipv4_nat_entry.rs
```diff
@@ -0,0 +1,495 @@
...
+        &self,
+        opctx: &OpContext,
+        nat_entry: Ipv4NatValues,
+    ) -> CreateResult<()> {
+        // Ensure that the record with the parameters we want exists in the
+        // database. If an entry exists with the same external ip but
+        // different target sled / vpc / mac, we will violate a uniqueness
+        // constraint (which is desired in this case).
+        //
+        // It seems that this isn't straightforward to express in diesel
```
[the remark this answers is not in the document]
internet-diglett: Will give this a try! Thanks for giving this a look!

smf/sled-agent/non-gimlet/config-rss.toml
internet-diglett: Local change that slipped into the PR, my bad 😂

schema/crdb/6.1.0/up1.sql
internet-diglett: This versioning was from my workaround to the previous issue we encountered regarding the anonymous constraints in the db schema, I can bump this to a proper major version number now that I have the fix (or will have the fix) in this branch.

nexus/db-queries/src/db/datastore/ipv4_nat_entry.rs
internet-diglett: This was originally intended for usage in saga unwinds, but I think we've since decided that we shouldn't try to unwind deletion of network configuration and instead consider it the "point of no return"

nexus/src/internal_api/http_entrypoints.rs
internet-diglett: I haven't seen a place where it has merit, so I'll remove it.

nexus/src/app/background/nat_cleanup.rs
internet-diglett: That's pretty close to how it works:
* Omicron periodically reads the latest generation in DB
* Omicron checks to see what generation numbers the various dendrite instances have (we are querying dendrite, not triggering any work in dendrite)
* Omicron only cleans up records older than the earliest found dendrite generation

The idea is that we don't clean up changes that a dendrite instance hasn't picked up yet. We also shouldn't have a detrimental TOCTOU race because the only way a dendrite instance moves backwards is if it crashes, and in that case it's okay that we're cleaning up old entries because we don't want that dendrite instance to re-apply them anyway.

Dendrite only catches up when:
* An event in the control plane calls the `trigger` endpoint on the dendrite API (a "doorbell" of sorts)
* Dendrite's internal RPW interval timer expires

In these events, yes, Dendrite will reach back via the internal API to catch up on the changes that have occurred.

nexus/db-model/src/ipv4_nat_entry.rs
```diff
@@ -0,0 +1,126 @@
...
+use schemars::JsonSchema;
+use serde::Serialize;
+use uuid::Uuid;
+
+// TODO correctness
+// If we're not going to store ipv4 and ipv6
+// NAT entries in the same table, and we don't
+// need any of the special properties of the IpNetwork
+// column type, does it make sense to use a different
+// column type?
```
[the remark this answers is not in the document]
internet-diglett: By keeping the tables for v4 and v6 separate, we maintain a 1:1 mapping with the tables in `dendrite`, which does make troubleshooting, observability, and state reconciliation a bit more straightforward, so my gut says "keep them separate"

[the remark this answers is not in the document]
internet-diglett: I'll try using `Ipv4Network` unless someone has an objection :)

internet-diglett:
Note that since we are changing some of the types used by `dendrite` via the internal API, the `deploy` task will probably fail since we must first push our changes here, update `dendrite`, then reference the updated `dendrite` in the package manifest here.

nexus/src/internal_api/http_entrypoints.rs
smklein: Looks like this still needs updating

nexus/src/internal_api/http_entrypoints.rs
smklein: I think I still might remove the `/rpw` prefix, the fact that Nexus processes this as an RPW is an implementation detail.

nexus/src/app/instance_network.rs
```diff
@@ -361,29 +368,55 @@ impl super::Nexus {
...
+                    mac: nexus_db_model::MacAddr(
+                        omicron_common::api::external::MacAddr(mac_address),
+                    ),
+                };
+                self.db_datastore
+                    .ensure_ipv4_nat_entry(opctx, nat_entry)
+                    .await?;
+            }
+            IpNetwork::V6(_v6net) => {
+                // TODO: implement handling of v6 nat.
```
smklein, line 417: If this is unimplemented, we should throw an error rather than silently doing nothing

nexus/db-model/src/ipv4_nat_entry.rs
```diff
@@ -0,0 +1,93 @@
...
+    pub fn first_port(&self) -> u16 {
+        self.first_port.into()
+    }
+
+    pub fn last_port(&self) -> u16 {
+        self.last_port.into()
+    }
+}
+
+/// Database representation of an Ipv4 NAT Generation.
```
smklein, line 53: Hey, it took me a while to figure this out -- this is a view of a sequence, not a table, and the `log_cnt` + `is_called` values are synthetically provided from CockroachDB.

That's worth documenting! I do not find that behavior obvious - I was trying to find out where these were used.

Related: Do we actually need to include `log_cnt` and `is_called` here? When we query for the Ipv4NatGen, we only read `last_value`, right?

smklein (approved):
Last few comments are mostly nitpicks, but this looks good! Thanks for the query cleanup!

nexus/db-model/src/ipv4_nat_entry.rs
internet-diglett: Yes, thus far we only care about `last_value`, I can delete the other two columns from the struct

internet-diglett: Actually, it seems that since we're not using this struct anymore thanks to some of the other changes that have been made :D

schema/crdb/dbinit.sql
```diff
@@ -2899,14 +2847,67 @@ CREATE TABLE IF NOT EXISTS omicron.public.db_metadata (
     CHECK (singleton = true)
 );
 
+
+CREATE SEQUENCE IF NOT EXISTS omicron.public.ipv4_nat_version START 1 INCREMENT 1;
```
smklein, line 2851: This is a super minor nitpick, but we did some work to ensure that the order of declarations in `dbinit.sql` should **mostly** not matter.

Can we move this stuff up earlier? I find it a little odd to have the ordering of dbinit be:

- Most of the structures we use
- The db_metadata table declaration
- The tables that were added for NAT
- INSERT into db_metadata

(I kinda think db_metadata's declaration + INSERT should be right next to each other, at the end of the file, if opssible)

smklein (approved):
One minor ordering comment aside, this looks good to merge to me!

schema/crdb/10.0.0/up1.sql
```diff
@@ -0,0 +1 @@
+CREATE SEQUENCE IF NOT EXISTS omicron.public.ipv4_nat_version START 1 INCREMENT 1;
```
smklein, line 1: Before you merge this -- looks like `10.0.0` is already used on main?

https://github.com/oxidecomputer/omicron/tree/main/schema/crdb/10.0.0

internet-diglett: Awaiting merge of https://github.com/oxidecomputer/dendrite/pull/627

internet-diglett: Dendrite PR merged, bumping Omicron dependency version and Schema version

PR #3807, opened 2023-08-01
Update clickhouse to v22.8.9.24-1 on illumos

citrus-it:


karencfv (approved):
Thanks! 

Just adding for context: This PR includes a small [fix](https://github.com/oxidecomputer/garbage-compactor/pull/14) on the ClickHouse Keeper (replica coordinator), which will have no effect on the current implementation, but is part of the upcoming work with node replication.

PR #3819, opened 2023-08-04
Update resource limits as per #3212

zephraph:
vCPUs: 32 -> 64
memory: 64 GiB -> 256 GiB
disk size: 1 TiB -> 1023 GiB


david-crespo:
The fact that no tests fail with this change is a little concerning!

zephraph:
The tests uses the constants to pivot off of. I don't see the value of hardcoding the limits in the tests. 

So you have situations like this: 

```
    let disk_size =
        ByteCount::try_from(MAX_DISK_SIZE_BYTES + (1 << 30)).unwrap();
```

where we use the limits to create a value over the limit to ensure that it fails.

david-crespo (approved):
Ah, that makes sense. Cool!

PR #3825, opened 2023-08-07
Enable timesync for the non-gimlet config and in CI

### LINKED ISSUE #3501: following how-to-run, Omicron does not sync clocks when using non-gimlet config

citrus-it:
This changes non-gimlet/config-rss.toml so that time synchronisation is enabled, and updates the CI deploy job to support that.

Fixes #3501

### DIFF
--- a/.github/buildomat/jobs/deploy.sh
+++ b/.github/buildomat/jobs/deploy.sh
@@ -7,6 +7,7 @@
 #:  "%/var/svc/log/oxide-sled-agent:default.log*",
 #:  "%/pool/ext/*/crypt/zone/oxz_*/root/var/svc/log/oxide-*.log*",
 #:  "%/pool/ext/*/crypt/zone/oxz_*/root/var/svc/log/system-illumos-*.log*",
+#:  "%/pool/ext/*/crypt/zone/oxz_ntp_*/root/var/log/chrony/*.log*",
 #:  "!/pool/ext/*/crypt/zone/oxz_propolis-server_*/root/var/svc/log/*.log*",
 #:  "%/pool/ext/*/crypt/debug/global/oxide-sled-agent:default.log.*",
 #:  "%/pool/ext/*/crypt/debug/oxz_*/oxide-*.log.*",
@@ -53,6 +54,7 @@ _exit_trap() {
 	z_swadm addr list
 	z_swadm route list
 	z_swadm arp list
+	z_swadm nat list
 
 	PORTS=$(pfexec /opt/oxide/opte/bin/opteadm list-ports | tail +2 | awk '{ print $1; }')
 	for p in $PORTS; do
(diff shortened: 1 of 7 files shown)


.github/buildomat/jobs/deploy.sh
```diff
@@ -166,7 +210,34 @@ ZPOOL_VDEV_DIR=/scratch ptime -m pfexec ./tools/create_virtual_hardware.sh
 tar xf out/omicron-sled-agent.tar pkg/config-rss.toml
 SILO_NAME="$(sed -n 's/silo_name = "\(.*\)"/\1/p' pkg/config-rss.toml)"
 EXTERNAL_DNS_DOMAIN="$(sed -n 's/external_dns_zone_name = "\(.*\)"/\1/p' pkg/config-rss.toml)"
-rm -f pkg/config-rss.toml
+
+# Substitute addresses from the external network range into the RSS config.
+sed -i~ "
```
luqmana, line 215: This is some impressive `sed`'ing 😅 but seems like this script could stand to not be bash soon (not a blocker for this PR to be clear)

.github/buildomat/jobs/deploy.sh
```diff
@@ -239,13 +310,10 @@ do
...
-# We also need to configure proxy arp for any services which use OPTE for external connectivity (e.g. Nexus)
-tar xf out/omicron-sled-agent.tar pkg/config-rss.toml
-SOFTNPU_MAC=$(dladm show-vnic sc0_1 -p -o macaddress | sed -E 's/[ :]/&0/g; s/0([^:]{2}(:|$))/\1/g')
-SERVICE_IP_POOL_START="$(sed -n 's/^first = "\(.*\)"/\1/p' pkg/config-rss.toml)"
-SERVICE_IP_POOL_END="$(sed -n 's/^last = "\(.*\)"/\1/p' pkg/config-rss.toml)"
-rm -r pkg
+# We also need to configure proxy arp for any services which use OPTE for
+# external connectivity (e.g. Nexus)
+SOFTNPU_MAC=$(dladm show-vnic sc0_1 -p -o macaddress \
+    | sed -E 's/[ :]/&0/g; s/0([^:]{2}(:|$))/\1/g')
```
luqmana, line 316: We can drop these (and subsequent) lines by just making sure `PXA_START/END` above also covers this range.

luqmana (commented):
woo, thanks for figuring this out andy!

.github/buildomat/jobs/deploy.sh
citrus-it: I didn't do that because I'm also using `PXA_START` - `PXE_END` to configure the IP pool through the API in the end-to-end tests bootstrap. I'll adjust that and simplify this script a bit.

.github/buildomat/jobs/deploy.sh
citrus-it: I didn't do that because I'm also using `PXA_START` - `PXE_END` to configure the IP pool through the API in the end-to-end tests bootstrap. I'll adjust that and simplify this script a bit.

luqmana (approved):

Issue #1598, opened 2022-08-16
[sled agent] Extract a Service Bundle from zones which are being destroyed


smklein:
When `RunningZone` objects are dropped, for whatever reason, the underlying zone is halted and removed:

https://github.com/oxidecomputer/omicron/blob/50c6ce1529a0fe075402efba813ee785259d32a2/sled-agent/src/illumos/running_zone.rs#L310-L321

This if often the case when:
- We attempt to boot a zone, but it fails to start
- A zone instance is destroyed

This helps prevent leakage of resources, but limits inspection of "why the zone died, and what it was up to".

It would be useful to extract log information from zones into a "service bundle" before we destroy the underlying zone.

davepacheco:
This would be great.  Things I'd consider including:

- ptree output for the whole zone
- pfiles, pstack, and core files for any of the key processes (e.g., in a CockroachDB zone, this would be just "cockroach" itself)
- SMF logs, other system logs, and any application-level logs
- uptime
- "last" output, "who" output
- `svcs -p` output for the whole zone (includes service states and timestamp of most recent state transitions)

Another tool that might be useful would be to preserve a limited number of such zones for interactive investigation by support.

leftwo:
As part of this, or maybe this can be another issue, gather the versions of software for the parts involved also.  Like Omicron, Sled-agent, propolis, crucible-agent/downstairs, Helios... maybe others too that I can't think of.

PR #3829, opened 2023-08-07
Adding automatic bundle on zone death

bnaecker:
- Moves zone bundle code to free functions in its own module, out of the `ServiceManager` itself.
- Adds handling of Propolis zones, by reworking the locking around the instance manager.
- Adds sled-agent endpoint for listing all zone bundles, even those not corresponding to an existing zone.
- Adds a "cause" to the zone bundle metadata, indicating why it was created.
- Some QoL improvements to `zone-bundle`, allowing listing bundles from zones matching a filter (or all), along with parseable output.
- Improves robustness of extracting `GATEWAY_MAC` from the ARP entries for the provided `GATEWAY_IP`, and adds warning if the proxy-arp entries are not provided.
- Extracts log files which may have been archived to a U.2 as well as the M.2-local log files
- Adds basic mechanism for running zone-specific commands. Not used yet.

### DIFF
--- a/sled-agent/src/instance.rs
+++ b/sled-agent/src/instance.rs
@@ -13,6 +13,7 @@ use crate::nexus::NexusClientWithResolver;
 use crate::params::{
     InstanceHardware, InstanceMigrationSourceParams,
     InstanceMigrationTargetParams, InstanceStateRequested, VpcFirewallRule,
+    ZoneBundleCause,
 };
 use crate::profile::*;
 use crate::storage_manager::StorageResources;
@@ -34,6 +35,9 @@ use omicron_common::api::internal::shared::{
 };
 use omicron_common::backoff;
 //use propolis_client::generated::DiskRequest;
+use crate::params::ZoneBundleMetadata;
+use crate::zone_bundle;
+use crate::zone_bundle::BundleError;
 use propolis_client::Client as PropolisClient;
 use rand::prelude::SliceRandom;
 use rand::SeedableRng;
@@ -526,12 +530,47 @@ impl InstanceInner {
     /// This routine is safe to call even if the instance's zone was never
     /// started. It is also safe to call multiple times on a single instance.
     async fn terminate(&mut self) -> Result<(), Error> {
+        let zname = propolis_zone_name(self.propolis_id());
+
+        // First fetch the running state.
+        //
+        // If there is nothing here, then there is no `RunningZone`, and so
+        // there's no zone or resources to clean up at all.
+        let mut running_state = if let Some(state) = self.running_state.take() {
+            state
+        } else {
+            debug!(
+                self.log,
+                "Instance::terminate() called with no running state"
+            );
+            return Ok(());
+        };
+
+        // Take a zone bundle whenever this instance stops.
+        let context = self
+            .storage
+            .zone_bundle_context(&zname, ZoneBundleCause::TerminatedInstance)
+            .await;
+        if let Err(e) = zone_bundle::create(
+            &self.log,
+            &running_state.running_zone,
+            &context,
+        )
+        .await
+        {
+            error!(
+                self.log,
+                "Failed to take zone bundle for terminated instance";
+                "zone_name" => &zname,
+                "reason" => ?e,
+            );
+        }
+
         // Ensure that no zone exists. This succeeds even if no zone was ever
         // created.
         // NOTE: we call`Zones::halt_and_remove_logged` directly instead of
         // `RunningZone::stop` in case we're called between creating the
         // zone and assigning `running_state`.
-        let zname = propolis_zone_name(self.propolis_id());
         warn!(self.log, "Halting and removing zone: {}", zname);
         Zones::halt_and_remove_logged(&self.log, &zname).await.unwrap();
 
@@ -539,12 +578,7 @@ impl InstanceInner {
         self.instance_ticket.terminate();
 
         // See if there are any runtime objects to clean up.
-        let mut running_state = if let Some(state) = self.running_state.take() {
-            state
-        } else {
-            return Ok(());
-        };
-
+        //
         // We already removed the zone above but mark it as stopped
         running_state.running_zone.stop().await.unwrap();
 
@@ -553,6 +587,20 @@ impl InstanceInner {
 
         Ok(())
     }
+}
+
+/// A result asking an instance to generate a zone bundle.
+#[derive(Debug)]
+pub enum InstanceZoneBundleResult {
+    /// A zone bundle was operation was performed on the requested zone.
+    Ok(Result<ZoneBundleMetadata, BundleError>),
+
+    /// A request to create a bundle for a zone whose name does not match this
+    /// one's Propolis zone.
+    NotMe,
+
+    /// The zone is not currently running.
+    NotRunning,
 }
 
 /// A reference to a single instance running a running Propolis server.
@@ -625,6 +673,39 @@ impl Instance {
         let inner = Arc::new(Mutex::new(instance));
 
         Ok(Instance { inner })
+    }
+
+    /// Create bundle from a zone matching the provided name.
+    ///
+    /// This returns an [`InstanceZoneBundleResult`] to indicate whether the
+    /// bundle could be correctly taken. See that enum for details.
+    pub async fn request_zone_bundle(
+        &self,
+        name: &str,
+    ) -> InstanceZoneBundleResult {
+        let inner = self.inner.lock().await;
+        let zone_name = propolis_zone_name(inner.propolis_id());
+        if zone_name != name {
+            return InstanceZoneBundleResult::NotMe;
+        }
+        match &*inner {
+            InstanceInner { running_state: None, .. } => {
+                InstanceZoneBundleResult::NotRunning
+            }
+            InstanceInner {
+                ref log,
+                running_state: Some(RunningState { ref running_zone, .. }),
+                ..
+            } => {
+                let context = inner
+                    .storage
+                    .zone_bundle_context(name, ZoneBundleCause::ExplicitRequest)
+                    .await;
+                InstanceZoneBundleResult::Ok(
+                    zone_bundle::create(log, running_zone, &context).await,
+                )
+            }
+        }
     }
 
     pub async fn current_state(&self) -> InstanceRuntimeState {
--- a/sled-agent/src/instance.rs
+++ b/sled-agent/src/instance.rs
@@ -13,7 +13,6 @@ use crate::nexus::NexusClientWithResolver;
 use crate::params::{
     InstanceHardware, InstanceMigrationSourceParams,
     InstanceMigrationTargetParams, InstanceStateRequested, VpcFirewallRule,
-    ZoneBundleCause,
 };
 use crate::profile::*;
 use crate::storage_manager::StorageResources;
@@ -35,6 +34,7 @@ use omicron_common::api::internal::shared::{
 };
 use omicron_common::backoff;
 //use propolis_client::generated::DiskRequest;
+use crate::params::ZoneBundleCause;
 use crate::params::ZoneBundleMetadata;
 use crate::zone_bundle;
 use crate::zone_bundle::BundleError;
@@ -155,7 +155,9 @@ fn fmri_name() -> String {
     format!("{}:default", service_name())
 }
 
-fn propolis_zone_name(id: &Uuid) -> String {
+/// Return the expected name of a Propolis zone managing an instace with the
+/// provided ID.
+pub fn propolis_zone_name(id: &Uuid) -> String {
     format!("{}{}", PROPOLIS_ZONE_PREFIX, id)
 }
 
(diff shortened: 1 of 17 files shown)


bnaecker:
This is part of #1598. It adds support for zone bundles to Propolis instances, and also automatically creates bundles for zones before they are destroyed. This took much longer than I anticipated, mostly due to the very awkward interface between the sled-agent and the underlying instances. The main sticking point is that the lock around the full `BTreeMap` of instances was a `std::sync::Mutex`. while the lock around the instances themselves is `tokio::sync::Mutex`. (The zone bundling code is also async.) It's nearly impossible to hold a sync lock across an await point, so this required moving the instance map lock to an async one, which had a bunch of other fallout.

The other issue I hit was part of #3713. That moves rotated log files out _off_ the M.2s and to some arbitrary U.2. I wasn't expecting that, and it's an example of the unintended coupling I was worried about in the writeup in #3388. Now the bundling logic is aware of that archival, and needs to go search the possible archive locations.

Anyway, that's a lot of implementation detail, but here's some high-level tests of the whole thing running now:

```bash
bnaecker@shale : ~/omicron $ ./target/release/zone-bundle --host fd00:1122:3344:101::1 ls
Zone                                                             Bundle ID                            Created
oxz_nexus_6a7b95ab-4195-44f3-a2a5-60064ce1587f                   650d86e9-e485-4582-b6be-66245b716765 2023-08-07 23:26:42.040415517 UTC
bnaecker@shale : ~/omicron $ ./target/release/zone-bundle --host fd00:1122:3344:101::1 list-zones
oxz_switch
oxz_internal_dns_d639d7a2-b290-499a-b09d-ba5c56ece97a
oxz_internal_dns_f97dc6be-e5c9-4f2b-a34d-cbf821c6e268
oxz_internal_dns_59171a80-b655-4201-a18a-2ae832b4f0a3
oxz_ntp_ef57a98d-159d-4fe2-a51b-9500c4bf8f29
oxz_cockroachdb_469567dc-8da6-44cd-8cc8-80a457691cca
oxz_cockroachdb_c3b6e935-124c-48a0-9576-9f13e76eaa61
oxz_cockroachdb_77134fbe-0e91-48a8-b54f-16a21cbefca7
oxz_cockroachdb_e8b944ee-39da-4e36-adfc-bf78cc2c21ff
oxz_cockroachdb_9a30f4d0-e7b6-4c24-a5c8-776e5ac451f1
oxz_crucible_936ff2c8-3648-428a-9d23-fea637916015
oxz_crucible_91e6cb07-887a-4257-8fff-8ddc5cf981aa
oxz_oximeter_5b36dff8-842d-4613-8e5e-3d677b370bc6
oxz_crucible_8b87ae1c-a99f-482c-bad8-48c98476cd04
oxz_crucible_e9e0fbae-fdbe-45f2-925e-adf217d46bcf
oxz_nexus_104852c6-06fb-4c1d-b78e-5742145acad1
oxz_crucible_pantry_d507ac9d-53b2-495a-a296-64da73107d80
oxz_crucible_pantry_a4ee219f-e4ad-45ca-9544-a0bd6b0cf1fe
oxz_crucible_pantry_4a2e059b-4826-48f8-ba34-e74f78772494
oxz_clickhouse_1fde5647-1a1e-42b1-8f2a-5b3191cb8284
oxz_external_dns_6193ec54-556f-4f31-940a-5d045337299c
oxz_external_dns_5317b71f-f649-48a0-960c-edce64f38de4
oxz_crucible_727f70ec-fd40-413c-b552-afe2b72ed22c
oxz_nexus_30570377-8dfd-4ae2-8c83-4bfd2dc32ba4
oxz_nexus_6a7b95ab-4195-44f3-a2a5-60064ce1587f
oxz_propolis-server_84dafabe-4ef3-49c9-8fce-4820d3dec68a
bnaecker@shale : ~/omicron $ ./target/release/zone-bundle --host fd00:1122:3344:101::1 create oxz_propolis-server_84dafabe-4ef3-49c9-8fce-4820d3dec68a
Created zone bundle: oxz_propolis-server_84dafabe-4ef3-49c9-8fce-4820d3dec68a/421b401b-cd8c-42eb-b73b-88d6bb4c573c
bnaecker@shale : ~/omicron $ ./target/release/zone-bundle --host fd00:1122:3344:101::1 ls
Zone                                                             Bundle ID                            Created
oxz_nexus_6a7b95ab-4195-44f3-a2a5-60064ce1587f                   650d86e9-e485-4582-b6be-66245b716765 2023-08-07 23:26:42.040415517 UTC
oxz_propolis-server_84dafabe-4ef3-49c9-8fce-4820d3dec68a         421b401b-cd8c-42eb-b73b-88d6bb4c573c 2023-08-07 23:29:09.291215570 UTC
bnaecker@shale : ~/omicron $ ../oxide.rs/target/release/oxide instance stop --project foo --instance inst0
success
Instance {
    description: "test instance",
    hostname: "inst0",
    id: 22090927-bcbd-43c2-b06f-0d77c183f9d7,
    memory: ByteCount(
        1073741824,
    ),
    name: Name(
        "inst0",
    ),
    ncpus: InstanceCpuCount(
        2,
    ),
    project_id: 3f5ac354-91d0-4bed-a52a-e265cab4a850,
    run_state: Stopping,
    time_created: 2023-08-07T23:28:34.220437Z,
    time_modified: 2023-08-07T23:28:34.220437Z,
    time_run_state_updated: 2023-08-07T23:29:23.561350Z,
}
bnaecker@shale : ~/omicron $ ../oxide.rs/target/release/oxide instance stop --project foo --instance inst0 -o zone-name,bundle-id,cause
error: unexpected argument '-o' found

Usage: oxide instance stop [OPTIONS] --instance <instance>

For more information, try '--help'.
bnaecker@shale : ~/omicron $ ../oxide.rs/target/release/oxide instance stop --project foo --instance inst0 -o zone-name,bundle-id,cause^C
bnaecker@shale : ~/omicron $ ./target/release/zone-bundle --host fd00:1122:3344:101::1 ls -o zone-name,bundle-id,cause
Zone                                                             Bundle ID                            Cause
oxz_nexus_6a7b95ab-4195-44f3-a2a5-60064ce1587f                   650d86e9-e485-4582-b6be-66245b716765 ExplicitRequest
oxz_propolis-server_84dafabe-4ef3-49c9-8fce-4820d3dec68a         421b401b-cd8c-42eb-b73b-88d6bb4c573c ExplicitRequest
oxz_propolis-server_84dafabe-4ef3-49c9-8fce-4820d3dec68a         b7e02d88-ecc2-4a10-a173-baa95d615945 TerminatedInstance
```

This shows that:

1. There was one zone bundle originally, for a Nexus zone.
2. I created one manually for the existing Propolis zone managing my instance.
3. A _new_ bundle was created automatically on zone death.

One side note: there's only one other time we actually destroy zones, which is when the sled-agent starts up and finds zones it didn't expect. I'm not really in a position to test that, though we do take bundles for those zones and mark them with a specific `ZoneBundleCause::UnexpectedZone`.

sled-agent/src/instance.rs
```diff
@@ -526,25 +530,55 @@ impl InstanceInner {
...
+            &context,
+        )
+        .await
+        {
+            error!(
+                self.log,
+                "Failed to take zone bundle for terminated instance";
+                "zone_name" => &zname,
+                "reason" => ?e,
+            );
```
smklein, line 566: So, on the one hand, doing this is great, because we'll capture a lot more info. On the other hand, it seems like it'll massively inflate the amount of traffic moving onto the M.2s (and U.2s, if I understand #3713 correctly).

Do we ever trim down the set of zone bundles? E.g., removing the oldest ones when we fill up too much space?

(honestly all other comments on this PR are nits, this is the one that IMO matters most)

tools/create_virtual_hardware.sh
```diff
@@ -59,7 +64,26 @@ function ensure_softnpu_zone {
     success "softnpu zone exists"
 }
 
+function warn_if_physical_link_and_no_proxy_arp {
```
smklein, line 67: These warning seem great, but should they be a separate PR?

tools/scrimlet/softnpu-init.sh
```diff
@@ -27,7 +27,24 @@ fi
 
 #   $ arp -a | grep 192.168.21.1
 #   e1000g1 192.168.21.1         255.255.255.255          90:ec:77:2e:70:27
-GATEWAY_MAC=${GATEWAY_MAC:=$(arp -a | grep "$GATEWAY_IP" | awk -F ' ' '{print $NF}')}
+#
+# Add an extrac space at the end of the search pattern passed to `grep`, so that
```
smklein, line 31:
```suggestion
# Add an extra space at the end of the search pattern passed to `grep`, so that
```

But see comment above -- should this be a separate PR?

sled-agent/src/instance_manager.rs
```diff
@@ -320,6 +327,27 @@ impl InstanceManager {
...
             .map_err(Error::from)
     }
+
+    /// Create a zone bundle from a named instance zone, if it exists.
+    pub async fn create_zone_bundle(
+        &self,
+        name: &str,
+    ) -> Result<ZoneBundleMetadata, BundleError> {
+        for (_propolis_id, instance) in
+            self.inner.instances.lock().await.values()
```
gjcolombo, line 337: Since the Propolis ID is part of the value in the map, I wonder if we could instead do something like this:

```rust
if let Some(instance) =
    self.inner
        .instances
        .lock()
        .unwrap()
        .values()
        .find(|(pid, _)| name == propolis_zone_name(pid))
        .cloned()
{
    match instance.request_zone_bundle(name).await {
        // we already know the name matches
        InstanceZoneBundleResult::NotMe => unreachable!(),

        // other cases as before
    }
}
```

Cloning the `Instance` is OK because it's just a wrapper around an `Arc` that contains the actual `InstanceInner`. The other instance manager routines do basically this to get around the sync/async mismatch; the difference is that they're looking up instances directly by key instead of trying to look for matching values, so they can call `get` whereas this code will have to `find`.

I think this allows you not to hold the instance manager lock while taking the bundle, which (if I'm understanding everything else correctly) should be enough to avoid having to turn that lock into a Tokio mutex. (It also saves us from holding a big lock over a possibly-long-running file copy operation.)

gjcolombo (commented):
I still need to look at `zone_bundle.rs` in more detail but had one comment about the instance manager locking that I wanted to get down before paging this out for the day.

sled-agent/src/instance_manager.rs
bnaecker: That's a great suggestion, thanks Greg. All the contortions were definitely predicated on not cloning, and so needing to hold both locks. That's not great, and thanks to this idea, not needed!

I think this should also get rid of the `InstanceZoneBundleResult`, since the `NotMe` variant will be obviated by the call to `find()`, and we can report `Unavailable` as a `BundleError` directly.

Thanks!

tools/create_virtual_hardware.sh
bnaecker: Sure, they can be. I definitely ran into them while testing this, but I can split them out as well, no problem.

sled-agent/src/instance.rs
bnaecker: Other than calls to `zone-bundle delete`, we do not ever remove bundles. That's an in-progress PR I'll be putting up next. I agree it's critical -- I debated putting it in this PR, but it was already borderline in terms of size for a review.

sled-agent/src/instance.rs
```diff
@@ -555,6 +589,20 @@ impl InstanceInner {
...
+pub enum InstanceZoneBundleResult {
+    /// A zone bundle was operation was performed on the requested zone.
+    Ok(Result<ZoneBundleMetadata, BundleError>),
+
+    /// A request to create a bundle for a zone whose name does not match this
+    /// one's Propolis zone.
+    NotMe,
+
+    /// The zone is not currently running.
+    NotRunning,
```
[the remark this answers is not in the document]
bnaecker: I think this enum will be removed entirely due to Greg's suggestion around locking below. This was mostly a way to report a name match from the `InstanceInner` out to the `InstanceManager`, but won't be needed I think.

sled-agent/src/instance.rs
bnaecker: Also, an argument could be made for _not_ taking automatic instance zone bundles here until the cleanup stuff is in place. I think enough folks have lost the race against the zone-cleanup machinery when an instance dies that it would be valuable. But it's certainly not a clear-cut decision.

smklein: Yup, we can land them separately, but I wouldn't want to ship this PR to customers without that follow-up, since that'll make us hit our resource limits with no reprieve. 

However, if that's already on your radar, I'm okay punting it -- as long as we're on the same page about urgency.

sled-agent/src/instance_manager.rs
bnaecker: So this exact construction doesn't quite work, because the lock on the instance map is still in scope inside the if-let. However! We can use let-else instead:

```rust
        let Some((_propolis_id, instance)) = self
            .inner
            .instances
            .lock()
            .unwrap()
            .values()
            .find(|(propolis_id, _instance)| name == propolis_zone_name(propolis_id))
            .cloned()
        else {
            return Err(BundleError::NoSuchZone { name: name.to_string() });
        };
        instance.request_zone_bundle().await
```

I really do love that syntax addition.

smklein (approved):

sled-agent/src/instance.rs
```diff
@@ -151,7 +155,9 @@ fn fmri_name() -> String {
     format!("{}:default", service_name())
 }
 
-fn propolis_zone_name(id: &Uuid) -> String {
+/// Return the expected name of a Propolis zone managing an instace with the
```
gjcolombo, line 158: nit: s/instace/instance

gjcolombo (approved):

PR #3832, opened 2023-08-08
[bootstrap-agent] Refactor: Inner -> SprocketsServer

jgallagher:
This is a nearly-trivial refactor: move the `Inner` raw TCP server out of `bootstrap/server.rs` into its own module. This is building up to #3827.

### DIFF
--- a/sled-agent/src/bootstrap/server.rs
+++ b/sled-agent/src/bootstrap/server.rs
@@ -6,32 +6,22 @@
 
 use super::agent::Agent;
 use super::config::Config;
-use super::params::version;
-use super::params::Request;
-use super::params::RequestEnvelope;
-use super::views::Response;
-use super::views::ResponseEnvelope;
 use crate::bootstrap::http_entrypoints::api as http_api;
 use crate::bootstrap::maghemite;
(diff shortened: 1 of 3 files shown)


sled-agent/src/bootstrap/server.rs
```diff
@@ -93,17 +83,19 @@ impl Server {
...
-            log.new(o!("component" => "rack init server (BootstrapAgent)"));
-        let rack_init_server_handle = Inner::start_rack_init_server(
-            Arc::clone(&bootstrap_agent),
-            inner_log,
-        )
-        .await?;
+        // Start the currently-misnamed sprockets server, which listens for raw
+        // TCP connections (which should ultimately be secured via sprockets).
+        let sprockets_server =
+            SprocketsServer::bind(Arc::clone(&bootstrap_agent), &log)
```
smklein, line 89: part of me wonders if the complexity of managing this raw TCP server means we'd be better served by "just using a dropshot interface" here, but that's a thing we can think about / clean-up elsewhere.

smklein (approved):

PR #3833, opened 2023-08-08
[bootstrap-agent] Refactor rack_ops

jgallagher:
Moves it from under `bootstrap/agent/` to just under `bootstrap/`, and removes its dependency on `Agent`. Part of #3827.

### DIFF
--- a/sled-agent/src/bootstrap/rack_ops.rs
+++ b/sled-agent/src/bootstrap/rack_ops.rs
@@ -4,15 +4,18 @@
 
 //! Internal API for rack-level bootstrap agent operations.
 
-use super::Agent;
 use crate::bootstrap::http_entrypoints::RackOperationStatus;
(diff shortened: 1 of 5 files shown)


sled-agent/src/bootstrap/rack_ops.rs
```diff
@@ -164,9 +167,12 @@ impl RssAccess {
...
     }
 
-    pub(super) fn start_initializing(
+    pub(crate) fn start_initializing(
         &self,
-        agent: &Arc<Agent>,
+        parent_log: &Logger,
+        global_zone_bootstrap_ip: Ipv6Addr,
+        storage_resources: &StorageResources,
+        bootstore_node_handle: &bootstore::NodeHandle,
```
smklein, line 175: Makes sense, seems like the crux of the refactor is pulling these values out of the agent itself.

smklein (approved):

PR #3858, opened 2023-08-09
RandomnWithDistinctSleds region allocation strategy

### LINKED ISSUE #3702: Allocate Crucible Datasets On Separate Sleds
As of https://github.com/oxidecomputer/omicron/pull/3650 we allocate datasets randomly across the rack. We generate an error if two datasets on the same zpool are selected (not currently possible, may be in future if another crucible dataset is added for some purpose). We do not factor the sled into account at all. We should. This will ensure resiliency against sled-level failures.

After speaking with sean on the matter (see some conversation in 3650), we both think it should be possible to do this in SQL with a few query steps in the CTE. This looks like the following code, approximately, but im probably getting something slightly wrong.

First, select one random zpool from each sled:

```
let our_random_ordering = /* dont feel like including it here, see existing code */;

zpool_dsl
  .inner_join(
    /* the query we have for making sure the zpools have enough space */
  )
  .order_by((zpool_dsl::sled_id, our_random_ordering))
  .distinct_on(zpool_dsl::sled_id)
  .select((zpool_dsl::id,));
```

Then, select one random dataset from each zpool. Note that this step should resolve the problem we have currently with potentially picking 2 datasets from a single pool during the selection, eliminating the need for the special error for that case. We could break that out into its own PR if we wanted.

```
dataset_dsl
  .inner_join(
    sled_selection.query_source()
      .on(sled_selection_dsl::zpool_id.eq(dataset_dsl::zpool_id))
  )
  .order_by((dataset_dsl::zpool_id, our_random_ordering))
  .distinct_on(dataset_dsl::zpool_id)
  .select((dataset_dsl::id,))
```

This gives us our list of possible datasets. Each dataset is from a distinct sled and zpool, and is on a zpool with enough space. Now we select 3 random candidates

```
dataset_selection /* or whatever name */
  .query_source()
  .order(our_random_ordering)
  .limit(REGION_REDUNDANCY_THRESHOLD)
  .select((dataset_selection_dsl::id,))
```

and then we should finally have the datasets we want.

The change is not quite that simple though. We may have tests that currently rely on being able to select 3 pool on a single sled. Additionally, being able to select 3 pools on one sled is a very nice feature to have for development. So we also will need to

- Modify tests accordingly, which tests remains to be seen
- Add a way for devs to configure nexus to use a 1-sled allocation strategy instead of a 3-sled allocation strategy.

3650 introduced the concept of an allocation strategy at lowest levels of the region allocation code, but did not run it through the rest of the codebase, so there's currently no way to configure it. This is a good time to add a way to do that. I do not know where that configuration should go (RSS config file?), or how I should get that configuration from wherever we set it to the place region allocation happens.

faithanalog:
I am submitting this draft now, but I am not 100% sure that the configuration location I have selected for the allocation strategy (PackageConfig toml) is the correct one. It would work for getting stuff onto dogfood/prod, but I don't know how to use a different one when compiling a build to run locally for dev purposes. That's a problem, because as devs we're all mostly running "1-sled" setups, so region allocation would just Not Work for any of us. 

If you know how to configure this differently for a local dev build, then maybe we need docs for that.

If that's just not in the cards, then maybe it should go into DeploymentConfig instead.

Either way, the core logic of this is fundamentally done. The tests are in there. The integration tests pass. And configuration _works right now_ using PackageConfig, so it is in a state where someone other than me can adjust the way configuration works or add docs if necessary. I will be gone for about a week, so I invite anyone to make changes around that config if they want to do that to get this merged while I'm out.

---

PR #3650 introduced the Random region allocation strategy to allocate regions randomly across the rack. This expands on that with the addition of the RandomWithDistinctSleds region allocation strategy. This strategy is the same, but requires the 3 crucible regions be allocated on 3 different sleds to improve resiliency against a whole-sled failure.

The Random strategy still exists, and does not require 3 distinct sleds. This is useful in one-sled environments such as the integration tests, and lab setups. This PR adds the ability to configure the allocation strategy in the Nexus PackageConfig toml. Anyone running in a one-sled setup will need to configure that to one-sled mode (as is done for the integration test environment).

This also fixes a shortcoming of #3650 whereby multiple datasets on a single zpool could be selected. That fix applies to both the old Random strategy and the new RandomWithDistinctSleds strategy.

`smf/nexus/config-partial.toml` is configured for
RandomWithDistinctSleds, as that is what we want to use on prod.

As I mentioned, the integration tests are not using the distinct sleds allocation strategy. I attempted to add 2 extra sleds to the simulated environment but found that this broke more things than I had the understanding to fix in this PR. It would be nice in the future for the sim environment to have 3 sleds in it though, not just for this but for anything else that might have different behaviors in a multi-sled setup.

In the present, I have unit tests that verify the allocation behavior works correctly with cockroachdb, and we can try it out on dogfood.

Fixes #3702


faithanalog:
I can't find it in the logs but I am assuming that the `deploy` job is not particularly happy with the "requiring 3 sleds" situation

nexus/db-queries/src/db/datastore/mod.rs
```diff
@@ -719,48 +690,69 @@ mod test {
...
+        datasets
     }
 
     #[tokio::test]
     /// Note that this test is currently non-deterministic. It can be made
     /// deterministic by generating deterministic *dataset* Uuids. The sled and
     /// pool IDs should not matter.
-    async fn test_region_allocation() {
+    async fn test_region_allocation_strat_random() {
         let logctx = dev::test_setup_log("test_region_allocation");
```
smklein, line 747: Nit: log doesn't match test name

nexus/db-queries/src/db/datastore/mod.rs
```diff
@@ -788,8 +782,79 @@ mod test {
...
+        let _ = db.cleanup().await;
+        logctx.cleanup_successful();
+    }
+
+    #[tokio::test]
+    /// Test the [`RegionAllocationStrategy::RandomWithDistinctSleds`] strategy.
+    /// It should always pick datasets where no two datasets are on the same
+    /// zpool and no two zpools are on the same sled.
+    async fn test_region_allocation_strat_random_with_distinct_sleds() {
+        let logctx = dev::test_setup_log("test_region_allocation");
```
smklein, line 811: Nit: log doesn't match test name

nexus/db-queries/src/db/datastore/mod.rs
```diff
@@ -798,8 +863,8 @@ mod test {
                 assert!(disk_zpools.insert(dataset.pool_id));
 
                 // Must be 3 unique sleds
-                // TODO: When allocation chooses 3 distinct sleds, uncomment this.
```
smklein, line 801: Yay!

nexus/db-queries/src/db/datastore/mod.rs
```diff
@@ -815,14 +880,70 @@ mod test {
         logctx.cleanup_successful();
     }
 
+    #[tokio::test]
+    /// Ensure the [`RegionAllocationStrategy::RandomWithDistinctSleds`]
+    /// strategy fails when there aren't enough distinct sleds.
+    async fn test_region_allocation_strat_random_with_distinct_sleds_fails() {
+        let logctx = dev::test_setup_log("test_region_allocation");
```
smklein, line 887: nit log name

nexus/db-queries/src/db/queries/region_allocation.rs
```diff
@@ -309,15 +309,40 @@ impl CandidateZpools {
...
+            .query_source()
+            .inner_join(with_zpool)
+            .filter(it_will_fit)
+            .select((old_zpool_usage::dsl::pool_id,));
+
+        let query = if distinct_sleds {
+            let seed_bytes = seed.to_le_bytes();
+
+            let query: Box<dyn CteQuery<SqlType = candidate_zpools::SqlType>> =
+                Box::new(
```
smklein, line 325: (unrelated to this specific spot) I just wanna say, thanks again for going so deep into making this CTE magic work. This is a non-trivial query, and I like the way you've done it. Great work.

nexus/tests/config.test.toml
```diff
@@ -89,3 +89,8 @@ dns_external.max_concurrent_server_updates = 5
 # certificates it will take _other_ Nexus instances to notice and stop serving
 # them (on a sunny day).
 external_endpoints.period_secs = 60
+
+[default_region_allocation_strategy]
+# we only have one sled in the test environment, so we need to use the
+# `Random` strategy, instead of `RandomWithDistinctSleds`
```
smklein, line 95: For most of our tests, I think this makes sense.

smf/nexus/config-partial.toml
```diff
@@ -38,3 +38,8 @@ dns_external.max_concurrent_server_updates = 5
 # certificates it will take _other_ Nexus instances to notice and stop serving
 # them (on a sunny day).
 external_endpoints.period_secs = 60
+
+[default_region_allocation_strategy]
+# by default, allocate across 3 distinct sleds
+# seed is omitted so a new seed will be chosen with every allocation.
+type = "random_with_distinct_sleds"
```
smklein, line 45: So I chatted with @iliana about this earlier, and here's the TL;DR of our conversation:

- Most of the "non-deploy" tests work using the config from `nexus/tests/config.test.toml`
- The "real prod config" is using this file (`smf/nexus/config-partial.toml`), and uses the distinct sled strategy, as we want

So the problem cases are the following:
- Developers trying to run a single sled
- The "deploy job", which typically runs a "non-gimlet" standalone sled (similar to the developer workflow)

These cases are painful, because we want **some** config options to be like prod (e.g., we use internal DNS! We have an underlay, kinda!) but others we **don't** want to look like our prod environment (most obviously: we're only running with a single sled).

iliana and I came to the conclusion that a couple ways to solve this would be the following:

## Add a new target option to `omicron-package` (my slight preference)

1. Add a new feature to the omicron-package target type named `cluster`, which can be either `single-sled` or `multi-sled`. This would involve updating `package/src/lib.rs` and `package/src/target.rs`. There is **some** overlap with the `gimlet` vs `gimlet-standalone` vs `non-gimlet` options, but I think it's probably worth just adding a new feature explicitly.
2. Split the Nexus config files into two: what previously was `smf/nexus/config-partial.toml` becomes `smf/nexus/multi-sled/config-partial.toml` and `smf/nexus/single-sled/config-partial.toml`. We can provide a different region allocation strategy in each, through the `PackageConfig`.
3. Update `package-manifest.toml` to reference the right `smf` subdirectory for Nexus, based on the `cluster` value (you can look at how we do this for the `sled-agent` target using the `machine` target).

Pros: Plumbs Nexus-specific options directly to Nexus
Cons: Possible to select confusing combinations of targets, like "gimlet-standalone" + "multi-sled"?

## Try to overload an existing target?

1. Use the `non-gimlet` and `gimlet-standalone` options, and "infer" that these are single-sled
2. Pass an argument from sled agent to Nexus (via `DeploymentConfig`) when this is the case

Pros: One fewer target?
Cons: Prevents us from effectively having a multi-sled setup without real gimlets (I think this would break the Canada cluster), kinda confusing that a Nexus-level option is plumbed through the sled agent config

faithanalog: you've sold me on adding a new target option to omicron-package i think.

nexus/db-queries/src/db/datastore/mod.rs
faithanalog: thanks for catching these, i hadnt even noticed the log initialization was taking this string at all

faithanalog:
@smklein I named it `rack_topology` before realizing you had suggested `cluster` as a name. I can switch it to cluster if you'd like though- sorry about that. Anyways, if CI passes (I am writing this before it finishes), how do you like the shape of this change?

smf/nexus/config-partial.toml
gjcolombo: What `cluster` option will I want to pass if I want to set up a multi-sled dev cluster with exactly two sleds? It seems like the `single-sled` option should still work for now (though it makes me wonder if there's a clearer label we could put on it); is that correct?

iliana:
The deploy CI failure seems to be caused by:

```
SledAgent: failed to initialize services: Failed to initialize 3 zones:
      - oxz_nexus_9b1b8d22-21f8-4b93-b487-3fbd84fe42a8: Failed to create OPTE port for service nexus: Failure interacting with the OPTE ioctl(2) interface: command CreateXde failed: MacExists { port: "opte115", vni: Vni { inner: 100 }, mac: MacAddr { inner: A8:40:25:FF:E4:EE } }
      - oxz_nexus_d64d9e21-fd6e-4eb0-94a6-c5e50421aa1e: Failed to create OPTE port for service nexus: Failure interacting with the OPTE ioctl(2) interface: command CreateXde failed: MacExists { port: "opte114", vni: Vni { inner: 100 }, mac: MacAddr { inner: A8:40:25:FF:A5:B5 } }
      - oxz_nexus_d5e0ff69-9935-4733-a913-70d0eeb40f3e: Failed to create OPTE port for service nexus: Failure interacting with the OPTE ioctl(2) interface: command CreateXde failed: MacExists { port: "opte116", vni: Vni { inner: 100 }, mac: MacAddr { inner: A8:40:25:FF:9A:04 } }
    file = sled-agent/src/http_entrypoints.rs:321
    sled_id = 5ed96990-c0b7-4d25-9d48-513b19d67f21
```

which I'm not sure if that's related?

The `nexus_config::test::test_repo_configs_are_valid` test is also failing.

faithanalog:
finally CI is happy. but at what cost

smklein (approved):
Apologies on the delay for this - LGTM, thanks for working through the config issues.

faithanalog:
alright ill merge this monday

PR #3865, opened 2023-08-10
Bump propolis-server package dependency

pfmooney:
This pulls in the latest propolis-server package in order to include the fix for oxidecomputer/propolis#474.


leftwo (commented):
Do the pointers to propolis in `Cargo.toml` also need to be updated?

leftwo (approved):

PR #3873, opened 2023-08-11
Define a saga for instance start

### LINKED ISSUE #2824: Instance start needs to be a saga

### LINKED ISSUE #3813: Instance start does not independently ensure v2p/dpd state exists
Repro steps:

1. start rack
2. create and start instance
3. stop instance
4. power rack down
5. cold boot the rack
6. restart the instance

Expected: instances have appropriate internal and external connectivity
Observed: they don't; closer examination shows the relevant Dendrite NAT mappings are missing (V2P mappings might be missing as well though I'm not sure)

Workaround: start the instance, stop it, and start it again; connectivity should be restored on the second boot.

---

The reason this happens is that Nexus's `instance_start` function doesn't actually try to establish any of these mappings: https://github.com/oxidecomputer/omicron/blob/ed70f1c8c6a83fed0d6b5fed2ba6c033bdc18c5b/nexus/src/app/instance.rs#L478-L514

The mappings also aren't removed when the instance is stopped. This means that, before introducing cold boot into the picture, we've been able to rely on having programmed the correct networking state for any non-deleted instance--the state is created once when the instance is created and deleted only when the instance goes away.

Stopping and restarting the instance works around the problem because instance destruction triggers a Propolis generation change in the sled agent via a call to clear the instance's current migration IDs: https://github.com/oxidecomputer/omicron/blob/ed70f1c8c6a83fed0d6b5fed2ba6c033bdc18c5b/sled-agent/src/common/instance.rs#L354-L360

Nexus assumes that *any* Propolis generation change is a sign that an instance might have moved between sleds, and uses it as an opportunity to propagate all the correct networking state: https://github.com/oxidecomputer/omicron/blob/ed70f1c8c6a83fed0d6b5fed2ba6c033bdc18c5b/nexus/src/app/instance.rs#L1162-L1220

---

This should be fixed by having `instance_start` ensure that all the correct mappings are created on the appropriate sleds--i.e. it needs to call `create_instance_v2p_mappings` and `instance_ensure_dpd_config` on instance start instead of hoping the right things happen when the instance stops. Related (once again) are #2315 and #2824--this behavior will become especially important when an instance can go from having a sled assignment to not having a sled assignment to having a new, different sled assignment after a start-stop-start sequence. When networking state updates are propagated through RPWs, the start sequence can just trigger the relevant RPWs instead of doing the work itself.

Naively dropping the configuration calls into `instance_start` *almost* works, but there's a lurking race condition because of the way the "Starting" state gets applied:

1. T1 begins to start the instance
2. T2 begins to delete the instance & starts the deletion saga
3. T2's saga deletes the instance's mappings
4. T2's saga marks the instance as deleted
5. T1 recreates the mappings
6. T1 actually tries to start the instance and fails because it has been deleted

To make this safe, we need to ensure the instance has transitioned from Stopped to Starting in `instance_start` to block any subsequent deletion attempts before we actually program any networking state. That's more complicated than it sounds because Nexus currently relies on sled agent to push the instance to the Starting state--we'll need some code up in Nexus to help manage that state transition.

gjcolombo:
Create a saga that starts instances. This has the following immediate benefits:

- It's no longer possible to leak an instance registration during start; previously this could happen if Nexus crashed while handling a start call.
- The saga synchronizes properly with concurrent attempts to delete an instance; the existing start routine may not be handling this correctly (it can look up an instance and decide it's OK to start, then start talking to sled agent about it while a deletion saga runs concurrently and deletes the instance).
- The saga establishes networking state (Dendrite NAT entries, OPTE V2P mappings) for a newly started instance if it wasn't previously established. This is a stopgap measure to ensure that this state exists when restarting an instance after a cluster is restarted. It should eventually be replaced by a step that triggers the appropriate networking RPW(s).

This saga can also be used, at least in theory, as a subsaga of the instance create saga to replace that saga's logic for starting a newly-created instance. This work isn't done in this PR, though. (The change isn't trivial because the new start saga expects a prior instance record as a parameter, and the create saga can't construct *a priori* the instance record it intends to insert into CRDB.)

Tested via assorted new cargo tests and by launching a dev cluster with the changes, stopping an instance, restarting it, and verifying that the instance restarted correctly and that Nexus logs contained the expected log lines.

Fixes #2824. Fixes #3813.

### DIFF
--- a/nexus/src/app/instance.rs
+++ b/nexus/src/app/instance.rs
@@ -475,41 +475,43 @@ impl super::Nexus {
         self.db_datastore.instance_refetch(opctx, &authz_instance).await
     }
 
-    /// Make sure the given Instance is running.
+    /// Attempts to start an instance if it is currently stopped.
     pub async fn instance_start(
-        &self,
+        self: &Arc<Self>,
         opctx: &OpContext,
         instance_lookup: &lookup::Instance<'_>,
     ) -> UpdateResult<db::model::Instance> {
-        // TODO(#2824): This needs to be a saga for crash resiliency
-        // purposes (otherwise the instance can be leaked if Nexus crashes
-        // between registration and instance start).
-        let (.., authz_instance, mut db_instance) =
-            instance_lookup.fetch().await?;
+        let (.., authz_instance, db_instance) =
+            instance_lookup.fetch_for(authz::Action::Modify).await?;
 
-        // The instance is not really being "created" (it already exists from
-        // the caller's perspective), but if it does not exist on its sled, the
-        // target sled agent will populate its instance manager with the
-        // contents of this modified record, and that record needs to allow a
-        // transition to the Starting state.
-        //
-        // If the instance does exist on this sled, this initial runtime state
-        // is ignored.
-        let initial_runtime = nexus_db_model::InstanceRuntimeState {
-            state: nexus_db_model::InstanceState(InstanceState::Creating),
-            ..db_instance.runtime_state
+        // If the instance is already starting or running, succeed immediately
+        // for idempotency. If the instance is stopped, try to start it. In all
+        // other cases return an error describing the state conflict.
+        match db_instance.runtime_state.state.0 {
+            InstanceState::Starting | InstanceState::Running => {
+                return Ok(db_instance)
+            }
+            InstanceState::Stopped => {}
+            _ => {
+                return Err(Error::conflict(&format!(
+                    "instance is in state {} but must be {} to be started",
+                    db_instance.runtime_state.state.0,
+                    InstanceState::Stopped
+                )))
+            }
+        }
+
+        let saga_params = sagas::instance_start::Params {
+            serialized_authn: authn::saga::Serialized::for_opctx(opctx),
+            instance: db_instance,
+            ensure_network: true,
         };
-        db_instance.runtime_state = initial_runtime;
-        self.instance_ensure_registered(opctx, &authz_instance, &db_instance)
-            .await?;
 
-        self.instance_request_state(
-            opctx,
-            &authz_instance,
-            &db_instance,
-            InstanceStateRequested::Running,
+        self.execute_saga::<sagas::instance_start::SagaInstanceStart>(
+            saga_params,
         )
         .await?;
+
         self.db_datastore.instance_refetch(opctx, &authz_instance).await
     }
 
@@ -1220,7 +1222,7 @@ impl super::Nexus {
     }
 
     // Switches with uplinks configured and boundary services enabled
-    async fn boundary_switches(
+    pub async fn boundary_switches(
         &self,
         opctx: &OpContext,
     ) -> Result<HashSet<SwitchLocation>, Error> {
--- a/nexus/src/app/instance.rs
+++ b/nexus/src/app/instance.rs
@@ -487,6 +487,17 @@ impl super::Nexus {
         // If the instance is already starting or running, succeed immediately
         // for idempotency. If the instance is stopped, try to start it. In all
         // other cases return an error describing the state conflict.
+        //
+        // The "Creating" state is not permitted here (even though a request to
+        // create can include a request to start the instance) because an
+        // instance that is still being created may not be ready to start yet
+        // (e.g. its disks may not yet be attached).
+        //
+        // If the instance is stopped, the start saga will try to change the
+        // instance's state to Starting and increment the instance's state
+        // generation number. If this increment fails (because someone else has
+        // changed the state), the saga fails. See the saga comments for more
+        // details on how this synchronization works.
         match db_instance.runtime_state.state.0 {
             InstanceState::Starting | InstanceState::Running => {
                 return Ok(db_instance)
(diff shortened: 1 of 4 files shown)


nexus/src/app/instance.rs
```diff
@@ -475,41 +475,43 @@ impl super::Nexus {
...
+            }
+            InstanceState::Stopped => {}
+            _ => {
+                return Err(Error::conflict(&format!(
+                    "instance is in state {} but must be {} to be started",
+                    db_instance.runtime_state.state.0,
+                    InstanceState::Stopped
+                )))
+            }
+        }
```
smklein, line 502: Any concern of TOCTTOU here? Kinda seems like this state checking could/should be part of the saga?

nexus/src/app/instance.rs
```diff
@@ -475,41 +475,43 @@ impl super::Nexus {
...
+            instance_lookup.fetch_for(authz::Action::Modify).await?;
+
+        // If the instance is already starting or running, succeed immediately
+        // for idempotency. If the instance is stopped, try to start it. In all
+        // other cases return an error describing the state conflict.
+        match db_instance.runtime_state.state.0 {
+            InstanceState::Starting | InstanceState::Running => {
+                return Ok(db_instance)
+            }
+            InstanceState::Stopped => {}
```
smklein, line 494: The saga itself seems fine accepting instances in the `Creating` state -- is that not acceptable here?

nexus/src/app/sagas/instance_start.rs
```diff
@@ -0,0 +1,710 @@
...
+    let params = sagactx.saga_params::<Params>()?;
+    let osagactx = sagactx.user_data();
+    info!(osagactx.log(), "start saga failed; returning instance to Stopped";
+          "instance_id" => %params.instance.id());
+
+    let runtime_state =
+        sagactx.lookup::<db::model::InstanceRuntimeState>("starting_state")?;
+
+    // Don't just restore the old state; if the instance was being created, and
+    // starting it failed, the instance is now stopped, not creating.
```
smklein, line 190: Good call

nexus/src/app/sagas/instance_start.rs
```diff
@@ -0,0 +1,710 @@
...
+        instance_start_register_actions(registry);
+    }
+
+    fn make_saga_dag(
+        _params: &Self::Params,
+        mut builder: steno::DagBuilder,
+    ) -> Result<steno::Dag, SagaInitError> {
+        builder.append(mark_as_starting_action());
+        builder.append(dpd_ensure_action());
+        builder.append(v2p_ensure_action());
```
smklein, line 86: Would both of these steps just become "trigger the Nexus RPW for NAT" once https://github.com/oxidecomputer/omicron/pull/3804 merges?

smklein (approved):
Looks great, thanks for these tests!

nexus/src/app/instance.rs
gjcolombo: I think we have to be careful about that here. There's a small window in the create saga where the instance record is in CRDB (and has the Creating state) while the objects that support it (external IPs, attached disks, etc.) are still being set up, and we don't want to allow a start request to go through while we're in that state.

The saga accepts Creating as a prior state so that it can be used as a subsaga from within the instance create saga (even though that's not hooked up yet). I could buy, though, that a better approach is to have the create saga fully create the instance, then unconditionally move it to Stopped, and then have it invoke the start saga if needed. That would make everything very consistent at the cost of having a Creating instance (where the create request has `start: true`) briefly go to Stopped before entering Starting. That seems to me like a smallish price to pay (and arguably it's strictly better because it's more accurate). I'll mull this over, but WDYT of this approach?

nexus/src/app/sagas/instance_start.rs
gjcolombo: Yep! There's a comment on line 47 about this; I can file a tracking issue and link up to that issue number (will leave this thread open until I do).

nexus/src/app/instance.rs
gjcolombo: We should be covered here by the record's state generation: if the state changes, the saga will fail to transition from Stopped to Starting (because its generation number will be outdated) and will bail. This still isn't a perfect scheme, since it's possible for the saga to hit the "oops the state changed" condition even if the state has come to rest at Stopped. To completely fix all this I think we need to fix another bug first (bet you can guess which one...).

I'll add a comment here about the possible TOCTTOU and how we avoid it. There's another comment in the saga explaining the other remaining problems with the synchronization scheme in this PR.

gjcolombo:
Thanks as always for the review! I've added some comments in d59b4cf to cover the topics discussed above.

I'll address the remaining open question about whether the instance create saga should move instances to Stopped before starting them in whatever follow-on PR connects the create saga to the start saga.

nexus/src/app/instance.rs
smklein: I like that approach. I was considering other options, but they all kinda seems like they incur an equivalent CRDB write to make the instance transition from "being constructed, not visible" to "constructed enough that we can correctly start it", which is basically what you're also proposing with the intermediate stopped state.

gjcolombo: Yeah, having slept on it, I'm comfortable saying the value of having a uniform start path outweighs the benefit of going directly from Creating to Starting without very briefly passing through Stopping first. Filed #3883 to track the remaining work here.

PR #3878, opened 2023-08-12
Enforce unique field names between targets and metrics for timeseries

### LINKED ISSUE #2929: `oximeter` queries will fail when target and metric share a field name
The general method for querying `oximeter` timeseries is to specify a list of "selectors", which are criteria on the fields. For example `foo!="bar"` will select all timeseries with a string field named `foo` whose value differs from `"bar"`. As part of building the main select query, we find all the timeseries keys that are compatible with the set of selectors provided, in the [`SelectQuery::field_query()` method](https://github.com/oxidecomputer/omicron/blob/00221b4221026e3a2e2c4c84bf1e6325ab992cca/oximeter/db/src/query.rs#L533). That is then used in a later `INNER JOIN` with the measurements tables, with an equality condition on the timeseries keys.

However, this will produce an incorrect result when the target and metric share a field name. For example, we do this in the integration tests: both the target and metric have a field imaginatively named `name`. Issuing the "field query" for this case we see:

```
feldspar :) select distinct * from fields_string;

SELECT DISTINCT *
FROM fields_string

Query id: c8a87a24-e1bd-4d21-a1ab-4beb7e3fafa6

┌─timeseries_name───────────────────────┬──────timeseries_key─┬─field_name─┬─field_value─────────────┐
│ integration_target:integration_metric │ 6257040532384654857 │ name       │ integration-test-metric │
│ integration_target:integration_metric │ 6257040532384654857 │ name       │ integration-test-target │
└───────────────────────────────────────┴─────────────────────┴────────────┴─────────────────────────┘

2 rows in set. Elapsed: 0.003 sec.

feldspar :) select filter0.timeseries_key as timeseries_key, filter0.field_name, filter0.field_value, filter1.field_name, filter1.field_value from (select distinct * from oximeter.fields_string where timeseries_name = 'integration_target:integration_metric' and field_name = 'name') as filter0 inner join (select distinct * from oximeter.fields_string where timeseries_name = 'integration_target:integration_metric' and field_name = 'name') as filter1 on (filter0.timeseries_name = filter1.timeseries_name and filter0.timeseries_key = filter1.timeseries_key) order by (filter0.timeseries_name, filter0.timeseries_key);

SELECT
    filter0.timeseries_key AS timeseries_key,
    filter0.field_name,
    filter0.field_value,
    filter1.field_name,
    filter1.field_value
FROM
(
    SELECT DISTINCT *
    FROM oximeter.fields_string
    WHERE (timeseries_name = 'integration_target:integration_metric') AND (field_name = 'name')
) AS filter0
INNER JOIN
(
    SELECT DISTINCT *
    FROM oximeter.fields_string
    WHERE (timeseries_name = 'integration_target:integration_metric') AND (field_name = 'name')
) AS filter1 ON (filter0.timeseries_name = filter1.timeseries_name) AND (filter0.timeseries_key = filter1.timeseries_key)
ORDER BY (filter0.timeseries_name, filter0.timeseries_key) ASC

Query id: 4f16e459-9488-4d21-b76a-8fb3631e7885

┌──────timeseries_key─┬─field_name─┬─field_value─────────────┬─filter1.field_name─┬─filter1.field_value─────┐
│ 6257040532384654857 │ name       │ integration-test-metric │ name               │ integration-test-metric │
│ 6257040532384654857 │ name       │ integration-test-metric │ name               │ integration-test-target │
│ 6257040532384654857 │ name       │ integration-test-target │ name               │ integration-test-metric │
│ 6257040532384654857 │ name       │ integration-test-target │ name               │ integration-test-target │
└─────────────────────┴────────────┴─────────────────────────┴────────────────────┴─────────────────────────┘

4 rows in set. Elapsed: 0.008 sec.
```

Note that the `DISTINCT` clause is not part of the query, but helps illustrate the problem. We can see that the `INNER JOIN` produced 4 rows, while we might expect two based on the query itself. The issue in this case is that we don't (and _cannot_) constrain the selector to the `name` field from the target or the metric. We've lost that information when we stored the fields into the field tables without it. We actually _have_ that data in the schema table, but we can't reconstruct it after we've inserted the records into the table.

I think the solution is to actually keep the source in the field table as well.

bnaecker:
Resolves #2929


common/Cargo.toml
```diff
@@ -20,6 +20,7 @@ hyper.workspace = true
 ipnetwork.workspace = true
 macaddr.workspace = true
 lazy_static.workspace = true
+oximeter.workspace = true
```
smklein, line 23: Would it be better to reverse this? Namely, to have `oximeter` depend on `common`, rather than the other way around?

(This current construction seems like it prevents oximeter from depending on common in the future, which it might reasonably want to do)

oximeter/oximeter/src/types.rs
```diff
@@ -591,6 +598,22 @@ impl Sample {
...
     }
+
+    // Check validity of field names for the target and metric. Currently this
+    // just verifies there are no duplicate names between them.
+    fn verify_field_names(
+        target: &FieldSet,
+        metric: &FieldSet,
+    ) -> Result<(), MetricsError> {
+        for name in target.fields.iter().map(|f| f.name.as_str()) {
+            if metric.fields.iter().any(|f| f.name == name) {
```
smklein, line 609: WDYT about making the `fields` member of `FieldSet` a BTreeSet?

This operation seems like it could be O(N), but I think it's currently  O(N**2)? I dunno if N is large enough yet to make a significant difference, but if we're planning on doing lookups regularly, it might be worthwhile.

smf/sled-agent/non-gimlet/config-rss.toml
```diff
@@ -97,7 +97,7 @@ infra_ip_last = "192.168.1.30"
 # You can configure multiple uplinks by repeating the following stanza
 [[rack_network_config.uplinks]]
 # The gateway IP for the rack's external network
-gateway_ip = "192.168.1.199"
```
smklein, line 100: Is this your local config? Probably shouldn't be checked-in

smf/sled-agent/non-gimlet/config.toml
```diff
@@ -68,7 +68,7 @@ switch_zone_maghemite_links = ["tfportrear0_0"]
 data_links = ["net0", "net1"]
 
 [log]
-level = "info"
+level = "debug"
```
smklein, line 71: Let's remove this

common/Cargo.toml
bnaecker: I wondered the same thing. I'll swap it around, thanks.

smf/sled-agent/non-gimlet/config-rss.toml
bnaecker: Good catch, thanks. I keep stashing this, but obviously the muscle memory around `git add -u` is strong :)

PR #3895, opened 2023-08-17
Get instance ID from parameters in start saga undo

### LINKED ISSUE #3894: Instance start saga's undo steps refer to nonexistent "instance_id" node
Seen in dogfood. `sis_v2p_ensure_undo` and `sis_ensure_registered_undo` both try to get an instance ID from a `sagactx.lookup` that refers to a nonexistent node. They should get the ID from the saga params instead like all the other nodes.

The fix itself is very simple; a more interesting problem is why the action unwind test for this saga didn't catch the problem. Investigating that now.

gjcolombo:
The instance start saga has two undo actions that try to read an instance ID from a nonexistent prior action (I incorrectly copied them from the instance create saga). The saga unwind test missed this because it erroneously checked only that its saga invocations were failing and not that they were failing at the correct step.

Fix the test so that it can reproduce this issue, then make the undo nodes read the saga IDs from the correct place.

Tested: cargo test, both with the test fix and no code fix (fails as expected) and with both fixes (passes as expected).

Fixes #3894.

### DIFF
--- a/nexus/src/app/sagas/instance_start.rs
+++ b/nexus/src/app/sagas/instance_start.rs
@@ -18,7 +18,6 @@ use serde::{Deserialize, Serialize};
 use sled_agent_client::types::InstanceStateRequested;
 use slog::info;
 use steno::ActionError;
-use uuid::Uuid;
 
 /// Parameters to the instance start saga.
 #[derive(Debug, Deserialize, Serialize)]
@@ -193,10 +192,16 @@ async fn sis_move_to_starting_undo(
         ..runtime_state
     };
 
-    let _ = osagactx
+    if !osagactx
         .datastore()
         .instance_update_runtime(&params.instance.id(), &new_runtime)
-        .await?;
+        .await?
+    {
+        info!(osagactx.log(),
+              "did not return instance to Stopped: old generation number";
+              "instance_id" => %params.instance.id());
+    }
+
     Ok(())
 }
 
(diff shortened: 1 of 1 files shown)


nexus/src/app/sagas/instance_start.rs
```diff
@@ -426,7 +431,7 @@ async fn sis_ensure_registered_undo(
     let osagactx = sagactx.user_data();
     let params = sagactx.saga_params::<Params>()?;
     let datastore = osagactx.datastore();
-    let instance_id = sagactx.lookup::<Uuid>("instance_id")?;
```
smklein, line 429: For my own edification: The removal of this lookup (and also the one on line 374) are the "fixes" for this PR. Everything else is clarity + tests.

smklein (approved):

nexus/src/app/sagas/instance_start.rs
gjcolombo: Correct, this part is the fix; everything else is fixing the tests and making the logs a little easier to understand. (Probably shouldn't have squashed the commits.) With only the test fixes in place, the unwind test fails because of a panic during undo that matches what we saw in dogfood.

nexus/src/app/sagas/instance_start.rs
```diff
@@ -616,17 +622,37 @@ mod test {
...
-
         let params = Params {
             serialized_authn: authn::saga::Serialized::for_opctx(&opctx),
             instance: db_instance,
             ensure_network: true,
         };
-
         let dag = create_saga_dag::<SagaInstanceStart>(params).unwrap();
-        for node in dag.get_nodes() {
+        let num_nodes = dag.get_nodes().count();
```
luqmana, line 636: (an aside it'd be nice if `SagaDagIterator` implemented `ExactSizeIterator` and a more efficient `Iterator::nth`)

luqmana (approved):

gjcolombo:
Thanks for the reviews! I'll merge once CI is green and I've run this on a dev cluster.

nexus/src/app/sagas/instance_start.rs
smklein: Filed https://github.com/oxidecomputer/steno/issues/164

gjcolombo:
I set up a dev cluster, successfully tricked it into going through the start saga unwind path (confirmed via logs), and observed that Nexus successfully unwound the saga without panicking.

PR #3898, opened 2023-08-17
ci/tuf-repo: use the caboose board tag as SP/ROT artifact names

iliana:
As a follow-up to #3887, use the artifact's `name` field for the board field for SP/ROT artifacts. This will allow Nexus / sled-agent / whatever to make decisions on which artifacts to fetch from the repository in a world where the control plane is accessing a TUF repo over HTTPS.

The original intent of the `name` field was to provide some key for the control plane to make decisions on. In the original update system mockup this was for zone names (i.e. your `kind` would be `KnownArtifactKind::Zone`, and then your `name` would be something like `omicron-nexus`). Given hindsight I probably would have named this something like `key`, or `identifier`, or whatever. (Maybe we still ought to?)

I had completely forgotten that this is what `name` was for, so I've written a comment for future me, and hopefully others, explaining it in `common/src/update.rs`. (For a while I was convinced I needed to plumb an optional `board` field all the way through the control plane!)

This doesn't matter to Wicket today, as it has the entire repository present and can read cabooses directly, and that is probably a safer fallback option anyway.

### DIFF
--- a/.github/buildomat/jobs/tuf-repo.sh
+++ b/.github/buildomat/jobs/tuf-repo.sh
@@ -134,6 +134,19 @@ git init /work/dvt-dock
     git checkout FETCH_HEAD
 )
 
+caboose_util_rot() {
+    # usage: caboose_util_rot ACTION IMAGE_A IMAGE_B
+    output_a=$(/work/caboose-util "$1" "$2")
+    output_b=$(/work/caboose-util "$1" "$3")
+    if [[ "$output_a" != "$output_b" ]]; then
+        >&2 echo "\`caboose-util $1\` mismatch:"
+        >&2 echo "  $2: $output_a"
+        >&2 echo "  $3: $output_b"
+        exit 1
+    fi
+    echo "$output_a"
+}
+
 add_hubris_artifacts() {
     series="$1"
     rot_dir="$2"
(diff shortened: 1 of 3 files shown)


.github/buildomat/jobs/tuf-repo.sh
```diff
@@ -174,8 +182,8 @@ path = "$rot_image_a"
 kind = "file"
 path = "$rot_image_b"
 [[artifact.${tufaceous_board}_sp]]
-name = "${tufaceous_board}_sp"
-version = "$sp_version"
+name = "$sp_caboose_board"
```
jgallagher, line 185: I think this is fine but just want to note that the SP `BORD` values don't actually include the string "sp" - they're `gimlet-c`, `gimlet-d`, etc. The RoT `BORD` value is more clear (`oxide-rot-1`).

jgallagher (approved):

.github/buildomat/jobs/tuf-repo.sh
iliana: Yeah, I think that's fine, since the artifact kind makes it clear that this is an SP image. We need a key that the control plane can reliably use to determine which SP image is correct for the hardware it's trying to update, and making it match what it can read from what's already installed is I think the most reliable we can get right now.

Issue #249, opened 2021-09-16
TLS integration


smklein:
We have a number of services which will need TLS integration. This issue tracks that integration:

## Nexus

Nexus exposes an [external server](https://github.com/oxidecomputer/omicron/blob/main/omicron-nexus/src/http_entrypoints_external.rs) for servicing client requests, as well as an [internal server](https://github.com/oxidecomputer/omicron/blob/main/omicron-nexus/src/http_entrypoints_internal.rs) for handling requests from sled agents. Both should exclusively use encrypted traffic.

## Agents

[Sled Agent](https://github.com/oxidecomputer/omicron/blob/main/omicron-sled-agent/src/http_entrypoints.rs), [Bootstrap Agent](https://github.com/oxidecomputer/omicron/blob/main/omicron-bootstrap-agent/src/http_entrypoints.rs), [Oximeter](https://github.com/oxidecomputer/omicron/blob/main/oximeter/oximeter/src/oximeter_server.rs) and others (e.g., the to-be-created SP agent, possibly?) employ cross-machine communication, and will need TLS integration to securely communicate.

## CockroachDB

For development purposes, we use the [--insecure](https://github.com/oxidecomputer/omicron/blob/96cb81fef84b2f418a5eabfaf32b0569b05d4e56/omicron-common/src/dev/db.rs#L104) flag when deploying CockroachDB instances. However, in the limit, we should ensure that we can safely deploy Cockroach instances using exclusively [encrypted traffic](https://www.cockroachlabs.com/docs/stable/encryption.html#encryption-in-flight).

## Blocked on...

- [x] TLS integration into Dropshot
- [ ] Having a proper story for key storage and retrieval
- [x] Externally-exposed services utilizing TLS
- [ ] Internally-exposed services utilizing TLS
- [ ] CockroachDB communicating utilizing TLS

davepacheco:
Mentioning here that there's a TODO-security that mentions this ticket that we probably also want to resolve for MVP:
https://github.com/oxidecomputer/omicron/blob/b062e95f5f917909b8c6d40200a4d0d80847694f/nexus/src/authn/external/session_cookie.rs#L59-L60

davepacheco:
See also #2367 and #2368.

PR #3902, opened 2023-08-18
Add `Secure` attribute to session cookie

david-crespo:
We had a TODO item in the code pointing to #249, but no dedicated issue. The `Secure` attribute means the browser will only include the cookie with encrypted requests. This is a no-brainer table stakes kind of thing we _must_ do. This was blocked on adding TLS to Nexus. 

We use `deployment.dropshot_external.tls` to decide whether to include `Secure` so things will still work in non-TLS environments like local dev.

~~Because the tests do not run in a browser, I think they should be indifferent to the `Secure` attribute, but I'll find out when CI runs.~~

~~The browser ignores this on `localhost`, so for a simple local Nexus instance without TLS this shouldn't be a problem. However, if people are running Nexus without TLS for other test and dev setups, that might be a problem — the web console won't work. I think we could get around that by looking at the config value for `deployment.dropshot_external.tls` and only including `Secure` if it's true.~~


davepacheco:
This seems like a good change.

> The browser ignores this on localhost, so for a simple local Nexus instance without TLS this shouldn't be a problem. However, if people are running Nexus without TLS for other test and dev setups, that might be a problem — the web console won't work. I think we could get around that by looking at the config value for deployment.dropshot_external.tls and only including Secure if it's true.

I think we might want to do this?  I don't know about other people but I definitely do run Omicron that way.  We could also lean into making even dev setups use TLS, but that requires a more real DNS setup that's a little tricky.

david-crespo:
Ok, I'll add the check for the config value. I agree local TLS is too much, at least for right now.

nexus/src/external_api/console_api.rs
```diff
@@ -432,6 +433,7 @@ pub async fn login_local(
                 // browser expiration is mostly for convenience, as the API will
                 // reject requests with an expired session regardless
                 apictx.session_absolute_timeout(),
+                apictx.external_tls_enabled,
```
david-crespo, line 436: I thought about having `session_cookie_header_value` take the whole `apictx`, but then the unit tests become less trivial, and it's nice for unit tests to be trivial.

nexus/src/context.rs
```diff
@@ -49,6 +49,8 @@ pub struct ServerContext {
     pub external_latencies: LatencyTracker,
     /// registry of metric producers
     pub producer_registry: ProducerRegistry,
+    /// TLS enabled on the external Dropshot server
+    pub external_tls_enabled: bool,
```
david-crespo, line 53: It feels a little sad to clutter up the context with another config value, but I wasn't sure whether I should get the answer from `nexus` itself because of the `Arc` around it. A method on `nexus` would have to look something like this. @davepacheco?

```rs
pub async fn external_tls_enabled(&self) -> bool {
    let external_server = self.external_server.lock().unwrap().take();
    if let Some(server) = external_server {
        return server.using_tls();
    }
    false
}
```

davepacheco: I think it's okay as-is.  One advantage of what's here over that method impl is that if someone asks the question before we've started the external server for some reason, it will report what's configured (which seems right) rather than panicking.

It seems like Dropshot could use a method on the `RequestContext` for this.  Then you could just check that in the handler.

davepacheco (approved):

PR #3905, opened 2023-08-18
fix how-to-run flow

davepacheco:
A few of us ran into some issues trying to setup Omicron from "main".

First, we misread the instructions and thought we did not need to set PXA_START/PXA_END if we're using the fake sled-local external network.  Then we had no proxy ARP and so no external connectivity worked.  This manifests as the `dig` step (looking up the Nexus IPs from the external DNS server) timing out.

This was after working around a `create_virtual_hardware.sh` failure:

```
+ dladm show-phys fake_external_stub0
dladm: failed to show physical link fake_external_stub0: invalid argument
+ return
```

I think the intent of this code was to return _success_ early if given a non-physical link.  But it just does `|| return`, which returns the previous (failure) code, which causes the script to exit (due to errexit).  But based on the above and feedback from @luqmana in chat, it sounds like you _do_ generally want to set PXA_START/PXA_END in this case.  So I'm just removing this condition altogether.

### DIFF
--- a/docs/how-to-run.adoc
+++ b/docs/how-to-run.adoc
@@ -104,9 +104,10 @@ As an example, in this doc:
 +
 * The network is 192.168.1.0/24.
 * The network's gateway is 192.168.1.199.
-* You can carve off 192.168.1.20 - 192.168.1.29 for externally-facing services provided by the system.
-* You can carve off 192.168.1.30 as the address used by SoftNPU on the external network.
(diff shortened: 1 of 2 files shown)


davepacheco:
This worked for me on maxwell using the fake external network config.

luqmana (approved):
Thanks for updating this to make things more clear!

Issue #3198, opened 2023-05-23
RSS / Rack Initialize: Fail fast and explicitly


rcgoodfellow:
When we send the rack-initialize call, before RSS sends out the signal for bootstrap agents to initialize the rack, there should be a preliminary check done first. In this check, RSS will contact each of the sleds it expects to be in the cluster on the bootstrap addresses provided in the rack initialize call. The RSS agent will ask each of the sled agents to verify that it's ready to initialize the rack. The individual sled/bootstrap agents then check all the preconditions for initializing the sled and report back to RSS. If there are failures, all of them should be reported back to the caller so they can take remedial action. This includes not being able to contact expected sleds. This should be a fast operation, on the order of a few seconds.

cc @andrewjstone

andrewjstone:
This will be done automatically by the trust quorum, which will run before any other RSS initialization. Bootstrap addresses will get passed and shares will get distributed and written to the M.2s before any other setup is done. A timeout will be set, such that if it doesn't complete we will report the failure, debug, and eventually reset the rack and try again.

askfongjojo:
@rcgoodfellow: Can we consider this closed now that trust quorum is in place?

rcgoodfellow:
Having a trust quorum has us in a much better place for the fast part. I think there is still ground to cover for the explicit part. At any time during rack initialization, the operator should be able to ask for a rack initialization report that includes pending and/or failed initialization tasks. E.g. operators should be able to explicitly ask why is the rack not initialized yet and get an explicit answer.

jgallagher:
Agreed with @rcgoodfellow - the only operartor-visible feedback is wicket's `Initializing` (with no further info). In addition to the internal-to-the-rack checks that trust quorum does, if RSS is stuck or doomed without external changes (to pull some possible problems out of a hat: our transceivers are failing to uplink, or we're stuck waiting on timesync because we're firewalled off from the NTP servers we're supposed to sync with), there's no way to know that without being logged into the support account and digging around.

PR #3929, opened 2023-08-22
[wicket] Add `wicket preflight uplink` check

jgallagher:
The goal of this is to aid in initial installation: we expect to see various issues with uplink connections and configuration (at multiple layers: can we get an L1 optical link? once we have an L1 link, can we talk to the external DNS/NTP servers?) that can impact rack setup in a way that is difficult to diagnose and debug while rack setup is running. Instead, _prior to running rack setup_, we can run this new `ssh -t wicket@... preflight uplink` check: it will walk through setting up the uplink and talking to the DNS and NTP servers more or less the same way a person (or RSS) would.

There are (at least) two major shortcomings with the implementation of this PR:

1. This is entirely an external / CLI-driven check; ideally we could drive this from within the wicket TUI.
2. It only performs an uplink check for the _current switch_. To check both switches, an operator must upload the RSS config to both wickets and then ssh into each wicket to run this check. This isn't hard but is annoying; ideally ssh'ing into one wicket would run checks on both scrimlets.

But I think this is already useful as-is.

This covers a portion (but not the entirety of) #3198.


andrewjstone (approved):
This is absolutely EPIC! It's a complicated set of steps that reads mostly linearly thanks to judicious use of the badass update engine by @sunshowers

Issue #3889, opened 2023-08-16
Test flake: silent failure to initialize cockroachdb?


jgallagher:
https://github.com/oxidecomputer/omicron/pull/3887/checks?check_run_id=15946171931 failed; the immediate symptom in the deploy job a timeout after trying to log in for 10 minutes:

```
673	2023-08-16T15:51:53.331Z	2023-08-16 15:51:52.586682700 UTC: login failed: logging in: error sending request for url (https://recovery.sys.oxide.test/v1/login/recovery/local): error trying to connect: operation timed out
674	2023-08-16T15:51:54.332Z	2023-08-16 15:51:53.587485600 UTC: attempting to log into API
675	2023-08-16T15:52:09.366Z	2023-08-16 15:52:08.621092329 UTC: login failed: logging in: error sending request for url (https://recovery.sys.oxide.test/v1/login/recovery/local): error trying to connect: operation timed out
676	2023-08-16T15:52:10.368Z	2023-08-16 15:52:09.622786274 UTC: attempting to log into API
677	2023-08-16T15:52:25.402Z	2023-08-16 15:52:24.656701751 UTC: login failed: logging in: error sending request for url (https://recovery.sys.oxide.test/v1/login/recovery/local): error trying to connect: operation timed out
678	2023-08-16T15:52:26.403Z	2023-08-16 15:52:25.657578343 UTC: attempting to log into API
679	2023-08-16T15:52:41.436Z	2023-08-16 15:52:40.691037367 UTC: login failed: logging in: error sending request for url (https://recovery.sys.oxide.test/v1/login/recovery/local): error trying to connect: operation timed out
680	2023-08-16T15:52:42.439Z	Error: logging in
681	2023-08-16T15:52:42.464Z	
682	2023-08-16T15:52:42.489Z	Caused by:
683	2023-08-16T15:52:43.040Z	    timed out after 609.325828065s
```

All three nexus logs have many warning that seem to indicate the database wasn't set up correctly, I think?

```
WARN	nexus: Cannot look up rack: Object (of type ById(24ed7902-3649-4e0e-8635-cd082ae9b0c0)) not found: rack
    file = nexus/src/app/rack.rs:555
```

but the sled-agent logs do not report any issues with initial CRDB setup:

```
328	2023-08-16T15:43:19.640Z	INFO	SledAgent (ServiceManager): Formatting CRDB
    file = sled-agent/src/services.rs:2248
...
[...]

PR #3933, opened 2023-08-23
service IP pool lookup returns wrong authz error

davepacheco:
I noticed this while reading code debugging #3889.  Basically, it's a wrong pattern to `map_err()` the result of `authorize()` to replace a `Forbidden` error with a `NotFound`.  Distinguishing between these two is the responsibility of the authz subsystem because it should be consistent for all callers of `authorize()`.  If we want it to be a NotFound, the authz subsystem supports that via the impl of `AuthorizedResource::on_unauthorized()` for the resource.  Concretely, we could preserve the existing behavior (while still delegating it to the authz subsystem) by customizing the impl for `IpPoolList`:
https://github.com/oxidecomputer/omicron/blob/94cde6d72d2603a3c5160b09d2cd04d958ae1881/nexus/db-queries/src/authz/api_resources.rs#L433-L441
to look more like the one for most non-synthetic resources (which do return NotFound when someone's not authorized to read them):
https://github.com/oxidecomputer/omicron/blob/94cde6d72d2603a3c5160b09d2cd04d958ae1881/nexus/db-queries/src/authz/api_resources.rs#L132-L154

I started looking at that when I saw that we're doing a `ListChildren` here, not a `Read`.  The resource itself ("the list of service IP pools") is _not_ really hidden from end users -- the URL is even published in the OpenAPI spec.  So I think it makes more sense to return Forbidden here, just like if you try to list IP pools when not authorized to do so.

### DIFF
--- a/nexus/db-queries/src/db/datastore/ip_pool.rs
+++ b/nexus/db-queries/src/db/datastore/ip_pool.rs
@@ -114,8 +114,6 @@ impl DataStore {
     ) -> LookupResult<(authz::IpPool, IpPool)> {
         use db::schema::ip_pool::dsl;
 
-        // Ensure the caller has the ability to look up these IP pools.
-        // If they don't, return "not found" instead of "forbidden".
         opctx
             .authorize(authz::Action::ListChildren, &authz::IP_POOL_LIST)
             .await?;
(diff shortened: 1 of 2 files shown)


nexus/db-queries/src/db/datastore/ip_pool.rs
```diff
@@ -118,14 +118,7 @@ impl DataStore {
         // If they don't, return "not found" instead of "forbidden".
```
david-crespo, line 118: This comment is no longer accurate

david-crespo (approved):
Looks good, reasoning makes sense. I also don't see a reason for special handling.

nexus/db-queries/src/db/datastore/ip_pool.rs
davepacheco: Good catch.

PR #3935, opened 2023-08-23
Add MPL-2.0 license to a handful of crates

jgallagher:
I was doing some experiments with `cargo-deny` and it noted a handful of crates in `omicron` that were missing licenses in the crate metadata. This fixes them.

`cargo-deny` might be too heavy handed to add to CI for now; it is unhappy with many of our dependencies, at least with the default config.

### DIFF
--- a/tufaceous/Cargo.toml
+++ b/tufaceous/Cargo.toml
@@ -2,6 +2,7 @@
 name = "tufaceous"
 version = "0.1.0"
 edition = "2021"
+license = "MPL-2.0"
 publish = false
 
 [dependencies]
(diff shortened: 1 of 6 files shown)


tufaceous/Cargo.toml
```diff
@@ -2,6 +2,7 @@
 name = "tufaceous"
 version = "0.1.0"
 edition = "2021"
+license = "MPL-2.0"
 publish = false
```
iliana, line 6: We may also want to ensure `publish = false` is present everywhere too (it prevents an accidental `cargo publish`). Although the usefulness of that for us is debatable, so probably not for this PR. (It's in tufaceous because I have a habit of adding it to every Cargo.toml I `init`.)

iliana (approved):

PR #3940, opened 2023-08-23
[wicketd] Add step to ensure RoT update landed before proceeding

### LINKED ISSUE #3937: wicketd: After updating+resetting an RoT, ensure we can communicate with it before proceeding
Today in wicketd, the final step in updating an RoT is [resetting it](https://github.com/oxidecomputer/omicron/blob/b644199132b316beb71c5bafad59b8cb0e6fbb87/wicketd/src/update_tracker.rs#L1934-L1951). This will wait until the SP has responded, but the SP [expects wicketd to check that the newly-installed image is actually running](https://github.com/oxidecomputer/hubris/blob/f88d569b36d96e30206ce9eace46d39216c4ddce/task/control-plane-agent/src/mgs_common.rs#L244-L245). We should do that before proceeding with updating the SP.

jgallagher:
Example failure when using an incorrectly-signed RoT image, causing bootleby to fall back to the original slot:

madrid

Fixes #3937.

### DIFF
--- a/wicket-common/src/update_events.rs
+++ b/wicket-common/src/update_events.rs
@@ -110,6 +110,7 @@ pub enum SpComponentUpdateStepId {
     Writing,
     SettingActiveBootSlot,
     Resetting,
+    CheckingActiveBootSlot,
 }
 
 impl StepSpec for SpComponentUpdateSpec {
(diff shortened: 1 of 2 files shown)


andrewjstone (approved):
Looks great! Thanks @jgallagher.

Issue #3939, opened 2023-08-23
convert `tools/install_prerequisites.sh` to an xtask


jordanhendricks:
A critical part of the development experience to build and run omicron involves ensuring the system in question has necessary packages (per the OS + distro's local package manager) and dependencies (binaries downloaded directly into the local workspace, such as clickhouse or cockroachdb).

Currently, packages and dependencies are installed by a pair of bash scripts, `tools/install_{builder,runner}_prerequisites.sh`, with the former being for a machine to build omicron, and the latter to run it (either with a simulated sled-agent or not). The script `tools/install_prerequisites.sh` runs both of these scripts, if you want to both build and run omicron on the same machine.

There is some notion of supported OSes in these tools (namely, macOS, linux, and helios), but there are limitations in that they assume a particular package manager, which forces folks to work around things by commenting out lines of the script or running bits manually.

I am going to take a stab at converting these tools into a single xtask command, if nothing else to get these scripts into rust and give us the many benefits related to better error handling, type safety, and ease of logging there. Some additional goals/features I have in mind are:
* the ability to inspect a system to determine whether it has the dependencies and packages needed already
* dry run options so that I can run the tool and know what it will install/do before I do it
* move dependency definitions (and maybe package definitions) into a TOML file so that changes to deps/packages are modifying a configuration file instead of code

There are several paper cuts I've learned of when researching this topic, including:
- #3938
- #2181 
- having to hack around the tool to use a different package manager for non-debian like distros (the scripts assume `apt`)

smklein:
With some recent changes, I bumped into this via https://github.com/oxidecomputer/omicron/issues/5401 . I'm happy to help pick this up, specifically for "create/destroy" virtual hardware

jordanhendricks:
Sure, if you'd like I can dig up my old branch and send your way?

smklein:
Sure! I think the virtual hardware things have shifted a bit, but I could also pick up the "prerequisite installation", which I haven't started on yet.

smklein:
I'm being slow today - I realized this whole issue is specifically about "installation" and setup, not virtual hardware creation, which is a separate mess of bash scripts (that I bumped into). I'm still interested in picking this up! But I have not started on any part of it, and I think all PRs you had here would still be relevant.

PR #3945, opened 2023-08-24
tools/ directory could use a README

jordanhendricks:
I find the sheer number of files in the `tools/` directory pretty overwhelming. Once #3939 lands, depending on how I end up structuring things, a number of these files should eventually be removed. At that point, it would be worth looking at whether we could add a bit more organization in terms of directory structure, which would also help. In the meantime, I figured I would write down my understanding of how these pieces fit together.

The tools here are used by a variety of people for lots of different things and it's totally possible I misrepresented something (or otherwise missed something big and useful to document): feedback welcome!

(For reviewer convenience, the rendered version of this change is [here](https://github.com/jordanhendricks/omicron/tree/doc-tools/tools).)

### DIFF
--- a/tools/README.adoc
+++ b/tools/README.adoc
@@ -1,0 +1,82 @@
+:showtitle:
+:toc: left
+:icons: font
+
+= Omicron Tools
+
+This directory is home to a number of tools related to Omicron development.
+There are some additional developer tools implemented as xtasks: Run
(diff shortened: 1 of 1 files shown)


davepacheco (approved):
Nice!  Thanks for doing this.

PR #3947, opened 2023-08-24
Fix ensuring snapshot creation and deletion

jmpesp:
If `project_ensure_snapshot` is passed a snapshot where the project and name matches a record that exists already, but the ID is different, then

    .on_conflict((dsl::project_id, dsl::name))
    .filter_target(dsl::time_deleted.is_null())
    .do_update()
    .set(dsl::time_modified.eq(dsl::time_modified)),

will incorrectly modify the existing record, leading callers of that function to erroneously believe they were the one to create that snapshot record. Change `project_ensure_snapshot` to check if an existing snapshot record has a matching project and name, and return ObjectAlreadyExists if so.

It's not appropriate to delete a snapshot in any state: if it's in state Creating, then a `snapshot_create` saga is working on it and concurrently running `snapshot_delete` will cause a conflict. Add an argument to `project_delete_snapshot` that describes the states that a snapshot can be deleted in, depending on on the call site.

### DIFF
--- a/nexus/db-queries/src/db/datastore/snapshot.rs
+++ b/nexus/db-queries/src/db/datastore/snapshot.rs
@@ -19,9 +19,14 @@ use crate::db::model::Snapshot;
 use crate::db::model::SnapshotState;
 use crate::db::pagination::paginated;
 use crate::db::update_and_check::UpdateAndCheck;
+use crate::db::TransactionError;
+use async_bb8_diesel::AsyncConnection;
 use async_bb8_diesel::AsyncRunQueryDsl;
+use async_bb8_diesel::ConnectionError;
 use chrono::Utc;
 use diesel::prelude::*;
+use diesel::result::Error as DieselError;
+use nexus_types::identity::Resource;
 use omicron_common::api::external::http_pagination::PaginatedBy;
 use omicron_common::api::external::CreateResult;
 use omicron_common::api::external::Error;
@@ -32,11 +37,6 @@ use omicron_common::api::external::UpdateResult;
 use omicron_common::bail_unless;
 use ref_cast::RefCast;
 use uuid::Uuid;
-use nexus_types::identity::Resource;
-use async_bb8_diesel::AsyncConnection;
-use crate::db::TransactionError;
-use async_bb8_diesel::ConnectionError;
-use diesel::result::Error as DieselError;
 
 impl DataStore {
     pub async fn project_ensure_snapshot(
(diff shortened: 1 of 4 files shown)


nexus/db-queries/src/db/datastore/snapshot.rs
```diff
@@ -43,28 +48,127 @@ impl DataStore {
...
-            AsyncInsertError::CollectionNotFound => Error::ObjectNotFound {
-                type_name: ResourceType::Project,
-                lookup_type: LookupType::ById(project_id),
-            },
-            AsyncInsertError::DatabaseError(e) => {
-                public_error_from_diesel_pool(e, ErrorHandler::Server)
-            }
-        })?;
+
+        let snapshot: Snapshot = self
```
smklein, line 65: (Looking at the old `Project::insert_resource` code) Could we do this without the transaction? `diesel::insert_into` should return the inserted row, which goes through `insert_and_get_result_async`.

It seems possible to call either:

- [insert_and_get_result_async](https://github.com/oxidecomputer/omicron/blob/9818f6fcf5f3f0019c1571e1bec9d3375d8867b2/nexus/db-queries/src/db/collection_insert.rs#L191), and check for "not found" on the result, or to call
- [insert_and_get_results_async](https://github.com/oxidecomputer/omicron/blob/9818f6fcf5f3f0019c1571e1bec9d3375d8867b2/nexus/db-queries/src/db/collection_insert.rs#L213C18-L213C46), and to check for an "empty vec" as the result (aka, nothing was inserted).

Something like:

```rust
       let snapshots: Vec<Snapshot> = Project::insert_resource(
            project_id,
            diesel::insert_into(dsl::snapshot)
                .values(snapshot)
                .on_conflict((dsl::project_id, dsl::name))
                .filter_target(dsl::time_deleted.is_null())
                .do_nothing(),
        )
        .insert_and_get_results_async(self.pool_authorized(opctx).await?)
        .await
        .map_err(|e| match e {
            AsyncInsertError::CollectionNotFound => Error::ObjectNotFound {
                type_name: ResourceType::Project,
                lookup_type: LookupType::ById(project_id),
            },
            AsyncInsertError::DatabaseError(e) => {
                public_error_from_diesel_pool(e, ErrorHandler::Server)
            }
        })?;
    let Some(snapshot) = snapshots.get(0) else {
         // Return an error -- we failed to insert the snapshot.
    };
```

jmpesp: I originally tried `insert_and_get_result_async` with the additional `filter` (once I got it working), but I saw that it returned `CollectionNotFound` instead of `DatabaseError`. I'm trying the second suggestion now.

nexus/db-queries/src/db/datastore/snapshot.rs
```diff
@@ -48,127 +44,49 @@ impl DataStore {
...
+        let snapshot_id = snapshot.id();
+
+        let mut snapshots: Vec<Snapshot> = Project::insert_resource(
+            project_id,
+            diesel::insert_into(dsl::snapshot)
+                .values(snapshot)
+                .on_conflict((dsl::project_id, dsl::name))
+                .filter_target(dsl::time_deleted.is_null())
+                .do_update()
+                .set(dsl::time_modified.eq(dsl::time_modified))
```
smklein, line 61: Do we still want to update the old record in this case?

jmpesp: Definitely not, but I'm at a loss as how to do that.

smklein: Can we use `on_conflict().do_nothing()`?

jmpesp: If I do

```rust
        let mut snapshots: Vec<Snapshot> = Project::insert_resource(
            project_id,
            diesel::insert_into(dsl::snapshot)
                .values(snapshot)
                .on_conflict((dsl::project_id, dsl::name))
                .filter_target(dsl::time_deleted.is_null())
                .do_nothing()
                .filter(dsl::id.eq(snapshot_id)),
        )
        .insert_and_get_results_async(self.pool_authorized(opctx).await?)
```

then I see a syntax error:

```
thread 'integration_tests::snapshots::test_create_snapshot_record_idempotent' panicked at 'called `Result::unwrap()` on an `Err` value: InternalError { internal_message: "unexpected database error: at or near \"where\": syntax error" }', nexus/tests/integration_tests/snapshots.rs:996:10
```

If I remove the `.filter(dsl::id.eq(snapshot_id)),` on the end of that (so it's just `on_conflict + do_nothing`, I see `test_create_snapshot_record_idempotent` failing on the second call here:

```rust
    // Test project_ensure_snapshot is idempotent

    let snapshot_created_1 = datastore
        .project_ensure_snapshot(&opctx, &authz_project, snapshot.clone())
        .await
        .unwrap();
                
    let snapshot_created_2 = datastore
        .project_ensure_snapshot(&opctx, &authz_project, snapshot.clone())
        .await
        .unwrap();
```

saying the snapshot already exists:

```
thread 'integration_tests::snapshots::test_create_snapshot_record_idempotent' panicked at 'called `Result::unwrap()` on an `Err` value: ObjectAlreadyExists { type_name: Snapshot, object_name: "snapshot" }', nexus/tests/integration_tests/snapshots.rs:1001:10
```

jmpesp:
We spoke in DMs, and I reverted to the transaction to avoid modifying the existing snapshot's `time_modified` column.

smklein (approved):

Issue #881, opened 2022-04-05
`integration_tests::disks::test_disk_create_disk_that_already_exists_fails` possibly flaky


smklein:
See: https://github.com/oxidecomputer/omicron/pull/865#issuecomment-1089265936

This merits (at least!) stress testing this particular test, to check if this is reproducible.

PR #879, opened 2022-04-05
[cleanup] Rename partition to dataset

smklein:
Fixes https://github.com/oxidecomputer/omicron/issues/878

Issue #878, opened 2022-04-05
[naming] s/partition/dataset


smklein:
See: https://github.com/oxidecomputer/omicron/pull/876/files/7cb0555ff91fa47209c9abf9ba6b615d7b7d6554#r842964923

names are hard

PR #876, opened 2022-04-05
[sled-agent] Decouple RSS from Bootstrap Agent, run in distinct task

smklein:

PR #3949, opened 2023-08-24
Update Crucible/Propolis

leftwo:
Updated sled-agent to no longer expect a
propolis_client::instance_spec::VolumeConstructionRequest when it is now just a Crucible VolumeConstructionRequest.

Changes in Crucible:
eliminate spurious metadb-related syncs (#881)
ACK the write after adding it to the work queue (#874) Use arc4random_buf or getrandom instead of ChaCha20Rng (#878) Fix crutest in README (#879)
Show client_id in panic messages (#843)
Reduce sqlite page cache size to 64KiB (#876)
Only flush extents that might actually be dirty. (#875)

Changes in Propolis:
Clean up and restructure CQE handling in NVMe
Improve error message when /dev/vmmctl not present (#506) new API endpoint for propolis-server to replace a crucible downstairs (#495) Update rustls-webpki for GHSA-fh2r-99q2-6mmg
Add MemAsync block backend
try reenabling PHD jobs (#484)
Define versioned instance specs (#472)
Update toml dependency to 0.7.x
Add more USDT probes to NVMe emulation
Add more to standalone-with-crucible (#490)
Update propolis with new Crucible Volume change (#485) Minor polish to standalone-crucible doc
Clean up bits for crucible in propolis-standalone
doc iteration for crucible and propolis-standalone Skeleton docs for using propolis with crucible disks in isolation propolis-standalone: Update expected crucible opts (#488) Split up README content for server and standalone
Add crucible config to propolis-standalone
Use libstd-provided OnceCell equivalent
Allow 64-vCPU instances on Helios (stlouis)
Elide test (and doctest) steps where not required
Clean up NVMe PRP parsing and add tests

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -489,7 +489,7 @@ dependencies = [
 [[package]]
 name = "bhyve_api"
 version = "0.0.0"
-source = "git+https://github.com/oxidecomputer/propolis?rev=77a580b7e2d3fee85ffe8513a86d2c464b7b5ed3#77a580b7e2d3fee85ffe8513a86d2c464b7b5ed3"
+source = "git+https://github.com/oxidecomputer/propolis?rev=4f371e971f9e33f5c02618b2ee95ad03cd531135#4f371e971f9e33f5c02618b2ee95ad03cd531135"
 dependencies = [
  "bhyve_api_sys",
  "libc",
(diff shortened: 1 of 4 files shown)


gjcolombo (approved):
LGTM--thanks for working through the build break.

PR #3950, opened 2023-08-24
[wicketd] switch to a watch channel for installinator reports

sunshowers:
This was originally changed to an unbounded receiver in #3579, but in
the comments to RFD 400 it was pointed out that maybe we shouldn't give
up on backpressure. Switch to a watch channel, since we only care about
the last element anyway.

(This isn't a hugely pressing issue, but I want to include it as a case
study in RFD 400.)

Also treat the receiver being closed as HTTP 410 Gone. We don't do
anything with this at the moment, but could abort installinator on
receiving this signal in the future.

### DIFF
--- a/wicketd/src/installinator_progress.rs
+++ b/wicketd/src/installinator_progress.rs
@@ -49,6 +49,8 @@ pub(crate) struct IprArtifactServer {
 }
 
 impl IprArtifactServer {
+    const CHANNEL_SIZE: usize = 16;
+
     pub(crate) fn report_progress(
         &self,
         update_id: Uuid,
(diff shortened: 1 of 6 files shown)


wicketd/src/installinator_progress.rs
```diff
@@ -195,15 +204,44 @@ impl RunningUpdate {
...
+                // all of its messages.
+                slog::warn!(
+                    log,
+                    "progress receiver is closed -- marking aborted";
+                    "last_seen" => ?report.last_seen,
+                );
+                (
+                    // Don't set the state to Closed here even if this is a
+                    // terminal update, because we want to keep returning
+                    // ReceiverClosed.
```
jgallagher, line 240: I don't entirely follow this comment, but I'm having a hard time finding where `Self::Closed` vs `Self::ReportsReceived(_)` is used or reported - can you point me in the right direction?

wicketd/src/installinator_progress.rs
```diff
@@ -72,30 +72,42 @@ impl IprArtifactServer {
...
+                    let (new_state, ret) = RunningUpdate::send_and_next_state(
+                        &self.log, sender, report,
+                    );
+                    *update = new_state;
+                    ret
                 }
                 RunningUpdate::Closed => {
                     // The sender has been closed; ignore the report.
                     *update = RunningUpdate::Closed;
+                    EventReportStatus::Processed
```
sunshowers, line 105: This is where the `ReportsReceived` and `Closed` statuses are read.

wicketd/src/installinator_progress.rs
sunshowers: I've expanded the comment -- hopefully this is clearer.

jgallagher: It is, thank you!

jgallagher (approved):

PR #3985, opened 2023-08-30
[nexus] Silo IP pools

### LINKED ISSUE #3926: Silo-scoped IP pools
Details to be added. Discussion in https://github.com/oxidecomputer/customer-support/issues/18.

Currently we have [fleet-scoped IP pools with CRUD ops](https://github.com/oxidecomputer/omicron/blob/1ed6f4b1288f7fcb76f1fc078159e3f735b7cba2/nexus/src/app/ip_pool.rs). We want to be able to do the same thing for each silo so instances created in a given silo are only drawing from the IP pools for that silo.

david-crespo:
Closes #3926 

A lot going on here. I will update this description as I finish things out. 

## Important background

* #3981 added `silo_id` and `project_id` columns to the IP pools table in the hope that that would be sufficient for the rest of the work schema-wise, but it turns out there was more needed.
* Before this change, the concept of a **default** IP pool was implemented through the `name` on the IP pool, i.e., the IP pool named `default` is the one used by default for instance IP allocation if no pool name is specified as part of the instance create POST
* Before this change the concept of an **internal** IP pool was implemented through a boolean `internal` column on the `ip_pool` table

## IP pool selection logic

There are two situations where we pick an IP pool to allocate IPs from. For an instance's **source NAT**, we always use the default pool. Before this change, with only fleet pools, this simply meant picking the one named `default` (analogous now to the one with `is_default == true`). With the possibility of silo pools, we now pick the most specific default available. That means that if there is a silo-scoped pool marked default _matching the current silo_, we use that. If not, we pick the fleet-level pool marked default, which should always exist (see possible todos at the bottom — we might want to take steps to guarantee this).

For instance ephemeral IPs, the instance create POST body takes an optional pool name. If none is specified, we follow the same defaulting logic as above — the most-specific pool marked `is_default`. We are leaving pool names globally unique (as opposed to per-scope) which IMO makes the following lookup logic easy to understand and implement: given a pool name, look up the pool by name. (That part we were already going.) The difference now with scopes is that we need to make sure that the scope of the selected pool (assuming it exists) **does not conflict** with the current scope, i.e., the current silo. In this situation, we do not care about what's marked default, and we are not trying to get an exact match on scope. We just need to disallow an instance from using an IP pool reserved for a different silo. We can revisit this, but as implemented here you can, for example, specify a non-default pool scoped to fleet or silo (if one exists) even if there exists a default pool scoped to your silo.

## DB migrations on `ip_pool` table

There are three migrations here based on guidance from @smklein based on [CRDB docs about limitations to online schema changes](https://www.cockroachlabs.com/docs/stable/online-schema-changes#limitations) and some conversations he had with them. It's possible they could be made into two. I don't think it can be done in one.

* Add `is_default` column and a unique index ensuring there is only one default IP pool per "scope" (unique `silo_id`, including null as a distinct value)
* Populate correct data in new columns
  * Populate `is_default = true` for any IP pools named `default` (there should be exactly one, but nothing depends on that)
  * `silo_id = INTERNAL_SILO_ID` for any IP pools marked `internal` (there should be exactly one, but nothing depends on that)
* Drop the `internal` column

## Code changes

- [x] Add [`similar-asserts`](https://crates.io/crates/similar-asserts) so we can get a usable diff when the schema migration tests fail. Without this you could get a 20k+ line diff with 4 relevant lines.
- [x] Use `silo_id == INTERNAL_SILO_ID` everywhere we were previously looking at the `internal` flag (thanks @luqmana for the [suggestion](https://github.com/oxidecomputer/omicron/pull/3981#discussion_r1309136684))
- [x] Add `silo_id` and `default` to `IpPoolCreate` (create POST params) and plumb them down the create chain
  * Leave off `project_id` for now, we can add that later
- [x] Fix index that is failing to prevent multiple `default` pools for a given scope (see comment https://github.com/oxidecomputer/omicron/pull/3985#discussion_r1309624130)
- [x] Source NAT IP allocation uses new defaulting logic to get the most specific available default IP Pool
- [x] Instance ephemeral IP allocation uses that default logic if no pool name specified in the create POST, otherwise look up pool by specified name (but can only get pools matching its scope, i.e., its project, silo, or the whole fleet)

### Limitations that we might want to turn into to-dos

* You can't update `default` on a pool, i.e., you can't make a pool default. You have to delete it and make a new one. This one isn't that hard — I would think of it like image promotion, where it's not a regular update pool, it's a special endpoint for making a pool default that can unset the current default if it's a different pool.
* Project-scoped IP pools endpoints are fake — they just return all IP pools. They were made in anticipation of being able to make them real. I'm thinking we should remove them or make them work, but I don't think we have time to make them work.
* Ensure somehow that there is always a fleet-level default pool to fall back to

### DIFF
--- a/nexus/db-model/src/schema.rs
+++ b/nexus/db-model/src/schema.rs
@@ -455,6 +455,7 @@ table! {
         rcgen -> Int8,
         silo_id -> Nullable<Uuid>,
         project_id -> Nullable<Uuid>,
+        default -> Bool,
     }
 }
 
--- a/nexus/db-model/src/schema.rs
+++ b/nexus/db-model/src/schema.rs
@@ -1131,7 +1131,7 @@ table! {
 ///
 /// This should be updated whenever the schema is changed. For more details,
 /// refer to: schema/crdb/README.adoc
-pub const SCHEMA_VERSION: SemverVersion = SemverVersion::new(3, 0, 2);
+pub const SCHEMA_VERSION: SemverVersion = SemverVersion::new(3, 0, 3);
 
 allow_tables_to_appear_in_same_query!(
     system_update,
--- a/nexus/db-model/src/schema.rs
+++ b/nexus/db-model/src/schema.rs
@@ -455,7 +455,7 @@ table! {
         rcgen -> Int8,
         silo_id -> Nullable<Uuid>,
         project_id -> Nullable<Uuid>,
-        default -> Bool,
+        is_default -> Bool,
     }
 }
 
--- a/nexus/db-model/src/schema.rs
+++ b/nexus/db-model/src/schema.rs
@@ -452,7 +452,6 @@ table! {
         time_created -> Timestamptz,
         time_modified -> Timestamptz,
         time_deleted -> Nullable<Timestamptz>,
-        internal -> Bool,
         rcgen -> Int8,
         silo_id -> Nullable<Uuid>,
         project_id -> Nullable<Uuid>,
@@ -1131,7 +1130,7 @@ table! {
 ///
 /// This should be updated whenever the schema is changed. For more details,
 /// refer to: schema/crdb/README.adoc
-pub const SCHEMA_VERSION: SemverVersion = SemverVersion::new(3, 0, 0);
+pub const SCHEMA_VERSION: SemverVersion = SemverVersion::new(3, 0, 2);
 
 allow_tables_to_appear_in_same_query!(
     system_update,
--- a/nexus/db-model/src/schema.rs
+++ b/nexus/db-model/src/schema.rs
@@ -454,7 +454,6 @@ table! {
         time_deleted -> Nullable<Timestamptz>,
         rcgen -> Int8,
         silo_id -> Nullable<Uuid>,
-        project_id -> Nullable<Uuid>,
         is_default -> Bool,
     }
 }
--- a/nexus/db-queries/src/db/datastore/external_ip.rs
+++ b/nexus/db-queries/src/db/datastore/external_ip.rs
@@ -50,21 +50,15 @@ impl DataStore {
         opctx: &OpContext,
         ip_id: Uuid,
         instance_id: Uuid,
-        project_id: Uuid,
         pool_name: Option<Name>,
     ) -> CreateResult<ExternalIp> {
         // If we have a pool name, look up the pool by name and return it
         // as long as its scopes don't conflict with the current scope.
         // Otherwise, not found.
         let pool = match pool_name {
-            Some(name) => {
-                self.ip_pools_fetch_for(&opctx, &name, Some(project_id)).await?
-            }
+            Some(name) => self.ip_pools_fetch_for(&opctx, &name).await?,
             // If no name given, use the default logic
-            None => {
-                self.ip_pools_fetch_default_for(&opctx, Some(project_id))
-                    .await?
-            }
+            None => self.ip_pools_fetch_default_for(&opctx).await?,
         };
 
         let pool_id = pool.identity.id;
--- a/nexus/db-queries/src/db/datastore/external_ip.rs
+++ b/nexus/db-queries/src/db/datastore/external_ip.rs
@@ -52,30 +52,15 @@ impl DataStore {
         instance_id: Uuid,
         pool_name: Option<Name>,
     ) -> CreateResult<ExternalIp> {
-        let current_silo = opctx.authn.silo_required()?;
-
         // If we have a pool name, look up the pool by name and return it
         // as long as its scopes don't conflict with the current scope.
         // Otherwise, not found.
         let pool = match pool_name {
-            Some(name) => {
-                self.ip_pools_fetch_for(
-                    &opctx,
-                    &name,
-                    current_silo.id(),
-                    None, // TODO: include current project ID
-                )
-                .await?
-            }
+            // TODO: include current project ID
+            Some(name) => self.ip_pools_fetch_for(&opctx, &name, None).await?,
             // If no name given, use the default logic
-            None => {
-                self.ip_pools_fetch_default_for(
-                    &opctx,
-                    Some(current_silo.id()),
-                    None, // TODO: include current project ID
-                )
-                .await?
-            }
+            // TODO: include current project ID
+            None => self.ip_pools_fetch_default_for(&opctx, None).await?,
         };
 
         let pool_id = pool.identity.id;
--- a/nexus/db-queries/src/db/datastore/external_ip.rs
+++ b/nexus/db-queries/src/db/datastore/external_ip.rs
@@ -64,7 +64,7 @@ impl DataStore {
                     &opctx,
                     authz::Action::CreateChild, // TODO: wtf
                     &name,
-                    Some(current_silo.id()),
+                    current_silo.id(),
                     None, // TODO: include current project ID
                 )
                 .await?
--- a/nexus/db-queries/src/db/datastore/external_ip.rs
+++ b/nexus/db-queries/src/db/datastore/external_ip.rs
@@ -58,6 +58,13 @@ impl DataStore {
         let name = pool_name.unwrap_or_else(|| {
             Name(ExternalName::from_str("default").unwrap())
         });
+
+        // TODO: this is the whole thing -- update this call to EITHER look up
+        // the pool by name (only allowing access to pools matching the scope of
+        // the current project, i.e., you can pick a pool by name only if it's
+        // scoped to your project, silo, or the whole fleet) or use the default
+        // logic in ip_pool_fetch_default_for
+
         let (.., pool) = self
             .ip_pools_fetch_for(opctx, authz::Action::CreateChild, &name)
             .await?;
--- a/nexus/db-queries/src/db/datastore/external_ip.rs
+++ b/nexus/db-queries/src/db/datastore/external_ip.rs
@@ -56,9 +56,9 @@ impl DataStore {
         // as long as its scopes don't conflict with the current scope.
         // Otherwise, not found.
         let pool = match pool_name {
-            Some(name) => self.ip_pools_fetch_for(&opctx, &name).await?,
+            Some(name) => self.ip_pools_fetch(&opctx, &name).await?,
             // If no name given, use the default logic
-            None => self.ip_pools_fetch_default_for(&opctx).await?,
+            None => self.ip_pools_fetch_default(&opctx).await?,
         };
 
         let pool_id = pool.identity.id;
--- a/nexus/db-queries/src/db/datastore/external_ip.rs
+++ b/nexus/db-queries/src/db/datastore/external_ip.rs
@@ -5,7 +5,6 @@
 //! [`DataStore`] methods on [`ExternalIp`]s.
 
 use super::DataStore;
-use crate::authz;
 use crate::context::OpContext;
 use crate::db;
 use crate::db::error::public_error_from_diesel_pool;
@@ -62,7 +61,6 @@ impl DataStore {
             Some(name) => {
                 self.ip_pools_fetch_for(
                     &opctx,
-                    authz::Action::CreateChild, // TODO: wtf
                     &name,
                     current_silo.id(),
                     None, // TODO: include current project ID
@@ -73,7 +71,6 @@ impl DataStore {
             None => {
                 self.ip_pools_fetch_default_for(
                     &opctx,
-                    authz::Action::CreateChild, // TODO: wtf
                     Some(current_silo.id()),
                     None, // TODO: include current project ID
                 )
--- a/nexus/db-queries/src/db/datastore/external_ip.rs
+++ b/nexus/db-queries/src/db/datastore/external_ip.rs
@@ -25,9 +25,7 @@ use nexus_types::identity::Resource;
 use omicron_common::api::external::CreateResult;
 use omicron_common::api::external::Error;
 use omicron_common::api::external::LookupResult;
-use omicron_common::api::external::Name as ExternalName;
 use std::net::IpAddr;
-use std::str::FromStr;
 use uuid::Uuid;
 
 impl DataStore {
@@ -55,21 +53,35 @@ impl DataStore {
         instance_id: Uuid,
         pool_name: Option<Name>,
     ) -> CreateResult<ExternalIp> {
-        let name = pool_name.unwrap_or_else(|| {
-            Name(ExternalName::from_str("default").unwrap())
-        });
+        let current_silo = opctx.authn.silo_required()?;
 
-        // TODO: this is the whole thing -- update this call to EITHER look up
-        // the pool by name (only allowing access to pools matching the scope of
-        // the current project, i.e., you can pick a pool by name only if it's
-        // scoped to your project, silo, or the whole fleet) or use the default
-        // logic in ip_pool_fetch_default_for
+        // If we have a pool name, look up the pool by name and return it
+        // as long as its scopes don't conflict with the current scope.
+        // Otherwise, not found.
+        let pool = match pool_name {
+            Some(name) => {
+                self.ip_pools_fetch_for(
+                    &opctx,
+                    authz::Action::CreateChild, // TODO: wtf
+                    &name,
+                    Some(current_silo.id()),
+                    None, // TODO: include current project ID
+                )
+                .await?
+            }
+            // If no name given, use the default logic
+            None => {
+                self.ip_pools_fetch_default_for(
+                    &opctx,
+                    authz::Action::CreateChild, // TODO: wtf
+                    Some(current_silo.id()),
+                    None, // TODO: include current project ID
+                )
+                .await?
+            }
+        };
 
-        let (.., pool) = self
-            .ip_pools_fetch_for(opctx, authz::Action::CreateChild, &name)
-            .await?;
         let pool_id = pool.identity.id;
-
         let data =
             IncompleteExternalIp::for_ephemeral(ip_id, instance_id, pool_id);
         self.allocate_external_ip(opctx, data).await
--- a/nexus/db-queries/src/db/datastore/external_ip.rs
+++ b/nexus/db-queries/src/db/datastore/external_ip.rs
@@ -50,17 +50,21 @@ impl DataStore {
         opctx: &OpContext,
         ip_id: Uuid,
         instance_id: Uuid,
+        project_id: Uuid,
         pool_name: Option<Name>,
     ) -> CreateResult<ExternalIp> {
         // If we have a pool name, look up the pool by name and return it
         // as long as its scopes don't conflict with the current scope.
         // Otherwise, not found.
         let pool = match pool_name {
-            // TODO: include current project ID
-            Some(name) => self.ip_pools_fetch_for(&opctx, &name, None).await?,
+            Some(name) => {
+                self.ip_pools_fetch_for(&opctx, &name, Some(project_id)).await?
+            }
             // If no name given, use the default logic
-            // TODO: include current project ID
-            None => self.ip_pools_fetch_default_for(&opctx, None).await?,
+            None => {
+                self.ip_pools_fetch_default_for(&opctx, Some(project_id))
+                    .await?
+            }
         };
 
         let pool_id = pool.identity.id;
--- a/nexus/db-queries/src/db/datastore/ip_pool.rs
+++ b/nexus/db-queries/src/db/datastore/ip_pool.rs
@@ -73,12 +73,15 @@ impl DataStore {
     }
 
     /// Looks up the default IP pool for a given scope, i.e., a given
-    /// combination of silo and project ID (or none). If there is no default at
-    /// a given scope, fall back up a level. There should always be a default at
-    /// fleet level, though this query can theoretically fail.
+    /// combination of silo and project ID. If there is no default at a given
+    /// scope, fall back up a level. There should always be a default at fleet
+    /// level, though this query can theoretically fail.
     pub async fn ip_pools_fetch_default_for(
         &self,
         opctx: &OpContext,
+        // Optional primarily because there are test contexts where we don't care
+        // about the project. If project ID is None, we will only get back pools
+        // that themselves have no project.
         project_id: Option<Uuid>,
     ) -> LookupResult<IpPool> {
         use db::schema::ip_pool::dsl;
@@ -123,7 +126,6 @@ impl DataStore {
         &self,
         opctx: &OpContext,
         name: &Name,
-        // TODO: project_id should always be defined, this is temporary
         project_id: Option<Uuid>,
     ) -> LookupResult<IpPool> {
         let authz_silo_id = opctx.authn.silo_required()?.id();
@@ -498,8 +500,11 @@ mod test {
         let mut db = test_setup_database(&logctx.log).await;
         let (opctx, datastore) = datastore_test(&logctx, &db).await;
 
+        // it doesn't matter whether the project exists
+        let project_id = Uuid::new_v4();
+
         // we start out with the default fleet-level pool already created,
-        // so when we ask for the fleet default (no silo or project) we get it back
+        // so when we ask for a default silo, we get it back
         let fleet_default_pool =
             datastore.ip_pools_fetch_default_for(&opctx, None).await.unwrap();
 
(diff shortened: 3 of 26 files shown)


nexus/src/app/sagas/instance_create.rs
```diff
@@ -826,8 +826,14 @@ async fn sic_allocate_instance_snat_ip(
...
-    let (.., pool) = datastore
-        .ip_pools_fetch_default_for(&opctx, authz::Action::CreateChild)
+    let pool = datastore
+        .ip_pools_fetch_default_for(
+            &opctx,
+            authz::Action::CreateChild,
+            // TODO: get these values right
+            None,
+            None,
+        )
```
david-crespo, line 836: TODO: pass the relevant silo ID and project ID here

nexus/db-queries/src/db/datastore/external_ip.rs
```diff
@@ -58,6 +58,13 @@ impl DataStore {
         let name = pool_name.unwrap_or_else(|| {
             Name(ExternalName::from_str("default").unwrap())
         });
+
+        // TODO: this is the whole thing -- update this call to EITHER look up
+        // the pool by name (only allowing access to pools matching the scope of
+        // the current project, i.e., you can pick a pool by name only if it's
+        // scoped to your project, silo, or the whole fleet) or use the default
+        // logic in ip_pool_fetch_default_for
+
```
david-crespo, line 67: TODO: kind of the whole point

nexus/tests/integration_tests/instances.rs
```diff
@@ -3502,6 +3503,8 @@ async fn test_instance_create_in_silo(cptestctx: &ControlPlaneTestContext) {
         .authn_as(AuthnMode::SiloUser(user_id))
         .execute()
         .await
+        // fails here with 403 that comes from the default IP pool query,
+        // which checks ListChildren on IP_POOL_LIST
```
david-crespo, line 3507: Need to figure out why trying to list the IP pools is causing a 403

schema/crdb/dbinit.sql
```diff
@@ -1480,32 +1480,33 @@ CREATE TABLE IF NOT EXISTS omicron.public.ip_pool (
...
 );
 
+/*
+ * Ensure there can only be one default pool for the fleet or a given silo or project
+ */
+CREATE UNIQUE INDEX IF NOT EXISTS one_default_pool_per_scope ON omicron.public.ip_pool (
+    silo_id, project_id
+) WHERE
+    "default" = true AND time_deleted IS NULL;
+
```
david-crespo, line 1509: Need to make sure this index treats `null` as a relevant distinct value, so that, e.g., at the `(null, null)` level (the fleet-wide level) there can only be one default pool.

david-crespo: Welp, it doesn't work if either or both is null because it doesn't treat null the way it treats normal values. Looks like this might work as a workaround:

```sql
CREATE OR REPLACE FUNCTION f_non_null_id(id uuid)
RETURNS uuid AS $$
BEGIN
  RETURN COALESCE(id, '00000000-0000-0000-0000-000000000000'::uuid);
END;

CREATE UNIQUE INDEX IF NOT EXISTS one_default_pool_per_scope ON omicron.public.ip_pool (
    f_non_null_id(silo_id), f_non_null_id(project_id)
) WHERE
    "default" = true AND time_deleted IS NULL;
```

_except_ CRDB doesn't support [`CREATE FUNCTION`](https://www.cockroachlabs.com/docs/v22.2/create-function) until 22.2.

Screenshot 2023-08-29 at 11 24 49 PM

david-crespo:
Only 3 tests are failing, which is surprisingly good.

```
FAIL [   6.275s] nexus-db-queries db::queries::external_ip::tests::test_ensure_pool_exhaustion_does_not_use_other_pool
FAIL [   4.879s] nexus-db-queries db::queries::external_ip::tests::test_next_external_ip_is_restricted_to_pools
FAIL [  28.225s] omicron-nexus::test_all integration_tests::instances::test_instance_create_in_silo
```

schema/crdb/dbinit.sql
```diff
@@ -1480,32 +1480,36 @@ CREATE TABLE IF NOT EXISTS omicron.public.ip_pool (
...
+/*
+ * Ensure there can only be one default pool for the fleet or a given silo or
+ * project. Coalesce is needed because otherwise different nulls are considered
+ * to be distinct from each other.
+ */
+CREATE UNIQUE INDEX IF NOT EXISTS one_default_pool_per_scope ON omicron.public.ip_pool (
+    COALESCE(silo_id, '00000000-0000-0000-0000-000000000000'::uuid), 
+    COALESCE(project_id, '00000000-0000-0000-0000-000000000000'::uuid) 
+) WHERE
+    "default" = true AND time_deleted IS NULL;
```
david-crespo, line 1511: Welp, how do we feel about this coalesce @smklein

david-crespo:
Still a bit of work to do, but all the existing tests pass. The biggest thing it needs is tests for instance IP allocation by name exercising the scope conflict checks.

nexus/db-queries/src/db/datastore/ip_pool.rs
```diff
@@ -65,44 +63,96 @@ impl DataStore {
...
+        use db::schema::ip_pool::dsl;
+
+        // TODO: Need auth check here. Only fleet viewers can list children on
+        // IP_POOL_LIST, so if we check that, nobody can make instances. This
+        // used to check CreateChild on an individual IP pool, but now we're not
+        // looking up by name so the check is more complicated
+        //
+        // opctx
+        //     .authorize(authz::Action::ListChildren, &authz::IP_POOL_LIST)
+        //     .await?;
```
david-crespo, line 94: Interesting TODO. Not sure how to do this auth check given the way perms are set up for IpPool and IpPoolList:

https://github.com/oxidecomputer/omicron/blob/eee942f601bbebf00352eb5c5f060e1e80a437e6/nexus/db-queries/src/authz/omicron.polar#L368-L390

david-crespo:
Worth noting that while nearly all the plumbing is there for project-scoped IP pools, my plan was to make it for now so you can't create one (there's no `project_id` param on `IpPoolCreate`) and you don't get a `project_id` back on the `IpPool` view. So it's like it doesn't exist. I would want to be more confident everything was correct.

nexus/db-queries/src/db/datastore/ip_pool.rs
```diff
@@ -65,44 +63,100 @@ impl DataStore {
...
-    ) -> LookupResult<(authz::IpPool, IpPool)> {
-        self.ip_pools_fetch_for(
-            opctx,
-            action,
-            &Name(ExternalName::from_str("default").unwrap()),
-        )
-        .await
+        // Optional primarily because there are test contexts where we don't care
+        // about the project. If project ID is None, we will only get back pools
+        // that themselves have no project.
```
ahl, line 84:
```suggestion
        // that themselves have no associated project.
```

... does that mean we'd just get fleet-wide pools?

nexus/db-queries/src/db/datastore/ip_pool.rs
ahl: Can we punt on this for now? Alternatively: if we don't do the most restrictive auth check right now are we concerned by the data exposed? Seems like a potentially reasonable punt.

ahl (commented):
looking good

nexus/db-queries/src/db/datastore/external_ip.rs
```diff
@@ -53,16 +50,24 @@ impl DataStore {
         opctx: &OpContext,
         ip_id: Uuid,
         instance_id: Uuid,
+        project_id: Uuid,
         pool_name: Option<Name>,
```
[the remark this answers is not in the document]
david-crespo: Yeah, I'm pretty sure it could be. I think the main reason it was name only was that the name `"default"` was significant, and now it isn't.

nexus/db-queries/src/db/datastore/ip_pool.rs
```diff
@@ -65,44 +63,100 @@ impl DataStore {
...
         .get_results_async(self.pool_authorized(opctx).await?)
         .await
         .map_err(|e| public_error_from_diesel_pool(e, ErrorHandler::Server))
     }
 
-    /// Looks up the default IP pool by name.
+    /// Looks up the default IP pool for a given scope, i.e., a given
+    /// combination of silo and project ID. If there is no default at a given
+    /// scope, fall back up a level. There should always be a default at fleet
+    /// level, though this query can theoretically fail.
```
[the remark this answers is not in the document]
david-crespo: Basically you would have to delete the fleet-wide (no silo and no project) IP pool we create by default, which you can't do if it has any IP ranges, and you can't delete those if they have any allocated IPs. So it's pretty hard to do that. Once we add the ability to make something a new default, you could conceivably create a new fleet-level pool, make it the default, and delete it while it's still empty. All very contrived situations, obviously. It might be cool to try to ensure that it's always there, but on the other hand I don't think it's expected that we can prevent every form of bad configuration here. In fact, we definitely can't because this all ties into customer network config that is outside our control.

nexus/db-queries/src/db/datastore/ip_pool.rs
david-crespo: Or silo-scoped pools. The signature here is a little misloading because it only takes project (because we are pulling the current silo off `opctx`) but the logic does look at both silo and project ID when picking defaults.

schema/crdb/dbinit.sql
smklein: I think postgres itself has better support here:

https://www.postgresql.org/docs/current/indexes-unique.html

They have a `CREATE UNIQUE INDEX ... ON TABLE ... NULLS DISTINCT` option, which can specify that NULLS are treated equally.

However, for Cockroachdb:

- https://www.cockroachlabs.com/docs/stable/create-index
- https://www.cockroachlabs.com/docs/stable/null-handling#nulls-and-unique-constraints

I don't see similar support.

So: Translating "NULL" to "The zero UUID" seems a little hacky, but probably fine.

schema/crdb/dbinit.sql
```diff
@@ -1480,32 +1480,36 @@ CREATE TABLE IF NOT EXISTS omicron.public.ip_pool (
...
-     * will (naturally) be the ID of the project's silo. Both must be null if
-     * internal is true, i.e., internal IP pools must be fleet-level pools.
+     * Fields representating association with a silo or project. silo_id must
+     * be non-null if project_id is non-null. silo_id is also used to mark an IP
+     * pool as "internal" by associating it with the oxide-internal silo.
      */
     silo_id UUID,
     project_id UUID,
 
     -- if silo_id is null, then project_id must be null
```
luqmana, line 1496: Does it ever make sense to have a non-NULL `project_id` and `silo_id` where the `silo_id` doesn't match the project's silo? If not, I think adjusting this so that at most one of them is non-null might be preferable. A `project_id` already uniquely identifies a project and therefore the silo it belongs to.

schema/crdb/3.0.0/up.sql
```diff
@@ -20,17 +12,36 @@ SELECT CAST(
 );
 
 ALTER TABLE omicron.public.ip_pool
+    ADD COLUMN IF NOT EXISTS is_default BOOLEAN NOT NULL DEFAULT FALSE,
```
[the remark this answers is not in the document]
david-crespo: Yes. I think I mentioned it at control plane huddle but I noted it was a bad idea. There isn't really a reason to do it this way, I can change it to use 3 new migrations.

nexus/src/app/ip_pool.rs
```diff
@@ -25,6 +27,10 @@ use omicron_common::api::external::ResourceType;
 use omicron_common::api::external::UpdateResult;
 use ref_cast::RefCast;
 
+fn is_internal(pool: &IpPool) -> bool {
```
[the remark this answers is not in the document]
david-crespo: I wanted to, but `models` can't import from `db-queries` because `db-queries` depends on `models`, and `db-queries` is where `INTERNAL_SILO_ID` is defined.

nexus/db-model/src/schema.rs
```diff
@@ -1131,7 +1131,7 @@ table! {
 ///
 /// This should be updated whenever the schema is changed. For more details,
 /// refer to: schema/crdb/README.adoc
-pub const SCHEMA_VERSION: SemverVersion = SemverVersion::new(3, 0, 0);
+pub const SCHEMA_VERSION: SemverVersion = SemverVersion::new(3, 0, 2);
```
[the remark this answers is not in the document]
david-crespo: I wanted to imply that these are related migrations, though I admit that is not at at all what semver patch versions are supposed to mean.

nexus/src/app/ip_pool.rs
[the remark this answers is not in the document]
luqmana: ah, fair enough

nexus/types/src/external_api/params.rs
```diff
@@ -731,6 +731,13 @@ impl std::fmt::Debug for CertificateCreate {
 pub struct IpPoolCreate {
     #[serde(flatten)]
     pub identity: IdentityMetadataCreateParams,
+
+    /// If an IP pool is associated with a silo, instance IP allocations in that
+    /// silo can draw from that pool.
+    pub silo: Option<NameOrId>,
+
+    #[serde(default)]
+    pub is_default: bool,
```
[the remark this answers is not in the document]
david-crespo: Added in [06b1af7](https://github.com/oxidecomputer/omicron/pull/3985/commits/06b1af7d2823f17865247b651078cf89952cbf42)

schema/crdb/3.0.0/up.sql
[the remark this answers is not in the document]
david-crespo: Changed to 3.0.1-3.0.3 instead of modifying 3.0.0 in [e3c56c4](https://github.com/oxidecomputer/omicron/pull/3985/commits/e3c56c45faabe21076da60488f283aeef2b9a5c5)

schema/crdb/dbinit.sql
david-crespo: True. My initial reaction was that if `silo_id` wasn't there, the query for the default pool would have to be more complicated.

```rs
dsl::ip_pool
    .filter(dsl::silo_id.eq(authz_silo_id).or(dsl::silo_id.is_null()))
    .filter(
        dsl::project_id.eq(project_id).or(dsl::project_id.is_null()),
    )
    .filter(dsl::is_default.eq(true))
    // this will sort by most specific first, i.e.,
    //
    //   (silo, project)
    //   (silo, null)
    //   (null, null)
    //
    // then by only taking the first result, we get the most specific one
    .order((
        dsl::project_id.asc().nulls_last(),
        dsl::silo_id.asc().nulls_last(),
    ))
    .select(IpPool::as_select())
    .first_async::<IpPool>(self.pool_authorized(opctx).await?)
    .await
    .map_err(|e| public_error_from_diesel_pool(e, ErrorHandler::Server))
```

But if at most one of the two was non-null, we could do something like

```sql
where default = true and (
  silo_id = authz_silo_id 
  or project_id = input_project_id
  or (silo_id is null and project_id is null)
)
```

and that should get us the same at-most-3 results. If anything the proposed modification is easier to understand rather than harder.

nexus/db-queries/src/db/datastore/ip_pool.rs
```diff
@@ -65,44 +63,100 @@ impl DataStore {
...
 
-        Ok((authz_pool, pool))
+        match (pool.project_id, project_id) {
+            (Some(pool_project_id), Some(current_project_id)) => {
+                if pool_project_id != current_project_id {
+                    return Err(authz_pool.not_found());
+                }
+            }
+            _ => {}
+        }
```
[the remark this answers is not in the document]
david-crespo: I don't understand what you mean, could you elaborate?

nexus/db-queries/src/db/datastore/ip_pool.rs
david-crespo: #3995

nexus/db-queries/src/db/datastore/ip_pool.rs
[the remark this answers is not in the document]
luqmana: Just imagining something tying into #3995 which I imagine would allow augmenting the `LookupPath::new(opctx, &self).ip_pool_name(&name)` call with `.project_id(project_id)`

nexus/db-queries/src/db/datastore/ip_pool.rs
```diff
@@ -65,44 +63,78 @@ impl DataStore {
...
         .map_err(|e| public_error_from_diesel_pool(e, ErrorHandler::Server))
     }
 
-    /// Looks up the default IP pool by name.
-    pub async fn ip_pools_fetch_default_for(
+    /// Look up the default IP pool for the current silo. If there is no default
+    /// at silo scope, fall back to the next level up, namely the fleet default.
+    /// There should always be a default pool at the fleet level, though this
+    /// query can theoretically fail if someone is able to delete that pool or
+    /// make another one the default and delete that.
```
luqmana, line 79: Not a blocker but maybe that's something we should guard accident.

david-crespo:
I have a test to add here and it's passing locally. Because it's only test code, I'm going to merge this so we can get things moving on dogfood and make that a second PR.

nexus/db-queries/src/db/datastore/ip_pool.rs
```diff
@@ -65,44 +63,78 @@ impl DataStore {
...
-    ) -> LookupResult<(authz::IpPool, IpPool)> {
+    ) -> LookupResult<IpPool> {
+        let authz_silo_id = opctx.authn.silo_required()?.id();
+
         let (.., authz_pool, pool) = LookupPath::new(opctx, &self)
             .ip_pool_name(&name)
-            .fetch_for(action)
+            // any authenticated user can CreateChild on an IP pool. this is
+            // meant to represent allocating an IP
+            .fetch_for(authz::Action::CreateChild)
```
luqmana, line 126: Why get rid of the `action` arg and hardcode `CreateChild` here? Makes it a little mislead as the method is named `ip_pools_fetch` but as you noted, the `CreateChild` action is for allocating an IP from the pool. So a caller that just wants to retrieve the pool but not allocate from it has a slightly stronger than necessary check.

nexus/db-queries/src/db/datastore/ip_pool.rs
```diff
@@ -65,44 +63,78 @@ impl DataStore {
...
             .await?;
-        if pool.internal {
-            return Err(authz_pool.not_found());
+
+        // You can't look up a pool by name if it conflicts with your current
+        // scope, i.e., if it has a silo it is different from your current silo
```
luqmana, line 130:
```suggestion
        // scope, i.e., if it has a silo that is different from your current silo
```

nexus/db-queries/src/db/datastore/ip_pool.rs
```diff
@@ -65,44 +63,78 @@ impl DataStore {
...
+        if let Some(pool_silo_id) = pool.silo_id {
+            if pool_silo_id != authz_silo_id {
+                return Err(authz_pool.not_found());
+            }
```
luqmana, line 134:
```suggestion
        if Some(authz_silo_id) != pool.silo_id {
            return Err(authz_pool.not_found());
```

nexus/db-queries/src/db/datastore/ip_pool.rs
```diff
@@ -425,3 +459,107 @@ impl DataStore {
...
+        let identity = IdentityMetadataCreateParams {
+            name: "non-default-for-silo".parse().unwrap(),
+            description: "".to_string(),
+        };
+        let _ = datastore
+            .ip_pool_create(
+                &opctx,
+                IpPool::new(&identity, Some(silo_id), /*default= */ false),
+            )
+            .await;
```
luqmana, line 518: check this call didn't fail?

nexus/db-queries/src/db/datastore/ip_pool.rs
```diff
@@ -425,3 +459,107 @@ impl DataStore {
...
+        assert_eq!(ip_pool.id(), fleet_default_pool.id());
+
+        // now create a default pool for the silo
+        let identity = IdentityMetadataCreateParams {
+            name: "default-for-silo".parse().unwrap(),
+            description: "".to_string(),
+        };
+        let _ = datastore
+            .ip_pool_create(&opctx, IpPool::new(&identity, Some(silo_id), true))
+            .await;
```
luqmana, line 535: check for failure here too

nexus/tests/integration_tests/ip_pools.rs
```diff
@@ -269,6 +273,77 @@ async fn test_ip_pool_basic_crud(cptestctx: &ControlPlaneTestContext) {
...
+            description: String::from(""),
+        },
+        silo: Some(NameOrId::Name(cptestctx.silo_name.clone())),
+        is_default: false,
+    };
+    let created_pool = create_pool(client, &params).await;
+    assert_eq!(created_pool.identity.name, "p0");
+    assert!(created_pool.silo_id.is_some());
+
+    let silo_id = created_pool.silo_id.unwrap();
```
luqmana, line 293: nit: Can just drop that assert since the unwrap would panic. `.expect` if you want a nice message.

schema/crdb/3.0.1/up.sql
```diff
@@ -0,0 +1,42 @@
...
+        'true',
+        'Invalid starting version for schema change'
+    ) AS BOOL
+);
+
+ALTER TABLE omicron.public.ip_pool
+    DROP COLUMN IF EXISTS project_id,
+    ADD COLUMN IF NOT EXISTS is_default BOOLEAN NOT NULL DEFAULT FALSE,
+    DROP CONSTRAINT IF EXISTS project_implies_silo,
+    DROP CONSTRAINT IF EXISTS internal_pools_have_null_silo_and_project;
```
luqmana, line 18: Is the ordering important here? i.e., is it ok for `DROP COLUMN` to come before the `DROP CONSTRAINT`'s?

schema/crdb/3.0.1/up.sql
```diff
@@ -0,0 +1,42 @@
...
+
+COMMIT;
+
+-- needs to be in its own transaction because of this thrilling bug
+-- https://github.com/cockroachdb/cockroach/issues/83593
```
luqmana, line 23: would a nested transaction also work? that would get rid of the duplicate schema version checks

luqmana (approved):
a couple nits/questions but looks good overall. Thanks!

schema/crdb/3.0.1/up.sql
david-crespo: The test that applies the migrations and makes sure they produce the right end result didn’t complain, so I think it’s ok.

nexus/db-queries/src/db/datastore/ip_pool.rs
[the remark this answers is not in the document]
david-crespo: https://github.com/oxidecomputer/omicron/issues/4003

nexus/db-queries/src/db/datastore/ip_pool.rs
david-crespo: This doesn't work when `pool.silo_id == None`. We only want to do the check if the fetched pool has a silo ID. If `pool.silo_id.is_none()`, then it can't conflict with the current silo because it's a fleet-wide pool.

nexus/db-queries/src/db/datastore/ip_pool.rs
david-crespo: https://github.com/oxidecomputer/omicron/issues/4003

nexus/db-queries/src/db/datastore/ip_pool.rs
david-crespo: I think the abstraction is just cut the wrong way here. I'm not sure whether I made it worse — it was already weird before. Both this auth check and the check against the silo scope are fundamentally about the use of this function for retrieving an IP pool in order to allocate IPs from it. Despite the name, this isn't a general-purpose function at all. It only has one non-test call site:

https://github.com/oxidecomputer/omicron/blob/01e730acf3efeb6ff2181d558cf373ded383a365/nexus/db-queries/src/db/datastore/external_ip.rs#L47-L68

I'm thinking the thing to do is inline this logic in that call site. `ip_pool_fetch_default` function involves a DB query so it makes sense as a datastore function, plus the `default` imples a "relative to what?" that I think makes the silo ID check make more sense. Plus the query needs to know about the silo ID in order to work.

PR #3987, opened 2023-08-30
Do not merge: Add repo-pvt2 with gimlet rev-e SP image

jgallagher:
Opening this so it can run through CI in case there are any issues using https://github.com/oxidecomputer/omicron/pull/3931 to update the PVT2 rack later today.


jgallagher:
Closing in favor of #3931.

PR #3989, opened 2023-08-30
[bootstore] Error when network config is current

andrewjstone:
This is a change to optimize for the RSS use case, which is currently the only place we update the bootstore network config.


leftwo (approved):
This was a very complicated change, and required some serious effort to understand
the impact and significance.  After several weeks of intense study, family neglect, malnutrition, and lack of vitamin D, I have reached the conclusion that this change
can be approved.  I'll be taking a six month vacation after this work to properly recover and
reenter society.

PR #3990, opened 2023-08-30
[tufaceous-lib] Reject duplicate artifact names

### LINKED ISSUE #3988: TUF repos have incorrect RoT images
https://github.com/oxidecomputer/omicron/commit/6a68451f0b7bd30ee4c8f2ad3e3a984971e3200e changed the TUF manifest we generate in CI to pull the RoT board name via hubtools, but it reports the same name for all three of gimlet/sidecar/psc (`oxide-rot-1`). This results in a TUF repo that has an entry for all three boards, but the actual artifact is the _same_ (i.e., we end up with a sidecar RoT image labels as the gimlet and sidecar and PSC RoT image). Skimming through tufaceou's source it looks like it's trying to guard against this:

https://github.com/oxidecomputer/omicron/blob/58e8c67e3449fe656a6d20721bbdab786034031e/tufaceous-lib/src/repository.rs#L278-L287

but that check isn't firing for some reason. We should:

a. Fix the check so it refuses to build repos with manifests like we're producing now
b. Fix our manifest so the three RoT images have different names

jgallagher:
Fixes #3988, albeit in a quick-and-dirty way. I have a couple concerns; will leave them as PR comments below.

### DIFF
--- a/tufaceous-lib/src/repository.rs
+++ b/tufaceous-lib/src/repository.rs
@@ -11,7 +11,7 @@ use omicron_common::{
     api::external::SemverVersion,
     update::{Artifact, ArtifactsDocument},
 };
-use std::num::NonZeroU64;
+use std::{collections::BTreeSet, num::NonZeroU64};
 use tough::{
     editor::{signed::SignedRole, RepositoryEditor},
     schema::{Root, Target},
@@ -202,20 +202,22 @@ pub struct OmicronRepoEditor {
     editor: RepositoryEditor,
     repo_path: Utf8PathBuf,
     artifacts: ArtifactsDocument,
-    existing_targets: Vec<TargetName>,
+    existing_target_filenames: BTreeSet<String>,
 }
 
 impl OmicronRepoEditor {
     fn new(repo: OmicronRepo) -> Result<Self> {
         let artifacts = repo.read_artifacts()?;
 
-        let existing_targets = repo
+        let existing_target_filenames = repo
             .repo
             .targets()
             .signed
             .targets_iter()
-            .map(|(name, _)| name.to_owned())
-            .collect::<Vec<_>>();
+            .flat_map(|(name, _)| {
+                [name.raw().to_string(), name.resolved().to_string()]
+            })
+            .collect::<BTreeSet<_>>();
 
         let editor = RepositoryEditor::from_repo(
             repo.repo_path
@@ -228,7 +230,7 @@ impl OmicronRepoEditor {
             editor,
             repo_path: repo.repo_path,
             artifacts,
-            existing_targets,
+            existing_target_filenames,
         })
     }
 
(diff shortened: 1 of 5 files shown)


tufaceous-lib/src/repository.rs
```diff
@@ -276,15 +278,13 @@ impl OmicronRepoEditor {
             artifact.target = filename.clone();
         } else {
```
jgallagher, line 279: In the `if` part of this branch, we're accepting (and replacing) an artifact if we're called with the same name+version+kind. I'm nervous about that - it certainly seems wrong if we're building from a manifest (since we shouldn't have duplicates like that, and the existence of a such a dup probably means we've made a mistake), but maybe it's necessary in general?

tufaceous-lib/src/repository.rs
```diff
@@ -276,15 +278,13 @@ impl OmicronRepoEditor {
...
             // if we don't, make sure we're not overriding another target.
-            if self.existing_targets.iter().any(|target_name| {
-                target_name.raw() == filename
-                    && target_name.resolved() == filename
-            }) {
+            if !self.existing_target_filenames.insert(filename.clone()) {
```
jgallagher, line 281: It seems strange to me that we have to do this check ourselves to avoid clobbering another file, even if the metadata (its `kind`) and contents (including the hash) are different. Are we (ab)using tuf in some way that makes this necessary?

iliana:
I think this is fine if we need to fix something quickly, but I don't think guarding against duplicate artifact names that belong to different kinds is correct.

- TUF requires that target names are different, even if they have different hashes[^hash]. This is because the `targets.json` is more or less [a JSON object keyed by the target name](https://github.com/theupdateframework/specification/blob/c51875f445d8a57efca9dadfbd5dbdece06d87e6/tuf-spec.md#file-formats-targetsjson-and-delegated-target-roles--file-formats-targets).
- An artifact is unique across the three values (name, version, kind). It is important that the name field be allowed to have the same value within a given repository because we will eventually have two different versions of a gimlet-e SP image in the same repository.

I am also somewhat confused about the underlying issue. I am going to attempt to understand it a bit better.

[^hash]: The presence of a hash in the filename is to allow for repositories to remain consistent while they're in the middle of being modified, as long as you change `timestamp.json` last.

tufaceous-lib/src/repository.rs
iliana: I think this should probably error, yeah. You're right that the existence of a duplicate name-version-kind triple means we've more likely made a mistake.

jgallagher: Changed to an error in [7898a35](https://github.com/oxidecomputer/omicron/pull/3990/commits/7898a35605936d620130316ff520f18f200fef67)

iliana (approved):

PR #3997, opened 2023-08-30
Prevent snapshot delete saga conflicts

jmpesp:
== Use resource locks to block other snapshot delete sagas ==

If a user issues many requests to delete a particular snapshot, this will trigger many snapshot-delete sagas, all of which will conflict with each other and cause problems. Use a resource lock (which is defined as a row in CRDB that has a lock id corresponding to a resource) to prevent this.

Without a resource lock like this, the snapshot delete saga would need a node that satisfied the following properties:

1. it would make some modification to the snapshot record in the database
2. that modification would block other sagas from executing if a user issued many delete requests for the same snapshot
3. if that saga node reran, it would execute ok

I couldn't find a way to satisfy both 2 and 3. In fact, at this point I think there is no way to satisfy both 2 and 3! A saga node that reexecutes after a crash is similar to that same node in a concurrently executing saga, and a mechanism that would block another saga from running a node would block the rerun of that saga node after a crash, no matter if the crashing saga "won" originally or not.

What was required was some unique information tied to the resource that was part of the saga, and thus a resource lock was born: the lock id is unique per saga, and using a row in the database as a lock would guarantee that only one saga can "win" and operate on a particular resource.

== Only clean up resources associated with your volume ==

When a read-only downstairs is no longer referenced by a volume, Nexus must issue DELETE calls to the appropriate Crucible agent to clean them up. As part of the snapshot delete saga, Nexus will decrement reference counts and return a list of regions and region_snapshots to delete (along with their datasets), but currently this list of region_snapshots will contain *any* resources to clean up, not just the argument volume's:

    volume_1 references to RO1, RO2, RO3
    volume_2 references to RO4, RO5, RO6

    resources_to_clean(volume_1) => [RO1, RO2, RO3, RO4, RO5, RO6]
    resources_to_clean(volume_2) => [RO1, RO2, RO3, RO4, RO5, RO6]

Handing these duplicate lists of resources to clean up is mostly ok: the Crucible agent is idempotent for equivalent requests, as it has to be to be used by Nexus in a saga. The one case where it isn't is when issuing a DELETE call for a read-only downstairs booted from a snapshot that doesn't exist: it will return 404. This makes sagas sad, and will cause them to unwind.

Fix this by only returning the resources associated with the argument volume:

    resources_to_clean(volume_1) => [RO1, RO2, RO3]
    resources_to_clean(volume_2) => [RO4, RO5, RO6]


nexus/db-queries/src/db/datastore/resource_lock.rs
```diff
@@ -0,0 +1,184 @@
...
+                        )
+                    })?;
+
+                if let Some(existing_lock) = existing_lock {
+                    if existing_lock.lock_id != lock_id {
+                        return Err(TransactionError::CustomError(
+                            Error::invalid_request(&format!(
+                            "resource {} already locked by another lock ID {}",
+                            resource_id, lock_id,
+                        )),
```
smklein, line 49: With this current scheme, we have no idea "who" took the lock, correct? So in a situation where we see the resource is locked, but no progress is being made, we really have no choice but to wait forever, correct?

For example, suppose we do the following:

- Start saga node, in which we intend to grab the lock
- Try to grab the lock, succeed
- Before we wrote that we finished the saga node, we crash
- Reboot
- Re-read saga log, try to replay node, see that someone else grabbed the lock
- Fail, never succeed again, because the node that "Grabbed" the lock has crashed and no longer exists

smklein: Oh, nevermind, I see that this is what the `lock_id` accomplishes - so if we see the same lock ID, we know it was our operation which triggered the action?

nexus/db-model/src/resource_lock.rs
```diff
@@ -0,0 +1,14 @@
...
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use crate::schema::resource_lock;
+use uuid::Uuid;
+
+/// Database representation of a resource lock
+#[derive(Queryable, Insertable, Debug, Clone, Selectable)]
+#[diesel(table_name = resource_lock)]
+pub struct ResourceLock {
+    pub resource_id: Uuid,
```
smklein, line 12: Might be worth documenting:

`resource_id` is the object being locked

I actually think `lock_id` is acting more like `owner_id` -- it should uniquely identify "who is asking for the lock", rather than the lock itself.

For example, in a theoretical situation where we grab multiple locks (... and, on that note, we should consider that deadlocks are possible now), it would be "normal" to use the same `lock_id` value across multiple `ResourceLock` objects.

PR #4007, opened 2023-09-01
Small IP pools fixes

### LINKED ISSUE #4005: IP Pools: include `is_default` value in the list/get ip pool response
The `is_default` attribute is currently missing from the get response body. It'll be useful for the admin to know whether a default pool is made for a silo and which one it is, instead of relying on the name or description to provide that information.

david-crespo:
Followup to #3985, closes #4005.

- Add `is_default` to IP pool response
- Inline `ip_pools_fetch` into the one callsite and delete it (per discussion https://github.com/oxidecomputer/omicron/pull/3985#discussion_r1312320845)
- Other small test tweaks suggested by @luqmana

### DIFF
--- a/nexus/db-queries/src/db/datastore/ip_pool.rs
+++ b/nexus/db-queries/src/db/datastore/ip_pool.rs
@@ -6,7 +6,6 @@
 
 use super::DataStore;
 use crate::authz;
-use crate::authz::ApiResource;
 use crate::context::OpContext;
 use crate::db;
 use crate::db::collection_insert::AsyncInsertError;
@@ -16,7 +15,6 @@ use crate::db::error::public_error_from_diesel_pool;
 use crate::db::error::ErrorHandler;
 use crate::db::fixed_data::silo::INTERNAL_SILO_ID;
 use crate::db::identity::Resource;
-use crate::db::lookup::LookupPath;
 use crate::db::model::IpPool;
 use crate::db::model::IpPoolRange;
 use crate::db::model::IpPoolUpdate;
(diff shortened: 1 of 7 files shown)


nexus/db-queries/src/db/datastore/ip_pool.rs
```diff
@@ -111,32 +109,6 @@ impl DataStore {
...
-        // scope, i.e., if it has a silo it is different from your current silo
-        if let Some(pool_silo_id) = pool.silo_id {
-            if pool_silo_id != authz_silo_id {
-                return Err(authz_pool.not_found());
-            }
-        }
-
-        Ok(pool)
-    }
-
```
david-crespo, line 139: This did not make sense as a function here. All the logic was related to the calling context.

nexus/db-queries/src/db/datastore/ip_pool.rs
```diff
@@ -541,13 +514,6 @@ mod test {
             .expect("Failed to get silo's default IP pool");
         assert_eq!(ip_pool.name().as_str(), "default-for-silo");
 
-        // if we ask for the fleet default by name, we can still get that one
-        let ip_pool = datastore
-            .ip_pools_fetch(&opctx, &Name("default".parse().unwrap()))
-            .await
-            .expect("Failed to get fleet default IP pool");
-        assert_eq!(ip_pool.id(), fleet_default_pool.id());
-
```
david-crespo, line 550: can't do this anymore because the method is gone, but this integration test covers it.

https://github.com/oxidecomputer/omicron/blob/01e730acf3efeb6ff2181d558cf373ded383a365/nexus/tests/integration_tests/instances.rs#L3442-L3451

PR #4027, opened 2023-09-02
Reduce unused dependencies

smklein:
- Downgraded multiple dependencies in `[dependencies]` that are only necessary in `[dev-dependencies]`
- Removed multiple dependencies that are unused

Used https://github.com/est31/cargo-udeps to perform this analysis


sunshowers (approved):
Thanks!

PR #4041, opened 2023-09-06
[wicketd] Miscellaneous work towards a smaller heap

jgallagher:
None of these commits directly impact wicketd's heap, but they're a stepping stone to the next PR (will open momentarily). It may be worth reviewing the commits individually; each should be a small and uncontroversial change, I think.


sunshowers (approved):
These look really good! Thanks.

jgallagher:
(This is not landing in `main` yet, so I'm going to go ahead and merge before CI is done so the real CI build can get going)

PR #4042, opened 2023-09-06
[wicketd] Avoid copying artifacts into memory during ingest

### LINKED ISSUE #3943: `wicketd` has a big heap after driving installation
In yesterday's dogfood update the captain updated the cubby 16 Scrimlet, then used it to update the rest of the rack, including the cubby 14 Scrimlet. In the aftermath, `pmap(1)` says that 16's `wicketd` is holding on to about 6.75 GiB of heap-related swap charges:

```
root@oxz_switch:~# pgrep wicketd | xargs pmap -S
1628:   /opt/oxide/wicketd/bin/wicketd run /var/svc/manifest/site/wicketd/conf
         Address     Kbytes       Swap Mode Mapped File
0000000000400000      33984          - r-x--  wicketd
000000000253F000       1192       1192 rw---  wicketd
0000000002669000    4150204    4150204 rw---    [ heap ]
00000000FFB58000    2916372    2916372 rw---    [ heap ]
```

Its sibling on 14 is much smaller:

```
2068:   /opt/oxide/wicketd/bin/wicketd run /var/svc/manifest/site/wicketd/conf
         Address     Kbytes       Swap Mode Mapped File
0000000000400000      33984          - r-x--  wicketd
000000000253F000       1192       1192 rw---  wicketd
0000000002669000      59716      59716 rw---    [ heap ]
```

RFD 413 proposes that each rack service be given a swap limit and reservation that ensures it won't OOM from a lack of available swap charges (as long as the service's processes are well-behaved). Under that scheme, we'd currently need to assume that `wicketd` needs a 7-8 GiB swap reservation. That's about 10% of the available swap on a sled in RFD 413's proposed budgets. This is workable for now, but I wanted to file an issue to track two items:

1. It'd be great to give the maximum heap size figure a haircut if at all possible.
2. More importantly, we need to understand where all this usage is coming from so that we can convince ourselves this maximum won't grow if our update repositories grow.

It'd also be good hygiene to make sure all this memory is freed after being used (assuming it isn't; it's possible that it's freed from the application's perspective but that the underlying heap implementation isn't actually releasing the underlying swap; I don't know enough about the illumos heap to know how likely this is). I don't think that affects the reservations we have to create (we have to reserve enough to accommodate peak swap usage even if that usage is transient).

jgallagher:
This continues on the work in #3953 to shrink `wicketd`'s heap usage by moving the TUF artifacts we ingest out of memory and onto disk (in a temporary directory).

Briefly recapping where we are on `main` after uploading a TUF repo:

```
root@oxz_switch:/opt/oxide# pgrep wicketd | xargs pmap -S | grep heap
0000000002979000    3390440    3390440 rw---    [ heap ]
00000000D1873000    2392080    2392080 rw---    [ heap ]
```

and where we are as of #3953 after uploading a TUF repo:

```
root@oxz_switch:/opt/oxide# pgrep wicketd | xargs pmap -S | grep heap
000000000297C000    1684440    1684440 rw---    [ heap ]
```

and now where we are as of this branch after uploading a TUF repo:

```
root@oxz_switch:/opt/oxide# pgrep wicketd | xargs pmap -S | grep heap
000000000296D000      48960      48960 rw---    [ heap ]
```

The big impact here is that when we're unpacking the host and trampoline artifacts, we no longer unpack them to memory then write them to disk; we stream them straight to the temporary directory as we unpack.

Fixes #3943.

### DIFF
--- a/wicketd/src/artifacts.rs
+++ b/wicketd/src/artifacts.rs
@@ -4,19 +4,17 @@
 
 use std::{
     borrow::Borrow,
-    collections::{btree_map, hash_map::Entry, BTreeMap, HashMap},
+    collections::{BTreeMap, HashMap},
     io, mem,
     sync::{Arc, Mutex},
 };
 
 use async_trait::async_trait;
-use bytes::Bytes;
 use camino::Utf8PathBuf;
 use camino_tempfile::Utf8TempDir;
 use debug_ignore::DebugIgnore;
 use display_error_chain::DisplayErrorChain;
 use dropshot::HttpError;
-use hubtools::RawHubrisArchive;
 use hyper::Body;
 use installinator_artifactd::{ArtifactGetter, EventReportStatus};
 use omicron_common::api::{
@@ -28,22 +26,30 @@ use omicron_common::update::{
 use slog::{error, info, Logger};
 use thiserror::Error;
 use tough::TargetName;
-use tufaceous_lib::{
-    ArchiveExtractor, HostPhaseImages, OmicronRepo, RotArchives,
-};
+use tufaceous_lib::{ArchiveExtractor, OmicronRepo};
 use uuid::Uuid;
 
-use crate::installinator_progress::IprArtifactServer;
+use crate::{
+    http_entrypoints::InstallableArtifacts,
+    installinator_progress::IprArtifactServer,
+};
 
 mod extracted_artifacts;
+mod update_plan;
 
 pub(crate) use self::extracted_artifacts::ExtractedArtifactDataHandle;
-use self::extracted_artifacts::ExtractedArtifacts;
+use self::update_plan::UpdatePlanBuilder;
 
 // A collection of artifacts along with an update plan using those artifacts.
 #[derive(Debug)]
 struct ArtifactsWithPlan {
-    by_id: DebugIgnore<HashMap<ArtifactId, ExtractedArtifactDataHandle>>,
+    // Map of top-level artifact IDs (present in the TUF repo) to the actual
+    // artifacts we're serving (e.g., a top-level RoT artifact will map to two
+    // artifact hashes: one for each of the A and B images).
+    //
+    // The sum of the lengths of the values of this map will match the number of
+    // entries in `by_hash`.
+    by_id: BTreeMap<ArtifactId, Vec<ArtifactHashId>>,
     by_hash: DebugIgnore<HashMap<ArtifactHashId, ExtractedArtifactDataHandle>>,
     plan: UpdatePlan,
 }
@@ -137,11 +143,18 @@ impl WicketdArtifactStore {
 
     pub(crate) fn system_version_and_artifact_ids(
         &self,
-    ) -> Option<(SemverVersion, Vec<ArtifactId>)> {
+    ) -> Option<(SemverVersion, Vec<InstallableArtifacts>)> {
         let artifacts = self.artifacts_with_plan.lock().unwrap();
         let artifacts = artifacts.as_ref()?;
         let system_version = artifacts.plan.system_version.clone();
-        let artifact_ids = artifacts.by_id.keys().cloned().collect();
+        let artifact_ids = artifacts
+            .by_id
+            .iter()
+            .map(|(k, v)| InstallableArtifacts {
+                artifact_id: k.clone(),
+                installable: v.clone(),
+            })
+            .collect();
         Some((system_version, artifact_ids))
     }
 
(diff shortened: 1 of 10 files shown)


wicketd/src/artifacts/update_plan.rs
```diff
@@ -0,0 +1,1016 @@
...
+    psc_rot_a: Option<ArtifactIdData>,
+    psc_rot_b: Option<ArtifactIdData>,
+    sidecar_sp: BTreeMap<Board, ArtifactIdData>,
+    sidecar_rot_a: Option<ArtifactIdData>,
+    sidecar_rot_b: Option<ArtifactIdData>,
+    host_phase_1: Option<ArtifactIdData>,
+    trampoline_phase_1: Option<ArtifactIdData>,
+    trampoline_phase_2: Option<ArtifactIdData>,
+    host_phase_2_hash: Option<ArtifactHash>,
+    control_plane_hash: Option<ArtifactHash>,
```
sunshowers, line 57: Why store `trampoline_phase_2` (i.e. not just a hash) and `host_phase_2_hash` here? Would be worth a comment.

wicketd/src/artifacts/update_plan.rs
```diff
@@ -0,0 +1,1016 @@
...
+            return self.add_unknown_artifact(
+                artifact_id,
+                artifact_hash,
+                reader,
+                by_id,
+                by_hash,
+            );
+        };
+
+        match artifact_kind {
```
sunshowers, line 112: It took me a while to understand the structure of this. I think one thing I got tripped up on is the "`(not trampoline!)`" and "`(not host!)`" comments. Could you add a comment explaining the general structure of this function?

wicketd/src/artifacts/update_plan.rs
```diff
@@ -0,0 +1,1016 @@
...
+                by_hash,
+            ),
+            KnownArtifactKind::ControlPlane => self.add_control_plane_artifact(
+                artifact_id,
+                artifact_kind,
+                artifact_hash,
+                reader,
+                by_id,
+                by_hash,
+            ),
```
sunshowers, line 153: Why pass in the `KnownArtifactKind` for these if we're always going to ensure it's one particular kind anyway?

wicketd/src/artifacts/update_plan.rs
```diff
@@ -0,0 +1,1016 @@
...
+        .entry(tuf_repo_artifact_id)
+        .or_default()
+        .push(by_hash_slot.key().clone());
+    by_hash_slot.insert(data);
+
+    Ok(())
+}
+
+#[cfg(test)]
+mod tests {
```
sunshowers, line 678: Are these tests mostly copied over?

sunshowers (approved):
Looks good. Thanks for waiting -- I spent a long time reading the PR.

wicketd/src/artifacts/update_plan.rs
jgallagher: Yes, with minor tweaks to match the structural changes.

wicketd/src/artifacts.rs
```diff
@@ -28,22 +26,30 @@ use omicron_common::update::{
...
 struct ArtifactsWithPlan {
-    by_id: DebugIgnore<HashMap<ArtifactId, ExtractedArtifactDataHandle>>,
+    // Map of top-level artifact IDs (present in the TUF repo) to the actual
+    // artifacts we're serving (e.g., a top-level RoT artifact will map to two
+    // artifact hashes: one for each of the A and B images).
+    //
+    // The sum of the lengths of the values of this map will match the number of
+    // entries in `by_hash`.
+    by_id: BTreeMap<ArtifactId, Vec<ArtifactHashId>>,
     by_hash: DebugIgnore<HashMap<ArtifactHashId, ExtractedArtifactDataHandle>>,
```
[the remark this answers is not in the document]
jgallagher: Certainly: added in c27dc85d

wicketd/src/artifacts/update_plan.rs
jgallagher: Added a comment in c27dc85d: we have to send the trampoline phase 2 to MGS, but installinator fetches the host phase 2 from us over the bootstrap network (we just have to send it the hash via MGS).

wicketd/src/artifacts/update_plan.rs
jgallagher: I added a bit of a comment and also removed the assertions (which were confusing at best as you pointed out in a later comment)

wicketd/src/artifacts/update_plan.rs
jgallagher: We do use it in the SP / RoT methods (since we need to known which of the gimlet/sidecar/psc kinds it is), and I just copied that over. Removed the arg from the host/trampoline/control plane methods in 88393ed9

jgallagher:
(This is not landing in `main` yet, so I'm going to go ahead and merge before CI is done so the real CI build can get going)

PR #4047, opened 2023-09-07
Only clean resources associated with your volume

jmpesp:
When a read-only downstairs is no longer referenced by a volume, Nexus must issue DELETE calls to the appropriate Crucible agent to clean them up. As part of the snapshot delete saga, Nexus will decrement reference counts and return a list of regions and region_snapshots to delete (along with their datasets), but currently this list of region_snapshots will contain *any* resources to clean up, not just the argument volume's:

    volume_1 references to RO1, RO2, RO3
    volume_2 references to RO4, RO5, RO6

    resources_to_clean(volume_1) => [RO1, RO2, RO3, RO4, RO5, RO6]
    resources_to_clean(volume_2) => [RO1, RO2, RO3, RO4, RO5, RO6]

Handing these duplicate lists of resources to clean up is mostly ok: the Crucible agent is idempotent for equivalent requests, as it has to be to be used by Nexus in a saga. The one case where it isn't is when issuing a DELETE call for a read-only downstairs booted from a snapshot that doesn't exist: it will return 404. This makes sagas sad, and will cause them to unwind.

Fix this by only returning the resources associated with the argument volume:

    resources_to_clean(volume_1) => [RO1, RO2, RO3]
    resources_to_clean(volume_2) => [RO4, RO5, RO6]

### DIFF
--- a/nexus/db-queries/src/db/datastore/volume.rs
+++ b/nexus/db-queries/src/db/datastore/volume.rs
@@ -640,15 +640,13 @@ impl DataStore {
                 // Decrease the number of uses for each referenced region snapshot.
                 use db::schema::region_snapshot::dsl;
 
-                for read_only_target in &crucible_targets.read_only_targets {
-                    diesel::update(dsl::region_snapshot)
-                        .filter(dsl::snapshot_addr.eq(read_only_target.clone()))
-                        .set(
-                            dsl::volume_references
-                                .eq(dsl::volume_references - 1),
-                        )
-                        .execute(conn)?;
-                }
+                diesel::update(dsl::region_snapshot)
+                    .filter(
+                        dsl::snapshot_addr
+                            .eq_any(&crucible_targets.read_only_targets),
+                    )
+                    .set(dsl::volume_references.eq(dsl::volume_references - 1))
+                    .execute(conn)?;
 
                 // Return what results can be cleaned up
                 let result = CrucibleResources::V1(CrucibleResourcesV1 {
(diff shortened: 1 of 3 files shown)


nexus/db-queries/src/db/datastore/volume.rs
```diff
@@ -640,15 +640,13 @@ impl DataStore {
...
-                        )
-                        .execute(conn)?;
-                }
+                diesel::update(dsl::region_snapshot)
+                    .filter(
+                        dsl::snapshot_addr
+                            .eq_any(&crucible_targets.read_only_targets),
+                    )
+                    .set(dsl::volume_references.eq(dsl::volume_references - 1))
+                    .execute(conn)?;
```
smklein, line 649: Nice

nexus/db-queries/src/db/datastore/volume.rs
```diff
@@ -688,15 +686,32 @@ impl DataStore {
...
+                            // volume containing these read-only targets.
+                            //
+                            // It's important to not return *every* region
+                            // snapshot with zero references: multiple volume
+                            // delete sub-sagas will then be issues duplicate
+                            // DELETE calls to Crucible agents, and a request to
+                            // delete a read-only downstairs running for a
+                            // snapshot that doesn't exist will return a 404,
+                            // causing the saga to error and unwind.
+                            .filter(
```
smklein, line 710: So down on line 735ish is where we actually store the "stuff to be cleaned up" -- in the case that multiple snapshots are sharing resources for a volume, it's basically the responsibility of the "last one deleted" to set this field with the set of regions to be deleted, right?

smklein (approved):

nexus/tests/integration_tests/snapshots.rs
```diff
@@ -1100,3 +1100,228 @@ async fn test_region_snapshot_create_idempotent(
...
+                // This must match what is in the snapshot_delete saga!
+                db::model::SnapshotState::Ready,
+                db::model::SnapshotState::Faulted,
+                db::model::SnapshotState::Destroyed,
+            ],
+        )
+        .await
+        .unwrap();
+
+    // Then, concurrently call
```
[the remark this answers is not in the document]
jmpesp: This comment could have been clearer: I meant this as "simulate the execution of these three `snapshot_delete` sagas overlapping, where  `decrease_crucible_resource_count_and_soft_delete_volume` happens one after each other.

PR #4049, opened 2023-09-07
Simulated Crucible agent should be stricter

jmpesp:
There's a few more info messages here that will aid debugging, plus a few ways the simulated Crucible agent is now stricter:

- it should be impossible to create a snapshot of a region in a bad state.

- it should be an error to try to delete a snapshot for a non-existent region (the real sled agent would return a 404)

- it should be an error to attempt creating a running snapshot for a snapshot that does not exist

The actual Crucible agent reuses freed ports, so the simulated one should too.


sled-agent/src/sim/storage.rs
```diff
@@ -258,28 +299,34 @@ impl CrucibleDataInner {
...
         let id = Uuid::from_str(&id.0).unwrap();
+        let port_number = self.get_free_port();
 
         let map =
             self.running_snapshots.entry(id).or_insert_with(|| HashMap::new());
 
         // If a running snapshot exists already, return it - this endpoint must
         // be idempotent.
         if let Some(running_snapshot) = map.get(&name.to_string()) {
+            self.used_ports.remove(&port_number);
```
leftwo, line 315: Nit: Could you put the `let port_number = ..` below this if and then you don't have to remove it here.

sled-agent/src/sim/storage.rs
```diff
@@ -317,15 +365,291 @@ impl CrucibleDataInner {
...
+        empty
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use omicron_test_utils::dev::test_setup_log;
+
+    /// Validate that the Crucible agent reuses ports
```
leftwo, line 397: Is this testing the sim agent, or the real agent?

sled-agent/src/sim/storage.rs
```diff
@@ -154,23 +167,47 @@ impl CrucibleDataInner {
...
+        snapshot_id: Uuid,
+    ) -> Result<Snapshot> {
         info!(self.log, "Creating region {} snapshot {}", id, snapshot_id);
-        self.snapshots
+
+        if let Some(region) = self.get(RegionId(id.to_string())) {
+            match region.state {
+                State::Failed | State::Destroyed | State::Tombstoned => {
+                    bail!(
+                        "cannot create snapshot of region in state {:?}",
```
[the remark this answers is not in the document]
jmpesp: Good idea, done in 2a3cbd545

sled-agent/src/sim/storage.rs
jmpesp: Good catch! Done in 4b705aff6

sled-agent/src/sim/storage.rs
leftwo: If it's all testing the sim, we should make it clear in the comments that is what we are testing, so I don't come back later and think this is testing a real agent :)

leftwo (approved):
Just the request to change the comments about sim testing should say it is testing sim.

sled-agent/src/sim/storage.rs
jmpesp: Done! see bbca204fe

PR #4052, opened 2023-09-07
Reduce public visibility of Nexus

smklein:
This PR originated as an attempt to reduce the amount of symbols exposed from Nexus.
Although that wasn't particularly fruitful, it's still worthwhile reducing this visibility.

This change exposed a couple areas where we have test-only code, and overall makes it slightly more difficult for consumers of the `omicron-nexus` library to rely on internal interfaces.


nexus/src/lib.rs
```diff
@@ -14,19 +14,18 @@
...
 pub mod external_api; // Public for testing
-pub mod internal_api; // Public for testing
+mod internal_api;
 mod populate;
 mod saga_interface;
-pub mod updates; // public for testing
+mod updates; // public for testing
 
 pub use app::test_interfaces::TestInterfaces;
 pub use app::Nexus;
```
smklein, line 26: I normally wouldn't rely on `pub(crate)` as much as I do in this PR, but this exposure (as well as `pub mod app`), which is used by the integration tests, makes it hard to determine "where things would be exposed out of the crate" at a module level.

jordanhendricks:
> Although that wasn't particularly fruitful, it's still worthwhile reducing this visibility.

By fruitful, do you mean as an attempt to affect incremental build time?

smklein:
> > Although that wasn't particularly fruitful, it's still worthwhile reducing this visibility.
> 
> By fruitful, do you mean as an attempt to affect incremental build time?

Yup. I was hoping that by dropping the number of publicly exported symbols, we might be able to reduce link time in particular. `nm -g` is still showing a fairly large number of extern symbols from this library, unfortunately.

nexus/src/saga_interface.rs
```diff
@@ -29,31 +29,32 @@ impl fmt::Debug for SagaContext {
...
         &self.nexus
     }
 
-    pub fn datastore(&self) -> &db::DataStore {
+    pub(crate) fn datastore(&self) -> &db::DataStore {
         self.nexus.datastore()
     }
 
-    pub async fn sled_client(
+    #[allow(dead_code)]
```
jgallagher, line 56: Similar as above - if this is truly dead, could we remove it?

nexus/src/lib.rs
```diff
@@ -41,10 +40,6 @@ use std::net::{SocketAddr, SocketAddrV6};
 use std::sync::Arc;
 use uuid::Uuid;
 
-// These modules used to be within nexus, but have been moved to
-// nexus-db-queries. Keeping these around temporarily for migration reasons.
-pub use nexus_db_queries::{authn, authz, db};
```
jgallagher, line 46: Ooh, thanks for addressing this. rust-analyzer helps a lot, but I always find the "the thing I'm importing is actually a re-export of something else" a little confusing when I run into them.

jgallagher (approved):
:+1: I'm in favor of minimizing visibility by default. This looked like it was pretty tedious - thanks for churning through it.

davepacheco:
Are there any consumers of `omicron_nexus` as a library crate, aside from its integration test crate?

smklein:
> Are there any consumers of `omicron_nexus` as a library crate, aside from its integration test crate?

Both of the binaries (`nexus` and `schema-updater`)

smklein:
> Are there any consumers of `omicron_nexus` as a library crate, aside from its integration test crate?

Ah, I forgot -- there's also a dependency on it from `dev-tools`, as part of the `cmd_run_all`, which starts an in-process Nexus server.

Cargo.toml
```diff
@@ -528,6 +528,10 @@ opt-level = 3
 #crucible-client-types = { path = "../crucible/crucible-client-types" }
 #crucible-pantry-client = { path = "../crucible/pantry-client" }
 #crucible-smf = { path = "../crucible/smf" }
+#[patch.crates-io]
```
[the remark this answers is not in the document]
smklein: Easy to fix, updated!

nexus/src/app/background/common.rs
```diff
@@ -281,6 +281,7 @@ impl Driver {
     }
 
     /// Returns the runtime status of the background task
+    #[allow(dead_code)]
     pub fn status(&self, task: &TaskHandle) -> TaskStatus {
```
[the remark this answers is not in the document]
smklein: It's being used by unit tests right now, though it looks "kinda reasonable to use from other contexts".

I marked it as `allow(dead_code)` rather than `cfg(test)` for that reason -- I think it would be reasonable for other parts of Nexus to act on this status, even though only tests are doing so right now.

nexus/src/saga_interface.rs
smklein: Ah, this one truly is dead. Removed.

nexus/src/lib.rs
smklein: Totally! I think the intermediate transition also made sense, but as these are becoming more established as separate crates, it makes more and more sense to keep detangling 'em.

PR #4054, opened 2023-09-08
Reduce direct dependencies on Diesel from nexus

smklein:
Instead, let nexus-db-queries handle them

### DIFF
--- a/nexus/Cargo.toml
+++ b/nexus/Cargo.toml
@@ -9,7 +9,6 @@ omicron-rpaths.workspace = true
 
 [dependencies]
 anyhow.workspace = true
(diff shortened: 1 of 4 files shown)


nexus/Cargo.toml
```diff
@@ -97,7 +92,9 @@ rustls = { workspace = true }
 
 [dev-dependencies]
 assert_matches.workspace = true
+async-bb8-diesel.workspace = true
 criterion.workspace = true
+diesel.workspace = true
```
smklein, line 97: Still being used from some saga tests to directly poke at the DB

jgallagher (approved):

PR #4055, opened 2023-09-08
Split regions ensure undo into separate saga node

jmpesp:
In both the disk_create and snapshot_create sagas, split the saga nodes to deal with `ensure_all_datasets_and_regions` partially failing. See RFD 419 for more information on this pattern. `delete_crucible_region` also has to change to see if the region never existed, because the real Crucible agent would return a 404 for a DELETE of a non-existent region.

Catch the case where we're trying to delete a Crucible region that never existed. As well, the real Crucible agent would not let a Failed region be deleted, so change that. This caused a test to fail, that should never have passed! We cannot delete Failed regions and this test was checking that we could. Correct this by causing a disk whose regions cannot be deleted to be Faulted.

Finally, ensure that there are no stuck sagas during our tests.


nexus/src/app/sagas/common_storage.rs
```diff
@@ -148,6 +148,35 @@ pub(super) async fn delete_crucible_region(
...
+
+            _ => {
+                return Err(Error::internal_error(
+                    "unexpected failure during `region_get`",
+                ));
+            }
+        }
+    }
+
+    // Past here, the region exists: ensure it is deleted.
```
gjcolombo, line 178: This probably predates this change, but out of curiosity, what happens if someone tries to delete a disk while a snapshot saga is in progress (or is unwinding)? Will `svd_delete_freed_crucible_regions` find the regions the snapshot saga's trying to unwind?

nexus/tests/integration_tests/volume_management.rs
```diff
@@ -1813,6 +1813,132 @@ async fn test_volume_checkout_updates_sparse_mid_multiple_gen(
...
+
+    // Set the third agent to fail creating the region
+    let zpool = &disk_test.zpools[2];
+    let dataset = &zpool.datasets[0];
+    disk_test
+        .sled_agent
+        .get_crucible_dataset(zpool.id, dataset.id)
+        .await
+        .set_region_creation_limit(1)
+        .await;
```
gjcolombo, line 1892: I assume this is set to 1 so that the disk creation will succeed while the snapshot request fails. The intent might be a little clearer if you just set it to 0 after creating the disk. (At that point you might not even want this to be a numerical limit: instead it could just be a `bool`, or, if you really want to gild the lily, you could tell the simulated sled agent exactly what kind of error to return instead of having it return 500 Internal Server Error unconditionally.)

nexus/src/app/sagas/common_storage.rs
jmpesp: You're not wrong, this is a problem. Stay tuned for a follow up PR!

nexus/src/app/sagas/disk_create.rs
```diff
@@ -863,6 +914,41 @@ pub(crate) mod test {
         assert_eq!(disk.project_id, project_id);
     }
 
+    async fn no_stuck_sagas(log: &Logger, datastore: &DataStore) -> bool {
```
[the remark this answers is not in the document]
jmpesp: nice, using that in 4a266bcbc

nexus/src/app/sagas/common_storage.rs
```diff
@@ -148,6 +148,35 @@ pub(super) async fn delete_crucible_region(
...
+                } else {
+                    return Err(Error::internal_error(&rv.message));
+                }
+            }
+
+            _ => {
+                return Err(Error::internal_error(
+                    "unexpected failure during `region_get`",
+                ));
+            }
```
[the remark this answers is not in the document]
jmpesp: good call, done in 37f4e6ef4

nexus/src/app/sagas/common_storage.rs
```diff
@@ -148,6 +148,35 @@ pub(super) async fn delete_crucible_region(
...
 ) -> Result<(), Error> {
+    // If the region never existed, then a `GET` will return 404, and so will a
+    // `DELETE`. Catch this case, and return Ok if the region never existed.
+    // This can occur if an `ensure_all_datasets_and_regions` partially fails.
+    let result = retry_until_known_result(log, || async {
+        client.region_get(&RegionId(region_id.to_string())).await
+    })
+    .await;
+
+    if let Err(e) = result {
```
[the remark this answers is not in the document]
jmpesp: yes! done in b06888421

nexus/tests/integration_tests/volume_management.rs
jmpesp: Good idea - I like the bool better, as the simulated sled agent is only returning 500s at that point. Changed in cc260f2f5

gjcolombo (approved):
LGTM--thanks for the changes!

Issue #2977, opened 2023-05-01
Nexus should pass request for "all services on a sled" as a single endpoint with a generation number


smklein:
https://github.com/oxidecomputer/omicron/pull/2972#issuecomment-1530453752 for some context, but it basically would let the "local ledger of services" on a sled be more cleanly a cache of data from CRDB.

davepacheco:
@smklein This was done already, right?

smklein:
Same as https://github.com/oxidecomputer/omicron/issues/2978#issuecomment-1711016927

davepacheco:
This behavior was folded into #4466.

PR #4061, opened 2023-09-09
[sled agent] Service API now uses generation numbers

smklein:
- Adds a generation number to the Sled Agent API when asking for services
- Sled Agent uses this generation number to discard out-of-date requests
- Sled Agent also responds with the existing generation number if requests arrive that are out-of-date

This is a necessary pre-requisite to performing #732 safely.

Fixes https://github.com/oxidecomputer/omicron/issues/2977 , https://github.com/oxidecomputer/omicron/issues/2978

### DIFF
--- a/common/src/api/external/mod.rs
+++ b/common/src/api/external/mod.rs
@@ -627,6 +627,7 @@ impl Generation {
         Generation(1)
     }
 
+    #[must_use]
     pub fn next(&self) -> Generation {
         // It should technically be an operational error if this wraps or even
         // exceeds the value allowed by an i64.  But it seems unlikely enough to
--- a/sled-agent/src/params.rs
+++ b/sled-agent/src/params.rs
@@ -8,6 +8,7 @@ pub use crate::zone_bundle::ZoneBundleId;
 pub use crate::zone_bundle::ZoneBundleMetadata;
 pub use illumos_utils::opte::params::VpcFirewallRule;
 pub use illumos_utils::opte::params::VpcFirewallRulesEnsureBody;
+use omicron_common::api::external::Generation;
 use omicron_common::api::internal::nexus::{
     DiskRuntimeState, InstanceRuntimeState,
 };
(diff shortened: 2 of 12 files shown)


sled-agent/src/params.rs
```diff
@@ -843,9 +844,25 @@ impl TryFrom<ServiceZoneService>
...
+    NotUpdated,
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema, PartialEq)]
+pub struct ServiceEnsureResponse {
+    // Returns the current generation number.
+    pub generation: Generation,
+    // Identifies whether or not this request updated the services on the sled.
+    pub status: ServiceEnsureStatus,
+}
```
smklein, line 864: Why respond with the generation number at all? Under "normal operating conditions", it would be nice if "whoever made calls always knew the existing generation number". Well...

1. We didn't embed this generation number in the RSS handoff request...
2. ... so Nexus doesn't know what it is...
3. ... and the way RSS does setup is more complex than just "requesting all services at once", so it's not reliably Generation::new().next() -- it may have been called a couple times.

So, this is my workaround. Sled Agent simply tells the caller if it did an update or not, and also tells the caller "by the way, this is my current generation number".

sled-agent/src/services.rs
```diff
@@ -270,15 +267,20 @@ impl Config {
...
 }
 
-// The filename of the ledger, within the provided directory.
-const SERVICES_LEDGER_FILENAME: &str = "services.json";
-
-// A wrapper around `ZoneRequest`, which allows it to be serialized
-// to a JSON file.
+/// A wrapper around `ZoneRequest`, which allows it to be serialized
+/// to a JSON file.
+// TODO: Should this be here? Could be in a different file
```
smklein, line 272: TODO(me): clean this up.

This is the "ledger" of requests that is stored on M.2s on the Sled. Fortunately, it didn't change, because I *already* snuck a generation number in here.

Unfortunately, it got hoisted up into the "sled_agent.rs", which means that "services.rs" is slightly-less-aware of it. Probably shouldn't live here.

common/src/api/external/mod.rs
```diff
@@ -627,6 +627,7 @@ impl Generation {
         Generation(1)
     }
 
+    #[must_use]
     pub fn next(&self) -> Generation {
```
smklein, line 631: I hit this issue locally, thinking that this was acting on `&mut self`, so I figured I'd the annotation to guard against a similar mistake.

sled-agent/src/rack_setup/service.rs
```diff
@@ -281,14 +282,31 @@ impl ServiceInner {
 
         info!(self.log, "sending service requests...");
         let services_put = || async {
+            let mut generation = Generation::new();
```
smklein, line 285: During rack setup, we don't really care about the generation numbers, so we basically just keep bumping this until it works.

sled-agent/src/sled_agent.rs
```diff
@@ -723,11 +793,37 @@ impl SledAgent {
                 .await?;
         }
 
-        self.inner
```
smklein, line 726: This was a big reason for the churn in this PR, and the rationale for "why did we pull up the ledger and put a lock on it?"

- The ledger covers not **only** the set of services, but also the datasets that should be used
- Whether or not we should perform updates depends on the generation number of the ledger, so we need to guard access to the ledger between "reading it" and "maybe updating it"

In the old version of this code, we "upserted filesystems" without a guard, before ensuring services. This wasn't protected by anything, and this portion of the code was not qualified by reading a generation number from the ledger.

sled-agent/src/params.rs
smklein: Follow-up option here -- we could just add a "nexus generation" as an optional to the thing stored in the ledger.

Should automatically be None, can be replaced by something more recent.

PR #4084, opened 2023-09-13
prototype Omicron debugger

davepacheco:
This PR:

- Adds *internal* Nexus APIs for fetching status of background tasks.  As part of doing this, I had to move a bunch of types from `omicron-nexus` into `nexus-types`.  I changed a few things around for clarity, like replacing `Option<Foo>` in a few places with an explicit enum with two variants.
- Adds`omdb`, a primordial Omicron debugger.  This started as a simple tool to print background task status.  I realized this could be a more general pattern and I wanted to flesh out a few other commands to see how they'd fit together.  There are lots of open questions here like "how do we manage interface stability", "should this have a REPL", and "how do we test this".  There's also some cleanup I'd like to do, like moving things around rather than adding a new top-level package.
- Fixes a bug where the internal API saga endpoints have not worked since #2916 because the endpoints were still looking for external API credentials despite now being on the internal API.

Despite the open questions, if we feel this is a step in a useful direction, I think it makes sense to land this sooner rather than later (before trying to figure all that out) so that we can get some experience with it and iterate on it.


andrewjstone:
@davepacheco AFAICT, this seems like a reasonable direction and builds on our existing background tasks stuff (that admittedly I hadn't looked at until today). 

Do you expect that we'll add to omdb for things like internal stats from other subsystems, like what you and I talked about with regards to exposure of bootstore stats from sled-agent?

davepacheco:
> @davepacheco AFAICT, this seems like a reasonable direction and builds on our existing background tasks stuff (that admittedly I hadn't looked at until today).
> 
> Do you expect that we'll add to omdb for things like internal stats from other subsystems, like what you and I talked about with regards to exposure of bootstore stats from sled-agent?

I think so!  But honestly I'm not sure where this thing will go.  I'm hopeful that it could be a general-purpose place to put any sort of thing that reaches out to a running component to get some state and print it out.  The target is a developer or support person.  I could see us adding loads of useful commands here to query all of our runtime components...or finding that it's not really the right pattern for some reason.

leftwo:
Do you have some example output showing how to use this tool?
I've kicked around various things for Crucible, but could not settle on something
that was light enough that I thought would do what I needed, so I'm curious to
see this `omdb` in action!

omdb/src/bin/omdb/db.rs
```diff
@@ -0,0 +1,370 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
jordanhendricks, line 1: Hurray for another rust tool!

omdb/src/bin/omdb/db.rs
```diff
@@ -0,0 +1,370 @@
...
+        &self,
+        log: &slog::Logger,
+    ) -> Result<(), anyhow::Error> {
+        let db_config = db::Config { url: self.db_url.clone() };
+        let pool = Arc::new(db::Pool::new(&log.clone(), &db_config));
+
+        // Being a dev tool, we want to try this operation even if the schema
+        // doesn't match what we expect.  So we use `DataStore::new_unchecked()`
+        // here.  We will then check the schema version explicitly and warn the
+        // user if it doesn't match.
```
jordanhendricks, line 77: nice

jordanhendricks (approved):
Gave this a look over like I said I would, though I didn't give it as much attention as I would other reviews since it's a prototype of a tool. I really like this direction (not surprisingly as a fan of mdb and similar debugging tools), but I mostly think this would be *great* to build on for debugging the dogfood rack. Something I do constantly is find where things are (e.g., where are the downstairs of this failed disk on X instance?), and you basically need a client for crdb to have any useful tooling there. It seems like we could build that pretty easily into this.

I am curious if there are things you would think make us think it's the wrong direction? Maybe that's too abstract, but just wondering what we might want to look out for.

omdb/src/bin/omdb/nexus.rs
```diff
@@ -0,0 +1,246 @@
...
+
+#[derive(Debug, Subcommand)]
+enum BackgroundTaskCommands {
+    /// Show documentation about background tasks
+    Doc,
+    /// Print a summary of the status of all background tasks
+    List,
+    /// Print human-readable summary of the status of each background task
+    Details,
+}
```
jordanhendricks, line 48: Another thing I like about this tool is how it describes some of the implementation in a useful way, which is educational to folks observing the system. These types of subcommands are great for that.

davepacheco:
> I am curious if there are things you would think make us think it's the wrong direction? Maybe that's too abstract, but just wondering what we might want to look out for.

Not exactly...the sorts of things I could imagine are like: we might decide we want to focus on a totally different approach instead (say opentracing-style data or log aggregation).  Or we decide it should have a REPL more like a traditional debugger.  (I think that'd be easy to add here, actually.)  Or we decide it's a bad idea to have something like this talk directly to the database.  I do think it's unlikely this would be _bad_ and it's a low-risk experiment.

PR #4088, opened 2023-09-13
VERY WIP background tasks to propagate service requests to sleds

smklein:
- not at all ready yet, plenty of TODOs, but it's the rough shape I want

### DIFF
--- a/nexus/src/app/background/init.rs
+++ b/nexus/src/app/background/init.rs
@@ -9,13 +9,17 @@ use super::dns_config;
 use super::dns_propagation;
 use super::dns_servers;
 use super::external_endpoints;
+use super::services_config;
+use super::services_propagation;
+use super::sled_servers;
 use nexus_db_model::DnsGroup;
 use nexus_db_queries::context::OpContext;
 use nexus_db_queries::db::DataStore;
 use omicron_common::nexus_config::BackgroundTaskConfig;
 use omicron_common::nexus_config::DnsTasksConfig;
+use once_cell::sync::OnceCell;
 use std::collections::BTreeMap;
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
 
 /// Describes ongoing background tasks and provides interfaces for working with
 /// them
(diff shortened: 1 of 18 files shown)


nexus/src/app/background/init.rs
```diff
@@ -25,7 +29,7 @@ use std::sync::Arc;
 pub struct BackgroundTasks {
     /// interface for working with background tasks (activation, checking
     /// status, etc.)
-    pub driver: common::Driver,
+    pub driver: Mutex<common::Driver>,
```
smklein, line 32: Feedback: Just have the tasks check if handoff is complete.

nexus/src/app/background/services_config.rs
```diff
@@ -0,0 +1,378 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
smklein, line 1: To the "service table":

- Add a global tracking service generation table
- Add a "added in" and "removed from" value in the service table

nexus/src/app/background/services_propagation.rs
```diff
@@ -0,0 +1,290 @@
...
+            info!(
+                log,
+                "Sled Services Propagated";
+                "address" => sled_addr,
+                "generation" => services.generation.to_string(),
+            );
+            Ok(())
+        }
+    }
+}
```
smklein, line 143: Something to be aware of: One bad sled could block us here? Maybe we should be tracking this "task" durably across the task "activate" calls, and we could drop if it we no longer care about this sled.

smklein: maybe punt on this, DNS propagation has the same problem

smklein:
Hey, you, reading this well before it's ready - Hi! I'm still poking around here, but figured out that there's a fair bit of accounting necessary before we can successfully send these requests.

I've started putting some of this detail into a doc here: https://docs.google.com/document/d/1AcK5ucGmzbG_JfVuZ8_CrLJjqgiVEcbZtS5fp6zm8GU/edit?usp=sharing , which I'll probably turn into an RFD eventually.

PR #4092, opened 2023-09-14
Bundle rotated log files

jmpesp:
Camino's Utf8Path only considers whole path components to match, and this was causing the zone bundle tool to miss rotated logfiles. Convert the paths into Strings, then use those object's `starts_with` instead.

### DIFF
--- a/illumos-utils/src/running_zone.rs
+++ b/illumos-utils/src/running_zone.rs
@@ -1026,10 +1026,12 @@ impl RunningZone {
             let path = entry.path();
 
             // Camino's Utf8Path only considers whole path components to match,
-            // so convert both paths into a String and use that object's
+            // so convert both paths into a &str and use that object's
             // starts_with. See the `camino_starts_with_behaviour` test.
+            let path_ref: &str = path.as_ref();
+            let current_log_file_ref: &str = current_log_file.as_ref();
             if path != current_log_file
-                && path.to_string().starts_with(&current_log_file.to_string())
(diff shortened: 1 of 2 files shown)


illumos-utils/src/running_zone.rs
```diff
@@ -1020,19 +1020,42 @@ impl RunningZone {
...
             let path = entry.path();
-            if path != current_log_file && path.starts_with(&current_log_file) {
-                rotated_files
-                    .push(root.join(path.strip_prefix("/").unwrap_or(path)));
+
+            // Camino's Utf8Path only considers whole path components to match,
+            // so convert both paths into a String and use that object's
+            // starts_with. See the `camino_starts_with_behaviour` test.
+            if path != current_log_file
+                && path.to_string().starts_with(&current_log_file.to_string())
```
bnaecker, line 1032: Since `Utf8PathBuf` implements `AsRef<str>`, I think we can accomplish this without temporaries? Something like:

```rust
let full: &str = path.as_ref();
if path != current_log_file && full.starts_with(current_log_file.as_ref()) {
    ...
}
```

bnaecker (approved):
Nice catch, thanks. I've one suggestion to avoid copies, but otherwise 🚢 !

illumos-utils/src/running_zone.rs
jmpesp: Nice! Done in 73989a9bb

PR #4095, opened 2023-09-15
Do not double count region snapshots records!

jmpesp:
`decrease_crucible_resource_count_and_soft_delete_volume` does not disambiguate cases where the snapshot_addr of a region_snapshot is duplicated with another one, which can occur due to the Crucible Agent reclaiming ports from destroyed daemons (see also #4049, which makes the simulated Crucible agent do this).

Several invocations of the snapshot create and snapshot delete sagas could race in such a way that one of these ports would be reclaimed, and then be used in a different snapshot, and the lifetime of both of these would overlap! This would confuse our reference counting, which was written with a naive assumption that this port reuse **wouldn't** occur with these overlapping lifetimes. Spoiler alert, it can:

    root@[fd00:1122:3344:101::3]:32221/omicron> select * from region_snapshot where snapshot_addr = '[fd00:1122:3344:102::7]:19016';
                   dataset_id              |              region_id               |             snapshot_id              |         snapshot_addr         | volume_references
    ---------------------------------------+--------------------------------------+--------------------------------------+-------------------------------+--------------------
      80790bfd-4b81-4381-9262-20912e3826cc | 0387bbb7-1d54-4683-943c-6c17d6804de9 | 1a800928-8f93-4cd3-9df1-4129582ffc20 | [fd00:1122:3344:102::7]:19016 |                 0
      80790bfd-4b81-4381-9262-20912e3826cc | ff20e066-8815-4eb6-ac84-fab9b9103462 | bdd9614e-f089-4a94-ae46-e10b96b79ba3 | [fd00:1122:3344:102::7]:19016 |                 0
    (2 rows)

One way to solve this would be to create a UNIQUE INDEX on `snapshot_addr` here, but then in these cases the snapshot creation would return a 500 error to the user.

This commit adds a sixth column: `deleting`, a boolean that is true when the region snapshot is part of a volume's `resources_to_clean_up`, and false otherwise. This is used to select (as part of the transaction for `decrease_crucible_resource_count_and_soft_delete_volume`) only the region_snapshot records that were decremented as part of that transaction, and skip re-deleting them otherwise.

This works because the overlapping lifetime of the records in the DB is **not** the overlapping lifetime of the actual read-only downstairs daemon: for the port to be reclaimed, the original daemon has to be DELETEd, which happens after the decrement transaction has already computed which resources to clean up:

1) a snapshot record is created:

```
            snapshot_id              |         snapshot_addr         | volume_references | deleted |
-------------------------------------+-------------------------------+-------------------+----------
1a800928-8f93-4cd3-9df1-4129582ffc20 | [fd00:1122:3344:102::7]:19016 |                 0 |  false  |
```

2) it is incremented as part of `volume_create`:

```
            snapshot_id              |         snapshot_addr         | volume_references | deleted |
-------------------------------------+-------------------------------+-------------------+----------
1a800928-8f93-4cd3-9df1-4129582ffc20 | [fd00:1122:3344:102::7]:19016 |                 1 |  false  |
```

3) when the volume is deleted, then the decrement transaction will:

  a) decrease `volume_references` by 1

```
            snapshot_id              |         snapshot_addr         | volume_references | deleted |
-------------------------------------+-------------------------------+-------------------+----------
1a800928-8f93-4cd3-9df1-4129582ffc20 | [fd00:1122:3344:102::7]:19016 |                 0 |  false  |
```

  b) note any `region_snapshot` records whose `volume_references` went to
     0 and have `deleted` = false, and return those in the list of
     resources to clean up:

     [ 1a800928-8f93-4cd3-9df1-4129582ffc20 ]

  c) set deleted = true for any region_snapshot records whose
     `volume_references` went to 0 and have deleted = false
```
            snapshot_id              |         snapshot_addr         | volume_references | deleted |
-------------------------------------+-------------------------------+-------------------+----------
1a800928-8f93-4cd3-9df1-4129582ffc20 | [fd00:1122:3344:102::7]:19016 |                 0 |   true  |
```

4) That read-only snapshot daemon is DELETEd, freeing up the port.
   Another snapshot creation occurs, using that reclaimed port:

```
            snapshot_id              |         snapshot_addr         | volume_references | deleted |
-------------------------------------+-------------------------------+-------------------+----------
1a800928-8f93-4cd3-9df1-4129582ffc20 | [fd00:1122:3344:102::7]:19016 |                 0 |   true  |
bdd9614e-f089-4a94-ae46-e10b96b79ba3 | [fd00:1122:3344:102::7]:19016 |                 0 |  false  |
```

5) That new snapshot is incremented as part of `volume_create`:

```
            snapshot_id              |         snapshot_addr         | volume_references | deleted |
-------------------------------------+-------------------------------+-------------------+----------
1a800928-8f93-4cd3-9df1-4129582ffc20 | [fd00:1122:3344:102::7]:19016 |                 0 |   true  |
bdd9614e-f089-4a94-ae46-e10b96b79ba3 | [fd00:1122:3344:102::7]:19016 |                 1 |  false  |
```

6) It is later deleted, and the decrement transaction will:

  a) decrease `volume_references` by 1:

```
j           snapshot_id              |         snapshot_addr         | volume_references | deleted |
-------------------------------------+-------------------------------+-------------------+----------
1a800928-8f93-4cd3-9df1-4129582ffc20 | [fd00:1122:3344:102::7]:19016 |                 0 |   true  |
bdd9614e-f089-4a94-ae46-e10b96b79ba3 | [fd00:1122:3344:102::7]:19016 |                 0 |  false  |
```

  b) note any `region_snapshot` records whose `volume_references` went to
     0 and have `deleted` = false, and return those in the list of
     resources to clean up:

     [ bdd9614e-f089-4a94-ae46-e10b96b79ba3 ]

  c) set deleted = true for any region_snapshot records whose
     `volume_references` went to 0 and have deleted = false

```
            snapshot_id              |         snapshot_addr         | volume_references | deleted |
-------------------------------------+-------------------------------+-------------------+----------
1a800928-8f93-4cd3-9df1-4129582ffc20 | [fd00:1122:3344:102::7]:19016 |                 0 |   true  |
bdd9614e-f089-4a94-ae46-e10b96b79ba3 | [fd00:1122:3344:102::7]:19016 |                 0 |   true  |
```


nexus/db-queries/src/db/datastore/volume.rs
```diff
@@ -707,12 +736,41 @@ impl DataStore {
...
+                            // XXX are any other filters required, except the
+                            // three from snapshots_to_delete?
                             .filter(
                                 dsl::snapshot_addr.eq_any(
                                     &crucible_targets.read_only_targets,
                                 ),
                             )
-                            .filter(dsl::volume_references.eq(0))
+                            // XXX is there a better way to do this, rather than
+                            // one filter per struct field?
```
jgallagher, line 748: Is it possible this could be incorrect (in addition to being awkward, which I assume is the reason for the comment)? E.g., if `snapshots_to_delete` contained the dataset_id/region_id/snapshot_id triples

```
(A, B, C),
(D, E, F),
```

and the database contained a row with `(A, B, F)`, then the three independent filters would all be true even though `(A, B, F)` itself is not one of the triples we're trying to delete. I _think_ such a contrived example is not actually possible, but I'm not entirely sure?

I'm not sure how to rework this, particularly in diesel. In SQL terms I think we'd want something like a subquery containing the `snapshots_to_delete` (instead of having them in a `Vec`) so that we could do a `WHERE EXISTS ...` on the three columns we care about, maybe? If we want to keep `snapshots_to_delete` in Rust, how bad would it be to loop over them and run this query once for each of them?

schema/crdb/5.0.0/up.sql
```diff
@@ -0,0 +1,25 @@
...
+            SELECT version = '4.0.0' and target_version = '5.0.0'
+            FROM omicron.public.db_metadata WHERE singleton = true
+        ),
+        'true',
+        'Invalid starting version for schema change'
+    ) AS BOOL
+);
+
+ALTER TABLE omicron.public.region_snapshot
+    ADD COLUMN IF NOT EXISTS deleting BOOL NOT NULL;
```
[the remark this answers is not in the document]
jmpesp: You're right, existing rows cause an error:
```
root@:26257/omicron> insert into region_snapshot values ('20047568-e275-4bb7-8e27-e5ac7fc4dd30', 'e0876e2c-0f50-4f29-9565-abd57a692e51', '0cb03485-0a26-40bb-99ff-85028980ebc0', '[fd00:1122:3344:101::9]:19000', 1, false);
INSERT 1

Time: 6ms total (execution 5ms / network 0ms)

root@:26257/omicron> alter table region_snapshot add column if not exists testing bool not null;
ERROR: null value in column "testing" violates not-null constraint
SQLSTATE: 23502

root@:26257/omicron> alter table region_snapshot add column if not exists testing bool not null default false;
ALTER TABLE

Time: 52ms total (execution 52ms / network 0ms)

root@:26257/omicron> alter table region_snapshot alter column testing drop default;
ALTER TABLE

Time: 25ms total (execution 25ms / network 0ms)
```

Added in 623e4478

nexus/db-queries/src/db/datastore/volume.rs
jmpesp: Turns out I can chain them together with `.and`: 042b6125

jgallagher: I think that still has the problem of `snapshots_to_delete = (A,B,C), (D,E,F)` matching `(A,B,F)`, right? But maybe that's fine / impossible in practice? It is an improvement over the chained filters for sure though.

PR #4098, opened 2023-09-15
flesh out `omdb`

davepacheco:
This PR has a bunch of small improvements to omdb:

* `omdb nexus background-tasks`:
    * prints the configured period
    * propagation task: shows per-server propagation results
    * lots of little display improvements
* adds `omdb db dns` (subcommands for `show`, `diff`, and `names`)
* adds support for configuring URLs through the environment
* adds some (very rough) guidance for people wanting to add stuff (in the form of some module-level comments)

### DIFF
--- a/omdb/src/bin/omdb/main.rs
+++ b/omdb/src/bin/omdb/main.rs
@@ -3,6 +3,21 @@
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 //! CLI for debugging Omicron internal state
+//!
+//! GROUND RULES:
+//!
+//! 1. There aren't a lot of ground rules here.  At least for now, this is a
+//!    place to put any kind of runtime tooling for Omicron that seems useful.
+//!    You can query the database directly (see notes in db.rs), use internal
+//!    APIs, etc.  To the degree that we can stick to stable interfaces, great.
+//!    But at this stage we'd rather have tools that work on latest than not
+//!    have them because we couldn't prioritize keeping them stable.
+//!
+//! 2. Where possible, when the tool encounters something unexpected, it should
+//!    print what it can (including the error message and bad data) and then
+//!    continue.  It generally shouldn't stop on the first error.  (We often
+//!    find strange things when debugging but we need our tools to tell us as
+//!    much as they can!)
 
 use anyhow::Context;
 use clap::Parser;
--- a/omdb/src/bin/omdb/nexus.rs
+++ b/omdb/src/bin/omdb/nexus.rs
@@ -6,6 +6,7 @@
 
 use anyhow::Context;
 use chrono::SecondsFormat;
+use chrono::Utc;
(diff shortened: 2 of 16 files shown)


omdb/src/bin/omdb/main.rs
```diff
@@ -3,6 +3,21 @@
...
+//! GROUND RULES:
+//!
+//! 1. There aren't a lot of ground rules here.  At least for now, this is a
+//!    place to put any kind of runtime tooling for Omicron that seems useful.
+//!    You can query the database directly (see notes in db.rs), use internal
+//!    APIs, etc.  To the degree that we can stick to stable interfaces, great.
+//!    But at this stage we'd rather have tools that work on latest than not
+//!    have them because we couldn't prioritize keeping them stable.
+//!
+//! 2. Where possible, when the tool encounters something unexpected, it should
```
andrewjstone, line 16: Very important!

omdb/src/bin/omdb/nexus.rs
```diff
@@ -4,21 +4,27 @@
...
+use std::collections::BTreeMap;
 use tabled::Tabled;
+use uuid::Uuid;
 
 /// Arguments to the "omdb nexus" subcommand
 #[derive(Debug, Args)]
 pub struct NexusArgs {
     /// URL of the Nexus internal API
-    nexus_internal_url: String,
+    #[clap(long, env("OMDB_NEXUS_URL"))]
```
andrewjstone, line 26: TIL

omdb/src/bin/omdb/nexus.rs
```diff
@@ -157,13 +193,285 @@ fn print_task(bgtask: &BackgroundTask) {
...
+    }
+    if let Ok(found_error) =
+        serde_json::from_value::<TaskError>(details.clone())
+    {
+        println!("    last completion reported error: {}", found_error.error);
+        println!("");
+        return;
+    }
+
+    // The rest is task-specific, keyed by the name.
```
andrewjstone, line 239: For things keyed by name, what if, if they don't deserialize, we instead format the json output rather than return just an error? One of the benefits of json is that it has named/tagged values, so we can just dump the value pretty-printed if we can't manually format it nicely.

If we don't want to clutter our output with pretty-printed json we could provide an option to skip this and just return the error.

andrewjstone (approved):
Great stuff. Love seeing more of this.

omdb/src/bin/omdb/nexus.rs
leftwo: lol!  Niiiiiice

omdb/src/bin/omdb/nexus.rs
davepacheco: Yes, great idea.  (I actually thought we were already doing that, but we only do that right now for tasks we don't recognize.)

PR #4100, opened 2023-09-15
Add hostname validation to certs uploaded to nexus

### LINKED ISSUE #4045: Validate CN and/or SAN in certificate matches external dns zone during RSS.

jgallagher:
This PR changes the `CertificateValidator` from taking "0 or 1" hostname to validate to taking a list of hostnames, and the cert passes validation if _any_ hostname from the list matches. This is to support a future world where we support adding and removing external DNS zone names, where an operator might perform an operation like:

1. The rack has an external name of `myrack.oxide.computer`, and every silo has a `$SILO.sys.myrack.oxide.computer` cert
2. Add a new external name of `something-else.oxide.computer`
3. Upload new certs to all silos with SAN `$SILO.sys.something-else.oxide.computer`
4. Remove the external name `myrack.oxide.computer`

To be clear this is a hypothetical world, because today we don't support modifying the external DNS name at all. But step 3 from this sequence is the argument for why the validation logic is "any of the hostnames" instead of "all of the hostnames": we want to add certs for the new hostname (that don't need to also work for the old hostname we're about to remove).

The check is performed three places:

1. When the recovery silo is created (which is maybe wrong? I could see an argument for omitting this check, although since wicket is also performing it it should certainly pass)
2. When any other silo is created
3. When a cert is uploaded to an existing silo

In all cases we build the list of hostnames for the validator by combining the list of external DNS zone names (which is currently always length 1) with the silo's `.sys` prefix. For the recovery silo specifically we build the list of length 1 by hand; for other silos and when adding to an existing silo, we query CRDB for all external DNS zone names.

I haven't yet deployed this to a full system and want to do that before merging, but this passes tests and we're about to have a busy week, so I wanted to go ahead and open it as an easy way to keep it on the radar.

Fixes #4045.

### DIFF
--- a/certificates/src/lib.rs
+++ b/certificates/src/lib.rs
@@ -9,6 +9,7 @@ use omicron_common::api::external::Error;
 use openssl::asn1::Asn1Time;
 use openssl::pkey::PKey;
 use openssl::x509::X509;
+use std::borrow::Borrow;
 use std::ffi::CString;
 
 mod openssl_ext;
@@ -103,15 +104,15 @@ impl CertificateValidator {
     /// `key` is expected to be the private key for the leaf certificate of
     /// `certs` in PEM format.
     ///
-    /// If `hostname` is not `None`, the leaf certificate of `certs` must be
-    /// valid for `hostname`, as determined by a dNSName entry in its subject
-    /// alternate names or (if there are no dNSName SANs) the cert's common
-    /// name.
-    pub fn validate(
+    /// If `possible_hostnames` is empty, no hostname validation is performed.
+    /// If `possible_hostnames` is not empty, we require _at least one_ of its
+    /// hostnames to match the SANs (or CN, of no SANs are present) of the leaf
+    /// certificate.
+    pub fn validate<S: Borrow<str>>(
         &self,
         certs: &[u8],
         key: &[u8],
-        hostname: Option<&str>,
+        possible_hostnames: &[S],
     ) -> Result<(), CertificateError> {
         // Checks on the certs themselves.
         let mut certs = X509::stack_from_pem(certs)
@@ -134,16 +135,25 @@ impl CertificateValidator {
         // to use with verifying the private key.
         let cert = certs.swap_remove(0);
 
-        if let Some(hostname) = hostname {
-            let c_hostname = CString::new(hostname).map_err(|_| {
-                CertificateError::InvalidValidationHostname(
-                    hostname.to_string(),
-                )
-            })?;
-            if !cert
-                .valid_for_hostname(&c_hostname)
-                .map_err(CertificateError::ErrorValidatingHostname)?
-            {
+        if !possible_hostnames.is_empty() {
+            let mut found_valid_hostname = false;
+            for hostname in possible_hostnames {
+                let hostname = hostname.borrow();
+                let c_hostname = CString::new(hostname).map_err(|_| {
+                    CertificateError::InvalidValidationHostname(
+                        hostname.to_string(),
+                    )
+                })?;
+                if cert
+                    .valid_for_hostname(&c_hostname)
+                    .map_err(CertificateError::ErrorValidatingHostname)?
+                {
+                    found_valid_hostname = true;
+                    break;
+                }
+            }
+
+            if !found_valid_hostname {
                 let cert_description =
                     cert.hostname_description().unwrap_or_else(|err| {
                         format!(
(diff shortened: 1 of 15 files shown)


certificates/src/lib.rs
```diff
@@ -103,15 +104,15 @@ impl CertificateValidator {
...
     /// `certs` in PEM format.
     ///
-    /// If `hostname` is not `None`, the leaf certificate of `certs` must be
-    /// valid for `hostname`, as determined by a dNSName entry in its subject
-    /// alternate names or (if there are no dNSName SANs) the cert's common
-    /// name.
-    pub fn validate(
+    /// If `possible_hostnames` is empty, no hostname validation is performed.
+    /// If `possible_hostnames` is not empty, we require _at least one_ of its
+    /// hostnames to match the SANs (or CN, of no SANs are present) of the leaf
```
luqmana, line 109: typo: "if no SANs are present"

certificates/src/lib.rs
```diff
@@ -103,15 +104,15 @@ impl CertificateValidator {
     /// `key` is expected to be the private key for the leaf certificate of
     /// `certs` in PEM format.
     ///
-    /// If `hostname` is not `None`, the leaf certificate of `certs` must be
-    /// valid for `hostname`, as determined by a dNSName entry in its subject
-    /// alternate names or (if there are no dNSName SANs) the cert's common
-    /// name.
-    pub fn validate(
+    /// If `possible_hostnames` is empty, no hostname validation is performed.
```
luqmana, line 107: Should `possible_hostnames` being empty just be an error?

certificates/src/lib.rs
```diff
@@ -134,16 +135,25 @@ impl CertificateValidator {
...
-            {
+        if !possible_hostnames.is_empty() {
+            let mut found_valid_hostname = false;
+            for hostname in possible_hostnames {
+                let hostname = hostname.borrow();
+                let c_hostname = CString::new(hostname).map_err(|_| {
+                    CertificateError::InvalidValidationHostname(
+                        hostname.to_string(),
+                    )
+                })?;
```
luqmana, line 146: Could we just warn on any invalid hostnames and continue with any others in the list instead of immediately erroring?

certificates/src/lib.rs
```diff
@@ -38,7 +39,7 @@ pub enum CertificateError {
     #[error("Error validating certificate hostname")]
     ErrorValidatingHostname(#[source] openssl::error::ErrorStack),
 
-    #[error("Certificate not valid for {hostname:?}: {cert_description:?}")]
+    #[error("Certificate not valid for {hostname:?}: {cert_description}")]
```
luqmana, line 42: Update error text to indicate possible set of hostnames?
```suggestion
    #[error("Certificate not valid for given hostnames ({hostname:?}): {cert_description}")]
```

nexus/src/app/certificate.rs
```diff
@@ -71,6 +78,25 @@ impl super::Nexus {
         }
     }
 
+    async fn silo_fq_dns_names(
```
luqmana, line 81: Nit: move to nexus/src/app/silo.rs

luqmana (commented):
Thanks for hooking up the rest of this!

certificates/src/lib.rs
jgallagher: There's one path in wicket where we currently validate a cert without a hostname: if the operator uploads a cert _before_ uploading the TOML config, we don't have a hostname to validate but can still check that the key matches the cert. We will still validate the cert with the recovery silo hostname prior to kicking off RSS. We could change that - it doesn't seem unreasonable to say "you must upload the TOML before uploading certs/keys"?

certificates/src/lib.rs
jgallagher: Is there a case where we would expect an invalid hostname here that's not indicative of some deeper problem? The hostnames here should always be of the form `{silo}.sys.{external_dns_zone_name}` (which we construct), so I _think_ the only way they could be invalid is if someone is specifying a bad silo or external DNS zone name. I wouldn't be surprised if we'd catch that earlier than here, but it doesn't seem harmful to fail ASAP.

I'm also not sure about warning - that won't end up visible to the operator, unless you mean building up a list of errors so we can send multiple failures at once?

certificates/src/lib.rs
luqmana: true, i think it's fine as-is then.

luqmana (approved):
Thanks John!

PR #4101, opened 2023-09-16
add smoke test for `omdb`

davepacheco:
This is a simple expectorate-based test to verify output of `omdb`.  The goal here is to help people iterate on omdb without unknowingly breaking basic functionality.  At this point, I feel we should feel totally comfortable changing the output.  The point of the test is to be able to tell when you're doing it (and how).  The test is far from exhaustive.

One annoying thing is that the information `omdb` prints necessarily varies between test runs (e.g., timestamps, internal service TCP port numbers).  So there's a process here for redacting that stuff so that the expectorate output looks the same.


omdb/src/bin/omdb/nexus.rs
```diff
@@ -34,17 +34,17 @@ pub struct NexusArgs {
 #[derive(Debug, Subcommand)]
 enum NexusCommands {
     /// print information about background tasks
-    BackgroundTask(BackgroundTaskArgs),
+    BackgroundTasks(BackgroundTasksArgs),
```
davepacheco, line 37: I just noticed the lack of plural was inconsistent with other subcommands.

jordanhendricks (approved):
Nice! Great idea.

Issue #4087, opened 2023-09-13
Cockroach failed to start after cold boot


jgallagher:
After a mupdate + reboot of the dogfood rack, one of the CRDB zones showed that CRDB failed to start:

```
BRM42220017 # zlogin oxz_cockroachdb_3237a532-acaa-4ebe-bf11-dde794fea739
root@oxz_cockroachdb_3237a532-acaa-4ebe-bf11-dde794fea739:~# svcs -xv
svc:/oxide/cockroachdb:default (CockroachDB)
 State: maintenance since Tue Sep 12 20:40:47 2023
Reason: Restarting too quickly.
   See: http://illumos.org/msg/SMF-8000-L5
   See: /var/svc/log/oxide-cockroachdb:default.log
Impact: This service is not running.
```

Logfiles have been copied to `/staff/core/omicron-4087`.

In the `svcs` log `rotated-zone-logs/3cafbb47-c194-4a42-99ff-34dfeab999ed/crypt/debug/oxz_cockroachdb_3237a532-acaa-4ebe-bf11-dde794fea739/oxide-cockroachdb:default.log.1694551247`, we see that we tried to start CRDB several times:

```
/staff/core/omicron-4087 $ grep ' Sep ' rotated-zone-logs/3cafbb47-c194-4a42-99ff-34dfeab999ed/crypt/debug/oxz_cockroachdb_3237a532-acaa-4ebe-bf11-dde794fea739/oxide-cockroachdb\:default.log.1694551247
[ Sep 12 20:40:12 Enabled. ]
[ Sep 12 20:40:12 Rereading configuration. ]
[ Sep 12 20:40:13 Rereading configuration. ]
[ Sep 12 20:40:24 Executing start method ("/opt/oxide/lib/svc/manifest/cockroachdb.sh"). ]
[ Sep 12 20:40:26 Method "start" exited with status 0. ]
[ Sep 12 20:40:28 Stopping because all processes in service exited. ]
[ Sep 12 20:40:28 Executing stop method (:kill). ]
[ Sep 12 20:40:28 Executing start method ("/opt/oxide/lib/svc/manifest/cockroachdb.sh"). ]
[ Sep 12 20:40:34 Method "start" exited with status 0. ]
[ Sep 12 20:40:36 Stopping because all processes in service exited. ]
[ Sep 12 20:40:36 Executing stop method (:kill). ]
[...]

PR #4119, opened 2023-09-19
ntp: check for reasonable time and stratum < 10

iliana:
Context from https://github.com/oxidecomputer/omicron/issues/4087#issuecomment-1726031959:

> What we think happened here is the boundary NTP zones considered themselves authoritative without having upstream connectivity, and responded to other NTP zones with a 1980-era timestamp. The telling factor here is the stratum 11 lines in tracking.log, which imply the boundary NTP service was serving stratum 10.
>
> We could most easily work around this by ensuring we don't consider the host in sync until the stratum is < 10. ... We could also set up an epoch before which time cannot be to consider ourselves in sync, which might be a good additional defense?

This adds these checks. The epoch I've chosen for no particular reason is UNIX time 1234567890 (Fri Feb 13 23:31:30 2009 UTC); there are plenty of times after this that would not make sense but it at least ensures it is not December 1986.

### DIFF
--- a/openapi/sled-agent.json
+++ b/openapi/sled-agent.json
@@ -2755,6 +2755,17 @@
             "format": "uint32",
             "minimum": 0
           },
+          "ref_time": {
+            "description": "The NTP reference time (i.e. what chrony thinks the current time is, not necessarily the current system time).",
+            "type": "number",
+            "format": "double"
+          },
+          "stratum": {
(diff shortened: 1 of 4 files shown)


citrus-it (approved):
Looks reasonable to me. Thanks for getting to the bottom of what happened here.

PR #4123, opened 2023-09-20
Move instance networking functions into their own module

gjcolombo:
`nexus/src/app/instance.rs` and `sled.rs` implement several networking-related functions that are less directly concerned with instance or sled management than their sibling routines. Tidy things up a bit by creating an `instance_network` module and moving instance V2P and NAT management functions there. Also, move instance NAT entry deletion logic into its own function that's called from the instance delete saga instead of implementing it inline in that saga.

These changes aim to reduce clutter in `instance.rs` and to move NAT entry deletion to a function that can be reused by subsequent changes to the way Nexus handles instance stop.

Except for some minor edits to error handling in `instance_delete_dpd_config` (needed because this function no longer returns a `steno::ActionError`), this PR only rearranges existing code and has no functional changes.


nexus/src/app/instance.rs
```diff
@@ -777,7 +773,6 @@ impl super::Nexus {
 
         // Gather the SSH public keys of the actor make the request so
         // that they may be injected into the new image via cloud-init.
-        // TODO-security: this should be replaced with a lookup based on
```
smklein, line 780: Did you mean to delete this?

smklein (approved):
I checked that functions signatures look the same (and they do, with the exception of the new function) but not that the bodies are the same. Trusting you that they are!

Thanks for the cleanup.

nexus/src/app/instance.rs
gjcolombo: Whoops. Thanks for the catch!

gjcolombo: Fixed in 39aa48547.

Issue #1881, opened 2022-10-28
sled agent could better expose information about what it's currently doing


davepacheco:
I realize this is pretty vague, and may turn into a project or just a bunch of more targeted tickets.

Under #1880, when we hit #1879, we spent kind of a while just trying to figure out what was _currently_ going on.  It would have been pretty helpful if we'd been able to query the Sled Agent process and see:

- RSS's state: it had found a pre-existing plan and was attempting to resume executing it
- that RSS was waiting on a specific bootstrap agent request
- that the bootstrap agent request was in a retry loop

You can figure a bunch of this out from the log, the code, and a lot of reasoning.  It's just time-consuming and error-prone.

I've had some success with past systems having an endpoint for fetching state specifically for debugging.  In this case we could report state like:

```
"subsystems": {
    "rss": {
        "plan": "found", /* alternatively: "uninitialized", "created" */
        "steps": [ {
                "name": "do_the_things",
                "started": $timestamp,
                "done": $timestamp,
                "result": "success",
        }, ..., {
                "name": "initialize_agents",
                "started": $timestamp,
                "attempts": 27,
                "last_attempt": $timestamp,
                "next_attempt": $timestamp,
                "last_attempt_result": { "error": "Sled Agent already initialized" }
        } ]
    }
}
```

At Joyent we had composable libraries that would emit the pieces of this.  For example [vasync](https://github.com/TritonDataCenter/node-vasync) was a control flow library that emitted something like the "steps" field above.  An analog for us might be to have a function that takes a `FuturesOrdered` and produces the "steps" output above or something like that?  The idea was to make this available in a common way and queryable via a single tool like [kang](https://github.com/TritonDataCenter/kang), although we didn't wind up doing much with kang.  (We _did_ wind up putting all this into some dashboards that were very useful.)

Thoughts?

andrewjstone:
@davepacheco I'm totally with you here. I find this code incredibly hard to reason about as well as it's bouncing back and forth between RSS, bootstrap agent, and sled agent. 

I think that RSS is going to eventually wind up being its own service endpoint driven by wicket, and all the bootstrap initialization will be queryable from that endpoint and from the bootstore itself. We'll be able to aggregately query the bootstore to see the current status. My plan was to make this holistically viewable, but I haven't gotten as far as I would have liked. The bootstore coordinator is going to be closely tied to RSS as to make storage of that information tolerant of failures we'll need it to be replicated consistently. 

[...]

PR #4126, opened 2023-09-20
[omdb] Add a couple simple Sled Agent queries

smklein:
Related to https://github.com/oxidecomputer/omicron/issues/1881

### DIFF
--- a/omdb/src/bin/omdb/main.rs
+++ b/omdb/src/bin/omdb/main.rs
@@ -25,7 +25,7 @@ use clap::Subcommand;
 
 mod db;
 mod nexus;
-mod sled;
+mod sled_agent;
 
 #[tokio::main]
 async fn main() -> Result<(), anyhow::Error> {
@@ -38,7 +38,7 @@ async fn main() -> Result<(), anyhow::Error> {
     match args.command {
         OmdbCommands::Db(db) => db.run_cmd(&log).await,
         OmdbCommands::Nexus(nexus) => nexus.run_cmd(&log).await,
-        OmdbCommands::Sled(sled) => sled.run_cmd(&log).await,
+        OmdbCommands::SledAgent(sled) => sled.run_cmd(&log).await,
     }
 }
 
@@ -70,7 +70,7 @@ enum OmdbCommands {
     /// Debug a specific Nexus instance
     Nexus(nexus::NexusArgs),
     /// Debug a specific Sled
-    Sled(sled::SledArgs),
+    SledAgent(sled_agent::SledAgentArgs),
 }
 
 fn parse_dropshot_log_level(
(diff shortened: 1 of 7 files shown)


omdb/src/bin/omdb/main.rs
```diff
@@ -35,8 +36,9 @@ async fn main() -> Result<(), anyhow::Error> {
         .context("failed to create logger")?;
 
     match args.command {
-        OmdbCommands::Nexus(nexus) => nexus.run_cmd(&log).await,
         OmdbCommands::Db(db) => db.run_cmd(&log).await,
+        OmdbCommands::Nexus(nexus) => nexus.run_cmd(&log).await,
+        OmdbCommands::Sled(sled) => sled.run_cmd(&log).await,
```
davepacheco, line 41: I only wonder if this should be `sled-agent` (there are other things on the sled).  If you do I'd carry that through everywhere -- e.g., the URL env var name too.

omdb/src/bin/omdb/sled.rs
```diff
@@ -0,0 +1,109 @@
...
+    Zones(ZoneCommands),
+
+    /// print information about zpools
+    #[clap(subcommand)]
+    Zpools(ZpoolCommands),
+}
+
+#[derive(Debug, Subcommand)]
+enum ZoneCommands {
+    /// Print list of all zones
```
davepacheco, line 37:
```suggestion
    /// Print list of control plane zones that the sled agent is configured to run
```

(just trying to be precise about what it's printing so that someone doesn't interpret this as the same as `zoneadm list`)

edit: I just checked and it looks like this is only the *running* zones that are the subset of zones the sled agent is configured to run, I think.

omdb/src/bin/omdb/sled.rs
```diff
@@ -0,0 +1,109 @@
...
+    List,
+}
+
+#[derive(Debug, Subcommand)]
+enum ZpoolCommands {
+    /// Print list of all zpools
```
davepacheco, line 43:
```suggestion
    /// Print list of all zpools managed by the sled agent
```

davepacheco (approved):
Awesome!  A few small suggestions above.

Thanks also for alphabetizing the commands and fixing the copy/pasted comments.

davepacheco:
I forgot to mention: you will need to merge with #4101 and regenerate the output with EXPECTORATE.

omdb/src/bin/omdb/main.rs
smklein: Done!

omdb/src/bin/omdb/sled.rs
smklein: Updated, specifying that these are "running control plane zones". I think we could **also** expose the set of zones the sled was **told** to run, but this will require modifying the sled agent API.

omdb/src/bin/omdb/sled.rs
smklein: Updated

PR #4133, opened 2023-09-21
omdb could be easier to configure

davepacheco:
Before this change:

- to use the `nexus` command you need to provide OMDB_NEXUS_URL in the environment or `--nexus-internal-url` on the command line
- to use the `db` command you need to provide OMDB_DB_URL in the environment or `--db-url` on the command line
- to use the `sled-agent` command you need to provide OMDB_SLED_AGENT_URL in the environment or `--sled-agent-url` on the command line

The database one is particularly annoying because it has pieces besides the IP address and port that are easy to forget or get wrong: the database name and `sslmode=disable`.

After this change:

- You _can_ still do all of the things above.  This is useful when for whatever reason you want to control precisely which component omdb talks to.
- For the `sled-agent` command, you still _have_ to provide the URL as described above.
- For the `nexus` and `db` commands, instead of providing the URLs for those, you can instead provide OMDB_DNS_SERVER in the environment or `--dns-server` on the command line.  Assuming you haven't _also_ provided a Nexus or DB URL (which still take precedence), then `omdb` will consult the DNS server to find database instances and Nexus instances.  As a side benefit, when connecting to the database, it uses _all_ of the addresses it finds, which means it should work as long as any of them is up without the user having to figure out which instances are up.
- If you provide none of these, then `omdb` falls back to the DNS server for `fd00:1122:3344::/48`.  This is not technically correct, in that Omicron can be deployed using a different base subnet for the underlay network.  But in practice, this is the right value for all deployments that I know about (including dogfood), so it allows the tool to work in most places with no configuration.

The upshot should be:
- In most real deployments, you don't need to configure _anything_ because `omdb` will consult the internal DNS server (which it already knows how to find) to find everything else.
- Even when one needs to override the DNS server, it should be more convenient to just have to specify OMDB_DNS_SERVER instead of a variety of other URLs.

Unrelated to all this, I bumped the built-in db query row limit from 100 to 500 records.  This value is somewhat arbitrary.  It's just there to avoid having to implement client pagination and to avoid us fetching a zillion records from the database at once. 100 was a little too small (@jgallagher ran into this listing DNS names in dogfood).  500 is enough so far.  A few thousand would probably also be fine.

### DIFF
--- a/dev-tools/omdb/tests/test_all_output.rs
+++ b/dev-tools/omdb/tests/test_all_output.rs
@@ -44,6 +44,9 @@ async fn test_omdb_usage_errors() {
         &["db", "services"],
         &["nexus"],
         &["nexus", "background-tasks"],
+        &["sled-agent"],
+        &["sled-agent", "zones"],
+        &["sled-agent", "zpools"],
     ];
 
     for args in invocations {
@@ -69,6 +72,9 @@ async fn test_omdb_success_cases(cptestctx: &ControlPlaneTestContext) {
         &["db", "sleds"],
         &["nexus", "background-tasks", "doc"],
         &["nexus", "background-tasks", "show"],
+        // We can't easily test the sled agent output because that's only
+        // provided by a real sled agent, which is not available in the
+        // ControlPlaneTestContext.
     ];
 
     for args in invocations {
@@ -102,6 +108,7 @@ async fn test_omdb_env_settings(cptestctx: &ControlPlaneTestContext) {
     let postgres_url = cptestctx.database.listen_url().to_string();
     let nexus_internal_url =
         format!("http://{}", cptestctx.internal_client.bind_address);
+    let dns_sockaddr = *cptestctx.internal_dns.dns_server.local_address();
     let mut output = String::new();
 
     // Database URL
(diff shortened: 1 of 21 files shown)


dev-tools/omdb/tests/test_all_output.rs
```diff
@@ -38,12 +38,14 @@ async fn test_omdb_usage_errors() {
         &["db"],
         &["db", "--help"],
         &["db", "dns"],
-        &["db", "dns", "show"],
```
davepacheco, line 41: I removed this test because the point was to show that we produced a useful error when you haven't provided any URLs.  But that's not an error any more.

internal-dns/src/resolver.rs
```diff
@@ -55,15 +55,15 @@ impl Resolver {
...
+        dns_addrs: &[SocketAddr],
     ) -> Result<Self, ResolveError> {
         info!(log, "new DNS resolver"; "addresses" => ?dns_addrs);
 
         let mut rc = ResolverConfig::new();
         let dns_server_count = dns_addrs.len();
         for socket_addr in dns_addrs.into_iter() {
             rc.add_name_server(NameServerConfig {
-                socket_addr,
+                socket_addr: *socket_addr,
```
jgallagher, line 66: This is a really silly nit, but if you find this unsightly, I think you can do

```rust
for &socket_addr in dns_addrs.into_iter() {
```

to avoid needing the deref here (since `SocketAddr` impls `Copy`).

internal-dns/src/resolver.rs
```diff
@@ -802,7 +790,7 @@ mod test {
         let dns_server = DnsServer::create(&logctx.log).await;
         let resolver = Resolver::new_from_addrs(
             logctx.log.clone(),
-            vec![*dns_server.dns_server.local_address()],
+            &[*dns_server.dns_server.local_address()],
```
jgallagher, line 793: Similar nit here - I think `local_address()` should probably return `SocketAddr` instead of `&SocketAddr`

jgallagher (approved):
This looks great, and I'm very excited to not manually find DB URLs. Thanks!

PR #4134, opened 2023-09-21
Created lookup for zpool_id and added to omdb

leftwo:
Updated the db-queries to support the zpool_id lookup.

Added a disks subcommand to the omdb db command, show information about disks in the database

Now @davepacheco , I know you said you wanted `get_disk()` and `get_instance()` in `omdb` directly.
but, do you really?  
I started to do that, but found I would need to import a bunch more stuff into omdb, which 
I can do, but also that `self.pool()` is not a pub, so I can't call it inside `omdb`.

Now, none of these issue are insurmountable, but I thought I would bring them to your
attention here and we can discuss further.

### DIFF
--- a/dev-tools/omdb/src/bin/omdb/db.rs
+++ b/dev-tools/omdb/src/bin/omdb/db.rs
@@ -23,10 +23,12 @@ use clap::ValueEnum;
 use diesel::expression::SelectableHelper;
 use diesel::query_dsl::QueryDsl;
 use diesel::ExpressionMethods;
+use nexus_db_model::Disk;
 use nexus_db_model::DnsGroup;
 use nexus_db_model::DnsName;
 use nexus_db_model::DnsVersion;
 use nexus_db_model::DnsZone;
+use nexus_db_model::Instance;
 use nexus_db_model::Sled;
 use nexus_db_queries::context::OpContext;
 use nexus_db_queries::db;
@@ -186,7 +188,7 @@ impl DbArgs {
                 command: DiskCommands::Info(uuid),
             }) => cmd_db_disk_info(&opctx, &datastore, uuid).await,
             DbCommands::Disks(DiskArgs { command: DiskCommands::List }) => {
-                cmd_db_disk_list(&opctx, &datastore, self.fetch_limit).await
+                cmd_db_disk_list(&datastore, self.fetch_limit).await
             }
             DbCommands::Dns(DnsArgs { command: DnsCommands::Show }) => {
                 cmd_db_dns_show(&opctx, &datastore, self.fetch_limit).await
@@ -300,7 +302,6 @@ fn first_page<'a, T>(limit: NonZeroU32) -> DataPageParams<'a, T> {
 
 /// Run `omdb db disk list`.
 async fn cmd_db_disk_list(
-    opctx: &OpContext,
     datastore: &DataStore,
     limit: NonZeroU32,
 ) -> Result<(), anyhow::Error> {
@@ -317,10 +318,14 @@ async fn cmd_db_disk_list(
     let mut rows = Vec::with_capacity(2);
     let ctx = || "listing disks".to_string();
 
-    let disks = datastore
-        .disk_list_all(opctx, &first_page(limit))
-        .await
-        .with_context(ctx)?;
+    use db::schema::disk::dsl;
+    let disks = dsl::disk
+            .filter(dsl::time_deleted.is_null())
+            .limit(i64::from(u32::from(limit)))
+            .select(Disk::as_select())
+            .load_async(datastore.pool_for_tests().await?)
+            .await
+            .context("loading disks")?;
 
     check_limit(&disks, limit, ctx);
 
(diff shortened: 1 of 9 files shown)


nexus/db-queries/src/db/datastore/disk.rs
```diff
@@ -151,6 +152,35 @@ impl DataStore {
...
+        use db::schema::disk::dsl as disk_dsl;
+
+        disk_dsl::disk
+            .filter(disk_dsl::id.eq(disk_uuid))
+            .select(Disk::as_select())
+            .first_async::<Disk>(self.pool())
+            .await
+            .map_err(|e| public_error_from_diesel_pool(e, ErrorHandler::Server))
+    }
+
```
davepacheco, line 183: Yeah, so the reason _not_ to do this is that it's exactly what `LookupPath` is _for_ except that LookupPath does an authz check.  I'm afraid by having this here, it would be very easy for people to use this function in Nexus and skip the authz check by accident, leaving a gaping security hole.

omdb's db.rs already imports a bunch of stuff to do db queries directly (e.g., for dns stuff).  Is it a lot more for disks?

If that's really a problem, we could modify `LookupPath` to add a `skip_authz()` option.  At least then people would have to go explicitly out of their way to skip authz.

nexus/db-queries/src/db/datastore/instance.rs
```diff
@@ -140,6 +140,21 @@ impl DataStore {
         .map_err(|e| public_error_from_diesel_pool(e, ErrorHandler::Server))
     }
 
+    /// Fetches the instance with the given UUID
+    pub async fn instance_get(
```
davepacheco, line 144: Same as with disks -- the lack of authz check here seems like a pretty big footgun (even if it's appropriate for omdb).

nexus/db-queries/src/db/datastore/disk.rs
leftwo: footgun removed from nexus, placed in omdb!

leftwo:
So, I fixed the test of `omdb` output, but this one I don't yet understand:

```
--- STDOUT:              nexus-db-queries authz::policy_test::test_iam_roles_behavior ---

running 1 test
creating user: fleet-admin
creating user: fleet-collaborator
creating user: fleet-viewer
creating user: silo1-admin
creating user: silo1-collaborator
creating user: silo1-viewer
creating user: silo1-proj1-admin
creating user: silo1-proj1-collaborator
creating user: silo1-proj1-viewer
test authz::policy_test::test_iam_roles_behavior ... FAILED

failures:

failures:
    authz::policy_test::test_iam_roles_behavior

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 147 filtered out; finished in 2.57s


--- STDERR:              nexus-db-queries authz::policy_test::test_iam_roles_behavior ---
log file: /tmp/nexus_db_queries-bb3b1f3e7e3c4490-test_iam_roles.6206.0.log
note: configured to log to "/tmp/nexus_db_queries-bb3b1f3e7e3c4490-test_iam_roles.6206.0.log"
thread 'authz::policy_test::test_iam_roles_behavior' panicked at 'these classes were not covered by the IAM role policy test: Zpool', nexus/db-queries/src/authz/policy_test/coverage.rs:88:13
WARN: dropped CockroachInstance without cleaning it up first (there may still be a child process running and a temporary directory leaked)
WARN: temporary directory leaked: /tmp/.tmprgbo26

------------
     Summary [   2.610s] 1 test run: 0 passed, 1 failed, 901 skipped
        FAIL [   2.609s] nexus-db-queries authz::policy_test::test_iam_roles_behavior
error: test run failed
```

It looks like the new `Zpool` needs a test, but I'm not sure where that test should be.

nexus/db-queries/src/db/datastore/instance.rs
leftwo: This should be addressed now.

dev-tools/omdb/src/bin/omdb/db.rs
```diff
@@ -266,6 +298,191 @@ fn first_page<'a, T>(limit: NonZeroU32) -> DataPageParams<'a, T> {
...
+        .filter(dsl::time_deleted.is_null())
+        .limit(i64::from(u32::from(limit)))
+        .select(Disk::as_select())
+        .load_async(datastore.pool_for_tests().await?)
+        .await
+        .context("loading disks")?;
+
+    check_limit(&disks, limit, ctx);
+
+    rows.extend(disks.into_iter().map(|disk| DiskRow {
```
davepacheco, line 332: nit: you could just `let rows = disks.into_iter()...` at this point, instead of creating a `mut rows` above and extending it here.

dev-tools/omdb/src/bin/omdb/db.rs
```diff
@@ -266,6 +298,191 @@ fn first_page<'a, T>(limit: NonZeroU32) -> DataPageParams<'a, T> {
...
+
+        let usr = UpstairsRow {
+            host_serial: my_sled.serial_number().to_string(),
+            disk_name: disk.name().to_string(),
+            instance_name,
+            propolis_zone: format!("oxz_propolis-server_{}", propolis_id),
+        };
+        rows.push(usr);
+    } else {
+        // The instance is not attached to anything, just print empty
```
davepacheco, line 430:
```suggestion
        // If the disk is not attached to anything, just print empty
```

I think?

davepacheco (approved):

dev-tools/omdb/src/bin/omdb/db.rs
leftwo: Done

dev-tools/omdb/src/bin/omdb/db.rs
leftwo: You think correctly.

PR #4139, opened 2023-09-25
nexus: don't require sled ID to delete v2p mappings

gjcolombo:
Nexus's `delete_instance_v2p_mappings` fetches the target instance's DB record to get the instance's current sled ID. This allows this function to (a) skip the sled when deleting V2P mappings, and (b) obtain the sled's physical host IP to pass into a `SetVirtualNetworkInterfaceHost` parameter block to pass to sled agent.

Change `delete_instance_v2p_mappings` not to need the sled ID for either of these purposes:

1. It's already safe to ask a sled to delete a V2P mapping that its XDE driver has already deleted. (Currently, this is true because explicitly asking to destroy a V2P mapping is actually a no-op. But even if it weren't, Nexus already assumes this request is idempotent, since it can be invoked from the instance deletion saga.)
2. The physical host IP is not necessary for the `oxide-vpc` lib to identify a V2P mapping for removal; only the VNI and virtual IP are needed (see `VpcMappings::del` in the OPTE repo).

This helps to clear a path for upcoming changes that will make an instance's sled ID optional (i.e. it is possible for an instance not to have one, or for the last-known location of a running instance to be more challenging to reason about than this routine presupposes).

Tested: cargo test, including integration test updates for tests that are affected by this behavior; ran an instance on a dev cluster, verified it was externally accessible, verified the expected sled agent calls were seen when stopping/deleting the instance, and verified a new instance that reused the previous instance's external IP was also reachable.

### DIFF
--- a/illumos-utils/src/opte/params.rs
+++ b/illumos-utils/src/opte/params.rs
@@ -38,3 +38,15 @@ pub struct SetVirtualNetworkInterfaceHost {
     pub physical_host_ip: Ipv6Addr,
     pub vni: external::Vni,
 }
+
+/// The data needed to identify a virtual IP for which a sled maintains an OPTE
+/// virtual-to-physical mapping such that that mapping can be deleted.
+#[derive(Clone, Debug, Serialize, Deserialize, JsonSchema, PartialEq)]
+pub struct DeleteVirtualNetworkInterfaceHost {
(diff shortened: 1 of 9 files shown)


bnaecker (approved):

gjcolombo:
Thanks for the review!

Issue #4132, opened 2023-09-21
[nexus] Access DB via "connection", not "pool"


smklein:
This is a gnarly implementation detail that arose out of a poor decision I made back in async-bb8-diesel.

In async-bb8-diesel, we define a trait named `AsyncConnection`, which allows callers to execute a variety of asynchronous Diesel operations. This trait is implemented for **both** the connection pool itself, as well as individual connections.

This was intended for developer convenience -- if you wanted to use a "pool" object directly to issue operations, you could. If you wanted to use a "connection" object to issue operations, you could check it out of the pool and use that too.

This approach has a couple of downsides:

- Generics. This hits build times, causes more codegen, etc. We should use only one implementation.
- Worse errors. Since acting on a "pool" can return a PoolError (e.g., did you fail to check out a connection?) in addition to the diesel errors (e.g. did your query fail?) it forces callers to **also** be generic over the type of error being used.

This can culminate in code like the following:

https://github.com/oxidecomputer/omicron/blob/46ffb37edf51f86fa7b0d46bd0c3a834eb213a75/nexus/db-queries/src/db/datastore/service.rs#L47-L55

If we universally just "checked out connections before using them", that code could be written as:

```rust
    pub(crate) async fn service_upsert_conn(
        &self,
        conn: ConcreteConnectionType,
        service: Service,
    ) -> CreateResult<Service> {
```

(and hey, look, that has no generics, will be faster to compile, and will be easier to reason about!)

PR #4140, opened 2023-09-25
[db] Access DB via connections, not via pool

smklein:
- Accesses the DB via connections explicitly, rather than using a pool
- This change reduces generics slightly, and simplifies error handling. Callers can deal with "pool errors" when checking out the connection, and can deal with "query errors" separately when issuing the queries themselves.

Depends on https://github.com/oxidecomputer/async-bb8-diesel/pull/52
Fixes https://github.com/oxidecomputer/omicron/issues/4132


Cargo.toml
```diff
@@ -133,7 +133,8 @@ api_identity = { path = "api_identity" }
 approx = "0.5.1"
 assert_matches = "1.5.0"
 assert_cmd = "2.0.12"
-async-bb8-diesel = { git = "https://github.com/oxidecomputer/async-bb8-diesel", rev = "be3d9bce50051d8c0e0c06078e8066cc27db3001" }
+# TODO: Update with a revision before merging
```
smklein, line 136: Note to self: 

- Merge https://github.com/oxidecomputer/async-bb8-diesel/pull/52
- Then update this rev

nexus/db-queries/src/db/datastore/mod.rs
```diff
@@ -200,29 +200,53 @@ impl DataStore {
...
-    // which gives us the chance to do a minimal security check before hitting
-    // the database.  Eventually, this function should only be used for doing
-    // authentication in the first place (since we can't do an authz check in
-    // that case).
-    fn pool(&self) -> &bb8::Pool<ConnectionManager<DbConnection>> {
-        self.pool.pool()
-    }
-
-    pub(super) async fn pool_authorized(
+    async fn pool_authorized(
```
smklein, line 203: This is sorta the meat-and-potatoes of this PR:

- We no longer expose the raw pool to most of Nexus
- Instead, we encourage callers to access a connection, via one of the `pool_connection...` functions below
- Queries can be issued against individual connections

This is nice, because we **must** act on connections from a transaction context anyway, so doing this unifies the way we issue queries -- regardless of whether you're writing code within a transaction, you'll always be acting on a single connection now, and you'll only need to deal with connection pool errors when acquiring the connection.

nexus/db-queries/src/db/datastore/mod.rs
```diff
@@ -200,29 +200,53 @@ impl DataStore {
...
+    pub(super) async fn pool_connection_authorized(
+        &self,
+        opctx: &OpContext,
+    ) -> Result<bb8::PooledConnection<ConnectionManager<DbConnection>>, Error>
+    {
+        let pool = self.pool_authorized(opctx).await?;
+        let connection = pool.get().await.map_err(|err| {
+            Error::internal_error(&format!(
+                "Failed to access DB connection: {err}"
+            ))
```
davepacheco, line 221: This is probably not an internal error, but rather a 503.  I think we had logic before in db/error.rs to figure this out?

nexus/db-queries/src/db/datastore/mod.rs
```diff
@@ -200,29 +200,53 @@ impl DataStore {
...
+    /// TODO-security: This should be deprecated in favor of
+    /// "pool_connection_authorized".
+    pub(super) async fn pool_connection_unauthorized(
+        &self,
+    ) -> Result<bb8::PooledConnection<ConnectionManager<DbConnection>>, Error>
+    {
+        let connection = self.pool.pool().get().await.map_err(|err| {
+            Error::internal_error(&format!(
+                "Failed to access DB connection: {err}"
+            ))
```
davepacheco, line 238: Same as above.

nexus/db-queries/src/db/datastore/mod.rs
```diff
@@ -200,29 +200,53 @@ impl DataStore {
...
+        })?;
+        Ok(connection)
+    }
+
+    /// Returns an unauthorized connection to a connection from the database
+    /// connection pool.
+    ///
+    /// TODO-security: This should be deprecated in favor of
+    /// "pool_connection_authorized".
+    pub(super) async fn pool_connection_unauthorized(
```
davepacheco, line 231: Is this a replacement for the old `pool()` function?

davepacheco (commented):
I agree this is a cleaner interface.  I reviewed the datastore/mod.rs, error.rs, and a few other areas, but I did not review most of nexus/db-queries/src/db because there's so much delta here that's mechanical but non-trivial.

nexus/db-queries/src/db/datastore/mod.rs
smklein: Good catch, I think the populator tests caught this too. Updated.

davepacheco (approved):
Thanks for doing this!

smklein:
I ran the following command:

```bash
cargo build -p omicron-nexus && touch nexus/src/lib.rs  && cargo build -p omicron-nexus --timings
```

(AKA, do a clean build, touch one file, rebuild validating timings).

This is what I saw on my Linux laptop on main:

image

This is what I saw with this PR:

image

I'm surprised by this result, and think I might need to do a deep dive into why this is happening.

smklein:
Huh, well, that's... interesting. I tried using https://blog.rust-lang.org/inside-rust/2020/02/25/intro-rustc-self-profile.html to dig into "why did this make the build slower", and here's what I found:

I basically wanted to run `cargo rustc -- -Zself-profile -Zself-profile-events=default,args` , but doing to caused rustc to ICE, so I just queries the `default` events, and things seemed to work. Note: Running these commands required nightly, and I used:

```bash
 $ rustup run nightly rustc --version
rustc 1.74.0-nightly (e7c502d93 2023-09-27)
```

I ran the following command:

```bash
cargo +nightly rustc -p omicron-nexus --lib && touch nexus/src/lib.rs && cargo +nightly rustc -p omicron-nexus --lib -- -Zself-profile -Zself-profile-events=default
```

(just like before -- build, touch one file, rebuild)

I ran this on both `main` and this branch. Comparing these results, I **was not able to reproduce this PR building slower** -- it was actually a small improvement over `main`!

Since I was forced to use the nightly compiler, I figured I'd redo the `build --timings` comparison too. Here are the results, **using the rustc 1.74.0-nightly (e7c502d93 2023-09-27) toolchain**:

### main
image

### This PR
image

## My takeaway

I think this PR is safe to land? It seems that the issues causing build speeds to regress are more related to our rust version than the commit itself.

Cargo.toml
smklein: Done

davepacheco:
In case it's helpful for anybody else, I had a branch with changes that needed to be updated after merging with this change and this worked to update my branch:

- replace calls to `public_error_from_diesel_pool()` with `public_error_from_diesel()`
- replace calls to `pool_authorized()` with calls to `pool_connection_authorized()`
- replace use of `TransactionError::Pool` with `TransactionError::Connection`

PR #4145, opened 2023-09-26
Add an SMF property for rack_id to MGS

andrewjstone:
A follow up commit will plumb the rack_id along with any discovered sleds to a new "hardware_sled" table in CRDB used for sled addition to an already initialized rack. This is going to be required when adding sleds to initialized racks, in order to create a `StartSledAgentRequest` for new sleds.


andrewjstone:
I still need to test this locally.

gateway/src/lib.rs
```diff
@@ -283,6 +285,24 @@ impl Server {
...
+    pub fn set_rack_id(&self, rack_id: Option<Uuid>) {
+        if let Some(rack_id) = rack_id {
+            let val = self.apictx.rack_id.get_or_init(|| rack_id);
+            if *val != rack_id {
+                error!(
+                    self.apictx.log,
+                    "Rack ID changed from {} to {}", val, rack_id
+                );
+            } else {
+                info!(self.apictx.log, "Set rack_id: {rack_id} for MGS");
```
jgallagher, line 299: Tiny nit - this log is just for MGS, so I think we can drop that bit. I might also shove the rack ID into a slog key/value? I haven't been consistent with this but I feel like it's a little cleaner; something like

```suggestion
                info!(self.apictx.log, "Set rack_id"; "rack_id" => %rack_id);
```

gateway/src/lib.rs
```diff
@@ -283,6 +285,24 @@ impl Server {
...
 
+    /// The rack_id will be set on a refresh of the SMF property when the sled
+    /// agent starts.
+    pub fn set_rack_id(&self, rack_id: Option<Uuid>) {
+        if let Some(rack_id) = rack_id {
+            let val = self.apictx.rack_id.get_or_init(|| rack_id);
+            if *val != rack_id {
+                error!(
+                    self.apictx.log,
+                    "Rack ID changed from {} to {}", val, rack_id
```
jgallagher, line 296: `get_or_init` doesn't change the existing value right? So maybe something like

```rust
error!(
    self.apictx.log,
    "Ignoring attempted change to rack ID";
    "current_rack_id" => %val,
    "ignored_new_rack_id" => %rack_id,
);
```

sled-agent/src/services.rs
```diff
@@ -2754,6 +2758,10 @@ impl ServiceManager {
                                 &format!("[{address}]:{MGS_PORT}"),
                             )?;
 
+                            if let Some(info) = self.inner.sled_info.get() {
```
jgallagher, line 2761: The fact that this is optional makes me slightly nervous - the underlay `address` is _not_ optional at this point, so at least per the types we could have an underlay addr but no rack ID. Is that actually possible in practice?

gateway/src/context.rs
```diff
@@ -6,29 +6,46 @@ use crate::error::StartupError;
...
         let mgmt_switch =
             ManagementSwitch::new(switch_config, &host_phase2_provider, log)
                 .await?;
 
+        let rack_id = OnceLock::new();
+        if let Some(id) = rack_id_config {
+            info!(log, "Setting rack_id: {id} for MGS");
+            match rack_id.set(id) {
+                Ok(()) => info!(log, "Setting rack_id: {id} for MGS"),
+                Err(existing) => error!(
```
jgallagher, line 38: Nit - I think we could avoid this error branch entirely via `OnceLock::from()`, which creates an already-initialized value; something like

```rust
let rack_id = if let Some(id) = rack_id_config {
    // ... logging ...
    OnceLock::from(id)
} else {
    OnceLock::new()
};

File: sled-agent/src/services.rs (hunk 3 above)
Comment (andrewjstone) [reply sled-agent/src/services.rs:2761]: I'm not sure if it's possible or not, but I suspect it is not possible in practice. The bootstrap server must have received a `StartSledAgentRequest` for it to start sled agent and learn about both the underlay network and the rack_id. It appears that the underlay address is only configured in [SledAgentInner](https://github.com/oxidecomputer/omicron/blob/main/sled-agent/src/bootstrap/server.rs#L289-L322) from the `StartSledAgentRequest`, and that this happens after [`sled_agent_started` is called with the `rack_id `](https://github.com/oxidecomputer/omicron/blob/main/sled-agent/src/sled_agent.rs#L416-L422) from the same request. 

In the cold boot case, the underlay address is only touched after the `SledAgentRequest` is [loaded](https://github.com/oxidecomputer/omicron/blob/main/sled-agent/src/bootstrap/server.rs#L290-L338).

I'm not really sure what we should do here. I can certainly call `unwrap` to indicate a programmer error, or just log an error, which will in the end result in not being able to add sleds, but also not crash existing working code.

File: gateway/src/bin/mgs.rs [hunk 5]
```diff
@@ -120,6 +123,7 @@ async fn do_run() -> Result<(), CmdError> {
                                 .map_err(|err| CmdError::Failure(
                                     format!("config refresh failed: {err}")
                                 ))?;
+                            server.set_rack_id(new_config.rack_id);
```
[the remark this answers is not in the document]
andrewjstone: Good call! Fixed.

gateway/src/lib.rs
andrewjstone: Fixed.

gateway/src/context.rs
andrewjstone: Excellent, this is exactly what I was looking for earlier! Fixed.

sled-agent/src/services.rs
jgallagher: I don't think I'd `unwrap()`. I'd definitely leave a comment as a breadcrumb for any future cleanup work, and I'd be fine with or without an error/warn log.

andrewjstone: Added in [347b378](https://github.com/oxidecomputer/omicron/pull/4145/commits/347b378a8dbfb860f0ce7e183abafcd7af9fefdb)

andrewjstone:
I tested manually on my helios box, and this is working as expected:

```
[ Sep 27 19:25:12 Rereading configuration. ]
[ Sep 27 19:25:12 Executing refresh method (:kill). ]
Sep 27 19:25:12.009 INFO Set rack_id, rack_id: ee589242-dfe2-4864-9b2e-21ad6a37d473, name: 358ab69b-ecbc-4f4a-881c-ee5245e6982c, file: gateway/src/lib.rs:300
Sep 27 19:25:12.010 INFO adjusting dropshot addresses; keeping existing addr, addr: [::1]:12225, name: 358ab69b-ecbc-4f4a-881c-ee5245e6982c, file: gateway/src/lib.rs:248
Sep 27 19:25:12.010 INFO adjusting dropshot addresses; starting new server, addr: [fd00:1122:3344:101::2]:12225, name: 358ab69b-ecbc-4f4a-881c-ee5245e6982c, file: gateway/src/lib.rs:255
```

jgallagher (approved):

PR #4147, opened 2023-09-27
Omdb networking

rcgoodfellow:
Adds an `omdb` command to list external network IP addresses along with their parent.

```
./omdb db network list-eips
note: using database URL postgresql://root@[fd00:1122:3344:109::3]:32221,[fd00:1122:3344:105::3]:32221,[fd00:1122:3344:10b::3]:32221,[fd00:1122:3344:107::3]:32221,[fd00:1122:3344:108::3]:32221/omicron?sslmode=disable
note: database schema version matches expected (4.0.0)
 ip               ports        kind       owner
 172.20.26.1/32   0/65535      Floating   Service ExternalDns
 172.20.26.2/32   0/65535      Floating   Service ExternalDns
 172.20.26.3/32   0/65535      Floating   Service Nexus
 172.20.26.4/32   0/65535      Floating   Service Nexus
 172.20.26.5/32   0/65535      Floating   Service Nexus
 172.20.26.6/32   0/16383      SNat       Service Ntp
 172.20.26.7/32   16384/32767  SNat       Service Ntp
 172.20.26.11/32  49152/65535  SNat       Instance try/sysbench-mysql-3
 172.20.26.11/32  32768/49151  SNat       Instance try/sysbench-mysql-2
 172.20.26.11/32  16384/32767  SNat       Instance try/test-upgrade-2
 172.20.26.13/32  0/65535      Ephemeral  Instance try/test-upgrade-2
 172.20.26.14/32  0/65535      Ephemeral  Instance try/sysbench-mysql-2
 172.20.26.15/32  0/65535      Ephemeral  Instance try/sysbench-mysql-3
 172.20.26.16/32  0/16383      SNat       Instance try/sysbench-mysql-4
 172.20.26.16/32  32768/49151  SNat       Instance try/sysbench-mysql-6
 172.20.26.16/32  49152/65535  SNat       Instance try/sysbench-mysql-7
```

Also comes with a verbose mode to print raw datastore structures.

```
BRM44220005 # ./omdb db network --verbose list-eips
note: using database URL postgresql://root@[fd00:1122:3344:109::3]:32221,[fd00:1122:3344:105::3]:32221,[fd00:1122:3344:10b::3]:32221,[fd00:1122:3344:107::3]:32221,[fd00:1122:3344:108::3]:32221/omicron?sslmode=disable
note: database schema version matches expected (4.0.0)
ExternalIp {
    id: 01043656-d049-449a-9b33-1bad7f4affdd,
    name: None,
    description: None,
    time_created: 2023-08-30T20:27:57.649708Z,
    time_modified: 2023-08-30T20:27:57.649708Z,
    time_deleted: None,
    ip_pool_id: cadb8535-e32f-4be2-9259-e45dec9fa3c7,
    ip_pool_range_id: 73aede53-8fe2-40c5-a6fb-23675269e979,
    is_service: false,
    parent_id: Some(
        33e6d4a4-6cc7-440f-8a9f-152d94136191,
    ),
    kind: SNat,
    ip: V4(
        Ipv4Network {
            addr: 172.20.26.41,
            prefix: 32,
        },
    ),
    first_port: SqlU16(
        16384,
    ),
    last_port: SqlU16(
        32767,
    ),
}
ExternalIp {
    id: 015fd6de-a47a-4804-bc96-2b5880c42ac3,
    name: None,
    description: None,
    time_created: 2023-08-30T19:48:56.391829Z,
    time_modified: 2023-08-30T19:48:56.391829Z,
    time_deleted: None,
...
```

### DIFF
--- a/dev-tools/omdb/src/bin/omdb/db.rs
+++ b/dev-tools/omdb/src/bin/omdb/db.rs
@@ -1201,10 +1201,19 @@ async fn cmd_db_eips(
     for ip in &ips {
         let owner = if let Some(owner_id) = ip.parent_id {
             if ip.is_service {
-                let service = LookupPath::new(opctx, datastore)
+                let service = match LookupPath::new(opctx, datastore)
                     .service_id(owner_id)
                     .fetch()
-                    .await?;
+                    .await
+                {
+                    Ok(instance) => instance,
+                    Err(e) => {
+                        eprintln!(
+                            "error looking up service with id {owner_id}: {e}"
+                        );
+                        continue;
+                    }
+                };
                 Owner::Service { kind: format!("{:?}", service.1.kind) }
             } else {
                 use db::schema::instance::dsl as instance_dsl;
--- a/dev-tools/omdb/src/bin/omdb/db.rs
+++ b/dev-tools/omdb/src/bin/omdb/db.rs
@@ -1134,12 +1134,6 @@ async fn cmd_db_eips(
     limit: NonZeroU32,
     verbose: bool,
 ) -> Result<(), anyhow::Error> {
-    /*
-    let ips = datastore
-        .lookup_external_ips(&opctx)
-        .await
-        .context("listing external ips")?;
-    */
     use db::schema::external_ip::dsl;
     let ips: Vec<ExternalIp> = dsl::external_ip
         .filter(dsl::time_deleted.is_null())
--- a/dev-tools/omdb/src/bin/omdb/db.rs
+++ b/dev-tools/omdb/src/bin/omdb/db.rs
@@ -181,6 +181,7 @@ struct NetworkArgs {
     #[command(subcommand)]
     command: NetworkCommands,
 
+    /// Print out raw data structures from the data store.
     #[clap(long)]
     verbose: bool,
 }
@@ -1177,14 +1178,18 @@ async fn cmd_db_eips(
         owner: Owner,
     }
 
+    if verbose {
+        for ip in &ips {
+            if verbose {
+                println!("{ip:#?}");
+            }
+        }
+        return Ok(());
+    }
+
     let mut rows = Vec::new();
 
     for ip in &ips {
-        if verbose {
-            println!("{ip:#?}");
-            continue;
-        }
-
         let owner = if let Some(owner_id) = ip.parent_id {
             if ip.is_service {
                 let service = LookupPath::new(opctx, datastore)
@@ -1202,7 +1207,7 @@ async fn cmd_db_eips(
                     .await
                     .context("loading requested instance")?
                     .pop()
-                    .context("loading requested instance")?;
+                    .context("requested instance not found")?;
 
                 use db::schema::project::dsl as project_dsl;
                 let project = project_dsl::project
@@ -1213,7 +1218,7 @@ async fn cmd_db_eips(
                     .await
                     .context("loading requested project")?
                     .pop()
-                    .context("loading requested instance")?;
+                    .context("requested project not found")?;
 
                 Owner::Instance {
                     project: project.name().to_string(),
@@ -1236,14 +1241,12 @@ async fn cmd_db_eips(
         rows.push(row);
     }
 
-    if !verbose {
-        rows.sort_by(|a, b| a.ip.cmp(&b.ip));
-        let table = tabled::Table::new(rows)
-            .with(tabled::settings::Style::empty())
-            .to_string();
+    rows.sort_by(|a, b| a.ip.cmp(&b.ip));
+    let table = tabled::Table::new(rows)
+        .with(tabled::settings::Style::empty())
+        .to_string();
 
-        println!("{}", table);
-    }
+    println!("{}", table);
 
     Ok(())
 }
--- a/dev-tools/omdb/src/bin/omdb/db.rs
+++ b/dev-tools/omdb/src/bin/omdb/db.rs
@@ -1144,7 +1144,7 @@ async fn cmd_db_eips(
     let ips: Vec<ExternalIp> = dsl::external_ip
         .filter(dsl::time_deleted.is_null())
         .select(ExternalIp::as_select())
-        .get_results_async(datastore.pool_for_tests().await?)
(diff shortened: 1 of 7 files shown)


dev-tools/omdb/tests/successes.out
```diff
@@ -84,6 +84,24 @@ stderr:
 note: using database URL postgresql://root@[::1]:REDACTED_PORT/omicron?sslmode=disable
 note: database schema version matches expected (4.0.0)
 =============================================
+EXECUTING COMMAND: omdb ["db", "network"]
```
davepacheco, line 87: It seems like this output would go in the "usage" test (if you want to cover it).  The point of the "successes" test is to sanity-check that the command produces reasonable-looking output against a real system.  You might add `omdb db network list-eips` to this test instead?  Or if it's too annoying to get that to work (you might have to add more redactions for the output to be stable across runs), skip it altogether?

nexus/db-queries/src/db/datastore/external_ip.rs
```diff
@@ -289,4 +289,18 @@ impl DataStore {
...
+        opctx: &OpContext,
+    ) -> LookupResult<Vec<ExternalIp>> {
+        use db::schema::external_ip::dsl;
+        dsl::external_ip
+            .filter(dsl::time_deleted.is_null())
+            .select(ExternalIp::as_select())
+            .get_results_async(self.pool_authorized(opctx).await?)
+            .await
+            .map_err(|e| public_error_from_diesel_pool(e, ErrorHandler::Server))
+    }
```
davepacheco, line 305: I think it's pretty dangerous to add a non-authz-gated datastore operation like this.  If this is a generally useful function, I'd just do the authz check.  If we don't think we'd need this kind of functionality in Nexus, or if authz is a problem, I think it's fine to just put this code straight into omdb (without an authz check).

dev-tools/omdb/src/bin/omdb/db.rs
```diff
@@ -933,6 +962,130 @@ async fn cmd_db_dns_names(
...
+                use db::schema::instance::dsl as instance_dsl;
+                let instance = instance_dsl::instance
+                    .filter(instance_dsl::id.eq(owner_id))
+                    .limit(1)
+                    .select(Instance::as_select())
+                    .load_async(datastore.pool_for_tests().await?)
+                    .await
+                    .context("loading requested instance")?
+                    .pop()
+                    .context("requested instance not found")?;
```
davepacheco, line 1045: Not critical but I'd consider continuing rather than bailing out for this case (so that if there's an errant "owner" in the database, we can still use the tool to see as much as we can).

dev-tools/omdb/src/bin/omdb/db.rs
```diff
@@ -933,6 +962,130 @@ async fn cmd_db_dns_names(
...
+                use db::schema::project::dsl as project_dsl;
+                let project = project_dsl::project
+                    .filter(project_dsl::id.eq(instance.project_id))
+                    .limit(1)
+                    .select(Project::as_select())
+                    .load_async(datastore.pool_for_tests().await?)
+                    .await
+                    .context("loading requested project")?
+                    .pop()
+                    .context("requested project not found")?;
```
davepacheco, line 1056: Same -- I'd considering continuing instead of bailing out here (but definitely not a blocker).

dev-tools/omdb/src/bin/omdb/db.rs
```diff
@@ -933,6 +962,130 @@ async fn cmd_db_dns_names(
...
+
+    impl Display for PortRange {
+        fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+            write!(f, "{}/{}", self.first, self.last)
+        }
+    }
+
+    #[derive(Tabled)]
+    enum Owner {
+        Instance { project: String, name: String },
```
davepacheco, line 991: I always wonder if the tool should report ids (which are immutable and unique across silos) or names like this that potentially save someone a separate lookup (or both, but that uses up valuable space too).  It's all tradeoffs -- just mentioning it to see what you think.

davepacheco:
Thanks for doing this!  This will be very useful.

dev-tools/omdb/src/bin/omdb/db.rs
rcgoodfellow: I explicitly chose names here - as my use case for this is mapping external IPs to something immediately tangible. I added an option for verbose output that provides the raw data structures from the DB that shows references based on UUID.

dev-tools/omdb/src/bin/omdb/db.rs
rcgoodfellow: Done

dev-tools/omdb/src/bin/omdb/db.rs
rcgoodfellow: Good call. Done

nexus/db-queries/src/db/datastore/external_ip.rs
rcgoodfellow: Sounds good. I just moved this into the omdb code.

dev-tools/omdb/tests/successes.out
rcgoodfellow: Moved this test to usage.

dev-tools/omdb/src/bin/omdb/db.rs
```diff
@@ -1098,6 +1128,162 @@ async fn cmd_db_dns_names(
...
+    datastore: &DataStore,
+    limit: NonZeroU32,
+    verbose: bool,
+) -> Result<(), anyhow::Error> {
+    /*
+    let ips = datastore
+        .lookup_external_ips(&opctx)
+        .await
+        .context("listing external ips")?;
+    */
```
davepacheco, line 1142: nit: probably want to remove this
I can't tell if this was here before because I can't access the incremental review.  (Maybe a force push?)

davepacheco (approved):
Just spotted one stray commented-out block but otherwise looks good!

davepacheco (approved):

PR #4150, opened 2023-09-27
[nexus-test-utils] replace build.rs with a new setup script

sunshowers:
Nextest now has support for setup scripts, which means that we can
replace the existing `build.rs` with a small binary.

I also took the liberty of switching some of the related code over to
camino.

Supersedes #4056.


nexus/test-utils/src/db.rs
```diff
@@ -16,8 +16,13 @@ use std::path::PathBuf;
...
 /// on the time spent performing this operation. Instead, we opt
 /// to copy the database from this seed location.
-fn seed_dir() -> PathBuf {
-    PathBuf::from(concat!(env!("OUT_DIR"), "/crdb-base"))
+fn seed_dir() -> Utf8PathBuf {
+    // The setup script should set this environment variable.
+    let seed_dir = std::env::var("CRDB_SEED_DIR")
+        .expect("CRDB_SEED_DIR missing -- are you running this test with `cargo nextest run`?");
+    // TODO: replace with temp dir written out by nextest
+    // Use "out/" for now.
```
smklein, line 24: Isn't this happening already?

crdb-seed/src/main.rs
```diff
@@ -0,0 +1,71 @@
...
+#[tokio::main]
+async fn main() {
+    // TODO: dropshot is v heavyweight for this, we should be able to pull in a
+    // smaller binary
+    let logctx = LogContext::new(
+        "crdb_seeding",
+        &ConfigLogging::StderrTerminal { level: ConfigLoggingLevel::Info },
+    );
+    let dir = ensure_seed_directory_exists(&logctx.log).await;
+    if let Ok(env_path) = std::env::var("NEXTEST_ENV") {
```
smklein, line 65: Out of curiosity, what sets this environment variable?

Also, if this environment variable cannot be found, should we throw an error?

crdb-seed/src/main.rs
```diff
@@ -0,0 +1,71 @@
...
+        .expect("Not a UTF-8 path")
+        .join("crdb-base");
+    std::fs::create_dir_all(&base_seed_dir).unwrap();
+    let desired_seed_dir = base_seed_dir.join(digest_unique_to_schema());
+
+    // If the directory didn't exist when we started, try to create it.
+    //
+    // Note that this may be executed concurrently by many tests, so
+    // we should consider it possible for another caller to create this
+    // seed directory before we finish setting it up ourselves.
```
smklein, line 32: Is this still true, or will nextest prevent the concurrent execution of this function?

smklein (approved):
Looks great, thanks for doing this!

crdb-seed/src/main.rs
sunshowers: > Out of curiosity, what sets this environment variable?

Nextest does -- it's the way setup scripts can communicate with nextest.

> Also, if this environment variable cannot be found, should we throw an error?

I figured folks might want to run the script by hand. Maybe we should warn in that case.

nexus/test-utils/src/db.rs
sunshowers: > Isn't this happening already?

Yes, lol. Will fix.

crdb-seed/src/main.rs
sunshowers: Nextest will execute it once, but it's possible that the user spins up two instances of nextest at the same time. I'll update the comment.

crdb-seed/src/main.rs
smklein: Gotcha. So someone could run this binary manually, and it would create the seed dir, but not pass the necessary info to nextest.

That makes sense -- I'm okay with this structure as-is, I just wanted to make sure it would be clear "where things went wrong" if the environment variable from nextest was not provided.

sunshowers: I added some more logging and a warning message to make this clearer.

jordanhendricks (approved):
This is great! Thanks!

PR #4152, opened 2023-09-27
[schema] Add more strict CRDB index comparison

### LINKED ISSUE #4143: CRDB schema change tests not covering constraints on indices sufficiently
See https://github.com/oxidecomputer/omicron/pull/4141 for context -- I haven't dug into the root issue here, but the minimal repro case is:

dbinit.sql:
```sql
CREATE INDEX IF NOT EXISTS lookup_zpool_by_disk on omicron.public.zpool (        
    physical_disk_id                                                             
) WHERE physical_disk_id IS NOT NULL AND time_deleted IS NULL;
```

up1.sql:
```sql
CREATE INDEX IF NOT EXISTS lookup_zpool_by_disk ON omicron.public.zpool (physical_disk_id, id) WHERE time_deleted IS NULL; 
```

These are not the same, but appear to be passing the `dbinit_equals_sum_of_all_up` test within https://github.com/oxidecomputer/omicron/blob/9a1f6bc5da1efed13dabbe14ec5066002cfd0851/nexus/tests/integration_tests/schema.rs .

My suspicion: More tables from https://www.cockroachlabs.com/docs/stable/information-schema need to be compared.

smklein:
Adds more schema comparison for DB schemas, and adds a couple regression tests for the specific index mismatch.

Fixes #4143

### DIFF
--- a/nexus/tests/integration_tests/schema.rs
+++ b/nexus/tests/integration_tests/schema.rs
@@ -15,7 +15,7 @@ use omicron_common::api::internal::shared::SwitchLocation;
 use omicron_common::nexus_config::Config;
 use omicron_common::nexus_config::SchemaConfig;
 use omicron_test_utils::dev::db::CockroachInstance;
-use pretty_assertions::assert_eq;
+use pretty_assertions::{assert_eq, assert_ne};
 use similar_asserts;
 use slog::Logger;
 use std::collections::{BTreeMap, BTreeSet};
@@ -132,6 +132,7 @@ enum AnySqlType {
     String(String),
     Bool(bool),
     Uuid(Uuid),
+    Int8(i64),
     // TODO: This isn't exhaustive, feel free to add more.
     //
     // These should only be necessary for rows where the database schema changes also choose to
@@ -166,6 +167,9 @@ impl<'a> tokio_postgres::types::FromSql<'a> for AnySqlType {
         }
         if Uuid::accepts(ty) {
             return Ok(AnySqlType::Uuid(Uuid::from_sql(ty, raw)?));
+        }
+        if i64::accepts(ty) {
+            return Ok(AnySqlType::Int8(i64::from_sql(ty, raw)?));
         }
         Err(anyhow::anyhow!(
             "Cannot parse type {ty}. If you're trying to use this type in a table which is populated \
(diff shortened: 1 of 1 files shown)


nexus/tests/integration_tests/schema.rs
```diff
@@ -432,6 +436,16 @@ const CHECK_CONSTRAINTS: [&'static str; 4] = [
     "check_clause",
 ];
 
+const CONSTRAINT_COLUMN_USAGE: [&'static str; 7] = [
```
smklein, line 439: Note: This wasn't used by the regression tests, but I figured I'd add it as "yet another check to validate that the DB schemas match".

See: https://www.cockroachlabs.com/docs/stable/information-schema#constraint_column_usage

nexus/tests/integration_tests/schema.rs
```diff
@@ -456,29 +470,50 @@ const REFERENTIAL_CONSTRAINTS: [&'static str; 8] = [
...
     "table_schema",
     "table_name",
     "non_unique",
     "index_schema",
     "index_name",
+    "seq_in_index",
     "column_name",
     "direction",
+    "storing",
+    "implicit",
```
smklein, line 484: "implicit" was the one we ended up using for the "column comparison" regression tests, but the others also seem useful to have.

nexus/tests/integration_tests/schema.rs
```diff
@@ -456,29 +470,50 @@ const REFERENTIAL_CONSTRAINTS: [&'static str; 8] = [
...
 ];
 
+const PG_INDEXES: [&'static str; 5] =
+    ["schemaname", "tablename", "indexname", "tablespace", "indexdef"];
```
smklein, line 488: Note: These are not actually from "InformationSchema", but rather, from https://www.cockroachlabs.com/docs/stable/pg-catalog .

This comparison was necessary to catch the differing "WHERE" clauses.

nexus/tests/integration_tests/schema.rs
```diff
@@ -456,29 +470,50 @@ const REFERENTIAL_CONSTRAINTS: [&'static str; 8] = [
...
+    "constraint_catalog",
+    "constraint_schema",
+    "constraint_name",
+    "table_catalog",
+    "table_schema",
+    "table_name",
+    "constraint_type",
+    "is_deferrable",
+    "initially_deferred",
+];
```
smklein, line 503: Similar to `CONSTRAINT_COLUMN_USAGE`, this wasn't necessary here, but it seems useful to ensure that it doesn't get out-of-sync regardless.

https://www.cockroachlabs.com/docs/stable/information-schema#table_constraints

smklein:
Added https://github.com/oxidecomputer/omicron/pull/4153 as a follow-up PR for more things that I wanted to test, but which were unrelated to the original issue.

leftwo (approved):
I'm not up to snuff on the precise details of how the database works, but I can follow
the flow here and for sure the specific case I had in 4143 would be caught here and
the test case confirms that.

PR #4154, opened 2023-09-27
[meta] add a workspace-hack package managed by hakari

sunshowers:
Similar to https://github.com/oxidecomputer/crucible/pull/956, this adds
a workspace-hack package managed by cargo-hakari. This should massively
cut down on incremental build times.

`cargo hakari` doesn't yet support writing out `workspace-hack`
dependency lines using workspace inheritance
(https://github.com/guppy-rs/guppy/issues/7), so skip over it in the
lint.

Reverting this change can be done by following the instructions at https://docs.rs/cargo-hakari/latest/cargo_hakari/#disabling-and-uninstalling.


smklein (approved):
I'm super on-board with this -- the ability to switch between building different packages quickly seems like a huge benefit.

Before merging, I do want to make sure @davepacheco gets a chance to weigh in - I heard you had some prior objections in this space, and I'm interested what they are (and how strongly you feel about them relative to build speed improvements).

davepacheco:
I think my previous objections were the stuff mentioned here:
https://docs.rs/cargo-hakari/0.9.27/cargo_hakari/about/index.html#drawbacks (thanks for writing that!)
and particularly the bit about losing track of which workspace packages require which features.  But after years of experience with workspace feature resolution issues, if this goes a ways towards addressing that, that'd be a big net win.

sunshowers:
Thanks. I'll land this since there are no conflicts.

Just wanted to reiterate that this change doesn't commit us to the future and can easily be reverted if we run into issues: https://docs.rs/cargo-hakari/latest/cargo_hakari/#disabling-and-uninstalling (though we can also look at fixing any issues we run into, of course).

PR #4163, opened 2023-09-28
[update-engine] fix buffer tests

sunshowers:
Apparently I'd made a couple of mistakes while writing tests:

* I was adding all events a second time by accident, which was hiding
  the fact that...
* A couple not signs were flipped, whoops.


sunshowers:
Err, need to make a change still.

andrewjstone (approved):

PR #4164, opened 2023-09-28
[update-engine] a few improvements to event buffers

sunshowers:
* For failures, track the parent key that failed (similar to aborts).
* Track the last root event index that causes data for a step to be
  updated -- we're going to use this in the line-based displayer.
* Add tests.

Depends on #4163.


update-engine/src/buffer.rs
```diff
@@ -744,47 +784,83 @@ impl<'buf, S: StepSpec> EventBufferSteps<'buf, S> {
 pub struct EventBufferStepData<S: StepSpec> {
     step_info: StepInfo<NestedSpec>,
     sort_key: StepSortKey,
+    // XXX: nest_level and total_steps are common to each execution, but are
```
andrewjstone, line 787: I think ideally I would tie them together if they always exist together and get mutated together, but I wouldn't hold the PR for that.

update-engine/src/buffer.rs
```diff
@@ -1832,6 +1993,78 @@ mod tests {
...
+                    data_index == root_event_index,
+                    "last_root_event_index should have been updated \
+                     but wasn't (actual: {data_index}, expected: {root_event_index}) \
+                     for step {step_key:?} (event: {event:?})",
+                );
+            } else {
+                ensure!(
+                    data_index < root_event_index,
+                    "last_root_event_index should *not* have been updated \
+                     but wasn't (current: {data_index}, new: {root_event_index}) \
```
andrewjstone, line 2016:
```suggestion
                     but was (current: {data_index}, new: {root_event_index}) \
```

andrewjstone (approved):

update-engine/src/buffer.rs
sunshowers: They always exist together but are immutable properties.

Issue #4165, opened 2023-09-29
CI `package` job fails due to transient buildomat errors


davepacheco:
Here's an example job:
https://buildomat.eng.oxide.computer/wg/0/details/01HBF0GRNYXR6RMJPSX95W8XMQ/A1WlUdHz8EzlmcZaxISu2UH2EYg66WKMc4cO1GbM1Dv04dRF/01HBF0K57K97G3Q1EHYFRGERWZ

I am not positive but looking at .github/buildomat/jobs/package.sh, we're running `ptime -m cargo run --locked --release --bin omicron-package -- -t test package`, we just finished the cargo build step, and `omicron-package` is emitting this error:

```


2087 | 2023-09-28T23:43:45.735Z | Error: failed reading response from https://buildomat.eng.oxide.computer/public/file/oxidecomputer/crucible/image/aeb69dda26c7e1a8b6eada425670cd4b83f91c07/crucible-pantry.tar.gz
-- | -- | --
2088 | 2023-09-28T23:43:45.932Z |  
2089 | 2023-09-28T23:43:46.303Z | Caused by:
2090 | 2023-09-28T23:43:47.327Z | 0: request or response body error: error reading a body from connection: end of file before message length reached
2091 | 2023-09-28T23:43:47.592Z | 1: error reading a body from connection: end of file before message length reached
2092 | 2023-09-28T23:43:47.869Z | 2: end of file before message length reached
```

I gather from @jclulow that the expectation of buildomat is like that of other cloud services: it may fail transiently at any time.  It seems like `omicron-package` needs to retry downloads of prebuilt artifacts?  @smklein does that sound right?

smklein:
Fix is up at https://github.com/oxidecomputer/omicron/pull/4168 , the retry behavior can be tweaked via the CLI.

PR #4168, opened 2023-09-29
[omicron-package] Retry failed downloads automatically

smklein:
- Automatically retries downloads, with a small amount of constant backoff
- Provides config options for users to customize these retry attempts / backoffs
- Extends progress bars to show when download failures occur

Tested manually, by running:

```bash
cargo run --bin=omicron-package -- --retry-count 5 package
```

Then cycling my machine's network connectivity. I observed the "Failed to download prebuilt messages", saw the attempts ticking down, and then reconnected. Once connectivity was restored, the download succeeded.

Fixes https://github.com/oxidecomputer/omicron/issues/4165

### DIFF
--- a/package/src/bin/omicron-package.rs
+++ b/package/src/bin/omicron-package.rs
@@ -41,6 +41,11 @@ enum SubCommand {
     Deploy(DeployCommand),
 }
 
+fn parse_duration_ms(arg: &str) -> Result<std::time::Duration> {
+    let ms = arg.parse()?;
+    Ok(std::time::Duration::from_millis(ms))
+}
+
 #[derive(Debug, Parser)]
 #[clap(name = "packaging tool")]
 struct Args {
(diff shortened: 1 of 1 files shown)


package/src/bin/omicron-package.rs
```diff
@@ -303,8 +325,63 @@ async fn get_sha256_digest(path: &PathBuf) -> Result<Digest> {
     Ok(context.finish())
 }
 
+async fn download_prebuilt(
```
smklein, line 328: This is basically just refactored to make re-calling it easier

davepacheco (approved):
Thanks for doing this.

It'd be nice if we could use the existing general-purpose retry stuff but I get that the policies they use today aren't appropriate here.

I also wonder if we're going to want to resume downloads from where they left off.  (If the server is failing somewhat often, we might never be able to do a full download, but we would be able to make progress if we resumed partial downloads.)  But we can always add this if we continue to see problems.

package/src/bin/omicron-package.rs
```diff
@@ -303,8 +325,63 @@ async fn get_sha256_digest(path: &PathBuf) -> Result<Digest> {
...
+        .get(&url)
+        .send()
+        .await
+        .with_context(|| format!("failed to get {url}"))?;
+    progress.set_length(
+        response
+            .content_length()
+            .ok_or_else(|| anyhow!("Missing Content Length"))?,
+    );
+    let mut file = tokio::fs::File::create(&path)
```
davepacheco, line 353: I just noticed that we don't explicitly remove the old file.  Does this remove it, or at least truncate it?

smklein: It truncates it on the second attempt. If it helps, the hash won't match on partial downloads, so it would know that it needs to retry if we re-execute.

jgallagher:
> I also wonder if we're going to want to resume downloads from where they left off. (If the server is failing somewhat often, we might never be able to do a full download, but we would be able to make progress if we resumed partial downloads.) But we can always add this if we continue to see problems.

This will probably require adding support in buildomat (https://github.com/oxidecomputer/buildomat/issues/31).

Issue #4166, opened 2023-09-29
Make renovate work with cargo-hakari


sunshowers:
Useful links:

* https://github.com/renovatebot/github-action for self-hosted Renovate
* https://docs.renovatebot.com/configuration-options/#postupgradetasks where we can run hakari after upgrades and before commits
* https://docs.renovatebot.com/configuration-options/#ignorepaths

ahl:
It looks like dependabot is just broken full stop:

https://github.com/oxidecomputer/omicron/network/updates/9290049/jobs

It would be good to get dependency updates working again soon.

sunshowers:
Actively working on this.

ahl:
Previously we had another problem, but now dependabot is just timing out. Perhaps the number of dependencies is beyond what it can handle. https://github.com/oxidecomputer/omicron/network/updates/729281002

sunshowers:
Weird, it looks like dependabot is stuck in some sort of maybe-quadratic resolution issue (in [this log](https://github.com/oxidecomputer/omicron/network/updates/729281002) `num-rational` and `futures-task` are fetched 78 times each -- why?)

I think part of what might be happening is that Dependabot doesn't know how to handle lockfile-only updates, and with the workspace-hack, a number of dependencies made their way from being transitive-only to being direct. So dependabot's issuing a lot of updates that it really should have been all this time, which is causing resolution slowdowns.

For people with access to Oxide RFDs, https://rfd.shared.oxide.computer/rfd/0434 talks about this and other issues.

sunshowers:
https://github.com/oxidecomputer/omicron/pull/4236 should address this.

sunshowers:
https://github.com/oxidecomputer/omicron/pull/4241 is an example of a PR which updates the workspace-hack.

We do need https://github.com/oxidecomputer/omicron/pull/4244.

PR #4170, opened 2023-10-02
we don't need workspace deps for dev-tools (and omdb is stale)

ahl:
We don't seem to need the `omicron-dev` dependency in `Cargo.toml` and the `omdb` one refers to its old location.


sunshowers (approved):
Thanks!

Hopefully this should start dependabot functioning again, though we still need to find a solution for workspace-hack as mentioned in #4166.

Issue #4193, opened 2023-10-03
Make `crdb-seed` more robust to missing files


david-crespo:
Trying to run tests on 6bc5e6062df7bffbe0eca1465b9d116ff7849e9f after previously successfully running them on (95% sure) the last commit from #4154, I got the following error:

```
--- STDERR:              omicron-nexus::test_all integration_tests::basic::test_basic_failures ---
log file: /var/folders/lk/vsx86g0545g3g2tsmvs50py40000gn/T/test_all-e3c1289ea1f963ce-test_basic_failures.32561.0.log
note: configured to log to "/var/folders/lk/vsx86g0545g3g2tsmvs50py40000gn/T/test_all-e3c1289ea1f963ce-test_basic_failures.32561.0.log"
thread 'integration_tests::basic::test_basic_failures' panicked at 'Cannot copy storage from seed directory: Failed to copy subdirectory /var/folders/lk/vsx86g0545g3g2tsmvs50py40000gn/T/crdb-base/c57f4ced1447ac7b5e0722b644563fcf94b4998591ad01a8c2cabf9ec1af380e/logs to /var/folders/lk/vsx86g0545g3g2tsmvs50py40000gn/T/.tmp2DCWRn/data/logs

Caused by:
    0: Failed to copy file at /var/folders/lk/vsx86g0545g3g2tsmvs50py40000gn/T/crdb-base/c57f4ced1447ac7b5e0722b644563fcf94b4998591ad01a8c2cabf9ec1af380e/logs/cockroach-pebble.log to /var/folders/lk/vsx86g0545g3g2tsmvs50py40000gn/T/.tmp2DCWRn/data/logs/cockroach-pebble.log
    1: No such file or directory (os error 2)', /Users/david/oxide/omicron/test-utils/src/dev/mod.rs:156:18
```

The file it's mad about is a symlink that points to a nonexistent file.

```
~/oxide/omicron $ l /var/folders/lk/vsx86g0545g3g2tsmvs50py40000gn/T/crdb-base/c57f4ced1447ac7b5e0722b644563fcf94b4998591ad01a8c2cabf9ec1af380e/logs
drwxr-x---  - david 27 Sep 20:51 goroutine_dump
drwxr-x---  - david 27 Sep 20:51 heap_profiler
drwxr-x---  - david 27 Sep 20:51 inflight_trace_dump
[...]

PR #4208, opened 2023-10-05
[crdb-seed] use a tarball, fix omicron-dev run-all

### LINKED ISSUE #4206: `omicron-dev run-all` relies on `CRDB_SEED_DIR`
Recent changes to crdb seeding stuff mean that `omicron-dev run-all` relies on an env var that is usually set by nextest. Ideally, `omicron-dev run-all` would handle this somehow and keep working as a one-liner.

```
~/oxide/omicron $ cargo run --bin=omicron-dev -- run-all
    Finished dev [unoptimized + debuginfo] target(s) in 1.04s
log file: /var/folders/lk/vsx86g0545g3g2tsmvs50py40000gn/T/omicron-dev-omicron-dev.26093.0.log
note: configured to log to "/var/folders/lk/vsx86g0545g3g2tsmvs50py40000gn/T/omicron-dev-omicron-dev.26093.0.log"
thread 'main' panicked at 'CRDB_SEED_DIR missing -- are you running this test with `cargo nextest run`?: NotPresent', nexus/test-utils/src/db.rs:22:10
zsh: abort      cargo run --bin=omicron-dev -- run-all
```

Indeed, I can make things work by running the seed command manually to get the env var and then manually setting it:

```
~/oxide/omicron $ cargo run -p crdb-seed 
    Finished dev [unoptimized + debuginfo] target(s) in 0.68s
     Running `target/debug/crdb-seed`
Oct 04 21:15:18.620 INFO Using existing seed directory: `/var/folders/lk/vsx86g0545g3g2tsmvs50py40000gn/T/crdb-base/c57f4ced1447ac7b5e0722b644563fcf94b4998591ad01a8c2cabf9ec1af380e`
Oct 04 21:15:18.620 WARN NEXTEST_ENV not set (is this script running under nextest?)

~/oxide/omicron $ CRDB_SEED_DIR=/var/folders/lk/vsx86g0545g3g2tsmvs50py40000gn/T/crdb-base/c57f4ced1447ac7b5e0722b644563fcf94b4998591ad01a8c2cabf9ec1af380e cargo run --bin=omicron-dev -- run-all
    Finished dev [unoptimized + debuginfo] target(s) in 0.92s
log file: /var/folders/lk/vsx86g0545g3g2tsmvs50py40000gn/T/omicron-dev-omicron-dev.26123.0.log
note: configured to log to "/var/folders/lk/vsx86g0545g3g2tsmvs50py40000gn/T/omicron-dev-omicron-dev.26123.0.log"
DB URL: postgresql://root@[::1]:58862/omicron?sslmode=disable
DB address: [::1]:58862
omicron-dev: nexus internal API:    [::1]:12221
etc etc 
```

This is the spot that relies on the env var. I'd try a fix myself, but I'm not sure what the best approach is. I assume manual runs of `run-all` should do something analogous to `cargo run -p crdb-seed` in order to get the seed dir location and pass it to the rest of the DB setup.

https://github.com/oxidecomputer/omicron/blob/a9104a0786f437e38d4945dd77470bc8fd4f974c/nexus/test-utils/src/db.rs#L19-L35

sunshowers:
Several changes:

1. In https://github.com/oxidecomputer/omicron/issues/4193, @david-crespo observed some missing files in the crdb-seed generated directory. My suspicion is that that is due to the `/tmp` cleaner that runs on macOS. @davepacheco suggested using a tarball to get atomicity (either the file exists or it doesn't), and it ended up being pretty simple to do that at the end.
2. Teach nexus-test-utils to ensure that the seed tarball exists, fixing `omicron-dev run-all` and anything else that uses nexus-test-utils (and avoiding a dependency on the environment).
3. Move `crdb-seed` to `dev-tools` (thanks Dave for pointing it out!)
4. Add a way to invalidate the cache with `CRDB_SEED_INVALIDATE=1` in the environment.
5. Add a readme for `crdb-seed`.

Fixes #4206.

Hopefully addresses #4193.


crdb-seed/Cargo.toml
```diff
@@ -5,12 +5,15 @@ edition = "2021"
...
+atomicwrites.workspace = true
+anyhow.workspace = true
 camino.workspace = true
 camino-tempfile.workspace = true
 dropshot.workspace = true
 hex.workspace = true
 omicron-test-utils.workspace = true
 ring.workspace = true
 slog.workspace = true
+tar.workspace = true
```
smklein, line 17: Same here, I'm only seeing this used by `test-utils`

test-utils/src/dev/mod.rs
```diff
@@ -118,13 +132,14 @@ pub async fn test_setup_database(
     source: StorageSource,
 ) -> db::CockroachInstance {
     usdt::register_probes().expect("Failed to register USDT DTrace probes");
-    setup_database(log, source).await
+    setup_database(log, source).await.unwrap()
 }
 
+// TODO: switch to anyhow entirely -- this function is currently a mishmash
```
smklein, line 138: Just to clarify, this appears to already be using an `anyhow::Result`?

sunshowers:
I tried migrating the function to `anyhow` so I could get better debugging, but couldn't get all the way -- there are still some `.expect`s littered in there. Hopefully that isn't too much of a problem since this is test-only code.

smklein (approved):
LGTM, leaving this as a single artifact seems simpler to manage.

Seems like we'd still be at-risk of a `/tmp` cleaner deleting this directory, say, "after nextest setup begins, but before the test executes", but I'm not sure we can ever deal with that problem while still using `/tmp`.

crdb-seed/Cargo.toml
```diff
@@ -5,12 +5,15 @@ edition = "2021"
 license = "MPL-2.0"
 
 [dependencies]
+atomicwrites.workspace = true
```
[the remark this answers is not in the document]
sunshowers: Fixed -- I moved the code around but forgot to remove this.

crdb-seed/Cargo.toml
sunshowers: Same.

test-utils/src/dev/mod.rs
sunshowers: So this is currently a mishmash of `anyhow::Result` and `unwrap`/`expect`. I've expanded the comment to make this clearer.

sunshowers:
I also added a way to invalidate the cache with `CRDB_SEED_INVALIDATE=1`. Any idea if/where this is worth documenting?

smklein:
> I also added a way to invalidate the cache with `CRDB_SEED_INVALIDATE=1`. Any idea if/where this is worth documenting?

A help message for the `crdb-seed` binary would help -- we could also document `NEXTEST_ENV` usage there too?

sunshowers:
> A help message for the crdb-seed binary would help -- we could also document NEXTEST_ENV usage there too?

Hmm, any thoughts on checking that `argv[1]` is `--help` vs just pulling in clap for that? clap seems a little overkill.

Maybe we could also add a README.

sunshowers:
Requesting review again because I added a bunch of new stuff to the PR.

test-utils/src/dev/seed.rs
```diff
@@ -0,0 +1,228 @@
...
+    why_invalidate: Option<&str>,
+) -> Result<(Utf8PathBuf, SeedTarballStatus)> {
+    // If the CRDB_SEED_TAR_ENV variable is set, return an error.
+    if let Ok(val) = std::env::var(CRDB_SEED_TAR_ENV) {
+        anyhow::bail!(
+            "{CRDB_SEED_TAR_ENV} is set to `{val}` -- implying that a test called \
+             ensure_seed_tarball_exists. Instead, tests should rely on the `crdb-seed` \
+             setup script."
+        );
+    }
```
smklein, line 84: This seems like a bit of a proxy for `#[ifcfg(not(test))]` -- is that too extreme of a conditional qualifier to put on this method?

It does seem like it would make `crdb-seed`, the binary, untestable, but I think it would enforce more accurately what the comments are describing.

test-utils/src/dev/seed.rs
```diff
@@ -0,0 +1,228 @@
...
+                .context("failed to remove seed tarball")?;
+            true
+        }
+        (true, None) => {
+            // The tarball exists. Update its mtime (i.e. `touch` it) to ensure
+            // that it doesn't get deleted by a /tmp cleaner like on macOS.
+            std::fs::OpenOptions::new()
+                .append(true)
+                .open(&desired_seed_tar)
+                .context("failed to touch seed tarball")?;
```
smklein, line 111: Is this sufficient to update mtime, if the file exists? If the file already exists, but nothing is written, I would have **suspected** that this does not update the underlying timestamp.

test-utils/src/dev/seed.rs
```diff
@@ -0,0 +1,228 @@
...
+            false
+        }
+        (false, None) => {
+            // The tarball doesn't exist.
+            false
+        }
+    };
+
+    // The tarball didn't exist when we started, so try to create it.
+    //
```
smklein, line 128: Nit: the tarball no longer exists

If `why_invalidated` is Some, it might have existed, but been removed

smklein (approved):
LGTM, mod one comment clarifying mtimes!

test-utils/src/dev/seed.rs
sunshowers: Really really good catch, thank you. I think I may end up using the filetime crate as in https://github.com/uutils/coreutils/blob/a596cda51612e72b5245ccb8c2fa6c05a89569a7/src/uu/touch/src/touch.rs#L205-L207.

test-utils/src/dev/seed.rs
sunshowers: Expanded this comment:

```rust
    // At this point the tarball does not exist (either because it didn't exist
    // in the first place or because it was deleted above), so try to create it.
```

test-utils/src/dev/seed.rs
sunshowers: Done, the filetime crate works great.

test-utils/src/dev/seed.rs
sunshowers: This is a dependency (not a leaf node) so I don't think `cfg(not(test))` would work? And if I try gating it behind a feature flag, the issue would be that `omicron-dev` needs this function and so feature unification would mean this gets included in test binaries if you're doing a top-level `cargo test`.

sunshowers: Actually, I ended up both preserving this check and gating it behind a feature flag. So we're triply sure this code isn't used by another other than omicron-dev and crdb-seed.

Issue #4207, opened 2023-10-04
CI run of `omicron-nexus::test_all integration_tests::schema::versions_have_idempotent_up` failed with transaction error


bnaecker:
The test `omicron-nexus::test_all integration_tests::schema::versions_have_idempotent_up` failed on a [recent CI run](https://buildomat.eng.oxide.computer/wg/0/details/01HBY3QJ3W9RFEYCFEAWERSFY3/Lvjsw71FVHUAje5HIFTf77IQSOc3lpbmZH3lMvwOfymYgAME/01HBY3S4VQKFEQJZ22VPBZZNTM#S4967) for a PR which changed unrelated code. I know we have a [number of issues](https://github.com/oxidecomputer/omicron/issues/4189) related to DB improvements, many of which are related to handling transactions more safely. I'm not sure if this is a dupe of one of those.

smklein:
Relevant error from the logs:

```
thread 'integration_tests::schema::versions_have_idempotent_up' panicked at 'Failed to COMMIT update:
Error {
  kind: Db,
  cause: Some(DbError { severity: "ERROR", parsed_severity: Some(Error), code: SqlState(E40001), message: "restart transaction: TransactionRetryWithProtoRefreshError: TransactionRetryError: retry txn (RETRY_SERIALIZABLE): \"sql txn\" meta=..., hint: Some("See: https://www.cockroachlabs.com/docs/v22.1/transaction-retry-error-reference.html#retry_serializable"), ... })
}', nexus/tests/integration_tests/schema.rs:98:18
```

Honestly kinda fucked up it's like CI knew I just wrote RFD 436. I'll fix this.

PR #4209, opened 2023-10-05
Add transaction retry to schema upgrade integration tests

smklein:
Fixes https://github.com/oxidecomputer/omicron/issues/4207


bnaecker (approved):
Looks great, thanks for the quick turnaround!

PR #4210, opened 2023-10-05
Simplify Diesel Error management

smklein:
Depends on https://github.com/oxidecomputer/async-bb8-diesel/pull/54

As of https://github.com/oxidecomputer/omicron/pull/4140 , we check out connections before issuing queries to the underlying database. This means that when we receive errors from the database, they are not overloaded as "connection checkout" OR "database" errors - they are now always database errors.


nexus/db-queries/src/db/error.rs
```diff
@@ -93,19 +85,6 @@ fn format_database_error(
...
-pub fn diesel_result_optional<T>(
-    result: Result<T, ConnectionError>,
-) -> Result<Option<T>, ConnectionError> {
-    match result {
-        Ok(v) => Ok(Some(v)),
-        Err(ConnectionError::Query(DieselError::NotFound)) => Ok(None),
-        Err(e) => Err(e),
-    }
-}
-
```
smklein, line 108: Since we're just using native Diesel errors, this isn't necessary -- we can just use Diesel's built-in `OptionalExtension`.

Cargo.toml
```diff
@@ -135,7 +135,7 @@ api_identity = { path = "api_identity" }
 approx = "0.5.1"
 assert_matches = "1.5.0"
 assert_cmd = "2.0.12"
-async-bb8-diesel = { git = "https://github.com/oxidecomputer/async-bb8-diesel", rev = "da04c087f835a51e0441addb19c5ef4986e1fcf2" }
+async-bb8-diesel = { git = "https://github.com/oxidecomputer/async-bb8-diesel", branch = "error-handling" }
```
smklein, line 138: TODO(me): Before merging, switch this back to the rev of https://github.com/oxidecomputer/async-bb8-diesel/pull/54

nexus/db-queries/src/db/true_or_cast_error.rs
```diff
@@ -93,10 +93,7 @@ pub fn matches_sentinel(
     match e {
         // Catch the specific errors designed to communicate the failures we
         // want to distinguish.
-        ConnectionError::Query(Error::DatabaseError(
-            DatabaseErrorKind::Unknown,
-            ref info,
-        )) => {
+        Error::DatabaseError(DatabaseErrorKind::Unknown, ref info) => {
```
sunshowers, line 96: Don't need `ref` here these days I think.

nexus/db-queries/src/db/queries/network_interface.rs
```diff
@@ -292,80 +289,76 @@ fn decode_database_error(
...
         }
         // This catches the UUID-cast failure intentionally introduced by
         // `push_instance_state_verification_subquery`, which verifies that
         // the instance doesn't even exist when running this query.
-        ConnectionError::Query(Error::DatabaseError(
-            DatabaseErrorKind::Unknown,
-            ref info,
-        )) if info.message() == NO_INSTANCE_ERROR_MESSAGE => {
+        DieselError::DatabaseError(DatabaseErrorKind::Unknown, ref info)
+            if info.message() == NO_INSTANCE_ERROR_MESSAGE =>
```
sunshowers, line 351: Don't need `ref` here these days (here or in the rest of the file)

sunshowers (approved):
Looks good!

nexus/db-queries/src/db/datastore/snapshot.rs
```diff
@@ -111,9 +110,7 @@ impl DataStore {
                     .await
                 {
                     Ok(v) => Ok(Some(v)),
-                    Err(ConnectionError::Query(DieselError::NotFound)) => {
-                        Ok(None)
-                    }
+                    Err(DieselError::NotFound) => Ok(None),
                     Err(e) => Err(e),
```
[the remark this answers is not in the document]
smklein: Nah, this is just regular error flattening. I can show you an example of the `OptionalExtension` elsewhere in this PR though....

nexus/db-queries/src/db/datastore/silo.rs
```diff
@@ -338,16 +337,15 @@ impl DataStore {
...
-        .map_err(|e| public_error_from_diesel(e, ErrorHandler::Server))?;
+        let project_found = project::dsl::project
+            .filter(project::dsl::silo_id.eq(id))
+            .filter(project::dsl::time_deleted.is_null())
+            .select(project::dsl::id)
+            .limit(1)
+            .first_async::<Uuid>(&*conn)
+            .await
+            .optional()
+            .map_err(|e| public_error_from_diesel(e, ErrorHandler::Server))?;
```
smklein, line 348: This is an example of us using https://docs.diesel.rs/master/diesel/prelude/trait.OptionalExtension.html

This works now because we're passing a https://docs.diesel.rs/master/diesel/result/enum.Error.html around, which can go directly to `public_error_from_diesel`.

nexus/db-queries/src/db/queries/network_interface.rs
smklein: I do need *one* of them in a case where we re-use the `err`  we're matching on, but I removed the rest

nexus/db-queries/src/db/true_or_cast_error.rs
smklein: Fixed!

nexus/db-queries/src/db/datastore/snapshot.rs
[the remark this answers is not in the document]
jmpesp: It looks to me like the [OptionalExtension](https://docs.diesel.rs/master/src/diesel/result.rs.html#283):

```rust
impl<T> OptionalExtension<T> for QueryResult<T> {
    fn optional(self) -> Result<Option<T>, Error> {
        match self {
            Ok(value) => Ok(Some(value)),
            Err(Error::NotFound) => Ok(None),
            Err(e) => Err(e),
        }
    }
}
```

jmpesp (approved):

nexus/db-queries/src/db/datastore/snapshot.rs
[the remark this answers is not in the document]
smklein: Oh, huh, good point, maybe I can / should use that here. Sorry for misinterpreting the original comment, @sunshowers !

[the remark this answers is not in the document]
smklein: Updated to use `optional()`, that simplifies this pathway a bit.

Issue #4211, opened 2023-10-05
Persist fault management data across reboots


citrus-it:
While @wesolows was digging into [stlouis#281](https://github.com/oxidecomputer/stlouis/issues/281)/cs#39 he found breadcrumbs in the kernel memory of the crash dump that indicated that the fault management system had done /something/ recently but, since the fault management logs and database are on non-persistent storage on the system ramdisk, that data was lost and we can only guess what occurred.

We should back the fault management data in `/var/fm/fmd` with a dataset on the current boot disk so that the fault management history is preserved. In the future, we should likely do the same for GZ system log files and other things.

Note that we have also previously seen a system fill up the root ramdisk (via `/var/fm/fmd`) due to a flood of ZFS errors so an appropriate quota should also be applied here.

citrus-it:
A few notes from testing on a bench gimlet:

```
gimlet-sn06 # cat `svcs -L sled-agent` | looker -c 'r.component == "BackingFs"'
09:48:42.276Z INFO SledAgent (BackingFs): Processing fmd
09:48:42.277Z INFO SledAgent (BackingFs): Ensuring dataset oxi_a462a7f7-b628-40fe-80ff-4e4189e2d62b/backing/fmd
09:48:42.430Z INFO SledAgent (BackingFs): Stopping service svc:/system/fmd:default
09:48:43.673Z INFO SledAgent (BackingFs): Mounting oxi_a462a7f7-b628-40fe-80ff-4e4189e2d62b/backing/fmd on /var/fm/fmd
09:48:43.721Z INFO SledAgent (BackingFs): Starting service svc:/system/fmd:default
```

```
gimlet-sn06# svcadm restart sled-agent
10:06:46.841Z INFO SledAgent (BackingFs): Processing fmd
10:06:46.841Z INFO SledAgent (BackingFs): Ensuring dataset oxi_a462a7f7-b628-40fe-80ff-4e4189e2d62b/backing/fmd
10:06:46.932Z INFO SledAgent (BackingFs): /var/fm/fmd is already mounted
```

```
gimlet-sn06 # zfs get mounted  `zfs list -Honame | grep backing`
NAME                                                  PROPERTY  VALUE    SOURCE
oxi_a462a7f7-b628-40fe-80ff-4e4189e2d62b/backing      mounted   yes      -
oxi_a462a7f7-b628-40fe-80ff-4e4189e2d62b/backing/fmd  mounted   yes      -
[...]

PR #4212, opened 2023-10-05
Back /var/fm/fmd with a dataset from the boot M.2

citrus-it:
`/var/fm/fmd` is where the illumos fault management system records data.
We want to preserve this data across system reboots and in real time
rather than via periodic data copying, so that the information is
available should the system panic shortly thereafter.

Fixes: https://github.com/oxidecomputer/omicron/issues/4211

### DIFF
--- a/sled-agent/src/backingfs.rs
+++ b/sled-agent/src/backingfs.rs
@@ -1,0 +1,178 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! Operations for dealing with persistent backing mounts for OS data
+
+// On Oxide hardware, the root filesystem is backed by a ramdisk and
+// non-persistent. However, there are several things within the root filesystem
+// which are useful to preserve across reboots, and these are backed persistent
+// datasets on the boot disk.
+//
+// Each boot disk contains a dataset sled_hardware::disk::M2_BACKING_DATASET
+// and for each backing mount, a child dataset is created under there that
+// is configured with the desired mountpoint in the root filesystem. Since
+// there are multiple disks which can be used to boot, these datasets are also
+// marked with the "canmount=noauto" attribute so that they do not all try to
+// mount automatically and race -- only one could ever succeed. This allows us
+// to come along later an specifically mount the one that we want (the one from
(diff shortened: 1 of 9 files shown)


sled-agent/src/backingfs.rs
```diff
@@ -0,0 +1,178 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! Operations for dealing with persistent backing mounts for OS data
+
+// On Oxide hardware, the root filesystem is backed by a ramdisk and
+// non-persistent. However, there are several things within the root filesystem
+// which are useful to preserve across reboots, and these are backed persistent
+// datasets on the boot disk.
```
jordanhendricks, line 10: > these are backed persistent datasets on the boot disk.

to be clear, that's done by this implementation?

sled-agent/src/backingfs.rs
```diff
@@ -0,0 +1,178 @@
...
+//
+// Each boot disk contains a dataset sled_hardware::disk::M2_BACKING_DATASET
+// and for each backing mount, a child dataset is created under there that
+// is configured with the desired mountpoint in the root filesystem. Since
+// there are multiple disks which can be used to boot, these datasets are also
+// marked with the "canmount=noauto" attribute so that they do not all try to
+// mount automatically and race -- only one could ever succeed. This allows us
+// to come along later an specifically mount the one that we want (the one from
```
jordanhendricks, line 18: "an" -> "and"

jordanhendricks (approved):
Looks good, just some nits.

sled-agent/src/backingfs.rs
```diff
@@ -0,0 +1,178 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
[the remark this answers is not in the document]
citrus-it: It originally was called that 😂 before I decided to drop the _ for some reason.
Moved it back.

sled-agent/src/backingfs.rs
citrus-it: Yes, that's right.

citrus-it:
> Looks good, just some nits.

Thanks for taking a look, I've tested this latest commit on a bench gimlet so I've set it to automerge once the CI deploy job passes.

PR #4224, opened 2023-10-06
Proxy Nexus's external API out the tech port via wicketd

jgallagher:
This implements one of the proposed solutions in RFD 431 (which I still need to get back to and refine a bit!):

* Nexus gains another dropshot instance that serves the external API (using the same shared server context as the existing external API servers) on the underlay network
* wicketd runs a TCP proxy on the tech port
* When wicketd receives a new connection on that proxy socket, it:
    * looks up Nexus in internal DNS (these records already exist for the _internal_ API, but that's good enough for us: we're just after the IP address here)
    * connects to any Nexus instance reported by the DNS query
    * proxies the connection between the tech port client and the Nexus instance

This is therefore exposing the "true" Nexus API, which still requires TLS, client auth, etc.

## Testing this PR

This is currently deployed on `madrid`, and the racklette is set up using dogfood's certs, and I manually tested (and demo'd) it there.

I'd be happy to add integration tests, but am not sure what would make sense. I could add a test where I spin up an internal DNS server, seed it with a fake Nexus entry, and ensure that wicketd will proxy connections to that fake Nexus, maybe? Is it worth trying to add tests that Nexus exposes the same API on both the external and "techport" interfaces?


nexus/src/lib.rs
```diff
@@ -131,23 +132,59 @@ impl Server {
...
+            let server_starter_external_techport =
+                dropshot::HttpServerStarter::new_with_tls(
+                    &techport_server_config,
+                    external_api(),
+                    Arc::clone(&apictx),
+                    &log.new(o!("component" => "dropshot_external_techport")),
+                    tls_config.map(dropshot::ConfigTls::Dynamic),
+                )
+                .map_err(|error| {
+                    format!("initializing external server: {}", error)
```
smklein, line 176: Nit, I might mention this is the techport server in the error message

wicketd/src/http_entrypoints.rs
```diff
@@ -1237,6 +1241,51 @@ async fn get_preflight_uplink_report(
...
+        HttpError::for_unavail(
+            None,
+            format!("failed to read SMF values: {err}"),
+        )
+    })?;
+
+    let Some(smf_values) = smf_values else {
+        return Err(HttpError::for_bad_request(
+            None,
+            "reloading config from SMF only available on illumos".to_string(),
```
smklein, line 1265: Oh, so on non-illumos platforms we'd expect it to `SmfConfigValues::read_current()` to return `None` rather than throwing an error?

smf/wicketd/manifest.xml
```diff
@@ -12,10 +12,25 @@
...
     timeout_seconds='0' />
   <exec_method type='method' name='stop' exec=':kill' timeout_seconds='0' />
 
+  <!--
+    This refresh method only exists to support changing `config/rack-subnet`,
+    and makes a couple critical assumptions:
+
+      * `config/address` is not changing
+      * the dropshot server listening at `config/address` is not configured
+        for TLS
```
smklein, line 25: We could validate these assumptions in code, within the body of `/reload-config`? It might be worthwhile to fail-hard in these cases if the error would otherwise be confusing.

wicketd/src/lib.rs
```diff
@@ -53,7 +57,50 @@ pub struct Args {
...
+            })?;
+            Some(Ipv6Subnet::new(addr))
+        };
+
+        Ok(Some(Self { rack_subnet }))
+    }
+
+    #[cfg(not(target_os = "illumos"))]
+    pub fn read_current() -> Result<Option<Self>> {
+        Ok(None)
```
smklein, line 102: Wouldn't it be better to return an error here? Is there a case where that breaks down?

smklein (approved):

smf/wicketd/manifest.xml
jgallagher: I don't think it's super easy to validate http vs https from within the endpoint, but that's maybe okay: the curl invocation itself will fail hard. I added a check that `config/address` is not changing in [53e0654](https://github.com/oxidecomputer/omicron/pull/4224/commits/53e06542b7c3b57722a12e672c35d49d3afb3d17)

wicketd/src/lib.rs
jgallagher: It would be better! This choice went along with the way the `wicketd` clap arguments were being parsed in an earlier version, and I didn't realize I could clean this up at this point. Changed in [53e0654](https://github.com/oxidecomputer/omicron/pull/4224/commits/53e06542b7c3b57722a12e672c35d49d3afb3d17)

PR #4236, opened 2023-10-10
[dependencies] add Renovate config

sunshowers:
* Add configuration for automatically creating dependencies, and for pinning GitHub Actions digests
* Add a post-upgrade script that runs cargo-hakari.

Things to check before we land this:

- [x] Make sure the Mend Renovate App isn't enabled for Omicron (Augustus verified this)

Depends on https://github.com/oxidecomputer/renovate-config/pull/5.

See [RFD 434](https://rfd.shared.oxide.computer/rfd/0434) and #4166.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,0 +1,8 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": [
+    "local>oxidecomputer/renovate-config",
+    "local>oxidecomputer/renovate-config//rust/autocreate",
+    "helpers:pinGitHubActionDigests"
+  ]
+}
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -3,6 +3,7 @@
   "extends": [
     "local>oxidecomputer/renovate-config",
     "local>oxidecomputer/renovate-config//rust/autocreate",
+    "local>oxidecomputer/renovate-config:post-upgrade",
     "helpers:pinGitHubActionDigests"
   ]
 }
--- a/tools/renovate-post-upgrade.sh
+++ b/tools/renovate-post-upgrade.sh
@@ -4,18 +4,37 @@
 
 set -euo pipefail
 
+# Function to retry a command up to 3 times.
+function retry_command {
+    local retries=3
+    local delay=5
+    local count=0
+    until "$@"; do
+        exit_code=$?
+        count=$((count+1))
+        if [ $count -lt $retries ]; then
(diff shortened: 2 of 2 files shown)


.github/renovate.json
```diff
@@ -0,0 +1,9 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": [
+    "local>oxidecomputer/renovate-config",
+    "local>oxidecomputer/renovate-config//rust/autocreate",
+    "local>oxidecomputer/renovate-config:post-upgrade",
```
smklein, line 6: By using this "extends" statement, we automatically parse https://github.com/oxidecomputer/renovate-config/pull/5 's `post-upgrade.json` and run the associated script? Or is there some other registration that's happening here?

tools/renovate-post-upgrade.sh
```diff
@@ -0,0 +1,23 @@
...
+    # Need cargo-binstall to install cargo-hakari.
+    if ! command -v cargo-binstall &> /dev/null; then
+        # Fetch cargo binstall.
+        echo "Installing cargo-binstall..."
+        curl -L --proto '=https' --tlsv1.2 -sSf https://raw.githubusercontent.com/cargo-bins/cargo-binstall/main/install-from-binstall-release.sh | bash
+    fi
+
+    # Install cargo-hakari.
+    echo "Installing cargo-hakari..."
+    cargo binstall cargo-hakari --no-confirm
```
smklein, line 18: @davepacheco mentioned that "we should expect all our cloud services to fail transiently at any time", in https://github.com/oxidecomputer/omicron/issues/4165 . Do you think we should build a retry-loop here, for these installation commands?

smklein (approved):

tools/renovate-post-upgrade.sh
sunshowers: Huh, for some reason I thought binstall already did retries but it turns out it does not (https://github.com/cargo-bins/cargo-binstall/pull/131 was rejected). Yes, retries are worth adding here for sure.

.github/renovate.json
sunshowers: Yes -- it just fetches that JSON file and includes that in the config.

tools/renovate-post-upgrade.sh
sunshowers: Added retries.

smklein (approved):
Looks great, thank you!

Issue #4221, opened 2023-10-06
Metrics API get requests return "Query must resolve to a single timeseries if limit is specified"


askfongjojo:
Console is no longer getting back any metrics (for utilization or disk metrics). It's hitting 500 errors in Nexus, e.g.
```
01:23:15.075Z INFO 65a11c18-7f59-41ac-b9e7-680627f996e7 (dropshot_external): request completed
    error_message_external = Internal Server Error
    error_message_internal = Query must resolve to a single timeseries if limit is specified
    file = /home/build/.cargo/git/checkouts/dropshot-a4a923d29dccc492/fa728d0/dropshot/src/server.rs:841
    latency_us = 429136
    local_addr = 172.30.2.5:443
    method = GET
    remote_addr = 172.20.17.42:58517
    req_id = 6620223f-f85a-4bba-9c1a-e4db78417764
    response_code = 500
    uri = https://oxide.sys.rack2.eng.oxide.computer/v1/disks/dbdata-ext4-primary/metrics/write?project=try&start_time=2023-10-05T01%3A23%3A14.591Z&end_time=2023-10-06T01%3A23%3A14.591Z&limit=3000
```

I invoked a similar query via API without the `limit` parameter to see what would happen and it's returning the same error:
```
01:25:12.706Z INFO 65a11c18-7f59-41ac-b9e7-680627f996e7 (dropshot_external): request completed
    error_message_external = Internal Server Error
    error_message_internal = Query must resolve to a single timeseries if limit is specified
    file = /home/build/.cargo/git/checkouts/dropshot-a4a923d29dccc492/fa728d0/dropshot/src/server.rs:841
    latency_us = 139882
    local_addr = 172.30.2.5:443
    method = GET
    remote_addr = 172.20.17.42:58551
    req_id = 5a8175d3-47b8-4add-bda7-f5d2734a954b
    response_code = 500
    uri = /v1/disks/c64f5fbe-8a2b-4f8c-a229-13a03dbf1c39/metrics/read_bytes?start_time=2023-10-04T16%3A34%3A22.508Z&end_time=2023-10-6T16%3A34%3A22.508Z
```

david-crespo:
Some notes on how this works. Will probably need @bnaecker's help to puzzle this out. What we'll probably want to do is look in Clickhouse to see why multiple timeseries with the same name (at least, I think that's what it is) exist.

[...]

PR #4246, opened 2023-10-11
[oximeter] Add minimal versioning support

smklein:
Provides a mechanism to explicitly version Oximeter schema used within Clickhouse, and to wipe that data
if any schema-breaking changes are made.

This is retroactively intended to solve the problems discussed within https://github.com/oxidecomputer/omicron/issues/4221 , where the `timeseries_key` generation was altered.


oximeter/db/src/client.rs
```diff
@@ -1339,6 +1429,126 @@ mod tests {
...
+            )
+            .await
+            .expect("Failed to SELECT from database")
+            .lines()
+            .count()
+    }
+
+    #[tokio::test]
+    async fn test_database_version_update_idempotent() {
+        let log = slog::Logger::root(slog::Discard, o!());
```
bnaecker, line 1446: I've been trying to move all the tests over to the test logging infrastructure from `omicron_test_utils`. The original tests predated all that stuff, I think, or I didn't use it for some reason.

Would you mind using the `test_setup_log` function a la L1312?

oximeter/db/src/client.rs
```diff
@@ -1339,6 +1429,126 @@ mod tests {
...
+            .expect("Failed to initialize timeseries database");
+
+        assert_eq!(1, get_schema_count(&client).await);
+
+        db.cleanup().await.expect("Failed to cleanup ClickHouse server");
+    }
+
+    #[tokio::test]
+    async fn test_database_version_will_not_downgrade() {
+        let log = slog::Logger::root(slog::Discard, o!());
```
bnaecker, line 1483: Same note about `test_setup_log` here.

oximeter/db/src/client.rs
```diff
@@ -1339,6 +1429,126 @@ mod tests {
...
+            .initialize_db_with_version(replicated, model::OXIMETER_VERSION)
+            .await
+            .expect_err("Should have failed, downgrades are not supported");
+
+        db.cleanup().await.expect("Failed to cleanup ClickHouse server");
+    }
+
+    #[tokio::test]
+    async fn test_database_version_wipes_old_version() {
+        let log = slog::Logger::root(slog::Discard, o!());
```
bnaecker, line 1518: Same note about `test_setup_log` here.

bnaecker (commented):
Looks good, thanks for the quick and helpful fix on this. Couple of small suggestions. Thanks!

oximeter/db/preprocessed_configs/config.xml
```diff
@@ -0,0 +1,41 @@
+<!-- This file was generated automatically.
```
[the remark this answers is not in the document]
smklein: Removed

oximeter/db/src/client.rs
```diff
@@ -269,7 +271,95 @@ impl Client {
         .map_err(|e| Error::Database(e.to_string()))
     }
 
-    // Verifies if instance is part of oximeter_cluster
+    /// Validates that the schema used by the DB matches the version used by
+    /// the executable using it.
+    ///
+    /// This function will wipe metrics data if the version stored within
```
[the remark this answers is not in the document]
smklein: Good catch, updated the comment to cover these cases.

oximeter/db/src/client.rs
```diff
@@ -269,7 +271,95 @@ impl Client {
...
+    }
+
+    async fn read_latest_version(&self) -> Result<u64, Error> {
+        let sql = format!(
+            "SELECT MAX(value) FROM {db_name}.version;",
+            db_name = crate::DATABASE_NAME,
+        );
+
+        let version = match self.execute_with_body(sql).await {
+            Ok(body) if body.is_empty() => {
```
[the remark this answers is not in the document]
smklein: So this case is actually unlikely, but I figured I'd include it regardless.

Here are the cases I'd expect us to use:

- When we upgrade the in-field databases, which exist, but don't have the "version" table, we'll see a database error containing `Table oximeter.version doesn't exist`.
- When we scan for versions in databases that haven't been formatted, we'll see a database error containing `Database oximeter doesn't exist`

This case of "no versions exist" would happen if we formatted the schema, but bailed out before we managed to insert the version (or crashed before we got that far).

I went ahead and updated the code + comments to make this a bit more explicit.

oximeter/db/src/client.rs
smklein: Sounds good, done for this and for several other callers in this file.

smklein:
I just added it in one of the more recent commits, but that check is part of this block:

https://github.com/oxidecomputer/omicron/blob/b3d57654ffdc29cc4f735bb3a6639d0e94bb6d9d/oximeter/db/src/client.rs#L346-L348

bnaecker:
Ah I see it now. Must have been looking at old commits, because of course that’s what GH shows by default. Thanks for addressing everything, feel free to merge when you’re ready.

david-crespo:
This looks good. The thing that makes me nervous is that any Oximeter client (which I understand to be tied to a control plane instance) can basically wipe Clickhouse at any time if it is running with a bumped version? Obviously we would not be bumping this version very often, but I feel there should be more documentation or _something_. At the very least, on the version itself, shouldn't it say something like HI IF YOU CHANGE THIS VALUE YOU WILL DELETE ALL THE DATA.

smklein:
> This looks good. The thing that makes me nervous is that any Oximeter client (which I understand to be tied to a control plane instance) can basically wipe Clickhouse at any time if it is running with a bumped version? Obviously we would not be bumping this version very often, but I feel there should be more documentation or _something_. At the very least, on the version itself, shouldn't it say something like HI IF YOU CHANGE THIS VALUE YOU WILL DELETE ALL THE DATA.

I added some documentation in https://github.com/oxidecomputer/omicron/pull/4246/commits/a6c1798f917a72c5c9ff3b1788d0ca55181f197d , including a link to https://github.com/oxidecomputer/omicron/issues/4271

david-crespo (approved):

Issue #4008, opened 2023-09-01
Timeseries need to use a stable hash algorithm for computing keys


bnaecker:
When `oximeter` fetches samples from a producer, it extracts a schema and a _timeseries key_. This is used to ID the specific instantiation of the timeseries within the schema, which is critical for being able to associate the records for a specific sample across the various tables. We use this to determine which fields go with which measurements.

We're currently using `std::collections::hash_map::DefaultHasher` to do this. There are no real stability guarantees around that though, which means we may lose the ability to associate records if that hash algorithm changes. We need to switch to a stable algorithm. XXHash or CityHash are attractive candidates.

I see no way to do this without breaking existing data. We need to do this ASAP, to minimize the data loss.

PR #4251, opened 2023-10-11
[oximeter] Use stable hash, stable format, for deriving timeseries_key

smklein:
- Uses [bcs](https://crates.io/crates/bcs) as a stable binary format for inputs to the `timeseries_key` function
- Uses [highway](https://crates.io/crates/highway) as a stable hash algorithm (it's keyed, fast, portable, and well-distributed).
- Additionally, adds an EXPECTORATE test validating the stability of timeseries_key values.

Fixes https://github.com/oxidecomputer/omicron/issues/4008 , and also addresses the issue raised in https://github.com/oxidecomputer/omicron/issues/4221 regarding stable input

NOTE: This PR itself *also* breaks the stability of the `timeseries_key` (hopefully for the last time), and will rely on https://github.com/oxidecomputer/omicron/pull/4246 to wipe the metrics DB for the next release.


oximeter/db/src/lib.rs
```diff
@@ -335,11 +335,13 @@ pub(crate) fn timeseries_key_for<'a>(
...
 ) -> TimeseriesKey {
-    use std::collections::hash_map::DefaultHasher;
+    use highway::HighwayHasher;
     use std::hash::{Hash, Hasher};
-    let mut hasher = DefaultHasher::new();
+    let mut hasher = HighwayHasher::default();
     for field in target_fields.chain(metric_fields) {
-        field.hash(&mut hasher);
+        bcs::to_bytes(&field)
+            .expect("Failed to serialized field to bytes")
```
smklein, line 343: This `expect` ideally would be cleaned up, but I was a bit lazy because of this comment:

> Serialization can fail if T’s implementation of Serialize decides to fail, if T contains sequences which are longer than MAX_SEQUENCE_LENGTH, or if T attempts to serialize an unsupported datatype such as a f32, f64, or char.

This seems unlikely to me because `MAX_SEQUENCE_LENGTH` is `2_147_483_647usize`, and we try to exhaustively test enum variants down below. But if y'all would prefer, I can plumb the error propagation back up.

bnaecker: Expect is fine to me.

oximeter/db/src/lib.rs
```diff
@@ -393,4 +395,90 @@ mod tests {
...
+            timeseries_key_for(
+                [&f("a", FieldValue::String("a".to_string()))].into_iter(),
+                [&f("b", FieldValue::String("b".to_string()))].into_iter(),
+            ),
+            timeseries_key_for(
+                [&f("b", FieldValue::String("b".to_string()))].into_iter(),
+                [&f("a", FieldValue::String("a".to_string()))].into_iter(),
+            ),
+        );
+    }
```
david-crespo, line 434: Kind of a weird test. I don't know that we buy too much by essentially confirming that `timeseries_key_for` concats the argument vecs in the order they're given.

oximeter/db/src/lib.rs
```diff
@@ -393,4 +395,90 @@ mod tests {
...
+            ("bool", FieldValue::Bool(true)),
+        ];
+
+        // Exhaustively testing enums is a bit tricky. Although it's easy to
+        // check "all variants of an enum are matched", it harder to test "all
+        // variants of an enum have been supplied".
+        //
+        // We use this as a proxy, confirming that each variant is represented
+        // here for the purposes of tracking stability.
+        assert_eq!(values.len(), FieldValue::COUNT);
```
david-crespo, line 468: clever

david-crespo (approved):
A+

oximeter/db/src/lib.rs
smklein: Agreed, I think I wanted this as a proxy for "is the `timeseries_key` for a `Sample` stable?"

We already have the tests for individual fields, but that doesn't tell us anything about "what if you have a sample that tries to change the order of fields?"

I went ahead and updated this test, to simply `EXPECTORATE` a `timeseries_key` value for a simple sample. That's the actual property I want to validate, so IMO it makes for a better test: 929e761c9210b501fe83126936eca85bbddc83de

oximeter/db/src/lib.rs
```diff
@@ -335,11 +335,13 @@ pub(crate) fn timeseries_key_for<'a>(
     target_fields: impl Iterator<Item = &'a Field>,
     metric_fields: impl Iterator<Item = &'a Field>,
```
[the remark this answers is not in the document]
smklein: One other thing -- I think the hashing is currently based **purely** from the `Field` structs:

https://github.com/oxidecomputer/omicron/blob/7e88bdffbae7a8796a4ebc6a5bfe80d1fedd4bb7/oximeter/oximeter/src/types.rs#L248-L255

However, the target and metric **also** have names, which could be distinct from the fields:

https://github.com/oxidecomputer/omicron/blob/7e88bdffbae7a8796a4ebc6a5bfe80d1fedd4bb7/oximeter/oximeter/src/types.rs#L644-L649

Should I be hashing this value too?


So in the case of:

```
Sample {
  measurement: N/A
  timeseries_name: "MyTimeseries",
  target: FieldSet {
    name: "MyTarget",
    fields: [ ("a", valueA), ("b", valueB)],
  },
  metric: FieldSet {
    name: "MyMetric",
    fields: [ ("c", valueC), ("d", valueD)],
  },
}
```

The thing we have been hashing -- even before this PR -- are the `Field` values, which are:

- `[ ("a", valueA), ("b", valueB), ("c", valueC), ("d", valueD)]`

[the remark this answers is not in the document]
smklein: WDYT about hashing the following for a sample, in-order:

1. The timeseries_name
2. The target name
3. The target fields
4. The metric name
5. The metric fields

oximeter/db/src/lib.rs
```diff
@@ -335,11 +335,13 @@ pub(crate) fn timeseries_key_for<'a>(
...
 ) -> TimeseriesKey {
-    use std::collections::hash_map::DefaultHasher;
+    use highway::HighwayHasher;
     use std::hash::{Hash, Hasher};
-    let mut hasher = DefaultHasher::new();
+    let mut hasher = HighwayHasher::default();
     for field in target_fields.chain(metric_fields) {
```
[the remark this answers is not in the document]
smklein: I've added this in 2e91139bb7699a1ca3e5e54306e8801a3aa23440 -- as expected, this does not change the hashes, since things are already sorted, but it does make it explicit.

I opted to alter the interface rather than re-sorting internally.

oximeter/db/src/lib.rs
```diff
@@ -393,4 +395,90 @@ mod tests {
...
+                [
+                    &f("a", FieldValue::String("a".to_string())),
+                    &f("b", FieldValue::String("b".to_string())),
+                ]
+                .into_iter(),
+                [].into_iter(),
+            ),
+        );
+
+        // However, order still matters ("a, b" != "b, a")
```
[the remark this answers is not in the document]
smklein: I have removed this test -- "stability of a sample" is more important -- but yes, this is fixed.

For what it's worth, we're sorting "the targets fields" and "the metric fields" separately, but I think we're on the same page here.

bnaecker:
The WIP I had to address this before the urgency of the 500s did that basically. Hash the timeseries name; target field names, types, and values; and then metric field names, types, and values; and datum type. 

I think with the nice canonical serialization strategy you’ve got here, we don’t need the types of the target and metric I think? That should be included by the serialized variant index, based on by reading of the bcs documentation. 

We don’t strictly need the names of the targets and metrics themselves. Those are concatenated into the timeseries name, so I think we can go without them.

So definitely the timeseries name, and field names and values for the target and metric. I strongly believe each of those lists of name-value pairs needs to be sorted before hashing. I also think we need to include the datum type. If that we’re to change, I think we’d consider that a different timeseries.

bnaecker:
Yep, sorting target fields and metric fields separately, then concatenations and hashing. Same page!

smklein:
> I think with the nice canonical serialization strategy you’ve got here, we don’t need the types of the target and metric I think? That should be included by the serialized variant index, based on by reading of the bcs documentation. We don’t strictly need the names of the targets and metrics themselves. Those are concatenated into the timeseries name, so I think we can go without them.

Oh, I see, we're saying "the timeseries name" is derived from the "target name + metric name", so it's implicit.

> So definitely the timeseries name, and field names and values for the target and metric. I strongly believe each of those lists of name-value pairs needs to be sorted before hashing. I also think we need to include the datum type. If that we’re to change, I think we’d consider that a different timeseries.

Gotcha, updated in c0110b7e94c4972157b6260474f06fcea2ea8545

bnaecker:
Sweet, this all looks very good! Thanks for addressing everything, and adding lots of comments around the details of the implementation. 

I’m happy with the whole change, so feel free to merge when you’re ready.

bnaecker (approved):

PR #4261, opened 2023-10-11
IP pools data model and API rework

### LINKED ISSUE #2148: Need API endpoint for listing IP Pools scoped to a project

### LINKED ISSUE #4002: IP pools: change/remove default
In #3985, IP pools gained an explicit property `is_default`, which can be set at create time. However, it cannot be changed through the IP pool update endpoint.

https://github.com/oxidecomputer/omicron/blob/01e730acf3efeb6ff2181d558cf373ded383a365/nexus/types/src/external_api/params.rs#L729-L752

Because there can be at most one default for a given scope (fleet-wide or silo-specific),

https://github.com/oxidecomputer/omicron/blob/01e730acf3efeb6ff2181d558cf373ded383a365/schema/crdb/dbinit.sql#L1497-L1505

## Proposal

I don't think it makes sense to add `is_default` to the update params. Instead, I think we should have a make default endpoint, sort of analogous to image promotion and demotion. This endpoint would

- set `is_default = false` on the current default pool (if there is one)
- set `is_default = true` on the target pool

We probably also want a make-not-default endpoint, but it should probably only work on silo-scoped pools, otherwise instance create could fail due to there being no default pool.

### LINKED ISSUE #4003: IP pools: disallow deleting fleet-wide default pool?
We create a default pool at the fleet level at rack init. In https://github.com/oxidecomputer/omicron/pull/3985#discussion_r1310642701 and the comment below I mentioned the possibility that you could delete the default fleet-wide pool and therefore have nothing to fall back to if there is no silo default.

https://github.com/oxidecomputer/omicron/blob/01e730acf3efeb6ff2181d558cf373ded383a365/nexus/db-queries/src/db/datastore/ip_pool.rs#L75-L80

A simple way to prevent this is to require there to always be a fleet-wide default pool by:

- disallowing deletion of a fleet-wide default
- only allowing make default on a fleet-wide pool, but not remove default (see #4002)

However, it may be that we actually _want_ to allow operators to remove the default fleet-level pool, thereby forcing end users to either use a silo default (if one is configured) or always explicitly specify a pool at instance create. We will have to listen to customers on this based on their experience with IP pools as currently implemented.

### LINKED ISSUE #4006: IP pools: update project IP pools endpoint with correct logic
As part of #3985 I did not update the list and view endpoints intended for non-operators to list the IP pools available to them, which for now means those IP pools which are fleet-scoped and those matching their silo.

david-crespo:
Closes #2148
Closes #4002
Closes #4003 
Closes #4006

## Background

#3985 (and followups #3998 and #4007) made it possible to associate an IP pool with a silo so that instances created in that silo would get their ephemeral IPs from said pool by default (i.e., without the user having to say anything other than "I want an ephemeral IP"). An IP pool associated with a silo was not accessible for ephemeral IP allocation from other silos — if a disallowed pool was specified by name at instance create time, the request would 404.

However! That was the quick version, and the data model left much to be desired. The relation was modeled by adding a nullable `silo_id` and sort-of-not-really-nullable `is_default` column directly on the IP pool table, which has the following limitations (and there are probably more):

* A given IP pool could only be associated with at most one silo, could not be shared
* The concept of `default` was treated as a property of the pool itself, rather than a property of the _association_ with another resource, which is quite strange. Even if you could associate the pool with multiple silos, you could not have it be the default for one and not for the other
* There is no way to create an IP pool without associating it with either the fleet or a silo
* Extending this model to allow association at the project level would be inelegant — we'd have to add a `project_id` column (which I did in #3981 before removing it in #3985)

More broadly (and vaguely), the idea of an IP pool "knowing" about silos or projects doesn't really make sense. Entities aren't really supposed to know about each other unless they have a parent-child relationship.

### No such thing as fleet-scoped pool, only silo

Thanks to @zephraph for encouraging me to make this change. It is dramatically easier to explain "link silo to IP pool" than it is to explain "link resource (fleet or silo) to IP pool". The way to recreate the behavior of a single default pool for the fleet is to simply associate a pool with all silos. Data migrations ensure that existing fleet-scoped pools will be associated with all silos. There can only be one default pool for a silo, so in the rare case where pool A is a fleet default and pool B is default on silo S, we associate both A and B with S, but only B is made silo default pool. 

### API

These endpoints are added. They're pretty self-explanatory.

```
ip_pool_silo_link                        POST     /v1/system/ip-pools/{pool}/silos
ip_pool_silo_list                        GET      /v1/system/ip-pools/{pool}/silos
ip_pool_silo_unlink                      DELETE   /v1/system/ip-pools/{pool}/silos/{silo}
ip_pool_silo_update                      PUT      /v1/system/ip-pools/{pool}/silos/{silo}
```

The `silo_id` and `is_default` fields are removed from the `IpPool` response as they are now a property of the `IpPoolLink`, not the pool itself.

I also fixed the silo-scoped IP pools list (`/v1/ip-pools`) and fetch (`/v1/ip-pools/{pool}`) endpoints, which a) did not actually filter for the current silo, allowing any user to fetch any pool, and b) took a spurious `project` query param that didn't do anything.

### DB

The association between IP pools and fleet or silo (or eventually projects, but not here) is now modeled through a polymorphic join table called `ip_pool_resource`:

ip_pool_id | resource_type | resource_id | is_default
-- | -- | -- | --
123 | silo | 23 | true
123 | silo | 4 | false
~~65~~ | ~~fleet~~ | ~~FLEET_ID~~ | ~~true~~

Now, instead of setting the association with a silo or fleet at IP pool create or update time, there are separate endpoints for adding and removing an association. A pool can be associated with any number of resources, but a unique index ensures that a given resource can only have one default pool.

### Default IP pool logic

If an instance ephemeral IP or a floating IP is created **with a pool specified**, we simply use that pool if it exists and is linked to the user's silo.

If an instance ephemeral IP or a floating IP is created **without a pool unspecified**, we look for a default pool for the current silo. If there is a pool linked with the current silo with `is_default=true`, use that. Otherwise, there is no default pool for the given scope and IP allocation will fail, which means the instance create or floating IP create request will fail.

The difference introduced in this PR is that we do not fall back to fleet default if there is no silo default because we have removed the concept of a fleet-scoped pool.

### Tests and test helpers

This is the source of a lot of noise in this PR. Because there can no longer be a fleet default pool, we can no longer rely on that for tests. The test setup was really confusing. We assumed a default IP pool existed, but we still had to populate it (add a range) if we had to do anything with it. Now, we don't assume it exists, we create it and add a range and associate it with a silo all in one helper.

## What do customers have to do when they upgrade?

They should not _have_ to do anything at upgrade time.

If they were relying on a single fleet default pool to automatically be used by new silos, when they create silos in the future they will have to manually associate each new silo with the desired pool. We are working on ways to make that easier or more automatic, but that's not in this change. It is less urgent because silo creation is an infrequent operation.

If they are _not_ using the previously fleet default IP pool named `default` and do not want it to exist, they can simply delete any IP ranges it contains, unlink it from all silos and delete it. If they are not using it, there should not be any IPs allocated from it (which means they can delete it).


schema/crdb/7.0.0/up5.sql
```diff
@@ -0,0 +1,6 @@
+-- copy existing ip_pool-to-silo assocations into association table
+INSERT INTO ip_pool_resource (ip_pool_id, resource_type, resource_id, is_default)
+SELECT id, 'silo', silo_id, is_default
+FROM ip_pool
+WHERE silo_id IS NOT null
+  AND time_deleted IS null;
```
david-crespo, line 6: @smklein `integration_tests::schema::versions_have_idempotent_up` is mad at me for `up4` and `up5` because the `is_default` and `silo_id` columns get deleted from `ip_pool` in `up7`, so these inserts fail when run a second time. I can't figure out how to make this insert conditional on those columns existing. Moving `up6` and `up7` into their own separate migration 8.0.0 migration makes the test pass. Is that good, though? Almost feels like too easy of a workaround.

david-crespo:
Tested updated association post body with the TS generator and it is working exactly as desired:

```ts
/**
 * Parameters for associating an IP pool with a resource (fleet, silo)
 */
export type IpPoolAssociationCreate =
  | { isDefault: boolean; resourceType: 'silo'; silo: NameOrId }
  | { isDefault: boolean; resourceType: 'fleet' }
```

schema/crdb/9.0.0/up1.sql
```diff
@@ -0,0 +1,4 @@
+CREATE TYPE IF NOT EXISTS omicron.public.ip_pool_resource_type AS ENUM (
+    'fleet',
+    'silo'
```
zephraph, line 3:
```suggestion
    'silo',
    'fleet'
```

The ordering here was wrong which could break things. Tricky because I assume this was automatically generated. Seems like it changing the order is a bug?

nexus/tests/integration_tests/ip_pools.rs
```diff
@@ -275,69 +282,226 @@ async fn test_ip_pool_basic_crud(cptestctx: &ControlPlaneTestContext) {
         .expect("Expected to be able to delete an empty IP Pool");
 }
 
+/// The internal IP pool, defined by its association with the internal silo,
+/// cannot be interacted with through the operator API. CRUD operations should
+/// all 404 except fetch by name or ID.
+#[nexus_test]
+async fn test_ip_pool_service_no_cud(cptestctx: &ControlPlaneTestContext) {
```
zephraph, line 289: Is `cud` here short for create, update, delete?

schema/crdb/9.0.0/up1.sql
david-crespo: This isn't automatically generated, it's manually written to produce the same thing as what's in dbinit.sql. The bug is me!

nexus/tests/integration_tests/ip_pools.rs
david-crespo: Yes. This is the sort of shenanigans that happen when you are not around

schema/crdb/9.0.0/up1.sql
david-crespo: Actually, though... the fact that it's not caught by the schema tests might be a bad thing. We've recently made the tests a little less sensitive (https://github.com/oxidecomputer/omicron/pull/4288), but I would not have expected that to affect enum order.

nexus/types/src/external_api/views.rs
```diff
@@ -244,7 +244,22 @@ pub struct VpcRouter {
...
+#[serde(rename_all = "snake_case")]
+pub enum IpPoolResourceType {
+    Fleet,
+    Silo,
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema, PartialEq)]
+pub struct IpPoolResource {
+    // TODO: is including the pool ID redundant? it's convenient to have and
+    // makes this response a cohesive whole
```
zephraph, line 259:
```suggestion
```

I would keep the pool id. It doesn't hurt to have and it makes the response more useful by itself.

david-crespo:
@FelixMcFelix could you take a look at 70724cb41ad0603d5394452894298634ce0af568 and 69a482a665354d9845dcb215471b793f8dd789bb, which build on #4559?

FelixMcFelix:
Thanks for this; I'm happy with the additional changes around floating IP management in 70724cb and 69a482a. Integrating silo link/unlink into the FIP tests is particularly useful!

schema/crdb/dbinit.sql
```diff
@@ -1572,6 +1552,33 @@ CREATE UNIQUE INDEX IF NOT EXISTS lookup_pool_by_name ON omicron.public.ip_pool
 ) WHERE
     time_deleted IS NULL;
 
+-- The order here is most-specific first, and it matters because we use this
+-- fact to select the most specific default in the case where there is both a
+-- silo default and a fleet default. If we were to add a project type, it should
+-- be added before silo.
+CREATE TYPE IF NOT EXISTS omicron.public.ip_pool_resource_type AS ENUM (
+    'silo'
+);
```
david-crespo, line 1561: I've kept this in even though I took out the possibility of a fleet association to keep the structure general, but note that it does not leak out to the API views — this stuff is on the model, but the API view is called `IpPoolSilo` and just says `silo_id`, no `resource_id` or `resource_type`. I could make that change all the way down here and stop talking about "resources" and "associations". That would clean things up, but it would take more time.

nexus/test-utils/src/resource_helpers.rs
```diff
@@ -55,6 +57,41 @@ where
...
+            .expect_status(Some(status)),
+    )
+    .authn_as(AuthnMode::PrivilegedUser)
+    .execute()
+    .await
+    .unwrap()
+    .parsed_body::<HttpErrorResponseBody>()
+    .unwrap()
+}
+
```
david-crespo, line 94: These might look silly but boy do they clean up the tests!

nexus/db-queries/src/db/datastore/ip_pool.rs
```diff
@@ -256,6 +375,312 @@ impl DataStore {
...
+        authz_ip_pool: &authz::IpPool,
+        authz_silo: &authz::Silo,
+        is_default: bool,
+    ) -> UpdateResult<IpPoolResource> {
+        use db::schema::ip_pool_resource::dsl;
+
+        // TODO: correct auth check
+        opctx
+            .authorize(authz::Action::CreateChild, &authz::IP_POOL_LIST)
+            .await?;
```
david-crespo, line 470: Pretty much all the auth checks in this PR are weird. In practice I think this works simply because you need fleet admin to do any of this stuff, and this is effectively checking for fleet admin. But we may need another synthetic resource representing IP pool/silo links to make this a bit more clear.

nexus/db-queries/src/db/datastore/project.rs
```diff
@@ -355,23 +356,30 @@ impl DataStore {
...
-        // != excludes nulls so we explicitly include them
-        .filter(dsl::silo_id.ne(*INTERNAL_SILO_ID).or(dsl::silo_id.is_null()))
-        .filter(dsl::time_deleted.is_null())
+        .inner_join(ip_pool_resource::table)
+        .filter(
+            ip_pool_resource::resource_type
+                .eq(IpPoolResourceType::Silo)
+                .and(ip_pool_resource::resource_id.eq(silo_id)),
+        )
+        .filter(ip_pool::time_deleted.is_null())
```
david-crespo, line 382: The logic in this function is correct in the sense that this is giving the list of IP pools available in the current project, but it's silly to write it this way because we are not actually filtering by project. This is pretty old and was meant to be a placeholder for a future in which pools could be linked to projects. I would like to change this to simply be "current silo IP pools". But that is noise and probably belongs in a followup. This PR is, at least, not making anything worse with regard to this function.

david-crespo (commented):
I went through and made some notes I think will be helpful on review. I'd say the biggest remaining gap in the PR is around making sure we're doing the right authz checks. For the most part everything around manipulating and linking pools requires fleet admin, and we are effectively checking for that, but not in a great way. I think mostly we're doing checks for `CreateChild` or `ListChildren` on `IP_POOL_LIST`.

nexus/db-queries/src/db/datastore/project.rs
david-crespo: Ended up doing this in bf2905294649a991acb42f974288c6cac996e2c1 anyway. It was easy and it cut 100 lines net.

nexus/src/app/ip_pool.rs
```diff
@@ -56,24 +71,125 @@ impl super::Nexus {
...
+        opctx: &OpContext,
+        pool_lookup: &lookup::IpPool<'_>,
+        silo_link: &params::IpPoolSiloLink,
+    ) -> CreateResult<db::model::IpPoolResource> {
+        let (.., authz_pool) =
+            pool_lookup.lookup_for(authz::Action::Modify).await?;
+        let (silo,) = self
+            .silo_lookup(&opctx, silo_link.silo.clone())?
+            .lookup_for(authz::Action::Read)
+            .await?;
```
david-crespo, line 134: Well this can't be right. Plus it's different from the perm check on dissociate.

nexus/src/app/project.rs
```diff
@@ -149,40 +147,4 @@ impl super::Nexus {
...
-                    .ip_pool_name(Name::ref_cast(name));
-                Ok(pool)
-            }
-            NameOrId::Id(id) => {
-                let pool =
-                    LookupPath::new(opctx, &self.db_datastore).ip_pool_id(*id);
-                Ok(pool)
-            }
-        }
-    }
```
david-crespo, line 187: this was the same as regular ip_pool_lookup. deleted, use the latter instead. see `silo_ip_pool_fetch`

nexus/src/app/project.rs
```diff
@@ -149,40 +147,4 @@ impl super::Nexus {
...
-        project_lookup: &lookup::Project<'_>,
-        pagparams: &PaginatedBy<'_>,
-    ) -> ListResultVec<db::model::IpPool> {
-        let (.., authz_project) =
-            project_lookup.lookup_for(authz::Action::ListChildren).await?;
-
-        self.db_datastore
-            .project_ip_pools_list(opctx, &authz_project, pagparams)
-            .await
-    }
```
david-crespo, line 165: moved to `nexus/src/app/ip_pools.rs`

nexus/src/app/ip_pool.rs
```diff
@@ -56,24 +71,125 @@ impl super::Nexus {
...
+            self.ip_pool_lookup(opctx, pool)?.fetch().await?;
+
+        // 404 if no link is found in the current silo
+        let link = self.db_datastore.ip_pool_fetch_link(opctx, pool.id()).await;
+        if link.is_err() {
+            return Err(authz_pool.not_found());
+        }
+
+        Ok(pool)
+    }
```
david-crespo, line 107: I don't think these silo-scoped ones (`silo_ip_pool_fetch` and `silo_ip_pools_list`) have the right permissions for regular silo users to get anything from them. Need to fix that.

nexus/db-queries/src/db/datastore/external_ip.rs
```diff
@@ -76,22 +76,18 @@ impl DataStore {
...
-                // exist. If the pool has no silo, it's fleet-scoped and can
-                // always be used.
-                let authz_silo_id = opctx.authn.silo_required()?.id();
-                if let Some(pool_silo_id) = pool.silo_id {
-                    if pool_silo_id != authz_silo_id {
-                        return Err(authz_pool.not_found());
-                    }
+                // If this pool is not linked to the current silo, 404
+                if self.ip_pool_fetch_link(opctx, pool.id()).await.is_err() {
+                    return Err(authz_pool.not_found());
```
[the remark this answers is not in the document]
zephraph: I was wondering if you'd need to pass through the `opctx_alloc` here to have permissions, but I see the only permissions check you're doing inside `ip_pool_fetch_link` is for the silo. Something I learned from the quotas work is that if you explicitly need higher permissions as a side effect of a lower perm thing you can use `self.opctx_alloc` inside the datastore and pass that as OpContext for elevated permissions. I wonder if that would be better with added authz checks in `ip_pool_fetch_link`.

nexus/db-queries/src/db/datastore/external_ip.rs
```diff
@@ -147,24 +143,29 @@ impl DataStore {
     ) -> CreateResult<ExternalIp> {
         let ip_id = Uuid::new_v4();
 
-        let pool_id = match params.pool {
+        // TODO: NameOrId resolution should happen a level higher, in the nexus function
+        let (.., authz_pool, pool) = match params.pool {
```
zephraph, line 147: Maybe we could just create a separate issue about this. I think there's a few places in the networking APIs around BGP that have the same thing.

nexus/db-queries/src/db/datastore/ip_pool.rs
```diff
@@ -46,29 +56,108 @@ impl DataStore {
...
+    /// the pool with that silo
+    pub async fn ip_pool_fetch_link(
+        &self,
+        opctx: &OpContext,
+        ip_pool_id: Uuid,
+    ) -> LookupResult<IpPoolResource> {
+        use db::schema::ip_pool;
+        use db::schema::ip_pool_resource;
+
+        let authz_silo = opctx.authn.silo_required()?;
```
zephraph, line 142: why no `internal_context("...")` call here? I think that gets spit out in the logs, can be useful.

nexus/db-queries/src/db/datastore/ip_pool.rs
```diff
@@ -256,6 +419,311 @@ impl DataStore {
...
+        .select(IpPoolResource::as_select())
+        .load_async::<IpPoolResource>(
+            &*self.pool_connection_authorized(opctx).await?,
+        )
+        .await
+        .map_err(|e| public_error_from_diesel(e, ErrorHandler::Server))
+    }
+
+    // TODO: should this error on conflict instead of updating?
+    pub async fn ip_pool_associate_resource(
```
zephraph, line 448: Minor nit, but should we use `link` here?

nexus/db-queries/src/db/datastore/ip_pool.rs
```diff
@@ -256,6 +419,311 @@ impl DataStore {
...
+        &self,
+        opctx: &OpContext,
+        ip_pool_resource: IpPoolResource,
+    ) -> CreateResult<IpPoolResource> {
+        use db::schema::ip_pool_resource::dsl;
+        opctx
+            .authorize(authz::Action::CreateChild, &authz::IP_POOL_LIST)
+            .await?;
```
zephraph, line 456: Perhaps unnecessary, but you could also check for modify on the silo? I assume mostly that if you have one you have the other but 🤷‍♂️

nexus/db-queries/src/db/datastore/ip_pool.rs
david-crespo: Yes. And it should really be called `link_silo`

nexus/db-queries/src/db/datastore/ip_pool.rs
zephraph: So you have a few options in these. The _correct_ way may likely be to create a dedicate authz resource representing pool/silo links and auth check that. Alternatively I think if you just checked modify on both resources that would be sufficient. I wouldn't do create child though.

nexus/db-queries/src/db/datastore/ip_pool.rs
```diff
@@ -256,6 +419,311 @@ impl DataStore {
...
+                    .execute_async(&conn)
+                    .await;
+                if let Err(e) = unset_default {
+                    return Err(TxnError::CustomError(
+                        IpPoolResourceUpdateError::FailedToUnsetDefault(e),
+                    ));
+                }
+            }
+
+            // TODO: test that this errors if the link doesn't exist already
```
zephraph, line 581: Don't forget / delete this if it's done?

nexus/db-queries/src/db/datastore/ip_pool.rs
```diff
@@ -256,6 +419,311 @@ impl DataStore {
...
+    /// IPs allocated from the pool in the associated silo
+    pub async fn ip_pool_dissociate_resource(
+        &self,
+        opctx: &OpContext,
+        association: &IpPoolResourceDelete,
+    ) -> DeleteResult {
+        use db::schema::ip_pool_resource;
+        opctx
+            .authorize(authz::Action::CreateChild, &authz::IP_POOL_LIST)
+            .await?;
```
zephraph, line 708: This doesn't feel like the right check. I think again, either creating an authz type for the relationship or checking modify on one or both of the resources being changed.

schema/crdb/21.0.0/up5.sql
```diff
@@ -0,0 +1,12 @@
+-- Preemptively turn any former fleet defaults into non-defaults if there's
+-- going to be a silo conflicting with it after up6
+UPDATE omicron.public.ip_pool_resource AS ipr
+SET is_default = false
+FROM omicron.public.ip_pool as ip
+WHERE ipr.resource_type = 'silo' -- technically unnecessary because there is only silo
+  AND ipr.is_default = true
+  AND ip.is_default = true -- both being default is the conflict being resolved
+  AND ip.silo_id IS NOT NULL
```
zephraph, line 9: Should this be `AND ip.silo_id IS null` b/c that's what indicates a fleet ip pool?

david-crespo: Nope, we're looking at silo pools in `ip_pool`, lining them up against entries in `ip_pool_resource` that are type=silo but came from a fleet pool in up4, preferring the former as a default.

nexus/db-queries/src/db/datastore/external_ip.rs
zephraph: I don't think you were planning on it, but let's save the fix for this for another PR.

nexus/db-queries/src/db/datastore/ip_pool.rs
```diff
@@ -91,23 +182,47 @@ impl DataStore {
...
+                public_error_from_diesel_lookup(
+                    e,
+                    ResourceType::IpPool,
+                    &lookup_type,
+                )
+            })
+            .map(|ip_pool| {
+                let authz_pool =
+                    authz::IpPool::new(authz::FLEET, ip_pool.id(), lookup_type);
+                (authz_pool, ip_pool)
```
zephraph, line 224: This is interesting, I haven't noticed where we've done something like this explicitly in the data store before. I'm not sure why the authz / resource tuple was necessary.

nexus/db-queries/src/db/datastore/ip_pool.rs
```diff
@@ -422,97 +899,168 @@ impl DataStore {
...
-            .expect("Failed to create silo default IP pool");
+            .expect("Failed to make IP pool default for silo");
 
-        // now when we ask for the default pool, we get the one we just made
-        let ip_pool = datastore
+        // now when we ask for the default pool again, we get that one
+        let (authz_pool1_for_silo, ip_pool) = datastore
             .ip_pools_fetch_default(&opctx)
             .await
             .expect("Failed to get silo's default IP pool");
```
zephraph, line 973: I see, because this doesn't have a normal lookup implemented this fetch implementation acts as one. Seems fine.

nexus/db-queries/src/db/datastore/ip_pool.rs
zephraph: Ahh, I found it below. This is a non-standard lookup so the id/name generated lookups that hang off of nexus wouldn't be available and thus its existence.

nexus/src/app/ip_pool.rs
zephraph: I think for now it certainly seems fine. We could go through and create a synthetic resource for the link which is probably the correct means of handling this. Worth an issue to investigate that.

nexus/db-queries/src/db/datastore/ip_pool.rs
david-crespo: Right. An edge case that doesn't really fit in the look up by name, id, or composite id model. Arguably it's a composite ID, which is why I'm using it as a sort of catchall for this.

nexus/db-queries/src/db/datastore/ip_pool.rs
david-crespo: Yeah, definitely. I need to change the signature to take the `authz_silo` and the `authz_pool`.

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -1414,6 +1394,128 @@ async fn ip_pool_update(
...
+/// Make an IP pool default or not-default for a silo
+///
+/// When a pool is made default for a silo, any existing default will remain
+/// linked to the silo, but will no longer be the default.
+#[endpoint {
+    method = PUT,
+    path = "/v1/system/ip-pools/{pool}/silos/{silo}",
+    tags = ["system/networking"],
+}]
+async fn ip_pool_silo_update(
```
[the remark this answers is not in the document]
david-crespo: Yeah, it could be `ip_pool_silo_set_default`. There are no other properties of the link that can be updated. Another option would be `ip_pool_silo_link_update` to make clearer it's about updating the link and not the pool or the silo.

zephraph (approved):
Overall I'm feeling pretty good about the state of this PR. There's undoubtedly areas that I've missed in review but it feels holistically better than where we where before it. 

Given we have time before the next release I feel it would be helpful to land this soon and be able to iterate on it as we feel it out through usage.

PR #4265, opened 2023-10-11
plumb customer-configured DNS into instances

iliana:
Currently [OPTE hardcodes DNS servers to 8.8.8.8](https://github.com/oxidecomputer/opte/issues/390), and we would prefer to not do this.

When setting up the rack, customers configure DNS servers. Currently this is added to Nexus's deployment configuration, so that Nexus can resolve external domain names for SAML. This change plumbs these DNS servers through to sled-agent and OPTE.

For non-instance OPTE ports, `DhcpCfg::default()` is used, which is equivalent to "no hostname, no domain name, no DNS servers, no search domains".

This is not the final state of the DHCP work we want to do; in talking with @rmustacc and @rcgoodfellow we agree that implementing this minimal DHCP option set is the correct thing to do urgently; and that we still need to design the broader inter-instance networking picture (whether we want an instance-facing recursive resolver on the rack and whether it should resolve domain names for other instances with VPC IPs, what those hostnames and search domains should be, and how customers can modify these settings).


iliana:
@gjcolombo FYI — requested a review from you because this was written prior to #4194 landing, and I did sort of the bare minimum merge from main just now. I'd like to hear if you think there's a more logical spot to put these changes to Nexus / Sled Agent.

sled-agent/src/instance.rs
```diff
@@ -610,6 +615,37 @@ impl Instance {
             zone_bundler,
         } = services;
 
+        let mut dhcp_config = DhcpCfg {
+            hostname: Some(
+                hardware
+                    .properties
+                    .hostname
```
gjcolombo, line 622: Not necessarily for this PR, but I think it might be worth an filing issue: I'm guessing that if this conversion fails it implies that the hostname in the instance create parameters was completely bogus such that this instance will never be able to start. If that's so we should see about rejecting invalid hostnames in `Nexus::project_instance_create` so that we reject bad hostnames up front instead of letting the create go through.

gjcolombo (approved):
The new instance plumbing LGTM (with my thanks for spending time picking through the merge). I didn't look at the OPTE or service plumbing at all.

sled-agent/src/instance.rs
iliana: I haven't gone and verified this is the case, but I _believe_ that our restrictions on the `Name` type make them all valid single-label hostnames. It would be good to prove this out though (or use a single consistent type across Omicron and OPTE).

iliana: Oh! Well. I'll see if that's not too bad to fix along with this PR.

iliana:
Two follow-up items I'd like to do are:
- instance hostname validation, as noted by Dave above
- verify resolv.conf contains our DNS servers in the end-to-end instance launch test

But we may want to land this before those tasks are done to avoid blocking the release.

luqmana (approved):

luqmana:
> But we may want to land this before those tasks are done to avoid blocking the release. I can turn these into follow-up issues if we like.

I think that makes sense. I wouldn't say those are blockers.

PR #4273, opened 2023-10-12
Combine host and trampoline OS images into one CI job

citrus-it:


.github/buildomat/jobs/host-image.sh
```diff
@@ -1,11 +1,12 @@
...
 #:
-#: name = "helios / build OS image"
+#: name = "helios / build OS images"
 #: variety = "basic"
 #: target = "helios-2.0"
 #: rust_toolchain = "1.72.1"
 #: output_rules = [
-#:	"=/work/helios/image/output/os.tar.gz",
+#:	"=/work/helios/upload/os-host.tar.gz",
+#:	"=/work/helios/upload/os-trampoline.tar.gz",
```
jclulow, line 9: Seeing as we're moving them around anyway, we could just as well:

```
mkdir -p /helios
```

and put them in there.  They don't need to be under `/work` etc.

citrus-it: It doesn't hurt to put them under `/work` though, I think. Then there is consistency in the tuf-repo script with the other things it's picking up from jobs.

It's a read-only root, so `/helios` doesn't work.

jclulow: It's definitely not a read-only root.  It's just a regular VM with a mutable disk installation; that's why you can use `pkg install` etc.  e.g., see https://github.com/oxidecomputer/pilot/blob/main/.github/buildomat/jobs/bins.sh

citrus-it: My mistake,  `pfexec` was what I needed there.

jgallagher (approved):

iliana (approved):

PR #4277, opened 2023-10-13
Update virtual provisioning counters on instance stop/start

### LINKED ISSUE #4257: Move virtual provisioning accounting for instances to instance start saga
This is a follow-up to #4194. Once that lands, instances should only count as 'provisioned' if they're actually running. To do this, we need to:

- move the virtual provisioning steps in the create saga to the start saga
- change the way instance virtual provisioning counters are keyed (today we use instance ID; now we'll need to use Propolis IDs; this is the virtual-provisioning equivalent of #2839)
- make sure these charges get cleaned up when instances are stopped (same place as we clean up sled resource reservations)

See the discussion in RFD 427 for more context.

gjcolombo:
Only charge virtual provisioning collections for instances when those instances are running. Charges are taken in the instance start saga and dropped when a sled agent tries to transition an instance to a stopped state. Unlike sled resource charges, provisioning charges are tied to instance states, not to VMM lifetimes. This ensures that a user is not charged twice for an instance (e.g. for quota management purposes) while the instance is migrating. See RFD 427 for more discussion.

Also fix a small idempotency issue in the cleanup path for VMM resources.

Tests: updated integration tests; manually checked virtual provisioning table values in a dev cluster & checked the values on the utilization graphs.

Fixes #4257.


david-crespo:
Huge, yet small! This is awesome.

nexus/src/app/instance.rs
```diff
@@ -1311,6 +1313,31 @@ impl super::Nexus {
...
+        // As with updating networking state, this must be done before
+        // committing the new runtime state to the database: once the DB is
+        // written, a new start saga can arrive and start the instance, which
+        // will try to create its own virtual provisioning charges, which will
+        // race with this operation.
+        if new_runtime_state.instance_state.propolis_id.is_none()
+            && new_runtime_state.instance_state.gen
+                >= db_instance.runtime().gen.0
+        {
+            self.db_datastore
```
smklein, line 1330: This check is made based off of `new_runtime_state` (from the sled agent) as well as `db_instance.runtime()` (from the DB), but both are cached -- couldn't **both** be out-of-date?

For example:

1. External API: Create instance, start it. Begin the call to "stop", which should send a request to stop down to the sled.
2. Sled Agent -> Nexus: Calls this function, indicating the instance should be stopped. Let's suppose sled agent thinks gen == N, the db thinks gen == N. But maybe calling `ensure_updated_instance_network_config` yields for a really long time.
3. Sled Agent -> Nexus: Calls this function again, from a retry-loop, but this time it succeeds. The instance is now stopped.
4. External API: Start the instance again. This re-provisions the resources used by the instance...
5. ... and, while the instance is running, the request from (2) comes back. It'll see a state of the world from before (3) and (4), so it'll call the `virtual_provisioning_collection_delete_instance` function. Now the accounting is wrong! This **is** a running instance, but we would have deleted the resources it's currently using.

smklein: I think this is related to what I was touching on here: https://github.com/oxidecomputer/omicron/pull/4194#discussion_r1348088206

If we're associating these resources with VMMs (AKA, "what is actually running") rather than Instances (the opaque thing that exists even without a propolis), should we alter the calls to `virtual_provisioning_collection_{create, delete}_instance` to act on VMM UUIDs instead, rather than instance UUIDs?

That way, it wouldn't be possible to do a "double delete" -- once a VMM is gone, the UUID for it should never be re-used/re-added, so repeated/delayed requests to delete should have no further effect after the first successful call.

gjcolombo: Hmm. Let me think about this one a little more--the race you've identified is definitely a problem. I'll mull it over to see if there's a good way to resolve it without dipping into transaction-land (such that we could say "only proceed with the deletion if the instance generation number is less than the one I think I'm trying to apply" or suchlike).

It's true that VMMs don't share that problem because they can't come back from a terminal state, but I want to be careful about charging counters on a per-VMM basis if those counters are going to be used for quota management purposes (which AIUI is part of the discussion in RFD 427). For example, suppose I have a quota of 32 vCPUs on my project; I start an instance with 16 vCPUs; the rack operator starts an upgrade that migrates my instance; then I try to start a second instance with 16 vCPUs; if we track per-VMM, this will fail because my quota's exhausted--but what do you *mean* it's exhausted? I only have the one running instance! (cc @askfongjojo to help weigh in on this part; if we're OK with charging on a per-VMM basis even though this can cause transient double-counting, then switching to VMM IDs is a much simpler option here.)

nexus/src/app/sagas/instance_start.rs
```diff
@@ -96,6 +101,7 @@ impl NexusSaga for SagaInstanceStart {
 
         builder.append(alloc_server_action());
         builder.append(alloc_propolis_ip_action());
+        builder.append(add_virtual_resources_action());
         builder.append(create_vmm_record_action());
         builder.append(mark_as_starting_action());
```
gjcolombo, line 106: This ordering is wrong. The saga should only add virtual resources once it's successfully marked the instance as starting, i.e., it should get through the "only one start can proceed at a time" interlock before charging for anything. Will fix in the next push.

nexus/src/app/instance.rs
gjcolombo: Fixed in 16b7fe6 by putting a sub-select into the instance provisioning counter deletion query that makes it apply only if the caller has furnished a sufficiently new generation number.

nexus/src/app/sagas/instance_start.rs
gjcolombo: Fixed in 16b7fe6.

nexus/src/app/instance.rs
askfongjojo: Transient double-counting is probably fine if it's indeed transient (i.e. seconds, not minutes). But if we foresee failure modes that can cause in-progress migration to double count for an extended period of time, it'll be better for the system rather than the user to "absorb" the migration-in-transit usage.

smklein (approved):
Thanks for the update - plumbing a sled-provided generation number seems like a valid way to mitigate the race I identified.

gjcolombo:
I've merged the latest main and retested on a dev cluster, and things still look OK, so I'm going to go ahead and land this one.

PR #4285, opened 2023-10-17
Add type conversions for EarlyNetworkConfig upgrade

andrewjstone:
This commit adds deserialization and conversion support for v0 and v1 of EarlyNetworkConfig types. Future commits will add support for writing the v1 version back to the bootstore ledger if necessary. Unit tests for conversion also coming soon.


common/src/api/internal/shared.rs
```diff
@@ -82,6 +90,41 @@ pub struct RackNetworkConfig {
...
 
+/// Deprecated, use `RackNetworkConfig` instead. Cannot actually deprecate due to
+/// <https://github.com/serde-rs/serde/issues/2195>
+///
+/// Our first version of `RackNetworkConfig`. If this exists in the bootstore, we
+/// upgrade out of it into `RackNetworkConfigV1` or later versions if possible.
+/// Initial network configuration
+///
+// TODO(AJS): Should this actually exist in the OpenAPI spec or should we just
+// hide the old version inside sled-agent/src/bootstrap/early_networking.rs
```
jgallagher, line 101: I think since we're not yet doing multiple-API-versions-supported-simultaneously, I would vote for "hide it from OpenAPI" so that the spec only shows what should currently be provided / returned.

common/src/api/internal/shared.rs
```diff
@@ -68,9 +68,17 @@ pub struct SourceNatConfig {
...
 
+// We alias [`RackNetworkConfig`] to the current version of the protocol, so
+// that we can convert between versions as necessary.
+pub type RackNetworkConfig = RackNetworkConfigV1;
+
 /// Initial network configuration
+///
+/// TODO(AJS): It's unclear if this should be serde renamed to `RackNetworkConfig`
+/// I *think* it's useful to have the version be explicit in the name, but I'm open
+/// to discussion.
```
jgallagher, line 79: I'm not sure serde cares about struct names, does it? At least I think that's true for JSON; if this is a top-level object, only the field names show up, and if this is embedded in another struct or enum, it's named based on that container's field/variant.

Maybe schemars would look at `serde(rename)` rules to pick a name in the OpenAPI spec though? I don't have strong feelings either way on what it's named there.

sled-agent/src/bootstrap/early_networking.rs
```diff
@@ -568,25 +569,67 @@ fn retry_policy_switch_mapping() -> ExponentialBackoff {
...
-    /// A copy of the initial rack network configuration when we are in
-    /// generation `1`.
+    // Rack network configuration as delivered from RSS or Nexus
     pub rack_network_config: Option<RackNetworkConfig>,
 }
 
+// The first production version of the EarlyNetworkConfig.
+//
+// If this version is in the bootstore than we need to convert it to
+// `EarlyNetworkConfig`. We intend to convert this data on
```
jgallagher, line 609: on what? :)

common/src/api/internal/shared.rs
andrewjstone: I added this comment because this rewrote the schema. I still think I prefer V1 in the schema explicitly though. WDYT?

common/src/api/internal/shared.rs
andrewjstone: Will do!

andrewjstone:
After discussion with @jgallagher, it seems unnecessary and potentially dangerous to write the updated schema back to the ledger since we can deterministically convert from `EarlyNetworkConfigV0` to `EarlyNetworkConfigV1`. Therefore, after some unit tests and small cleanup this should be good to go on the sled-agent side.

common/src/api/internal/shared.rs
jgallagher: Leaving the V1 in the schema explicitly seems fine to me :+1:

sled-agent/src/bootstrap/early_networking.rs
andrewjstone: Line removed in [a992afa](https://github.com/oxidecomputer/omicron/pull/4285/commits/a992afac9c5cb9622da4b8368a91d810cad58cc9)

sled-agent/src/bootstrap/early_networking.rs
```diff
@@ -595,13 +639,77 @@ impl From<EarlyNetworkConfig> for bootstore::NetworkConfig {
...
+    pub infra_ip_last: Ipv4Addr,
+    /// Uplinks for connecting the rack to external networks
+    pub uplinks: Vec<UplinkConfig>,
+}
+
+impl RackNetworkConfigV0 {
+    /// Convert from `RackNetworkConfigV0` to `RackNetworkConfigV1`
+    ///
+    /// We cannot use `From<RackNetworkConfigV0> for `RackNetworkConfigV1`
+    /// because the `rack_network` field does not exist in `RackNetworkConfigV0`
```
andrewjstone, line 695:
```suggestion
    /// because the `rack_subnet` field does not exist in `RackNetworkConfigV0`
```

rcgoodfellow (approved):

Issue #4286, opened 2023-10-17
dbinit is getting too order dependent


smklein:
Following up from some issues observed by @internet-diglett -- we're observing that the `dbinit_equals_sum_of_all_up` test is failing more frequently (not flake, just with new PRs) for schema changes which cause constraint changes.

This was referenced by @gjcolombo in https://github.com/oxidecomputer/omicron/tree/main/schema/crdb#adding-tables-and-views , but it's reproducible in some surprising cases.

For example, right now, with `main` at ee8d93b39399c8f764f67101826bc20c89003942 , if I add the following:

```sql
CREATE TABLE IF NOT EXISTS omicron.public.foo (
    id UUID CONSTRAINT pk PRIMARY KEY
);
```

To both `8.0.0/up.sql`, and also to the end of `dbinit.sql`, I see test failures comparing table constraints.

```
thread 'integration_tests::schema::dbinit_equals_sum_of_all_up' panicked at 'assertion failed: `(left == right)`'
  left: `"[Row { values: [NamedSqlValue { column: \"constraint_catalog\", value: Some(String(\"omicron\")) }, NamedSqlValue { column: \"constraint_schema\", value: Some(String(\"public\")) }, NamedSqlValue { column: \"c..."` (truncated)
 right: `"[Row { values: [NamedSqlValue { column: \"constraint_catalog\", value: Some(String(\"omicron\")) }, NamedSqlValue { column: \"constraint_schema\", value: Some(String(\"public\")) }, NamedSqlValue { column: \"c..."` (truncated)

Differences (-left|+right):
             NamedSqlValue {
                 column: "constraint_name",
                 value: Some(
                     String(
-                        "4101115737_243_1_not_null",
+                        "4101115737_242_1_not_null",
                     ),
                 ),
             },
             NamedSqlValue {

', nexus/tests/integration_tests/schema.rs:579:9
```

internet-diglett:
It *seems* that there may be some changes in the `7.0.0` migration that causes the value generated for the `constraint_name` to be off by a predictable value. It may be due to a `NOT NULL` constraint being dropped, or another column being dropped or some combination of statements. Tests pass after moving the migration from `8.0.0` to `6.1.0`, and adjusting the placement of the statements in `dbinit.sql` accordingly.

david-crespo:
[...]

PR #4288, opened 2023-10-17
Make dbinit.sql slightly less order-dependent

smklein:
This PR does two things:

1. It avoids checking for the order-dependent NOT NULL constraint in information schema, opting to parse the output of `SHOW CONSTRAINTS` and  [information_schema.columna.is_nullable](https://www.cockroachlabs.com/docs/v23.1/information-schema#columns) instead.
2. It re-orders the db_metadata creation to the end of dbinit.sql (mostly for clarity)

Fixes https://github.com/oxidecomputer/omicron/issues/4286


gjcolombo:
> It re-orders the db_metadata creation to the end of dbinit.sql (mostly for clarity)

Hooray!

I've only skimmed the PR but will ask anyway: If this is now legal, is the guideline in the schema README to add new tables only to the end of dbinit.sql still relevant? (If it's not, we should remove it and consider replacing it with guidance that db_metadata should always be last.)

smklein:
> > It re-orders the db_metadata creation to the end of dbinit.sql (mostly for clarity)
> 
> Hooray!
> 
> I've only skimmed the PR but will ask anyway: If this is now legal, is the guideline in the schema README to add new tables only to the end of dbinit.sql still relevant? (If it's not, we should remove it and consider replacing it with guidance that db_metadata should always be last.)

Sure thing, I just pushed a small update to the schema/crdb README.

david-crespo (approved):
Looks good. I confirmed this change makes the schema tests pass in https://github.com/oxidecomputer/omicron/pull/4261.

PR #4289, opened 2023-10-17
[wicketd] Accept TUF repos with RoT archives signed with different keys

jgallagher:
As of this PR, wicketd will (a) accept TUF repos containing multiple RoT archives for the same board target (e.g., multiple gimlet RoT images), and when performing a mupdate, it will ask the RoT for its currently-active CMPA and CFPA pages and search for an RoT archive that matches.

After this is deployed to all fielded systems, we'll be able to drop the `-rot-staging-dev` and `-prod-rel` TUF repos from CI, and only build a single TUF repo with all RoT images. This PR adds a new `-rot-all` TUF repo publishing step but does not remove the old ones, as we'll need them to update into this version of wicketd.

### DIFF
--- a/.github/buildomat/jobs/tuf-repo.sh
+++ b/.github/buildomat/jobs/tuf-repo.sh
@@ -25,6 +25,21 @@
 #: job = "helios / build trampoline OS image"
 #:
 #: [[publish]]
+#: series = "rot-all"
+#: name = "repo.zip.parta"
+#: from_output = "/work/repo-rot-all.zip.parta"
+#:
+#: [[publish]]
+#: series = "rot-all"
+#: name = "repo.zip.partb"
+#: from_output = "/work/repo-rot-all.zip.partb"
+#:
+#: [[publish]]
+#: series = "rot-all"
+#: name = "repo.zip.sha256.txt"
+#: from_output = "/work/repo-rot-all.zip.sha256.txt"
+#:
+#: [[publish]]
 #: series = "rot-prod-rel"
 #: name = "repo.zip.parta"
 #: from_output = "/work/repo-rot-prod-rel.zip.parta"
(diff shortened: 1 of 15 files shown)


.github/buildomat/jobs/tuf-repo.sh
```diff
@@ -168,6 +183,38 @@ caboose_util_rot() {
 }
 
 SERIES_LIST=()
+
+# Create an initial `manifest-rot-all.toml` containing the SP images for all
+# boards. While we still need to build multiple TUF repos,
+# `add_hubris_artifacts` below will append RoT images to this manifest (in
+# addition to the single-RoT manifest it creates).
+prep_rot_all_series() {
```
jgallagher, line 191: @iliana You probably care most about the changes here. They feel pretty janky, but (a) it wasn't clear how to fit this into the existing `add_hubris_artifacts` and (b) once we don't need the separate staging/dev and prod/rel TUF repos, we can probably drop most of what makes it ugly. Happy to do any cleanup if there's a nicer way to do this for now, though!

iliana (approved):

.github/buildomat/jobs/tuf-repo.sh
iliana: No more janky than the rest of the script. :) I am excited to clean this up after all known racks have updated beyond this PR!

wicketd/src/update_tracker.rs
```diff
@@ -1582,6 +1582,127 @@ impl UpdateContext {
...
+                })?;
+            let archive = RawHubrisArchive::from_vec(image).map_err(|err| {
+                UpdateTerminalError::FailedFindingSignedRotImage {
+                    error: anyhow::Error::new(err).context(format!(
+                        "failed to read hubris archive for {:?}",
+                        artifact.id
+                    )),
+                }
+            })?;
+            match archive.verify(&cmpa, &cfpa) {
```
andrewjstone, line 1676: It's handy that this function exists! 😄

andrewjstone (approved):

PR #4291, opened 2023-10-18
initial inventory for automated update

davepacheco:
This PR implements the first round of hardware/software inventory for automated update.  See [RFD 433](https://rfd.shared.oxide.computer/rfd/433) for background.  There's a summary of the new data model in [dbinit.sql](https://github.com/oxidecomputer/omicron/blob/0f01172693b25e7b513ed02a2f743e0eec530949/schema/crdb/dbinit.sql#L2512-L2567).

I'm sorry this change is so big.  Here are the key pieces:

* in `nexus/types`: type related to software inventory (used in a few places)
* in `schema/crdb` and `nexus/db-model`: database schema/model described in RFD 433
* in `nexus/db-queries`: datastore queries to insert or delete an entire inventory `Collection`
* in `nexus/inventory`: new crate with Collector and builder interface.  This crate only collects inventory -- it doesn't do anything with the database.
* in `nexus/src/app/background`: a new background task that uses these other pieces to collect inventory, write it to the database, and clean up old collections
* `omdb` support for showing inventory data from the database
* implementation of all this for service processors, roots of trust, and cabooses, all fetched from MGS

Still to-do here:

* [ ] add path for schema update (I only updated dbinit.sql so far)
* [ ] automated tests
* [ ] maybe: replace `inv_caboose` table with NULLable fields on `inv_service_processor` and `inv_root_of_trust`
* [ ] what chicken switches do we want to add?  I'm thinking it'd be useful to have a way to turn this thing off altogether if needed but I'm not sure the best way to do that.  It'd be easy to add a Nexus config option but I'm not sure how easy that is to use on a deployed system.

What's not here (and will be in future PRs, not this one):

* any other hardware/software inventory, including anything from sled agent: host OS versions, physical disks, Omicron zones, other hardware, etc.
* Nothing uses any of this data yet.

Some other stuff came along for the ride here.  I'm happy to separate these if that's useful but they're each pretty small:

* `omicron-dev run-all`, as well as all tests that set up a `ControlPlaneTestContext`, now run a Management Gateway Service backed by the same simulated SPs used in the existing MGS tests.  This was easy to do, convenient for future inventory work, and it was necessary to test the `omdb` changes.
* I noticed while debugging some SQL queries that `omdb` does not call `usdt::register_probes()`, so we don't have (for example) the diesel-dtrace probes in `omdb`.  I added a call in `pool.rs` to cover these.  This isn't quite once per process, but it's close, and ensures that anybody who uses our database layer will get these probes.  This was a one line change.
* In datastore/mod.rs, I removed `pool_authorized()` because it was non-`pub` and there was only one caller and I found the name confusing.

---

Here's some example output.  I used `omicron-dev run-all` to get everything going:

```console
$ cargo run --bin=omicron-dev -- run-all
...
    Finished dev [unoptimized + debuginfo] target(s) in 4m 36s
log file: /dangerzone/omicron_tmp/omicron-dev-omicron-dev.20844.0.log
note: configured to log to "/dangerzone/omicron_tmp/omicron-dev-omicron-dev.20844.0.log"
DB URL: postgresql://root@[::1]:49896/omicron?sslmode=disable
DB address: [::1]:49896
log file: /dangerzone/omicron_tmp/omicron-dev-omicron-dev.20844.1.log
note: configured to log to "/dangerzone/omicron_tmp/omicron-dev-omicron-dev.20844.1.log"
omicron-dev: nexus internal API:    [::1]:12221
omicron-dev: cockroachdb pid:       21010
omicron-dev: cockroachdb URL:       postgresql://root@[::1]:49896/omicron?sslmode=disable
omicron-dev: cockroachdb directory: /dangerzone/omicron_tmp/.tmpvU7h5N
omicron-dev: internal DNS HTTP:     http://[::1]:56273
omicron-dev: internal DNS:          [::1]:37290
omicron-dev: external DNS name:     oxide-dev.test
omicron-dev: external DNS HTTP:     http://[::1]:36365
omicron-dev: external DNS:          [::1]:55005
omicron-dev:   e.g. `dig @::1 -p 55005 test-suite-silo.sys.oxide-dev.test`
omicron-dev: management gateway:    http://[::1]:57254
omicron-dev: silo name:             test-suite-silo
omicron-dev: privileged user name:  test-privileged
```

Here's `omdb nexus background-tasks show` for the new "inventory" task:

```console
$ export OMDB_DNS_SERVER=[::1]:37290
$ omdb nexus background-tasks show
note: using Nexus URL http://[::1]:12221
...
task: "inventory_collection"
  configured period: every 10m
  currently executing: no
  last completed activation: iter 3, triggered by an explicit signal
    started at 2023-10-18T03:12:39.479Z (139s ago) and ran for 210ms
    last collection id:      ac9edf3e-c420-4fa1-ae05-547f05d60574
    last collection started: 2023-10-18T03:12:39Z
    last collection done:    2023-10-18T03:12:39Z
```

Here's using `omdb` to poke around the inventory data:

```console
$ omdb db inventory
Print information about collected hardware/software inventory

Usage: omdb db inventory <COMMAND>

Commands:
  baseboard-ids  list all baseboards ever found
  cabooses       list all cabooses ever found
  collections    list and show details from particular collections
  help           Print this message or the help of the given subcommand(s)

Options:
  -h, --help  Print help

$ omdb db inventory baseboard-ids
note: using database URL postgresql://root@[::1]:49896/omicron?sslmode=disable
note: database schema version matches expected (7.0.0)
ID                                   PART_NUMBER      SERIAL_NUMBER 
a2418980-2188-4613-9787-df7d96ee94eb FAKE_SIM_GIMLET  SimGimlet00   
9684c118-6bae-461d-bbee-005265aee0f2 FAKE_SIM_GIMLET  SimGimlet01   
e4350b7a-49d2-4858-b4f2-74e52decfa80 FAKE_SIM_SIDECAR SimSidecar0   
c3f5ce3a-a536-44dd-b699-7354e778494d FAKE_SIM_SIDECAR SimSidecar1   

$ omdb db inventory cabooses
note: using database URL postgresql://root@[::1]:49896/omicron?sslmode=disable
note: database schema version matches expected (7.0.0)
ID                                   BOARD         GIT_COMMIT NAME       VERSION 
0f14c32f-9e59-44bd-a842-d99c499ff99c SimGimletSp   ffffffff   SimGimlet  0.0.1   
1cee4409-9021-42d3-b742-cf1a21c01918 SimSidecarRot eeeeeeee   SimSidecar 0.0.1   
2e7545b4-e4b2-4060-ade7-247e6ce64a70 SimSidecarSp  ffffffff   SimSidecar 0.0.1   
95d067e3-daf6-475e-8b01-01f6b93c1dd6 SimGimletRot  eeeeeeee   SimGimlet  0.0.1   

$ omdb db inventory collections
list and show details from particular collections

Usage: omdb db inventory collections <COMMAND>

Commands:
  list  list collections
  show  show what was found in a particular collection
  help  Print this message or the help of the given subcommand(s)

Options:
  -h, --help  Print help

$ omdb db inventory collections list
note: using database URL postgresql://root@[::1]:49896/omicron?sslmode=disable
note: database schema version matches expected (7.0.0)
ID                                   STARTED              TOOK  NSPS NERRORS 
ac9edf3e-c420-4fa1-ae05-547f05d60574 2023-10-18T03:12:39Z 17 ms 4    0       

$ omdb db inventory collections show ac9edf3e-c420-4fa1-ae05-547f05d60574
note: using database URL postgresql://root@[::1]:49896/omicron?sslmode=disable
note: database schema version matches expected (7.0.0)
collection: ac9edf3e-c420-4fa1-ae05-547f05d60574
collector:  e6bff1ff-24fb-49dc-a54e-c6a350cd4d6c (likely a Nexus instance)
started:    2023-10-18T03:12:39.543Z
done:       2023-10-18T03:12:39.560Z
errors:     0

Sled SimGimlet00
    part number: FAKE_SIM_GIMLET
    power:    A2
    revision: 0
    MGS slot: Sled 0 (cubby 0)
    found at: 2023-10-18 03:12:39.553986 UTC from http://[::1]:57254
    cabooses:
        SLOT       BOARD        NAME      VERSION GIT_COMMIT 
         SP slot 0 SimGimletSp  SimGimlet 0.0.1   ffffffff   
         SP slot 1 SimGimletSp  SimGimlet 0.0.1   ffffffff   
        RoT slot A SimGimletRot SimGimlet 0.0.1   eeeeeeee   
        RoT slot B SimGimletRot SimGimlet 0.0.1   eeeeeeee   
    RoT: active slot: slot A
    RoT: persistent boot preference: slot A
    RoT: pending persistent boot preference: -
    RoT: transient boot preference: -
    RoT: slot A SHA3-256: -
    RoT: slot B SHA3-256: -

Sled SimGimlet01
    part number: FAKE_SIM_GIMLET
    power:    A2
    revision: 0
    MGS slot: Sled 1 (cubby 1)
    found at: 2023-10-18 03:12:39.557752 UTC from http://[::1]:57254
    cabooses:
        SLOT       BOARD        NAME      VERSION GIT_COMMIT 
         SP slot 0 SimGimletSp  SimGimlet 0.0.1   ffffffff   
         SP slot 1 SimGimletSp  SimGimlet 0.0.1   ffffffff   
        RoT slot A SimGimletRot SimGimlet 0.0.1   eeeeeeee   
        RoT slot B SimGimletRot SimGimlet 0.0.1   eeeeeeee   
    RoT: active slot: slot A
    RoT: persistent boot preference: slot A
    RoT: pending persistent boot preference: -
    RoT: transient boot preference: -
    RoT: slot A SHA3-256: -
    RoT: slot B SHA3-256: -

Switch SimSidecar0
    part number: FAKE_SIM_SIDECAR
    power:    A2
    revision: 0
    MGS slot: Switch 0
    found at: 2023-10-18 03:12:39.546153 UTC from http://[::1]:57254
    cabooses:
        SLOT       BOARD         NAME       VERSION GIT_COMMIT 
         SP slot 0 SimSidecarSp  SimSidecar 0.0.1   ffffffff   
         SP slot 1 SimSidecarSp  SimSidecar 0.0.1   ffffffff   
        RoT slot A SimSidecarRot SimSidecar 0.0.1   eeeeeeee   
        RoT slot B SimSidecarRot SimSidecar 0.0.1   eeeeeeee   
    RoT: active slot: slot A
    RoT: persistent boot preference: slot A
    RoT: pending persistent boot preference: -
    RoT: transient boot preference: -
    RoT: slot A SHA3-256: -
    RoT: slot B SHA3-256: -

Switch SimSidecar1
    part number: FAKE_SIM_SIDECAR
    power:    A2
    revision: 0
    MGS slot: Switch 1
    found at: 2023-10-18 03:12:39.550225 UTC from http://[::1]:57254
    cabooses:
        SLOT       BOARD         NAME       VERSION GIT_COMMIT 
         SP slot 0 SimSidecarSp  SimSidecar 0.0.1   ffffffff   
         SP slot 1 SimSidecarSp  SimSidecar 0.0.1   ffffffff   
        RoT slot A SimSidecarRot SimSidecar 0.0.1   eeeeeeee   
        RoT slot B SimSidecarRot SimSidecar 0.0.1   eeeeeeee   
    RoT: active slot: slot A
    RoT: persistent boot preference: slot A
    RoT: pending persistent boot preference: -
    RoT: transient boot preference: -
    RoT: slot A SHA3-256: -
    RoT: slot B SHA3-256: -

```


andrewjstone:
> what chicken switches do we want to add?

I first read that as "chicken sandwiches" and was very confused. I'm not even that hungry!

nexus/db-queries/src/db/datastore/inventory.rs
```diff
@@ -0,0 +1,914 @@
...
+        &self,
+        opctx: &OpContext,
+        collection_id: Uuid,
+    ) -> Result<(), Error> {
+        // As with inserting a whole collection, we remove it in one big
+        // transaction for simplicity.  Similar considerations apply.  We could
+        // break it up if these transactions become too big.  But we'd need a
+        // way to stop other clients from discovering a collection after we
+        // start removing it and we'd also need to make sure we didn't leak a
+        // collection if we crash while deleting it.
```
jgallagher, line 587: Do we also need to prune no-longer-referenced `hw_baseboard_id` or `sw_caboose` rows? It's a little hard to imagine `hw_baseboard_id` getting "too big" since it only gets a row for each physical component the rack sees, but maybe in a large/long-lived multirack deployment? `sw_caboose` gets a few new rows for every update, so probably grows somewhat faster but still not all that quickly.

nexus/db-queries/src/db/datastore/inventory.rs
```diff
@@ -0,0 +1,914 @@
...
+            // table name as an identifier.  This is the same for both "FROM"
+            // and "INSERT" clauses.  And diesel internally does the same thing
+            // here (see the type of `InsertStatement::into_clause`).
+            into_inv_caboose: db::schema::inv_caboose::table.from_clause(),
+        }
+    }
+}
+
+impl diesel::query_builder::QueryFragment<diesel::pg::Pg> for InvCabooseInsert {
+    fn walk_ast<'b>(
```
jgallagher, line 787: I have no useful suggestion here, but would like to register my complaint that this is much harder to read than the equivalent query written out in raw SQL would be.

nexus/inventory/src/builder.rs
```diff
@@ -0,0 +1,278 @@
...
+                anyhow!("reported multiple times (same value)",)
+            } else {
+                anyhow!(
+                    "reported caboose multiple times (previously {:?}, \
+                    now {:?})",
+                    previous,
+                    sw_caboose
+                )
+            };
+            Err(error.context(format!(
```
jgallagher, line 241: I don't understand a sequence by which we could see this error message - what is this guarding against?

nexus/inventory/src/builder.rs
```diff
@@ -0,0 +1,278 @@
...
+            Ok(())
+        }
+    }
+
+    /// Helper function for normalizing items
+    ///
+    /// If `item` (or its equivalent) is not already in `items`, insert it.
+    /// Either way, return the item from `items`.  (This will either be `item`
+    /// itself or whatever was already in `items`.)
+    fn enum_item<T: Clone + Ord>(
```
jgallagher, line 255: Nitpick, but I don't love this name, particularly `enum`. Maybe `share_item`, or even `arcify_item`?

schema/crdb/dbinit.sql
```diff
@@ -2508,6 +2508,217 @@ CREATE TABLE IF NOT EXISTS omicron.public.switch_port_settings_address_config (
...
+    slot_boot_pref_transient omicron.public.hw_rot_slot, -- nullable
+    slot_boot_pref_persistent omicron.public.hw_rot_slot NOT NULL,
+    slot_boot_pref_persistent_pending omicron.public.hw_rot_slot, -- nullable
+    slot_a_sha3_256 TEXT, -- nullable
+    slot_b_sha3_256 TEXT, -- nullable
+
+    PRIMARY KEY (inv_collection_id, hw_baseboard_id)
+);
+
+CREATE TYPE IF NOT EXISTS omicron.public.caboose_which AS ENUM (
```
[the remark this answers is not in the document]
davepacheco: I went ahead and did this in 58c010f5f238f6ae5f17b93c16e045d0a4843aba.  Honestly, I could go either way on the result.  `CabooseWhich` does feel janky.  But it also reflects exactly what we're getting from MGS: that is, each row in this table reflects one response from the get-caboose endpoint, and that essentially represents the parameters to that request.  It's kind of a nice property that no row in an `inv_*` table represents data from multiple collection requests:

* it guides the schema design -- there's one table for each kind of observation (with maybe additional tables if there are a bunch of fields that can be present or absent together)
* it makes it easy to map the collection request responses to database rows
* it makes it easy to have the uniform set of (inv_collection_id, time_collected, source) fields.  We do have those here, but it's arguably misleading because the source and time_collected fields on `inv_service_processor` and `inv_root_of_trust` don't apply to the caboose fields
* This example sounds awfully specific but I feel like there's something general here: imagine if we wanted in the future to update the database _as_ we collect data instead of all at once at the end.  We'd have this unfortunate situation of having to either insert an `inv_service_processor` record and then update it later or else hang onto it (don't insert it) until we've tried to collected all the things that might go into it.

I don't think these are big deals for this particular case.  Rather, I came to this after exploring much different ways to structure this -- like an `inv_sled` table that might include pieces of information from both the sled agent (like the current host OS) and the SP (like the current host flash contents), etc.  I really disliked this because in the face of partial failures you have all these partial rows and then everything has to be NULL-able.  That's how I got to the "rows should not include data from multiple sources" rule.

Put differently: I don't think this specific violation of that rule is that bad, but without that rule, I found myself spinning in circles for a long time about how to design the schema.  It's pretty compelling to just say "each source of observation is a table; each observation is a row; then apply the usual database normalization rules".

All that said, I'm kind of ambivalent in the end.  I think I slightly prefer the previous thing with `caboose_which` but I'm interested in your thoughts.

nexus/inventory/src/builder.rs
jgallagher: Yeah, that sounds good :+1:

nexus/inventory/src/collector.rs
```diff
@@ -0,0 +1,202 @@
...
+        //
+        // This choice is debatable.  It's conceivable that an SP could be
+        // functioning but not visible to ignition.  In that case, we'd be
+        // better off trying to ask MGS about it even though ignition reports it
+        // powered off.  But in practice, if ignition can't see it, it's much
+        // more likely that there's just nothing plugged in.  And in that case,
+        // if we try to ask MGS about it, we have to wait for MGS to time out
+        // its attempt to reach it (currently several seconds).  This choice
+        // enables inventory to complete much faster, at the expense of not
+        // being able to identify this particular condition.
```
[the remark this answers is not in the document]
jgallagher: No argument on deferring it. Maybe create an issue after this lands so we don't lose track? Seems like the kind of thing that would only happen on an already-bad day.

nexus/db-queries/src/db/datastore/inventory.rs
davepacheco: In principle, eventually, yes.  I  think this is not urgent and we will notice before it becomes so.

nexus/db-queries/src/db/datastore/inventory.rs
```diff
@@ -0,0 +1,914 @@
...
+        // will only remove a complete collection if it has seen a newer
+        // complete one.  This cannot result in keeping fewer than "nkeep"
+        // collections because any Nexus will only remove a collection if there
+        // are "nkeep" newer ones.  In both of these cases, another Nexus might
+        // remove one of the ones that the first Nexus was counting on keeping,
+        // but only if there was a newer one to replace it.
+
+        opctx.authorize(authz::Action::Modify, &authz::INVENTORY).await?;
+
+        loop {
```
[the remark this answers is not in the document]
davepacheco: I think this is a promising approach.  But I'm a little worried it'll take a while to make this real (first real SQL, then real Diesel code), and also that when we do, we may find the database winds up doing a table scan (or rejecting the SQL because we've configured it to disallow that).  As an example: one of the stated assumptions is that the number of collections here _could_ be huge.  In that case, the highest-level subquery here will produce only 3 rows, but the query itself will be trying to select any collections _not_ in that set, which will in turn return a very large result set.  So we'll a `LIMIT` there.  We also want to start with the oldest ones, so we'll want an `ORDER BY` timestamp.  At this point, there are enough variables here that I'm not sure what query plan Cockroach will use.  I think ideal would be to do the subquery, then scan the index (by timestamp) and just skip over any rows that are in the subquery and stop when we reach the limit.  I hope it will do that but it's hard to be sure until we do that work.

I think this is probably all solvable (or else we'll find out why it's not), but if the current code is at least correct and not pathological, I'd rather defer this than spend the time now to work all this out.

schema/crdb/dbinit.sql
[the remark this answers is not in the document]
davepacheco: > I think it's the same either way. 

Yeah, from a representation perspective, that's true.  I had in my mind an implicit rule that we wouldn't want to write a record and then update it later in the same operation.  But that's somewhat arbitrary, too.  (_Not_ doing this does make it more complicated to infer anything from partially-inserted collections, and to measure progress based on what's present, but now we're talking about several layers of hypotheticals that aren't worth dealing with now.)

davepacheco:
@jgallagher I've made a bunch of changes since your review, but hopefully no surprises.  Besides the stuff that came up in your review:

- added a bunch of automated tests (which required yet more database layer boilerplate)
- rewrote the `omdb` commands to use functions I created for the automated tests
- replaced the CTE with regular Diesel code when I realized I could do a join within `INSERT INTO ... SELECT`
- added a "disable" chicken switch in the Nexus config, in case of some critical reason we need to stop inventory from running
- added the database schema changes to the update path

That's all I've got planned so I think this is ready for re-review.

jgallagher:
Looks like the helios deploy test failure is legit (or at least related to the changes):

```
error_message_internal = Failed to initialize CockroachDb:
 Error running command in zone 'oxz_cockroachdb_13a01e9b-3937-4043-b973-eef03d18a34f':
 Command [/opt/oxide/cockroachdb/bin/cockroach sql --insecure --host [fd00:1122:3344:101::7]:32221 --file /opt/oxide/cockroachdb/sql/dbinit.sql] executed and failed with status:
 exit status: 1
 stdout: ... snip snip snip ... CREATE TABLE\n  stderr: ERROR: relation "omicron.public.inv_collection" does not exist\nSQLSTATE: 42P01\nERROR: relation "omicron.public.inv_collection" does not exist\nSQLSTATE: 42P01\nFailed running "sql"\n

File: nexus/db-queries/src/db/datastore/inventory.rs [hunk 8]
```diff
@@ -0,0 +1,1568 @@
...
+            "collection_id" => collection.id.to_string(),
+        );
+
+        Ok(())
+    }
+
+    /// Prune inventory collections stored in the database, keeping at least
+    /// `nkeep`.
+    ///
+    /// This function removes as many collections as possible while preserving
```
andrewjstone, line 498: The latest `nkeep` are provided by timestamps, which aren't really global. Right now collection is at 10 min intervals, so as long as only 1 nexus performs a collection per interval this should be fine. However, I could see problems arising around order, although quite unlikely due to our 500ms limitation around syncing. 

I don't really think this is something worth worrying about but figured I'd ask for completeness sake. Are there autoincrementing IDs we could use for collections rather than UUIDs as foreign keys and then sort by those? Would this present other issues with dueling Nexuses?

andrewjstone (commented):
@davepacheco This looks great. I only skimmed most of the DB transaction stuff, but the overall picture makes sense to me. I'll leave it to John to approve due to my skimming and his expertise.

nexus/db-queries/src/db/datastore/inventory.rs
davepacheco: Yeah, using timestamps is definitely a little fuzzy.  In this case though I think that reflects the reality that collections are not atomic and they don't have a total order.  Two Nexus instances could totally run collections concurrently that have start/done times that overlap (and I think that's fine).  Consumers can decide if they want the most-recently-started or most-recently-finished (or even the-one-containing-the-most-recent-collection-time-for-the-specific-item-that-I-care-about).  We could potentially use a [sequence](https://www.cockroachlabs.com/docs/stable/create-sequence) to assign a total order to these but I don't think it would have a useful semantic meaning -- at best it'd be a proxy for "which one committed to the database first" and I'm not sure that's useful.

Okay so my argument is basically "report the facts (the start/done timestamp) and let consumers decide what they want".  But that just punts your question to "okay, well, which ones _should_ we keep when we're pruning them?".  And I think the answer here is to tune both the frequency and `nkeep` such that it doesn't really matter if we choose "wrong" -- i.e., if two collections start/finish at about the same time but for some reason a consumer might reasonably want either one, we should probably just keep both.  But my expectation here is that all consumers for now would probably want the same thing, which is the latest "time_started" one, and as long as "nkeep" is more than 1 then it doesn't matter which ones we keep if two overlap because there's always a newer one which is what consumers actually want.

andrewjstone: Ah, ok. That makes sense. I was actually thinking that we could eliminate the overlapping collections to a degree by having each nexus check that there isn't a collection currently running - or rather that one hasn't started within some bound (say collection_interval / 2) before kicking off another. With that, collections should be very close to totally ordered by time if not always so.

nexus/db-queries/src/db/datastore/inventory.rs
```diff
@@ -0,0 +1,1568 @@
...
+            "nrots" => nrots,
+            "ncabooses" => ncabooses,
+            "nerrors" => nerrors,
+        );
+
+        Ok(())
+    }
+}
+
+/// A SQL common table expression (CTE) used to insert into `inv_caboose`
```
jgallagher, line 803: I think this block comment is referencing code that no longer exists, right?

jgallagher (approved):
This looks great! Just a handful of small nits.

nexus/src/app/background/init.rs
```diff
@@ -88,6 +96,30 @@ impl BackgroundTasks {
             (task, watcher_channel)
         };
 
+        // Background task: inventory collector
+        let task_inventory_collection = {
+            let watcher = inventory_collection::InventoryCollector::new(
```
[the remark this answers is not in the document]
davepacheco: Yeah, fixed in a55216d40.

nexus/db-model/src/inventory.rs
```diff
@@ -0,0 +1,443 @@
...
+/// See [`nexus_types::inventory::BaseboardId`].
+#[derive(Queryable, Insertable, Clone, Debug, Selectable)]
+#[diesel(table_name = hw_baseboard_id)]
+pub struct HwBaseboardId {
+    pub id: Uuid,
+    pub part_number: String,
+    pub serial_number: String,
+}
+
+impl<'a> From<&'a BaseboardId> for HwBaseboardId {
```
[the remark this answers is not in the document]
davepacheco: Agreed.  Fixed in a55216d40.  I changed this one and the `SwCaboose` one.  I did not change the `InvCollection` one because in that case, the source object (a `Collection`) is potentially huge.

nexus/src/app/background/inventory_collection.rs
```diff
@@ -0,0 +1,240 @@
...
+    #[nexus_test(server = crate::Server)]
+    async fn test_basic(cptestctx: &ControlPlaneTestContext) {
+        let nexus = &cptestctx.server.apictx().nexus;
+        let datastore = nexus.datastore();
+        let opctx = OpContext::for_tests(
+            cptestctx.logctx.log.clone(),
+            datastore.clone(),
+        );
+
+        // Nexus starts our very background task, so we should find a collection
```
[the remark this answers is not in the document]
davepacheco: Not a typo, but poorly written.  I reworded it in a55216d40.

nexus/db-queries/src/db/datastore/inventory.rs
```diff
@@ -0,0 +1,1568 @@
...
+            .collect::<Vec<_>>();
+        let error_values = collection
+            .errors
+            .iter()
+            .enumerate()
+            .map(|(i, message)| {
+                let index = u16::try_from(i).map_err(|e| {
+                    Error::internal_error(&format!(
+                        "failed to convert error index to u16 (too \
+                                many errors in inventory collection?): {}",
```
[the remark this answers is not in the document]
davepacheco: Ugh.  This keeps happening and I don't notice.  I'm not sure why.  I wonder if it happens when some other change (like a symbol rename) causes this block to be reformatted when I'm not actually working on it.  Anyway, fixed in a55216d40.

nexus/db-queries/src/db/datastore/inventory.rs
davepacheco: Yikes!  Yes.  Removed in a55216d40.

nexus/src/app/background/inventory_collection.rs
```diff
@@ -0,0 +1,240 @@
...
+use serde_json::json;
+use std::sync::Arc;
+
+/// Background task that reads inventory for the rack
+pub struct InventoryCollector {
+    datastore: Arc<DataStore>,
+    resolver: internal_dns::resolver::Resolver,
+    creator: String,
+    nkeep: u32,
+    disable: bool,
```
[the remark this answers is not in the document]
davepacheco: That's right.  To my knowledge we do not yet have a way to apply dynamic config at runtime.  The intent here is that if we really needed to, we could modify the TOML file inside each Nexus zone to disable this task.  Then we'd restart Nexus.  It's obviously not great but I've frequently found these sorts of facilities essential in the mitigation of production incidents in the past.  (A step up might be a support API for pausing any background task in a particular Nexus instance by name.  But that wouldn't survive Nexus restart without storing that config somewhere.)

davepacheco:
I think I've addressed the outstanding feedback and I intend to land this once the repo re-opens after the latest customer update.

PR #4298, opened 2023-10-18
Limit size of the search range for VNIs when creating VPCs

### LINKED ISSUE #4283: Memory budget exceeded error when attempting to create a new vpc
The error was encountered on rack2 when I tried to create a new project:
```
17:28:52.650Z INFO 65a11c18-7f59-41ac-b9e7-680627f996e7 (dropshot_external): request completed
    error_message_internal = saga ACTION error at node "vpc": database error (kind = Unknown): root: memory budget exceeded: 50328352 bytes requested, 117717552 currently allocated, 134217728 bytes in budget\nHINT: Consider increasing --max-sql-memory startup parameter.\n
    file = /home/build/.cargo/git/checkouts/dropshot-a4a923d29dccc492/fa728d0/dropshot/src/server.rs:841
    latency_us = 4283076
    local_addr = 172.30.2.5:443
    method = POST
    remote_addr = 172.20.17.42:58746
    req_id = 68591636-0a4e-4e81-80b3-0a846f0e9aa5
    response_code = 500
    uri = https://oxide-local.sys.rack2.eng.oxide.computer/v1/projects
```
Given the error was related to vpc, I then tried to create a new vpc in an existing project and hit the same error.

bnaecker:
- Fixes #4283.
- Adds a relatively small limit to the `NextItem` query used for finding a free VNI during VPC creation. This limits the memory consumption to something very reasonable, but is big enough that we should be extremely unlikely to find _no_ available VNIs in the range.
- Add an application-level retry loop when inserting _customer_ VPCs, which catches the unlikely event that there really are no VNIs available, and retries a few times.
- Adds tests for the computation of the limited search range.
- Adds tests for the actual exhaustion-detection and retry behavior.

### DIFF
--- a/nexus/db-queries/src/db/datastore/vpc.rs
+++ b/nexus/db-queries/src/db/datastore/vpc.rs
@@ -34,12 +34,15 @@ use crate::db::model::VpcUpdate;
 use crate::db::model::{Ipv4Net, Ipv6Net};
 use crate::db::pagination::paginated;
 use crate::db::queries::vpc::InsertVpcQuery;
+use crate::db::queries::vpc::VniSearchIter;
 use crate::db::queries::vpc_subnet::FilterConflictingVpcSubnetRangesQuery;
 use crate::db::queries::vpc_subnet::SubnetError;
 use async_bb8_diesel::AsyncConnection;
 use async_bb8_diesel::AsyncRunQueryDsl;
 use chrono::Utc;
 use diesel::prelude::*;
+use diesel::result::DatabaseErrorKind;
+use diesel::result::Error as DieselError;
 use ipnetwork::IpNetwork;
 use omicron_common::api::external::http_pagination::PaginatedBy;
 use omicron_common::api::external::CreateResult;
@@ -85,18 +88,23 @@ impl DataStore {
             SERVICES_VPC.clone(),
             Some(Vni(ExternalVni::SERVICES_VNI)),
         );
-        let authz_vpc = self
+        let authz_vpc = match self
             .project_create_vpc_raw(opctx, &authz_project, vpc_query)
             .await
-            .map(|(authz_vpc, _)| authz_vpc)
-            .or_else(|e| match e {
-                Error::ObjectAlreadyExists { .. } => Ok(authz::Vpc::new(
-                    authz_project.clone(),
-                    *SERVICES_VPC_ID,
-                    LookupType::ByName(SERVICES_VPC.identity.name.to_string()),
-                )),
-                _ => Err(e),
-            })?;
+        {
+            Ok(None) => {
+                let msg = "VNI exhaustion detected when creating built-in VPCs";
+                error!(opctx.log, "{}", msg);
+                Err(Error::internal_error(msg))
+            }
+            Ok(Some((authz_vpc, _))) => Ok(authz_vpc),
+            Err(Error::ObjectAlreadyExists { .. }) => Ok(authz::Vpc::new(
+                authz_project.clone(),
+                *SERVICES_VPC_ID,
+                LookupType::ByName(SERVICES_VPC.identity.name.to_string()),
+            )),
+            Err(e) => Err(e),
+        }?;
 
         // Also add the system router and internet gateway route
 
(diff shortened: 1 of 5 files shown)


nexus/db-queries/src/db/queries/vpc.rs
```diff
@@ -278,3 +270,79 @@ impl NextVni {
...
+#[cfg(not(test))]
+pub const MAX_VNI_SEARCH_RANGE_SIZE: u32 = 2048;
+#[cfg(test)]
+pub const MAX_VNI_SEARCH_RANGE_SIZE: u32 = 10;
+
+impl VniShifts {
+    fn new(vni: Vni) -> Self {
+        let base_u32 = u32::from(vni.0);
+        let range_end = base_u32 + MAX_VNI_SEARCH_RANGE_SIZE;
+
```
smklein, line 301: Should we add a https://docs.rs/static_assertions/latest/static_assertions/macro.const_assert.html that `MAX_VNI_SEARCH_RANGE_SIZE <= (MAX_VNI - MIN_GUEST_VNI)` ? Otherwise it seems like our wrap-around "min_shift" might collide with the part of the range we previously checked.

nexus/db-model/src/vpc.rs
```diff
@@ -110,6 +110,14 @@ impl IncompleteVpc {
             subnet_gen: Generation::new(),
         })
     }
+
+    /// Create a copy of self, but with a new random VNI.
+    ///
+    /// This is used to retry insertion of a VPC in the case where we can't find
+    /// an available VNI.
+    pub fn with_new_vni(self) -> Self {
```
smklein, line 118: Super nitpick, this is `with_random_vni`, as it could randomly be the same VNI as before

nexus/db-queries/src/db/datastore/vpc.rs
```diff
@@ -287,22 +294,60 @@ impl DataStore {
...
+                        "original_vni" => ?original_vni,
+                        "new_vni" => ?vpc.vni,
+                    );
+                    attempt += 1;
+                }
+            }
+        }
+
+        // We've failed to find a VNI within our retry limit. We'll return a 503
+        // in this case.
```
smklein, line 331: This is certainly better than what we had before -- I wonder if there's any point to a "slower, more exhaustive search" once we hit this condition.

Don't get me wrong, I'd be happy to land this change as-is and avoid the immediate issues with memory we're seeing, but it seems like it would be frustrating to keep retrying "project create" to try to find a free VNI, on a system that's very heavily loaded (e.g., a high % of VNIs are allocated).

Either way, happy punting on that until later.

smklein (approved):
Great tests!

nexus/db-queries/src/db/datastore/vpc.rs
bnaecker: I guess it depends on what that slower, more exhaustive search looks like. One very easy option would be to increase the size of the search range, either statically or at each failure. We could certainly pick this to be something in between the entire VNI range (which it used to be) and 0.1% of the range (in this PR). Is that feasible?

What would a more exhaustive search look like? I think avoiding the memory pressure would require restructuring the query, which I'd definitely like to defer.

nexus/db-queries/src/db/queries/vpc.rs
bnaecker: Good idea, done in ca6d7ccdf

nexus/db-model/src/vpc.rs
bnaecker: Done in ca6d7ccdf

nexus/db-queries/src/db/datastore/vpc.rs
smklein: So, in a world with 100 buckets, we're basically doing one of the following:

- Expensive full search: Search all 100 buckets for a free bucket in a single query. This runs out of memory for a larger number of buckets.
- This PR: Pick a random index, scan the next ~5 buckets. If we fail, retry 3 times. This observes a fraction of the total number of buckets, but in a not-super-loaded system, works fine. However, at worst case, this only scans 5% of the total buckets (in the case where "random" returns the same starting index 3 times in a row), and at best case, it scans 15% of the total number of buckets. It's possible this happens when there are free buckets.

I think there are some alternatives here:

1. **Building an efficient allocator**. Perhaps we track the "free VNIs" through a set of ranges, and treat it as a proper range-based allocator. This shares a lot of the problems we'd encounter with a memory allocator, but we'd do the ops in SQL. Rather than scanning, we could simply look up the smallest range in our free list, and pop a VNI out of it.

Pros: It's efficient. It's potentially re-usable for other allocations, like IPs. As long as there is space available, we shouldn't fail.
Cons: It's expensive in engineering time to build. It's complex.

2. **Scanning the entire VNI space in batches**. Rather than "randomly" picking three indices, we "randomly" pick the starting index, from which we scan, and simply iterate over the entire VNI space, one window at a time. For example: we randomly start on bucket 50, scan from 50 -> 55, find nothing free, so we then scan from 55 -> 60, find bucket 56 is free, and return it.

Pros: It's simple. As long as there is space available, we shouldn't fail.
Cons: It's slow on a contended system.

smklein: I'm not even proposing blocking this PR on these features, but I just want us to be aware of the tradeoff in the following conditions:

| Method | Memory Usage | Cost-to-build | Allocates if space is free? | Efficient under contention |
| -- | -- | -- | -- | -- |
| Full search | Too High | N/A (on `main`) | Yes | Not really? | 
| Random Index + Retry | Low | Cheap | Not always | No, but it fails fast. Though a caller will need to retry! |
| Random Index + Scan everything in batches | Low | Cheap | Yes | No, it would loop for a while | 
| Efficient Allocator | Lowish? | Expensive | Yes | It should be! |

bnaecker: Thanks for laying those out. The one I was missing was the batch-based approach, which has a lot of appeal!

There's also another option, I think. The VNIs are a global range -- all VPCs share them. They're also a single number. This is in contrast to the IP addresses, where I intended this query to be used. There, there's a _subnet_ we usually need to search within, and also can be overlap in the subnets between different VPCs.

So what if we put an index on VNIs (I think we already have one), and then restructure the query to have a subquery which takes the next available VNI? I'm not sure if we should avoid allocating sequentially for some reason, but other than that, I don't see a downside. The query is not super complicated, something like

```sql
SELECT IF(
    (SELECT COALESCE(max(vni), 1023) + 1 FROM vpc) > 16777216, -- check if the next VNI is beyond the max
    NULL, -- If so, take NULL which will fail with a `NotNullViolation` error
    (SELECT COALESCE(max(vni), 1023) + 1 FROM vpc) -- or insert the next one
) AS "vni"
```

I'm sure it could be improved, but that would work and handle no entries at all and the table being full.

bnaecker: The downside of course is that we're mucking with the Diesel query at the level of raw SQL, which always takes longer than I want it to.

bnaecker: Oh, duh. This doesn't wrap at all, so we'll never reallocate VNIs at the bottom of the range once they're freed. Silly me. I like the scan-in-batches approach. It's not very far from this PR (just a bit more bookkeeping around the VNI search range), and should avoid the poor client experience of needing a retry.

bnaecker: Ok, I've done the exhaustive search in 3be825e15. Let me know if you have other suggestions, and thanks for the eyes  @smklein!

bnaecker:
Hmm looks like this hit the NTP test flake John noted earlier today. I’ll try the check again, but let me know if you’ve any suggestions in the meantime @smklein

nexus/db-queries/src/db/datastore/vpc.rs
smklein: Just to confirm -- are we still picking a random starting offset? (That seems like a minor optimization, but one that might spread out 'where allocations occur').

smklein: Other than confirming that piece -- looks good to me!

bnaecker: Yep starts with a random offset the first time the query is constructed. Then we loop over fixed sized ranges until we successfully done a free VNI.

PR #4304, opened 2023-10-19
bootstore: don't use hardcoded ports in tests

luqmana:
Fix some test flakiness by not hardcoding specific ports to use.


bootstore/src/schemes/v0/peer.rs
```diff
@@ -1217,80 +1419,60 @@ mod tests {
...
-        // Ensure the peer's generation numbers don't get bumped. The learner
-        // will ask the same sled for a share first, which it already handed
-        // out.
-        assert!(
-            peer0_gen_new == peer0_gen_new_2
-                && peer1_gen_new == peer1_gen_new_2
-        );
+            nodes[1].get_status().await.unwrap().fsm_ledger_generation;
+        assert_eq!(peer0_gen_new, peer0_gen_new_2);
+        assert_eq!(peer1_gen_new, peer1_gen_new_2);
```
luqmana, line 1448: This probably warrants a close look to make sure the test is still valid. Without it, there was still some flakiness based since every time a node restarted it would have a new port and thus impact choice of peer.

bootstore/src/schemes/v0/peer.rs
```diff
@@ -1217,80 +1419,60 @@ mod tests {
...
+        // bumped and stop it.
+        let non_learner_peer = if peer0_gen_new == peer0_gen { 0 } else { 1 };
+        nodes.shutdown_node(non_learner_peer).await;
+
+        // Now we can stop the learner, wipe its ledger, and restart it.
+        nodes.shutdown_learner(true).await;
+        nodes.start_learner().await;
+
+        // Now restart the other peer and tell the learner to relearn its share
+        nodes.start_node(non_learner_peer).await;
```
andrewjstone, line 1437: I believe this is non-deterministic, and we are just getting lucky. When `init_learner()` is called it chooses a selected peer in a `BTreeSet<Baseboard>`. For success, starting a node may not complete the connection before 
the [learner selects a peer](https://github.com/oxidecomputer/omicron/blob/9191af67b9aff710dac93ada6012b1e9b7d0c79c/bootstore/src/schemes/v0/fsm.rs#L287) or the Baseboard of the `non_learner_peer` may sort after the other peer. 

I think we should swap lines 1437 and 1438 so that the proper node is always chosen, but @luqmana claimed this causes hangs. I am going to dig into this in between BGP testing on madrid.

andrewjstone (commented):
Thanks so much for this @luqmana. The tests are much shorter and cleaner with the new setup code! I'm still checking for correctness as detailed in a comment below. Other than that one thing though, this looks great!

bootstore/src/schemes/v0/peer.rs
andrewjstone: I confirmed the hang and it makes sense. In order for a share to be handed out from one node to a learner, the encrypted 'extra shares' must be decrypted by recomputing the rack secret. If we don't start the second peer back up there's only one peer running and the rack secret cannot be recomputed, and a share is never handed out. The learner tries indefinitely and only returns once it learns a share. This is confirmed by the logs.

If we have connections at `learner-2` from peer `A` and peer `B`, the learner will always select A to learn from since it is sorted first the in the `peer.connections` BTreeSet. What this test was trying to do was to figure out the sort order dynamically and let the connections land where they may, then force a send to the same connection. We can simplify the test by always forcing the learner to pick peer A by asking the learner which connections it currently has and waiting until it has A, and only then trying to call `init_learner`. I'll make that change, PR it to this and then we should be good to go!

andrewjstone (approved):
Good to go! Thanks again @luqmana

PR #4308, opened 2023-10-20
Make sure project_delete_snapshot does its job

jmpesp:
`project_delete_snapshot`'s job is to soft-delete snapshots but due to the combination of the time_deleted filter and `check_if_exists` it was not doing this: `test_create_snapshot_record_idempotent` would fail if there was a check that the snapshot was actually deleted.

This was causing a snapshot delete request to proceed for a snapshot that was in state Creating, causing a case where a volume was hard deleted before it had its associated resource accounting performed, leading to regions associated with deleted resources that would never be cleaned up.


nexus/db-queries/src/db/datastore/snapshot.rs
```diff
@@ -253,15 +252,13 @@ impl DataStore {
         use db::schema::snapshot::dsl;
 
         let updated_rows = diesel::update(dsl::snapshot)
-            .filter(dsl::time_deleted.is_null())
```
smklein, line 256: It kinda seems to me like the problem is our inability to distinguish the "state is wrong" vs "object is already deleted" cases.

We use this pattern of `filter(time_deleted.is_null())` before `check_if_exists` in the following spots:

- nexus/db-queries/src/db/datastore/silo_user.rs
- nexus/db-queries/src/db/datastore/vmm.rs
- nexus/db-queries/src/db/datastore/external_ip.rs
- nexus/db-queries/src/db/datastore/ssh_key.rs
- nexus/db-queries/src/db/datastore/disk.rs
- nexus/db-queries/src/db/datastore/instance.rs

But AFAICT, most of those don't also check for state validity, which should propagate a unique user error.

| Snapshot Exists (by UUID) | Snapshot is already Soft Deleted | Valid state to be deleted | diesel::update result | Why | 
| -- | -- | -- | -- | -- |
| No | N/A | N/A | Error from `check_if_exists` -> Not found | Object doesn't exist |
| Yes | No | Yes | `check_if_exists` returns `UpdateStatus::Updated` | Deletion succeeds |
| Yes | No | No | `check_if_exists` returns `UpdateStatus::NotUpdatedButExists` | Deletion failed because bad state. |
| Yes | Yes | N/A | `check_if_exists` returns `UpdateStatus::NotUpdatedButExists` | Deletion failed because the object was soft deleted |

smklein: I think I understand what this PR is trying to do, but I think it has the following problem:

1. We soft delete a snapshot. This works.
2. We re-issue the soft deletion of the snapshot. This now "appears idempotent", and returns the `snapshot_id`.
3. We hard delete the snapshot.
4. We re-issue the soft deletion of the snapshot. **Now** it fails, because the record has been deleted.

That's weird, right??? Like, the result of hard-deleting the snapshot could suddenly change the outcome of this function.

I understand that we probably want idempotency here for the sake of the saga, but I think it's the responsibility of the calling saga to do the following:

- Issue a deletion request
- Map "not found" -> OK

This is, for example, how the instance deletion saga works.

smklein: Source for the instance deletion saga's idempotency: https://github.com/oxidecomputer/omicron/blob/9191af67b9aff710dac93ada6012b1e9b7d0c79c/nexus/src/app/sagas/instance_delete.rs#L80-L92

jmpesp:
Thanks for the comments - `project_delete_snapshot` was indeed using `check_if_exists` incorrectly. The result was being coerced into an integer instead of the result struct because `project_delete_snapshot` was using `execute_async` instead of `execute_and_check`. 0b4f0d4d9 changes `project_delete_snapshot` to mimic how `project_delete_disk_no_auth` correctly uses `check_if_exists`. I've also added an idempotence test for `project_delete_disk_no_auth` in d90c9ee8a to ensure this behaviour in a test.

nexus/db-queries/src/db/datastore/snapshot.rs
```diff
@@ -252,30 +253,70 @@ impl DataStore {
...
+                public_error_from_diesel(
+                    e,
+                    ErrorHandler::NotFoundByLookup(
+                        ResourceType::Snapshot,
+                        LookupType::ById(snapshot_id),
+                    ),
+                )
+            })?;
+
+        match result.status {
```
smklein, line 278: This new version looks great, thanks for updating!

smklein (approved):

nexus/db-queries/src/db/datastore/snapshot.rs
jmpesp: > That's weird, right??? Like, the result of hard-deleting the snapshot could suddenly change the outcome of this function.

FWIW, I don't think this problem is limited to snapshots - we don't hard-delete most resources at the moment but I imagine that other code that uses `UpdateAndCheck` in this way would respond the same if the resource was hard deleted.

PR #4309, opened 2023-10-20
[nexus] Improve logging for schema changes

smklein:
Improves visibility of logging, indicating:
- What version are we upgrading into, throughout the process
- Logging the names of SQL files used for upgrade, as they occur


david-crespo (approved):
Glad you also ran into this so I don't have to figure it out!

PR #4310, opened 2023-10-21
Ensure bootstore test isn't just getting lucky

andrewjstone:
For Luqman's test fix there was one spot, where the learner must ask the same peer for a share to ensure that the same share is returned. We were getting lucky based on connection speed. This change tightens that up and fixes the problem described [here](https://github.com/oxidecomputer/ omicron/pull/4304#discussion_r1367481111).


andrewjstone:
I removed one of the waits and confirmed non-determinism, then ran the test in a loop for a while to watch it pass.

Issue #4220, opened 2023-10-06
Oximeter producer for sidecar hits "Schema mismatch for timeseries 'data_link:link_up'" error


askfongjojo:
Errors like this were seen on rack2 which was updated this morning from main:
```
21:33:58.422Z ERRO oximeter (oximeter-agent): timeseries schema mismatch, sample will be skipped: Schema mismatch for timeseries 'data_link:link_up', expected fields {"link_id": I64, "port_id": String, "rack_id": Uuid, "sidecar_id": Uuid, "sled_id": Uuid} found fields {"link_id": I64, "port_id": String, "rack_id": Uuid, "sidecar_id": Uuid, "sled_id": Uuid}
    collector_id = da510a57-3af1-4d2b-b2ed-2e8849f27d8b
    file = oximeter/db/src/client.rs:300
    id = ef3a153a-de65-4ba4-be3d-eee1fa152907
```

bnaecker:
This error message is extremely unhelpful, since it seems to show the same field names / values in the found and expected schema. I'm going to add a small PR which improves the error reporting in `oximeter-collector`, to hopefully smoke this out more clearly.

david-crespo:
Must be comparing them with different orders but then printing with sorted keys.

bnaecker:
Yeah, it looks like another consequence of the switch from a `Vec` to the `BTreeMap` for storing fields. Testing now.

bnaecker:
Well that's not what I expected:

```
16:19:47.512Z ERRO oximeter (oximeter-agent): timeseries schema mismatch, sample will be skipped
    actual = TimeseriesSchema { timeseries_name: TimeseriesName("data_link:link_up"), field_schema: {FieldSchema { name: "link_id", ty: I64, source: Target }, FieldSchema { name: "port_id", ty: String, source: Target }, FieldSchema { name: "rack_id", ty: Uuid, source: Target }, FieldSchema { name: "sidecar_id", ty: Uuid, source: Target }, FieldSchema { name: "sled_id", ty: Uuid, source: Target }}, datum_type: CumulativeI64, created: 2023-10-22T16:19:47.512167943Z }
    collector_id = 82b3c525-cf15-40e1-8a9d-9bf5eb4cb3af
[...]

PR #4312, opened 2023-10-21
Sort fields when extracting timeseries schema

bnaecker:
- Fields are reported in a sample via the implementations of `Target` and `Metric`, which may or may not be sorted. They'll be in declaration order, if folks derive the traits. When deriving a schema from the sample, collect fields into a set to ignore order.
- Convert between a `DbFieldList` and `BTreeSet` when inserting / reading fields from the nested tables in ClickHouse.
- Add sanity test that we're sorting field schema correctly.
- Errors for schema mismatches report entire schema, not just fields.


bnaecker:
This is intended to help smoke out #4220, but may actually resolve that as well. We collect fields in the order their defined from the sample, which may not be sorted. This stores them in a set to ensure sorting by name.

bnaecker:
Ok, I'm looking more at this and am confused. The fields that are causing the issue are all from the target. A `Sample` does in fact store and return those sorted by name, so that means we must have derived a schema where the fields are sorted by name. Unless I'm missing something, this means we have data in the database where they're _not_ sorted by name, so that the derived sample doesn't match. That's pretty surprising, since we previously wiped the database as part of the updates around #4008.

I'm going to wait for dogfood to come back up and look in the database. I may scale this back to only change how the error is printed if I can't get a good theory for where this error comes from.

bnaecker:
Ok, I've tested this on `madrid` several times. The big upshot is that https://github.com/oxidecomputer/omicron/issues/4220 would have been much easier to debug:

```
20:09:13.761Z ERRO oximeter (oximeter-agent): timeseries schema mismatch, sample will be skipped
    actual = TimeseriesSchema { timeseries_name: TimeseriesName("data_link:link_up"), field_schema: {FieldSchema { name: "link_id", ty: I64, source: Target }, FieldSchema { name: "port_id", ty: String, source: Target }, FieldSchema { name: "rack_id", ty: Uuid, source: Target }, FieldSchema { name: "sidecar_id", ty: Uuid, source: Target }, FieldSchema { name: "sled_id", ty: Uuid, source: Target }}, datum_type: CumulativeI64, created: 2023-10-23T20:09:13.761288693Z }
    collector_id = 82b3c525-cf15-40e1-8a9d-9bf5eb4cb3af
    expected = TimeseriesSchema { timeseries_name: TimeseriesName("data_link:link_up"), field_schema: {FieldSchema { name: "link_id", ty: I64, source: Target }, FieldSchema { name: "port_id", ty: String, source: Target }, FieldSchema { name: "rack_id", ty: Uuid, source: Target }, FieldSchema { name: "sidecar_id", ty: Uuid, source: Target }, FieldSchema { name: "sled_id", ty: Uuid, source: Target }}, datum_type: Bool, created: 2023-10-20T21:20:50.954995640Z }
    file = oximeter/db/src/client.rs:401
    id = 17b64209-cdcd-490c-b961-2c4212821cbd
    sample = Sample { measurement: Measurement { timestamp: 2023-10-23T20:09:13.630990476Z, datum: CumulativeI64(Cumulative { start_time: 2023-10-20T23:06:25.888054024Z, value: 1 }) }, timeseries_name: "data_link:link_up", target: FieldSet { name: "data_link", fields: {"link_id": Field { name: "link_id", value: I64(0) }, "port_id": Field { name: "port_id", value: String("qsfp0") }, "rack_id": Field { name: "rack_id", value: Uuid(0b26ba18-7ca7-42c6-8450-d0fcdcaad0c0) }, "sidecar_id": Field { name: "sidecar_id", value: Uuid(00000000-0000-0000-0802-5dbb797061d4) }, "sled_id": Field { name: "sled_id", value: Uuid(c3bce666-0593-4e24-baca-c8bd121c9cda) }} }, metric: FieldSet { name: "link_up", fields: {} } }
```

Second, we now do _not_ insert additional copies of the timeseries schema into the database, each time we see a sample for the first time. Instead, we check the internal cache and the database, only inserting if the schema appears in neither. I have a unit test which checks this, but I've also seen that the number of entries in the schema table the live `madrid` ClickHosue database does not increase if we restart `oximeter`.

oximeter/db/src/client.rs
```diff
@@ -503,10 +516,12 @@ impl Client {
         response
     }
 
-    async fn get_schema(&self) -> Result<(), Error> {
+    async fn get_schema_locked(
```
david-crespo, line 519: I find this name a little confusing since the function doesn't seem to itself have anything to do with locking, but rather assumes the passed in schema is locked. Is that right? Maybe a comment to clarify if so, if not a name change.

bnaecker: Sure, I can do that. I think that's a Rust-ism, used when the method can only be called after acquiring a lock, in this case around the schema map.

bnaecker: Done in a67f22617

oximeter/db/src/client.rs
```diff
@@ -384,30 +380,47 @@ impl Client {
...
+        let mut schema = self.schema.lock().await;
+
+        // We need to possibly check that this schema is in the local cache, or
+        // in the database, all while we hold the lock to ensure there's no
+        // concurrent additions. This containment check is needed so that we
+        // check both the local cache and the database, to avoid adding a schema
+        // a second time.
+        if !schema.contains_key(&name) {
+            self.get_schema_locked(&mut schema).await?;
+        }
```
david-crespo, line 394: Why is checking `schema.contains_key(&name)` sufficient to decide whether we need to go update the cached value from the DB? Can there not be a situation where the name is already in the cache but the associated schema in the DB is different? Or is that guarantee what locking buys us?

david-crespo: I think that first paragraph is a much better comment than what's there!

david-crespo: Or both!

david-crespo (approved):

oximeter/db/src/client.rs
bnaecker: Sweet, will update

bnaecker: Done in c1b287520

PR #4319, opened 2023-10-23
Re-introduce support for setting a VLAN for an uplink from RSS

luqmana:
Going from the previous `UplinkConfig` format to the `PortConfigV1` introduced with the BGP work seems to have lost the ability to configure the VLAN ID for reaching a gateway. The actual APIs and functionality is still there so just needed to be wired back up.

Adding a vlan id to the default route in `config-rss.toml`, I was able to stand up omicron and confirm both inbound & outbound connectivity.

### DIFF
--- a/common/src/api/internal/shared.rs
+++ b/common/src/api/internal/shared.rs
@@ -111,6 +111,8 @@ pub struct RouteConfig {
     pub destination: IpNetwork,
     /// The nexthop/gateway address.
     pub nexthop: IpAddr,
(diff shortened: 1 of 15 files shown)


andrewjstone (approved):
Awesome fix. Thanks @luqmana!

rcgoodfellow:
Note that setting the VLAN id on routes is going away entirely. The implementation was too costly on the Tofino, and we had to remove it for another feature as it did not wind up getting used.

PR #4325, opened 2023-10-24
Refactor InstalledZone::install to use a builder pattern, per TODO.

lifning:
Additionally, make a builder-factory with an option to create fake builders, in service of refactoring some things to enable some unit tests being written.

### DIFF
--- a/sled-agent/src/instance.rs
+++ b/sled-agent/src/instance.rs
@@ -26,7 +26,7 @@ use futures::lock::{Mutex, MutexGuard};
 use illumos_utils::dladm::Etherstub;
 use illumos_utils::link::VnicAllocator;
 use illumos_utils::opte::{DhcpCfg, PortManager};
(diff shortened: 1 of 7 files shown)


sled-agent/src/instance.rs
```diff
@@ -905,31 +910,27 @@ impl Instance {
...
+            .builder()
+            .with_log(inner.log.clone())
+            .with_underlay_vnic_allocator(&inner.vnic_allocator)
+            .with_zone_root_path(&root)
+            .with_zone_image_paths(&["/opt/oxide".into()])
+            .with_zone_type("propolis-server")
+            .with_datasets(&[])
+            .with_filesystems(&[])
+            .with_data_links(&[])
+            .with_devices(&[
```
citrus-it, line 924: This is missing `.with_unique_name()` - was that intentional?

lifning: sure wasn't! thanks

citrus-it (approved):

PR #4326, opened 2023-10-24
[wicketd] Fall back to old behavior if SP is too old to support reading RoT CMPA/CFPA

jgallagher:
We hit this on rack3 but did not hit it on dogfood due to more frequent updates to dogfood's SP/RoT. Prior to this PR, wicketd expected to be able to ask an SP for its RoT's CMPA/CFPA pages, but if a rack is jumping from the 1.0.2 release to current master, its SPs are too old to understand that message. With this change, we will fall back to wicketd's previous behavior of requiring exactly 1 RoT archive if we fail to fetch the CMPA from the target component.

### DIFF
--- a/wicketd/src/update_tracker.rs
+++ b/wicketd/src/update_tracker.rs
@@ -1645,11 +1645,19 @@ impl UpdateContext {
                     artifact_to_apply = Some(available_artifacts[0].clone());
                     None
                 } else {
+                    error!(
+                        self.log,
+                        "Failed to get RoT CMPA; unable to choose from \
+                         multiple available RoT artifacts";
+                        "err" => %err,
+                        "num_rot_artifacts" => available_artifacts.len(),
+                    );
                     return Err(UpdateTerminalError::GetRotCmpaFailed {
                         error: err.into(),
                     });
                 }
             }
+            // For any other error (e.g., comms failures), just fail as normal.
             Err(err) => {
                 return Err(UpdateTerminalError::GetRotCmpaFailed {
                     error: err.into(),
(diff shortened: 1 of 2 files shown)


wicketd/src/update_tracker.rs
```diff
@@ -1608,90 +1608,131 @@ impl UpdateContext {
...
-                Err(err) => {
-                    // We log this but don't fail - we want to continue looking
-                    // for a verifiable artifact.
-                    info!(
-                        self.log, "RoT archive verification failed";
-                        "artifact" => ?artifact.id,
-                        "err" => %DisplayErrorChain::new(&err),
-                    );
+            }
+            Err(err) => {
```
mkeeter, line 1653: Add a comment here that this covers communication errors, etc?

wicketd/src/update_tracker.rs
```diff
@@ -1608,90 +1608,131 @@ impl UpdateContext {
...
-                        "failed to read hubris archive for {:?}",
-                        artifact.id
-                    )),
-                }
-            })?;
-            match archive.verify(&cmpa, &cfpa) {
-                Ok(()) => {
+                })?)
+            }
+            // TODO is there a better way to check the _specific_ error response
```
mkeeter, line 1632: I think this is fine; it's no worse than before!

wicketd/src/update_tracker.rs
```diff
@@ -1608,90 +1608,131 @@ impl UpdateContext {
...
+                        self.log,
+                        "Failed to get RoT CMPA page; \
+                         using only available RoT artifact";
+                        "err" => %err,
                     );
-                    artifact_to_apply = Some(artifact.clone());
-                    break;
+                    artifact_to_apply = Some(available_artifacts[0].clone());
+                    None
+                } else {
```
mkeeter, line 1647: Add some logging here, since it's a weird corner case

PR #4329, opened 2023-10-24
[nexus] [minor] Bring BGP endpoint doc comments in line with conventions

david-crespo:
I noticed there were periods in the docs sidebar. Also used the word "list" as appropriate.

### DIFF
--- a/nexus/src/external_api/http_entrypoints.rs
+++ b/nexus/src/external_api/http_entrypoints.rs
@@ -2609,7 +2609,7 @@ async fn networking_loopback_address_delete(
     apictx.external_latencies.instrument_dropshot_handler(&rqctx, handler).await
 }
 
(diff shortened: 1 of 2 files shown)


nexus/src/external_api/http_entrypoints.rs
```diff
@@ -2609,7 +2609,7 @@ async fn networking_loopback_address_delete(
     apictx.external_latencies.instrument_dropshot_handler(&rqctx, handler).await
 }
 
-/// Get loopback addresses, optionally filtering by id
+/// List loopback addresses
```
david-crespo, line 2612: Didn't think "optionally filtering by id" is necessary, also not sure it's even true

rcgoodfellow (approved):

PR #4338, opened 2023-10-25
Update propolis mock-server dependencies

pfmooney:
This switches simulated sled-agent to used the now-independent propolis mock server from oxidecomputer/propolis#556.

Once that has landed, this PR will need to be updated w/ the as-committed git hash.


pfmooney:
Integrated as part of #4473

PR #4343, opened 2023-10-25
Port settings fixes

### LINKED ISSUE #4327: A failure in the port settings apply saga on update can lead to disassociation of port settings.
I've observed that when port settings are updated through the API, and `dpd` rejects the update, the saga fails and unwinds - but the ports that reference the settings group being updated lose their association with that group.

### LINKED ISSUE #4328: Invalid port settings updates return an HTTP 500 error instead of 400
When an invalid port settings update takes place, for example trying to change the link speed of an active port (this requires deleting and re-creating the port) this comes back as an internal server error with no useful information to the API user. This should be a 400 error with helpful information on why the request was invalid.

rcgoodfellow:
- Fixes  #4327 
- Fixes #4328
- Pulls in 
  - oxidecomputer/dendrite#656
  - oxidecomputer/dendrite#654
  - oxidecomputer/dendrite#662
  - oxidecomputer/dendrite#663
  - oxidecomputer/maghemite#81

## Depends on
- oxidecomputer/dendrite#664


Nieuwejaar (approved):
I was literally about to send you a review with about half of this stuff in it already.

Issue #4335, opened 2023-10-24
`oximeter` can have schema in its cache that are not in the database


bnaecker:
I'm testing some code that adds a new `oximeter` producer into the sled agent. I spun up Omicron on my non-Gimlet machine, to see how the new timeseries find their way into ClickHouse. `oximeter` pulled the samples from the sled-agent and extracted their schema. It found no matches in its own internal cache, so it inserted them. But then the request to the DB itself to insert those into the `timeseries_schema` table failed. Here's a snippet from the `oximeter` logs:

```
{"msg":"new timeseries schema: \"{\\\"timeseries_name\\\":\\\"physical_data_link:packets_sent\\\",\\\"fields.name\\\":[\\\"hostname\\\",\\\"link_name\\\",\\\"rack_id\\\",\\\"serial\\\",\\\"sled_id\\\"],\\\"fields.type\\\":[\\\"String\\\",\\\"String\\\",\\\"Uuid\\\",\\\"String\\\",\\\"Uuid\\\"],\\\"fields.source\\\":[\\\"Target\\\",\\\"Target\\\",\\\"Target\\\",\\\"Target\\\",\\\"Target\\\"],\\\"datum_type\\\":\\\"CumulativeU64\\\",\\\"created\\\":\\\"2023-10-24 22:00:02.549459360\\\"}\"","v":0,"name":"oximeter","level":20,"time":"2023-10-24T22:00:02.549522771Z","hostname":"oxz_oximeter_cf4204d9-7033-4455-ae92-19fcd8635fcc","pid":25112,"id":"7c50c14e-41a8-447a-9db4-a8a54ce0f289","component":"clickhouse-client","collector_id":"cf4204d9-7033-4455-ae92-19fcd8635fcc","component":"oximeter-agent"}
{"msg":"inserting 6 new timeseries schema","v":0,"name":"oximeter","level":20,"time":"2023-10-24T22:00:02.571792413Z","hostname":"oxz_oximeter_cf4204d9-7033-4455-ae92-19fcd8635fcc","pid":25112,"id":"7c50c14e-41a8-447a-9db4-a8a54ce0f289","component":"clickhouse-client","collector_id":"cf4204d9-7033-4455-ae92-19fcd8635fcc","component":"oximeter-agent"}
[...]

PR #4348, opened 2023-10-25
Remove timeseries schema from cache on insertion failure

bnaecker:
- Fixes https://github.com/oxidecomputer/omicron/issues/4335
- Factors out internals of `oximeter_db::Client::insert_samples()` to aid testing, and add regression
- Remove new timeseries schema from the internal cache, if we fail to insert them into the schema table.

### DIFF
--- a/oximeter/db/src/client.rs
+++ b/oximeter/db/src/client.rs
@@ -629,7 +629,7 @@ impl Client {
         UnrolledSampleRows { new_schema, rows }
     }
 
-    // Flush new schema to the database, or remove them from the cache on
+    // Save new schema to the database, or remove them from the cache on
     // failure.
     //
     // This attempts to insert the provided schema into the timeseries schema
@@ -647,7 +647,7 @@ impl Client {
     //
     // NOTE: This is an issue even in the case where the schema don't conflict. Two clients may
     // receive a sample with a new schema, and both would then try to insert that schema.
-    async fn flush_new_schema_or_remove(
+    async fn save_new_schema_or_remove(
         &self,
         new_schema: BTreeMap<TimeseriesName, String>,
     ) -> Result<(), Error> {
@@ -761,7 +761,7 @@ impl DbWrite for Client {
         debug!(self.log, "unrolling {} total samples", samples.len());
         let UnrolledSampleRows { new_schema, rows } =
             self.unroll_samples(samples).await;
-        self.flush_new_schema_or_remove(new_schema).await?;
+        self.save_new_schema_or_remove(new_schema).await?;
         self.insert_unrolled_samples(rows).await
     }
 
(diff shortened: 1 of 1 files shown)


oximeter/db/src/client.rs
```diff
@@ -642,34 +626,78 @@ impl DbWrite for Client {
...
+                    self.log,
+                    "failed to insert new schema, removing from cache";
+                    "error" => ?e,
+                );
+                let mut schema = self.schema.lock().await;
+                for name in new_schema.keys() {
+                    schema
+                        .remove(name)
+                        .expect("New schema should have been cached");
+                }
```
david-crespo, line 689: Is it correct to say that overzealously removing from the cache is low-risk, because the next time we try to do anything with a schema with a given name, we will check the DB for it anyway? I'm thinking about a scenario where there is an old value `V0` for `name` in `schema` and a new one for it `V1` in `new_schema`, we fail to insert `new_schema`, and then here we remove `(name, V0)` from `schema` even though it's not the one that failed to be inserted.

oximeter/db/src/client.rs
```diff
@@ -642,34 +626,78 @@ impl DbWrite for Client {
...
+    // tables. It's not clear how to handle this, since ClickHouse provides no transactions.
+    // This is unlikely to happen at this point, because the design is such that there will be
+    // a single `oximeter` instance, which has one client object, connected to a single
+    // ClickHouse server. But once we start replicating data, the window within which the race
+    // can occur is much larger, since it includes the time it takes ClickHouse to replicate
+    // data between nodes.
+    //
+    // NOTE: This is an issue even in the case where the schema don't conflict. Two clients may
+    // receive a sample with a new schema, and both would then try to insert that schema.
+    async fn flush_new_schema_or_remove(
```
david-crespo, line 650: To me `flush` connotes moving something from one place to another, so like if we're flushing to the DB, we're removing it from memory. Since I don't think that's the case here, I would call this `write` or `save` or something.

oximeter/db/src/client.rs
bnaecker: I think it is low-risk, yeah. I considered just nuking the cache entirely, but that seems a bigger change than I wanted to fit in right before a release.

I'm not sure that the scenario you describe is possible though. If we get to the point of finding two schema for `name`, `V0` and `V1`, then they have to be equal or we ignore the sample in any case. So at this point in the code, we have exactly one schema in the cache, and at most one in the database. That sound right?

oximeter/db/src/client.rs
bnaecker: `save` is fine too, yeah. I meant it in the same way _flush_ is used to describe a disk operation -- take all the things I asked the disk to do, which may have been buffered, and wait until they've been actually performed by the underlying storage medium.

bnaecker: Renamed in 7699c2572

oximeter/db/src/client.rs
david-crespo: Sounds good!

david-crespo (approved):

Issue #4336, opened 2023-10-24
`oximeter` is missing `datum_type`s in its schema table


bnaecker:
As part of #4091, we added support for a bunch of new numeric types to `oximeter`. We added new tables to support each those, and a bunch of tests that we can insert and recall the values from the right table. However, we neglected to put those new types in [the `timeseries_schema.datum_type` column](https://github.com/oxidecomputer/omicron/blob/9daf99b870865044ed781107605791752f0eec94/oximeter/db/src/db-single-node-init.sql#L470-L480), which means we cannot store the _schema_ for timeseries using those new types. Trying yields this in the `oximeter` logs:

```
22:24:58.347Z DEBG oximeter (oximeter-agent): inserting 6 new timeseries schema
    collector_id = cf4204d9-7033-4455-ae92-19fcd8635fcc
    id = fab4dcd7-6652-4485-9b6a-b76691917bc2
22:24:58.366Z WARN oximeter (oximeter-agent): failed to insert some results into metric DB: Error interacting with telemetry database: Code: 36. DB::Exception: Unknown element 'CumulativeU64' for enum, maybe you meant: ['CumulativeI64']: (while reading the value of key datum_type): While executing ParallelParsingBlockInputFormat. (BAD_ARGUMENTS) (version 22.8.9.1)
    collector_id = cf4204d9-7033-4455-ae92-19fcd8635fcc
    file = oximeter/collector/src/lib.rs:369
```

We need to either update that table definition, and bump the version, or build tooling to run `ALTER TABLE` statements at runtime for changes that are backwards compatible like this one.

karencfv:
Related: https://github.com/oxidecomputer/omicron/issues/4048

PR #4349, opened 2023-10-25
Add new datum types to timeseries schema table

bnaecker:
- Fixes https://github.com/oxidecomputer/omicron/issues/4336
- Adds regression test making sure a query selecting all datum types from the schema table succeeds, even if it returns zero results. This uses an iterator over the datum type enum, so it should catch future changes to the type.
- Adds new datum types to schema table
- Bumps oximeter database version number


bnaecker:
Note that this will _wipe the existing database_. This should take effect as part of the one (planned) wipe at customer sites, but not need another one.

david-crespo (approved):

PR #4371, opened 2023-10-26
omdb could care less about the current database version

davepacheco:
`omdb` has a cheesy test to check that its output doesn't change accidentally.  It currently captures the current database schema version.  We should redact this the way we do timestamps, uuids, etc. so that people don't have to update this test when they bump the database version.


leftwo (approved):
LG, Thanks for this.

Issue #4148, opened 2023-09-27
[ClickHouse] Remove single node functionality


karencfv:
Once replicated mode has been [enabled](https://github.com/oxidecomputer/omicron/issues/4000), there will be no further use for single node ClickHouse. The init SQL is wildly different between replicated and single node mode, this means that any testing and/or development done on a single node will be unreliable on a replicated node setup.

Several parts of the codebase will need to be modified:

- [ ] Init SQL files
- [ ] Unit tests - https://github.com/oxidecomputer/omicron/pull/4372
- [ ] Integration tests
- [ ] Simulated omicron - https://github.com/oxidecomputer/omicron/pull/4422
- [ ] Oximeter cli (OG oxdb)

karencfv:
Unassigned myself from this as I have to focus on other work for the time being

bnaecker:
I recommend that we close this as not planned. I believe we'll always want the ability to run single-node functionality for tests that interact with the `oximeter` database itself. Many of those tests do not care if their running against a replicated or single-node table, e.g., something that just inserts data and fetches it back out to verify serialization. For those tests which _do_ care, we already have the ability to start a replicated cluster.

karencfv:
Heya @bnaecker! I raise this point in the "Open Questions" [section](https://rfd.shared.oxide.computer/rfd/0468#_open_questions) of RFD 468. Do you mind writing your concerns there so we can discuss there and keep a record of why we made a choice over another?

bnaecker:
What I'm talking about is orthogonal to how we deploy ClickHouse on production racks. I don't have much insight to add to whether we should remove deployment of single-node ClickHouse zones in those environments. You and @andrewjstone have thought much more deeply about that at this point that me!

I'm asking more about test and development environments. This issue seems to say we should remove the ability to run single-node ClickHouse **completely**, including in unit tests, integration tests, or other development situations. I disagree. I find it extremely valuable to quickly run tests that rely on an `oximeter` database, and spinning up a single-node version is many times faster and more reliable than a cluster. I also think it's a non-starter to completely serialize every single Nexus integration test, as we'd have to if we want to reliably run ClickHouse clusters during those tests.

I'd be fine keeping this issue open, if you'd prefer to refine it so that it refers to production environments, e.g., the deployment of single-node ClickHouse zones by the sled-agent's service manager.

karencfv:
[...]

Issue #4001, opened 2023-08-31
[ClickHouse] Improve replicated mode testing


karencfv:
In https://github.com/oxidecomputer/omicron/pull/3494 tests that validate the replicated db_init file [were added](https://github.com/oxidecomputer/omicron/pull/3494/files#diff-6f5af870905bd92fc2c62db62d674d1e033edee57adbe7ea70d929c79cd03ba1R671), and  [checks](https://github.com/oxidecomputer/omicron/pull/3494/files#diff-6f5af870905bd92fc2c62db62d674d1e033edee57adbe7ea70d929c79cd03ba1R664) were incorporated in tests that validate whether a CH instance is part of a cluster or not.

More testing should be implemented and the current tests could be refined.

karencfv:
All single node tests run against a multinode cluster now

PR #4372, opened 2023-10-27
[oximeter] Port all single node ClickHouse tests to a replicated set up

karencfv:
This is a follow up to https://github.com/oxidecomputer/omicron/pull/4149 . In this commit all single node tests are now run against a clustered set up. Additionally, a few bugs on the replicated init SQL file have been fixed.

A few things to note:

- All of the individual field and measurement tests all run under a single test now. The reason for this is test times. Each test needs to wipe out it's database before the next test runs. In a replicated installation, this means that all nodes must sync, which takes time. Before I merged all the field/measurement tests into a single one, the testing time for a replicated set up took about 10 minutes, which really is too much. Now it takes ~3.5 minutes. It is still a lot, but only a temporary measure until we can run these tests in parallel again. An unintended bonus of this approach is that running the tests this way means that you are testing that the data is being inserted in the correct table.
- ~There is a strange behaviour that occurs when running the `recall_measurement_f32_test` test. This test is unique in the fact that it has to insert an f64 value to the `oximeter.measurements_f32` table. For some strange reason the data is inserted into the `oximeter.measurements_f64` table as well. I honestly could not find out why. It seems to be some strange ClickHouse quirk, so I have disabled the test temporarily.~

  ~I tweaked the test to return 2 rows instead of one and these are the results. As you can see, an additional row is inserted into the `oximeter.measurements_f64` at the same time the `recall_measurement_f32_test` test is run.~

  ~Output from the log:~

  ~{"msg":"executing SQL query: INSERT INTO oximeter.measurements_f32 FORMAT JSONEachRow {\"datum\":1.1,\"timeseries_key\":101,\"timeseries_name\":\"foo:bar\",\"timestamp\":\"2023-10-27 00:52:35.601821001\"}","v":0,"name":"test_recall_of_all_fields","level":10,"time":"2023-10-27T00:52:35.602139961Z","hostname":"pop-os","pid":846472,"id":"47b615cf-5e40-4ff8-8700-a48cc97a4830","component":"clickhouse-client"}~
  ~{"msg":"executing SQL query: SELECT * FROM oximeter.measurements_f32 LIMIT 2 FORMAT JSONEachRow;","v":0,"name":"test_recall_of_all_fields","level":10,"time":"2023-10-27T00:52:35.613187591Z","hostname":"pop-os","pid":846472,"id":"47b615cf-5e40-4ff8-8700-a48cc97a4830","component":"clickhouse-client"}~
  ~{"msg":"executing SQL query: INSERT INTO oximeter.measurements_f64 FORMAT JSONEachRow {\"datum\":1.1,\"timeseries_key\":101,\"timeseries_name\":\"foo:bar\",\"timestamp\":\"2023-10-27 00:52:35.614849622\"}","v":0,"name":"test_recall_of_all_fields","level":10,"time":"2023-10-27T00:52:35.614945438Z","hostname":"pop-os","pid":846472,"id":"47b615cf-5e40-4ff8-8700-a48cc97a4830","component":"clickhouse-client"}~
  ~{"msg":"executing SQL query: SELECT * FROM oximeter.measurements_f64 LIMIT 2 FORMAT JSONEachRow;","v":0,"name":"test_recall_of_all_fields","level":10,"time":"2023-10-27T00:52:35.624995473Z","hostname":"pop-os","pid":846472,"id":"47b615cf-5e40-4ff8-8700-a48cc97a4830","component":"clickhouse-client"}~

  ~Test output:~
  ~Object {"datum": Number(1), "timeseries_key": Number(101), "timeseries_name": String("foo:bar"), "timestamp": String("2023-10-27 00:52:35.588811834")}~
  ~Object {"datum": Number(1), "timeseries_key": Number(101), "timeseries_name": String("foo:bar"), "timestamp": String("2023-10-27 00:52:35.588811834")}~
  ~{"timeseries_name":"foo:bar","timeseries_key":101,"timestamp":"2023-10-27 00:52:35.601821001","datum":1.1}~
  
  ~Object {"datum": Number(1.1), "timeseries_key": Number(101), "timeseries_name": String("foo:bar"), "timestamp": String("2023-10-27 00:52:35.601821001")}~
  ~Object {"datum": Number(1.1), "timeseries_key": Number(101), "timeseries_name": String("foo:bar"), "timestamp": String("2023-10-27 00:52:35.601821001")}~
  ~{"timeseries_name":"foo:bar","timeseries_key":101,"timestamp":"2023-10-27 00:52:35.601821001","datum":1.100000023841858}~
  ~{"timeseries_name":"foo:bar","timeseries_key":101,"timestamp":"2023-10-27 00:52:35.614849622","datum":1.1}~

UPDATE: Tyniest of typos in the init sql file was to blame for the above bug

Related: https://github.com/oxidecomputer/omicron/issues/4148
Related: https://github.com/oxidecomputer/omicron/issues/4001

### DIFF
--- a/oximeter/db/src/client.rs
+++ b/oximeter/db/src/client.rs
@@ -865,10 +865,14 @@ mod tests {
 
         pub async fn wipe_db(&self, client: &Client) -> Result<(), Error> {
             match *self {
-                InstallationType::SingleNode => {
-                    client.wipe_single_node_db().await.expect("Failed to remove timeseries database")
-                }
-                InstallationType::Cluster => client.wipe_replicated_db().await.expect("Failed to remove timeseries database"),
+                InstallationType::SingleNode => client
+                    .wipe_single_node_db()
+                    .await
+                    .expect("Failed to remove timeseries database"),
+                InstallationType::Cluster => client
+                    .wipe_replicated_db()
+                    .await
+                    .expect("Failed to remove timeseries database"),
             }
             Ok(())
         }
@@ -3054,7 +3058,7 @@ mod tests {
             .await
             .expect_err("Should have failed, downgrades are not supported");
 
-            db_type.wipe_db(&client).await.unwrap();
+        db_type.wipe_db(&client).await.unwrap();
         logctx.cleanup_successful();
         Ok(())
     }
--- a/oximeter/db/src/client.rs
+++ b/oximeter/db/src/client.rs
@@ -873,151 +873,347 @@ mod tests {
             .unwrap();
 
         // Tests that data can be inserted via the client
-        insert_samples_test(db.address).await.unwrap();
+        insert_samples_test(db.address, InstallationType::SingleNode)
+            .await
+            .unwrap();
 
         // Tests for a schema mismatch
-        schema_mismatch_test(db.address).await.unwrap();
+        schema_mismatch_test(db.address, InstallationType::SingleNode)
+            .await
+            .unwrap();
 
         // Tests for a schema update
-        schema_updated_test(db.address).await.unwrap();
+        schema_updated_test(db.address, InstallationType::SingleNode)
+            .await
+            .unwrap();
 
         // Tests for specific timeseries selection
-        client_select_timeseries_one_test(db.address).await.unwrap();
+        client_select_timeseries_one_test(
+            db.address,
+            InstallationType::SingleNode,
+        )
+        .await
+        .unwrap();
 
         // Tests for specific timeseries selection
-        field_record_count_test(db.address).await.unwrap();
+        field_record_count_test(db.address, InstallationType::SingleNode)
+            .await
+            .unwrap();
 
         // ClickHouse regression test
-        unquoted_64bit_integers_test(db.address).await.unwrap();
+        unquoted_64bit_integers_test(db.address, InstallationType::SingleNode)
+            .await
+            .unwrap();
 
         // Tests to verify that we can distinguish between metrics by name
-        differentiate_by_timeseries_name_test(db.address).await.unwrap();
+        differentiate_by_timeseries_name_test(
+            db.address,
+            InstallationType::SingleNode,
+        )
+        .await
+        .unwrap();
 
         // Tests selecting a single timeseries
-        select_timeseries_with_select_one_test(db.address).await.unwrap();
+        select_timeseries_with_select_one_test(
+            db.address,
+            InstallationType::SingleNode,
+        )
+        .await
+        .unwrap();
 
         // Tests selecting two timeseries
         select_timeseries_with_select_one_field_with_multiple_values_test(
             db.address,
+            InstallationType::SingleNode,
         )
         .await
         .unwrap();
 
         // Tests selecting multiple timeseries
-        select_timeseries_with_select_multiple_fields_with_multiple_values_test(db.address).await.unwrap();
+        select_timeseries_with_select_multiple_fields_with_multiple_values_test(db.address, InstallationType::SingleNode).await.unwrap();
 
         // Tests selecting all timeseries
-        select_timeseries_with_all_test(db.address).await.unwrap();
+        select_timeseries_with_all_test(
+            db.address,
+            InstallationType::SingleNode,
+        )
+        .await
+        .unwrap();
 
         // Tests selecting all timeseries with start time
-        select_timeseries_with_start_time_test(db.address).await.unwrap();
+        select_timeseries_with_start_time_test(
+            db.address,
+            InstallationType::SingleNode,
+        )
+        .await
+        .unwrap();
 
         // Tests selecting all timeseries with start time
-        select_timeseries_with_limit_test(db.address).await.unwrap();
+        select_timeseries_with_limit_test(
+            db.address,
+            InstallationType::SingleNode,
+        )
+        .await
+        .unwrap();
 
         // Tests selecting all timeseries with order
-        select_timeseries_with_order_test(db.address).await.unwrap();
+        select_timeseries_with_order_test(
+            db.address,
+            InstallationType::SingleNode,
+        )
+        .await
+        .unwrap();
 
         // Tests schema does not change
-        get_schema_no_new_values_test(db.address).await.unwrap();
+        get_schema_no_new_values_test(db.address, InstallationType::SingleNode)
+            .await
+            .unwrap();
 
         // Tests listing timeseries schema
-        timeseries_schema_list_test(db.address).await.unwrap();
+        timeseries_schema_list_test(db.address, InstallationType::SingleNode)
+            .await
+            .unwrap();
 
         // Tests listing timeseries
-        list_timeseries_test(db.address).await.unwrap();
+        list_timeseries_test(db.address, InstallationType::SingleNode)
+            .await
+            .unwrap();
 
         // Tests no changes are made when version is not updated
-        database_version_update_idempotent_test(db.address).await.unwrap();
+        database_version_update_idempotent_test(
+            db.address,
+            InstallationType::SingleNode,
+        )
+        .await
+        .unwrap();
 
         // Tests that downgrading is impossible
-        database_version_will_not_downgrade_test(db.address).await.unwrap();
+        database_version_will_not_downgrade_test(
+            db.address,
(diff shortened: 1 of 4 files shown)


oximeter/db/src/client.rs
```diff
@@ -854,198 +858,524 @@ mod tests {
...
+        let result = client_2.execute_with_body(sql.clone()).await.unwrap();
+        assert!(result.contains("hiya"));
 
-        db.cleanup().await.expect("Failed to cleanup ClickHouse server");
+        client_1.wipe_replicated_db().await?;
         logctx.cleanup_successful();
+
+        // TODO(https://github.com/oxidecomputer/omicron/issues/4001): With distributed
+        // engine, it can take a long time to sync the data. This means it's tricky to
+        // test that the data exists on both nodes.
```
karencfv, line 1334: https://github.com/oxidecomputer/omicron/pull/4149#discussion_r1370775228 didn't really work unfortunately :( I will continue to think about this one, but I don't want to stop this PR as I think it's important to have the rest of the replicated tests running on CI.

bnaecker: What do you mean by "didn't work"? The [`FLUSH DISTRIBUTED`](https://clickhouse.com/docs/en/sql-reference/statements/system#flush-distributed) statement makes a call that synchronously flushes all data to the other nodes. IIUC, we could call that here, and if / when it returns, all data should have been replicated.

bnaecker:
>  There is a strange behaviour that occurs when running the recall_measurement_f32_test test. This test is unique in the fact that it has to insert an f64 value to the oximeter.measurements_f32 table. For some strange reason the data is inserted into the oximeter.measurements_f64 table as well. I honestly could not find out why. It seems to be some strange ClickHouse quirk, so I have disabled the test temporarily.

This seems like something we need to understand. The test implementation [picks the right table here](https://github.com/oxidecomputer/omicron/blob/561f54502ec7dcced9fdd54fc7ab6407343974db/oximeter/db/src/client.rs#L2600C37-L2600C37), and if there's an issue with either that function or the code, we should probably get to ground on it.

oximeter/db/src/client.rs
bnaecker: Also, if we're really not going to test that data inserted on one node is visible on the other, we should not be calling this `data_is_replicated_test()`.

oximeter/db/src/client.rs
```diff
@@ -854,198 +858,524 @@ mod tests {
...
+        Ok(())
     }
 
-    async fn is_not_oximeter_cluster_test(
+    async fn is_oximeter_cluster_test(
         address: SocketAddr,
+        db_type: InstallationType,
     ) -> Result<(), Error> {
         let logctx = test_setup_log("test_is_not_oximeter_cluster");
```
bnaecker, line 1342: Looks like the test name changed, but the log name didn't?

oximeter/db/src/client.rs
```diff
@@ -2101,148 +2631,114 @@ mod tests {
...
-        address: SocketAddr,
+        client: &Client,
     ) -> Result<(), Error> {
         let field = FieldValue::String("foo".into());
         let as_json = serde_json::Value::from("foo");
-        test_recall_field_value_impl(address, field, as_json).await?;
-        Ok(())
-    }
-
-    async fn recall_field_value_ipv4addr_test(
```
bnaecker, line 2199: Did we lose this test?

oximeter/db/src/client.rs
```diff
@@ -2282,157 +2778,141 @@ mod tests {
...
         Ok(())
     }
 
-    async fn recall_measurement_f32_test(
-        address: SocketAddr,
-    ) -> Result<(), Error> {
+    // NOTE: This test is ignored intentionally.
+    //
+    // For some reason when the value is an f64 it inserts in both measurements_f32
+    // and measurements_f64 tables. This behaviour is unexpected, and should be investigated
```
bnaecker, line 2861: This is not something we can ignore in this PR. We can't be inserting measurements into the wrong table. I've looked through the code, and am not sure how this could happen. Do we see it with any other measurement types? The test is run and passes on `main`, so might it be possible that this was due to multiple tests run concurrently?

test-utils/src/dev/clickhouse.rs
```diff
@@ -252,14 +254,15 @@ impl ClickHouseInstance {
             })?;
```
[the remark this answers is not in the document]
karencfv: I'll file a bug report in the ClickHouse repo :)

karencfv:
Thanks for the review @bnaecker :bowing_woman:  I think I've addressed all of your comments, let me know what you think!

bnaecker:
My point wasn’t that we need to file a bug, but that we need to understand the behavior. It may indeed be a bug, but I would be a bit surprised that the server programs would care about the specific addresses they’re using, as long as they can listen on them. 

Can you describe exactly what behavior you see? How did you run the data and keeper servers, and what was the exact error you saw? If we do file a bug against ClickHouse then they’ll want that information as well, so I think it’ll be time well spent.

karencfv:
@bnaecker After looking into this for a while I realised ClickHouse servers have a small quirk, where when setting the keeper hosts as ipV6 localhost addresses in the replica configuration file, they must be wrapped in square brackets (only the keeper hosts, the other hosts are set without square brackets :woman_shrugging: ). Otherwise, when running any query, a "Service not found" error appears.

karencfv:
It's really odd, you use IPv6 addresses without brackets everywhere else. See: https://clickhouse.com/docs/en/operations/server-configuration-parameters/settings#interserver-listen-host, and https://clickhouse.com/docs/en/operations/server-configuration-parameters/settings#listen_host

bnaecker:
@karencfv Thank you for tracking that down. It looks to me like this _is_ a ClickHouse bug. On [this line](https://github.com/ClickHouse/ClickHouse/blob/a011990fd75628c63c7995c4f15475f1d4125d10/src/Coordination/KeeperStateManager.cpp#L149), we can see that the hostname for the keeper server (a string) is concatenated directly with the port converted to a string. So for an IPv6 localhost address, _provided as a numeric value_, this would result in a socket address like `::1:12345`. That might be correct for a symbolic hostname, like `localhost:12345`, but for IPv6, which separates segments with `:`, one needs to surround it with brackets, so that we get `[::1]:12345`.

This is good to have discovered. We need to make sure to provide the keeper addresses as either symbolic hostnames for the server to resolve, or surround them with brackets ourselves. I will also file a bug against CH, since this could certainly be fixed internally.

karencfv:
> This is good to have discovered. We need to make sure to provide the keeper addresses as either symbolic hostnames for the server to resolve, or surround them with brackets ourselves. I will also file a bug against CH, since this could certainly be fixed internally.

Thanks @bnaecker! I've already made the changes in this commit https://github.com/oxidecomputer/omicron/pull/4372/commits/df652c8e0427d482c5e29ecdafede06d401d0782. If everything looks good to you could I get a an approval so I can get this merged?

oximeter/db/src/client.rs
bnaecker: Could you please point me to the commit / line which fixes this one?

karencfv: https://github.com/oxidecomputer/omicron/pull/4372/commits/0b77159cfe6184417be62ac01b13e6e7537087c2#diff-d7e0c19781820dffa1f107bb7e5b9b78bc9f79d85255c3457010f989987d83d3R209

oximeter/db/src/client.rs
bnaecker: Ok, I think that's reasonable for this particular test, thanks.

In general, I think minimizing the test duration is not the right tradeoff. The absolute duration numbers are quite small (a few seconds or 10s of seconds), so a large relative change doesn't make for a substantially worse developer experience. In any case, by running all the test _implementations_ in a single Rust `#[test]` function, we're losing the ability to run exactly which tests we want. I think this actually _slows down_ iteration, since I can't just run one test to confirm new features or bug fixes. I need to run all of them and sift through the output to find that for the test I'm actually working on. I feel that robust, complete, and repeatable testing is more important than _fast_ tests as we implement multi-node ClickHouse.

bnaecker (approved):

PR #4378, opened 2023-10-27
Add sagas subcommand to omdb

### LINKED ISSUE #4157: omdb: Saga state tracking
During a debug session on 9/28, we jumped into a cockroachdb shell to manually query for saga state.

Of note:
- What sagas are running?
- What sagas failed?
- For sagas that failed, what were they? Where did they fail? What resources were they acting on?

This information would be a great candidate for an omdb query command -- if we could grab this info without using `cockroach sql`, we would.

jmpesp:
Provide a `sagas` subcommand to omdb:

    $ ./target/debug/omdb db sagas --help
    Print information related to sagas

    Usage: omdb db sagas <COMMAND>

    Commands:
      list         List all sagas
      stuck        List sagas that failed to unwind
      show         Show the execution of a saga
      unwound      List sagas that encountered an error and successfully unwound
      failing      Show any failing nodes
      overlapping  List any sagas which overlap execution with this one
      interleave   Show multiple saga executions along a timeline
      help         Print this message or the help of the given subcommand(s)

    Options:
      -h, --help  Print help

Also, add `region` subcommand:

    $ ./target/debug/omdb db regions --help
    Print information related to regions

    Usage: omdb db regions <COMMAND>

    Commands:
      list     List all regions
      used-by  Find what is using a region
      help     Print this message or the help of the given subcommand(s)

    Options:
      -h, --help  Print help

Fixes #4157.

### DIFF
--- a/dev-tools/omdb/src/bin/omdb/db.rs
+++ b/dev-tools/omdb/src/bin/omdb/db.rs
@@ -19,7 +19,9 @@ use crate::Omdb;
 use anyhow::anyhow;
 use anyhow::bail;
 use anyhow::Context;
+use async_bb8_diesel::AsyncConnection;
 use async_bb8_diesel::AsyncRunQueryDsl;
+use async_bb8_diesel::AsyncSimpleConnection;
 use chrono::SecondsFormat;
 use clap::Args;
 use clap::Subcommand;
@@ -30,6 +32,9 @@ use diesel::BoolExpressionMethods;
 use diesel::ExpressionMethods;
 use diesel::JoinOnDsl;
 use diesel::NullableExpressionMethods;
+use nexus_db_model::saga_types::Saga;
+use nexus_db_model::saga_types::SagaId;
+use nexus_db_model::saga_types::SagaNodeEvent;
 use nexus_db_model::Dataset;
 use nexus_db_model::Disk;
 use nexus_db_model::DnsGroup;
@@ -37,10 +42,12 @@ use nexus_db_model::DnsName;
 use nexus_db_model::DnsVersion;
 use nexus_db_model::DnsZone;
 use nexus_db_model::ExternalIp;
+use nexus_db_model::Image;
 use nexus_db_model::Instance;
 use nexus_db_model::Project;
 use nexus_db_model::Region;
 use nexus_db_model::Sled;
+use nexus_db_model::Snapshot;
 use nexus_db_model::Vmm;
 use nexus_db_model::Zpool;
 use nexus_db_queries::context::OpContext;
@@ -56,6 +63,10 @@ use nexus_types::internal_api::params::Srv;
 use omicron_common::api::external::DataPageParams;
 use omicron_common::api::external::Generation;
 use omicron_common::postgres_config::PostgresConfigWithUrl;
+use petgraph::graph::NodeIndex;
+use petgraph::Graph;
+use serde::Deserialize;
+use serde::Serialize;
 use std::cmp::Ordering;
 use std::collections::BTreeMap;
 use std::collections::HashSet;
@@ -136,6 +147,10 @@ enum DbCommands {
     Instances,
     /// Print information about the network
     Network(NetworkArgs),
+    /// Print information related to sagas
+    Sagas(SagaArgs),
+    /// Print information related to regions
+    Regions(RegionArgs),
 }
 
 #[derive(Debug, Args)]
@@ -234,6 +249,89 @@ struct NetworkArgs {
 enum NetworkCommands {
     /// List external IPs
     ListEips,
+}
+
+#[derive(Debug, Args)]
+struct SagaArgs {
+    #[command(subcommand)]
+    command: SagaCommands,
+}
+
+#[derive(Debug, Subcommand)]
+enum SagaCommands {
+    /// List all sagas
+    List(SagaListArgs),
+
+    /// List sagas that failed to unwind
+    Stuck,
+
+    /// Show the execution of a saga
+    Show(SagaShowArgs),
+
+    /// List sagas that encountered an error and successfully unwound
+    Unwound,
+
+    /// Show any failing nodes
+    Failing,
+
+    /// List any sagas which overlap execution with this one
+    Overlapping(SagaOverlappingArgs),
+
+    /// Show multiple saga executions along a timeline
+    Interleave(SagaInterleaveArgs),
+}
+
+#[derive(Debug, Args)]
+struct SagaListArgs {
+    /// Show saga starting node params
+    #[arg(short)]
+    show_params: bool,
+}
+
+#[derive(Debug, Args)]
+struct SagaShowArgs {
+    saga_id: Uuid,
+}
+
+#[derive(Debug, Args)]
+struct SagaOverlappingArgs {
+    saga_id: Uuid,
+
+    /// Print saga IDs only
+    #[arg(short)]
+    id_only: bool,
+}
+
+#[derive(Debug, Args)]
+struct SagaInterleaveArgs {
+    saga_id: Vec<Uuid>,
+}
+
+#[derive(Debug, Args)]
+struct RegionArgs {
+    #[command(subcommand)]
+    command: RegionCommands,
+}
+
+#[derive(Debug, Subcommand)]
+enum RegionCommands {
+    /// List all regions
+    List(RegionListArgs),
+
+    /// Find what is using a region
+    UsedBy(RegionUsedByArgs),
+}
+
+#[derive(Debug, Args)]
+struct RegionListArgs {
+    /// Print region IDs only
+    #[arg(short)]
+    id_only: bool,
+}
+
+#[derive(Debug, Args)]
+struct RegionUsedByArgs {
+    region_id: Vec<Uuid>,
 }
 
 impl DbArgs {
@@ -341,6 +439,67 @@ impl DbArgs {
             }) => {
                 cmd_db_eips(&opctx, &datastore, self.fetch_limit, *verbose)
                     .await
+            }
+            DbCommands::Sagas(SagaArgs {
+                command: SagaCommands::List(saga_list_args),
+            }) => {
+                cmd_db_sagas_list(&datastore, self.fetch_limit, saga_list_args)
+                    .await
+            }
+            DbCommands::Sagas(SagaArgs { command: SagaCommands::Stuck }) => {
+                cmd_db_sagas_list_stuck(&datastore, self.fetch_limit).await
+            }
+            DbCommands::Sagas(SagaArgs {
+                command: SagaCommands::Show(saga_show_args),
+            }) => {
+                cmd_db_sagas_show(&datastore, self.fetch_limit, saga_show_args)
+                    .await
+            }
+            DbCommands::Sagas(SagaArgs { command: SagaCommands::Unwound }) => {
+                cmd_db_sagas_list_unwound(&datastore, self.fetch_limit).await
+            }
+            DbCommands::Sagas(SagaArgs { command: SagaCommands::Failing }) => {
+                cmd_db_sagas_list_failing(&datastore, self.fetch_limit).await
+            }
+            DbCommands::Sagas(SagaArgs {
+                command: SagaCommands::Overlapping(saga_overlapping_args),
+            }) => {
+                cmd_db_sagas_list_overlapping(
+                    &datastore,
+                    self.fetch_limit,
+                    saga_overlapping_args,
+                )
+                .await
+            }
+            DbCommands::Sagas(SagaArgs {
+                command: SagaCommands::Interleave(saga_interleave_args),
+            }) => {
+                cmd_db_sagas_list_interleave(
+                    &datastore,
+                    self.fetch_limit,
+                    saga_interleave_args,
+                )
+                .await
+            }
+            DbCommands::Regions(RegionArgs {
+                command: RegionCommands::List(region_list_args),
+            }) => {
+                cmd_db_regions_list(
+                    &datastore,
+                    self.fetch_limit,
+                    region_list_args,
+                )
+                .await
+            }
+            DbCommands::Regions(RegionArgs {
+                command: RegionCommands::UsedBy(region_used_by_args),
+            }) => {
+                cmd_db_regions_used_by(
+                    &datastore,
+                    self.fetch_limit,
+                    region_used_by_args,
+                )
+                .await
             }
         }
     }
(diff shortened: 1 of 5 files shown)


davepacheco:
Assuming this is the same implementation that's being used to collect data from customers: I see some dates printed like "2023-10-26 21:50:27.623424 UTC".  What do you think of printing that as `2023-10-26T21:50:27.623424Z`?  I haven't checked whether the first one _isn't_ ISO 8601, but the latter is what I see much more often.

davepacheco:
(feel free to defer all of this for this PR -- I just wasn't sure where else to put this feedback)

- In the printed saga event log, I think it'd be useful to distinguish between the action name and the node name.  The action name is needed to find the source code but the node name is needed to line up the output of one node with an input of another.
- It'd be useful to report which Nexus was responsible for a saga (this is in the saga table).  (That'd let you go straight to the right log file instead of grep'ing all of the Nexus log files for that saga id.)

jmpesp:
@davepacheco I agree with your suggestions and have addressed them, along with merging with main :)

jmpesp:
@davepacheco what do you think? :)

dev-tools/omdb/src/bin/omdb/db.rs
```diff
@@ -425,6 +442,92 @@ enum NetworkCommands {
...
+    Unwound,
+
+    /// Show any failing nodes
+    Failing,
+
+    /// List any sagas which overlap execution with this one
+    Overlapping(SagaOverlappingArgs),
+
+    /// Show multiple saga executions along a timeline
+    Interleave(SagaInterleaveArgs),
```
davepacheco, line 502: These sound pretty neat!  Do you have example output?  (I guess it would be hard to write an automated test for this?)

dev-tools/omdb/src/bin/omdb/db.rs
```diff
@@ -425,6 +442,92 @@ enum NetworkCommands {
...
+}
+
+#[derive(Debug, Args)]
+struct SagaListArgs {
+    /// Show saga starting node params
+    #[arg(short)]
```
davepacheco, line 508: Is there any reason to not have a long arg name here?  (I think I'd do that for everything.)

If we're going to have a short one, maybe `-p` (for "params") would be better than the default `-s` for "show"?

dev-tools/omdb/src/bin/omdb/db.rs
```diff
@@ -425,6 +442,92 @@ enum NetworkCommands {
...
+    saga_id: Uuid,
+
+    /// Print saga IDs only
+    #[arg(short)]
+    id_only: bool,
+}
+
+#[derive(Debug, Args)]
+struct SagaInterleaveArgs {
+    saga_id: Vec<Uuid>,
```
davepacheco, line 528:
```suggestion
    saga_ids: Vec<Uuid>,
```

dev-tools/omdb/src/bin/omdb/db.rs
```diff
@@ -2249,6 +2416,1221 @@ async fn cmd_db_network_list_vnics(
     Ok(())
 }
 
+// Copy some types from Steno, because steno uses pub(crate) everywhere. We
+// don't want to change Steno to make the internals public, but these types
+// should be fairly stable.
```
davepacheco, line 2421: We don't have to do this here but I think exporting the types from Steno (maybe under a "debug" feature to discourage general use? I don't know) would be preferable to duplicating them.  I don't think there's reason to assume these are particularly stable.  We don't change steno that much but we did change these types significantly last year and this is definitely in the bucket of things I would assume we could change in Steno without breaking people.  By duplicating them we won't realize we've broken them.

(edit: more below)

dev-tools/omdb/src/bin/omdb/db.rs
```diff
@@ -2249,6 +2416,1221 @@ async fn cmd_db_network_list_vnics(
...
+    Action { name: String, label: String, action_name: String },
+    Constant { name: String, value: Arc<serde_json::Value> },
+    SubsagaStart { saga_name: String, params_node_name: String },
+    SubsagaEnd { name: String },
+}
+
+fn print_sagas(sagas: Vec<Saga>, with_start_params: bool) {
+    if with_start_params {
+        // Using Tabled isn't recommended: the data column could be huge, and Tabled
+        // will print spaces in order to make each column the same size
```
davepacheco, line 2457: I don't quite follow this.  What does this do differently from `Tabled` here?

dev-tools/omdb/src/bin/omdb/db.rs
```diff
@@ -2249,6 +2416,1221 @@ async fn cmd_db_network_list_vnics(
...
+    args: &SagaListArgs,
+) -> Result<(), anyhow::Error> {
+    let sagas = datastore
+        .pool_connection_for_tests()
+        .await?
+        .transaction_async(|conn| async move {
+            use db::schema::saga::dsl;
+
+            // Sorting by time_created requires a full table scan with the
+            // current index definitions.
```
davepacheco, line 2882: I think I'd just add the index here.  It would really suck if we ran this on a production system that had run a lot of sagas and it just ate up a bunch of database memory or tied up resources for a while executing the query.

dev-tools/omdb/src/bin/omdb/db.rs
```diff
@@ -2249,6 +2416,1221 @@ async fn cmd_db_network_list_vnics(
...
+            conn.batch_execute_async(ALLOW_FULL_TABLE_SCAN_SQL).await?;
+
+            db::paginated(
```
davepacheco, line 2885: I think if you're only making one query, you don't really need `paginated` right?  I don't think we use that elsewhere here?

dev-tools/omdb/src/bin/omdb/db.rs
```diff
@@ -2249,6 +2416,1221 @@ async fn cmd_db_network_list_vnics(
...
+            .await?
+            .into_iter()
+            .collect();
+
+            // A saga successfully unwound if it started unwinding and didn't
+            // fail for any of the nodes.
+            let sagas: Vec<_> = undo_started_sagas
+                .difference(&undo_failed_sagas)
+                .cloned()
+                .collect();
```
davepacheco, line 3038: This seems like an example of steno internals that could change.  Just thinking out loud: what if instead of either duplicating or exporting Steno-internal types, it had a way of loading up records for a saga (i.e., you'd give it the same objects that it gives you in https://docs.rs/steno/latest/steno/trait.SecStore.html) and then you could ask it the various questions that you're trying to ask here (like: did this one unwind? did it get stuck? what's the name of this node?)

davepacheco (commented):
So sorry that I missed the previous ping on this!

This is obviously pretty low risk so none of this feedback is that important.  There are some nitty suggestions here but my main worry is around duplicating Steno types and embedding logic based around Steno internals.  Fixing this is probably a bunch more work though and obviously it's useful to have these commands so I'm not sure it makes sense to go fix that now.  Maybe we just accept this potential brittleness and see how much of a problem it becomes?  Or land this now but go do that work in parallel and then update this?

dev-tools/omdb/src/bin/omdb/db.rs
davepacheco: While I do think it'd be useful for Steno to expose debug interfaces like this, I'm not sure how much effort it will be and I'm reluctant to block landing this PR on having done that.

I don't think it's too bad if we interpret the saga node events here, but I *would* change the `String::from("undo_started")` stuff to use some value that we know is in sync with steno.  Unfortunately I see that this column is currently typed as a String, which makes this more annoying.  But in nexus/db-model/src/saga_types.rs, there's an `impl TryFrom<SagaNodeEvent> for steno::SagaNodeEvent`, and it translates specific string values that we use sinto their Steno counterparts.  Maybe we could make those constants and use those here?  Or: I see that in the database we use `steno_saga_node_event.event_type.label().to_string()`, so maybe we could use `steno::SagaNodeEvent::UndoStarted.label().to_string()` here?

dev-tools/omdb/src/bin/omdb/db.rs
```diff
@@ -425,6 +442,92 @@ enum NetworkCommands {
...
+    region_id: Vec<Uuid>,
+}
+
+#[derive(Debug, Args)]
+struct SagaArgs {
+    #[command(subcommand)]
+    command: SagaCommands,
+}
+
+#[derive(Debug, Subcommand)]
```
davepacheco, line 485: Just a suggestion: this file is getting pretty huge and I think we should start splitting it up in modules in the same way that the top-level omdb commands are split into modules.  I did that in https://github.com/oxidecomputer/omicron/pull/7695.  Basically, I'd put your new subcommands into `db/saga.rs` and `db/region.rs` and only expose the subcommand enums from them.

dev-tools/omdb/src/bin/omdb/db.rs
```diff
@@ -146,6 +154,11 @@ impl Display for MaybeSledId {
     }
 }
 
+/// Format a chrono::DateTime like "2018-01-26T18:30:09.453Z"
+fn chrono_to_rfc_3999(time: &chrono::DateTime<chrono::Utc>) -> String {
+    time.to_rfc3339_opts(SecondsFormat::Millis, true)
```
davepacheco, line 159: I think this duplicates `datetime_rfc3339_concise()`.

Also, the name is slightly wrong (3999 vs 3339).

dev-tools/omdb/src/bin/omdb/db.rs
davepacheco: Thinking about this more: it does make me nervous to duplicate these types wholesale like this.  If it's just a matter of making them `pub` we could do that.  The feature I suggested is probably not worth it because all of our consumers will wind up getting it because of workspace resolution.

But how much of this do you actually need?  I got a lot of the useful output from the existing pub interfaces in #7695.

dev-tools/omdb/src/bin/omdb/db.rs
davepacheco: On re-reading, I figured this out.

```suggestion
        // We avoid `Tabled` here because some columns could be very wide and its auto-sizing would cause all rows to be that wide, resulting in unreadable output.
```

dev-tools/omdb/src/bin/omdb/db.rs
```diff
@@ -2249,6 +2416,1221 @@ async fn cmd_db_network_list_vnics(
...
+            data,
+        })
+    }
+
+    if rows.is_empty() {
+        return;
+    }
+
+    // Using Tabled isn't recommended: the data column could be huge, and Tabled
+    // will print spaces in order to make each column the same size
```
davepacheco, line 2793: same as above

dev-tools/omdb/src/bin/omdb/db.rs
davepacheco: These days you can use `Paginator` for this to get all the sagas in a paginated way.  There should be code you can use in #7695.

dev-tools/omdb/src/bin/omdb/db.rs
davepacheco: Edit: if you're going to collect all the sagas, you can sort in Rust instead.

I would really try to avoid using ALLOW_FULL_TABLE_SCAN_SQL in all these places if we can.

dev-tools/omdb/src/bin/omdb/db.rs
```diff
@@ -2249,6 +2416,1221 @@ async fn cmd_db_network_list_vnics(
...
+                dsl::time_created,
+                &first_page::<dsl::time_created>(fetch_opts.fetch_limit),
+            )
+            .load_async(&conn)
+            .await
+        })
+        .await?;
+
+    check_limit(&sagas, fetch_opts.fetch_limit, || {
+        String::from("listing sagas")
```
davepacheco, line 2909: You can also strike this if you use pagination to get all the sagas.

dev-tools/omdb/src/bin/omdb/db.rs
```diff
@@ -2249,6 +2416,1221 @@ async fn cmd_db_network_list_vnics(
...
+) -> Result<(), anyhow::Error> {
+    let sagas = datastore
+        .pool_connection_for_tests()
+        .await?
+        .transaction_async(|conn| async move {
+            use db::schema::saga_node_event::dsl;
+
+            // Sorting by time_created requires a full table scan with the
+            // current index definitions.
+            conn.batch_execute_async(ALLOW_FULL_TABLE_SCAN_SQL).await?;
```
davepacheco, line 2930: Same as above.

dev-tools/omdb/src/bin/omdb/db.rs
```diff
@@ -2249,6 +2416,1221 @@ async fn cmd_db_network_list_vnics(
...
+
+            let stuck_sagas: Vec<SagaId> = db::paginated(
```
davepacheco, line 2932: As above: if you're going to get all the results, you want to use `Paginator` I think.  If not, you don't need `db::paginated` (but then you should probably use `check_limit` to warn the user if your results are truncated).

You could also do both of these in one query (something like `SELECT saga_id from saga s, saga_node_event sne WHERE s.id = sne.saga_id AND sne.event_type == 'undo_failed').  I think this should be possible with regular diesel but may be a pain and definitely not necessary here.

It seems like we use this pattern in a few places (fetch saga ids (or maybe even entire records?) for sagas with a node event of type `type`).  It might help to make a helper function.

dev-tools/omdb/src/bin/omdb/db.rs
```diff
@@ -2249,6 +2416,1221 @@ async fn cmd_db_network_list_vnics(
...
+
+/// List all stuck sagas
+async fn cmd_db_sagas_list_stuck(
+    datastore: &DataStore,
+    fetch_opts: &DbFetchOptions,
+) -> Result<(), anyhow::Error> {
+    let sagas = datastore
+        .pool_connection_for_tests()
+        .await?
+        .transaction_async(|conn| async move {
```
davepacheco, line 2925: I don't think these need to be in an explicit transaction.

davepacheco:
Also, I noticed the singular/plural in `omdb db` is kind of a mess:

```
$ ./target/debug/omdb db
Query the control plane database (CockroachDB)

Usage: omdb db [OPTIONS] <COMMAND>

Commands:
  replacements-to-do           Print any Crucible resources that are located on expunged physical disks
  rack                         Print information about the rack
  disks                        Print information about virtual disks
  dns                          Print information about internal and external DNS
  inventory                    Print information about collected hardware/software inventory
  physical-disks               Print information about physical disks
  region                       Print information about regions
  region-replacement           Query for information about region replacements, optionally manually triggering one
  region-snapshot-replacement  Query for information about region snapshot replacements, optionally manually triggering one
  sleds                        Print information about sleds
  instance                     Print information about customer instances
  instances                    Alias to `omdb instance list`
  network                      Print information about the network
  migrations                   Print information about migrations
  snapshots                    Print information about snapshots
  validate                     Validate the contents of the database
  volumes                      Print information about volumes
  vmm                          Print information about Propolis virtual machine manager (VMM) processes
  vmms                         Alias to `omdb db vmm list`
  help                         Print this message or the help of the given subcommand(s)
```

- `instance` and `vmm`: are subcommands, have plural aliases (`instances`, `vmms`) that alias to `list` under the subcommand
- `sleds`, `physical-disks`: no subcommands at all
- `migrations`, `snapshots`, `volumes`: plural subcommands

I was thinking this should be `saga` but now I realize either way we've got a lot of inconsistency either way.

PR #4393, opened 2023-10-30
update database schema update docs

davepacheco:
I found some of the schema update docs confusing so I took a swing at clarifying them.  I still had some questions (will comment inline below).

edit: I fixed the line wrapping here in a separate commit so that you can ignore the re-wrapping changes by just ignoring commit 47ddd43.  (The file is hard-wrapped, but it was wrapped inconsistently at anywhere from 80 to 90 columns, which was surprisingly disruptive to read (at least in my editor, because I saw both the hard wraps and the soft wraps, so every other line was like 5 columns long).)


schema/crdb/README.adoc
```diff
@@ -7,28 +7,31 @@ This directory describes the schema(s) used by CockroachDB.
 
 We use the following conventions:
 
-* `schema/crdb/VERSION/up.sql`: The necessary idempotent migrations to
-  transition from the previous version of CockroachDB to this version. These
```
davepacheco: I was particularly confused by "These migrations will always be placed within one transaction per file."  I wasn't sure what "migration" meant (one statement?  all the statements in one file?  all of the `up*.sql` files for one version?).  So I avoided this term and used "statements" and "all of the statements in a given file" instead.

I'm also not positive I interpreted this right.  I interpreted this to mean "the framework will take all of the statements in one file and execute them in one transaction."  It may have meant "you are responsible for putting exactly one transaction in the file" (i.e., it should have `BEGIN` and `END`, etc.)  I think this was once the case, but it's empirically not the case right now.

schema/crdb/README.adoc
```diff
@@ -7,28 +7,31 @@ This directory describes the schema(s) used by CockroachDB.
...
+  this version.  All of the statements in a given file will be executed
+  together in one transaction; however, usually only one statement should
+  appear in each file.  More on this below.
+** If there's only one statement required, we put it into `up.sql`.
+** If more than one change is needed, any number of files starting with `up`
+   and ending with `.sql` may be used. These files will be sorted in
+   lexicographic order before being executed.  Each will be executed in a
    separate transaction.
 ** CockroachDB documentation recommends the following: "Execute schema
-   changes... in a single explicit transaction consisting of the single schema
```
davepacheco: I'm not sure if the CockroachDB docs changed, but the text as-is here does not (quite) appear in the linked docs.  I only noticed because I was trying to learn more and looked for the source and had a hard time finding it.  There's an extra word "single" here.

schema/crdb/README.adoc
```diff
@@ -58,13 +61,20 @@ Process:
 
 * Choose a `NEW_VERSION` number. This should almost certainly be a major
   version bump over `OLD_VERSION`.
-* Add a file to `schema/crdb/NEW_VERSION/up.sql` with your changes to the
-  schema.
-** This file should validate the expected current version transactionally.
```
davepacheco, line 58: > This file should validate the expected current version transactionally.

I'm not sure what this means.  I didn't think anything like this should be necessary so I removed it, but I'd like a +1 that I'm not missing something important!

schema/crdb/README.adoc
```diff
@@ -68,20 +85,21 @@ Process:
 SQL Validation, via Automated Tests:
 
 * The `SCHEMA_VERSION` matches the version used in `dbinit.sql`
-* The combination of all `up.sql` files results in the same schema as `dbinit.sql`
+* The combination of all `up.sql` files results in the same schema as
+  `dbinit.sql`
 * All `up.sql` files can be applied twice without error
 
-==== Handling common schema changes
+==== General notes
```
davepacheco, line 92: This is kind of a vague heading but the one that was here didn't seem to describe the contents of this section.

schema/crdb/README.adoc
```diff
@@ -68,20 +85,21 @@ Process:
...
+tables and constraints should generally not be important.
 
-As convention, however, we recommend keeping the `db_metadata` file at the end of
-`dbinit.sql`, so that the database does not contain a version until it is fully
-populated.
+As convention, however, we recommend keeping the `db_metadata` file at the end
+of `dbinit.sql`, so that the database does not contain a version until it is
+fully populated.
 
 ==== Adding new source tables to an existing view
```
davepacheco, line 104: This section threw me off.  What's "the correct position" (i.e., position in what order? why does it matter?)  Is this still relevant after #4288?  If it is, maybe it'd help to move this section and the next one under a common heading like "Gotchas around specific kinds of schema changes" so that it's clearer that these are unusual cases and you can skip it altogether if you're obviously not in one of them.

schema/crdb/README.adoc
david-crespo: You are correct — it used to be the case ([example](https://github.com/oxidecomputer/omicron/blob/cec576fc6b3c56e782e655465340e050636d7cb2/schema/crdb/3.0.0/up.sql)) that schema changes had to contain an explicit `BEGIN` and `COMMIT`, but now the migration doer handles wrapping each `up*.sql` in its own transaction.

david-crespo (approved):
Changes look good. I share the question about the "Adding new source tables to an existing view" and I agree about grouping those last sections under a "Scenario-specific gotchas" heading.

schema/crdb/README.adoc
smklein: I agree with @david-crespo -- all the more reason to update these docs.

schema/crdb/README.adoc
```diff
@@ -53,11 +59,22 @@ Assumptions:
...
+* Create directory `schema/crdb/NEW_VERSION`.
+* If only one SQL statement is necessary to get from `OLD_VERSION` to
+  `NEW_VERSION`, put that statement into `schema/crdb/NEW_VERSION/up.sql`.  If
+  multiple statements are required, put each one into a separate file, naming
+  these `schema/crdb/NEW_VERSION/upN.sql` for as many `N` as you need.
+** Each file should contain _either_ one schema-modifying statement _or_ some
+   number of data-modifying statements.  You can combine multiple data-modifying
+   statements.  But you should not mix schema-modifying statements and
+   data-modifying statements in one file.  And you should not include multiple
+   schema-modifying statements in one file.
```
smklein, line 73: Great summary, this is spot-on. 

Technically there are [exceptions](https://www.cockroachlabs.com/docs/stable/online-schema-changes#schema-changes-within-transactions):

> Some schema change operations can be run within explicit, multiple statement transactions. CREATE TABLE and CREATE INDEX statements can be run within the same transaction with the same atomicity guarantees as other SQL statements. There are no performance or rollback issues when using these statements within a multiple statement transaction.

But I think this is an apt summary, unless someone performing a schema change can cite more-specific docs from CRDB for a multi-statement transaction.

schema/crdb/README.adoc
smklein: I think this is a good question for @gjcolombo -- this was added in 876e8ca7c6 with some of the instance work.

I **think** this is obsolete now that ordering no longer matters, as you say, after #4288.

schema/crdb/README.adoc
smklein: Maybe "Writing Schema Change Statements"? This appears to be advice for making particular statements possible, assuming that the reader already understands the high-level flow of how they're applied.

schema/crdb/README.adoc
smklein: > I'm also not positive I interpreted this right. I interpreted this to mean "the framework will take all of the statements in one file and execute them in one transaction."

This is correct -- the framework also tracks some version information within this transaction, to make it safe for multiple Nexus instances to perform these updates concurrently.

> It may have meant "you are responsible for putting exactly one transaction in the file" (i.e., it should have BEGIN and END, etc.) I think this was once the case, but it's empirically not the case right now.

Yeah, this **was** the case, but it is not the case any longer.

smklein (approved):
Thanks for cleaning this up!

schema/crdb/README.adoc
gjcolombo: I concur--I think the "Adding new source tables to an existing view" section is obsoleted by #4288 and can be removed.

davepacheco: Thanks!  Done in [51fcf64](https://github.com/oxidecomputer/omicron/pull/4393/commits/51fcf6476f978b851fa594ebd18623287627f6da).

davepacheco:
> Changes look good. I share the question about the "Adding new source tables to an existing view" and I agree about grouping those last sections under a "Scenario-specific gotchas" heading.

I like this title (whether you meant it seriously or not).  Done in 658f19505.

davepacheco:
Enabling auto-merge because I think I've addressed all the feedback.

PR #4404, opened 2023-11-01
Update `how-to-run.adoc` post-BGP

FelixMcFelix:
Small edits needed to bring this in line with how routes are configured now. `gateway_ip` is no longer explicitly named as such, so I've added an explicit comment to signpost how $GATEWAY_IP should be used w.r.t. the default route.

### DIFF
--- a/docs/how-to-run.adoc
+++ b/docs/how-to-run.adoc
@@ -274,6 +274,8 @@ You will also need to update route information if your `$GATEWAY_IP` differs fro
 # outside.  This block assumes that you're following option (2) above: putting
 # your Oxide system on an existing network that you control.
 [rack_network_config]
+# An internal-only IPv6 address block which contains AZ-wide services.
+# This does not need to be changed.
 rack_subnet = "fd00:1122:3344:01::/56"
 # A range of IP addresses used by Boundary Services on the network.  In a real
 # system, these would be addresses of the uplink ports on the Sidecar.  With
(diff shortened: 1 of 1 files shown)


docs/how-to-run.adoc
```diff
@@ -266,26 +266,44 @@ last = "192.168.1.29"
...
 # system, these would be addresses of the uplink ports on the Sidecar.  With
 # softnpu, only one address is used.
 infra_ip_first = "192.168.1.30"
 infra_ip_last = "192.168.1.30"
-# Name of the port.  This should always be "qsfp0" when using softnpu.
-uplink_port = "qsfp0"
-uplink_port_speed = "40G"
-uplink_port_fec="none"
+
+# Configurations for BGP routers to run on the scrimlets.
```
jordanhendricks, line 284: For this value and the ones below, it's not clear to me how I know what values to set here, particularly for a clustero n a home network, where I assume BGP is not a factor. Could we add some color on that?

jordanhendricks (commented):
Thanks for doing this! I had a couple questions about the changes.

docs/how-to-run.adoc
FelixMcFelix: Thanks! I've extended these descriptions after quickly following `common::api::internal::shared::{RackNetworkRackNetworkConfigServices, BgpConfig}` around sled-agent. I haven't had any time to play around with BGP as yet, maybe @rcgoodfellow can verify this makes sense?

rcgoodfellow: Yeah, this makes sense. I'll add an update to the [BGP Docs PR](https://github.com/oxidecomputer/docs/pull/216) that updates the [network preparations RSS docs](https://docs-git-bgp-guide-oxidecomputer.vercel.app/guides/system/network-preparations#_complete_configuration) in addition to the [BGP Guide](https://docs-git-bgp-guide-oxidecomputer.vercel.app/guides/operator/configuring-bgp). Then we can just provide a link to the user/operator facing docs here.

FelixMcFelix:
I've added links to Ry's newly published BGP resources, could you please take a look again @jordanhendricks ?

jordanhendricks (approved):
Thanks! The comments you added are helpful and what I was hoping for.

PR #4407, opened 2023-11-01
Create a new version of StartSledAgentRequest

andrewjstone:
In preparation for sled-addition, a new version of `StartSledAgentRequest` was created and the old version coverted to V0. Deterministic conversion from v0 to v1 is possible because we know that `ntp_servers` and `dns_servers` fields are not in use. We also know the values of the new fields. `is_lrtq_learner` must be false because we only support RSS addition on existing sleds.

We then implement a deserialize method that will return the new version regardless of what's on disk. This is similar to how we handled the bootstore upgrade for `EarlyNetworkConfig`. Also similar to that, we extended the format of `StartSledAgentRequest` to ease future deserializations. In those deserializations we can deserialize the "header" and use [serde_json::value::RawValue](https://docs.rs/serde_json/latest/serde_json/value/struct.RawValue.html) in order to defer deserialization of the `body` field until we know the `schema_version`.

### DIFF
--- a/common/src/ledger.rs
+++ b/common/src/ledger.rs
@@ -170,7 +170,7 @@ pub trait Ledgerable: DeserializeOwned + Serialize + Send + Sync {
     async fn read_from(log: &Logger, path: &Utf8Path) -> Result<Self, Error> {
         if path.exists() {
             debug!(log, "Reading ledger from {}", path);
-            serde_json::from_str(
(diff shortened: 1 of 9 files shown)


jgallagher (approved):

andrewjstone:
Tested successfully on madrid via mupdate based upgrade.

PR #4410, opened 2023-11-01
Various volume management fixes

jmpesp:
This commit bundles up a few fixes related to volume management:

- `find_deleted_volume_regions` now returns the `Option<RegionSnapshot>` that resulted from the `left_join` in the query for freed regions. This is then consulted to see if sending a DELETE for that region is safe: if the `Option` is `Some`, then the region snapshot has not been deleted yet, and sending a region DELETE will surely result in a `must delete snapshots first` error from the corresponding Crucible agent.

- Fix a few typos in nexus/src/app/sagas/common_storage.rs

- Nexus now waits for the Agent's `zfs destroy` of a snapshot to take place. Otherwise if illumos doesn't immediately remove the snapshot it may be returned by a subsequent `zfs list` later.

- Either `decrease_crucible_resource_count_and_soft_delete_volume` or `volume_hard_delete` should be called when unwinding a saga, calling both is not required.

- In the snapshot deletion saga, use `append_parallel` for the two volume delete sub sagas: in order to _not_ orphan Crucible resources, it's important that both volumes be soft deleted, and that failing to delete one volume's Crucible resources does not cause the other to _not_ be soft deleted.

- Also fix a very confusing typo when building the destination volume delete sub saga.

### DIFF
--- a/nexus/src/app/sagas/common_storage.rs
+++ b/nexus/src/app/sagas/common_storage.rs
@@ -338,7 +338,7 @@ pub(super) async fn delete_crucible_running_snapshot(
     .map_err(|e| {
         error!(
             log,
-            "delete_crucible_snapshot: region_delete_running_snapshot saw {:?}",
+            "delete_crucible_running_snapshot: region_delete_running_snapshot saw {:?}",
             e
         );
         match e {
@@ -377,7 +377,7 @@ pub(super) async fn delete_crucible_running_snapshot(
                 })
                 .await
                 .map_err(|e| {
-                    error!(log, "delete_crucible_snapshot: region_get_snapshots saw {:?}", e);
+                    error!(log, "delete_crucible_running_snapshot: region_get_snapshots saw {:?}", e);
                     match e {
                         crucible_agent_client::Error::ErrorResponse(rv) => {
                             match rv.status() {
@@ -494,7 +494,19 @@ pub(super) async fn delete_crucible_snapshot(
     region_id: Uuid,
     snapshot_id: Uuid,
 ) -> Result<(), Error> {
-    // delete snapshot - this endpoint is synchronous, it is not only a request
+    // Unlike other Crucible agent endpoints, this one is synchronous in that it
+    // is not only a request to the Crucible agent: `zfs destroy` is performed
+    // right away. However this is still a request to illumos that may not take
+    // effect right away. Wait until the snapshot no longer appears in the list
+    // of region snapshots, meaning it was not returned from `zfs list`.
+
+    info!(
+        log,
+        "deleting region {} snapshot {}",
+        region_id.to_string(),
+        snapshot_id.to_string(),
+    );
+
     retry_until_known_result(log, || async {
         client
             .region_delete_snapshot(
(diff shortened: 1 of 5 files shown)


nexus/src/app/sagas/common_storage.rs
```diff
@@ -524,7 +536,94 @@ pub(super) async fn delete_crucible_snapshot(
...
+
+            if response
+                .snapshots
+                .iter()
+                .any(|x| x.name == snapshot_id.to_string())
+            {
+                info!(
+                    log,
+                    "region {} snapshot {} still exists, waiting",
+                    region_id.to_string(),
```
papertigers, line 592: Calling `.to_string()` is unnecessary here. We could just go with `{region_id}` and `{snapshot_id}` since they both are `Uuid` and implement `Display`.  Or we could simply drop the `.to_string()`.

nexus/src/app/sagas/common_storage.rs
```diff
@@ -524,7 +536,94 @@ pub(super) async fn delete_crucible_snapshot(
...
+                    snapshot_id.to_string(),
+                );
+
+                Err(BackoffError::transient(WaitError::Transient(anyhow!(
+                    "region {} snapshot {} not deleted yet",
+                    region_id.to_string(),
+                    snapshot_id.to_string(),
```
papertigers, line 599: We can drop `.to_string()` or used named parameters here as well.

nexus/src/app/sagas/common_storage.rs
```diff
@@ -524,7 +536,94 @@ pub(super) async fn delete_crucible_snapshot(
...
+                ))))
+            } else {
+                info!(
+                    log,
+                    "region {} snapshot {} deleted",
+                    region_id.to_string(),
+                    snapshot_id.to_string(),
```
papertigers, line 606: We can drop `.to_string()` or used named parameters here as well.

papertigers (commented):
Just a handful of nits.

nexus/src/app/sagas/common_storage.rs
```diff
@@ -494,7 +494,19 @@ pub(super) async fn delete_crucible_snapshot(
...
+    // is not only a request to the Crucible agent: `zfs destroy` is performed
+    // right away. However this is still a request to illumos that may not take
+    // effect right away. Wait until the snapshot no longer appears in the list
+    // of region snapshots, meaning it was not returned from `zfs list`.
+
+    info!(
+        log,
+        "deleting region {} snapshot {}",
+        region_id.to_string(),
+        snapshot_id.to_string(),
```
papertigers, line 507: We can drop `.to_string()` or used named parameters here as well.

gjcolombo:
@jmpesp and I discussed these mechanisms offline. A summary of that discussion:

- It wasn't 100% clear why the new "region garbage collection" procedure at the end of the volume delete saga is necessary. It would be interesting to see how the snapshot stress test performs without it. Another part of this change ensures that when Nexus asks a Crucible agent to delete a snapshot, the relevant Nexus function doesn't return until the agent successfully observes that the host OS has actually deleted the snapshot; this hints at a class of bugs along these lines:
  - Nexus asks to delete a snapshot (e.g. during unwind of the snapshot create saga), gets an "OK" response from the agent, then tries to delete the corresponding region; this fails because there's still an extant snapshot; if unwind ignores this and proceeds anyway, the region will be leaked.
  - This problem goes away if (as in this change) Nexus waits for the snapshot to be deleted before proceeding with unwind.
- That said, the "wait for snapshot deletion" part of the change doesn't help in cases where a region has already been leaked. In that case the garbage-collecting behavior in this PR is still useful. However, it might make more sense to run it as a background task/RPW than as a rider on the volume delete saga (since if no one deletes a volume the collector will never run!).

gjcolombo (commented):
(See my other comment on the PR; leaving this here to clear the review bit.)

nexus/src/app/sagas/common_storage.rs
jmpesp: Traditionally I haven't done this, thanks for the prompting - done in 2fb0fb3a4

nexus/src/app/sagas/common_storage.rs
jmpesp: done in 2fb0fb3a4

nexus/src/app/sagas/common_storage.rs
jmpesp: done in 2fb0fb3a4

nexus/src/app/sagas/common_storage.rs
jmpesp: done in 2fb0fb3a4

jmpesp:
@gjcolombo thanks for the summary of our discussion - let me know if you're OK with these changes :)

gjcolombo (approved):

nexus/src/app/sagas/common_storage.rs
```diff
@@ -252,27 +252,23 @@ pub(super) async fn delete_crucible_region(
...
-                    ))))
-                }
+                _ => Err(BackoffError::transient(WaitError::Transient(
+                    anyhow!("region {region_id} unexpected state"),
+                ))),
             }
         },
         |e: WaitError, delay| {
-            info!(log, "{:?}, trying again in {:?}", e, delay,);
+            info!(log, "{:?}, trying again in {:?}", e, delay);
```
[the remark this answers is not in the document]
jmpesp: Agreed, I added slog key value mappings for the messages in 9e141be4a

PR #4415, opened 2023-11-02
Implement sled-agent API for adding a sled

andrewjstone:
This code largely reuses the existing code paths for starting sled-agent with a small change to configured the bootstore as a learner node if trust quorum is in use.

This code path does not attempt any retries, as it should be idempotent as long as the same sled and rack UUIDs are used.This is a command driven by an operator and we want to know right away if it succeeded or not. Multiple sleds can be added by Nexus with individual calls, so that individual errors can be returned without added complexity at the sled- agent level. These requests can also be sent to different sled-agents, since the sled-agent is just acting as a proxy to a remote bootstrap agent.


sled-agent/src/http_entrypoints.rs
```diff
@@ -706,3 +708,40 @@ async fn write_network_bootstore_config(
...
+    }
+
+    crate::sled_agent::add_sled_to_initialized_rack(
+        sa.logger().clone(),
+        request.sled_id,
+        request.start_request,
+    )
+    .await
+    .map_err(|e| {
+        HttpError::for_internal_error(format!(
```
jgallagher, line 742: Nit / question: `for_internal_error` will only put this error message in the log, and returns an opaque "500 internal error" to the client. Since this is a pretty low-level API anyway, do we want to push a more detailed error out to the client if something goes wrong? We can still use a 500-level code with a more detailed error; it's just slightly more verbose ([example from wicketd](https://github.com/oxidecomputer/omicron/blob/dbf01fddbfc9c9b836c173f70ed80340c9230d09/wicketd/src/http_entrypoints.rs#L449-L456)).

sled-agent/src/sled_agent.rs
```diff
@@ -942,3 +947,91 @@ impl SledAgent {
...
+    let bootstrap_addr =
+        target_ip.ok_or_else(|| AddSledError::NotFound(sled_id.clone()))?;
+    let bootstrap_addr =
+        SocketAddrV6::new(bootstrap_addr, BOOTSTRAP_AGENT_RACK_INIT_PORT, 0, 0);
+    let client = crate::bootstrap::client::Client::new(
+        bootstrap_addr,
+        log.new(o!("BootstrapAgentClient" => bootstrap_addr.to_string())),
+    );
+
+    client.start_sled_agent(&request).await.map_err(|err| {
```
jgallagher, line 1031: What happens if we send this request to a sled that's already in the cluster?

jgallagher (approved):
Looks great - this is a lot less than I expected it to be! Just a couple minor questions.

sled-agent/src/http_entrypoints.rs
andrewjstone: Ooh, that's much better! Added in 6552bb8f82c6289130879f1c41b6b7e2960deb51

andrewjstone: Oops, and I just fixed the error code in f151bb876959bf3bbd8415c1db754ffd267e9b80

sled-agent/src/sled_agent.rs
andrewjstone: Great question. 

We'll end up in the [`ServerStarted` clause](https://github.com/oxidecomputer/omicron/blob/main/sled-agent/src/bootstrap/server.rs#L751) which tries to be idempotent but only on a few fields. It will currently return an error if the sled-id or address changed. 

I think it would probably be better to make this truly idempotent and check the whole request. What do you think?

jgallagher: > I think it would probably be better to make this truly idempotent and check the whole request. What do you think?

It looks like as written, if we get a request with matching fields that are checked but non-matching other fields, we just ignore them entirely and return `Ok(_)`, which would presumably leave the client confused (it could reasonably assume all the params it passed are in use, but really the sled-agent was still running with some other set of params)? I agree making it check the whole request seems better.

andrewjstone:
I tested adding sled 15 to an existing madrid cluster of sleds (0,1,14) and it worked. The sled learned its key share and ended up starting sled agent which registered itself in CRDB via Nexus.

I also ensured that the command is idempotent and we properly error when trying to add a sled that is already part of the cluster.

PR #4416, opened 2023-11-02
[update-engine] ensure in-order processing of events

sunshowers:
Previously, there were situations where events could be received out of
order. Solve this in two different ways:

* For nested engines, send data over the payload channel rather than
  having another channel in between.
* For `send_progress` and `send_nested_report`, use a oneshot channel to
  synchronize state before returning. (Trying to use a solution similar
  to the one for nested engines ends up being very difficult--since
  the overall channel is user-controlled, what happens if it errors
  out? Retaining the payload channel makes those errors easier to
  handle.)

Make a couple of other related changes:

* Determine `now` for events as soon as the function is called, and include it in the payload.
* The particular arrangement of code required a bunch of `'static` annotations, so include them. These shouldn't be an issue since `StepSpec`s are typically uninhabited types and errors are almost always `'static`.

The test ensures this -- it was failing before this PR and succeeds
after it.

### DIFF
--- a/update-engine/src/buffer.rs
+++ b/update-engine/src/buffer.rs
@@ -1500,7 +1500,7 @@ mod tests {
     use tokio_stream::wrappers::ReceiverStream;
 
     use crate::{
-        events::{ProgressUnits, StepProgress},
+        events::{ProgressCounter, ProgressUnits, StepProgress},
         test_utils::TestSpec,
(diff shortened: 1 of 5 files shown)


andrewjstone (approved):
Looks great! Just some leftover names to cleanup.

PR #4417, opened 2023-11-02
[update-engine] record more info about parent steps that reached a terminal state

sunshowers:
Record information about any parent steps that reached a terminal state.
We need this as part of the line display output to ensure correct
ordering.

Since we may make a bunch of copies of `CompletionInfo`, `FailureInfo`
and `AbortInfo`, wrap them in an `Arc`.


andrewjstone (approved):
Looks great. The extra context will be good to have.

PR #4422, opened 2023-11-03
Implement ClickHouse replicated cluster option for simulated omicron

karencfv:
New `--replicated` flag for the `omicron-dev ch-run` command. This will allow us to spin up a ClickHouse replicated cluster containing 2 replicas and 3 keepers.

```console
$ cargo run --bin omicron-dev -- ch-run --replicated
    Finished dev [unoptimized + debuginfo] target(s) in 0.31s
     Running `target/debug/omicron-dev ch-run --replicated`
omicron-dev: running ClickHouse cluster with configuration files:
 replicas: /home/{user}/src/omicron/oximeter/db/src/configs/replica_config.xml
 keepers: /home/{user}/src/omicron/oximeter/db/src/configs/keeper_config.xml
omicron-dev: ClickHouse cluster is running with PIDs: 1113482, 1113681, 1113387, 1113451, 1113419
omicron-dev: ClickHouse HTTP servers listening on ports: 8123, 8124
omicron-dev: using /tmp/.tmpFH6v8h and /tmp/.tmpkUjDji for ClickHouse data storage
```

Related: https://github.com/oxidecomputer/omicron/issues/4148

### DIFF
--- a/dev-tools/omicron-dev/src/bin/omicron-dev.rs
+++ b/dev-tools/omicron-dev/src/bin/omicron-dev.rs
@@ -265,18 +265,30 @@ struct ChRunArgs {
     /// The HTTP port on which the server will listen
     #[clap(short, long, default_value = "8123", action)]
     port: u16,
+    /// Starts a ClickHouse replicated cluster of 2 replicas and 3 keeper nodes
+    #[clap(long, conflicts_with = "port", action)]
+    replicated: bool,
 }
 
 async fn cmd_clickhouse_run(args: &ChRunArgs) -> Result<(), anyhow::Error> {
+    if args.replicated {
+        start_replicated_cluster().await?;
+    } else {
+        start_single_node(args.port).await?;
+    }
+    Ok(())
+}
+
+async fn start_single_node(port: u16) -> Result<(), anyhow::Error> {
     // Start a stream listening for SIGINT
     let signals = Signals::new(&[SIGINT]).expect("failed to wait for SIGINT");
     let mut signal_stream = signals.fuse();
 
     // Start the database server process, possibly on a specific port
     let mut db_instance =
-        dev::clickhouse::ClickHouseInstance::new_single_node(args.port).await?;
+        dev::clickhouse::ClickHouseInstance::new_single_node(port).await?;
     println!(
-        "omicron-dev: running ClickHouse with full command:\n\"clickhouse {}\"",
+        "\nomicron-dev: running ClickHouse with full command:\n\"clickhouse {}\"",
         db_instance.cmdline().join(" ")
     );
     println!(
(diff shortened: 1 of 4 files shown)


dev-tools/omicron-dev/src/bin/omicron-dev.rs
```diff
@@ -320,6 +332,116 @@ async fn cmd_clickhouse_run(args: &ChRunArgs) -> Result<(), anyhow::Error> {
...
 
+async fn start_replicated_cluster() -> Result<(), anyhow::Error> {
+    // Start a stream listening for SIGINT
+    let signals = Signals::new(&[SIGINT]).expect("failed to wait for SIGINT");
+    let mut signal_stream = signals.fuse();
+
+    // Start the database server and keeper processes
+    let cur_dir = std::env::current_dir().unwrap();
+    let replica_config =
+        cur_dir.as_path().join("oximeter/db/src/configs/replica_config.xml");
```
bnaecker, line 343: It seems like we might want to use the project directory structure here, rather than CWD. E.g, something like [`CARGO_MANIFEST_DIR`](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates), which points to the workspace root. That would allow folks to run this from more than just the project root itself.

dev-tools/omicron-dev/src/bin/omicron-dev.rs
```diff
@@ -320,6 +332,116 @@ async fn cmd_clickhouse_run(args: &ChRunArgs) -> Result<(), anyhow::Error> {
...
+    let cur_dir = std::env::current_dir().unwrap();
+    let keeper_config =
```
bnaecker, line 345: Same note here about CWD

dev-tools/omicron-dev/src/bin/omicron-dev.rs
```diff
@@ -320,6 +332,116 @@ async fn cmd_clickhouse_run(args: &ChRunArgs) -> Result<(), anyhow::Error> {
...
+            .await?;
+    println!(
+        "omicron-dev: running ClickHouse cluster with configuration files:\n \
+        replicas: {}\n keepers: {}",
+        cluster.replica_config_path().display(),
+        cluster.keeper_config_path().display()
+    );
+    let pid_error_msg = "Failed to get process PID, it may not have started";
+    println!(
+        "omicron-dev: ClickHouse cluster is running with PIDs: {}, {}, {}, {}, {}",
```
bnaecker, line 359: Could we print out which kind of process each corresponds to? E.g., `keeper PIDS = [x, y, z]`.

oximeter/db/src/client.rs
```diff
@@ -1064,7 +1064,14 @@ mod tests {
 
     #[tokio::test]
     async fn test_replicated() {
-        let mut cluster = ClickHouseCluster::new()
+        let cur_dir = std::env::current_dir().unwrap();
+        let replica_config =
+            cur_dir.as_path().join("src/configs/replica_config.xml");
```
bnaecker, line 1069: Same note about CWD.

oximeter/db/src/client.rs
```diff
@@ -1064,7 +1064,14 @@ mod tests {
...
+        let cur_dir = std::env::current_dir().unwrap();
+        let keeper_config =
```
bnaecker, line 1071: Same note about CWD

test-utils/src/dev/clickhouse.rs
```diff
@@ -330,25 +320,33 @@ impl Drop for ClickHouseInstance {
...
 
         // Start all replica nodes
-        let cur_dir = std::env::current_dir().unwrap();
-        let replica_config =
-            cur_dir.as_path().join("src/configs/replica_config.xml");
-
         let replica_amount = 2;
         let mut replicas =
-            Self::new_replica_set(replica_amount, replica_config).await?;
+            Self::new_replica_set(replica_amount, replica_config.clone())
```
bnaecker, line 348: Kind of a nit, but I'd recommend passing the configuration as a reference rather than as an owned value, both in this method and `new_keeper_set`. It's cloned internally for each replica.

karencfv:
Thanks for the review @bnaecker! I think I've addressed all your comments :)

bnaecker (approved):
Looks good, thanks!

PR #4427, opened 2023-11-03
Add preliminary `SpUpdater` to Nexus

jgallagher:
This is not wired up in any way, and is only accessible by Nexus's integration tests (for now). The majority of this PR backs those tests: both the tests themselves and the changes to `sp-sim` that allow it to receive updates dwarf the actual `SpUpdater` implementation itself.


nexus/src/app/update/sp_updater.rs
```diff
@@ -0,0 +1,345 @@
...
+}
+
+pub struct SpUpdater {
+    log: Logger,
+    progress: watch::Sender<Option<UpdateProgress>>,
+    sp_type: SpType,
+    sp_slot: u32,
+    update_id: Uuid,
+    // TODO-clarity maybe a newtype for this? TBD where we how we get this from
+    // wherever it's stored, which might give us a stronger type already.
```
davepacheco, line 48:
```suggestion
    // TODO-clarity maybe a newtype for this? TBD how we get this from
    // wherever it's stored, which might give us a stronger type already.
```

nexus/src/app/update/sp_updater.rs
```diff
@@ -0,0 +1,345 @@
...
+        // here is if all clients failed with connection errors. Return the
+        // error from the last MGS we tried.
+        Err(GatewayClientError::CommunicationError(last_err.unwrap()))
+    }
+
+    async fn start_update_one_mgs(
+        &self,
+        client: &gateway_client::Client,
+    ) -> Result<(), GatewayClientError> {
+        // The SP has two firmware slots, but they're aren't invidually labled.
```
davepacheco, line 157:
```suggestion
        // The SP has two firmware slots, but they're aren't individually labeled.
```

nexus/src/app/update/sp_updater.rs
```diff
@@ -0,0 +1,345 @@
...
+            };
+
+            self.progress.send_replace(Some(UpdateProgress::Failed(
+                error_message.clone(),
+            )));
+            return Err(SpUpdateError::FailedToComplete(error_message));
+        }
+    }
+
+    async fn finalize_update_via_reset_one_mgs(
```
davepacheco, line 325: What do you think about waiting for it to come back up and verifying that it reports that it's running the new version?  I'm just thinking that if this ever _didn't_ happen, we'd want to know about it promptly as part of the update process.  Also, it'd be nice to know that after a successful SP update, the SP is running.  If we don't wait for that I could imagine the next step in an update process failing spuriously with "something is wrong -- the SP is not up".

sp-sim/src/lib.rs
```diff
@@ -50,6 +55,44 @@ pub trait SimulatedSp {
...
+
+    /// Get a watch channel on which this simulated SP will publish a
+    /// monotonically increasing count of how many responses it has successfully
+    /// sent.
+    ///
+    /// Returns `None` if called before the SP has set up its sockets to handle
+    /// requests.
+    fn responses_sent_count(&self) -> Option<watch::Receiver<usize>>;
+
+    /// Inject a UDP-level "throttler" on this simualted SP.
```
davepacheco, line 76: This sounds like basically a semaphore.  Is that right?

sp-sim/src/gimlet.rs
```diff
@@ -115,43 +148,44 @@ impl Gimlet {
...
 
-                    serial_console_addrs.insert(
+                    if let Some(addr) = component_config.serial_console {
+                        let listener =
+                            TcpListener::bind(addr).await.with_context(
+                                || format!("failed to bind to {}", addr),
+                            )?;
+                        info!(
+                            log, "bound fake serial console to TCP port";
+                            "addr" => %addr,
```
davepacheco, line 184: (can't put this comment in the right place -- thanks GitHub)

Lines 189-170 look incorrectly aligned.

nexus/tests/integration_tests/sp_updater.rs
```diff
@@ -0,0 +1,445 @@
+//! Nexus integration tests
```
davepacheco, line 1: Missing license header.  And this comment may not apply here?

nexus/tests/integration_tests/sp_updater.rs
```diff
@@ -0,0 +1,445 @@
...
+//!
```
davepacheco, line 2: Out of curiosity, could/should these be unit tests rather than integration tests?

nexus/tests/integration_tests/sp_updater.rs
```diff
@@ -0,0 +1,445 @@
...
+        &*last_update_image,
+        "simulated SP update contents (len {}) \
+         do not match test generated fake image (len {})",
+        last_update_image.len(),
+        hubris_archive.image.data.len()
+    );
+}
+
+#[tokio::test]
+async fn test_sp_updater_remembers_successful_mgs_instance() {
```
davepacheco, line 154: Nice test.  Is it worth one that makes sure that if an MGS instance fails partway through, then we do go back to the other one?  I could go either way -- we're going to need broader integration-level tests for stuff like that and it might not make sense to test that level of detail at this layer.

gateway-test-utils/src/setup.rs
```diff
@@ -104,7 +104,9 @@ pub async fn test_setup_with_config(
     let localhost_port_0 = SocketAddrV6::new(Ipv6Addr::LOCALHOST, 0, 0, 0);
     let mut addresses = vec![localhost_port_0];
     if let Some(addr) = listen_addr {
-        addresses.push(addr);
+        if addr != localhost_port_0 {
```
davepacheco, line 107: This doesn't seem like a problem but I'm not sure what's going on here.  It seems like the `listen_addr` arg is an extra address on which MGS is supposed to listen, and we always configure it to listen on `[::1]:0`.  Why would the caller pass a `Some` with the same address here?

davepacheco (approved):
Looks great!  Just some minor comments.

nexus/src/app/update/sp_updater.rs
jgallagher: That's a good idea; I'll add that. FWIW the `reset` operation already waits until the SP comes back and is able to answer messages, so we'd already fail if the SP didn't come up at all. But without the extra check we could still be running the old version (at least in theory, although I don't think that's possible in practice today given how the SP boots).

sp-sim/src/lib.rs
jgallagher: Yeah, I'll reword this. I think when I first wrote this I was using a tokio task to use this to throttle the SP to N responses/second, but it definitely is implementing a sempahore.

nexus/tests/integration_tests/sp_updater.rs
jgallagher: Not sure; they definitely need all the MGS-with-simulated-SPs stuff, which I thought tipped them over into integration tests. Do we have unit tests that stand up that much supporting machinery?

nexus/tests/integration_tests/sp_updater.rs
jgallagher: Seems reasonable! I'll take a stab at it; happy to add it if it's relatively easy (which I think it will be).

nexus/tests/integration_tests/sp_updater.rs
davepacheco: We do (see all the uses of `#[nexus_test]` in the unit tests -- that thing stands up all kinds of stuff).  It does make me wonder what the distinction means for us given that we're not really testing a library crate in the usual sense.  I definitely wouldn't worry too much about it.  I was just wondering.

sp-sim/src/lib.rs
davepacheco: Up to you.  I was just checking my understanding.

gateway-test-utils/src/setup.rs
jgallagher: I think this is just me not completely understanding all the test utils we have; looking at it again I'm being overly explicit in the tests and can clean that up (which will make this change unnecessary - I'll revert it).

nexus/src/app/update/sp_updater.rs
jgallagher: We chatted about this offline and opted _not_ to do this: `SpUpdater` doesn't currently know _what_ version it's updating the SP to, so it would either need to open the SP archive with `hubtools` and inspect, or accept the version from the caller (and trust that the caller told it a version that matches the SP archive). We don't yet have the mechanism by which `SpUpdater` gets its data, but it will almost certainly require using `hubtools` to inspect the archive _before_ we get here, and we don't really want to repeat that inspection. The next inventory collection taken after any SP update will report the version it's running, and `SpUpdater`'s caller should be in a position to trigger such a collection.

nexus/tests/integration_tests/sp_updater.rs
jgallagher: Added in 53fc4d8dc

Issue #4419, opened 2023-11-02
CmdError error messages should provide more information


papertigers:
While testing on a bench gimlet I ran into a version mismatch for what sled-agent was expecting vs what version helios was using.

```
{"msg":"Sending prefix to ddmd for advertisement","v":0,"name":"SledAgent","level":30,"time":"2023-11-02T21:04:18.748039834Z","hostname":"gimlet-sn06","pid":104370,"DdmAdminClient":"[::1]:8000","component":"BootstrapAgentStartup","file":"clients/ddm-admin-client/src/lib.rs:91","prefix":"Ipv6Prefix { addr: fdb0:e8ea:6a09:6c32::, len: 64 }"}
sled-agent: Failed to delete all XDE devices
[ Nov  2 21:04:18 Stopping because all processes in service exited. ]
[ Nov  2 21:04:18 Executing stop method (:kill). ]
[ Nov  2 21:04:18 Restarting too quickly, changing state to maintenance. ]
```

Looking at where that error comes from pointed me at:
```rust
    illumos_utils::opte::delete_all_xde_devices(&log)
        .map_err(StartError::DeleteXdeDevices)?;
```

If we manually truss sled-agent we can see the ioctl call fails.
```
/129:   ioctl(3, DP_POLL, 0xFFFFF5FFC93FF880)           = 1
/1:     open("/dev/xde", O_RDWR|O_CLOEXEC)              = 11
/1:     ioctl(11, _IOWRN('w', 1, 119), 0xFFFFF5FFFFDDBB68) Err#71 EPROTO
```

Looking at my omicron repo that I packaged from:
```
$ cat tools/opte_version
0.25.183
```

vs what `opteadm` reports:
```
# /opt/oxide/opte/bin/opteadm -V
opteadm 0.23.181
```

It seems like this info should be included with the Rust type provided by thiserror but we don't see any information about the failed ioctl call in sled-agent's logs.
```
    #[error("Failed to delete all XDE devices")]
    DeleteXdeDevices(#[source] illumos_utils::opte::Error),
```

luqmana:
Ok yea, we definitely have the mismatched error but I misremembered where, it's in opte and not omicron.

We'd either see the [`IoctlFailed`](https://github.com/oxidecomputer/opte/blob/faaf3320e673167722c2c7e6fb2a86bd84f04a8a/lib/opte-ioctl/src/lib.rs#L40-L42) variant
```rust
[...]

PR #4428, opened 2023-11-03
CmdError error messages should provide more information

papertigers:
Fixes: #4419

This commit is a little invasive at various call sites so I would appreciate some feedback.

Before this commit you would see:
```
sled-agent: Failed to delete all XDE devices
```

After this commit we now see:
```
sled-agent: Failed to delete all XDE devices

Caused by:
    0: Failure interacting with the OPTE ioctl(2) interface: command ListPorts failed: BadApiVersion { user: 25, kernel: 23 }
    1: command ListPorts failed: BadApiVersion { user: 25, kernel: 23 }
```

### DIFF
--- a/gateway/src/bin/mgs.rs
+++ b/gateway/src/bin/mgs.rs
@@ -4,6 +4,7 @@
 
 //! Executable program to run gateway, the management gateway service
 
+use anyhow::{anyhow, Context};
 use clap::Parser;
 use futures::StreamExt;
 use omicron_common::cmd::{fatal, CmdError};
@@ -70,26 +71,24 @@ async fn do_run() -> Result<(), CmdError> {
     let args = Args::parse();
 
     match args {
-        Args::Openapi => run_openapi().map_err(CmdError::Failure),
+        Args::Openapi => {
+            run_openapi().map_err(|e| CmdError::Failure(anyhow!("{e}")))
+        }
         Args::Run {
             config_file_path,
             id_and_address_from_smf,
             id,
             address,
         } => {
-            let config = Config::from_file(&config_file_path).map_err(|e| {
-                CmdError::Failure(format!(
-                    "failed to parse {}: {}",
-                    config_file_path.display(),
-                    e
-                ))
-            })?;
+            let config = Config::from_file(&config_file_path)
+                .with_context(|| {
+                    format!("failed to parse {}", config_file_path.display())
+                })
+                .map_err(CmdError::Failure)?;
 
-            let mut signals = Signals::new([signal::SIGUSR1]).map_err(|e| {
-                CmdError::Failure(format!(
-                    "failed to set up signal handler: {e}"
-                ))
-            })?;
+            let mut signals = Signals::new([signal::SIGUSR1])
+                .context("failed to set up signal handler")
+                .map_err(CmdError::Failure)?;
 
             let (id, addresses, rack_id) = if id_and_address_from_smf {
                 let config = read_smf_config()?;
@@ -102,8 +101,9 @@ async fn do_run() -> Result<(), CmdError> {
                 (id.unwrap(), vec![address.unwrap()], rack_id)
             };
             let args = MgsArguments { id, addresses, rack_id };
-            let mut server =
-                start_server(config, args).await.map_err(CmdError::Failure)?;
+            let mut server = start_server(config, args)
+                .await
+                .map_err(|e| CmdError::Failure(anyhow!(e)))?;
 
             loop {
                 tokio::select! {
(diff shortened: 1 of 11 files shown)


papertigers:
CI failed because I neglected the non illumos cfg sections.  Is there a good way to make sure one doesn't miss these in the future before CI runs?

gateway/src/bin/mgs.rs
```diff
@@ -70,26 +71,24 @@ async fn do_run() -> Result<(), CmdError> {
     let args = Args::parse();
 
     match args {
-        Args::Openapi => run_openapi().map_err(CmdError::Failure),
+        Args::Openapi => {
+            run_openapi().map_err(|e| CmdError::Failure(anyhow!("{e}")))
```
jgallagher, line 75: Should this be

```suggestion
            run_openapi().map_err(|e| CmdError::Failure(anyhow!(e)))
```

instead? That's the form used down on line 106, but I'm not sure what the difference is off the top of my head.

I'll leave the same suggestion other places I see it under the assumption that it's nicer to have `anyhow` wrapping an actual error than wrapping an error converted to a string; I _think_ as written we might lose any further error sources in the chain, because anyhow would just see a string here?

I also see a fair amount of `CmdError::Failure(e.into())`, which I think is equivalent to `CmdError::Failure(anyhow!(e))`. I don't mind both of those being present; I have a very, very slight preference for `anyhow!(e)` over `e.into()` but wouldn't bother changing the `e.into()`s.

oximeter/collector/src/bin/oximeter.rs
```diff
@@ -132,7 +133,9 @@ async fn main() {
 async fn do_run() -> Result<(), CmdError> {
     let args = Args::parse();
     match args {
-        Args::Openapi => run_openapi().map_err(CmdError::Failure),
+        Args::Openapi => {
+            run_openapi().map_err(|err| CmdError::Failure(anyhow!("{err}")))
```
jgallagher, line 137:
```suggestion
            run_openapi().map_err(|err| CmdError::Failure(anyhow!(err)))
```

gateway/src/bin/mgs.rs
```diff
@@ -102,35 +101,31 @@ async fn do_run() -> Result<(), CmdError> {
...
+                                .map_err(|err| CmdError::Failure(anyhow!("config refresh failed: {err}")))?;
                         }
                         // We only register `SIGUSR1` and never close the
                         // handle, so we never expect `None` or any other
                         // signal.
                         _ => unreachable!("invalid signal: {signal:?}"),
                     },
                     result = server.wait_for_finish() => {
-                        return result.map_err(CmdError::Failure)
+                        return result.map_err(|err| CmdError::Failure(anyhow!("{err}")))
```
jgallagher, line 128:
```suggestion
                        return result.map_err(|err| CmdError::Failure(anyhow!(err)))
```

sled-agent/src/bin/sled-agent.rs
```diff
@@ -51,16 +52,14 @@ async fn do_run() -> Result<(), CmdError> {
...
-            OpenapiFlavor::Sled => {
-                sled_server::run_openapi().map_err(CmdError::Failure)
-            }
-            OpenapiFlavor::Bootstrap => {
-                bootstrap_server::run_openapi().map_err(CmdError::Failure)
-            }
+            OpenapiFlavor::Sled => sled_server::run_openapi()
+                .map_err(|err| CmdError::Failure(anyhow!("{err}"))),
+            OpenapiFlavor::Bootstrap => bootstrap_server::run_openapi()
+                .map_err(|err| CmdError::Failure(anyhow!("{err}"))),
```
jgallagher, line 58:
```suggestion
                .map_err(|err| CmdError::Failure(anyhow!(err))),
```

sled-agent/src/bin/sled-agent.rs
```diff
@@ -103,12 +102,15 @@ async fn do_run() -> Result<(), CmdError> {
...
+                        return Err(CmdError::Failure(e.into()));
                     }
                 }
             }
 
-            server.wait_for_finish().await.map_err(CmdError::Failure)?;
+            server
+                .wait_for_finish()
+                .await
+                .map_err(|err| CmdError::Failure(anyhow!("{err}")))?;
```
jgallagher, line 113:
```suggestion
                .map_err(|err| CmdError::Failure(anyhow!(err)))?;
```

jgallagher (commented):
Thanks! These feel like kinda tedious changes, but better error messages are a huge :+1: for everybody.

jgallagher:
> CI failed because I neglected the non illumos cfg sections. Is there a good way to make sure one doesn't miss these in the future before CI runs?

In theory one can do something like `cargo check --target x86_64-unknown-linux-gnu`, but I suspect that won't work for annoying reasons (OpenSSL for one). If I'm working in `cfg`-controlled areas like this I end up just building on both platforms locally, which isn't super satisfying. Would be happy to hear from others if there's a better option.

citrus-it:
> Thanks! These feel like kinda tedious changes, but better error messages are a huge 👍 for everybody.

Speaking as someone who hit exactly this today and only managed to resolve it quickly because I was chatting to @papertigers - most definitely!

papertigers:
> > Thanks! These feel like kinda tedious changes, but better error messages are a huge 👍 for everybody.
> 
> Speaking as someone who hit exactly this today and only managed to resolve it quickly because I was chatting to @papertigers - most definitely!

And because we were speaking I realized that `./tools/install_prerequisites.sh` gets you a working env on macOS making the cfg illumos bits easier to spot locally.  Thanks!

papertigers:
Sorry for all the commit noise on this. I used a macro to fix up all of the `e.into()` -> `anyhow!(e)` so that we are consistent everywhere. I also applied all of your other suggested fixes @jgallagher.  Hopefully this is in better shape now -- I will try to do one more manual run on a bench gimlet before merging this.

jgallagher (approved):
Looks great - thanks!

papertigers:
With the latest patchset things still look good:
```
sled-agent: Failed to delete all XDE devices

Caused by:
    0: Failure interacting with the OPTE ioctl(2) interface: command ListPorts failed: BadApiVersion { user: 25, kernel: 23 }
    1: command ListPorts failed: BadApiVersion { user: 25, kernel: 23 }
```

jordanhendricks (approved):
Thanks for jumping on this so quickly! Will be so nice to have when someone hits this again.

PR #4430, opened 2023-11-04
[wicketd] make clear_update_state take multiple SPs

sunshowers:
This ensures atomicity and will be used for a new CLI interface.


wicketd/src/update_tracker.rs
```diff
@@ -695,16 +715,16 @@ pub enum StartUpdateError {
 
 #[derive(Debug, Clone, Error, Eq, PartialEq)]
 pub enum ClearUpdateStateError {
-    #[error("target is currently being updated")]
-    UpdateInProgress,
+    #[error("target are currently being updated: {}", sps_to_string(.0))]
```
jgallagher, line 718:
```suggestion
    #[error("targets are currently being updated: {}", sps_to_string(.0))]
```

jgallagher (approved):

wicketd/src/update_tracker.rs
sunshowers: done, thanks!

PR #4437, opened 2023-11-04
Bump system version to 1.0.4

iliana:
In preparation for the next release, bump the system version.

### DIFF
--- a/.github/buildomat/jobs/package.sh
+++ b/.github/buildomat/jobs/package.sh
@@ -37,7 +37,7 @@ rustc --version
 # trampoline global zone images.
 #
 COMMIT=$(git rev-parse HEAD)
-VERSION="1.0.3-0.ci+git${COMMIT:0:11}"
+VERSION="1.0.4-0.ci+git${COMMIT:0:11}"
 echo "$VERSION" >/work/version.txt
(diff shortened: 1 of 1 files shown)


askfongjojo (approved):

leftwo (approved):

iliana:
I realized a few hours ago that this change does have the effect of also bumping all our control plane component versions to "4.0.0". This is probably not ideal.

I might prefer to change this to keep the version stamped into zones and whatnot at the current version, and replace the system version in the tuf-repo.sh job.

askfongjojo:
How about we make it v1.0.4 first? It should not result in confusion as long as we are good about linking this to the marketing version in our internal documentation.

iliana:
As in, control plane components versioned v1.0.4, while still adjusting the system version to v4.0.0?

askfongjojo:
Sorry for being unclear. Just bumping the system version to 1.0.4, exactly like how we did it for 1.0.3.

iliana:
OK. I'll adjust this PR and we'll work out the finer details.

askfongjojo:
For the record - LGTM after all the back-and-forth.

PR #4450, opened 2023-11-07
[wicket] make CLI more usable in wicketd integration tests

sunshowers:
Make it so that wicket's CLI is usable in wicketd's integration tests. To
achieve this, make a few changes:

1. Factor out `exec` into `exec_with_args`.
2. Create the Tokio runtime outside of `exec_with_args`, and make
   `exec_with_args` async. This is so that wicketd's
   integration tests can run `exec_with_args` with a preexisting runtime.
3. Allow capturing any output not written to logs. For now this is only
   used by the `rack_update` commands.

Depends on #4062 (mostly because it is more convenient for me that way).


wicket/src/dispatch.rs
```diff
@@ -12,34 +12,82 @@ use clap::Parser;
...
+            let log = setup_log(&log_path()?, WithStderr::No)?;
+            Runner::new(log, wicketd_addr).run()
+        }
+    }
+}
+
+/// Enables capturing of wicket's output.
+pub enum OutputKind<'a> {
+    /// Captures output to the provided log, as well as a buffer.
+    Captured {
```
jgallagher, line 51: It doesn't look like this variant is ever constructed - is that coming in a subsequent PR?

jgallagher (approved):

wicket/src/dispatch.rs
sunshowers: Yeah, in #4451.

PR #4459, opened 2023-11-07
[wicketd] Don't check RoT CMPA/CFPA until we've decided to update it

### LINKED ISSUE #4420: Mupdate fails on locked Rev E gimlet
Sled 15 on madrid is a Rev E gimlet with a production locked RoT. Even when not attempting to update the RoT, the mupdate will fail immediately due to an image check. The check shouldn't prevent mupdate when the RoT isn't being updated.

image

jgallagher:
Fixes #4420, which is a bit of an edge case: if the TUF repo contains RoT images with the same version as the target sled, but does not contain an RoT image signed with a correct key for the target sled, we would previously fail to mupdate the sled even if the user chose _not_ to update the RoT (since its version number already matched), due to missing a correctly-signed image.

With this PR, we postpone the key check (implemented as fetching the CMPA/CFPA pages to verify against each archive) until _after_ we've decided whether or not to try updating the RoT at all.

An additional check added with this PR that is always true today with TUF repos built in CI is that for each class of Hubris archive where we accept multiple options (e.g., "gimlet SP" where we accept multiple board revisions, "gimlet RoT", "sidecar RoT", etc.), we require all the archives of that class to have the same version. We will still allow SP images to have different versions than RoT images, and even allow "gimlet RoT" images to have different versions of "sidecar RoT" images, but we do not allow something like "gimlet RoT version 1.0.1 signed with a dev key and gimlet RoT version 1.0.2 signed with a production key".

Todo before merging:

- [x] Test on `madrid` **Edit:** Test completed. I uploaded a TUF repo containing only 1.0.2 prod RoT images to `madrid` and attempted to update a sled with a 1.0.2 staging-dev RoT image on `main` with force update RoT _unchecked_, and it failed in the way described in #4420. After updating wicketd to this branch, the update succeeded: the RoT update was skipped as expected, and the rest of the update proceeded normally.

### DIFF
--- a/wicketd/src/artifacts/error.rs
+++ b/wicketd/src/artifacts/error.rs
@@ -5,6 +5,7 @@
 use camino::Utf8PathBuf;
 use display_error_chain::DisplayErrorChain;
 use dropshot::HttpError;
+use omicron_common::api::external::SemverVersion;
 use omicron_common::api::internal::nexus::KnownArtifactKind;
 use omicron_common::update::{ArtifactHashId, ArtifactId, ArtifactKind};
 use slog::error;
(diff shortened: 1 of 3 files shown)


andrewjstone (approved):
Looks great!

PR #4462, opened 2023-11-08
[ci] switch GHA to using the PR tip by default

### LINKED ISSUE #4461: Switch GitHub Actions to checking out the PR tip by default
On running a `pull_request` action, GitHub Actions, in a decision which appears to be somewhat surprising at first, does not check out the pull request tip. Instead, it checks out a **merge commit** of the PR tip with the target branch as of the time the PR was created. See actions/checkout#426 for more discussion.

Buildomat currently always uses the PR tip.

While there are reasonable arguments and trade-offs in both directions, the most important thing is to make sure that Buildomat and GitHub Actions use the same scheme. The most straightforward way to do that is to switch GHA to using the PR tip.

sunshowers:
Fixes #4461 (see that issue for more).

### DIFF
--- a/.github/workflows/check-opte-ver.yml
+++ b/.github/workflows/check-opte-ver.yml
@@ -10,6 +10,8 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3.6.0
+        with:
+          ref: ${{ github.event.pull_request.head.sha }} # see omicron#4461
       - name: Install jq
         run: sudo apt-get install -y jq
       - name: Install toml-cli
--- a/.github/workflows/check-workspace-deps.yml
+++ b/.github/workflows/check-workspace-deps.yml
@@ -11,5 +11,7 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3.6.0
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}  # see omicron#4461
       - name: Check Workspace Dependencies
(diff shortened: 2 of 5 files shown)


sunshowers:
This commit: cca4bd6036a8e0ab43a47926983170ccdcf7e6a2

GitHub: https://github.com/oxidecomputer/omicron/actions/runs/6792747869/job/18466514889?pr=4462#step:2:473 (uses cca4bd6036a8e0ab43a47926983170ccdcf7e6a2)

Buildomat: https://buildomat.eng.oxide.computer/wg/0/details/01HEPA9XY2BCFQ3KW6RMBKYDK3/7F73YbYapfBQ6b1auISxhdzl5691bZvqvfYIw5G2ASCWKX7J/01HEPAAQFMBWWCW6XWTSSFYYMF#S74 (uses cca4bd6036a8e0ab43a47926983170ccdcf7e6a2 as well)

.github/workflows/check-opte-ver.yml
```diff
@@ -10,6 +10,8 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3.6.0
+        with:
+          ref: ${{ github.event.pull_request.head.sha }} # see omicron#4461
```
davepacheco, line 14: Just a general comment: for something that's not a workaround, I like to put the explanation directly in a comment rather than reference an issue.  Up to you -- it's not important to change this here.

davepacheco (approved):
Thanks for doing this!

.github/workflows/check-opte-ver.yml
sunshowers: Ah I generally do agree, but in this case it was just that it needed to be referenced in a bunch of places. I'll leave it as is.

PR #4469, opened 2023-11-08
Fix reflector-bot merge conflict handling

augustuswm:
The update workflows for dendrite and maghemite are currently failing to merge main into their respective integration branches. The intended behavior is that the workflows start by merging main to pick up any new changes, and overwriting any changes to the `*_openapi_version` files that had been made on the integration branches with the versions from main (in the event of a conflict). This gets the integration branch into a clean spot which it can then apply the update scripts to.

The workflows as is though are not correctly passing the list of paths for which the the main branch should be preferred. This leaves the integration branch with a conflict causing the action to fail.

These changes should fix that and allow merges to complete cleanly.

### DIFF
--- a/.github/workflows/update-dendrite.yml
+++ b/.github/workflows/update-dendrite.yml
@@ -39,7 +39,7 @@ jobs:
           . ./tools/reflector/helpers.sh
 
           PATHS=("tools")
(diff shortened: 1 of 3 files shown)


sunshowers (approved):
TIL about local -n. Thanks!

PR #4475, opened 2023-11-09
Support a restricted SQL subset for querying timeseries

bnaecker:
- Add methods for querying oximeter timeseries using a limited subset of SQL. The raw string is heavily validated, and only simple SELECT queries are currently supported, with a limited subset of ClickHouse functions. Still, this allows running many different kinds of queries, including aggregations, window functions, and joins.
- Adds a few types for reading out tabular data from the query result, and parsing basic metadata for understanding resource usage of the queries.
- Add a barebones SQL shell for running oximeter SQL queries, using the new `Client::query` method.
- Include a bunch of tests for the restricted SQL subset as well as correctness of the actual returned queries against the DB.


bnaecker:
This one has been hanging around for a while, so I'm not sure if folks have any feedback about it. On one hand, there is basically zero risk right now. SQL support is only available through the internal `oxdb` development tool, not the public API. I've tried to ensure that all queries must be read-only, and there are a number of other guardrails in place (e.g., a function allow-list). On the other hand, if this isn't a good direction, we may not want to even start down it.

oximeter/db/src/sql/mod.rs
```diff
@@ -0,0 +1,1351 @@
...
+            || query.fetch.is_some()
+            || !query.locks.is_empty()
+        {
+            return unsupported!(
+                "CTEs, FETCH and LOCKS are not currently supported"
+            );
+        }
+        let SetExpr::Select(select) = &mut *query.body else {
+            return unsupported!("Only SELECT queries are currently supported");
+        };
```
david-crespo, line 825: was not aware of let/else. that is handy!

bnaecker: Yeah, might be my favorite syntax enhancement from the last few Rust releases!

david-crespo (approved):
Sorry, been busy! I think this looks great, and it will be very helpful to exercise it as a dev tool to inform the much harder decision about whether to incorporate it into the API somehow.

PR #4482, opened 2023-11-10
Remove a bunch of unused dependencies

smklein:
I used `cargo machete` to find quite a few of these (https://github.com/bnjbvr/cargo-machete) .

Notably this incurs a significant amount of false positives for the `client/` crates using progenitor. I *think* this is because progenitor provides a macro, and simply expects clients to have the necessary dependencies on the crates being used (including: `chrono`, `futures`, `regress`, `serde`, etc).


nexus/Cargo.toml
```diff
@@ -36,16 +34,12 @@ ipnetwork.workspace = true
...
 mime_guess.workspace = true
-newtype_derive.workspace = true
 # Not under "dev-dependencies"; these also need to be implemented for
 # integration tests.
 nexus-test-interface.workspace = true
 num-integer.workspace = true
 once_cell.workspace = true
 openssl.workspace = true
-openssl-sys.workspace = true
-openssl-probe.workspace = true
```
smklein, line 47: @jmpesp other than "tests pass" is there a way confirm this doesn't break anything? I see these are transitive deps of samael, but Nexus doesn't appear to depend on them directly.

smklein:
The `build-and-test (helios)` target on main has been taking about ~82 minutes on `main`, this is ~80 minutes. This is probably within "just noise" range, but it's still a worthwhile cleanup IMO.

nexus/Cargo.toml
jmpesp: The samael / SAML tests are pretty extensive - I'd be more worried about incompatible openssl versions (or libxmlsec1, like [here](https://github.com/oxidecomputer/garbage-compactor/pull/3)), something that I think we test by deploying to a helios machine during CI. I don't think there's anything an actual end-to-end SAML login would exercise over the unit tests though.

jmpesp (approved):
This looks good to me!

nexus/Cargo.toml
smklein: Gotcha, I'll trust those end-to-end tests then!

PR #4483, opened 2023-11-10
Disable incremental builds on CI

smklein:
With the intent to speed up CI times, this follows the advice of: https://matklad.github.io/2021/09/04/fast-rust-builds.html#CI-Workflow

> Disable incremental compilation. CI builds often are closer to from-scratch builds, as changes are typically much bigger than from a local edit-compile cycle. For from-scratch builds, incremental adds an extra dependency-tracking overhead. It also significantly increases the amount of IO and the size of ./target, which make caching less effective.


smklein:
The `build-and-test (helios)` target on main has been taking about ~82 minutes on `main`, this is ~76 minutes.

This is a small but apparently notable bump. Every bit counts.

david-crespo (approved):

iliana:
Nice!!

Issue #3814, opened 2023-08-02
Nexus should retry all transactions CRDB tells it to


jmpesp:
https://github.com/oxidecomputer/omicron/pull/3754 adds a `retry_transaction` function on `TransactionError` that returns `true` if CRDB is telling us to, based on the instructions at https://www.cockroachlabs.com/docs/v23.1/transaction-retry-error-reference#client-side-retry-handling.

Part of the review of that PR asked: why not to this for all transactions? If CRDB is telling us that we should be retrying a transaction, then we should. This issue tracks that work. Retrying in general seems like a safe thing to do but part of this issue should be convincing ourselves that we're not creating unbounded work for CRDB by retrying every transaction until it succeeds. [Note this may also be impossible](https://en.wikipedia.org/wiki/Halting_problem)

davepacheco:
I think this is related to #973, which talks about having a way to put a bunch of statements into a non-interactive transaction.  CockroachDB calls this [batched statements](https://www.cockroachlabs.com/docs/v23.1/transactions#batched-statements).  [This allows CockroachDB to retry such errors automatically.](https://www.cockroachlabs.com/docs/v23.1/transactions#transaction-retries)  This would have all the benefits mentioned in RFD 192 about avoiding interactive transactions, plus I'd expect it to minimize the window for transaction conflicts (since it avoids a round-trip to Nexus, plus all the time associated with doing the client-side retry).

davepacheco:
[...]

PR #4487, opened 2023-11-11
[nexus] Make most transactions automatically retry

smklein:
Integrates automatic transaction retry into Nexus for most transactions.

Additionally, this PR provides a "RetryHelper" object to help standardize how transaction retry is performed.
Currently, after a short randomized wait (up to an upper bound), we retry unconditionally, emitting each
attempt to Oximeter for further analysis.

- [x] Depends on https://github.com/oxidecomputer/async-bb8-diesel/pull/58
- [x] As noted in https://github.com/oxidecomputer/async-bb8-diesel/pull/58, this will require customizing CRDB session variables to work correctly. (Edit: this is done on each transaction)

Part of https://github.com/oxidecomputer/customer-support/issues/46 
Part of https://github.com/oxidecomputer/omicron/issues/3814


nexus/db-queries/src/db/datastore/silo_group.rs
```diff
@@ -147,34 +147,40 @@ impl DataStore {
...
+                            .values(silo_group_memberships)
+                            .execute_async(&conn)
+                            .await?;
 
-                Ok(())
-            })
+                        Ok(())
+                    }
+                },
+                || async { true },
```
smklein, line 182: This can hopefully be replaced by a shared function which:
 - Keeps a counter
 - Logs information to Oximeter, perhaps?
 
 Though maybe this should happen outside the scope of the transaction -- when this closure is called, the transaction is still in-progress!

nexus/db-queries/src/transaction_retry.rs
```diff
@@ -0,0 +1,141 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
smklein, line 1: This file is worth reviewing, independently of the other conversions.

The "big" structs to consider here are:

1. `Producer`, which emits information about retries to Oximeter
1. `RetryHelper`, which callers can use to automate the transaction retry decision-making logic.

nexus/db-queries/src/transaction_retry.rs
```diff
@@ -0,0 +1,141 @@
...
+impl RetryHelperInner {
+    fn new() -> Self {
+        Self { start: Utc::now(), attempts: 1 }
+    }
+
+    fn tick(&mut self) -> Self {
+        let start = self.start;
+        let attempts = self.attempts;
+
+        self.start = Utc::now();
```
smklein, line 67: Resetting `start` ensures that each `RetryData` metric shows up separately, rather than each one appearing to take longer and longer.

nexus/db-queries/src/transaction_retry.rs
```diff
@@ -0,0 +1,141 @@
...
+            .await;
+
+        return inner.attempts < MAX_RETRY_ATTEMPTS;
+    }
+
+    /// Converts this function to a retryable callback that can be used from
+    /// "transaction_async_with_retry".
+    pub(crate) fn as_callback(
+        self,
+    ) -> impl Fn() -> futures::future::BoxFuture<'static, bool> {
```
smklein, line 124: I really want: `impl Fn() -> (impl Future<Output = bool>)` but I don't know how to express that without boxing.

It's not really generic -- we only have one implementation here -- but the type is tricky to express.

nexus/db-queries/src/transaction_retry.rs
```diff
@@ -0,0 +1,141 @@
...
+
+/// Helper utility for tracking retry attempts and latency.
+/// Intended to be used from within "transaction_async_with_retry".
+pub(crate) struct RetryHelper {
+    producer: Producer,
+    name: &'static str,
+    inner: Mutex<RetryHelperInner>,
+}
+
+const MAX_RETRY_ATTEMPTS: usize = 10;
```
smklein, line 82: Totally arbitrary, but it's bounded.

nexus/db-queries/src/transaction_retry.rs
```diff
@@ -0,0 +1,141 @@
...
+    // This function:
+    // - Appends a metric identifying the duration of the transaction operation
+    // - Performs a randomly generated backoff (limited to less than 50 ms)
+    // - Returns "true" if the transaction should be restarted
+    async fn retry_callback(&self) -> bool {
+        let inner = self.inner.lock().unwrap().tick();
+
+        let _ = self.producer.append(
+            &Transaction { name: self.name.into() },
+            (Utc::now() - inner.start).to_std().unwrap_or(Duration::ZERO),
```
bnaecker, line 106: So if I'm following all this, the start time is taken pretty early, in some cases before we even take a connection from the pool. Is it possible to tighten those bounds at all, say by taking the timestamp just before the transaction closure is actually run? That may not be possible without more coordination in the actual connect pool implementation.

If not, I think it'd help to have a note that the latency is an upper bound, since it includes setup and network time.

nexus/db-queries/src/transaction_retry.rs
```diff
@@ -0,0 +1,141 @@
...
+        }
+    }
+}
+
+impl oximeter::Producer for Producer {
+    fn produce(
+        &mut self,
+    ) -> Result<Box<dyn Iterator<Item = Sample> + 'static>, MetricsError> {
+        let samples =
+            std::mem::replace(&mut *self.samples.lock().unwrap(), vec![]);
```
bnaecker, line 138: Nit, but `std::mem::take()` might be slightly more clear.

nexus/db-queries/src/transaction_retry.rs
```diff
@@ -0,0 +1,141 @@
...
+#[derive(Debug, Clone, Target)]
+struct Transaction {
+    name: String,
+}
+
+// Identifies that a retry has occurred, and track how long
+// the transaction took (either since starting, or since the last
+// retry failure was recorded).
+#[derive(Debug, Clone, Metric)]
+struct RetryData {
```
bnaecker, line 22: Is it worth including some data that would let us know which attempt this latency corresponds to?

bnaecker (commented):
I focused on the Oximeter integration portions in `transaction_retry.rs`. I've a few suggestions and comments, but overall this looks pretty solid. Thanks for wrangling this massive diff!

nexus/db-queries/src/db/datastore/silo_group.rs
smklein: This is implemented now in RetryHelper, marking resolved

nexus/db-queries/src/transaction_retry.rs
```diff
@@ -0,0 +1,141 @@
...
+//! Helper types for performing automatic transaction retries
+
+use chrono::Utc;
+use oximeter::{types::Sample, Metric, MetricsError, Target};
+use std::sync::{Arc, Mutex};
+use std::time::Duration;
+
+// Identifies "which" transaction is retrying
+#[derive(Debug, Clone, Target)]
+struct Transaction {
```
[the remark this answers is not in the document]
smklein: Sure thing, updated!

nexus/db-queries/src/transaction_retry.rs
```diff
@@ -0,0 +1,141 @@
...
+    name: String,
```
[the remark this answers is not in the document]
smklein: I'm not sure "which CRDB server we're connected to" is stable, so I don't think it should be part of the Target.

I figured this would be enough to *start*, simply because seeing a transaction name on here means "it's worth making that transaction leaner" -- the timestamps should also paint a pretty relevant picture of timing.

nexus/db-queries/src/transaction_retry.rs
```diff
@@ -0,0 +1,141 @@
...
+            &Transaction { name: self.name.into() },
+            (Utc::now() - inner.start).to_std().unwrap_or(Duration::ZERO),
+        );
+
+        // This backoff is not exponential, but I'm not sure we actually want
+        // that much backoff here. If we're repeatedly failing, it would
+        // probably be better to fail the operation, at which point Oximeter
+        // will keep track of the failing transaction and identify that it's a
+        // high-priority target for CTE conversion.
+        tokio::time::sleep(Duration::from_millis(rand::random::<u64>() % 50))
```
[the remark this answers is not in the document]
smklein: Great, I'll use that! Included via `gen_range`.

nexus/db-queries/src/transaction_retry.rs
smklein: Yeah, this is definitely an issue here. I figured it wouldn't be *that* bad, because this is basically a developer-facing issue, and we care about:

- Which transactions are open for a long time and hitting retry errors (should they be optimized?)
- Which transactions are open for a short time and hitting retry errors (should they be converted to CTEs?)

But beyond that, I don't really know how much precision matters. I think, by far, the most important aspect is just going to be a ranking of "which transactions are retrying the most"?

I do think it's possible to get a more picture-perfect accurate view, but I don't *think* the accuracy matters enough to do it -- it would require basically making a call to the `RetryHelper` object within the body of the function passed to `transaction_async_with_callback`, which would also require wrapping it in an `Arc` across all usages. I don't *think* the tradeoff is worth it? I'd bet the cost of checking out a connection is small relative to the cost of most transactions.

If you disagree, I can sketch out a more real plan, but given our usage here (dev only visibility for metrics, to help us prioritize) I'm kinda making the bet this shortcut is worth it.

nexus/db-queries/src/transaction_retry.rs
smklein: Added!

nexus/db-queries/src/transaction_retry.rs
bnaecker: I think you're probably right that it's not worth much complexity at this point, but let's add a note that this includes a bit of time that's not strictly part of the transaction in the database itself.

nexus/db-queries/src/db/datastore/rack.rs
```diff
@@ -89,15 +91,12 @@ enum RackInitError {
 }
 type TxnError = TransactionError<RackInitError>;
```
luqmana, line 92: Seems like only `rack_create_recovery_silo` remains using this and could be given the same treatment as `rack_populate_service_records`?

nexus/db-queries/src/db/datastore/address_lot.rs
```diff
@@ -45,51 +47,61 @@ impl DataStore {
...
-                let blocks: Vec<AddressLotBlock> = params
-                    .blocks
-                    .iter()
-                    .map(|b| {
-                        AddressLotBlock::new(
-                            db_lot.id(),
-                            b.first_address.into(),
-                            b.last_address.into(),
-                        )
+            .transaction_async_with_retry(
```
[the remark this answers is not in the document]
luqmana: Looks like most calls to `transaction_async_with_retry` use `RetryHelper`, the only exception I quickly found being `apply_schema_update`. Worth just encompassing the boiler plate for the common case (take the producer and a name as params and dropping the retry callback)?

luqmana (commented):
Took a look with a focus on some of the parts I've been in more (rack & network_interface) and looks good overall. A couple of comments but not blockers

nexus/db-queries/src/db/datastore/rack.rs
smklein: I've gone ahead and just plumbed through `RackInitError` here, though I've added a variant to identify transaction retry errors.

nexus/db-queries/src/db/datastore/rack.rs
```diff
@@ -623,42 +643,65 @@ impl DataStore {
...
-                    .dns_zones_list_all_on_connection(
-                        opctx,
-                        &conn,
-                        DnsGroup::External,
-                    )
-                    .await?;
-
-                Ok((ips, dns_zones))
-            })
+            .transaction_async_with_retry(
```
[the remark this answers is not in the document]
smklein: I agree with you that these could become a single transaction, but I'm not 100% sure they should be -- there's logic to construct DNS updates via `DnsVersionUpdateBuilder` in between these two transactions, and by combining them, we're forcing that logic to happen in a transactional context.

nexus/db-queries/src/db/datastore/address_lot.rs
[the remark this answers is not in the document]
smklein: Good call, I went ahead and wrapped this stuff into a little helper in:

https://github.com/oxidecomputer/omicron/pull/4487/commits/da330175cf9a28642edb12fe7c4761f9227631ab
and
https://github.com/oxidecomputer/omicron/pull/4487/commits/eb3097791351abcd5907d0888e168e25271f6105

Hopefully this just means a little less boilerplate for everyone using these retryable transactions.

nexus/db-queries/src/transaction_retry.rs
```diff
@@ -0,0 +1,341 @@
...
+}
+
+/// Helper utility for passing non-retryable errors out-of-band from
+/// transactions.
+///
+/// Transactions prefer to act on the `diesel::result::Error` type,
+/// but transaction users may want more meaningful error types.
+/// This utility helps callers safely propagate back Diesel errors while
+/// retaining auxiliary error info.
+pub struct OptionalError<E>(Arc<Mutex<Option<E>>>);
```
smklein, line 183: I added this as a helper to simplify a bunch of the spots where we observe a diesel error, consider bailing immediately if it's retryable, and otherwise set an out-of-band error while returning `RollbackTransaction`  from the Diesel closure.

nexus/db-queries/src/db/datastore/address_lot.rs
```diff
@@ -113,46 +125,60 @@ impl DataStore {
...
-            let now = Utc::now();
-            diesel::update(lot_dsl::address_lot)
-                .filter(lot_dsl::time_deleted.is_null())
-                .filter(lot_dsl::id.eq(id))
-                .set(lot_dsl::time_deleted.eq(now))
-                .execute_async(&conn)
-                .await?;
+        conn.transaction_async_with_retry(
+            |conn| {
+                let err = err.clone();
```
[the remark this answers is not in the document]
luqmana: Would it be reasonable to move `TransactionError` into `async-bb8-diesel` (maybe as `TransactionRetryError`) allow this use case? One benefit is making it less likely that `OptionalError` doesn't get checked. Or were you worried about like generics bloat?

nexus/db-queries/src/db/datastore/snapshot.rs
```diff
@@ -54,108 +53,109 @@ impl DataStore {
...
-            .map_err(|e: TxnError| match e {
-                TxnError::CustomError(e) => match e {
-                    CustomError::ResourceAlreadyExists => {
-                        Error::ObjectAlreadyExists {
-                            type_name: ResourceType::Snapshot,
-                            object_name: snapshot_name,
+                    if let Some(existing_snapshot_id) = existing_snapshot_id {
+                        if existing_snapshot_id != snapshot.id() {
+                            return Err(
+                                err.bail(CustomError::ResourceAlreadyExists)
```
luqmana, line 119: I think you can set `Error::ObjectAlreadyExists` here and just drop `CustomError` altogether

smklein: Good catch, updated!

nexus/db-queries/src/db/datastore/address_lot.rs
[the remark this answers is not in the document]
luqmana: Ok yea, that tracks. I think it's fine as-is now, we can revisit later if needed. (Besides, I'd feel a lil bad making you go and refactor all over again :P)

PR #4488, opened 2023-11-11
Update dropshot and openapiv3 deps

pfmooney:
The json specification generated by the updated openapiv3 is ordered slightly differently, but not in a way that matters to anything but "brute force" diff checks like expectorate.  This updates the dependency and applies the reordering to the recorded OpenAPI specs.

### DIFF
--- a/nexus/tests/output/cmd-nexus-noargs-stderr
+++ b/nexus/tests/output/cmd-nexus-noargs-stderr
@@ -1,6 +1,13 @@
-error: the following required arguments were not provided:
-  <CONFIG_FILE_PATH>
+See README.adoc for more information
 
-Usage: nexus <CONFIG_FILE_PATH>
+Usage: nexus [OPTIONS] [CONFIG_FILE_PATH]
 
-For more information, try '--help'.
+Arguments:
+  [CONFIG_FILE_PATH]  
+
+Options:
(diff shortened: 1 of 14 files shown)


nexus/tests/output/cmd-nexus-noargs-stderr
```diff
@@ -1,6 +1,13 @@
-error: the following required arguments were not provided:
-  <CONFIG_FILE_PATH>
+See README.adoc for more information
 
-Usage: nexus <CONFIG_FILE_PATH>
+Usage: nexus [OPTIONS] [CONFIG_FILE_PATH]
```
leftwo, line 3: I think what is supported is:
`nexus -i`
or
`nexus -o`
or
`nexus <config_file_path>`

I don't know how to make Clap enforce that though, and what you have here essentially does the same
thing anyway.

leftwo (approved):
I think what you are trying to do is have

nexus/tests/output/cmd-nexus-noargs-stderr
pfmooney: Frankly, I'd like to change nexus to operate like `nexus run <cfg>` or `nexus openapi` like all (?) of the other openapi servers in omicron.  Updating this was a stop-gap so I could get it to emit openapi docs for me without having to rustle up a valid configuration.

Issue #3927, opened 2023-08-21
Propolis zone installation took 81 seconds and caused instance start to time out


gjcolombo:
Seen in selfhost. Relevant log snippet:

```
21:19:35.260Z INFO SledAgent (InstanceManager): ensuring instance is registered
    file = sled-agent/src/instance_manager.rs:174
    instance_id = d49bdb29-700a-4714-9501-d9626c1e89b2
    propolis_id = 0f8796d9-c4df-4c72-89a2-2fd231fc395e
21:19:35.261Z INFO SledAgent (InstanceManager): registering new instance
    file = sled-agent/src/instance_manager.rs:204
    instance_id = d49bdb29-700a-4714-9501-d9626c1e89b2
21:19:36.043Z INFO SledAgent (InstanceManager): Ensuring new instance
    file = sled-agent/src/instance.rs:725
    instance_id = d49bdb29-700a-4714-9501-d9626c1e89b2
21:19:36.162Z INFO SledAgent (InstanceManager): Publishing instance state update to Nexus
    file = sled-agent/src/instance.rs:275
    instance_id = d49bdb29-700a-4714-9501-d9626c1e89b2
    state = InstanceRuntimeState { run_state: Starting, sled_id: ca79c08b-d31c-4769-95dc-df47aa190475, propolis_id: 0f8796d9-c4df-4c72-89a2-2fd231fc395e, dst_propolis_id: None, propolis_addr: Some([fd00:1122:3344:105::53]:12400), migration_id: None, propolis_gen: Generation(1), ncpus: InstanceCpuCount(32), memory: ByteCount(68719476736), hostname: "hubris", gen: Generation(2), time_updated: 2023-08-21T21:19:36.043175230Z }
21:19:38.477Z INFO SledAgent (InstanceManager): Configuring new Omicron zone: oxz_propolis-server_0f8796d9-c4df-4c72-89a2-2fd231fc395e
    file = illumos-utils/src/zone.rs:314
    instance_id = d49bdb29-700a-4714-9501-d9626c1e89b2
21:19:42.189Z INFO SledAgent (InstanceManager): Installing Omicron zone: oxz_propolis-server_0f8796d9-c4df-4c72-89a2-2fd231fc395e
    file = illumos-utils/src/zone.rs:353
    instance_id = d49bdb29-700a-4714-9501-d9626c1e89b2
[...]

PR #4489, opened 2023-11-11
Add some unit tests for sled-agent Instance creation

lifning:
Depends on #4325 for faking zone creation.

At time of writing, instance creation roughly looks like:

- nexus -> sled-agent: `instance_put_state`
  - sled-agent: `InstanceManager::ensure_state`
    - sled-agent: `Instance::propolis_ensure`
      - sled-agent -> nexus: `cpapi_instances_put` (if not migrating)
      - sled-agent: `Instance::setup_propolis_locked` (*blocking!*)
        - `RunningZone::install` and `Zones::boot`
        - `illumos_utils::svc::wait_for_service`
        - `self::wait_for_http_server` for propolis-server itself
      - sled-agent: `Instance::ensure_propolis_and_tasks`
        - sled-agent: spawn `Instance::monitor_state_task`
      - sled-agent -> nexus: `cpapi_instances_put` (if not migrating)
    - sled-agent: return ok result
- nexus: `handle_instance_put_result`

Or at least, it does in the happy path. #3927 saw propolis zone
creation take longer than the minute nexus's call to sled-agent's
`instance_put_state`. That might've looked something like:

- nexus -> sled-agent: `instance_put_state`
  - sled-agent: `InstanceManager::ensure_state`
    - sled-agent: `Instance::propolis_ensure`
      - sled-agent -> nexus: `cpapi_instances_put` (if not migrating)
      - sled-agent: `Instance::setup_propolis_locked` (*blocking!*)
        - `RunningZone::install` and `Zones::boot`
- nexus: i've been waiting a whole minute for this. connection timeout! `handle_instance_put_result`
    - [...]
    - sled-agent: return... oh, they hung up. :(

To avoid this timeout being implicit at the *Dropshot configuration*
layer (that is to say, we should still have *some* timeout),
we could consider a small refactor to make `instance_put_state` not a
blocking call -- especially since it's already sending nexus updates on
its progress via out-of-band `cpapi_instances_put` calls! That might look
something like:

- nexus -> sled-agent: `instance_put_state`
  - sled-agent: `InstanceManager::ensure_state`
    - sled-agent: spawn {
      - sled-agent: `Instance::propolis_ensure`
        - sled-agent -> nexus: `cpapi_instances_put` (if not migrating)
        - sled-agent: `Instance::setup_propolis_locked` (blocking!)
        - sled-agent: `Instance::ensure_propolis_and_tasks`
          - sled-agent: spawn `Instance::monitor_state_task`
        - sled-agent -> nexus: `cpapi_instances_put` (if not migrating)
        - sled-agent -> nexus: a cpapi call equivalent to the `handle_instance_put_result` nexus currently invokes after getting the response from the blocking call

(With a way for nexus to cancel an instance creation by ID, and a timeout
in sled-agent itself for terminating the attempt and reporting the failure
back to nexus, and a shorter threshold for logging the event of an instance
creation taking a long time.)

Before such a change, though, we should really have some more tests around
sled-agent's instance creation code at all! So here's a few.

### DIFF
--- a/illumos-utils/src/running_zone.rs
+++ b/illumos-utils/src/running_zone.rs
@@ -1169,29 +1169,55 @@ impl ZoneBuilderFactory {
 /// Created by [ZoneBuilderFactory].
 #[derive(Default)]
 pub struct ZoneBuilder<'a> {
+    /// Logger to which status messages are written during zone installation.
     log: Option<Logger>,
+    /// Allocates the NIC used for control plane communication.
     underlay_vnic_allocator: Option<&'a VnicAllocator<Etherstub>>,
+    /// Filesystem path at which the installed zone will reside.
     zone_root_path: Option<&'a Utf8Path>,
+    /// The directories that will be searched for the image tarball for the
+    /// provided zone type ([`Self::with_zone_type`]).
     zone_image_paths: Option<&'a [Utf8PathBuf]>,
+    /// The name of the type of zone being created (e.g. "propolis-server")
     zone_type: Option<&'a str>,
-    unique_name: Option<Uuid>, // actually optional
+    /// Unique ID of the instance of the zone being created. (optional)
+    // *actually* optional (in contrast to other fields that are `Option` for
+    // builder purposes - that is, skipping this field in the builder will
+    // still result in an `Ok(InstalledZone)` from `.install()`, rather than
+    // an `Err(InstallZoneError::IncompleteBuilder)`.
+    unique_name: Option<Uuid>,
+    /// ZFS datasets to be accessed from within the zone.
     datasets: Option<&'a [zone::Dataset]>,
+    /// Filesystems to mount within the zone.
     filesystems: Option<&'a [zone::Fs]>,
+    /// Additional network device names to add to the zone.
     data_links: Option<&'a [String]>,
+    /// Device nodes to pass through to the zone.
     devices: Option<&'a [zone::Device]>,
+    /// OPTE devices for the guest network interfaces.
     opte_ports: Option<Vec<(Port, PortTicket)>>,
-    bootstrap_vnic: Option<Link>, // actually optional
+    /// NIC to use for creating a bootstrap address on the switch zone.
+    // actually optional (as above)
+    bootstrap_vnic: Option<Link>,
+    /// Physical NICs possibly provisioned to the zone.
     links: Option<Vec<Link>>,
+    /// The maximum set of privileges any process in this zone can obtain.
     limit_priv: Option<Vec<String>>,
+    /// For unit tests only: if `Some`, then no actual zones will be installed
+    /// by this builder, and minimal facsimiles of them will be placed in
+    /// temporary directories according to the contents of the provided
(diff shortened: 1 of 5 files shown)


illumos-utils/src/running_zone.rs
```diff
@@ -1118,24 +1123,208 @@ impl InstalledZone {
...
+/// Builder-pattern construct for creating an [InstalledZone].
+/// Created by [ZoneBuilderFactory].
+#[derive(Default)]
+pub struct ZoneBuilder<'a> {
+    log: Option<Logger>,
+    underlay_vnic_allocator: Option<&'a VnicAllocator<Etherstub>>,
+    zone_root_path: Option<&'a Utf8Path>,
+    zone_image_paths: Option<&'a [Utf8PathBuf]>,
+    zone_type: Option<&'a str>,
+    unique_name: Option<Uuid>, // actually optional
```
jordanhendricks, line 1176: I think it's worth clarifying what "actually optional" means here? (same for the other field).

illumos-utils/src/running_zone.rs
```diff
@@ -1118,24 +1123,208 @@ impl InstalledZone {
...
+    /// Create a [ZoneBuilder] that inherits this factory's fakeness.
+    pub fn builder<'a>(&self) -> ZoneBuilder<'a> {
+        ZoneBuilder { fake_cfg: self.fake_cfg.clone(), ..Default::default() }
+    }
+}
+
+/// Builder-pattern construct for creating an [InstalledZone].
+/// Created by [ZoneBuilderFactory].
+#[derive(Default)]
+pub struct ZoneBuilder<'a> {
```
jordanhendricks, line 1170: I think it's probably worth putting some doc comments on what at least some of these fields mean.

sled-agent/src/instance.rs
```diff
@@ -1088,3 +1090,476 @@ impl Instance {
...
+        .await
+        .expect("starting DNS server");
+
+        let dns_dropshot_client = dns_service_client::Client::new(
+            &format!("http://{}", dns_dropshot.local_addr()),
+            logctx.log.new(o!("component" => "DnsDropshotClient")),
+        );
+        let mut dns_config = internal_dns::DnsConfigBuilder::new();
+        let IpAddr::V6(nexus_ip_addr) = nexus_server.local_addr().ip() else {
+            panic!()
```
jordanhendricks, line 1232: add a panic message here?

illumos-utils/src/running_zone.rs
lifning: agree, though i might put the doc comments on the `pub fn`'s of the builder rather than the private fields themselves

jordanhendricks: Even though they're private fields, I've found that having the doc comments on struct fields is still useful for IDEs like rust-analyzer and so on. (I know we've done that in other areas of omicron.) I expect it's more likely that someone will want documentation on the field than the setter for the builder object.

lifning:
closing this because it makes more sense to just pull it in as part of #4691

PR #4493, opened 2023-11-13
Unpublish VPC router related endpoints

david-crespo:
These endpoints aren't currently used for anything, so we don't want them showing up in either the docs, the generated clients, or the CLI.

Per https://github.com/oxidecomputer/docs/issues/232 (private repo)


karencfv:
Wait!!! Don't you need to be able to delete these in order to delete Subnets, so you can delete VPCs, so you can delete projects?

david-crespo:
> Wait!!! Don't you need to be able to delete these in order to delete Subnets, so you can delete VPCs, so you can delete projects?

Great point. I forgot to check. I was surprised to find everything is fine.

VPC routers are children of VPCs, not subnets. The child of subnet is a NIC, so delete subnet checks for child NICs but not child routers.

https://github.com/oxidecomputer/omicron/blob/6bbaa804fd2a8e04dcff07aa76782a2f3cbd1146/nexus/db-queries/src/db/datastore/vpc.rs#L761-L818

I think the relationship between subnets and routers and you might have had in mind is this: creating a subnet causes a route to be created in the system router, and deleting a subnet causes that route to be deleted. At least, that's supposed to be what happens — it turns out that is an undone TODO:

https://github.com/oxidecomputer/omicron/blob/6bbaa804fd2a8e04dcff07aa76782a2f3cbd1146/nexus/src/app/vpc_subnet.rs#L66-L68

https://github.com/oxidecomputer/omicron/blob/6bbaa804fd2a8e04dcff07aa76782a2f3cbd1146/nexus/src/app/vpc_subnet.rs#L242-L244

VPC delete checks for child subnets but not child routers:

https://github.com/oxidecomputer/omicron/blob/6bbaa804fd2a8e04dcff07aa76782a2f3cbd1146/nexus/db-queries/src/db/datastore/vpc.rs#L434-L504

Delete router itself doesn't have any conditions.

https://github.com/oxidecomputer/omicron/blob/6bbaa804fd2a8e04dcff07aa76782a2f3cbd1146/nexus/db-queries/src/db/datastore/vpc.rs#L939-L961

Finally, the delete VPC function in nexus/app also deletes all routers.

https://github.com/oxidecomputer/omicron/blob/6bbaa804fd2a8e04dcff07aa76782a2f3cbd1146/nexus/src/app/vpc.rs#L130-L165

I tested this on dogfood and was able to delete a VPC after deleting subnets, did not have to delete routers. The VPCs integration test also covers this when it deletes a VPC after deleting subnets.

https://github.com/oxidecomputer/omicron/blob/6bbaa804fd2a8e04dcff07aa76782a2f3cbd1146/nexus/tests/integration_tests/vpcs.rs#L189-L200

karencfv (approved):
> Great point. I forgot to check. I was surprised to find everything is fine.

phew :sweat_smile: carry on then

zephraph (approved):

rmustacc:
An additional note on the API surface is that every VPC has a System VPC Router, which always exists with the VPC. The bits that weren't there yet while show information about that router also allow for custom VPC Routers to be created which are then applied to individual subnets. So in the future, when there are custom VPC routers that can apply to subnets, that'll likely be something we consider blocking a VPC delete, but since none of those exist today and the System VPC router is not something that can be created or destroyed (from a theoretical perspective, obviously this doesn't exist hence the unpublish).

Issue #3812, opened 2023-08-01
[nexus] De-register an instance's producer with `oximeter` when the instance is destroyed


bnaecker:
See #3808 for the genesis of this.

`nexus` needs to de-register a producer with `oximeter` when a guest instance is destroyed. This will notify `oximeter` to stop collecting from the Propolis server managing the instance, since no data will ever be produced again. This should be part of instance-deletion saga most likely.

bnaecker:
Closed by #4495

Issue #3742, opened 2023-07-21
Tracking: Instance Lifecycle Overhaul


smklein:
- [ ] **Updating Instance State Information within Nexus**
  - [ ] "Sled Agent registering itself with Nexus" should also transfer information about "Here are the instances the sled agent knows about". It can start as an empty set. See https://github.com/oxidecomputer/omicron/issues/3633 for a lot more detail here.
  - [ ] The Sled Agent should refuse to handle instance requests until it successfully registers itself with Nexus. This would help avoid race conditions where: Nexus sends a request to a rebooting sled, at the same time as the sled registers with nexus and identifies that "all instances are dead now", inadvertently marking a very new instance as failed.
  - [ ] Nexus should look up all instances that **should have** been running on the sled and mark them failed.
  - [ ]  **Later** Nexus can use an RPW to look for instances that are marked as "failed + auto_boot_on_fault", and re-provision them in the background.
  - [ ] Idea: We could plausible update the "normal" instance provisioning workflow to rely on this RPW for provisioning, too. This would let "instance create" return much faster, and leave the work of "finding an appropriate sled and starting the instance" to a background task that could tolerate slower APIs to the backend.
- [ ] **Instances without Sleds**
  - [ ] We need to make it possible for Instances to **not** have a propolis ID / sled ID, in the case that they are stopped.
  - [ ] We also have the cleanup to do, ensuring that the virtual resources consumed by instances are no longer consumed in the case when an instance is stopped, but not deleted.
- [ ] **Handling Failed Instances**
  - [ ] Confirm that instances can be forcefully deleted after being marked failed
  - [ ] Plumb through the sled agent API @gjcolombo mentioned to "force-stop an instance" through the public-facing API for this failed case, to ensure that the instance is truly destroyed.

gjcolombo:
#2315 also tracks the "instances without sleds" work. It probably depends on #2824, since starting an instance with no resource reservation is a multi-step process.

> Nexus can use an RPW to look for instances that are marked as "failed + auto_boot_on_fault", and re-provision them in the background.

If we use the existing `Failed` state for this, we'll need to make sure that

- instances we've marked as Failed and are restarting have definitely been torn down (e.g. what happens if a Nexus-to-sled-agent call fails and causes Nexus to move an instance to the Failed state, but the problem was transient and the instance is actually alive?)
[...]

PR #4495, opened 2023-11-14
Stop collecting Propolis metrics on instance stop

bnaecker:
When Nexus responds to a sled-agent notification that the instance is stopped and its Propolis server is gone, hard-delete the assignment record and ask `oximeter` to stop collecting from it.

### DIFF
--- a/nexus/src/app/instance.rs
+++ b/nexus/src/app/instance.rs
@@ -1320,7 +1320,9 @@ impl super::Nexus {
         .await?;
 
         // If the supplied instance state indicates that the instance no longer
-        // has an active VMM, attempt to delete the virtual provisioning record
+        // has an active VMM, attempt to delete the virtual provisioning record,
+        // and the assignment of the Propolis metric producer to an oximeter
+        // collector.
         //
         // As with updating networking state, this must be done before
         // committing the new runtime state to the database: once the DB is
@@ -1338,6 +1340,7 @@ impl super::Nexus {
                     (&new_runtime_state.instance_state.gen).into(),
                 )
                 .await?;
+            self.unassign_producer(instance_id).await?;
         }
 
         // Write the new instance and VMM states back to CRDB. This needs to be
--- a/nexus/src/app/instance.rs
+++ b/nexus/src/app/instance.rs
@@ -1340,6 +1340,20 @@ impl super::Nexus {
                     (&new_runtime_state.instance_state.gen).into(),
                 )
                 .await?;
+
+            // TODO-correctness: The `notify_instance_updated` method can run
+            // concurrently with itself in some situations, such as where a
+            // sled-agent attempts to update Nexus about a stopped instance;
+            // that times out; and it makes another request to a different
+            // Nexus. The call to `unassign_producer` is racy in those
+            // situations, and we may end with instances with no metrics.
+            //
+            // This unfortunate case should be handled as part of
+            // instance-lifecycle improvements, notably using a reliable
+            // persistent workflow to correctly update the oximete assignment as
+            // an instance's state changes.
+            //
+            // Tracked in https://github.com/oxidecomputer/omicron/issues/3742.
             self.unassign_producer(instance_id).await?;
         }
 
--- a/nexus/db-model/src/oximeter_info.rs
+++ b/nexus/db-model/src/oximeter_info.rs
@@ -8,7 +8,7 @@ use chrono::{DateTime, Utc};
 use nexus_types::internal_api;
 use uuid::Uuid;
 
-/// Message used to notify Nexus that this oximeter instance is up and running.
+/// A record representing a registered `oximeter` collector.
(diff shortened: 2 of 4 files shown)


nexus/src/app/instance.rs
```diff
@@ -1338,6 +1340,7 @@ impl super::Nexus {
                     (&new_runtime_state.instance_state.gen).into(),
                 )
                 .await?;
+            self.unassign_producer(instance_id).await?;
```
gjcolombo, line 1343: There's a race in this path caused by the fact that this function can run concurrently with itself:

1. Sled agent calls Nexus N1 to send a state update that retires a VMM
2. Call times out
3. Sled agent retries and calls Nexus N2 to send the same state update
4. N2 completes the update, unassigns the producer, and clears the VMM ID from the instance record
5. User restarts the instance (possible because N2 ran to completion)
6. N1 finally gets around to processing the request and calls `unassign_producer` again
7. Now the producer is unassigned while the instance is active. Oops.

I think that ideally, we'd avoid this sort of problem by making stopped instance cleanup happen in a background task and not in a retriable notification from sled agent, but I think it will take a while to do that (and it's way out of the scope of this PR). 

In the absence of such a task, I think there are a few possible ways to deal with this:

1. Do something similar to what `virtual_provisioning_collection_delete_instance` does: have a variant of `producer_endpoint_delete` that only allows the endpoint to be deleted if the prospective Instance update hasn't already been performed by someone else
2. Try to make the *VMM* the nominal producer of metrics instead of the instance: VMM IDs are unique to instance generations, so in the race above, step 6 won't delete the producer associated with the generation of the instance that started in step 5
3. ¯\_(ツ)_/¯ for now, file an issue, and leave a comment here about this possible problem (and note that fixing up this part of the instance lifecycle is probably the best way to solve it)

To be terribly honest, I think I'm actually leaning toward option 3 here, because I think 1 and 2 are kind of bad:

1. Hacking an instance generation check into `virtual_provisioning_collection_delete_instance` was actually not that hard because the relevant database statement already had an appropriate CTE in place that already referred to IDs in the Instance table. That's not the case here: there's no existing subquery structure, and (more importantly) not all producer IDs are instance IDs, so you'd need a special variant of `producer_endpoint_delete` just for this.
2. I assume the producer ID is used not just for uniqueness but for *correlation*--i.e., it's important that all incarnations of a single Instance use the same instance ID as their producer ID, so that we can associate metrics with an instance across multiple start/stop cycles. Switching to the VMM ID prevents that unless there's some other way to set a "correlation ID" across samples from different producers (I haven't looked).

Anyway: if (1) or (2) seem more reasonable to you than they do to me, it'd be great to prevent this problem entirely. And I'm assuming it's not a huge problem to unregister a producer that's actually alive (nothing will get collected, sure, but nothing will explode, either). Under those circumstances I think I'd rather have "we miss collections in the comparatively unlikely event that sled agent retries a notifications" than "Oximeter memory usage grows boundlessly" and would accordingly be OK with just leaving a comment here and calling it a day. But wanted to put all the options out there for you to think about.

bnaecker: Thanks for pointing this out, I agree it's a race I hadn't considered. There is a lot of uncertainty here, but I generally agree that (3) might be the most expedient approach, even though it means we lose instance metrics in the sequence you laid out.

I've thought a lot about the RPW-like approach, and I think it makes the most sense long-term. I have a WIP that adds a new `kind` column to the `metric_producer` table which should help make that a reality. For Oxide-specific services that are managed and deployed by `nexus`, the handling of the oximeter assignment could be part of the normal (but not-yet-existing) update process. That is, as `nexus` decides to deploy or remove some service, it would assign or unassign that service's `oximeter` instance as part of that process. An instance-cleanup RPW could also fit in this, and would know to look for records in the `metric_producer` table with the instance ID and possibly using the `kind` to validate its operation.

I don't see the CTE approach as tenable, for the reasons you point out. The IDs aren't always instance IDs, meaning one would potentially need to search many tables. The CTE would be extremely complicated, if it's even possible. The instance-only code path is also not great, as you point out.

Is there an existing issue for the instance cleanup RPW? I can make this part a sub-task of that, or create a new issue if one doesn't exist.

gjcolombo: Sounds good--let's leave an explanatory note and an issue link and then get the improvement in. Thanks for thinking it over!

> Is there an existing issue for the instance cleanup RPW? I can make this part a sub-task of that, or create a new issue if one doesn't exist.

Could you tack it onto the list in #3742?

gjcolombo (approved):

bnaecker:
Thanks @gjcolombo, I've added an item about ensuring metric registrations follow the instance's state in #3742, and have added a note referring to that in b866993c. Appreciate the thoughtful review, as always!

bnaecker:
This closes #3812

PR #4496, opened 2023-11-14
Nexus inventory: Add collection of RoT CMPA and CFPA pages

jgallagher:
The RoT can report four different 512-byte pages (CMPA, and CFPA active/inactive/scratch). Given multiple RoT artifacts that are viable (match the right board, etc.) but are signed with different keys, these pages are required to identify which archive was signed with a key that the RoT will accept. This PR adds collection of these pages to the inventory system added in #4291.

The implementation here is fairly bulky but very mechanical, and is implemented almost identically to the way we collect cabooses: there's an `rot_page_which` to identify which of the four kinds of page it is, and a table for storing the relatively small number of raw page data values. Most of the changes in this PR resulted from "find where we're doing something for cabooses, then do the analogous thing for RoT pages". There are a couple minor quibbles in the unit tests that I'll point out by leaving comments below.

The RoT pages now show up when viewing a collection through omdb (note that the quite long base64 string is truncated; there's a command line flag to override the truncation and show the full string):

```console
$ omdb db inventory collections show e2f84867-010d-4ac3-bbf3-bc1e865da16b > x.txt
note: using database URL postgresql://root@[::1]:43301/omicron?sslmode=disable
note: database schema version matches expected (11.0.0)
collection: e2f84867-010d-4ac3-bbf3-bc1e865da16b
collector:  e6bff1ff-24fb-49dc-a54e-c6a350cd4d6c (likely a Nexus instance)
started:    2023-11-14T18:51:54.900Z
done:       2023-11-14T18:51:54.942Z
errors:     0

Sled SimGimlet00
    part number: FAKE_SIM_GIMLET
    power:    A2
    revision: 0
    MGS slot: Sled 0 (cubby 0)
    found at: 2023-11-14 18:51:54.924602 UTC from http://[::1]:42341
    cabooses:
        SLOT     BOARD        NAME      VERSION GIT_COMMIT 
        SpSlot0  SimGimletSp  SimGimlet 0.0.1   ffffffff   
        SpSlot1  SimGimletSp  SimGimlet 0.0.1   ffffffff   
        RotSlotA SimGimletRot SimGimlet 0.0.1   eeeeeeee   
        RotSlotB SimGimletRot SimGimlet 0.0.1   eeeeeeee   
    RoT pages:
        SLOT         DATA_BASE64                         
        Cmpa         Z2ltbGV0LWNtcGEAAAAAAAAAAAAAAAAA... 
        CfpaActive   Z2ltbGV0LWNmcGEtYWN0aXZlAAAAAAAA... 
        CfpaInactive Z2ltbGV0LWNmcGEtaW5hY3RpdmUAAAAA... 
        CfpaScratch  Z2ltbGV0LWNmcGEtc2NyYXRjaAAAAAAA... 
    RoT: active slot: slot A
    RoT: persistent boot preference: slot A
    RoT: pending persistent boot preference: -
    RoT: transient boot preference: -
    RoT: slot A SHA3-256: -
    RoT: slot B SHA3-256: -

Sled SimGimlet01
    part number: FAKE_SIM_GIMLET
    power:    A2
    revision: 0
    MGS slot: Sled 1 (cubby 1)
    found at: 2023-11-14 18:51:54.935038 UTC from http://[::1]:42341
    cabooses:
        SLOT     BOARD        NAME      VERSION GIT_COMMIT 
        SpSlot0  SimGimletSp  SimGimlet 0.0.1   ffffffff   
        SpSlot1  SimGimletSp  SimGimlet 0.0.1   ffffffff   
        RotSlotA SimGimletRot SimGimlet 0.0.1   eeeeeeee   
        RotSlotB SimGimletRot SimGimlet 0.0.1   eeeeeeee   
    RoT pages:
        SLOT         DATA_BASE64                         
        Cmpa         Z2ltbGV0LWNtcGEAAAAAAAAAAAAAAAAA... 
        CfpaActive   Z2ltbGV0LWNmcGEtYWN0aXZlAAAAAAAA... 
        CfpaInactive Z2ltbGV0LWNmcGEtaW5hY3RpdmUAAAAA... 
        CfpaScratch  Z2ltbGV0LWNmcGEtc2NyYXRjaAAAAAAA... 
    RoT: active slot: slot A
    RoT: persistent boot preference: slot A
    RoT: pending persistent boot preference: -
    RoT: transient boot preference: -
    RoT: slot A SHA3-256: -
    RoT: slot B SHA3-256: -

Switch SimSidecar0
    part number: FAKE_SIM_SIDECAR
    power:    A2
    revision: 0
    MGS slot: Switch 0
    found at: 2023-11-14 18:51:54.904 UTC from http://[::1]:42341
    cabooses:
        SLOT     BOARD         NAME       VERSION GIT_COMMIT 
        SpSlot0  SimSidecarSp  SimSidecar 0.0.1   ffffffff   
        SpSlot1  SimSidecarSp  SimSidecar 0.0.1   ffffffff   
        RotSlotA SimSidecarRot SimSidecar 0.0.1   eeeeeeee   
        RotSlotB SimSidecarRot SimSidecar 0.0.1   eeeeeeee   
    RoT pages:
        SLOT         DATA_BASE64                         
        Cmpa         c2lkZWNhci1jbXBhAAAAAAAAAAAAAAAA... 
        CfpaActive   c2lkZWNhci1jZnBhLWFjdGl2ZQAAAAAA... 
        CfpaInactive c2lkZWNhci1jZnBhLWluYWN0aXZlAAAA... 
        CfpaScratch  c2lkZWNhci1jZnBhLXNjcmF0Y2gAAAAA... 
    RoT: active slot: slot A
    RoT: persistent boot preference: slot A
    RoT: pending persistent boot preference: -
    RoT: transient boot preference: -
    RoT: slot A SHA3-256: -
    RoT: slot B SHA3-256: -

Switch SimSidecar1
    part number: FAKE_SIM_SIDECAR
    power:    A2
    revision: 0
    MGS slot: Switch 1
    found at: 2023-11-14 18:51:54.915680 UTC from http://[::1]:42341
    cabooses:
        SLOT     BOARD         NAME       VERSION GIT_COMMIT 
        SpSlot0  SimSidecarSp  SimSidecar 0.0.1   ffffffff   
        SpSlot1  SimSidecarSp  SimSidecar 0.0.1   ffffffff   
        RotSlotA SimSidecarRot SimSidecar 0.0.1   eeeeeeee   
        RotSlotB SimSidecarRot SimSidecar 0.0.1   eeeeeeee   
    RoT pages:
        SLOT         DATA_BASE64                         
        Cmpa         c2lkZWNhci1jbXBhAAAAAAAAAAAAAAAA... 
        CfpaActive   c2lkZWNhci1jZnBhLWFjdGl2ZQAAAAAA... 
        CfpaInactive c2lkZWNhci1jZnBhLWluYWN0aXZlAAAA... 
        CfpaScratch  c2lkZWNhci1jZnBhLXNjcmF0Y2gAAAAA... 
    RoT: active slot: slot A
    RoT: persistent boot preference: slot A
    RoT: pending persistent boot preference: -
    RoT: transient boot preference: -
    RoT: slot A SHA3-256: -
    RoT: slot B SHA3-256: -
```

There's also a new `omdb` subcommand to report the RoT pages (which does not truncate, but if we think it should that'd be easy to change):

```console
$ omdb db inventory rot-pages
note: using database URL postgresql://root@[::1]:43301/omicron?sslmode=disable
note: database schema version matches expected (11.0.0)
ID                                   DATA_BASE64                                                                                                                                                                                                                                                                                                                                                                                                                                                                            
099ba572-a978-4592-ae7a-452629377904 c2lkZWNhci1jZnBhLWluYWN0aXZlAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=
0e9dc5b0-b190-43da-acb6-84450fdfdb94 c2lkZWNhci1jbXBhAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=
80923bac-fbcc-46e0-b861-9dba906c14f7 Z2ltbGV0LWNmcGEtaW5hY3RpdmUAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=
98cc4225-a791-4092-99c6-81e27e8d8ffa c2lkZWNhci1jZnBhLWFjdGl2ZQAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=
a32eaf95-a20e-4570-8860-e0fb584a2ff1 c2lkZWNhci1jZnBhLXNjcmF0Y2gAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=
c941810a-1c6a-4dda-9c71-41a0caf62ace Z2ltbGV0LWNtcGEAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=
e96042d0-ae8a-435c-9118-1b71e8a9a651 Z2ltbGV0LWNmcGEtYWN0aXZlAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=
fdc27064-4338-4cbe-bfe5-622b11a9afbc Z2ltbGV0LWNmcGEtc2NyYXRjaAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=

### CODE REVIEW

File: nexus/db-queries/src/db/datastore/inventory.rs [hunk 1]
```diff
@@ -852,12 +1013,60 @@ impl DataStoreInventoryTest for DataStore {
...
                     .select(dsl::id)
                     .order_by(dsl::time_started)
                     .load_async(&conn)
                     .await
-                    .context("failed to list collections")
+                    .context("failed to list collections")?;
+
+                // Ensure we don't have any stray collection IDs in other
+                // inventory tables that reference collection IDs that are not
+                // present in `inv_collection`.
```
jgallagher, line 1025: This is the first of the two test issues I mentioned in the PR description. I had forgotten to remove rows in `inv_root_of_trust_page` when deleting a collection, but the existing tests still passed. I added this bit to check that we don't have any leftover collection IDs in tables other than `inv_collection`, but maybe this should be a separate method that the tests call explicitly? Open to suggestions.

nexus/inventory/src/builder.rs
```diff
@@ -769,6 +990,28 @@ mod test {
...
+        assert!(collection.rot_pages.contains(&rot_page.page));
+
+        // TODO-correctness Is this test correct? We reported the same RoT page
+        // with different data (rot_page1, then rot_page2). The second
+        // `found_rot_page` returned an error, but we overwrote the original
+        // data and did not record the error in `collection.errors`. Should we
+        // either have kept the original data or returned Ok while returning an
+        // error? It seems a little strange we returned Err but accepted the new
+        // data.
+        assert_eq!(rot_page.page.data_base64, rot_page2.data_base64);
```
jgallagher, line 1004: This is the second test issue I mentioned in the PR description. The caboose behaves this way so I made it match, but I'm not sure it's right - should we at least be logging a collection error if we hit this?

dev-tools/omdb/src/bin/omdb/db.rs
```diff
@@ -235,6 +238,8 @@ enum InventoryCommands {
     BaseboardIds,
     /// list all cabooses ever found
     Cabooses,
+    /// list all root of trust pages every found
```
davepacheco, line 241:
```suggestion
    /// list all root of trust pages ever found
```

dev-tools/omdb/src/bin/omdb/db.rs
```diff
@@ -1763,12 +1771,29 @@ async fn cmd_db_inventory(
         InventoryCommands::Cabooses => {
             cmd_db_inventory_cabooses(&conn, limit).await
         }
+        InventoryCommands::RotPages => {
```
davepacheco, line 1774: nit: keep the variants alphabetical?

dev-tools/omdb/src/bin/omdb/db.rs
```diff
@@ -235,6 +238,8 @@ enum InventoryCommands {
     BaseboardIds,
     /// list all cabooses ever found
     Cabooses,
+    /// list all root of trust pages every found
+    RotPages,
```
davepacheco, line 242: nit: keep the variants alphabetical?

nexus/inventory/src/builder.rs
davepacheco: If I'm understanding it right: I think the current behavior is okay and also the improvements you alluded to would be good, too.  My thinking is this:

- It seems very possible (although probably uncommon) to get two different responses here due to the underlying data having changed in between the two calls (e.g., if we did an inventory collection during an update, we might see the caboose contents change).  In this case, there's an intrinsic race, and I don't think it matters much which data we take, especially as long as we record which one it was (source and timestamp).  A consumer that was depending on getting the later data really needs to wait for a collection to complete that _starts_ after whatever operation they want it to be after.  (I realize in the current implementation we'll take the later one anyway.  Similarly if someone was depending on it being the earlier one, they'd have to look at a collection that finished before whatever operation they care about happened.)
- The builder returns an error so that the caller can at least tell that something unusual happened and can report it.  What level of reporting is appropriate?  I think calling this a collection error is misleading.  Those are intended to communicate that a particular collection is likely to be incomplete or invalid, which isn't the case in the example I mentioned and I don't think we want that sort of race to invalidate any concurrent collection.  I think a warning-level log message (and counter bump, ideally) would be appropriate.  If we wanted to augment collections with warnings or something like that, I think that'd also be fine, but I'm not sure it's worth doing right now.
- The other possibility is that two MGS's reliably report inconsistent data for the same device even though the underlying data isn't changing.  I can imagine this happening (i.e., due to a bug in one of them), but dealing with it in general seems pretty hard and it doesn't seem all that likely to me.  Being able to tell in a support context that it's happened seems like the right level of effort.

That's just an explanation of how I think we got here.  We could certainly change any of this!

nexus/types/src/inventory.rs
```diff
@@ -177,3 +200,57 @@ pub enum CabooseWhich {
...
     RotSlotB,
 }
+
+/// Root of trust page contents found during a collection
+///
+/// These are normalized in the database.  Each distinct `RotPage` is assigned a
+/// uuid and shared across many possible collections that reference it.
+#[derive(Clone, Debug, Ord, Eq, PartialOrd, PartialEq)]
+pub struct RotPage {
+    pub data_base64: String,
```
davepacheco, line 210: I'm a little surprised this doesn't have some more specific type but I see that the `base64` crate does just give you back a String.

schema/crdb/11.0.0/up2.sql
```diff
@@ -0,0 +1,2 @@
+CREATE UNIQUE INDEX IF NOT EXISTS root_of_trust_page_properties
```
davepacheco, line 1: This does make me slightly nervous (because the values are so big and the domain of possible values is so big) but I guess it's not really worse than any other string and in practice we don't expect to have that many values here.

nexus/db-queries/src/db/datastore/inventory.rs
```diff
@@ -469,6 +491,128 @@ impl DataStore {
                 }
             }
 
+            // Insert rows for the root of trust pages that we found. This is
+            // almost identical to inserting cabooses above, and just like for
+            // cabooses, we do this using INSERT INTO ... SELECT. We have these
+            // three tables:
```
davepacheco, line 497: I'd probably avoid duplicating the existing comment and just say "it looks like the other case".  It's not a big deal.  It just makes it look to me like there's something more special about this case than [I think] there is.

nexus/db-queries/src/db/datastore/inventory.rs
davepacheco: Yeah I thought about this too but couldn't come up with a great way to solve the problem of forgetting to update the inventory delete code.  Would this test have caught your issue?  I wouldn't think so because I assume you'd also have forgotten to update the test to check your new table.

davepacheco (approved):
Nice!

nexus/db-queries/src/db/datastore/inventory.rs
jgallagher: It would not have caught my issue without updating the test to check the new table, no. I did make a pass through updating the tests in this file before realizing I'd forgotten to update the delete code, though, so maybe I would have added the table here and then seen a test failure? Hard to know, and I'm not sure whether keeping this around is worth it - thoughts?

I can only think of terrible ways to try to ensure we've checked all the tables that might hold dangling collection IDs, none of which are worth voicing aloud I don't think.

nexus/inventory/src/builder.rs
jgallagher: This makes sense, thanks. I was only thinking of your third bullet point (or slight variations, such as an SP / RoT reporting inconsistent data itself, even if MGS isn't buggy), but you're right that this seems we're much more likely to hit this by racing with an update, which shouldn't cause a failure.

> I think a warning-level log message (and counter bump, ideally) would be appropriate.

`Collector` does log this (at `error!`) level currently - want me to change them to `warn!`, and possibly include a "was an update happening simultaneously?" (either in the log message itself or in a comment around where the log message is generated)?

What do you mean by a counter bump?

schema/crdb/11.0.0/up2.sql
jgallagher: I'm also slightly nervous about this, but (a) I believe these change even less frequently than cabooses (e.g., when keys are revoked, for which we don't yet have a process) and (b) I'm not sure what an alternative would be. Without this index we wouldn't be able to ensure uniqueness "by hand", right, since querying to see if a given page exists would require a full table scan?

nexus/db-queries/src/db/datastore/inventory.rs
davepacheco: I don't think it's worth keeping but I don't mind if you want to.

This is getting off-topic but the bestthing I've come to is this: this issue does come up a fair bit and I could imagine a mechanism where we take a database dump before and after a test and verify that it hasn't left detritus around.  This mechanism would have to be configurable since the system in general _does_ produce stuff unrelated to the test.  For example, every _other_ Nexus test will wind up having an inventory collection in the database after it runs and it should ignore that.  But these tests shouldn't.  Similarly, these tests should ignore stuff generated by other subsystems, but _not_ the inventory stuff.  I could imagine grouping tables into categories and having the mechanism say "this category should now be empty" or something.  Even this only really solves this problem if we enforce that every table _have_ a category so that we can't accidentally forget to add new tables like in the case here.

Anyway, I think this is probably a good idea and well beyond the scope of this PR.

nexus/inventory/src/builder.rs
davepacheco: I wouldn't bother changing it.

By counter bump I mean adding Oximeter-collected metrics to this subsystem and having an error counter for this case.  That'd be welcome but obviously beyond the scope of this PR!

schema/crdb/11.0.0/up2.sql
davepacheco: Agreed.

PR #4497, opened 2023-11-14
Add a producer kind to oximeter metric producers

bnaecker:
- Adds the `kind` enum to metric producer information, including DB schema, model, and various client parameter types. This records the supported types of metric producers, and is intended to aid debugging and future work around updates and instance lifecycle management.
- Add schema update files which create the DB enum type and add it as a column to the `metric_producer` table. This currently _drops_ the existing table and recreates it with the new column, rather than adding the column using `ALTER TABLE`. That is intended to remove old entries in bulk, since nothing previously removed the records for Propolis servers when their instance was stopped.

This is the initial PR in a sequence that will eventually make this field _required_ in both the database and API requests. As there are consumers of this API outside of the Omicron repository, this field needs to start as optional, to avoid introducing a commit with incompatible clients. Here are the planned steps:

- This PR introduces an optional `kind` field into the `metric_producer` table and `ProducerEndpoint` registration request body
- A PR in [Crucible](https://github.com/oxidecomputer/crucible/pull/1039) which includes the optional field in its API requests
- A PR in [Dendrite](https://github.com/oxidecomputer/dendrite) which includes the optional field in its API requests
- A PR in [Propolis](https://github.com/oxidecomputer/propolis) which includes the optional field in its API requests
- A PR in Omicron which makes the database and API request fields required, and points to the new versions of the three previous dependencies
- A PR in Dendrite, Crucible, and Propolis which adapts to the new required field from the previous step
- A final PR in Omicron which points to the new versions of the three previous dependencies

### DIFF
--- a/common/src/api/internal/nexus.rs
+++ b/common/src/api/internal/nexus.rs
@@ -84,13 +84,34 @@ pub struct SledInstanceState {
 
 // Oximeter producer/collector objects.
 
+/// The kind of metric producer this is.
+#[derive(Clone, Copy, Debug, Deserialize, JsonSchema, PartialEq, Serialize)]
+#[serde(rename_all = "snake_case")]
+pub enum ProducerKind {
+    /// The producer is a sled-agent.
+    SledAgent,
+    /// The producer is an Omicron-managed service.
+    Service,
+    /// The producer is a Propolis VMM managing a guest instance.
+    Instance,
+}
+
 /// Information announced by a metric server, used so that clients can contact it and collect
 /// available metric data from it.
 #[derive(Clone, Debug, Deserialize, JsonSchema, Serialize, PartialEq)]
 pub struct ProducerEndpoint {
+    /// A unique ID for this producer.
     pub id: Uuid,
+    /// The kind of producer.
+    pub kind: Option<ProducerKind>,
+    /// The IP address and port at which `oximeter` can collect metrics from the
+    /// producer.
     pub address: SocketAddr,
+    /// The API base route from which `oximeter` can collect metrics.
+    ///
+    /// The full route is `{base_route}/{id}`.
     pub base_route: String,
+    /// The interval on which `oximeter` should collect metrics.
     pub interval: Duration,
 }
 
--- a/schema/crdb/dbinit.sql
+++ b/schema/crdb/dbinit.sql
@@ -1109,12 +1109,25 @@ CREATE TABLE IF NOT EXISTS omicron.public.oximeter (
 );
 
 /*
+ * The kind of metric producer each record corresponds to.
+ */
+CREATE TYPE IF NOT EXISTS omicron.public.producer_kind AS ENUM (
+    -- A sled agent for an entry in the sled table.
+    'sled_agent',
+    -- A service in the omicron.public.service table
+    'service',
+    -- A Propolis VMM for an instance in the omicron.public.instance table
+    'instance'
+);
+
+/*
  * Information about registered metric producers.
  */
 CREATE TABLE IF NOT EXISTS omicron.public.metric_producer (
     id UUID PRIMARY KEY,
     time_created TIMESTAMPTZ NOT NULL,
     time_modified TIMESTAMPTZ NOT NULL,
+    kind omicron.public.producer_kind,
     ip INET NOT NULL,
     port INT4 CHECK (port BETWEEN 0 AND 65535) NOT NULL,
     interval FLOAT NOT NULL,
@@ -2906,7 +2919,7 @@ INSERT INTO omicron.public.db_metadata (
     version,
     target_version
 ) VALUES
-    ( TRUE, NOW(), NOW(), '11.0.0', NULL)
+    ( TRUE, NOW(), NOW(), '12.0.0', NULL)
 ON CONFLICT DO NOTHING;
 
 COMMIT;
--- a/clients/nexus-client/src/lib.rs
+++ b/clients/nexus-client/src/lib.rs
@@ -202,6 +202,19 @@ impl From<&types::InstanceState>
     }
 }
 
+impl From<omicron_common::api::internal::nexus::ProducerKind>
+    for types::ProducerKind
+{
+    fn from(kind: omicron_common::api::internal::nexus::ProducerKind) -> Self {
+        use omicron_common::api::internal::nexus::ProducerKind;
+        match kind {
+            ProducerKind::SledAgent => Self::SledAgent,
+            ProducerKind::Service => Self::Service,
+            ProducerKind::Instance => Self::Instance,
+        }
+    }
+}
+
 impl From<&omicron_common::api::internal::nexus::ProducerEndpoint>
     for types::ProducerEndpoint
 {
@@ -212,6 +225,7 @@ impl From<&omicron_common::api::internal::nexus::ProducerEndpoint>
             address: s.address.to_string(),
             base_route: s.base_route.clone(),
             id: s.id,
+            kind: s.kind.map(Into::into),
             interval: s.interval.into(),
         }
     }
--- a/clients/oximeter-client/src/lib.rs
+++ b/clients/oximeter-client/src/lib.rs
@@ -20,6 +20,19 @@ impl From<std::time::Duration> for types::Duration {
     }
 }
 
+impl From<omicron_common::api::internal::nexus::ProducerKind>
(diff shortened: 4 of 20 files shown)


common/src/api/internal/nexus.rs
```diff
@@ -84,13 +84,34 @@ pub struct SledInstanceState {
...
+}
+
 /// Information announced by a metric server, used so that clients can contact it and collect
 /// available metric data from it.
 #[derive(Clone, Debug, Deserialize, JsonSchema, Serialize, PartialEq)]
 pub struct ProducerEndpoint {
+    /// A unique ID for this producer.
     pub id: Uuid,
+    /// The kind of producer.
+    pub kind: Option<ProducerKind>,
```
davepacheco, line 106: Should this be required?  Is it just that we're worried about API compatibility?

common/src/api/internal/nexus.rs
```diff
@@ -84,13 +84,34 @@ pub struct SledInstanceState {
 
 // Oximeter producer/collector objects.
 
+/// The kind of metric producer this is.
+#[derive(Clone, Copy, Debug, Deserialize, JsonSchema, PartialEq, Serialize)]
+#[serde(rename_all = "snake_case")]
+pub enum ProducerKind {
+    /// The producer is a sled-agent.
+    SledAgent,
+    /// The producer is an Oxide-managed service.
```
davepacheco, line 93:
```suggestion
    /// The producer is an Omicron-managed service.
```

schema/crdb/11.0.0/up01.sql
```diff
@@ -0,0 +1,23 @@
...
+ * Drop the entire metric producer assignment table.
+ *
+ * Programs wishing to produce metrics need to register with Nexus. That creates
+ * an assignment of the producer to a collector, which is recorded in this
+ * table. That registration is idempotent, and every _current_ producer will
+ * register when it restarts. For example, `dpd` includes a task that registers
+ * with Nexus, so each time it (re)starts, that registration will happen.
+ *
+ * With that in mind, dropping this table is safe, _because all updates are
+ * currently offline_. The current metric producers are:
```
davepacheco, line 11:
```suggestion
 * With that in mind, dropping this table is safe, because as of today all software updates require that the whole control plane be offline, so we know that these entries will be re-created shortly. The current metric producers are:
```

schema/crdb/11.0.0/up01.sql
```diff
@@ -0,0 +1,23 @@
+/*
+ * Drop the entire metric producer assignment table.
```
davepacheco, line 2: It might be worth mentioning that we're doing this because we want to add a _required_ column and we don't have a great way to backfill the data -- _but_ we also don't need to because the data will be immediately repopulated.

schema/crdb/dbinit.sql
```diff
@@ -1108,13 +1108,26 @@ CREATE TABLE IF NOT EXISTS omicron.public.oximeter (
...
+);
+
 /*
  * Information about registered metric producers.
  */
 CREATE TABLE IF NOT EXISTS omicron.public.metric_producer (
     id UUID PRIMARY KEY,
     time_created TIMESTAMPTZ NOT NULL,
     time_modified TIMESTAMPTZ NOT NULL,
+    kind omicron.public.producer_kind,
```
davepacheco, line 1130: Same as for the Rust code: shouldn't this be required?

common/src/api/internal/nexus.rs
bnaecker: Yeah, I was working under the assumption that this needed to be a multi-step process: introduce an optional field in the API and DB; update Propolis and Dendrite to send that field; and then make the field non-optional. If it is required, I think we have a commit which can't possibly work. Maybe that's OK?

schema/crdb/dbinit.sql
bnaecker: I think the same answer applies here. Yes, I want to make this required, but I'm not sure how we can do that in one commit because of the producers outside the repository. Or at least, without introducing a commit in which Propolis and Dendrite will never be able to produce metrics, which might be fine.

common/src/api/internal/nexus.rs
davepacheco: Got it -- this is because of the consumers outside of Omicron that are then used in other parts of Omicron.  I hadn't realized (or forgot) there was going to be a follow-on PR.  I think it might be worth writing this plan down in the PR description so it's clear what's going to happen and also what the state is at each point.  (For example, what actually happens if you were to build a TUF repo after landing this PR but before the others?)

My suggestion for stuff like this (and you may have been planning to do this already) is to get all the PRs lined up and not land any of them until we're happy with all of them and the sequence itself.  I sometimes find a problem when I get to the later ones that necessitates fixing an "earlier" one.

bnaecker: > this is because of the consumers outside of Omicron

Yeah, exactly.

> I think it might be worth writing this plan down in the PR description

Sounds will, I'll update that now

> get all the PRs lined up

Yep, I've got branches in the Dendrite and Propolis repositories which add the new optional field to their registration requests.

bnaecker:
This whole dance requires a sequence of updates, laid out in the PR description. There are currently a few other revisions that need to be orchestrated between Crucible, Propolis, and Omicron, so I'm going to wait until those are through the pipeline before starting on this sequence.

bnaecker:
Friendly ping on this @davepacheco, I'd like to tie this off before the holiday, during a quiet few days, if possible.

davepacheco:
I think the change is fine and I'm just trying to work through the sequence of PRs to make sure we aren't going to accidentally break something.  Sorry to be tedious about it but every time I _don't_ spell it out like this I get it wrong and break stuff.

1. **This PR to `omicron` repo: add an _optional_ field to the Nexus internal API and update any consumers in this repo.**  Other repos with consumers of this API (Crucible, Propolis, and Dendrite) continue to build using the older API version and client.  So builds all work.  If you deploy the TUF repo built from this version, you'll get a Nexus that accepts the optional parameter and the out-of-repo components (Crucible, Propolis, Dendrite) will _not_ be providing it.
2. **PR to `propolis` that updates the Nexus client to the newer version and provides the new parameter as `Some(some_kind)`.**  Nothing changes in the build or deployment of any other repo.   If you rebuild the full TUF repo at this point, there's no change from the previous state since `omicron` is still pointing at the previous Propolis.
3. **PR to `dendrite` that does the same, with similar results.**  Again, at this point, if you rebuilt and deployed the TUF repo, you'd be getting a Nexus that accepts the optional parameter and components that do _not_ provide it. 
4. **PR to `crucible` that does the same.**  etc.
5. **PR to `omicron` repo to make the field required, both in the API and the database schema.**  At this point, all repo _builds_ continue to work.  Crucible, Propolis, and Dendrite are still using a Nexus client where the parameter is optional.  If you deploy at this point, you're deploying a Nexus that _requires_ the parameter and a bunch of components that provide it _thinking_ it's optional.  (Which is fine!  This works because providing an optional parameter and a required parameter looks the same on the wire.)
6. **PR to `propolis` that updates the Nexus client to the newest version and provides the new parameter as `some_kind`** (compare with `Some(some_kind)` above).  Nothing changes in the build or deployment of any other repo.  If you rebiuld the full TUF repo at this point, there's no change from the previous state since `omicron` is still pointing at the previous Propolis.
7. **PR to `dendrite` that does the same, with similar results.**  Again, at this point, if you rebuilt and deployed the TUF repo, you'd be getting a Nexus that requires the parameter and components that _do_ provide it, albeit believing that it's optional.
8. **PR to `crucible` that does the same.**  etc.
9. **PR to `omicron` once again that updates the versions of `propolis`, `dendrite`, and `crucible` that it pulls down.**  At this point if you build and deploy the TUF repo, you'll get components that are all aligned on the fact that this parameter is required.

By contrast, I asked offline what would happen if we skipped the "optional" phase altogether.  If we did, then in step 1, if you built and deployed the TUF repo, you'd get a Nexus that requires the parameter and a Propolis that doesn't provide it and that would not work.

davepacheco (approved):

bnaecker:
Update: Steps 1-4 have been completed. Omicron has support for the optional parameter, as well as the 3 out-of-tree consumers. I haven't yet pointed Omicron at those updated commits.

bnaecker:
#4571 is up for review now, which takes this through step 5/

PR #4500, opened 2023-11-15
Send oximeter all existing producer assignments on registration

### LINKED ISSUE #4498: Oximeter producer registration needs to register harder
When an `oximeter` collector registers itself with `nexus`, the latter attempts to reassign all the existing metric producers to that `oximeter` instance. This is to make sure that `oximeter` restarts are handled pretty gracefully, and that it starts collecting again from everything it needs to.

That code is here:

https://github.com/oxidecomputer/omicron/blob/a8a49c396b90083f248b103580724dd42dca297d/nexus/src/app/oximeter.rs#L93-L104

It's also wrong. We're paginating that list from the DB, but only ever taking the first list. We're potentially abandoning any producers after that first page. We can actually see this on dogfood. There are really around 2200 records in the `metric_producer` table right now, but the one `oximeter` collector we have only knows about 255 of them:

```
root@oxz_switch:~# /tmp/omdb oximeter --oximeter-url http://[fd00:1122:3344:10a::3]:12223 list-producers | wc -l
     255
```

Presumably anything after the first page are producers that were registered after the last restart of `oximeter`.

This code needs to be updated to actually paginate the entire list and make sure those are all sent to `oximeter`.

bnaecker:
- Fixes #4498
- When `oximeter` registers with `nexus`, we send it any producer assignments so that it can continue collecting after a restart. The list of assignments is paginated from the database, but we previously only sent the first page. This ensures we send all records, and adds a regression test.

### DIFF
--- a/nexus/src/app/oximeter.rs
+++ b/nexus/src/app/oximeter.rs
@@ -87,32 +87,43 @@ impl super::Nexus {
             "address" => oximeter_info.address,
         );
 
-        // Regardless, notify the collector of any assigned metric producers. This should be empty
-        // if this Oximeter collector is registering for the first time, but may not be if the
-        // service is re-registering after failure.
-        let pagparams = DataPageParams {
-            marker: None,
-            direction: PaginationOrder::Ascending,
-            limit: std::num::NonZeroU32::new(100).unwrap(),
-        };
-        let producers = self
-            .db_datastore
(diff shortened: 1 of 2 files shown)


smklein (approved):
heh, I love ways to improve scalability like this. "read the rest of the pages" is easy enough to review!

Thanks for the regression test!

Issue #3514, opened 2023-07-06
wtmp record mangling could use some tuning


wesolows:
After #2516 we note have `boottime_rewrite` and co. to rewrite the boot time record in utmp/wtmp; see https://github.com/oxidecomputer/omicron/blob/c2de4805acfd0c20bef9d87afacac9f8e821ad87/sled-agent/src/services.rs#L2525-L2565.  The original intent of this was to make code that reads this record (notably but probably not exclusively uptime(1)) work "normally" to reduce surprises for post-mortem debugging with respect to the contents of support bundles, as well as anyone doing bringup or engineering work on Oxide-owned machines who might be able to execute commands.

As it stands today, this has two flaws:

1. It writes the *current* wall time as the boot time.  This is kind of almost right, but invariably sled-agent starts up -- and NTP completes synching the wall clock -- some seconds or more likely minutes after the kernel actually started keeping time.  This isn't catastrophic since at least the record is vaguely plausible, unlike 28 Dec 1986, but it's just wrong enough that it might confuse someone.  We should instead subtract from the current NTP-supplied wall time the amount of time the sled has been running.  This should allow us sub-second accuracy which is basically fine and probably as good as we get on other illumos systems anyway.  In particular, `CLOCK_MONOTONIC` on illumos is `CLOCK_HIGHRES` which should not be affected by leap seconds or NTP adjustments (whether slow or fast).  Both `gethrtime()` and `clock_gettime(CLOCK_HIGHRES)` should supply the absolute time since the kernel started its clock on illumos.  `CLOCK_MONOTONIC` is not guaranteed to do this by POSIX, and in particular it does not on Linux, but this code isn't intended to be platform-agnostic and if for some reason (test mocks?) it needed to be, there are nonportable alternatives for most other platforms we might care about.
[...]

PR #4501, opened 2023-11-15
Improve wtmp record mangling.

citrus-it:
This fixes two problems with the current wtmp/utmp record mangling. First it is not idempotent, the system's notion of boot time is set to the current time of day whenever sled agent confirms that time is synchronised. Secondly, this is only approximate even for the first sled agent start, but it's plain wrong if sled agent restarts later.

In conjunction with changes to stlouis, the `tmpx` utility is now able to process all zones itself, and uses the true system boot time for each zone when updating records.

Fixes: https://github.com/oxidecomputer/omicron/issues/3514

### DIFF
--- a/sled-agent/src/services.rs
+++ b/sled-agent/src/services.rs
@@ -100,13 +100,11 @@ use sled_storage::manager::StorageHandle;
 use slog::Logger;
 use std::collections::BTreeMap;
 use std::collections::HashSet;
-use std::iter;
 use std::iter::FromIterator;
 use std::net::{IpAddr, Ipv6Addr, SocketAddr};
 use std::str::FromStr;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
-use std::time::{SystemTime, UNIX_EPOCH};
 use tokio::io::AsyncWriteExt;
 use tokio::sync::oneshot;
 use tokio::sync::Mutex;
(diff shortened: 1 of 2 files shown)


sled-agent/src/services.rs
```diff
@@ -2403,33 +2398,13 @@ impl ServiceManager {
...
-                    info!(self.inner.log, "Updated {}", &file);
-                }
-            }
+        // Call out to the 'tmpx' utility program which will rewrite the wtmpx
+        // and utmpx databases in every zone, including the global zone, to
+        // reflect the adjusted system boot time.
+        let mut command = std::process::Command::new(PFEXEC);
+        let cmd = command.args(&["/usr/platform/oxide/bin/tmpx", "-Z"]);
+        if let Err(e) = execute(cmd) {
+            warn!(self.inner.log, "Updating [wu]tmpx databases failed: {}", e);
```
jgallagher, line 2407: We lost the `info!` log if this succeeds (although maybe it's less interesting to log "we successfully told `tmpx` to do its thing" than "we told tmpx to modify a specific file"?). I have no strong feeling either way but just wanted to confirm this is intentional.

jgallagher (approved):
I don't know anything about `tmpx`, but the local sled-agent changes LGTM

sled-agent/src/services.rs
citrus-it: Yes, I had the same thought. The new `tmpx` is fire and forget, it will traverse all zones and idempotently fix up boot times, producing an error if anything goes wrong. I didn't think it was valuable to keep the success line (but I also have no strong feeling)

PR #4503, opened 2023-11-16
Pare down nexus-client dependencies

pfmooney:
Adding sled-hardware and sled-storage as dependencies to nexus-client bring in a large portion of unrelated crates, all in service of some (albeit convenient) type conversion routines.  For the sake of nexus-client consumers outside of omicron, we can manually implement those few conversions in sled-agent where they are consumed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3934,8 +3934,6 @@ dependencies = [
  "schemars",
  "serde",
  "serde_json",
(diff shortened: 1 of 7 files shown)


andrewjstone (approved):
Looks great. Thanks for cleaning up my mess @pfmooney

PR #4504, opened 2023-11-16
[nexus] List all uninitialized sleds

andrewjstone:
As part of adding a sled to an already initialized rack, we need a way for operators to be able to list sleds that are not part of a rack. This PR adds an external nexus endpoint for doing just that. Like the `sleds` endpoint, this endpoint is not tied to a rack.

The way this works is by looking at the SPs in the latest inventory collection and finding all the Baseboards that are not in the `sled` table in CRDB.

A follow up commit will allow an operator to add uninitialized sleds to a rack with a new external nexus endpoint.

### DIFF
--- a/nexus/db-model/src/rack.rs
+++ b/nexus/db-model/src/rack.rs
@@ -4,8 +4,9 @@
 
 use crate::schema::rack;
 use db_macros::Asset;
-use ipnetwork::IpNetwork;
+use ipnetwork::{IpNetwork, Ipv6Network};
 use nexus_types::{external_api::views, identity::Asset};
+use omicron_common::api;
 use uuid::Uuid;
 
 /// Information about a local rack.
@@ -26,6 +27,22 @@ impl Rack {
             initialized: false,
             tuf_base_url: None,
             rack_subnet: None,
+        }
+    }
+
+    pub fn get_subnet(&self) -> Result<Ipv6Network, api::external::Error> {
+        match self.rack_subnet {
+            Some(IpNetwork::V6(subnet)) => Ok(subnet),
+            Some(IpNetwork::V4(_)) => {
+                return Err(api::external::Error::InternalError {
+                    internal_message: "rack subnet not IPv6".into(),
+                })
+            }
+            None => {
+                return Err(api::external::Error::InternalError {
+                    internal_message: "rack subnet not set".into(),
+                })
+            }
         }
     }
 }
--- a/nexus/db-model/src/rack.rs
+++ b/nexus/db-model/src/rack.rs
@@ -30,7 +30,7 @@ impl Rack {
         }
     }
 
-    pub fn get_subnet(&self) -> Result<Ipv6Network, api::external::Error> {
+    pub fn subnet(&self) -> Result<Ipv6Network, api::external::Error> {
         match self.rack_subnet {
             Some(IpNetwork::V6(subnet)) => Ok(subnet),
             Some(IpNetwork::V4(_)) => {
--- a/nexus/db-queries/src/db/datastore/inventory.rs
+++ b/nexus/db-queries/src/db/datastore/inventory.rs
@@ -1236,6 +1236,18 @@ mod test {
         assert_eq!(collection2.baseboards.len(), nbaseboards);
         assert_eq!(collection2.cabooses.len(), ncabooses);
 
+        // Check that we get an error on the limit being reached for
+        // `read_all_or_nothing`
+        let limit = NonZeroU32::new(1).unwrap();
+        assert!(datastore
+            .inventory_collection_read_all_or_nothing(
+                &opctx,
+                collection2.id,
+                limit
+            )
+            .await
+            .is_err());
+
         // Now insert an equivalent collection again.  Verify the distinct
         // baseboards and cabooses again.  This is important: the insertion
         // process should re-use the baseboards and cabooses from the previous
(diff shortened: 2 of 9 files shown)


nexus/db-model/src/rack.rs
```diff
@@ -28,6 +29,22 @@ impl Rack {
             rack_subnet: None,
         }
     }
+
+    pub fn get_subnet(&self) -> Result<Ipv6Network, api::external::Error> {
```
davepacheco, line 33: super nit: I think we usually don't use `get_` prefixes (per https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter)

nexus/db-model/src/rack.rs
```diff
@@ -28,6 +29,22 @@ impl Rack {
             rack_subnet: None,
         }
     }
+
```
davepacheco, line 32: (can't put this comment in the right place, which is L20)

I think maybe `rack_subnet` be non-`pub` if we want people to use the new getter.

nexus/src/app/rack.rs
```diff
@@ -724,4 +716,51 @@ impl super::Nexus {
...
     }
+
+    /// Return the list of sleds that are inserted into an initialized rack
+    /// but not yet initialized as part of a rack.
+    pub(crate) async fn uninitialized_sled_list(
+        &self,
+        opctx: &OpContext,
+    ) -> ListResultVec<UninitializedSled> {
+        // Grab the SPs from the last collection
+        let limit = NonZeroU32::new(50).unwrap();
```
davepacheco, line 727: Can we fail more gracefully if this limit is wrong for some reason?  Like should we produce a 500 if we actually get `limit` results back?

nexus/src/app/rack.rs
```diff
@@ -724,4 +716,51 @@ impl super::Nexus {
...
+        let collection = self
+            .db_datastore
+            .inventory_get_latest_collection(opctx, limit)
+            .await?;
+        let pagparams = DataPageParams {
+            marker: None,
+            direction: dropshot::PaginationOrder::Descending,
+            // TODO: This limit is only suitable for a single sled cluster
+            limit: NonZeroU32::new(32).unwrap(),
```
davepacheco, line 736: Same here

nexus/db-queries/src/db/datastore/inventory.rs
```diff
@@ -798,54 +797,311 @@ impl DataStore {
...
+}
+
 impl DataStoreInventoryTest for DataStore {
     fn inventory_collections(&self) -> BoxFuture<anyhow::Result<Vec<Uuid>>> {
         async {
             let conn = self
                 .pool_connection_for_tests()
                 .await
-                .context("getting connectoin")?;
+                .context("getting connection")?;
```
davepacheco, line 1104: oops!  thanks

nexus/db-queries/src/db/datastore/inventory.rs
```diff
@@ -863,257 +1119,11 @@ impl DataStoreInventoryTest for DataStore {
...
-                limit_reached,
-            ))
-        }
-        .boxed()
-    }
 }
 
 #[cfg(test)]
 mod test {
+    use crate::context::OpContext;
```
davepacheco, line 1126: What do you think about adding a test that uses read_all_or_nothing with too-short a limit to make sure it fails in the expected way?  (I'm hopeful this could be like 5 lines in the existing test.)

davepacheco (approved):
I have a few small suggestions but this looks great.

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -4382,6 +4383,25 @@ async fn rack_view(
...
 
+/// List uninitialized sleds in a given rack
+#[endpoint {
+    method = GET,
+    path = "/v1/system/hardware/uninitialized-sleds",
+    tags = ["system/hardware"]
+}]
+async fn uninitialized_sled_list(
+    rqctx: RequestContext<Arc<ServerContext>>,
+) -> Result<HttpResponseOk<Vec<UninitializedSled>>, HttpError> {
```
karencfv, line 4394: Should we return a paginated response here? I realise the number of unitialised sleds probably won't be enough to warrant pagination, but it seems like an unwritten convention to always return a `ResultsPage<T>` with "list" endpoints.

andrewjstone: That was my original take as well @karencfv. But @davepacheco and I talked about it, and it seems like overkill given the physical limitations.

nexus/db-model/src/rack.rs
andrewjstone: The app code for setting the field when syncing the bootstore uses the field directly, and that seems correct to me.

nexus/src/app/rack.rs
andrewjstone: I believe we do fail gracefully if the limit is reached here. We pass this limit into `inventory_collection_read_all_or_nothing` which ends up [calling `bail_unless` if the limit is reached](https://github.com/oxidecomputer/omicron/pull/4504/files#diff-5d05ec64d061b45e7c1e131a7914eb2213e423e5ca75a0c6c5983a6e7721f462R838)

nexus/src/app/rack.rs
andrewjstone: I started to do this, but we can safely hit the limit of 32 here. I could change this to be 33 so that we error on multi-rack I suppose, but by that point we'll have many other issues in testing.

nexus/db-model/src/rack.rs
andrewjstone: Done in [8ad101f](https://github.com/oxidecomputer/omicron/pull/4504/commits/8ad101fff837f1373947e10b207367b34bf7932f)

nexus/db-queries/src/db/datastore/inventory.rs
andrewjstone: Done in [8ad101f](https://github.com/oxidecomputer/omicron/pull/4504/commits/8ad101fff837f1373947e10b207367b34bf7932f)

PR #4506, opened 2023-11-16
[db-queries] Avoid interactive transaction for all authz checks

smklein:
While working on https://github.com/oxidecomputer/customer-support/issues/46 , I noticed that we perform an interactive transaction for every authz check perfomed by the database.

This PR re-writes this code to avoid using an interactive transaction altogether.


luqmana (approved):
Oh yea, this is an easy win here!

davepacheco:
Good call!

PR #4510, opened 2023-11-17
[wicket] produce a better error message if MGS isn't available

sunshowers:
Currently, we just time out in this case. Produce a better error message by
waiting for 80% of WICKETD_TIMEOUT.

### DIFF
--- a/wicketd/src/http_entrypoints.rs
+++ b/wicketd/src/http_entrypoints.rs
@@ -904,7 +904,8 @@ async fn post_start_update(
     //    out of an abundance of caution).
     //
     // First, get our most-recently-cached inventory view. (Only wait 80% of
-    // WICKETD_TIMEOUT for this -- if the inventory isn't available, then we
+    // WICKETD_TIMEOUT for this: if even a cached inventory isn't available,
+    // it's because we've never established contact with MGS. In that case, we
     // should produce a useful error message rather than timing out on the
     // client.)
     let inventory = match tokio::time::timeout(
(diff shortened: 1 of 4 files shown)


wicketd/src/http_entrypoints.rs
```diff
@@ -896,21 +897,37 @@ async fn post_start_update(
...
+        Ok(Err(ShutdownInProgress)) => {
             return Err(HttpError::for_unavail(
                 None,
                 "Server is shutting down".into(),
             ));
         }
+        Err(_) => {
+            // Use 400 Bad Request instead of 503 Service Unavailable so that
+            // the error message is propagated to the client.
+            return Err(HttpError::for_bad_request(
```
jgallagher, line 926: We can (and probably should in this case, right?) return a 503 while still providing a client-visible error message; it's just more verbose: https://github.com/oxidecomputer/omicron/blob/ca9d90a79c2a7a9ffdca12a705d007d7f28ce61f/wicketd/src/http_entrypoints.rs#L389-L395

`external_message` specifically is returned to the client. In the [helper method](https://github.com/oxidecomputer/dropshot/blob/1e187aff50fc8559330c63e6693cc4deade22e28/dropshot/src/error.rs#L182-L185) it's set to "Service Unavailable", but the fields are intentionally pub to allow this, I believe.

wicketd/src/http_entrypoints.rs
```diff
@@ -896,21 +897,37 @@ async fn post_start_update(
...
-    // First, get our most-recently-cached inventory view.
-    let inventory = match rqctx.mgs_handle.get_cached_inventory().await {
-        Ok(inventory) => inventory,
-        Err(ShutdownInProgress) => {
+    // First, get our most-recently-cached inventory view. (Only wait 80% of
+    // WICKETD_TIMEOUT for this -- if the inventory isn't available, then we
+    // should produce a useful error message rather than timing out on the
+    // client.)
+    let inventory = match tokio::time::timeout(
+        WICKETD_TIMEOUT.mul_f32(0.8),
```
jgallagher, line 911: I was going to ask if this was artificially shortening the timeout for just this endpoint, but since we're trying to get the _cached_ inventory we only expect this to timeout if MGS has never been available, right? I.e., if we've ever gotten inventory from MGS this will return immediately?

jgallagher (approved):

wicketd/src/http_entrypoints.rs
sunshowers: Ah interesting! Thanks for the pointer.

wicketd/src/http_entrypoints.rs
sunshowers: Yeah. I'll add a comment to this effect.

Issue #2483, opened 2023-03-03
Need a way to take a sled out of the provisioning pool


askfongjojo:
There are several situations when we do not want new VM instances to be placed on a certain sled:
1. when the operator or update/migrate program needs to drain a sled prior to SW/HW maintenance
2. when the migrate program wants to reserve the sled so that it has the expected capacity for VMs it's trying to migrate over
3. when the operator wants to reserve the sled to migrate in it a "noisy neighbor" from another sled

askfongjojo:
Per suggestion in RFD 378, we can tentatively introduce `reserved` as a sled status (and `reserve` as the action in API/Console).

jordanhendricks:
@askfongjojo and I chatted a bit about scoping this work out more cleanly for MVP. I believe these are the determinations we arrived at.

There are a couple of use cases I believe we need to prioritize for MVP with a feature that removes a sled from consideration for provisioning:
1. stopping new provisions to a sled so that we can migrate instances off of it, update the sled, and reboot it
2. noticing a sled is unusable (due to a crash, physical pull from the cubby, whatever) and stopping new provisions there (see also: #2986)

Of these two cases, it seems to me that (1) is much more critical for MVP; if we can't stop new provisions to a sled, we create more unbounded work for migrating instances from it during the update path. The situation for (2) is not ideal, but I believe the worst thing that happens in the current provisioning scheme today (i.e., selecting a random sled) is that we would return an error, and it is possible to retry such an operation.

As a starting point, I propose we do the following:
- add a boolean column to the sled table indicating whether the sled is provisionable (and I think we likely want an associated timestamp for its last update, and at least a log message when it is toggled)
- modify `sled_reservation_create` to consider this field in its query to determine an appropriate sled for provisioning

There are some open questions here about if we want an API endpoint here (internal, or external for fleet operators) to change a sled's state in this way. As I see it, the most important consumer for now is the update process, and my understanding is that this process is still being ironed out. So for now, I propose that we punt on an API for this and I can add the plumbing to update the database schema and query the new field for a first PR. As update evolves and we have a better sense of what is needed, we can make determinations for what type of API we need to support around this behavior.

Lastly, this ticket mentions more complex status of sled state, such as goals outlined in RFD 378. One could also imagine us evolving the "provisioning pool" concept to multiple types of pools that might be operator-specified. I think that is something we can also punt on for the MVP timeframe, and as we move toward supporting such a world, we could design a more general-purpose feature.

[...]

PR #4520, opened 2023-11-20
[nexus] add sled provision state

sunshowers:
Add the notion of a sled provision state to Nexus. Currently, we will only use this to prevent new resources and regions from being provisioned to sleds.

This PR includes:

1. Database updates and schema migrations.
2. Database APIs in `nexus-db-queries`.
3. An HTTP API.
4. Tests for resource and region allocation.

Questions:

- [x] Verify transactionality -- I think this is good enough as is but we do look up the old sled value and update it in two separate txns.
- [x] I haven't touched this before, so I might have made some newbie mistakes! I feel like I got a decent sense.

<details><summary>TODO list</summary>
<p>
TODO:

- [x] Figure out whether [`db::model::Sled`](https://github.com/oxidecomputer/omicron/blob/4951c339a8987c757596207c81afdfada8bbf7c7/nexus/db-model/src/sled.rs#L39) should track provision state -- it seems like it is created by sled_agent API calls, which don't know the provision state. But `sled_list` should return the provision state. Do we need to split up the type into two, one for creation and one for listing? Are there existing patterns for this?
  - Addressed via #4533.
- [x] I noticed that `db::model::Sled` has a `last_used_address` parameter that is [assigned here](https://github.com/oxidecomputer/omicron/blob/4951c339a8987c757596207c81afdfada8bbf7c7/nexus/db-model/src/sled.rs#L72-L76) but [not used in `sled_upsert`](https://github.com/oxidecomputer/omicron/blob/4951c339a8987c757596207c81afdfada8bbf7c7/nexus/db-queries/src/db/datastore/sled.rs#L32). This seems maybe a bit dodgy?
  - Addressed via #4533.
- [x] Add APIs for updating the sled provision state.
  - [x] Do we need to return the old provision state?
  - [ ] Verify transactionality/idempotency.
- [x] Add tests.
  - [x] Region allocation.
  - [x] Resource allocation.
- [x] Add schema migration.
</p>
</details> 

Fixes (maybe partially) #2483.


nexus/db-queries/src/db/datastore/sled.rs
```diff
@@ -217,6 +232,34 @@ impl DataStore {
...
+        // Ensure that opctx is authorized to modify this sled.
+        let (_, sled) = sled_lookup.fetch_for(authz::Action::Modify).await?;
+        let old_state = sled.provision_state();
+        if old_state == state {
+            return Ok(old_state);
+        }
+
+        use db::schema::sled::dsl;
+        // XXX: Do we need transaction semantics here? Probably not because
+        // this is idempotent?
```
smklein, line 252: There are a couple very small tweaks I might make here:

Add `filter(dsl::time_deleted.is_null())` when searching for this sled. If we've decided the sled should be deleted, we should not enable further modifications.

On transactions: This is a single statement, so it is implicitly treated as a transaction: https://www.cockroachlabs.com/docs/stable/transactions#individual-statements . The only part of this that *isn't* considered transactional is that in-memory check of "if old_state == state" that's being done in memory, because someone could modify the DB state while we're making that state.

This is a very small concern, but I'll mention it because I think it's a good example for cases that could be more complex:

Suppose we have multiple operators trying to update provision state from "provisionable" to "not provisionable" at the same time. If both get past the `if old_state == state` check before getting to the `UPDATE`, then they'll both update the sled state (to the same value!), and both see the `old_state` return value of "provisionable", even though (according to a serialized view of the world) "one of them actually made the sled "not provisionable".

(I don't actually think this is that bad, but it *does* raise the question of "what does that return value mean? **should** we actually be exposing that information? How much do we care about providing genuine "compare-and-swap" semantics here?)

This is one of the reasons why it's kinda nice to perform as many checks as we'd like within "purely SQL" -- the single statement is an implicit transaction, so we avoid races like this by doing this kinda work in memory.

- If we add a `dsl::sled.filter(dsl::provision_state.ne(state))` clause to the `UPDATE`, then we'll only perform this update if we actually have work to do, by extending the `WHERE` clause. However, this has a downside -- if the sled **exists**, but is already in the desired state, it'll be filtered out, and we'll see "no rows updated".
- We can use use the `RETURNING` clause to access the DB value, but this only returns the **new** value, and not the old one.

Fortunately, there is a CTE that we have built to help queries do the following: https://github.com/oxidecomputer/omicron/blob/main/nexus/db-queries/src/db/update_and_check.rs

This lets you...
- Filter for the exact condition you want to check
- ... but if the old object exists, you can read it from the DB: https://github.com/oxidecomputer/omicron/blob/55b39533cfe9a3f2fc1185adaa9c2118efaee6bf/nexus/db-queries/src/db/update_and_check.rs#L118-L123

This basically lets you do an update **and** read the old value at the same time, if you want. Look for `check_if_exists` if you wanna find some example usage within the codebase.

That said -- we only really want this if we want to return the old state value! If we simply want to omit that from the return value, I think we could mostly ignore this machinery.

nexus/db-queries/src/db/queries/region_allocation.rs
```diff
@@ -310,13 +311,20 @@ impl CandidateZpools {
...
+            .inner_join(with_sled);
+
+        let sled_is_provisionable = sled_dsl::provision_state
+            .eq(crate::db::model::SledProvisionState::Provisionable);
 
         let base_query = old_zpool_usage
             .query_source()
             .inner_join(with_zpool)
             .filter(it_will_fit)
+            .filter(sled_is_provisionable)
```
smklein, line 327: Looks great, this 100% makes sense to me

nexus/tests/integration_tests/schema.rs
```diff
@@ -629,7 +629,14 @@ impl InformationSchema {
...
             other.referential_constraints
         );
-        similar_asserts::assert_eq!(self.statistics, other.statistics);
+        similar_asserts::assert_eq!(
+            self.statistics,
+            other.statistics,
+            "Statistics did not match. This often means that in dbinit.sql, a new \
+            column was added into the middle of a table rather than to the end. \
+            If that is the case, change dbinit.sql to add the column to the end \
+            of the table."
```
smklein, line 638: hell yeah thank you for making this more understandable

nexus/types/src/external_api/params.rs
```diff
@@ -75,6 +75,23 @@ pub struct SledSelector {
...
+}
+
+/// Response to `sled_set_provision_state`.
+#[derive(Clone, Debug, Serialize, Deserialize, JsonSchema, PartialEq)]
+pub struct SledProvisionStateResponse {
+    /// The old provision state.
+    pub old_state: super::views::SledProvisionState,
+
+    /// The new provision state.
+    pub new_state: super::views::SledProvisionState,
```
smklein, line 92: See my prior comments -- we *can* support this return value, but it means we'll need to be more careful with the database transactional semantics.

smklein (commented):
Looks great! I have a smattering of comments, but my big two are:

1. We gotta do a minor rename to make the schema change work
2. The transactional semantics of altering the `provisionable` state need some tweaking to make the return value consistent -- but I have a small essay discussing some of our options there

schema/crdb/13.0.0/up5.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TYPE IF NOT EXISTS omicron.public.sled_provision_state AS ENUM (
```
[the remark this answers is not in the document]
sunshowers: This is really weird but I swore that at the time I started writing this it was at version 13.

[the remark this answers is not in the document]
sunshowers: Oh, I see -- *each schema update* is its own version. Got it! (I thought that the versions were tied to releases or something, and that our schema updater was storing (version, counter) pairs.

nexus/db-queries/src/db/datastore/sled.rs
```diff
@@ -194,6 +201,14 @@ impl DataStore {
...
             .map_err(|e| match e {
                 TxnError::CustomError(SledReservationError::NotFound) => {
+                    // XXX: Should this be a different error? 503 Service
+                    // Unavailable is often for transient errors, and this
+                    // message isn't exposed to users (which seems wrong?)
+                    //
+                    // Maybe 409 Conflict as documented at
+                    // https://stackoverflow.com/a/11093566, since this request
+                    // if completed would put the system into an inconsistent
+                    // state.
```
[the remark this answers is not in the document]
sunshowers: I think my main remaining concern is that the error message doesn't make its way to clients -- for something like this it seems like it would be useful.

nexus/db-queries/src/db/datastore/sled.rs
sunshowers: Thank you, the CTE is exactly what I was looking for, I think!

sunshowers: All right, I updated the PR to use `update_and_check` but I'm seeing a mysterious error: https://gist.github.com/sunshowers/5e0829d0a29f09ae03f5782fc094206f#file-gistfile0-txt-L28

I verified that it is the `update_and_check` that's causing this by commenting it out and seeing these tests pass.

sunshowers: This was https://github.com/oxidecomputer/omicron/issues/4570.

nexus/db-queries/src/db/datastore/sled.rs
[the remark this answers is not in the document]
sunshowers: Going to turn this into a separate discussion.

nexus/db-queries/src/db/datastore/sled.rs
sunshowers: Addressed using `update_and_check`, works great.

schema/crdb/dbinit.sql
```diff
@@ -213,7 +223,7 @@ CREATE TABLE IF NOT EXISTS omicron.public.switch (
     time_created TIMESTAMPTZ NOT NULL,
     time_modified TIMESTAMPTZ NOT NULL,
     time_deleted TIMESTAMPTZ,
-    rcgen INT NOT NULL,
+    rcgen INT8 NOT NULL,
```
smklein, line 226: At the tail end of this file, there is a `14.0.0` that needs to be updated to `15.0.0` for the tests to pass

nexus/db-queries/src/db/datastore/sled.rs
```diff
@@ -217,6 +223,40 @@ impl DataStore {
...
+            .filter(dsl::time_deleted.is_null())
+            .filter(dsl::id.eq(sled_id))
+            .filter(dsl::provision_state.ne(state))
+            .set((
+                dsl::provision_state.eq(state),
+                dsl::time_modified.eq(Utc::now()),
+            ))
+            .check_if_exists::<Sled>(sled_id);
+
+        println!("{}", diesel::debug_query::<diesel::pg::Pg, _>(&query));
```
smklein, line 251: Let's remove this before merging

smklein (approved):
Looks good to me -- thanks for the fixes!

nexus/db-queries/src/db/datastore/sled.rs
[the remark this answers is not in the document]
sunshowers: https://github.com/oxidecomputer/omicron/pull/4573

schema/crdb/dbinit.sql
sunshowers: Oh I should revert the `rcgen` change, it's not relevant (though maybe we do want to do that as a blanket change)

sunshowers: Done.

openapi/nexus.json
```diff
@@ -13099,6 +13157,75 @@
...
+              "non_provisionable"
+            ]
+          },
+          {
+            "description": "This is a state that isn't known yet.\n\nThis is defined to avoid API breakage.",
+            "type": "string",
+            "enum": [
+              "unknown"
+            ]
+          }
```
ahl, line 13183: Is this how you intended this state to appear in the OpenAPI doc? I'm not clear on how this will prevent API breakage.

sunshowers: Addressed in https://github.com/oxidecomputer/omicron/pull/4608

PR #4528, opened 2023-11-20
Fix #4509 and #4512

### LINKED ISSUE #4509: Code reuse between switch port settings apply/clear sagas results in serialization error
When we call this code from the port settings **clear** saga

https://github.com/oxidecomputer/omicron/blob/ca9d90a79c2a7a9ffdca12a705d007d7f28ce61f/nexus/src/app/sagas/switch_port_settings_clear.rs#L154

We land here in the **apply** saga code

https://github.com/oxidecomputer/omicron/blob/ca9d90a79c2a7a9ffdca12a705d007d7f28ce61f/nexus/src/app/sagas/switch_port_settings_apply.rs#L728-L737

Line 732 is problematic there as the `Params` type is different in the **clear** and **apply** sagas and we wind up with this puddle of disappointment.

```
02:51:25.959Z INFO 65a11c18-7f59-41ac-b9e7-680627f996e7 (dropshot_external): request completed
    error_message_internal = saga ACTION error at node "switch_port_settings": deserialize failed: saga params ({"port_name":"qsfp0","serialized_authn":{"kind":{"Authenticated":[{"actor":{"SiloUser":{"silo_id":"bb5700e7-c5f7-4c45-a1bc-bd610c03c55b","silo_user_id":"5d3758e4-2360-4386-99cc-d22bef03e827"}}},{"mapped_fleet_roles":{"admin":["admin"]}}]}},"switch_port_id":"348b7e16-d806-4edc-b4ac-ba48014c14b5"}): missing field `switch_port_settings_id`
    file = /home/build/.cargo/git/checkouts/dropshot-a4a923d29dccc492/ff87a01/dropshot/src/server.rs:841
    latency_us = 189133
    local_addr = 172.30.2.5:443
    method = DELETE
    remote_addr = 172.20.16.254:62131
    req_id = d86a8413-78aa-4ad1-b5fa-263386b7dbf7
    response_code = 500
    uri = //v1/system/hardware/switch-port/qsfp0/settings?rack_id=de608e01-b8e4-4d93-b972-a7dbed36dd22&switch_location=switch1
```

### LINKED ISSUE #4512: Boundary switches database query only considers static routes.
From below, we can see that a switch is only considered a boundary switch if it has static routes.

https://github.com/oxidecomputer/omicron/blob/781ed1236489bb536506fb03ba1be39b34ed81a5/nexus/db-queries/src/db/datastore/switch_port.rs#L1107-L1135

This falls apart if the rack only has BGP routes. This function should consider the presence of BGP configuration in addition to static routes. In the limit, this general approach should give way to the mechanisms described in RFD 404.

rcgoodfellow:
- Fixes #4509 
- Fixes #4512

Most of the diff in this PR is just moving code around. To fix #4509 I factored out a common file for both port settings sagas to use, making sure saga-specific elements do not leak into that file.

Both fixes come with regression tests.

### DIFF
--- a/nexus/db-model/src/schema.rs
+++ b/nexus/db-model/src/schema.rs
@@ -1328,3 +1328,8 @@ allow_tables_to_appear_in_same_query!(
     switch_port,
     switch_port_settings_route_config
 );
+
+allow_tables_to_appear_in_same_query!(
+    switch_port,
+    switch_port_settings_bgp_peer_config
+);
--- a/nexus/db-queries/src/db/datastore/switch_port.rs
+++ b/nexus/db-queries/src/db/datastore/switch_port.rs
@@ -23,8 +23,8 @@ use crate::db::pagination::paginated;
 use async_bb8_diesel::{AsyncConnection, AsyncRunQueryDsl};
 use diesel::result::Error as DieselError;
 use diesel::{
-    ExpressionMethods, JoinOnDsl, NullableExpressionMethods, QueryDsl,
-    SelectableHelper,
+    CombineDsl, ExpressionMethods, JoinOnDsl, NullableExpressionMethods,
+    QueryDsl, SelectableHelper,
 };
 use nexus_types::external_api::params;
 use omicron_common::api::external::http_pagination::PaginatedBy;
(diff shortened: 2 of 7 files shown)


jmpesp (approved):
:rocket:

Issue #4392, opened 2023-10-30
deps unexpectedly rebuilt after running `cargo nextest run` differently


davepacheco:
I noticed all this working on branch "dap/nexus-inventory-2".  Commit 376a37bb21d8bab98a2c4cd8747fc3a5ee3140d3 is sync'd up with "main" from this morning.  From there, I ran a full test suite run that failed one test:

```
$ ptime -m cargo nextest run --no-fail-fast
info: experimental features enabled: setup-scripts
    Finished test [unoptimized + debuginfo] target(s) in 1.08s
    Starting 892 tests across 99 binaries (2 skipped)
       SETUP [      1/1] crdb-seed: cargo run -p crdb-seed
             [ 00:00:00] [                                                                                            ]   0/894:          Finished dev [unoptimized + debuginfo] target(s) in 0.53s
     Running `target/debug/crdb-seed`
Oct 30 20:25:03.262 INFO Using existing CRDB seed tarball: `/dangerzone/omicron_tmp/crdb-base-dap/3af8e1b0d6d99d1eb564ed4f736ee088e760afd006b09b3fb54a50ef98d608b1.tar`
  SETUP PASS [      1/1] crdb-seed: cargo run -p crdb-seed
        PASS [   0.083s] authz-macros::proc-macro/authz-macros test_authz_dump
...
------------
     Summary [ 403.723s] 892 tests run: 891 passed (3 slow, 1 leaky), 1 failed, 2 skipped
        FAIL [  18.704s] omicron-nexus::test_all integration_tests::schema::dbinit_equals_sum_of_all_up
error: test run failed

real     6:47.092637560
user  1:19:44.454769533
sys     14:08.574488939
trap       41.302586601
tflt        0.500335124
dflt        3.788205567
kflt        0.048879750
lock 365:06:39.566636121
slp  16:25:43.533571797
lat   1:03:17.653685989
stop    26:13.939275810
```

I then tried to fix this test, which changed these files: `nexus/db-model/src/schema.rs`, `schema/crdb/dbinit.sql`, `schema/crdb/9.0.0/` (a whole tree).

Then I tried to rerun just the failing test.  But it looks like a whole lot of stuff got rebuilt, including a bunch of deps from outside the workspace:

```
$ ptime -m cargo nextest run -p omicron-nexus integration_tests::schema::dbinit_equals_sum_of_all_up
info: experimental features enabled: setup-scripts
   Compiling serde_json v1.0.107
[...]

PR #4535, opened 2023-11-21
[meta] a few changes to prevent duplicate dep builds

sunshowers:
This PR has a few changes that make builds and test runs significantly faster:

1. Remove `xtask` from the list of default-members. This makes it so that `cargo nextest run` and `cargo nextest run -p <package>` use more dependency feature sets in common.
2. Move `opt-level` settings from `profile.test` to `profile.dev`. Again, this results in more cache hits.
3. Set `profile.dev.panic` to `unwind`. This is to unify build units across dev and test builds: tests are always built with `panic = "unwind"` so that proper backtraces can be printed out. Release builds stay as `abort`.
4. For a belt-and-suspenders approach, make the `crdb-seed` script use the `test` profile. If there are any divergences between `dev` and `test` in the future, then crdb-seed should share its build cache with the tests it was presumably invoked for.
5. Set `profile.dev.build-override.debug` to `line-tables-only`. This, along with 3, means that target (normal/dev) and build (host) dependencies are now unified.

All of this comes together for a pretty sweet improvement.

See #4392 for more details and how I investigated this issue.

## Impact

With a fresh build on Linux with mold, I ran three commands in sequence:

1. `cargo nextest run --no-run`
2. `cargo nextest run -p nexus-db-queries`
3. `cargo build -p omicron-nexus`

The results were:

|               **command**               | **phase**         | **before** | **before, cumul.** | **after** | **after, cumul.** |
|-----------------------------------------|-------------------|-----------:|-------------------:|----------:|------------------:|
| `cargo nextest run`                     | build             |       173s |               173s |      158s |              158s |
| `cargo nextest run -p nexus-db-queries` | build             |        61s |               234s |       51s |              209s |
| `cargo nextest run -p nexus-db-queries` | `crdb-seed` build |        21s |               255s |        1s |              210s |
| `cargo build -p omicron-nexus`          | build             |        99s |               354s |       69s |              279s |

So the cumulative time spent on these three commands went from 354s to 279s. That's a 1.26x speedup. And this should also make other commands better as well (omicron-nexus is a bit of a weird case because it takes a very long time to compile by itself, and that 69s in the "after" column is entirely building omicron-nexus).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -393,6 +393,11 @@ zeroize = { version = "1.6.0", features = ["zeroize_derive", "std"] }
 zip = { version = "0.6.6", default-features = false, features = ["deflate","bzip2"] }
 zone = { version = "0.3", default-features = false, features = ["async"] }
 
+# NOTE: The test profile inherits from the dev profile, so settings under
+# profile.dev get inherited. AVOID setting anything under profile.test: that
+# will cause dev and test builds to diverge, which will cause more Cargo build
+# cache misses.
+
 [profile.dev]
 # Note: This used to be panic = "abort" earlier, but that caused a lot of
 # duplicate dependency builds. Letting panic be "unwind" causes dependencies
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -391,13 +393,27 @@ zeroize = { version = "1.6.0", features = ["zeroize_derive", "std"] }
 zip = { version = "0.6.6", default-features = false, features = ["deflate","bzip2"] }
 zone = { version = "0.3", default-features = false, features = ["async"] }
 
+# NOTE: The test profile inherits from the dev profile, so settings under
+# profile.dev get inherited. AVOID setting anything under profile.test: that
+# will cause dev and test builds to diverge, which will cause more Cargo build
+# cache misses.
```
smklein, line 399: Good catch, and AFAICT, that's a large part of the rest of this PR, correct? Ensuring that "build for dev" and "build for test" are aligned, so they only build once?

Cargo.toml
```diff
@@ -391,13 +393,27 @@ zeroize = { version = "1.6.0", features = ["zeroize_derive", "std"] }
...
+
 # See https://github.com/oxidecomputer/omicron/issues/4009 for some background context here.
 # By reducing the debug level (though keeping enough to have meaningful
 # backtraces), we reduce incremental build time and binary size significantly.
 debug = "line-tables-only"
 
+[profile.dev.build-override]
+# Setting this to line-tables-only results in a large improvement in build
+# times, because it allows target and host dependencies to be unified.
+debug = "line-tables-only"
```
smklein, line 415: I understand why we want to set `line-tables-only` here, and I understand that we want it set for development builds. I also understand why we'd want to align development and test builds (e.g. with `panic = 'unwind'"` above).

Should we *also* be aligning `profile.dev` and `profile.dev.build-override` across other settings?

Should we, for example be setting `panic = "unwind"` for `build-override`?

smklein (approved):
Thanks for doing this investigation and adjustment -- this is a significant boost, happy to see us re-use portions of the build where we can / should!

rcgoodfellow:
Unfortunately, this does not appear to move the needle on illumos.

| command | before full-compile | after full-compile | before incremental-recompile | after incremental recompile |
|---|---|---|---|---|
| `cargo nextest run --no-run` | 986s | 985s | 646s | 664s |

Incremental compile time measured by adding a comment in the file `neuxs/src/app/sagas/switch_port_settings_apply.rs` and re-running `cargo nextest run --no-run`

### More details

#### System
- AMD Ryzen Threadripper 3990x
- 256GB Memory
- Samsung 970 Evo NVMe 2Tb
- helios-2.0.22298

#### Before full-compile

```
real    16:25.599274380
user  1:59:36.518864609
sys     19:15.753319431
```
#### Before incremental recompile

```
real    10:45.830306626
user    29:00.931336328
sys      1:31.905661961
```
#### After full-compile

```
real    16:25.070892103
user  1:48:28.561607858
sys     13:22.269607043
```
#### After incremental recompile

```
real    11:03.987974145
user    40:35.817986189
sys      1:21.120225119
```

Cargo.toml
sunshowers: Yes, the biggest chunk by loc is removing divergence between dev and test builds.

Cargo.toml
sunshowers: By default, [build-override for dev builds only sets](https://doc.rust-lang.org/cargo/reference/profiles.html#build-dependencies):

1. `opt-level = 0`. This is the same as the default opt-level for dev builds. We override the opt-level for some packages, but based on my testing it looks like those overrides are also carried through to build deps. If we ever set the global opt-level to something other than 0, however, we'll want to also set it for build-override.
* `codegen-units = 256`. In the unit graph the value appears to always be `null` so this doesn't appear to make a difference.
* `debug = false`. We're changing this setting as part of this PR.

sunshowers:
Discussed this on #oxide-control-plane in Matrix -- unfortunately the listed case would not benefit (nor be affected negatively) by this change. Where an improvement would occur is with e.g. `cargo nextest run` followed by `cargo nextest run -p omicron-sled-agent`.

PR #4529, opened 2023-11-20
Restore schema 10.0.0

jgallagher:
This was accidentally merged in with 11.0.0 in #3804

Fixes #4530 (but see #4531 for followup work)

PR #4546, opened 2023-11-21
Reject bad combinations of `up*.sql` migrations

### LINKED ISSUE #4531: Schema migrations: detect and reject conflicting schema names
One symptom of https://github.com/oxidecomputer/omicron/issues/4530 was that we had a migration directory containing both `up1.sql` and `up01.sql` (and additional higher numbers), caused by incorrect git merge / move operations while trying to land two unrelated PRs that both added new migrations. We should detect this, ideally at CI test time, to at least catch "this directory seems to contain two different sets of migrations".

jgallagher:
Fixes #4531.

If I cherry-pick these changes onto `main` just prior to #4529 (i.e., when the `11.0.0` directory contained upsql files for both 10.0.0 and 11.0.0), the `dbinit_equals_sum_of_all_up` integration test fails as desired:

```
thread 'integration_tests::schema::dbinit_equals_sum_of_all_up' panicked at nexus/tests/integration_tests/schema.rs:133:58:
called `Result::unwrap()` on an `Err` value: "invalid `up*.sql` combination: /data/github/omicron/nexus/../schema/crdb/11.0.0/up01.sql, /data/github/omicron/nexus/../schema/crdb/11.0.0/up1.sql"
```


david-crespo (approved):

nexus/db-queries/src/db/datastore/db_metadata.rs
```diff
@@ -54,23 +63,83 @@ pub async fn all_sql_for_version_migration<P: AsRef<Utf8Path>>(
...
+                if up_number == previous_up_number + 1 {
+                    prev_up_number = Some((path, up_number));
+                } else {
+                    return Err(format!(
+                        "invalid `up*.sql` combination: \
+                         {previous_path}, {path}"
+                    ));
+                }
+            }
+        }
```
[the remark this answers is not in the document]
jgallagher: haha, thanks? Someone would've had to do it eventually; too easy to mess up across branches!

david-crespo:
A simpler but less efficient way of checking the list of numbers could be (this is pseudocode, forgive me)

```
let migration_nums = filenames.map(extract_num)
let n = migration_nums.len()
// error if migration_nums is not either [0] or [1..n]
```

nexus/db-queries/src/db/datastore/db_metadata.rs
```diff
@@ -26,23 +26,32 @@ use std::str::FromStr;
...
-/// part of the migration, and fully read to a string.
+/// Files that do not begin with "up" and end with ".sql" are ignored. The
+/// collection of `up*.sql` files must fall into one of these two conventions:
 ///
-/// These are sorted lexicographically.
+/// * "up.sql" with no other files
+/// * "up1.sql", "up2.sql", ..., beginning from 1, optionally with leading
+///   zeroes (e.g., "up01.sql", "up02.sql", ...). There is no maximum value, but
+///   there may not be any gaps (e.g., if "up2.sql" and "up4.sql" exist, so must
+///   "up3.sql") and there must not be any repeats (e.g., if "up1.sql" exists,
```
smklein, line 51: I think this comment got cut off

smklein (approved):
Thank you for adding this, and thank you more for the tests!

nexus/db-queries/src/db/datastore/db_metadata.rs
```diff
@@ -499,8 +679,9 @@ mod test {
             .await;
 
         // Show that the datastores can be created concurrently.
-        let config =
-            SchemaConfig { schema_dir: config_dir.path().to_path_buf() };
+        let config = SchemaConfig {
+            schema_dir: config_dir.into_path().into_std_path_buf(),
```
smklein, line 683: I think this is failing CI: https://docs.rs/camino-tempfile/1.0.2/camino_tempfile/struct.Utf8TempDir.html#method.into_path

> This consumes the Utf8TempDir without deleting directory on the filesystem, meaning that the directory will no longer be automatically deleted.

jgallagher: Whoops! Thanks for pointing this out.

jgallagher:
> An easier to understand but probably less efficient way of checking the list of numbers could be (this is pseudocode, forgive me)
> 
> ```
> let migration_nums = filenames.map(extract_num)
> let n = migration_nums.len()
> // error if migration_nums is not either [0] or [1..n]
> ```

I like that, but wanted to keep the fairly nice error messages. I took a stab at the best of both worlds in https://github.com/oxidecomputer/omicron/pull/4546/commits/2503da5e2a1fbd40f2b5d586c1ef2fc1b5871d7d, which had a couple unexpected benefits:

1. I got to use the "new" slice matching stuff for the first time
2. I realized we weren't handling a directory with no `up*.sql`s at all, which is now an error

PR #4547, opened 2023-11-21
Detect and un-delete phantom disks

jmpesp:
A "phantom" disk is a disk where the disk delete saga has run for it but unwound: this leaves that disk soft-deleted, but the resources and accounting for that disk remain. Users cannot request that the disk be deleted again, and it remains a phantom.

There are two fixes for this:

1. Change the disk delete saga to undo the disk's soft delete and set the disk to faulted during an unwind. This way, users can request that disks be deleted repeatedly until it works.

2. Create a background task that detects these phantom disks and does the same thing: un-delete them and set them to faulted.

This requires adding an index on `id` to the `disk` table, so the schema is bumped to 12.0.1.

Fixes oxidecomputer/customer-support#58.


schema/crdb/12.0.1/up01.sql
```diff
@@ -0,0 +1,4 @@
+CREATE UNIQUE INDEX IF NOT EXISTS lookup_deleted_disk ON omicron.public.disk (
```
jgallagher, line 1: With my apologies: I just merged a `13.0.0` schema migration, so I think this will have to become `14.0.0` (or maybe even `15.0.0` if #4545 lands first)

david-crespo: This is a very annoying problem. @davepacheco’s idea of named migrations and a canonical order list is sounding pretty good to me!

jmpesp: I'll wait for #4545, that's a higher priority.

dev-tools/omdb/src/bin/omdb/nexus.rs
```diff
@@ -515,6 +515,22 @@ fn print_task_details(bgtask: &BackgroundTask, details: &serde_json::Value) {
                 );
             }
         };
+    } else if name == "phantom_disks" {
+        #[derive(Deserialize)]
+        struct TaskSuccess {
+            /// how many phantom disks were found
+            ok: usize,
```
bnaecker, line 522: A more accurate name could be helpful, maybe `n_disks` or `disk_count`?

dev-tools/omdb/src/bin/omdb/nexus.rs
```diff
@@ -515,6 +515,22 @@ fn print_task_details(bgtask: &BackgroundTask, details: &serde_json::Value) {
...
+        }
+
+        match serde_json::from_value::<TaskSuccess>(details.clone()) {
+            Err(error) => eprintln!(
+                "warning: failed to interpret task details: {:?}: {:?}",
+                error, details
+            ),
+            Ok(success) => {
+                println!("    phantom disks found: {}", success.ok);
```
bnaecker, line 531: Maybe `"number of phantom disks found: {}"` to make clear it's a count.

nexus/db-model/src/schema.rs
```diff
@@ -1299,7 +1299,7 @@ table! {
 ///
 /// This should be updated whenever the schema is changed. For more details,
 /// refer to: schema/crdb/README.adoc
-pub const SCHEMA_VERSION: SemverVersion = SemverVersion::new(14, 0, 0);
+pub const SCHEMA_VERSION: SemverVersion = SemverVersion::new(15, 0, 0);
```
bnaecker: I'm sure you're aware, but we're now targeting v18 I think? Definitely past 15.

nexus/db-queries/src/db/datastore/disk.rs
```diff
@@ -652,4 +655,143 @@ impl DataStore {
...
+        disk_id: &Uuid,
+    ) -> Result<(), Error> {
+        use db::schema::disk::dsl;
+        let conn = self.pool_connection_unauthorized().await?;
+
+        let faulted = api::external::DiskState::Faulted.label();
+
+        let result = diesel::update(dsl::disk)
+            .filter(dsl::time_deleted.is_not_null())
+            .filter(dsl::id.eq(*disk_id))
```
bnaecker, line 676: This is the primary key, right? I think `diesel::update(dsl::disk).find(*disk_id)` might be more clear here, unless I'm missing something about the DSL that precludes that.

bnaecker: Ah, I think you might have included the filter on `time_deleted IS NOT NULL` to ensure you're only operating on deleted disks, for the benefit of the `execute_and_check()` mechanism. Ignore the previous suggestion if that's the case!

nexus/db-queries/src/db/datastore/disk.rs
```diff
@@ -652,4 +655,143 @@ impl DataStore {
...
+        // 1. the disk and volume are soft-deleted
+        // 2. the `virtual_provisioning_resource` exists
+        //
+        // It's important not to conflict with any currently running disk delete
+        // saga.
+
+        Ok(potential_phantom_disks
+            .into_iter()
+            .filter(|(disk, resource, volume)| {
+                if let Some(volume) = volume {
```
bnaecker, line 772: It looks like splitting out these conditions is useful from a documentation and clarity perspective, but they _could_ be collapsed to:

```rust
disk.time_deleted().is_some()
    && resource.is_some() 
    && volume.map(|vol| vol.time_deleted().is_some()).unwrap_or(true)
```

nexus/db-queries/src/db/datastore/disk.rs
```diff
@@ -652,4 +655,143 @@ impl DataStore {
...
+                    // In this branch, the volume record was soft-deleted. The
```
bnaecker, line 773: I get that the volume record could not have been _hard_ deleted, but how do we know it _is_ soft-deleted? Could it be neither of those?

nexus/src/app/background/phantom_disks.rs
```diff
@@ -0,0 +1,84 @@
...
+//! Background task for detecting and un-deleting phantom disks
+//!
+//! A "phantom" disk is one where a disk delete saga partially completed but
+//! unwound: before a fix for customer-support#58, this would leave disks
+//! deleted but would also leave a `virtual_provisioning_resource` record for
+//! that disk. There would be no way to re-trigger the disk delete saga as the
+//! disk was deleted, so the project that disk was in could not be deleted.
+//!
+//! This background task detects that case and sets the disk to "faulted". The
+//! fix for custoemr-support#58 is similar: it changes the disk delete saga's
```
bnaecker, line 14: Spelling: custoemer -> customer

nexus/src/app/background/phantom_disks.rs
```diff
@@ -0,0 +1,84 @@
...
+//! unwind to also un-delete the disk and set it to faulted. This enables it to
```
bnaecker, line 15: I could be misreading this, but this seems to imply that the changes to the disk-delete saga and this task have the same goal: Take a disk that could not be completely deleted; undelete it; set it to faulted; somehow notify the user that they can try again. Is that accurate? What I'm getting at is it seems that this task should not be needed if the saga is also doing this set of steps. It might be helpful to lay out why we need both of them.

nexus/src/app/background/phantom_disks.rs
```diff
@@ -0,0 +1,84 @@
...
+
+                if let Err(e) = result {
+                    error!(
+                        &log,
+                        "error undeleting disk and setting to faulted: {:#}", e
+                    );
+                }
+            }
+
+            json!({ "ok": rv })
```
bnaecker, line 80: It seems like we may want to return the number of successfully and _unsuccessfully_ deleted phantom disks here. Would that be helpful?

nexus/tests/integration_tests/disks.rs
```diff
@@ -1241,6 +1241,138 @@ async fn test_disk_virtual_provisioning_collection(
...
+
+    // Request disk delete again
+    NexusRequest::new(
+        RequestBuilder::new(client, Method::DELETE, &disk_url)
+            .expect_status(Some(StatusCode::NO_CONTENT)),
+    )
+    .authn_as(AuthnMode::PrivilegedUser)
+    .execute()
+    .await
+    .expect("unexpected failure deleting 1 GiB disk");
```
bnaecker, line 1346: Is there a way we can exercise the idempotency-path in nexus/db-queries/src/db/datastore/disk.rs at L700-ish?

schema/crdb/18.0.0/up01.sql
```diff
@@ -0,0 +1,4 @@
+CREATE UNIQUE INDEX IF NOT EXISTS lookup_deleted_disk ON omicron.public.disk (
```
bnaecker, line 1: I'm a bit surprised by this index. The `id` is the primary key, so do we need an additional index in order to look things up that are deleted?

bnaecker (commented):
Nice work, this looks like a solid improvement! I've a few questions around clarity and naming and such, but the overall structure looks very reasonable. Thanks!

schema/crdb/18.0.0/up01.sql
bnaecker: Interesting! I think I see what's causing the full scan. The meat of the query selects from the `disk` table without any filter on `id`, and then uses that in the join expression. That's the last two nodes in the first explain output, which are the inputs to the merge join node. That makes sense to me, since the only filter we're applying is the `time_deleted IS NOT NULL`. 

If it's true that we don't have any more filtering criteria, like the disk primary key or something, then I think we _do_ need an index. I think indices always [store the primary key implicitly](https://www.cockroachlabs.com/docs/v23.1/schema-design-indexes#example), so I"m guessing the two options are equivalent or nearly so.

I have one niggling thought though. The first explain output says "missing stats". Can we try to generate them, and see if the explain output changes? I think [`CREATE STATISTICS`](https://www.cockroachlabs.com/docs/v23.1/create-statistics) will do that, but I'm not positive.

bnaecker: I think that seems plausible, I'm just wondering if the presence of the statistics would change the behavior of the query planner. It's not a huge deal, I think the index is pretty low cost and seems necessary to run the query we want in a scalable way.

nexus/db-queries/src/db/datastore/disk.rs
jmpesp: :+1:

nexus/db-queries/src/db/datastore/disk.rs
jmpesp: They could, yeah, but I prefer this form with the documentation :)

dev-tools/omdb/src/bin/omdb/nexus.rs
jmpesp: agreed, this was me blindly copying from the background task output. done in e64ec0850

dev-tools/omdb/src/bin/omdb/nexus.rs
jmpesp: also done in e64ec0850

nexus/src/app/background/phantom_disks.rs
jmpesp: 615679aed

nexus/src/app/background/phantom_disks.rs
jmpesp: You're not wrong! The background task is there to catch previous cases where the saga unwound and caused these problems. Added this bit in 6e0827c13

nexus/src/app/background/phantom_disks.rs
jmpesp: yeah, that would: done in 1c4b61e0ff017909d4b657a207cf47e4126418a4

nexus/db-queries/src/db/datastore/disk.rs
bnaecker: Sweet, that comment is very helpful.

bnaecker (approved):
Thanks for the back-and-forth, this looks good to me!

askfongjojo:
@jmpesp, sorry this is a late-breaking question - what happens if some of the undeleted disks in the same project have the same name? Would they be renamed to honor the unique-name constraint?

jmpesp:
@askfongjojo I had not considered that! As written in this PR the background task will not rename the disk, but I believe that the un-delete would fail due to the existing `lookup_disk_by_project` UNIQUE INDEX:

```sql
CREATE UNIQUE INDEX IF NOT EXISTS lookup_disk_by_project ON omicron.public.disk (
    project_id,
    name
) WHERE
    time_deleted IS NULL;
```

This would mean that a disk with the same name would shadow the phantom disk, consuming virtual resources and counting towards quotas. I'll open an issue to track fixing this.

PR #4548, opened 2023-11-21
[nexus] Add `HostPhase1Updater`

jgallagher:
Continues the work from #4427 and #4502 adding types to Nexus that can deliver updates to components managed by the SP, and contains a similar amount of duplication from those two. Non-duplication changes are mostly refactoring:

* Removed component-specific error types in favor of `SpComponentUpdateError`
* Extracted the "start the update and poll MGS until it's done" logic into a new `common_sp_update` module

The `test_host_phase1_updater_delivers_progress` is subtly different from the RoT/SP versions of the same test, but not in a way that's particularly interesting (see the "Unlike the SP and RoT cases" comment for details).

### DIFF
--- a/nexus/src/app/test_interfaces.rs
+++ b/nexus/src/app/test_interfaces.rs
@@ -10,6 +10,7 @@ use sled_agent_client::Client as SledAgentClient;
 use std::sync::Arc;
 use uuid::Uuid;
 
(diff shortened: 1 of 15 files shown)


davepacheco (approved):
I like the commonization here!

PR #4557, opened 2023-11-23
[omicron-dev] increase test timeout to 30 seconds

sunshowers:
On my machine (Ryzen 7950X) I saw that under load (32 tests running at the same
time), the timeout would quite reliably be hit likely because cockroach was
starved. Increasing it seems pretty harmless.

### DIFF
--- a/dev-tools/omicron-dev/tests/test_omicron_dev.rs
+++ b/dev-tools/omicron-dev/tests/test_omicron_dev.rs
@@ -27,7 +27,7 @@ use subprocess::Redirection;
 const CMD_OMICRON_DEV: &str = env!("CARGO_BIN_EXE_omicron-dev");
 
 /// timeout used for various things that should be pretty quick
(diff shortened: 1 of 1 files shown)


davepacheco (approved):
It looks like this timeout only covers the period between when we send SIGINT to `omicron-dev db-run` and when `wait(2)` (or some other `wait` syscall) returns an exit status.  15 seconds seems pretty long to wait for that?  I don't think there's much harm in bumping this up -- I'm just worried something deeper is wrong.

sunshowers:
Yeah, `omicron-dev db-run` has [its own signal handler](https://github.com/oxidecomputer/omicron/blob/36f6abe06548eae0f954f1b968f98def846f7d93/dev-tools/omicron-dev/src/bin/omicron-dev.rs#L184-L201) for `SIGINT` that waits for the database to shut down. That's what takes more than 15 seconds.

PR #4559, opened 2023-11-24
(Multiple) Floating IP Support

### LINKED ISSUE #1334: Implement Floating IPs

### LINKED ISSUE #1467: Handle multiple external IP addresses per instance
We're adding support for Ephemeral IP addresses in https://github.com/oxidecomputer/omicron/pull/1458, but that's currently limited to one IP address. That's mostly because we're currently abusing the source NAT configuration in OPTE meant for _outbound_ connectivity, as a workaround to implement inbound connectivity. Also, OPTE doesn't have a way to express the IP address(es) used for 1-1 NAT in its API yet. (No issue, add link when that's created.)

This tracks relaxing the limit of 1 external IP address per instance, and passing all those to OPTE. See [here](https://github.com/oxidecomputer/omicron/blob/a6af80cc8cc2e28c90d6fc174dd9fa70ed6a2036/nexus/src/app/instance.rs#L500) for the related source, and [here](https://github.com/oxidecomputer/omicron/pull/1458#discussion_r924698039) for related discussion.

FelixMcFelix:
This PR adds external API endpoints to create and manipulate floating IP objects within a project, and ensures they will be attached/detached during the instance create/destroy sagas. External IPs served to instances are now partitioned into ephemeral and floating when talking to sled-agent, as OPTE will prefer floating IP addresses for outbound traffic. The new endpoints are:
* POST     `/v1/floating-ips`
* DELETE   `/v1/floating-ips/{floating_ip}`
* GET      `/v1/floating-ips`
* GET      `/v1/floating-ips/{floating_ip}`

Currently, this only allows floating IPs to be attached/detached during instance create. Separate per-instance attach/detach endpoints for live instances will follow in another PR, since I suspect there's extra saga work needing done to get it right.

This is closely tied with [opte#420](https://github.com/oxidecomputer/opte/pull/420) -- the semantics of having external IPs of either/both type are described on that PR. `helios / deploy` will fail until that is merged and we can create a new helios + OPTE image.

Will close #1467 and close #1334.


nexus/types/src/external_api/views.rs
```diff
@@ -265,6 +265,22 @@ pub struct ExternalIp {
     pub kind: IpKind,
```
FelixMcFelix, line 265: Re: `views::ExternalIp`, I don't know if we want to change this to an enum (rather than rely on IpKind) to directly hand over a `views::FloatingIp` instead of needing users to (painfully) look it up again. The console works with this as-is, so as long as `ip` is flattened out then we can make this change.

rcgoodfellow:
It would be good to have docs for this in the form of a [guide](https://docs.oxide.computer/guides/introduction#_overview) on managing floating IPs.

rcgoodfellow:
Since there are new endpoints, we also need updates to [oxide.rs](https://github.com/oxidecomputer/oxide.rs) to add those endpoints.

rcgoodfellow:
I deployed this branch and was able to launch an instance with a floating IP and an ephemeral IP :tada:

The following worked for me.

```
$ oxide api /v1/floating-ips?project=murphy --method POST --input floater.json
```
```
$ cat floater.json
{
    "name": "floater",
    "description": "a floater",
    "address": "192.168.20.47"
}
```
```
$ oxide api /v1/instances?project=murphy --method POST --input - <<EOF
{
  "name": "myinst",
  "description": "my inst",
  "hostname": "myinst",
  "memory": 1073741824,
  "ncpus": 2,
  "disks": [
    {
      "type": "attach",
      "name": "bookworm-boot"
    }
  ],
  "external_ips": [{"type": "ephemeral"}, {"type": "floating", "floating_ip_name": "floater"}]
}
EOF
```

I had several missteps along the way, but the errors coming back from the API were very helpful and the corrections needed were clear :+1:.

Upon launching the instance, I was able to ssh in via both the floating and ephemeral IP.

I was also able to verify that OPTE was preferring the floating IP over the ephemeral on egress for ICMP, UDP and TCP traffic.

Basic kicking of the tires all looks good here. Great work, I know this is a lot of integration work in addition to the underlying OPTE bits.

nexus/db-queries/src/db/datastore/external_ip.rs
```diff
@@ -281,4 +376,167 @@ impl DataStore {
...
+    pub async fn floating_ip_delete(
+        &self,
+        opctx: &OpContext,
+        authz_fip: &authz::FloatingIp,
+        db_fip: &FloatingIp,
+    ) -> DeleteResult {
+        use db::schema::external_ip::dsl;
+
+        // Verify this FIP is not attached to any instances/services.
+        if db_fip.parent_id.is_some() {
```
rcgoodfellow, line 419: Should getting the `db_fip` value and acting on the value for the detach action below be done in a transaction to ensure that we can't get a race condition between attaching and deleting?

nexus/db-queries/src/db/datastore/external_ip.rs
```diff
@@ -281,4 +376,167 @@ impl DataStore {
...
+        &self,
+        opctx: &OpContext,
+        authz_fip: &authz::FloatingIp,
+        db_fip: &FloatingIp,
+        instance_id: Uuid,
+    ) -> UpdateResult<FloatingIp> {
+        use db::schema::external_ip::dsl;
+
+        // Verify this FIP is not attached to any instances/services.
+        if db_fip.parent_id.is_some() {
```
rcgoodfellow, line 462: Same question as the delete case, does the parent id check need to be done in a transaction context with the attach to avoid races?

nexus/db-queries/src/db/datastore/external_ip.rs
```diff
@@ -281,4 +376,167 @@ impl DataStore {
...
+    /// Detaches a Floating IP address from an instance.
+    pub async fn floating_ip_detach(
+        &self,
+        opctx: &OpContext,
+        authz_fip: &authz::FloatingIp,
+        db_fip: &FloatingIp,
+    ) -> UpdateResult<FloatingIp> {
+        use db::schema::external_ip::dsl;
+
+        let Some(instance_id) = db_fip.parent_id else {
```
rcgoodfellow, line 507: Same question on needing to be in transaction for `db_fip` check.

nexus/src/app/mod.rs
```diff
@@ -79,8 +79,10 @@ pub(crate) use nexus_db_queries::db::queries::disk::MAX_DISKS_PER_INSTANCE;
 
 pub(crate) const MAX_NICS_PER_INSTANCE: usize = 8;
 
-// TODO-completeness: Support multiple external IPs
-pub(crate) const MAX_EXTERNAL_IPS_PER_INSTANCE: usize = 1;
+// XXX: Might want to recast as max *floating* IPs, we have at most one
+//      ephemeral (so bounded in saga by design).
+pub(crate) const MAX_EXTERNAL_IPS_PER_INSTANCE: usize = 32;
```
rcgoodfellow, line 84: Where does the upper bound of 32 come from?

nexus/db-queries/src/db/datastore/external_ip.rs
FelixMcFelix: I think it would be improved by doing so, yes. The actual `diesel::update` attach/detach/delete operations all include a filter on parent_id, so currently if preempted by another write they will back out without doing anything unintended, but I'd prefer a transaction too.

nexus/src/app/mod.rs
rcgoodfellow: Ok. This additional color should probably be captured in a comment here in case we need to come back and revisit this limit. We'll also want to note this limit in the user-facing docs.

rcgoodfellow (approved):
LGTM and testing checks out, thanks!

I think it would be good to have someone who understands the instance lifecycle much better than me take a look. Perhaps @gjcolombo?

nexus/db-queries/src/db/datastore/external_ip.rs
bnaecker: The [`update_and_check()`](https://github.com/oxidecomputer/omicron/blob/bcd7ac5d5e10355774310aa583aa59253eb880e5/nexus/db-queries/src/db/update_and_check.rs#L1) machinery is probably going to be helpful here. It lets one do conditional updates, say on the parent ID being NULL, and returning a detectable error in the case where that condition is not met.

rcgoodfellow:
In the interest of getting things lined up for the release, I'm merging this.

PR #4561, opened 2023-11-27
[update-engine] fix GroupDisplayStats to avoid integer underflow

### LINKED ISSUE #4507: [wicket] "Running" count wrong in rack-update attach
image

Looks like an underflow?

sunshowers:
This could happen if an empty `EventReport` is passed in -- in that case we'd
transition to `Running` but return `NotStarted`.

Fix this by not transitioning `self.kind` to `Running` if we're going to return
`NotStarted`. This does bloat up the code a little but I think is clearer
overall.

Thanks to @jgallagher for all the help debugging this!

Also clean up some related logic and add tests.

Fixes #4507.

### DIFF
--- a/update-engine/src/display/group_display.rs
+++ b/update-engine/src/display/group_display.rs
@@ -163,7 +163,7 @@ impl<K: Eq + Ord, W: std::io::Write, S: StepSpec> GroupDisplay<K, W, S> {
                     "root_total_elapsed" => ?result.root_total_elapsed,
                 );
             } else {
-                slog::debug!(
+                slog::trace!(
                     self.log,
                     "add_event_report called, state did not change";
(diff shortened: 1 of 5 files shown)


update-engine/src/display/group_display.rs
```diff
@@ -236,18 +236,9 @@ impl GroupDisplayStats {
     }
 
     fn apply_result(&mut self, result: AddEventReportResult) {
-        // Process result.after first to avoid integer underflow.
```
sunshowers, line 239: This wasn't really good enough to avoid integer underflow 😅, so I decided to just check for equality and then do the more obvious thing of applying `before` before `after`.

jordanhendricks (approved):
nice find!

PR #4566, opened 2023-11-28
Set version to 5.0.0

iliana:
[Last time around I raised a concern about jumping to x.0.0](https://github.com/oxidecomputer/omicron/pull/4437#issuecomment-1793652280) but we resolved that in the next updates meeting, and decided it's fine for all the inner control plane components to be versioned this way too for the time being. The actual updates system will rely less on implied semver semantics and will instead be more explicit, probably on ranges of versions.

### DIFF
--- a/.github/buildomat/jobs/package.sh
+++ b/.github/buildomat/jobs/package.sh
@@ -37,7 +37,7 @@ rustc --version
 # trampoline global zone images.
 #
 COMMIT=$(git rev-parse HEAD)
-VERSION="1.0.4-0.ci+git${COMMIT:0:11}"
+VERSION="5.0.0-0.ci+git${COMMIT:0:11}"
 echo "$VERSION" >/work/version.txt
 
 ptime -m ./tools/install_builder_prerequisites.sh -yp


leftwo (approved):
I don't know why that Oximeter test failed.  It does not seem related but we should
probably try to understand why it failed.

iliana:
Gonna hit retry on that. Links for the failed run, for my convenience:
https://github.com/oxidecomputer/omicron/pull/4566/checks?check_run_id=19107648858
https://buildomat.eng.oxide.computer/wg/0/details/01HGBHS5T9K24XNZBN5P935SRP/DqUsGcd1lsSgGXJZ9Qk7oMcSd0UOmoSjfc6NTe6dSp1BesXj/01HGBHSNG2W4FDB4QM0S40PMCR

bnaecker:
I've looked at the failure a bit. The test uses `tokio`'s time facilities to pause and advance time, and then checks that an expected number of events has occurred. I've generally found the time-manipulation functions to be robust, so long as the duration one uses in `tokio::time::advance` is much smaller than the duration of the other timers or intervals under test. The tick resolution is 1/10th the interval of the various timers, which apparently isn't small enough. I can't get this exact situation to repro on my machine, but I definitely can if I increase the interval.

I'll make a small PR increasing the tick resolution in the call to `tokio::time::advance()`.

Issue #4570, opened 2023-11-28
Need a check that field order in schema.rs matches order in dbinit.sql


sunshowers:
While developing https://github.com/oxidecomputer/omicron/pull/4520 I encountered a mysterious error `Received more than 8 bytes decoding i64. Was an expression of a different type misidentified as BigInt?` [(full error)](https://gist.github.com/sunshowers/5e0829d0a29f09ae03f5782fc094206f#file-gistfile0-txt-L28). It took a bit of debugging work to figure out that this was happening due to a mismatch in the field order between `dbinit.sql` and `schema.rs`.

We should have a check to ensure that the field order (and maybe types, if possible) match.

davepacheco:
Maybe something for #1534.

PR #4572, opened 2023-11-29
[nexus] Make 'update_and_check' CTE explicitly request columns

smklein:
Related to https://github.com/oxidecomputer/omicron/issues/4570 , but not a direct fix for it

This PR removes a usage of ".\*" from a SQL query. Using ".\*" in sql queries is somewhat risky -- it makes an implicit dependency on order, and can make backwards compatibility difficult in certain circumstances.

Instead, this PR provides a `ColumnWalker`, for converting a tuple of columns to an iterator, and requests the expected columns explicitly.

### DIFF
--- a/nexus/db-queries/src/db/column_walker.rs
+++ b/nexus/db-queries/src/db/column_walker.rs
@@ -1,0 +1,114 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! CTE utility for iterating over all columns in a table.
+
+use diesel::prelude::*;
+use std::marker::PhantomData;
+
+/// Used to iterate over a tuple of columns ("T").
+///
+/// Diesel exposes "AllColumns" as a tuple, which is difficult to iterate over
+/// -- after all, all the types are distinct. However, each of these types
+/// implements "Column", so we can use a macro to provide a
+/// "convertion-to-iterator" implemenation for our expected tuples.
+pub(crate) struct ColumnWalker<T> {
+    remaining: PhantomData<T>,
+}
+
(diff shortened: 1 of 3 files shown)


nexus/db-queries/src/db/column_walker.rs
```diff
@@ -0,0 +1,114 @@
...
+macro_rules! impl_column_walker {
+    ( $($column:ident)+ ) => (
+        impl<$($column: Column),+> IntoIterator for ColumnWalker<($($column,)+)> {
+            type Item = &'static str;
+            type IntoIter = std::vec::IntoIter<Self::Item>;
+
+            fn into_iter(self) -> Self::IntoIter {
+                // TODO: don't convert to vec? You'll need to figure
+                // out how to state the type of IntoIter.
+                [$($column::NAME,)+].to_vec().into_iter()
```
smklein, line 35: I would like to turn this into the following

```suggestion
                [$($column::NAME,)+].into_iter()
```

... but this would require changing:

```rust
type IntoIter = std::vec::IntoIter<Self::Item>;
```

Into:

```rust
type IntoIter = std::array::IntoIter<Self::Item, N>;
```

I don't know how to specify the type of "N" here, even though it **should** be knowable -- it's the length of `$column`.

sunshowers: I think either of the first two options in https://danielkeep.github.io/tlborm/book/blk-counting.html would be fine. A better approach would be https://github.com/rust-lang/rust/issues/83527 which isn't stable yet.

sunshowers: Actually in this case you can just pass in the length as an argument to `impl_column_walker`.

smklein: Ah, good call, I'll pass the length explicitly.

sunshowers (approved):
Looks great!

PR #4573, opened 2023-11-29
[nexus] improve external messages and make more available to clients

sunshowers:
While developing #4520, I observed that we were producing a number of error messages that were:

* 503 Service Unavailable,
* With only an internal message attached
* But where the message is both safe and useful to display to clients.

This is my attempt to make the situation slightly better. To achieve this, I made a few changes:

1. Make all the client errors carry a new `MessagePair` struct, which consists of an external message. (Along the way, correct the definition of e.g. the `Conflict` variant: it actually is an external message, not an internal one.)
2. Define a new `InsufficientCapacity` variant that consists of both an external and an internal message. This variant resolves to a 507 Insufficient Storage error, and has a more helpful message than just "Service Unavailable".
3. Turn some current 503 errors into client errors so that the message is available externally.

Looking for feedback on this approach!


davepacheco (commented):
Thanks for taking a look at this.  Is the motivation here or aid debugging for ourselves or to help end users and operators better understand why a 503 failed?  I ask because I can definitely see this being useful for developers debugging failures or for customers for some kinds of failures, but I worry this could easily wind up confusing customers more than it helps them when messages refer to implementation details instead of user-level abstractions that we've provided.  Example: if the system fails to create a Crucible volume because all available disk space has been allocated, it makes sense to me that internally we'd report something like "we couldn't find a zpool [or dataset] with enough space to allocate a region".  But those aren't terms that an end user or even operator would be expected to know about, right?  Ultimately I think we want the system to communicate something like: "the sum total of disks, images, etc. that exist have consumed all available storage capacity of the rack.  You must add capacity or delete some of these Disks."  That said, the request handler is not necessarily in a position to make the leap from "I couldn't find storage" to "there is no storage and you need to add more".  Ideally, I think we'd communicate something like "we couldn't find capacity" and we'd have other interfaces (alerts? a status page?) that would communicate the capacity situation and documentation that says "an error message about being unable to find capacity may mean that you're actually out -- check this status page".  I appreciate this is a lot more complicated and even we decide to go this route we don't have all these things today.

I think most of the messages that are made external in this PR don't feel suitable for consumption by an operator or an end user.

What do you think of this: instead of allowing `ServiceUnavailable` to have external messages, create new `Error` variants for the ones that we want to have external messages, and map those to 503 (preserving the external message) on the way out?  We could add an `InsufficientCapacity` variant with a way to express what resource we failed to allocate (like "storage space" or "IP addresses" or something).  The reason I suggest this is to make it harder to accidentally leak implementation details in error messages.  If a developer wants to expose something to the end user, they'd have to make a new variant (which I hope will cause people to consider more exactly what they're communicating to the API consumer).

I _think_ "out of capacity" (including both physical resources and virtual ones like IP addresses) covers all the cases we'd want to expose external messages for?  If we set those aside, I feel like a 503 is almost never going to be actionable for an end user and rarely so for an operator, too.  The cases I usually think about for a 503 are (1) the system suffered some very transient issue (e.g., the CockroachDB node crashed and a query failed during the request) or relatedly (2) the system has suffered too many failures to do what was requested (e.g., you tried to do something with a Crucible volume, all of whose "downstairs" instances are on sleds that were removed or panicked or whatever).  I don't think an end user can do anything about these.  For an operator, I think we've got to communicate this some better way (e.g., alerts, a status page showing such problems, etc.).  So for all these cases I feel like a more specific message would be more confusing than helpful.  The capacity-related ones do feel different because they are actionable, even to end users.  They might then go check how much capacity they're using, free up capacity, or just stop trying to do what they're doing.

sunshowers:
Thanks for the detailed review, @davepacheco! Yes, I'm thinking that operators are the most likely consumers of these messages, and that end users will ask operators for help if they run into an issue. My general view (mostly informed by devtools) is that status panes and such are all really useful, but providing synchronous, immediate "inner loop" feedback is crucial.

My initial work was definitely touching on "insufficient capacity" messages (just today there's been discussion in `#oxide-dogfood` about 503s that don't have a message attached.) I think it's fair to say that for external 503 errors, we should write informative messages.

I'll go over all the cases you pointed out and see what I think of them.

sunshowers:
@davepacheco I've addressed your specific feedback, and also broadened the scope to fix some of the other issues I spotted along the way (see updated PR description, e.g. the fact that `Conflict`'s message was mislabeled as internal when it should have been external). Would love feedback on the update.

common/src/api/external/error.rs
```diff
@@ -66,6 +69,88 @@ pub enum Error {
...
+}
+
+struct MessageVariantDisplay<'a>(&'a MessageVariant);
+
+impl<'a> Display for MessageVariantDisplay<'a> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self.0 {
+            MessageVariant::Internal(msg) => write!(f, "(internal: {})", msg),
+            MessageVariant::External { message, internal_context } => {
+                write!(f, ": {message}")?;
```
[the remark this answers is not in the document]
sunshowers: Changed this significantly, and followed this approach in the new version.

zephraph:
Thanks for putting this together @sunshowers! As I was working on #4605 I came across the need for a new error code to indicate to users that they can't provision more resources because it would exceed their set capacity. I'd also created `InsufficientCapacity` as an apt error so I'm excited to see this. The only question I have is if a user who's trying to provision a resource passed their granted capacity should see a 503 or a 400 error. There might be a slight difference in semantics between an internal operation failing due to running out of capacity in an unexpected / uncontrollable way vs a user hitting against their quota. 

Any thoughts? Ultimately I'm in favor of landing this with or without tweaking the response code as that's something we could always iterate on.

david-crespo:
@zephraph I don’t like 503 for what is ultimately IMO a non-exceptional  occurrence. Everything is fine, they’re just out of space in their corner of the world.

common/src/api/external/error.rs
```diff
@@ -55,15 +55,90 @@ pub enum Error {
...
+    }
+}
+
+struct MessagePairDisplayInternal<'a>(&'a MessagePair);
+
+impl<'a> Display for MessagePairDisplayInternal<'a> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.0.external_message)?;
+        if !self.0.internal_context.is_empty() {
+            write!(f, " (with internal context: {})", self.0.internal_context)?;
```
davepacheco, line 138: Was this change in formatting intentional?  It feels like kind of a big one but maybe I overestimate it.  Do you happen to have any examples handy from testing?

common/src/api/external/error.rs
```diff
@@ -292,28 +411,42 @@ impl From<Error> for HttpError {
                 internal_message,
             },
 
-            Error::InvalidRequest { message } => HttpError::for_bad_request(
-                Some(String::from("InvalidRequest")),
-                message,
-            ),
+            Error::InvalidRequest { message } => {
```
davepacheco, line 414: (applies to this whole function)

aside from the internal message, are any of these changing at all?

nexus/db-queries/src/db/datastore/external_ip.rs
```diff
@@ -151,6 +151,7 @@ impl DataStore {
                             "Requested external IP address not available",
                         )
                     } else {
+                        // XXX Should this be insufficient_capacity?
```
davepacheco, line 154: I think so.

nexus/db-queries/src/db/queries/region_allocation.rs
```diff
@@ -46,19 +46,23 @@ pub fn from_diesel(e: DieselError) -> external::Error {
         NOT_ENOUGH_UNIQUE_ZPOOLS_SENTINEL,
     ];
     if let Some(sentinel) = matches_sentinel(&e, &sentinels) {
+        let external_message = "Not enough disk space";
```
davepacheco, line 49:
```suggestion
        let external_message = "Not enough storage";
```

nexus/src/app/instance.rs
```diff
@@ -1583,28 +1580,22 @@ impl super::Nexus {
...
-                }),
-                InstanceState::Destroyed => Err(Error::ServiceUnavailable {
-                    internal_message: format!(
-                        "cannot connect to serial console of instance in state \
-                        {:?}",
-                        InstanceState::Stopped),
-                }),
+                | InstanceState::Failed => {
+                    Err(Error::invalid_request(format!(
+                        "cannot connect to serial console of {} instance",
```
davepacheco, line 1585: nitty, but I like the previous format a little better.  "cannot connect to serial console of instance in state \"Starting\"" communicates clearly that "Starting" is a discrete state that the user might know about and could go look up what it means and when it would change and how to check for it, etc.  "cannot connect to serial console of starting instance" makes me thing "what do you consider starting?"

nexus/src/app/vpc_router.rs
```diff
@@ -129,9 +129,9 @@ impl super::Nexus {
         // router kind cannot be changed, but it might be able to save us a
         // database round-trip.
         if db_router.kind == VpcRouterKind::System {
-            return Err(Error::MethodNotAllowed {
-                internal_message: "Cannot delete system router".to_string(),
-            });
+            return Err(Error::method_not_allowed(
```
davepacheco, line 132: I'm not sure if this would break anything but I think I'd make this a 403 or 400 instead and remove the MethodNotAllowed variant.  405 feels like a lower-level error -- I'd expect a client to be pretty confused by that.  This is just my intuition -- I checked docs and I don't think 405 is necessarily wrong here.

davepacheco (commented):
Cool -- I like this direction a lot.

FYI, there are some text and semantic changes here that seem important to UX (and compatibility) but they're a little buried in more mechanical changes that made it a little hard to make sure I spotted everything.

Thanks again for taking this on -- and fixing several cases in the process!

sunshowers:
> FYI, there are some text and semantic changes here that seem important to UX (and compatibility) but they're a little buried in more mechanical changes that made it a little hard to make sure I spotted everything.

I do apologize for that, I thought the change would be small but it ballooned very quickly.

common/src/api/external/error.rs
```diff
@@ -166,6 +254,32 @@ impl Error {
...
+        Error::InsufficientCapacity {
+            message: MessagePair::new_full(
+                external_message.into(),
+                internal_message.into(),
+            ),
+        }
+    }
+
+    /// Generates an [`Error::MethodNotAllowed`] error with an external message.
+    pub fn method_not_allowed(message: impl Into<String>) -> Error {
```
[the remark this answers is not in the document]
sunshowers: Yeah it's definitely a bit weird. I think 405 Method Not Allowed is typically used in spots where the method is *unconditionally* not permitted (e.g. `POST` on a `GET`-only endpoint). But in this case we're using it in a case where the method is *conditionally* not permitted, based on the input parameters. There's also the "Allow" header which [RFC 9110](https://httpwg.org/specs/rfc9110.html#status.405) says MUST be returned, but we don't do that.

Based on all this, maybe this should just be 400 Invalid Request.

common/src/api/external/error.rs
sunshowers: I think this isn't so much a formatting change as much as it is an addition. Previously, we just wouldn't track any internal context for the error variants that had associated external messages, and now we do.

In any case, this only affects internal logging; user-visible messages don't change.

common/src/api/external/error.rs
sunshowers: No, no changes.

nexus/db-queries/src/db/datastore/external_ip.rs
sunshowers: Changed, thanks.

nexus/db-queries/src/db/queries/region_allocation.rs
sunshowers: Done, thanks.

nexus/src/app/instance.rs
sunshowers: Sure, I wanted to align this with `cannot connect to serial console of destroyed instance` below, but I think your reasoning makes sense. As a compromise I ended up using the `Display` impl rather than the `Debug` one, so it says `cannot connect to serial console of instance in state "starting"`. (Not wedded to any particular outcome here at all, to be clear.)

nexus/src/app/vpc_router.rs
sunshowers: Yeah, agreed. Not providing the "Allow" header was the clincher for me. Turned this into 400.

david-crespo:
I don't love 507 for this, mostly because it's rare and weird and kinda WebDAV-specific. It does feel like a shame that there isn't something better. On reflection, despite what I said, 400 _is_ a stretch since there isn't anything in the request that's wrong —  you could make the exact same request 5 minutes later and it would be fine. I guess a 503 would be ok. My worry is that when I see a 503, I figure the whole system is borked rather than there being some limitation on the specific operation. But with the message clarifying the cause, 503 might be best.

sunshowers:
Hmm, I think a thing to consider is that 507 is a very specific error that won't be used (now or in the future) for anything else. Agreed that it comes from WebDAV but I've found other WebDAV-originating codes like [422 Unprocessable Entity/Content](https://httpwg.org/specs/rfc9110.html#status.422) to be very useful.

I think if 400 is ruled out, between 503 and 507 I would prefer 507.

david-crespo:
That's an interesting point, that the weirdness is a strength — no one will assume they know what it means. Persuasive!

common/src/api/external/error.rs
davepacheco: Ah, I think I misunderstood what this was used for.  I saw that this comes from `with_internal_context`, which elsewhere I think prepends a context message, and I thought this was replacing that.  I see that's basically still there at L110.  So I think I was just confused -- sorry.

nexus/db-queries/src/db/datastore/external_ip.rs
```diff
@@ -219,10 +219,13 @@ impl DataStore {
                             "Requested external IP address not available",
                         ))
                     } else {
-                        // XXX Should this be insufficient_capacity?
-                        Error::invalid_request(
-                            "No external IP addresses available",
-                        ))
+                        TransactionError::CustomError(
+                            // XXX Should this be insufficient_capacity?
```
davepacheco, line 223: Can drop this XXX

davepacheco (approved):

nexus/db-queries/src/db/datastore/external_ip.rs
sunshowers: Thanks for catching this!

zephraph:
@ahl suggested I dig into how AWS/GCP handles errors a bit to get a sense of how we might differentiate errors that are human controllable (a user is trying to allocate resources that would bust their quota) vs things outside of their control (not enough hardware to take the operation). AWS has some good documentation on their errors (bifurcated on client vs server) with rich descriptions of each: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/errors-overview.html. As I'd somewhat suggested, AWS gives 400 for client errors which include things like trying to allocate a resource that would exceed your resource limit. They give 500 errors for things outside the user's control. 

As @sunshowers and I were discussing this we both agree that this means we likely need to continue increasing the explicitness of our errors.

PR #4577, opened 2023-11-29
Improve oximeter self-stat tests

bnaecker:
Reduces the tick interval in calls to `tokio::time::advance()` to ensure all timers complete reliably. See #4566 for context.

### DIFF
--- a/oximeter/collector/src/agent.rs
+++ b/oximeter/collector/src/agent.rs
@@ -659,6 +659,23 @@ mod tests {
     use tokio::time::Instant;
     use uuid::Uuid;
 
+    // Interval on which oximeter collects from producers in these tests.
+    const COLLECTION_INTERVAL: Duration = Duration::from_secs(1);
+
+    // Interval in calls to `tokio::time::advance`. This must be sufficiently
+    // small relative to `COLLECTION_INTERVAL` to ensure all ticks of internal
+    // timers complete as expected.
+    const TICK_INTERVAL: Duration = Duration::from_millis(10);
+
+    // Total number of collection attempts.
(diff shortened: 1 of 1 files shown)


oximeter/collector/src/agent.rs
```diff
@@ -659,6 +659,23 @@ mod tests {
...
+    const TICK_INTERVAL: Duration = Duration::from_millis(10);
+
+    // Total number of collection attempts.
+    const N_COLLECTIONS: usize = 5;
+
+    // Period these tests wait using `tokio::time::advance()` before checking
+    // their test conditions.
+    fn test_wait_period() -> Duration {
+        COLLECTION_INTERVAL * N_COLLECTIONS as u32 + COLLECTION_INTERVAL / 2
+    }
```
iliana, line 677: I was trying to figure out if I could make this a `const` expr and ended up with

```rust
const TEST_WAIT_PERIOD: Duration = Duration::from_millis(
    COLLECTION_INTERVAL.as_millis() as u64 * N_COLLECTIONS as u64
        + COLLECTION_INTERVAL.as_millis() as u64 / 2,
);
```

which is not necessarily better. (It sure is strange that `.as_millis()` returns u128 but `.from_millis()` takes a u64.)

iliana (approved):

oximeter/collector/src/agent.rs
bnaecker: Yeah, I was disappointed I couldn't find a clean const expr either. This one is pretty good though, thanks!

bnaecker: Fixed up in 5e76841ee

PR #4585, opened 2023-11-30
Add on-demand communication probes.

rcgoodfellow:
This PR adds the following.

## Probes

A new first-class Omicron element called a probe is introduced. A probe is similar to an instance but is underpinned by a zone instead of an HVM. They are managed much like instances in terms of lifecycle. They have network interfaces on a VPC that are externally reachable via ephemeral IP addresses. They also have network interfaces on the underlay network. The primary function of probes in this initial PR is for network testing. They come with daemons like `thundermuffin` pre-installed as SMF services to facilitate communications testing both through boundary services and within the rack. Probes may become more general over time to facilitate more kinds of testing beyond networking.

The idea for probes came from the desire to test the Oxide stack in an environment where the hardware is constructed by interconnected virtual machines. Nested virtualization is undesirable for a number of reasons, so probes plus virtual rack topologies give us the ability to test a significant surface area of the control plane and the underlying systems it manages, such as networking and storage. While the motivation for probes was to test in virtual environments, they're just as applicable to hardware environments. Because they're a first-class Omicron element, the same tests written using probes in virtual environments can run on real racks.

Probes currently require fleet admin privileges. This is enforced in the API handlers.

## A 4-gimlet 2-sidecar CI test

A primary goal of building the probes mechanism is to have automated multi-switch multi-sled tests that run in CI. This PR also takes the first step along that path. There are two new CI jobs added. The `a4x2-prepare` job builds and packages omicron for each of the 4 gimlets in the topology. This includes RSS configuration, individual sled configuration, and faux-mgs configuration for each scrimlet. The omicron packages and their configuration bits are then tarred up and provided as artifacts for the `a4x2-deploy` job.

The `a4x2-deploy` job extracts the artifacts from `a4x2-prepare` into a set of folders with omicron configuration and deployment archives for each gimlet in the topology. This is in a folder called `cargo-bay`. The cargo bay contains a top-level folder for each virtual gimlet. The [a4x2 falcon topology](https://github.com/oxidecomputer/testbed/tree/main/a4x2) knows to look for these folders and mounts each into the corresponding gimlet via P9fs mounts. In the cargo bay for each gimlet, there is also an initialization script that installs omicron in the VM using `omicron-install`. This initialization script is run automatically by falcon (this is in the [code](https://github.com/oxidecomputer/testbed/blob/main/a4x2/src/main.rs#L57-L71) for the topology itself, not something that is hard coded). Once the topology has been launched the control plane is on it's way toward coming up.

The way that we test the virtual rack is functional is through probes. There is a new testing program in this PR called `commtest` in the `end-to-end-tests` directory. This program takes an address to reach the Oxide API at as a parameter and waits for the `/ping` endpoint to become responsive. Once that happens, a probe is launched on each sled in the topology and a basic connectivity test is run against each probe using ICMP. The test checks for packet loss within a configurable threshold for each probe. If the test passes, then the CI job passes.

In addition to the virtual rack, the test topology also contains a pair of routers running the Arista network stack that the sidecars are connected to. The connection between the rack and these routers is facilitated by BGP. There is also a third router called the customer edge (`ce`) that connects to both Arista routers. This router is running Linux+FRR. It advertises a default route to both Arista routers that propagates via BGP to the Oxide rack routers. Similarly, the IP pool address block the Oxide rack advertises propagates to the customer edge router via the Arista routers through BGP. This customer edge VM is also running an `iptables` configuration to decouple the testing network from the host network on the lab machine the entire topology is running on. All that needs to be done on the lab machine to connect to the rack is create a route to the IP pool block the rack is using that uses the IP address of the external interface on the customer edge machine as a nexthop. This is done in the `a4x2-deploy` job.

Depends on
- https://github.com/oxidecomputer/helios/pull/143


rcgoodfellow:
Threre are still a few `TODO`s in here to take care of, but I think this is generally ready for review.

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -5967,6 +5975,125 @@ async fn current_user_ssh_key_delete(
...
+            &|_, p: &ProbeInfo| match paginated_by {
+                PaginatedBy::Id(_) => NameOrId::Id(p.id),
+                PaginatedBy::Name(_) => NameOrId::Name(p.name.clone().into()),
+            },
+        )?))
+    };
+    apictx.external_latencies.instrument_dropshot_handler(&rqctx, handler).await
+}
+
+/// View an instrumentation probe.
```
david-crespo, line 6016:
```suggestion
/// View instrumentation probe
```

etc. for the others. No articles, no periods

david-crespo:
I'm not sure how to think about fleet admin being required but the probes being project-scoped. I don't think we have anything else like that — everything else that requires fleet admin is in `/system/*` endpoints and is not silo-scoped. In the short term it's probably fine as this is a pretty advanced feature. You should probably give these endpoints a tag starting with `system/` (like `system/probes`) even if the path stays silo-scoped (i.e., the path doesn't have `/v1/system/` in front) so they get grouped with the operator stuff in the API docs sidebar.

* How useful would it be for these probes to show up in the web console for users who do have the appropriate permissions? We could in theory gate their presence a similar way we gate whether to turn the top left into a system/silo switcher (we check if the user has fleet viewer perms, we could find a way to check if they're an admin).
* How important is it that they are project-scoped? (And therefore silo-scoped) At a glance I can see that probes get ephemeral IPs, which need to be drawn from an IP pool, and IP pools are associated with silos. So being in a silo is the only way to currently to be associated with an IP pool. Maybe you could imagine using the service pool for this, though even the service pool works by association with a special internal silo.

rcgoodfellow:
> How important is it that they are project-scoped? (And therefore silo-scoped) At a glance I can see that probes get ephemeral IPs, which need to be drawn from an IP pool, and IP pools are associated with silos. So being in a silo is currently the only way to be associated with an IP pool. Maybe you could imagine using the service pool for this, though even the service pool works by association with a special internal silo.

The goal is for probes to look as much like instances as possible. So having them be a part of a project is a part of that. 

> I'm not sure how to think about fleet admin being required but the probes being project-scoped. I don't think we have anything else like that — everything else that requires fleet admin is in /system/* endpoints and is not silo-scoped. In the short term it's probably fine as this is a pretty advanced feature. 

One of the unique aspects of probes, is that we let the caller place the probe on a particular sled. That requires having access to sled IDs which I don't think ordinary users have.

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -5967,6 +5975,125 @@ async fn current_user_ssh_key_delete(
     apictx.external_latencies.instrument_dropshot_handler(&rqctx, handler).await
 }
 
+/// List instrumentation probes.
+#[endpoint {
+    method = GET,
+    path = "/v1/probes",
+    tags = ["probes"],
```
david-crespo, line 5982:
```suggestion
    tags = ["system/probes"],
```

The `system/` prefix is how we determine the grouping on the docs site.

https://github.com/oxidecomputer/docs/blob/e605c8a871d0e42e6eed4c512b9fcdbeb22f2682/app/routes/api.tsx#L62-L68

rcgoodfellow: Done

ahl:
Are probes intended to be something about which we'd educate customers? If so, under what scenarios would we imagine customers making use of these (i.e. outside of the context of an issue on which support is engaged)? This is (to me) an unfamiliar concept. If it's something we want in the product for 1. validation and 2. diagnostics in the field, would we prefer to have this either unpublished in the API (i.e. not in the OpenAPI / docs / SDK / CLI) and/or in some distinct API (aka "the double secret API" aka "the support API" aka "the experimental API" that might come with its own SDK / CLI / CLI subcommand?

rcgoodfellow:
Probes could be useful as a tool to narrow down problems for operators. If an end user reports a connectivity problem with a particular instance, an operator could launch a probe on the same sled, drawing an external address from the same IP pool, to help determine if the connectivity issue is at the instance level, the sled level or they could launch several probes to see if there is a broader connectivity issue.

That being said, the primary use for probes right now is to ensure working end-to-end networking for PRs in CI.

nexus/src/external_api/http_entrypoints.rs
rcgoodfellow: Done

nexus/tests/integration_tests/saml.rs
```diff
@@ -35,6 +35,8 @@ pub const SAML_IDP_DESCRIPTOR_NO_KEYS: &str =
     include_str!("data/saml_idp_descriptor_no_keys.xml");
 
 // Create a SAML IdP
+//TODO(ry) this is broken on my system
```
rcgoodfellow, line 38: All these saml tests are failing on my workstation with.

```
--- STDERR:              omicron-nexus::test_all integration_tests::saml::test_post_saml_response_with_relay_state ---
log file: /tmp/test_all-cff3f951c9d1b075-test_post_saml_response_with_relay_state.10581.0.log
note: configured to log to "/tmp/test_all-cff3f951c9d1b075-test_post_saml_response_with_relay_state.10581.0.log"
DB URL: postgresql://root@[::1]:41288/omicron?sslmode=disable
DB address: [::1]:41288
log file: /tmp/test_all-cff3f951c9d1b075-test_post_saml_response_with_relay_state.10581.2.log
note: configured to log to "/tmp/test_all-cff3f951c9d1b075-test_post_saml_response_with_relay_state.10581.2.log"
func=xmlSecOpenSSLAppCertLoadBIO:file=app.c:line=1294:obj=unknown:subj=d2i_X509_bio:error=4:crypto library function failed:openssl error: error:00000000:lib(0)::reason(0)
func=xmlSecOpenSSLAppKeyFromCertLoadBIO:file=app.c:line=1014:obj=unknown:subj=xmlSecOpenSSLAppCertLoadBIO:error=1:xmlsec library function failed:
func=xmlSecOpenSSLAppKeyLoadBIO:file=app.c:line=414:obj=unknown:subj=xmlSecOpenSSLAppKeyFromCertLoadBIO:error=1:xmlsec library function failed:
func=xmlSecOpenSSLAppKeyLoadMemory:file=app.c:line=306:obj=unknown:subj=xmlSecOpenSSLAppKeyLoadBIO:error=1:xmlsec library function failed:
thread 'integration_tests::saml::test_post_saml_response_with_relay_state' panicked at nexus/tests/integration_tests/saml.rs:1339:6:
expected success: expected status code 303 See Other, found 400 Bad Request
WARN: dropped CockroachInstance without cleaning it up first (there may still be a child process running and a temporary directory leaked)
WARN: temporary directory leaked: "/tmp/.tmp5scdly"
        If you would like to access the database for debugging, run the following:

        # Run the database
        cargo run --bin omicron-dev db-run --no-populate --store-dir "/tmp/.tmp5scdly/data"
        # Access the database. Note the port may change if you run multiple databases.
        cockroach sql --host=localhost:32221 --insecure
WARN: dropped ClickHouseInstance without cleaning it up first (there may still be a child process running and a temporary directory leaked)

   Canceling due to test failure
------------
     Summary [  18.944s] 1 test run: 0 passed, 1 failed, 1072 skipped
     SIGSEGV [   5.898s] omicron-nexus::test_all integration_tests::saml::test_post_saml_response_with_relay_state
```

I'm running stock helios that I update regularly.

rcgoodfellow: Here is what I'm running locally

```
ry@masaka:~$ uname -a
SunOS masaka 5.11 helios-2.0.22500 i86pc i386 i86pc
```
```
ry@masaka:~$ cat /etc/versions/build
heads/stlouis-0-g9318fd43fb
```

ahl:
> Probes could be useful as a tool to narrow down problems for operators. If an end user reports a connectivity problem with a particular instance, an operator could launch a probe on the same sled, drawing an external address from the same IP pool, to help determine if the connectivity issue is at the instance level, the sled level or they could launch several probes to see if there is a broader connectivity issue.
> 
> That being said, the primary use for probes right now is to ensure working end-to-end networking for PRs in CI.

In that case, my suggestion would be to keep it out of the documented API / CLI / etc. What do you think?

nexus/tests/integration_tests/oximeter.rs
```diff
@@ -152,7 +152,7 @@ async fn test_oximeter_reregistration() {
 
     // Timeouts for checks
     const POLL_INTERVAL: Duration = Duration::from_millis(100);
-    const POLL_DURATION: Duration = Duration::from_secs(30);
+    const POLL_DURATION: Duration = Duration::from_secs(60);
```
rcgoodfellow, line 155: These checks time out for me on occasion. Bumping to 60 seconds helps.

nexus/tests/integration_tests/saml.rs
rcgoodfellow: This appears to have been due use of the `mold` linker on illumos.

rcgoodfellow:
> In that case, my suggestion would be to keep it out of the documented API / CLI / etc. What do you think?

I've moved the probes API endpoints to an "experimental" API in 4f58a4a43ff27bcf392f0e3b1f8b0999b2e3bb0b

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -352,10 +352,33 @@ pub(crate) fn external_api() -> NexusApiDescription {
...
+    fn register_experimental<T>(
+        api: &mut NexusApiDescription,
+        endpoint: T,
+    ) -> Result<(), String>
+    where
+        T: Into<ApiEndpoint<Arc<ServerContext>>>,
+    {
+        let mut ep: ApiEndpoint<Arc<ServerContext>> = endpoint.into();
+        // only one tag is allowed
+        ep.tags = vec![String::from("hidden")];
```
sunshowers, line 367: Does this need to fail on the presence of any existing tags?

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -352,10 +352,33 @@ pub(crate) fn external_api() -> NexusApiDescription {
...
+        let experimental = String::from("/experimental");
+        if !ep.path.starts_with(&experimental) {
+            ep.path = experimental + &ep.path;
+        }
```
sunshowers, line 371: I'm a little bit unsure of this kind of conditional magic to be honest -- I'd rather either expect that the path begins with `/experimental`, or expect that it doesn't and add it.

sunshowers (commented):
(Just a comment on experimental API endpoints.)

Thank you for taking this on! I've been wanting something like this for a while, and I really like this solution (my own half-baked solution was to use sets of tags, but this is easier to read I think.)

nexus/src/external_api/http_entrypoints.rs
rcgoodfellow: the posture i was going for here was to be able to move endpoint definitions unmodified from experimental -> normal

e.g., the `#[endpoint(... stuff ...)]`  does not have to change

so this is an overwrite rather than a check on pre-existing conditions type of thing

nexus/src/external_api/http_entrypoints.rs
rcgoodfellow: This has been updated as we discussed in chat to be an unconditional update.

FelixMcFelix (commented):
Thanks Ry, from what I've looked over this is a very nice addition. I haven't yet had the chance to stand this up and test it locally, I'm hoping to do so shortly. I've left some questions throughout.

On 'what' we can test with this, I guess this gives us really good mileage for any traffic carried by a VPC (Instance<->Probe and Probe<->Rack-external). If we wanted to ask more targeted questions about the underlay itself, would we create probes and `zlogin` ?

nexus/db-queries/src/db/datastore/probe.rs
```diff
@@ -0,0 +1,411 @@
...
+        )
+        .unwrap();
+        let internal_default_name = db::model::Name::from(default_name.clone());
+
+        let (.., db_subnet) = LookupPath::new(opctx, self)
+            .project_id(authz_project.id())
+            .vpc_name(&internal_default_name)
+            .vpc_subnet_name(&internal_default_name)
+            .fetch()
+            .await?;
```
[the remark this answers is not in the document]
rcgoodfellow: Created a follow-up issue for this
- https://github.com/oxidecomputer/omicron/issues/5225

dev-tools/omdb/src/bin/omdb/db.rs
```diff
@@ -620,6 +621,24 @@ async fn lookup_instance(
         .with_context(|| format!("loading instance {instance_id}"))
 }
 
+/// Helper function to looks up an probe with the given ID.
```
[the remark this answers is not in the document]
rcgoodfellow: Done

end-to-end-tests/Cargo.toml
```diff
@@ -15,13 +15,22 @@ omicron-sled-agent.workspace = true
 omicron-test-utils.workspace = true
 oxide-client.workspace = true
 rand.workspace = true
-reqwest.workspace = true
-russh = "0.42.0"
-russh-keys = "0.42.0"
+russh = "0.40.2"
+reqwest = { workspace = true, features = ["cookies"] }
+russh-keys = "0.40.1"
```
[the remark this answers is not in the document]
rcgoodfellow: Yep, looks like it. Fixed.

end-to-end-tests/src/helpers/icmp.rs
```diff
@@ -0,0 +1,306 @@
...
+                        .saturating_sub(tgt.rx_count)
+                        as usize;
+                }
+            }
+            sleep(Duration::from_millis(10));
+        });
+    }
+
+    fn rx(self: Arc<Self>) {
+        // Spawn a background thread to receive ICMP replys and do the
```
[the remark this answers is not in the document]
rcgoodfellow: Done

end-to-end-tests/src/helpers/icmp.rs
```diff
@@ -0,0 +1,306 @@
...
+        // necessary accounting.
+        spawn(move || loop {
+            let mut ubuf = [MaybeUninit::new(0); 10240];
+            if let Ok((sz, _)) = self.sock.recv_from(&mut ubuf) {
+                let buf = unsafe { &slice_assume_init_ref(&ubuf) };
```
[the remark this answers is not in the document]
rcgoodfellow: Good call. Done.

nexus/blueprint-execution/src/resource_allocation.rs
```diff
@@ -888,9 +889,14 @@ mod tests {
                             "invalid NIC kind (expected service, got instance)"
                         )
                     }
-                    NetworkInterfaceKind::Service(id) => {
+                    NetworkInterfaceKind::Probe { .. } => {
+                        panic!(
+                            "invalid NIC kind (expected service, got instance)"
```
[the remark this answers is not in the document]
rcgoodfellow: I think this code no longer exists after rebase.

nexus/db-queries/src/db/datastore/probe.rs
```diff
@@ -0,0 +1,411 @@
...
+            interface.into_internal(db_subnet.ipv4_block.0.into());
+        interface.vni = vni.0;
+
+        Ok(ProbeInfo {
+            id: probe.id(),
+            name: probe.name().clone().into(),
+            sled: probe.sled,
+            interface,
+            external_ips,
+        })
```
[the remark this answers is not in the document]
rcgoodfellow: Done

nexus/db-queries/src/db/queries/network_interface.rs
```diff
@@ -652,7 +658,7 @@ impl NextMacAddress {
                     NextMacShifts::for_guest();
                 (base.into(), max_shift, min_shift)
             }
-            NetworkInterfaceKind::Service => {
+            NetworkInterfaceKind::Service | NetworkInterfaceKind::Probe => {
```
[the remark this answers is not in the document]
rcgoodfellow: Good catch fixed.

nexus/src/app/sagas/switch_port_settings_apply.rs
```diff
@@ -2,6 +2,8 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+// muffins
```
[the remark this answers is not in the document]
rcgoodfellow: LOL. I did this to run a quick incremental compile benchmark for someone and forgot it was there. Removed.

nexus/types/src/external_api/params.rs
```diff
@@ -2036,3 +2037,21 @@ pub struct UpdatesGetRepositoryParams {
...
+
+// Probes
+
+/// Create time parameters for probes.
+#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema)]
+pub struct ProbeCreate {
+    #[serde(flatten)]
+    pub identity: IdentityMetadataCreateParams,
+    pub sled: Uuid,
+    pub ip_pool: Option<Name>,
```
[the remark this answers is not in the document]
rcgoodfellow: Done

nexus/db-model/src/external_ip.rs
```diff
@@ -124,6 +126,7 @@ pub struct ExternalIp {
     // Only Some(_) for instance Floating IPs
     pub project_id: Option<Uuid>,
     pub state: IpAttachState,
+    pub is_probe: bool,
```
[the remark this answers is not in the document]
FelixMcFelix: Opened a separate issue for this: https://github.com/oxidecomputer/omicron/issues/5228

FelixMcFelix (approved):
Thanks for handling the nits/papercuts etc., looks good!

rcgoodfellow:
Unfortunately, I need to disable the `a4x2` CI jobs in this initial commit. The zones and services within VMs come up too unreliably and would be a major source of friction for day-to-day development. We'll continue to push to make things more reliable within VMs. 

The primary issue appears to be SMF service startup failures within zones. Sometimes it's CRDB, sometimes it's more basic things like ndp. These service start failures seem to correspond to periods of heavy I/O. I did try running `zpool trim` as a precursor to a4x2 CI runs, and that sped things up significantly - but unfortunately, executing several runs with the benefit of trim shows reliability issues remain. I see these issues much less often on my local workstation, so it does appear to be system performance related in some way.

The CI tests and machinery remain in place within this PR, but the buildomat `enable` property is set to false.

PR #4588, opened 2023-11-30
fix phased startup of zones when sled agent starts up

davepacheco:
When sled agent starts up, it starts up the zones it's responsible for in a specific order:

- internal DNS
- NTP (_starting_ boundary NTP zones implicitly requires that internal DNS be up)
- (wait for time sync)
- start everything else

At least, that's what the comments say, and that appears to be what it should do.  But the first call to start internal DNS _also_ starts the NTP zones today.  This appears to work by accident because the DNS zones happen to get processed first.

This was discussed on #4466:
https://github.com/oxidecomputer/omicron/pull/4466#discussion_r1409989281
https://github.com/oxidecomputer/omicron/pull/4466#discussion_r1409586932

This change fixes that.


jgallagher (approved):
Thanks!

citrus-it:
> NTP (starting boundary NTP zones implicitly requires that internal DNS be up)

What's the dependency here, being able to configure the boundary services for external network access?
(Chrony itself is not dependent on internal DNS, hence the question).

davepacheco:
> > NTP (starting boundary NTP zones implicitly requires that internal DNS be up)
> 
> What's the dependency here, being able to configure the boundary services for external network access? (Chrony itself is not dependent on internal DNS, hence the question).

It's a dependency in the sled agent code that _starts_ the boundary NTP zones, not the zones themselves.  The comment from the code explains:
https://github.com/oxidecomputer/omicron/blob/e3887d54e959401a169b40ae88fa9a675e2bdfff/sled-agent/src/services.rs#L839-L843

davepacheco:
Besides CI, I tested the tip of this branch (3320239) on madrid:

- mupdate all sleds
- deploy Omicron
- provision instance
- ssh into instance
- cold boot (`uadmin 2 1` on all four sleds)
- wait for it to come up
- provision instance
- ssh into instance

That all worked.

common/src/ledger.rs
```diff
@@ -184,7 +184,7 @@ pub trait Ledgerable: DeserializeOwned + Serialize + Send + Sync {
                 err,
             })
         } else {
-            warn!(log, "No ledger in {path}");
+            info!(log, "No ledger in {path}");
```
davepacheco, line 187: This also happens during normal operation prior to the ledgers being initially written (during initial setup, if nothing else) so I found it misleading to make these warnings.

PR #4597, opened 2023-12-01
Update Cargo.toml to match package-manifest for crucible

leftwo:


bnaecker:
The CI tests are failing, we think because [this entrypoint](https://github.com/oxidecomputer/omicron/blob/92aed1a25a57d8483fc212ba245d871514aeadee/sled-agent/src/sim/http_entrypoints_pantry.rs#L216) and possibly others are using the older snake-case naming convention. These effectively duplicate the Crucible Pantry HTTP API, but in a separate place, making it tricky to keep things in sync.

bnaecker (approved):
Thanks for tracking this down!

leftwo:
In testing on a bench gimlet with these bits, I'm now able to upload an image using the CLI:
```
EVT22200005 # ${OXIDE} disk import --project alan --description "jammy disk" --path /alan/jammy-server-cloudimg-amd64.raw --disk jammysource
[00:12:21] [████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████] 2.20 GiB/2.20 GiB (3.04 MiB/s, 0s)done!
EVT22200005 # 
```

Issue #4330, opened 2023-10-24
Want `oximeter` tooling for detecting timeseries conflicts during tests


bnaecker:
In #4220, a programming error in Dendrite manifested as a schema mismatch error in the `oximeter` collector logs. It's unfortunate that the source of the error is separated from its reporting by such a distance. It would be extremely helpful to have tooling which allows developers to run tests in their own repo which can detect this kind of error. Initially, catching schema conflicts is the goal, though one could imagine extending this to ensuring schema _updates_ follow particular constraints, when we build in functionality to update timeseries schema.

I talked with @davepacheco about this a bit, and we hit on a couple of different approaches: (1) write timeseries definitions in a DSL, and use that to generate the Rust types and (2) provide an associated function in the existing traits which spits out the full schema for each object. I'll expand on these below:

### Background

Currently, timeseries are constructed from a _target_ and _metric_. These are traits in `oximeter` which can be derived, and include methods to list the name of the object and details about its fields (names and types). All of these methods take `&self`, which may become relevant when we consider option (2).

### Timeseries DSL

In this approach, we move "back" one step in the compilation process. Rather than developers writing the Rust types, they write a small DSL from which we generate the Rust types. Using totally made-up syntax:

```toml
[[target]]
name = "foo"
fields = [ { name = "field0", type = "type0" }, ... ]

[[metric]]
name = "bar"
fields = [ { name = "other_field0", type = "type0" }, ... ]
datum_type = "cumulative_i64"

[[timeseries]]
target = "foo"
metric = "bar"
```

From this, we'd generate Rust types like:

```rust
#[derive(Target)]
struct Foo {
    field0: type0,
    ...
}

#[derive(Metric)]
struct Bar {
    other_field0: type0,
    ...
}
```

The Rust code is exactly what one might write manually today. This would allow us to run some checks against the TOML file, though, such as making sure that each timeseries has unique names, for example, which would have caught #4220. If that file is checked in to source control, this also enables checking changes against the current version.

### Associated schema function

In option (2), we'd either augment the existing `Target` and `Metric` traits, or add a new one. That would include a method like:

```rust
trait Target {
    fn schema() -> TargetSchema where Self: Sized;
}

trait Metric {
    fn schema() -> MetricSchema where Self: Sized;
}
```

[...]

PR #4602, opened 2023-12-01
Add functions to catch timeseries schema changes

bnaecker:
- Move schema types from the `oximeter-db` crate to `oximeter` proper
- Add a `SchemaSet` for storing a unique set of timeseries schema, and comparing against a set deserialized from file. This works like expectorate, with a few tweaks, and allows developers to catch any changes. If we want to relax this to catching _compatible_ changes, for some definition of that, we can do that pretty easily.


bnaecker:
Fixes #4330

oximeter/collector/src/self_stats.rs
```diff
@@ -168,4 +175,50 @@ mod tests {
...
+            producer_port: 12345,
+            base_route: String::from("/"),
+            reason: FailureReason::Unreachable.to_string(),
+            datum: Cumulative::new(0),
+        }
+    }
+
+    // Check that the self-stat timeseries schema have not changed.
+    #[test]
+    fn test_no_schema_changes() {
```
bnaecker, line 210: This is the one place I'm currently exercising the new test function. I can roll it out to other timeseries schema in follow-up PRs.

oximeter/db/src/lib.rs
```diff
@@ -117,136 +125,6 @@ pub enum Error {
     NonSequentialSchemaVersions,
 }
 
-/// A timeseries name.
```
bnaecker, line 120: Almost all of this negative diff was just moved into `oximeter/src/types.rs` and re-exported here.

oximeter/oximeter/src/schema.rs
```diff
@@ -0,0 +1,640 @@
...
+use serde::Deserialize;
+use serde::Serialize;
+use std::collections::btree_map::Entry;
+use std::collections::BTreeMap;
+use std::collections::BTreeSet;
+use std::fmt::Write;
+use std::path::Path;
+
+/// The name and type information for a field of a timeseries schema.
+#[derive(
```
bnaecker, line 27: Almost all of this positive diff was simply moved from `oximeter/db/src/lib.rs`.

oximeter/oximeter/src/schema.rs
```diff
@@ -0,0 +1,640 @@
...
+    /// and field sources.
+    ///
+    /// This is intentionally similar to `expectorate::assert_contents()`. If
+    /// the provided file doesn't exist, it's treated as empty. If it does, a
+    /// `SchemaSet` is deserialized from it and a comparison between that and
+    /// `self` is done.
+    ///
+    /// You can use `EXPECTORATE=overwrite` to overwrite the existing file,
+    /// rather than panicking.
+    pub fn assert_contents(&self, path: impl AsRef<Path>) {
```
bnaecker, line 325: This is the meat of the PR, the function that loads a checked-in schema filed and asserts it matches the ones in `self`. Note that this is a bit different from `expectorate` itself, since we're checking equality in a way that ignores the creation timestamp of the schema.

oximeter/oximeter/src/schema.rs
```diff
@@ -0,0 +1,640 @@
...
+    //
+    // This was exacerbated by the error reporting, where we did in fact sort
+    // all fields across the target and metric, making it difficult to tell how
+    // the derived schema was different, if at all.
+    //
+    // This test generates a sample with a schema where the target and metric
+    // fields are sorted within them, but not across them. We check that the
+    // derived schema are actually equal, which means we've imposed that
+    // ordering when deriving the schema.
+    #[test]
```
smklein, line 582: Great test, way to call this out now (the regression case was gnarly, and I'm happy we're guarding against it like this)

smklein (approved):

oximeter/oximeter/src/schema.rs
bnaecker: Ha thanks! I put this one in during that whole debugging issue, exactly because of how painful it was!

PR #4605, opened 2023-12-04
Add resource limits

zephraph:
This PR aims to introduce `quotas` as a concept into Nexus for allowing operators to enforce virtual resource limits at the silo level. The initial implementation will be limited to checks during instance start, disk creation, and snapshot creation. We will _not_ being doing advanced quota recalculation as system resources change. We will _not yet_ be enforcing intelligent quota caps where the sum of all quotas must be less than the theoretical available system virtual resources. 

The implementation of this functionality is shaped by [RFD-427](https://rfd.shared.oxide.computer/rfd/0427) but some desired functionality will be deferred given time/complexity constraints. 

Longer term I believe the shape of quotas and perhaps even their relationship to silos may change. This PR implements a simplified version that matches closely to how the virtual resource provisioning tables are already built out. I know there's some oddness around the shape of the quotas table with it not having its own ID and otherwise being mildly divergent from other resources, but this was largely to ensure we could migrate to another solution _and_ not overcomplicate the initial implementation. 

## TODO
- [x] Add quota creation as a step of silo creation
- [x] Add initialization checks in CTEs for instance create, etc to only proceed when quota unmet
- [x] Wire up CTE sentinels in upstream callsites
- [x] Add backfill migration for existing customers
- [x] Add tests for quota enforcement
- [x] Delete the quotas when the silo is deleted


nexus/db-queries/src/db/queries/virtual_provisioning_collection_update.rs
```diff
@@ -13,16 +13,52 @@ use crate::db::pool::DbConnection;
...
+    let sentinels = [
+        NOT_ENOUGH_CPUS_SENTINEL,
+        NOT_ENOUGH_MEMORY_SENTINEL,
+        NOT_ENOUGH_STORAGE_SENTINEL,
+    ];
+    if let Some(sentinel) = matches_sentinel(&e, &sentinels) {
+        match sentinel {
+            NOT_ENOUGH_CPUS_SENTINEL => {
+                return external::Error::InvalidRequest { message: "Insufficient Capacity: Not enough CPUs to complete request. Either stop unused instances to free up resources or contact the rack operator to request a capacity increase.".to_string() }
+            }
```
zephraph, line 50: Eventually it would be nice to use @sunshowers' `InsufficientCapacity` error from https://github.com/oxidecomputer/omicron/pull/4573

augustuswm:
Looking at the APIs, what is the intended plan for if an operator wanted to remove a quote, or a single aspect of a quota? Would it be to set that field to an arbitrarily high number?

zephraph:
@augustuswm an operator can't remove a quota. All silos are required to have quotas as it effectively determines what capacity the silo has. _Today_ an operator could set an arbitrarily high value such that the quota is effectively meaningless because the silo would starve the system of resources before hitting the quota. That's obviously not going to be how it is longer term. Eventually we will have some calculated total virtual capacity which the sum of all quotas must be less than or equal to.

openapi/nexus.json
```diff
@@ -6309,6 +6368,90 @@
...
+          "5XX": {
+            "$ref": "#/components/responses/Error"
+          }
+        }
+      },
+      "put": {
+        "tags": [
+          "system/quotas"
+        ],
+        "summary": "Update the resource quotas of a given silo",
```
ahl, line 6412: I assume that omitting a particular resource (e.g. just specifying `memory`) means that it would be unaffected; this seems worth including in the documentation somewhere.

nexus/tests/output/nexus_tags.txt
```diff
@@ -164,6 +164,12 @@ networking_switch_port_settings_delete   DELETE   /v1/system/networking/switch-p
 networking_switch_port_settings_list     GET      /v1/system/networking/switch-port-settings
 networking_switch_port_settings_view     GET      /v1/system/networking/switch-port-settings/{port}
 
+API operations found with tag "system/quotas"
```
ahl, line 167: I think just put it under `system/silos`

ahl (commented):
Great shit and I can't thank you enough for taking this on.

`/v1/system/quotas` seems redundant with list all silos, GET `/v1/system/silos/{silo}/quotas` on each. Maybe it's worth it for a common use case to avoid the N+1 problem. I don't know.

How do we envision accessing use against those resources? It seems like often I'm going to ask: what are my limits and how much am I using right now? Do we imagine that as two call in two distinct parts of the API?

nexus/db-queries/src/db/datastore/silo.rs
```diff
@@ -28,6 +28,7 @@ use chrono::Utc;
 use diesel::prelude::*;
 use nexus_db_model::Certificate;
 use nexus_db_model::ServiceKind;
+use nexus_db_model::SiloQuotas;
```
smklein, line 31: (as discussed on our call) It's down in `load_builtin_silo` but we need to insert the quotas here too!

nexus/db-queries/src/db/datastore/quota.rs
```diff
@@ -0,0 +1,117 @@
...
+        let silo_id = authz_silo.id();
+        use db::schema::silo_quotas::dsl;
+
+        let result = conn
+            .transaction_async(|c| async move {
+                diesel::insert_into(dsl::silo_quotas)
+                    .values(quotas)
+                    .execute_async(&c)
+                    .await
+                    .map_err(TransactionError::CustomError)
```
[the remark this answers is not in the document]
zephraph: It just needs to use the connection (because it's inside the silo create transaction). I'll check for a different way to specify that.

nexus/db-queries/src/db/queries/virtual_provisioning_collection_update.rs
zephraph: On second thought, maybe not. I think we should distinguish between hitting up against human set capacity limits vs hardware limits.

zephraph:
> Great shit and I can't thank you enough for taking this on.

🙏 

> `/v1/system/quotas` seems redundant with list all silos, GET `/v1/system/silos/{silo}/quotas` on each. Maybe it's worth it for a common use case to avoid the N+1 problem. I don't know.

I think it's worth it for now but its usefulness will be reduced when I land the capacity and utilization work. We can revisit it in that PR.

> How do we envision accessing use against those resources? It seems like often I'm going to ask: what are my limits and how much am I using right now? Do we imagine that as two call in two distinct parts of the API?

To your last question, no, I think that should be the same API. Seeing only the quotas in isolation isn't (in my mind) very useful.

nexus/db-queries/src/db/datastore/quota.rs
```diff
@@ -0,0 +1,103 @@
...
+            .map_err(|e| {
+                public_error_from_diesel(
+                    e,
+                    ErrorHandler::Conflict(
+                        ResourceType::SiloQuotas,
+                        &silo_id.to_string(),
+                    ),
+                )
+            })
+            .map(|_| ())
```
zephraph, line 47: @smklein this look okay?

nexus/db-queries/src/db/datastore/silo.rs
zephraph: TODO for myself: create an issue for the bifurcation of silo creation

nexus/db-queries/src/db/datastore/quota.rs
smklein: Seems fine to me!

nexus/types/src/external_api/params.rs
```diff
@@ -297,6 +303,50 @@ pub struct SiloCreate {
...
+
+    /// 30% of CPUs and memory reserved for internal usage.
+    /// Storage calculated at (total / 3.5) to account for redundancy / bookkeeping.
+    pub fn half_rack() -> Self {
+        Self {
+            cpus: 90,
+            memory: ByteCount::from_gibibytes_u32(708),
+            storage: ByteCount::from_gibibytes_u32(850),
+        }
+    }
```
zephraph, line 335: Just remembered this is only the resources for a single sled. Need to multiply this by... 16?

nexus/db-queries/src/db/queries/virtual_provisioning_collection_update.rs
sunshowers: @zephraph and I chatted about this and post- this is what I suggested:

* Have the `InsufficientCapacity` variant carry around an enum, e.g. `InsufficientCapacityKind::UserQuota` and `::System`, and the `insufficient_capacity` constructor accept that as an argument (forcing callers to specify which `InsufficientCapacityKind` variant they want)
* In the message, include the kind: "Insufficient capacity (user quota)" and "... (system)"
* Consider returning different HTTP error codes for each variant.

cc @askfongjojo who I think has some thoughts about this.

askfongjojo: @sunshowers @zephraph: Thanks for looping me in. Once silo quotas are fully in place, user should rarely hit the `System` capacity error, assuming that the sum of all silo quotas should not exceed the total usable capacity of the rack. But I can think of two situations when the `System` capacity error would be hit before the `UserQuota` one:
1. Some sleds are in maintenance or taken out of the provisioning pool
2. All sleds are relatively full and the user is trying to provision a very large VM.

As such, perhaps the `UserQuota` kind error will be a HTTP 400 (end user needs to free up some capacity themselves) whereas the `System` one is a 507 (operator needs to migrate some instances, check the sled status, force some silos to reduce their usage temporarily, etc.)?

nexus/types/src/external_api/params.rs
zephraph: It's basically a shortcut to set a decent top line quota that (will when I fix it) equate out to the calculated capacity of a half rack. Right now it's basically used for tests and other places we need to allocate some amount of resources that can't be 0.

zephraph: I've updated this function to be called `from_sled_count` and to take in a number of sleds. So the usage is like `SiloQuotaCreate::from_sled_count(16)` for a half rack. 

This function is mostly a helper/placeholder. All of the places we have current (including the default silo) are largely for testing. I don't think it's worth doing anything fancy for the default silo b/c we should ultimately be removing it. Further, this number isn't really all that accurate. To do this correctly, you'd want to query all the sleds, get there hardware based numbers, and use those to calculate the totals with the slush factors. We will do that at some point but even then these usages might not be worth updating. 

Regardless, this doesn't need to be smart right now and it's here b/c to aid in creating quotas with non-empty silos.

nexus/db-queries/src/db/queries/virtual_provisioning_collection_update.rs
zephraph: For now I'm just using the `InsufficientCapacity` error wholesale. This isn't the ideal state, it would be good to follow up with a better error state, but it's sufficient for the time being.

zephraph: I created an issue to denote the need for better error messages: https://github.com/oxidecomputer/omicron/issues/4680

nexus/db-queries/src/db/queries/virtual_provisioning_collection_update.rs
```diff
@@ -195,6 +375,13 @@ where
     }
 }
 
+/// The virtual resource collection is only updated when a resource is inserted
+/// or deleted from the resource provisioning table for idempotency.
```
smklein, line 379:
```suggestion
/// The virtual resource collection is only updated when a resource is inserted
/// or deleted from the resource provisioning table. By probing for the presence
/// or absence of a resource, we can update collections at the same time as we
/// create or destroy the resource, which helps make the operation idempotent.
```

nexus/tests/integration_tests/quotas.rs
```diff
@@ -0,0 +1,280 @@
...
+        .execute()
+        .await
+    }
+
+    async fn cleanup_instance(
+        &self,
+        client: &ClientTestContext,
+        name: &str,
+    ) -> TestResponse {
+        // Stop instance if it's started... can probably ignore errors here
```
smklein, line 113: This comment seems out-of-date, I think it's okay to `expect` the error

nexus/tests/integration_tests/quotas.rs
```diff
@@ -0,0 +1,280 @@
...
+        "quota-test-silo",
+        params::SiloQuotasCreate::empty(),
+    )
+    .await;
+
+    // Ensure trying to provision an instance with empty quotas fails
+    system
+        .provision_instance(client, "instance", 1, 1)
+        .await
+        .expect_err("should've failed with insufficient CPU quota");
```
smklein, line 245: tiiiiiiny nitpick.

Whenever you do `expect_err`, we just expect **any** error, not necessarily the one you expected.

If we updated this to:

```rust
let err = system.provision_instance(...).await.expect_err("Should have failed!").to_string();
assert!(err.contains("Insufficient CPU quota"), "Unexpected error: {err}");
```

The test would be a little safer -- it would (correctly!) fail if a different error was getting spit out here

nexus/tests/integration_tests/quotas.rs
```diff
@@ -0,0 +1,280 @@
...
+
+    // Allocating instance should now succeed
+    system
+        .provision_instance(client, "instance", 2, 10)
+        .await
+        .expect("Instance should've had enough resources to be provisioned");
+
+    system.provision_disk(client, "disk", 3).await.expect_err(
+        "Disk should not be provisioned because it exceeds the quota",
+    );
```
smklein, line 274: Same comment re: `expect_err`

schema/crdb/20.0.0/up02.sql
```diff
@@ -0,0 +1,30 @@
...
+SELECT
+  s.id AS silo_id,
+  NOW() AS time_created,
+  NOW() AS time_modified,
+  -- ~70% of 128 threads leaving 30% for internal use
+  90 AS cpus,
+  -- 708 GiB (~70% of memory leaving 30% for internal use)
+  760209211392 AS memory_bytes,
+  -- 850 GiB (total storage / 3.5)
+  912680550400 AS storage_bytes
```
askfongjojo, line 25: There are on ongoing discussions about abandoning the use of "assumed" usable capacity figures. But if we still want to use them, these numbers be multiplied by the number of sleds and physical disks (excluding the M2s). So a half-rack will provide 90*16 = 1440 threads, and so on.

smklein (approved):
This looks great, thank you for taking on the heroic effort of tackling this. The CTE and DB migration code make sense to me -- this cut all the way across Nexus, and looks great.

A few comments below, but the only thing I'd consider a "blocker" is the `from_sled_count` function, which is IMO kinda misleading. Everything else is just nits!

nexus/types/src/external_api/params.rs
```diff
@@ -297,6 +303,54 @@ pub struct SiloCreate {
...
+    ///
+    /// 30% of CPUs and memory reserved for internal usage.
+    /// Storage calculated at (total / 3.5) to account for redundancy / bookkeeping.
+    pub fn from_sled_count(num_sleds: u32) -> Self {
+        Self {
+            cpus: 90 * i64::from(num_sleds),
+            memory: ByteCount::from_gibibytes_u32(708 * num_sleds),
+            storage: ByteCount::from_gibibytes_u32(850 * num_sleds),
+        }
+    }
```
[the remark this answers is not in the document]
zephraph: As per our conversations in chat, I've dropped this `from_sled_count` function in favor of a function called `arbitrarily_high_default` which just returns a lot of 9s for each value. I've updated the migration to use the same number of 9s so it's a consistently identifiable set of numbers.

nexus/tests/integration_tests/quotas.rs
zephraph: Thanks for the reminder. I had this in my notes from our pairing session but forgot to go back and add it. Doing that now.

schema/crdb/20.0.0/up02.sql
zephraph: This is replaced with an arbitrarily high limit, significantly more than can be provisioned on even a full rack.

nexus/types/src/external_api/params.rs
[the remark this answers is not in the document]
smklein: Looks great, thanks for patching!

nexus/db-queries/src/db/datastore/silo.rs
```diff
@@ -61,13 +62,32 @@ impl DataStore {
...
-            .do_nothing()
-            .execute_async(&*self.pool_connection_authorized(opctx).await?)
+        use db::schema::silo_quotas::dsl as quotas_dsl;
+        let count = self
+            .pool_connection_authorized(opctx)
+            .await?
+            .transaction_async(|conn| async move {
+                diesel::insert_into(quotas_dsl::silo_quotas)
+                    .values(SiloQuotas::arbitrarily_high_default(
+                        DEFAULT_SILO.id(),
```
david-crespo, line 72: Could use a comment on the possibly-surprising-in-the-future fact that we do this for default silo but not internal silo.

david-crespo (approved):
Unbelievable work. The core diesel bit is wild. Had a couple of comments, nothing important.

Issue #4604, opened 2023-12-02
Wicketd: Make refresh method more robust


sunshowers:
During the dogfood mupdate yesterday (2023-11-30), we saw that wicketd was stuck in the maintenance state because the smf `refresh` method failed at startup with these logs:

```
[ Dec 28 02:21:23 Executing start method ("ctrun -l child -o noorphan,regent /opt/oxide/wicketd/bin/wicketd run /var/
svc/manifest/site/wicketd/config.toml --address [::1]:12226 --artifact-address [fdb0:a840:2504:195::2]:12227 --mgs-ad
dress [::1]:12225 --nexus-proxy-address [::]:12229 --baseboard-file /opt/oxide/baseboard.json --read-smf-config &"). 
]
[ Dec 28 02:21:23 Method "start" exited with status 0. ]
[ Dec 28 02:21:25 Rereading configuration. ]
[ Dec 28 02:21:25 Executing refresh method ("curl -X POST http://[::1]:12226/reload-config"). ]
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
^M  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0^M  0     0    0     0    0     0    
  0      0 --:--:-- --:--:-- --:--:--     0
curl: (7) Failed to connect to ::1 port 12226 after 0 ms: Couldn't connect to server
[ Dec 28 02:21:25 Method "refresh" exited with status 7. ]
[ Dec 28 02:21:25 Executing refresh method ("curl -X POST http://[::1]:12226/reload-config"). ]
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
^M  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0^M  0     0    0     0    0     0    
  0      0 --:--:-- --:--:-- --:--:--     0
curl: (7) Failed to connect to ::1 port 12226 after 0 ms: Couldn't connect to server
[ Dec 28 02:21:25 Method "refresh" exited with status 7. ]
[ Dec 28 02:21:25 Executing refresh method ("curl -X POST http://[::1]:12226/reload-config"). ]
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
[...]

PR #4606, opened 2023-12-04
[wicketd] Add refresh-config subcommand instead of using curl 

jgallagher:
This allows us to add some time for refreshing a wicketd that hasn't fully started (and therefore isn't reachable on its dropshot server yet).

Fixes https://github.com/oxidecomputer/omicron/issues/4604.

Testing this is a little awkward because config refresh is only support on illumos via SMF, so I tested this by hand on `madrid`:

0. Disable wicketd (`svcadm disable wicketd`)
1. Install the new wicketd binary and `manifest.xml` in the switch zone
2. Import the new manifest.xml (`svccfg import /var/svc/manifest/site/wicketd/manifest.xml`)
3. (I expected to need to restore the config properties that sled-agent had set, but they persisted)
4. Enable wicketd and immediately try to refresh it (`svcadm enable wicketd && svcadm refresh wicketd`)

In the wicketd logs, we see that the first refresh attempt failed because it fired before the dropshot server was up:

```
[ Dec  4 18:34:22 Executing start method ("ctrun -l child -o noorphan,regent /opt/oxide/wicketd/bin/wicketd run /var/svc/manifest/site/wicketd/config.toml --address [::1]:12226 --artifact-address [fdb0:a840:2504:355::2]:12227 --mgs-address [::1]:12225 --nex
us-proxy-address [::]:12229 --baseboard-file /opt/oxide/baseboard.json --read-smf-config &"). ]
[ Dec  4 18:34:22 Method "start" exited with status 0. ]
[ Dec  4 18:34:22 Rereading configuration. ]
note: configured to log to "/dev/stdout"
[ Dec  4 18:34:22 Executing refresh method ("/opt/oxide/wicketd/bin/wicketd refresh-config /var/svc/manifest/site/wicketd/config.toml --address [::1]:12226"). ]
note: configured to log to "/dev/stdout"
18:34:22.330Z WARN wicketd: failed to refresh wicketd config (attempt 1 of 3); will retry after 5s
    err = Communication Error: error sending request for url (http://[::1]:12226/reload-config): error trying to connect: tcp connect error: Connection refused (os error 146)\nCaused by:\n  -> error sending request for url (http://[::1]:12226/reload-config)
: error trying to connect: tcp connect error: Connection refused (os error 146)\n  -> error trying to connect: tcp connect error: Connection refused (os error 146)\n  -> tcp connect error: Connection refused (os error 146)\n  -> Connection refused (os error
 146)
18:34:22.396Z INFO wicketd (dropshot (wicketd)): listening
    file = /home/john/.cargo/git/checkouts/dropshot-a4a923d29dccc492/ff87a01/dropshot/src/server.rs:195
    local_addr = [::1]:12226
```

10 seconds later, we see the successful connection, POST, and exit of the SMF `refresh`:

```
18:34:32.332Z INFO wicketd (dropshot (wicketd)): accepted connection
    file = /home/john/.cargo/git/checkouts/dropshot-a4a923d29dccc492/ff87a01/dropshot/src/server.rs:769
    local_addr = [::1]:12226
    remote_addr = [::1]:32976
18:34:32.388Z INFO wicketd (dropshot (wicketd)): request completed
    file = /home/john/.cargo/git/checkouts/dropshot-a4a923d29dccc492/ff87a01/dropshot/src/server.rs:853
    latency_us = 30475
    local_addr = [::1]:12226
    method = POST
    remote_addr = [::1]:32976
    req_id = e0c2034a-0a99-45c1-a651-57249ca258f0
    response_code = 204
    uri = /reload-config
[ Dec  4 18:34:32 Method "refresh" exited with status 0. ]
```

### DIFF
--- a/wicketd/src/bin/wicketd.rs
+++ b/wicketd/src/bin/wicketd.rs
@@ -26,7 +26,7 @@ enum Args {
         #[clap(name = "CONFIG_FILE_PATH", action)]
         config_file_path: PathBuf,
 
-        /// The address for the technician port
+        /// The address on which the main wicketd dropshot server should listen
         #[clap(short, long, action)]
         address: SocketAddrV6,
 
@@ -56,6 +56,19 @@ enum Args {
         /// via `--read-smf-config` or an SMF refresh
         #[clap(long, action, conflicts_with("read_smf_config"))]
         rack_subnet: Option<Ipv6Addr>,
+    },
+
+    /// Instruct a running wicketd server to refresh its config
+    ///
+    /// Mechanically, this hits a specific endpoint served by wicketd's dropshot
+    /// server
+    RefreshConfig {
+        #[clap(name = "CONFIG_FILE_PATH", action)]
+        config_file_path: PathBuf,
+
+        /// The address of the server to refresh
+        #[clap(short, long, action)]
+        address: SocketAddrV6,
     },
 }
 
@@ -139,6 +152,23 @@ async fn do_run() -> Result<(), CmdError> {
                 .wait_for_finish()
                 .await
                 .map_err(|err| CmdError::Failure(anyhow!(err)))
+        }
+        Args::RefreshConfig { config_file_path, address } => {
+            let config = Config::from_file(&config_file_path)
+                .with_context(|| {
+                    format!("failed to parse {}", config_file_path.display())
+                })
+                .map_err(CmdError::Failure)?;
+
+            let log = config
+                .log
+                .to_logger("wicketd")
+                .context("failed to initialize logger")
+                .map_err(CmdError::Failure)?;
+
+            Server::refresh_config(log, address)
+                .await
+                .map_err(CmdError::Failure)
         }
     }
 }
--- a/wicketd/src/bin/wicketd.rs
+++ b/wicketd/src/bin/wicketd.rs
@@ -5,6 +5,7 @@
 //! Executable for wicketd: technician port based management service
 
 use anyhow::{anyhow, Context};
+use camino::Utf8PathBuf;
 use clap::Parser;
 use omicron_common::{
     address::Ipv6Subnet,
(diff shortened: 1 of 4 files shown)


wicketd/src/bin/wicketd.rs
```diff
@@ -57,6 +57,19 @@ enum Args {
...
         rack_subnet: Option<Ipv6Addr>,
     },
+
+    /// Instruct a running wicketd server to refresh its config
+    ///
+    /// Mechanically, this hits a specific endpoint served by wicketd's dropshot
+    /// server
+    RefreshConfig {
+        #[clap(name = "CONFIG_FILE_PATH", action)]
+        config_file_path: PathBuf,
```
sunshowers, line 67: Can use `Utf8PathBuf` here

wicketd/src/bin/wicketd.rs
```diff
@@ -140,5 +153,22 @@ async fn do_run() -> Result<(), CmdError> {
...
+
+            let log = config
+                .log
+                .to_logger("wicketd")
+                .context("failed to initialize logger")
+                .map_err(CmdError::Failure)?;
+
+            Server::refresh_config(log, address)
+                .await
+                .map_err(CmdError::Failure)
```
sunshowers, line 171: Could you put in some explicit documentation here about exit codes? I believe this exits with either 1 or 2, which is fine because it doesn't collide with any of the exit codes in https://illumos.org/man/7/smf_method.

sunshowers (approved):

jgallagher:
> This also feels like maybe it can have an integration test?

We chatted about this offline and decided this probably isn't worth it. In all our current testing environments the wicketd we talk to wouldn't actually be able to reload its config:

1. Reloading config [requires illumos](https://github.com/oxidecomputer/omicron/blob/0a781f9957f5da1ef4b2ca7dce1d7bf13be5b739/wicketd/src/lib.rs#L112-L115)
2. Even on illumos, reloading config requires that wicketd be [running under SMF](https://github.com/oxidecomputer/omicron/blob/0a781f9957f5da1ef4b2ca7dce1d7bf13be5b739/wicketd/src/lib.rs#L80-L82)

We could add a test that `wicketd reload-config` attempts to hit the correct endpoint which then reports an error, but that isn't testing the parts of this change that are more likely to be wrong (e.g., the `manifest.xml` change).

PR #4608, opened 2023-12-04
[nexus] remove views::SledProvisionState::Unknown

sunshowers:
This doesn't quite work as expected:

* As an input type, if an unknown `SledProvisionState` is specified, we immediately produce an error as soon as we enter the HTTP entrypoint. There's no functional difference between that and producing an error at deserialization time.
* As an output type, progenitor doesn't support `#[serde(other)]` so the unknown type doesn't work.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4179,7 +4179,6 @@ dependencies = [
  "schemars",
  "serde",
  "serde_json",
- "serde_with",
  "steno",
  "strum",
  "uuid",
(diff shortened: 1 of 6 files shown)


ahl (approved):
Thanks! FWIW while it's true that progenitor doesn't do anything special with regard to unknown variants, it also doesn't choke on them! I think we might want the json schema to reflect the unknown/catchall nature of a variant like this if we were to do this in the future.

PR #4620, opened 2023-12-06
API updates for oxidecomputer/maghemite#88 fix

rcgoodfellow:
This PR updates Maghemite and makes the necessary API changes for the oxidecomputer/maghemite#88 fix.

- Depends on oxidecomputer/maghemite#96

The primary change here is that the [`ApplyRequest`](https://github.com/oxidecomputer/maghemite/pull/96/files#diff-d6c62eabcffd50242753cc5d8eae2265ecc08763166275856872b5fae243c3deR541-R558) parameter in the `mgd` API changed from

```rust
pub struct ApplyRequest {
    pub peer_group: String,
    pub peers: Vec<BgpPeerConfig>
}
```

to

```rust
pub struct ApplyRequest {
    /// ASN to apply changes to.
    pub asn: u32,
    /// Complete set of prefixes to originate. Any active prefixes not in this
    /// list will be removed. All prefixes in this list are ensured to be in
    /// the originating set.
    pub originate: Vec<Prefix4>,
    /// Lists of peers indexed by peer group. Set's within a peer group key are
    /// a total set. For example, the value
    ///
    /// ```text
    /// {"foo": [a, b, d]}
    /// ```
    /// Means that the peer group "foo" only contains the peers `a`, `b` and
    /// `d`. If there is a peer `c` currently in the peer group "foo", it will
    /// be removed.
    pub peers: HashMap<String, Vec<BgpPeerConfig>>,
 }
```

This data structure change pulls originated prefixes out of individual peers into the top-level object. It also organizes peer configs in a map indexed by peer group.

The primary changes in this PR are around code that translates Omicron API requests into this new Maghemite API parameter structure.


nexus/src/app/sagas/switch_port_settings_common.rs
```diff
@@ -454,11 +456,44 @@ pub(crate) async fn ensure_switch_port_bgp_settings(
...
+            hold_time: peer.hold_time.0.into(),
+            idle_hold_time: peer.idle_hold_time.0.into(),
+            delay_open: peer.delay_open.0.into(),
+            connect_retry: peer.connect_retry.0.into(),
+            keepalive: peer.keepalive.0.into(),
+            resolution: BGP_SESSION_RESOLUTION,
+            passive: false,
+        };
+
+        match bgp_peer_configs.get_mut(&switch_port_name) {
```
FelixMcFelix, line 481: You could use `HashMap::entry` with `Entry::or_insert_with` to be a bit more succint, but I suppose you'd need to clone `switch_port_name` unconditionally.

FelixMcFelix (approved):
Thanks, this looks good with one optional nit.

nexus/src/app/sagas/switch_port_settings_common.rs
rcgoodfellow: Yeah I think this is the equivalent using the `Entity::or_*` API

```rust
bgp_peer_configs.entry(switch_port_name.clone()).or_default().push(bpc);
```

And while that is more concise, I would say it's bordering on terse/dense, and it's less clear what the intent is. It's also very possible that I'm too dense to understand how to use this API in a clear way 😆.

Issue #4621, opened 2023-12-06
500 error when attempting to list uninitialized sleds


askfongjojo:
I ran into this on rack2 when invoking the Nexus API for getting a list of uninitialized sleds:
```
00:32:34.861Z INFO 65a11c18-7f59-41ac-b9e7-680627f996e7 (dropshot_external): request completed
    error_message_external = Internal Server Error
    error_message_internal = hit limit of 50 records while loading collection
    file = /home/build/.cargo/git/checkouts/dropshot-a4a923d29dccc492/ff87a01/dropshot/src/server.rs:841
    latency_us = 67030
    local_addr = 172.30.2.5:443
    method = GET
    remote_addr = 172.20.17.42:53378
    req_id = 61f13361-3699-4d64-9af9-308ded641a60
    response_code = 500
    uri = /v1/system/hardware/uninitialized-sleds
```

I'm not entirely sure what the offending table is (definitely not the sled table which has fewer than 60 rows). This is the content of the `inv_service_processor` table
```
root@[fd00:1122:3344:105::3]:32221/omicron> select * from inv_service_processor order by sp_type, sp_slot;
           inv_collection_id           |           hw_baseboard_id            |        time_collected         |                source                | sp_type | sp_slot | baseboard_revision | hubris_archive_id | power_state
---------------------------------------+--------------------------------------+-------------------------------+--------------------------------------+---------+---------+--------------------+-------------------+--------------
  a15e1a80-d9aa-4679-ba21-896ba86d9de8 | c6313404-bde5-42f8-a884-9de41c06843f | 2023-11-30 18:51:35.736285+00 | http://[fd00:1122:3344:10b::2]:12225 | sled    |       8 |                  6 | 62882ff9ad414aae  | A0
  5247e759-2478-4be8-b8ed-795e1d2ed15c | c6313404-bde5-42f8-a884-9de41c06843f | 2023-12-06 01:00:25.684695+00 | http://[fd00:1122:3344:10b::2]:12225 | sled    |       8 |                  6 | 62882ff9ad414aae  | A0
  68b2d1c0-248d-4f00-9a20-a3153a20c47f | c6313404-bde5-42f8-a884-9de41c06843f | 2023-12-06 00:59:43.813077+00 | http://[fd00:1122:3344:10b::2]:12225 | sled    |       8 |                  6 | 62882ff9ad414aae  | A0
[...]

PR #4631, opened 2023-12-06
[omdb] typo fix in error message

jgallagher:
Noticed this while investigating #4621:

```
warning: 16 collection errorwere s reported above
```


davepacheco (approved):
Eek!  Thanks.

Issue #4629, opened 2023-12-06
inventory fetch should query database in batches


davepacheco:
For context, see #4621.  The list-uninitialized-sleds code was hardcoding a database limit of "50" to read the latest inventory collection, but that isn't large enough for even dogfood, let alone a full rack.  It's easy to bump the limit (and we will for now -- see #4626), but the required limit could change in the future as we build out more inventory.

At the very least, we should write a test that will fail if the limit that we use isn't enough to read a full rack's standard inventory.  But with #4632 we have a relatively easy way to query in batches and we should use that instead.

davepacheco:
## Background

The inventory process generates a _collection_ that describes everything in the system.  That starts with baseboards and includes cabooses and other data about fixed software components (RoT slots, SP slots, etc.).

When some code wants to use inventory data, it reads a complete collection.  This involves reading a bunch of rows from a bunch of different inventory tables.  In general, when querying database tables, we use a limit because lots of bad things can happen if the query returns a lot more rows than we expected.

## Limits, pagination

What limit should we use?  There are two options: use a limit large enough to read everything in one query (per table) or pick any old limit and implement client-side pagination (by "client-side" here, I mean "database client-side", which is Nexus).

Pagination sounds like the obvious answer.  I have generally avoided doing client-side pagination within Nexus but not for great reasons.  I don't think we have a facility for doing this generally.  I'm not sure how easy it would be to build.  And it's not great to copy the (non-trivial) boilerplate to every place that needs it.  Here, my assumption was that if these collections actually get large, we'll probably want to re-think whether it even makes sense to read the entire collection in one go.

Plus, for the foreseeable future of just one rack, it's hard to imagine any of these tables containing so much data that we actually _need_ to paginate it.  Even if we implemented pagination, if we chose a typical limit of say 1000 records, it seems likely that the code would only ever make one query.  I figured a pragmatic approach would be: assume that will be the case (by not implementing pagination but checking if we hit the limit) and fail gracefully if that turns out to be wrong.  That's how we got here.  

**The crux of the problem right now is: this limit is currently chosen by app-level code paths, and that if the limit is too small, we won't find out until runtime.**

## Solution: commonize and test the limit

A good bang-for-the-buck improvement would be:

- Move the limit selection to a common code path.  (Example: make the `limit` argument to the `inventory_read_collection` family of functions _optional_.  Have the functions themselves choose a limit we think is big enough for a full rack.)
[...]

PR #4632, opened 2023-12-06
add a helper for querying the database in batches

davepacheco:
I've wanted this in lots of places and worked around it usually by setting a "high enough" limit and checking whether it had been reached.  See #4629 for an example.

I thought this was going to be a pain because I wasn't sure how to express this at the Diesel level.  @smklein had the idea to do this at the `DataPageParams` level, which makes a lot more sense anyway since it lets us reuse existing "list" functions.  I had wanted to write a function to drive the whole operation, but expressing the consumer's "list" operation as a closure ran into lifetime issues we couldn't figure out how to solve.

This approach leaves the loop control flow in the hands of the caller but tries to make it easy to use right and hard to use wrong.

I've only converted one spot so far, which is one that was already doing the paginated queries "by hand".  So this particular PR is kind of a no-op but adds the function so that we can convert more in the future.

### DIFF
--- a/nexus/db-queries/src/db/pagination.rs
+++ b/nexus/db-queries/src/db/pagination.rs
@@ -16,6 +16,7 @@ use diesel::AppearsOnTable;
 use diesel::Column;
 use diesel::{ExpressionMethods, QueryDsl};
 use omicron_common::api::external::DataPageParams;
+use std::num::NonZeroU32;
 
 // Shorthand alias for "the SQL type of the whole table".
 type TableSqlType<T> = <T as AsQuery>::SqlType;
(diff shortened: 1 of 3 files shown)


nexus/db-queries/src/db/pagination.rs
```diff
@@ -169,6 +170,145 @@ where
...
+/// ## Design notes
+///
+/// The separation of `Paginator` and `PaginatorHelper` is aimed at making it
+/// harder to misuse this interface.  We could skip the helper altogether and
+/// just have `Paginator::next()` return the DatePageParams directly.  But you'd
+/// still need a `Paginator::found_batch()`.  And it would be easy to forget to
+/// call this, leading to an infinite loop at runtime.  To avoid this mistake,
+/// `Paginator::next()` consumes `self`.  You can't get another `Paginator` back
+/// until you use `PaginatorHelper::found_batch()`.  That also consumes `self`
+/// so that you can't keep using the old `DataPageParams`.
```
jgallagher, line 237: I really like this `Design Notes` section :+1:

jgallagher (approved):
Looks great!

PR #4636, opened 2023-12-06
[sled-agent] address EarlyNetworkConfig issues from 20231130 dogfood mupdate

sunshowers:
* Log deserialization as both v1 and v0 failing.
* If an error occurs with deserialization as both v1 and v0, then preferentially return the error produced by v1 rather than v0.
* Add tests for bootstore blobs from earlier versions. If more fields are added in the future, this test should catch any issues that occur.


sled-agent/tests/integration_tests/early_network.rs
```diff
@@ -0,0 +1,154 @@
...
+    EarlyNetworkConfig, EarlyNetworkConfigBody,
+};
+use omicron_test_utils::dev::test_setup_log;
+
+const BLOB_PATH: &str = "tests/data/early_network_blobs.txt";
+
+/// Test that previous and current versions of `EarlyNetworkConfig` blobs
+/// deserialize correctly.
+#[test]
+fn early_network_blobs_deserialize() {
```
rcgoodfellow, line 27: FWIW: for little parsers like this, I really like using [nom](https://github.com/rust-bakery/nom).

sunshowers: Ah definitely if it's a bit more complex than this. In this case I figured it was just splitting by comma.

PR #4648, opened 2023-12-07
add --include-deleted switch to `omdb db instances`

### LINKED ISSUE #4647: want `omdb db instances` to exclude deleted instances by default
By default, the `instances` command includes soft-deleted instances, which currently don't get garbage-collected. This more or less requires users to add the `--fetch-limit` switch when invoking `omdb` (and requires the user to guess what fetch limit will be high enough to include all instances to make sure the instances of interest are in the command output).

I would wager (at least based on my own experience) that most of the time, users of `omdb db instances` are interested in instances that haven't been deleted yet (e.g. to find the sled/VMM that are hosting a running instance). If that's so, we can probably fix this paper cut by making `omdb` exclude deleted instances by default (while adding an option to include them if that's what the user really wants).

gjcolombo:
By default `omdb db instances` includes all instances, including all deleted instances. Since deleted instances don't currently get garbage-collected, this adds a lot of visual clutter to the command output and more or less requires the use of the `--fetch-limit` switch to ensure that the instance of interest actually appears in the command output.

Assume that more often than not the instance of interest is one that hasn't yet been deleted, make OMDB default to filtering out soft-deleted instances, and add an `--include-deleted` switch to revert to the old behavior if it's desired.

Tested on the dogfood rack, verifying in particular that if I wanted to find all instances matching a specific name pattern, including deleted instances required me to add `--fetch-limit 3750` to get them all, whereas excluding them (by running the command with no special arguments) returned all the instances of interest with the default fetch limit.

Fixes #4647.

### DIFF
--- a/dev-tools/omdb/src/bin/omdb/db.rs
+++ b/dev-tools/omdb/src/bin/omdb/db.rs
@@ -142,7 +142,7 @@ pub struct DbArgs {
 
 #[derive(Debug, Args)]
 pub struct DbFetchOptions {
-    /// The maximum number of rows a DB query should be allowed to try to fetch.
+    /// limit to apply to queries that fetch rows
     #[clap(
         long = "fetch-limit",
         default_value_t = NonZeroU32::new(500).unwrap(),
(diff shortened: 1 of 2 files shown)


luqmana:
Looks good but seems like this would be a useful flag & behaviour for all/most `db` subcommands? In which case it'd be worth dealing with it uniformly

gjcolombo:
> Looks good but seems like this would be a useful flag & behaviour for all/most `db` subcommands? In which case it'd be worth dealing with it uniformly

Good call. Fixed in 3f3b7c6 by making `--include-deleted` into a top-level fetch option (similar to the fetch limit) and applying it wherever it made sense and was straightforward to do so.

luqmana (approved):
Thanks!

PR #4654, opened 2023-12-08
Explicitly use the alloc op context for dpd nat updates

zephraph:
There was an issue in main where deleting an instance failed w/ a 403. Digging in further it looked like the 403 was coming from the instance delete saga, specifically the `InstanceDeleteNat` step that was added in #4610. It turns out that called `instance_delete_dpd_config` which called `boundary_switches` with the normal opcontext. Boundary switches calls `list_switch_ports_with_uplinks` and that requires fleet level permissions. 

The fix here was to pass the `opcontext_alloc` which gives higher level permissions down the `boundary_switches` call chain. 

I believe that we should _ultimately_ try to figure out some way to encode our authz state into the type system such that you can't just pass down a generic opcontext. I'm not exactly sure _how_ we would do that, but it seems worth pursuing.

### DIFF
--- a/nexus/src/app/instance_network.rs
+++ b/nexus/src/app/instance_network.rs
@@ -438,6 +438,7 @@ impl super::Nexus {
     pub(crate) async fn instance_delete_dpd_config(
         &self,
         opctx: &OpContext,
+        opctx_alloc: &OpContext,
         authz_instance: &authz::Instance,
     ) -> Result<(), Error> {
         let log = &self.log;
@@ -450,8 +451,6 @@ impl super::Nexus {
             .db_datastore
             .instance_lookup_external_ips(opctx, instance_id)
             .await?;
-
-        let boundary_switches = self.boundary_switches(opctx).await?;
 
         let mut errors = vec![];
         for entry in external_ips {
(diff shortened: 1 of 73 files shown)


zephraph:
I think there's a test failure here. If anyone feels like getting this over the line, please, be my guest.

zephraph:
@jmpesp I appreciate you!

nexus/src/app/instance_network.rs
```diff
@@ -438,6 +438,7 @@ impl super::Nexus {
     pub(crate) async fn instance_delete_dpd_config(
         &self,
         opctx: &OpContext,
+        opctx_alloc: &OpContext,
```
smklein, line 441: I think the usage of this `opctx_alloc` is correct - as in, it appears to be the right opctx we should use -- but this is a method of the `Nexus ` struct which contains the `opctx_alloc` field already (see: `nexus/src/app/mod.rs` for the definition).

Could we just access `self.opctx_alloc`? Seems weird to have it be passed as a parameter, if we don't really want the caller to customize it.

zephraph: Oh! Yes, that would make sense.

smklein (approved):
LGTM, only caveat being "maybe we don't pass the opctx as an argument". But the test makes sense!

zephraph:
I updated to used the recommended approach (which I just realized was also done in other places).

PR #4669, opened 2023-12-11
[nexus] Allow silo admins to upload new certs

### LINKED ISSUE #4532: silo admin cannot manage TLS certificates (but is supposed to be able to)
@augustuswm reported trying to update a rack's TLS certificates and getting an unepected 403.  Details about the user:

```
oxide --insecure current-user view
success
CurrentUser {
    display_name: "augustus@oxidecomputer.com",
    id: 8f7df011-ef5f-41c1-8f7a-c87dd90f723a,
    silo_id: a705588a-3e62-4c19-a227-c073476c33fc,
    silo_name: Name(
        "sandbox",
    ),
}

oxide --insecure current-user groups
Group {
    display_name: "all",
    id: 61b8fdd3-b32a-4d16-be6b-81633ad9c657,
    silo_id: a705588a-3e62-4c19-a227-c073476c33fc,
}

oxide --insecure silo policy view --silo a705588a-3e62-4c19-a227-c073476c33fc
success
SiloRolePolicy {
    role_assignments: [
        SiloRoleRoleAssignment {
            identity_id: 61b8fdd3-b32a-4d16-be6b-81633ad9c657,
            identity_type: SiloGroup,
            role_name: Admin,
        },
        SiloRoleRoleAssignment {
            identity_id: 8f7df011-ef5f-41c1-8f7a-c87dd90f723a,
            identity_type: SiloUser,
            role_name: Admin,
        },
    ],
}
```

He expected (correctly) that Silo Administrators are supposed to be able to modify TLS certificates.  And indeed, the authz check for creating a certificate checks the `CreateChild` action on the `SiloCertificateList`:
https://github.com/oxidecomputer/omicron/blob/e37f3d003fd11571d296db0fe6217d768a5a1341/nexus/db-queries/src/db/datastore/certificate.rs#L43

and the Oso policy file grants the corresponding permission to Silo admins:
https://github.com/oxidecomputer/omicron/blob/e37f3d003fd11571d296db0fe6217d768a5a1341/nexus/db-queries/src/authz/omicron.polar#L426-L431

The failure was actually here:

```
19:10:22.371Z DEBG 1cfdb5b6-e568-436a-a85f-7fecf1b8eef2 (dropshot_external): roles
    actor_id = 8f7df011-ef5f-41c1-8f7a-c87dd90f723a
    authenticated = true
    local_addr = 172.30.2.6:443
    method = POST
    remote_addr = 75.72.239.35:51658
    req_id = 1a1f3341-f246-49b2-9476-7de61e34a4ca
    roles = RoleSet { roles: {} }
    uri = //v1/certificates
19:10:22.374Z DEBG 1cfdb5b6-e568-436a-a85f-7fecf1b8eef2 (dropshot_external): authorize result
    action = Read
    actor = Some(Actor::SiloUser { silo_user_id: 8f7df011-ef5f-41c1-8f7a-c87dd90f723a, silo_id: a705588a-3e62-4c19-a227-c073476c33fc, .. })
    actor_id = 8f7df011-ef5f-41c1-8f7a-c87dd90f723a
    authenticated = true
    local_addr = 172.30.2.6:443
    method = POST
    remote_addr = 75.72.239.35:51658
    req_id = 1a1f3341-f246-49b2-9476-7de61e34a4ca
    resource = DnsConfig
    result = Err(Forbidden)
    uri = //v1/certificates
19:10:22.374Z INFO 1cfdb5b6-e568-436a-a85f-7fecf1b8eef2 (dropshot_external): request completed
    error_message_external = Forbidden
    error_message_internal = Forbidden
    file = /home/build/.cargo/git/checkouts/dropshot-a4a923d29dccc492/fa728d0/dropshot/src/server.rs:841
    latency_us = 43717
    local_addr = 172.30.2.6:443
    method = POST
    remote_addr = 75.72.239.35:51658
    req_id = 1a1f3341-f246-49b2-9476-7de61e34a4ca
    response_code = 403
    uri = //v1/certificates
```

It's a failure to `read` the `DnsConfig`.  I expect this is because before we call into the datastore to create the certificate (which is the authz check I linked above), we first validate it, which requires fetching the list of external DNS zones to figure out what the silo's external DNS names are going to be:

https://github.com/oxidecomputer/omicron/blob/e37f3d003fd11571d296db0fe6217d768a5a1341/nexus/src/app/certificate.rs#L50-L64

that goes into `silo_fq_dns_names()`, which indeed winds up here, checking the "read" privilege on `DNS_CONFIG`:
https://github.com/oxidecomputer/omicron/blob/e37f3d003fd11571d296db0fe6217d768a5a1341/nexus/db-queries/src/db/datastore/dns.rs#L84

I suspect this was introduced among the recent changes to validate the TLS certificates better.  @jgallagher does this sound plausible?  It's hard because we don't have a lot of testing around minimum privileges -- see #1374 for more on that.

In terms of fixes: right now, we use one resource (`DnsConfig`) to cover all of the DNS data, including the list of external DNS zone names _and_ all the DNS names in both internal and external DNS.  While it seems reasonable to me to say that Silo Admins should be able to see the list of external DNS zone names (i.e., the DNS domains that have been delegated to the rack), it seems excessive to grant them permission to see all the DNS names in both internal and external DNS.  So one option would be to split this authz resource into two parts, grant privileges on one of them to "Silo Admin", and use that when listing external DNS zones.  That would solve this problem but doesn't feel particularly easy or clean.

Another option is to use the pattern we use elsewhere, where Nexus assumes a privileged, internal-only identity to fetch this information _on behalf_ of the authenticated user.  We'd want to be careful to make sure that we're not exposing anything we shouldn't by doing this.  (For example, if we decide the list of external DNS zone names _is_ privileged information, we probably wouldn't want to let you discover it by trying to upload certificates for various possible names and seeing if they get rejected?)  But my intuition is that a Silo Admin already needs to know the list of external DNS zone names in order to know what TLS certificates they need to provide (or what DNS names those certificates are valid for).

jgallagher:
The additional cert validation added in #4100 broke the ability for silo admins to upload new certs, because it introduced a call to fetch the rack DNS configuration (in order to assemble the FQDNs for the silo to check that the cert is valid for them). This PR fixes that by using an elevated internal privilege for that DNS config lookup.

Fixes #4532.

~~TODO before merging~~ Done:

- [x] ~~confirm fix on madrid (the integration test makes me think this is _probably_ not necessary, but I'd rather be sure)~~ Tested on madrid; was able to use the CLI `oxide certificate create` command while logged in as a silo admin, which 403'd prior to this PR

### DIFF
--- a/nexus/src/app/certificate.rs
+++ b/nexus/src/app/certificate.rs
@@ -6,6 +6,7 @@
 
 use crate::external_api::params;
 use crate::external_api::shared;
+use nexus_db_queries::authn;
 use nexus_db_queries::authz;
 use nexus_db_queries::context::OpContext;
 use nexus_db_queries::db;
@@ -20,6 +21,7 @@ use omicron_common::api::external::InternalContext;
 use omicron_common::api::external::ListResultVec;
 use omicron_common::api::external::NameOrId;
 use ref_cast::RefCast;
+use std::sync::Arc;
 use uuid::Uuid;
 
 impl super::Nexus {
@@ -47,8 +49,29 @@ impl super::Nexus {
             .silo_required()
             .internal_context("creating a Certificate")?;
 
-        let silo_fq_dns_names =
-            self.silo_fq_dns_names(opctx, authz_silo.id()).await?;
+        // The `opctx` we received is going to be checked for permission to
+        // create a cert below in `db_datastore.certificate_create`, but first
+        // we need to look up this silo's fully-qualified domain names in order
+        // to check that the cert we've been given is valid for this silo.
+        // Looking up DNS names requires reading the DNS configuration of the
+        // _rack_, which this user may not be able to do (even if they have
+        // permission to upload new certs, which almost certainly implies a
+        // silo-level admin. We'll construct a new opctx here to allow this DNS
+        // config lookup, because we believe it does not leak any information
+        // that a silo admin doesn't already know (the external DNS name(s) of
+        // the rack, which leads to their silo's DNS name(s)).
+        //
+        // See https://github.com/oxidecomputer/omicron/issues/4532 for
+        // additional background.
+        let silo_fq_dns_names = {
+            let dns_opctx = OpContext::for_background(
+                opctx.log.clone(),
+                Arc::clone(&self.authz),
+                authn::Context::internal_service_balancer(),
+                Arc::clone(self.datastore()),
+            );
+            self.silo_fq_dns_names(&dns_opctx, authz_silo.id()).await?
+        };
 
         let kind = params.service;
         let new_certificate = db::model::Certificate::new(
--- a/nexus/src/app/certificate.rs
+++ b/nexus/src/app/certificate.rs
@@ -6,7 +6,6 @@
 
 use crate::external_api::params;
 use crate::external_api::shared;
-use nexus_db_queries::authn;
 use nexus_db_queries::authz;
 use nexus_db_queries::context::OpContext;
 use nexus_db_queries::db;
@@ -21,7 +20,6 @@ use omicron_common::api::external::InternalContext;
 use omicron_common::api::external::ListResultVec;
 use omicron_common::api::external::NameOrId;
 use ref_cast::RefCast;
-use std::sync::Arc;
 use uuid::Uuid;
 
 impl super::Nexus {
@@ -56,22 +54,15 @@ impl super::Nexus {
         // Looking up DNS names requires reading the DNS configuration of the
         // _rack_, which this user may not be able to do (even if they have
         // permission to upload new certs, which almost certainly implies a
-        // silo-level admin. We'll construct a new opctx here to allow this DNS
-        // config lookup, because we believe it does not leak any information
-        // that a silo admin doesn't already know (the external DNS name(s) of
-        // the rack, which leads to their silo's DNS name(s)).
-        //
-        // See https://github.com/oxidecomputer/omicron/issues/4532 for
-        // additional background.
-        let silo_fq_dns_names = {
-            let dns_opctx = OpContext::for_background(
-                opctx.log.clone(),
-                Arc::clone(&self.authz),
-                authn::Context::internal_service_balancer(),
-                Arc::clone(self.datastore()),
-            );
-            self.silo_fq_dns_names(&dns_opctx, authz_silo.id()).await?
-        };
+        // silo-level admin. We'll use our `opctx_external_authn()` context,
+        // which is the same context used to create a silo. This is a higher
+        // privelege than the current user may have, but we believe it does not
+        // leak any information that a silo admin doesn't already know (the
+        // external DNS name(s) of the rack, which leads to their silo's DNS
+        // name(s)).
+        let silo_fq_dns_names = self
+            .silo_fq_dns_names(self.opctx_external_authn(), authz_silo.id())
+            .await?;
 
         let kind = params.service;
         let new_certificate = db::model::Certificate::new(
--- a/nexus/src/app/certificate.rs
+++ b/nexus/src/app/certificate.rs
@@ -54,9 +54,9 @@ impl super::Nexus {
         // Looking up DNS names requires reading the DNS configuration of the
         // _rack_, which this user may not be able to do (even if they have
         // permission to upload new certs, which almost certainly implies a
-        // silo-level admin. We'll use our `opctx_external_authn()` context,
+        // silo-level admin). We'll use our `opctx_external_authn()` context,
         // which is the same context used to create a silo. This is a higher
-        // privelege than the current user may have, but we believe it does not
+        // privilege than the current user may have, but we believe it does not
         // leak any information that a silo admin doesn't already know (the
         // external DNS name(s) of the rack, which leads to their silo's DNS
         // name(s)).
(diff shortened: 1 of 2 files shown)


nexus/tests/integration_tests/silos.rs
```diff
@@ -2437,3 +2440,85 @@ async fn check_fleet_privileges(
...
+// Test that a silo admin can create new certificates for their silo
+//
+// Internally, the certificate validation check requires the `authz::DNS_CONFIG`
+// resource (to check that the certificate is valid for
+// `{silo_name}.{external_dns_zone_name}`), which silo admins may not have. We
+// have to use an alternate, elevated context to perform that check, and this
+// test confirms we do so.
+//
+// Ensures no regression on
+// https://github.com/oxidecomputer/omicron/issues/4532.
```
davepacheco, line 2453: Personally I don't bother referencing past bugs -- I think it's enough to explain why this is useful/necessary for correctness.

nexus/src/app/certificate.rs
```diff
@@ -47,8 +49,29 @@ impl super::Nexus {
...
+        //
+        // See https://github.com/oxidecomputer/omicron/issues/4532 for
+        // additional background.
+        let silo_fq_dns_names = {
+            let dns_opctx = OpContext::for_background(
+                opctx.log.clone(),
+                Arc::clone(&self.authz),
+                authn::Context::internal_service_balancer(),
+                Arc::clone(self.datastore()),
+            );
```
davepacheco, line 72: I see that we've been a little inconsistent on how we create OpContexts that Nexus uses to do privileged things on behalf of unprivileged users.

- For the "instance allocator" and "external authn" OpContexts, we create them once at startup and store them into `Nexus`.  Callers can access the former directly as a field.  They can access the latter that way or using `self.opctx_external_authn()`.
- For the service balancer and internal API OpContexts, we have similar methods `self.opctx_for_service_balancer()` and `self.opctx_for_internal_api()`, but these construct the `OpContext` anew each time rather than just doing that once on startup.

I generally prefer the first approach but clearly it's common to do the latter too.  I was going to suggest _not_ doing this logic inline here, creating a helper instead, although maybe you did this on purpose to preserve the existing request log (in `opctx.log`)?

There's also the question of whether this is the _right_ OpContext.  This definitely _isn't_ the service balancer...but neither is the other user of this OpContext that I found.  In a sense, it really doesn't matter.  These internal identities only exist to limit the privileges of code paths.  You _could_ create a new identity here that can only read DNS and use that.  But it might be overkill to do that everywhere we need to use this pattern.

I see that we have an "internal-read" identity that only gets "fleet viewer", not "fleet admin".  I also see that we already have `Nexus.opctx_alloc` using that identity.  You could use that here and it'd be a little safer (fewer privileges) and also would avoid constructing a new OpContext inline with the request.  The only downside is that the label attached to the log for that one is "InstanceAllocator", which would be a little misleading.  You could also just go ahead and construct a new OpContext here using `authn::Context::internal_read()` instead, or add a helper to do it.

I don't have a strong feeling about all these options.  I would at least use the `internal_read()` identity because there's not much work and I don't think there's any downside.  As for whether to create a new identity, whether to create the OpContext inline or use a helper or use the existing one, these all have tradeoffs and one could go either way.

jgallagher: Summarizing an offline chat:

* We can't use "internal-read", because the DNS config policy only grants read to "fleet admin" (which "internal-read" isn't).
* Per [the comments on DnsConfig](https://github.com/oxidecomputer/omicron/blob/07eb7dafc20e35e44edf429fcbb759cbb33edd5f/nexus/db-queries/src/authz/omicron.polar#L355-L359), Nexus uses the "external-authenticator" role when creating silos, so it seems reasonable to use it here as well.

Changed in [fa74a7a](https://github.com/oxidecomputer/omicron/pull/4669/commits/fa74a7aa263a2ed092867b3e952c22300a7e05d5).

nexus/src/app/certificate.rs
```diff
@@ -47,8 +47,22 @@ impl super::Nexus {
...
-        let silo_fq_dns_names =
-            self.silo_fq_dns_names(opctx, authz_silo.id()).await?;
+        // The `opctx` we received is going to be checked for permission to
+        // create a cert below in `db_datastore.certificate_create`, but first
+        // we need to look up this silo's fully-qualified domain names in order
+        // to check that the cert we've been given is valid for this silo.
+        // Looking up DNS names requires reading the DNS configuration of the
+        // _rack_, which this user may not be able to do (even if they have
+        // permission to upload new certs, which almost certainly implies a
+        // silo-level admin. We'll use our `opctx_external_authn()` context,
```
davepacheco, line 57:
```suggestion
        // silo-level admin). We'll use our `opctx_external_authn()` context,
```

(sorry for missing this before)

nexus/src/app/certificate.rs
```diff
@@ -47,8 +47,22 @@ impl super::Nexus {
...
+        // which is the same context used to create a silo. This is a higher
+        // privelege than the current user may have, but we believe it does not
```
davepacheco, line 59:
```suggestion
        // privilege than the current user may have, but we believe it does not
```

davepacheco (approved):

PR #4670, opened 2023-12-11
Oximeter self-stat tests don't need to be based on time

### LINKED ISSUE #4657: Test flake: oximeter-collector agent::tests::test_self_stat_collection_count
[Example](https://buildomat.eng.oxide.computer/wg/0/details/01HH5SD0AVG263G639APACANK1/T0SjTwswcNwb7xuW50RusoIcoGmCwX48vgzKmD2HP1omct7t/01HH5SE7HJGZJHG4169N5KKJP0#S4377)

bnaecker:
- Fixes #4657
- Oximeter self stat tests run a few collections, and verify that the relevant counters keep track of those. Previously, that was based on _time_, advancing time forward until the expected number of collections actually occur. There's nothing really guaranteeing that those collections do occur, since the tasks may not run exactly when we expect. This commit updates the tests so that the producer explicitly updates a shared counter every time they receive a request. The tests then reduce to asserting this counter equals the self-stat counter.

### DIFF
--- a/oximeter/collector/src/agent.rs
+++ b/oximeter/collector/src/agent.rs
@@ -654,6 +654,8 @@ mod tests {
     use std::net::Ipv6Addr;
     use std::net::SocketAddr;
     use std::net::SocketAddrV6;
+    use std::sync::atomic::AtomicU64;
+    use std::sync::atomic::Ordering;
     use std::time::Duration;
     use tokio::sync::oneshot;
     use tokio::time::Instant;
@@ -667,7 +669,8 @@ mod tests {
     // timers complete as expected.
     const TICK_INTERVAL: Duration = Duration::from_millis(10);
 
-    // Total number of collection attempts.
+    // Total number of collection attempts, and the expected number of
+    // collections which fail in the "unreachability" test below.
     const N_COLLECTIONS: u64 = 5;
 
     // Period these tests wait using `tokio::time::advance()` before checking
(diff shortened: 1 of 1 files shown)


sunshowers (approved):
Thanks for fixing this!

I'm wondering whether we want to provide the time source as a dependency rather than directly relying on system time. For example, providing a stream of ticks as an argument. Then that could be standard time for production and a custom, manually controlled impl for tests.

bnaecker:
Thanks for the review @sunshowers!

> I'm wondering whether we want to provide the time source as a dependency rather than directly relying on system time. For example, providing a stream of ticks as an argument. Then that could be standard time for production and a custom, manually controlled impl for tests.

I'm not sure I follow this, sorry! This stream-of-ticks is an argument to what, exactly? It seems like there is a test-specific notion of time, since we use `tokio::time::{pause,advance}` and friends in tests, but that may be different than what you're suggesting.

sunshowers:
> This stream-of-ticks is an argument to what, exactly?

To the service sending oximeter events. I assumed the events were sent on a regular schedule.

The tokio::time stuff is cool but it's a pretty big hammer, I think.

bnaecker:
> To the service sending oximeter events. I assumed the events were sent on a regular schedule.

Ok, I think what you're proposing is that instead of waiting for a timer internally, the collection task would wait on a channel onto which an external entity could send those notifications to collect. Is that accurate?

If so, I think we have something similar already: The tasks wait on a timer expiration, and _also_ a channel on which we can request explicit collections. I could have updated this test to use that directly, I think I just forgot about it until now.

sunshowers:
> Ok, I think what you're proposing is that instead of waiting for a timer internally, the collection task would wait on a channel onto which an external entity could send those notifications to collect. Is that accurate?

Yes.

> If so, I think we have something similar already: The tasks wait on a timer expiration, and also a channel on which we can request explicit collections. I could have updated this test to use that directly, I think I just forgot about it until now.

Ahh. Yeah, maybe the test could just disable timer-based collection (if that makes sense) and always rely on explicit collections. Or it could model the timer-based collection as another stream or channel, if you want to include that as part of the test.

PR #4671, opened 2023-12-11
Restore larger max request body size

jmpesp:
76b835d5d68657265564f40a4667e8ffc7026091 removed the dropshot configuration that bumped the sled agent's max request size to 1M. The sled agent needs to be able to handler larger POSTs, so restore that parameter.

### DIFF
--- a/sled-agent/src/server.rs
+++ b/sled-agent/src/server.rs
@@ -72,6 +72,10 @@ impl Server {
 
         let dropshot_config = dropshot::ConfigDropshot {
             bind_address: SocketAddr::V6(sled_address),
+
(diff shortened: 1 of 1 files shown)


sunshowers:
Yeah, it's configured to be 2 GiB here: https://github.com/oxidecomputer/omicron/commit/76b835d5d68657265564f40a4667e8ffc7026091#diff-586795faca15dc17935afd21020891e71fba46678340f7f08c50082bacc9ff3f

jmpesp:
Ah, I missed this due to how I deploy and test builds - which is _not_ with those toml files! Sorry for all the noise folks!

PR #4672, opened 2023-12-11
inventory builder could better distinguish runtime errors from API misuse

davepacheco:
The inventory builder has two forms of error reporting: `found_error()` is used to report things that are really part of the collection itself: a failure to collect some important information, or unexpected data, etc.  Returned errors are used to signal to the caller that they've done something wrong.  (These could as well be panics, except that the program state is not so wrong that it's dangerous to proceed.)  This PR changes these to use simple newtypes.  More could probably be done here (maybe in terms of making some of the `CollectorBug`s actual compile errors) but I think this is a useful step.

### DIFF
--- a/nexus/inventory/src/builder.rs
+++ b/nexus/inventory/src/builder.rs
@@ -27,7 +27,39 @@ use nexus_types::inventory::ServiceProcessor;
 use std::collections::BTreeMap;
 use std::collections::BTreeSet;
 use std::sync::Arc;
+use thiserror::Error;
 use uuid::Uuid;
+
+/// Describes an operational error encountered during the collection process
+///
+/// Examples include a down MGS instance, failure to parse a response from some
+/// other service, etc.  We currently don't need to distinguish these
+/// programmatically.
+#[derive(Debug, Error)]
+pub struct InventoryError(#[from] anyhow::Error);
+
+impl std::fmt::Display for InventoryError {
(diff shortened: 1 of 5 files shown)


jgallagher (approved):

nexus/inventory/src/builder.rs
```diff
@@ -351,11 +379,12 @@ impl CollectionBuilder {
...
-    /// process (e.g., a down MGS instance).  It's not intended for mis-uses of
-    /// this API, which are conveyed instead through returned errors (and should
-    /// probably cause the caller to stop collection altogether).
-    pub fn found_error(&mut self, error: anyhow::Error) {
+    /// See [`InventoryError`] for more on what kinds of errors are reported
+    /// this way.  These errors are stored as part of the collection so that
+    /// future readers can see what problems might make the collection
+    /// incomplete.  By contrast, [`CollectorBug`]s are not reported and stored
+    /// this way.
+    pub fn found_error(&mut self, error: InventoryError) {
```
jgallagher, line 387: rustdoc is complaining about the doc links to private types; I'm surprised clippy isn't likewise complaining about `InventoryError`'s presence here in a `pub` API. Maybe make this (and the error types) `pub(crate)`?

davepacheco: Thanks for looking.  I wound up exposing these instead.  At this point I think it's useful to be able to fully use a `CollectionBuilder` from the outside (for tests, if nothing else).

PR #4534, opened 2023-11-21
[sled-agent] Oximeter self-assembling zone

karencfv:
Related https://github.com/oxidecomputer/omicron/issues/1898

Closes: https://github.com/oxidecomputer/omicron/issues/2883

Issue #1898, opened 2022-11-01
Tracking issue for "Self-assembling Zones"


smklein:
This topic was discussed in the latter half of https://drive.google.com/file/d/185bFxdvDo_1aA5-T5ywp9t3ZCJbc52B7/view , but I'll briefly summarize here:

- Currently, after sled agent boots zones, they require additional configuration before they can be made usable.
- This configuration includes: Creating IP addresses, setting up routes, setting SMF configuration options, etc.
- These configuration changes are made by invoking `zlogin` to access the zone.
- Across reboot, a portion of this information (namely the SMF properties) is saved into `/var/oxide`. This instructs the sled agent how to relaunch zones across reboots.

Instead, we should do the following:
- Make zones "self-assembling" as much as possible.
  - *Before* the zone is booted, the sled agent should inject a file into `/var/svc/profile/site.xml` to load an [SMF profile](https://illumos.org/man/7/smf) with the run-time parameters
  - The zone should rely on services starting from the `manifest-import` service
  - These dynamic parameters can then be processed by a "method script"  inside the zone
  - As a result, we should not need to call `zlogin` with a set of commands to run inside the zone

What are the advantages of doing this?
- **Less likely to have race conditions with SMF**. We have had issues in the past, like https://github.com/oxidecomputer/omicron/issues/1115 , related to asynchronous setting of SMF properties relative to launching the service. Additionally, we've dealt with race conditions between the sled agent and the`manifest-import` service. With this new mechanism, sled agent will set all configuration *before* starting the zone, so the ownership of the zone is more clear.
- **Less configuration *outside* the zone to re-launch the zone**. Management of zones across reboots is a fair bit easier, no longer requiring auxiliary information in `/var/oxide` to know how to re-launch the zone.
- **Less calls to zlogin**. This should be a minor efficiency boost.
- **Sled agent should become simpler**. Hopefully. Let's see!

Tracking pieces:
- [x] Add support for dynamically generating an SMF profile (added within https://github.com/oxidecomputer/omicron/pull/1902)
- [x] Create zone network setup service so individual zones don't do this themselves https://github.com/oxidecomputer/omicron/pull/4677
- [x] Generate SMF profiles (and use method scripts) to launch the following services: 
  - [x] Storage
    - [x] Cockroach: https://github.com/oxidecomputer/omicron/pull/1902
    - [x] Clickhouse: https://github.com/oxidecomputer/omicron/pull/1902
[...]

PR #4677, opened 2023-12-12
Zone network configuration service

karencfv:
As part of the work for [self assembling zones](https://github.com/oxidecomputer/omicron/issues/1898), it was [suggested](https://github.com/oxidecomputer/omicron/pull/4534#issuecomment-1821457494) to break the network configuration out into a separate service.

## Implementation

This PR introduces a new SMF service `oxide/zone-network-setup`, which sets up the common initial zone networking configuration for each self assembled zone. 

Each of the "self assembled zone" services will now depend on this new service to run, and all properties relating to zone network configuration have been removed from these services.

The executable which does the actual zone networking setup, is built as a tiny CLI. It takes advantage of clap's parsing validation to make sure we have all of the properties present, and in the format they are intended to be.

## Caveats

There are two remaining self assembled zones that don't depend on this new service yet (crucible and crucible-pantry). As these two zones need coordinated PRs with the crucible repo, I'd like to implement these in a follow up PR once this one is approved and merged.


smf/cockroachdb/method_script.sh
```diff
@@ -18,15 +18,17 @@ if [[ $DATALINK == unknown ]] || [[ $GATEWAY == unknown ]]; then
 fi
 
 # TODO remove when https://github.com/oxidecomputer/stlouis/issues/435 is addressed
```
karencfv, line 20: @jmpesp I see this issue has been closed, are we OK to remove this part now?

jclulow:
I have had a quick squiz at the PR, and I think adding a Rust program to do the work is indeed much better than a shell script!  Thanks for digging into this.  Before looking too closely at the code, I have some architectural thoughts.

### Separate SMF service

I think we should do the networking setup in a wholly separate **[SMF](https://illumos.org/man/7/smf) service**.  In the PR as it stands, we've replaced some direct **ipadm** invocations in the method script for the Cockroach DB SMF service -- but the moving parts are still essentially a part of the Cockroach DB service.

What I would sort of expect is to add a new SMF service, with just a default instance; e.g., **svc:/oxide/network-setup:default**.  This would be enabled by default.  It should be of the **transient** duration (see [svc.startd(8)](https://illumos.org/man/8/svc.startd)), exiting successfully when it completes setup.  It should be idempotent, so that if it is restarted (whether after interruption or not) it does the right thing.  It probably does not _need_ to tear things down in a **stop** method, and it may not presently need to provide a functional **refresh** method at this time.

To the extent that we need to communicate any information to this service from outside (e.g., addresses or subnets or interface names) we would do so via properties in the default service instance.  The name, value, type, and behaviour of each of those properties (and indeed the FMRI of the whole service instance) would essentially become a cross-version stable interface; if sled agent needs to populate them, we should commit to them so that the version of sled agent and the version of the zone image aren't locked together.

I suspect this new service would depend on the existing **svc:/network/physical** service that ships with the operating system.  We would probably then make the OS milestone service, **svc:/milestone/network**, depend on our new service here.  Then we would redo any other services, like Cockroach DB, that probably don't want to start until the network is configured, such that they depend on the network milestone.  That's [roughly what we do](https://github.com/illumos/metadata-agent/blob/master/metadata.xml#L20-L46) with a similar piece of software, the illumos metadata agent, except in a different context -- the metadata agent is a replacement for **cloud-init**, and runs in the global zone of virtualised guests on cloud platforms.  In this case, the network setup service is running inside a non-global zone on top of an Omicron-managed system. 

### No shell at all

I don't think we need to use a shell script as the start/stop method for the new service.  I would restructure the new Rust program to be invoked as the start method directly, rather than invoked several times from a new shell script.  I also wouldn't do any argument construction or parsing, except for perhaps passing the method name (see **%m** in **Method Tokens** of [smf_method(7)](https://illumos.org/man/7/smf_method#method-tokens)), but instead just use the [**smf**](https://github.com/illumos/smf-rs) crate (which uses [libscf(3LIB)](https://illumos.org/man/3LIB/libscf) directly) to access any properties that we need.

---

There is, I suspect, a lot to digest here.  I've tried to link some manual pages that contain some of the details.  When reading and writing SMF manifests, it may also be worth looking at the [service_bundle(5)](https://illumos.org/man/5/service_bundle) XML DTD, which is shipped in the OS and also [in the source base](https://github.com/illumos/illumos-gate/blob/master/usr/src/cmd/svc/dtd/service_bundle.dtd.1); it contains a lot of comments about the structure of those files.

Definitely happy to answer any questions or talk more about this whenever it would be helpful!  Let me know!

karencfv:
> I think we should do the networking setup in a wholly separate [SMF](https://illumos.org/man/7/smf) service. In the PR as it stands, we've replaced some direct ipadm invocations in the method script for the Cockroach DB SMF service -- but the moving parts are still essentially a part of the Cockroach DB service.

@jclulow, as I understand it, as it stands today, a new SMF service would have to be created by RSS. This would mean that we cannot have a new service running on the current racks until zones are provisioned by Nexus instead of RSS :disappointed:. Hopefully I'm wrong about this, WDYT?

jclulow:
I don't know much about RSS, but an SMF service is defined by an XML file that can just be shipped in the correct place in the zone image.  If you put it in `/lib/svc/manifest/site` it ought, I expect, to be imported early enough to impose itself as a dependency of the network milestone.  I believe manifests in that directory are imported by the *early-manifest-import* service at zone boot.

jgallagher:
> @jclulow, as I understand it, as it stands today, a new SMF service would have to be created by RSS. This would mean that we cannot have a new service running on the current racks until zones are provisioned by Nexus instead of RSS 😞. Hopefully I'm wrong about this, WDYT?

There's a bit of term overloading here - we can't currently start new _control plane_ services / service zones on deployed racks because only RSS sets them up today, but SMF services are a lower level thing. Sled agent already configures and starts various SMF services within many of the zones it starts (e.g., all the services within the switch zone, where the list of switch zone SMF services has grown over time without having to deal with RSS or control plane service management). I think the hard line is probably "does sled-agent have all the information it needs to configure this {SMF,control plane} service", which is usually/always true for "SMF service" and false for "control plane service".

karencfv:
Ah, perfect! Thanks for the clarification @jclulow @jgallagher  :bowing_woman:

karencfv:
@jclulow thanks a bunch for your help! I've updated the PR to reflect the feedback you gave me.

If this architecture makes more sense, I'll update the rest of the self assembled zones to follow this pattern and move the PR out of draft mode.

Let me know what you think :)

smf/zone-network-setup/manifest.xml
```diff
@@ -0,0 +1,54 @@
...
+  </dependency>
+
+  <dependency name='multi_user' grouping='require_all' restart_on='none'
+    type='service'>
+  <service_fmri value='svc:/milestone/multi-user:default' />
+  </dependency>
+
+  <!-- TODO: Make svc:/milestone/network depend on this service? -->
+
+<!--TODO: Do the actual thing we want to do-->
```
karencfv, line 22: Gargh, I'll get rid of this line. Just want to wait for CI to finish

sled-agent/src/services.rs
```diff
@@ -1376,6 +1376,25 @@ impl ServiceManager {
             .add_instance(ServiceInstanceBuilder::new("default")))
     }
 
+    async fn zone_network_setup_install(
```
citrus-it, line 1379: Is there any reason this needs to be an async function?

smf/zone-network-setup/manifest.xml
```diff
@@ -0,0 +1,46 @@
...
+    timeout_seconds='0' />
+  
+  <property_group name='startd' type='framework'>
+    <propval name='duration' type='astring' value='transient' />
+  </property_group>
+
+  <property_group name='config' type='application'>
+    <propval name='datalink' type='astring' value='unknown' />
+    <propval name='gateway' type='astring' value='unknown' />
+    <propval name='listen_addr' type='astring' value='unknown' />
```
citrus-it, line 31: I think I'd be inclined to find a different name for this property as `listen_addr` is a property that the other services have and configures the address that they listen on. Perhaps just `static_addr` - it could even be a list of static addresses but I don't think there are any services that need that yet.

smf/zone-network-setup/method_script.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
```
citrus-it, line 1: I would just not use a shell script for the service method (if you do, it should be in `ksh`, not `bash` since it is run from SMF fairly early).

I would use property tokens in the start method in the manifest, that is, the start method would look something like the following, then you don't need this script at all.

```xml
    exec='/opt/oxide/zone-network-setup/bin/zone-networking -d %{config/datalink} -l %{config/listen_addr} -g %{config/gateway}'
    timeout_seconds='0' />
```

citrus-it (changes requested):
Thanks for doing this, it's going to make things a lot cleaner around self-assembly zones.
I have a few comments, some of which are just suggestions (Github doesn't make it easy to differentiate). If anything isn't clear or you don't agree, please let me know.

(and apologies for you now having to re-base this on top of my disable-sshd PR that just merged!)

sled-agent/src/services.rs
karencfv: Not really, I'll change that

smf/zone-network-setup/manifest.xml
karencfv: Ah yeah, that property name makes much more sense. I'm not sure I'd like to have a list of addresses yet if none of the services require this. We could modify it later if the need arises. What do you think?

karencfv:
Thanks for the review @citrus-it ! I think I've addressed all of your comments, let me know what you think :)

package-manifest.toml
```diff
@@ -613,3 +613,16 @@ source.packages = [
...
+
+[package.zone-network-setup]
+service_name = "zone-network-setup"
+only_for_targets.image = "standard"
+source.type = "local"
+source.rust.binary_names = ["zone-networking"]
+source.rust.release = true
+source.paths = [
+    { from = "smf/zone-network-setup/manifest.xml", to = "/var/svc/manifest/site/zone-network-setup/manifest.xml" },
+    { from = "smf/zone-network-setup/method_script.sh", to = "/opt/oxide/lib/svc/manifest/zone-network-setup.sh" },
```
citrus-it, line 625: You need to drop this line now that there is no method script.

citrus-it:
> Thanks for the review @citrus-it ! I think I've addressed all of your comments, let me know what you think :)

This looks good, thanks. Just added a comment about the `package-manifest.toml`.

I think it's fine to just have a single static address for now, since there's no consumer for multiple addresses.

package-manifest.toml
karencfv: ha! whoops! forgot about that bit :sweat_smile:

citrus-it (commented):
Thanks for the updates, this is looking good to me.

karencfv:
Thanks for the review @citrus-it!

Tiny ping for @jclulow , @davepacheco @jgallagher or @smklein to review on the rust side of things, and I'll merge this :)

jclulow:
I will take a look soon!

karencfv:
heyo! :wave: 

Sorry for the pressure here I know there's so much going on! Would it be possible to get a review on the rust side of things? @citrus-it kindly went over the Illumos side of this PR, so there's less to review! 

The thing is, I need this approved so I can move on with the rest of the work for the self assembled zones.

smklein (approved):
Looks good to me!

sled-agent/src/services.rs
```diff
@@ -1419,6 +1419,25 @@ impl ServiceManager {
...
+        let mut config_builder = PropertyGroupBuilder::new("config");
+        config_builder = config_builder
+            .add_property("datalink", "astring", datalink)
+            .add_property("gateway", "astring", gateway)
+            .add_property("static_addr", "astring", static_addr);
+
+        Ok(ServiceBuilder::new("oxide/zone-network-setup")
+            .add_property_group(config_builder)
+            .add_instance(ServiceInstanceBuilder::new("default")))
+    }
```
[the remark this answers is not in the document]
karencfv: :bowing_woman:

karencfv:
Thanks everyone!

PR #4679, opened 2023-12-12
Fix oximeter test flake due to timestamp collision

### LINKED ISSUE #4678: Oximeter test fails locally due to successive `Utc::now()` having the same value
## Phenomenon

While debugging the test failure I got [here](https://github.com/oxidecomputer/omicron/pull/4261/checks?check_run_id=19537307726) after merging main into #4261, I found I could reproduce the failure locally (on a Mac) not only on that branch, but also on `main`, going all the way back to [9b666e73](https://github.com/oxidecomputer/omicron/commits/9b666e73dec06f2a645a714cdfbb21d63a2f3504), the merge commit for #4552.

I get the failure as long as any two of these lines are allowed to run.

https://github.com/oxidecomputer/omicron/blob/9e67e3b0d2c4e7333f35381e7e9d920e21f41851/oximeter/db/src/client.rs#L3351-L3367

The failure occurs when Oximeter tries to parse the following two JSON results (there's a `\n` in the middle) as a single JSON value.

```
[oximeter/db/src/model.rs:1486] line = "{\"timeseries_name\":\"foo:bar\",\"timeseries_key\":101,\"start_time\":\"2023-12-12 03:44:15.808208000\",\"timestamp\":\"2023-12-12 03:44:15.808210000\",\"bins\":[-128,0,1,2],\"counts\":[0,0,0,0]}\n{\"timeseries_name\":\"foo:bar\",\"timeseries_key\":101,\"start_time\":\"2023-12-12 03:44:15.808210000\",\"timestamp\":\"2023-12-12 03:44:15.808210000\",\"bins\":[],\"counts\":[]}\n"
```

## A theory

The test involves selecting by timestamp:

https://github.com/oxidecomputer/omicron/blob/9e67e3b0d2c4e7333f35381e7e9d920e21f41851/oximeter/db/src/client.rs#L3276

And I think these `Utc::now()` are coming out to have the same value when they're not supposed to, which means we get two results when we expect one. 

https://github.com/oxidecomputer/omicron/blob/9e67e3b0d2c4e7333f35381e7e9d920e21f41851/oximeter/db/src/client.rs#L3138-L3157

## Testing the theory

When I pull the times out and log them:

```diff
diff --git a/oximeter/db/src/client.rs b/oximeter/db/src/client.rs
index d295d0dcd..334d8979c 100644
--- a/oximeter/db/src/client.rs
+++ b/oximeter/db/src/client.rs
@@ -3145,12 +3145,15 @@ mod tests {
         serde_json::Value: From<T>,
     {
         let datum = Datum::from(hist);
-        let measurement = Measurement::new(Utc::now(), datum);
+        let t1 = Utc::now();
+        let measurement = Measurement::new(t1, datum);
+        let t2 = Utc::now();
         let missing_datum = Datum::Missing(
-            MissingDatum::new(measurement.datum_type(), Some(Utc::now()))
-                .unwrap(),
+            MissingDatum::new(measurement.datum_type(), Some(t2)).unwrap(),
         );
-        let missing_measurement = Measurement::new(Utc::now(), missing_datum);
+        let t3 = Utc::now();
+        let missing_measurement = Measurement::new(t3, missing_datum);
+        dbg!(t1, t2, t3);
         test_recall_measurement_impl(measurement, client).await?;
         test_recall_measurement_impl(missing_measurement, client).await?;
         Ok(())
```
 
I get three of the same time.

```
[oximeter/db/src/client.rs:3156] t1 = 2023-12-12T05:29:13.239922Z
[oximeter/db/src/client.rs:3156] t2 = 2023-12-12T05:29:13.239922Z
[oximeter/db/src/client.rs:3156] t3 = 2023-12-12T05:29:13.239922Z
```

When I artificially make them a nanosecond apart, the tests pass.

```diff
diff --git a/oximeter/db/src/client.rs b/oximeter/db/src/client.rs
index d295d0dcd..9ac91e7dd 100644
--- a/oximeter/db/src/client.rs
+++ b/oximeter/db/src/client.rs
@@ -3145,12 +3145,14 @@ mod tests {
         serde_json::Value: From<T>,
     {
         let datum = Datum::from(hist);
-        let measurement = Measurement::new(Utc::now(), datum);
+        let t1 = Utc::now();
+        let measurement = Measurement::new(t1, datum);
+        let t2 = t1 + chrono::Duration::nanoseconds(1);
         let missing_datum = Datum::Missing(
-            MissingDatum::new(measurement.datum_type(), Some(Utc::now()))
-                .unwrap(),
+            MissingDatum::new(measurement.datum_type(), Some(t2)).unwrap(),
         );
-        let missing_measurement = Measurement::new(Utc::now(), missing_datum);
+        let missing_measurement = Measurement::new(t2, missing_datum);
+        dbg!(t1, t2);
         test_recall_measurement_impl(measurement, client).await?;
         test_recall_measurement_impl(missing_measurement, client).await?;
         Ok(())
```

```
[oximeter/db/src/client.rs:3155] t1 = 2023-12-12T05:48:07.684980Z
[oximeter/db/src/client.rs:3155] t2 = 2023-12-12T05:48:07.684980001Z
```

## Discussion

In order to explain why 

1. this fails consistently for me on my Mac, 
2. it also failed on my branch on Ubuntu but not Helios, and
3. it has not failed on omicron main

I think it would have to be a) something weird about how time works on Mac that makes the `Utc::now` come out the same consistently, b) it was a flukey flake in the Ubuntu case. I'm re-running the failed CI [here](https://github.com/oxidecomputer/omicron/pull/4261/checks?check_run_id=19543819173). If they pass this time, that would be encouraging.

david-crespo:
Closes #4678

### DIFF
--- a/oximeter/db/src/client.rs
+++ b/oximeter/db/src/client.rs
@@ -3145,12 +3145,18 @@ mod tests {
         serde_json::Value: From<T>,
     {
         let datum = Datum::from(hist);
-        let measurement = Measurement::new(Utc::now(), datum);
+
+        // We artificially give different timestamps to avoid a test flake in
+        // CI (reproducible reliably on macOS) where the two Utc::now() are the
+        // same, which means we get two results on retrieval when we expect one
+        let t1 = Utc::now();
+        let t2 = t1 + Duration::from_nanos(1);
+
+        let measurement = Measurement::new(t1, datum);
         let missing_datum = Datum::Missing(
-            MissingDatum::new(measurement.datum_type(), Some(Utc::now()))
-                .unwrap(),
+            MissingDatum::new(measurement.datum_type(), Some(t2)).unwrap(),
         );
-        let missing_measurement = Measurement::new(Utc::now(), missing_datum);
+        let missing_measurement = Measurement::new(t2, missing_datum);
         test_recall_measurement_impl(measurement, client).await?;
         test_recall_measurement_impl(missing_measurement, client).await?;
         Ok(())


oximeter/db/src/client.rs
```diff
@@ -3145,12 +3145,18 @@ mod tests {
         serde_json::Value: From<T>,
     {
         let datum = Datum::from(hist);
-        let measurement = Measurement::new(Utc::now(), datum);
+
+        // We artificially give different timestamps to avoid a test flake in
+        // CI (reproducible reliably on macOS) where the two Utc::now() are the
+        // same, which means we get two results on retrieval when we expect one
+        let t1 = Utc::now();
+        let t2 = t1 + Duration::from_nanos(1);
```
bnaecker, line 3153: So this particular test structure, inserting a real and missing sample, occurs in many places. Is it worth it to make this a method or something, and then call that in the other `*_test_impl` functions?

bnaecker (commented):
Thanks for finding and fixing this. It looks like a good, easy fix, I just wonder if we shouldn't do the same thing in more of the tests.

oximeter/db/src/client.rs
david-crespo: I'll take a look. What is odd is that those don't fail for me locally even though in many cases they seem to be using not two `Utc::now()` calls but literally the same timestamp. So I might need your help understanding what that's about.

bnaecker: Ok, I think I understand why that's the case. This particular test of histograms works a bit differently from the others, which tests scalars. Those construct a sample, run the test for recalling it, and then create the missing sample and run that recall test. This one constructs both samples back-to-back, and then runs each recall test.

So I think the fix you have here is fine, since this one test is the only test where they're constructed back-to-back.

bnaecker (approved):

PR #4681, opened 2023-12-12
Rename disks when un-deleting and faulting

### LINKED ISSUE #4673: Cannot un-delete a disk with the same name
Disks for which their delete saga unwound prior to the fixes in https://github.com/oxidecomputer/omicron/pull/4547 would be "phantom" disks: they would be soft-deleted but would consume virtual resources and count towards quotas. If the user created a disk with the same name _after_ attempting to delete the original disk, this new disk would prevent the phantom disk from being un-deleted due to the constraint that disk names within a project must be unique:

```sql
CREATE UNIQUE INDEX IF NOT EXISTS lookup_disk_by_project ON omicron.public.disk (
    project_id,
    name
) WHERE
    time_deleted IS NULL;
```

As written today, the background task does not rename the disk, so the un-delete step would fail every time the background task is triggered.

Possible fixes:

- rename the disk to something unique during un-deletion
- trigger the disk delete saga instead of un-deleting the disk

jmpesp:
When un-deleting phantom disks and setting them to faulted, use a new name that includes the disk's UUID: this ensures that even if a user created another disk with the same name in the project, the phantom disk can still be un-deleted and faulted, and eventually cleaned up.

Fixes #4673


nexus/db-queries/src/db/datastore/disk.rs
```diff
@@ -657,7 +657,10 @@ impl DataStore {
     /// If the disk delete saga unwinds, then the disk should _not_ remain
     /// deleted: disk delete saga should be triggered again in order to fully
     /// complete, and the only way to do that is to un-delete the disk. Set it
-    /// to faulted to ensure that it won't be used.
+    /// to faulted to ensure that it won't be used. Use the disk's UUID as part
+    /// of its new name to ensure that even if a user created another disk that
+    /// shadows this "phantom" disk the original can still be un-deleted and
+    /// faulted.
```
smklein, line 663: It might be worth calling out - no matter how silly - that like, technically the user could be using this name for another too. Seems unlikely though!

smklein (approved):

nexus/db-queries/src/db/datastore/disk.rs
jmpesp: You're not wrong! added something in 1ac073e8a

Issue #4481, opened 2023-11-09
`static USE_MOCKS` is gross and should be removed


andrewjstone:
During the implementation of https://github.com/oxidecomputer/omicron/pull/4332, I added a static variable that allows selecting between using mocks and not using mocks in tests. This kludge was needed because of feature unification causing the tests in `sled-agent` to fail when run with all other tests at the top-level. Further details can be found in this comment: https://github.com/oxidecomputer/omicron/pull/4332#discussion_r1388623062

We want to eventually remove this disgrace, and the way forward is most likely to refactor the ServiceManager such that it can be tested without mocks in a manner similar to how we restructured `StorageManager` in #4332

Issue #4226, opened 2023-10-07
Revisit moving instances to Failed in `handle_instance_put_result`


gjcolombo:
Follow-up from the refactoring in #4194.

If Nexus calls sled agent to do something that registers an instance or changes its state, sled agent will try to return a `Result<Option<SledInstanceState>, Error>`. Nexus will then get back one of the following values:

- `Ok(None)` if the call succeeded and this didn't change what sled agent thinks the instance/VMM state is
- `Ok(Some(new_state))` if the call succeeded and sled agent wants Nexus to try to write some new state back to CRDB
- `Err(e)` if the call failed, where `e: progenitor_client::Error<sled_agent_client::Error>`

The first two cases are handled by simply trying to write a state update to CRDB and aren't important here. The error case, however, walks a tightrope: https://github.com/oxidecomputer/omicron/blob/2bb0fdc6237ce6b024fb21dc1c1ebe54a4247d41/nexus/src/app/instance.rs#L891-L943

Nexus will take the Progenitor error it got back, feed it through a `From` impl to convert it to an Omicron external API error, and then see if that error is a Bad Request error or something else. In the former case, the update is dropped; in the latter case, Nexus assumes some catastrophe has happened and that the instance has to be marked as Failed.

The relevant From impl is here:

https://github.com/oxidecomputer/omicron/blob/c03029373982a528d372c68f3015506f15dd2a07/common/src/api/external/error.rs#L355-L418

Notice the following:

- Progenitor errors that didn't produce a response from sled agent--e.g. disconnections due to timeouts or network weather--all get turned into internal errors. These will cause an instance to move to Failed even if the problem was completely transient.
- Sled agent responses get flattened into either 400 Bad Request (for all 4XX errors) or 500 Internal Server Error (everything else) unless they happen to be 503 Service Unavailable, which is preserved. (See #3238.)

This creates a great many ways for an instance to be marked as Failed, even if the VM is actually perfectly healthy. The Failed state is also a pain to deal with because a Failed instance can only be deleted and not recovered in any other way (see #2825 and especially https://github.com/oxidecomputer/omicron/issues/2825#issuecomment-1650601195).

Apart from all of this, an instance can be marked as Failed because of a *Propolis* failure: for example, if Propolis hits an error while setting up a VM component, it will publish the Failed state and that will bubble back up to Nexus via sled agent.

[...]

Issue #4004, opened 2023-08-31
want sled-agent timeout on completion of requests to stop a Propolis VM


gjcolombo:
Today, requests to stop a running instance must by necessity involve the instance's active Propolis (sled agent sends the stop request to Propolis; the instance is stopped when Propolis says so, at which point sled agent cleans up the Propolis zone and all related objects). If an instance's Propolis is not responding, or there is no active Propolis, there is no obvious way to clean up the instance and recover.

A short-term workaround is to grant some form of API access to sled agent's "unregister instance" API, which forcibly executes the termination path (tearing down the Propolis zone and removing the instance from the sled's instance table) and can get force an instance into a stopped state.

In the long run instance lifecycle management needs to be made more robust to Propolis failure and/or non-responsiveness.

jordanhendricks:
Related: #3209

gjcolombo:
In #4194, sled agent's "instance unregister" API assumes that it can produce the correct posterior VMM and instance states by emulating a "Propolis destroyed" state transition. (That is, sled agent's "rudely terminate this instance" function computes the next state by pretending that it immediately got a message from the current VMM that says "I am destroyed and my current migration has failed.")

This is fine today because the unregister API is only used when unwinding start and migrate sagas, where the VMMs that are subject to unregistration have by definition not gotten to do anything interesting yet. It's less fine if Propolis has already begun to run, and especially not fine if we're force-quitting an instance that's a migration target:

1. Source and target VMMs agree that migration is done
2. Just before they tell their respective sled agents, Nexus shows up and rudely terminates the source
3. The emulated "VMM destroyed, migration failed" transition retires the source and the migration, which makes it possible to start the instance again...
4. ...but the target might actually begin to run while this is happening!

This seems dangerous. We probably want to adjust the synchronization here to be something more like the following:

1. Nexus "disowns" a particular VMM, e.g. by replacing its ID in an instance record with a zeroed ID (or some other set of sentinel values that prevents the instance from restarting)
2. Nexus terminates all of the disowned VMMs
3. Nexus clears out the sentinels to allow the instance to start again

This would need to be done in a saga to ensure the whole process runs to completion. More design work's needed here. We just need to do that work before we hook up any external APIs to the existing `instance_ensure_unregistered` sled agent interface.

davepacheco:
Also related: #4872

gjcolombo:
[...]

PR #4691, opened 2023-12-14
sled-agent instance creation asynchronous refactor

lifning:
WIP/draft. requires (and described in the description of) https://github.com/oxidecomputer/omicron/pull/4489/files


illumos-utils/src/running_zone.rs
```diff
@@ -1175,29 +1175,55 @@ impl ZoneBuilderFactory {
 /// Created by [ZoneBuilderFactory].
 #[derive(Default)]
 pub struct ZoneBuilder<'a> {
+    /// Logger to which status messages are written during zone installation.
```
gjcolombo, line 1178: Thanks for adding the doc comments here!

gjcolombo (commented):
Thanks for putting this together! I have a few implementation-specific comments and one larger observation:

On my reading, the new code preserves Nexus's existing rule that if a Propolis zone hasn't come up within X seconds, then the instance should be marked as Failed. The differences are mostly in the value of X (120 seconds in the PR vs. 60 now) and in the mechanism by which the Failed state is applied (HTTP request timeout now vs. expiration of a sleep on a Nexus task in the PR). The biggest additional difference that I can see is that in the PR, if the instance has Failed by the time sled agent reports its status, sled agent will explicitly clean up the zone it created (though this unfortunately doesn't free up space on the sled from Nexus's perspective because of #4226).

I might just be missing something about how this works, but if I have it right, I'm worried that the new behavioral juice isn't worth the squeeze of the complexity this PR adds (in the form of an extra sled agent/Nexus round trip whose synchronization and error semantics we have to think about and in the form of extra constraints on when and how instance generation numbers must change).

What would you think of changing Nexus's behavior so that it doesn't enforce a timeout on Propolis zone startup at all? That is, on instance start:

- Nexus moves the relevant Instance to Starting and sets its VMM ID
- Nexus calls the relevant sled's "ensure registered" endpoint (this already returns quickly)
- Nexus calls the sled's "ensure state" endpoint to start the instance; this should now return right away without blocking on a response from the sled agent instance task
- Later, when the sled's Propolis zone comes up and Propolis starts, it will publish a transition to Running that will be handled by the normal means (sled agent will call `cpapi_instances_put` to change the VMM state and the instance's effective state will change accordingly)

I think this is more or less what the sled agent portion of the PR does today, but without the need for an additional Nexus task to enforce the start timeout or extra internal APIs to receive status updates from sled agents.

The main downside of doing this is of course that if something *does* go wrong in instance startup (e.g., the sled bounces), the instance will get stuck in Starting instead of going to Failed, and it's harder (read: next to impossible) to recover an instance that's stuck in Starting. I think that may be justifiable here, though, on the following reasoning:

- Having fewer instances go to Failed outweighs the marginal risk of an instance that's stuck in Starting
  - Failed is better than Starting, but Failed instances are still pretty painful to deal with; you can delete them but that still leaks resources (per the discussion in and TODOs linked to #4226 and #4872)
  - Sled/sled agent bounces are much rarer than long-running zone startup commands, so the pain we save from having fewer instances spuriously go to Failed outweighs the risk of them getting stuck in Starting
- Simplifying this PR frees up some complexity budget that we can spend on #4004, which if successful would give users a way to fix *all* "stuck in Starting"/"stuck in Stopping" problems (not just those caused by zone startup timeouts)

WDYT of this approach?

lifning:
> Thanks for putting this together! I have a few implementation-specific comments and one larger observation:
> [...]
> I might just be missing something about how this works, but if I have it right, I'm worried that the new behavioral juice isn't worth the squeeze of the complexity this PR adds (in the form of an extra sled agent/Nexus round trip whose synchronization and error semantics we have to think about and in the form of extra constraints on when and how instance generation numbers must change).
> 
> What would you think of changing Nexus's behavior so that it doesn't enforce a timeout on Propolis zone startup at all? That is, on instance start:
>   * Nexus moves the relevant Instance to Starting and sets its VMM ID
>   * Nexus calls the relevant sled's "ensure registered" endpoint (this already returns quickly)
>   * Nexus calls the sled's "ensure state" endpoint to start the instance; this should now return right away without blocking on a response from the sled agent instance task
>   * Later, when the sled's Propolis zone comes up and Propolis starts, it will publish a transition to Running that will be handled by the normal means (sled agent will call `cpapi_instances_put` to change the VMM state and the instance's effective state will change accordingly)
> 
> I think this is more or less what the sled agent portion of the PR does today, but without the need for an additional Nexus task to enforce the start timeout or extra internal APIs to receive status updates from sled agents.
> [...]
> 
> WDYT of this approach?

that certainly sounds like a better compromise!  and later, if we have a nexus background task monitoring *all* the known instances for any who've been in Starting too long, we can act on it then with the existing calls.  but yeah, i'm good with throwing out the return-trip of this as long as the resulting behavioral change of "we simply don't time out instance creation at all any more" is acceptable.

sled-agent/src/instance_manager.rs
```diff
@@ -194,7 +195,20 @@ impl InstanceManager {
...
+            // (see InstanceRunner::put_state)
+            InstanceStateRequested::MigrationTarget(_)
+            | InstanceStateRequested::Running => {
+                // don't error on channel being closed
+                tokio::spawn(rx);
+                Ok(InstancePutStateResponse { updated_runtime: None })
+            }
+            InstanceStateRequested::Stopped
+            | InstanceStateRequested::Reboot => rx.await?,
+        }
```
lifning, line 211: with the nexus-side out-of-band timeout stuff removed (and thanks to the recent restructure of InstanceManager to have its own worker task), the crux of this change is now just this block.

sled-agent/src/instance.rs
```diff
@@ -1521,3 +1525,656 @@ impl InstanceRunner {
...
+        )
+        .await
+        .expect("timed out creating Instance struct");
+
+        let (put_tx, put_rx) = oneshot::channel();
+
+        inst.put_state(put_tx, InstanceStateRequested::Running)
+            .await
+            .expect("failed to send Instance::put_state");
+
```
andrewjstone, line 1988: Can we call [`tokio::time::advance`](https://docs.rs/tokio/latest/tokio/time/fn.advance.html) here, so that we don't have to wait for the full `TIMEOUT_DURATION`? This will also allow us to increase that duration so without making the success case necessarily take longer, but still giving plenty of time so we don't get flaky runs in CI.

I think to do this, you'll have to create the timeout future, then advance the time, and then await the future.

andrewjstone (approved):
@lifning This looks really good. The tests are super thorough. I appreciate the hassle you had to go through to setup all the params to pass into the Instance constructor!

I only had a few minor quibbles. The ones that referenced @smklein he should take a quick look at. If he's fine leaving them I think this is good to merge once my other comments are resolved.

sled-agent/src/instance.rs
```diff
@@ -1521,3 +1525,656 @@ impl InstanceRunner {
...
+    use dns_server::dns_server::ServerHandle as DnsServerHandle;
+    use dropshot::test_util::LogContext;
+    use dropshot::{HandlerTaskMode, HttpServer};
+    use illumos_utils::dladm::MockDladm;
+    use illumos_utils::dladm::__mock_MockDladm::__create_vnic::Context as MockDladmCreateVnicContext;
+    use illumos_utils::dladm::__mock_MockDladm::__delete_vnic::Context as MockDladmDeleteVnicContext;
+    use illumos_utils::opte::params::DhcpConfig;
+    use illumos_utils::svc::__wait_for_service::Context as MockWaitForServiceContext;
+    use illumos_utils::zone::MockZones;
+    use illumos_utils::zone::__mock_MockZones::__boot::Context as MockZonesBootContext;
```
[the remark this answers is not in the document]
lifning: right, i think ideally we eventually replace all the uses of mockall in illumos\_utils with fakes that we can plumb through anything that uses them - which may be a bit cumbersome, as currently a great deal of that functionality (including what's used here) are static methods.

though if i understand correctly, nextest runs each test case in its own process, and if so these simple stubs returning a static value shouldn't be stepping on anything substantial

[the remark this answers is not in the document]
andrewjstone: The problem that I ran into that was somewhat alluded to was https://github.com/oxidecomputer/omicron/issues/4481. It doesn't appear to be a problem specifically here though.

Issue #4628, opened 2023-12-06
Floating IP delete/attach/detach should be fully contained in transactions


FelixMcFelix:
As of #4559, the above 3 operations use a fetched `FloatingIp`/`ExternalIp` to provide user-friendly error messages based on the current state of `parent_id`, before firing an UPDATE query conditioned on the correct `parent_id` field.

While this is not incorrect, this will provide a less user-friendly error message in the event of concurrent modification. We should move these datastore methods into a `transaction` and/or `transaction_retry_wrapper` to clean this up.

bnaecker:
I'm pretty confident one can do this without an interactive transaction (with multiple DB roundtrips). Instead, the `update_and_check()` machinery could be used to attach or detach a floating IP with the desired conditions, such as `parent_id IS NULL` or similar.

FelixMcFelix:
That looks like just the ticket, thanks! I hadn't yet come across it -- I'll see about working that in as a solution alongside #4630.

PR #4694, opened 2023-12-14
Live attach/detach of external IPs

### LINKED ISSUE #4630: Need to support live attach/detach of floating IPs from instances
#4559 defines the API and plumbing for floating IPs to be used by instances. However, they can only be attached and detached from instances during their creation and deletion, respectively. We need to be able to perform these modifications on running instances.

At a high level (based on my current prototype), I think this needs:
* New endpoints on nexus, likely modeled on disk-attach.
* New endpoints on sled-agent to add/remove an external IP.
* Additional sagas to coordinate nexus and sled-agent operations.

There's a question of how this will interact with migration which Greg, Mike, Trey, Luqman and myself discussed earlier this week. I think in an ideal world we would send sled-agent messages to both the source and target sled (in addition to ensuring new NAT bindings are propagated out to DPD for the current active VM). For an initial implementation, we are probably going to be best-served by returning a 503+retry-after if a migration is in progress, while *possibly* blocking new migrations using a sentinel all-zeroes `migration_id`.

FelixMcFelix:
This PR adds new endpoints to attach and detach external IPs to/from an individual instance at runtime, when instances are either stopped or started. These new endpoints are:

* POST `/v1/floating-ips/{floating_ip}/attach`
* POST `/v1/floating-ips/{floating_ip}/detach`
* POST `/v1/instances/{instance}/external-ips/ephemeral`
* DELETE `/v1/instances/{instance}/external-ips/ephemeral`

These follow and enforce the same rules as external IPs registered during instance creation: at most one ephemeral IP, and at most 32 external IPs total.

`/v1/floating-ips/{floating_ip}/attach` includes a `kind` field to account for future API resources which a FIP may be bound to -- such as internet gateways, load balancers, and services.

## Interaction with other instance lifecycle changes and sagas

Both external IP modify sagas begin with an atomic update to external IP attach state conditioned on $\mathit{state}\in[ \mathit{started},\mathit{stopped}]$. As a result, we know that an external IP saga can only ever start before any other instance state change occurs. We then only need to think about how these other sagas/events must behave when called *during* an attach/detach, keeping in mind that these are worst-case orderings: attach/detach are likely to complete quickly.

### Instance start & migrate

Both of these sagas alter an instance's functional sled ID, which controls whether NAT entry insertion and OPTE port state updates are performed. If an IP attach/detach is incomplete when either saga reaches `instance_ensure_dpd_config` or `instance_ensure_registered` (e.g., any IP associated with the target instance is in attaching/detaching state), the start/migrate will unwind with an HTTP 503.

Generally, neither should undo in practice since IP attach/detach are fast operations -- particularly when an instance is formerly stopped. This is used solely to guarantee that only one saga is accessing a given external IP at a time, and that the update target remains unchanged.

### Instance stop & delete

These operations are either not sagaized (stop), or cannot unwind (delete), and so we cannot block them using IP attach state. IP attach/detach will unwind if a given sled-agent is no longer responsible for an instance. Instance delete will force-detach IP addresses bound to an instance, and if this is seen then IP attach will deliberately unwind to potentially clean up NAT state. OPTE/DPD undo operations are best-effort in such a case to prevent stuck sagas.

Instance stop and IP attach may interleave such that the latter adds additional NAT entries after other network state is cleared. Because we cannot unwind in this case, `instance_ensure_dpd_config` will now attempt to remove leftover conflicting RPW entries if they are detected, since we know they are a deviation from intended state.

## Additional/supporting changes

* Pool/floating IP specifiers in instance create now take `NameOrId`, parameter names changed to match.
* External IP create/bind in instance create no longer double-resolves name on saga unwind.
* `views::ExternalIp` can now contain `FloatingIp` body.
* DPD NAT insert/remove functions now perform single rule update via ID instead of index into the EIP list -- index-based was unstable under live addition/removal.
* NAT RPW ensure is now more authoritative, and will remove conflicting entries if an initial insert fails.
* Pool `NameOrId` resolution for floating IP allocation pulled up from `Datastore` into `Nexus`.

---

Closes #4630 and #4628.


nexus/src/external_api/http_entrypoints.rs
```diff
@@ -3727,6 +3729,78 @@ async fn instance_external_ip_list(
...
+#[endpoint {
+    method = POST,
+    path = "/v1/instances/{instance}/external-ips/detach",
+    tags = ["instances"],
+}]
+async fn instance_external_ip_detach(
+    rqctx: RequestContext<Arc<ServerContext>>,
+    path_params: Path<params::InstancePath>,
+    query_params: Query<params::OptionalProjectSelector>,
+    ip_to_detach: TypedBody<params::ExternalIpDelete>,
```
ahl, line 3778: `ExternalIpDetach`?

nexus/types/src/external_api/params.rs
```diff
@@ -947,6 +947,13 @@ pub enum ExternalIpCreate {
     Floating { floating_ip_name: Name },
 }
 
+#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema)]
+#[serde(tag = "type", rename_all = "snake_case")]
+pub enum ExternalIpDelete {
+    Ephemeral,
+    Floating { floating_ip_name: Name },
+}
```
ahl, line 955: Why aren't we detaching by id?

ahl (commented):
some minor API feedback

nexus/types/src/external_api/params.rs
FelixMcFelix: This is mostly symmetrical to `ExternalIpCreate`, which is used in instance creation as well as IP attach. I'd prefer to specify a `NameOrId` in both structs (for both the target floating IP and ephemeral IP pool), but I'm not sure what the policy is for a breaking change there.

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -3727,6 +3729,78 @@ async fn instance_external_ip_list(
...
+) -> Result<HttpResponseAccepted<Option<views::ExternalIp>>, HttpError> {
```
david-crespo, line 3779: As mentioned [here](https://github.com/oxidecomputer/oxide.rs/pull/499#discussion_r1442168889), this should be something like

```rs
struct ExternalIpDetachment {
  external_ip: Option<ExternalIp>
}
```

nexus/types/src/external_api/params.rs
ahl: I think the policy is BREAK THE API. We have basically zero consumers so now is the time to get the API right.

nexus/src/external_api/http_entrypoints.rs
ahl: I agreed with @david-crespo that that's not the structure we probably want, however, I think the fix is to return an error response rather than a success response that indicates that there was no such entity. @david-crespo wrote up guidance for this: https://github.com/oxidecomputer/omicron/blob/main/docs/http-status-codes.adoc

nexus/types/src/external_api/params.rs
david-crespo: Yeah, we need to fix the other one, not make this one equally bad

nexus/src/external_api/http_entrypoints.rs
david-crespo: Agreed, I didn’t read far enough to see why it might be None

FelixMcFelix: That sounds reasonable enough for this case. My understanding is that the saga return value needs to remain `Ok(_)` due to idempotency requirements, but converting to another status code is simple just outside that.

nexus/src/app/instance_network.rs
```diff
@@ -344,47 +343,71 @@ impl super::Nexus {
...
                 )));
             }
+        } else {
+            (&ips[..], true)
+        };
+
+        // This is performed so that an IP attach/detach will block the
+        // instance_start saga. Return service unavailable to indicate
+        // the request is retryable.
+        if must_all_be_attached
```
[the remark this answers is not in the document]
FelixMcFelix: Re-ensuring an existing NAT entry for the RPW is a no-op, and should only cost a row lookup. So there's nothing wrong with always pushing all `Attached` IPs apart from wasted effort.

We do still require an override for at least one other entry in an arbitrary state, because of saga unwind of attach/detach: an IP might be `Detaching`, unwind when informing OPTE, and then need to be specifically readded to NAT. Similarly an `Attaching` IP might unwind at the same step, and need to be removed from NAT. As a result, we don't want other executing sagas to try to do anything with IPs in a state they don't expect or IPs they don't 'own'. We'd also need to ensure that instance_start sees "all attached", and that attach/detach don't care about other IPs' states because they can safely run concurrently for different IP objects on the same instance.

Unrelated: Looking over this again I've also seen that `ensure_nat_entry` is needlessly duplicated per IP, which is again a no-op. This is also indirectly duplicated today in `main` (instance_start: `for switch in boundary_switches {... instance_ensure_dpd_config ...}` -> `for ip in ips { ... ensure_nat_entry ... }`), but we can fix this here now.

nexus/src/app/sagas/instance_ip_attach.rs
```diff
@@ -0,0 +1,527 @@
...
+        &params.serialized_authn,
+        IpAttachState::Attaching,
+        IpAttachState::Attached,
+    )
+    .await?
+    {
+        warn!(
+            log,
+            "siia_complete_attach: external IP was deleted or call was idempotent"
+        )
```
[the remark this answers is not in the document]
FelixMcFelix: I had intentionally designed for Case 2, but had partly overlooked Case 1 (assuming that both are stop->delete) -- thanks a lot for catching that. I agree on that unwind pathway; although I'd (maybe incorrectly) assumed that the instance would be guaranteed to be removed from InstanceManager before moving from `Stopping->Stopped`, since any OPTE endpoint on sled-agent will then give us `Error::NoSuchInstance`.

I'd also be happier to unwind here rather than to push the limits of what we can get away with. Examining it a bit more closely however, there is another issue with the Case 1 race. The NAT entry will not be removed on unwind -- it's no longer attached to the instance, so `instance_delete_dpd_config` will not see it. We need a tweak there to handle this and not wipe out a potential followup attach on another instance.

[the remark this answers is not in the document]
FelixMcFelix: Okay, I've resolved this by unwinding if we observe a delete at the end of the attach saga. I've put in some extra work to clean up straggler NAT entries more precisely (needed in the delete-unwind case) and to make `instance_ensure_dpd_config` more aggressive in pushing routes where an old conflict may exist. I think doing both of these means more useful buffers between us and catastrophe.

FelixMcFelix:
@ahl @david-crespo On the API design front, I've written up some thoughts after the call on Tuesday into a strawman we can pick apart: https://gist.github.com/FelixMcFelix/18d20262a918ccf691a325a8d948379d.

gjcolombo (approved):
Thanks for adding the new checks here--LGTM for the instance lifecycle pieces (I haven't looked at the external API bits at all).

FelixMcFelix:
Attach and detach APIs for floating/ephemeral have been retooled according to our last meeting -- split by ephemeral/floating, and floating IP control now lives on the FIP itself to support future network API developments. If there are naming tweaks needed then they shouldn't be too onerous to make. :)

david-crespo (approved):
Didn't review implementation closely, but the APIs look good!

PR #4696, opened 2023-12-14
Capacity and utilization

zephraph:
This PR is a follow up to #4605 which adds views into capacity and utilization both at the silo and system level. 

API: 
|op|method|url|
|--|--|--|
|silo_utilization_list | GET    |  /v1/system/utilization/silos |
|silo_utilization_view        |            GET    |  /v1/system/utilization/silos/{silo} |
|utilization_view          |               GET  |    /v1/utilization |

I'm not entirely satisfied w/ the silo utilization endpoints. They could be this instead:

|op|method|url|
|--|--|--|
|silo_utilization_list | GET    |  /v1/system/silos-utilization |
|silo_utilization_view        |            GET    |  /v1/system/silos/{silo}/utilization |

Also take special note of the views

```rust
// For the eyes of end users
/// View of the current silo's resource utilization and capacity
#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema)]
pub struct Utilization {
    /// Accounts for resources allocated to running instances or storage allocated via disks or snapshots
    /// Note that CPU and memory resources associated with a stopped instances are not counted here
    /// whereas associated disks will still be counted
    pub provisioned: VirtualResourceCounts,
    /// The total amount of resources that can be provisioned in this silo
    /// Actions that would exceed this limit will fail
    pub capacity: VirtualResourceCounts,
}

// For the eyes of an operator
/// View of a silo's resource utilization and capacity
#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema)]
pub struct SiloUtilization {
    pub silo_id: Uuid,
    pub silo_name: Name,
    /// Accounts for resources allocated by in silos like CPU or memory for running instances and storage for disks and snapshots
    /// Note that CPU and memory resources associated with a stopped instances are not counted here
    pub provisioned: VirtualResourceCounts,
    /// Accounts for the total amount of resources reserved for silos via their quotas
    pub allocated: VirtualResourceCounts,
}
```

For users in the silo I use `provisioned` and `capacity` as the language. Their `capacity` is represented by the quota set by an operator. For the operator `provisioned` is the same but `allocated` is used to denote the amount of resources allotted via quotas. 

---

Note: I had planned to add a full system utilization endpoint to this PR but that would increase the scope. Instead will ship that API as a part of the next release. We can calculate some version of the full system utilization on the client by listing all the silos and their utilization.


common/src/api/external/mod.rs
```diff
@@ -71,6 +71,23 @@ pub trait ObjectIdentity {
...
+impl<T: ObjectIdentity> SimpleIdentity for T {
+    fn id(&self) -> Uuid {
+        self.identity().id
+    }
+
+    fn name(&self) -> &Name {
+        &self.identity().name
+    }
+}
+
```
david-crespo, line 90: sneaky

nexus/db-model/src/quota.rs
```diff
@@ -78,9 +80,9 @@ impl From<views::SiloQuotas> for SiloQuotas {
...
             time_modified: Utc::now(),
-            cpus: silo_quotas.cpus,
-            memory: silo_quotas.memory.into(),
-            storage: silo_quotas.storage.into(),
+            cpus: silo_quotas.limits.cpus,
+            memory: silo_quotas.limits.memory.into(),
+            storage: silo_quotas.limits.storage.into(),
         }
     }
 }
```
david-crespo, line 88: question stands from the previous PR, is this From used? it doesn't make sense to go from view to model

zephraph: I deleted it and ran the quotas tests which passed so... no? I don't recall why I added that initially.

openapi/nexus.json
```diff
@@ -6883,6 +6980,33 @@
         }
       }
     },
+    "/v1/utilization": {
```
karencfv, line 6983: Should this not be `/v1/system/utilization`? It also has the tag "silos" like the other endpoints. Conversely, if we want to keep this one as `/v1/utilization`, perhaps we could change the other endpoints to be `/v1/system/utilization`. Otherwise it feels like one set of endpoints is very explicit and the other is not. WDYT?

karencfv (commented):
Almost there! Thanks for pushing through with this :)

openapi/nexus.json
zephraph: This is the silo admin's view of the utilization inside the silo. So this is a developers way of seeing how many resources their silo can provision. I dropped `/v1/system/utilization` (which summarizes the entirety of the system utilization) from this PR as it was more complicated to deliver. 

In my mind `/v1/system/...` means operator whereas `/v1/...`

common/src/api/external/mod.rs
zephraph: Yeah, the constraint of implementing `ObjectIdentity` felt a little unnecessary. We have several of these not-really-resource resources and we should still be able to treat as if they were (even when they don't have a resource/asset identity)

openapi/nexus.json
karencfv: > I dropped /v1/system/utilization (which summarizes the entirety of the system utilization) from this PR as it was more complicated to deliver.

Ah! I think this was the piece of information I was missing. If we want to have this endpoint in the future then the current implementation makes sense, thanks!

schema/crdb/dbinit.sql
```diff
@@ -836,6 +836,31 @@ CREATE TABLE IF NOT EXISTS omicron.public.silo_quotas (
...
+    s.name AS silo_name,
+    c.cpus_provisioned AS cpus_provisioned,
+    c.ram_provisioned AS memory_provisioned,
+    c.virtual_disk_bytes_provisioned AS storage_provisioned,
+    q.cpus AS cpus_allocated,
+    q.memory_bytes AS memory_allocated,
+    q.storage_bytes AS storage_allocated
+FROM
+    omicron.public.virtual_provisioning_collection AS c
+    RIGHT JOIN omicron.public.silo_quotas AS q
```
david-crespo, line 855: Is this a right join because a silo which has never had a provisioning even would have no entry in `virtual_provisioning_collection`?

david-crespo (approved):
Very clean! Would like to see integration tests on the new endpoints. I figure you have those on the way.

schema/crdb/dbinit.sql
zephraph: A provisioning collection can be the fleet, a silo, or a project. I want all `silo_quotas` and whatever entries in the `virtual_provisioning_collection` as match.

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -515,6 +518,87 @@ async fn policy_update(
...
+        Ok(HttpResponseOk(utilization.into()))
+    };
+    apictx.external_latencies.instrument_dropshot_handler(&rqctx, handler).await
+}
+
+/// View the current utilization of a given silo
+#[endpoint {
+    method = GET,
+    path = "/v1/system/utilization/silos/{silo}",
+    tags = ["silos"],
```
david-crespo, line 547: I think these should be `system/silos` instead of `silos`

nexus/types/src/external_api/views.rs
```diff
@@ -49,14 +49,64 @@ pub struct Silo {
...
+}
+
+// For the eyes of end users
+/// View of the current silo's resource utilization and capacity
+#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema)]
+pub struct Utilization {
+    /// Accounts for resources allocated to running instances or storage allocated via disks or snapshots
+    /// Note that CPU and memory resources associated with a stopped instances are not counted here
+    /// whereas associated disks will still be counted
+    pub provisioned: VirtualResourceCounts,
```
david-crespo, line 78: Are these supposed to have `#[serde(flatten)]` on them too? I think this is why `VirtualResourceCounts` is showing up in the spec. On the other hand it's kind of cool to not flatten it and share the structure, but it's possible the other language clients might not like it as much.

david-crespo: Whoops, I see now why these are not. They're supposed to be under these keys. In any case this is still probably why `VirtualResourceCounts` is making it into the spec.

Issue #4607, opened 2023-12-04
[API] `add_sled_to_initialized_rack` and `uninitialized_sled_list` need some work


ahl:
- [x] Consider naming and path of `uninitialized_sled_list`
- [x] `uninitialized_sled_list` should be paginated.
- [x] Consider naming of `add_sled_to_initialized_rack`

---

Here are the current `system/hardware` operations:

```
API operations found with tag "system/hardware"   
OPERATION ID                             METHOD   URL PATH
add_sled_to_initialized_rack             POST     /v1/system/hardware/sleds
networking_switch_port_apply_settings    POST     /v1/system/hardware/switch-port/{port}/settings
networking_switch_port_clear_settings    DELETE   /v1/system/hardware/switch-port/{port}/settings
networking_switch_port_list              GET      /v1/system/hardware/switch-port
physical_disk_list                       GET      /v1/system/hardware/disks
rack_list                                GET      /v1/system/hardware/racks
rack_view                                GET      /v1/system/hardware/racks/{rack_id}
sled_instance_list                       GET      /v1/system/hardware/sleds/{sled_id}/instances
sled_list                                GET      /v1/system/hardware/sleds
sled_physical_disk_list                  GET      /v1/system/hardware/sleds/{sled_id}/disks
sled_set_provision_state                 PUT      /v1/system/hardware/sleds/{sled_id}/provision-state
sled_view                                GET      /v1/system/hardware/sleds/{sled_id}
switch_list                              GET      /v1/system/hardware/switches
switch_view                              GET      /v1/system/hardware/switches/{switch_id}
uninitialized_sled_list                  GET      /v1/system/hardware/uninitialized-sleds
```

1.
I think that `uninitialized_sled_list` may be a little out of place (but it might be fine). One option is to have a query param to `sled_list` that determines whether all sleds, initialized sleds, uninitialized sleds, both/all/whatever are listed. Another is the a-grammatical `sled_uninitialized_list`.

2.
The operation returns an array of items; it should probably be paginated.

3.
`add_sled_to_initialized_rack` is similarly different. I'd suggest `sled_add_uninitialized` or perhaps simply `sled_add`. Note also that this operation is `/v1/system/hardware/sleds`; would it make sense for `uninitialized_sled_list` to have a more consistent path for consistency with this operation on uninitialized sleds?

david-crespo:
1. I like query param on `sled_list` if it really is just a filter on that list
[...]

PR #4698, opened 2023-12-14
Cleanup list uninitialized sled API

andrewjstone:
* Rename `uninitialized_sled_list` to `sled_list_uninitialized`
* Add fake pagination to this API endpoint

Fixes part of #4607

### DIFF
--- a/nexus/src/external_api/http_entrypoints.rs
+++ b/nexus/src/external_api/http_entrypoints.rs
@@ -4570,7 +4570,7 @@ async fn rack_view(
 /// List uninitialized sleds in a given rack
 #[endpoint {
     method = GET,
-    path = "/v1/system/hardware/uninitialized-sleds",
+    path = "/v1/system/hardware/sleds-uninitialized",
     tags = ["system/hardware"]
 }]
 async fn sled_list_uninitialized(
--- a/nexus/src/external_api/http_entrypoints.rs
+++ b/nexus/src/external_api/http_entrypoints.rs
@@ -228,7 +228,7 @@ pub(crate) fn external_api() -> NexusApiDescription {
         api.register(physical_disk_list)?;
         api.register(switch_list)?;
         api.register(switch_view)?;
-        api.register(uninitialized_sled_list)?;
+        api.register(sled_list_uninitialized)?;
         api.register(add_sled_to_initialized_rack)?;
 
         api.register(user_builtin_list)?;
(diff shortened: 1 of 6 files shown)


nexus/tests/output/nexus_tags.txt
```diff
@@ -125,12 +125,12 @@ rack_list                                GET      /v1/system/hardware/racks
 rack_view                                GET      /v1/system/hardware/racks/{rack_id}
 sled_instance_list                       GET      /v1/system/hardware/sleds/{sled_id}/instances
 sled_list                                GET      /v1/system/hardware/sleds
+sled_list_uninitialized                  GET      /v1/system/hardware/uninitialized-sleds
```
zephraph, line 128: Should we changed the endpoint to be `/v1/system/hardware/sleds-uninitialized`? This follows a similar pattern where in cases we may have wanted the API to be `/sleds/uninitialized` but can't because of naming conflicts, we use `-` instead.

andrewjstone: If that's the consensus, I can go ahead and make this change.

david-crespo: I agree with Justin. This pattern is not obvious in the API because where it really shows up is the console, e.g., [`/projects`](https://oxide-console-preview.vercel.app/projects) vs. [`/projects-new`](https://oxide-console-preview.vercel.app/projects-new).

There are a few endpoints in the API that are similar, like `/v1/system/silo-quotas` and `/v1/vpc-subnets`, but both of those are hierarchical noun-noun (parent-child) pairs so not quite the same thing. But yes, the basic principle is to line up the names so they're obviously related:

```
/sleds
/sleds-uninitialized
```

david-crespo (approved):
looks good pending the discussed change to the path

andrewjstone:
Thanks for the review and tips @zephraph @david-crespo! This was my first new nexus endpoint, so I'm not surprised I ran into something like this.

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -4570,18 +4570,18 @@ async fn rack_view(
 /// List uninitialized sleds in a given rack
 #[endpoint {
     method = GET,
-    path = "/v1/system/hardware/uninitialized-sleds",
+    path = "/v1/system/hardware/sleds-uninitialized",
     tags = ["system/hardware"]
 }]
-async fn uninitialized_sled_list(
+async fn sled_list_uninitialized(
     rqctx: RequestContext<Arc<ServerContext>>,
```
ahl, line 4577: this is insufficient to make this into a paginated endpoint. We also need the `Query<..>` even though we'll ignore it.

openapi/nexus.json
```diff
@@ -4015,6 +4015,33 @@
...
+              }
+            }
+          },
+          "4XX": {
+            "$ref": "#/components/responses/Error"
+          },
+          "5XX": {
+            "$ref": "#/components/responses/Error"
+          }
+        }
```
ahl, line 4042: we'd expect to see "x-dropshot-paginated" in here

nexus/src/external_api/http_entrypoints.rs
andrewjstone: Sorry about that and thanks for the tip! I will update it in a follow up.

ahl: I'm sorry I didn't get this comment in in time!

andrewjstone: Fixed in https://github.com/oxidecomputer/omicron/pull/4720

openapi/nexus.json
andrewjstone: `x-dropshot-pagination` shows up as of https://github.com/oxidecomputer/omicron/pull/4720

PR #4703, opened 2023-12-15
[trivial] Move `SCHEMA_VERSION` to the top of `schema.rs`

david-crespo:
Digging for this in the middle of the file when I want to bump the version drives me nuts. Will obviously hold until after release.


andrewjstone (commented):
Wise man!

PR #4704, opened 2023-12-15
Rename API to sled_add and take UninitializedSledId as param

andrewjstone:
This is the second half of the fix for #4607


ahl (commented):
API looks good to me; thanks for doing this!

nexus/src/app/rack.rs
```diff
@@ -933,33 +935,76 @@ impl super::Nexus {
...
+        //
+        // We set a limit of 200 here to give us some breathing room when
+        // querying for cabooses and RoT pages, each of which is "4 per SP/RoT",
+        // which in a single fully populated rack works out to (32 sleds + 2
+        // switches + 1 psc) * 4 = 140.
+        //
+        // This feels bad and probably needs more thought; see
+        // https://github.com/oxidecomputer/omicron/issues/4621 where this limit
+        // being too low bit us, and it will link to a more general followup
+        // issue.
```
davepacheco, line 974: It seems like this same code was added in #4626 -- did it move here or something?

nexus/src/app/rack.rs
```diff
@@ -933,33 +935,76 @@ impl super::Nexus {
...
+
+        // If there isn't a collection, we don't know about the sled
+        let Some(collection) = collection else {
+            return Err(Error::ObjectNotFound {
+                type_name: ResourceType::Sled,
+                lookup_type:
+                    omicron_common::api::external::LookupType::ByCompositeId(
+                        format!("{sled:?}"),
+                    ),
+            });
```
davepacheco, line 989: I think I might make this a 500 or 503 with internal message "no inventory data available".  The problem here doesn't seem like it's that they gave us a sled that doesn't exist -- it's that something is seriously broken because we have no inventory data at all.

nexus/src/app/rack.rs
```diff
@@ -933,33 +935,76 @@ impl super::Nexus {
...
+        // Find the revision
+        let Some(sp) = collection.sps.get(&baseboard_id) else {
+            return Err(Error::ObjectNotFound {
+                type_name: ResourceType::Sled,
+                lookup_type:
+                    omicron_common::api::external::LookupType::ByCompositeId(
+                        format!("{sled:?}"),
+                    ),
+            });
+        };
```
davepacheco, line 1001: This at least is an explicit lookup, but it wasn't by sled id.  edit: oh! we're using the debug formatter for the UninitializedSled, which _would_ have the baseboard properties...that seems okay, although maybe a little weird to expose the Rust debug format to the end user.

nexus/src/app/rack.rs
```diff
@@ -985,13 +1030,11 @@ impl super::Nexus {
...
-                internal_message: format!(
-                    "failed to add sled with baseboard {:?} to rack {}: {e}",
-                    sled.baseboard, allocation.rack_id
-                ),
-            }
+        sa.sled_add(&req).await.map_err(|e| Error::InternalError {
+            internal_message: format!(
+                "failed to add sled with baseboard {:?} to rack {}: {e}",
+                sled, allocation.rack_id
+            ),
```
davepacheco, line 1037: No worse than before, but this would incorrectly report a transient (e.g., network) issue as a 500 when it should be a 503.

davepacheco (approved):
A few suggestions here but it looks good!

nexus/src/app/rack.rs
andrewjstone: This code is just a copy of the call to `db_datastore.inventory_get_latest_collection()`. I just figured I better copy the comment as well to reflect why we chose a limit of 200. I suppose, I could move the comment and limit to a const local to this file as well. Wdyt?

nexus/src/app/rack.rs
andrewjstone: Would you prefer I change this to a 503 or be more specific and check to see which error is returned? I'm not sure the sled-agent differentiates between a 500 and 503 for this endpoint and in most cases 500 is probably better.

nexus/src/app/rack.rs
andrewjstone: Fixed in [d14ca51](https://github.com/oxidecomputer/omicron/pull/4704/commits/d14ca5157c6a7eb130b613a057042f011a1a1fb8)

nexus/src/app/rack.rs
davepacheco: Ah okay.  What do you think of putting the "200" into a const in this file and putting the comment there instead?

nexus/src/app/rack.rs
davepacheco: I think this is fine.  I was going to suggest formatting it in JSON instead of Rust Debug output, but I looked around for other users of `ByCompositeId` and they're all doing various things, including using the debug output, so it's probably not worth worrying about here.

nexus/src/app/rack.rs
davepacheco: Ideally we'd check which error was returned.  It might not even be an error from the sled agent.  For example right now if we got DNS resolution error or a TCP connection failure of some kind, we'd report that as a 500, but that should probably be a 503 (more on this [here](https://github.com/oxidecomputer/omicron/blob/main/docs/http-status-codes.adoc#34-server-errors)).

I'm not sure it's worth dealing with in this PR, since the problem predates this and I think the fix is probably more systemic.  We probably want to make a helper function that we can use in a bunch of places (similar in spirit to [this one](https://github.com/oxidecomputer/omicron/blob/de152304c3f551914db05bda1ddbc4b139e1f766/clients/dns-service-client/src/lib.rs#L27-L83) and then use it everywhere we make a Progenitor call (similar to the way we use `public_error_from_diesel` every time we make a database call...though I just noticed that one seems to have the same problem!)

PR #4705, opened 2023-12-15
Fix FIP creation being able to access IP Pools in other silos

FelixMcFelix:
As raised in Matrix, this backports a fix from #4261 where a new floating IP can be allocated using the name or ID of an IP pool which is bound to another silo.


david-crespo (approved):
😅

PR #4707, opened 2023-12-15
[authz-macros] accept an optional input_key argument

sunshowers:
In some cases including composite keys, it can be better to make the outside
representation of the primary key some kind of struct, rather than passing
around tuples of various types.

Enable that in the `authz_resource` macro by allowing users to specify an
optional `input_key` argument, which represents a better view into the primary
key.

I'm not entirely sure that the `From` trait is the right thing to use here, but
it seems like a pretty low-cost decision to change in the future.

As part of this PR I also switched to the prettyplease crate, which as the
README explains is more suitable for generated code than rustfmt:
https://crates.io/crates/prettyplease

### DIFF
--- a/nexus/authz-macros/src/lib.rs
+++ b/nexus/authz-macros/src/lib.rs
@@ -480,6 +480,8 @@ mod tests {
             name = "Instance",
             parent = "Project",
             primary_key = (String, String),
+            // The SomeCompositeId type doesn't exist, but that's okay because
+            // this code is never compiled, just printed out.
             input_key = SomeCompositeId,
             roles_allowed = false,
             polar_snippet = InProject,
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6147,9 +6147,9 @@ dependencies = [
 
 [[package]]
 name = "prettyplease"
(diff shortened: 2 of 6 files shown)


smklein (approved):
Makes sense to me, thanks for the additional doc comment!

nexus/authz-macros/src/lib.rs
```diff
@@ -411,15 +462,34 @@ fn do_authz_resource(
...
+            polar_snippet = Custom,
+        })
+        .unwrap();
+        println!("{}", pretty_format(output));
+
+        let output = do_authz_resource(quote! {
+            name = "Instance",
+            parent = "Project",
+            primary_key = (String, String),
+            input_key = SomeCompositeId,
```
smklein, line 483: Okay the only part of this I find odd is that `SomeCompositeId` doesn't seem to exist as a struct here -- is this generating code that doesn't compile?

sunshowers: Yeah, this is generating code just to be printed out. I'll add a comment here.

Issue #3238, opened 2023-05-26
sled agent and Nexus frequently flatten errors into 500 Internal Server Error


gjcolombo:
Sled agent has a relatively rich set of internal error types, but at the Dropshot HTTP error boundary, it converts almost all of them to 500 errors:

https://github.com/oxidecomputer/omicron/blob/cab09253e6e6e27761d1ef99f8087f86ccad2e8b/sled-agent/src/sled_agent.rs#L102-L137

This makes it hard for callers to reason about the cause or permanence of any of these errors. This was a pebble in the shoe of PR #2892 and is coming up again in #3230 (sled agent is making a call to Nexus whose response depends in part on calls back down into sled agent, and there's no way to be sure about the permanence of any errors returned from Nexus in that path because all errors are getting flattened into a single error code).

There are similar paths through Nexus that flatten errors this way that the fix for #3230 will have to take into account:

- Updating instance state can update the instance's Dendrite configuration: https://github.com/oxidecomputer/omicron/blob/cab09253e6e6e27761d1ef99f8087f86ccad2e8b/nexus/src/app/instance.rs#L1138-L1144
- `instance_ensure_dpd_config` calls `dpd_client.ensure_nat_entry` and maps all errors to 500s: https://github.com/oxidecomputer/omicron/blob/cab09253e6e6e27761d1ef99f8087f86ccad2e8b/nexus/src/app/instance.rs#L1243-L1259
- `ensure_nat_entry` calls Dendrite daemon routines that can (I presume) produce transient failures (e.g. transient communication failures), e.g.: https://github.com/oxidecomputer/omicron/blob/cab09253e6e6e27761d1ef99f8087f86ccad2e8b/dpd-client/src/lib.rs#L111-L127

Revisiting absolutely every single error conversion in sled agent and Nexus all at once is probably a non-starter, but we can likely start by improving sled agent's internal-error-to-Dropshot-error conversion and look for opportunities to remove other error-flattening cases as they arise.

gjcolombo:
#3334 is a first step toward addressing this issue.

hawkw:
Looking closer at this, it looks like a lot of the work here has been addressed in #3334 and #4682. It's not particularly clear what else is left to do in _this_ issue on the `nexus/`sled-agent` side. Perhaps @gjcolombo will have more input into this when he's back from leave.

We do need to get propolis#649 merged, though, which is _sort of_ related...

hawkw:
[...]

Issue #3206, opened 2023-05-24
sled-agent failed failed to notice propolis-server panicked


jordanhendricks:
An instance on rack2 panicked, but its state continued to be reported in the console as "Running". The `instance_state_monitor` task in sled-agent is responsible for observing state changes from propolis, which loops calling the `instance_state_monitor` endpoint. If propolis-server panics, then propolis will stop returning requests from that endpoint. I am unsure of what happens in this case; I assume the request will timeout and report an error, which will exit the task and result in a similar failure mode as #2727.

jordanhendricks:
There isn't anything in the sled-agent logs to indicate what happened after the instance was started:

```
[2023-05-23T21:08:05.313743363Z]  INFO: SledAgent/InstanceManager/691 on BRM42220026: Configuring new Omicron zone: oxz_propolis-server_7ffa6f90-cbd3-407a-9c9a-aac81e975a40 (instance_id=642c108e-5cbe-4569-bf43-b682905a090c)
[2023-05-23T21:08:05.341565987Z]  INFO: SledAgent/InstanceManager/691 on BRM42220026: Installing Omicron zone: oxz_propolis-server_7ffa6f90-cbd3-407a-9c9a-aac81e975a40 (instance_id=642c108e-5cbe-4569-bf43-b682905a090c)
[2023-05-23T21:08:07.063699793Z]  INFO: SledAgent/InstanceManager/691 on BRM42220026: Zone booting (instance_id=642c108e-5cbe-4569-bf43-b682905a090c)
    zone: oxz_propolis-server_7ffa6f90-cbd3-407a-9c9a-aac81e975a40
[2023-05-23T21:08:14.850305984Z]  INFO: SledAgent/InstanceManager/691 on BRM42220026: Adding address: Static(V6(Ipv6Network { addr: fd00:1122:3344:101::23, prefix: 64 })) (instance_id=642c108e-5cbe-4569-bf43-b682905a090c)
    zone: oxz_propolis-server_7ffa6f90-cbd3-407a-9c9a-aac81e975a40
[2023-05-23T21:08:15.437467031Z]  INFO: SledAgent/InstanceManager/691 on BRM42220026: Created address fd00:1122:3344:101::23/64 for zone: oxz_propolis-server_7ffa6f90-cbd3-407a-9c9a-aac81e975a40 (instance_id=642c108e-5cbe-4569-bf43-b682905a090c)
[...]

PR #4711, opened 2023-12-18
Move instances to Failed if sled agent returns an "unhealthy" error type from calls to stop/reboot

### LINKED ISSUE #4709: Instances no longer transition to failed state when propolis zone has crashed or is gone
The behavior change appears to be related to https://github.com/oxidecomputer/omicron/pull/4682. Prior to that, instances were moved to failed state by sled-agent or when user attempted to reboot the instance.

The ability to handle propolis zone gone situation may not be by design (though we rely on it quite a bit now during sled planned or unplanned reboot). The ability to handle propolis zone crashing due to unhandled hypervisor or OS exception is probably something the system should have.

gjcolombo:
Restore `instance_reboot` and `instance_stop` to their prior behavior: if these routines try to contact sled agent and get back a server error, mark the instance as unhealthy and move it to the Failed state.

Also use `#[source]` instead of message interpolation in `InstanceStateChangeError::SledAgent`.

This restores the status quo ante from #4682 in anticipation of reaching a better overall mechanism for dealing with failures to communicate about instances with sled agents. See #3206, #3238, and #4226 for more discussion.

Tests: new integration test; stood up a dev cluster, started an instance, killed the zone with `zoneadm halt`, and verified that calls to reboot/stop the instance eventually marked it as Failed (due to a timeout attempting to contact the Propolis zone).

Fixes #4709.

### DIFF
--- a/nexus/src/app/instance.rs
+++ b/nexus/src/app/instance.rs
@@ -65,7 +65,7 @@ type SledAgentClientError =
     sled_agent_client::Error<sled_agent_client::types::Error>;
 
 // Newtype wrapper to avoid the orphan type rule.
-#[derive(Debug)]
+#[derive(Debug, thiserror::Error)]
 pub struct SledAgentInstancePutError(pub SledAgentClientError);
 
 impl std::fmt::Display for SledAgentInstancePutError {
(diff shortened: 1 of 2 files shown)


gjcolombo:
This needs to be verified in a dev cluster before it will be ready for review.

gjcolombo:
This "works" (tested in a dev cluster now) but feels very dodgy--we're relying on the fact that if a Propolis zone disappears, then Nexus calls to sled agent that try to talk to Propolis will time out and produce a Progenitor error case that Nexus interprets as an "instance unhealthy" case. I think it does restore the regressed behavior from #4709, so I'm going to open this up for review, but I think our handling of these sorts of cases is very close to the breaking point...

smklein (approved):
As mentioned in DMs: I'm okay with this now, but we should really move the responsibility of "monitoring VMs and marking them as failed" to an RPW, because it's weird that we only update this info "when you asked to change the state anyway".

Happy to help with that work ASAP.

PR #4717, opened 2023-12-19
[mgs] Use slog-error-chain to clean up error types and logging

jgallagher:
These are all pretty mechanical changes:

* Use `#[source]` and `#[from]` correctly
* Add a bit of context to a few error cases (e.g., including the `SpIdentifier` when returning an `SpCommunicationFailed` error)
* Use `InlineErrorChain` instead of `anyhow` to convert error chains into strings (avoiding the intermediate `anyhow::Error` heap allocation)
* Switch to `Utf8PathBuf` for command line args and related errors

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4698,6 +4698,7 @@ version = "0.1.0"
 dependencies = [
  "anyhow",
  "base64",
+ "camino",
  "clap 4.4.3",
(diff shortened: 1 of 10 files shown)


andrewjstone (approved):
Nice cleanup!

PR #4718, opened 2023-12-19
fix Name regex in json schema

ahl:
Before this change, this string was rejected "a-B-c" and this string passed "a-".


zephraph (approved):
More tests would be helpful!

```regex
^[a-z]([a-zA-Z0-9-]*[a-zA-Z0-9]+)?
```

So the `+` at the end ensures there's _at least_ one character without a dash. Adding `A-Z` to the middle group ensures we can also use a mix of upper case and dashes after the first letter. Overall I think it makes sense, thanks for catching this.

ahl:
NB: we didn't actually use this regex for validation within nexus so this change should not affect existing data; the visible consequences will be:

- use of valid names that this regex excluded will now be permitted by validated endpoints such as the CLI or Rust SDK
- use of invalid names will be caught in validated clients such as the CLI and Rust SDK rather than generating a server-side validation failure

PR #4720, opened 2023-12-19
Fix fake pagination for sled_list_uninitialized

andrewjstone:


nexus/src/external_api/http_entrypoints.rs
```diff
@@ -4659,6 +4659,7 @@ async fn rack_view(
 }]
 async fn sled_list_uninitialized(
     rqctx: RequestContext<Arc<ServerContext>>,
+    _query_params: Query<PaginatedById>,
 ) -> Result<HttpResponseOk<ResultsPage<UninitializedSled>>, HttpError> {
     let apictx = rqctx.context();
     let handler = async {
```
ahl, line 4665: - I don't think we want to offer a `sort_by` query parameter that we ignore
- I'm fine ignoring the `limit` parameter, but perhaps we could use it below
- I think we want to produce an error if a caller specifies a page token

```suggestion
    query: Query<PaginationParams<EmptyScanParams, ()>>,
) -> Result<HttpResponseOk<ResultsPage<UninitializedSled>>, HttpError> {
    let pag_params = query.into_inner();
    if let dropshot::WhichPage::Next(last_seen) = &pag_params.page {
        return Err(Error::invalid_value(last_seen.clone(), "bad page token").into());
    };
    let apictx = rqctx.context();
    let handler = async {
```

openapi/nexus.json
```diff
@@ -4022,6 +4022,35 @@
...
+              "nullable": true,
+              "type": "string"
+            }
+          },
+          {
+            "in": "query",
+            "name": "sort_by",
+            "schema": {
+              "$ref": "#/components/schemas/IdSortMode"
+            }
```
ahl, line 4051: I think ideally this should not be here

openapi/nexus.json
```diff
@@ -4039,6 +4068,9 @@
           "5XX": {
             "$ref": "#/components/responses/Error"
           }
+        },
+        "x-dropshot-pagination": {
+          "required": []
```
ahl, line 4073: nice

ahl (commented):
I think this is very close; thanks for doing it

nexus/src/external_api/http_entrypoints.rs
andrewjstone: Thanks for the tip! Fixed in 06a385178.

openapi/nexus.json
andrewjstone: Fixed in 06a385178.

ahl (approved):
Lfgtm

PR #4725, opened 2023-12-21
Update Sidecar-lite, ignore ARP hostnames in SoftNPU init

FelixMcFelix:
Bumps to the most recent sidecar-lite, which includes a minor fix for manually added routes, and adds the `-n` flag when checking `arp` for the gateway's MAC address during SoftNPU init. This is really only needed if the gateway is also acting as a DNS server and names itself something cutesy, but it can come up:

```
kyle@farme:~/gits/omicron$ arp -a
Net to Media Table: IPv4
Device   IP Address               Mask      Flags      Phys Addr
------ -------------------- --------------- -------- ---------------
rge0   hub.home.arpa        255.255.255.255          b8:6a:f1:28:cd:00

...

kyle@farme:~/gits/omicron$ arp -an
Net to Media Table: IPv4
Device   IP Address               Mask      Flags      Phys Addr
------ -------------------- --------------- -------- ---------------
rge0   10.0.0.1             255.255.255.255          b8:6a:f1:28:cd:00
```

### DIFF
--- a/tools/create_virtual_hardware.sh
+++ b/tools/create_virtual_hardware.sh
@@ -63,7 +63,7 @@ function ensure_softnpu_zone {
             --omicron-zone \
             --ports sc0_0,tfportrear0_0 \
             --ports sc0_1,tfportqsfp0_0 \
-            --sidecar-lite-commit 45ed98fea5824feb4d42f45bbf218e597dc9fc58 \
+            --sidecar-lite-commit 6007a1b0ffe57ee738222867c343bde7e47bbf60 \
             --softnpu-commit dbab082dfa89da5db5ca2325c257089d2f130092
      }
     "$SOURCE_DIR"/scrimlet/softnpu-init.sh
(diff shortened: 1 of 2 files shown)


tools/create_virtual_hardware.sh
```diff
@@ -63,7 +63,7 @@ function ensure_softnpu_zone {
             --omicron-zone \
             --ports sc0_0,tfportrear0_0 \
             --ports sc0_1,tfportqsfp0_0 \
-            --sidecar-lite-commit 45ed98fea5824feb4d42f45bbf218e597dc9fc58 \
+            --sidecar-lite-commit 6007a1b0ffe57ee738222867c343bde7e47bbf60 \
```
rcgoodfellow, line 66: i don't see this commit in the sidecar-lite commit history.

FelixMcFelix: Ack, that was the pre-merge branch. Should be remedied now.

rcgoodfellow (approved):

PR #4732, opened 2023-12-27
Add a background task for update plan execution

andrewjstone:
Note: I expect this to lead to some discussion before getting merged.

This PR is the first step in creating a background task that is capable of taking a `Blueprint` and then reifying that blueprint into deployed or updated software. This PR includes the initial version of a blueprint that is only capable of describing deployable omicron zones. A basic executor that sends the related `OmicronZonesConfig` to the appropriate sled-agents for newly added sleds was created.

A test is included that shows how a hypothetical planner for an `add-sled` workflow will deploy Omicron zones in a manner similar to RSS, where first the internal DNS zone is deployed and then the internal DNS and NTP zones are deployed. Deployment alwyas contains all zones expected to be running on the sled-agent. Any zones running that are not included are expected to be shut down.

This background task is not yet wired up into `nexus/src/app/background/ init.rs`, because I believe this depends on a planner being added that is capable of providing Blueprints to drive the task forward. The first planner can be super simple and triggered from an operator HTTP request, but we probably still want a basic planner. One reason is that the planner itself will almost certainly want to wait for timesync before it proceeds to installing the rest of the zones.


nexus/src/app/background/plan_execution.rs
```diff
@@ -0,0 +1,327 @@
...
+    ) -> PlanExecutor {
+        PlanExecutor { rx_blueprint }
+    }
+}
+
+impl BackgroundTask for PlanExecutor {
+    fn activate<'a, 'b, 'c>(
+        &'a mut self,
+        opctx: &'b OpContext,
+    ) -> BoxFuture<'c, serde_json::Value>
```
sunshowers, line 43: Hmm, this can just be `activate<'a>(&'a mut self, opctx: &'a OpContext) -> BoxFuture<'a, serde_json::Value>`. That's because all the individual lifetime parameters here are [covariant](https://lifetime-variance.sunshowers.io/ch01-02-formalizing-variance.html).

nexus/src/app/background/plan_execution.rs
```diff
@@ -0,0 +1,327 @@
...
+        async {
+            // Get the latest blueprint, cloning to prevent holding a read lock
+            // on the watch.
+            let blueprint = self.rx_blueprint.borrow().clone();
+
+            let Some(blueprint) = blueprint else {
+                warn!(&opctx.log,
+                      "Plan execution: skipped";
+                      "reason" => "no blueprint");
+                return json!({"error": "no blueprint" });
```
sunshowers, line 57: Is this the right design for the trait method? I thought maybe this would return a `Result` type (though maybe this was discussed already).

nexus/src/app/background/plan_execution.rs
```diff
@@ -0,0 +1,327 @@
...
+    zones: &BTreeMap<Uuid, (SocketAddr, OmicronZonesConfig)>,
+) -> Result<(), Vec<anyhow::Error>> {
+    let errors: Vec<_> = stream::iter(zones.clone())
+        .filter_map(|(sled_id, (addr, config))| async move {
+            let client = sled_agent_client::Client::new(
+                &format!("http://{}", addr),
+                log.clone(),
+            );
+            let result =
+                client.omicron_zones_put(&config).await.with_context(|| {
```
sunshowers, line 101: For my understanding -- what are the semantics of `omicron_zones_put`? Will it remove any zones that are currently running but not in the new config?

nexus/src/app/background/plan_execution.rs
sunshowers: https://github.com/oxidecomputer/omicron/pull/4877

nexus/src/app/background/plan_execution.rs
davepacheco: The caller doesn't do anything different with the outcome.  (If that's surprising, the [design notes](https://github.com/oxidecomputer/omicron/blob/255cb83822098f80c2f60811152922259ca8e82c/nexus/src/app/background/common.rs#L65-L129) may shed some light on that part.)  And often it's not obvious how to classify the outcome.  Using DNS propagation as the canonical example: the task may propagate the latest DNS config to 3 out of 5 servers.  That's not obviously either a success or failure -- and fortunately, we don't need to decide since the caller doesn't care.  The most important consumer of this is a human looking at the system to understand what it did: "DNS generation X was propagated to these servers, but we couldn't reach these other servers and here are the errors we encountered talking to each one."  [For that kind of summary, the success case looks pretty similar to the case where some errors were encountered.](https://github.com/oxidecomputer/omicron/blob/255cb83822098f80c2f60811152922259ca8e82c/nexus/src/app/background/dns_propagation.rs#L93-L125)  So right now this is just a JSON value to give maximum flexibility.  (omdb just knows how to interpret those.)  I don't love it -- obviously this could be a more typed value, but the cost seemed to me to outweigh the benefit here.  If we end up wanting programmatic consumers of this information, we might want to revisit, but then we'll have to deal with the fact that background tasks return different types, even though we want to treat them uniformly, etc.

(I think this is also outside the scope of the PR, but it's a good question!)

sunshowers: That makes sense! Sorry for the distraction; this is definitely a hard problem.

nexus/src/app/background/plan_execution.rs
andrewjstone: Correct. `omicron_zones_put` will create a zone if one doesn't already exist in the set, and remove any that are not in the new config.

andrewjstone:
Closing in favor of #4891

Issue #4733, opened 2023-12-28
Make `nexus_test` tests less slow


smklein:
I noticed that a bunch of our tests are taking several seconds to run, even when doing little work, which seems slow. Let's dig into "why".

I'm providing data from running:

`cargo nt --no-capture -- integration_tests::basic::test_ping`

This is a relatively simple test that hits and endpoint and exits - on my machine, it's taking ~9 seconds, **after** initial CRDB population. If this could be optimized, it would have a positive multiplicative on our test suite.

A good portion of this test setup seems to consist of: Start Nexus, run internal jobs, populate initial data, etc, which also uses authz checks (and therefore triggers additional database workload).

Popping open the CockroachDB web API, and sorting by "% of All Runtime" I noticed the following: 

![image](https://github.com/oxidecomputer/omicron/assets/3258857/d79dc229-d4d4-426b-9441-b693a9e95a6a)

There is a query consuming the vast majority of our runtime here:

```sql
SELECT pg_type.oid, pg_type.typarray FROM pg_type WHERE pg_type.oid = quote_ident($1)::REGTYPE::OID LIMIT $1
```

I don't know if it's blocking and sequential, but if it is, it represents a ~3.5 second buffer of time in all invocations of `#[nexus_test]`.

I believe this query originates from here: https://github.com/diesel-rs/diesel/blob/ca3ba763746d2312e23de98d57f99c353dcc0fc2/diesel/src/pg/metadata_lookup.rs#L101-L131

This function is used within implementations of https://docs.diesel.rs/master/diesel/sql_types/trait.HasSqlType.html -- specifically, it's called in the [derivation of SqlType](https://github.com/diesel-rs/diesel/blob/ca3ba763746d2312e23de98d57f99c353dcc0fc2/diesel_derives/src/sql_type.rs#L107-L116).

This seems to be derived mostly for our SQL `ENUM` types.

Some ideas to mitigate:

- This looks like it's cached on each connection, but we're using a connection pool with the default number of sessions (according to bb8, 10. This matches what I see on the CRDB web console). If we could share this metadata between sessions, we'd only need to look it up once.
- This information looks like it's being queried from diesel "one OID at a time". We do know all the enums used in a schema ahead-of-time, so we could pre-populate this cache with a batch operation to fetch all of them?

PR #4735, opened 2023-12-29
Optimize OID lookup for user-defined types

smklein:
# Background

Looking up OIDs for ENUMs consumes a non-trival amount of time during initialization, since they are synchronously fetched one-by-one. Although this is amortized for long-lived connections, it incurs a cost on all tests using CRDB, and can be optimized. Additionally, these OID caches are per-connection, which incurs redundant work when we have a connection pool (and we do, with bb8). See #4733 for additional context.

Feedback from the Diesel maintainers:

> I think it would be reasonable to share this information at pool level...
>
> [OIDs are] currently queried by one type at the time as we argued that it won't matter for long running applications. There you do the overhead once per connection and then you just hit the cache. I don't think we did really consider short running applications (like tests) for that.

# This PR

1. Performs a one-time-lookup of user-defined ENUM OIDs on the first connection, and shares it between subsequent connections. This pre-populates the cache used by Diesel.
2. Modifies all `SqlType` derives to ensure that the `schema` is specified as `public` (this ensures a cache hit)
3. Adds a test to compare the "list of user-defined enums to look up" with the current DB schema, and ensures they're in-sync.

# Results

I focused on optimizing `integration_tests::basic::test_ping`, as this is very close to a "no-op" version of the `#[nexus_test]`. I wrote a script to repeatedly execute this test in a loop (excluding the first result, as a "warm-up" run to populate CRDB).

- Before: Average time over 10 runs: 10.42 seconds
- After: Average time over 10 runs: 9.47 seconds

That's a small difference, but hopefully it'll provide a dent over "all our tests", and avoid additional user-defined types from slowing down all tests.

Part of https://github.com/oxidecomputer/omicron/issues/4733


smklein:
Checking the last few commits to main:

- build-and-test (ubuntu-22.04) Successful in 52m - Success!
- build-and-test (ubuntu-22.04) Successful in 49m - Success!
- build-and-test (ubuntu-22.04) Successful in 50m - Success!

And from nextest:

```
5206	2023-12-22T09:02:24.159Z	     Summary [1982.003s] 970 tests run: 970 passed (13 slow, 1 leaky), 3 skipped
5207	2023-12-22T09:02:24.164Z		Command being timed: "timeout 2h cargo nextest run --profile ci --locked --verbose"
5208	2023-12-22T09:02:24.168Z		User time (seconds): 9423.74
5209	2023-12-22T09:02:24.168Z		System time (seconds): 1377.74
5210	2023-12-22T09:02:24.168Z		Percent of CPU this job got: 543%
5211	2023-12-22T09:02:24.168Z		Elapsed (wall clock) time (h:mm:ss or m:ss): 33:06.69
```

This PR:
- build-and-test (ubuntu-22.04) Successful in 47m — Success!

```
5202	2023-12-29T00:46:42.395Z	     Summary [1853.552s] 972 tests run: 972 passed (9 slow, 1 leaky), 3 skipped
5203	2023-12-29T00:46:42.405Z		Command being timed: "timeout 2h cargo nextest run --profile ci --locked --verbose"
5204	2023-12-29T00:46:42.408Z		User time (seconds): 7717.26
5205	2023-12-29T00:46:42.408Z		System time (seconds): 1386.81
5206	2023-12-29T00:46:42.408Z		Percent of CPU this job got: 489%
5207	2023-12-29T00:46:42.408Z		Elapsed (wall clock) time (h:mm:ss or m:ss): 30:59.41
```

Seems like a small but "probably worth it" step - wall time drops by a couple minutes, and the total "User time" drops significantly.

andrewjstone (commented):
@smklein This looks good. I like the new abstraction and the time savings are great. Thank you for doing this!

My one concern is about using this in production upgrade scenarios once we are running multiple versions of CRDB. I could guess at a couple of possible scenarios where we could mistakenly treat one enum value for another if the OIDs change. According to https://www.cockroachlabs.com/docs/v23.1/oid, OIDs are not guaranteed to be stable across versions. Could they also change with an alter table statement? For the first issue of version changes, diesel's original connection caching should continue to work. For the latter, even diesel as it works per connection would have this issue I believe.

I have no idea if these are actual problems that can exist, but figured I'd bring it up here as I was thinking about it.

smklein:
> My one concern is about using this in production upgrade scenarios once we are running multiple versions of CRDB. I could guess at a couple of possible scenarios where we could mistakenly treat one enum value for another if the OIDs change. According to https://www.cockroachlabs.com/docs/v23.1/oid, OIDs are not guaranteed to be stable across versions. Could they also change with an alter table statement? For the first issue of version changes, diesel's original connection caching should continue to work. For the latter, even diesel as it works per connection would have this issue I believe.

These are good questions.

- On "OID stability across versions": I do agree that this is a potential problem. I tried to address this in fdf8dffa29b5a4e2edb0e3fafbfc0b0b7542b314 , by explicitly key-ing off the CRDB version, so "each time we access a new CRDB version, we treat it as a distinct cache". This gives me a *slightly* slower "average" time for the ping test: **Average time over 10 runs: 9.65 seconds.** However, this seems like a worthwhile trade-off, and should mitigate the potential instability across CRDB releases.

- On "ALTER TABLE statements": I *think*, reading https://github.com/cockroachdb/cockroach/pull/33697 , that these would be stable? As a reminder, the OID lookup isn't happening for arbitrary tables, but rather, just for user-created types, which is a slightly different scope -- I think this would actually be relevant for an `ALTER TYPE` statement instead.

I tried this manually on a single-node CRDB, and "adding a value, removing a value, and renaming the type" for an ENUM does not appear to change the OID. This isn't exhaustive, but at least it isn't trivially broken.

andrewjstone:
> > My one concern is about using this in production upgrade scenarios once we are running multiple versions of CRDB. I could guess at a couple of possible scenarios where we could mistakenly treat one enum value for another if the OIDs change. According to https://www.cockroachlabs.com/docs/v23.1/oid, OIDs are not guaranteed to be stable across versions. Could they also change with an alter table statement? For the first issue of version changes, diesel's original connection caching should continue to work. For the latter, even diesel as it works per connection would have this issue I believe.
> 
> These are good questions.
> 
> * On "OID stability across versions": I do agree that this is a potential problem. I tried to address this in [fdf8dff](https://github.com/oxidecomputer/omicron/commit/fdf8dffa29b5a4e2edb0e3fafbfc0b0b7542b314) , by explicitly key-ing off the CRDB version, so "each time we access a new CRDB version, we treat it as a distinct cache". This gives me a _slightly_ slower "average" time for the ping test: **Average time over 10 runs: 9.65 seconds.** However, this seems like a worthwhile trade-off, and should mitigate the potential instability across CRDB releases.
> * On "ALTER TABLE statements": I _think_, reading [sql: make pg_catalog OIDs stable and refer to the descriptor IDs directly cockroachdb/cockroach#33697](https://github.com/cockroachdb/cockroach/pull/33697) , that these would be stable? As a reminder, the OID lookup isn't happening for arbitrary tables, but rather, just for user-created types, which is a slightly different scope -- I think this would actually be relevant for an `ALTER TYPE` statement instead.
> 
> I tried this manually on a single-node CRDB, and "adding a value, removing a value, and renaming the type" for an ENUM does not appear to change the OID. This isn't exhaustive, but at least it isn't trivially broken.

Wow, quick fix for the first issue! Thanks!

As for the second issue, you are right that `ALTER TYPE` is what we should be concerned about potentially. With that test, things seem to work as one would hope :)

andrewjstone (approved):
Looks great. Thank you @smklein!

PR #4742, opened 2024-01-02
Fix endpoint description casing

benjaminleonard:
Noticed some inconsistencies with the endpoint descriptions. Fixes https://github.com/oxidecomputer/docs/issues/253


zephraph:
Thanks for doing this!

zephraph (approved):

PR #4754, opened 2024-01-04
sled agent client could use some primitives from omicron_common

davepacheco:
sled_agent_client generates types like `Generation` and `ByteCount` that ultimately come from omicron_common.  These are more like primitive types than API-specific types.  I think it makes more sense to re-use the omicron_common ones here.

There are a bunch of other types here from omicron_common that could get the same treatment: all the instance, disk, and net-related ones.  I have _not_ done that here.  These seem like the sorts of things we might well expect to evolve over time and we do want to be more careful about them, not just assume our local copy matches the server's.

### DIFF
--- a/clients/sled-agent-client/src/lib.rs
+++ b/clients/sled-agent-client/src/lib.rs
@@ -25,6 +25,9 @@ progenitor::generate_api!(
     //TODO trade the manual transformations later in this file for the
     //     replace directives below?
     replace = {
+        ByteCount = omicron_common::api::external::ByteCount,
+        Generation = omicron_common::api::external::Generation,
+        Name = omicron_common::api::external::Name,
         SwitchLocation = omicron_common::api::external::SwitchLocation,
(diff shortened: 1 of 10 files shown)


clients/sled-agent-client/src/lib.rs
```diff
@@ -25,6 +25,9 @@ progenitor::generate_api!(
     //TODO trade the manual transformations later in this file for the
     //     replace directives below?
```
ahl, line 26: it seems like there are still quite a few of them, so I assume this comment is no less applicable than before your change; is that right?

ahl (approved):
makes sense.

clients/sled-agent-client/src/lib.rs
davepacheco: That's right.  Much of what's left are stuff I wouldn't consider "primitives" (e.g., InstanceRuntimeState).  There's some net stuff that could get the same treatment but I wasn't sure so I left it alone.

PR #974, opened 2022-04-25
Pass SSH public keys through to cloud-init

plotnick:
Builds on #911 by plumbing SSH public keys through to the cloud-init meta-data.

Right now, we collect only the keys belonging to the actor making the request. Once role assignments for Silo users are in place, we can ask for, e.g., all the keys belonging to all the users with an appropriate role (`vms.instance.create`, probably) on the instance.

It would be nice to have an automated test for this, but I think we can take that as a follow-up. Tested manually with local setup using Debian-cloud image from [this gist](https://gist.github.com/iliana/7911bd754b0bb3cf21fe15c16601bd9a).

PR #4764, opened 2024-01-05
User specified SSH keys to inject at instance create time

### LINKED ISSUE #3056: Let user specify SSH keys to inject at instance create time
The current implementation automatically [retrieves](https://github.com/oxidecomputer/omicron/blob/6f8cc328740b0f535df01c91efc38108a34edad4/nexus/src/app/instance.rs#L713-L748) and [inserts](https://github.com/oxidecomputer/omicron/blob/6f8cc328740b0f535df01c91efc38108a34edad4/nexus/src/cidata.rs#L17-L36) a user's SSH keys into the instances they create. However, it would be beneficial for users to have control over which keys are added, as well as the ability to pass an SSH key not associated with their account.

I propose adding an option to the [instance_create](https://docs.oxide.computer/api/instance_create) API that accepts an array of attached SSH keys identified by name or ID, or a custom SSH key provided as a string (which will not be linked to the user's account).

For an enhanced user experience, we could also include an option that automatically incorporates a user's SSH keys. This would be particularly useful for the CLI, where retrieving SSH keys and adding them as an option is more cumbersome.

It might be reasonable to defer implementation until after FCS. In the meantime, I could work on adding a console message to inform users that all their SSH keys will be injected into the instance (where cloud-init is supported).

_Proposed console designs:_
image

benjaminleonard:
Fixes #3056

This PR adds an optional `ssh_keys` parameter to the instance create payload. If it's empty the same behavior as we have today is retained: all ssh keys from a user's profile are transferred over. If a vector of ssh key name or IDs are passed then only transfers over those given keys. If an empty vector is passed no ssh keys will be transferred. 

There's also a `/v1/instances/<instance>/ssh-public-keys` API that lists all public keys that were sent via cloud-init to the instance at instance create. I talked to @iliana a bit about the proper response to this API. Currently it returns the full `SshKey` view. We may want to alter that to transmit fewer details. 

Other changes:
- Pulled SSH keys app code into its own file

```[tasklist]
### Remaining Work
- [x] Wire up join table model
- [x] Update instance create saga to associate keys
- [x] Update key registry to be a simple list
- [x] Do something with the limit
- [x] tests
```

### Followup Work
```[tasklist]
- [ ] https://github.com/oxidecomputer/omicron/issues/4866
- [ ] https://github.com/oxidecomputer/omicron/issues/4918
- [ ] https://github.com/oxidecomputer/omicron/issues/4919
```


nexus/test-utils/src/resource_helpers.rs
```diff
@@ -410,6 +410,7 @@ pub async fn create_instance_with(
             user_data:
                 b"#cloud-config\nsystem_info:\n  default_user:\n    name: oxide"
                     .to_vec(),
+            public_keys: Some(Vec::new()),
```
david-crespo, line 413: this seems like the use case for `None` unless that has a special meaning

nexus/types/src/external_api/params.rs
```diff
@@ -1000,9 +1000,9 @@ pub struct InstanceCreate {
     #[serde(default)]
     pub disks: Vec<InstanceDiskAttachment>,
 
-    /// List of public keys (identified by ID or name) to be associated with the instance.
+    /// List of public SSH keys (identified by ID or name) to be associated with the instance.
     #[serde(default)]
-    pub public_keys: Option<Vec<NameOrId>>,
+    pub ssh_keys: Option<Vec<NameOrId>>,
```
karencfv, line 1005: I think I like the field to specify that the keys are public (purely from a security point of view), how about `public_ssh_keys` instead?

jclulow: If you're going to do that, it should be **ssh_public_keys** probably.  **Public Key** is the thing and SSH is a qualifier.

david-crespo: I like that, I like having both adjectives in there.

zephraph: Something I talked through with with @charliepark too. I think whatever we do, it's worth keeping the resource label consistent across usages.

zephraph: I'd note that GitHub just calls them SSH Keys. I think (hope) the _public_ aspect is implied and that folks aren't in the habit of submitting private keys. 

I think we could go further than this and augment our clients to _explicitly_ reject keys formatted like a private key. They typically have beginning/ending tags of various forms and we can reject based on the presence of that.

karencfv: > I'd note that GitHub just calls them SSH Keys. I think (hope) the public aspect is implied and that folks aren't in the habit of submitting private keys.

Is there any harm in being explicit and calling them `ssh_public_keys`? As far as I'm aware we're expecting to have users with varying degree of technical expertise. Being explicit can be helpful.

nexus/db-model/src/schema.rs
```diff
@@ -712,6 +712,13 @@ table! {
     }
 }
 
+table! {
+    instance_ssk_key (instance_id, ssh_key_id) {
```
benjaminleonard, line 716:
```suggestion
    instance_ssh_key (instance_id, ssh_key_id) {
```

(unless I'm missing something and this is deliberate)

nexus/db-queries/src/db/datastore/ssh_key.rs
```diff
@@ -21,11 +21,98 @@ use diesel::prelude::*;
...
+                    }
+                }
+            }
+        }
+
+        return Ok(result.iter().map(|&(id, _)| id).collect());
+    }
+
+    /// Given a list of IDs for SSH public keys, fetches the keys that belong to
+    /// the user and aren't deleted. Does not fail if keys are missing.
```
benjaminleonard, line 96: I like this approach, handles missing keys on instance create without being unnecessarily brittle elsewhere.

nexus/db-model/src/schema.rs
zephraph: that was not deliberate! Thanks for the catch

nexus/types/src/external_api/params.rs
zephraph: The rename touches a lot of files so I've pulled it out into a separate issue to be done in a follow up PR: https://github.com/oxidecomputer/omicron/issues/4866

nexus/db-queries/src/db/datastore/ssh_key.rs
```diff
@@ -21,11 +21,200 @@ use diesel::prelude::*;
...
+            .select((dsl::id, dsl::name))
+            .get_results_async(&*self.pool_connection_authorized(opctx).await?)
+            .await
+            .map_err(|e| public_error_from_diesel(e, ErrorHandler::Server))?;
+
+        // If a key isn't present in the result that was present in the input that means it either
+        // doesn't exist or isn't owned by the user. Either way we want to give a specific lookup error
+        // for at least the first result. It would be nice to include an aggregate error with all the missing
+        // keys.
+        // TODO: Optimize the performance of this
```
karencfv, line 68: Can you open an issue to track these TODOs and copy the issue number here?

nexus/db-queries/src/db/datastore/ssh_key.rs
```diff
@@ -21,11 +21,200 @@ use diesel::prelude::*;
...
+        opctx: &OpContext,
+        authz_user: &authz::SiloUser,
+        instance_id: Uuid,
+        keys: &Option<Vec<Uuid>>,
+    ) -> UpdateResult<()> {
+        opctx.authorize(authz::Action::ListChildren, authz_user).await?;
+
+        let instance_ssh_keys: Vec<db::model::InstanceSshKey> = match keys {
+            // If the keys are None, use the fallback behavior of assigning all the users keys
+            None => {
```
karencfv, line 131: Would this not confuse users thinking that `None` means add no keys? Perhaps we may want to revisit why the default behaviour is to add all keys.

There are a few reasons we may want to consider this default behaviour:
- From a security standpoint, people may have forgotten to remove keys from their user account that have been compromised.
- Additionally, users may not be aware that `None` means all, it doesn't feel particularly intuitive (often people don't read the docs!). They may end up adding keys they did not intend to add, and be unaware of it if the instance was not created through the console.
- Do we expect our customers to particularly _want_ to have all keys automatically added to the instances they create?

I think it'd also be good to have some input from @ahl and/or @askfongjojo  on this

nexus/src/app/mod.rs
```diff
@@ -100,6 +101,8 @@ pub const MAX_MEMORY_BYTES_PER_INSTANCE: u64 = 256 * (1 << 30); // 256 GiB
 pub const MIN_DISK_SIZE_BYTES: u32 = 1 << 30; // 1 GiB
 pub const MAX_DISK_SIZE_BYTES: u64 = 1023 * (1 << 30); // 1023 GiB
 
+pub const MAX_SSH_KEYS_PER_INSTANCE: u32 = 10;
```
karencfv, line 104: Why 10? Perhaps it would be good to add a comment here with the reasoning behind this number.

nexus/src/app/sagas/instance_create.rs
```diff
@@ -238,6 +244,17 @@ impl NexusSaga for SagaInstanceCreate {
...
+        // let mut sagabuilder = DagBuilder::new(subsaga_name);
+        // subsaga_builder.append(Node::action("output", "AssociateSshKeys", ASSOCIATE_SSH_KEYS.as_ref(),
+        // ));
+        // subsaga_append(
+        //     "ssh_keys".into(),
+        //     subsaga_builder.build()?,
+        //     &mut builder,
+        //     params.clone(),
+        //     0,
+        // )?;
```
karencfv, line 257: Should this be deleted?

nexus/src/app/sagas/instance_create.rs
```diff
@@ -291,6 +308,77 @@ impl NexusSaga for SagaInstanceCreate {
...
+
+    let opctx = crate::context::op_context_for_saga_action(
+        &sagactx,
+        &saga_params.serialized_authn,
+    );
+    let instance_id = sagactx.lookup::<Uuid>("instance_id")?;
+
+    // Gather the SSH public keys of the actor make the request so
+    // that they may be injected into the new image via cloud-init.
+    // TODO-security: this should be replaced with a lookup based on
```
karencfv, line 326: As above. Can you open an issue to track this and reference it here so this isn't forgotten about?

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -2666,6 +2667,50 @@ async fn instance_serial_console_stream(
     }
 }
 
+/// List the SSH public keys sent to the instance via cloud-init during instance creation
```
karencfv, line 2670: nit :
```suggestion
/// List the SSH public keys added to the instance via cloud-init during instance creation
```

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -2666,6 +2667,50 @@ async fn instance_serial_console_stream(
...
+///
+/// Note that this list is a snapshot in time and will not reflect updates made after
+/// the instance is created.
```
karencfv, line 2673: Nice! I like this clarification

schema/crdb/dbinit.sql
```diff
@@ -827,6 +827,19 @@ CREATE UNIQUE INDEX IF NOT EXISTS lookup_ssh_key_by_silo_user ON omicron.public.
 ) WHERE
     time_deleted IS NULL;
 
+/**
+ * Represents the SSH keys copied to an instance at create time by cloud-init.
+ * Entries are added here when an instance is created (with configured SSH keys)
+ * and removed when the instance is destroyed.
+ *
+ * TODO: Should this have time created / time deleted
```
karencfv, line 835: Just commenting here since the PR is set as ready to review. This is intended to be added in this PR yes?

karencfv (commented):
I'm really looking forward to this feature! Thanks a bunch for taking the time to flesh it out so thoroughly

nexus/db-queries/src/db/datastore/ssh_key.rs
karencfv: Also, if users have more than 10 keys saved in their user account, every attempt at creating an instance with the default behaviour of adding all keys would result in an error. 

This means that someone with more than 10 keys would have to set which keys they want added every single time to avoid a failure.

nexus/src/app/sagas/instance_create.rs
zephraph: It should be, thanks!

schema/crdb/dbinit.sql
zephraph: Potentially, but not necessarily. Currently I have this relationship configured to be hard deleted when the instance is (soft) deleted. This was a weird case being that it's both a join _and_ not something a user can directly change. It's created at instance create and deleted at instance delete. 

If I _do_ need to add the deleted timestamp, I certainly will.

nexus/db-model/src/schema.rs
```diff
@@ -713,6 +713,13 @@ table! {
     }
 }
 
+table! {
+    instance_ssh_key (instance_id, ssh_key_id) {
+        instance_id -> Uuid,
+        ssh_key_id -> Uuid,
```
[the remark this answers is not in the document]
zephraph: I wasn't planning on modifying the non-public references.

nexus/db-queries/src/db/datastore/ssh_key.rs
zephraph: Will do

nexus/src/app/mod.rs
zephraph: Thanks for pointing this out, I meant to update it. The limit is absolutely arbitrary which I should comment as such. It's 8 in main (and silently ignores any SSH key over 8 in your profile when you're transferring it over). I've had a few conversations about these limits. There are a few places they are relevant.

1. What's the maximum number of keys a user can create in their account?
2. What's the maximum number of keys we want to support transferring to the instance?

If 2 is less than 1 then there's an edge case for users who have large amounts of SSH keys that want to just send them all to the instance. In the near term it might actually be good just to unify these limits. Just limit it to say... 100 SSH keys per account/instance (which was actually suggested by @iliana [in chat](https://matrix.to/#/!oSMXTSrgambaNFEemN:oxide.computer/$LuhOLmLVhoQNI7unHloUI_C66po73zfGHQJUWwcQWnc?via=oxide.computer&via=unix.house&via=matrix.org))

nexus/src/app/sagas/instance_create.rs
zephraph: I didn't add this TODO, just copied it over from where it was originally. I don't actually know what the original intent is here and I somewhat assume it's out of date. @davepacheco, I think you originally wrote this. Can you give any more context? I'm happy to file an issue, but I'd just like to understand more what the replacement should be (if it's even still relevant)

nexus/db-queries/src/db/datastore/ssh_key.rs
zephraph: Heh, yeah, I've had a lot of conversations about this. @iliana and I [discussed](https://matrix.to/#/!oSMXTSrgambaNFEemN:oxide.computer/$SN8Tm7fxeQwbmlZhkrIua6pRWIwrWK8OEpnvl5v3QwQ?via=oxide.computer&via=unix.house&via=matrix.org) this in chat. This implementation was xer original suggestion. 

There's a few things to consider here. 
1. Should this be a required field? If it's required then it's one extra thing folks _have_ to specify when creating an instance. Originally I was worried about the repercussions of this decision (and others like it) on the CLI. That's mostly moot given that most args have to be provided via a json file. If it's _not_ required, I think the only sane default is to send them all (as we do today). Having the default mean none get sent would lead to workflow situations where users forget and then lock themselves out (which is worse in my mind than having to go and remove some keys manually after the fact)
2. Should we even have the option for users to send all keys to the instance? I feel like _yes_ is probably the answer here because I suspect most users will have few keys. That's also another reason I've leaned towards this behavior as the default. If we didn't preserve this option then users would _always_ have to specify keys by name/id in the API (lest they can't SSH into the instance) and that feels like a pretty big DX regression. That said, I would appreciate more feedback from folks on this.

So, yeah, an explicit enum having `AllFromProfile | Only[keys] | NoKeys` is certainly an option I can implement in this PR if that's the consensus. I do see the value of it, especially around _forcing_ users to make a choice. If the choice that they're forced to make is often the same as the default of this PR though, that feels not great.

iliana: I really don't like breaking old clients. Requiring an explicit enum will be a flag day for any existing code. This is, I think, an extremely good candidate for being our first version bump on an API. We did give ourselves a number of backstops to be able to declare a new version of the API, but we haven't used any of them yet, so that sort of explodes the work needed to be done here.

If we want to land this in the immediate term, we should keep the implicit behavior to use all of the user's keys if the field is not present (i.e. `None`). But ideally we have a clearer v2 either way.

karencfv: Thanks for the comments both! These are my thoughts:

> Should this be a required field? If it's required then it's one extra thing folks have to specify when creating an instance. Originally I was worried about the repercussions of this decision (and others like it) on the CLI. That's mostly moot given that most args have to be provided via a json file. If it's not required, I think the only sane default is to send them all (as we do today). Having the default mean none get sent would lead to workflow situations where users forget and then lock themselves out (which is worse in my mind than having to go and remove some keys manually after the fact)

To be perfectly honest, I don't see this being a required field as a problem. IMO it's better to be explicit about what you're doing than get surprises down the line. Alternatively, if we don't want this to be a required field, we could also have a concept of a "default" SSH key which gets added to any instance where SSH keys haven't been defined. WDYT?

We also need to think of the use case where new users are migrating their workloads from their previous vendor. For example: a user that didn't have support for adding SSH keys at create time (*cough* VMware *cough*) will have had to either have a user data init script that creates authentication, or have an image with a predefined user, and add additional SSH keys later. In this scenario, users may want to initially conserve their current workflows as part of a big lift and shift operation. These migrations are no small thing. and we really need to think about how to make it as seamless as possible for our customers to migrate over without unintended consequences.

Automatically adding every SSH key the user has saved feels quite nuclear, and can lead to easily giving unintended access. Especially if you're in the type or role where your job is to set up infrastructure for others to use and have saved other people's public keys in your account.

FWIW, I don't see being locked out of your instance as a necessarily terrible thing. This is something the user would automatically notice and would only need to delete and recreate their instance.

> Should we even have the option for users to send all keys to the instance? I feel like yes is probably the answer here because I suspect most users will have few keys. That's also another reason I've leaned towards this behavior as the default. If we didn't preserve this option then users would always have to specify keys by name/id in the API (lest they can't SSH into the instance) and that feels like a pretty big DX regression. That said, I would appreciate more feedback from folks on this.

I wouldn't necessarily say most users would have a small amount of keys, it really depends. Also, users that manage infrastructure will have ssh keys of others. So it's not just how many keys are saved, but also to whom those keys are giving access to. Frankly, at this point I don't think we have enough information about what most users will be doing. To me the big questions for now are:

- Does anyone really need to have every key they have saved added to their instance on create?
- What is more annoying? Adding a flag to define ssh keys on instance create, or having to go into an instance and remove SSH keys that you realised you really shouldn't have in there later down the track.
- Which repercussions are worse? Unintentionally giving access to keys that shouldn't have this access (could be compromised keys the user forgot to remove or people who shouldn't have access), or having to recreate an instance because you forgot to define a key and got locked out (this could be mitigated by having a "default" ssh key).

Additionally, by automatically adding all keys we'd be making it harder for our users to follow the [principle of least privilege](https://en.wikipedia.org/wiki/Principle_of_least_privilege) 

> I really don't like breaking old clients. Requiring an explicit enum will be a flag day for any existing code. This is, I think, an extremely good candidate for being our first version bump on an API. We did give ourselves a number of backstops to be able to declare a new version of the API, but we haven't used any of them yet (as far as I'm aware?), so that sort of explodes the work needed to be done here.

As for the concern about breaking APIs, I think the consensus is that now is the moment to be breaking the API to make sure we get it right. Posting a recent comment here -> https://github.com/oxidecomputer/omicron/pull/4694#discussion_r1442402379

benjaminleonard: I agree on breaking the API and including no keys if none are defined.

However making it required might be a poor user experience in the CLI. If the user doesn't want to include keys, they shouldn't need that argument.

zephraph: I appreciate all the discussion here. @karencfv you bring up a lot of great points and I feel like there's just ultimately some more product discovery we need to do here. 

I have higher priorities than this PR and I've already sank too much time into it. I'm just going to create an issue to encapsulate the context here and we can solve this in a follow up. The default behavior as implemented is the same as what we have today so it shouldn't be disruptive to include in this release while also giving us the time we need to figure out how we want this to be.

zephraph: Here's the follow up issue to tackle this work: https://github.com/oxidecomputer/omicron/issues/4919

nexus/src/app/sagas/instance_create.rs
davepacheco: I'm not sure.  Looking through the history I believe this came in with #974.  Maybe @plotnick remembers?

plotnick: Zowie, that was a while ago; thanks, @davepacheco for digging up that PR (#974). That TODO is mentioned in its description:

> Right now, we collect only the keys belonging to the actor making the request. Once role assignments for Silo users are in place, we can ask for, e.g., all the keys belonging to all the users with an appropriate role (vms.instance.create, probably) on the instance.

I don't remember exactly where that desire came from, other than generally wanting to allow customers to assign permissions at the role/group level rather than for individual users (as the current lookup would require); but we didn't have any kind of roles at the time, so couldn't do it then. If it's still desirable and we now have something like Silo/IdP roles (I've been away from that layer of Omicron for a while, so don't really know), it could certainly be spun out into an issue.

nexus/src/app/sagas/instance_create.rs
```diff
@@ -291,6 +297,77 @@ impl NexusSaga for SagaInstanceCreate {
...
+    let datastore = osagactx.datastore();
+    let saga_params = sagactx.saga_params::<Params>()?;
+
+    let opctx = crate::context::op_context_for_saga_action(
+        &sagactx,
+        &saga_params.serialized_authn,
+    );
+    let instance_id = sagactx.lookup::<Uuid>("instance_id")?;
+
+    // Gather the SSH public keys of the actor make the request so
```
plotnick, line 313: Old typo:
```suggestion
    // Gather the SSH public keys of the actor making the request so
```

karencfv (approved):
Thanks for creating follow up issues for product discovery for instance SSH keys. I'm OK with revising all of that at a later date, and merging this PR (once the tests pass) for now so we can introduce new functionality in this release.

Thanks again for taking the time to look into this!

end-to-end-tests/src/instance_launch.rs
```diff
@@ -132,7 +132,10 @@ async fn instance_launch() -> Result<()> {
         .split_once("-----BEGIN SSH HOST KEY KEYS-----")
         .and_then(|(_, s)| s.split_once("-----END SSH HOST KEY KEYS-----"))
         .and_then(|(lines, _)| {
-            lines.trim().lines().find(|line| line.starts_with("ssh-ed25519"))
+            lines
+                .trim()
+                .lines()
+                .find(|line| line.contains(public_key_str.clone().as_str()))
```
iliana, line 138: This doesn't seem correct. This section of code is trying to extract the host's SSH key from the console output, so we can authenticate the host when we try to connect to it. That is not the same key as the one we created and sent to the instance to authenticate our user.

If we are failing to extract a host SSH key from the console output, either we aren't waiting long enough or something else is wrong.

zephraph: It was extracting a key, but the key it was extracting was wrong. I thought that for some reason there were multiple.

iliana:
Looking at [the deploy test run for 88ddaa6](https://github.com/oxidecomputer/omicron/runs/20999988955), [the failure](https://buildomat.eng.oxide.computer/wg/0/details/01HNBHCAWK7BGG1SXQJJ0V4M8F/1GvfPSRUYCOfivpe1OsPa15HY8rGTHM5AaNATnHVZHCX6kls/01HNC5110Y8GMFX8AW2MK4J15K#S1041) does not implicate the host key parsing. The relevant snippet of stdout:

```
generate SSH key
create SSH key: ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIBx78Gk6UIlPOJT7JC888SVHUOKwFJ29yJp2cCOv1TVZ
create disk
create instance
instance external IP: 10.151.1.176
waiting for serial console
host key: ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIHeStG8cgDDVntEMVlMHWkDlwZhMQFBlkja61JQk9gCi
connecting ssh
authenticating ssh
Error: authentication failed
```

This implicates this code:

https://github.com/oxidecomputer/omicron/blob/88ddaa630b0e20fefaed0216fe0d42ed7e56aed3/end-to-end-tests/src/instance_launch.rs#L148-L159

Which implies the connection was established, the host key was authenticated, and what failed to occur was authenticating with our private key. If we extract the `cloud_init_bytes` from [the sled-agent log](https://buildomat.eng.oxide.computer/wg/0/artefact/01HNBHCAWK7BGG1SXQJJ0V4M8F/1GvfPSRUYCOfivpe1OsPa15HY8rGTHM5AaNATnHVZHCX6kls/01HNC5110Y8GMFX8AW2MK4J15K/01HNC6F4MTMR0WT9VD1QAMH5MJ/oxide-sled-agent:default.log?format=x-bunyan#L1783), the key is present in meta-data:

```
debian@meow:~$ sudo mount ./wat.img /mnt
debian@meow:~$ cat /mnt/meta-data
{"instance-id":"5d3b43e8-7186-4356-8bb3-45731e23a1c2","local-hostname":"localshark","ssh-keys":["ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIBx78Gk6UIlPOJT7JC888SVHUOKwFJ29yJp2cCOv1TVZ"]}
```

However, that is the wrong JSON key. cloud-init expects `public-keys`:

```
debian@meow:~$ sudo mount /dev/vda /mnt
mount: /mnt: WARNING: source write-protected, mounted read-only.
debian@meow:~$ cat /mnt/meta-data
{"instance-id":"302d68c3-c7d1-4b1c-a154-b96db67fbb6b","local-hostname":"meow","public-keys":["ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIOGmLom09EvwjDo6hSwzBf3kNdf/sMw/lPiswy3u4HR9 iliana@horizon","ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIECO1ghEFVs0WIFJ5mXvMq0GqIaBb4CTbexL5IYLohZ1 iliana@keysmash","ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIDDC/oZFROia+ElMQ0cp3GD2g3/06YoZhA5EsrlKxT2N iliana@redwood"]}
```

The changes to `nexus/src/cidata.rs` are causing the failure. I'm looking for a specific reference on what this file needs to contain from cloud-init's perspective so we can add an instructive comment.

iliana:
> I'm looking for a specific reference on what this file needs to contain from cloud-init's perspective so we can add an instructive comment.

Well, I'm not finding any, so I might just have to write it myself...

Issue #4765, opened 2024-01-05
use separate underlay IP ranges for control plane and propolis instances


davepacheco:
There are a few consumers of IP addresses on each sled's /64 subnet of the underlay network:

- The sled's GZ IP: always a fixed address within the subnet.
- The switch zone IP, if any: always reserved at a fixed address within the subnet
- Control plane zones: these are currently all allocated during RSS, before any propolis instances could be running.
- Propolis instances: these are allocated dynamically as customer Instances are booted and halted.  These are the only ones that get allocated after rack setup.  [Allocation happens by picking the next address that hasn't already been used.](https://github.com/oxidecomputer/omicron/blob/3a69cb0159764a4de932f3e3dd6da1385bf9095f/nexus/db-queries/src/db/datastore/mod.rs#L278-L312)

To be able to provision control plane services after rack setup, we'll need to coordinate allocation of IPs between these last two.  A simple way to do this would be to reserve parts of this subnet for different purposes.  Say:

- reserve the first block for sled-assigned addresses (this would cover the global zone and switch zone addresses, plus give us room for adding other things here)
- reserve the next block for control plane services
- reserve the rest for propolis instances

I _think_ implementing this might be as simple as:

- decide how big we want each of these blocks to be
- for new sleds (including all sleds in new deployments), set "last_used_address" to the minimum address in the propolis block
- for existing deployments, have a db migration step that bumps any "last_used_address" values for existing sleds to at least the minimum address in the propolis block

(This was discussed in 2023-12-14 _ad hoc_ update call.  The recording might have more details.)

davepacheco:
(closing because GitHub didn't notice that #4777 fixes this)

PR #4777, opened 2024-01-08
Addresses for propolis instances at SLED_PREFIX + 0xFFFF

andrewjstone:
Rather than allocating instance IPs starting at `SLED_PREFIX` + `RSS_RESERVED_ADDRESSES` + 1 where the `1` is the sled-agent allocated address of the GZ, we begin allocation from a larger block: `SLED_PREFIX` + `CP_SERVICES_RESERVED_ADDRESSES`. This gives us more room for nexus to allocate control plane services.

Implements #4765


nexus/tests/integration_tests/schema.rs
```diff
@@ -1024,6 +1042,57 @@ fn after_23_0_0(client: &Client) -> BoxFuture<'_, ()> {
...
+            ))
+            .await
+            .expect("Failed to create sleds");
+    })
+}
+
+fn after_24_0_0(client: &Client) -> BoxFuture<'_, ()> {
+    Box::pin(async {
+        // Confirm that the ip_pool_resource objects have been created
+        // by the migration.
```
smklein, line 1073: This looks like a copy-paste comment that isn't quite right here

smklein: This looks like a copy-paste comment that isn't quite right here

andrewjstone: Whoops. Fixed.

common/src/address.rs
```diff
@@ -165,6 +165,9 @@ const GZ_ADDRESS_INDEX: usize = 2;
 /// The maximum number of addresses per sled reserved for RSS.
 pub const RSS_RESERVED_ADDRESSES: u16 = 32;
 
+// The maximum number of addresses per sled reserved for control plane services.
+pub const CP_SERVICES_RESERVED_ADDRESSES: u16 = 0xFFFF;
```
smklein, line 169: What about instances which have been allocated below this threshold that are currently deployed on systems?

Is it an issue that they exist within this reserved range? I don't think *future* allocations will land there, but just trying to ensure we don't make assumptions about prior ones.

andrewjstone: I asked about this in dogfood chat and got this response from @bnaecker:
> It does look like the Propolis server IP address is assigned on instance start, not creation.

Since we currently perform schema upgrades with shutdown instances this shouldn't be a problem. It should probably be tested on real hardware with real instances.

bnaecker: That's based on this, for reference:

https://github.com/oxidecomputer/omicron/blob/9eba46df9980aa76a5043cf0300feddc6e79202c/nexus/src/app/sagas/instance_start.rs#L153-L163

andrewjstone: > That's based on this, for reference:
> 
> https://github.com/oxidecomputer/omicron/blob/9eba46df9980aa76a5043cf0300feddc6e79202c/nexus/src/app/sagas/instance_start.rs#L153-L163

Ha, I just dug that up and was about to put that here :) Thanks!

smklein:
> The plan I've been working from (which is really just an idea we've talked about, plus an implementation in my branch, so still somewhat speculative) is that the store of record for IP addresses in use by Omicron is the current target blueprint. Thus in generating a new blueprint the planner can pick any addresses it wants as long as they don't overlap with each other.

Gotcha, so this conflict resolution could all happen in-memory then, and we wouldn't need to do anything in the DB to help us do the allocation? That makes sense to me.

smklein:
> * `schema/crdb/dbinit.sql` - there's a comment on `last_used_address` suggesting this is for an "Oxide service" - now it isn't, it's for instances exclusively.
> * `nexus/db-model/src/sled.rs` - there's a similar comment on `last_used_address` that needs updating
> * `nexus/db-queries/src/db/datastore/mod.rs` - there's another comment on `next_ipv6_address` that indicates "pulling from last_used_address" is for Oxide services. This isn't true anymore, again, it's service-only.
> * That `next_ipv6_address` method is called by a function called `allocate_sled_ipv6`, which again, is kinda a misnomer, since this should now **only** be used for instances.

@andrewjstone , if we can resolve these nits, this PR LGTM!

common/src/address.rs
smklein: Gotcha, that makes sense. And later, when we do migrations, we'd allocate a new instance IP anyway, so this all checks out to me.

andrewjstone:
> > * `schema/crdb/dbinit.sql` - there's a comment on `last_used_address` suggesting this is for an "Oxide service" - now it isn't, it's for instances exclusively.
> > * `nexus/db-model/src/sled.rs` - there's a similar comment on `last_used_address` that needs updating
> > * `nexus/db-queries/src/db/datastore/mod.rs` - there's another comment on `next_ipv6_address` that indicates "pulling from last_used_address" is for Oxide services. This isn't true anymore, again, it's service-only.
> > * That `next_ipv6_address` method is called by a function called `allocate_sled_ipv6`, which again, is kinda a misnomer, since this should now **only** be used for instances.
> 
> @andrewjstone , if we can resolve these nits, this PR LGTM!

These are great catches! I will fix them. Thank you very much for the thorough review!

smklein (approved):

Issue #4779, opened 2024-01-08
Some tests in omicron-nexus::test_all stall out and hang


sunshowers:
Some example buildomat runs:

On main: https://github.com/oxidecomputer/omicron/runs/20227366854

```
omicron-nexus::test_all integration_tests::projects::test_project_deletion_with_vpc
omicron-nexus::test_all integration_tests::silos::test_saml_idp_metadata_data_valid
```

On a PR: https://github.com/oxidecomputer/omicron/pull/4773/checks?check_run_id=20231856893

```
omicron-nexus::test_all integration_tests::console_api::test_session_me_groups
```

rcgoodfellow:
I have what feels like a similar issue, but with different tests hanging. In this run, the following tests hung.

```
omicron-nexus::test_all integration_tests::rack::test_rack_initialization
omicron-nexus app::update::tests::test_update_deployments
```

Here is a [link to the CI job](https://github.com/oxidecomputer/omicron/pull/4585/checks?check_run_id=20282444863). In this run, the illumos integration tests passed, but the Linux ones did not. This also comes from a [commit](https://github.com/oxidecomputer/omicron/pull/4585/commits/7c93705680b267931a4ff32cc56eaf727f7ca07a) that only changed an unrelated buildomat script and the previous commit on that branch passed CI on the Linux build-and-test job.

smklein:
Just observed this failure here: https://github.com/oxidecomputer/omicron/pull/4783/checks?check_run_id=20286750269

`omicron-nexus::test_all integration_tests::images::test_make_disk_from_image` appears to have been running non-stop for almost two hours. Looks like this happened on Ubuntu-22.04.

Reading what's going on from just the Nexus log is getting kinda difficult - the rendered log is 25,000 lines long, of which ~20,000 of those lines are from `BackgroundTasks`.

sunshowers:
https://github.com/oxidecomputer/omicron/pull/4780 should hopefully land very soon.

smklein:
Couple more cases that just showed up:  https://github.com/oxidecomputer/omicron/pull/4783
```
[...]

PR #4780, opened 2024-01-08
[tests] time out tests in test_all after a duration

sunshowers:
We're seeing some tests in `test_all` hang forever (#4779). Set a reasonable
upper bound on test duration.

This will also cause stdout and stderr for failing tests to be printed. Doing
so on SIGTERM in general is tracked at
https://github.com/nextest-rs/nextest/issues/1208.

Also, bump up the required nextest version to 0.9.64 to make use of the
`binary_id` predicate.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -3,7 +3,7 @@
 #
 # The required version should be bumped up if we need new features, performance
 # improvements or bugfixes that are present in newer versions of nextest.
-nextest-version = { required = "0.9.59", recommended = "0.9.64" }
+nextest-version = { required = "0.9.64", recommended = "0.9.64" }
 
 experimental = ["setup-scripts"]
 
@@ -30,5 +30,11 @@ command = 'cargo run -p crdb-seed --profile test'
 clickhouse-cluster = { max-threads = 1 }
 
 [[profile.default.overrides]]
-filter = 'package(oximeter-db) and test(replicated)'
+filter = 'binary_id(oximeter-db::replicated)'
 test-group = 'clickhouse-cluster'
+
+[[profile.default.overrides]]
+filter = 'binary_id(omicron-nexus::test_all)'
+# As of 2023-01-08, the slowest test in test_all takes 196s on a Ryzen 7950X.
+# 900s is a good upper limit that adds a comfortable buffer.
+slow-timeout = { period = '60s', terminate-after = 15 }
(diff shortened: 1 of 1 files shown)


.config/nextest.toml
```diff
@@ -30,5 +30,11 @@ command = 'cargo run -p crdb-seed --profile test'
...
 [[profile.default.overrides]]
-filter = 'package(oximeter-db) and test(replicated)'
+filter = 'binary_id(oximeter-db::replicated)'
 test-group = 'clickhouse-cluster'
+
+[[profile.default.overrides]]
+filter = 'binary_id(omicron-nexus::test_all)'
+# As of 2023-01-08, the slowest test in test_all takes 196s on a Ryzen 7950X.
+# 900s is a good upper limit that adds a comfortable buffer.
+slow-timeout = { period = '60s', terminate-after = 15 }
```
smklein, line 40: It's not great, but I've seen tests on my laptop that take up to 1000 seconds -- the saga 'test_action_failure_can_unwind' tests are particularly slow, and definitely need optimizing.

I'm not even sure I'd say this timeout is "wrong" - that test needs fixing - but figured I'd mention it regardless

smklein, line 40: It's not great, but I've seen tests on my laptop that take up to 1000 seconds -- the saga 'test_action_failure_can_unwind' tests are particularly slow, and definitely need optimizing.

I'm not even sure I'd say this timeout is "wrong" - that test needs fixing - but figured I'd mention it regardless

smklein (approved):

davepacheco:
FWIW there is a timeout already on the whole test run and we went with a few hours:
https://github.com/oxidecomputer/omicron/blob/cde9b1547ea928c1b6ce0c11bfd09690c2502549/.github/buildomat/build-and-test.sh#L62

Clulow's Lament applies here: it's a tradeoff between false positives (spurious test failures because we didn't wait long enough) vs. taking too long to get feedback when they're hung.  Personally, if a test hangs when I'm running it locally, I'd always rather it hang (and not kill it) because that will preserve a lot of useful runtime state I might want to debug it.  In CI I don't think I care either way.  But false positives here would certainly suck.

.config/nextest.toml
sunshowers: Hmm, that test isn't part of `test_all`.

sunshowers:
> In CI I don't think I care either way. But false positives here would certainly suck.

Fair, I'll restrict it to CI.

PR #4781, opened 2024-01-08
[tufaceous-lib] make DeserializedManifest public and serializable, fix serialization

sunshowers:
This is prep work for #4690. I'd like to add support for making changes to a
DeserializedManifest, then serializing it.

However, it turned out that the bytesize crate does not serialize bytes
correctly (https://github.com/hyunsik/bytesize/issues/40). To address this,
just use a u64 and write our own deserializer, using the alternative parse-size
crate.

Also add a test to ensure that serialization of the fake manifest roundtrips
correctly.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -708,15 +708,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "bytesize"
-version = "1.3.0"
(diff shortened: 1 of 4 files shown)


andrewjstone (approved):
Looks great!

Issue #4747, opened 2024-01-02
Want a way to test data migrations


david-crespo:
While working on #4261, I found it hard to get confidence in my data migrations using existing tooling. One thing I found helpful was using DB Fiddle to write some similar tables and running similar migrations on them to make sure I got the expected results.

https://www.db-fiddle.com/f/pzmM69My6mZWq1NEKm56PG/1

However, this is lacking in ways that formal tooling could improve on:

1. It's Postgres 15, not CRDB, and especially not the version of CRDB we're running
2. The tables are simplified because making them match exactly is annoying to do by hand, even though it would be easy to do automatically

So what I have in mind is essentially a way of:

1. Running migrations up to but not including the one under test
2. Running some SQL to insert rows in the DB
3. Running the migration under test
4. Asserting stuff about the resulting DB state

smklein:
https://github.com/oxidecomputer/omicron/blob/cfcf209c4f4b278bbfe27d19c115d657c63b5d76/nexus/tests/integration_tests/schema.rs#L479-L506 and https://github.com/oxidecomputer/omicron/blob/cfcf209c4f4b278bbfe27d19c115d657c63b5d76/nexus/tests/integration_tests/schema.rs#L808-L850 seem *close* to what you're describing for (1), but it's unfortunately not super cheap to perform "all schema upgrades in order from 1.0.0 to latest". I've been eyeing these tests anyway for optimization purposes, so I wonder if we could share that workload somehow.

(That said, the idea of "update to right before my schema change, populate some test data, do the change, then validate the state" seems super reasonable)

PR #4783, opened 2024-01-09
[schema][test] Add a data migration validation test

smklein:
Adds a schema test with "before" / "after" hooks, and adds an example specifically for the "23.0.0" migration.

My intent is that this can be used for any other schema migrations that would like to execute arbitrary SQL checks
against the new schema too.

Fixes https://github.com/oxidecomputer/omicron/issues/4747

### DIFF
--- a/nexus/tests/integration_tests/schema.rs
+++ b/nexus/tests/integration_tests/schema.rs
@@ -5,6 +5,7 @@
 use camino::Utf8PathBuf;
 use chrono::{DateTime, Utc};
 use dropshot::test_util::LogContext;
+use futures::future::BoxFuture;
 use nexus_db_model::schema::SCHEMA_VERSION as LATEST_SCHEMA_VERSION;
 use nexus_db_queries::db::datastore::{
     all_sql_for_version_migration, EARLIEST_SUPPORTED_VERSION,
@@ -14,7 +15,7 @@ use omicron_common::api::external::SemverVersion;
 use omicron_common::api::internal::shared::SwitchLocation;
 use omicron_common::nexus_config::Config;
 use omicron_common::nexus_config::SchemaConfig;
-use omicron_test_utils::dev::db::CockroachInstance;
+use omicron_test_utils::dev::db::{Client, CockroachInstance};
 use pretty_assertions::{assert_eq, assert_ne};
 use similar_asserts;
 use slog::Logger;
@@ -163,6 +164,18 @@ async fn query_crdb_schema_version(crdb: &CockroachInstance) -> String {
     version
 }
 
+#[derive(PartialEq, Clone, Debug)]
+struct SqlEnum {
+    name: String,
+    variant: String,
+}
+
+impl From<(&str, &str)> for SqlEnum {
+    fn from((name, variant): (&str, &str)) -> Self {
+        Self { name: name.to_string(), variant: variant.to_string() }
+    }
+}
+
 // A newtype wrapper around a string, which allows us to more liberally
 // interpret SQL types.
 //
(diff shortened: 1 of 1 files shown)


nexus/tests/integration_tests/schema.rs
```diff
@@ -218,31 +269,53 @@ impl<'a> tokio_postgres::types::FromSql<'a> for AnySqlType {
         if f32::accepts(ty) {
             return Ok(AnySqlType::Float4(f32::from_sql(ty, raw)?));
         }
+        if serde_json::value::Value::accepts(ty) {
```
smklein, line 272: Urgh. So I feel like I'm basically implementing OID checking here, which I'd rather do via a match statement? But reading https://docs.rs/tokio-postgres/0.7.10/tokio_postgres/types/index.html , I'm struggling to find a way to match on the [Type](https://docs.rs/tokio-postgres/0.7.10/tokio_postgres/types/struct.Type.html) type, rather than have this chain of if-statements.

- [Type::kind](https://docs.rs/tokio-postgres/0.7.10/tokio_postgres/types/enum.Kind.html) returns an enum of possible "flavors of type", but this isn't enough to infer the conversion to one of our "rust types" in the `AnySqlType` variant.
- [Type::oid](https://docs.rs/tokio-postgres/0.7.10/tokio_postgres/types/struct.Type.html#method.oid) can return a `u32` that exactly tells us the resulting oid. But... it's not a const method? So I can't actually access it from all the constant types declared in the "Associated Constants" of https://docs.rs/tokio-postgres/0.7.10/tokio_postgres/types/struct.Type.html#implementations. 

Frustratingly, this info **does** exist in the postgres_types crate: https://docs.rs/postgres-types/0.2.6/src/postgres_types/type_gen.rs.html#207 but it isn't exposed publically.

nexus/tests/integration_tests/schema.rs
```diff
@@ -218,31 +269,53 @@ impl<'a> tokio_postgres::types::FromSql<'a> for AnySqlType {
...
-            ).into())
+
+        use tokio_postgres::types::Kind;
+        match ty.kind() {
+            Kind::Enum(_) => {
+                Ok(AnySqlType::Enum(SqlEnum {
+                    name: ty.name().to_string(),
+                    variant: std::str::from_utf8(raw)?.to_string(),
+                }))
+            },
```
smklein, line 290: This is the laziest possible de-serialization of a user-defined type possible.

"If it's a user-defined type, just throw it into some strings, and do no further checks".

The tests below can perform this validation if they want it -- we don't need compile-time checks on old versions of the DB.

nexus/tests/integration_tests/schema.rs
```diff
@@ -218,31 +269,53 @@ impl<'a> tokio_postgres::types::FromSql<'a> for AnySqlType {
...
     }
 }
 
+// It's a little redunant to include the column name alongside each value,
+// but it results in a prettier diff.
 #[derive(PartialEq, Debug)]
-struct NamedSqlValue {
-    // It's a little redunant to include the column name alongside each value,
-    // but it results in a prettier diff.
+struct ColumnValue {
```
smklein, line 309: Only tangentially related to this change, but IMO makes the usage a little more clear.

nexus/tests/integration_tests/schema.rs
```diff
@@ -849,6 +922,178 @@ async fn dbinit_equals_sum_of_all_up() {
...
+// - Your migration check will run in the same test as all other migration
+// checks, because performing schema migrations isn't that fast. If you
+// perform an operation that could be disruptive to subsequent checks, I
+// recommend cleaning up after yourself (e.g., DELETE relevant rows).
+// - I recommend using schema checks that are NOT strongly-typed. When you
+// add a migration check, it'll happen to match the "latest" static schemas
+// defined by Nexus, but that won't always be the case. As the schema
+// continues to change (maybe a table you're trying to check gets a new column
+// in a later version), your code should continue operating on the OLD version,
+// and as such, should avoid needing any updates.
```
smklein, line 1046: This is worth reading - if any part of the test is still unclear, I'm happy to update this documentation.

nexus/tests/integration_tests/schema.rs
```diff
@@ -849,6 +922,178 @@ async fn dbinit_equals_sum_of_all_up() {
...
+        // created resource.
+        //
+        // The same relationship is true for pool2 / silo2.
+        assert_eq!(
+            ip_pool_resources[0].values,
+            vec![
+                ColumnValue::new("ip_pool_id", POOL1),
+                ColumnValue::new("resource_type", type_silo.clone()),
+                ColumnValue::new("resource_id", SILO1),
+                ColumnValue::new("is_default", true),
```
smklein, line 986: Note that by using weakly-typed access to SQL here, this will continue to work, even if the "new version" of the SQL table changes (Same above during the insertions -- those will be operating on the 23.0.0 version, even if we later upgrade to 30.0.0 and keep changing all these tables).

andrewjstone (approved):
This is great. Thank you Sean. It's clarifying to see how this all fits together and that I would indeed just use sql directly from the client as I started to play with yesterday.

nexus/tests/integration_tests/schema.rs
smklein: Added some extra doc comments!

smklein: Added some extra doc comments!

smklein:
I have high confidence the failures here are due to https://github.com/oxidecomputer/omicron/issues/4779 , going to merge even though we're seeing limited timeouts.

PR #4789, opened 2024-01-10
[nexus-test-utils] set 60s timeouts for each init step

sunshowers:
In #4779 we're tracking what appears to be a ClickHouse initialization failure
during Nexus startup. Set a timeout of 60s for each step in the initialization
process.

These steps should usually not take more than 5 seconds each, so 60s is a
really comfortable buffer.


nexus/test-utils/src/lib.rs
```diff
@@ -604,18 +636,21 @@ impl<'a, N: NexusServer> ControlPlaneTestContextBuilder<'a, N> {
         dns_config_client.dns_config_put(&dns_config).await.expect(
             "Failed to send initial DNS records to internal DNS server",
         );
-        dns_config
+        self.dns_config = Some(dns_config);
```
smklein, line 639: This is just to avoid a `move` between setup steps, yeah?

nexus/test-utils/src/lib.rs
```diff
@@ -290,11 +293,40 @@ impl<'a, N: NexusServer> ControlPlaneTestContextBuilder<'a, N> {
...
             silo_name: None,
             user_name: None,
         }
     }
 
+    pub async fn init_with_steps(
+        &mut self,
+        steps: Vec<(
+            &str,
+            Box<dyn for<'b> FnOnce(&'b mut Self) -> BoxFuture<'b, ()>>,
```
sunshowers, line 306: Yay for no async closures :(

smklein: I know, also been suffering from this recently. Sucks to need to lean on box so heavily to get this all done.

smklein (approved):
Neat, strongly in favor of the better visibility during test setup!

nexus/test-utils/src/lib.rs
sunshowers: With this pattern, we no longer have a way to return a value that the next step can use. This is the easiest way to work around that (and follows the pattern the rest of the builder uses anyway).

I ran into something similar with the update-engine and came up with the design to use step handles that can be passed around, but only resolved once steps start executing. For example, https://github.com/oxidecomputer/omicron/blob/45b665184ad6254f5b2bb969abefbf476e0803c8/update-engine/examples/update-engine-basic/main.rs#L308-L310. But that's trying to do a lot more than we need here.

PR #4790, opened 2024-01-10
Provide mechanism for RPWs to execute sagas

jmpesp:
This commit adds a `SagaRequest` channel that can be shared with reliable persistent workflow (RPW) background tasks so that a saga can be executed as part of those task's activation.

This functionality will be used as part of downstairs replacement: when a physical disk is marked as no longer in use, a series of RPWs and sagas will ensure that any Crucible resources that were allocated to that physical disk are moved, and make any necessary volume changes.


nexus/src/app/sagas/mod.rs
```diff
@@ -400,3 +401,18 @@ where
...
+/// activation by sending a SagaRequest through a supplied channel to Nexus.
+pub enum SagaRequest {
+    #[cfg(test)]
+    TestOnly,
+}
+
+impl SagaRequest {
+    pub fn channel() -> (mpsc::Sender<SagaRequest>, mpsc::Receiver<SagaRequest>)
+    {
+        // XXX this is a magic number!
```
smklein, line 415: Could we have a simple explanation for how this was chosen?

Shoving this into a named const with a comment like "This limits _____, and was chosen arbitrarily" would be fine - at least that would identify who gets blocked on it!

nexus/src/app/background/region_replacement.rs
```diff
@@ -0,0 +1,61 @@
...
+
+impl RegionReplacementDetector {
+    pub fn new(
+        datastore: Arc<DataStore>,
+        saga_request: Sender<SagaRequest>,
+    ) -> Self {
+        RegionReplacementDetector {
+            _datastore: datastore,
+            _saga_request: saga_request,
+        }
```
smklein, line 33: It's not clear to me why we're including the stub task for this `RegionReplacementDetector` - you mentioned in chat that this task is necessary, but we aren't even doing anything with this channel, and we **can't** do anything with it, because it only contains a test-only variant?

jmpesp: To be clear, it's necessary to put the `Sender<SagaRequest>` somewhere so that it's not dropped, as that would cause Nexus to panic (with the changes in this PR).

nexus/src/app/sagas/mod.rs
```diff
@@ -400,3 +401,18 @@ where
     )
     .await
 }
+
+/// Reliable persistent workflows can request that sagas be run as part of their
+/// activation by sending a SagaRequest through a supplied channel to Nexus.
+pub enum SagaRequest {
+    #[cfg(test)]
+    TestOnly,
+}
```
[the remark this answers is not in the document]
jmpesp: The intention is for this enum to have a variant for each saga that could be requested, along with the params:

```rust
    DiskCreate { params: disk_create::Params },
```

and for Nexus to simply be unwrapping this:

```rust
match saga_request {
    ....
    DiskCreate { params: disk_create::Params } => {
        self.execute_saga::<sagas::disk_create::SagaDiskCreate>(params).await?;
    }
    ....
}
```

Unfortunately I don't have a task that uses this along with this PR: `RegionReplacementDetector` is currently a stub until physical disks have a state that this background task can query for, I simply wanted to get this work in separately to make subsequent PRs smaller :)

I believe it could help this PR make sense to wait for physical disks to have this state, and to have an implementation to go along with this.

nexus/src/app/background/region_replacement.rs
smklein: I understand that those are the mechanics of monitoring a channel, but from my perspective, we're plumbing a channel that we can **never** use in production because it only contains a test-only variant. I'm just not sure what I'm reviewing -- what is this channel **supposed** to be passing around? Why is it necessary?

nexus/src/app/sagas/mod.rs
jmpesp: I've changed this in 5ee010555, this is a good point.

nexus/src/app/sagas/mod.rs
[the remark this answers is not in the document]
smklein: I see. So the TL;DR is that this PR provides a mechanism for background tasks to access `Arc<Nexus>`, more-or-less, without them storing a bunch of https://doc.rust-lang.org/std/sync/struct.Weak.html ? And we use channels with requests to Nexus to ask it to execute each saga?

That makes sense!

nexus/src/app/mod.rs
```diff
@@ -477,6 +482,27 @@ impl Nexus {
...
+            let nexus = nexus.clone();
+            tokio::spawn(async move {
+                loop {
+                    match saga_request_recv.recv().await {
+                        None => {
+                            // If this channel is closed, then RPWs will not be
+                            // able to request that sagas be run. Panic Nexus so
+                            // that it can be restarted.
+                            panic!("saga request channel closed!");
+                        }
```
smklein, line 496: Could this be a signal that we should shut down? Might be better to "log a warning, then exit" rather than panicking.

smklein (approved):

nexus/src/app/mod.rs
jmpesp: This is a good point, changed to `error!` in b2be97f6e

PR #4802, opened 2024-01-11
Create /etc/inet/hosts as part of zone networking setup

citrus-it:


karencfv (approved):
Thanks for this!

citrus-it:
I have tested this by deploying to a bench gimlet and confirming that the hosts files within the self assembly zones are correct, and also that the previous messages from auditing about being unable to resolve the host's primary hostname no longer occur.

Issue #4762, opened 2024-01-05
Internal IP pool should not be linkable to other silos


david-crespo:
Followup to #4261. See `no_cud` integration test. Currently linking of internal pool to silos other than the internal silo is allowed and should be prevented.

PR #4808, opened 2024-01-12
Fix duplicate entries in IP pools list

david-crespo:
Bugfix for #4261. Using `distinct` to eliminate dupes from the left outer join works, but it's better to just use the well-known name of the service IP pool to exclude it from whatever operations it needs to be excluded from.

Potential followup related to #4762: if the ID was well-known too, we could do the `is_internal` check without having to hit the DB.

### DIFF
--- a/nexus/tests/integration_tests/endpoints.rs
+++ b/nexus/tests/integration_tests/endpoints.rs
@@ -1048,7 +1048,7 @@ pub static VERIFY_ENDPOINTS: Lazy<Vec<VerifyEndpoint>> = Lazy::new(|| {
         // IP Pool endpoint (Oxide services)
         VerifyEndpoint {
             url: &DEMO_IP_POOL_SERVICE_URL,
-            visibility: Visibility::Public,
+            visibility: Visibility::Protected,
             unprivileged_access: UnprivilegedAccess::None,
             allowed_methods: vec![
                 AllowedMethod::Get
@@ -1058,7 +1058,7 @@ pub static VERIFY_ENDPOINTS: Lazy<Vec<VerifyEndpoint>> = Lazy::new(|| {
         // IP Pool ranges endpoint (Oxide services)
         VerifyEndpoint {
             url: &DEMO_IP_POOL_SERVICE_RANGES_URL,
-            visibility: Visibility::Public,
+            visibility: Visibility::Protected,
             unprivileged_access: UnprivilegedAccess::None,
             allowed_methods: vec![
                 AllowedMethod::Get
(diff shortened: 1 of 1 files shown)


nexus/db-queries/src/db/datastore/ip_pool.rs
```diff
@@ -81,6 +81,7 @@ impl DataStore {
         )
         .filter(ip_pool::time_deleted.is_null())
         .select(IpPool::as_select())
+        .distinct()
```
david-crespo, line 84: This is the easy fix, but I'm wondering if instead of using linkage to the internal silo to identify the internal pool, we could leave out the join altogether if we just had a well-known name for the internal pool and filtered that out.

zephraph: I believe the `SERVICE_IP_POOL_NAME` const should be usable for this.

https://github.com/oxidecomputer/omicron/blob/009c4ae24d7394550f779dc02d83097293d87873/nexus/db-queries/src/db/datastore/mod.rs#L117

https://github.com/oxidecomputer/omicron/blob/009c4ae24d7394550f779dc02d83097293d87873/nexus/db-queries/src/db/datastore/rack.rs#L756-L760

david-crespo: Will attempt.

nexus/db-queries/src/db/datastore/ip_pool.rs
```diff
@@ -225,48 +218,15 @@ impl DataStore {
...
-                        LookupType::ByCompositeId(
-                            "Service IP Pool".to_string(),
-                        ),
-                    ),
-                    ip_pool,
-                )
-            })?;
-        Ok((authz_pool, pool))
+        let name = SERVICE_IP_POOL_NAME.parse().unwrap();
+        LookupPath::new(&opctx, self).ip_pool_name(&Name(name)).fetch().await
```
david-crespo, line 229: 🤦

nexus/db-queries/src/db/datastore/ip_pool.rs
```diff
@@ -71,14 +71,7 @@ impl DataStore {
...
         }
-        .left_outer_join(ip_pool_resource::table)
-        .filter(
-            ip_pool_resource::resource_id
-                .ne(*INTERNAL_SILO_ID)
-                // resource_id is not nullable -- null here means the
-                // pool has no entry in the join table
-                .or(ip_pool_resource::resource_id.is_null()),
-        )
+        .filter(ip_pool::name.ne(SERVICE_IP_POOL_NAME))
```
zephraph, line 74: Yeah, so much better 🙌

nexus/db-queries/src/db/datastore/ip_pool.rs
zephraph: I count this as a win, haha.

PR #4810, opened 2024-01-12
Add oxlog tool and library

andrewjstone:


dev-tools/oxlog/src/lib.rs
```diff
@@ -0,0 +1,308 @@
...
+pub struct SvcLogs {
+    pub current: Option<LogFile>,
+    pub archived: Vec<LogFile>,
+
+    // Logs in non-standard places and logs that aren't necessarily oxide logs
+    pub extra: Vec<LogFile>,
+}
+
+// These probably don't warrant newtypes. They are just to make the nested
+// maps in `Logs` a bit easier to read.
```
andrewjstone, line 103: TODO: Update comment. No more `Logs`

dev-tools/oxlog/src/lib.rs
```diff
@@ -0,0 +1,308 @@
...
+            load_svc_logs(dir, &mut output);
+        }
+        for (svc_name, dir) in paths.extra.clone() {
+            load_extra_logs(dir, svc_name, &mut output);
+        }
+        output
+    }
+}
+
+fn load_svc_logs(dir: Utf8PathBuf, logs: &mut BTreeMap<ServiceName, SvcLogs>) {
```
bnaecker, line 220: There is some similar logic [here](https://github.com/oxidecomputer/omicron/blob/86c64a2fe8167de354a827425b2d48cfe0f33f33/illumos-utils/src/running_zone.rs#L1419) for finding logs during zone-bundling, it might be a good idea to share it somehow. Calling into this library in the `running_zone` mod seems reasonable.

dev-tools/oxlog/src/bin/oxlog.rs
```diff
@@ -0,0 +1,116 @@
...
+        #[arg(short, long)]
+        metadata: bool,
+
+        /// Print only the current log file
+        #[arg(short, long)]
+        current: bool,
+
+        /// Print only the archive log files
+        #[arg(short, long)]
+        archived: bool,
```
bnaecker, line 40: I think there is a third class, rotated log files that have not yet been archived. These probably don't live very long, since the archiver runs frequently. But we may want to consider including them and a CLI switch to filter them.

andrewjstone: These are actually included with the archived files right now. I"m happy to separate them out though if you think that makes sense.

dev-tools/oxlog/src/lib.rs
andrewjstone: I can certainly add a helper function to do this. However, right now oxlog only looks at logs with the `oxide-` prefix. There are a lot more with the `system-`prefix. Should I make that a separate type of retrieval?

bnaecker: I'm not sure. The `system-illumos-` prefix was included because there are some of our services that still use the original naming of the SMF services, like `svc:/system/illumos/propolis-server`. (I think some of the networking daemons are still named this way too.)

Searching for that prefix becomes more reasonable when we're only looking for logs that come from a specific, named service, which is the case of the zone bundling process. That might be too permissive for `oxlog`, I'm really not sure. If that's true, then maybe we can commonize the portion of the checks that make sense (say, for SMF services named like `svc:/oxide/blah`). It seems like you might also be trying to capture _all_ log files for _any_ service, not just the "primary" Oxide services in zone. In that case, the commonization might be less helpful.

andrewjstone: > Searching for that prefix becomes more reasonable when we're only looking for logs that come from a specific, named service, which is the case of the zone bundling process. That might be too permissive for oxlog, I'm really not sure.

I think this is likely to be a common use case, but we also capture all logs in a zone. This means that we'll suck up all 'system-' prefixed logs unless we hardcode a list of services that can be named like this. I've so far avoiding hardcoding the names of any services and deducing service names by the names of the log-files. It turns out that strategy falls apart because of this dichotomy in naming. It seems to me that we would be better off changing the prefix for things like propolis to be `oxide-`, but in the meantime I can suck up only files that match a hardcoded set of services so we don't miss them. I specifically do not want to include all system service files as there are so many and mostly we won't want to dig through them in the common case. Again, though, we likely will want to be able to retrieve those files in some debugging scenarios though. Getting them through sled-agent integration via oxlog is preferable than having to ssh in. I will add some support for this as well.

>  It seems like you might also be trying to capture all log files for any service, not just the "primary" Oxide services in zone. In that case, the commonization might be less helpful.

I'm trying to gather any oxide specific service log in a zone, ideally without knowing the name of the service ahead of time. That may be too ambitious.

bnaecker: > I'm trying to gather any oxide specific service log in a zone, ideally without knowing the name of the service ahead of time. That may be too ambitious.

So the way we did that in zone-bundling is to list all SMF FMRIs matching the patterns `svc:/oxide/*` or `svc:/system/illumos/*`, and then listing the logfiles associated with each of those (with `svcs -L`). 

Based on your answer, it does really sound like you want to use the same logic as in the zone bundling, or at least, use that logic possibly in addition to some other. Note that the prefix used in the zone-bundling is `system-illumos-*` _not_ `system-*`. The latter is definitely very permissive, and will get you a bunch of stuff unrelated to Oxide services. If the goal is Oxide-specific log files, then I think the logic can be shared with zone-bundling :)

andrewjstone: >  Note that the prefix used in the zone-bundling is system-illumos-* not system-*. The latter is definitely very permissive, and will get you a bunch of stuff unrelated to Oxide services. If the goal is Oxide-specific log files, then I think the logic can be shared with zone-bundling :)

Aha! Alright, so far the only dependency for oxlog is using the filesystem. But I can definitely parse using `system-illumos-*` and include those files. Then we can share the logic.

andrewjstone: I added this support in [c38d7e4](https://github.com/oxidecomputer/omicron/pull/4810/commits/c38d7e4c1f036546bf302d44e5c036d7fc278baf)

dev-tools/oxlog/src/lib.rs
```diff
@@ -0,0 +1,442 @@
...
+
+#[derive(Debug)]
+pub struct Pools {
+    pub internal: Vec<Uuid>,
+    pub external: Vec<Uuid>,
+}
+
+impl Pools {
+    pub fn read() -> anyhow::Result<Pools> {
+        let internal = read_dir("/pool/int/")
```
andrewjstone, line 47: We don't actually read anything from the internal_pools, but I don't think it hurts to keep this type around for now. We may end up reading if we expand this tool to grab other files.

This was one of the first bits of code I wrote for this PR and I left it through all the iterations. Happy to change/remove if desired though.

dev-tools/oxlog/src/lib.rs
```diff
@@ -0,0 +1,442 @@
...
+            continue;
+        };
+        let filename = entry.file_name();
+        let Some(filename) = filename.to_str() else {
+            continue;
+        };
+        let mut path = dir.clone();
+        path.push(filename);
+        let mut logfile = LogFile::new(path);
+        if let Ok(metadata) = entry.metadata() {
```
andrewjstone, line 367: Managing metadata can be deduped here and in `load_svc_logs` by adding a method to `LogFile`. We *may* want to add a filter for metadata as well so that we don't bother collecting it if we aren't asked for it.

dev-tools/oxlog/src/lib.rs
```diff
@@ -0,0 +1,442 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    pub use super::is_oxide_smf_log_file;
+    pub use super::oxide_smf_service_name_from_log_file_name;
+
+    #[test]
+    fn test_is_oxide_smf_log_file() {
```
andrewjstone, line 388: I copied this from the zone_bundle code along with the related method. I added an extra test at the end to ensure we don't also collect `system-*` along with `system-illumos-*` files.

dev-tools/oxlog/src/lib.rs
```diff
@@ -0,0 +1,442 @@
...
+}
+
+const OX_SMF_PREFIXES: [&str; 2] = ["oxide-", "system-illumos-"];
+
+/// Return true if the provided file name appears to be a valid log file for an
+/// Oxide-managed SMF service.
+///
+/// Note that this operates on the _file name_. Any leading path components will
+/// cause this check to return `false`.
+pub fn is_oxide_smf_log_file(filename: impl AsRef<str>) -> bool {
```
andrewjstone, line 260: @bnaecker I copied this over from `illumos_utils`. If you want I can go ahead and update the `zone-bundle` code to use the method here instead, and remove it from `illumos_utils`.  I can do that in this PR or a follow up. 

It also looks like (according to grep at least) that we aren't using `illumos_utils::is_oxide_smf_service`. Should I just remove it altogether? I can leave it in `illumos-utils` but it will seem like a hanging chad after removing `is_oxide_smf_log_file`.

dev-tools/oxlog/src/lib.rs
```diff
@@ -0,0 +1,442 @@
...
+    pub fn zone_logs(
+        &self,
+        zone: &str,
+        filter: Filter,
+    ) -> BTreeMap<ServiceName, SvcLogs> {
+        let mut output = BTreeMap::new();
+        let Some(paths) = self.zones.get(zone) else {
+            return BTreeMap::new();
+        };
+        load_svc_logs(paths.primary.clone(), &mut output);
```
andrewjstone, line 237: I specifically did not add a filter check before this line because:
 1. This is fast
 2. Logs that have been rotated but not yet moved into the `debug` directory will be added to archive paths, so I didn't want to only filter on `current`. 

However, I now realize I could filter such that we don't load the logs if `filter.current` and `filter.archived` are both false.

andrewjstone:
I also need to clean up the packaging into the global zone as @smklein pointed out: https://github.com/oxidecomputer/omicron/pull/4840#issuecomment-1899010237

dev-tools/oxlog/src/lib.rs
andrewjstone: When I was writing this code I realized that filtering on metadata would prevent us from checking the size or we'd end up just not reading the modified date. I don't think this is important right now, so skipped it.

dev-tools/oxlog/src/lib.rs
andrewjstone: Fixed in [f2b2f8a](https://github.com/oxidecomputer/omicron/pull/4810/commits/f2b2f8a245aebd32243937a72ae2ef0603f15f15)

dev-tools/oxlog/src/lib.rs
andrewjstone: `illumos-utils` now uses the method which I copied into oxlog from oxlog as of [caf3cd5](https://github.com/oxidecomputer/omicron/pull/4810/commits/caf3cd5f74be8313f00d617d9cdb4d08d11a9c41)

dev-tools/oxlog/src/lib.rs
andrewjstone: Nevermind. I misread the comment above. `illumos_utils::running_zone` now calls into the `oxlog` method.

andrewjstone:
> I also need to clean up the packaging into the global zone as @smklein pointed out: [#4840 (comment)](https://github.com/oxidecomputer/omicron/pull/4840#issuecomment-1899010237)

Implemented in [09ad1b5](https://github.com/oxidecomputer/omicron/pull/4810/commits/09ad1b5580a5865c0d4dce5b672fb1994155b108)

dev-tools/oxlog/src/lib.rs
```diff
@@ -0,0 +1,448 @@
...
+
+//! A tool to show oxide related log file paths
+//!
+//! All data is based off of reading the filesystem
+
+use anyhow::Context;
+use camino::Utf8PathBuf;
+use chrono::{DateTime, Utc};
+use std::collections::BTreeMap;
+use std::fs::{read_dir, DirEntry};
```
smklein, line 13: I believe camino supports a version of read_dir that does path parsing, and exposes a UTF-8 based DirEntry:
https://docs.rs/camino/1.1.6/camino/struct.Utf8Path.html#method.read_dir

That might be useful here?

smklein (approved):

dev-tools/oxlog/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+camino.workspace = true
+chrono.workspace = true
+clap.workspace = true
+uuid.workspace = true
+omicron-workspace-hack.workspace = true
+
+# Disable doc builds by default for our binaries to work around issue
+# rust-lang/cargo#8373.  These docs would not be very useful anyway.
+[[bin]]
+name = "oxlog"
```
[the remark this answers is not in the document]
andrewjstone: I definitely cargo-culted this from elsewhere. I'll get rid of the reference.

dev-tools/oxlog/src/lib.rs
```diff
@@ -0,0 +1,448 @@
...
+///
+/// Each field in the filter is additive.
+///
+/// The filter was added to the library and not just the CLI because in some
+/// cases searching for archived logs is pretty expensive.
+#[derive(Clone, Copy, Debug)]
+pub struct Filter {
+    pub current: bool,
+    pub archived: bool,
+    pub extra: bool,
```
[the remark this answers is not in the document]
andrewjstone: Done in [835d0cd](https://github.com/oxidecomputer/omicron/pull/4810/commits/835d0cd021ec0d63b779fb16542b7a250a7a8130)

dev-tools/oxlog/src/lib.rs
andrewjstone: Great idea! This saves a 3 line check when getting the filename! Done in [5e48e48](https://github.com/oxidecomputer/omicron/pull/4810/commits/5e48e4814f0c0d5a6beebc81f123e0bd7d86faf8)

PR #4816, opened 2024-01-15
[package] Use topological sorting from omicron-zone-package

smklein:
This functionality is now provided in `omicron-zone-package`, as of https://github.com/oxidecomputer/omicron-package/pull/57


luqmana (approved):
Thanks for pulling that out. Great to see it being useful elsewhere too!

Issue #4819, opened 2024-01-16
[sled agent] Co-locate zone filesystems with the datasets managed by zones


smklein:
Currently, when the sled agent provisions zones onto datasets, it picks a "random U.2" to distribute these zone filesystems onto arbitrary U.2s:

https://github.com/oxidecomputer/omicron/blob/146aa5fdec6c527382af26c9369227d7c76ed8f8/sled-agent/src/services.rs#L2929-L2939

Unlike the more durable "`/data` datasets", these zone filesystems are destroyed between reboots, so this choice was fairly arbitrary.

Unfortunately, this choice means that "For zones which already datasets", their fault domain is now spread across two different U.2s. We could make this allocation policy slightly smarter, to use a U.2 for the existing dataset if one already exists.

PR #4820, opened 2024-01-16
Co-locate 'data' datasets with 'zone' datasets

smklein:
Fixes https://github.com/oxidecomputer/omicron/issues/4819

### DIFF
--- a/sled-agent/src/services.rs
+++ b/sled-agent/src/services.rs
@@ -2887,12 +2887,10 @@ impl ServiceManager {
         }
 
         // Create zones that should be running
-        let all_u2_roots = self
-            .inner
-            .storage
-            .get_latest_resources()
-            .await
-            .all_u2_mountpoints(ZONE_DATASET);
+        let storage = self.inner.storage.get_latest_resources().await;
+        let all_u2_pools = storage.all_u2_zpools();
+        let all_u2_roots = storage.all_u2_mountpoints(ZONE_DATASET);
+
         let mut new_zones = Vec::new();
         for zone in zones_to_be_added {
             // Check if we think the zone should already be running
@@ -2932,11 +2930,32 @@ impl ServiceManager {
             //
             // This is (currently) intentional, as the zone filesystem should
             // be destroyed between reboots.
-            let mut rng = rand::thread_rng();
-            let root = all_u2_roots
-                .choose(&mut rng)
-                .ok_or_else(|| Error::U2NotFound)?
-                .clone();
+            let root = if let Some(dataset) = zone.dataset_name() {
+                // If the zone happens to already manage a dataset, then
+                // we co-locate the zone dataset on the same zpool.
+                //
+                // This slightly reduces the underlying fault domain for the
+                // service.
+                let data_pool = dataset.pool();
+                if !all_u2_pools.contains(&data_pool) {
+                    warn!(
+                        log,
+                        "zone dataset requested on a zpool which doesn't exist";
+                        "zone" => &name,
+                        "zpool" => %data_pool
+                    );
+                    return Err(Error::MissingDevice {
+                        device: format!("zpool: {data_pool}"),
+                    });
+                }
+                data_pool.dataset_mountpoint(ZONE_DATASET)
+            } else {
+                let mut rng = rand::thread_rng();
+                all_u2_roots
+                    .choose(&mut rng)
+                    .ok_or_else(|| Error::U2NotFound)?
+                    .clone()
+            };
 
             new_zones.push(OmicronZoneConfigLocal { zone: zone.clone(), root });
         }
--- a/sled-agent/src/services.rs
+++ b/sled-agent/src/services.rs
@@ -2889,7 +2889,6 @@ impl ServiceManager {
         // Create zones that should be running
         let storage = self.inner.storage.get_latest_resources().await;
         let all_u2_pools = storage.all_u2_zpools();
-        let all_u2_roots = storage.all_u2_mountpoints(ZONE_DATASET);
 
         let mut new_zones = Vec::new();
         for zone in zones_to_be_added {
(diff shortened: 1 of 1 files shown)


sled-agent/src/services.rs
```diff
@@ -2932,11 +2929,33 @@ impl ServiceManager {
...
+                // service.
+                let data_pool = dataset.pool();
+                if !all_u2_pools.contains(&data_pool) {
+                    warn!(
+                        log,
+                        "zone dataset requested on a zpool which doesn't exist";
+                        "zone" => &name,
+                        "zpool" => %data_pool
+                    );
+                    return Err(Error::MissingDevice {
```
andrewjstone, line 2946: Is there ever a case where we'd want to crate the zone filesystem even if the zpool dataset wasn't found? I can't think of one and believe this makes sense, just putting the question out there for completeness.

sled-agent/src/services.rs
```diff
@@ -2932,11 +2929,33 @@ impl ServiceManager {
             //
             // This is (currently) intentional, as the zone filesystem should
             // be destroyed between reboots.
-            let mut rng = rand::thread_rng();
```
andrewjstone, line 2935: Should probably remove the above comment and/or move part of it to the else clause below.

andrewjstone (approved):
LGTM, with one nit for the comment.

sled-agent/src/services.rs
smklein: To be super pedantic:

> even if the zpool dataset wasn't found

This logic is not looking for a dataset, it's looking for a *zpool*. If the zpool doesn't exist, we should not expect to be able to provision any new datasets on it, for any reason.

The sled agent auto-formats new disks to have the necessary zpools for further allocations **before** those disks are made available to the control plane at large, so it should be necessary for us to process these pools before we could make this kinda allocation.

The only situation where I can imagine this causing a "new" error would be:

1. The sled agent boots, formats zpools, informs Nexus about them
2. The sled reboots, we haven't locally parsed which disks are up yet
3. Nexus tries sending a request to provision a new service to that sled
4. This logic runs, and we'd report "the U.2 you're asking for doesn't appear to exist"
5. If Nexus re-sent the request, it should work later, once the sled agent notices that the zpool exists

I *think* the value of the "early error in cases where the disk was truly removed" is more useful, but I could be persuaded to omit this error case, if you think it's a regression?

smklein:
Manually testing on a local helios box:

# Before this PR

```bash
 $ zfs list -o name | grep cockroach                                            
oxp_14b4dc87-ab46-49fb-a4b4-d361ae214c03/cockroachdb
oxp_14b4dc87-ab46-49fb-a4b4-d361ae214c03/crypt/zone/oxz_cockroachdb_44c03fb8-0370-4ba0-9eb0-6dfb24da1afd
oxp_24b4dc87-ab46-49fb-a4b4-d361ae214c03/cockroachdb  # Not co-located with zone dataset
oxp_31bd71cd-4736-4a12-a387-9b74b050396f/cockroachdb
oxp_31bd71cd-4736-4a12-a387-9b74b050396f/crypt/zone/oxz_cockroachdb_f0c56182-9e76-4c81-afb7-47dfed16d051
oxp_616b26df-e62a-4c68-b506-f4a923d8aaf7/cockroachdb
oxp_616b26df-e62a-4c68-b506-f4a923d8aaf7/crypt/zone/oxz_cockroachdb_52e7363b-4409-4877-aa69-5bdece01dace
oxp_cd70d7f6-2354-4bf2-8012-55bf9eaf7930/cockroachdb  # Not co-located with zone dataset
oxp_e4b4dc87-ab46-49fb-a4b4-d361ae214c03/crypt/zone/oxz_cockroachdb_51b8666d-527f-4b4e-8269-4196eb48f27a
oxp_f4b4dc87-ab46-49fb-a4b4-d361ae214c03/crypt/zone/oxz_cockroachdb_74a8f98d-746e-43da-aa91-bc4740f1b9b1
```

# After this PR (all datasets co-located)

```bash
 $ zfs list -o name | grep cockroach
oxp_14b4dc87-ab46-49fb-a4b4-d361ae214c03/cockroachdb
oxp_14b4dc87-ab46-49fb-a4b4-d361ae214c03/crypt/zone/oxz_cockroachdb_5fa30186-8f64-47ef-9320-9d5c9fad6d11

oxp_24b4dc87-ab46-49fb-a4b4-d361ae214c03/cockroachdb
oxp_24b4dc87-ab46-49fb-a4b4-d361ae214c03/crypt/zone/oxz_cockroachdb_eb6efb03-0fec-4d1b-8260-7966d65b7a63

oxp_31bd71cd-4736-4a12-a387-9b74b050396f/cockroachdb
oxp_31bd71cd-4736-4a12-a387-9b74b050396f/crypt/zone/oxz_cockroachdb_4907da9c-6b8f-4203-a6b1-e331a7596a3a

oxp_616b26df-e62a-4c68-b506-f4a923d8aaf7/cockroachdb
oxp_616b26df-e62a-4c68-b506-f4a923d8aaf7/crypt/zone/oxz_cockroachdb_56cbfdb4-cd27-4eb0-84e9-f031a1817229

oxp_cd70d7f6-2354-4bf2-8012-55bf9eaf7930/cockroachdb
oxp_cd70d7f6-2354-4bf2-8012-55bf9eaf7930/crypt/zone/oxz_cockroachdb_fa5379b6-1519-4c7c-a191-930ad0c535cd
```

sled-agent/src/services.rs
andrewjstone: Thanks for the details and pedantry. Always appreciated. :) I think the error is good to keep and in that case nexus should retry anyway I believe, whether with this sled or another. Eventually Inventory will handle the availability of disks and the planner will correct for this scenario I believe

sled-agent/src/services.rs
smklein: Done in 61d94dce573b33d019dfe41023e4685285e410e1

PR #4821, opened 2024-01-16
don't enable chrony in the gz, it runs in the ntp zone

rcgoodfellow:
When chrony runs in both the gz and in the ntp zone, they fight each other and time becomes unstable, causing issues for subsystems like CRDB that depend on stable time. This PR removes enablement of chrony in the gz in the prerequisites install script.


andrewjstone (approved):
Thanks @rcgoodfellow

PR #4822, opened 2024-01-16
Rpws for all networking

### LINKED ISSUE #4650: Service NAT entries missing after dendrite restart
I recently restarted a `dpd` daemon. All the instance NAT entries were repopulated. However, service NAT entries with a VNI of 100 were missing.

### LINKED ISSUE #4715: Network RPWs for All
This is a tracking issue for the remaining work that needs to be done to completely migrate our network configuration to RPWs.

Currently we are only handing post-Nexus NAT configuration with RPWs. We now need to ensure that pre-Nexus NAT information is managed via RPW. We also need to move all other network configuration items, pre- and post-Nexus, to RPWs.

The "novel" part of this project is that we will also need to determine how to perform synchronization of post-Nexus state to bootstore, meaning that we have some bi-directional synchronization occurring:
1. Cold boot process performs initial configuration
2. Initial configuration is passed to Nexus for recording in RPW tables
3. If any relevant changes happen after Nexus is up, that must be updated in the RPW tables _and_ the bootstore

Currently, it seems that all service zone NAT uses VNI 100, so if we can be sure that is a long term design decision that will not change, we can use that (or a similar mechanism) to differentiate which NAT entries need to be recorded in the bootstore vs. which ones should wait for post-Nexus mechanism to kick in.

At the moment, it is my understanding that all link and BGP configuration will need to be persisted in both mechanisms.

TODO
---
- [x] Pre-Nexus NAT should also be persisted via RPW
- [ ] Link / Port Configurations should be persisted via RPW
- [ ] BGP Configurations should be persisted via RPW

Related Issues
---
Closes #4650
Closes #3631

internet-diglett:
Overview
---
This PR ensures the rest of our `dpd` configuration is covered by a RPW to help recover state in the event of `dendrite` crashing, the switch zone restarting / being replaced, or the sled restarting. This is accomplished via a background task in Nexus that periodically ensures `dpd` is up to date with the tables in Nexus. The tradeoffs of this design is that we don't track versioning and reconcile the entire state every time, but since the actual number of ports will never be that high (relative to something like NAT entries) the tradeoff of less efficiency for much greater simplicity seems to make sense today, and it requires much less rework in Nexus and Dendrite should we choose to replace this strategy down the road.

Tasks
---
- [x] Ensure that Service Zones configured during rss, cold boot, and nexus have their NAT entries added to the NAT RPW table (extracted into  #4857)
- [x] Create background task that periodically reconciles switch port configuration for dendrite instances
- [x] Move switch zone uplink SMF property updates to RPW
- [x] Move routing updates (via mg) to RPW
    - [x] Static Routing
    - [x] BGP
- [x] Move bootstore updates to RPW
- [x] Move loopback address management to RPW
- [x] Move Nexus-side switch zone service on-demand lookups as outlined in #5092

Verifications Performed
---
- [x] Basic instance deployment
- [x] Loopback Address Creation
- [ ] BGP configuration (a4x2)
- [ ] BGP configuration modification (a4x2)
- [x] Static routing
- [x] Static routing configuration modification

Related
---
Closes #4715
Closes #4650

Depends on https://github.com/oxidecomputer/dendrite/pull/838


sled-agent/src/params.rs
```diff
@@ -526,6 +526,8 @@ pub enum OmicronZoneType {
         /// The service vNIC providing outbound connectivity using OPTE.
         nic: NetworkInterface,
         /// The SNAT configuration for outbound connections.
+        // TODO: internet-diglett
+        // needs NAT entry persisted in RPW
```
internet-diglett, line 530:
```suggestion
```

sled-agent/src/params.rs
```diff
@@ -585,6 +587,8 @@ pub enum OmicronZoneType {
         /// The address at which the internal nexus server is reachable.
         internal_address: SocketAddrV6,
         /// The address at which the external nexus server is reachable.
+        // TODO: internet-diglett
+        // needs NAT entry persisted in RPW
```
internet-diglett, line 591:
```suggestion
```

nexus/src/app/background/inventory_collection.rs
```diff
@@ -153,64 +151,14 @@ async fn inventory_activate(
     Ok(collection)
 }
 
-/// Determine which sleds to inventory based on what's in the database
```
internet-diglett, line 156: I'm probably going to undo the commit containing this change (moving the sled querying logic to common). Initially I thought that I should potentially query the sled agents directly, but it seems that using the records stored by this background job is sufficient even though it adds some latency.

rcgoodfellow:
A few initial high-level observations.

- I think all the port settings saga code can be thrown out now?
    - [switch_port_settings_apply.rs](https://github.com/oxidecomputer/omicron/blob/rpws-for-all-networking/nexus/src/app/sagas/switch_port_settings_apply.rs)
    - [switch_port_settings_clear.rs](https://github.com/oxidecomputer/omicron/blob/rpws-for-all-networking/nexus/src/app/sagas/switch_port_settings_clear.rs)
    - [switch_port_settings_common.rs](https://github.com/oxidecomputer/omicron/blob/rpws-for-all-networking/nexus/src/app/sagas/switch_port_settings_common.rs)
    - Along with all the saga plumbing in the API handlers like [this](https://github.com/oxidecomputer/omicron/blob/rpws-for-all-networking/nexus/src/app/switch_port.rs#L111-L128).

- When certain API actions are taken, I think we'll want to trigger the RPW instead of waiting for the next period. I did this for BFD like [this](https://github.com/oxidecomputer/omicron/blob/main/nexus/src/app/bfd.rs#L39-L41).

internet-diglett:
@rcgoodfellow I actually had a similar thought, was going to talk to you about it but it seems you are thinking similarly. I will pull out the saga and update with RPW triggers and give it a test.

nexus/src/app/background/sync_switch_port_settings.rs
```diff
@@ -0,0 +1,1236 @@
...
+                    }
+                }
+            }
+            json!({})
+        }
+        .boxed()
+    }
+}
+
+fn uplinks(
```
rcgoodfellow, line 769: Just a note that wee need to re-evaluate

- oxidecomputer/customer-support#95

with this new code path for uplink setup in place.

nexus/src/app/background/sync_switch_port_settings.rs
```diff
@@ -0,0 +1,1236 @@
...
+                    for peer in &settings.bgp_peers {
+                        let bgp_config_id = peer.bgp_config_id;
+
+                        // since we only have one bgp config per switch, we only need to fetch it once
+                        let bgp_config = match switch_bgp_config.entry(*location) {
+                            Entry::Occupied(occupied_entry) => {
+                                let (existing_id, existing_config) = occupied_entry.get().clone();
+                                // verify peers don't have differing configs
+                                if existing_id != bgp_config_id {
+                                    // should we flag the switch and not do *any* updates to it?
```
rcgoodfellow, line 397: I think logging and error and continuing is good here. We should think about how to get this information to operators. It should be this case that the RPW will converge to the intended state when remedial action is taken by an operator.

nexus/src/app/background/sync_switch_port_settings.rs
```diff
@@ -0,0 +1,1236 @@
...
+                                },
+                            };
+
+                            let mut prefixes: Vec<Prefix4> = vec![];
+
+                            for announcement in &announcements {
+                                let value = match announcement.network.ip() {
+                                    IpAddr::V4(value) => value,
+                                    IpAddr::V6(a) => {
+                                        error!(log, "bad request, only ipv4 supported at this time"; "requested_address" => ?a);
```
rcgoodfellow, line 472: Just a note that we should validate this in the API call handler and never actually get here. We should file an issue for this if there is not one already.

nexus/src/app/background/sync_switch_port_settings.rs
```diff
@@ -0,0 +1,1236 @@
...
+                // calculate and apply bootstore changes
+                //
+
+                // check downstream bootstore version
+                let mut downstream_config: Option<EarlyNetworkConfig> = None;
+
+                // Since we update the first scrimlet we can reach (we failover to the second one
+                // if updating the first one fails) we need to check them both.
+                for (_location, client) in &sled_agent_clients {
+                    let scrimlet_cfg  = match client.read_network_bootstore_config_cache().await {
```
rcgoodfellow, line 583: I don't think it's correct to read a bootstore config from a sled agent and then use the generation number as the basis for whether we need to bump the config as we're doing below with the `bootstore_needs_update` check. The sole source of truth is the nexus DB and the bootstore/sled agents below are in a constant state of eventual consistency convergence, so we should really be treating them unconditional write-only entities. @andrewjstone can keep me honest here.

nexus/src/app/background/sync_switch_port_settings.rs
```diff
@@ -0,0 +1,1236 @@
...
+                        error!(log, "rack subnet must be ipv6"; "rack" => ?rack);
+                        continue;
+                    },
+                    None => {
+                        error!(log, "rack subnet not set"; "rack" => ?rack);
+                        continue;
+                    }
+                };
+
+                // TODO: @rcgoodfellow is this correct? Do we place the BgpConfig for both switches in a single Vec to send to the bootstore?
```
rcgoodfellow, line 620: The primary concern is that this is consistent with that the sled-agent is expecting during early networking. So that is where I would go check.

nexus/src/app/background/sync_switch_port_settings.rs
```diff
@@ -0,0 +1,1236 @@
...
+                            // inconsistent with a generic set of addresses on ports.
+                            infra_ip_first: Ipv4Addr::UNSPECIFIED,
+                            infra_ip_last: Ipv4Addr::UNSPECIFIED,
+                            ports,
+                            bgp,
+                        }),
+                    },
+                };
+
+                // check to see if our config is different(?)
```
rcgoodfellow, line 709: While we do need to make this check, per my comments above, it needs to be based on the bootstore version in the DB, not one that is being reported by a particular sled-agent.

nexus/src/app/background/sync_switch_port_settings.rs
```diff
@@ -0,0 +1,1236 @@
...
+                        &log,
+                        "updating bootstore config";
+                        "old config" => ?downstream_config,
+                        "new config" => ?upstream_config,
+                    );
+
+                    // push the updates to both scrimlets
+                    // if both scrimlets are down, bootstore updates aren't happening anyway
+                    for (location, client) in &sled_agent_clients {
+                        if let Err(e) = client.write_network_bootstore_config(&upstream_config).await {
```
rcgoodfellow, line 751: Maybe this can be relaxed to have at least one successful write semantics. I think if we write the bootstore to one sled agent, does not matter which, it's the bootstore protocol's responsibility to propagate to others. cc: @andrewjstone.

nexus/src/app/background/sync_switch_zone_smf.rs
```diff
@@ -0,0 +1,39 @@
...
+    datastore: Arc<DataStore>,
+}
+
+impl SwitchZoneSmfManager {
+    pub fn new(datastore: Arc<DataStore>) -> Self {
+        Self { datastore }
+    }
+}
+
+impl BackgroundTask for SwitchZoneSmfManager {
```
rcgoodfellow, line 27: Task does nothing?

tools/dendrite_openapi_version
```diff
@@ -1,2 +1,2 @@
-COMMIT="3618dd6017b363c5d34399273453cf50b9c9a43e"
-SHA2="aa670165e5b459fab4caba36ae4d382a09264ff5cf6a2dac0dae0a0db39a378e"
+COMMIT="d17bc65973d2329643482332dcf7fa2dc568cf00"
```
rcgoodfellow, line 1: please leave a comment in the Omicron PR with a link to your Dendrite PR

wicketd/src/preflight_check/uplink.rs
```diff
@@ -788,13 +779,10 @@ fn build_port_settings(
...
     for r in &uplink.routes {
-        if let (IpNetwork::V4(dst), IpAddr::V4(nexthop)) =
+        if let (IpNetwork::V4(_dst), IpAddr::V4(_nexthop)) =
             (r.destination, r.nexthop)
         {
-            port_settings.v4_routes.insert(
-                dst.to_string(),
-                vec![RouteSettingsV4 { link_id: link_id.0, nexthop }],
-            );
+            // TODO: do we need to create config for mgd?
```
rcgoodfellow, line 785: Yes, I believe we should handle that here.

rcgoodfellow (commented):
Thanks Levon, this looks pretty good. Initial testing in `a4x2` checks out. Even did a scrimlet cold boot and that went well.

A few comments below, along with solicitation for feedback from folks who know various subsystems much better than me.

nexus/src/app/background/sync_bootstore_network_settings.rs
```diff
@@ -0,0 +1,37 @@
...
+    datastore: Arc<DataStore>,
+}
+
+impl BootstoreNetworkSettingsManager {
+    pub fn new(datastore: Arc<DataStore>) -> Self {
+        Self { datastore }
+    }
+}
+
+impl BackgroundTask for BootstoreNetworkSettingsManager {
```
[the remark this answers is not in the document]
internet-diglett: Ah, yes there was an initial thought to have separate reconcilers, but I realized they all depended on a lot of the same state so they ended up in the same background task. I'll clean these up.

nexus/src/app/background/sync_switch_port_settings.rs
```diff
@@ -0,0 +1,1236 @@
...
+        async move {
+            let log = &opctx.log;
+
+            // Check to see if task has been enabled
+            // This task is not enabled until Nexus completes the rack initialization post-RSS,
+            // otherwise there is a race where Nexus has not finished populating the DB with
+            // information from the handoff message but the RPW is attempting to reconcile against
+            // information in the DB. This happens because RPWs are initialized *before* Nexus is
+            // initialized.
+            match self.datastore.get_background_task_toggle(opctx, SYNC_SWITCH_PORT_SETTINGS.into()).await {
```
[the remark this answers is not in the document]
internet-diglett: Yeah, the initial implementation is very much a test of "does this idea work well in practice?" and I'm totally okay with workshopping this a bit more.

nexus/src/app/background/sync_switch_port_settings.rs
internet-diglett: @andrewjstone and I actually discussed this a bit last week, but it's probably good to also discuss it here. The challenge we currently have is that we don't know _when_ the "db bootstore config" has actually drifted from what's on the bootstore, so we don't know when to bump the version number and push a change. Previously this logic was in a saga, and the saga only ran when there was a change, so we always bumped the number and pushed a new bootstore config to the sled agents. Here, we run every N seconds even if there wasn't a change. 

> I don't think it's correct to read a bootstore config from a sled agent and then use the generation number as the basis for whether we need to bump the config as we're doing below with the bootstore_needs_update check. 

So comparing the version is actually a mistake that explains a bug I just noticed! 😂 

The rest of the logic in that block was to address a shortcoming we have with our bootstore config tracking in Nexus:

The logic here was _intended_ to check both scrimlets to see who has the highest config, and then compare the other contents of what's in the db (ntp_servers & rack_network_config) against that config to see if a drift has occurred. IIUC, there _shouldn't_ be a scenario where the scrimlets are lagging behind any other sleds. Admittedly it's a bit of a hack. If we want to properly track changes to the bootstore config we'll probably need to add another table)

The bug I noticed is that the RPW is _always_ bumping the version number and updating the bootstore config, even if it hasn't changed, because our "db bootstore config" is being initialized with a version of `0` and then set to the next generation number via the `bump_version` datastore method if the RPW believes the bootstore needs to be updated (which it believes to be true because it's comparing the wrong field). This still results in a system that converges to the correct settings, but it means that the sled-agents a propagating bootstore configs far more often than they need to. 😂

nexus/src/app/background/sync_switch_port_settings.rs
internet-diglett: I believe this is the relaxed implementation? If any (or all) of the requests fail, we simply log the error and continue. We try both scrimlets just in case one is down for a prolonged period of time.

internet-diglett: I think I better understand what you're saying, which is if we're successful on one, we don't need to try the second. I'm ok with that, but I think the eagerly updating multiple agents falls into the bootstore "seeding" idea that has been discussed offline, where the propagation happens faster if multiple agents receive the update instead of one, but I also defer to @andrewjstone on this.

nexus/db-queries/src/db/datastore/background_task_toggle.rs
```diff
@@ -0,0 +1,94 @@
...
+    },
+};
+use async_bb8_diesel::AsyncRunQueryDsl;
+use diesel::{prelude::*, ExpressionMethods, SelectableHelper};
+use omicron_common::api::external::{CreateResult, Error, LookupResult};
+
+use super::DataStore;
+
+impl DataStore {
+    pub async fn set_background_task_toggle(
```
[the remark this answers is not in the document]
internet-diglett: Updated the docs in the struct definition with a bit more commentary

nexus/src/app/background/sync_switch_port_settings.rs
internet-diglett: Related:
* https://github.com/oxidecomputer/omicron/issues/5085
* https://github.com/oxidecomputer/omicron/pull/5107

internet-diglett: I'll check to see if there is an issue specific to BGP as well.

nexus/src/app/background/sync_switch_port_settings.rs
```diff
@@ -0,0 +1,1236 @@
...
+                };
+
+                apply_switch_port_changes(dpd_clients, &changes, log).await;
+
+                //
+                // calculate and apply routing changes
+                //
+
+                // get the static routes on each switch
+                let downstream_static_routes =
```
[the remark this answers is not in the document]
internet-diglett: Apologies for this, my brain switched to "upstream and downstream" in terms of repositories / source-of-truth. I've made this more consistent with `<things>_on_switch` and `desired_<things>`,

nexus/src/app/background/sync_switch_port_settings.rs
andrewjstone: Yes, this is what I called "seeding". It's a mechanism used by some other protocols including [rapid](https://research.vmware.com/publications/stable-and-consistent-membership-at-scale-with-rapid). However, it's really just for reliability here and not for performance. If the original node dies permanently the data won't replicate. You get better odds with N replicas that such a thing won't happen.

FelixMcFelix (commented):
I had a quick look over and I think the design of the RPW task is clear! The comment on sync_switch_port_settings.rs:191-209 really explains the context around some of the extra machinery you need here, which I found useful as I've not had too much time to dig into RSS/early networking yet. :) I don't have anything to add that's not been raised by other folks.

nexus/src/app/background/sync_switch_port_settings.rs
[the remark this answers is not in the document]
davepacheco: Thanks for checking in!  tl;dr: it makes sense to me that we need to store in the database a record of the transition from "the database state is not yet authoritative [and the background task must not try to propagate it]" to "the database state is authoritative [and the background task must propagate it]".  I don't love the idea of generalizing this to an on/off switch for background tasks.  How about storing a purpose-specific record somewhere that says "we have finished loading the pre-RSS networking state" and then just check for that?  To me, describing this as a generic "toggle" suggests (wrongly) that it's a question of administrative/operational policy whether the switch is "on" or "off", and that someone who knows what they're doing might reasonably flip it.  That's not the case here, though: it's a state transition that happens at a very precise time and (as far as I can tell?) never goes back.  So I think a purpose-specific record in the database makes more sense.

To elaborate (this probably doesn't add anything but writing it has helped me figure out what makes me nervous about this): I'm admittedly bringing some of my own expectations to this.  I have long expected that we _would_ build switches in the database for disabling background tasks for emergency situations -- chicken switches, basically, to give us extra tools when responding to complex, unforeseen problems.  This kind of stuff has been tremendously valuable for me in the past during incident response.  This _looks_ a lot like that kind of thing.  But it's very different: with an operational override, there's one "normal" position, and one "overridden by a human due to emergency" position.  The system would generally only ever run in one position unless a human decided to override things.  There's no semantics associated with "on" or "off" other than "normal" and "overridden".  None of that's true for this switch.  The mechanism you need here is not a policy question of "should this thing run or not" but rather "is it safe/valid/correct for this thing to run" vs. "is it safe/valid/correct for it not to run".  Hence my suggestion to make this a purpose-specific database record: the basic idea (a barrier stored in the database) makes sense to me but the generalization as an on/off switch feels problematic.

nexus/src/app/background/sync_switch_port_settings.rs
andrewjstone: I forgot to follow up here. Levon and I did chat about this, and what he was doing reading from the bootstore was safe, but it's still not how it was intended to be used. Really, all reads and writes should go through CRDB to linearize them and then they should be pushed to bootstore. Levon is going to take a crack at adding a table for the bootstore data so a history of updates can be maintained and then only writing to bootstore.

internet-diglett:
> I have long expected that we would build switches in the database for disabling background tasks for emergency situations -- chicken switches, basically, to give us extra tools when responding to complex, unforeseen problems. This kind of stuff has been tremendously valuable for me in the past during incident response. This looks a lot like that kind of thing. But it's very different: with an operational override, there's one "normal" position, and one "overridden by a human due to emergency" position.

@davepacheco it's interesting that you mention that, because I actually left a comment on the "db model" struct for the toggle expressing a similar idea. I historically have viewed toggles in software as a "generic switch" that can be flipped on or off for various reasons, not necessarily _only_ for emergencies, but for the kind of software we're writing here I can see why we'd want to make an explicit distinction.

>  I don't love the idea of generalizing this to an on/off switch for background tasks. How about storing a purpose-specific record somewhere that says "we have finished loading the pre-RSS networking state" and then just check for that?

Combining the feedback you've provided, would it be acceptable to change the existing table to something like:

```SQL
CREATE TABLE IF NOT EXISTS system_flags (
    name STRING PRIMARY KEY,
    -- other stuff if needed
)
```

Which would then look something like this in practice:

```rust
// for situations like waiting for RSS -> handoff to complete, we can do something like the following:
if self.datastore.get_system_flags(opctx, HANDOFF_TO_NEXUS_COMPLETED).await?.is_none() {
  // do no-op
}

// maybe for situations where we want to stop a particular task in case of emergency, we can do the following?
if self.datastore.get_system_flags(opctx, DISABLE_BACKGROUND_TASK_FOO).await?.is_some() {
  // do no-op
}
```

So now the name of the flag would would indicate its intended use, and its presence in the DB would be what
triggers the behavior we want.

nexus/src/app/background/sync_switch_port_settings.rs
```diff
@@ -0,0 +1,1240 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! Background task for propagating per-port switch settings to management daemons
```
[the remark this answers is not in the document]
internet-diglett: I have already begun breaking some of the steps down into functions due to similar thoughts, the primary driver being that I want other to be able to read the code in `activate()` and be able to understand what it is doing at a high level. The bootstore and BGP stuff hasn't been broken down yet because it's going through another round of refactoring (primarily the bootstore update logic, which depends on the BGP update logic).

davepacheco:
> > I don't love the idea of generalizing this to an on/off switch for background tasks. How about storing a purpose-specific record somewhere that says "we have finished loading the pre-RSS networking state" and then just check for that?
> 
> Combining the feedback you've provided, would it be acceptable to change the existing table to something like:
> 
> ```sql
> CREATE TABLE IF NOT EXISTS system_flags (
>     name STRING PRIMARY KEY,
>     -- other stuff if needed
> )
> ```
> 
> Which would then look something like this in practice:
> 
> ```rust
> // for situations like waiting for RSS -> handoff to complete, we can do something like the following:
> if self.datastore.get_system_flags(opctx, HANDOFF_TO_NEXUS_COMPLETED).await?.is_none() {
>   // do no-op
> }
> 
> // maybe for situations where we want to stop a particular task in case of emergency, we can do the following?
> if self.datastore.get_system_flags(opctx, DISABLE_BACKGROUND_TASK_FOO).await?.is_some() {
>   // do no-op
> }
> ```
> 
> So now the name of the flag would would indicate its intended use, and its presence in the DB would be what triggers the behavior we want.

That does seem a little harder to accidentally misuse but it still puts "critical application state that's only supposed to be changed in specific ways by the application" into the same bucket as "operator/support policy that could be reasonably changed by someone who knows what they're doing at any time".  My intuition is to avoid generalizing this mechanism before we've got a second consumer.  For example, if/when we go add support for disabling background tasks, we might end up wanting something richer than this anyway (e.g., configure the period, too? record the time? keep an audit log?) and then not use this mechanism.

I do get the desire to generalize it because it seems weird to have a table with just one bit of information.  But it seems to me that this really is part of ordinary application state, not a grab bag of global tunables, and I'd consider just creating a one-off table with at most one record in it.  We could even put some debugging information in there, like:

```sql
CREATE TABLE networking_handoff (
    nexus_id UUID NOT NULL,
    -- See `db_metadata` table for more on the `singleton` pattern.
    singleton BOOL NOT NULL PRIMARY KEY,
    CHECK (singleton = true)
);
```

---

Also, elsewhere we assume "handoff to Nexus completed" == "there is a record in the Rack table".  For example see [here](https://github.com/oxidecomputer/omicron/blob/rpws-for-all-networking/nexus/src/app/rack.rs#L627).  This is also specified in RFD 278 ("The transfer-of-control operation should synchronously write some state to the database that allows both this Nexus instance and others to tell that transfer-of-control has happened. Transfer of control happens when this database transaction commits, even if the API call gets interrupted, times out, or fails.")  I looked at the implementation of `Nexus::rack_initialize()` and I see the networking state you're using is written _outside_ of that transaction, after the transaction.  So if I'm understanding right, the event you're waiting for is not "handoff to Nexus completed", but rather something that happens shortly after that.  I called it "networking handoff" above; you could call it something else but it's not quite "RSS handoff to Nexus".

That said, if you could move all this into the same rack initialization database transaction, then you could use the existing "rack" record and not need to write anything else.  This could be a ton of work though depending on how that code is currently factored.

schema/crdb/dbinit.sql
```diff
@@ -3545,14 +3545,29 @@ SELECT
...
+CREATE TABLE IF NOT EXISTS omicron.public.background_task_toggles (
+    id UUID NOT NULL DEFAULT uuid_generate_v4(),
+    name STRING NOT NULL,
+    enabled BOOLEAN NOT NULL,
+    time_created TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    time_modified TIMESTAMPTZ ON UPDATE NOW() NOT NULL DEFAULT NOW(),
+    time_deleted TIMESTAMPTZ
+);
+
+CREATE UNIQUE INDEX IF NOT EXISTS background_task_toggles_name
```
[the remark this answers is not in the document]
internet-diglett: > Why bother with an id here? Why not make name the primary key?
For some reason I recall there being a bit of discussion debating whether we should always use surrogate keys or if it was okay to opt for using natural keys. I don't really mind either way.

> (I think we generally have avoided default values in the database and specify them from Rust instead. Looking at dbinit.sql I see that we've been a little inconsistent about this but there are only 5 instances in that file where time_created or time_modified has DEFAULT NOW() compared with about 100 where they don't.)

I'm okay with changing this as well, but I guess I'm a bit fuzzy on how loose or strict we're being with our timestamps, as I see calls to `chrono::Utc::now()` as well as diesel's `dsl::some_timestamp_column.eq(Utc::now())` on the Rust side. I'm guessing when you say "set it in Rust" you're referring to doing it via a diesel query, which I'm guessing does `NOW()` in the query builder?

I'm asking about this distinction because I'm concerned about the following:

When a particular instance of nexus creates / modifies a record and sets the timestamp via Rust code via `chrono::Utc::now()`, that record has not yet been created / modified from the perspective of all of the other nexus instances. As far as I can tell, the _real_ `time_*` is when the DB commits the transaction for all of the other Nexus instances to see. When we let the DB set `time_created` and `time_modified`, we are getting a more accurate representation of when these events are visible and actionable from the perspective of all of the other Nexus instances.

Consider the following circumstance:
1. Nexus A timestamps a record, in Rust, with Timestamp1
2. Nexus B timestamps a record with Timestamp2
3. They both insert at roughly the same time, but for whatever reason (latency, disk, etc.) the DB commits the record from Nexus B _first_.
4. When operators / developers look at the DB, we are led to believe that the record from Nexus A was created _before_ Nexus B, but the truth is the that the record for Nexus B was there first, and could have been affecting application behavior before Nexus A.

Being that our application embraces a lot of idempotent & eventually-consistent patterns, I don't _think_ this will get us into trouble, but I think it's worth considering this scenario because it could create some confusing moments down the road whenever we try to use the timestamp data.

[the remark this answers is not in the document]
davepacheco: Yeah, timestamps cannot be used for correctness for the reasons you mentioned.  Even when they're generated by the database, it doesn't reflect the time at which the data became visible to other transactions.  (That's not even well-defined, since different CockroachDB nodes will have slightly different times.)  I think the database evaluates `NOW()` when parsing or processing the query but decidedly before the transaction is committed.

I tried this: in one shell, `cockroach demo` and set up a table:

```
dap@zathras ~ $ cockroach demo
#
# Welcome to the CockroachDB demo database!
#
# You are connected to a temporary, in-memory CockroachDB cluster of 1 node.
#
# This demo session will attempt to enable enterprise features
# by acquiring a temporary license from Cockroach Labs in the background.
# To disable this behavior, set the environment variable
# COCKROACH_SKIP_ENABLING_DIAGNOSTIC_REPORTING=true.
#
# Beginning initialization of the movr dataset, please wait...
#
# The cluster has been preloaded with the "movr" dataset
# (MovR is a fictional vehicle sharing company).
#
# Reminder: your changes to data stored in the demo session will not be saved!
#
# If you wish to access this demo cluster using another tool, you will need
# the following details:
#
#   - Connection parameters:
#     (webui)    http://127.0.0.1:8080/demologin?password=demo75304&username=demo
#     (sql)      postgres://demo:demo75304@127.0.0.1:26257?sslmode=require
#     (sql/unix) postgres://demo:demo75304@?host=%2Fvar%2Ffolders%2Fx3%2F6tbrhnss6zz677z4zjsyvkqc0000gn%2FT%2Fdemo351111831&port=26257
#   
#   - Username: "demo", password: "demo75304"
#   - Directory with certificate files (for certain SQL drivers/tools): /var/folders/x3/6tbrhnss6zz677z4zjsyvkqc0000gn/T/demo351111831
#
# Server version: CockroachDB CCL v21.1.10 (x86_64-apple-darwin19, built 2021/10/07 02:42:10, go1.15.14) (same version as client)
# Cluster ID: 9034e646-3342-4a6a-b2cd-d0da89bc3ebe
# Organization: Cockroach Demo
#
# Enter \? for a brief introduction.
#
demo@127.0.0.1:26257/movr> create table t1 (c INT, v TIMESTAMPTZ);
CREATE TABLE

Time: 7ms total (execution 7ms / network 0ms)

```

In that shell, open a transaction, do an INSERT with `NOW()`, but don't commit it yet:

```
demo@127.0.0.1:26257/movr> BEGIN;
BEGIN

Time: 0ms total (execution 0ms / network 0ms)

demo@127.0.0.1:26257/movr  OPEN> INSERT INTO t1 (c, v) VALUES (1, NOW());
INSERT 1

Time: 1ms total (execution 1ms / network 0ms)


```

Now wait like 10 seconds.  Then in another shell, run `cockroach sql` to connect to the same database and do a similar INSERT but right away:

```
dap@zathras ~ $ cockroach sql --url 'postgres://demo:demo75304@127.0.0.1:26257?sslmode=require'
#
# Welcome to the CockroachDB SQL shell.
# All statements must be terminated by a semicolon.
# To exit, type: \q.
#
# Server version: CockroachDB CCL v21.1.10 (x86_64-apple-darwin19, built 2021/10/07 02:42:10, go1.15.14) (same version as client)
# Cluster ID: 9034e646-3342-4a6a-b2cd-d0da89bc3ebe
# Organization: Cockroach Demo
#
# Enter \? for a brief introduction.
#
demo@127.0.0.1:26257/defaultdb> use movr;
SET

Time: 1ms total (execution 1ms / network 0ms)

demo@127.0.0.1:26257/movr> INSERT INTO t1 (c, v) VALUES (2, NOW());
INSERT 1

Time: 2ms total (execution 2ms / network 0ms)

```

As soon as possible, in the first one, COMMIT:

```
demo@127.0.0.1:26257/movr  OPEN> COMMIT;
COMMIT

Time: 1ms total (execution 1ms / network 0ms)
```

So we did the INSERTs about 10 seconds apart but the commits about the same time.  This is what we get:

```
demo@127.0.0.1:26257/movr> SELECT * from t1;
  c |               v
----+--------------------------------
  1 | 2024-03-06 20:11:35.018453+00
  2 | 2024-03-06 20:11:49.489064+00
(2 rows)

Time: 1ms total (execution 1ms / network 0ms)
```

i.e., the timestamps are 14 seconds apart, which means the timestamp of the first one was 14 seconds before it was comitted.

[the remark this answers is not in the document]
davepacheco: I tried to go further and show an intermediate state where the *later* row is visible before the earlier row is.  This is hard because of CockroachDB's serializeable guarantee.  What happens instead is that if you try to list the table, it blocks until you commit the open transaction.  I'm fairly sure I've seen this sort of thing with PostgreSQL but it's not as easy for me to test that.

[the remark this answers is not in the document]
internet-diglett: :open_mouth: 

_the entire world is a lie_

But seriously, thanks for taking the time to provide that demo! Yet another interesting nuance of building distributed systems! :joy:

nexus/db-queries/src/db/datastore/address_lot.rs
```diff
@@ -53,12 +53,33 @@ impl DataStore {
...
-                    diesel::insert_into(lot_dsl::address_lot)
-                        .values(lot)
-                        .returning(AddressLot::as_returning())
-                        .get_result_async(&conn)
-                        .await?;
+                // @internet-diglett says:
+                // I hate this. I know how to replace this transaction with
+                // CTEs but for the life of me I can't get it to work in
+                // diesel. I gave up and just extended the logic inside
+                // of the transaction instead chasing diesel trait bound errors.
```
smklein, line 60: While it's fresh: I don't want to block your PR, but if you have the equivalent SQL for a CTE, could you post it in an issue? I can collaborate with you on making that work separately from this PR.

nexus/db-queries/src/db/datastore/address_lot.rs
```diff
@@ -53,12 +53,33 @@ impl DataStore {
...
+                            .eq(Name::from(params.identity.name.clone())),
+                    )
+                    .filter(lot_dsl::time_deleted.is_null())
+                    .select(AddressLot::as_select())
+                    .limit(1)
+                    .first_async(&conn)
+                    .await
+                    .ok();
+
+                let db_lot = match found_lot {
```
smklein, line 73: In the context of a transaction, this "SELECT or else INSERT" behavior seems fine to me. "INSERT ... ON CONFLICT ... RETURNING" can really only tell you the value you *just* inserted anyway, so it'll omit any value which exists in the DB already.

(As you're already well-aware, CTEs are the right way to reduce this load with a smaller number of queries)

nexus/db-queries/src/db/datastore/address_lot.rs
```diff
@@ -225,6 +278,35 @@ impl DataStore {
...
+        &self,
+        opctx: &OpContext,
+        name: String,
+    ) -> LookupResult<Vec<AddressLotBlock>> {
+        let conn = self.pool_connection_authorized(opctx).await?;
+
+        use db::schema::address_lot::dsl as lot_dsl;
+        use db::schema::address_lot_block::dsl as block_dsl;
+
+        let address_lot_id = lot_dsl::address_lot
```
smklein, line 293: nitpick: This could be a single query with a join, but I think this is also fine as-is

nexus/db-queries/src/db/datastore/address_lot.rs
```diff
@@ -72,6 +93,38 @@ impl DataStore {
...
+
+                let blocks = blocks
+                    .into_iter()
+                    .filter(|b| {
+                        !db_blocks.iter().any(|db_b| {
+                            db_b.first_address == b.first_address
+                                || db_b.last_address == b.last_address
+                        })
+                    })
+                    .collect::<Vec<_>>();
```
smklein, line 126: To be clear, this is looking for cases where:

- We have `blocks` that come from params
- In the database, we looked up `db_blocks`, which come from the same address lot and happen to have the same "first address" or the same "last address" as the parameter-supplied `blocks` (lines 99-114)

And this particular filter identifies:
- If there are **any** `blocks` where `db_blocks` has the same "first address" or "last address", we filter them out of `blocks`?

Am I misunderstanding this? Aren't we explicitly looking for that overlap - why then filter it out?

internet-diglett: We're _supposed_ to be filtering for blocks to insert, but I may have gotten my wires crossed here. I'll take a second look here.

internet-diglett: Yeah, this actually doesn't work quite right for what we're planning to return below. I'll need to tweak it. It should be _inserting_ the correct stuff though.

internet-diglett: @smklein this should be fixed now

nexus/db-queries/src/db/datastore/address_lot.rs
internet-diglett: https://github.com/oxidecomputer/omicron/issues/5277

rcgoodfellow (approved):
Things look good in a4x2. Onward to on-rack testing.

PR #4823, opened 2024-01-17
[tufaceous-lib] add fake artifacts with an mtime of zero

sunshowers:
In #4690 I'd like to be able to generate a tufaceous repo twice and expect that each artifact has the same hash.

Writing tests for this is a bit annoying at the moment because at the moment the overall repo has a different hash when generated again (because TUF bakes timestamps into its format), but each artifact has the same hash. #4690 has other work going on which makes writing this test quite simple, so I'll include the test there.

### DIFF
--- a/tufaceous-lib/src/oxide_metadata.rs
+++ b/tufaceous-lib/src/oxide_metadata.rs
@@ -10,10 +10,7 @@
  * Copyright 2023 Oxide Computer Company
  */
 
-use std::{
-    collections::HashMap,
-    time::{SystemTime, UNIX_EPOCH},
-};
+use std::collections::HashMap;
 
 use anyhow::{bail, Result};
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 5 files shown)


tufaceous-lib/src/oxide_metadata.rs
```diff
@@ -52,8 +49,9 @@ impl Metadata {
         let mut b = serde_json::to_vec(self)?;
         b.push(b'\n');
 
-        let mtime =
-            SystemTime::now().duration_since(UNIX_EPOCH).unwrap().as_secs();
+        // XXX This was changed from upstream to add oxide.json with a zero
+        // timestamp, to ensure stability of fake manifests.
```
jgallagher, line 53: Should this method take an `MtimeSource`?

jgallagher (approved):
LGTM, just one question about the `XXX` comment

tufaceous-lib/src/oxide_metadata.rs
sunshowers: Hmm, I feel like I tried that and it didn't quite work, but I might be wrong -- let me check.

sunshowers: Oh yeah, the thing I ran into is that at the moment we don't have mtime source annotations in a few spots. For example,

```toml
[[artifact.host]]
name = "fake-host"
version = "1.0.0"
[artifact.host.source]
kind = "composite-host"
phase_1 = { kind = "fake", size = "512KiB" }
phase_2 = { kind = "file", path = "foo/bar" }
```

In this case, what should the mtime of `oxide.json` be? I punted on that.

However, on a re-look I realized that we can just base this on whether all of the sub-artifacts are fake. Updated the PR with this new approach.

Thanks for pushing me on this!

PR #4831, opened 2024-01-17
[common] switch nexus config to using camino paths

sunshowers:
In #4690 I was dealing with some of these paths, and conversion to and fro is a
bit annoying.

There is no loss of generality here because all of these paths are read from a
config file -- and are therefore expected to be UTF-8 anyway.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4738,6 +4738,7 @@ dependencies = [
  "base64",
  "buf-list",
  "camino",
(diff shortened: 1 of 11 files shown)


smklein (approved):
Thanks for cleaning this up!

david-crespo (approved):
console API bits look good and I appreciate the tweaks

PR #4833, opened 2024-01-18
add issue templates for flaky tests

davepacheco:


davepacheco:
Nobody ever reviewed this but I think it's pretty inconsequential and would be useful to me so I'm planning to just land it.

PR #4834, opened 2024-01-18
more specific address reservation errors

rcgoodfellow:
I was trying to use the rack today to do some dev/test. I started by creating a VM with the requested address `10.0.0.2` and was greeted with this error.

```
Address 10.0.0.2 not in subnet 10.0.0.0/24 or is reserved for rack services
```

The former is not true, which leaves me with the latter. Then the question is, what addresses are reserved for rack services? This PR changes the error message to be more direct and specific about the issue at hand.

Possible errors are now

```rust
pub enum RequestAddressError {
    #[error("{} is outside subnet {}", .0, .1)]
    OutsideSubnet(IpAddr, ipnetwork::IpNetwork),
    #[error(
        "The first {} addresses of a subnet are reserved",
        NUM_INITIAL_RESERVED_IP_ADDRESSES
    )]
    Reserved,
    #[error("Cannot request a broadcast address")]
    Broadcast,
}
```

### DIFF
--- a/nexus/db-model/src/ipv4net.rs
+++ b/nexus/db-model/src/ipv4net.rs
@@ -2,6 +2,7 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+use crate::vpc_subnet::RequestAddressError;
 use diesel::backend::Backend;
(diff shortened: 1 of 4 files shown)


FelixMcFelix (approved):
Thanks, this makes the error condition here far clearer.

PR #4840, opened 2024-01-18
Package omdb into global zone

andrewjstone:


leftwo:
Thanks x 500 for adding this

I wonder if some sort of "tools" directory might be worth considering.  This, oxlog, and possibly other tools we want to
build could all be there, and we could add that path to the PATH.  Maybe not for this PR, but something to consider.

davepacheco:
💯 on doing this.  I don't think I grok what actually happens with this change.  I gather we get a new Omicron package with the binary in it.  Does that wind up in the TUF repo somehow?  What exactly winds up delivered onto sleds that go through like a mupdate or the usual setup process?

andrewjstone:
> 💯 on doing this. I don't think I grok what actually happens with this change. I gather we get a new Omicron package with the binary in it. Does that wind up in the TUF repo somehow? What exactly winds up delivered onto sleds that go through like a mupdate or the usual setup process?

I'm not 100% sure. I looked through the package and tuf-repo buildomat steps and it appears like omicron-package is run and then tuf repackages the output zones. That's how I assumed it worked and nothing so far has disproven that.

andrewjstone:
I'm going to go ahead and also add this to the switch zone, as that's where its most useful.

smklein:
Presumably you want this bundled with the host OS, right? 

The buildomat [package job](https://github.com/oxidecomputer/omicron/blob/main/.github/buildomat/jobs/package.sh) creates [several files that end up going into the global zone](https://github.com/oxidecomputer/omicron/blob/6b20cd25cae650c35f0a5c1d02f8ab5d86a6223d/.github/buildomat/jobs/package.sh#L56-L97).

It appears to invoke the [build-global-zone-package script](https://github.com/oxidecomputer/omicron/blob/main/tools/build-global-zone-packages.sh) to do this (and there appears to be some dependency checking [here](https://github.com/oxidecomputer/omicron/blob/main/tools/build-global-zone-packages.sh#L12-L18C2), if you wanna omdb to it).

These global zone packages are then shoved into the host OS image:

https://github.com/oxidecomputer/omicron/blob/6b20cd25cae650c35f0a5c1d02f8ab5d86a6223d/.github/buildomat/jobs/host-image.sh#L78-L84

Which is picked up by the TUF repo job.

andrewjstone:
> Presumably you want this bundled with the host OS, right?
> 
> The buildomat [package job](https://github.com/oxidecomputer/omicron/blob/main/.github/buildomat/jobs/package.sh) creates [several files that end up going into the global zone](https://github.com/oxidecomputer/omicron/blob/6b20cd25cae650c35f0a5c1d02f8ab5d86a6223d/.github/buildomat/jobs/package.sh#L56-L97).
> 
> It appears to invoke the [build-global-zone-package script](https://github.com/oxidecomputer/omicron/blob/main/tools/build-global-zone-packages.sh) to do this (and there appears to be some dependency checking [here](https://github.com/oxidecomputer/omicron/blob/main/tools/build-global-zone-packages.sh#L12-L18C2), if you wanna omdb to it).
> 
> These global zone packages are then shoved into the host OS image:
> 
> https://github.com/oxidecomputer/omicron/blob/6b20cd25cae650c35f0a5c1d02f8ab5d86a6223d/.github/buildomat/jobs/host-image.sh#L78-L84
> 
> Which is picked up by the TUF repo job.

Thanks Sean. This was very helpful information. However, I moved omdb to the switch zone in [4fa8e74](https://github.com/oxidecomputer/omicron/pull/4840/commits/4fa8e74adb27a1454ede281cdfe08c343a334ad7), and I don't believe I need to do anything special for installation regarding tuf now.

smklein:
Sounds good -- looks like your change now uses that same "package into TUF package" flow, by piggy-backing onto the switch zone, which is totally fine!

smklein (approved):

andrewjstone:
I tested on 4 node testbed and omdb does get installed although I get the following error, which I'm looking into.

andrewjstone:
Accidental button press. I was saying, I got an error trying to run omdb on the testbed.  I'm looking into it.

```
root@oxz_switch:~# omdb
ld.so.1: omdb: fatal: libpq.so.5: open failed: No such file or directory
Killed
```

smklein:
I think `libpq` must only have been bundled into the Nexus zone, and perhaps not the switch zone? I suppose it's probably installed in the global zone... somewhere.

Nexus copies all postgres shared libraries via:
https://github.com/oxidecomputer/omicron/blob/main/package-manifest.toml#L107C1-L108C1

Perhaps the switch zone now needs to do something similar?

smklein:
Also you could probably include all this info -- **and** omdb -- in the `switch_zone_setup` package, so that you don't need to re-specify these dependencies in each distinct flavor of the switch zone.

andrewjstone:
> Also you could probably include all this info -- **and** omdb -- in the `switch_zone_setup` package, so that you don't need to re-specify these dependencies in each distinct flavor of the switch zone.

This was a great idea. I also took the liberty of moving some other common packages into `switch_zone_setup` in [3abad93](https://github.com/oxidecomputer/omicron/pull/4840/commits/3abad936781bf38d181a4f71e40be6f904a26169)

Unfortunately, I'm still seeing the same missng postgres lib error. However, I believe it's due more to the testbed setup than anything else. In the switch zones of the g0 and g4 nodes of my testbed there is actually no `/opt/ooce` at all. I logged into a scrimlet on dogfood and this is not true on the scrimlet. In fact, not only does `/opt/ooce` exist there, but all the postgres libraries are present in the switch zone already.

We should test this on a bench gimlet to confirm.

andrewjstone:
> Unfortunately, I'm still seeing the same missng postgres lib error. However, I believe it's due more to the testbed setup than anything else. In the switch zones of the g0 and g4 nodes of my testbed there is actually no `/opt/ooce` at all. I logged into a scrimlet on dogfood and this is not true on the scrimlet. In fact, not only does `/opt/ooce` exist there, but all the postgres libraries are present in the switch zone already.

So, I don't think it has anything to do with the testbed. In order to include omdb (and the other switch zone packages) in the `switch_zone_setup` package I had to change `source.type` from `local` to `composite`. `omicron-package` [does not call add_paths](https://github.com/oxidecomputer/omicron-package/blob/main/src/package.rs#L739) for composite packages, and so the `source.paths` field in `package-manifest.toml` get ignored.  

With omicron-package as it stands I have to put omdb back into each different version of the switch zone package and then add the pgsql libs in `source.paths` of `switch_zone_setup` while keeping that package `local`.

Building packages and launching testbeds is a time consuming way to figure this all out. @smklein do you think it would make sense to change `omicron-package` to allow path additions for composite packages or alternately to return an error if `source.paths` is mistakenly used for a composite package?

andrewjstone:
> With omicron-package as it stands I have to put omdb back into each different version of the switch zone package and then add the pgsql libs in source.paths of switch_zone_setup while keeping that package local.

Yup! This worked!

```
root@oxz_switch:~# omdb mgs inventory
note: MGS URL not specified.  Will pick one from DNS.
note: using MGS URL http://[fd00:1122:3344:104::2]:12225
ALL CONFIGURED SPs

    TYPE   SLOT
    Sled   0
    Sled   1
    Switch 0

SPs FOUND THROUGH IGNITION

    TYPE   SLOT SYSTEM_TYPE
    Sled   0    Gimlet
    Sled   1    -
    Switch 0    Gimlet

SERVICE PROCESSOR STATES

    TYPE   SLOT MODEL            SERIAL      REV HUBRIS           PWR ROT_ACTIVE
    Sled   0    FAKE_SIM_GIMLET  SimGimlet0  0   0000000000000000 A2  slot A
    Switch 0    FAKE_SIM_SIDECAR SimSidecar0 0   0000000000000000 A2  slot A

SP DETAILS: type "Sled" slot 0

    ROOT OF TRUST

        NAME                               VALUE
        active slot                        slot A
        persistent boot preference         slot A
        pending persistent boot preference -
        transient boot preference          -
        slot A SHA3 256 digest             -
        slot B SHA3 256 digest             -

    COMPONENTS

        NAME         DESCRIPTION   DEVICE       PRESENCE SERIAL
        sp3-host-cpu FAKE host cpu sp3-host-cpu Present  None

    CABOOSES: none found

SP DETAILS: type "Switch" slot 0

    ROOT OF TRUST

        NAME                               VALUE
        active slot                        slot A
        persistent boot preference         slot A
        pending persistent boot preference -
        transient boot preference          -
        slot A SHA3 256 digest             -
        slot B SHA3 256 digest             -

    COMPONENTS: none found
```

andrewjstone:
Just to show how pumped I am to be able to run omdb on a software testbed:

```
root@oxz_switch:~# omdb nexus background-tasks list
note: using Nexus URL http://[fd00:1122:3344:104::5]:12221
task_name                PGEN# PSTART               Psecs P CSTART C
external_endpoints       3     2024-01-19T07:05:01Z   0.1 S -      -
phantom_disks            3     2024-01-19T07:04:50Z   0.0 T -      -
dns_config_internal      3     2024-01-19T07:05:01Z   0.0 S -      -
dns_propagation_internal 5     2024-01-19T07:05:01Z   1.6 D -      -
dns_servers_internal     3     2024-01-19T07:05:01Z   0.0 S -      -
inventory_collection     3     2024-01-19T07:05:01Z   1.5 S -      -
dns_servers_external     3     2024-01-19T07:05:01Z   0.0 S -      -
dns_propagation_external 5     2024-01-19T07:05:01Z   0.2 D -      -
dns_config_external      3     2024-01-19T07:05:01Z   0.0 S -      -
nat_v4_garbage_collector 3     2024-01-19T07:04:50Z   0.0 T -      -
```

```
root@oxz_switch:~# omdb db sleds
note: using database URL postgresql://root@[fd00:1122:3344:102::3]:32221,[fd00:1122:3344:103::3]:32221,[fd00:1122:3344:101::3]:32221,[fd00:1122:3344:103::4]:32221,[fd00:1122:3344:104::3]:32221/omicron?sslmode=disable
note: database schema version matches expected (24.0.0)
SERIAL                           IP                            ROLE     ID
b18cbc44066946e9ba8a0d98a9da3dbf [fd00:1122:3344:103::1]:12345 -        76c488c6-159b-4517-806b-0e608f2d13bb
0195c23def7a4683b0737093cec283b3 [fd00:1122:3344:102::1]:12345 -        804b1165-d9b4-48ba-8d56-bf97595168e3
cbdcc98421c84ec691f88bedc6caef10 [fd00:1122:3344:104::1]:12345 scrimlet 812abc1d-2039-4201-89d0-f54a9eeabe3e
9f7c2bd074d342fcbbac03f25d5a138e [fd00:1122:3344:101::1]:12345 scrimlet ab0e694d-c0cf-40bb-a669-7a6513a37ed3
```

PR #4843, opened 2024-01-18
[nexus] Endpoint to list IP pools for silo, add `is_default` to silo-scoped IP pools list

### LINKED ISSUE #4752: Give end user a way to tell which IP pool is silo default
#4261 makes `is_default` a property of the link to the silo (`IpPoolSilo`) and not of the `IpPool` itself, which means that when `/v1/ip-pools` lists `IpPool`s for the current silo, there is no way to tell which is the default, i.e., which one instance create will pull an IP from if they don't specify a pool.

The fastest way to do this would be to add `is_default: true` to the default pool in the list (if there is one). What gives me pause is that it means the endpoint no longer returns the same `IpPool` object that we are using in all the other endpoints — this one now has an `is_default` property, so it would need to be a new struct. Maybe that's fine.

Another (probably worse) approach would be to add another endpoint like `/v1/ip-pools/default` (that path only makes sense if we don't intend to add `/v1/ip-pools/{name-or-id}`, and we probably do intend to for symmetry even though it's not very important). That way `/v1/ip-pools` would not have to change, but we could still hit this other endpoint in the console to show which is the default. I am inclined against this because only the console would really benefit from it — a CLI or SDK consumer would have to do extra work to match what we do in the console.

### LINKED ISSUE #4763: Need operator endpoint to list IP pools for a silo
#4261 allows operators to list linked silos for a specified IP pool and allows non-operators to list available pools in their silo only. But it is probably more natural for the operator to think in the latter way as well: for a given silo, what are the IP pools available (and which is the default).

david-crespo:
Fixes #4752
Fixes #4763 

The main trick here is introducing `views::SiloIpPool`, which is the same as `views::IpPool` except it also has `is_default` on it. It only makes sense in the context of a particular silo because `is_default` is only defined for a (pool, silo) pair, not for a pool alone. 

- [x] Add `GET /v1/system/silos/{silo}/ip-pools`
- [x] `/v1/ip-pools` and `/v1/ip-pools/{pool}` should return `SiloIpPool` too
- [x] Tests for `/v1/system/silos/{silo}/ip-pools`
- [x] We can't have both `SiloIpPool` and `IpPoolSilo`, cleaned up by changing:
  - `views::IpPoolSilo` -> `views::SiloIpSiloLink`
  - `params::IpPoolSiloLink` -> `views::IpPoolLinkSilo`

### DIFF
--- a/nexus/db-queries/src/db/datastore/ip_pool.rs
+++ b/nexus/db-queries/src/db/datastore/ip_pool.rs
@@ -403,17 +403,22 @@ impl DataStore {
         &self,
         opctx: &OpContext,
         authz_silo: &authz::Silo,
-        pagparams: &DataPageParams<'_, Uuid>,
+        pagparams: &PaginatedBy<'_>,
     ) -> ListResultVec<(IpPool, IpPoolResource)> {
         use db::schema::ip_pool;
         use db::schema::ip_pool_resource;
 
-        paginated(
-            ip_pool_resource::table,
-            ip_pool_resource::ip_pool_id,
-            pagparams,
-        )
-        .inner_join(ip_pool::table)
+        match pagparams {
+            PaginatedBy::Id(pagparams) => {
+                paginated(ip_pool::table, ip_pool::id, pagparams)
+            }
+            PaginatedBy::Name(pagparams) => paginated(
+                ip_pool::table,
+                ip_pool::name,
+                &pagparams.map_name(|n| Name::ref_cast(n)),
+            ),
+        }
+        .inner_join(ip_pool_resource::table)
         .filter(ip_pool_resource::resource_id.eq(authz_silo.id()))
         .filter(ip_pool_resource::resource_type.eq(IpPoolResourceType::Silo))
         .filter(ip_pool::time_deleted.is_null())
--- a/nexus/db-queries/src/db/datastore/ip_pool.rs
+++ b/nexus/db-queries/src/db/datastore/ip_pool.rs
@@ -84,7 +84,7 @@ impl DataStore {
         &self,
         opctx: &OpContext,
         pagparams: &PaginatedBy<'_>,
-    ) -> ListResultVec<db::model::IpPool> {
+    ) -> ListResultVec<(IpPool, IpPoolResource)> {
         use db::schema::ip_pool;
         use db::schema::ip_pool_resource;
 
@@ -108,13 +108,10 @@ impl DataStore {
             ),
         }
         .inner_join(ip_pool_resource::table)
-        .filter(
-            ip_pool_resource::resource_type
-                .eq(IpPoolResourceType::Silo)
-                .and(ip_pool_resource::resource_id.eq(silo_id)),
-        )
+        .filter(ip_pool_resource::resource_type.eq(IpPoolResourceType::Silo))
+        .filter(ip_pool_resource::resource_id.eq(silo_id))
         .filter(ip_pool::time_deleted.is_null())
-        .select(db::model::IpPool::as_select())
+        .select(<(IpPool, IpPoolResource)>::as_select())
         .get_results_async(&*self.pool_connection_authorized(opctx).await?)
         .await
         .map_err(|e| public_error_from_diesel(e, ErrorHandler::Server))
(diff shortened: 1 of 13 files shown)


nexus/db-queries/src/db/datastore/ip_pool.rs
```diff
@@ -108,13 +108,10 @@ impl DataStore {
...
-        .filter(
-            ip_pool_resource::resource_type
-                .eq(IpPoolResourceType::Silo)
-                .and(ip_pool_resource::resource_id.eq(silo_id)),
-        )
+        .filter(ip_pool_resource::resource_type.eq(IpPoolResourceType::Silo))
+        .filter(ip_pool_resource::resource_id.eq(silo_id))
         .filter(ip_pool::time_deleted.is_null())
-        .select(db::model::IpPool::as_select())
+        .select(<(IpPool, IpPoolResource)>::as_select())
```
david-crespo, line 114: neat trick. initially I tried `(IpPool::as_select(), IpPoolResource::as_select())` but it wanted the above

nexus/db-queries/src/db/datastore/ip_pool.rs
```diff
@@ -400,6 +397,39 @@ impl DataStore {
...
+        .filter(ip_pool_resource::resource_id.eq(authz_silo.id()))
+        .filter(ip_pool_resource::resource_type.eq(IpPoolResourceType::Silo))
+        .filter(ip_pool::time_deleted.is_null())
+        .select(<(IpPool, IpPoolResource)>::as_select())
+        .load_async::<(IpPool, IpPoolResource)>(
+            &*self.pool_connection_authorized(opctx).await?,
+        )
+        .await
+        .map_err(|e| public_error_from_diesel(e, ErrorHandler::Server))
+    }
```
david-crespo, line 431: I just realized this query is exactly the same as the one above — the only difference is whether you're passing in a silo or pulling the current silo from auth context. I will extract a shared function.

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -1489,7 +1539,14 @@ async fn ip_pool_silo_list(
...
     query_params: Query<PaginatedById>,
-) -> Result<HttpResponseOk<ResultsPage<views::IpPoolSilo>>, HttpError> {
+    // TODO: this could just list views::Silo -- it's not like knowing silo_id
+    // and nothing else is particularly useful -- except we also want to say
+    // whether the pool is marked default on each silo. So one option would
+    // be  to do the same as we did with SiloIpPool -- include is_default on
+    // whatever the thing is. Still... all we'd have to do to make this usable
+    // in both places would be to make it { ...IpPool, silo_id, silo_name,
+    // is_default }
+) -> Result<HttpResponseOk<ResultsPage<views::IpPoolSiloLink>>, HttpError> {
```
zephraph, line 1549: I wonder if `views::Silo` shouldn't just be updated with `default_ip_pool: Option<Uuid>` and the silo just be returned here. There's only ever one default pool, it seems like a good thing to latch onto the silo.

david-crespo: That would mean every query for silos would have to join to the ip_pool_resource table and return both the silo and the link so we can tack on the pool ID.

nexus/db-queries/src/db/datastore/ip_pool.rs
zephraph: This came up in some of my work too. I just dropped the `current` version altogether and only left on that takes a given argument. The app layer can pass in the current silo when needed.

zephraph (approved):
Other than removing the duplicate query, this mostly looks good.

nexus/src/external_api/http_entrypoints.rs
zephraph: Despite the increased cost I still think it'd be a UX improvement for the API. That said, it doesn't need to be done here. I'm not a _huge_ fan of returning the link, but it's fine for now. It gives them all the info they need to get whatever else they need.

nexus/src/app/ip_pool.rs
```diff
@@ -109,14 +118,34 @@ impl super::Nexus {
     ) -> ListResultVec<db::model::IpPoolResource> {
         let (.., authz_pool) =
             pool_lookup.lookup_for(authz::Action::ListChildren).await?;
+
+        // check ability to list silos in general
+        opctx.authorize(authz::Action::ListChildren, &authz::FLEET).await?;
+
```
david-crespo, line 124: Noticed this check was missing here

PR #4853, opened 2024-01-19
[sled-agent] Encrypt a specific set of U.2 datasets

smklein:
This PR does the following:

- As a part of processing U.2s during initialization, `ensure_zpool_datasets_are_encrypted` is invoked. This identifies all datasets which should be encrypted (`cockroachdb`, `clickhouse`, `internal_dns`, `external_dns`, `clickhouse_keeper`) and performs a one-way migration from unencrypted to encrypted dataset.
- Additionally, during zone launching, the sled agent verifies properties about datasets that it expects should be encrypted. This helps prevent these encrypted dataset from being used before their transfer has finished, and also prevents these zones from ever using unencrypted datasets in the future.


sled-agent/src/services.rs
```diff
@@ -2930,6 +2948,35 @@ impl ServiceManager {
...
+                            prop_name: String::from(name),
+                            prop_value: actual,
+                            prop_value_expected: String::from(expected),
+                        });
+                    }
+                    return Ok(());
+                };
+                check_property("zoned", zoned, "on")?;
+                check_property("canmount", canmount, "on")?;
+                if dataset.dataset().dataset_should_be_encrypted() {
```
smklein, line 2976: Need this check because crucible's dataset is still unencrypted.

sled-storage/src/dataset.rs
```diff
@@ -364,6 +430,352 @@ pub(crate) async fn ensure_zpool_has_datasets(
...
+        find_all_unencrypted_datasets_directly_within_pool(&log, &zpool_name)
+            .await?;
+
+    // TODO: Could do this in parallel?
+    for dataset in unencrypted_datasets {
+        let log = &log.new(slog::o!("dataset" => dataset.clone()));
+        info!(log, "Found unencrypted dataset");
+
+        ensure_zpool_dataset_is_encrypted(&log, &zpool_name, &dataset).await?;
+    }
```
smklein, line 485: I'm not sure if we *actually* want to parallelize this, but it would be easy to do.

Seems like we should make disk addition parallel at least, so we can fire up all U.2s at once -- but we might get less benefit from putting extra workload on a U.2 already performing a migration. That's all speculative though.

andrewjstone: I'm going to start adding some concurrency support later today, or Monday at the latest. Got a touch delayed.

smklein: All good - I think it would be nice to have, at any time, but I don't think it needs to be a blocker here.

andrewjstone (approved):
This looks fantastic @smklein!

sled-storage/src/dataset.rs
```diff
@@ -364,6 +430,352 @@ pub(crate) async fn ensure_zpool_has_datasets(
...
+        let mut iter = line.split_whitespace();
+        let Some(dataset) = iter.next() else {
+            continue;
+        };
+        let log = log.new(slog::o!("dataset" => dataset.to_string()));
+
+        let Some(encryption) = iter.next() else {
+            continue;
+        };
+        let encrypted = encryption != "off";
```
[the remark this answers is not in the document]
smklein: From what I can tell, these are the valid options for encryption:

[source](https://illumos.org/man/8/zfs)
```
      encryption=on|off|aes-128-ccm|aes-192-ccm|aes-256-ccm|aes-128-gcm|aes-192-gcm|aes-256-gcm
       Controls the encryption cipher suite (block cipher, key length, and mode)
       used for this dataset.  Requires the encryption feature to be enabled on
       the pool.  Requires a keyformat to be set at dataset creation time.

       Selecting encryption=on when creating a dataset indicates that the
       default encryption suite will be selected, which is currently
       aes-256-ccm.  In order to provide consistent data protection, encryption
       must be specified at dataset creation time and it cannot be changed
       afterwards.
```

Regardless, I think I still want to be conservative here, for reasons outlined in: ed8bea7ed6e5647aa538e6d7104a61dd328e51fd

(But that should catch the "off" and "-" cases both as "I don't think this is encrypted").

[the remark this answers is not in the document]
andrewjstone: LGTM. Thank you!

Issue #4851, opened 2024-01-19
Publish per-instance vCPU usage information to `oximeter`


bnaecker:
Information about vCPU usage for each instance is maintained by the hypervisor in a set of `kstat`s:

- `vmm` is the module, with an integer instance that is unique per VMM.
- `vmm::vm:vm_name` lists the instance's UUID as its name. We may not need to collect this through `kstat`s since we already have it in the sled agent, but it exists.
- There are per-vCPU statistics which provide the cumulative time (in nanoseconds, through `gethrtime()`) in a number of microstates:
  - `time_emu_kern` tracks emulation performed in kernel space
  - `time_emu_user` tracks emulation performed in userspace
  - `time_idle` tracks idle time, such as when the guest is halted
  - `time_init` is time prior to the guest running at all
  - `time_run` is, well, when the guest is running
  - `time_sched` is time off-CPU, such as during for an interrupt or waiting for a lock

We should publish these to `oximeter`. It might make sense to track the state as an enum (string in ClickHouse), so that we can easily aggregate over it, or break out the usage by state. 

We should also include metadata about the instance itself, including at least the silo, project, and instance UUIDs.

gjcolombo:
A couple of thoughts: Do we want to gather these from sled agent or from Propolis itself? (*Can* we gather them from Propolis given that's running in a zone?) To me there's something Neat and Tidy about having VMMs produce statistics describing their own components, but I'm not sure there are any *technical* advantages to that, at least as far as kstat gathering is concerned. (Would implementing this in sled agent require that sled agent remember the number of CPUs each instance has? I'm not sure it actually remembers that today.)

bnaecker:
It looks like the zones are indeed able to get `kstat`s, at least through the command line tool. That suggests that Propolis could do the work too. One advantage to the sled-agent doing the work is that it already has to troll through `kstat`s for the physical data link stats (and presumably more in the future). But I don't feel strongly about it.

As far as the number of vCPUs, the sled-agent currently does accept that in the instance-creation request, and it looks like it stores it in the `InstanceInner.properties.vcpus` field. But even if it did not do that, the vCPU is indexed in the `kstat` name as well, such as `vmm:0:vcpu0:time_idle`, so we could pull it out of there if needed. I think it'd be better to use the instance properties to find the right kstats though.

bnaecker:
[...]

PR #4855, opened 2024-01-20
Publish instance vCPU usage statistics to `oximeter`

bnaecker:
- Adds the silo and project IDs to the instance-ensure request from Nexus to the sled-agent. These are used as fields on the instance-related statistics.
- Defines a `VirtualMachine` oximeter target and `VcpuUsage` metric. The latter has a `state` field which corresponds to the named kstats published by the hypervisor that accumulate the time spent in a number of vCPU microstates. The combination of these should allow us to aggregate or break down vCPU usage by silo, project, instance, vCPU ID, and CPU state.
- Adds APIs to the `MetricsManager` for starting / stopping tracking instance-related metrics, and plumbs the type through the `InstanceManager` and `Instance` (and their internal friends), so that new instances can control when data is produced from them. Currently, we'll start producing as soon as we get a non-terminate response from Propolis in the `instance_state_monitor()` task, and stop when the instance is terminated.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -168,7 +168,7 @@ dependencies = [
  "omicron-workspace-hack",
  "proc-macro2",
  "quote",
- "syn 2.0.46",
+ "syn 2.0.48",
 ]
 
 [[package]]
@@ -269,7 +269,7 @@ checksum = "5fd55a5ba1179988837d24ab4c7cc8ed6efdeff578ede0416b4225a5fca35bd0"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.46",
+ "syn 2.0.48",
 ]
 
 [[package]]
@@ -291,7 +291,7 @@ checksum = "16e62a023e7c117e27523144c5d2459f4397fcc3cab0085af8e2224f643a0193"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.46",
+ "syn 2.0.48",
 ]
 
 [[package]]
@@ -302,7 +302,7 @@ checksum = "c980ee35e870bd1a4d2c8294d4c04d0499e67bca1e4b5cefcc693c2fa00caea9"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.46",
+ "syn 2.0.48",
 ]
 
 [[package]]
@@ -353,7 +353,7 @@ dependencies = [
  "quote",
  "serde",
  "serde_tokenstream 0.2.0",
- "syn 2.0.46",
+ "syn 2.0.48",
 ]
 
 [[package]]
@@ -490,7 +490,7 @@ dependencies = [
  "regex",
  "rustc-hash",
  "shlex",
- "syn 2.0.46",
+ "syn 2.0.48",
  "which",
 ]
 
@@ -986,7 +986,7 @@ dependencies = [
  "heck 0.4.1",
  "proc-macro2",
  "quote",
- "syn 2.0.46",
+ "syn 2.0.48",
 ]
 
 [[package]]
@@ -1412,7 +1412,7 @@ checksum = "83fdaf97f4804dcebfa5862639bc9ce4121e82140bec2a987ac5140294865b5b"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.46",
+ "syn 2.0.48",
 ]
 
 [[package]]
@@ -1460,7 +1460,7 @@ dependencies = [
  "proc-macro2",
  "quote",
  "strsim 0.10.0",
- "syn 2.0.46",
+ "syn 2.0.48",
 ]
 
 [[package]]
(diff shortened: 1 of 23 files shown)


bnaecker:
This still needs more testing around the instance lifecycle itself, but here are some manual tests I ran to verify the new code for pulling kstats themselves from the `KstatSampler`.

Testing this is hard, since it requires an actual VM. Instead, I wrote a tiny
CLI tool to fake up the important parts. That is here:

```rust
use oximeter_instruments::kstat::KstatTarget;

fn main() {
    let ctl = kstat_rs::Ctl::new().unwrap();

    // Find a VMM kstat.
    //
    // We'll use this to cons up a `VirtualMachine` that matches, to test our
    // shit.
    let Some(id) = ctl.iter().find_map(|mut kstat| {
        if kstat.ks_module == "vmm" && kstat.ks_name == "vm" {
            let data = ctl.read(&mut kstat).unwrap();
            let kstat_rs::Data::Named(named) = data else {
                return None;
            };
            named.iter().find_map(|nv| {
                if nv.name == "vm_name" {
                    let kstat_rs::NamedData::String(vm_name) = nv.value else {
                        return None;
                    };
                    Some(vm_name.to_string())
                } else {
                    None
                }
            })
        } else {
            None
        }
    }) else
    {
        eprintln!("No VMM kstats found");
        return;
    };

    println!("Found VMM kstats for VM with name: {id}");
    let vm = oximeter_instruments::kstat::virtual_machine::VirtualMachine {
        silo_id: uuid::Uuid::new_v4(),
        project_id: uuid::Uuid::new_v4(),
        instance_id: id.parse().unwrap(),
    };

    // Filter the kstats again to those in the `vmm` module, so that we can call
    // `to_samples()` as it normally will be
    let now = chrono::Utc::now();
    let kstats: Vec<_> = ctl.iter().filter(|kstat| kstat.ks_module == "vmm").map(|mut kstat| {
        (now, kstat, ctl.read(&mut kstat).unwrap())
    })
    .collect();

    let samples = vm.to_samples(&kstats).unwrap();
    println!("produced samples:\n{samples:#?}");
}
```
So it finds a VM name first, so that we can construct a `VirtualMachine` target.
It then filters the list of kstats, which is normally the job of the
`KstatSampler`. This is all so we can test the `to_samples()` method, which
converts the kstats into the `VcpuUsage` metric samples we care about. There is
one per (vCPU, state). Here is some of the output of that CLI run on the Gimlet
in cubby 10 of the dogfood rack:

```
 BRM42220009 # /tmp/vcpu | head -60
Found VMM kstats for VM with name: 0fe5243a-ae48-44f3-8a42-3e55c8ea1d40produced samples:
[
    Sample {
        measurement: Measurement {
            timestamp: 2024-01-19T22:53:30.652303313Z,
            datum: CumulativeU64(
                Cumulative {
                    start_time: 2024-01-19T22:53:34.520006783Z,
                    value: 595956991,
                },
            ),
        },
        timeseries_name: "virtual_machine:vcpu_usage",
        target: FieldSet {
            name: "virtual_machine",
            fields: {
                "instance_id": Field {
                    name: "instance_id",
                    value: Uuid(
                        0fe5243a-ae48-44f3-8a42-3e55c8ea1d40,
                    ),
                },
                "project_id": Field {
                    name: "project_id",
                    value: Uuid(
                        e23d5dbd-460a-40a8-b01d-3a914ce322d5,
                    ),
                },
                "silo_id": Field {
                    name: "silo_id",
                    value: Uuid(
                        ddc3dcc7-39c2-4200-8d4c-be3803f9994c,
                    ),
                },
            },
        },
        metric: FieldSet {
            name: "vcpu_usage",
            fields: {
                "state": Field {
                    name: "state",
                    value: String(
                        "init",
                    ),
                },
                "vcpu_id": Field {
                    name: "vcpu_id",
                    value: U32(
                        0,
                    ),
                },
            },
        },
    },
    Sample {
        measurement: Measurement {
            timestamp: 2024-01-19T22:53:30.652303313Z,
            datum: CumulativeU64(
                Cumulative {
                    start_time: 2024-01-19T22:53:34.520006783Z,
$
```
So that measured the cumulative time in the `time_init` microstate as
`595956991`, which is what `kstat` itself reports:

```
BRM42220009 # kstat vmm:1:vm:vm_name
module: vmm                             instance: 1
name:   vm                              class:    misc
        vm_name                         0fe5243a-ae48-44f3-8a42-3e55c8ea1d40

BRM42220009 # kstat vmm:1:vcpu0:time_init
module: vmm                             instance: 1
name:   vcpu0                           class:    misc
        time_init                       595956991

```

So this all seems to work OK. I'm going to do more testing locally starting and stopping a bunch of VMs to check that the statistics continue to work and that they show up in ClickHouse as I expect.

bnaecker:
Fixes #4851

bnaecker:
I'm going to close this, and instead re-open a slightly smaller PR just for the first chunk of work listed in #4851. That will add the `oximeter` metric definitions and include the instance metadata in the sled-agent HTTP API.

PR #4856, opened 2024-01-20
Concurrent disk creation

andrewjstone:
Make any sequential for loops for U.2 disk additions concurrent inside the storage manager. Now that we have some data migration going on during disk creation, we don't want to wait for each of the 10 disks in a sled to make progress before moving on. 

I believe the existing tests cover these changes well, as the external behavior of queuing and adding should not have changed.

### DIFF
--- a/sled-storage/src/manager.rs
+++ b/sled-storage/src/manager.rs
@@ -11,6 +11,8 @@ use crate::disk::{Disk, DiskError, RawDisk};
 use crate::error::Error;
 use crate::resources::{AddDiskResult, StorageResources};
 use camino::Utf8PathBuf;
+use futures::stream::FuturesUnordered;
+use futures::StreamExt;
 use illumos_utils::zfs::{Mountpoint, Zfs};
 use illumos_utils::zpool::ZpoolName;
 use key_manager::StorageKeyRequester;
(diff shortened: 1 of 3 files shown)


sled-storage/src/manager.rs
```diff
@@ -398,35 +400,75 @@ impl StorageManager {
...
-                        to_dequeue.insert(disk);
+                        self.queued_u2_drives.remove(&raw_disk);
                     }
                     Ok(AddDiskResult::DiskAlreadyInserted) => {
-                        to_dequeue.insert(disk);
+                        self.queued_u2_drives.remove(&raw_disk);
                     }
-                    Ok(AddDiskResult::DiskQueued) => (),
+                    Ok(AddDiskResult::DiskQueued) => {
+                        error!(
```
andrewjstone, line 444: Should we panic here? If for some reason this error code came back, we probably should set `self.state = StorageManagerState::QueueingDisks`

sled-storage/src/manager.rs
```diff
@@ -539,21 +581,37 @@ impl StorageManager {
...
-            match self.add_disk(raw_disk).await {
-                Ok(AddDiskResult::DiskInserted) => should_update = true,
-                Ok(_) => (),
-                Err(err) => {
-                    warn!(
-                        self.log,
-                        "Failed to add disk to storage resources: {err}";
-                        "disk_id" => ?disk_id
-                    );
+            match raw_disk.variant() {
```
smklein, line 591: Just to confirm - have you tested an "unencrypted" -> "encrypted" migration with this PR? I unfortunately didn't really have automated tests for that process

sled-storage/src/manager.rs
smklein: I don't really like the idea of panicking here, but I'm not sure I understand all these pathways.

This codepath looks pretty redundant with the call to `add_u2_disk`, which calls `Disk::new` and also sets the `QueueingDisks` state when we match on an underlying error from `self.resources.insert_disk`.

Is there a reason we aren't using that method anymore? I'm kinda nervous by so much state-setting and checking from different possible pathways. Specifically: If I were to edit this codebase, I'm really unclear "who is responsible for setting `self.state`, caller, callee, or my current function".

sled-storage/src/manager.rs
```diff
@@ -539,21 +581,37 @@ impl StorageManager {
             }
         }
 
+        // We want to add all U.2 disks concurrently rather than sequentially.
```
smklein, line 584: IMO it's a little arbitrary to be doing this for U.2s, and not M.2s, right?

I think I see that we made this choice because these functions act on `&mut self`? (If I'm wrong, please correct me) and that would make parallelized calls not possible

But I wonder if this is worth a slight refactor, to split:

- `add_disk`, which is `&mut self`, and currently:
  - `Disk::new` +
  - Update "self" (either via `self.resources`, queues, state, etc)

Into something like:

- (&mut self) "Check we're ready for this disk, or enqueue it for later if not"
- (&self) "Actually call `Disk::new`"
- (&mut self) "Update state" (call `insert_disks`, update `self.queued_drives`, `self.state`, etc)

This would let us keep the calls to `Disk::new` parallel, while avoiding some of the splayed state management

smklein (commented):
Looks good, but a couple questions below about refactors to maybe make state management more clear

andrewjstone:
This is pretty out of date, and it's unlikely to ever get merged.

PR #4857, opened 2024-01-20
background task for service zone nat

### LINKED ISSUE #4650: Service NAT entries missing after dendrite restart

internet-diglett:
Currently the logic for configuring NAT for service zones is deeply nested and crosses sled-agent http API boundaries. The cleanest way to deliver eventual consistency for service zone nat entries was to pull the zone information from inventory and use that to generate nat entries to reconcile against the `ipv4_nat_entry` table. This covers us in the following scenarios:

### RSS:
* User provides configuration to RSS
* RSS process ultimately creates a sled plan and service plan
* Application of service plan by sled-agents creates zones
* zone create makes direct calls to dendrite to configure NAT (it is the only way it can be done at this time)
* eventually the Nexus zones are launched and handoff to Nexus is complete
* inventory task is run, recording zone locations to db
* service zone nat background task reads inventory from db and uses the data to generate records for `ipv4_nat_entry` table, then triggers dendrite sync.
* sync is ultimately a noop because nat entries already exist in dendrite (dendrite operations are idempotent)
 
### Cold boot:
* sled-agents create switch zones if they are managing a scrimlet, and subsequently create zones written to their ledgers. This may result in direct calls to dendrite.
* Once nexus is back up, inventory will resume being collected
* service zone nat background task will read inventory from db to reconcile entries in `ipv4_nat_entry` table and then trigger dendrite sync.
* If nat is out of date on dendrite, it will be updated on trigger.

### Dendrite crash
* If dendrite crashes and restarts, it will immediately contact Nexus for re-sync (pre-existing logic from earlier NAT RPW work)
* service zone and instance nat entries are now present in rpw table, so all nat entries will be restored

### Migration / Relocation of service zone
* New zone gets created on a sled in the rack. Direct call to dendrite will be made (it uses the same logic as pre-nexus to create zone).
* Inventory task will record new location of service zone
* Service zone nat background task will use inventory to update table, adding and removing the necessary nat entries and triggering a dendrite update

Considerations
---
Because this relies on data from the inventory task which runs on a periodic timer (600s), and because this task also runs on a periodic timer (30s), there may be some latency for picking up changes. A few potential avenues for improvement:

* Plumb additional logic into service zone nat configuration that enables direct updates to the `ipv4_nat_entry` table once nexus is online. Of note, this would further bifurcate the logic of pre-nexus and post-nexus state management. At this moment, it seems that this is the most painful approach. An argument can be made that we ultimately should be lifting the nat configuration logic _out_ of the service zone creation instead.

* Decrease the timer for the inventory task. This is the simplest change, however this would result in more frequent collection, increasing overhead. I do not know _how much_ this would increase overhead. Maybe it is negligible.

* Plumb in the ability to trigger the inventory collection task for interesting control plane events. This would allow us to keep the _relatively_ infrequent timing intervals but allow us to refresh on-demand when needed.

Related
---
Closes #4650 
Extracted from #4822


nexus/src/app/background/sync_service_zone_nat.rs
```diff
@@ -0,0 +1,233 @@
...
+                    ipnetwork::Ipv6Network::new(*sled.ip, 128).unwrap(),
+                );
+
+                let zones_config: sled_agent_client::types::OmicronZonesConfig =
+                    zones_found.zones;
+                let zones: Vec<sled_agent_client::types::OmicronZoneConfig> =
+                    zones_config.zones;
+                for zone in zones {
+                    let zone_type: OmicronZoneType = zone.zone_type;
+                    match zone_type {
```
rcgoodfellow, line 122: ExternalDns is missing from the cases being handled below.

rcgoodfellow (commented):
Thanks Levon!

The approach for using inventory as the source of truth for reconciling the service entries seems solid.

I took this for a spin in `a4x2`. After the system came up, I rebooted `scrimlet0`. When it came back up most of the NAT entries came back with it, and I was able to hit the Nexus API successfully without any manual prodding 🎉 . The NAT entry for external DNS was missing however. I noted where I believe the issue is in the code review.

nexus/db-queries/src/db/datastore/ip_pool.rs
```diff
@@ -185,6 +185,9 @@ impl DataStore {
         // join ip_pool to ip_pool_resource and filter
 
         // used in both success and error outcomes
+        // TODO: Correctness
```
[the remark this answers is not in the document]
internet-diglett: This is an issue @zephraph and I ran into while troubleshooting an error during instance creation. When this lookup is attempted and fails, the logs will show "Default pool for current silo" instead of the id used to attempt the lookup. I'm not quite sure how this is _supposed_ to work, but I think it's safe to say that the error message should at least indicate the name or uuid used to attempt the lookup.

nexus/src/app/background/sync_service_zone_nat.rs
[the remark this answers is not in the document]
internet-diglett: I considered matching against every zone explicitly instead of using `_` to provide some sort of signal in that situation, I guess I'll implement it that way so they at least get a compiler error.

internet-diglett: Good catch!

nexus/src/app/background/sync_service_zone_nat.rs
```diff
@@ -0,0 +1,233 @@
...
+    pub fn new(
+        datastore: Arc<DataStore>,
+        dpd_clients: Vec<Arc<dpd_client::Client>>,
+    ) -> Self {
+        Self { datastore, dpd_clients }
+    }
+}
+
+impl BackgroundTask for ServiceZoneNatTracker {
+    fn activate<'a, 'b, 'c>(
```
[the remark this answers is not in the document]
internet-diglett: These lifetimes are required by the `BackgroundTask` trait in `nexus/src/app/background/common.rs`. My rough guess is that the returned value may have some dependency on data in `self` or `OpContext` in certain background tasks.

nexus/src/app/background/sync_service_zone_nat.rs
```diff
@@ -0,0 +1,233 @@
...
+                            // Append ipv4 nat entry
+                            ipv4_nat_values.push(nat_value);
+                        }
+
+                        _ => continue,
+                    }
+                }
+            }
+
+            if ipv4_nat_values.is_empty() {
```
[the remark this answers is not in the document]
internet-diglett: If I understand things correctly, as of today this should never be empty because Nexus must be running in order to execute the background task. If Nexus is running, we should have some entries here.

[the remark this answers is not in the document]
internet-diglett: I probably need to wordsmith the error message to indicate this more clearly

nexus/src/app/background/init.rs
```diff
@@ -163,6 +171,23 @@ impl BackgroundTasks {
             task
         };
 
+        let task_service_zone_nat_tracker = {
+            driver.register(
+                "service_zone_nat_tracker".to_string(),
+                String::from(
+                    "ensures service zone nat records are recorded in NAT RPW table",
+                ),
+                // TODO create separate config for this task
```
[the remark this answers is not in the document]
internet-diglett: I believe I have actually done this and just failed to come back and update this call 😂

nexus/src/app/background/sync_service_zone_nat.rs
internet-diglett: Fixed

dns zone vnics
---
```
sled03 ➜  omicron git:(rpw-for-service-zone-nat) ✗ pfexec zlogin oxz_external_dns_4fbac8b1-8242-4907-b720-7796c34ecbb3 dladm show-vnic
LINK         OVER         SPEED    MACADDRESS        MACADDRTYPE         VID
vopte4       ?            0        a8:40:25:ff:f7:a9 fixed               0
oxControlService28 ?      0        2:8:20:83:ff:75   random              0
sled03 ➜  omicron git:(rpw-for-service-zone-nat) ✗ pfexec zlogin oxz_external_dns_9ea8ca5a-cb91-40a4-8f10-838424abc1ba dladm show-vnic
LINK         OVER         SPEED    MACADDRESS        MACADDRTYPE         VID
vopte2       ?            0        a8:40:25:ff:c4:3d fixed               0
oxControlService25 ?      0        2:8:20:52:c1:2a   random              0
```

nat entries persisting after restart of dendrite
---
```
root@oxz_switch:~# swadm nat list
External IP  Port low  Port high  Internal IP            Inner MAC          VNI
10.85.0.20   0         65535      fd00:1122:3344:101::1  a8:40:25:ff:c4:3d  100
10.85.0.21   0         65535      fd00:1122:3344:101::1  a8:40:25:ff:f7:a9  100
10.85.0.22   0         65535      fd00:1122:3344:101::1  a8:40:25:ff:9b:eb  100
10.85.0.23   0         65535      fd00:1122:3344:101::1  a8:40:25:ff:86:cd  100
10.85.0.24   0         65535      fd00:1122:3344:101::1  a8:40:25:ff:8f:0b  100
10.85.0.25   0         16383      fd00:1122:3344:101::1  a8:40:25:ff:f6:79  100

root@oxz_switch:~# svcadm restart dendrite
root@oxz_switch:~# swadm nat list
External IP  Port low  Port high  Internal IP            Inner MAC          VNI
10.85.0.20   0         65535      fd00:1122:3344:101::1  a8:40:25:ff:c4:3d  100
10.85.0.21   0         65535      fd00:1122:3344:101::1  a8:40:25:ff:f7:a9  100
10.85.0.22   0         65535      fd00:1122:3344:101::1  a8:40:25:ff:9b:eb  100
10.85.0.23   0         65535      fd00:1122:3344:101::1  a8:40:25:ff:86:cd  100
10.85.0.24   0         65535      fd00:1122:3344:101::1  a8:40:25:ff:8f:0b  100
10.85.0.25   0         16383      fd00:1122:3344:101::1  a8:40:25:ff:f6:79  100
```

rcgoodfellow: Thanks. I've retested this in `a4x2` and all the NAT entries came back up. I even rebooted both switches (one at a time) and connectivity came back up on both just fine.

nexus/src/app/background/sync_service_zone_nat.rs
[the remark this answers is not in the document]
rcgoodfellow: Ok yeah, so the lifetimes come directly from the trait definition, so I guess the rationale would go there.

rcgoodfellow (approved):
LGTM, thanks Levon! Let's just get this rebased on main and run through another round of testing before we merge.

nexus/src/app/background/sync_service_zone_nat.rs
```diff
@@ -0,0 +1,315 @@
...
+                    &log,
+                    "nexus is running but no service zone nat values could be generated from inventory";
+                );
+                return json!({
+                    "error": "nexus is running but no service zone nat values could be generated from inventory"
+                });
+            }
+
+            // reconcile service zone nat entries
+            let result = match self.datastore.ipv4_nat_sync_service_zones(opctx, &ipv4_nat_values).await {
```
smklein, line 269: We discussed this in chat, but this whole flow has some implications for our calls to `ensure_nat_entry` / `ensure_ipv4_nat_entry`:

- If you're writing code in Nexus to make an **instance** with a NAT entry, you must call these functions explicitly. No one else will add your NAT entry to the DB!
- If you're writing code in Nexus to make a **service** with a NAT entry, you **should not call these functions**, and should rely on the "Nexus -> Sled Agent -> Inventory -> Service Zone NAT RPW" pathway" to ensure that these entries get populated. If you *tried* to add the entry to the DB, you'd risk a race condition between "collection via inventory / RPW" and "explicitly inserting the record".

I think that's okay, but we should add documentation around the `ensure_nat_entry` function to make that distinction extremely clear! I think it's kinda subtle that the same table is treated quite differently depending on the purpose of the NAT entry.

smklein: Also, for posterity, I created this drawing to map out the data flow here.

My "TLDR" of the above is that I wanted to ensure we avoided having loops in this data flow graph:

image

internet-diglett: I think this is great feedback, adding some documentation to it now. I think one day we may get to a place where we can lift the NAT logic out of the service zone creation functionality in sled-agent, which could help us move towards a more consistent pattern of interacting with this table.

nexus/db-queries/src/db/datastore/ip_pool.rs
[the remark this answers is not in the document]
zephraph: cc @david-crespo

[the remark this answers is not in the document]
david-crespo: I think Ryan is correct that it's not a correctness issue. But it should be fixed, thanks for flagging it as a source of pain. I see #4864. I'm still working on IP pools stuff this week so I'll roll in an improvement.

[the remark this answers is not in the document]
internet-diglett: @david-crespo After reading the comment you left in #4864 I understand things better now, many thanks!

internet-diglett:
I've added a stopgap that was discussed in chat, we are now checking to ensure that we have a minimum number of each service zone type that we're interested in before reconciling the entries.

nexus/src/app/background/sync_service_zone_nat.rs
```diff
@@ -21,6 +21,18 @@ use std::net::{IpAddr, SocketAddr};
 use std::num::NonZeroU32;
 use std::sync::Arc;
 
+// Minumum number of boundary NTP zones that should be present in a valid
+// set of service zone nat configurations.
+const MIN_NTP_COUNT: usize = 1;
+
+// Minumum number of nexus zones that should be present in a valid
+// set of service zone nat configurations.
+const MIN_NEXUS_COUNT: usize = 3;
```
smklein, line 30: I'm not sure we should do this -- it's possible that one of the three Nexus instances have gone away, and we'd still want our NAT entries to be up-to-date.

This is a reasonable goal to try to achieve in the "graceful shutdown" case -- if we want three Nexus instances to run, we should provision a 4th one before removing one of the original 3 -- but if a sled is yanked from the rack, we **do** have two Nexus instances. That's just a truth! We should aspire to have the blueprint creator create a new Nexus service, provision it, and get the NAT entries populated, but it's **very possible** to run under our redundancy expectations. That's why we use redundancy!

nexus/src/app/background/sync_service_zone_nat.rs
```diff
@@ -265,6 +284,38 @@ impl BackgroundTask for ServiceZoneNatTracker {
...
+                    &log,
+                    "generated config for fewer than the minimum allowed number of ntp zones";
+                );
+                return json!({
+                    "error": "generated config for fewer than the minimum allowed number of ntp zones"
+
+                });
+            }
+
+            if nexus_count < MIN_NEXUS_COUNT {
```
smklein, line 308: Example where this could go awry:

- Sled pulled (or crashes, or catches fire) we mark it as "Removed"
- We identify that the services on the sled are not present
- Inventory reports that two Nexus instances are running
- 2 < 3
- So we fail the RPW here, and never update **any** NAT entries until redundancy is fully restored?

internet-diglett: Yes, that is correct. Will we not attempt to move the service zone to a new sled in this situation?

smklein: As soon as we implement service re-provisioning, this would happen, but I think the scope of "new service provisions" is going to start with only Crucible and NTP.

Until that is fully implemented, this check would just stop NAT propagation

nexus/src/app/background/sync_service_zone_nat.rs
internet-diglett: Would setting all of the minimums to `1` be a reasonable compromise? If we don't have at least 1 `Nexus`, `NTP`, and `ExternalDns` zone that can be found in inventory, I would think we have more serious problems, no?

smklein: Chatting with some folks in the update sync today, it sounds like the inventory system -- which gathers the inventory in a "collection", which contains a lot of objects -- may, at any time, simply "not report a sled" within that collection. Theoretically, that means we could see an "inventory collection" that doesn't contain any sleds which contain Nexus, NTP, external DNS, etc.

This has some weird implications for depending on the inventory system as the source-of-truth here, but without a full implementation of blueprints, I acknowledge that there isn't a great alternative yet.

So: "Could we set all the minimums to 1?" That would stop us from propagating the state of the inventory system if we saw a blip that eliminated all of these critical zones. So in that sense, it's arguably better than not doing the check! I also think it avoids "breaking" NAT when we're under-provisioned, as I mentioned in the case below.

If we're on the same page that, eventually, the right source-of-truth is "info from the blueprint, somehow", I think this is a reasonable intermediate step.

smklein: (also, I appreciate you tolerating all this churn. Getting RPWs + NAT propagation right is tricky, and having this portion of the system be "not-totally-ready" does make this extra hairy. Thank you for pushing through regardless)

internet-diglett: No worries! I know this is _critical_ so I appreciate the extra eyes and feedback!

rcgoodfellow:
I tested this after d3501dc, and things look much better. I restarted both scrimlets one after the other, and after each restart, all the NAT entries came back on the scrimlet that had been restarted.

PR #4858, opened 2024-01-20
[update-engine] reuse parent_key_and_child_index from existing steps

sunshowers:
During a dogfood mupdate on 2024-01-18, I saw some really weird output with
wicket's `rack-update attach` that looked like:

```
[sled 8 00:20:48]   Running .... 12a 5s 1/3) Writing host phase 2 to slot B
```

The "5s" is all wrong -- the letter there is supposed to indicate, for an
engine that has one or more nested engines, the index of that nested engine. So
for example, if a step 12 has two nested engines, they would be marked "12a"
and "12b".

"5s" indicates that that's the 19th nested engine for that step, and we
definitely have nowhere near 19 nested engines for a step anywhere in wicketd.

This turned out to be because we weren't reusing child indexes from earlier
steps in the sequence. Fix that, and also add:

* tests which catch this issue
* some dev-only code to wicket which made it easy to debug this locally

### DIFF
--- a/wicket/src/cli/rack_update.rs
+++ b/wicket/src/cli/rack_update.rs
@@ -8,23 +8,29 @@
 
 use std::{
     collections::{BTreeMap, BTreeSet},
+    io::{BufReader, Write},
     net::SocketAddrV6,
     time::Duration,
 };
 
 use anyhow::{anyhow, bail, Context, Result};
+use camino::Utf8PathBuf;
 use clap::{Args, Subcommand, ValueEnum};
 use slog::Logger;
 use tokio::{sync::watch, task::JoinHandle};
 use update_engine::{
     display::{GroupDisplay, LineDisplayStyles},
-    NestedError,
+    EventBuffer, NestedError,
 };
 use wicket_common::{
-    rack_update::ClearUpdateStateResponse, update_events::EventReport,
+    rack_update::ClearUpdateStateResponse,
+    update_events::{EventReport, WicketdEngineSpec},
     WICKETD_TIMEOUT,
 };
-use wicketd_client::types::{ClearUpdateStateParams, StartUpdateParams};
+use wicketd_client::types::{
+    ClearUpdateStateParams, GetArtifactsAndEventReportsResponse,
+    StartUpdateParams,
+};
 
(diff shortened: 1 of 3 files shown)


wicket/src/cli/rack_update.rs
```diff
@@ -380,6 +404,155 @@ async fn do_clear_update_state(
...
+
+                        // If display.add_event_report errors out, it's for a report for a
+                        // component we weren't interested in. Ignore it.
+                        _ = display.add_event_report(&id, report);
+
+                        display.write_events()?;
+                    }
+                }
+            }
+            ReplayStrategy::Idempotent => {
```
sunshowers, line 535: This is the best name I could come up with for this strategy -- would love ideas for better names.

wicket/src/cli/rack_update.rs
```diff
@@ -41,10 +47,22 @@ use super::command::CommandOutput;
...
+
     /// Attach to one or more running updates.
     Attach(AttachArgs),
+
     /// Clear updates.
     Clear(ClearArgs),
+
+    /// Dump artifacts and event reports from wicketd.
+    ///
+    /// Debug-only, intended for development.
```
andrewjstone, line 59: Nice :)

Do you have any concerns about this being available in production?

wicket/src/cli/rack_update.rs
andrewjstone: `Idempotent` seems fine to me.  The only alternative I could think of is `Full` or `Complete`.

andrewjstone (approved):

wicket/src/cli/rack_update.rs
sunshowers: No not really. There's nothing here that isn't in the UI, and this is clearly marked as a "debug" interface.

wicket/src/cli/rack_update.rs
sunshowers: I'd considered "full", but I think it isn't clearly differentiated from "oneshot" -- "idempotent" is slightly weird at first but makes sense when you see what it's doing.

Issue #4740, opened 2024-01-01
Use versioned progenitor dependency


rcgoodfellow:
Currently, we import progenitor in `Cargo.toml` using the progenitor repository main branch and depend on the Omicron lockfile for tracking the revision we're using. This can cause downstream issues like [this](https://github.com/oxidecomputer/maghemite/runs/20068283477) when updating lockfiles that have dependencies in Omicron.

I tried this simple change, as we're on a commit between the 4.1 and 5.0 releases.

```diff
diff --git a/Cargo.toml b/Cargo.toml
index f7256ce8b..64d1c0908 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -291,8 +291,8 @@ pretty_assertions = "1.4.0"
 pretty-hex = "0.4.0"
 prettyplease = "0.2.15"
 proc-macro2 = "1.0"
-progenitor = { git = "https://github.com/oxidecomputer/progenitor", branch = "main" }
-progenitor-client = { git = "https://github.com/oxidecomputer/progenitor", branch = "main" }
+progenitor = "0.4"
+progenitor-client = "0.4"
 bhyve_api = { git = "https://github.com/oxidecomputer/propolis", rev = "f1571ce141421cff3d3328f43e7722f5df96fdda" }
 propolis-client = { git = "https://github.com/oxidecomputer/propolis", rev = "f1571ce141421cff3d3328f43e7722f5df96fdda" }
 propolis-mock-server = { git = "https://github.com/oxidecomputer/propolis", rev = "f1571ce141421cff3d3328f43e7722f5df96fdda" }
```

However, that resulted in a bunch of errors, as it seems we depend on features on the main branch after the 4.1 release but before 5.0.

ahl:
We hadn't previously used the crates.io version because progenitor was evolving quickly and publishing a new revision for every turn of the crank was more of a burden. The rate of change has slowed (read: it basically works), so--sure, I think that would be reasonable for omicron and its satellite repos such as crucible and propolis.

sunshowers:
We use progenitor from crates.io now -- ok to close this issue?

rcgoodfellow:
yep

PR #4874, opened 2024-01-23
Update progenitor from v0.4.0 -> v0.5.0

bnaecker:


pfmooney:
This would definitely be nice to get squared away, as things pulling in Omicron and Progenitor as deps need to manually downgrade the latter to 0.4.x in order to avoid compiler errors due to the change in types.

bnaecker:
I agree! I've got a WIP in Propolis for publishing the vCPU kstats which necessitated the update anyway. I've got a few more pieces to tidy up on those WIPs before I tackle this, but I will definitely get to it this afternoon.

common/src/api/external/error.rs
```diff
@@ -523,6 +523,7 @@ impl<T: ClientError> From<progenitor::progenitor_client::Error<T>> for Error {
             // our version constraints (i.e. that the call was to a newer
             // service with an incompatible response).
             progenitor::progenitor_client::Error::InvalidResponsePayload(
+                _bytes,
                 ee,
             ) => Error::internal_error(&format!(
                 "InvalidResponsePayload: {}",
```
ahl, line 529: I think print out the bytes

ahl: do we know why we're not just using the `Display` impl on error? these all seem close to
https://github.com/oxidecomputer/progenitor/blob/main/progenitor-client/src/progenitor_client.rs#L307-L337

nexus/src/app/sagas/switch_port_settings_apply.rs
```diff
@@ -223,7 +223,7 @@ async fn spa_undo_ensure_switch_port_settings(
     let log = sagactx.user_data().log();
 
     let port_id: PortId = PortId::from_str(&params.switch_port_name)
-        .map_err(|e| external::Error::internal_error(e))?;
+        .map_err(|e| external::Error::internal_error(e.to_string().as_str()))?;
```
ahl, line 226: is this the typify change to replace the `Result<T, String>` with a real error type?

nexus/src/app/sagas/switch_port_settings_clear.rs
```diff
@@ -179,7 +179,7 @@ async fn spa_undo_clear_switch_port_settings(
     let log = sagactx.user_data().log();
 
     let port_id: PortId = PortId::from_str(&params.port_name)
-        .map_err(|e| external::Error::internal_error(e))?;
+        .map_err(|e| external::Error::internal_error(e.to_string().as_str()))?;
```
ahl, line 182: maybe change `internal_error`?
image

Maybe these should accept `impl ToString` ? or `AsRef<str>`

sled-agent/src/sim/http_entrypoints_pantry.rs
```diff
@@ -365,6 +365,16 @@ mod tests {
                     );
                 };
                 for (key, value) in map.iter() {
+                    // We intentionally skip the "description" key, provided
+                    // that the value is also a true String. This is mostly a
+                    // one-off for the udpate to Progenitor 0.5.0, which caused
+                    // this key to be added. But it's also pretty harmless,
+                    // since it's not possible to get this key-value combination
+                    // in a real JSON schema.
+                    if key == "description" && matches!(value, Value::String(_))
```
ahl, line 374:
```suggestion
                    if key == "description" && value.is_string())
```

perhaps?

common/src/api/external/error.rs
bnaecker: I don't know the answer to your question, no. I mostly did it for compatibility with the other arms, though it does look exactly like the display impl.

> I think print out the bytes

With `{:?}` formatting or something? I worry about size or contents, since IIUC this can be literally anything, right?

nexus/src/app/sagas/switch_port_settings_apply.rs
bnaecker: The `e` here seems to have type `ConversionError`, so I think that's right. Is simply stringifying it the best approach?

nexus/src/app/sagas/switch_port_settings_clear.rs
bnaecker: I agree that would be a good change, I've found myself often calling `to_string()` only to throw that away immediately, where the internal error variant then allocates a new one. I would prefer to do it in another PR though, since this variant is used _extensively_.

sled-agent/src/sim/http_entrypoints_pantry.rs
bnaecker: Oh neat! TIL, thanks.

nexus/src/app/sagas/switch_port_settings_clear.rs
bnaecker: https://github.com/oxidecomputer/omicron/issues/4892

common/src/api/external/error.rs
ahl: This error would be a pretty serious problem. It would mean that we got back data that didn't match the format we expected i.e. we expected a response body of a particular type but the deserialization failed. As the comment notes, this would be reflective of a service incompatibility that would represent a pretty serious violation of an important constraint (namely that services are able to talk to each other if they're deployed together).

Yes, the output would be a pile of bytes, but if we hit this I think we'll want to know what that pile of bytes was!

nexus/src/app/sagas/switch_port_settings_apply.rs
ahl: Yeah I think the error is really just a Cow<'static, str>

nexus/src/app/sagas/switch_port_settings_clear.rs
ahl: Sounds good

ahl:
meta note: we're still using the git dependency so the fact that it's 0.4.0 or 0.5.0 is less important than getting to a more recent commit from main...

common/src/api/external/error.rs
bnaecker: Makes sense, I'll add that to the output as well.

ahl (approved):

rcgoodfellow:
Is there a reason we can't use the numbered versions on crates.io? I logged a ticket on this a few weeks back when I had a bit of breakage downstream.

- #4740

PR #4876, opened 2024-01-24
[wicket] update to ratatui 0.25.0

sunshowers:
This was mostly a smooth upgrade, with just a small change required to our use
of tui-tree-widget.

Tested the TUI locally -- works great.


andrewjstone (approved):
Thanks @sunshowers

PR #4878, opened 2024-01-24
[meta] update samael to 0.0.14

sunshowers:
Required to unblock a bunch of other updates.

The behavior of a test changed, but in a way that to my understanding based on [the Duo article](https://duo.com/blog/duo-finds-saml-vulnerabilities-affecting-multiple-implementations) is still safe. See the comment included in the PR for more.


sunshowers:
Looks like a genuine regression -- the new version of samael appears to no longer be rejecting SAML responses with embedded comments.

sunshowers:
Looks like it was a behavior change and the new behavior is fine.

jmpesp (approved):
That test comment is excellent :)

PR #814, opened 2022-03-25
Introduce concept of Silos

jmpesp:
Add silos, which will isolate organizations, and provide a namespace for
users and groups.

This required adding Silo id to Actor, so users that have authenticated
now have an associated Silo id that can be used to restrict organization
lookup.

Silos can be created, read, and deleted. Modification is a TODO. Silos
can be marked discoverable or not, to support a tenancy model where
users and resources are strongly isolated from each other.

A few tests have been modified to use authn_as because an earlier
version of this branch added OpContext to every endpoint, but that was
reverted because the blast radius of the PR would have been too large.
What remains are a few modified tests that make authenticated calls.

When all endpoints are protected and each datastore function has an
OpContext, Silo can be looked up on Actor. For now, there are places
hard coding as the built-in Silo.

Still TODO:
- authz for silos and silo users
  - some testing is dependent on ^
- PUT /silos/{name}
- building on top of silos

PR #4880, opened 2024-01-24
Better error message when there's no default IP pool

### LINKED ISSUE #4864: More detail in default ip-pool lookup failures
Recently when troubleshooting the cause of an instance failing to deploy, I saw that I was hitting an error that showed the `ip-pool` could not be found. I was looking the the logs to see which `ip-pool`, I only saw "Default pool for current silo" in the log message. Should we also provide the UUID or name being used in the query, so we actually know _what_ failed to be found?

https://github.com/oxidecomputer/omicron/blob/c794b1a33f6fbb500723b8c3ced9595b390f66a7/nexus/db-queries/src/db/datastore/ip_pool.rs#L187-L193

david-crespo:
Closes #4864 

This is a bad error message to get when the problem is that there is no default IP pool configured for your current silo:

```
not found: ip-pool with id "Default pool for current silo"
```

"Default pool for current silo" is not an id, so why would we call it one? This is better:

```
not found: default IP pool for current silo
```

This PR is just making that possible.

### DIFF
--- a/common/src/api/external/error.rs
+++ b/common/src/api/external/error.rs
@@ -144,11 +144,11 @@ pub enum LookupType {
     ByName(String),
     /// a specific id was requested
     ById(Uuid),
-    /// a session token was requested
-    BySessionToken(String),
     /// a specific id was requested with some composite type
     /// (caller summarizes it)
     ByCompositeId(String),
+    /// object selected by criteria that would be confusing to call an ID
+    ByOther(String),
 }
 
 impl LookupType {
@@ -359,23 +359,22 @@ impl From<Error> for HttpError {
     fn from(error: Error) -> HttpError {
         match error {
             Error::ObjectNotFound { type_name: t, lookup_type: lt } => {
-                // TODO-cleanup is there a better way to express this?
-                let (lookup_field, lookup_value) = match lt {
-                    LookupType::ByName(name) => ("name", name),
-                    LookupType::ById(id) => ("id", id.to_string()),
-                    LookupType::ByCompositeId(label) => ("id", label),
-                    LookupType::BySessionToken(token) => {
-                        ("session token", token)
+                let message = match lt {
+                    LookupType::ByName(name) => {
+                        format!("{} with name \"{}\"", t, name)
                     }
+                    LookupType::ById(id) => {
+                        format!("{} with id \"{}\"", t, id)
+                    }
+                    LookupType::ByCompositeId(label) => {
+                        format!("{} with id \"{}\"", t, label)
+                    }
+                    LookupType::ByOther(msg) => msg,
                 };
-                let message = format!(
-                    "not found: {} with {} \"{}\"",
-                    t, lookup_field, lookup_value
-                );
                 HttpError::for_client_error(
                     Some(String::from("ObjectNotFound")),
                     http::StatusCode::NOT_FOUND,
-                    message,
+                    format!("not found: {}", message),
                 )
             }
 
--- a/nexus/db-queries/src/db/datastore/ip_pool.rs
+++ b/nexus/db-queries/src/db/datastore/ip_pool.rs
@@ -134,12 +134,8 @@ impl DataStore {
         //     .authorize(authz::Action::ListChildren, &authz::IP_POOL_LIST)
         //     .await?;
 
-        // join ip_pool to ip_pool_resource and filter
-
-        // used in both success and error outcomes
-        let lookup_type = LookupType::ByCompositeId(
-            "Default pool for current silo".to_string(),
-        );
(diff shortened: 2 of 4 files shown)


common/src/api/external/error.rs
```diff
@@ -144,11 +144,11 @@ pub enum LookupType {
     ByName(String),
     /// a specific id was requested
     ById(Uuid),
-    /// a session token was requested
-    BySessionToken(String),
```
david-crespo, line 148: This is not used. What's weird is it was introduced in #814, but it is also not used there.

common/src/api/external/error.rs
```diff
@@ -144,11 +144,11 @@ pub enum LookupType {
...
     /// a specific id was requested with some composite type
     /// (caller summarizes it)
     ByCompositeId(String),
+    /// object selected by criteria that would be confusing to call an ID
+    ByOther(String),
```
david-crespo, line 151: A catchall name like this is surely a code smell, but I can't think of a better name.

common/src/api/external/error.rs
```diff
@@ -359,23 +359,22 @@ impl From<Error> for HttpError {
...
+                    LookupType::ByName(name) => {
+                        format!("{} with name \"{}\"", t, name)
                     }
+                    LookupType::ById(id) => {
+                        format!("{} with id \"{}\"", t, id)
+                    }
+                    LookupType::ByCompositeId(label) => {
+                        format!("{} with id \"{}\"", t, label)
+                    }
+                    LookupType::ByOther(msg) => msg,
```
david-crespo, line 372: This is the point of all this — I need a variant that doesn't stick all that extra structure in the message.

nexus/db-queries/src/db/datastore/ip_pool.rs
```diff
@@ -134,12 +134,8 @@ impl DataStore {
...
         //     .await?;
 
-        // join ip_pool to ip_pool_resource and filter
-
-        // used in both success and error outcomes
-        let lookup_type = LookupType::ByCompositeId(
-            "Default pool for current silo".to_string(),
-        );
+        let lookup_type =
+            LookupType::ByOther("default IP pool for current silo".to_string());
```
zephraph, line 138: Other than stripping the ID part of the string (which is helpful because it'll not longer look like an error) I don't know that this provides that much more actionable context. Something like this might be more helpful in the logs:

```rust
LookupType::ByQuery(format!("ip_pool where ip_pool_resource { resource_id: {}, is_default: true }", authz_silo_id));
```

It doesn't have to be exactly that. I know this doesn't communicate in plain english the error (as you've done above), but it does communicate some specifics of what tables to look at if something unexpected is happening. 

cc @internet-diglett do you have thoughts?

common/src/api/external/error.rs
zephraph: Thanks for removing this. I'm not sure how much sense it made in this context to begin with.

nexus/db-queries/src/db/datastore/ip_pool.rs
david-crespo: I considered options more like that, giving the stuff we select by, but to a certain extent I think it’s revealing too much about how the concept of default pool is implemented. It feels weird for an API to be talking about a join table by name.

zephraph: How about combining the two? As with other errors in the system, there's both an internal and external message here. Internally we want enough context to know what to dig into if something's gone wrong. Externally we want a general message for the users to know what's failed at a high level.

david-crespo: True, interesting idea. Mostly only the 500-level errors have internal messages, except `InvalidValue` is a 400-level one used for things like disk size being too big and it does have that. It would be pretty invasive to change `ObjectNotFound` to have that.

https://github.com/oxidecomputer/omicron/blob/c392c76f0cc8a0bdaca5cf7690c6d95dd071e7db/common/src/api/external/error.rs#L28-L73

zephraph: Changing `ObjectNotFound` might be the right way to do it, but that's not what I specifically had in mind. 

image

In the logs above, there's the external event message as you'd expect but also a serialized version of the actual error object. Even if the "query" part of this never showed up in an error, we'd still see it in the logs (we, at least under the context at which this was initially seen). 

Now, it could be that we really only see errors serialized out like that in the logs under very special circumstances (maybe only saga events?). Were that the case, it may not be worth it.

david-crespo:
Merging this as a net improvement. Will think about making it more better later.

PR #4891, opened 2024-01-24
Add a background task for update plan execution

andrewjstone:
This PR is the first step in creating a background task that is capable of taking a `Blueprint` and then reifying that blueprint into deployed or updated software. This PR uses the initial version of a Blueprint introduced in #4804. A basic executor that sends the related `OmicronZonesConfig` to the appropriate sled-agents for newly added sleds was created.

A test is included that shows how a hypothetical planner for an `add-sled` workflow will deploy Omicron zones in a manner similar to RSS, where first the internal DNS zone is deployed and then the internal DNS and NTP zones are deployed. Deployment alwyas contains all zones expected to be running on the sled-agent. Any zones running that are not included are expected to be shut down.

I still need to hook this into `nexus/src/app/background/init.rs`


nexus/src/app/background/plan_execution.rs
```diff
@@ -0,0 +1,408 @@
...
+        log: &Logger,
+        opctx: &OpContext,
+        zones: &BTreeMap<Uuid, OmicronZonesConfig>,
+    ) -> Result<(), Vec<anyhow::Error>> {
+        let errors: Vec<_> = stream::iter(zones.clone())
+            .filter_map(|(sled_id, config)| async move {
+                let client = match self.sled_client(&opctx, &sled_id).await {
+                    Ok(client) => client,
+                    Err(err) => {
+                        warn!(log, "{}", err);
```
jgallagher, line 95: Nit: probably want 

```suggestion
                        warn!(log, "{err:#}");
```

to get the full error chain.

nexus/src/app/background/plan_execution.rs
```diff
@@ -0,0 +1,408 @@
...
+
+impl BackgroundTask for PlanExecutor {
+    fn activate<'a>(
+        &'a mut self,
+        opctx: &'a OpContext,
+    ) -> BoxFuture<'a, serde_json::Value> {
+        async {
+            // Get the latest blueprint, cloning to prevent holding a read lock
+            // on the watch.
+            let blueprint = self.rx_blueprint.borrow().clone();
```
jgallagher, line 141: I'm not sure if a `watch::Receiver` is the final version of how the current blueprint gets into this task, but just in general, should this be `borrow_and_update()` so `.changed()` will work if we end up using it?

nexus/src/app/background/plan_execution.rs
```diff
@@ -0,0 +1,408 @@
...
+        // We re-use the same one to not obfuscate things
+        let dataset = OmicronZoneDataset {
+            pool_name: format!("oxp_{}", Uuid::new_v4()).parse().unwrap(),
+        };
+
+        let generation = Generation::new();
+
+        // Zones are updated in a particular order, but each request contains
+        // the full set of zones that must be running.
+        // See https://github.com/oxidecomputer/omicron/blob/main/sled-agent/src/rack_setup/service.rs#L976-L998
```
jgallagher, line 270: Linking to `main` will atrophy over time, but I'm not sure I like a link to a specific commit much better. Could we reference a method name instead, maybe? If that changes there's a chance this comment gets updated when someone is searching for the method name.

nexus/src/app/background/plan_execution.rs
```diff
@@ -0,0 +1,408 @@
...
+            zones_in_service: BTreeSet::new(),
+            parent_blueprint_id: None,
+            time_created: chrono::Utc::now(),
+            creator: "test".to_string(),
+            comment: "test blueprint".to_string(),
+        }
+    }
+
+    #[nexus_test(server = crate::Server)]
+    async fn test_deploy_omicron_zones(cptestctx: &ControlPlaneTestContext) {
```
jgallagher, line 204: This test is really easy to follow :+1:

nexus/src/app/background/plan_execution.rs
andrewjstone: Hmm. I don't think it's strictly necessary, but it may prevent extra unnecessary calls to `activate` of this taks. I was copying the usage in `dns_propagation` which is how I ended up with a call to `borrow`. 

We actually do use `changed` inside the [GenericWatcher](https://github.com/oxidecomputer/omicron/blob/main/nexus/src/app/background/common.rs#L446-L462) which is the only spot we actually wait for the receiver. Then when the value has changed, which the call to `changed` in `GenericWatcher` marks as seen, the task is activated and we just read the latest value. However, as you surmised, if we see a different value and don't mark it seen, then changed will fire again unnecessarily. I'll modify this code to use `borrow_and_update` which seems like the better option.

CC @davepacheco

nexus/src/app/background/plan_execution.rs
andrewjstone: Fixed in 24b34f70f76a5795d51987442f392dc374cc71eb

nexus/src/app/background/plan_execution.rs
```diff
@@ -0,0 +1,408 @@
...
+                let result = client
+                    .omicron_zones_put(&config)
+                    .await
+                    .with_context(|| {
+                        format!("Failed to put {config:#?} to sled {sled_id}")
+                    });
+
+                match result {
+                    Err(error) => {
+                        warn!(log, "{}", error);
```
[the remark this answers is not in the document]
andrewjstone: Fixed in 24b34f70f76a5795d51987442f392dc374cc71eb

nexus/src/app/background/plan_execution.rs
andrewjstone: Fixed in 24b34f70f76a5795d51987442f392dc374cc71eb

nexus/src/app/background/plan_execution.rs
andrewjstone: Fixed in 24b34f70f76a5795d51987442f392dc374cc71eb

andrewjstone:
This is probably good enough to review. There are still 1-2 more automated tests that should be added: one for the background task that loads the target blueprint, and probably one for testing interaction between the two background tasks, similar to what's done with DNS. I would have already written those, but instead I have spent all weekend trying to get the whole thing running on the a4x2 testbed to see if we could rapidly iterate on testing e2e add-sled. I'm pretty close I think, but am stuck currently with some weird rack init setup errors related to lrtq that need more detailed debugging. I expect a few small PRs to come out of that effort as well.

nexus/src/app/background/blueprint_load.rs
```diff
@@ -0,0 +1,159 @@
...
+                    // but log an error. We'll retry when we're activated again.
+                    let message = format!("{:#}", error);
+                    warn!(
+                        &log,
+                        "failed to read target blueprint";
+                        "error" => &message
+                    );
+                    json!({
+                        "error": 
+                        format!("failed to read target blueprint: {message}")})
```
davepacheco, line 72: nit: the formatting here looks a little weird.

nexus/src/app/background/blueprint_load.rs
```diff
@@ -0,0 +1,159 @@
...
+                (Some(old), Ok(None)) => {
+                    // We have transitioned from having a blueprint to not
+                    // having one. This should not happen.
+                    let message = format!(
+                        "target blueprint with id {} was removed. There is no \
+                        longer any target blueprint",
+                        old.id
+                    );
+                    error!(&log, "{}", message);
+                    json!({"error": message})
```
davepacheco, line 87: Agreed this shouldn't happen.  If it did, I wonder if we should set the current one to `None` anyway?  I'm not really sure to be honest but if Nexus restarted after this point, that's what would happen anyway.

If we're going to keep the existing one, I feel like we should include that information in the JSON object too.

nexus/src/app/background/blueprint_load.rs
```diff
@@ -0,0 +1,159 @@
...
+                }
+                (None, Ok(Some((_, new_target)))) => {
+                    // We've found a new target blueprint for the first time.
```
davepacheco, line 90:
```suggestion
                    // We've found a target blueprint for the first time.
```

nexus/src/app/background/blueprint_load.rs
```diff
@@ -0,0 +1,159 @@
...
+                            log,
+                            "found new target blueprint";
+                            "target_id" => &target_id,
+                            "time_created" => &time_created
+                        );
+                        self.last = Some(Arc::new(new));
+                        self.tx.send_replace(self.last.clone());
+                        json!({
+                            "target_id": target_id,
+                            "time_created": time_created
```
davepacheco, line 121: Suggestion: add `time_found: Utc::now()` (so that if we're debugging some weird behavior, we can see easily when this Nexus first saw this blueprint)

nexus/src/app/background/blueprint_load.rs
```diff
@@ -0,0 +1,159 @@
...
+                        // blueprint to change, but we check to catch possible
+                        // bugs further up the stack.
+                        if **old != new {
+                            let message = format!(
+                                "blueprint for id {} changed. \
+                                Blueprints are supposed to be immutable.",
+                                target_id
+                            );
+                            error!(&log, "{}", message);
+                            json!({"error": message})
```
davepacheco, line 137: Similar to above: if we're going to keep the existing one in this case, then I feel like we should include that information in the JSON object too.

nexus/src/app/background/blueprint_load.rs
```diff
@@ -0,0 +1,159 @@
...
+                            "found new target blueprint";
+                            "target_id" => &target_id,
+                            "time_created" => &time_created
+                        );
+                        self.last = Some(Arc::new(new));
+                        self.tx.send_replace(self.last.clone());
+                        json!({
+                            "target_id": target_id,
+                            "time_created": time_created
+                        })
```
davepacheco, line 122: Also: it'd be great to add code to `omdb` to parse this: a mention [here](https://github.com/oxidecomputer/omicron/blob/45df2e6ce738eba9a66e2f885bdc509d8932834b/dev-tools/omdb/src/bin/omdb/nexus.rs#L258) but mostly [here](https://github.com/oxidecomputer/omicron/blob/45df2e6ce738eba9a66e2f885bdc509d8932834b/dev-tools/omdb/src/bin/omdb/nexus.rs#L327-L339).  This isn't a blocker because omdb will do its best to print what it finds anyway but it'd be nice to have a human-readable summary there.

nexus/src/app/background/blueprint_load.rs
```diff
@@ -0,0 +1,159 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! Background task for loading the target blueprint from the DB
+//!
+//! This task triggers the `plan_execution` background task when the blueprint
```
davepacheco, line 7:
```suggestion
//! This task triggers the `blueprint_execution` background task when the blueprint
```

nexus/src/app/background/blueprint_execution.rs
```diff
@@ -0,0 +1,407 @@
...
+impl BlueprintExecutor {
+    // Temporary until we wire up the background task
+    #[allow(unused)]
+    pub fn new(
+        datastore: Arc<DataStore>,
+        rx_blueprint: watch::Receiver<Option<Arc<Blueprint>>>,
+    ) -> BlueprintExecutor {
+        BlueprintExecutor { datastore, rx_blueprint }
+    }
+
```
davepacheco, line 41: Is this still needed / used?

nexus/src/app/background/blueprint_execution.rs
```diff
@@ -0,0 +1,407 @@
...
+            &format!("http://{}", sled.address()),
+            client,
+            opctx.log.clone(),
+        ))
+    }
+
+    async fn realize_blueprint(
+        &self,
+        opctx: &OpContext,
+        blueprint: &std::sync::Arc<Blueprint>,
```
davepacheco, line 78: nit: I think you could accept `&Blueprint` here.

nexus/src/app/background/blueprint_execution.rs
```diff
@@ -0,0 +1,407 @@
...
+
+            // Return the result as a `serde_json::Value`
+            match result {
+                Ok(()) => json!({}),
+                Err(errors) => {
+                    let errors: Vec<_> = errors
+                        .into_iter()
+                        .map(|e| format!("{:#}", e))
+                        .collect();
+                    json!({"errors": errors})
```
davepacheco, line 160: I think it'd be useful to report which blueprint we tried to execute.

nexus/src/app/background/blueprint_execution.rs
```diff
@@ -0,0 +1,407 @@
...
+            zones_in_service: BTreeSet::new(),
+            parent_blueprint_id: None,
+            time_created: chrono::Utc::now(),
+            creator: "test".to_string(),
+            comment: "test blueprint".to_string(),
+        }
+    }
+
+    #[nexus_test(server = crate::Server)]
+    async fn test_deploy_omicron_zones(cptestctx: &ControlPlaneTestContext) {
```
davepacheco, line 204: Nice test!

I think we could also/instead have one that uses an actual simulated sled agent and fetches the inventory back to verify it.  I'm not sure there'd be much advantage here, except that eventually we'll probably want to make sure that works so we can test higher-level stuff.  (I wouldn't really worry about this now, just mentioning it.)

nexus/src/app/background/blueprint_execution.rs
```diff
@@ -0,0 +1,407 @@
...
+use std::collections::BTreeMap;
+use std::sync::Arc;
+use tokio::sync::watch;
+use uuid::Uuid;
+
+/// Background task that takes a [`Blueprint`] and realizes the change to
+/// the state of the system based on the `Blueprint`.
+pub struct BlueprintExecutor {
+    datastore: Arc<DataStore>,
+    rx_blueprint: watch::Receiver<Option<Arc<Blueprint>>>,
```
davepacheco, line 29: I think it's pretty important that we have a way to turn this thing off at runtime if we find the system is doing something harmful.  I believe we already have an `enabled` bool on the BlueprintTarget for this purpose.   So I think maybe this thing should accept `(BlueprintTarget, BlueprintTarget)` (or something similar, instead of just `Blueprint`).  Then this task could just do nothing if `!target.enabled`.

davepacheco (commented):
Very nice!

nexus/src/app/background/blueprint_load.rs
andrewjstone: fixed in [596c9be](https://github.com/oxidecomputer/omicron/pull/4891/commits/596c9be466c0ba54c41201edf0c18d9b8c6c6280)

nexus/src/app/background/blueprint_load.rs
andrewjstone: Good idea. Done in [596c9be](https://github.com/oxidecomputer/omicron/pull/4891/commits/596c9be466c0ba54c41201edf0c18d9b8c6c6280)

nexus/src/app/background/blueprint_load.rs
andrewjstone: Good idea. Done in [596c9be](https://github.com/oxidecomputer/omicron/pull/4891/commits/596c9be466c0ba54c41201edf0c18d9b8c6c6280)

nexus/src/app/background/blueprint_load.rs
andrewjstone: Whoops, meant to add a `status` field here as done elsewhere. Will do in next commit.

andrewjstone: Added more details in `[5d72471](https://github.com/oxidecomputer/omicron/pull/4891/commits/5d72471ad30c517f19c94d1eab9274fde74bf61d)

nexus/src/app/background/blueprint_load.rs
andrewjstone: Added in [5d72471](https://github.com/oxidecomputer/omicron/pull/4891/commits/5d72471ad30c517f19c94d1eab9274fde74bf61d)

nexus/src/app/background/blueprint_execution.rs
andrewjstone: Good catch!

nexus/src/app/background/blueprint_execution.rs
andrewjstone: Done in 596c9be466c0ba54c41201edf0c18d9b8c6c6280

nexus/src/app/background/blueprint_load.rs
andrewjstone: Hmm, I started to do this, but the fields in each json object aren't necessarily the same and the whole thing would be a bit of a mess. Let me think about it.

nexus/src/app/background/blueprint_execution.rs
andrewjstone: I particularly didn't do that, because I just wanted to test the behavior of the background task. My hope is that I can get the whole thing running in the a4x2 job and we can use the real sled-agent in CI.

nexus/src/app/background/blueprint_load.rs
```diff
@@ -0,0 +1,176 @@
...
+
+impl TargetBlueprintLoader {
+    pub fn new(datastore: Arc<DataStore>) -> TargetBlueprintLoader {
+        let (tx, rx) = watch::channel(None);
+        TargetBlueprintLoader { datastore, last: None, tx, rx }
+    }
+
+    /// Expose the target blueprint
+    pub fn watcher(&self) -> watch::Receiver<Option<Arc<Blueprint>>> {
+        self.rx.clone()
```
jgallagher, line 35: Tiny nit - I think we could use `self.tx.subscribe()` here and drop `self.rx` entirely (unless somewhere else we're assuming there's always at least one receiver, which I'm not seeing at a glance)

nexus/src/app/background/blueprint_load.rs
```diff
@@ -0,0 +1,176 @@
...
+                (Some(old), Ok(Some((_, new)))) => {
+                    let target_id = new.id.to_string();
+                    let time_created = new.time_created.to_string();
+                    if old.id != new.id {
+                        // The current target blueprint has been updated
+                        info!(
+                            log,
+                            "found new target blueprint";
+                            "target_id" => &target_id,
+                            "time_created" => &time_created
```
jgallagher, line 125: Tiny nit - I'd maybe prefix these as `new_target_id` and `new_time_created` to contrast with the `current_*` properties set on the logger when we created it above.

nexus/src/app/background/blueprint_execution.rs
andrewjstone: > (I assume the suggestion of `(BlueprintTarget, BlueprintTarget)` is a typo and should've been `(Blueprint, BlueprintTarget)`.)
> 
I also assumed this, but reversed the types in the tuple ;)

> Just thinking out loud here, this may dumb: Should we squish `BlueprintTarget` into `Blueprint` somehow, or possibly add a new type for that combination? The db method for reading the current target also returns the 2-tuple, which felt a little awkward. I _think_ the only thing from `BlueprintTarget` that anything other than a debugging human will care about is whether it's enabled, right? Some options in no particular order:
> 
> * Add an `enabled` field to `Blueprint` (I think this doesn't make sense, because what does this field mean if the blueprint isn't the target? but maybe there's some way to make it make sense)
> * Add a `blueprint` field to `BlueprintTarget`
> * Add a struct that has `Blueprint` and `BlueprintTarget` as fields
> * Add `EnabledBlueprint` / `DisabledBlueprint` newtypes around `Blueprint`, which can be created from a `BlueprintTarget` and its `Blueprint`

I think dealing with tuples is a pain in the ass, but I don't like the idea of squishing these in the DB. Ideally the Blueprint is immutable, and the pointer to it (the target) is mutable. This would change that. The newtype idea could work though.

nexus/src/app/background/blueprint_load.rs
jgallagher: Happy to defer to y'all on saga output, but would it make sense to have some kind of serializable status enum that both this crate and omdb could use? It would round-trip through `serde_json` but would address the immediate issue of all the different fields.

nexus/src/app/background/blueprint_execution.rs
jgallagher: Ah, I didn't mean squishing in the db itself, just in the types / return values from the db query. But yeah I think the newtype is maybe my preference, although maybe it should be an enum? Something like

```rust
enum CurrentTargetBlueprint {
    Enabled(Blueprint),
    Disabled(Blueprint),
}
```

?

jgallagher: Hm, this loses the extra debugging info that's present in `BlueprintTarget`. Maybe I'm now leaning toward squishing the tuple into a struct just to give it a name?

```rust
struct CurrentTargetBlueprint {
    blueprint: Blueprint,
    metadata: BlueprintTarget,
}
```

or something.

andrewjstone: Deleted my last comment, because it made no sense. lol.

andrewjstone: I implemented this using the current tuple version in [9d95fd8](https://github.com/oxidecomputer/omicron/pull/4891/commits/9d95fd84dc23eee190c1924cf57ff7574617acb7)

I'm happy to change to use to a new type if we agree, although it can probably wait for a follow up.

nexus/src/app/background/blueprint_load.rs
andrewjstone: I dislike prefixing all the properties. What if I changed `current_*` to `original_*` and left everything else as is.

nexus/src/app/background/blueprint_load.rs
andrewjstone: Good call! Fixed in 769cda3c5c3a5209c40b4f635ac4bcfd7fefbf47

nexus/src/app/background/blueprint_load.rs
davepacheco: Yeah, there's:

- put the actual type into the OpenAPI spec and use a progenitor-generated type (seems ideal, but this seems hard to do without making things a lot more complicated?)
- define a struct to use on both sides (seems fine?)
- duplicate the struct on both sides (this is what we do today for the most part, and it's obviously not great, but the existing structs are also pretty simple, and the cost of getting it wrong is low)

They seem to all have their tradeoffs and I don't have a strong feeling.

andrewjstone:
@davepacheco @jgallagher 
I believe I've resolved all your comments, minus the omdb changes. I also added another test. This is ready for another round of review.

davepacheco:
I noticed about the test failure: [this one](https://buildomat.eng.oxide.computer/wg/0/details/01HNEA6EGWR0FEQ8P80S8BH2JC/etd3RSSjAaK2K601rJ7Z7XtP463KM5ls9chXtfepSy71T6BS/01HNEA7RQH1E9H6CP6ZX1PZKHY#S4625) seems to be because your new tasks get printed in non-deterministic order.  Adding your tasks to [this list](https://github.com/oxidecomputer/omicron/blob/45df2e6ce738eba9a66e2f885bdc509d8932834b/dev-tools/omdb/src/bin/omdb/nexus.rs#L258) should fix this.  (omdb could probably print the tasks it doesn't know about in alphabetical order to avoid having to do this.  Sorry.)

I can't tell if you have another problem here which is that `blueprint_loader` ran twice instead of once _while_ the test was running, which spuriously affected the output.  I think the easiest way to deal with this would be to tune the period up in the test suite's config file (which I think is separate from the other ones, so that's hopefully easy).

davepacheco (approved):

andrewjstone:
> I noticed about the test failure: [this one](https://buildomat.eng.oxide.computer/wg/0/details/01HNEA6EGWR0FEQ8P80S8BH2JC/etd3RSSjAaK2K601rJ7Z7XtP463KM5ls9chXtfepSy71T6BS/01HNEA7RQH1E9H6CP6ZX1PZKHY#S4625) seems to be because your new tasks get printed in non-deterministic order. Adding your tasks to [this list](https://github.com/oxidecomputer/omicron/blob/45df2e6ce738eba9a66e2f885bdc509d8932834b/dev-tools/omdb/src/bin/omdb/nexus.rs#L258) should fix this. (omdb could probably print the tasks it doesn't know about in alphabetical order to avoid having to do this. Sorry.)

I actually had those tasks listed and removed them because I was getting that error. It's possible I used the wrong names though. I'll try agian.

> 
> I can't tell if you have another problem here which is that `blueprint_loader` ran twice instead of once _while_ the test was running, which spuriously affected the output. I think the easiest way to deal with this would be to tune the period up in the test suite's config file (which I think is separate from the other ones, so that's hopefully easy).

I will do that! Thanks.

davepacheco:
> > I noticed about the test failure: [this one](https://buildomat.eng.oxide.computer/wg/0/details/01HNEA6EGWR0FEQ8P80S8BH2JC/etd3RSSjAaK2K601rJ7Z7XtP463KM5ls9chXtfepSy71T6BS/01HNEA7RQH1E9H6CP6ZX1PZKHY#S4625) seems to be because your new tasks get printed in non-deterministic order. Adding your tasks to [this list](https://github.com/oxidecomputer/omicron/blob/45df2e6ce738eba9a66e2f885bdc509d8932834b/dev-tools/omdb/src/bin/omdb/nexus.rs#L258) should fix this. (omdb could probably print the tasks it doesn't know about in alphabetical order to avoid having to do this. Sorry.)
> 
> I actually had those tasks listed and removed them because I was getting that error. It's possible I used the wrong names though. I'll try agian.

Ah, I didn't look closely enough at the omdb code.  I think it _does_ print the unknown tasks in alphabetical order.  So I'm guessing what happened here is the checked-in expectorate output is from a run where they _were_ in that list in the other order.

PR #4898, opened 2024-01-25
Update to USDT 0.5.0

bnaecker:
- Update diesel-dtrace
- Update usdt
- Handle API change in how providers are named, since dunders are no longer translated to dashes.


bnaecker:
Blargh, I forgot this is blocked on at least two other repos also updating, `slog-dtrace` and `dropshot`

PR #4903, opened 2024-01-26
Fix IP pools data migration

### LINKED ISSUE #4875: IP pool migration failed to keep fleet default as default
This bit of data migration from #4261 did not behave as expected on dogfood. The `default` pool was a fleet-level default, and there was a second pool `oxide-pool` associated directly with the `oxide` silo but as non-default. In this situation, as the comment describes, the migration ends up associating both pool `default` and `oxide-pool` with silo `oxide`, with `default` keeping `is_default = true` and `oxide-pool` getting `is_default = false`. When we ran the update, however, both had `is_default = false`, leaving silo `oxide` without a default pool.

https://github.com/oxidecomputer/omicron/blob/624fbba20d7eb754e2c1a3ca7f1c5676fd2f1e0d/schema/crdb/23.0.0/up4.sql#L14-L28

This is despite our lovely test for this very scenario. Indeed, this is the change these tests were added for.

https://github.com/oxidecomputer/omicron/blob/624fbba20d7eb754e2c1a3ca7f1c5676fd2f1e0d/nexus/tests/integration_tests/schema.rs#L1022-L1026

david-crespo:
Closes #4875

## Problem

After the IP pools migrations on the dogfood rack, the `default` pool was not marked `is_default=true` for the `oxide` silo when it should have been.

## Diagnosis

When checking for silo-scoped default pools overriding a fleet-scoped default, I neglected to require that the silo-scoped defaults in question were non-deleted. This means that if there was a deleted pool with `silo_id=<oxide silo id>` and `is_default=true`, that would be considered an overriding default and leave us with `is_default=false` on the `default` pool.

Well, I can't check `silo_id` and `is_default` on the pools because those columns have been dropped, but there is a deleted pool called `oxide-default` that says in the description it was meant as the default pool for only the `oxide` silo. 

```
oot@[fd00:1122:3344:105::3]:32221/omicron> select * from omicron.public.ip_pool;
                   id                  |        name        |          description           |         time_created          |         time_modified         |         time_deleted          | rcgen
---------------------------------------+--------------------+--------------------------------+-------------------------------+-------------------------------+-------------------------------+--------
  1efa49a2-3f3a-43ab-97ac-d38658069c39 | oxide-default      | oxide silo-only pool - default | 2023-08-31 05:33:00.11079+00  | 2023-08-31 05:33:00.11079+00  | 2023-08-31 06:03:22.426488+00 |     1
```

I think we can be pretty confident this is what got us.

## Fix

Add `AND time_deleted IS NULL` to the subquery.

## Mitigation in existing systems

Already done. Dogfood is the only long-running system where the bad migration ran, and all I had to do there was use the API to set `is_default=true` for the (`default` pool, `oxide` silo) link.

### DIFF
--- a/nexus/tests/integration_tests/schema.rs
+++ b/nexus/tests/integration_tests/schema.rs
@@ -954,6 +954,7 @@ const SILO2: Uuid = Uuid::from_u128(0x222251F0_5c3d_4647_83b0_8f3515da7be1);
 const POOL1: Uuid = Uuid::from_u128(0x11116001_5c3d_4647_83b0_8f3515da7be1);
 const POOL2: Uuid = Uuid::from_u128(0x22226001_5c3d_4647_83b0_8f3515da7be1);
 const POOL3: Uuid = Uuid::from_u128(0x33336001_5c3d_4647_83b0_8f3515da7be1);
+const POOL4: Uuid = Uuid::from_u128(0x44446001_5c3d_4647_83b0_8f3515da7be1);
 
 // "513D" -> "Sled"
 const SLED1: Uuid = Uuid::from_u128(0x1111513d_5c3d_4647_83b0_8f3515da7be1);
@@ -977,7 +978,8 @@ fn before_23_0_0(client: &Client) -> BoxFuture<'_, ()> {
             (id, name, description, time_created, time_modified, time_deleted, rcgen, silo_id, is_default) VALUES
           ('{POOL1}', 'pool1', '', now(), now(), NULL, 1, '{SILO1}', true),
           ('{POOL2}', 'pool2', '', now(), now(), NULL, 1, '{SILO2}', false),
-          ('{POOL3}', 'pool3', '', now(), now(), NULL, 1, null, true);
+          ('{POOL3}', 'pool3', '', now(), now(), NULL, 1, null, true),
+          ('{POOL4}', 'pool4', '', now(), now(), NULL, 1, null, false);
         ")).await.expect("Failed to create IP Pool");
     })
 }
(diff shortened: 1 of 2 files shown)


schema/crdb/23.0.0/up4.sql
```diff
@@ -23,8 +23,11 @@ SELECT
   -- AND NOT EXISTS here causes is_default to be false in row 1 if there is a
   -- conflicting silo default pool. row 2 is inserted in up5.
   p.is_default AND NOT EXISTS (
-    SELECT 1 FROM omicron.public.ip_pool 
-    WHERE silo_id = s.id AND is_default
+    SELECT 1 
+    FROM omicron.public.ip_pool p0
+    WHERE p0.silo_id = s.id 
+      AND p0.is_default
+      AND p0.time_deleted IS NULL
```
david-crespo, line 30: `and time_deleted is null` is the fix.

nexus/tests/integration_tests/schema.rs
```diff
@@ -975,9 +977,11 @@ fn before_23_0_0(client: &Client) -> BoxFuture<'_, ()> {
         // no corresponding silo.
         client.batch_execute(&format!("INSERT INTO ip_pool
             (id, name, description, time_created, time_modified, time_deleted, rcgen, silo_id, is_default) VALUES
+          ('{POOL0}', 'pool2', '', now(), now(), now(), 1, '{SILO2}', true),
```
david-crespo, line 980: With this deleted pool added to the test data but the **fix commented out**, the test fails in the way we expect, which matches the problem on dogfood: the (`pool3`, `silo2`) link gets `is_default=false` instead of `is_default=true`.

Issue #4911, opened 2024-01-27
test flake: 'integration_tests::sp_updater::test_sp_updater_delivers_progress' panicked at nexus/tests/integration_tests/sp_updater.rs:574:18


rcgoodfellow:
See 
- https://buildomat.eng.oxide.computer/wg/0/details/01HN4PESXPQF9QBHDEF7BDDCAW/JkbzZ4e3pCTe4Rf0JkcawMf5VsCMkGdlnSvzzfSrVvC4ph2E/01HN4PFAFKAMHTDE91X4W57BGK

jgallagher:
I think I see from the logs what happened. Working on reproducing / fixing.

PR #4922, opened 2024-01-29
Rework {Sp,Rot,HostPhase1}Updater tests to be less flaky 

jgallagher:
There were three tests that were mostly copy/pasted, that all tried to tame an inherent race between two tokio tasks: one is delivering an update to the simulated SP, and the other is polling that simulated SP for the progress of that update. The taming was by throttling the simulated SP to only reply to a single message at a time, then checking some internal state to infer what kind of message the SP had just received, and using that inference to check some invariants. The inference was typically right, but as evidenced by the name of this PR, not perfect.

I took a fairly big hammer to the Rot and HostPhase1 updater tests: they now just loop waiting to see the "Complete" progress message, and no longer try to single-step through messages. Under normal conditions (e.g., not heavily-loaded-CI systems), it's likely these tests will skip from "no progress" to "complete", but that's ok.

I tried to rework the SpUpdater test to keep the same "single-step through messages" to carefully check the progress reports by adding more details to the simulated SP, so the test can know exactly when the specific message it was trying to infer (i.e., `SpUpdater`'s requests for the update status) arrives. This should fix the flake; if it doesn't, we can replace this one with the same impl that the Rot and HostPhase1 tests now have.

Fixes https://github.com/oxidecomputer/omicron/issues/4911

### DIFF
--- a/nexus/tests/integration_tests/host_phase1_updater.rs
+++ b/nexus/tests/integration_tests/host_phase1_updater.rs
@@ -6,7 +6,7 @@
 //! MGS to SP.
 
 use gateway_client::types::SpType;
-use gateway_messages::{SpPort, UpdateInProgressStatus, UpdateStatus};
+use gateway_messages::SpPort;
 use gateway_test_utils::setup as mgs_setup;
 use omicron_nexus::app::test_interfaces::{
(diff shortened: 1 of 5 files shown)


andrewjstone (approved):
Seems reasonable to me.

PR #4932, opened 2024-01-30
Perform inventory fetch queries in batches

### LINKED ISSUE #4629: inventory fetch should query database in batches

hawkw:
When querying the database for inventory data, we read a large number of
rows from a lot of different tables. Therefore, we want to ensure that
there is always an upper bound on the amount of data read per query.
Currently, an upper bound is ensured using _limits_: we add a limit to
the query, and if more rows would be returned, we fail the query. This
means that we can ensure an upper bound. However, the issue with this is
that the behavior when the upper bound is reached is to *fail* --- this
means that it's possible to select an upper bound that's too small, and
upon hitting that upper bound, we just fail...and there's no way to know
whether the selected limit is big enough until we hit it at runtime.
Instead of using hard limits, it's preferable to use _pagination_. By
paginating these queries, we still set an upper bound on the number of
rows returned by a single query, but rather than failing if that bound
is reached, we instead perform multiple queries.

This branch changes inventory fetch queries to use pagination with a
batch size limit, rather than a hard limit on query size, using the
pagination utilities added in #4632. I've replaced the
`inventory_collection_read_best_effort` and
`inventory_collection_read_all_or_nothing` methods, which limit the size
of the query, with a new `inventory_collection_read_batched` method,
which uses the limit parameter as a maximum size per batch. I've then
changed the existing `inventory_get_latest_collection` method to use
`inventory_collection_read_batched` rather than
`inventory_collection_read_all_or_nothing`.

Closes #4629


nexus/db-queries/src/db/datastore/inventory.rs
```diff
@@ -1195,14 +1196,14 @@ impl DataStore {
...
-    /// records
+    /// Attempt to read the latest collection while limiting queries to `batch_size`
+    /// records at a time.
     ///
     /// If there aren't any collections, return `Ok(None)`.
     pub async fn inventory_get_latest_collection(
         &self,
         opctx: &OpContext,
-        limit: NonZeroU32,
+        batch_size: NonZeroU32,
```
davepacheco, line 1206: I suspect it's not important to allow consumers to control the batch size.  I think I'd probably just define a `const` in this file of like 500 or 1000 and use that here.  (You'd get to remove one or two apologetic block comments associated with arbitrary constants like [this one](https://github.com/oxidecomputer/omicron/blob/main/nexus/src/app/rack.rs#L74-L85)!).  If there's a consumer that needs to control it, maybe we could accept `Option<NonZeroU32>` and default to a `const` in this file?  Basically I don't want to have N consumers have to make this decision when it really shouldn't matter for most of them.

Update: I just found [this example](https://github.com/oxidecomputer/omicron/blob/5ba928fb9b661fb46142be8c63552e9e7d7c6278/nexus/src/app/background/sync_service_zone_nat.rs#L65), which is definitely dangerous, and so I'm more convinced we *don't* want callers to have to provide something here.

nexus/db-queries/src/db/datastore/inventory.rs
```diff
@@ -1220,48 +1221,26 @@ impl DataStore {
...
+    /// Attempt to read the current collection.
+    ///
+    /// Queries are limited to `batch_size` records at a time, performing
+    /// multiple queries if more than `batch_size` records exist.
+    pub async fn inventory_collection_read_batched(
         &self,
         opctx: &OpContext,
         id: Uuid,
-        limit: NonZeroU32,
+        batch_size: NonZeroU32,
```
davepacheco, line 1241: Same -- if this is `pub`, I'd drop the `batch_size` or make it an optional override.

nexus/db-queries/src/db/datastore/inventory.rs
```diff
@@ -1285,73 +1264,115 @@ impl DataStore {
...
                 .load_async(&*conn)
                 .await
-                .map_err(|e| public_error_from_diesel(e, ErrorHandler::Server))?
-                .into_iter()
-                .map(|sp_row| {
-                    let baseboard_id = sp_row.hw_baseboard_id;
+                .map_err(|e| {
+                    public_error_from_diesel(e, ErrorHandler::Server)
+                })?;
+                paginator = p.found_batch(&batch, &|row| row.time_collected);
```
davepacheco, line 1309: The key here really should be unique, but `time_collected` isn't.  How about using `hw_baseboard_id`?

nexus/db-queries/src/db/datastore/inventory.rs
```diff
@@ -1285,73 +1264,115 @@ impl DataStore {
...
                 .select(InvRootOfTrust::as_select())
                 .load_async(&*conn)
                 .await
-                .map_err(|e| public_error_from_diesel(e, ErrorHandler::Server))?
-                .into_iter()
-                .map(|rot_row| {
+                .map_err(|e| {
+                    public_error_from_diesel(e, ErrorHandler::Server)
+                })?;
+                paginator = p.found_batch(&batch, &|row| row.time_collected);
```
davepacheco, line 1340: Same here -- use `hw_baseboard_id`?

nexus/db-queries/src/db/datastore/inventory.rs
```diff
@@ -1285,73 +1264,115 @@ impl DataStore {
...
                 .filter(dsl::inv_collection_id.eq(id))
-                .limit(sql_limit)
                 .select(InvSledAgent::as_select())
                 .load_async(&*conn)
                 .await
                 .map_err(|e| {
                     public_error_from_diesel(e, ErrorHandler::Server)
-                })?
+                })?;
+                paginator = p.found_batch(&batch, &|row| row.time_collected);
```
davepacheco, line 1371: Same -- use `sled_id`?

nexus/db-queries/src/db/datastore/inventory.rs
```diff
@@ -1457,17 +1489,29 @@ impl DataStore {
...
                 .filter(dsl::inv_collection_id.eq(id))
-                .limit(sql_limit)
                 .select(InvCaboose::as_select())
                 .load_async(&*conn)
                 .await
                 .map_err(|e| {
                     public_error_from_diesel(e, ErrorHandler::Server)
-                })?
+                })?;
+                paginator = p.found_batch(&batch, &|row| row.time_collected);
```
davepacheco, line 1509: Same -- but this is trickier because I think the unique key is really (`inv_collection_id` (which is implied by the filter), `hw_baseboard_id`, `which`).  Maybe we can use `paginated_multicolumn` here?

nexus/db-queries/src/db/datastore/inventory.rs
```diff
@@ -1537,17 +1589,29 @@ impl DataStore {
...
                 .filter(dsl::inv_collection_id.eq(id))
-                .limit(sql_limit)
                 .select(InvRotPage::as_select())
                 .load_async(&*conn)
                 .await
                 .map_err(|e| {
                     public_error_from_diesel(e, ErrorHandler::Server)
-                })?
+                })?;
+                paginator = p.found_batch(&batch, &|row| row.time_collected);
```
davepacheco, line 1609: Same -- and this one's also multicolumn I think.  (When in doubt for what should go here, check out the unique indexes / primary keys in schema/crdb/dbinit.sql.)

nexus/db-queries/src/db/datastore/inventory.rs
```diff
@@ -1626,62 +1701,98 @@ impl DataStore {
...
                 .select(InvSledOmicronZones::as_select())
                 .load_async(&*conn)
                 .await
-                .map_err(|e| public_error_from_diesel(e, ErrorHandler::Server))?
-                .into_iter()
-                .map(|sled_zones_config| {
+                .map_err(|e| {
+                    public_error_from_diesel(e, ErrorHandler::Server)
+                })?;
+                paginator = p.found_batch(&batch, &|row| row.time_collected);
```
davepacheco, line 1721: I think you can use `sled_id` here.

nexus/db-queries/src/db/datastore/inventory.rs
```diff
@@ -1939,18 +2047,6 @@ mod test {
         assert_eq!(collection2.cabooses.len(), coll_counts.cabooses);
         assert_eq!(collection2.rot_pages.len(), coll_counts.rot_pages);
 
-        // Check that we get an error on the limit being reached for
```
davepacheco, line 1942: How about adding a test that uses a batch size of 1 and makes sure we get the same result for some reasonably representative collection?  (That way we know the pagination stuff is working.)

dev-tools/omdb/src/bin/omdb/db.rs
```diff
@@ -2499,13 +2499,10 @@ async fn cmd_db_inventory_collections_show(
     limit: NonZeroU32,
     long_string_formatter: LongStringFormatter,
 ) -> Result<(), anyhow::Error> {
-    let (collection, incomplete) = datastore
-        .inventory_collection_read_best_effort(opctx, id, limit)
+    let collection = datastore
+        .inventory_collection_read_batched(opctx, id, limit)
```
davepacheco, line 2503: Related to my other comment: I don't think it's important that `omdb` pass through its `limit` argument here.  That argument exists for the many cases that _don't_ do pagination because there the `omdb` user has to make a tradeoff between completeness and (in theory) risk of making a super expensive query.  In cases that are doing client-side pagination, we don't have that tradeoff (hooray!) so it's fine to ignore that parameter.

nexus/db-queries/src/db/datastore/inventory.rs
hawkw: Ah, that's what I'd done initially, but when I read the documentation for `Paginator`, I got the impression that this was supposed to be how the rows were _ordered_. I'll change it to the unique ID here (and elsewhere).

nexus/db-queries/src/db/datastore/inventory.rs
hawkw: Hmm, if we follow your advice from above about not making the batch size
configurable, we'd have to factor out a private function that takes a batch size
parameter...do you think it's worth doing that just for testing this?

nexus/db-queries/src/db/datastore/inventory.rs
davepacheco: Ah, yeah, it should be both: we want an index that's ordered the way we want to paginate through them, and where each item can be uniquely identified by one the last value in the index.

nexus/db-queries/src/db/datastore/inventory.rs
davepacheco: Yeah, I think that's easy enough to do and worthwhile.  I've been burned by a surprising number of pagination bugs in the past!

nexus/db-queries/src/db/datastore/inventory.rs
hawkw: Got it, thanks. I wasn't sure if the UUIDs met the ordering requirement in this case, but based on your suggestion above I'm assuming they do?

davepacheco: That's a good question.  If we take this example of `inv_service_processor.hw_baseboard_id`, [we have a composite primary key of `inv_collection_id` (which is part of the filter here) and then `hw_baseboard_id`](https://github.com/oxidecomputer/omicron/blob/5439ce8f391e1f2f1679378171c5a3d880f7ee37/schema/crdb/dbinit.sql#L2784), so yes, we'll have a unique index sorted first by `inv_service_processor`, then by `hw_baseboard_id`, which means we can (1) fetch any page of rows of `inv_service_processor` by going directly to the corresponding part of the index, and then (2) uniquely identify the last row on any page.  (By "page" here I mean the sense used in this code, not an OS page or filesystem page or database page or anything like that.)

We'd want to check that for all the other cases, too.  In some cases, it might not be the primary key, but there may be another unique index.  And if there isn't already a suitable unique index, we'll probably want to add one to the schema.  I'm happy to talk through this live if that's useful, especially if we find we have to add an index.

hawkw: I think everything should be using the right keys for pagination as of 02f39d2ae8b1898cad45c6914f0d3f072dd2a3dd --- if I'm wrong about that, happy to chat about it either in review or synchronously!

davepacheco (commented):
The changes look good!  Were you also going to add the new test?

hawkw:
> The changes look good! Were you also going to add the new test?

Yup, I just didn't get to it yesterday. Planning to add that today!

davepacheco:
Oops -- didn't mean to jump the gun!

davepacheco (approved):

PR #4933, opened 2024-01-30
Support Omicron development on NixOS

hawkw:
I use NixOS on my dev machine, and I'd like to be able to do some
development tasks locally (such as `cargo check`, running
`rust-analyzer`, et cetera). This branch makes the following changes to
better support Omicron development on NixOS:

- **Add `flake.nix` for Nix users** (484323c)

  In order to build Omicron, including running `cargo check` or
  `rust-analyzer`, a number of non-Rust dependencies must be present.
  This commit adds a [Nix flake] file which users of the Nix package
  manager can [use to enter a dev shell][nix-develop] that contains
  those dependencies, making it easy for Nix and NixOS users to develop
  Omicron on NixOS.

  The Nix flake configured here creates a build environment with `clang`
  as the C compiler toolchain, and ensures that the following libraries
  are available:

  - `libclang` (of a version matching the `clang` C toolchain)
  - `libpq`
  - `libxml2`
  - `libxmlsec`
  - `openssl`
  - `sqlite`

  In addition, it provides `pkg-config`, `cmake`, and `rustup`, and
  ensures that the various environment variables that the Rust bindings
  for these libraries use to find the libraries are set correctly.

  This makes it possible to run `cargo check`, `cargo build`, and
  `rust-analyzer` on NixOS.

  I've also added to the `.envrc` file so that `direnv` will
  automatically use the Nix flake's dev environment, if a `nix` binary
  is present on the user's system and that `nix` binary supports flakes.

  [Nix flake]: https://nixos.wiki/wiki/Flakes
  [nix-develop]:
      https://nixos.wiki/wiki/Development_environment_with_nix-shell#nix_develop

- **Use `/usr/bin/env bash` rather than `/bin/bash`** (9b438c1)

  On some operating systems, such as [NixOS][1], there may not be a
  `bash` executable in `/bin`, but a `bash` executable may be located
  elsewhere. Using a `#!/bin/bash` shebang line in scripts may not work
  correctly on these systems.

  Instead, a `#!/usr/bin/env bash` shebang is [generally considered][2]
  the preferable way to write portable shell scripts that use Bash.
  Therefore, this commit changes the shell scripts in `tools/` to use
  `#!/usr/bin/env bash` shebang lines, rather than `#!/bin/bash`.

  [1]: https://discourse.nixos.org/t/shebang-locations/28992
  [2]:
      https://www.cyberciti.biz/tips/finding-bash-perl-python-portably-using-env.html
      
Hopefully, checking config files for supporting specific dev tools and
environments into Git is considered acceptable. Nix flakes require the
flake and its lockfile be tracked in Git, as Git revisions are used for
caching. And, I believe a few other folks at Oxide also use Nix, so I
thought having the flake in Git would be helpful for them as well.
However, if people object to this being checked into Git, I can also use
an (untracked) legacy `shell.nix` configuration for `nix-shell`, instead
(which I'd be happy to share with other Nix users).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ debug.out
 rusty-tags.vi
 *.sw*
 tags
+.direnv
--- a/flake.nix
+++ b/flake.nix
@@ -54,6 +54,10 @@
               LIBCLANG_PATH = "${libclang.lib}/lib";
               OPENSSL_DIR = "${openssl.dev}";
               OPENSSL_LIB_DIR = "${openssl.out}/lib";
+
+              # Needed by rustfmt-wrapper, see:
+              # https://github.com/oxidecomputer/rustfmt-wrapper/blob/main/src/lib.rs
+              RUSTFMT = "${rustToolchain}/bin/rustfmt";
             };
         }
       );
--- a/flake.nix
+++ b/flake.nix
@@ -52,11 +52,8 @@
               name = "omicron";
               DEP_PQ_LIBDIRS = " ${postgresql.lib}/lib";
               LIBCLANG_PATH = "${libclang.lib}/lib";
-              LOCALE_ARCHIVE = "${glibcLocales}/lib/locale/locale-archive";
-              LC_ALL = "en_US.UTF-8";
               OPENSSL_DIR = "${openssl.dev}";
               OPENSSL_LIB_DIR = "${openssl.out}/lib";
-              CARGO_TERM_COLOR = "always";
             };
         }
       );
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,63 @@
+{
+  description = "Development environment for Omicron";
+
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
+    flake-utils.url = "github:numtide/flake-utils";
+    rust-overlay = {
+      url = "github:oxalica/rust-overlay";
+      inputs = {
+        nixpkgs.follows = "nixpkgs";
+        flake-utils.follows = "flake-utils";
+      };
+    };
+  };
+
+  outputs = { self, nixpkgs, flake-utils, rust-overlay }:
+    flake-utils.lib.eachDefaultSystem
+      (system:
+        let
+          overlays = [ (import rust-overlay) ];
+          pkgs = import nixpkgs {
+            inherit system overlays;
+          };
+          # use the Rust toolchain defined in the `rust-toolchain.toml` file.
+          rustToolchain = pkgs.pkgsBuildHost.rust-bin.fromRustupToolchainFile ./rust-toolchain.toml;
+          nativeBuildInputs = with pkgs; [
+            rustToolchain
+            cmake
+            stdenv
+            pkg-config
+          ];
+          buildInputs = with pkgs; [
+            # libs
+            openssl
+            postgresql
+            xmlsec
+            sqlite
+            libclang
+            libxml2
+          ];
+        in
+        with pkgs;
+        {
+          devShells.default = mkShell.override
+            {
+              # use Clang as the C compiler for all C libraries
+              stdenv = clangStdenv;
+            }
+            {
+              inherit buildInputs nativeBuildInputs;
+
+              name = "omicron";
+              DEP_PQ_LIBDIRS = " ${postgresql.lib}/lib";
+              LIBCLANG_PATH = "${libclang.lib}/lib";
+              LOCALE_ARCHIVE = "${glibcLocales}/lib/locale/locale-archive";
+              LC_ALL = "en_US.UTF-8";
+              OPENSSL_DIR = "${openssl.dev}";
+              OPENSSL_LIB_DIR = "${openssl.out}/lib";
+              CARGO_TERM_COLOR = "always";
+            };
+        }
+      );
+}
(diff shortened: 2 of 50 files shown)


hawkw:
@sunshowers and @faithanalog, I've tagged you both as reviewers for this branch as IIRC you're both using Nix. If anyone else ought to review this as well, please let me know!

hawkw:
> Though, it could also be cool if the flake handled getting those binaries to make `install_builder_prequisites.sh` completely unnecessary

Yeah, I was also thinking about this --- I got everything building for now by running the `tools/ci_download_maghemite_openapi` and `tools/ci_download_dendrite_openapi` scripts, but I'd like to have the flake handle that for you as well. My plan was to do that in a separate PR, but I'm happy to work on it now and include that here if you'd prefer.

sunshowers:
lgtm! though I'll let @faithanalog accept it

.gitignore
```diff
@@ -15,3 +15,4 @@ debug.out
 rusty-tags.vi
 *.sw*
 tags
+.direnv
```
sunshowers, line 18: Is this a directory that gets added to the working copy due to this PR?

faithanalog: yes.

```
vi@philomena ~/o/omicron (eliza/nixos-build)> ls .direnv/
flake-profile@  flake-profile-1-link@
```

hawkw: Yes, if using the flake via direnv, a bunch of stuff gets cached in `.direnv`.

hawkw:
@faithanalog for now, I've changed `install_builder_prerequisites` as you've suggested. In the future, I'd like to see if the external requirements can be represented as inputs to the flake, so that `install_builder_prerequisites` isn't necessary at all when using Nix. But, my goal here is to get a minimal working Nix dev environment, so having to run `install_builder_prerequisites` manually seems acceptable for a first pass...

flake.nix
```diff
@@ -0,0 +1,63 @@
...
+              inherit buildInputs nativeBuildInputs;
+
+              name = "omicron";
+              DEP_PQ_LIBDIRS = " ${postgresql.lib}/lib";
+              LIBCLANG_PATH = "${libclang.lib}/lib";
+              LOCALE_ARCHIVE = "${glibcLocales}/lib/locale/locale-archive";
+              LC_ALL = "en_US.UTF-8";
+              OPENSSL_DIR = "${openssl.dev}";
+              OPENSSL_LIB_DIR = "${openssl.out}/lib";
+              CARGO_TERM_COLOR = "always";
```
sunshowers, line 59: Hmm, what's the reason this is required?

flake.nix
```diff
@@ -0,0 +1,63 @@
...
+              stdenv = clangStdenv;
+            }
+            {
+              inherit buildInputs nativeBuildInputs;
+
+              name = "omicron";
+              DEP_PQ_LIBDIRS = " ${postgresql.lib}/lib";
+              LIBCLANG_PATH = "${libclang.lib}/lib";
+              LOCALE_ARCHIVE = "${glibcLocales}/lib/locale/locale-archive";
+              LC_ALL = "en_US.UTF-8";
```
sunshowers, line 56: Interesting!

flake.nix
hawkw: this was just copied and pasted from another dev shell, i think it can be removed.

flake.nix
hawkw: we don't actually need this either, it also slipped in when i copied and pasted
a big blob of stuff from a different nix shell. my bad!

faithanalog:
While testing on a fresh VM, I learned that we also need to set `RUSTFMT` for [rustfmt-wrapper](https://github.com/oxidecomputer/rustfmt-wrapper/blob/main/src/lib.rs) which is used by dpd-client, otherwise I get an error about rustfmt missing:

```
error: failed to run custom build command for `dpd-client v0.1.0 (/root/omicron/clients/dpd-client)`

Caused by:
  process didn't exit successfully: `/root/omicron/target/debug/build/dpd-client-ee1ad752558d9c24/build-script-build` (exit status: 1)
  --- stdout
  cargo:rerun-if-changed=../../package-manifest.toml
  cargo:rerun-if-changed=../../out/downloads/dpd-fd159136c552d8b4ec4d49dd9bae7e38f6a636e6.json

  --- stderr
  Error: rustfmt failed on progenitor code from ../../out/downloads/dpd-fd159136c552d8b4ec4d49dd9bae7e38f6a636e6.json

  Caused by:
      rustfmt is not installed
```


I fixed it with this diff:

```
diff --git a/flake.nix b/flake.nix
index 576b038fc..637a1ce64 100644
--- a/flake.nix
+++ b/flake.nix
@@ -54,6 +54,9 @@
               LIBCLANG_PATH = "${libclang.lib}/lib";
               OPENSSL_DIR = "${openssl.dev}";
               OPENSSL_LIB_DIR = "${openssl.out}/lib";
+
+              # Needed by rustfmt-wrapper, see https://github.com/oxidecomputer/rustfmt-wrapper/blob/main/src/lib.rs
+              RUSTFMT = "${rustToolchain}/bin/rustfmt";
             };
         }
       );
```

Other than that:

- `cargo build --release` fails trying to link omdb against libltdl. Unfortunately it looks like nixpkgs doesn't currently provide libltdl.
  - I don't think it matters since the purpose of this PR is to make it easier to run editor tooling, but noting it here.
- direnv, tests, and build all still work fine on my illumos workstation

So I say add the RUSTFMT var, and then everything else looks good to me!

hawkw:
Great, thanks @faithanalog! The `libltdl` thing seems like a bit of a bother, but I agree that it's not a hard blocker for the initial purpose of just being able to run `rust-analyzer`. I can look into it later.

faithanalog (approved):

PR #4934, opened 2024-01-30
[nexus] basic scaffolding for typed UUIDs + a couple of simple examples

sunshowers:
This PR contains a basic implementation of typed UUIDs using the [newtype-uuid](https://crates.io/crates/newtype-uuid) crate, plus a simple conversion of two UUID types to this model: `LoopbackAddress` and `TufRepo`.

I picked these two types based on the fact that there were just a few places that used them.

## Determinations

This PR introduces a new crate called `omicron-uuid-kinds`. This effectively forms a registry of all the different kinds of UUIDs we have, one that (to a reasonable extent) can be shared across Oxide repos.

### One big registry or several smaller ones?

One question that came up was whether we want a central registry or several smaller ones. The best answer to that is currently unclear, but putting them in the same crate has several advantages:

* The macros updated in this PR know where uuid kinds will be found, so they let you specify just a bare name as the kind.
* It's simpler.

The disadvantage of this is that any change to this registry will cause omicron to be rebuilt. Hopefully, once most UUIDs are converted over, this crate won't be touched too much.

This is not an entirely commitment-free decision because turning one registry into several will require at least a few hours of work -- for example, macros may need to be updated to handle multiple sources.

Crates which aren't part of the omicron cinematic universe, like dropshot and steno, will probably have their own registries (probably within the crates themselves).

### Where should the registry live?

Once we've decided on that, the next question is where this registry should be stored. There was some debate in `#oxide-control-plane`, and two options stood out:

1. In omicron itself.

    * The upside of this is that it is very to add new UUID kinds within omicron, especially as we transition over existing UUIDs.
    * The downside is that we also want to use these kinds in crucible and propolis -- and pulling in `omicron-uuid-kinds` as a Git dependency would typically cause two copies of `omicron-uuid-kinds` to be floating around. That can result in a lot of pain, including extremely confusing errors.

2. In a separate repo. While it eliminates the issue with duplicated deps, it does add a fair bit more friction because users updating omicron would now have to make two separate PRs.

In this PR, I've chosen 1 because I believe that at this point agility within omicron is the most important thing. The downside is mitigated by using a `[patch]` directive -- see `Cargo.toml` for more. (I've also written the crate in a way that makes it possible for no-std environments to pull in, if necessary.)

This is straightforward to change. If we find that 1 is too much of a bother, we'll easily be able to switch to 2 while keeping the name of the crate the same (so minimal churn).


sunshowers:
Notes from update sync:

* move `TypedUuid` into a separate crate so that crucible and propolis etc can use it
* consider e.g. `type SledId = TypedUuid<SledType>`.

sunshowers:
> consider e.g. type SledId = TypedUuid<SledKind>.

After giving it some thought, I'm honestly leaning against this a bit because it adds an extra name to the type system. Someone new to the codebase would have to understand that `SledId` is the same as `TypedUuid<SledKind>`.

The main benefit of type aliases is being able to describe complicated types succinctly. However, I personally don't feel `TypedUuid<SledKind>` is complex enough to justify the cost. (This is based purely on judgment, and I'm interested in hearing if others feel differently).

andrewjstone:
> > consider e.g. type SledId = TypedUuid.
> 
> After giving it some thought, I'm honestly leaning against this a bit because it adds an extra name to the type system. Someone new to the codebase would have to understand that `SledId` is the same as `TypedUuid<SledKind>`.
> 
> The main benefit of type aliases is being able to describe complicated types succinctly, without going through the motions of implementing a wrapper newtype. However, I personally don't feel `TypedUuid<SledKind>` is complex enough to justify the cost. (This is based purely on judgment, and I'm interested in hearing if others feel differently).

I agree with you Rain. I don't think the TypedUuid<SledKind> adds much of a burden. If it's used a ton in a given place, a local unexported type alias (per file) can be made.

sunshowers:
> move TypedUuid into a separate crate so that crucible and propolis etc can use it

Done: https://crates.io/crates/newtype-uuid

zephraph:
Thanks for starting this @sunshowers, it looks really interesting. One thing I wanted to note is that when we were doing the API rewrite we were also looking to support [typed UUIDs as user input](https://rfd.shared.oxide.computer/rfd/0322#_ids) similar to Stripe's [typed ids](https://gist.github.com/fnky/76f533366f75cf75802c8052b577e2a5). There's an [open source library](https://github.com/jetpack-io/typeid) that has a similar approach as Stripe's. 

We have some extra constraints though. Given that we allow the selection of resources via the API by either name or ID we have restrictions such that a name _can't_ be a UUID. It chatting with @ahl about this early on we'd loosely spec'd out `<resource>:<uuid>` (a light [urn](https://en.wikipedia.org/wiki/Uniform_Resource_Name)) given that `:` is an invalid character for names.

 To outline how I think about the flow of identifiers in the API:
1. A request comes into an API handler. At that point we should resolve the `NameOrId` reference into a `Lookup` struct.
2. The `Lookup` gets passed to the Nexus app layer and is used to perform a lookup and resolve an `authz` resource. 
3. The data layer should accept an `authz` resource (_not a raw UUID_) which can then be used to perform a query by ID via `authz_resource.id()`

I've definitely had issues with UUIDs (and names) being passed to places and potentially being the wrong types. I believe strongly that we shouldn't be passing bare specifiers (names/ids) in _most_ cases in the external API flow. That said, I understand sagas have to pass around IDs and the typed IDs are perfect for that. 

So here's what I'd like to see with the typed IDs in the short term: `authz_resource.id()` should return a typed ID of the relevant type and (if possible) places in diesel to be updated to expect a typed ID. Essentially the longer term goal here should be to push raw UUIDs out to the very edges of the system with the hope of eventually dropping those. Now, that doesn't mean changing how we _store_ UUIDs in the DB, I strongly believe that should remain unaffected. 

**Some changes we may want to make in this PR based on the above**:
1. It would be nice to be able to directly construct typed IDs for a resource kind (`TypedId::Instance(uuid: UUID)`)
2. It would be nice to have a conversion method from string to typed ID. It could be a `try_from` or a `FromStr` impl like name has. Essentially `instance:<uuid>` -> `TypedId<Instance>` or w/e. We can incrementally update this as more types are ported.

Eventually I want to update `NameOrId` to have a `TypedId` as an arm so we can start supporting this directly from the API.

common/src/typed_uuid.rs
```diff
@@ -0,0 +1,26 @@
...
+        $(
+            #[derive(JsonSchema)]
+            pub enum $kind {}
+
+            impl TypedUuidKind for $kind {
+                #[inline]
+                fn tag() -> TypedUuidTag {
+                    TypedUuidTag::new($tag)
+                }
+            }
```
zephraph, line 19: I don't fully understand this. Why do we need a unique enum for every type of uuid? (I suspect this is just a rust pattern I'm ignorant of, I'd love to learn more)

nexus/db-model/src/switch_interface.rs
```diff
@@ -118,7 +121,7 @@ pub struct LoopbackAddress {
 
 impl LoopbackAddress {
     pub fn new(
-        id: Option<Uuid>,
+        id: Option<TypedUuid<LoopbackAddressKind>>,
```
zephraph, line 124: I feel like `TypedUuid::LoopbackAddress` might be better as an API, even if that causes a lot more noise in the crate.

nexus/db-macros/src/lookup.rs
```diff
@@ -951,6 +997,17 @@ mod test {
...
+        let output = lookup_resource(quote! {
+            name = "Sled",
+            ancestors = [],
+            children = [],
+            lookup_by_name = false,
+            soft_deletes = true,
+            primary_key_columns = [ { column_name = "id", uuid_kind = ::omicron_common::api::SledType } ]
+        })
+        .unwrap();
+        println!("{}", pretty_format(output));
```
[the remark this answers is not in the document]
sunshowers: Yeah, thanks for catching this.

common/src/typed_uuid.rs
sunshowers: The idea behind the PR is to have a marker type which indicates that a UUID represents a particular identity. The marker type can be anything that implements `TypedUuidKind`. So we'd have a type for sled IDs, one for project IDs, etc, etc. Once we have that, we can rely on the compiler to check that we aren't mixing up different kinds of IDs.

Now, note that the marker type does not need to store any data, so a zero-sized type is ideal. One option is an empty struct, e.g. `MyKind(())`, and that's fine. But in our case we don't actually need to ever construct an instance of that type! So, in the interest of making illegal states unrepresentable, we're looking to define a type which _cannot be constructed_.

What kind of type can never be constructed? The [never type (`!`)](https://doc.rust-lang.org/std/primitive.never.html) is one answer, certainly, but in this case we want to define lots of types that are different but cannot be constructed. The answer is that we define an enum without any variants. (It is easy to see that the enum can never be constructed -- if there aren't any variants, how would you specify one?)

This ties into some ideas behind type systems. You might have heard that a struct is called a "product type" and an enum is called a "sum type". One way to see how these names came about is to look at the cardinality of these types--i.e. how many values each type can take.

For example:

```rust
struct MyStruct(u8, bool);

enum MyEnum {
    Integer(u8),
    Boolean(bool),
}
```

How many different values can `MyStruct` and `MyEnum` represent?

* For `MyStruct`, there's two elements, a `u8` (256 elements) and a `bool` (2 elements). The total number of different values `MyStruct` can take is **256 * 2 = 512**.
* For `MyEnum`, the value can either be a `u8` or a `bool`. The total number of different values is **256 + 2 = 258**.

Now consider an "empty" struct and an "empty" enum.

```rust
struct EmptyStruct;

enum EmptyEnum {}
```

So, what's the cardinality of `EmptyStruct`? It's the identity for multiplication, which is **1**. `EmptyStruct` can only take one possible value; it can only be defined one way.

And similarly, the cardinality of `EmptyEnum`, being the identity for addition, is **0**. In other words, `EmptyEnum` can simply never take a value. It must always be zero.

---

There is a downside to this, though, which is that we can't define a UUID that's "either this or that", e.g. "either a sled or a project". For that, we'd need to move beyond marker types -- we'd need to store some kind of value along with the type. Practical use cases for this aren't entirely clear to me, but if you (or someone else) has some we should switch over to that. (It would complicate our APIs a bit, so if we don't need that I'd rather we not have that.)

(Apologies if this was too long-winded and you were aware of most or all of it!)

nexus/db-model/src/switch_interface.rs
sunshowers: Hmm I'm not quite sure what you mean. `TypedUuid::LoopbackAddress` would be an enum variant (runtime representation), not a type parameter (compile-time representation).

common/src/typed_uuid.rs
zephraph: Very instructive, thank you!

sunshowers: I wrote up https://github.com/oxidecomputer/newtype-uuid/issues/7 as a response to the "why marker types" question.

sunshowers:
> To outline how I think about the flow of identifiers in the API:

Adding a tag to our UUID serialization format makes a lot of sense in the long term, but it would require some careful thought. I think it's out of scope for this PR in particular, but I wrote up https://github.com/oxidecomputer/newtype-uuid/issues/8 with some ideas for how to achieve this.

nexus/authz-macros/src/lib.rs
```diff
@@ -160,8 +182,97 @@ struct Input {
...
+//                 &content.cursor().token_stream(),
+//             )?;
+//             let kind_param = parsed.uuid_kind.into_inner();
+//             let full_ty = parse_quote!(::newtype_uuid::TypedUuid::<::omicron_uuid_kinds::#kind_param>);
+//             Ok(Self::UuidKind { full_ty })
+//         } else {
+//             Ok(Self::Standard(input.parse()?))
+//         }
+//     }
+// }
```
smklein, line 272: Dead code?

nexus/db-queries/src/authz/api_resources.rs
```diff
@@ -869,7 +869,8 @@ authz_resource! {
 authz_resource! {
     name = "LoopbackAddress",
     parent = "Fleet",
-    primary_key = Uuid,
+    // TODO: make it easier to specify the typed UUID kind
```
smklein, line 872: Does this still need work? I'm not 100% sure I follow what additional work is needed here

nexus/db-queries/src/db/datastore/switch_interface.rs
```diff
@@ -130,7 +133,7 @@ impl DataStore {
         self.transaction_retry_wrapper("loopback_address_delete")
             .transaction(&conn, |conn| async move {
                 let la = diesel::delete(dsl::loopback_address)
-                    .filter(dsl::id.eq(id))
+                    .filter(dsl::id.eq(to_db_typed_uuid(id)))
```
smklein, line 136: I looked into "what would it take to not need to specify this converter wrapper", and frankly, I think you've made the right call. 

[`.eq`](https://docs.diesel.rs/master/diesel/expression_methods/trait.ExpressionMethods.html#method.eq) requires the argument type to have `T: AsExpression<Self::SqlType>`, which can either be derived (as you've done for the `DbTypedUuid`, or which is implemented generically:

```rust
impl<T, ST> AsExpression<ST> for T
where
  T: Expression<SqlType = ST>,
  ST: SqlType + TypedExpressionType`
```

([source](https://docs.diesel.rs/master/diesel/expression/trait.AsExpression.html#impl-AsExpression%3CST%3E-for-T))

I think this is how "Uuid" works natively here -- it implements `SqlType + TypedExpressionType`.
If we wanted to, we probably *could* implement either `AsExpression` or `SqlType` for `TypedUuid<T>`, but then it forces the UUID registry to depend on Diesel, which would be a bummer.

uuid-kinds/README.adoc
```diff
@@ -0,0 +1,73 @@
+# omicron-uuid-kinds
```
smklein, line 1: Good call to include this documentation explicitly!

uuid-kinds/src/lib.rs
```diff
@@ -0,0 +1,45 @@
...
+// This should generally be an append-only list. Removing items from this list
+// will not break things for now (because newtype-uuid does not currently alter
+// any serialization formats), but it may involve some degree of churn across
+// repos.
+//
+// Please keep this list in alphabetical order.
+
+impl_typed_uuid_kind! {
+    LoopbackAddressKind => "loopback_address",
+    TufRepoKind => "tuf_repo",
```
smklein, line 44: Nitpick: What do you think about dropping the `Kind` suffix? `TypedUuid<FooKind>` seems redundant compared to `TypedUuid<Foo>`. Arguably it's implied by the crate name anyway

smklein (approved):
Damn, it's impressive you got this working through all the macros so quickly! Thanks for kicking this off!

sunshowers:
> Damn, it's impressive you got this working through all the macros so quickly! Thanks for kicking this off!

Haha it was a lot of sweat and tears!

nexus/authz-macros/src/lib.rs
sunshowers: Yeah, thanks for catching this. I switched from using `syn::Parse` to `serde_tokenstream` but forgot to clean this up.

nexus/db-queries/src/authz/api_resources.rs
sunshowers: Nothing else needed here.

nexus/db-queries/src/db/datastore/switch_interface.rs
sunshowers: Yeah, I definitely don't want the UUID registry to depend on Diesel. I actually filed https://github.com/diesel-rs/diesel/discussions/3919, and I still need to get back to that.

uuid-kinds/src/lib.rs
sunshowers: I thought about this -- the problem is that `LoopbackAddress` already exists as a type, and it would be unfortunate for this marker type to be mixed up with the real type.

sunshowers:
Before landing this I'd like to release newtype-uuid v1. I realized over the weekend that trying to bump the major version on it is going to be very painful, requiring a whole bunch of coordination across several repos.

I've asked @hawkw to do an API review before releasing it, and anyone else is welcome to do the same of course: https://github.com/oxidecomputer/newtype-uuid

Cargo.toml
```diff
@@ -627,6 +627,11 @@ branch = "oxide/omicron"
 [patch.crates-io.omicron-workspace-hack]
 path = "workspace-hack"
 
+# Pulls in https://github.com/njaremko/samael/pull/41
+[patch.crates-io.samael]
+git = "https://github.com/oxidecomputer/samael"
+branch = "oxide/omicron"
```
hawkw, line 633: should we be pointing this at a rev instead of a branch, so that the build is reproducible from this commit forever?

sunshowers: That's not part of this commit -- it's a merge artifact. [current main](https://github.com/oxidecomputer/omicron/blob/fedd8c87ccbf17af6251433a122652cf0962464e/Cargo.toml#L616-L619) has it.

sunshowers: (But also -- doesn't `Cargo.lock` ensure reproducibility?)

Issue #4931, opened 2024-01-30
[sled agent] Instance creation will deadlock, stop all future instances from being provisioned, if you give any bad hostnames


smklein:
TL;DR:

- If you create a new instance, containing either an invalid hostname, host_domain, or domain_search_list...
- We do some parsing within `Instance::new`, which will return an error.
- This is actually a Bad Thing, because `Instance::new` holds ownership of an `InstanceTicket`, which, while being dropped, grabs a `std::Mutex` to remove itself from the "Map of managed instances"...
- ... which is particularly nasty, because `Instance::new` is called from a context where the "Map of managed instances" is already locked. This results in a deadlock. This also blocks all future provisions, by preventing access to this instance map.

Context:

- Where we first lock the instances in the "instance manager": https://github.com/oxidecomputer/omicron/blob/62547d22e74c4bb134b8d65e373ace793de7529f/sled-agent/src/instance_manager.rs#L240
- Where we create the "instance ticket": https://github.com/oxidecomputer/omicron/blob/62547d22e74c4bb134b8d65e373ace793de7529f/sled-agent/src/instance_manager.rs#L266-L267
- Where we call "instance::new": https://github.com/oxidecomputer/omicron/blob/62547d22e74c4bb134b8d65e373ace793de7529f/sled-agent/src/instance_manager.rs#L288-L295
- Where we can fail within "Instance::new", triggering the deadlock: https://github.com/oxidecomputer/omicron/blob/62547d22e74c4bb134b8d65e373ace793de7529f/sled-agent/src/instance.rs#L728-L747

bnaecker:
Here is some MDB data to prove that this is indeed a deadlock as laid out above.

We took a core of the sled-agent on dogfood, sled 11 (hostname `BRM42220006`), and pointed `mdb` at it. Looking at the stacks we see:

```
BRM42220006 # mdb core.651
Loading modules: [ libumem.so.1 libnvpair.so.1 libc.so.1 ld.so.1 ]
> $G
C++ symbol demangling enabled
> ::stacks
THREAD           STATE    SOBJ                COUNT
2                PARKED   CV                    122
                 libc.so.1`cond_wait_queue+0x55
                 libc.so.1`__cond_wait+0xba
                 libc.so.1`cond_wait+0x2e
                 libc.so.1`pthread_cond_wait+0x15
                 parking_lot::condvar::Condvar::wait_until_internal::hab79cd95889c5f9b+0x27b
[...]

PR #4936, opened 2024-01-30
[sled-agent] Drop usage of locks from instances to grab fewer locks during drop

smklein:
This PR refactors both `Instance` and `InstanceManager` to a message-passing structure to avoid the need for locking.

We do the following:
- Most of the `Instance` / `InstanceManager` internals have been moved to the `{Instance, InstanceManager}Runner` structure, which is elusively owned by a `tokio::task`.
- The `{Instance,InstanceManager}` sends requests to this task to process ongoing work.
- The `InstanceTicket` has been re-worked -- rather than removing an instance from an inner `BTreeMap`, it sends a request for the `tokio::task` to do this removal. This avoids any need for locking. 

Fixes https://github.com/oxidecomputer/omicron/issues/4931


gjcolombo:
I think the message passing approach is generally righteous, but (if I understand the change correctly) it introduces a new failure mode that we have to be careful of: now all instance-related requests to a sled are handled serially, so if one of those takes a long time, instance management functionality is unavailable for the entire sled for the duration. (I've dreamt for a while of an architecture where there's a message handler task *per instance* that owns all of the objects related to that instance, which I think would mitigate that issue somewhat, but I haven't thought at all about how to get there from here.)

---

I only realized what was going on with this bug after most of the discussion of it had been had (mea maxima culpa), so maybe this idea was discussed and rejected, but is it possible to avoid recursively acquiring the `InstanceTicket` lock without changing the overall architecture? IIUC, the problem is that

- `ensure_registered` takes the instance manager lock
- `ensure_registered` creates a full-fledged `InstanceTicket` with a reference to that instance manager
  - when this is dropped the `Drop` impl will try to take the manager lock
- `ensure_registered` calls `Instance::new`, which fails, so the ticket doesn't get transferred to the Instance
- control leaves the scope where the ticket was created; this causes it to be dropped while the manager lock is still held

If that's right, would it make sense to have `Instance::new` take an `Arc<InstanceManagerInternal>` and then call `InstanceTicket::new` only when it knows instance creation has succeeded? ISTM that would avoid this particular deadlock without changing much else about `InstanceManager` (though we should inspect it to be sure that this sort of issue can't arise anywhere else...). This might provide a way to staunch the bleeding without having to worry about any "can't operate on instances in parallel" issues.

---

One other note: this is the sort of change where you might find https://github.com/oxidecomputer/omicron-stress especially useful for testing (probably worth extending it to pass bogus hostnames sometimes). LMK if I can help get you started with the tools there.

hawkw:
> think the message passing approach is generally righteous, but (if I understand the change correctly) it introduces a new failure mode that we have to be careful of: now all instance-related requests to a sled are handled serially, so if one of those takes a long time, instance management functionality is unavailable for the entire sled for the duration.

Depending on the scope of the lock, this may also be the case when using a `Mutex` rather than a queue.  If operations hold a lock for the entire duration of the operation, the mutex also forces those operations to be performed serially...

N.B. that I'm not at all familiar with the previous code, and would have to look closer to see whether this was actually happening based on the way locks were actually used. But, it's worth bringing up...

smklein:
Appreciate the feedback, @gjcolombo and @hawkw ! I've gone ahead and updated this PR with a new commit that, I *believe* avoids blocking by using message passing through to `instance.rs` as well. 

> I've dreamt for a while of an architecture where there's a message handler task per instance that owns all of the objects related to that instance, which I think would mitigate that issue somewhat, but I haven't thought at all about how to get there from here.

Yeah, I basically just tried implementing this! I realized that for it to not block the "instance manager", I needed to use "request pipelining", which basically takes the form:

- InstanceManager exposes an API like `async fn add_external_ip() -> Result<...>`
- Internally, this creates a `tokio::sync::oneshot::Sender` that gets sent to a background task. The caller waits on `tokio::sync::oneshot::Receiver`, if they want to get that result back out.
- The InstanceManager's background task has a handler for `add_external_ip()` -- and if it wants to make a call a specific instance, it **explicitly passes the `tokio::sync::oneshot::Sender` as an argument**.
- This means that the InstanceManager's background task is no longer blocked on anything -- the `Instance`-specific background task can then respond directly to whomever accessed the `InstanceManager` API originally.

> One other note: this is the sort of change where you might find https://github.com/oxidecomputer/omicron-stress especially useful for testing (probably worth extending it to pass bogus hostnames sometimes). LMK if I can help get you started with the tools there.

Yes I'm *extremely* interested and would appreciate some of your time to get a little tutorial for this PR!

> Depending on the scope of the lock, this may also be the case when using a Mutex rather than a queue. If operations hold a lock for the entire duration of the operation, the mutex also forces those operations to be performed serially...

Totally agreed! I've **mostly** removed mutexes in this portion of the codebase now that we're using message passing, but I think the tasks are "blocked" in similar spots -- we just now "process a message for an instance" when we previously would have "locked that instance".

sled-agent/src/instance.rs
```diff
@@ -186,9 +198,117 @@ struct PropolisSetup {
...
+// the state of Propolis for updates.
+//
+// This task communicates with the "InstanceRunner" task to report status.
+struct InstanceMonitorRunner {
+    client: Arc<PropolisClient>,
+    tx_monitor: mpsc::Sender<InstanceMonitorRequest>,
+}
+
+impl InstanceMonitorRunner {
+    async fn run(self) -> Result<(), anyhow::Error> {
```
smklein, line 246: See "monitor_state_task" below for how this was modified.

It's similar to the original, but rather than accessing the "instance" state via `Arc`, we chat with it on the `tx_monitor/rx_monitor` channel.

sled-agent/src/instance.rs
```diff
@@ -186,9 +198,117 @@ struct PropolisSetup {
...
     log: Logger,
 
+    // A oneshot which, when closed, will cause the "run" loop to terminate
+    // without processing any subsequent messages.
+    //
+    // We need the tx side of this thing to be alive to prevent "terminate_rx"
+    // from shutting down early.
+    #[allow(dead_code)]
+    terminate_tx: oneshot::Sender<()>,
+    terminate_rx: oneshot::Receiver<()>,
```
smklein, line 302: I swear there must be a simpler solution here, but this is basically trying to cover the need for:

- I have a "runner" object, which is basically a loop + `tokio::select`
- I want to let the runner terminate itself . No one else needs this ability externally.

It feels like it must be possible to do this with a smaller / more lightweight object, but both ends of the oneshot also seemed very straightforward.

sled-agent/src/instance.rs
```diff
@@ -186,9 +198,117 @@ struct PropolisSetup {
...
+            // Now that we have the response from Propolis' HTTP server, we
+            // forward that to the InstanceRunner.
+            //
+            // It will decide the new state, provide that info to Nexus,
+            // and possibly identify if we should terminate.
+            let (tx, rx) = oneshot::channel();
+            self.tx_monitor
+                .send(InstanceMonitorRequest::Update { state: response, tx })
+                .await
+                .map_err(|e| anyhow!(e))?;
```
smklein, line 272: This bit is new -- we used to do this in-line, acting on an `Arc<Instance>`, but now the instance is owned by the `InstanceRunner` exclusively, so we ask it to do work for us.

sled-agent/src/instance.rs
```diff
@@ -237,7 +357,92 @@ struct InstanceInner {
     instance_ticket: InstanceTicket,
 }
 
-impl InstanceInner {
+impl InstanceRunner {
+    async fn run(mut self) -> Result<(), Error> {
+        loop {
+            tokio::select! {
+                biased;
```
smklein, line 364: The priority here is:

1. If we should terminate, do that.
2. Otherwise, keep an eye on any changes self-reported by the VMM.
3. Finally, handle user-requests.

sled-agent/src/instance.rs
```diff
@@ -518,7 +724,7 @@ impl InstanceInner {
 
             // Ensure the instance is removed from the instance manager's table
             // so that a new instance can take its place.
-            self.instance_ticket.terminate();
+            self.instance_ticket.deregister();
```
smklein, line 727: Too many things called "terminate", so I renamed the de-registration from the `InstanceManager` BTreeMap to "deregister".

sled-agent/src/instance.rs
```diff
@@ -755,8 +960,18 @@ impl Instance {
             }
         }
 
-        let instance = InstanceInner {
+        let (terminate_tx, terminate_rx) = oneshot::channel();
+        let (tx, rx) = mpsc::channel(QUEUE_SIZE);
+        let (tx_monitor, rx_monitor) = mpsc::channel(1);
```
smklein, line 965: The `InstanceMonitorRunner` waits for the `InstanceRunner` to respond before processing the next state change anyway -- it needs to know how to react -- so I think it's fine that this is only "size of one".

sled-agent/src/instance.rs
hawkw: A `oneshot` is [actually quite lightweight](https://docs.rs/tokio/latest/src/tokio/sync/oneshot.rs.html)  The other option in `tokio::sync` would be an `Arc<tokio::sync::Notify>`, which can be used to wake a task _without_ sending data, but since a `Notify` can be used to wake a task multiple times, the implementation is actually much more complex.

In this specific case, though, since the termination is only sent from "inside" the task, you don't really need to wake up the task --- the task is (AFAICT) always running when it terminates itself. So, as far as I can tell, you should be able to just replace this with `terminating: bool`, and then just put
```rust
if self.terminating {
    return Ok(());
}
```
at the top of the `loop` in `InstanceRunner::run`, before the `tokio::select!`. You'll hit that every time the loop iterates, before you enter the `select!` and possibly yield.

hawkw: ...actually, rather than writing `if self.terminating`, you could just change the loop to `while !self.terminating`.

smklein: Ah, that's a good point! I used a `oneshot` to make sure I wouldn't miss this as a "wakeup call", but it's true that if a task is only flagging this for itself, this can be much simpler.

hawkw: Although, hmm. It is worth noting that if the terminating oneshot is sent, and then the task `.await`s something else that isn't ready, it will have woken itself and the `select!` will fall through into the terminating branch. This isn't the case with a `bool`. So, if you represent termination with a bool, set the flag, and then `.await` something, you're not going to "see" the termination until the thing you await wakes you.

If we don't `.await` anything after setting the flag, that's not necessarily an issue. But, it's worth calling out...

smklein: Just to make sure I'm grokking this:

- If we use a boolean that we check before the `select` call, the `InstanceRunner` can self-terminate, but only after one of the select arms has "fully completed"
- If we use a oneshot, we can bail-out before any other select arms "fully complete"

For example, something like:


```rust
while !self.terminating {
  tokio::select! {
    let a = self.foo() => ...
    let b = self.bar() => ...
  }
}

async fn foo(&mut self) {
  do_stuff.await;

  self.terminating = true;

  // Other `select` arms could continue being queried concurrently while we're awaiting this.
  // If we closed a oneshot, however, we'd exit before getting to query this.
  do_stuff.await;
}
```

hawkw: @smklein yup, that's correct.

andrewjstone (approved):
Looks great! It's awesome to see this change, despite what forced it.

gjcolombo (approved):
All this makes sense to me--thanks for putting this together!

sled-agent/src/instance_manager.rs
```diff
@@ -352,152 +677,141 @@ impl InstanceManager {
...
         &self,
+        tx: oneshot::Sender<Result<(), Error>>,
         instance_id: Uuid,
         ip: &InstanceExternalIpBody,
     ) -> Result<(), Error> {
-        let instance = {
-            let instances = self.inner.instances.lock().unwrap();
-            instances.get(&instance_id).map(|(_id, v)| v.clone())
-        };
+        let instance = { self.instances.get(&instance_id).map(|(_id, v)| v) };
```
hawkw, line 759: nit, take it or leave it: it looks like there are a few places where we do a `get` and ignore the ID field of the tuple, do you think it's worth having a method (`self.get_instance()` maybe?) that just does that?

sled-agent/src/instance.rs
```diff
@@ -186,9 +198,117 @@ struct PropolisSetup {
...
+            // known to Propolis.
+            let response = self
+                .client
+                .instance_state_monitor()
+                .body(propolis_client::types::InstanceStateMonitorRequest {
+                    gen,
+                })
+                .send()
+                .await
+                .map_err(|e| anyhow!(e))?
```
hawkw, line 259: style nit, not important: is this necessary?  `?` will call `anyhow::Error::from(e)` automagically, so I don't think we need to explicitly wrap it as an `anyhow` error:
```suggestion
                .await?
```

sled-agent/src/instance.rs
hawkw, line 272: style nit, not important: i think this can just be
```suggestion
                .await?;
```

sled-agent/src/instance.rs
```diff
@@ -186,9 +198,117 @@ struct PropolisSetup {
...
+            //
+            // It will decide the new state, provide that info to Nexus,
+            // and possibly identify if we should terminate.
+            let (tx, rx) = oneshot::channel();
+            self.tx_monitor
+                .send(InstanceMonitorRequest::Update { state: response, tx })
+                .await
+                .map_err(|e| anyhow!(e))?;
+
+            if let Reaction::Terminate = rx.await.map_err(|e| anyhow!(e))? {
```
hawkw, line 274: and i think this can probably also just be 
```suggestion
            if let Reaction::Terminate = rx.await? {
```

sled-agent/src/instance.rs
```diff
@@ -237,7 +357,92 @@ struct InstanceInner {
...
+                // Handle messages from our own "Monitor the VMM" task.
+                request = self.rx_monitor.recv() => {
+                    use InstanceMonitorRequest::*;
+                    match request {
+                        Some(Update { state, tx }) => {
+                            let observed = ObservedPropolisState::new(
+                                self.state.instance(),
+                                &state,
+                            );
+                            let reaction = self.observe_state(&observed).await?;
```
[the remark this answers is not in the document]
hawkw: :+1: i agree with Greg here.

sled-agent/src/instance.rs
smklein: If that's the case, I *think* the boolean approach here is okay. We will only set this value on ourselves, and completing "our dispatched operation" is okay. We *do* want to e.g. post our state to Nexus if we can, before terminating the runner, which is another asynchronous operation.

sled-agent/src/instance_manager.rs
```diff
@@ -199,6 +222,316 @@ impl InstanceManager {
...
+        rx.await?
+    }
+}
+
+// Most requests that can be sent to the "InstanceManagerRunner" task.
+//
+// These messages are sent by "InstanceManager"'s interface, and processed by
+// the runner task.
+//
+// By convention, responses are sent on the "tx" oneshot.
```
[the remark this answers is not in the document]
smklein: Yup, definitely inspired by your code using mpsc + oneshot. I'm using `tx` because of how universal the replies are, so I don't *think* it's ambiguous right now.

sled-agent/src/instance_manager.rs
```diff
@@ -199,6 +222,316 @@ impl InstanceManager {
...
+    terminate_tx: mpsc::UnboundedSender<InstanceDeregisterRequest>,
+    terminate_rx: mpsc::UnboundedReceiver<InstanceDeregisterRequest>,
+
+    nexus_client: NexusClientWithResolver,
+
+    // TODO: If we held an object representing an enum of "Created OR Running"
+    // instance, we could avoid the methods within "instance.rs" that panic
+    // if the Propolis client hasn't been initialized.
+    /// A mapping from a Sled Agent "Instance ID" to ("Propolis ID", [Instance]).
+    instances: BTreeMap<Uuid, (Uuid, Instance)>,
```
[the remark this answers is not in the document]
smklein: Agreed, though I might save this one until after this PR. This type should probably flow from the HTTP layer onwards.

sled-agent/src/instance_manager.rs
```diff
@@ -199,6 +222,316 @@ impl InstanceManager {
...
+                // If anyone has made a request to remove an instance from our
+                // state tracking, we do so before processing subsequent
+                // requests. This ensure that "add, remove, add" of the same
+                // instance always works.
+                request = self.terminate_rx.recv() => {
+                    match request {
+                        Some(request) => {
+                            self.instances.remove(&request.id);
+                        }
+                        None => break,
```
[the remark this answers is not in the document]
smklein: Logging added for all of these "closed channel" cases.

sled-agent/src/instance_manager.rs
```diff
@@ -90,6 +82,22 @@ pub(crate) struct InstanceManagerServices {
...
 }
 
+// Describes the internals of the "InstanceManager", though most of the
+// instance manager's state exists within the "InstanceManagerRunner" structure.
+struct InstanceManagerInternal {
+    log: Logger,
+    tx: mpsc::Sender<InstanceManagerRequest>,
+    // NOTE: Arguably, this field could be "owned" by the InstanceManagerRunner.
+    // It was not moved there, and the reservoir functions were not converted to
+    // use the message-apssing interface (see: "InstanceManagerRequest") because
```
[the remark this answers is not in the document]
smklein: Fixed

sled-agent/src/instance.rs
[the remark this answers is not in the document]
smklein: This is mostly true -- I'll prevent the `Result` from propagating through types, since we're not using it -- but as a warning, `terminate_inner` appears to be unwrapping rather than throwing errors. It should probably be made more protected against "things going wrong during instance teardown".

Either way - updated so that the main runner returns `()`, rather than a result.

sled-agent/src/instance_manager.rs
smklein: Added that method, and using it -- thanks for the suggestion!

sled-agent/src/instance.rs
smklein: Good call, updated

sled-agent/src/instance.rs
smklein: Updated

sled-agent/src/instance.rs
```diff
@@ -237,7 +357,92 @@ struct InstanceInner {
...
+                            self.publish_state_to_nexus().await;
+
+                            // NOTE: If we fail to send here, the
+                            // InstanceMonitorRunner has stopped running for some
+                            // reason. We may want to handle that case.
+                            let _ = tx.send(reaction);
```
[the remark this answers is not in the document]
smklein: Done in 0b060f6fc6ead023bd1a91c1f0fac691257630ae - both here and in the instance manager, I at least log if the client disconnects.

smklein:
Thanks everyone for the feedback! I believe I've incorporated all the recommendations

hawkw (approved):
@smklein hopefully you weren't still waiting for my :thumbsup: --- looks good to me!

smklein:
> @smklein hopefully you weren't still waiting for my 👍 --- looks good to me!

No worries, I was waiting for the next release to be cut -- but I appreciate the approval regardless!

PR #4937, opened 2024-01-30
Use hostname for Baseboard::Pc::identifier field

andrewjstone:
Doing this allows easier testbed setup and usage since we identify new sleds to add via their baseboard. In case we can't get a hostname we revert to using a UUID, as in the prior version of code.


sled-hardware/Cargo.toml
```diff
@@ -10,6 +10,7 @@ anyhow.workspace = true
 camino.workspace = true
 cfg-if.workspace = true
 futures.workspace = true
+gethostname = "0.4.3"
```
davepacheco, line 13: I'm somewhat shocked that this isn't built in.  Would it be better to use to use [nix's implementation](https://docs.rs/nix/latest/nix/unistd/fn.gethostname.html) (since we're already using `nix` I think)?

Either way, I'd think we'd want to add this to the workspace-level deps.

davepacheco:
I don't know enough to know if this change would create any problems.  I think all the stuff I've done explicitly ignores non-Oxide hardware so I don't think that'll care.  But it seems like this punts the uniqueness requirement to the deployer, who needs to ensure that hostnames are set properly and unique.  I wonder if @jmpesp or @smklein has an opinion here?

jclulow:
I think you probably want to try and use the SMBIOS UUID for this?

```
 $ smbios -t 1
ID    SIZE TYPE
1     79   SMB_TYPE_SYSTEM (type 1) (system information)

  Manufacturer: TYAN
  Product: B8036T65AV4E24HR
  Version: empty
  Serial Number: TCK2118F0004

  UUID: 00008036-4354-4245-3244-4b3033303031
  Wake-Up Event: 0x6 (power switch)
  SKU Number: empty
  Family: empty
```

It is often a unique property of the physical or virtual system on which you are running.

rmustacc:
@jclulow we can't today due to it missing from Propolis (https://github.com/oxidecomputer/propolis/issues/628). Which means that Falcon and co. don't have it.

jclulow:
Ah!  Very well.

sled-hardware/Cargo.toml
andrewjstone: Good idea. I'll use nix's impl!

andrewjstone:
> I don't know enough to know if this change would create any problems. I think all the stuff I've done explicitly ignores non-Oxide hardware so I don't think that'll care. But it seems like this punts the uniqueness requirement to the deployer, who needs to ensure that hostnames are set properly and unique. I wonder if @jmpesp or @smklein has an opinion here?

I tagged @jmpesp because he does the same thing for his petfood cluster in northern america.

sled-hardware/Cargo.toml
andrewjstone: Turns out we are not using nix. It is a transitive dep though.

jmpesp (approved):
:dog: :canned_food: approved

PR #4938, opened 2024-01-30
Validate instance hostnames in the Nexus API

bnaecker:
- Add the `Hostname` type to the external API, which checks that a string conforms to RFC 1035 (and others) describing valid Internet hostnames. Add some tests for that.
- Use the new type everywhere in Nexus and its integration tests, though we still serialize to a String in the database and when passing the instance information to the sled-agent. The sled-agent does its own validation internally through OPTE.
- Add a few regression tests for creating instances with known-bad hostnames.


bnaecker:
I'm beefing up the tests here, I think these are not actually testing what I want.

bnaecker:
Ok folks, this is ready for a review.

common/src/api/external/mod.rs
```diff
@@ -696,6 +696,101 @@ impl TryFrom<i64> for Generation {
...
+                        .to_string(),
+                ),
+                ..Default::default()
+            })),
+            instance_type: Some(schemars::schema::SingleOrVec::Single(
+                Box::new(schemars::schema::InstanceType::String),
+            )),
+            string: Some(Box::new(schemars::schema::StringValidation {
+                max_length: Some(HOSTNAME_MAX_LEN),
+                min_length: None,
```
jmpesp, line 786: Submitting `"hostname": ""` was something that would trigger the problem: should this be non-zero?

bnaecker: Done in 49609cfde

nexus/db-queries/src/db/datastore/instance.rs
```diff
@@ -97,7 +97,11 @@ impl From<InstanceAndActiveVmm> for omicron_common::api::external::Instance {
             project_id: value.instance.project_id,
             ncpus: value.instance.ncpus.into(),
             memory: value.instance.memory.into(),
-            hostname: value.instance.hostname,
+            hostname: value
+                .instance
+                .hostname
+                .parse()
+                .expect("found invalid hostname in the database"),
```
jmpesp, line 104: What if there are existing hostnames in the DB that went in prior to this validation?

bnaecker: It's a good question. The change that revealed the [deadlock in the sled-agent](https://github.com/oxidecomputer/omicron/issues/4931) was added in Oct 2023. I am not positive, but it seems unlikely that there are any such instances still in existence.

If we're concerned about it, or want to be more cautious (reasonable!), I can change this `From` implementation to be a `TryFrom` and bubble up a 500 rather than panicking Nexus.

jmpesp: > If we're concerned about it, or want to be more cautious (reasonable!), I can change this From implementation to be a TryFrom and bubble up a 500 rather than panicking Nexus.

I think this is the right move - isn't it guaranteed that any instances that had a non-compliant hostname would be in state Failed? We should make sure that those can be deleted, even with the `TryFrom` bubbling up the 500.

bnaecker: >  isn't it guaranteed that any instances that had a non-compliant hostname would be in state Failed?

I think that's right, for instances created after Oct 2023. The sled-agent would have deadlocked, and the request from Nexus to it timed out. The result would be a failed instance.

I'll change this now, thanks.

smklein (commented):
I think the current build failures are real;

> end-to-end-tests/src/instance_launch.rs:67

looks like it's using a hostname. Might not be built by default within the workspace?

smklein (approved):

bnaecker:
Thanks Sean, nice catch. Fixed.

openapi/nexus.json
```diff
@@ -11975,6 +11975,13 @@
           "start_time"
         ]
       },
+      "Hostname": {
+        "title": "An RFC-1035-compliant hostname",
+        "description": "A hostname identifies a host on a network, and is usually a dot-delimited sequence of labels, where each label contains only letters, digits, or the hyphen. See RFCs 1035 an 952 for more details.",
```
karencfv, line 11980:
```suggestion
        "description": "A hostname identifies a host on a network, and is usually a dot-delimited sequence of labels, where each label contains only letters, digits, or the hyphen. See RFCs 1035 and 952 for more details.",
```

bnaecker: Nice catch, thanks @karencfv

karencfv:
Heya @bnaecker, I think you forgot to generate the openapi/nexus.json file in this last commit https://github.com/oxidecomputer/omicron/pull/4938/commits/0b2a761301c69817253b67ebae6b5e73498aea21. This would make the tests fail I think

nexus/db-queries/src/db/datastore/instance.rs
bnaecker: This (rightfully) opened a pretty big can of worms for me. The general question is: what do we do with existing invalid hostnames?

The first question I think we need to understand is whether we _have_ invalid hostnames. If we do not, then writing a large amount of code to handle it is not useful.

If we do, there are several ways we can handle it:

- Some kind of data migration during the upgrade. The big pro here is very little new Rust code in the control plane itself. The big con is a lot of manual work, some of which may not even be possible in SQL. For example, I'm not sure how to even detect invalid hostnames in SQL, since the regex engine needs to support lookarounds to work correctly. If it doesn't, we'd need to run more complicated SQL, or potentially run some Rust code directly. We also need to figure out what to do with invalid hostnames.
- Detect and automatically rewrite invalid hostnames when reading them from the database. For example, we could try to validate the hostname every time we read the record, and if that _fails_ we could simply use the instance's `name` field, which is a valid hostname. The pro here is that it's pretty simple to write. The con is obvious: it's confusing and janky, since the instance itself would get a hostname from DHCP that doesn't match the one it was created with.
- The most heavy-weight but careful solution would be to handle both valid and invalid hostnames for some migration period. On writing new instance records, we only insert known-valid hostnames. On read, we would extract the hostname into an enum, something like:

    ```rust
    enum MaybeHostname {
        Valid(Hostname),
        Invalid(String),
    }
    ```

    If we're doing something like fetching or listing the instance for the public API, we could serialize this into a less-rigid view type, which retains the hostname as an unvalidated `String`. If we're doing basically any other operation, most saliently, sending the instance to the sled-agent for it to start, we first validate the hostname.

    The pro here is safety and robustness, at the obvious expense of a large amount of code. We'd need to carry that for some time, but presumably we could remove it again when we're confident there are no extant instances with invalid hostnames.

smklein: Are we sure that the hostname validation being done by the sled agent is as strict as what you're adding in this PR?

I think it's okay if "no", but just throwing it out there, if your PR rejects a name like `my_host`, and the sled agent would accept it, then the likelihood of invalid hostnames being used seems a lot higher.

My two cents: I like your "Detect and automatically rewrite invalid hostnames" approach, with one small tweak: we could just throw an error, rather than re-writing anything.

For example:

If someone created an instance with an invalid hostname, we could parse the hostname in Nexus during "instance start", and throw an error before sending anything to the sled. Something like "Hostname validation has become more strict, and this name is no longer valid. Please update the hostname or create a new instance".

IMO the burden here isn't terrible -- after an update, we require all instances to stop anyway, so we can still get away with migrations like this. That basically leaves the hostnames alone, but prevents the "bad ones in the field" from being used.

bnaecker: > Are we sure that the hostname validation being done by the sled agent is as strict as what you're adding in this PR?
>
> I think it's okay if "no", but just throwing it out there, if your PR rejects a name like my_host, and the sled agent      would accept it, then the likelihood of invalid hostnames being used seems a lot higher.

OPTE's hostname validation is less strict than this PR's. 


> My two cents: I like your "Detect and automatically rewrite invalid hostnames" approach, with one small tweak: we could just throw an error, rather than re-writing anything.

I agree this is a practical approach. It seems like we might be able to get away with checking this in just a few places, such as `Nexus::instance_start()` and maybe `Nexus::instance_reboot()`.

bnaecker: So I think I'm going to take this approach, in the interest of urgency. One issue is that the check cannot be when we read instance records from the database, as that would prevent ever operating on those instances at all, including to delete them. Instead, we have to:

- Check hostnames when we start an instance. Luckily, I think there is only one path here, `Nexus::instance_ensure_registered()`, which is called from both the instance start and migration sagas.
- Deserialize hostnames as strings, without validation, from the DB, and send them as strings out the Nexus API. That's to support listing instances and ideally deleting them, even if they have invalid hostnames. We would need to change the instance view type to use a validated hostname in the future, if it matters.

nexus/src/app/instance.rs
```diff
@@ -1010,6 +1011,23 @@ impl super::Nexus {
     ) -> Result<(), Error> {
         opctx.authorize(authz::Action::Modify, authz_instance).await?;
 
+        // Check that the hostname is valid.
+        //
+        // TODO-cleanup: This can be removed when we are confident that no
+        // instances exist prior to the addition of strict hostname validation
+        // in the API.
+        if db_instance.hostname.parse::<Hostname>().is_err() {
```
bnaecker, line 1019: @smklein This is the shorter-term fix we discussed offline. This is code-path is common to instance start and instance migrate, though because udpates bounce all instances, the former is the relevant one.

It would be nice to beef up the rest of the code (mostly around `nexus_db_model::Instance`) to require a validated hostname everywhere. I'll make an issue if we're happy landing this as-is, and we can tackle that once we're confident no invalid hostnames still exist.

smklein: Rad, as long as this is visible to customers, this seems like a very appropriate fix to me.

One other option would be validation early in the "instance start" saga -- that's where this `db_instance` comes from anyway, I think, since it's a parameter to that saga -- but this works too, and is about halfway through the saga.

bnaecker: Yeah, either would work. I wanted to make sure I didn't miss any code path that calls this, though I don't see any. I'm running a test now to exercise this, manually dorking with the instance record in the DB.

bnaecker:
Ok, I've run a quick test locally to confirm that this does what I expect. It's a bit nasty, hold your hat.

I launched the control plane as usual on my dev machine, using `omicron-package {package,install}`. Once things were up, I created all the necessary preconditions: quotas, IP pool linked to the silo, project, Alpine-based image and disk. I then made an instance, which with the new API validation requires a valid hostname:

```
bnaecker@shale : ~/omicron $ oxide instance list --project $PROJECT
Instance {
    description: "my inst",
    hostname: "test",
    id: 0d027324-bce4-4498-b00d-eea57ac0fa64,
    memory: ByteCount(
        1073741824,
    ),
    name: Name(
        "test",
    ),
    ncpus: InstanceCpuCount(
        2,
    ),
    project_id: 23afd905-24e3-4222-9e37-47836a18697d,
    run_state: Stopped,
    time_created: 2024-02-01T17:49:09.254629Z,
    time_modified: 2024-02-01T17:49:09.254629Z,
    time_run_state_updated: 2024-02-01T17:49:09.254629Z,
}
```

Note that I created the instance but did not start it. Now, I went into the DB and manually dorked with the hostname, to make it invalid. This should simulate exactly the case of a user having previously created an instance with an invalid hostname, and the applying the offline update past this Omicron commit.

```
bnaecker@shale : ~/omicron $ ./tools/cockroachdb/bin/cockroach sql --url postgresql://[fd00:1122:3344:101::6]:32221/omicron?sslmode=disable --insecure
#
# Welcome to the CockroachDB SQL shell.
# All statements must be terminated by a semicolon.
# To exit, type: \q.
#
# Client version: CockroachDB OSS v21.1.10 (x86_64-pc-solaris2.11, built 2021/10/12 22:52:33, go1.16.5)
# Server version: CockroachDB OSS v22.1.9-dirty (x86_64-pc-solaris2.11, built 2022/10/26 21:17:46, go1.17.13)
# Cluster ID: cf550c23-a61a-46f6-88de-c63703c144af
#
# Enter \? for a brief introduction.
#
root@[fd00:1122:3344:101::6]:32221/omicron> select hostname from instance where id = '0d027324-bce4-4498-b00d-eea57ac0fa64';
  hostname
------------
  test
(1 row)

Time: 2ms total (execution 1ms / network 0ms)
root@[fd00:1122:3344:101::6]:32221/omicron> UPDATE instance SET hostname = 'well_shit' WHERE id = '0d027324-bce4-4498-b00d-eea57ac0fa64';
UPDATE 1

Time: 8ms total (execution 8ms / network 0ms)

root@[fd00:1122:3344:101::6]:32221/omicron> select hostname from instance where id = '0d027324-bce4-4498-b00d-eea57ac0fa64';
  hostname
-------------
  well_shit
(1 row)

Time: 1ms total (execution 1ms / network 0ms)

root@[fd00:1122:3344:101::6]:32221/omicron> ^D
bnaecker@shale : ~/omicron $ oxide instance list --project $PROJECT
Instance {
    description: "my inst",
    hostname: "well_shit",
    id: 0d027324-bce4-4498-b00d-eea57ac0fa64,
    memory: ByteCount(
        1073741824,
    ),
    name: Name(
        "test",
    ),
    ncpus: InstanceCpuCount(
        2,
    ),
    project_id: 23afd905-24e3-4222-9e37-47836a18697d,
    run_state: Stopped,
    time_created: 2024-02-01T17:49:09.254629Z,
    time_modified: 2024-02-01T17:49:09.254629Z,
    time_run_state_updated: 2024-02-01T17:49:09.254629Z,
}
```

Now let's start the instance:

```
bnaecker@shale : ~/omicron $ oxide instance start --instance 0d027324-bce4-4498-b00d-eea57ac0fa64
error
Error Response: status: 400 Bad Request; headers: {"content-type": "application/json", "x-request-id": "2fa558fa-e79f-4557-9972-f5168bd69730", "content-length": "322", "date": "Thu, 01 Feb 2024 17:52:36 GMT"}; value: Error { error_code: Some("InvalidRequest"), message: "The instance hostname 'well_shit' is no longer valid. To access the data on its disks, this instance must be deleted, and a new one created with the relevant disks. The new hostname will be validated at that time.", request_id: "2fa558fa-e79f-4557-9972-f5168bd69730" }
```

So, we're at least preventing users from torpedoing an entire sled, by triggering #4931. We're also providing a reasonably clear set of steps to recover the instance's data, if needed.

nexus/db-queries/src/db/datastore/instance.rs
bnaecker: Fixed in [4546ed3](https://github.com/oxidecomputer/omicron/pull/4938/commits/4546ed3b7af7131ed0fef2d206e7ee511538bc9d), see testing notes below.

smklein (approved):
New approach LGTM too!

bnaecker:
Failed the OpenAPI check yet again, since the new commits return a String as the hostname when fetching an instance.

jmpesp (approved):
:shipit:

PR #4940, opened 2024-01-31
Integrate Package cache

smklein:
This PR pulls in the changes introduced in https://github.com/oxidecomputer/omicron-package/pull/60, which has been published as a breaking version of omicron-package.

These changes include caching, a non-TUI logfile, and camino paths.

Caching can optionally be disabled with:

```bash
omicron-package package --disable-cache
```

Though the cache is now enabled by default


steveklabnik (approved):
Looks good to me!

PR #4852, opened 2024-01-19
BFD support

rcgoodfellow:
This is a staging PR and should most likely be pulled into 
- #4822.

Here we simply pass through BFD commands to the underlying `mgd` daemons on the switches. No attempt is made to add BFD to the database schema or persist BFD information. As that would likely conflict with #4822.

The purpose of the RFD is to set up the scaffolding and API interfaces for BFD to work end-to-end, and to do some interim testing without the benefit of persistence.

Depends on

- oxidecomputer/maghemite#67

PR #4943, opened 2024-01-31
Hide internal silos in utilization

### LINKED ISSUE #4708: Non-discoverable silos should have zero quotas and be excluded from utilization query
Currently all hidden silos show up in the console (because they are returned in the `/v1/system/utilization/silos` result set). This doesn't seem right especially because list silos never displayed these silos.

It is not a showstopper but we probably want to follow up with:
- setting these silos to a zero quota
- excluding them from utilization queries

I can see the argument of allowing the operator to see if any resources are being consumed by those silos (especially because they are NOT prohibited from creating VMs and disks). We may need to think about a way to make them less prominent.

zephraph:
Fixes #4708.

I've updated the `/v1/system/utilization/silos` endpoint to only return non-discoverable silos if they have a quota set. I believe the `default-silo` _does_ get a quota set currently which is non-ideal, but that should be the only one that shows up on the list. 

I need specific eyes on the migration b/c I've never written a view migration before.


nexus/db-model/src/utilization.rs
```diff
@@ -9,6 +9,7 @@ use uuid::Uuid;
 pub struct SiloUtilization {
     pub silo_id: Uuid,
     pub silo_name: Name,
+    pub silo_discoverable: bool,
```
david-crespo, line 12: Strictly speaking, while this is fine to add, you don't actually need this line, right? In the query you're using it in SQL directly.

david-crespo (approved):

nexus/db-model/src/schema.rs
```diff
@@ -13,7 +13,7 @@ use omicron_common::api::external::SemverVersion;
 ///
 /// This should be updated whenever the schema is changed. For more details,
 /// refer to: schema/crdb/README.adoc
-pub const SCHEMA_VERSION: SemverVersion = SemverVersion::new(30, 0, 0);
+pub const SCHEMA_VERSION: SemverVersion = SemverVersion::new(31, 0, 0);
```
david-crespo, line 16: You need to be 32 so https://github.com/oxidecomputer/omicron/pull/4852 can be 31

augustuswm:
Logically this makes sense to me. I don't know what specifically cockroachdb supports but if this is similar to postgres you may need to use `CREATE OR REPLACE VIEW` instead of `ALTER VIEW`

smklein (approved):

augustuswm (approved):

zephraph:
@augustuswm pointed out using `CREATE OR REPLACE VIEW` so I changed it to that instead. Something I'm not sure about is what [CRDB's docs](https://www.cockroachlabs.com/docs/v23.2/create-view#parameters) say about this:

> In order to replace an existing view, the new view must have the same columns as the existing view, or more. If the new view has additional columns, the old columns must be a prefix of the new columns. For example, if the existing view has columns a, b, the new view can have an additional column c, but must have columns a, b as a prefix. In this case, CREATE OR REPLACE VIEW myview (a, b, c) would be allowed, but CREATE OR REPLACE VIEW myview (b, a, c) would not.

nexus/db-model/src/utilization.rs
zephraph: Hmmm, perhaps. I thought this struct is what generated the type to use for the `dsl` but that might actually be from the `table!` call. Yeah, regardless, it shouldn't matter.

zephraph:
I've updated the PR w/ a comment in the migration about what's being added and bumped the schema version an extra number to avoid clashing with #4852

augustuswm:
Looks like based on the failure it might need a drop / create instead of a replace.

zephraph:
> Looks like based on the failure it might need a drop / create instead of a replace.

I reordered it such that the new column is at the end of the view and that causes it to pass. Seems like view migrations are order sensitive (which is what the docs I quoted above implied, I just didn't know if that meant for the select statement too).

PR #4950, opened 2024-02-01
[wicketd] replace snafu with thiserror

sunshowers:
This is currently the only use of snafu in omicron (outside of transitive
deps), and I figure it is simple enough that we can replace it with thiserror,
reducing use of a direct dependency.

This was the first time I'd encountered snafu so I read a bit about it. As far
as I understand, the main benefit of snafu is that it pushes you heavily
towards an error-per-module pattern. However, thiserror does permit that
pattern as well, and in practice it is only a little more verbose than snafu to
do right (`map_err` vs `context`, though `snafu` introduces new types that
aren't in the source code like `IoSnafu` and `ParseSnafu`).

I'd love to get your take on it though @andrewjstone.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10291,7 +10291,6 @@ dependencies = [
  "sled-hardware",
  "slog",
  "slog-dtrace",
- "snafu",
  "subprocess",
  "tar",
  "thiserror",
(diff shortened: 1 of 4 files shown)


andrewjstone (approved):
I forget exactly why I went down this road. I don't see any real benefit to snafu at this point. I agree it's better to keep things standardized with thiserror. Thanks for the fix!

Issue #4886, opened 2024-01-24
add sled: planner support for deploying Nexus


davepacheco:
The blueprint planner needs to support deploying new Nexus zones as needed.

For real systems, the policy would generally be to make sure there are 3 Nexus zones in the fleet (per the compile-time constant used by RSS).  To test this more easily, we discussed (on yesterday's update call) having a policy of "put one Nexus on every sled".

To deploy a new Nexus, we need to construct a new `OmicronZoneConfig` with `OmicronZoneType::Nexus`, which will require an external IP from the services IP pool.  (Do these need to be allocated separately somehow or can we use the same approach that we use for underlay IPs, where their presence in the current target blueprint is what defines allocated vs. not?)  As I understand it, with oxidecomputer/omicron#4715, we don't need to do anything else to configure boundary services to route this IP appropriately.  This will happen automatically some time after the zone is deployed and things are reflected in inventory.  We might want to trigger an inventory collection after this execution.

PR #4959, opened 2024-02-01
blueprint planner: add nexus on every sled

jgallagher:
_Most_ of the work in this PR is to give `BlueprintBuilder` the ability to add Nexus to a sled at all. A very minor bit to enable testing is a change to the planner to attempt to add a Nexus zone to every sled.

I think (?) we probably don't want to merge this into `main` with the planner making that decision, but I'm opening this for review because (a) maybe I'm wrong, since we're driving all of this by hand, (b) the bulk of the PR is ready for review, and (c) we want a TUF repo with this change to test on `madrid`.


jgallagher:
Whoops, missed some tests. Working on fixing that and rebasing onto `main`, please hold the reviews...

jgallagher:
> Whoops, missed some tests. Working on fixing that and rebasing onto `main`, please hold the reviews...

Ok, rebasing done and tests may be fixed (running locally too, but still slow). Clear for reviews; if there are more test breaks I won't rebase again.

nexus/deployment/src/blueprint_builder.rs
```diff
@@ -329,6 +455,118 @@ impl<'a> BlueprintBuilder<'a> {
...
+        if has_nexus {
+            return Ok(Ensure::NotNeeded);
+        }
+
+        // Whether Nexus should use TLS and what the external DNS servers it
+        // should use are currently provided at rack-setup time, and should be
+        // consistent across all Nexus instances. We'll assume we can copy them
+        // from any other Nexus zone in our parent blueprint.
+        //
+        // TODO-correctness Once these properties can be changed by a rack
```
andrewjstone, line 480: I'm not sure these settings need to be part of policy. They are really customer configuration, and while they could live in policy I'm not sure if it makes sense to put them there or to keep them in a separate customer configuration.

The way I've been thinking of the reconfigurator is as a control system. The policy, inventory, current database state, and current blueprint all feed into the planner to generate a new blueprint which then triggers changes which feeback into the planner. With this, we can have any number of different inputs, and we don't have to force them all into the policy, inventory, or blueprint. However, I'm unsure if that is helpful or if it simplifies things to move items used by the planner into either policy, blueprint, or inventory.

With all that being said, I don't think we can rely on the prior blueprint for this information. It can be configured by an operator and should be read by the planner directly from the database in its "raw" form, or as part of the updated policy. We'll likely also have to update other existing nexus zones, right?

You should also open an issue for this so we don't lose track of it.

nexus/deployment/src/blueprint_builder.rs
```diff
@@ -146,16 +174,114 @@ impl<'a> BlueprintBuilder<'a> {
...
+        let mut used_external_ips: HashSet<IpAddr> = HashSet::new();
+        let mut used_macs: HashSet<MacAddr> = HashSet::new();
+
+        for zone_config in parent_blueprint.omicron_zones.values() {
+            for z in &zone_config.zones {
+                if let OmicronZoneType::Nexus { nic, .. } = &z.zone_type {
+                    match nic.ip {
+                        IpAddr::V4(ip) => {
+                            if !existing_nexus_v4_ips.insert(ip) {
+                                bail!("duplicate Nexus NIC IP: {ip}");
```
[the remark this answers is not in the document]
jgallagher: It's definitely bad! The two ways this could happen are:

* A bug in blueprint generation where we added a Nexus with a dupe IP (preventing this is the whole point of `existing_nexus_v4_ips`, etc)
* An initial blueprint created from a collection contains dupes; this shouldn't be possible in practice given how RSS works

I don't think we'll ever see either of these, and I was tempted to make this a panic. But panicking nexus on "we should never see this" seems bad, hence the current `bail!`.

nexus/deployment/src/blueprint_builder.rs
jgallagher: > I'm not sure these settings need to be part of policy. They are really customer configuration, and while they could live in policy I'm not sure if it makes sense to put them there or to keep them in a separate customer configuration.

I've been thinking of the policy as the customer configuration. It's currently "the sleds that should be part of the rack" and (as of this PR) "the service external IP pool", both of which are provided at RSS and only modified by an operator. But I don't feel super strongly about what we call it and whether parts should live in something else.

> With all that being said, I don't think we can rely on the prior blueprint for this information. It can be configured by an operator and should be read by the planner directly from the database in its "raw" form, or as part of the updated policy. We'll likely also have to update other existing nexus zones, right?

I agree in the limit, but it can't be configured by an operator today, so there's nothing to read from the database. Definitely fine on opening an issue; I think moving the storage of these into the DB (and eventually making them operator-adjustable) shouldn't be rolled into this PR. Are you okay with this as-is for now?

andrewjstone: Sorry, I didn't mean to give you the impression I wanted you to make changes related to this on this PR. I was just riffing on design for the future. This is good as is for now.

andrewjstone (approved):

nexus/deployment/src/planner.rs
```diff
@@ -133,7 +133,30 @@ impl<'a> Planner<'a> {
                 }
             }
 
-            if ncrucibles_added > 0 {
+            // TODO XXX To test Nexus zone addition, attempt to ensure every
+            // sled has a Nexus zone. This is NOT a sensible policy, and should
+            // be removed before this lands on `main`.
```
davepacheco, line 138: (discussed offline) we should probably do _something_ about this, even if it's just to change the comment.

nexus/deployment/src/planner.rs
```diff
@@ -133,7 +133,30 @@ impl<'a> Planner<'a> {
...
+            //
+            // Adding a new Nexus currently requires the parent blueprint to
+            // have had at least one Nexus. By trying to ensure a Nexus on every
```
davepacheco, line 141: I'm confused about why this paragraph and the next are here, but maybe it's obviated by the policy changes we talked about offline.

nexus/deployment/src/blueprint_builder.rs
```diff
@@ -81,6 +99,16 @@ pub struct BlueprintBuilder<'a> {
     zones_in_service: BTreeSet<Uuid>,
     creator: String,
     comments: Vec<String>,
+
+    // These fields emulate how RSS choses addresses for zone NICs.
```
davepacheco, line 103:
```suggestion
    // These fields emulate how RSS chooses addresses for zone NICs.
```

nexus/deployment/src/blueprint_builder.rs
```diff
@@ -146,16 +174,114 @@ impl<'a> BlueprintBuilder<'a> {
...
+        // in a way that expects the addition to reuse resources from the
+        // removal; we will won't want to attempt to reuse resources from a zone
+        // until we know it's been fully removed.
+        let mut existing_nexus_v4_ips: HashSet<Ipv4Addr> = HashSet::new();
+        let mut existing_nexus_v6_ips: HashSet<Ipv6Addr> = HashSet::new();
+        let mut used_external_ips: HashSet<IpAddr> = HashSet::new();
+        let mut used_macs: HashSet<MacAddr> = HashSet::new();
+
+        for zone_config in parent_blueprint.omicron_zones.values() {
+            for z in &zone_config.zones {
```
davepacheco, line 216: I think you could use `parent_blueprint.all_omicron_zones()` here.

nexus/deployment/src/blueprint_builder.rs
```diff
@@ -329,6 +455,118 @@ impl<'a> BlueprintBuilder<'a> {
         Ok(Ensure::Added)
     }
 
+    pub fn sled_ensure_zone_nexus(
+        &mut self,
+        sled_id: Uuid,
+    ) -> Result<Ensure, Error> {
+        // If there's already a Nexus zone on this sled, do nothing.
```
davepacheco, line 462: I don't think we want this.  This is different from NTP (where two can't really coexist on the same sled) and Crucible (same, for the same zpool).  It's very useful in dev/test to have multiple Nexus instances on one sled.

nexus/deployment/src/blueprint_builder.rs
davepacheco: Yeah, I think we've been using "policy" to mean any kind of configuration that would be input to planning -- i.e., anything other than the current state or previous blueprint.  Agreed that in an ideal world, this would be configurable by the customer, and so it would be stored somewhere in the database, and we'd use that, and that absent that, we should just look at what other Nexus instances have set here.

nexus/deployment/src/blueprint_builder.rs
```diff
@@ -329,6 +455,118 @@ impl<'a> BlueprintBuilder<'a> {
...
+            },
+        };
+        self.sled_add_zone(sled_id, zone)?;
+        Ok(Ensure::Added)
+    }
+
+    fn random_mac(&mut self) -> MacAddr {
+        let mut mac = MacAddr::random_system();
+
+        // TODO-performance if the size of `used_macs` starts to approach some
```
davepacheco, line 559: Feel free to ignore this.  It's not a well fleshed-out thought.  I worry about using randomness in the planner because I think that'll make it harder to test, and easier to introduce thrashing behavior (where two Nexuses -- or even one! -- fight over an arbitrary decision).  You could as well argue the other way around, that by using randomness we'll be forced to encounter and deal with thrashing.  I guess it's just a gut feel that randomness in the planner is going to make our lives harder.

nexus/deployment/src/planner.rs
jgallagher: Yep. Trying to briefly summarize the offline discussion, as I'm working on changes here that are nontrivial:

* Instead of adding a Nexus to every sled, we'll try to ensure there are 3 Nexuses. (For testing on madrid, we can easily build a local version that wants 4 instead of 3.)
* If it needs to add new Nexuses, the planner should prioritize one of the sleds with the minimum number of Nexus zones. If there are more sleds than Nexuses (the typical case), it'll pick sleds with no Nexus. If there are fewer than 3 sleds (common in dev/test), it'll be forced to assign multiple Nexus instances to one sled.

nexus/deployment/src/blueprint_builder.rs
jgallagher: Hmm! It's a fair concern. My gut feeling is that there are going to be other places where we want "just pick a random one" (e.g., which of the N sleds should get an extra Nexus/cockroach/etc.), and if we do something _too_ deterministic (e.g., sort by sled ID) we'll make our lives unnecessarily difficult by picking the same sled for too many things.

Maybe a bad idea: what if the planner seeded an RNG with something deterministic, maybe the parent blueprint ID? Then it has deterministic behavior for any single given parent, but we can still freely use randomness.

andrewjstone: > Maybe a bad idea: what if the planner seeded an RNG with something deterministic, maybe the parent blueprint ID? Then it has deterministic behavior for any single given parent, but we can still freely use randomness.

I am a fan of this! It's basically how proptest and all the other related tools give you replay determinism.

davepacheco: > we'll make our lives unnecessarily difficult by picking the same sled for too many things

I don't think randomness really solves that problem, but admittedly it might be good enough for a while.  I think I'd propose an approach where we narrow the set of sleds to which ones are acceptable for a choice, then pick deterministically among those.  That deterministic process could initially be "pick the lexicographically-first sled id", but it could also be "pick the one with fewer zones on it" or "pick the one with the least CPU/disk/memory already allocated by the control plane" or "assign each zone type a score based on how much load we expect it to generate and then pick the sled with the lowest total score".  That is, we can do something simple (pick the first) or sophisticated (if we find we need to deal better with distribution) but it's still deterministic.

This is a lot more complicated than generating MAC addresses, though.  Again, feel free to ignore this but I mention it because the sequential allocator seems no worse here and eliminates the only current source of randomness in the planner (I think).

> Maybe a bad idea: what if the planner seeded an RNG with something deterministic, maybe the parent blueprint ID? Then it has deterministic behavior for any single given parent, but we can still freely use randomness.

I'm not sure.  I could see that going either way too.  It doesn't help test determinism unless we ensure that we use deterministic UUID generation too.  I looked into that briefly while writing tests but it seemed like it'd require a lot of surgery.

sunshowers: > Maybe a bad idea: what if the planner seeded an RNG with something deterministic, maybe the parent blueprint ID? Then it has deterministic behavior for any single given parent, but we can still freely use randomness.

I think a deterministic source of randomness stored in the blueprint would be cool, and would probably end up getting used for a lot of things.

> It doesn't help test determinism unless we ensure that we use deterministic UUID generation too. I looked into that briefly while writing tests but it seemed like it'd require a lot of surgery.

Deterministic UUID generation would be pretty useful as well I think, effectively requiring that a source of UUIDs be passed down.

jgallagher: > This is a lot more complicated than generating MAC addresses, though. Again, feel free to ignore this but I mention it because the sequential allocator seems no worse here and eliminates the only current source of randomness in the planner (I think).

Oh, /facepalm, MACs are not the only source of random output of blueprints, specifically because of UUIDs (which you and @sunshowers mentioned - thank you). We already generate the NTP and Crucible zone IDs, and this PR adds two more UUID generations: the Nexus zone ID and the ID for the Nexus zone NIC.

I think it would not be a huge lift to add deterministic random to `BlueprintBuilder` and use it for both UUIDs and MACs. Probably the trickiest bit will be ensuring future changes continue to use the deterministic random, but probably we can catch that in unit tests? I'll give it a whirl and see what it looks like.

sunshowers: Yeah, thinking about it, it's a bit hard to enforce the invariant that people don't use the `Uuid::new_v4` method long-term. One way is probably to not enable the `v4` feature, but that seems excessive (using the system RNG is completely fine in most other spots, just not here).

I guess a unit test could generate blueprints with a fixed seed several times, and ensure that the generated UUIDs stay the same.

sunshowers: Another idea really is to just ban `new_v4` via clippy: https://rust-lang.github.io/rust-clippy/master/#/disallowed_methods

davepacheco: I think it's fine if we want to have a deterministic source of randomness for blueprints.  At the same time, I think randomness is not a great solution to any of the problems we've mentioned, or at least not better than an alternative (like sequential MAC address allocation), so I'm not sure it's worth the effort.

I think my aversion to randomness here comes down to a few things:

- For automated testing, it's more annoying to verify that the planner generated the blueprint we expected because we can't just hardcode the values we know it will choose. Deterministic-random doesn't _necessarily_ help here but it could.  But we already wrote a higher-level `diff` thing that can tell if two blueprints are different, ignoring things like the "id" and "time_created" (which will always be different anyway).  So this isn't that big a deal.
- When we get to making this more autonomous, it will become important for Nexus instances to be able to look at some other instance's plan and say "is that okay?"  That _might_ be the same as "is that equivalent to the thing I would do?" -- or it might not.  I alluded to this earlier.  Choosing deterministically makes it easier for us to punt on this for longer (because two Nexus instances will tend to create the same plans), but it doesn't really solve the problem.  Also: we're talking about hypotheticals on hypotheticals at this point and I really don't think we want to spend a lot of time on such things now.
- Probably the biggest thing: during general operation and especially in support contexts: it means the system's behavior is less predictable.  I can imagine wanting to set a customer's expectations -- or even our own -- about what would happen when the planner is enabled or when given some input.  With randomness, even deterministic randomness, that's harder.
- I've often reached for random approaches in the past and found that they work okay for a while, but in the limit / at scale, the variance winds up being a problem.

Honestly, none of these is that compelling.  If there were a compelling reason to use randomness here, I'd say go for it.  But I don't see any reason _to_ use randomness, either, aside from being slightly quicker to implement than sequential.  Either way, I don't think it's worth investing a lot of time on this part right now.

jgallagher: I went through this and basically copied what RSS was doing, so I ended up with sequential OPTE IP address generation and random MAC generation. It's a small change to switch over to sequential MAC generation; I can put that in this PR so I'm not introducing a new kind of randomness, and we can punt on whether we should have deterministic UUIDs for a followup.

jgallagher:
I think I've addressed most of the comments above. The changes are substantial enough that this warrants a second look, particularly 029dd2bd4ec9fa148f483389356a689db1e8fef6 which switches from "allocate one Nexus on every sled" to "allocate a total of `Policy.target_nexus_zone_count` Nexus zones, prioritizing sleds that have the fewest Nexus zones currently".

nexus/deployment/src/blueprint_builder.rs
jgallagher: [9748940](https://github.com/oxidecomputer/omicron/pull/4959/commits/97489400c6221be65d165ee400ad264dc9dfbda7) switches to sequential MAC allocation. I'm not going to mark this conversation resolved because I don't want to hide it, but I think it's resolved as far as this PR is concerned; please correct me if that's wrong!

davepacheco:
Note: this will fix #4886, I assume.

nexus/types/src/deployment.rs
```diff
@@ -55,6 +55,13 @@ pub struct Policy {
     /// ranges specified by the IP pool for externally-visible control plane
     /// services (e.g., external DNS, Nexus, boundary NTP)
     pub service_ip_pool_ranges: Vec<IpRange>,
+
+    /// desired total number of deployed Nexus zones
+    // This number probably doesn't make sense alone once we get to a world
+    // where we're deploying multiple _versions_ of Nexus and need to ensure a
+    // minimum number in service at any given time, but for now all we know how
+    // to do is deploy a fixed number of all-the-same-version.
```
davepacheco, line 63: Sort of?  I would assume in that world this reflects the target number of zones in-service at whatever version is specified in the policy.  So if this were "3", I'd expect us to either deploy 1, remove 1, deploy 1, remove 1, etc. to maintain 3 in service at all times; or else deploy 3, remove 3.  Either way I think this parameter will still make sense.

nexus/deployment/src/planner.rs
```diff
@@ -168,6 +162,126 @@ impl<'a> Planner<'a> {
...
+
+        'outer: for _ in 0..nexus_to_add {
+            // `sleds_by_num_nexus` is sorted by key already, and we want to
+            // pick from the lowest-numbered bin. We can just loop over its
+            // keys, expecting to stop on the first iteration, with the only
+            // exception being when we've removed all the sleds from a bin.
+            for (&num_nexus, sleds) in sleds_by_num_nexus.iter_mut() {
+                // TODO choose more smartly than effectively ordering by
+                // sled_id? See
+                // https://github.com/oxidecomputer/omicron/pull/4959#discussion_r1476795735.
```
davepacheco, line 228: I'd probably strike this

nexus/deployment/src/planner.rs
```diff
@@ -168,6 +162,126 @@ impl<'a> Planner<'a> {
...
+                // to it" case). That's fine.
+                sleds_to_change.insert(sled_id, num_nexus + 1);
+
+                // Put this sled back in our map, but now with one more Nexus.
+                sleds_by_num_nexus
+                    .entry(num_nexus + 1)
+                    .or_default()
+                    .push(sled_id);
+
+                continue 'outer;
```
davepacheco, line 245: I'm not sure if it's any better, but it seems like you could avoid the loop label with just a `break` here and removing the `unreachable!` below.

nexus/deployment/src/planner.rs
```diff
@@ -168,6 +162,126 @@ impl<'a> Planner<'a> {
...
+    fn ensure_correct_number_of_nexus_zones(
+        &mut self,
+        sleds_ineligible_for_services: &BTreeSet<Uuid>,
+    ) -> Result<(), Error> {
+        // Bin every sled by the number of Nexus zones it currently has while
+        // counting the total number of Nexus zones.
+        let mut num_total_nexus = 0;
+        let mut sleds_by_num_nexus: BTreeMap<usize, Vec<Uuid>> =
+            BTreeMap::new();
+        for &sled_id in self.policy.sleds.keys() {
```
davepacheco, line 181: This is all totally fine.  I'm just thinking out loud: would this be a lot simpler using https://doc.rust-lang.org/std/collections/struct.BinaryHeap.html?  I was figuring you'd throw all the sleds in a heap ordered by the num_nexus count and while you're still looking for another one to deploy, pull it out of the heap, deploy the zone, and put it back in with the new value.  (That assumes a min-heap.)

I'm not asking you to rewrite this -- I was just curious if you'd considered this or if you thought it could be simpler?

davepacheco (approved):

nexus/deployment/src/planner.rs
jgallagher: Because we don't want to do it, or because as we add more services and smarts to the planner it will be obvious that we need to do it? (E.g., if we duplicated this function for "cockroach" and "clickhouse", we'd end up ordering the sleds within each bin the same way.)

nexus/deployment/src/planner.rs
jgallagher: That feels slightly worse to me because it turns what really should be `unreachable!` (i.e., iterating all the way through this inner loop without finding a sled) into a silent fall through, which I think would result in an infinite loop.

nexus/deployment/src/planner.rs
jgallagher: I did think about that, and thought it had a couple downsides:

* There's some small machinery to create a struct that holds both the ID and the Nexus count that sorts as we expect to store in the binary heap.
* The standard binary heap doesn't (AFAICT?) have a way to say "give me all the tied-for-lowest-count values", only "give me one of the tied-for-lowest-count values". If we want to pick from the sleds with some specific criteria (e.g., total control plane services on the sled) we're going to want the former, which is what this whole binning thing is for.

That second point is building toward a hypothetical future, and is overkill / moot if we never do the smarter choice thing. But I think it's fairly likely we'll want to do something like that?

nexus/deployment/src/planner.rs
davepacheco: Most of the time I feel like it's more distracting than useful to mention alternatives / possible future work in the source, unless it's really useful for understanding the current implementation.  If we want to keep track of this, I think we're better off filing an issue.  This is obviously subjective and it's just a suggestion.

sled-agent/Cargo.toml
```diff
@@ -44,6 +44,7 @@ libc.workspace = true
 macaddr.workspace = true
 mg-admin-client.workspace = true
 nexus-client.workspace = true
+nexus-deployment.workspace = true
```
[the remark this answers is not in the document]
jgallagher: Moved (and killed sled-agent's dependency on nexus-deployment) in d33215495

nexus/deployment/src/planner.rs
jgallagher: I reworded this to just briefly describe what it's currently doing

jgallagher:
> Note: this will fix #4886, I assume.

Not quite - we still need a change on the executor side to insert db records for the external IP and NIC required by Nexus. I'm working on that now.

PR #4961, opened 2024-02-01
nix flake: Manage clickhouse, Cockroachdb, Dendrite-Stub, and Maghemite deps

hawkw:
Currently, running Omicron's test suite requires Clickhouse,
Cockroachdb, Dendrite-stub, and Maghemite binaries to be present in
`./out`. These are downloaded by a set of scripts in `tools/`, such as
`tools/ci_download_clickhouse`, et cetera. This branch updates the Nix
flake to handle downloading these dependencies as part of the Nix
configuration. This way, Nix users will not need to re-run these scripts
when the dependencies on these binaries change, as the flake is
automatically updated by `direnv`. The flake wraps the `clickhouse`,
`clockroach`, `mgd`, `dpd`, and `swadm` binaries using
`autoPatchelfHook`, to ensure that dynamically loaded libraries will
work on NixOS. Then, the binaries are symlinked into the `out` directory
identically to how the `tools/ci_download_${foo}` scripts work.

After this change, I can now successfully use `omicron-dev run-all` on
my NixOS machine! The contents of my `./out` directory looks like this:

```console
$ tree ./out
out
├── clickhouse
│   ├── clickhouse ⇒ /nix/store/p0a6hx0l27v6d4n1x1ddy6fp48q2gw39-clickhouse/bin/clickhouse
│   └── config.xml ⇒ /nix/store/01qhirxkns7x8rwh0d87hw47a3gc7hkg-clickhouse/etc/config.xml
├── cockroachdb
│   └── bin ⇒ /nix/store/adzvkw7d373md98msa2p4ri2miirndbv-cockroachdb/bin
├── dendrite-stub ⇒ /nix/store/84fs1l2njmc2v1hkhjjk7kyspncriqnh-dendrite-stub
└── mgd ⇒ /nix/store/rr23ga4m8yckqvmccc7p9jwqy326a8hb-mgd
```


faithanalog (approved):
yup that's some nix. works on my non-NixOS machine as well.

sunshowers:
Pretty sure clickhouse just needs to be on the PATH, though CI would tell you for certain.

sunshowers:
I do feel like it's important to use the right version of clickhouse, and I'm not sure what version gets pulled in via Nix.

hawkw:
> I do feel like it's important to use the right version of clickhouse, and I'm not sure what version gets pulled in via Nix.

Unfortunately, since nixpkgs doesn't have old versions of packages, there's not a nice way to get the "right" version of clickhouse. We'd have to write our own Clickhouse derivation that downloads a binary and patches it the same way nixpkgs does. Which...we could do, but it might be better left to a follow-up.

hawkw:
@sunshowers  & @faithanalog, I kept hacking on this branch a bit and I've now also added code for downloading the Maghemite and Dendrite OpenAPI from Buildomat as part of the flake, as well. Let me know what you think? 

I'm also happy to pull that change out into a separate branch, if that's preferred.

sunshowers:
We discussed this in `#oxide-control-plane` yesterday, and came to the conclusion that the clickhouse version divergence between the pinned version in omicron, and what's available in nixpkgs, is already causing issues. @hawkw said she'll keep hacking on this.

faithanalog:
OMFG ok the steam-run hack is beautiful but also it doesn't work out of the box on my gentoo sadly

```
error: a 'i686-linux' with features {} is required to build '/nix/store/13kvr5dymmrmxccdfni1irni4l9lwpll-libidn2-2.3.4.drv', but I am a 'x86_64-linux' with features {benchmark, big-parallel, nixos-test, uid-range}
```

I'm getting a cache-miss on the libidn2 binary or something i guess

hawkw:
> OMFG ok the steam-run hack is beautiful but also it doesn't work out of the box on my gentoo sadly
> 
> ```
> error: a 'i686-linux' with features {} is required to build '/nix/store/13kvr5dymmrmxccdfni1irni4l9lwpll-libidn2-2.3.4.drv', but I am a 'x86_64-linux' with features {benchmark, big-parallel, nixos-test, uid-range}
> ```
> 
> I'm getting a cache-miss on the 32-bit libs or something i guess

Ugh, that's unfortunate. I believe `steam-run` is just kind of a wrapper around `buildFHSEnv`, so I can try to figure out a way to get clickhouse to work without requiring 32-bit libs.

hawkw:
The definition of `steam-run` is here: https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/games/steam/fhsenv.nix

We probably don't need *all* those libs to run clickhouse anyway...

sunshowers:
Seems like we may be able to copy and adapt this: https://github.com/NixOS/nixpkgs/blob/nixos-23.05/pkgs/games/steam/fhsenv.nix#L310

hawkw:
i'm kind of obsessed with this file:
image

hawkw:
I also took a crack at trying `nix-alien` to handle the use of dynamically linked libraries automagically, but it seems that it doesn't support binaries that do their own `dlopen`ing...I'm gonna continue trying with the `buildFHSEnv` approach.

hawkw:
Okay, so it turns out that all I need to do to get it working without `steam-run` on _my_ system is to stick `clickhouse` inside a `buildFHSEnv`. We don't even need to stick any other libs in that `FHSEnv`; as it seems like clickhouse just uses `dlopen` to load libc stuff? @faithanalog, can you confirm that 0037c1cf63c104069a577eb96db87e3ae2fb52b3 works on your system as well?

sunshowers:
This is awesome!

If we add a new binary in the future, is there some place we can put documentation about how to update the flake to handle it? Or is it just a matter of consulting either of you?

faithanalog:
- nix flake evaluates fine
- `cargo build --release` works; binaries are all linking against /nix paths so i know it's got the environment set up right.
- `cargo nextest r` works
- `cargo check` works

all without even having to run the scripts to install builder prereqs. love it

faithanalog:
> This is awesome!
> 
> If we add a new binary in the future, is there some place we can put documentation about how to update the flake to handle it? Or is it just a matter of consulting either of you?

its kind of annoying. this is one of the biggest painpoints of nix. the way it works is basically that you update the source URL. then you run a nix build. it'll fail because your sha256 hash is wrong and give you the new hash. you delete the old one in the .nix file and copy the new one in. and then you're good to go. 

i think there's actually a command that automates this but i dont remember what it is

sunshowers:
OK -- I think since it's not part of CI for now it won't affect devs not using Nix, but we may have to document this: both adding a new version of a tool, and adding a new tool.

hawkw:
> its kind of annoying. this is one of the biggest painpoints of nix. the way it works is basically that you update the source URL. then you run a nix build. it'll fail because your sha256 hash is wrong but it'll print the new hash for you. you delete the old one in the .nix file and copy the new one in. and then you're good to go.

For the Maghemite and Dendrite-stub dependencies (and their OpenAPI definitions), we actually use the SHAs from the `dendrite_stub_checksums` and `maghemite_checksums` files in `tools/`; so those don't require touching the flake in order to update. I'd like to do this for the Clickhouse and Cockroachdb deps, as well, but those currently use MD5s rather than SHA256 for some reason. Ideally, we could replace the MD5 hashes with SHA256s, and the flake can then automatically consume those versions as inputs without requiring any work from non-Nix users.

faithanalog:
https://github.com/oxidecomputer/omicron/pull/4961/commits/644f46af3b5909f1dd5e8d0dd3a23b19e26c47bb breaks the build by commenting out `LIBCLANG_PATH`:

```
  --- stderr
  thread 'main' panicked at /home/vi/.cargo/registry/src/index.crates.io-6f17d22bba15001f/bindgen-0.69.2/lib.rs:620:31:
  Unable to find libclang: "the `libclang` shared library at /usr/lib/llvm/17/lib64/libclang.so.17.0.6 could not be opened: libstdc++.so.6: cannot open shared object file: No such file or directory"
```

Bringing it back fixes the build:

```
diff --git a/flake.nix b/flake.nix
index 728739ff7..5eb7e1c4a 100644
--- a/flake.nix
+++ b/flake.nix
@@ -384,7 +384,7 @@
 
             name = "omicron";
             DEP_PQ_LIBDIRS = "${pkgs.postgresql.lib}/lib";
-            # LIBCLANG_PATH = "${libclang.lib}/lib";
+            LIBCLANG_PATH = "${pkgs.libclang.lib}/lib";
             OPENSSL_DIR = "${pkgs.openssl.dev}";
             OPENSSL_LIB_DIR = "${pkgs.openssl.out}/lib";

```

Other than that, everything stills works on my machines and I've convinced myself I understand the nix code, so it seems good to me

hawkw:
> [644f46a](https://github.com/oxidecomputer/omicron/commit/644f46af3b5909f1dd5e8d0dd3a23b19e26c47bb) breaks the build by commenting out `LIBCLANG_PATH`:

oops!

faithanalog (approved):
basically yeah this seems good. the manual update bumps for the stuff with md5s is annoying but improving that can be a thing for a different PR.

faithanalog:
oh one sec actually i think it should be `pkgs.libclang.lib`, not `libclang.lib`. because stuff moved around

hawkw:
> oh one sec actually i think it should be `pkgs.libclang.lib`, not `libclang.lib`. because stuff moved around

good catch --- i changed it to not have `with pkgs;` because i wanted to call them `clickhouse` and `cockroachdb` without getting the ones from nixpkgs.

sunshowers:
On my Pop OS 22.04 system, I'm seeing with this PR:

```
% cargo xtask clippy
    Finished dev [unoptimized + debuginfo] target(s) in 0.17s
     Running `target/debug/xtask clippy`
running: "/nix/store/6m89why8xnnak39marjim5fza38fvpmg-cargo-1.74.1-x86_64-unknown-linux-gnu/bin/cargo" "clippy" "--all-targets" "--" "--deny" "warnings" "--allow" "clippy::style"
    Checking omicron-test-utils v0.1.0 (/home/rain/dev/oxide/omicron/test-utils)
    Checking sled-agent-client v0.1.0 (/home/rain/dev/oxide/omicron/clients/sled-agent-client)
    Checking illumos-utils v0.1.0 (/home/rain/dev/oxide/omicron/illumos-utils)
    Checking internal-dns v0.1.0 (/home/rain/dev/oxide/omicron/internal-dns)
    Checking oximeter v0.1.0 (/home/rain/dev/oxide/omicron/oximeter/oximeter)
    Checking key-manager v0.1.0 (/home/rain/dev/oxide/omicron/key-manager)
   Compiling ddm-admin-client v0.1.0 (/home/rain/dev/oxide/omicron/clients/ddm-admin-client)
   Compiling dpd-client v0.1.0 (/home/rain/dev/oxide/omicron/clients/dpd-client)
    Checking ipcc v0.1.0 (/home/rain/dev/oxide/omicron/ipcc)
   Compiling mg-admin-client v0.1.0 (/home/rain/dev/oxide/omicron/clients/mg-admin-client)
error: failed to run custom build command for `ddm-admin-client v0.1.0 (/home/rain/dev/oxide/omicron/clients/ddm-admin-client)`

Caused by:
  process didn't exit successfully: `/home/rain/dev/oxide/omicron/target/debug/build/ddm-admin-client-df450aa6b594d553/build-script-build` (exit status: 127)
  --- stderr
  /home/rain/dev/oxide/omicron/target/debug/build/ddm-admin-client-df450aa6b594d553/build-script-build: error while loading shared libraries: libssl.so.3: cannot open shared object file: No such file or directory
warning: build failed, waiting for other jobs to finish...
error: failed to run custom build command for `dpd-client v0.1.0 (/home/rain/dev/oxide/omicron/clients/dpd-client)`

Caused by:
  process didn't exit successfully: `/home/rain/dev/oxide/omicron/target/debug/build/dpd-client-703c93ada12b9250/build-script-build` (exit status: 127)
  --- stderr
  /home/rain/dev/oxide/omicron/target/debug/build/dpd-client-703c93ada12b9250/build-script-build: error while loading shared libraries: libssl.so.3: cannot open shared object file: No such file or directory
error: failed to run custom build command for `mg-admin-client v0.1.0 (/home/rain/dev/oxide/omicron/clients/mg-admin-client)`

Caused by:
  process didn't exit successfully: `/home/rain/dev/oxide/omicron/target/debug/build/mg-admin-client-a278af5536c4f804/build-script-build` (exit status: 127)
  --- stderr
  /home/rain/dev/oxide/omicron/target/debug/build/mg-admin-client-a278af5536c4f804/build-script-build: error while loading shared libraries: libssl.so.3: cannot open shared object file: No such file or directory
Error: clippy failed: exit status: 101
```

sunshowers:
hmm:

```
% ldd /home/rain/dev/oxide/omicron/target/debug/build/dpd-client-703c93ada12b9250/build-script-build
        linux-vdso.so.1 (0x00007fffcf367000)
        libssl.so.3 => not found
        libcrypto.so.3 => not found
        libgcc_s.so.1 => /nix/store/bncqz1jnd3mkssg2rbx3qgqy55bi0m7g-xgcc-13.2.0-libgcc/lib/libgcc_s.so.1 (0x0000778d2c553000)
        libm.so.6 => /nix/store/7jiqcrg061xi5clniy7z5pvkc4jiaqav-glibc-2.38-27/lib/libm.so.6 (0x0000778d2c471000)
        libc.so.6 => /nix/store/7jiqcrg061xi5clniy7z5pvkc4jiaqav-glibc-2.38-27/lib/libc.so.6 (0x0000778d2b217000)
        /nix/store/7jiqcrg061xi5clniy7z5pvkc4jiaqav-glibc-2.38-27/lib/ld-linux-x86-64.so.2 => /nix/store/7jiqcrg061xi5clniy7z5pvkc4jiaqav-glibc-2.38-27/lib64/ld-linux-x86-64.so.2 (0x0000778d2c57a000)
```

sunshowers:
Interesting:

```
% ls -al $OPENSSL_LIB_DIR 
total 6000
dr-xr-xr-x 1 root root     134 Dec 31  1969 .
dr-xr-xr-x 1 root root      12 Dec 31  1969 ..
dr-xr-xr-x 1 root root      80 Dec 31  1969 engines-3
lrwxrwxrwx 4 root root      14 Dec 31  1969 libcrypto.so -> libcrypto.so.3
-r-xr-xr-x 2 root root 5320112 Dec 31  1969 libcrypto.so.3
lrwxrwxrwx 4 root root      11 Dec 31  1969 libssl.so -> libssl.so.3
-r-xr-xr-x 2 root root  812312 Dec 31  1969 libssl.so.3
dr-xr-xr-x 1 root root      18 Dec 31  1969 ossl-modules
```

So libssl exists but the dynamic loader isn't able to see it?

faithanalog:
interesting, ill spin up a VM and see if i see the same

sunshowers:
Thanks. Meanwhile is there a way I can disable use of the flake locally?

faithanalog:
we could change the .envrc to require opt-in by setting an environment var. like this (just tried this out on my system)

```
vi@philomena ~/o/omicron (eliza/flakehouse)> cat .envrc
# For use with direnv: https://direnv.net
# See also: ./env.sh

PATH_add out/cockroachdb/bin
PATH_add out/clickhouse
PATH_add out/dendrite-stub/bin
PATH_add out/mgd/root/opt/oxide/mgd/bin

if [ "$OMICRON_USE_FLAKE" = 1 ] && nix flake show &> /dev/null
then
    use flake;
fi
```

sunshowers:
Perfect, that would work. (I'd make it opt-in for now)

sunshowers:
@faithanalog done in https://github.com/oxidecomputer/omicron/pull/5011

sunshowers:
(Note that I had the same issue with flakes on main, so this PR doesn't change much)

hawkw:
Hmm...it's possible that the `patchelf` stuff is problematic on systems with Nix that aren't NixOS. I'll look into that a bit.

sunshowers:
Yeah, `readelf --binary` for the build scripts didn't have `$OPENSSL_LIB_DIR` in their runpath.

faithanalog:
FWIW ive done all my testing on gentoo machines because I don't have any NixOS ones handy. so im curious why im not seeing issues but popos is

hawkw:
> Interesting:
> 
> ```
> % ls -al $OPENSSL_LIB_DIR 
> total 6000
> dr-xr-xr-x 1 root root     134 Dec 31  1969 .
> dr-xr-xr-x 1 root root      12 Dec 31  1969 ..
> dr-xr-xr-x 1 root root      80 Dec 31  1969 engines-3
> lrwxrwxrwx 4 root root      14 Dec 31  1969 libcrypto.so -> libcrypto.so.3
> -r-xr-xr-x 2 root root 5320112 Dec 31  1969 libcrypto.so.3
> lrwxrwxrwx 4 root root      11 Dec 31  1969 libssl.so -> libssl.so.3
> -r-xr-xr-x 2 root root  812312 Dec 31  1969 libssl.so.3
> dr-xr-xr-x 1 root root      18 Dec 31  1969 ossl-modules
> ```
> 
> So libssl exists but the dynamic loader isn't able to see it?

@sunshowers what's the value of `$OPENSSL_LIB_DIR` on your machine?

flake.nix
```diff
@@ -3,62 +3,420 @@
...
+              rm -r out/cockroachdb
+
+              mkdir -p out/clickhouse
+              mkdir -p out/cockroachdb/
+
+              ln -s ${mgd.out} -T out/mgd
+              ln -s ${dendrite-stub.out} -T out/dendrite-stub
+              ln -s ${clickhouse.out}/bin/clickhouse out/clickhouse/clickhouse
+              ln -s ${clickhouse.out}/etc/config.xml out/clickhouse
+              ln -s ${cockroachdb.out}/bin out/cockroachdb/bin
```
hawkw, line 408: fwiw, we don't _strictly_ need to symlink the `clickhouse` and `cockroach` executables into `out/`, because `omicron-dev` happily runs them from the `$PATH`; it's just Dendrite and Maghemite that need to be in `out/`, because Dendrite apparently has other stuff it needs in its filesysten,

but, i thought it was nicer to symlink them as well, to more closely mimic the behavior of the various `ci_download_whatever` scripts.

ideally, we could also just point everything at Dendrite and Maghemite's `nix-store` paths via env variables and not have to do any symlinking, but that would require changing other code that i'm not deeply familiar with yet. i might look into that next.

hawkw:
> ```
> % echo $OPENSSL_LIB_DIR
> /nix/store/1l3a02nqq5b5v7rhchj89hi7plmbza5r-openssl-3.0.12/lib
> 
> % readelf --dynamic /home/rain/dev/oxide/omicron2/target/debug/build/ddm-admin-client-df450aa6b594d553/build-script-build
> 
> Dynamic section at offset 0xf98218 contains 33 entries:
>   Tag        Type                         Name/Value
>  0x0000000000000001 (NEEDED)             Shared library: [libssl.so.3]
>  0x0000000000000001 (NEEDED)             Shared library: [libcrypto.so.3]
>  0x0000000000000001 (NEEDED)             Shared library: [libgcc_s.so.1]
>  0x0000000000000001 (NEEDED)             Shared library: [libm.so.6]
>  0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
>  0x0000000000000001 (NEEDED)             Shared library: [ld-linux-x86-64.so.2]
>  0x000000000000001d (RUNPATH)            Library runpath: [/home/rain/dev/oxide/omicron2/outputs/out/lib]
>  0x0000000000000007 (RELA)               0xbaa0
>  0x0000000000000008 (RELASZ)             1234848 (bytes)
>  0x0000000000000009 (RELAENT)            24 (bytes)
>  0x0000000000000017 (JMPREL)             0x139240
>  0x0000000000000002 (PLTRELSZ)           24 (bytes)
>  0x0000000000000014 (PLTREL)             RELA
>  0x0000000000000003 (PLTGOT)             0xf9a1f8
>  0x0000000000000006 (SYMTAB)             0x1fd0
>  0x000000000000000b (SYMENT)             24 (bytes)
>  0x0000000000000005 (STRTAB)             0x7460
>  0x000000000000000a (STRSZ)              15597 (bytes)
>  0x0000000000000019 (INIT_ARRAY)         0xff8638
>  0x000000000000001b (INIT_ARRAYSZ)       16 (bytes)
>  0x000000000000001a (FINI_ARRAY)         0xff8630
>  0x000000000000001c (FINI_ARRAYSZ)       8 (bytes)
>  0x000000006ffffff0 (VERSYM)             0xb14e
>  0x000000006ffffffe (VERNEED)            0xb860
>  0x000000006fffffff (VERNEEDNUM)         6
>  0x000000000000000c (INIT)               0x3d8ee0
>  0x000000000000000d (FINI)               0x3d8ed0
>  0x0000000000000004 (HASH)               0x378
>  0x000000006ffffef5 (GNU_HASH)           0x1fb0
>  0x000000000000001e (FLAGS)              BIND_NOW
>  0x000000006ffffffb (FLAGS_1)            Flags: NOW PIE
>  0x0000000000000015 (DEBUG)              0x0
>  0x0000000000000000 (NULL)               0x0
> ```

Ah, it looks like your runpath is the problem...i'm not totally sure why it doesn't have nix-store paths on it when @faithanalog's build on Gentoo does. Gonna do some googling.

hawkw:
Hmm, https://discourse.nixos.org/t/when-we-are-in-a-nix-develop-shell-how-does-it-make-the-libraries-available-to-the-binaries/18870/4 looks relevant to @sunshowers' problems on non-NixOS distros; in particular [this comment](https://discourse.nixos.org/t/when-we-are-in-a-nix-develop-shell-how-does-it-make-the-libraries-available-to-the-binaries/18870/7). But, I'm not totally sure what the _solution_ is...

hawkw:
LADIES AND GENTLEMEN...WE GOT HIM!

It turns out the problem is:
1. @sunshowers is using `mold` as the linker
2. presumably, Rainmold` was not installed from `nixpkgs`
3. unwrapped `mold` doesn't seem to be able to find Nix libraries, c.f.: https://discourse.nixos.org/t/using-mold-as-linker-prevents-libraries-from-being-found/18530
4. it turns out that the guy who wrote the logging library we use also packaged a wrapped version of `mold` that Just Does The Right Thing on NixOS (https://github.com/NixOS/nixpkgs/pull/257158), so presumably if Rain installs that, everything will just be fine (at least, i hope?)
5. alternatively, Rain could just use a normal linker like the rest of us.

I'm going to remove the explicit setting of `LD_LIBRARY_PATH` that i added in 717fe2db0cccc37818a61d558395177b3b40d826 in an attempt to make things work on Rain's computer, because it just makes their `lld` unhappy when _not_ using `mold`, and is generally a bad thing to do.

faithanalog:
WOW

sunshowers:
And specifically, I've aliased `cargo` to `mold -run cargo` which does `LD_PRELOAD` hacks to redirect `ld` calls to `mold`. Which understandably don't work with Nix.

Issue #4955, opened 2024-02-01
Exclude non-discoverable silos from list of linked silos for IP pool


david-crespo:
Note that the fix here might be to avoid linking them to this pool in the first place rather than to directly exclude them from the list when we fetch them.

---

We show the `default-silo` and `oxide-internal` on the list of linked silos:
 
![Screenshot 2024-02-01 at 9 50 51 AM](https://github.com/oxidecomputer/omicron/assets/3612203/f7e0144a-571e-4a75-a7bf-c31ec5a36bae)

And we can even click on them and see a detail page:

<img width="1324" alt="image" src="https://github.com/oxidecomputer/omicron/assets/3612203/26226eb6-39cd-4710-9d66-c284104f2585">

But neither of these silos actually shows up in the list of silos shown to the operator, and rightly so!

![Screenshot 2024-02-01 at 9 52 58 AM](https://github.com/oxidecomputer/omicron/assets/3612203/35a66dcc-9341-4528-9665-c8148fe20e4b)

zephraph:
non-discoverable silos strike again!

PR #4867, opened 2024-01-23
Delete IP pool links on silo delete and IP pool delete

david-crespo:
Closes #4849 

Before this change, IP pool delete is blocked if there are any outstanding silo links, which would mean the user would have to unlink every silo before deleting a pool. This is annoying. The main insight here, discussed in #4849, is that once we get past the other checks to ensure that the pool or silo is not in use (pool contains no IP ranges, silo contains no projects), there is no need to block on links, and it is fine for us to delete any associated links when we delete the thing.

PR #4962, opened 2024-02-01
Exclude non-discoverable silos from IP pool silos list

david-crespo:
They're still linkable, just like they're still GETable by name or ID. They just don't show up in lists.

Closes https://github.com/oxidecomputer/omicron/issues/4955

### DIFF
--- a/nexus/db-queries/src/db/datastore/ip_pool.rs
+++ b/nexus/db-queries/src/db/datastore/ip_pool.rs
@@ -330,6 +330,7 @@ impl DataStore {
     ) -> ListResultVec<IpPoolResource> {
         use db::schema::ip_pool;
         use db::schema::ip_pool_resource;
+        use db::schema::silo;
 
         paginated(
             ip_pool_resource::table,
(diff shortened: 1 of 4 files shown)


nexus/db-queries/src/db/datastore/ip_pool.rs
```diff
@@ -330,15 +330,19 @@ impl DataStore {
...
             ip_pool_resource::table,
             ip_pool_resource::ip_pool_id,
             pagparams,
         )
         .inner_join(ip_pool::table)
+        .inner_join(silo::table.on(silo::id.eq(ip_pool_resource::resource_id)))
         .filter(ip_pool::id.eq(authz_pool.id()))
         .filter(ip_pool::time_deleted.is_null())
+        .filter(silo::time_deleted.is_null())
+        .filter(silo::discoverable.eq(true))
```
david-crespo, line 345: Matching the logic in `silos_list`. Since we're not joining on silo, we can also filter out links to deleted silos, even though it should be impossible to have any such links after #4867.

https://github.com/oxidecomputer/omicron/blob/0531d6161ee2b670a0ab2a72f26727d4ec3cacaf/nexus/db-queries/src/db/datastore/silo.rs#L322-L351

schema/crdb/23.0.0/up4.sql
```diff
@@ -36,6 +36,7 @@ CROSS JOIN omicron.public.silo AS s
 WHERE p.time_deleted IS null 
   AND p.silo_id IS null -- means it's a fleet pool
   AND s.time_deleted IS null
+  -- TODO: AND NOT s.discoverable ???
```
david-crespo, line 39: Leaning toward not adding this additional filter. The goal is to convert existing data to the new model while keeping the logic the same, and if P0 is a fleet default pool before this migration, it is accessible from non-discoverable silos. So in order to keep the behavior the same _after_ the migration, P0 should be linked to those silos.

zephraph (approved):

PR #4963, opened 2024-02-01
move blueprint execution to its own package

davepacheco:
Some or all of this might be a bad idea, but my thought here is to move the guts of blueprint execution into a new `nexus-blueprint-execution` package.  The goal is to separate the concerns of blueprint execution from running that blueprint execution in a background task.  After this change, the background task itself is a very thin wrapper.  I have found this pattern convenient for the other background tasks I've worked on.  I expect it will be easier to test, maybe incorporate into a CLI tool for development, maybe incorporate parts into RSS so that we can share code where it makes sense, etc.  Another side effect is that I hope (but haven't measured) that taking incremental laps on the test might be a lot faster because it won't need to rebuild and relink all of Nexus.


davepacheco:
So far, this PR doesn't address the test.  If you look at the code in 1249c8d, ignore the test for now.  Logically, the guts of the test now belong in `nexus-blueprint-execution`, which is pretty straightforward.  Then in the background task what we'd be testing is basically _that_ it invokes the real implementation, that it provides the right argument, and that it translates the result into the JSON value that we expect.  I haven't figured out a satisfying way to do this.  Here are some things I considered:

1. Move the guts of the test to `nexus-blueprint-execution` and create a new test for the background task that just exercises what it's responsible for.  There are a few ways to do this:
    1. Do basically the same thing it does today.  This would duplicate a lot of the behavior of the `nexus-blueprint-execution` test.  That's okay but I think we'd want to package up and export a bunch of the pieces from the `nexus-blueprint-execution` test: a couple of canned blueprints plus the fake sled agent servers plus the httpmock expectations that we use for those fake sled agent servers.
    2. Create a trait to wrap the `realize_blueprint` function.  For the real thing, use an impl that calls into `nexus-blueprint-execution`.  For testing, use a faked-up impl, created either with `mockall` or by hand or something.  I normally hate hiding functionality behind traits solely to be able to swap it out for tests.  But maybe it's worthwhile here?
2. Keep the test in the background task subsystem.  This is a little weird because it's testing functionality from its dependency, and it also defeats part of the point, which was quicker iteration time on that test by not having to rebuild all of Nexus.
3. Move the test to `nexus-blueprint-execution`, leaving no test in the background task.  This covers much of the behavior but means we'd be losing coverage of what the background task does.

davepacheco:
I went with a slightly different variant of option 1 above: the background task has a test, it exercises the output as well as the "enabled" property, it always uses the real implementation, and like the other test, it sets up some fake sled agents.  This involved some duplication, but not as much as I feared because we don't need to exercise as many different blueprints or sled agent behaviors in this test.  This seems like a reasonable middle ground to me.

davepacheco:
I forgot to add: I lost the raw data, but: this helps incremental test time a bit, but it's still painful.  On my system, an incremental run of the background task test (touch the file, rerun `cargo nextest run -p omicron-nexus` for just that one test) took over 8 minutes.  For the `nexus-blueprint-execution` package, it took about 5 minutes.  This pattern makes a much bigger difference for nexus-deployment (incremental test run rebuild took 6 seconds) and nexus-inventory (incremental test run rebuild took 55s).  I suspect it doesn't help as much here because the test still pulls in all of omicron-nexus to run a whole `ControlPlaneTestContext`.  We could potentially slim this down and create just a DataStore, or even plug the list of sleds into the executor so it doesn't even need that.  I bet that would make a huge improvement for the incremental test run time.  But I'm not sure if that will make sense in the long term, since execution may rely on more of the stack having been set up (e.g., an initial DNS version in the db).

nexus/blueprint-execution/src/omicron_zones.rs
```diff
@@ -0,0 +1,346 @@
...
+use sled_agent_client::Client as SledAgentClient;
+use slog::info;
+use slog::warn;
+use slog::Logger;
+use std::collections::BTreeMap;
+use uuid::Uuid;
+
+/// Idempotently ensure that the specified Omicron zones are deployed to the
+/// corresponding sleds
+pub async fn deploy_zones(
```
jgallagher, line 23: I'm tempted to say this should be `pub(crate)` since the only caller is the main `realize_blueprint` entry point, but I don't feel strongly about it (e.g., if you think there might be a reason to call this directly from some other crate's test suite or something).

nexus/blueprint-execution/src/omicron_zones.rs
```diff
@@ -0,0 +1,346 @@
...
+                Err(error) => {
+                    warn!(log, "{error:#}");
+                    Some(error)
+                }
+                Ok(_) => {
+                    info!(
+                        log,
+                        "Successfully deployed zones for sled agent";
+                        "sled_id" => %sled_id,
+                        "generation" => config.generation.to_string()
```
jgallagher, line 53: Tiny nit, since `generation` impls `Display`:

```suggestion
                        "generation" => %config.generation,
```

(This might really not matter if `slog` just calls `.to_string()` anyway, but it matches `%sled_id` :man_shrugging:)

nexus/blueprint-execution/src/omicron_zones.rs
```diff
@@ -0,0 +1,346 @@
...
+                    // be only a single zone.
+                    request::body(json_decoded(|c: &OmicronZonesConfig| {
+                        c.generation == 1u32.into() && c.zones.len() == 1
+                    }))
+                ])
+                .respond_with(status_code(204)),
+            );
+        }
+
+        // Executre it.
```
jgallagher, line 252:
```suggestion
        // Execute it.
```

nexus/blueprint-execution/tests/config.test.toml
```diff
@@ -0,0 +1 @@
+../../tests/config.test.toml
```
jgallagher, line 1: Accidental addition maybe?

jgallagher (approved):
The middle ground you landed on for the tests LGTM :+1:

nexus/blueprint-execution/src/omicron_zones.rs
davepacheco: 👍 done in cdc88d0b7.

nexus/blueprint-execution/src/omicron_zones.rs
davepacheco: 👍 done in cdc88d0b7.

nexus/blueprint-execution/src/omicron_zones.rs
davepacheco: 👍 done in cdc88d0b7.

nexus/blueprint-execution/tests/config.test.toml
davepacheco: Surprisingly, no.  The nexus-test-utils stuff loads a config file from a relative path and this enables it to work for this package.  omdb has a similar symlink.  (There may be a way to clean this up.)

andrewjstone (approved):
Excellent. Thanks @davepacheco

PR #4965, opened 2024-02-01
[nexus-macros] convert dump tests to snapshot tests

sunshowers:
Currently, these "dump" tests are mainly for debugging. However, the output of
these macros is expected to be stable, so we can turn them into snapshot tests.

(Ordinarily I'd use cargo-insta for snapshot tests, but we use expectorate
which works fine.)

### DIFF
--- a/nexus/db-macros/Cargo.toml
+++ b/nexus/db-macros/Cargo.toml
@@ -18,4 +18,5 @@ syn = { workspace = true, features = ["extra-traits"] }
 omicron-workspace-hack.workspace = true
 
 [dev-dependencies]
+expectorate.workspace = true
 prettyplease.workspace = true
--- a/nexus/db-macros/outputs/update_artifact.txt
+++ b/nexus/db-macros/outputs/update_artifact.txt
@@ -1,0 +1,206 @@
+///Selects a resource of type UpdateArtifact (or any of its children, using the functions on this struct) for lookup or fetch
+pub enum UpdateArtifact<'a> {
+    /// An error occurred while selecting the resource
+    ///
+    /// This error will be returned by any lookup/fetch attempts.
+    Error(Root<'a>, Error),
+    /// We're looking for a resource with the given primary key
+    ///
+    /// This has no parent container -- a by-id lookup is always global
+    PrimaryKey(Root<'a>, String, i64, KnownArtifactKind),
+}
+impl<'a> UpdateArtifact<'a> {
+    /// Fetch the record corresponding to the selected resource
+    ///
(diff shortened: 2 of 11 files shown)


davepacheco (approved):

smklein (approved):

nexus/db-macros/outputs/update_artifact.txt
```diff
@@ -0,0 +1,206 @@
...
+                            ),
+                        ),
+                    ),
+                )
+            })?;
+        let authz_update_artifact = Self::make_authz(
+            &&authz::FLEET,
+            &db_row,
+            LookupType::ByCompositeId(
+                format!("name = {:?}, version = {:?}, kind = {:?}", v0, v1, v2,),
```
smklein, line 201: Well, the tests are certainly working!

```suggestion
                format!("name = {:?}, version = {:?}, kind = {:?}", v0, v1, v2),
```

Should fix CI, I think

sunshowers: Ummmm I can't repro this locally, the test passes for me

sunshowers: This looks like prettyplease doing its thing? But why would it behave differently on my machine?

nexus/db-macros/Cargo.toml
```diff
@@ -18,4 +18,5 @@ syn = { workspace = true, features = ["extra-traits"] }
 omicron-workspace-hack.workspace = true
 
 [dev-dependencies]
+expectorate.workspace = true
 prettyplease.workspace = true
```
dtolnay, line 22:
```suggestion
prettyplease = { workspace = true, features = ["verbatim"] }
```

My guess is this will resolve the divergence. I will leave it to you to look into how come the feature can be disabled on @sunshowers's machine and enabled in the CI machine.

dtolnay (changes requested):

nexus/db-macros/outputs/update_artifact.txt
sunshowers: Fixed by enabling the verbatim feature.

It looks like prettyplease is being built two different ways, once with the verbatim feature and once without. I don't know why hakari isn't catching that -- will diagnose this later.

nexus/db-macros/Cargo.toml
sunshowers: Hah, thanks. Yeah, this looks like a feature unification thing.

ahl: @dtolnay if we say "serde" into a mirror three times, do you appear?

PR #4973, opened 2024-02-03
omdb: add "mgs sensors" and "mgs dashboard" commands

bcantrill:


dev-tools/omdb/src/bin/omdb/mgs/dashboard.rs
```diff
@@ -0,0 +1,1115 @@
...
+/// Runs `omdb mgs dashboard`
+///
+pub(crate) async fn cmd_mgs_dashboard(
+    omdb: &crate::Omdb,
+    log: &slog::Logger,
+    mgs_args: &crate::mgs::MgsArgs,
+    args: &DashboardArgs,
+) -> Result<(), anyhow::Error> {
+    let mut input = if let Some(ref input) = args.sensors_args.input {
+        let file = File::open(input)?;
```
jgallagher, line 797: Probably want something like

```suggestion
        let file = File::open(input).with_context(|| format!("failed to open {input}"))?;
```

so the error message includes the path.

`.context()` / `.with_context()` _might_ be useful on other `?`s in this function, but I'm less sure about them (and also suspect it's a lot less expected for any of them to fail in practice?).

dev-tools/omdb/src/bin/omdb/mgs/dashboard.rs
```diff
@@ -0,0 +1,1115 @@
...
+        .constraints(
+            [Constraint::Percentage(100 - r), Constraint::Percentage(r)]
+                .as_ref(),
+        )
+        .split(parent);
+
+    let latest = now as i64 - graph.offs as i64;
+    let earliest = Local.timestamp_opt(latest - graph.width as i64, 0).unwrap();
+    let latest = Local.timestamp_opt(latest, 0).unwrap();
+    let fmt = "%Y-%m-%d %H:%M:%S";
```
jgallagher, line 926: I don't see anything wrong with this, but wondering: could we use [.to_rfc3339()](https://docs.rs/chrono/0.4.33/chrono/struct.DateTime.html#method.to_rfc3339) instead? Or maybe `to_rfc3339_opts` with `use_z: true`?

dev-tools/omdb/src/bin/omdb/mgs/sensors.rs
```diff
@@ -0,0 +1,954 @@
...
+    fn format(&self, value: f32, parseable: bool) -> String {
+        if parseable {
+            format!("{value}")
+        } else {
+            match self.kind {
+                MeasurementKind::Speed => {
+                    format!("{value:0} RPM")
+                }
+                _ => {
+                    format!("{value:.2}{}", self.units())
```
jgallagher, line 138: Do you want a space between these like there is for `RPM`?

dev-tools/omdb/src/bin/omdb/mgs/sensors.rs
```diff
@@ -0,0 +1,954 @@
...
+pub(crate) enum DeviceIdentifier {
+    Field(usize),
+    Device(String),
+}
+
+impl DeviceIdentifier {
+    fn device(&self) -> &String {
+        match self {
+            Self::Device(ref device) => device,
+            _ => panic!(),
```
jgallagher, line 221: This seems a little fishy (likewise the panic in `field()`). I'm tempted to propose one of two changes:

* Make this return `Option<&String>` (or `Option<&str>` since `&String` is kinda weird), and let the caller unwrap if they know somehow that this must be a device.
* (Maybe? This might be really annoying) Drop this enum entirely? If its users always know exactly which case they have (which they have to to use these methods), could they use the underlying types instead?

dev-tools/omdb/src/bin/omdb/mgs/sensors.rs
```diff
@@ -0,0 +1,954 @@
...
+    let now = std::time::Instant::now();
+
+    let mut handles = vec![];
+    for sp_id in sp_list {
+        let mgs_client = mgs_client.clone();
+        let type_ = sp_id.type_;
+        let slot = sp_id.slot;
+
+        let handle =
+            tokio::spawn(async move { sp_info(mgs_client, type_, slot).await });
```
jgallagher, line 337: Tiny nit: `sp_info` returns a future that can be spawned directly; we don't have to wrap it in an async block:

```suggestion
            tokio::spawn(sp_info(mgs_client, type_, slot));
```

(I think after making this change you could probably drop the 3 locals too, and make it `spawn(sp_info(mgs_client.clone(), sp_id.type_, sp_id.slot))`, since you no longer need to move them into a block.)

dev-tools/omdb/src/bin/omdb/mgs/sensors.rs
```diff
@@ -0,0 +1,954 @@
...
+        SensorMetadata {
+            sensors_by_sensor,
+            sensors_by_sensor_and_sp,
+            sensors_by_id,
+            sensors_by_sp,
+            work_by_sp,
+            start_time: args.start,
+            end_time: match args.end {
+                Some(end) => Some(end),
+                None => match args.duration {
```
jgallagher, line 646: I'd probably use `.map` here to avoid the nested match:

```suggestion
                None => args.duration.map(|duration| time + duration),
```

dev-tools/omdb/src/bin/omdb/mgs/sensors.rs
```diff
@@ -0,0 +1,954 @@
...
+                }
+                _ => None,
+            })
+            .zip(ids.iter())
+        {
+            rval.push((*id, value));
+        }
+    }
+
+    Ok((rval, std::time::Instant::now().duration_since(start)))
```
jgallagher, line 690:
```suggestion
    Ok((rval, start.elapsed()))
```

dev-tools/omdb/src/bin/omdb/mgs/sensors.rs
```diff
@@ -0,0 +1,954 @@
...
+/// Runs `omdb mgs sensors`
+///
+pub(crate) async fn cmd_mgs_sensors(
+    omdb: &crate::Omdb,
+    log: &slog::Logger,
+    mgs_args: &crate::mgs::MgsArgs,
+    args: &SensorsArgs,
+) -> Result<(), anyhow::Error> {
+    let mut input = if let Some(ref input) = args.input {
+        let file = File::open(input)?;
```
jgallagher, line 820:
```suggestion
        let file = File::open(input).with_context(|| format!("failed to open {input}"))?;
```

dev-tools/omdb/src/bin/omdb/mgs/sensors.rs
```diff
@@ -0,0 +1,954 @@
...
+        )
+    } else {
+        SensorInput::MgsClient(mgs_args.mgs_client(omdb, log).await?)
+    };
+
+    let (metadata, mut values) = sensor_metadata(&mut input, args).await?;
+
+    //
+    // A bit of shenangians to force metadata to be 'static -- which allows
+    // us to share it with tasks.
```
jgallagher, line 833: We don't need to leak this to share it with tasks - the general thing here is to stash it into an `Arc`. The changes are spread out but pretty trivial:

```diff
--- a/dev-tools/omdb/src/bin/omdb/mgs/dashboard.rs
+++ b/dev-tools/omdb/src/bin/omdb/mgs/dashboard.rs
@@ -486,7 +486,7 @@ struct Dashboard {
 }

 impl Dashboard {
-    fn new(metadata: &'static SensorMetadata) -> Result<Dashboard> {
+    fn new(metadata: &SensorMetadata) -> Result<Dashboard> {
         let mut sps =
             metadata.sensors_by_sp.keys().map(|m| *m).collect::<Vec<_>>();
         let mut graphs = HashMap::new();
@@ -806,14 +806,7 @@ pub(crate) async fn cmd_mgs_dashboard(
     let (metadata, values) =
         sensor_metadata(&mut input, &args.sensors_args).await?;

-    //
-    // A bit of shenangians to force metadata to be 'static -- which allows
-    // us to share it with tasks.
-    //
-    let metadata = Box::leak(Box::new(metadata));
-    let metadata: &_ = metadata;
-
-    let mut dashboard = Dashboard::new(metadata)?;
+    let mut dashboard = Dashboard::new(&metadata)?;
     let mut last = values.time;
     let mut force = true;
     let mut update = true;
@@ -822,7 +815,7 @@ pub(crate) async fn cmd_mgs_dashboard(

     if args.sensors_args.input.is_some() && !args.simulate_realtime {
         loop {
-            let values = sensor_data(&mut input, metadata).await?;
+            let values = sensor_data(&mut input, &metadata).await?;

             if values.time == 0 {
                 break;
@@ -862,7 +855,7 @@ pub(crate) async fn cmd_mgs_dashboard(

         if update && now != last {
             let kicked = Instant::now();
-            let f = sensor_data(&mut input, metadata);
+            let f = sensor_data(&mut input, &metadata);
             last = now;

             while Instant::now().duration_since(kicked).as_millis() < 800 {
--- a/dev-tools/omdb/src/bin/omdb/mgs/sensors.rs
+++ b/dev-tools/omdb/src/bin/omdb/mgs/sensors.rs
@@ -15,6 +15,7 @@ use gateway_client::types::SpType;
 use multimap::MultiMap;
 use std::collections::{HashMap, HashSet};
 use std::fs::File;
+use std::sync::Arc;
 use std::time::{Duration, SystemTime, UNIX_EPOCH};

 #[derive(Debug, Args)]
@@ -543,7 +544,7 @@ fn sp_info_csv<R: std::io::Read>(
 pub(crate) async fn sensor_metadata<R: std::io::Read>(
     input: &mut SensorInput<R>,
     args: &SensorsArgs,
-) -> Result<(SensorMetadata, SensorValues), anyhow::Error> {
+) -> Result<(Arc<SensorMetadata>, SensorValues), anyhow::Error> {
     let by_kind = if let Some(types) = &args.types {
         let mut h = HashSet::new();

@@ -634,7 +635,7 @@ pub(crate) async fn sensor_metadata<R: std::io::Read>(
     }

     Ok((
-        SensorMetadata {
+        Arc::new(SensorMetadata {
             sensors_by_sensor,
             sensors_by_sensor_and_sp,
             sensors_by_id,
@@ -648,7 +649,7 @@ pub(crate) async fn sensor_metadata<R: std::io::Read>(
                     None => None,
                 },
             },
-        },
+        }),
         SensorValues { values, time, latencies: info.latencies },
     ))
 }
@@ -692,7 +693,7 @@ async fn sp_read_sensors(

 async fn sp_data_mgs(
     mgs_client: &gateway_client::Client,
-    metadata: &'static SensorMetadata,
+    metadata: &Arc<SensorMetadata>,
 ) -> Result<SensorValues, anyhow::Error> {
     let mut values = HashMap::new();
     let mut latencies = HashMap::new();
@@ -701,9 +702,10 @@ async fn sp_data_mgs(
     for sp_id in metadata.sensors_by_sp.keys() {
         let mgs_client = mgs_client.clone();
         let id = *sp_id;
+        let metadata = Arc::clone(&metadata);

         let handle = tokio::spawn(async move {
-            sp_read_sensors(&mgs_client, &id, metadata).await
+            sp_read_sensors(&mgs_client, &id, &metadata).await
         });

         handles.push((id, handle));
@@ -729,7 +731,7 @@ async fn sp_data_mgs(
 fn sp_data_csv<R: std::io::Read + std::io::Seek>(
     reader: &mut csv::Reader<R>,
     position: &mut csv::Position,
-    metadata: &'static SensorMetadata,
+    metadata: &SensorMetadata,
 ) -> Result<SensorValues, anyhow::Error> {
     let headers = reader.headers()?;
     let hlen = headers.len();
@@ -795,14 +797,14 @@ fn sp_data_csv<R: std::io::Read + std::io::Seek>(

 pub(crate) async fn sensor_data<R: std::io::Read + std::io::Seek>(
     input: &mut SensorInput<R>,
-    metadata: &'static SensorMetadata,
+    metadata: &Arc<SensorMetadata>,
 ) -> Result<SensorValues, anyhow::Error> {
     match input {
         SensorInput::MgsClient(ref mgs_client) => {
             sp_data_mgs(mgs_client, metadata).await
         }
         SensorInput::CsvReader(reader, position) => {
-            sp_data_csv(reader, position, metadata)
+            sp_data_csv(reader, position, &metadata)
         }
     }
 }
@@ -828,13 +830,6 @@ pub(crate) async fn cmd_mgs_sensors(

     let (metadata, mut values) = sensor_metadata(&mut input, args).await?;

-    //
-    // A bit of shenangians to force metadata to be 'static -- which allows
-    // us to share it with tasks.
-    //
-    let metadata = Box::leak(Box::new(metadata));
-    let metadata: &_ = metadata;
-
     let mut sensors = metadata.sensors_by_sensor.keys().collect::<Vec<_>>();
     sensors.sort();

@@ -939,7 +934,7 @@ pub(crate) async fn cmd_mgs_sensors(
             wakeup += tokio::time::Duration::from_millis(1000);
         }

-        values = sensor_data(&mut input, metadata).await?;
+        values = sensor_data(&mut input, &metadata).await?;

         if args.input.is_some() && values.time == 0 {
             break;
```

jgallagher (approved):
LGTM - almost entirely nitpicks for clarity / error reporting.

dev-tools/omdb/src/bin/omdb/mgs/dashboard.rs
bcantrill: Yeah, that's a little nicer. As it stood, it didn't leave *too* much to the imagination:

```console
$ omdb mgs dashboard -i nonono
Error: No such file or directory (os error 2)
```

But being more explicit is better -- thank you!

dev-tools/omdb/src/bin/omdb/mgs/dashboard.rs
bcantrill: It's a matter of horizontal screen real estate -- we don't get *quite* the exact format we want with `to_rfc3339_opts` in that it consumes maximum width for the TZ offset. I'll add a comment.

dev-tools/omdb/src/bin/omdb/mgs/dashboard.rs
```diff
@@ -0,0 +1,1115 @@
...
+            let values = f.await?;
+            dashboard.values(&values);
+            force = true;
+            continue;
+        }
+
+        tokio::time::sleep(Duration::from_millis(10)).await;
+    };
+
+    // restore terminal
```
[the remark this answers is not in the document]
jgallagher: Huh, I wonder if this is terminal dependent? I just tried

```diff
--- a/dev-tools/omdb/src/bin/omdb/mgs/dashboard.rs
+++ b/dev-tools/omdb/src/bin/omdb/mgs/dashboard.rs
@@ -866,6 +866,7 @@ pub(crate) async fn cmd_mgs_dashboard(
             last = now;

             while Instant::now().duration_since(kicked).as_millis() < 800 {
+                anyhow::bail!("kaboom");
                 tokio::time::sleep(Duration::from_millis(10)).await;

                 match run_dashboard(&mut terminal, &mut dashboard, force) {
```

and I got stuck with a raw mode terminal.

dev-tools/omdb/src/bin/omdb/mgs/dashboard.rs
```diff
@@ -0,0 +1,1115 @@
...
+                        break 'outer Err(err);
+                    }
+                    Ok(true) => {
+                        break 'outer Ok(());
+                    }
+                    _ => {}
+                }
+            }
+
+            let values = f.await?;
```
jgallagher, line 882: Looking at this again while I was poking around with the terminal restore on error question: why do we construct `f` before the while loop but not `.await` it until here? I think it would be equivalent to get rid of `f` and call `sensor_data(...).await` directly here, but maybe there was a different intent with `f`?

dev-tools/omdb/src/bin/omdb/mgs/sensors.rs
bcantrill: So, I'm not as in to either of those. To me, this makes the consuming code more complicated -- and the `unwrap` to me would introduce clutter without meaningfully changing anything. Inclined to add a comment here rather than restructure or drop it...

jgallagher: Sure; this isn't a hill I'm dying on, since this isn't a public API and misuse is pretty unlikely.

dev-tools/omdb/src/bin/omdb/mgs/sensors.rs
bcantrill: Yeah: for the other units (`°C`, `V`, `A`), it looks more natural with the number next to the unit -- but it looks decidedly unnatural with `RPM`; will add a comment.

dev-tools/omdb/src/bin/omdb/mgs/sensors.rs
bcantrill: I added a comment to clarify.

dev-tools/omdb/src/bin/omdb/mgs/sensors.rs
bcantrill: Thank you for this! I had tried to get this working earlier, but had (clearly) failed!

dev-tools/omdb/src/bin/omdb/mgs/dashboard.rs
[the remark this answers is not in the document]
bcantrill: Ha -- I realized that my attempt to induce this was being caught by my other code that prevents it. ;)  I rephrased this slightly differently (making sure that there isn't a return -- either explicit or via the question mark operator -- from within that block).

PR #4983, opened 2024-02-04
Bootstore related log message adjustments

andrewjstone:
Actual bootstore messages sent and received were not logged. We now log them. These messages are quite infrequent, but have massive debugging value.

We also get rid of logging bootstrap network prefixes everytime we look at them. This happens constantly, since bootstore polls them every few seconds indefinitely so it can discover new nodes. It's also unnecessary for bootstore in particular, because bootstore will already log when the prefixes change.

The method the log is removed from is used in a few other places, including installinator and wicketd, but there are enough logs and related contexts to debug things there. In any case, failure there is a support scenario and we can use ddm-admin to debug. Additional contextual logs can also be added as needed.

### DIFF
--- a/bootstore/src/schemes/v0/peer.rs
+++ b/bootstore/src/schemes/v0/peer.rs
@@ -647,7 +647,7 @@ impl Node {
             .iter()
             .filter(|(id, _)| Some(*id) != excluded_peer)
         {
-            debug!(
+            info!(
                 self.log,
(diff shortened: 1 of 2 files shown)


jgallagher (approved):
Thank you! Been meaning to do this for a while, and never got over the "well I'll just `cat log | grep -v 'Received prefixes'` again" hump.

PR #4986, opened 2024-02-05
Better error messages on invalid IP range

### LINKED ISSUE #4968: IP range JSON validation error is not helpful
I tried various kinds of bad IP range and they all have the same horrible message, even when the addresses themselves are valid but first > last or one is IPv4 and the other is IPv6. 

```ts
> (await oxide.ipPoolRangeRemove({ path: { pool: 'default' }, body: { first: '', last: '' } })).data.message
"unable to parse JSON body: data did not match any variant of untagged enum IpRange"

> (await oxide.ipPoolRangeRemove({ path: { pool: 'default' }, body: { first: 'abc', last: 'def' } })).data.message
"unable to parse JSON body: data did not match any variant of untagged enum IpRange"

> (await oxide.ipPoolRangeRemove({ path: { pool: 'default' }, body: { first: '1.1.1.2', last: '1.1.1.1' } })).data.message
"unable to parse JSON body: data did not match any variant of untagged enum IpRange"

> (await oxide.ipPoolRangeRemove({ path: { pool: 'default' }, body: { first: '1.1.1.2', last: '2001:db8:3333:4444:CCCC:DDDD:EEEE:FFFF' } })).data.message
"unable to parse JSON body: data did not match any variant of untagged enum IpRange" 
```

david-crespo:
Closes #4968 

Just a first pass. TODOs indicate clear room for improvement.

### DIFF
--- a/common/src/address.rs
+++ b/common/src/address.rs
@@ -11,7 +11,7 @@ use crate::api::external::{self, Error, Ipv4Net, Ipv6Net};
 use ipnetwork::{Ipv4Network, Ipv6Network};
 use once_cell::sync::Lazy;
 use schemars::JsonSchema;
-use serde::{Deserialize, Serialize};
+use serde::{de, Deserialize, Serialize};
 use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddrV6};
 
 pub const AZ_PREFIX: u8 = 48;
@@ -347,7 +347,7 @@ pub fn get_64_subnet(
 ///
 /// The first address in the range is guaranteed to be no greater than the last
 /// address.
-#[derive(Clone, Copy, Debug, PartialEq, Deserialize, Serialize)]
+#[derive(Clone, Copy, Debug, PartialEq, Serialize)]
 #[serde(untagged)]
 pub enum IpRange {
     V4(Ipv4Range),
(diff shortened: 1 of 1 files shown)


david-crespo:
@ahl I would appreciate a quick look since this is your area of expertise.

common/src/address.rs
```diff
@@ -386,6 +386,65 @@ impl JsonSchema for IpRange {
...
+        D: serde::Deserializer<'de>,
+    {
+        #[derive(Deserialize)]
+        struct RawIpRange {
+            first: String,
+            last: String,
+        }
+
+        let raw = RawIpRange::deserialize(deserializer)?;
+
```
ahl, line 403: You could try something like this:

```rust
let first = raw.first.parse::<IpAddr>();
let last = raw.last.parse::<IpAddr>();

match (first, last) {
    (Err(_), Err(_)) => // both bad
    (Ok(IpAddr::V4(_)), Ok(IpAddr::V6(_))) => //nah
    (Ok(IpAddr::V4(first)), Ok(IpAddr::V4(last)) => IpRange::V4 { first, last },

// ...
```

ahl (commented):
Seems good; I added a suggestion for an approach you might try that may or may not be tidier.

FYI here's something similar in the CLI: https://github.com/oxidecomputer/oxide.rs/blob/bdde14a547f66247fb84610114afda531dc8e0e5/cli/src/main.rs#L71

common/src/address.rs
```diff
@@ -386,6 +386,65 @@ impl JsonSchema for IpRange {
...
+        // Attempt to parse all the kinds
+        // TODO: this should prob be done by parsing as Ipv4Range and Ipv6Range
```
ahl, line 405: I don't think so? That precludes you from intelligible errors when someone (probably me) mixes ipv4 and ipv6 here.

common/src/address.rs
ahl: Actually, you could do:

```rust
struct RawIpRange {
    first: IpAddr,
    last: IpAddr
}
```

PR #4987, opened 2024-02-05
Don't constrain baseboard when adding sled

andrewjstone:
We were artificially limiting how we added sleds to a rack by forcing them to be of type `Baseboard::Gimlet`. Instead of constructing a `Baseboard` inside nexus, we instead send down the serial and part numbers and use those to uniquely identify the sled. We ignore whether it's a PC or Gimlet as long as the ids match. This is similar to how the inventory works and allows adding a sled to a rack on the falcon a4x2 testbed.

### DIFF
--- a/nexus/src/app/rack.rs
+++ b/nexus/src/app/rack.rs
@@ -871,36 +871,15 @@ impl super::Nexus {
             )
             .await?;
 
-        // Grab the SPs from the last collection
-        let collection =
-            self.db_datastore.inventory_get_latest_collection(opctx).await?;
-
-        // If there isn't a collection, we don't know about the sled
-        let Some(collection) = collection else {
-            return Err(Error::unavail("no inventory data available"));
-        };
-
-        // Find the revision
-        let Some(sp) = collection.sps.get(&baseboard_id) else {
-            return Err(Error::ObjectNotFound {
-                type_name: ResourceType::Sled,
-                lookup_type:
-                    omicron_common::api::external::LookupType::ByCompositeId(
-                        format!("{sled:?}"),
-                    ),
-            });
-        };
-
-        // Convert the baseboard as necessary
-        let baseboard = sled_agent_client::types::Baseboard::Gimlet {
-            identifier: sled.serial.clone(),
-            model: sled.part.clone(),
-            revision: sp.baseboard_revision.into(),
+        // Convert `UninitializedSledId` to the sled-agent type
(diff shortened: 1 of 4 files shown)


davepacheco:
This is probably fine but after the discussion in chat on Friday I'm not sure it's the right direction for the testbed.  It sounded like the direction we wanted was "make the testbed look indistinguishable from real hardware" rather than changing the software to better support the testbed?  (CC @rcgoodfellow)

andrewjstone:
> This is probably fine but after the discussion in chat on Friday I'm not sure it's the right direction for the testbed. It sounded like the direction we wanted was "make the testbed look indistinguishable from real hardware" rather than changing the software to better support the testbed? (CC @rcgoodfellow)

I actually don't think this violates that constraint at all. The testbed code acts exactly like the omicron code. There was never a need to specify "Gimlet" at the nexus layer and so we don't anymore. The constraint is that the `BaseboardId` is unique, which is what we do in both hardware and testbed.

I also don't think that constraint means we would never change production code to run on the testbed, similar to how we would change it to work on new generations of hardware. Not all sleds will be gimlets in the future and so this change allows new hardware to work as well.

Additionally, this literally allows me to run our current update on the testbed *now*. I just tested it via adding a sled and generating blueprints via OMDB. That's a big win for a small change.

davepacheco:
Ah, I think I understand this better now.  We're not changing the Nexus external API here.  It's just that before, Nexus was fetching the revision from the inventory, solely to fill in the "revision" that Bootstrap Agent wanted.  Now Sled Agent accepts what we actually want, which is a part/serial to identify the sled (no revision -- it was not needed), and we only compare those two fields.

My only question then is whether there's any impact to existing systems.  I'm guessing not since we still never run mixed versions of Nexus/Sled Agent?

andrewjstone:
> Ah, I think I understand this better now. We're not changing the Nexus external API here. It's just that before, Nexus was fetching the revision from the inventory, solely to fill in the "revision" that Bootstrap Agent wanted. Now Sled Agent accepts what we actually want, which is a part/serial to identify the sled (no revision -- it was not needed), and we only compare those two fields.

You got it. I had this epiphany last night, and realized it would make my questions on friday moot! 😃 

> 
> My only question then is whether there's any impact to existing systems. I'm guessing not since we still never run mixed versions of Nexus/Sled Agent?

Exactly. And in addition to that, the only time this path gets called is when *adding* a sled, which we haven't done yet anywhere.

davepacheco (approved):

PR #4989, opened 2024-02-05
update internal DNS during blueprint execution

### LINKED ISSUE #4887: blueprints: Nexus needs to update internal DNS
For many services, being in internal DNS determines what's in-service vs. not.
 
- Blueprints need to contain enough information for blueprint execution to reliably compute the complete internal DNS contents.  (This might already be true?)
- Blueprint execution needs to update DNS as needed

The basic scheme we've discussed is:

- We store in the blueprint: the current internal DNS version.  We could either fetch this during planning or else take it from the parent blueprint, if we assume we only ever change internal DNS through blueprints (which is probably a good idea, but also more brittle).
- We store in the blueprint: for each Omicron zone whether it's in-service or not.  This is separate from whether it exists or not.  (We could store a list of the uuids in-service, or a list of the uuids not-in-service, or whatever, but ultimately we need to be able to tell exactly which zones are in service during execution.)
- During execution, we:
    - Fetch the current DNS config, including version.  If the version differs from the blueprint's stored DNS version, abort.  This means someone else has beat us to it.
    - Compute the expected contents of the DNS config.  If it matches the actual DNS config, do nothing.
    - Otherwise, create a `DnsVersionUpdateBuilder` and use add/remove to create what the new DNS config should look like.
    - Execute the DNS update.  On conflict, abort.  This means someone else beat us to it.

This should work even if a Nexus attempts to execute an older (no-longer-the-target blueprint) because we'll wind up in one of the abort cases.

davepacheco:
Fixes #4887.


internal-dns/src/config.rs
```diff
@@ -378,23 +419,6 @@ impl DnsConfigBuilder {
...
-                });
-                (ServiceName::Scrimlet(location).dns_name(), vec![srv])
-            });
-
-        let scrimlet_aaaa_records =
-            self.scrimlets.into_iter().map(|(location, addr)| {
-                let aaaa = DnsRecord::Aaaa(*addr.ip());
-                (format!("{location}.scrimlet"), vec![aaaa])
-            });
-
```
davepacheco, line 397: tl;dr: I brought this up in oxide-control-plane chat and sync'd with @rcgoodfellow and @internet-diglett and I believe these DNS names can be removed.

More details:

- I could not find any references to these DNS names anywhere in Omicron.  I looked for "switch0.scrimlet" and "switch1.scrimlet" and even ".scrimlet" (in case it was being filled into a template).  If they were used programmatically, they'd show up as `ServiceName::Scrimlet`.  The only place that's used is [a comment explaining some context](https://github.com/oxidecomputer/omicron/blob/ddd036760d2bef6fdda4c1c18469f253333e4a29/nexus/src/app/sagas/switch_port_settings_common.rs#L504-L517).
- What I got from chat is what that comment says: these DNS names were added at some point after dogfood was set up.  It was then discovered that they couldn't be used for whatever they were going to be used for because these DNS names only get set up at RSS-time, so systems like dogfood were just never going to have them.  Instead, a different mechanism was created to find these zones when needed.
- From what I can tell, the RSS code that assigned these DNS names chose the switch location (switch0 vs. switch1) arbitrarily, based on which one's sled id came lexicographically first.  That's potentially inconsistent with what MGS reported as switch0 and switch1.  So these names were potentially wrong on any systems where they _were_ set up.
- It should be possible to augment the code in this PR to reliably establish correct DNS names for these, provided we have an inventory collection with the information that we need from MGS.  This way, in the future, we _can_ use DNS for this.  I didn't add this to this PR.

schema/crdb/dbinit.sql
```diff
@@ -3082,6 +3082,11 @@ CREATE TABLE IF NOT EXISTS omicron.public.blueprint (
     -- now gone.
     parent_blueprint_id UUID,
 
+    -- identifies the latest internal DNS version when blueprint planning began
+    -- XXX-dap when it comes time for the migration, make sure to update
```
davepacheco, line 3086: When this PR has approval, I'll implement the db migration part.  That part needs to make sure we provide a value for the `internal_dns_version` field for existing rows.  I'm planning to set it to 1.  That should be the current DNS version deployed everywhere.  It should also be safe even if it weren't -- having one that's too old is always safe.  It just means you won't be able to execute that blueprint and you need to re-plan to make a new one.

sled-agent/src/rack_setup/plan/service.rs
```diff
@@ -298,36 +297,17 @@ impl Plan {
...
-                .unwrap();
-
-            // TODO only works for single rack
-            let sled_address = get_sled_address(sled.subnet);
-            let switch_location = if i == 0 {
-                SwitchLocation::Switch0
-            } else {
-                SwitchLocation::Switch1
-            };
-            dns_builder.host_scrimlet(switch_location, sled_address).unwrap();
```
davepacheco, line 330: This is the bit I mentioned in my other comment.  These DNS names were not used, and they were also not being assigned correctly.

nexus/types/src/inventory.rs
```diff
@@ -134,6 +134,14 @@ impl Collection {
     ) -> impl Iterator<Item = &OmicronZoneConfig> {
         self.omicron_zones.values().flat_map(|z| z.zones.zones.iter())
     }
+
+    /// Iterate over the sled ids of sleds identified as Scrimlets
+    pub fn scrimlets(&self) -> impl Iterator<Item = Uuid> + '_ {
+        self.sled_agents
+            .iter()
+            .filter(|(_, inventory)| inventory.sled_role == SledRole::Scrimlet)
+            .map(|(sled_id, _)| *sled_id)
```
jgallagher, line 143: Tiny nit - if you want to squish this down to `filter_map`, you could use https://doc.rust-lang.org/std/primitive.bool.html#method.then_some (which I only learned about recently):

```rust
.filter_map(|(sled_id, inventory)| (inventory.sled_role == SledRole::Scrimlet).then_some(*sled_id))

File: nexus/blueprint-execution/src/dns.rs [hunk 5]
```diff
@@ -0,0 +1,744 @@
...
+                scrimlet.id,
+                switch_zone_ip,
+                DENDRITE_PORT,
+                MGS_PORT,
+                MGD_PORT,
+            )
+            .unwrap();
+    }
+
+    dns_builder.generation(blueprint.internal_dns_version.next());
```
jgallagher, line 253: This threw me a bit, because it _looks_ like the returned DNS config has always has a generation bump, which means it will be marching up every time this RPW runs. But making sure I understand: that doesn't happen because the config we return here is compared against the existing config using `DnsDiff`, which does _not_ look at the generation number, only the contents, right? That seems a little implicit, but I don't have a great suggestion off the top of my head for making it clearer. We could wait to bump the generation until after the comparison, maybe? That might be a little more obvious?

nexus/blueprint-execution/src/dns.rs
```diff
@@ -0,0 +1,744 @@
...
+                        addr, zone_id
+                    );
+                    expected_srv_targets.insert(format!(
+                        "{}.{}",
+                        name, blueprint_dns_zone.zone_name
+                    ));
+                    continue;
+                }
+
+                if let Some(scrimlet_id) = switch_sleds_by_ip.remove(addr) {
```
jgallagher, line 559: Just for my own understanding: if we remove the scrimlet ID the first time we get here for a particular address, why don't we trip up later due to switch zones having multiple DNS records (dendrite / mgs / mgd)?

nexus/blueprint-execution/src/dns.rs
```diff
@@ -0,0 +1,744 @@
...
+use omicron_common::address::NTP_PORT;
+use omicron_common::address::OXIMETER_PORT;
+use omicron_common::api::external::Error;
+use omicron_common::api::external::Generation;
+use omicron_common::api::external::InternalContext;
+use slog::{debug, info, o};
+use std::collections::BTreeMap;
+use uuid::Uuid;
+
+pub async fn deploy_dns<S>(
```
jgallagher, line 38: Two tiny nits:

* I'd make this `pub(crate)`
* Given the former, I think you could drop `<S>` and take `creator: String`, and let the caller do the conversion. That keeps the ergonomics of the public API without needing the generic on this inner function.

jgallagher (approved):
This looks great; only small nits and clarification questions :+1:

nexus/types/src/inventory.rs
davepacheco: This is neat and I didn't know it.  But I think I like what's here better in this case.

nexus/blueprint-execution/src/dns.rs
davepacheco: I see what you mean.  I think it would be an improvement if instead of `DnsConfigParams`, we passed around the equivalent of `dns_config_params.zones[0].records` -- i.e., just the _records_ that are in the internal DNS zone.  Then we'd assemble that into a final `DnsConfigParams` at the last minute, before writing it to the database.  Then it would be clearer what's being generated and diff'd, etc.  But right now there's no type for that, just `HashMap<String, Vec<DnsRecord>>`.  I think at minimum we'd want a newtype for this, and probably a richer one, too.  `DnsConfigBuilder` maybe should produce one of those instead of `DnsConfigParams`.  Anyway I think this would be better but I'd like to _not_ do it right now.

I have at least added a few comments to explain this though.

(I inferred your suggestion to "wait to bump the generation until after the comparison" to mean "return something with the same generation as what's there, then do the comparison, then mutate it if we find we need to bump it".  I don't love that.  It seems misleading in a different way.)

nexus/blueprint-execution/src/dns.rs
davepacheco: Good question!  I believe there are a few different records here:

- one AAAA record for something like `dendrite-$sled_id.control-plane.oxide.internal`
- three SRV records for stuff like `_mgs._tcp.control-plane.oxide.internal`, `_dendrite._tcp.control-plane.oxide.internal`, and `_mgd._tcp.control-plane.oxide.internal`.  These have a `target` field which points at `dendrite-$sled_id.control-plane.oxide.internal`, not the IP address directly.

This loop iterates over just the names having `AAAA` records.  There should be only one of those for each Scrimlet.  It's the next loop that iterates over the `SRV` records and checks those.

PR #5001, opened 2024-02-06
Unbreak the build (ratatui x omdb)

smklein:
I suspect we had a race condition in our merge-queue-less world, where ratatui was updated to `0.26.0`, and needed slightly different parameters. This PR fixes that compatibility error.


bcantrill (approved):
Looks great -- thank you.  I have verified that the change builds (of course) and that the resulting binary works as expected (also no surprise!).

sunshowers:
Looks like there was a race between @bcantrill and myself, sorry!

sunshowers (approved):

PR #5004, opened 2024-02-06
bump external API version number

ahl:
This is a semi-cowardly approach. Rather than actually implementing the non-integer we [determined](https://rfd.shared.oxide.computer/rfd/0299#_determinations), this moves us from 0.0.1 to 0.0.6 (6 for release 6). This is to dry run the oxide.rs [versioned release process](https://github.com/oxidecomputer/oxide.rs?tab=readme-ov-file#determine-the-version-number).


iliana (approved):

zephraph:
It would be nice to have clarification on if this is a temporary measure to work out the versioning plumbing or if we're abandoning the version scheme laid out in the RFD.

ahl:
Intended as a temporary measure

Issue #4990, opened 2024-02-05
Planner is generating duplicate zone IPs


andrewjstone:
After adding a sled to a testbed cluster and creating a couple of blueprints: 1 to generate the ntp zone to add on the sled and the other including the crucible zones, a duplicate ip was issued for the ntp zone and one of the crucible zones.

```
root@oxz_crucible_ded43e9b:~# Feb  5 23:25:32 unknown last message repeated 434 times
Feb  5 23:25:33 unknown ip: WARNING: oxControlService4:1 has duplicate address fd00:1122:3344:21::21 (in use by 02:08:20:d6:ac:65); disabled
```

```
root@oxz_ntp_ec3274db:~# ipadm
ADDROBJ           TYPE     STATE        ADDR
lo0/v4            static   ok           127.0.0.1/8
lo0/v6            static   ok           ::1/128
oxControlService0/ll addrconf ok        fe80::8:20ff:fed6:ac65%oxControlService0/10
oxControlService0/omicron6 static ok    fd00:1122:3344:21::21/64
root@oxz_ntp_ec3274db:~# exit
logout

[Connection to zone 'oxz_ntp_ec3274db-1d79-47d9-9664-022f2c53c81a' pts/1 closed]
root@g2:~# zlogin oxz_crucible_ded43e9b-5fff-4c6a-91c4-d528b9579b70
[Connected to zone 'oxz_crucible_ded43e9b-5fff-4c6a-91c4-d528b9579b70' pts/1]
The illumos Project     helios-2.0.22418        January 2024
root@oxz_crucible_ded43e9b:~# ipadm
ADDROBJ           TYPE     STATE        ADDR
lo0/v4            static   ok           127.0.0.1/8
lo0/v6            static   ok           ::1/128
oxControlService4/ll addrconf tentative fe80::8:20ff:fe75:e486%oxControlService4/10
oxControlService4/? static duplicate    fd00:1122:3344:21::21/64
```

andrewjstone:
The following shows that the same ips were generated for the same sled in the same blueprint. I regenerated blueprints from the same parent twice, but only set one to the target.
```
root@[fd00:1122:3344:102::3]:32221/omicron> select blueprint_id,sled_id,zone_type,underlay_address from bp_omicron_zone;                                                                                                                             
              blueprint_id             |               sled_id                |    zone_type    |   underlay_address
---------------------------------------+--------------------------------------+-----------------+------------------------
[...]

PR #5005, opened 2024-02-06
planner generated duplicate IPs

davepacheco:
This addresses one of the two problems mentioned in #4990.

For the record: the fix is pretty trivial.  Much of the delta here is:

- I updated the diff output to include IPs so that we could see this particular issue more easily
- I updated the tests to check this condition
- I refactored the way we track Omicron zones as @jgallagher suggested to make this kind of bug harder to introduce

### DIFF
--- a/nexus/deployment/src/blueprint_builder.rs
+++ b/nexus/deployment/src/blueprint_builder.rs
@@ -77,7 +77,7 @@ pub struct BlueprintBuilder<'a> {
 
     // These fields will become part of the final blueprint.  See the
     // corresponding fields in `Blueprint`.
-    omicron_zones: BTreeMap<Uuid, OmicronZonesConfig>,
+    zones: BlueprintZones<'a>,
     zones_in_service: BTreeSet<Uuid>,
     creator: String,
     comments: Vec<String>,
@@ -151,7 +151,7 @@ impl<'a> BlueprintBuilder<'a> {
             parent_blueprint,
             policy,
             sled_ip_allocators: BTreeMap::new(),
-            omicron_zones: BTreeMap::new(),
+            zones: BlueprintZones::new(parent_blueprint),
             zones_in_service: parent_blueprint.zones_in_service.clone(),
             creator: creator.to_owned(),
             comments: Vec::new(),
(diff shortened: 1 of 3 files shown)


nexus/deployment/src/blueprint_builder.rs
```diff
@@ -398,7 +398,11 @@ impl<'a> BlueprintBuilder<'a> {
 
                 // Record each of the sled's zones' underlay addresses as
                 // allocated.
-                if let Some(sled_zones) = self.omicron_zones.get(&sled_id) {
+                let sled_zones =
+                    self.omicron_zones.get(&sled_id).or_else(|| {
+                        self.parent_blueprint.omicron_zones.get(&sled_id)
+                    });
+                if let Some(sled_zones) = sled_zones {
```
davepacheco, line 405: I tried to replace this with the new helper `sled_current_omicron_zones()` below but I get:

```
error[E0502]: cannot borrow `*self` as immutable because it is also borrowed as mutable
   --> nexus/deployment/src/blueprint_builder.rs:377:67
    |
377 |             self.sled_ip_allocators.entry(sled_id).or_insert_with(|| {
    |             -----------------------                -------------- ^^ immutable borrow occurs here
    |             |                                      |
    |             |                                      mutable borrow later used by call
    |             mutable borrow occurs here
...
401 |                 for z in self.sled_current_omicron_zones(sled_id) {
    |                          ---- second borrow occurs due to use of `*self` in closure

For more information about this error, try `rustc --explain E0502`.
```

I understand the complaint and I've run into this a lot but I don't know generally how to avoid it.

nexus/deployment/src/blueprint_builder.rs
```diff
@@ -420,10 +357,109 @@ impl<'a> BlueprintBuilder<'a> {
...
+        BlueprintZones {
+            changed_zones: BTreeMap::new(),
+            parent_zones: &parent_blueprint.omicron_zones,
+        }
+    }
+
+    /// Returns a mutable reference to a sled's Omicron zones *because* we're
+    /// going to change them.  It's essential that the caller _does_ change them
+    /// because we will have bumped the generation number and we don't want to
+    /// do that if no changes are being made.
```
jgallagher, line 384: I can think of some ways we could try to enforce this in the API, but they'd be worse than what you have, so I have a question instead. By "essential" here you mean "the control plane will end up doing unnecessary work if we bump the generation without making any changes", right? It's not essential from a correctness point of view?

jgallagher (approved):
I really like how `BlueprintZones` turned out :+1:

nexus/deployment/src/blueprint_builder.rs
davepacheco: I think that's right.  And yeah, same: I thought about just tracking the zone list here, and figuring out whether it's changed at the end, and bumping the generation if so, but it didn't seem worth the complexity given the cost of getting this wrong.

andrewjstone (approved):
Thanks for the quick fix @davepacheco

Issue #5002, opened 2024-02-06
Sled agent fails to get past ensuring ledgered zones when a dataset goes missing


augustuswm:
#4853 added checks for properties on datasets during sled-agent startup (`ensure_all_omicron_zones`). This requires that datasets that have been recorded to the `services.json` ledger are available on the next startup. If any dataset is not ready, `ensure_all_omicron_zones` will bail out and restart after a timeout. If a dataset has gone missing (i.e. disk removal) then this function can never complete and will loop forever, stopping the remaining zones from starting, and any other work sled-agent needs to do after the zones start.

While we don't support drive removal currently we likely want to allow getting past this invalid state in the meantime.

Log message that is repeated:
```
17:29:21.438Z WARN SledAgent: Failed to load services, will retry in 223.810606507s
    error = GetZfsValue { zone: "oxz_crucible_41f07d39-fcc0-4796-8b7c-7cfcd9135f78", source: GetValueError { filesystem: "oxp_4b90abdc-3348-4158-bedc-5bcd56e281d8/crucible", name: "[\\"zoned\\", \\"canmount\\", \\"encryption\\"]", err: Execution(CommandFailure(CommandFailureInfo { command: "/usr/sbin/zfs get -Ho value zoned,canmount,encryption oxp_4b90abdc-3348-4158-bedc-5bcd56e281d8/crucible", status: ExitStatus(unix_wait_status(256)), stdout: "", stderr: "cannot open 'oxp_4b90abdc-3348-4158-bedc-5bcd56e281d8/crucible': dataset does not exist\\n" })) } }
    file = sled-agent/src/sled_agent.rs:612
    sled_id = da89d292-1e33-4c21-8cc2-65b6ef9cbb1b

```

smklein:
See also: https://github.com/oxidecomputer/omicron/issues/2973

In `ensure_all_omicron_zones`, the check we added [validates that the dataset being used by the zone exists before using it](https://github.com/oxidecomputer/omicron/blob/7e0ce9905ecf46e815f370b4cb5b1dfc05a36096/sled-agent/src/services.rs#L2985-L3048), but this is basically intended to give a better error message. We expect that starting the zone would also fail if it's configured to use a missing dataset. I guess we must have been "starting" the zone, but not seeing the failure to boot.

I agree that "unplugging a disk on a sled that's powered off" needs to be a case we support. There's more work upstack to make sure nexus fully reconciles the view of the world, but in the meantime, I'll do the following:

[...]

PR #5008, opened 2024-02-06
[sled-agent] Ignore zones with invalid storage

smklein:
Partial fix to https://github.com/oxidecomputer/omicron/issues/5002

Implements a solution from @davepacheco : 

> if a zone depends on a U.2 that is not available, we skip it altogether (we treat it as though it's not in the list of zones to start)


smklein:
Closing this in favor of https://github.com/oxidecomputer/omicron/pull/5012. See #5002 for rationale

PR #5012, opened 2024-02-07
[sled-agent] Initialize zones more independently of each other

smklein:
Resolves https://github.com/oxidecomputer/omicron/issues/5002

This change attempts to make zone setup simpler and more independent. Namely: If any particular zone cannot start, due to NTP timesync, internal DNS lookup, or a missing disk, it should be able to fail without necessarily preventing all other zones from initializing.


sled-agent/src/services.rs
```diff
@@ -893,84 +965,9 @@ impl ServiceManager {
...
-                    "Time not yet synchronised (retrying in {:?})",
-                    delay;
-                    "error" => ?error
-                );
-            },
-        )
-        .await
-        .expect("Expected an infinite retry loop syncing time");
-
-        let mut existing_zones = self.inner.zones.lock().await;
```
smklein, line 966: This ordering has been removed under the assumption that:

- The NTP zone, which depends on internal DNS in the call to `opte_ports_needed`, will either successfully access *some* internal DNS server (possibly not on this sled!) or fail with a timeout before starting.
- Zones which depend on time being synchronized (which is true for all zones other than NTP and DNS) will fail to start within `start_omicron_zone`, when we check the `time_is_synchronized` boolean.

If these assumptions are upheld, I think this "order-independent" approach to calling `ensure_all_omicron_zones` seems cleaner. It can just be re-called as many times as it needs to be, and zones can keep getting added.

sled-agent/src/services.rs
```diff
@@ -2878,211 +2927,224 @@ impl ServiceManager {
...
+    // created (or destroyed) intentionally. As a result, even in error cases,
+    // it's possible that the set of `existing_zones` changes. However, this set
+    // will only change in the direction of `new_request`: zones will only be
+    // removed if they ARE NOT part of `new_request`, and zones will only be
+    // added if they ARE part of `new_request`.
+    // - Zones are not updated in-place: two zone configurations that differ
+    // in any way are treated as entirely distinct.
+    // - This method does not record any information such that these services
+    // are re-instantiated on boot.
+    async fn ensure_all_omicron_zones(
```
smklein, line 2943: I think a fair amount of complexity with the old version of this method came it trying to do a three-way merge between:

- "What was the old config (values in `old_config`)"?
- "What is the new config (values in `new_request`)"?
- "What are the set of zones that the Sled Agent knows about (`existing_zones`)"?
- (and arguably, a fourth: "What are the actual set of zones that the OS reports?")

While re-working this, I realized that the information contained within `old_config` **does not matter** -- we don't really care what the old configuration was trying to do, we really just need to move "current state of the system" to try to track "new request" as best as we can.

By cutting out both `old_config` and the "order-dependent filtering", IMO, this method is a little simpler.

sled-agent/src/services.rs
```diff
@@ -2878,211 +2927,224 @@ impl ServiceManager {
...
+            if let Err(e) = zone.runtime.stop().await {
+                error!(log, "Failed to stop zone {}: {e}", zone.name());
+            }
+        } else {
+            warn!(log, "Expected to remove zone, but could not find it");
+        }
+    }
+
+    // Ensures that if a zone is about to be installed, it does not exist.
+    async fn ensure_removed(
```
smklein, line 3039: This method is *almost* the same as it was before. We used to only check `existing_zones` to decide which zones to look at, and then we used to "ignore" running zones, because if it's already up, the zone must have already been successfully launched by someone.

In retrospect, I think this wasn't the right choice. We only called this code in a context where we want to **explicitly** add a new zone that we don't know about (see: `zones_to_be_added`), and the goal is to remove detritus from the system ("Get to a clean slate") before trying to start a new zone. Trying to handle the state of "the zone could be okay, or it could be bad an need removal, and we need to decide which is which" made this hard to reason about.

In reality, the calls to `zone_bundle_and_try_remove` for `zones_to_be_removed` should be the "normal" way of removing zones, but this slightly updated method just acts as a conservative back-stop to clear anything that could prevent a new zone from being started.

sled-agent/src/services.rs
```diff
@@ -4536,88 +4632,6 @@ mod test {
         logctx.cleanup_successful();
     }
 
-    #[tokio::test]
-    async fn test_old_ledger_migration_continue() {
```
smklein, line 4540: I think this test passed because of somewhat erroneous implementation details (see: my comments on the "three way merge of configs and zones" being way too complicated).

The set of zones in `tests/old-service-ledgers/rack2-sled10.json` contains many zones -- like CockroachDB, Crucible, etc -- that require explicit datasets to be present. We do nothing in the test to create those datasets. As a result, IMO, the call to `ensure_all_omicron_zones_persistent` below **should** be failing, because we cannot re-instantiate this configuration on a device missing those datasets.

That's what's happening now, but it undermines the value of this test, so I removed it.

sled-agent/src/services.rs
```diff
@@ -551,7 +597,33 @@ enum SledLocalZone {
...
-type ZoneMap = BTreeMap<String, RunningZone>;
+// The return type for `start_omicron_zones`.
+//
+// When multiple zones are started concurrently, some can fail while others
+// succeed. This structure allows the function to return this nuanced
+// information.
+#[must_use]
+struct StartZonesResult {
+    // The set of zones which have successfully started.
+    zones: Vec<OmicronZone>,
```
jgallagher, line 608: Tiny nit - maybe name this field indicating success (`zones_started` or `successful_zones` or somesuch)? I could imagine thinking that this is "the whole set of zones we tried to start" if I were skimming through this.

sled-agent/src/services.rs
```diff
@@ -2688,56 +2685,103 @@ impl ServiceManager {
...
+        all_u2_pools: &Vec<ZpoolName>,
+    ) -> Result<OmicronZone, Error> {
+        // Ensure the zone has been fully removed before we try to boot it.
+        //
+        // This ensures that old "partially booted/stopped" zones do not
+        // interfere with our installation.
+        self.ensure_removed(&zone).await?;
+
+        // If this zone requires timesync and we aren't ready, fail it early.
+        if zone_requires_timesync(&zone.zone_type) && !time_is_synchronized {
```
jgallagher, line 2715: Tiny nit - `zone_requires_timesync` might be a bit cleaner as a method on `OmicronZoneType`?

sled-agent/src/services.rs
```diff
@@ -2688,56 +2685,103 @@ impl ServiceManager {
...
-        let futures = requests.iter().map(|request| {
-            async move {
-                self.initialize_zone(
-                    ZoneArgs::Omicron(request),
-                    // filesystems=
-                    &[],
-                    // data_links=
-                    &[],
-                )
+        let futures = requests.clone().map(|zone| async move {
```
jgallagher, line 2762: I think this clone isn't needed

```suggestion
        let futures = requests.map(|zone| async move {
```

sled-agent/src/services.rs
```diff
@@ -2878,211 +2927,224 @@ impl ServiceManager {
...
-                    .inner
-                    .zone_bundler
-                    .create(&zone, ZoneBundleCause::UnexpectedZone)
-                    .await
+            }
+            if let Err(e) = zone.runtime.stop().await {
+                error!(log, "Failed to stop zone {}: {e}", zone.name());
+            }
+        } else {
+            warn!(log, "Expected to remove zone, but could not find it");
```
jgallagher, line 3034: Maybe add `expected_zone_name` to this log?

Tiny style nitpick: You could structure this as

```rust
let Some(mut zone) = existing_zones.remove(&expected_zone_name) else {
    warn!(...);
    return;
};
```

to shave an indentation level off the bulk of the method.

jgallagher (commented):
These changes LGTM; all my comments are ignorable nitpicks in the interest of urgency.

Should we run this through RSS and cold boot on madrid to confirms we didn't miss something subtle?

andrewjstone (approved):
These changes look good to me. All the behavior seems to track with retries happening from within `load_services` rather than nested just for timesync. Nice work @smklein!

andrewjstone:
> Should we run this through RSS and cold boot on madrid to confirms we didn't miss something subtle?

I'll give this a run on the testbed as a sanity check. I'll leave it up to you guys to decide if it should also run on madrid.

andrewjstone:
> > Should we run this through RSS and cold boot on madrid to confirms we didn't miss something subtle?
> 
> I'll give this a run on the testbed as a sanity check. I'll leave it up to you guys to decide if it should also run on madrid.

I ran this branch through RSS and all zones came up as expected. I also coldbooted sled `g0` and all zones relaunched correctly.

sled-agent/src/services.rs
smklein: Renamed to `new_zones`!

sled-agent/src/services.rs
smklein: Agreed, updated to become a method there.

sled-agent/src/services.rs
smklein: Removed!

sled-agent/src/services.rs
smklein: Done and done

sled-agent/src/services.rs
```diff
@@ -2878,211 +2927,224 @@ impl ServiceManager {
...
+            requested_zones_set.difference(&old_zone_configs);
 
-        // For each new zone request, ensure that we've sufficiently
-        // synchronized time.
-        //
-        // NOTE: This imposes a constraint, during initial setup, cold boot,
-        // etc, that NTP and the internal DNS system it depends on MUST be
-        // initialized prior to other zones.
+        // Destroy zones that should not be running
+        for zone in zones_to_be_removed {
```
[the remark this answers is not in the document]
smklein: Agreed, tracking in #5024

augustuswm:
When this is ready I'm available to get the build on to dogfood, and then the colo.

PR #5007, opened 2024-02-06
pin libxmlsec1 1.3.2 in CI on macOS

iliana:
A [temporary](https://xkcd.com/2730/) hack to keep the macOS job mildly happy while I muster the energy to fix this upstream in samael.

PR #5013, opened 2024-02-07
bump maghemite

rcgoodfellow:


rcgoodfellow:
looks like macos issues will be fixed in 
- #5007

PR #5015, opened 2024-02-07
Remove articles from API endpoint summaries

### LINKED ISSUE #4741: Remove articles from endpoint summaries
They're just noise IMO.

image

image

Example spot:

https://github.com/oxidecomputer/omicron/blob/9274beb0112e50014e8f14cba2f3fbec17a64421/nexus/src/external_api/http_entrypoints.rs#L1911

david-crespo:
Closes #4741 

The process here was pretty cool. I fed all the lines to GPT-4 and got it to give me a giant `sed` command to update all the summaries. It had a couple of issues as written, but I fixed them. Details [here](https://gist.github.com/david-crespo/b5cd941045dc518a61e1eab1a5d1f50e). Then I manually went through and tweaked everything.

### DIFF
--- a/nexus/src/external_api/http_entrypoints.rs
+++ b/nexus/src/external_api/http_entrypoints.rs
@@ -411,7 +411,7 @@ async fn ping(
     Ok(HttpResponseOk(views::Ping { status: views::PingStatus::Ok }))
 }
 
-/// Fetch the top-level IAM policy
+/// Fetch top-level IAM policy
 #[endpoint {
     method = GET,
     path = "/v1/system/policy",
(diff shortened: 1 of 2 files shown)


nexus/src/external_api/http_entrypoints.rs
```diff
@@ -651,7 +651,7 @@ async fn silo_quotas_view(
     apictx.external_latencies.instrument_dropshot_handler(&rqctx, handler).await
 }
 
-/// Update the resource quotas of a given silo
+/// Update resource quotas for silo
```
benjaminleonard, line 654: Was going to suggest getting this smaller with: "Update silo resource quotas" – but probably less readable.

benjaminleonard (approved):
Love, thanks!

zephraph (approved):
This is great! Thanks!

Issue #4920, opened 2024-01-29
macOS homebrew `libxmlsec1` 1.3.3 incompatible with `samael`


david-crespo:
```
   Compiling samael v0.0.14
error[E0432]: unresolved import `crate::bindings::xmlSecOpenSSLAppKeyLoad`
  --> /Users/david/.cargo/registry/src/index.crates.io-6f17d22bba15001f/samael-0.0.14/src/xmlsec/backend.rs:14:9
   |
14 |         xmlSecOpenSSLAppKeyLoad as xmlSecCryptoAppKeyLoad,
   |         -----------------------^^^^^^^^^^^^^^^^^^^^^^^^^^
   |         |
   |         no `xmlSecOpenSSLAppKeyLoad` in `bindings`
   |         help: a similar name exists in the module: `xmlSecOpenSSLAppKeyLoadEx`

   Compiling mg-admin-client v0.1.0 (/Users/david/oxide/omicron/clients/mg-admin-client)
   Compiling dpd-client v0.1.0 (/Users/david/oxide/omicron/clients/dpd-client)
For more information about this error, try `rustc --explain E0432`.
error: could not compile `samael` (lib) due to previous error
```

[xmlsec](https://www.aleksey.com/xmlsec/) 1.3.3 came out on 2024-01-04, but `libxmlsec1` on homebrew was only bumped from 1.3.2 to 1.3.3 [yesterday](https://github.com/Homebrew/homebrew-core/commits/f5c6266aaf51084a7f9e5d1917cb8f566fee0395/Formula/lib/libxmlsec1.rb). The samael dep in omicron was bumped to 0.0.14 [five days ago](https://github.com/oxidecomputer/omicron/commit/e261a960cb365ad92f103a35b262713118ea6441), and I'm pretty sure I had #4903 building locally on top of that, so I believe this is primarily about `libxmlsec1`.

There seems to be no way to force homebrew to install `1.3.2` to test the problem. However, I went into `/Users/david/.cargo/registry/src/index.crates.io-6f17d22bba15001f/samael-0.0.14/src/xmlsec/backend.rs` and changed the [offending line](https://github.com/njaremko/samael/blob/4b4811d97ae324da1376624e77e82af18ee108f3/src/xmlsec/backend.rs#L14) to reference `xmlSecOpenSSLAppKeyLoadEx` (I know this is not good) and it did let it compile and let me run an integration test (that does not use SAML).

[...]

PR #5017, opened 2024-02-07
tools: use SHA256 in `ci_download_cockroachdb`

hawkw:
Depends on #4961.

Currently, the `tools/ci_download_cockroachdb` script uses MD5 as the checksum for the Cockroachdb tarball. This is unfortunate for two reasons:

1. Upstream Cockroachdb (and the corresponding Illumos build) publish SHA256 digests for these tarballs, rather than MD5s (see https://www.cockroachlabs.com/docs/releases/ and https://illumos.org/downloads/). Using SHA256 rather than MD5 digests should make updating to a new version easier.

2. Nix requires SHA256 checksums for files downloaded as build inputs. Currently, the Nix flake can use the SHA256 checksums for Maghemite and Dendrite from `tools/maghemite_mgd_checksums` and `tools/dendrite_stub_checksums`, meaning that updating these versions does not require manually changing the Nix flake. However, because we use MD5 rather than SHA256 checksums for Cockroachdb, updating the Cockroachdb version requires manually changing the version in the Nix flake, which is a shame, especially if someone unfamiliar with Nix has to do it...

This commit changes `tools/cockroachdb_checksums` and the corresponding `tools/ci_download_cockroachdb` script to use SHA256 rather than MD5. I've changed the Nix flake to read the hash from this file rather than hard-coding it, so now, the cockroachdb version can be updated without touching the flake.


smklein (approved):
Rad, thanks for updating this!

/me wonders about blake3 being here someday

faithanalog:
yeah the macOS build failure is #4920 dont worry about it

PR #5018, opened 2024-02-07
[RSS] fix split-brain rack subnet

### LINKED ISSUE #5009: Newly added sleds are not on rack subnet
I added a sled `g2` to an existing rack. The subnet should be `fd00:1122:3344:100/56`, but I ended up with a sled ip of `fd00:1122:3344:21::1`

jgallagher:
Prior to this commit, RSS accepted a set of parameters that included a `rack_subnet` and an optional (but not really) `rack_network_config`; the `rack_network_config` _also_ contained a `rack_subnet` property. The first `rack_subnet` was used by RSS to pick sled addresses, but the second is what it handed off to Nexus to record in the database.

This PR makes three changes to parameters (the bulk of the diff is the expected fallout from them):

* Removes the top-level `rack_subnet` field; only the `rack_network_config.rack_subnet` remains.
* Makes `rack_network_config` non-optional. The handoff to Nexus would fail if this was `None`, so now it's always required. (This was only a little annoying in a few tests where we now have to cons up a fake network config.)
* Changes wicket/wicket to accept a subset of `rack_network_config` that does _not_ include a `rack_subnet`; this is a value the control plane should choose on its own.

One potentially-dangerous change is that the RSS parameters changed are not just passed when RSS is run; they're also serialized to disk as `rss-sled-plan.json`. We have a test to ensure changes don't affect the schema of this plan, but I believe the changes here are backwards compatible (an old plan that has a no-longer-present `rack_subet` is fine, and the JSON representation of the optional `RackNetworkConfig` is that it can be either `null` or an object; we'll fail to read any plans with `null`, but those would have failed to handoff to Nexus anyway as noted above). To check this is right, I pulled the `rss-sled-plan.json` off of madrid, censored the certs, replaced the password hash with one of our test hashes, and added a test that we can still read it.

---

Changes that might make sense but I didn't attempt:

* Changing the `rack_network_config` in the early networking bootstore to be non-optional. I think this would be correct, but is probably more trouble than it's worth to migrate. We might consider this the next time we make other, unrelated changes here though.
* Removing the `rack_subnet` field not just from user -> wicket, but also from {wicket,developer} -> RSS. We could make RSS pick its own rack subnet, maybe? This seemed dubious enough I stopped. This does mean the TOML files used to automatically launch RSS still have a `rack_subnet` value, but now it's only one (under the rack network config) instead of two.
* Changing the rack subnet choice to be random. wicket continues to use the hardcoded value we've been using.

---

I also fixed a handful of places where we define the rack subnet `fd00:1122:3344:01::/56`; I believe this is just wrong / a typo. The `:01` at the end is equivalent to `:0001`, which is equivalent to the /56 `fd00:1122:3344:0000::/56`. Every place we had this we meant to use `fd00:1122:3344:0100::/56`, so I changed all them (I think!).

Fixes #5009, but only for any racks that run RSS after this change. I am not attempting to retroactively correct any racks that had the wrong `rack_subnet` recorded in the database, as I believe all such deployed racks are dev systems that are frequently wiped and reinstalled.


andrewjstone:
> Changes that might make sense but I didn't attempt:

Changing the rack_network_config in the early networking bootstore to be non-optional. I think this would be correct, but is probably more trouble than it's worth to migrate. We might consider this the next time we make other, unrelated changes here though.
Removing the rack_subnet field not just from user -> wicket, but also from {wicket,developer} -> RSS. We could make RSS pick its own rack subnet, maybe? This seemed dubious enough I stopped. This does mean the TOML files used to automatically launch RSS still have a rack_subnet value, but now it's only one (under the rack network config) instead of two.
Changing the rack subnet choice to be random. wicket continues to use the hardcoded value we've been using

FWIW, I think all three of these decisions are absolutely the right call at this juncture. I also think you are correct not to change the rack subnet in CRDB at runtime on existing setups. We should validate with customers that the value in the DB is what we expect and can correct this manually if necessary, which I don't believe it will be considering the timing of when this change in behavior was introduced.

sled-agent/src/bootstrap/params.rs
```diff
@@ -14,7 +14,7 @@ use serde::{Deserialize, Serialize};
 use sha3::{Digest, Sha3_256};
 use sled_hardware::Baseboard;
 use std::borrow::Cow;
-use std::collections::HashSet;
+use std::collections::BTreeSet;
```
smklein, line 17: Ah, is this for consistent order?

smklein (approved):
Makes sense to me, thanks for including the config from Madrid as a test case!

andrewjstone (approved):
Looks great! Thanks John!

sled-agent/src/bootstrap/params.rs
jgallagher: Yeah, for the expectorate test I added with madrid's config. With `HashSet`/`HashMap` it's a lot more annoying to check the parsed value.

smf/sled-agent/gimlet-standalone/config-rss.toml
```diff
@@ -88,7 +80,14 @@ last = "192.168.1.29"
 # Configuration to bring up Boundary Services and make Nexus reachable from the
 # outside.  See docs/how-to-run.adoc for more on what to put here.
 [rack_network_config]
-rack_subnet = "fd00:1122:3344:01::/56"
+# The /56 subnet for this rack.  This subnet is internal to the rack and fully
+# managed by Omicron, so you can pick anything you want within the IPv6 Unique
+# Local Address (ULA) range.  The rack-specific /56 subnet also implies the
+# parent /48 AZ subnet.
+#              |............|    <- This /48 is the AZ Subnet
```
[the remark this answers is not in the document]
jgallagher: Thanks, but not mine. :) I just moved this from above.

PR #5021, opened 2024-02-07
[omdb] Add `db rack list`

jgallagher:
In the context of #5009, it would be good to be able to quickly confirm whether a rack has the correct or incorrect subnet recorded in CRDB.

Madrid is incorrect (as expected):

```
root@oxz_switch1:~# /var/tmp/john/omdb db rack list
ID                                   INITIALIZED TUF_BASE_URL RACK_SUBNET
ed6bcf59-9620-491d-8ebd-4a4eebf2e136 true        -            fd00:1122:3344:1::/56
```

Rack2 is correct (not surprising, since this would've been set after RSS ran, which means it wouldn't hit #5009, although I don't know _exactly_ how this did get set in practice):

```
root@oxz_switch1:~# /var/tmp/john/omdb db rack list
ID                                   INITIALIZED TUF_BASE_URL RACK_SUBNET
de608e01-b8e4-4d93-b972-a7dbed36dd22 true        -            fd00:1122:3344:100::/56
```

### DIFF
--- a/dev-tools/omdb/src/bin/omdb/db.rs
+++ b/dev-tools/omdb/src/bin/omdb/db.rs
@@ -160,6 +160,8 @@ pub struct DbFetchOptions {
 /// Subcommands that query or update the database
 #[derive(Debug, Subcommand)]
 enum DbCommands {
+    /// Print information about the rack
+    Rack(RackArgs),
     /// Print information about disks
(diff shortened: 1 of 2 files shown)


andrewjstone (approved):
Looks awesome. I'm guessing the limit around racks is not an issue right now :)

Issue #5014, opened 2024-02-07
Blueprint Execution Zone Deployment rewrites `omicron-zones.json` unnecessarily


andrewjstone:
Almost every time we sync, we seem to be committing a new version of the ledger even though nothing has changed. This will cause unnecessary wear on the U.2s on a real system, and is generally just bad form.

PR #5025, opened 2024-02-08
[sled-agent] Avoid unnecessary ledger commit

smklein:
Fixes https://github.com/oxidecomputer/omicron/issues/5014


andrewjstone (approved):
Ha, that was the exact fix I was going to make this morning before our long discussion and my punt to you! Glad we agree :D

PR #5029, opened 2024-02-08
[nexus] Improve logging when transactions retry

smklein:
I was trying to determine whether or not transaction retries spiked alongside a CRDB crash, and found that the current mechanism of tracking "which transactions did we retry" could use a little improvement.

Although we do send these retry attempts to Clickhouse (and I'm able to see them for historical data) this doesn't help us if Nexus crashes before those queries can be pulled by Oximeter.

This PR adds logging for these cases:
- It logs with `warn` if we retry at all
- It logs with `info` if a retried transaction completes

### DIFF
--- a/nexus/db-queries/src/transaction_retry.rs
+++ b/nexus/db-queries/src/transaction_retry.rs
@@ -9,6 +9,7 @@ use chrono::Utc;
 use diesel::result::Error as DieselError;
 use oximeter::{types::Sample, Metric, MetricsError, Target};
 use rand::{thread_rng, Rng};
+use slog::{info, warn, Logger};
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
@@ -60,6 +61,10 @@ impl RetryHelperInner {
         Self { start: Utc::now(), attempts: 1 }
     }
 
+    fn has_retried(&self) -> bool {
+        self.attempts > 1
+    }
+
     fn tick(&mut self) -> Self {
         let start = self.start;
         let attempts = self.attempts;
@@ -74,6 +79,7 @@ impl RetryHelperInner {
 /// Helper utility for tracking retry attempts and latency.
 /// Intended to be used from within "transaction_async_with_retry".
 pub struct RetryHelper {
+    log: Logger,
     producer: Producer,
     name: &'static str,
     inner: Mutex<RetryHelperInner>,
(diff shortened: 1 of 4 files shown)


smklein:
I'll get a doc up for "how to poke at clickhouse for this info" too, since this took me a sec, and is probably also useful to write down somewhere.

nexus/db-queries/src/transaction_retry.rs
```diff
@@ -151,14 +182,13 @@ impl RetryHelper {
...
-        self,
+    // Converts this function to a retryable callback that can be used from
+    // "transaction_async_with_retry".
+    fn as_callback(
+        self: Arc<Self>,
     ) -> impl Fn() -> futures::future::BoxFuture<'static, bool> {
-        let r = Arc::new(self);
         move || {
-            let r = r.clone();
+            let r = self.clone();
```
sunshowers, line 191: Is this clone still necessary?

sunshowers (approved):

nexus/db-queries/src/transaction_retry.rs
smklein: Without the clone:

```
error[E0507]: cannot move out of `self`, a captured variable in an `Fn` closure
   --> nexus/db-queries/src/transaction_retry.rs:191:21
    |
188 |         self: Arc<Self>,
    |         ---- captured outer variable
189 |     ) -> impl Fn() -> futures::future::BoxFuture<'static, bool> {
190 |         move || {
    |         ------- captured by this `Fn` closure
191 |             let r = self;
    |                     ^^^^ move occurs because `self` has type `Arc<RetryHelper>`, which does not implement the `Copy` trait
    |
help: consider borrowing here
    |
191 |             let r = &self;
    |                     +

For more information about this error, try `rustc --explain E0507`.
error: could not compile `nexus-db-queries` (lib) due to previous error
```

I don't actually want to give a reference there, because I want to "move the Arc into the closure", but for whatever reason, that `move` isn't letting rust know to move the arc into the closure...

sunshowers: Oh, because the return value is an `impl Fn` which can be called many times. Never mind

smklein: Ah, yeah, it has to be -- it's retryable, since that's the whole point of this.

PR #5030, opened 2024-02-08
Add documentation example for querying Clickhouse

smklein:
This is somewhat of a corollary to https://github.com/oxidecomputer/omicron/pull/5029. While debugging this info,
it took me a while to figure out what exactly to poke, so I figured I'd document that experience.

### DIFF
--- a/docs/clickhouse-debugging.adoc
+++ b/docs/clickhouse-debugging.adoc
@@ -31,7 +31,7 @@ $ pilot host login <cubby where clickhouse exists>
 $ pfexec zlogin oxz_clickhouse_<UUID>
 ----
 
-. **Identify the IP address of Clickhouse**. This is psosible using `ipadm`:
+. **Identify the IP address of Clickhouse**. This is possible using `ipadm`:
 +
 [source,bash]
 ----
--- a/docs/clickhouse-debugging.adoc
+++ b/docs/clickhouse-debugging.adoc
@@ -1,0 +1,199 @@
+:showtitle:
+:numbered:
+:toc: left
+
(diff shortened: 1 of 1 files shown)


david-crespo (approved):

docs/clickhouse-debugging.adoc
```diff
@@ -0,0 +1,199 @@
...
+lo0/v6            static   ok           ::1/128
+oxControlService8/ll addrconf ok        fe80::8:20ff:fe35:6b0a%oxControlService8/10
+oxControlService8/omicron6 static ok    fd00:1122:3344:107::4/64   <-- It's this one!
+----
+. **Log into Clickhouse using the CLI**
++
+[source,bash]
+----
+# Run from within the Clickhouse zone
+$ /opt/oxide/clickhouse/clickhouse client --host fd00:1122:3344:107::4
```
sunshowers, line 51: Can this process be automated?

sunshowers (approved):
Looks great!

docs/clickhouse-debugging.adoc
smklein: I want it to be. I'm kinda curious if omdb can handle a lot of this stuff, or if `oxdb` can help here - I know that's a tool that can be used to query clickhouse, but I didn't see it in any of the packaging stuff, so I wasn't sure if it was available on deployed racks.

But anyway, yeah, my dream would be:
- omdb has a command to find clickhouse via DNS
- omdb can then either call into oxdb, or re-use that functionality as a library...
- ... and as a caller, running something like `omdb clickhouse query <timeseries_name>` and seeing all data would be rad

PR #5035, opened 2024-02-08
[nix flake] add `rust-src` component

hawkw:
Currently, when using the Nix flake managed Rust toolchain, `rust-analyzer` complains that it can't expand proc macros, because it can't find the sysroot. Adding the `rust-src` Rustup component fixes this. This PR adds the `rust-src` component to the flake's Rust toolchain.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -17,7 +17,11 @@
         system = "x86_64-linux";
       };
       # use the Rust toolchain defined in the `rust-toolchain.toml` file.
-      rustToolchain = pkgs.pkgsBuildHost.rust-bin.fromRustupToolchainFile ./rust-toolchain.toml;
+      rustToolchain = (pkgs.pkgsBuildHost.rust-bin.fromRustupToolchainFile ./rust-toolchain.toml).override {
+        extensions = [
+          "rust-src" # for rust-analyzer
+        ];
+      };
 
       buildInputs = with pkgs; [
         # libs


sunshowers:
Should this be specified in `rust-toolchain.toml`?

iliana:
> Should this be specified in `rust-toolchain.toml`?

Since this is only needed for rust-analyzer usage, which is an optional workflow, probably not.

hawkw:
> Should this be specified in `rust-toolchain.toml`?

It could be, I'll change that.

iliana (approved):

hawkw:
> > Should this be specified in `rust-toolchain.toml`?
> 
> Since this is only needed for rust-analyzer usage, which is an optional workflow, probably not.

Thinking about that, it occurs to me that it might be more Ideologically Correct to only add it in the `devShell`, so that if we added a package to the Nix flake for building Omicron itself, that wouldn't need to download the `rust-src` component. But, we don't currently have a package for building Omicron itself.

sunshowers:
Discussed in #oxide-control-plane on Matrix -- the crucial difference here is that Nix installations need that directive [in order to set an environment variable](https://nixos.wiki/wiki/Rust#Shell.nix_example), while non-Nix users can simply run `rustup component add rust-src` and have things work.

sunshowers (approved):

PR #5036, opened 2024-02-09
Add ordering to nat changeset

internet-diglett:
When querying the DB for nat entries, it is not sufficient to
filter for entries with a version_added and version_removed greater
than a specified value, then order. The sorted columns need to be
interleaved as well during the query, or we may accidentally skip
entries. This commit adds a view to the db that enables such a query.


schema/crdb/dbinit.sql
```diff
@@ -3436,14 +3436,57 @@ STORING (
   time_deleted
 );
 
+/**
+ * A view of the ipv4 nat change history
```
rcgoodfellow, line 3440: Not blocking but a comment of what this union is doing and why it's here would be nice.

rcgoodfellow (approved):

schema/crdb/dbinit.sql
internet-diglett: I actually found a bug in the original view while writing the comments 😭 
So I'm actually very thankful that you had me go back and write some comments 😂

andrewjstone (approved):
Looks good. Solid test!

PR #5038, opened 2024-02-09
Remove thing-flinger

### LINKED ISSUE #5034: Remove thing-flinger
This was a point-in-time creation and is no longer needed with the falcon a4x2 testbed and wicket.

andrewjstone:
Thing-flinger was used to deploy multinode omicron clusters running on commodity hardware for early testing needs. This was primarily used for early experimentation with trust-quorum secret sharing and other preliminary multi-node experiments. It deployed via SSH and was never intended to be used forever. It was a stop-gap that nobody has used in a while now that we have real hardware with wicket and the a4x2 falcon testbed.

Fixes #5034


smf/mgs-sim/config.toml
```diff
@@ -16,7 +16,7 @@ request_body_max_bytes = 536870912
 [switch]
```
smklein, line 16: These SMF config changes don't seem related to thing-flinger -- are they?

smf/sled-agent/non-gimlet/config.toml
```diff
@@ -35,46 +35,36 @@ zpools = [
...
-swap_device_size_gb = 64
+vmm_reservoir_percentage = 0
 
 # An optional data link from which we extract a MAC address.
 # This is used as a unique identifier for the bootstrap address.
 #
 # If empty, this will be equivalent to the first result from:
 # $ dladm show-phys -p -o LINK
-# data_link = "igb0"
+data_link = "vioif1"
```
smklein, line 49: Pretty sure this is causing the deployment job to fail

smf/mgs-sim/config.toml
andrewjstone: Oh jeez. This is what I get for pushing at 2 am. I must have done a `git add .` with the mods from the testbed. Thanks Sean!

smklein (approved):

PR #5040, opened 2024-02-09
`/lookup/*` route for console

david-crespo:
Needed to make https://github.com/oxidecomputer/console/pull/1944 work. Adding one top-level route `/lookup/*` so we can freely add more routes under it in the console without adding anything here.

Will hold on merging until after the release is finalized.


zephraph:
I was thinking of something along the same lines. Funny that it's somewhat a throwback to the original by-id endpoints.

zephraph (approved):

Issue #3722, opened 2023-07-20
Expose LRTQ live debug/status information


andrewjstone:
Talk to @davepacheco about his plans

andrewjstone:
Implemented by #5041

PR #5041, opened 2024-02-10
Add `omdb sled-agent bootstore status`

### LINKED ISSUE #4295: Add omdb plugin for bootstore

andrewjstone:
I ended up creating an API type, `BootstoreStatus`, that is suitable for JSON / OpenAPI.

Fixes #4295 and #3722

### DIFF
--- a/bootstore/src/schemes/v0/peer.rs
+++ b/bootstore/src/schemes/v0/peer.rs
@@ -62,6 +62,15 @@ pub enum NodeRequestError {
     },
 }
 
+impl From<NodeRequestError> for omicron_common::api::external::Error {
+    fn from(error: NodeRequestError) -> Self {
+        omicron_common::api::external::Error::internal_error(&format!(
+            "{:#}",
+            error
+        ))
+    }
+}
+
 /// A request sent to the `Node` task from the `NodeHandle`
 pub enum NodeApiRequest {
     /// Initialize a rack at the behest of RSS running on the same scrimlet as
--- a/bootstore/src/schemes/v0/peer.rs
+++ b/bootstore/src/schemes/v0/peer.rs
@@ -65,8 +65,7 @@ pub enum NodeRequestError {
 impl From<NodeRequestError> for omicron_common::api::external::Error {
     fn from(error: NodeRequestError) -> Self {
         omicron_common::api::external::Error::internal_error(&format!(
-            "{:#}",
-            error
+            "{error}"
         ))
     }
 }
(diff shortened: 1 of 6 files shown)


sled-agent/src/http_entrypoints.rs
```diff
@@ -972,3 +973,23 @@ async fn inventory(
...
+async fn bootstore_status(
+    request_context: RequestContext<SledAgent>,
+) -> Result<HttpResponseOk<String>, HttpError> {
+    let sa = request_context.context();
+    let bootstore = sa.bootstore();
+    let status = bootstore.get_status().await.map_err(|e| {
+        HttpError::from(omicron_common::api::external::Error::from(e))
+    })?;
+    // Status is full of maps with non-string keys, and using `serde_as`
+    // doesn't work with JSON. Since this is just used for omdb right now,
```
andrewjstone, line 991: should say doesn't work with `JsonSchema`

andrewjstone:
Output from running this on the testbed against the g0 sled-agent.

```
root@oxz_switch:~# omdb sled-agent --sled-agent-url http://[fd00:1122:3344:101::1]:12345 bootstore status
fsm ledger generation: 1
network config ledger generation: Some(1)
fsm state: initial_member
peers (found by ddmd):
    [fdb0:a840:2500:3::1]:12347
    [fdb0:a840:2500:5::1]:12347
    [fdb0:a840:2500:7::1]:12347
established connections:
     Pc { identifier: "g1", model: "i86pc" } : [fdb0:a840:2500:3::1]:12347
     Pc { identifier: "g2", model: "i86pc" } : [fdb0:a840:2500:5::1]:12347
     Pc { identifier: "g3", model: "i86pc" } : [fdb0:a840:2500:7::1]:12347
accepted connections:
    <none>
negotiating connections:
    <none>
```

bootstore/src/schemes/v0/peer.rs
```diff
@@ -62,6 +62,15 @@ pub enum NodeRequestError {
     },
 }
 
+impl From<NodeRequestError> for omicron_common::api::external::Error {
+    fn from(error: NodeRequestError) -> Self {
+        omicron_common::api::external::Error::internal_error(&format!(
+            "{:#}",
```
jgallagher, line 68: I don't think `external::Error` supports an alternate format; this could be just `{error}`. Arguably we should use slog-error-chain here to print the chain, but in practice it doesn't matter currently, because `external::Error` doesn't have any cases with chained errors.

jgallagher (approved):
LGTM :+1: 

I think your test failure is legit, but just an expectorate one

bootstore/src/schemes/v0/peer.rs
andrewjstone: A quick search shows we have `:#` forms littered throughout the codebase, although arguably this doesn't matter as it just ends up being a no-op.

jgallagher: Well, we also have `anyhow::Error`s littered throughout the codebase too. For those it _does_ matter. :man_shrugging:

Issue #4947, opened 2024-01-31
retire "services" table


davepacheco:
The `services` table stores information about Omicron zones deployed by RSS.  I think the intent was so that Nexus would know what was supposed to exist so that over time it could reconfigure the set of zones deployed.  Unfortunately, we discovered this table doesn't have enough information to fully specify these zones so it cannot be used (by itself) to describe either what zones _do_ exist nor what zones _should_ exist.  But fortunately we've already addressed this problem:

- A fuller specification of each Omicron zone is fetched from each sled agent as part of the inventory process.  This allows Nexus to know what zones are actually out there (as of some particular collection).
- Blueprints include this same representation of Omicron zones.  This allows Nexus to know what zones are _supposed_ to be out there.
- We use this representation in sled agent's `PUT /omicron-zones` API so we know it's sufficient to fully specify these zones.
- I think we've built enough of a blueprint planner to have some confidence that the overall representation works to evolve the set of deployed zones.

Given that, I think we can just drop the `services` table altogether.  There is only one thing that uses it as far as I know, and that's DNS propagation.  After oxidecomputer/omicron#4889, we should be free to drop this table altogether.

CC @smklein does this sound reasonable?

smklein:
Makes sense to me, I think the combination of inventory ("what exists") and blueprints ("what we want to exist") makes sense, and the services table is redundant with both of those.

jgallagher:
Summarizing discussion from yesterday's update sync: there are (at least) two other clients of the `service` table:

1. `omdb db network list-eips` uses it to map from an external IP to the kind of service (nexus vs ntp)
2.  `DataStore::vpc_resolve_to_sleds()` uses it to determine which sleds are running services that are using a service VPC (currently only `SERVICE_VPC_ID`, but we could imagine more fine-grained VPCs in the future)

#5202 changes both of those to also check the current target blueprint, but doesn't remove their use of `service`. To do so, I think we would need to first guarantee that every system has a current target blueprint. This should be straightforward for deployed systems, as we could set it to a _disabled_ blueprint based on the latest collection, but would require work to ensure it will be true for new systems going through RSS.

Separately, there's a question of whether these clients ought to be looking at the latest inventory instead of (or in addition to?) the current target blueprint. Both inventory and blueprints are potentially misleading, but in different ways (inventory collections are lossy, blueprints are aspirational).

davepacheco:
I think I found another use here:
[...]

PR #5044, opened 2024-02-12
[nexus] remove `services` queries & OMDB commands

hawkw:
Depends on #5033.

As described in #4947, we would like to remove the `services` table, as its only remaining use is DNS propagation. PR #5033 changes DNS propagation to no longer use the `services` table. Now that we're no longer consuming this table, this commit removes the one method which queries that table (`Datastore::services_list_kind`) and its two remaining callers, the OMDB commands that query the table. I've also removed the test for inserting and querying this table.

Note that I have *not* yet removed the code in RSS that actually populates this table when the rack is set up. I thought it might be desirable to still populate this data in case we have to roll back to a previous version of Nexus that uses the `services` table. If this isn't something we care about, I can remove that code as well, allowing us to remove the `Datastore` methods for inserting to `services`.


smklein:
> Note that I have _not_ yet removed the code in RSS that actually populates this table when the rack is set up. I thought it might be desirable to still populate this data in case we have to roll back to a previous version of Nexus that uses the `services` table. If this isn't something we care about, I can remove that code as well, allowing us to remove the `Datastore` methods for inserting to `services`.

I'm okay with doing this incrementally!

jgallagher:
> Now that we're no longer consuming this table, this commit removes the one method which queries that table (`Datastore::services_list_kind`) and its two remaining callers, the OMDB commands that query the table. I've also removed the test for inserting and querying this table.

I ran into two other places that are consuming this table.

1. [omdb db network list-eips](https://github.com/oxidecomputer/omicron/blob/55a0760a77a37333c51594229e06c9b598349c90/dev-tools/omdb/src/bin/omdb/db.rs#L1866-L1869) indirectly "joins" against `service` to attach a service kind to each service external IP. We could maybe use `external_ip.description` instead, although it looks like it's only populated for `nexus` and `external-dns` (??). Another option might be to join against the `service_network_interface` table, which has a service_id column, and we populate both the name and the description. From madrid:

```
/omicron> SELECT eip.ip,eip.name,eip.description,sni.name,sni.description FROM external_ip AS eip LEFT JOIN service_network_interface AS sni ON eip.parent_id = sni.service_id WHERE eip.is_service;
      ip      |                       name                        | description  |                       name                        |        description
--------------+---------------------------------------------------+--------------+---------------------------------------------------+----------------------------
  172.20.28.1 | external-dns-d3471b34-6c67-4aef-bf53-3fe70b65efb8 | external_dns | external-dns-d3471b34-6c67-4aef-bf53-3fe70b65efb8 | external_dns service vNIC
  172.20.28.6 | NULL                                              | NULL         | ntp-af51b1eb-1266-4235-b2fd-6526f6ef8122          | ntp service vNIC
  172.20.28.5 | NULL                                              | NULL         | ntp-aaef9e47-a704-4fcd-b333-027bfd7e1ac9          | ntp service vNIC
  172.20.28.2 | nexus-6b0cf0b4-b5c6-4eda-a57f-7d110cb46324        | nexus        | nexus-6b0cf0b4-b5c6-4eda-a57f-7d110cb46324        | nexus service vNIC
  172.20.28.3 | nexus-a365f750-1804-4021-ada8-e0532c973c32        | nexus        | nexus-a365f750-1804-4021-ada8-e0532c973c32        | nexus service vNIC
  172.20.28.4 | nexus-a87d43d8-8a94-4a18-95ff-12206ee833d9        | nexus        | nexus-a87d43d8-8a94-4a18-95ff-12206ee833d9        | nexus service vNIC
  172.20.28.7 | nexus-5acb41a9-f76c-4b3c-8313-2a74ce37c32f        | nexus        | nexus-5acb41a9-f76c-4b3c-8313-2a74ce37c32f        | nexus service vNIC
(7 rows)
```

2. [vpc_resolve_to_sleds](https://github.com/oxidecomputer/omicron/blob/55a0760a77a37333c51594229e06c9b598349c90/nexus/db-queries/src/db/datastore/vpc.rs#L674-L682) joins the `service_network_interface` table against `service` to map service vNICs to sleds. I'm not sure off the top of my head what to do about this one.

Before we stop populating the `service` table, we probably need to try deleting it from `schema.rs` and seeing if there are other consumers lurking around.

jgallagher:
Closing this in favor of #5287

PR #5045, opened 2024-02-12
Reconfigurator: Record external networking allocations when realizing a blueprint

### LINKED ISSUE #4886: add sled: planner support for deploying Nexus

jgallagher:
This PR expands blueprint execution to record external IPs and created NICs for Nexus, Boundary NTP, and External DNS in crdb _before_ sending new zone requests to sled-agents.

The implementation has a very obvious TOCTOU race, but I think it's okay (as explained in the comments inline). If two Nexuses try to realize the same blueprint simultaneously and both see no records present, only one will succeed to insert, and the other will spuriously fail. Assuming that failure causes a retry, subsequent attempts to realize the same blueprint will succeed, as the required records will be present. If this seems wrong, please holler!

I'd like to give this a spin on either madrid or one of the software testbeds before merging, but I think this is ready for review.


andrewjstone (approved):
Looks great!

nexus/blueprint-execution/src/resource_allocation.rs
```diff
@@ -0,0 +1,998 @@
...
+            // to have a single IP. This service already has (at least) one:
+            // make sure this list includes the one we want, or return an error.
+            for allocated_ip in &allocated_ips {
+                if allocated_ip.ip.ip() == external_ip
+                    && port_range
+                        .map(|(first, last)| {
+                            allocated_ip.first_port == SqlU16(first)
+                                && allocated_ip.last_port == SqlU16(last)
+                        })
+                        .unwrap_or(true)
```
sunshowers, line 120: What does `port_range` being None mean here?

nexus/blueprint-execution/src/resource_allocation.rs
```diff
@@ -0,0 +1,998 @@
...
+                description,
+                service_id,
+                external_ip,
+            )
+            .await
+            .with_context(|| {
+                format!(
+                    "failed to allocate IP to {zone_type} {service_id}: \
+                         {external_ip}"
+                )
```
sunshowers, line 267: Can we add a note here saying that this is expected in case of a race? (Or maybe this is already done at a higher level.)

I'm thinking about future debugging sessions where a simple error in the logs can be confusing.

nexus/blueprint-execution/src/resource_allocation.rs
```diff
@@ -0,0 +1,998 @@
...
+                snat.ip,
+                (snat.first_port, snat.last_port),
+            )
+            .await
+            .with_context(|| {
+                format!(
+                    "failed to allocate snat IP to {zone_type} {service_id}: \
+                     {snat:?}"
+                )
+            })?;
```
sunshowers, line 322: Same comment applies here I guess, and maybe it does suggest that there is (or should be) a higher-level note.

nexus/blueprint-execution/src/resource_allocation.rs
```diff
@@ -0,0 +1,998 @@
...
+
+    // Nexus and ExternalDns both use non-SNAT service IPs; this method is used
+    // to allocate external networking for both of them.
+    async fn ensure_external_service_ip(
+        &self,
+        zone_type: &'static str,
+        service_id: Uuid,
+        external_ip: IpAddr,
+        ip_name: &Name,
+    ) -> anyhow::Result<()> {
```
sunshowers, line 227: Thoughts on `use anyhow::Result`?

nexus/blueprint-execution/src/resource_allocation.rs
jgallagher: There are two kinds of external IPs that have separate methods (`ensure_external_service_ip` and `ensure_external_service_snat_ip`) that both want to call this method. In the former case, all we have is an IP address, no port range, so when it calls this function it passes `None`. In the latter case, we have a `SourceNatConfig` value, which includes an IP and a port range, so when it calls this function it passes `Some(_)`.

In practice, the non-SNAT IPs are given a port range of `(0, 65535)`, but it seemed weird to hardcode that here when it's a decision being made elsewhere.

nexus/blueprint-execution/src/resource_allocation.rs
jgallagher: Hmm, I hope the error we get from the datastore would make it clear whether a failure is due to a duplication issue or some other transient db issue. I'm hesitant to label this as "expected if we're racing" when the underlying error might be some other kind of db problem.

nexus/blueprint-execution/src/resource_allocation.rs
jgallagher: No objection, I just have a light preference for this because when I see `Result<()>` I have to either remember or look up which typealias it is.

nexus/blueprint-execution/src/resource_allocation.rs
sunshowers: Hmm interesting, should we use a type that's isomorphic to Option but not actually that? Definitely was surprised.

sunshowers: (In particular, Ord doesn't make sense I think)

nexus/blueprint-execution/src/resource_allocation.rs
sunshowers: Okay, just trying to think it through.

nexus/blueprint-execution/src/resource_allocation.rs
jgallagher: Sorry, where is Ord coming into play?

sunshowers: Sorry, what I mean is that if we define our own type, we can choose to not implement `Ord` on it, since it doesn't make sense in this case. We could also define other operations like `subset_of`/`superset_of`, which is what I think this code is trying to do.

sunshowers: (But this is followup material, doesn't have to be done here.)

sunshowers (approved):

andrewjstone:
On testbed, I'm seeing this warning as soon as I generate a blueprint for g2, which certainly doesn't have a nexus zone yet:

```
00:25:24.307Z WARN c33a7560-3972-4cc0-8418-aa5db7cd5026 (ServerContext): zone has unexpected NICs allocated
    allocated_nics = [ServiceNetworkInterface { identity: ServiceNetworkInterfaceIdentity { id: 3fac0bfd-5bba-402c-940c-99b7f1bda592, name: Name(Name("nexus-5a581037-15b9-4196-9d0b-cf20a77e86b5")), description: "nexus service vNIC", time_created: 2024-02-13T23:28:19.951897Z, time_modified: 2024-02-13T23:28:19.951897Z, time_deleted: None }, service_id: 5a581037-15b9-4196-9d0b-cf20a77e86b5, vpc_id: 001de000-074c-4000-8000-000000000000, subnet_id: 001de000-c470-4000-8000-000000000002, mac: MacAddr(MacAddr(MacAddr6([168, 64, 37, 255, 194, 97]))), ip: V4(Ipv4Network { addr: 172.30.2.5, prefix: 32 }), slot: 1, primary: true }]
    background_task = blueprint_executor
    comment =
    file = nexus/blueprint-execution/src/resource_allocation.rs:195
    want_nic = NetworkInterface { id: 3fac0bfd-5bba-402c-940c-99b7f1bda592, ip: 172.30.2.5, kind: Service(5a581037-15b9-4196-9d0b-cf20a
```

andrewjstone:
I think the previous error is preventing blueprint execution. No `omicron-zones.json` is showing up and the ntp zone isn't starting. I'm going to test on main to see if I can get nexus deploying without this code.

bnaecker:
@andrewjstone is this possibly being caused by https://github.com/oxidecomputer/omicron/issues/5056?

andrewjstone:
> @andrewjstone is this possibly being caused by #5056?

Ah very possibly.

jgallagher:
Yeah, it definitely is, and I strongly suspect the NIC it's warning about is for one of the existing Nexus zones. This PR attempts to reconcile _all_ of the external IP allocations, which should be a no-op for anything set up during RSS, but is failing because of the mismatched slot number. Fixing #5056 is at the top of my list.

nexus/blueprint-execution/src/resource_allocation.rs
jgallagher: Hmm, we don't want subset/superset here; this is just a straight equality check (except for having to convert to `SqlU16` to actually do the comparison). Is there a clearer way to express that?

jgallagher:
@andrewjstone I've caught up with main which has fixed #5056, if you want to give this another try.

andrewjstone:
> @andrewjstone I've caught up with main which has fixed #5056, if you want to give this another try.

Will do!

andrewjstone:
> > @andrewjstone I've caught up with main which has fixed #5056, if you want to give this another try.
> 
> Will do!

I re-ran this and blueprints are executing successfully. Unfortunately I didn't remember to add an extra nexus node to the code, so I'm stuck at 3 nexuses for the original RSS nodes. However, the bug appears to be fixed regarding #5056, since the executor is successfully deploying zones.

jgallagher:
I ran this through a test on madrid, and it was almost successful. I was able to add a 4th Nexus, it did go to the one sled that didn't have one from RSS, the allocations for external IP and vNIC looked valid and were recorded in cockroach, and the background NAT RPW set up the NAT rules for it after the inventory collection picked it up. I ran into two problems (one extremely minor that may not even be worth fixing, and one serious):

1. `omdb db network list-eips` failed to show the new external IP, because it tries to map the service name from the `services` table. We're not creating an entry in `services` due to its planned removal (#4947). I'll see if I can fix up omdb here.
2. `sled-agent` didn't get OPTE firewall rules for the new service, so even though Nexus existed and had NAT plumbing, it wasn't accepting inbound connections.

I'm going to poke around 2 a bit and will open an issue soon.

jgallagher:
Haha :man_facepalming: the test failure on the latest commit is legit. I think "always request new firewall rules when getting zones" is probably okay _once Nexus is up_, but can't gate RSS. I need to think about this more.

nexus/blueprint-execution/src/resource_allocation.rs
sunshowers: Ahh, no I think it was just that I misunderstood what this was doing :)

PR #5052, opened 2024-02-13
[nexus-db-queries] remove uses of filter_target

sunshowers:
`filter_target` is required for partial indexes but is easy to misuse for full indexes. To work around that, define an `as_partial_index` method, switch to it, and ban general uses of `filter_target` via clippy.

`as_partial_index` adds a `WHERE false` clause. See the comments included in the PR for why that works.

The rest of the PR:

* Makes sure clippy's `disallowed_` lints are active. `-A clippy::style` disables them.
* Adds some missing tests (thanks @jmpesp)
* Fixes a `RunnableQuery` that often did not return anything, replacing it with `RunnableQueryNoReturn`.


nexus/db-queries/src/db/datastore/silo_group.rs
```diff
@@ -41,7 +41,12 @@ impl DataStore {
         Ok(diesel::insert_into(dsl::silo_group)
             .values(silo_group)
             .on_conflict((dsl::silo_id, dsl::external_id))
-            .filter_target(dsl::time_deleted.is_null())
+            // This used to have a `filter_target` condition on
+            // `dsl::time_deleted.is_null()`, but that's been removed since
+            // filter_target doesn't work with CockroachDB:
+            // https://github.com/oxidecomputer/omicron/issues/5047.
+            //
+            // XXX: is this correct?
```
sunshowers, line 49: @jmpesp I was trying to figure out was this was trying to do -- how much do we care about deleted rows here?

nexus/db-queries/src/db/datastore/snapshot.rs
```diff
@@ -113,26 +119,29 @@ impl DataStore {
...
 
-                    Project::insert_resource(
-                        project_id,
+                    let query = {
+                        use diesel::query_dsl::methods::FilterDsl;
+
                         diesel::insert_into(dsl::snapshot)
                             .values(snapshot)
                             .on_conflict((dsl::project_id, dsl::name))
-                            .filter_target(dsl::time_deleted.is_null())
```
sunshowers, line 121: I think you wrote this as well @jmpesp -- is the replacement correct?

nexus/db-queries/src/db/datastore/silo_group.rs
sunshowers: I'm also wondering whether returning the silo is the right thing to do -- with `ON CONFLICT DO NOTHING`, no rows are returned if the row already exists, and any code that attempts to get an item will fail.

jmpesp: This filter_target allows us to delete a silo group, then re-create another one later (through JIT user and group creation) using the same group string. I don't think it's correct to remove this check, but it also looks like the original `filter_target` was being ignored :(

nexus/db-queries/src/db/datastore/snapshot.rs
jmpesp: I don't think so - `test_create_snapshot_record_idempotent` fails for me if I run this PR locally:

```
thread 'integration_tests::snapshots::test_create_snapshot_record_idempotent' panicked at nexus/tests/integration_tests/snapshots.rs:946:10:
called `Result::unwrap()` on an `Err` value: InternalError { internal_message: "unexpected database error: there is no unique or exclusion constraint matching the ON CONFLICT specification" }
```

nexus/db-queries/src/db/datastore/silo_group.rs
jmpesp: > I'm also wondering whether returning the silo is the right thing to do -- with `ON CONFLICT DO NOTHING`, no rows are returned if the row already exists, and any code that attempts to get an item will fail.

Yeah, this is probably not the right thing to do then.

sunshowers:
Thanks, looking through the comments now.

clippy.toml
```diff
@@ -0,0 +1,9 @@
+disallowed-methods = [
+    # `ON CONFLICT ... WHERE ... DO UPDATE` does not work with CockroachDB --
+    # see https://github.com/oxidecomputer/omicron/issues/5047.
+    #
+    # Instead, import diesel::query_dsl::methods::FilterDsl, then use
+    # `do_update().set(...).filter(...)`. (This is not part of the
+    # documentation as of Diesel 2.1.4, but it works.)
+    "diesel::upsert::DecoratableTarget::filter_target",
```
hawkw, line 8: I wonder if it's possible to write a simple test which would break if, when updating to a new Diesel version, `FilterDsl`  changed to no longer generate SQL that works correctly with crdb? Is that worth the effort?

nexus/db-queries/src/db/datastore/snapshot.rs
```diff
@@ -93,6 +93,11 @@ impl DataStore {
                     // If this function is passed a snapshot with an ID that
                     // does not match, but a project and name that does, return
                     // ObjectAlreadyExists here.
+                    //
+                    // NOTE: an earlier version of this comment suggested using
+                    // filter_target for time_deleted, which does not work with
+                    // CockroachDB:
+                    // https://github.com/oxidecomputer/omicron/issues/5047
```
hawkw, line 100: is this really necessary? i feel like this information is kind of captured in the Git history already...

nexus/db-queries/src/db/datastore/silo_group.rs
sunshowers: Digging into this more -- it turns out that this is actually required if the constraint you're matching against has a WHERE clause. In this case, the WHERE clause is https://github.com/oxidecomputer/omicron/blob/6c9c9b73d82c91389f0c8ccb46e33aaaf0afdf03/schema/crdb/dbinit.sql#L706-L707.

Also, you can truly pass in _anything_ which implies the WHERE clause in the constraint, according to Aristotelian (classical) logic. For example, you can pass in `WHERE false = true` or `WHERE 1 = 2`, since in Aristotelian logic [a contradiction implies any proposition](https://en.wikipedia.org/wiki/Principle_of_explosion).

nexus/db-queries/src/db/datastore/snapshot.rs
sunshowers: removed! (Honestly I was trying to wrap my head around this)

clippy.toml
sunshowers: I think existing tests would catch this, right?

sunshowers:
As discussed today, we do actually need `filter_target` here because this is a partial index. Replace our uses of it with this thing that's not _not_ a hack, but I think is significantly harder to misuse.

I've verified via CockroachDB `EXPLAIN` that this compiles down to the same query.

I'm happy to write extensive comments explaining this, but would love to get a thumbs up on the approach.

nexus/db-queries/src/db/on_conflict_ext.rs
```diff
@@ -0,0 +1,36 @@
...
+use diesel::expression::SqlLiteral;
+use diesel::prelude::*;
+use diesel::sql_types::Bool;
+use diesel::upsert::IncompleteOnConflict;
+
+/// An extension trait for incomplete [`ON CONFLICT`] expressions.
+pub trait IncompleteOnConflictExt {
+    type AsPartialIndexOutput;
+
+    // TODO: explain what this does.
```
sunshowers, line 15: This is where I'd put in a pretty long comment.

nexus/db-queries/src/db/datastore/silo_group.rs
sunshowers: Replaced with `as_partial_index`.

sunshowers:
I spent the evening thinking about this.

For reference, the approaches we've discussed so far are:

1. Require that users of `filter_target` use `allow(clippy::disallowed_methods)`.
2. Largely abandon this PR, and trust that (now that we've flagged it in such memorable fashion) we'll catch it during review.
3. ~~Use `ON CONFLICT ON CONSTRAINT`~~ (this does not work for partial indexes)
4. Use this `WHERE false` hack.

1, 2 and 3 were discussed in the meeting today. 4 came later.

3 would be amazing if it worked, but it sadly does not. Out of the valid approaches discussed in the meeting today (1 and 2), there seemed to be general consensus about 1.

---

The best argument I can come up with for this is that it is the best way we know of to achieve the thing what we really want -- which is to just be able to name the columns that are constrained, and have the database figure out the specific index constraint.

One argument against it is that basically that we're relying on a 2000-year old bug in classical logic, as Eliza put it. I verified that crdb 23.2.0 has the same behavior, but it is possible that Cockroach errors on this in the future. Though in that case:

* It's not a hugely difficult revert.
* There will hopefully be a better way to express this as part of that effort. (Why can't `ON CONFLICT ON CONSTRAINT` just work...)

nexus/tests/integration_tests/snapshots.rs
```diff
@@ -1057,6 +1055,36 @@ async fn test_create_snapshot_record_idempotent(
         )
         .await
         .unwrap();
+
+    // Test that a new snapshot can be added with the same name as a deleted
```
smklein, line 1059: Thanks for adding this test!

nexus/db-queries/src/db/on_conflict_ext.rs
smklein: I'm interested. Happy to see that your issue to CRDB is going to get us more docs from them too! https://github.com/cockroachdb/docs/issues/18292 .

nexus/db-queries/src/db/on_conflict_ext.rs
```diff
@@ -0,0 +1,36 @@
...
+where
+    Target: DecoratableTarget<SqlLiteral<Bool>>,
+{
+    type AsPartialIndexOutput =
+        IncompleteOnConflict<Stmt, DecoratedAsPartialIndex<Target>>;
+
+    fn as_partial_index(self) -> Self::AsPartialIndexOutput {
+        #[allow(clippy::disallowed_methods)]
+        {
+            self.filter_target(sql::<Bool>("false"))
```
smklein, line 30: So I do understand (somewhat) based on the discussion yesterday that "contradiction implies YOLO", so `false` **implies** all partial indexes.

Is there ever a valid reason to select a subset of partial indices? Would this be desirable where someone wants conflict on one index, but not another?

smklein (approved):

nexus/db-queries/src/db/on_conflict_ext.rs
sunshowers: > So I do understand (somewhat) based on the discussion yesterday that "contradiction implies YOLO", so false implies all partial indexes.

Specifically, `false` implies all the _constraints_ on the partial indexes for the given columns. For example, in this case we have a partial index:

```sql
CREATE UNIQUE INDEX IF NOT EXISTS lookup_silo_user_by_silo ON omicron.public.silo_user (
    silo_id,
    external_id
) WHERE
    time_deleted IS NULL;
```

If you specify `silo_id` and `external_id`, and pass in `WHERE false`, the database will be able to infer that you mean the constraint `time_deleted IS NULL`. If you specify a set of columns that doesn't have any kind of unique constraint on it, the query will error out.

> Is there ever a valid reason to select a subset of partial indices? Would this be desirable where someone wants conflict on one index, but not another?

From my testing against crdb, the set of columns has to match one or more `UNIQUE INDEX`es exactly. If there are multiple `UNIQUE INDEX` instances on a set of columns, then `WHERE false` will match all of them.

I think there are theoretical cases where it's possible that being able to specify some narrower constraint is useful, but I can't imagine we'll be needing it in omicron any time soon.

sunshowers:
<details>

<summary>Rendered documentation</summary>

Screenshot 2024-02-14 at 18-25-53 IncompleteOnConflictExt in nexus_db_queries db - Rust

</details>

nexus/db-queries/src/db/on_conflict_ext.rs
sunshowers: Done -- see the latest version of the PR, and the screenshot of a rendered version of it.

Issue #5053, opened 2024-02-13
Any services failing to load -> Firewall rules don't come up


smklein:
I believe this is the root cause of https://github.com/oxidecomputer/colo/issues/63

# Scenario

- You have a sled which is cold booting, which has a set of services. Some of these services need firewall rules (e.g., External DNS, Nexus) to operate correctly.
- When the sled is cold booting, not all services come back online. This can happen, for example, when a U.2 has been removed from a service that needs it. (You'd be stuck in a retry loop [here](https://github.com/oxidecomputer/omicron/blob/5e9e59c312cd787ba50cf6776f220951917dfa14/sled-agent/src/sled_agent.rs#L602-L620)).
- This prevents the sled from calling `notify_nexus_about_self`, a function which [also plumbs firewall rules back to the sled](https://github.com/oxidecomputer/omicron/blob/5e9e59c312cd787ba50cf6776f220951917dfa14/sled-agent/src/sled_agent.rs#L622-L625).
- Therefore: the services on this sled which need firewall rules don't sufficiently come up. For Nexus, this can prevent the rack from appearing available.

## Why is this complicated

There are a couple subtleties to draw out here.

- "Why don't we `notify_nexus_about_self` concurrently with loading services" -> We can, and actually *do* this (we also happen to call `notify_nexus_about_self` if any hardware info has changed, on a separate code-path). This actually triggers Nexus to send firewall rules before services are necessarily initialized.
- "Why don't we just retry calls to `notify_nexus_about_self` until they work?" -> Because they'd succeed "too early". This call causes Nexus to look up firewall rules and send them back to the sled, which eventually bottoms out [here](https://github.com/oxidecomputer/omicron/blob/5e9e59c312cd787ba50cf6776f220951917dfa14/illumos-utils/src/opte/port_manager.rs#L517-L556). Note: this function **only acts on OPTE ports that already exist**! So if this call was performed **before** we created the Nexus zone, nexus-specific firewall rules would just be ignored.

So: We can't "notify Nexus + get firewall rules" concurrently with loading services, because it might trigger to early, and we can't really rely on a "retry method" with this call, because it won't fail if we called it too early.

## Proposals

1. **Nexus Internal Endpoint to Ask for FW Rules, Triggered After Individual Zones Launch**. Rather than waiting for "all zones to launch", the Sled Agent could just re-apply firewall rules after **any** zones that need FW rules come up.
[...]

PR #5054, opened 2024-02-13
[nexus][sled-agent] Try more to load firewall rules during service bringup

smklein:
Fixes https://github.com/oxidecomputer/omicron/issues/5053 , with this solution implemented:

> Nexus Internal Endpoint to Ask for FW Rules, Triggered After Individual Zones Launch. Rather than waiting for "all zones to launch", the Sled Agent could just re-apply firewall rules after any zones that need FW rules come up.

I used a bit of a brute-force solution here -- "just do it each time we try to launch a set of zones". This could certainly be more fine-grained, but it gets the job done.

### DIFF
--- a/sled-agent/src/sled_agent.rs
+++ b/sled-agent/src/sled_agent.rs
@@ -105,6 +105,9 @@ pub enum Error {
     #[error("Failed to operate on underlay device: {0}")]
     Underlay(#[from] underlay::Error),
 
+    #[error("Failed to request firewall rules")]
+    FirewallRequest(#[source] nexus_client::Error<nexus_client::types::Error>),
+
     #[error(transparent)]
     Services(#[from] crate::services::Error),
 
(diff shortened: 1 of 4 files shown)


sled-agent/src/sled_agent.rs
```diff
@@ -602,11 +605,24 @@ impl SledAgent {
...
+                // partially load some services.
+                let firewall_result = self
+                    .request_firewall_update()
                     .await
-                    .map_err(|err| BackoffError::transient(err))
+                    .map_err(|err| BackoffError::transient(err));
+
+                // Only complete if we have loaded all services and firewall
+                // rules successfully.
+                load_services_result.and(firewall_result)
```
smklein, line 625: By doing this operations back-to-back, we basically "load as many firewall rules as we can" for a set of services.

I could make this trigger more individually for "services which need firewall rules" (e.g., the request is only made after **adding** a Nexus / External DNS / boundary NTP zone), but felt that this was more "obviously not missing anything".

I could push down the triggers into the `ServiceManager` if folks felt strongly about this.

luqmana: Hmm, like this we'll continue to hit the endpoint if we get stuck in the disk error loop right? Wonder if cause any issues constantly trying to update fw rules. Oh but there's the backoff in the retry loop. So, should be fine

luqmana (approved):
Thanks for putting this together so quick Sean!

FelixMcFelix (approved):
Thanks for the very quick turnaround! This looks about right to me from what we talked over.

PR #5061, opened 2024-02-14
[docs] Guide to creating/editing SMF services in omicron

karencfv:
This is probably incomplete, we can add more stuff down the track. 

Just wanted to add some information while it's fresh in my mind

### DIFF
--- a/docs/adding-or-modifying-smf-services.adoc
+++ b/docs/adding-or-modifying-smf-services.adoc
@@ -66,7 +66,7 @@ output.intermediate_only = true
 
 == Sled agent services
 
-To make sure all of the service properties are populated, you'll have to edit the `sled-agent/src/services.rs` file. This is where all of the omicron zones are built.
+Service properties are values which can only be known at run-time. To make sure all of them are populated, you'll have to edit the `sled-agent/src/services.rs` file. This is where all of the omicron zones are built.
 
 If your service is new, it may look something like the following code. As you can see, the properties we are populating are the same as the ones on the service's manifest.xml file.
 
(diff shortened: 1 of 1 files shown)


docs/adding-or-modifying-smf-services.adoc
```diff
@@ -0,0 +1,112 @@
...
+    { from = "smf/oximeter", to = "/var/svc/manifest/site/oximeter" },
+    { from = "oximeter/db/schema", to = "/opt/oxide/oximeter/schema" },
+]
+output.type = "zone"
+output.intermediate_only = true
+```
+
+== Sled agent services
+
+To make sure all of the service properties are populated, you'll have to edit the `sled-agent/src/services.rs` file. This is where all of the omicron zones are built.
```
smklein, line 69: Might be worth noting -- these are valuable for properties that "can only be known at run-time"

docs/adding-or-modifying-smf-services.adoc
```diff
@@ -0,0 +1,112 @@
...
+
+= Adding or modifying SMF services
+
+== SMF services
+
+Go to the `smf/` directory, there you will find all current services or you create a new directory if you're working on a new service.
+
+Each directory can contain a myriad of files depending on the service, but only a `manifest.xml` file is necessary. This file contains the complete set of properties associated with the service instance.
+
+== Packaging
```
smklein, line 13: Thanks for documenting these short-comings. These are default decisions that were made ("rust package" comes from "package name", and "service_name" means "zone name" for zones) but this also serves as good data points for "how the manifest could be made more clear + explicit in the future".

smklein (approved):
Thanks for adding this!

karencfv:
Thanks for the review @smklein :bowing_woman:

PR #5063, opened 2024-02-14
[db-queries] Decouples CTE usage from Diesel

smklein:


nexus/db-queries/src/db/raw_query_builder.rs
```diff
@@ -0,0 +1,151 @@
...
+}
+
+impl TrustedStr {
+    /// Explicitly constructs a string, with a name that hopefully
+    /// gives callers some pause when calling this API.
+    ///
+    /// If arbitrary user input is provided here, this string COULD
+    /// cause SQL injection attacks, so each call-site should have a
+    /// justification for "why it's safe".
+    pub fn this_string_wont_cause_sql_injections(s: String) -> Self {
```
david-crespo, line 45: “this string could” reads like more of a warning that “this string won’t”, which could be misinterpreted as a reassurance

smklein: Updated to `i_take_responsibility_for_validating_this_string` - I'm okay wordsmithing this, since IMO it should be long, obvious, and hard to accidentally call.

Also wrapped this in a struct so a caller can't create the enum. I always forget that all enum variants are always accessible for construction.

smklein:
Weirdly this PR seems to be causing failures on the integration tests, but not on the more "targeted to the query" tests in `nexus-db-queries`. I need to investigate further to see what's happening.

As a side-note: It's been kinda useful to use `cockroachdb sqlfmt` on the output of `diesel::debug_query`, emitted via expectorate. I haven't wired up the auto-formatting yet, but that seems useful to easily compare diffs of complicated queries.

smklein:
I think I found the underlying issue breaking the tests. https://docs.rs/diesel/latest/diesel/dsl/fn.sql_query.html relies on Diesel's `QueryableByName`, which "gets the results wrong" when parsing the output of the query. AFAICT, the query itself is acting correctly, we're just parsing the output badly.

Basically: It returns a `(SELECT dataset) UNION (SELECT region)`. If both `dataset` and `region` contain a field named `id` (they do), then we pick the "first column named id", which leads to obvious problems (e.g., we never read the `region`'s `id` correctly).

I really wish there was a way to have columns queried by Diesel returned to a struct in the order they were parsed. I don't know why `sql_query` is imposing this constraint on us.

nexus/db-queries/src/db/raw_query_builder.rs
```diff
@@ -0,0 +1,195 @@
...
+pub struct TrustedStr(TrustedStrVariants);
+
+impl TrustedStr {
+    /// Explicitly constructs a string, with a name that hopefully
+    /// gives callers some pause when calling this API.
+    ///
+    /// If arbitrary user input is provided here, this string COULD
+    /// cause SQL injection attacks, so each call-site should have a
+    /// justification for "why it's safe".
+    pub fn i_take_responsibility_for_validating_this_string(s: String) -> Self {
```
hawkw, line 46: this is wonderful :)

smklein:
image

> This means that you cannot deserialize this query into a tuple, and any structs used must implement QueryableByName.

so_that_was_a_lie.jpeg

Turns out you can use the newtype pattern, re-implement some Diesel traits (https://docs.diesel.rs/master/diesel/prelude/trait.RunQueryDsl.html to indicate that it's executable, https://docs.diesel.rs/master/diesel/query_builder/trait.Query.html , with specific SQL type to say "what the output should be deserialized from", and https://docs.diesel.rs/master/diesel/query_builder/trait.QueryFragment.html to plumb through the underlying `SqlQuery` so we can execute the raw SQL we constructed)

This seems to work. I'm using a new type named `TypedSqlQuery` that should be generally available for this purpose, of exposing a "strongly typed SQL fragment" on top of a "not-very-strongly typed construction".

nexus/db-queries/src/db/raw_query_builder.rs
smklein: I hope it's the right direction, and I'd be stoked to move things more and more into "const fn"s as we can.

I think the choices made by Diesel (e.g. https://docs.diesel.rs/master/diesel/query_builder/struct.SqlQuery.html#method.sql, which just takes anything that impls `AsRef<str>` ) makes this a little too easy to foot-gun via SQL injection when building queries that side-step the norms of Diesel's type system.

I know it's possible to subvert this with `&'static str` too, and I wish there was a way to more explicitly say "compile-time-constructed string", but that's as close a proxy as I think I can get right now.

jmpesp:
My two cents: I really hope I'm not causing us to move away from what Diesel gives us. The entire time I had where things wouldn't compile was due to me *missing a join* on a table that I later used in a WHERE filter: this *was* a legit error that diesel was flagging. Once that was cleared up, everything worked.

I much prefer working with the older CTE, though I do love the other things this PR adds.

> Generally, the disconnect between "what SQL do I want to write" and "what invocations will make Diesel happy" has been "not really worth it" in this space.

Even though I had a frustrating time trying to figure out what the problem was, I'd still argue that Diesel gives us more than just raw SQL. It's especially useful if we need to refactor or reshape things so we can share code, etc.

smklein:
> My two cents: I really hope I'm not causing us to move away from what Diesel gives us. The entire time I had where things wouldn't compile was due to me _missing a join_ on a table that I later used in a WHERE filter: this _was_ a legit error that diesel was flagging. Once that was cleared up, everything worked.

This is definitely not just you -- I can clarify the PR description? But I have heard from many people interfacing with CTEs and Diesel and struggling (@bnaecker , @jgallagher , @internet-diglett , and more) because the workflow is "figure out what SQL to write", and then "figure out how to convince Diesel to accept it". If translating SQL to Diesel bought us guarantees that we could not get any other way, I'd be more convinced, but for more complex queries like this, I think we can get sufficient protection -- and a flatter structure -- by using something much closer to raw SQL.

> Even though I had a frustrating time trying to figure out what the problem was, I'd still argue that Diesel gives us more than just raw SQL. It's especially useful if we need to refactor or reshape things so we can share code, etc.

To expand a little bit on the value-proposition: When I originally added the CTE utilities, I had the same hopes. I wanted to use Diesel's type-safety guarantees to help us ensure things like:

- Are you only selecting columns from tables that exist in your FROM clauses?
- Similarly, when selecting columns from subqueries, do *those* types and names match?
- Do the types of your columns used as bind parameters match the underlying types within the SQL query itself?
- Is the SQL being constructed "valid"?

I still think these are all valid protections, that we should ensure! I also think that the EXPLAIN-based tests, in addition to very simple integration tests, provide protection against all these things. Furthermore, they do so with fewer types, without any need for aliasing, without any need for JOIN-based macros, and immediately with support for arbitrary Postgres / CockroachDB support that may or may not be represented in Diesel.

nexus/db-queries/src/db/raw_query_builder.rs
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    /// Takes the final boxed query
+    pub fn query<T>(self) -> TypedSqlQuery<T> {
+        TypedSqlQuery { inner: self.query, _phantom: PhantomData }
+    }
+}
+
+/// Diesel's [diesel::query_builder::BoxedSqlQuery] has a few drawbacks that
+/// make wrapper more palatable:
```
jmpesp, line 162: "make *this* wrapper", maybe?

nexus/db-queries/src/db/queries/region_allocation.rs
```diff
@@ -77,611 +64,344 @@ pub fn from_diesel(e: DieselError) -> external::Error {
...
+    // - Data redundancy: If a drive fails it should only take one of the 3
+    //   regions with it
+    // - Risk of overallocation: We only check that each zpool as enough
+    //   room for one region, so we should not allocate more than one region
+    //   to it.
+    //
+    // Selecting two datasets on the same zpool will not initially be
+    // possible, as at the time of writing each zpool only has one dataset.
+    // Additionally, we intend to modify the allocation strategy to select
+    // from 3 distinct sleds, removing the possibility entirely. But, if we
```
jmpesp, line 247: This modification has been done already right?

jmpesp (approved):
Looks good to me!

nexus/db-queries/src/db/raw_query_builder.rs
smklein: Updated

nexus/db-queries/src/db/queries/region_allocation.rs
smklein: I'll update the wording -- it has been done, but as a configuration option, so it's not *always* turned on

Issue #5056, opened 2024-02-13
Inserting a `NetworkInterface` should allow the caller to specify the slot


jgallagher:
When RSS is creating `NetworkInterface`s to send to sled-agent, it chooses a slot number (and always chooses `slot: 0`; e.g., [here for Nexus instances](https://github.com/oxidecomputer/omicron/blob/6c9c9b73d82c91389f0c8ccb46e33aaaf0afdf03/sled-agent/src/rack_setup/plan/service.rs#L1018)). After Nexus receives the handoff from RSS upon completion RSS copmletion, Nexus [converts the RSS-created `NetworkInterface`s to `IncompleteNetworkInterface`s](https://github.com/oxidecomputer/omicron/blob/8b4d2e9338a8c98290ca2845cf23f82ceee3f6eb/nexus/db-queries/src/db/datastore/rack.rs#L477-L487). `IncompleteNetworkInterface` allows the caller to specify the NIC ID, MAC, and IP, all of which we do; however, it does not allow specifying the `slot`. The slot is chosen by a `NextItem` query helper, which may choose any slot from 0-7; see #5055.

This leads to some inconsistency that may be a little painful as we work on the reconfiguration system. The inventory system collects `NetworkInterface`s as a part of collecting each sled's service zone configuration; these are the values chosen by RSS. On dogfood, we see that each has a `slot` of 0, as expected:

```
root@[fd00:1122:3344:109::3]:32221/omicron  OPEN> select * from inv_omicron_zone_nic where inv_collection_id = '373b8b21-d878-4da0-8eb7-6f45fb04f49c';
           inv_collection_id           |                  id                  |                       name                        |     ip     |       mac       |    subnet     | vni | is_primary | slot
---------------------------------------+--------------------------------------+---------------------------------------------------+------------+-----------------+---------------+-----+------------+-------
  373b8b21-d878-4da0-8eb7-6f45fb04f49c | 2e9a412e-c79a-48fe-8fa4-f5a6afed1040 | nexus-2898657e-4141-4c05-851b-147bffc6bbbd        | 172.30.2.7 | 184993468892761 | 172.30.2.0/24 | 100 |    true    |    0
[...]

Issue #5055, opened 2024-02-13
`NextItem` query may not return available items in series order


gjcolombo:
Nexus's [`NextItem` query fragment](https://github.com/oxidecomputer/omicron/blob/6c9c9b73d82c91389f0c8ccb46e33aaaf0afdf03/nexus/db-queries/src/db/queries/next_item.rs#L165) allows callers to select an "available" item from a pool of items that may have been previously assigned to some other entity. For example, disk attachment uses `NextItem` to select an available PCI slot to which to attach a new disk.

Perhaps unintuitively, `NextItem` makes no guarantees about the order in which it returns eligible items from its pool. For example, here's a mock `NextItem` query returning the available disk slots for an instance with no disks currently attached:

```
root@[fd00:1122:3344:107::3]:32221/omicron> SELECT slot FROM
(SELECT 0 + shift AS slot FROM
(SELECT generate_series(0, 8) AS shift UNION ALL SELECT generate_series(0, -1) AS shift)
LEFT OUTER JOIN disk
ON (attach_instance_id, slot, time_deleted IS NULL) = ('97afec53-7d1c-41ef-a583-47521c6ed0c8', 0 + shift, TRUE)
WHERE slot IS NULL);
  slot
--------
     1
     2
     4
     5
     6
     7
     8
     0
     3
(9 rows)
```

This is weird if you (like me) read `NextItem` to mean "the *next* available item in the set, if they are considered in order" and not just "any available item."

We should give `NextItem` these semantics (i.e. "return the lowest-numbered available item") by sticking an `ORDER BY` on the outermost SELECT before returning the first row.

Note that this is (as far as I know) only a hygiene issue so far--for VMs, disk and NIC attachment work just fine when you plug a device into slot 1 but not slot 0, and this doesn't cause a single slot to be used for more than one disk/NIC on a given instance. But changing this behavior (or the name of `NextItem`, I suppose) will make it harder to write code in the future that incorrectly assumes that the query returns items in order.

PR #5065, opened 2024-02-14
`IncompleteNetworkInterface`: allow user-specified explicit `slot`

jgallagher:
This makes a several minor changes to plumb slots through:

* `IncompleteNetworkInterface` now stores an optional slot, just like it stores an optional IP/MAC address
* In `network_interfaces::InsertQuery`, if the incoming slot is set, we use it directly instead of running the `NextItem`-based subquery
* Adds a partial index to ensure uniqueness of a slot within a single `parent_id` (I believe this is correct, but would love confirmation from someone more familiar!)
* `IncompleteNetworkInterface::new_service()` now takes a _non-optional_ IP, MAC address, and slot. This matches how it was called in all non-test code.
* Tweaked the Nexus internal API used for RSS handoff to include the slot in the description of NICs.

This is a partial fix for #5056, and should produce correct behavior on new systems that run through RSS, even without a fix for #5055 (because we bypass `NextItem` altogether with this change). In particular, I think this should unblock testing of #5045 on madrid / testbeds. It does not address the already-recorded-NICs-with-incorrect-slots on systems like dogfood; I'll take care of that in a subsequent PR.


nexus/db-model/src/network_interface.rs
```diff
@@ -17,6 +17,13 @@ use nexus_types::identity::Resource;
 use omicron_common::api::external;
 use uuid::Uuid;
 
+/// The max number of interfaces that may be associated with a resource,
+/// e.g., instance or service.
+///
+/// RFD 135 caps instances at 8 interfaces and we use the same limit for
+/// all types of interfaces for simplicity.
+pub const MAX_NICS: usize = 8;
```
bnaecker, line 25: There exists the same constant in Nexus itself, `crate::nexus::MAX_NICS_PER_INSTANCE`. I think this was moved, not new, but just checking we want to duplicate the const rather than re-export one.

nexus/db-model/src/network_interface.rs
```diff
@@ -221,24 +230,35 @@ impl IncompleteNetworkInterface {
...
             }
-            (Some(mac), NetworkInterfaceKind::Service) if !mac.is_system() => {
-                return Err(external::Error::invalid_request(&format!(
-                    "invalid MAC address {} for service NIC",
-                    mac
+        }
+        if let Some(slot) = slot {
+            if usize::from(slot) >= MAX_NICS {
+                return Err(external::Error::invalid_request(format!(
+                    "invalid slot {slot} for NIC",
```
bnaecker, line 259: Possibly include the maximum here?

nexus/db-queries/src/db/queries/network_interface.rs
```diff
@@ -984,6 +996,7 @@ pub struct InsertQuery {
     parent_id_str: String,
     ip_sql: Option<IpNetwork>,
     mac_sql: Option<db::model::MacAddr>,
+    slot_sql: Option<i16>,
```
bnaecker, line 999: Nit, but we we may want to use the `SqlU8` type here.

nexus/db-queries/src/db/queries/network_interface.rs
```diff
@@ -386,6 +392,12 @@ fn decode_database_error(
                 let mac = interface.mac.unwrap_or_else(|| MacAddr::from_i64(0));
                 InsertError::MacAddressNotAvailable(mac)
             }
+            // Constraint violated if a user-requested slot has
+            // already been assigned within the same instance or service.
+            Some(constraint) if constraint == SLOT_NOT_AVAILABLE_CONSTRAINT => {
+                let slot = interface.slot.unwrap_or(0);
```
bnaecker, line 398: Do we care about the case where the slot is `None` here? That shouldn't be possible, since we'll only get this error if the user requested one, meaning slot is `Some(_)`.

schema/crdb/34.0.0/up.sql
```diff
@@ -0,0 +1,6 @@
+CREATE UNIQUE INDEX IF NOT EXISTS network_interface_parent_id_slot_key ON omicron.public.network_interface (
```
bnaecker, line 1: This is correct. We effectively had a uniqueness guarantee from the `NextItem` subquery, but this is the right way to enforce that when the user can also provide one explicitly.

bnaecker (approved):
Looks good, the tests are excellent!

nexus/db-queries/src/db/queries/network_interface.rs
jgallagher: Agreed, and I think we came to the same conclusion with `MAC_NOT_AVAILABLE_CONSTRAINT` (which is immediately above this new case). We could be more pedantic and store the option instead of having to `unwrap_or()`, but I'm not sure it's really useful.

nexus/db-model/src/network_interface.rs
jgallagher: By usage they definitely have to be the same value, so it doesn't seem good to have two constants. I'll rename this one `MAX_NICS_PER_INSTANCE`, and change nexus to re-export it instead of defining its own. Nice catch.

PR #5067, opened 2024-02-14
Order `NextItem` subquery results predictably

### LINKED ISSUE #5055: `NextItem` query may not return available items in series order

bnaecker:
- Fixes #5055
- Add a new column, `index` to the `NextItem` subquery, which indexes the shifts from 0..n.
- Add an `ORDER BY index` clause to guarantee order.
- Add test

### DIFF
--- a/nexus/db-queries/src/db/queries/next_item.rs
+++ b/nexus/db-queries/src/db/queries/next_item.rs
@@ -35,7 +35,7 @@ use uuid::Uuid;
 /// SELECT
 ///     <first_address> + shift AS ip
 /// FROM
-///     generate_series(0, <max_shift>) AS shift
+///     generate_series(0, <max_shift>) AS shift,
 /// LEFT OUTER JOIN
 ///     network_interface
 /// ON
@@ -43,21 +43,22 @@ use uuid::Uuid;
 ///     (<subnet_id>, <first_address> + shift, TRUE)
 /// WHERE
 ///     ip IS NULL
+/// ORDER BY ip
 /// LIMIT 1
 /// ```
 ///
-/// This query selects the lowest address in the IP subnet that's not already
-/// allocated to a guest interface. Note that the query is linear in the number
-/// of _allocated_ guest addresses. <first_address> and <max_shift> are chosen
-/// based on the subnet and its size, and take into account reserved IP
-/// addresses (such as the broadcast address).
+/// This query selects the next address after <first_address> in the IP subnet
+/// that's not already allocated to a guest interface. Note that the query is
+/// linear in the number of _allocated_ guest addresses. <first_address> and
+/// <max_shift> are chosen based on the subnet and its size, and take into
+/// account reserved IP addresses (such as the broadcast address).
 ///
 /// General query structure
 /// -----------------------
 ///
 /// Much of the value of this type comes from the ability to specify the
 /// starting point for a scan for the next item. In the case above, of an IP
-/// address for a guest NIC, we always try to allocate the lowest available
+/// address for a guest NIC, we always try to allocate the next available
 /// address. This implies the search is linear in the number of allocated IP
 /// addresses, but that runtime cost is acceptable for a few reasons. First, the
 /// predictability of the addresses is nice. Second, the subnets can generally
@@ -81,16 +82,16 @@ use uuid::Uuid;
 /// FROM
 ///     (
 ///         SELECT
-///             shift
+///             "index", shift
 ///         FROM
-///             generate_series(0, <max_shift>)
-///         AS shift
+///             generate_series(0, <n>) AS "index"
+///             generate_series(0, <max_shift>) AS shift,
 ///         UNION ALL
 ///         SELECT
 ///             shift
 ///         FROM
-///             generate_series(<min_shift>, -1)
-///         AS shift
+///             generate_series(<n + 1>, <range_size>) AS "index"
+///             generate_series(<min_shift>, -1) AS shift,
 /// LEFT OUTER JOIN
 ///     <table>
 /// ON
@@ -98,6 +99,7 @@ use uuid::Uuid;
 ///     (<scope_key>, <base> + shift, TRUE)
 /// WHERE
 ///     <item_column> IS NULL
+/// ORDER BY "index"
 /// LIMIT 1
 /// ```
 ///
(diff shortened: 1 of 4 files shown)


bnaecker:
All the test failures uncovered another bug. Looking at [this log](https://buildomat.eng.oxide.computer/wg/0/details/01HPMN1MCKE75WCHT9967K119Q/PpuARdlVS8A7PU7zS62jumAMdN5yMco5nW9U7XfqoC6GxNRD/01HPMN2AA4B73CGF7NFK6F00R1#S7674), for example, we can see we're tripping a new check I added in this PR because I was paranoid. The `NextItem` query takes a "min shift" and a "max shift". Those are the shifts _to the minimum_ and _to the maximum_ of the searched range, respectively. So the max must really be >=0 and the minimum <= 0.

The MAC address query [here](https://github.com/oxidecomputer/omicron/blob/4fdedceec5f62525b0b907a9010695ba1cb62931/nexus/db-queries/src/db/queries/network_interface.rs#L600-L614) is passing a positive number for the min shift. We've been potentially searching kind of arbitrary ranges for a while now. Why I didn't ensure this in the type system to begin with, I can't say. I am going to look into doing that now, so that the generator requires passing in two `u32`s, for the maximum leftward and rightward shifts, relative to the base.

bnaecker:
Ok, I'm going to punt on changing the logic around the shifts in this particular PR. I have fixed the bug in the system MAC allocation I found. I do think the API for expressing the range of the `NextItem` query is difficult to use correctly, and I'd like to change it in a follow up PR. Doing so now will be more invasive than I'd like.

bnaecker:
@gjcolombo This is ready for a review now

nexus/db-queries/src/db/queries/next_item.rs
```diff
@@ -81,23 +82,24 @@ use uuid::Uuid;
...
-///             shift
+///             index, shift
 ///         FROM
-///             generate_series(0, <max_shift>)
-///         AS shift
+///             generate_series(0, <max_shift>) AS shift,
+///             generate_series(0, <n>) AS index
 ///         UNION ALL
 ///         SELECT
 ///             shift
```
gjcolombo, line 91: nit: `index, shift`?

nexus/db-queries/src/db/queries/next_item.rs
```diff
@@ -81,23 +82,24 @@ use uuid::Uuid;
 /// FROM
 ///     (
 ///         SELECT
-///             shift
+///             index, shift
 ///         FROM
-///             generate_series(0, <max_shift>)
-///         AS shift
+///             generate_series(0, <max_shift>) AS shift,
+///             generate_series(0, <n>) AS index
```
gjcolombo, line 88: turbo doc nit: if I play around with this query in a CockroachDB cluster I have to quote `"index"` in order to get the query to work; presumably Diesel is doing this for us, but it's nice to be able to muck around with the query by hand to understand how it works

gjcolombo (approved):
This makes sense to me. Thanks for fixing this up! Unfortunately the query that was reproducing random disk slot orderings for me the other day is no longer doing so, so I wasn't able to test this out directly, but the overall logic of it makes sense.

A couple of notes:

- I did manage to find the place in the CRDB docs where row ordering (or lack thereof) is specified, if you want to refer to it in here somewhere: https://www.cockroachlabs.com/docs/stable/select-clause#sorting-and-limiting-query-results ("rows are not sorted with any consistent criteria").
- I'm probably just being pedantic, but one thing I want to make sure to note: if/when we make a subsequent change to try to fix any unexpected slot number assignments that came from the old query, we **should not touch** the assigned slot numbers for existing attached disks--these need to remain stable because there may be guest data (PCI slot numbers stored on the EFI system partition) that refers to them.

nexus/db-queries/src/db/queries/next_item.rs
bnaecker: Done, here's what we get now, as seen by the DTrace query-start probe:

```
INSERT INTO "test_schema"."item" ("id", "value", "time_deleted") VALUES (gen_random_uuid(), (SELECT $1 + "shift" AS "value" FROM (SELECT generate_series(0, $2) AS "index", generate_series(0, $3) AS "shift" UNION ALL SELECT generate_series($4, $5) AS "index", generate_series($6, -1) AS "shift") LEFT OUTER JOIN "test_schema"."item" ON ("value", "time_deleted" IS NULL) = ($7 + "shift", TRUE)  WHERE "value" IS NULL ORDER BY "index" LIMIT 1), NULL) RETURNING "test_schema"."item"."id", "test_schema"."item"."value", "test_schema"."item"."time_deleted" -- binds: [5, 5, 5, 6, 10, -5, 5]
```

bnaecker:
I'm holding on this for a bit. I did some digging into questions on #5085 and am concerned we'll hit some of the assertions this PR adds if users add an instance with an external IPv6 address. Doing so used to fail pretty early with a 400, but that seems to have changed since I last worked on that part of the code. I'm looking now, but don't want to merge this if Nexus can panic due to a currently-valid external request.

bnaecker:
Ok, I've convinced myself that the concerns I had were not related to this PR specifically. Merging shortly.

PR #5071, opened 2024-02-15
Update progenitor to bc0bb4b

bnaecker:


ahl:
looks good; I'd love to see this either 1. happen automatically (@sunshowers ? @augustuswm ?) or 2. for us to switch to the crates.io version.

bnaecker:
Yeah, that'd be nice. We're already ahead of the crates.io version, so we could switch to that on its next release there. Or include this in the configurations for one of the bots.

augustuswm:
I can set up some bot configuration if the goal is to track against a specific branch (i.e. `main`) for progenitor if that is needed. Do we usually rely on progenitor features that are not included in the current crates.io release?

bnaecker:
I'm not specifically relying on new features, but the fact that this is a git dependency does often create trouble for me when making changes that span repositories. Propolis, Crucible, and Omicron are the usual culprit. It's also very possible I'm doing something silly when making those changes!

sunshowers:
oh crap, accidentally hit the disable auto merge button, sorry!

sunshowers:
Yeah I'd love it if we switched to the crates.io version. Renovate doesn't understand git dependencies at the moment—i was hoping to add that at some point but then other priorities happened.

bnaecker:
@sunshowers Ok, that sounds good. I think we can do that after the next version bump on crates.io, since we're ahead of it right now.

PR #5078, opened 2024-02-15
Add PEM encoded certificate to external api response

augustuswm:
* Add pem encoded certificate to certificate list and view endpoints

When managing multiple silos, it quickly becomes desirable to be able to use the oxide cli and/or sdk to manage them. Handling certificate rotation currently requires storing external state about what certificate is registered with which silo as our external API only returns identity metadata about the certificate. This PR proposes adding the the public certificate to the fields returned for a certificate.

This initial implementation tries to take the simplest approach to this:
1. Return the exact representation that we store in the database (PEM-encoded string)
2. Returns the certificate on both list and view endpoints

An alternative approach may be to only return the full certificate as a part of the view endpoint, but I'm not sure if that is a pattern that we use.

The other consideration is how this should handle the failure to construct a String from the bytes form of a certificate. Currently it is very disruptive, and hard to debug if bad data were to be stored in the database.

### DIFF
--- a/nexus/db-model/src/certificate.rs
+++ b/nexus/db-model/src/certificate.rs
@@ -82,6 +82,16 @@ impl TryFrom<Certificate> for views::Certificate {
         Ok(Self {
             identity: cert.identity(),
             service: cert.service.try_into()?,
+            cert: String::from_utf8(cert.cert)
+                .map_err(|err| {
+                    Error::InternalError {
+                        internal_message: format!(
+                            "Failed to construct string from stored certificate: {}",
+                            err
+                        )
+                    }
+                }
+            )?,
         })
     }
 }
--- a/nexus/db-model/src/certificate.rs
+++ b/nexus/db-model/src/certificate.rs
@@ -86,11 +86,8 @@ impl TryFrom<Certificate> for views::Certificate {
             // database with PEM encoding which are essentially bundles of Base64 encoded text.
             // The only cases in which this conversion should fail is when our internal database
             // representation of the certificate is invalid.
-            cert: String::from_utf8(cert.cert).map_err(|err| {
-                Error::internal_error(&format!(
-                    "Failed to construct string from stored certificate: {}",
-                    err
-                ))
+            cert: String::from_utf8(cert.cert).map_err(|_| {
+                Error::internal_error("Certificate is not valid UTF-8")
             })?,
         })
     }
--- a/nexus/db-model/src/certificate.rs
+++ b/nexus/db-model/src/certificate.rs
@@ -82,16 +82,16 @@ impl TryFrom<Certificate> for views::Certificate {
         Ok(Self {
             identity: cert.identity(),
             service: cert.service.try_into()?,
-            cert: String::from_utf8(cert.cert)
-                .map_err(|err| {
-                    Error::InternalError {
-                        internal_message: format!(
-                            "Failed to construct string from stored certificate: {}",
-                            err
-                        )
-                    }
-                }
-            )?,
+            // This is expected to succeed in normal circumstances. Certificates are stored in the
+            // database with PEM encoding which are essentially bundles of Base64 encoded text.
+            // The only cases in which this conversion should fail is when our internal database
+            // representation of the certificate is invalid.
+            cert: String::from_utf8(cert.cert).map_err(|err| {
+                Error::internal_error(&format!(
+                    "Failed to construct string from stored certificate: {}",
+                    err
+                ))
+            })?,
         })
     }
 }
--- a/nexus/types/src/external_api/views.rs
+++ b/nexus/types/src/external_api/views.rs
@@ -177,7 +177,10 @@ pub struct Project {
 pub struct Certificate {
     #[serde(flatten)]
     pub identity: IdentityMetadata,
+    /// The service using this certificate
     pub service: ServiceUsingCertificate,
+    /// PEM-formatted string containing public certificate chain
+    pub cert: String,
 }
 
 // IMAGES
--- a/openapi/nexus.json
+++ b/openapi/nexus.json
@@ -11021,6 +11021,10 @@
         "description": "View of a Certificate",
         "type": "object",
         "properties": {
+          "cert": {
+            "description": "PEM-formatted string containing public certificate chain",
+            "type": "string"
+          },
           "description": {
             "description": "human-readable free-form text about a resource",
             "type": "string"
@@ -11039,7 +11043,12 @@
             ]
           },
           "service": {
-            "$ref": "#/components/schemas/ServiceUsingCertificate"
+            "description": "The service using this certificate",
+            "allOf": [
+              {
+                "$ref": "#/components/schemas/ServiceUsingCertificate"
+              }
+            ]
           },
           "time_created": {
             "description": "timestamp when this resource was created",
@@ -11053,6 +11062,7 @@
           }
         },
         "required": [
+          "cert",
           "description",
           "id",
           "name",


nexus/db-model/src/certificate.rs
```diff
@@ -82,6 +82,16 @@ impl TryFrom<Certificate> for views::Certificate {
...
+            cert: String::from_utf8(cert.cert)
+                .map_err(|err| {
+                    Error::InternalError {
+                        internal_message: format!(
+                            "Failed to construct string from stored certificate: {}",
+                            err
+                        )
+                    }
+                }
+            )?,
```
david-crespo, line 94: Worth having a comment about why this conversion is expected to always succeed. As far as I can tell, we should expect this conversion to work 100% of the time because it comes in as a `String`. 

https://github.com/oxidecomputer/omicron/blob/5176bb3c0398488d0c9669fa049e85b33fbacd54/nexus/types/src/external_api/params.rs#L811-L823

I figure the 500 is appropriate. Maybe be more concrete in the error, like: `cert was not valid UTF-8`. Though the `err` you stick in there should say that too. In general using this constructor function is slightly preferred over the struct constructor, but it doesn't matter that much, and it won't be much less noisy.

https://github.com/oxidecomputer/omicron/blob/5176bb3c0398488d0c9669fa049e85b33fbacd54/common/src/api/external/error.rs#L211-L219

augustuswm: Updated the error here and added a comment on why I expect the conversion not to fail. I think your more direct error explaining that the certificate is not valid utf-8 is better. The 500 felt off to me initially too. But It is probably correct as it it should be something that the control plane did that got us in to the error case.

davepacheco:
Good idea!  I'd love another +1 from someone more familiar with working with certificates.

I wonder if it makes more sense to put a fingerprint here?  If so, that'd make it less unwieldy to keep in the "list" endpoint.  As it is, it does seem less than ideal to include it there.  Maybe @ahl has an opinion?

augustuswm:
@ahl Any thoughts on this? Should we use an identifier in the list endpoint and then display the full contents in a view endpoint?

Ran in to this again while doing a round of cert updates, and forgot that I had started on it.

ahl:
For what it's worth, the `view` and `list` endpoints do seem to mostly (if not entirely) use the same type. If this seems unwieldy we can invent a summary type for the list endpoint. If the concern is consuming terminal space in the output, perhaps we can solve that with a custom command.

If we think most consumers of the API would want this as part of list then we should probably include it.

david-crespo:
+1 to adding it in the list. I don't see the extra bytes being a big problem. There's always https://github.com/oxidecomputer/dropshot/issues/221 (auto gzip if the accept header says to).

augustuswm:
@davepacheco - Thoughts on getting this in to release 9?

davepacheco:
Thanks for the bump.  Reading back through the comments:

- There's agreement that we'd like to include this in the "list" endpoint.  I don't love having the full certificate bytes in the list endpoint because of the size of this request (which can have many certificates, at least in principle) and general unwieldiness of having giant requests.  It's kind of a vague concern and if everyone else feels like it's fine, that's fine with me.
- But as a user, if I want to compare certificates, I think I want to compare fingerprints anyway.  No?  (I guess I'm looking for more input here, too.)
- Unrelated to that, I'd still like to get a +1 from somebody more fluent with best practices around transmitting certificates.  Maybe @iliana?

iliana:
Returning full certificates (not the private keys!) via the API is fine. The full certificate is sent to any HTTPS client anyway, whether the user is authenticated or not (for example, run `openssl s_client -connect github.com:443` and see the `-----BEGIN CERTIFICATE-----` blob, or [the steps to open Firefox's certificate viewer for a site](https://support.mozilla.org/en-US/kb/secure-website-certificate#w_viewing-a-certificate)).

I'm not sure I agree that the size of the certificates are that much of a concern; the wildcard certificates we use on our racks are less than 2 KiB PEM-encoded (so, already base64'd). An eventual snapshot block download endpoint would almost certainly return more than 10 or 20 times that that for efficiently backing up disks off-rack.

I don't think fingerprints are particularly useful. (Certificate serial numbers would honestly be more useful, and they're still not that useful on their own.) There's supposed to be a standard way to calculate fingerprints but I think different tools have failed to completely agree on that.

augustuswm:
I do think that it is important to be able to retrieve the actual contents of the certificate (via some endpoint), and as @iliana noted making an https request retrieves the data of the the certificate that the Nexus decides to serve.

The intention with returning the certificate contents is to be able to analyze the individual certificate fields (in particular expiration dates). This allows storing the key and certificate "only" in the silo, and no where else. Then when we near expiration we can generate a new key pair to rotate in to the silo.

davepacheco:
Sounds good!  So the remaining thing then is: we accept certificates as bytes, not strings; here, we're returning them as strings, which is a little inconsistent and introduces the failure mode that they're not UTF-8 encoded (which produces a 500, which is a little weird, but there's not much else we can do).  Should we just return them in bytes the same way we accept them?  (@ahl?)

ahl:
> I'm not sure I agree that the size of the certificates are that much of a concern; the wildcard certificates we use on our racks are less than 2 KiB PEM-encoded (so, already base64'd). An eventual snapshot block download endpoint would almost certainly return more than 10 or 20 times that that for efficiently backing up disks off-rack.

I think the concern regard the size is not, say, the bandwidth but rather than it would make the CLI cumbersome. I think we can sort that out client-side if it becomes an issue--let's just make the API what we expect to be most useful. If fingerprints aren't useful then let's return the full cert.

> So the remaining thing then is: we accept certificates as bytes, not strings; here, we're returning them as strings, which is a little inconsistent and introduces the failure mode that they're not UTF-8 encoded (which produces a 500, which is a little weird, but there's not much else we can do). Should we just return them in bytes the same way we accept them?

Ugh! Should we only accept them as a string? Practically is there a way to create that 500?

iliana:
Is that difference because we accept certs in PEM or DER, but then always encode them internally as PEM?

augustuswm:
I thought that we only accepted new certificates as PEM encoded Strings, and then write them as-is to the database:

https://github.com/oxidecomputer/omicron/blob/main/nexus/src/app/certificate.rs#L43
https://github.com/oxidecomputer/omicron/blob/main/nexus/types/src/external_api/params.rs#L821
https://github.com/oxidecomputer/omicron/blob/main/nexus/db-queries/src/db/datastore/certificate.rs#L46

Thought is there a different location where we accept them as bytes? I'm certainly not as familiar with Nexus.

iliana:
I am also not familiar with the full certificate lifecycle, but if we're already only accepting PEM-encoded strings, that seems right to me.

davepacheco:
Ugh, sorry -- I'm living in the past (pre-#3396).  These have been Strings for a while now.  Great!

[We do still store them as BYTES](https://github.com/oxidecomputer/omicron/blob/7fb6cb47c7b194b7088744ede0572d5e5be703a6/schema/crdb/dbinit.sql#L397-L401) in the database, unfortunately.  But that's out of scope here.  Could you file an issue on that @augustuswm?

davepacheco (approved):

davepacheco:
Ugh, sorry for yet more confusion.  I think my "approval" posted a comment I had drafted months ago and thought I sent but apparently didn't...and which we already talked about and resolved.  Sorry again!  This looks good.

augustuswm:
All good! I'll re-run the failed tests here and get an issue open on looking into the database schema.

PR #5079, opened 2024-02-15
Make snapshot delete's SPACE_ACCOUNT idempotent

jmpesp:
The SPACE_ACCOUNT node in the snapshot delete saga is not idempotent, and will fail to compensate for the forward action. This will lead to undercounting if the snapshot delete unwinds and the user deletes the snapshot again.


jmpesp:
After thinking about this, if this PR is merged then we need to mimic what is done during the disk delete unwind, namely that we should transition the snapshot to `Faulted` so that the delete can be retriggered by the user.

I'd like to add more tests for this.

PR #5081, opened 2024-02-15
Add EXPECTORATE tests for virtual_provisioning_collection CTE

smklein:
This is a direct follow-up to https://github.com/oxidecomputer/omicron/pull/5063, focused on the `virtual_provisioning_collection` CTE.

This PR adds a test which validates the current SQL query, before re-structuring it to use `TypedSqlQuery`.


nexus/db-queries/src/db/queries/virtual_provisioning_collection_update.rs
```diff
@@ -646,3 +646,71 @@ impl Query for VirtualProvisioningCollectionUpdate {
...
+            max_instance_gen,
+            cpus_diff,
+            ram_diff,
+            project_id,
+        );
+
+        expectorate_query_contents(
+            &query,
+            "tests/output/virtual_provisioning_collection_update_delete_instance.sql",
+        ).await;
```
hawkw, line 714: this is kind of nitpicky, but: what do you think about having a separate test for each query, rather than one for all of them? having a separate test per query will make it more obvious which query changed in the `nextest` output, and also let us see *all* the queries that have changed if we broke multiple queries at the same time. with one test checking all of them, we're instead in a situation where if i break one and see that a test failed, i don't see that i've broken other queries as well until i actually fix the first failure and the test can continue to the next one...

hawkw (approved):
looks good to me overall! i had one note about how to structure the test, let me knw what you thinl

nexus/db-queries/src/db/queries/virtual_provisioning_collection_update.rs
smklein: Good idea, happy to split these up!

PR #5089, opened 2024-02-16
[db-queries] Convert virtual provisioning CTE to use raw SQL

smklein:
Builds on:
- https://github.com/oxidecomputer/omicron/pull/5063
- https://github.com/oxidecomputer/omicron/pull/5081

Converts the virtual provisioning CTE to use SQL more directly


nexus/db-queries/src/db/queries/virtual_provisioning_collection_update.rs
```diff
@@ -713,4 +568,68 @@ mod test {
...
+            id,
+            max_instance_gen,
+            cpus_diff,
+            ram_diff,
+            project_id,
+        );
+        let _ = query
+            .explain_async(&conn)
+            .await
+            .expect("Failed to explain query - is it valid SQL?");
```
hawkw, line 630: similar to my suggestion on https://github.com/oxidecomputer/omicron/pull/5081#discussion_r1617990794, i might personally prefer to break up the test for each query into its own test function?

nexus/db-queries/src/db/queries/virtual_provisioning_collection_update.rs
```diff
@@ -399,129 +108,359 @@ impl VirtualProvisioningCollectionUpdate {
...
-            *crate::db::fixed_data::FLEET_ID,
-        );
-
-        let quotas = Quotas::new(&parent_silo);
-        let silo_provisioned = SiloProvisioned::new(&parent_silo);
+    ) -> TypedSqlQuery<SelectableSql<VirtualProvisioningCollection>> {
+        let query = QueryBuilder::new().sql("
+WITH
+  parent_silo AS (SELECT project.silo_id AS id FROM project WHERE project.id = ").param().sql("),")
+            .bind::<sql_types::Uuid, _>(project_id).sql("
```
hawkw, line 118: IIUC it shouldn't really be necessary to format the query strings here identically to the `format_sql` output, since we're still going to call `format_sql` on them prior to testing that they match?

hawkw (approved):
seems fine to me --- it's a tossup over which code is easier to understand, but if the raw SQL helps to avoid fighting the type checker, it's probably worth it. it might be worth getting a second opinion from someone who's not me, but i'm :+1: on this.

nexus/db-queries/src/db/queries/virtual_provisioning_collection_update.rs
smklein: Sounds good, will split these up!

PR #5113, opened 2024-02-22
"expensive operation" check was too aggressive

davepacheco:
See #5111.  In #4989 I added a check so that we couldn't list all sleds in the rack (or do other expensive things) from an API handler.  But right now, blueprint planning always happens inside an internal API handler.  So for now I'm suggesting we relax this constraint to prevent it only from the external API.


davepacheco:
Merging after getting approval from @jgallagher offline.

PR #5125, opened 2024-02-23
[test-utils] Give a little hint about running CockroachDB on test failures

smklein:
This is a pattern I use a lot, but I want to help make it more visible to folks during test failures.

Digging into the state of Cockroach after a test failure is a useful debugging tool, so I want to make
it obvious "how to do that".

### DIFF
--- a/test-utils/src/dev/db.rs
+++ b/test-utils/src/dev/db.rs
@@ -641,11 +641,11 @@ impl Drop for CockroachInstance {
                 let path = temp_dir.into_path();
                 eprintln!(
                     "WARN: temporary directory leaked: {path:?}\n\
-                     \tIf you would like to access the database for debugging, run the following:\n\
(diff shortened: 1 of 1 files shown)


david-crespo (approved):
genius

davepacheco:
This is what `omicron-dev db-run --store-dir` was originally for.  (no objection to this PR as is, just thought it would be useful to have a tool to do that for you)

davepacheco (approved):
Let me make that stronger: I think it's a great idea to direct people how to do this after this kind of test failure!

I do wonder if we should we print out the exact command line that was used to run the test?  We log that earlier for the same reason.  We could also direct them to `omicron-dev` which I think would do the same thing?  I don't think it's a big deal either way.

smklein:
Updated the instructions to advise folks to use `omicron-dev db-run`!

PR #5127, opened 2024-02-23
Upgrade clickhouse from v22.8.9.24 to v23.8.7.24

citrus-it:
The tests all now pass and I've deployed this to a bench gimlet and done some basic smoke tests.

- [x] Single node deployment;
- [x] Clustered deployment;
- [x] Upgrade from old version to new;
- [x] More tests, TBD

### DIFF
--- a/test-utils/src/dev/clickhouse.rs
+++ b/test-utils/src/dev/clickhouse.rs
@@ -722,7 +722,7 @@ async fn find_clickhouse_port_in_log(
 pub async fn wait_for_ready(
     log_path: Utf8PathBuf,
     timeout: Duration,
-    needle: &'static str,
+    needle: &str,
 ) -> Result<(), anyhow::Error> {
     let p = poll::wait_for_condition(
         || async {
@@ -758,7 +758,7 @@ pub async fn wait_for_ready(
 async fn discover_ready(
     path: &Utf8Path,
     timeout: Duration,
-    needle: &'static str,
+    needle: &str,
 ) -> Result<(), ClickHouseError> {
     let timeout = Instant::now() + timeout;
     tokio::time::timeout_at(timeout, clickhouse_ready_from_log(path, needle))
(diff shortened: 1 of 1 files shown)


oximeter/db/src/client.rs
```diff
@@ -857,9 +857,11 @@ impl Client {
             Err(Error::Database(err))
                 // Case 1: The database has not been created.
                 if err.contains("Database oximeter doesn't exist") ||
+                    err.contains("Database oximeter does not exist") ||
```
bnaecker, line 860: Nit, but maybe extract this to a const?

oximeter/db/src/configs/keeper_config.xml
```diff
@@ -9,13 +9,14 @@
...
     <!-- To allow all use :: -->
-    <listen_host from_env="CH_LISTEN_ADDR"/> 
+    <listen_host from_env="CH_LISTEN_ADDR"/>
     <path replace="true" from_env="CH_DATASTORE"/>
 
     <keeper_server>
         <tcp_port from_env="CH_LISTEN_PORT"/>
         <server_id from_env="CH_KEEPER_ID_CURRENT"/>
         <log_storage_path from_env="CH_LOG_STORAGE_PATH"/>
+        <storage_path from_env="CH_STORAGE_PATH"/>
```
bnaecker, line 19: I can't find any documentation for this new setting, have you? What does it refer to?

bnaecker (approved):
Looks good to me, thanks for pushing on this!

oximeter/db/src/configs/keeper_config.xml
citrus-it: I meant to remove that, it was a red herring when I was troubleshooting earlier.
If it isn't explicitly set, it just gets set to the parent directory of `<log_storage_path/>` - it seems to be a way of just defining a parent directory for all of the other pieces, but doesn't seem to be documented.
I expect the tests will pass without it.

karencfv (approved):
Thanks for looking into this, and also for cleaning up the tests!

oximeter/db/src/client.rs
citrus-it: Done

Issue #5058, opened 2024-02-14
Allow direct triggering of inventory collection?


andrewjstone:
Currently, when adding a sled to a rack, I"m not able to generate a blueprint from a collection that exists. I have to wait for the next round of inventory collection which can take up to 10 minutes. I believe this is because the new sled isn't in the old inventory. Once the next collection is made things work as we expect.

I'm wondering if we should add an OMDB command to trigger a collection to speed up testing.

```
root@oxz_switch:~# omdb nexus blueprints generate-from-collection 0fe01f0a-9d18-4575-8d33-91b56f00ca12
note: Nexus URL not specified.  Will pick one from DNS.
note: using DNS server for subnet fd00:1122:3344::/48
note: (if this is not right, use --dns-server to specify an alternate DNS server)
note: using Nexus URL http://[fd00:1122:3344:102::5]:12221
Error: creating blueprint from collection id

Caused by:
    Error Response: status: 500 Internal Server Error; headers: {"content-type": "application/json", "x-request-id": "233de1ba-672b-410f-947a-7a52c90fc879", "content-length": "124", "date": "Wed, 14 Feb 2024 00:17:54 GMT"}; value: Error { error_code: Some("Internal"), message: "Internal Server Error", request_id: "233de1ba-672b-410f-947a-7a52c90fc879" }
root@oxz_switch:~# omdb db inventory collections list
note: database URL not specified.  Will search DNS.
note: (override with --db-url or OMDB_DB_URL)
note: using DNS server for subnet fd00:1122:3344::/48
note: (if this is not right, use --dns-server to specify an alternate DNS server)
note: using database URL postgresql://root@[fd00:1122:3344:103::3]:32221,[fd00:1122:3344:102::3]:32221,[fd00:1122:3344:102::4]:32221,[fd00:1122:3344:101::4]:32221,[fd00:1122:3344:101::3]:32221/omicron?sslmode=disable
note: database schema version matches expected (33.0.1)
ID                                   STARTED              TOOK   NSPS NERRORS
0fe01f0a-9d18-4575-8d33-91b56f00ca12 2024-02-14T00:07:18Z 148 ms 5    0
0551af6b-4de7-4719-8bde-01a97c8d990f 2024-02-14T00:07:18Z 143 ms 5    0
36d00d44-91f2-41b5-97bf-9963adc4781f 2024-02-14T00:08:24Z 129 ms 5    0
[...]

PR #5130, opened 2024-02-23
trigger inventory collection after blueprint execution

davepacheco:
This came up in the context of #5111 and [here in particular](https://github.com/oxidecomputer/omicron/issues/5111#issuecomment-1959346002):

> > FWIW, there were 2 periods where I ended up waiting on a collection:
> 
>     * After add sled, as discussed above and we now understand
> 
>     * In between making the blueprint that added an NTP zone to the new sled and regenerating the next blueprint
> 
> 
> In the second case, even though the sled had already started the NTP zone and successfully timesync'd, I was still waiting for an inventory collection because the planner [won't add anything else to the sled until the latest collection shows the NTP zone](https://github.com/oxidecomputer/omicron/blob/42844311eb688e63ea23361dfb4279b44e65f792/nexus/deployment/src/planner.rs#L106-L139). I'm not sure where we could trigger a collection here, though - if sled-agent told us in the response whether the `PUT /omicron-zones` actually started new zones, could the reconfigurator task trigger a collection if new zones were added?

I figured that more generally, any time we execute a blueprint, successfully or otherwise, we may have made some changes and we ought to re-collect inventory.

### DIFF
--- a/nexus/src/app/background/blueprint_execution.rs
+++ b/nexus/src/app/background/blueprint_execution.rs
@@ -20,6 +20,8 @@ pub struct BlueprintExecutor {
     datastore: Arc<DataStore>,
     rx_blueprint: watch::Receiver<Option<Arc<(BlueprintTarget, Blueprint)>>>,
     nexus_label: String,
+    tx: watch::Sender<usize>,
+    count: usize,
 }
 
 impl BlueprintExecutor {
@@ -30,7 +32,12 @@ impl BlueprintExecutor {
         >,
         nexus_label: String,
     ) -> BlueprintExecutor {
-        BlueprintExecutor { datastore, rx_blueprint, nexus_label }
+        let (tx, _) = watch::channel(0);
+        BlueprintExecutor { datastore, rx_blueprint, nexus_label, tx, count: 0 }
+    }
+
+    pub fn watcher(&self) -> watch::Receiver<usize> {
+        self.tx.subscribe()
     }
 }
 
@@ -70,6 +77,10 @@ impl BackgroundTask for BlueprintExecutor {
                 &self.nexus_label,
             )
             .await;
+
+            // Trigger anybody waiting for this to finish.
+            self.count = self.count + 1;
+            self.tx.send(self.count);
 
             // Return the result as a `serde_json::Value`
             match result {
--- a/nexus/src/app/background/blueprint_execution.rs
+++ b/nexus/src/app/background/blueprint_execution.rs
@@ -21,7 +21,6 @@ pub struct BlueprintExecutor {
     rx_blueprint: watch::Receiver<Option<Arc<(BlueprintTarget, Blueprint)>>>,
     nexus_label: String,
     tx: watch::Sender<usize>,
-    count: usize,
 }
 
 impl BlueprintExecutor {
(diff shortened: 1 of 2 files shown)


nexus/src/app/background/blueprint_execution.rs
```diff
@@ -71,6 +78,10 @@ impl BackgroundTask for BlueprintExecutor {
             )
             .await;
 
+            // Trigger anybody waiting for this to finish.
+            self.count = self.count + 1;
+            self.tx.send(self.count);
```
jgallagher, line 83: I'll take "`watch` is hard to use correctly for $200, Alex"; this should probably be `send_replace` or `send_modify` so it doesn't spuriously fail if there are no subscribers.

jgallagher: Actually, if we used `send_modify` could we drop `self.count` altogether?

```rust
self.tx.send_modify(|count| *count += 1)

File: nexus/src/app/background/init.rs [hunk 2]
```diff
@@ -239,6 +216,37 @@ impl BackgroundTasks {
...
+            let task = driver.register(
+                String::from("inventory_collection"),
+                String::from(
+                    "collects hardware and software inventory data from the \
+                    whole system",
+                ),
+                config.inventory.period_secs,
+                Box::new(collector),
+                opctx.child(BTreeMap::new()),
+                vec![Box::new(rx_blueprint_exec)],
```
jgallagher, line 244: I think we currently have blueprint exec set to every 60s, and inventory collection set to every 600s. Is this going to 10x the inventory collection frequency?

nexus/src/app/background/blueprint_execution.rs
davepacheco: Good call!

nexus/src/app/background/init.rs
davepacheco: If a target is set, yes.  You could vie the 600s after this as "if there's nothing going on, collect at least this often".

I imagine we'll want to revisit these triggers and timeouts when we get to fully automating this but I think this is okay for now.

nexus/src/app/background/blueprint_execution.rs
davepacheco: Fixed in #70dfed3d

nexus/src/app/background/init.rs
jgallagher: Even as we're doing this by hand, we'll go from 600s to 60s as soon as we set a target, but it will never go back, right? Once we've set a target there will always be a target.

jgallagher: I guess what I'm kinda wishing for here is "only trigger the collection if blueprint realization actually changed something", but that might be quite difficult.

davepacheco: Agreed, but yeah, that seems hard to know.  Certainly if we get almost any errors, we have to assume something may have changed.  We could have each module in the execution part return information about whether it _might_ have done anything that would require re-inventorying.  Concretely, we'd need sled agent to tell us whether it made any changes when we did `PUT /omicron-zones`.  This seems like a potentially fine optimization, but also just an optimization (and bugs in this area would be really annoying).  I think it makes sense to wait until this becomes more of a problem.  What do you think?

jgallagher (approved):

andrewjstone:
This is great. Thanks @davepacheco. While it still may be worthy it to implement https://github.com/oxidecomputer/omicron/issues/5058, it is now much less of a priority!

PR #5135, opened 2024-02-23
Accept notifications from Crucible

### LINKED ISSUE #5120: Accept repair status reports from Crucible
Implement the server side of https://github.com/oxidecomputer/crucible/issues/839 by accepting the following notifications:

- live repair has started
- live repair has completed either ok or with an error

jmpesp:
Allow any Upstairs to notify Nexus about the start or completion (plus status) of live repairs. The motivation for this was to be used in the final stage of region replacement to notify Nexus that the replacement has finished, but more generally this can be used to keep track of how many times repair occurs for each region.

Fixes #5120


schema/crdb/37.0.0/up02.sql
```diff
@@ -0,0 +1,19 @@
+CREATE TABLE IF NOT EXISTS live_repair_notification (
+    time TIMESTAMPTZ NOT NULL,
```
leftwo, line 2: It this timestamp the timestamp when the record was last updated?

uuid-kinds/src/lib.rs
```diff
@@ -45,6 +45,10 @@ macro_rules! impl_typed_uuid_kind {
 // Please keep this list in alphabetical order.
 
 impl_typed_uuid_kind! {
+    DownstairsRegionKind => "downstairs_region",
```
leftwo, line 48: This won't apply to any existing UUIDs for regions, right?  Only the new stuff you are
adding here?

leftwo (commented):
I have some questions about this, to make sure I understand it :)

The plan is that we will be storing information about a volume under repair
in a new table in the database?  But I'm not clear on how we connect this table
back to a specific volume.

I do think we should not call this `LiveRepair..` and just call it `Repair..` everywhere.
We can re-use all of this for the same mechanism that happens when an upstairs
first starts and needs to reconcile the three downstairs with each other. 

It seems like it would not be too difficult to add an `update` or `status` endpoint
that could report to nexus the progress of a repair.  Maybe we want to put a
column in the database now to allow us to start using it later?

Also, thinking about migration cases, do we want another column in the the
database to indicate why?  I've not given this much thought yet, and I don't
think this particular table is going to be exposed to the user.  I'm just thinking
about future us trying to determine why a `Repair` was started.  Having some
fields in the database to leave us breadcrumbs might be helpful.

nexus/db-model/src/live_repair.rs
```diff
@@ -0,0 +1,95 @@
...
+    Started => b"started"
+    Succeeded => b"succeeded"
+    Failed => b"failed"
+);
+
+/// A record of Crucible live repair notifications: when a live repair started,
+/// succeeded, failed, etc.
+///
+/// Each live repair attempt is uniquely identified by the repair ID, upstairs
+/// ID, session ID, and region ID. How those change tells Nexus about what is
```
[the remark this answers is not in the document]
jmpesp: It's something I still have to add in the corresponding Crucible PR - it is unique per repair, and is chosen by the Upstairs.

nexus/db-model/src/live_repair.rs
```diff
@@ -0,0 +1,95 @@
...
+#[derive(Queryable, Insertable, Debug, Clone, Selectable)]
+#[diesel(table_name = live_repair_notification)]
+pub struct LiveRepairNotification {
+    pub time: DateTime<Utc>,
+
+    pub repair_id: DbTypedUuid<LiveRepairKind>,
+    pub upstairs_id: DbTypedUuid<UpstairsKind>,
+    pub session_id: DbTypedUuid<UpstairsSessionKind>,
+
+    pub region_id: DbTypedUuid<DownstairsRegionKind>,
```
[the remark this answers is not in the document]
jmpesp: True! But Nexus is keeping track of repairs per region, so it has to create multiple records for this. Nexus will be able to tell that multiple downstairs are being repaired at the same time using the repair id.

nexus/src/app/volume.rs
```diff
@@ -30,4 +36,78 @@ impl super::Nexus {
...
+                            LiveRepairNotificationType::Failed
+                        } else {
+                            LiveRepairNotificationType::Succeeded
+                        },
+                    ),
+                )
+                .await?;
+
+            if !repair_finish_info.aborted {
+                // TODO-followup if there's an active region replacement
```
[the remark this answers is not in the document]
jmpesp: Right :) The `TODO-followup` is for me to fill this code in by kicking off the "region_replacement_finish" saga, which was separated out of this PR.

schema/crdb/37.0.0/up02.sql
jmpesp: Yes

uuid-kinds/src/lib.rs
jmpesp: Correct, that's still work that is required.

nexus/db-queries/src/db/datastore/volume.rs
```diff
@@ -809,6 +812,101 @@ impl DataStore {
...
+                                LiveRepairNotification,
+                            > = dsl::live_repair_notification
+                                .filter(dsl::repair_id.eq(record.repair_id))
+                                .filter(dsl::upstairs_id.eq(record.upstairs_id))
+                                .filter(dsl::session_id.eq(record.session_id))
+                                .filter(dsl::region_id.eq(record.region_id))
+                                .filter(dsl::notification_type.eq_any(vec![
+                                    LiveRepairNotificationType::Succeeded,
+                                    LiveRepairNotificationType::Failed,
+                                ]))
```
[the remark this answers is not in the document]
jmpesp: This does touch on a problem. Each Upstairs should probably retry notifying Nexus until there's a known result of that notification, like saga nodes do. Currently though the Upstairs is stateless - if it crashes, then the state of that notification retry (or any notifications) is lost. Without a place to store that state the potential for lost notifications exists.

Even without a crash it is conceivable that due to some asymmetric network outage an Upstairs would also _not_ be able to contact Nexus, but still able to receive, via propolis, volume replacement commands. It would then try to send out those notifications only to spin retrying them.

It's possible that Nexus could detect this case. I know the PR isn't up yet but the forthcoming one will contain region replacement request records. If Nexus completes the "region replacement start" saga then (ignoring the other cases for a sec) it will have sent a volume replacement request to an instance's propolis. Assuming the Propolis server stays up, there could be some action taken if Nexus does not receive any "live repair started" or "live repair finished" despite having successfully sent a volume replacement request.

There's more thinking to be done here, and probably more hairy cases to work though. For now I think it's ok to accept a Success/Failed message without a Started message to account for any transient network problems getting in the way of the Started notification.

jmpesp:
> The plan is that we will be storing information about a volume under repair in a new table in the database? But I'm not clear on how we connect this table back to a specific volume.

There's going to be a region replacement request that can be tied back to the Volume by checking the connection addresses in the sub volumes. For read-write region replacements, the region under replacement can only be tied back to one volume, and because of the current region allocation strategy each region is allocated on a separate sled.

Note this is to work around us not storing the region ID in the VCR, and because we have to assume that the region's Crucible agent will not be accessible.

> I do think we should not call this LiveRepair.. and just call it Repair.. everywhere. We can re-use all of this for the same mechanism that happens when an upstairs first starts and needs to reconcile the three downstairs with each other.

Agreed, done in ff08b0001

> It seems like it would not be too difficult to add an update or status endpoint that could report to nexus the progress of a repair. Maybe we want to put a column in the database now to allow us to start using it later?

:+1:, done in 0cd8601af

> Also, thinking about migration cases, do we want another column in the the database to indicate why? I've not given this much thought yet, and I don't think this particular table is going to be exposed to the user. I'm just thinking about future us trying to determine why a Repair was started. Having some fields in the database to leave us breadcrumbs might be helpful.

I'm also going to think more about this tonight.

nexus/db-queries/src/db/datastore/volume.rs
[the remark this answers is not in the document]
leftwo: Yeah, I think allowing it seems like the best thing to do here.

schema/crdb/37.0.0/up01.sql
```diff
@@ -1,5 +1,13 @@
-CREATE TYPE IF NOT EXISTS omicron.public.live_repair_notification_type AS ENUM (
-  'started',
-  'succeeded',
-  'failed'
+-- The disposition for a particular sled. This is updated solely by the
```
leftwo, line 1: Is this showing up in my diff because of the merge (and now I see the actual 37)
where this PR now has 38?

leftwo (commented):
I know I asked for it, but now I have a complication for you.

When doing an initial reconciliation, it's possible that all three downstairs have 
different extents that need correction.  Unlikely, but possible.

How would we represent this?  By having all three regions report they are being
repaired?   I think that can work, but it may look strange from the outside.

schema/crdb/37.0.0/up01.sql
jmpesp: Yeah, I think so!

nexus/db-model/src/schema.rs
```diff
@@ -1519,18 +1519,28 @@ table! {
...
+        notification_type -> crate::UpstairsRepairNotificationTypeEnum,
+    }
+}
+
+table! {
+    upstairs_repair_progress (repair_id, time, current_item, total_items) {
+        repair_id -> Uuid,
+        time -> Timestamptz,
+        current_item -> Int8,
+        total_items -> Int8,
```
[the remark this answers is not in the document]
jmpesp: It is, yeah

nexus/db-model/src/live_repair.rs
```diff
@@ -0,0 +1,95 @@
...
+///   repair ID is different, then the same Upstairs is trying to repair that
+///   region again. This could be due to a failed first attempt, or that
+///   downstairs may have been kicked out again.
+///
+/// - if the upstairs ID and region ID are the same, but the session ID and
+///   repair ID are different, then a different session of the same Upstairs is
+///   trying to repair that Downstairs. Session IDs change each time the
+///   Upstairs is created, so it could have crashed, or it could have been
+///   migrated and the destination Propolis' Upstairs is attempting to repair
+///   the same region.
```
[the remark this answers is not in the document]
leftwo: If there is an existing newer record, we should be able to tell the difference between an unfinished
repair and a later repair with "almost" the same data that finished it.  Also, as the driver of "what needs
a repair" is not this table, I don't think there is any issue if things are left behind here.

jmpesp:
> When doing an initial reconciliation, it's possible that all three downstairs have different extents that need correction. Unlikely, but possible.
>
> How would we represent this? By having all three regions report they are being repaired? I think that can work, but it may look strange from the outside.

This would look like three `Reconciliation` records all with the same repair id, which we can then tell is this unlikely scenario :)

nexus/db-queries/src/db/datastore/volume.rs
```diff
@@ -819,47 +824,68 @@ impl DataStore {
...
                 async move {
+                    // Return 409 if a repair ID does not match types
+                    let mismatched_record_type_count: usize =
+                        dsl::upstairs_repair_notification
+                            .filter(dsl::repair_id.eq(record.repair_id))
+                            .filter(dsl::repair_type.ne(record.repair_type))
+                            .execute_async(&conn)
+                            .await?;
+
+                    if mismatched_record_type_count == 0 {
```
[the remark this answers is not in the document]
jmpesp: Yeeeep 832e649f9

jmpesp:
> I'm just thinking about future us trying to determine why a Repair was started. Having some fields in the database to leave us breadcrumbs might be helpful.

Back to this - I was thinking what a breadcrumb would get us over now having the repair type enum.

If it's `Live`, then there are two reasons for starting live repair (that I know of): receiving an IO error for a write / flush / extent repair op from a Downstairs, or the queue of live work exceeds `IO_OUTSTANDING_MAX`. 

If it's `Reconciliation`, then that's when extents don't match at startup.

We could plumb up `ClientStopReason`?

leftwo:
> > Back to this - I was thinking what a breadcrumb would get us over now having the repair type enum.
> > If it's `Live`, then there are two reasons for starting live repair (that I know of): receiving an IO error for a write / flush / extent repair op from a Downstairs, or the queue of live work exceeds `IO_OUTSTANDING_MAX`.
> > If it's `Reconciliation`, then that's when extents don't match at startup.
> > We could plumb up `ClientStopReason`?
> 
> Yeah, I guess we would have to make up types for each possible reason.

Oh, if Nexus tells the Upstairs that it has to replace a downstairs due to a pending upgrade on that
sled, that would be another reason.  The Upstairs should know the difference though between a
LiveRepair that started as a result of a replacement, and a LiveRepair that started because there was
an error.

nexus/db-queries/src/db/datastore/volume.rs
```diff
@@ -907,6 +928,66 @@ impl DataStore {
...
+                        notification_dsl::upstairs_repair_notification
+                            .filter(notification_dsl::repair_id.eq(nexus_db_model::to_db_typed_uuid(repair_id)))
+                            .filter(notification_dsl::upstairs_id.eq(nexus_db_model::to_db_typed_uuid(upstairs_id)))
+                            .filter(notification_dsl::notification_type.eq(UpstairsRepairNotificationType::Started))
+                            .get_result_async(&conn)
+                            .await
+                            .optional()?;
+
+                    if matching_repair.is_none() {
+                        // XXX should be 404
```
leftwo, line 963: If this should be 404, can we make it?  Or that's just not possible from this place?

nexus/tests/integration_tests/volume_management.rs
```diff
@@ -2552,3 +2559,619 @@ async fn test_volume_hard_delete_idempotent(
...
+    let region_id: TypedUuid<DownstairsRegionKind> = TypedUuid::new_v4();
+
+    // Notify start
+    let notify_url = format!("/upstairs/{upstairs_id}/repair-start");
+
+    int_client
+        .make_request(
+            Method::POST,
+            &notify_url,
+            Some(internal::nexus::RepairStartInfo {
```
leftwo, line 2584: Nit: it seems like this `internal::nexus::RepairStartInfo` and the one called
below are the same.  Could we pull this out to a variable and call the
`make_request` with it like that?  It would be easier to see right away that
the test was sending the same content twice (and fewer lines of code :) )

leftwo (approved):
We may need to fine tune this as we get further down the path with ds replacement, 
but it seems like this is what we think we need now.  If we need to update it later
we can do that.

nexus/db-model/src/schema.rs
```diff
@@ -13,7 +13,7 @@ use omicron_common::api::external::SemverVersion;
 ///
 /// This should be updated whenever the schema is changed. For more details,
 /// refer to: schema/crdb/README.adoc
-pub const SCHEMA_VERSION: SemverVersion = SemverVersion::new(37, 0, 0);
+pub const SCHEMA_VERSION: SemverVersion = SemverVersion::new(38, 0, 0);
```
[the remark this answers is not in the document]
jmpesp: yeeeeeeeeeeeep

nexus/db-queries/src/db/datastore/volume.rs
jmpesp: Done in 0b37b63bb752243298e894bb186eb8d6e907374b

nexus/tests/integration_tests/volume_management.rs
jmpesp: done in 3dfabf027

nexus/db-model/src/live_repair.rs
```diff
@@ -0,0 +1,95 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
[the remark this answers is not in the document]
jmpesp: They're noted as optional in https://rfd.shared.oxide.computer/rfd/0224#_copyright_licensing_notices

schema/crdb/41.0.0/up05.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TYPE IF NOT EXISTS omicron.public.downstairs_client_stop_reason_type AS ENUM (
+  'replacing',
+  'disabled',
+  'failed_reconcile',
+  'io_error',
+  'bad_negotiation_order',
+  'incompatible',
+  'failed_live_repair',
+  'too_many_outstanding_jobs',
+  'deactivated'
```
leftwo, line 10: Do we need a `timeout` choice too?  It could be either a connection closed from the remote side, or just nothing and we gave up on it.

leftwo (approved):

schema/crdb/41.0.0/up05.sql
jmpesp: I think so, yeah - `ClientStopReason` is recorded now but not if the client task has a ClientRunResult. I'll do that now.

jmpesp: df79ff2dc adds separate endpoints for when an Upstairs requests a stop, and when a Downstairs client task stops.

Issue #5051, opened 2024-02-13
[test flake] test_instance_serial failed: port re-use?


smklein:
Saw the following failure locally:

```
--- STDERR:              omicron-nexus::test_all integration_tests::instances::test_instance_serial ---                
log file: /tmp/test_all-89aaaa31bb79b02a-test_instance_serial.470459.0.log                                             
note: configured to log to "/tmp/test_all-89aaaa31bb79b02a-test_instance_serial.470459.0.log"                          
DB URL: postgresql://root@[::1]:44551/omicron?sslmode=disable                                                          
DB address: [::1]:44551                                                                                                
log file: /tmp/test_all-89aaaa31bb79b02a-test_instance_serial.470459.2.log                                             
note: configured to log to "/tmp/test_all-89aaaa31bb79b02a-test_instance_serial.470459.2.log"                          
thread 'integration_tests::instances::test_instance_serial' panicked at nexus/tests/integration_tests/instances.rs:3770:10:
called `Result::unwrap()` on an `Err` value: ServiceUnavailable { internal_message: "initializing propolis-server: error creating server listener: Address already in use (os error 98)" }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace                                          
WARN: dropped CockroachInstance without cleaning it up first (there may still be a child process running and a temporary directory leaked)
WARN: temporary directory leaked: /tmp/.tmpao5INd                                                                      
WARN: dropped ClickHouseInstance without cleaning it up first (there may still be a child process running and a temporary directory leaked)
WARN: dropped DendriteInstance without cleaning it up first (there may still be a child process running and a temporary directory leaked)
WARN: dendrite temporary directory leaked: /tmp/.tmpXfxBIx                                                             
WARN: dropped DendriteInstance without cleaning it up first (there may still be a child process running and a temporary directory leaked)
WARN: dendrite temporary directory leaked: /tmp/.tmpB5PPSJ                                                             
WARN: dropped MgdInstance without cleaning it up first (there may still be a child process running and a temporary directory leaked)
WARN: mgd temporary directory leaked: /tmp/.tmp5XnP9L                                                                  
WARN: dropped MgdInstance without cleaning it up first (there may still be a child process running and a temporary directory leaked)
WARN: mgd temporary directory leaked: /tmp/.tmpE1LFfU  
```

[...]

PR #5138, opened 2024-02-24
Push hardcoded PROPOLIS_PORT out of all but the instance creation/migration sagas

lifning:
(addressing #5051 )

Adds a database column for Nexus to track `propolis_port` in addition to the `propolis_ip`, such that tests using sled-agent-sim don't collide.

Restrict hardcoded PROPOLIS_PORT to simply being the default in sagas. It remains 12400 in production, since there's no reason for it not to be, but for scenarios wherein each "propolis-server" API isn't necessarily served from an illumos zone with its own IP address, we can let the port be automatically assigned.


nexus/db-model/src/vmm.rs
```diff
@@ -43,6 +43,9 @@ pub struct Vmm {
     /// The IP address at which this VMM is serving the Propolis server API.
     pub propolis_ip: ipnetwork::IpNetwork,
 
+    /// The socket port on which this VMM is serving the Propolis server API.
+    pub propolis_port: i32,
```
gjcolombo, line 47: I'd look at using `nexus_db_model::SqlU16` for this--it has some helper impls that might cut down on the number of `u16`/`i32` conversions Nexus needs to do.

gjcolombo (approved):
LGTM, with one small suggestion in `struct Vmm` that I think will cut down on a bit of type conversion boilerplate.

nexus/db-model/src/vmm.rs
lifning: oh! good to know, thanks

PR #5143, opened 2024-02-26
Blueprint execution: Add dataset records for Crucible zones

### LINKED ISSUE #5118: Reconfigurator needs to create datasets for new crucible zones
(This is from problem 7 of #5111)

After adding a new sled and going through the reconfigurator to set up zones, instances were able to launch on that sled, but its crucible zones were never chosen for new disks. If I understand correctly, the region allocation starts by building a list of [datasets and amount of space used](https://github.com/oxidecomputer/omicron/blob/42844311eb688e63ea23361dfb4279b44e65f792/nexus/db-queries/src/db/queries/region_allocation.rs#L263-L280). On the new sled, before the crucible zones were added, the sled's self-report to Nexus had triggered the insertion of relevant rows in the `zpools` table, which is what the reconfigurator [keyed on to create new crucible zones](https://github.com/oxidecomputer/omicron/blob/42844311eb688e63ea23361dfb4279b44e65f792/nexus/deployment/src/planner.rs#L141-L157). However, creation of crucible zones on the sled doesn't trigger any insertions into the `dataset` table, so I don't believe allocation would ever consider them.

During the RSS handoff, [Nexus inserts all datasets it got from RSS with an `on_conflict(id).do_update(...)`](https://github.com/oxidecomputer/omicron/blob/42844311eb688e63ea23361dfb4279b44e65f792/nexus/db-queries/src/db/datastore/rack.rs#L648-L675). RSS builds this list of datasets [based on all services in the rack-wide service plan](https://github.com/oxidecomputer/omicron/blob/42844311eb688e63ea23361dfb4279b44e65f792/sled-agent/src/rack_setup/service.rs#L572-L596).

I think reconfigurator needs to create these dataset rows for crucible zones it creates, but in general it's trying to operate in an idempotent fashion and may attempt to execute the same blueprint many times. I'm also not sure about the timing of when it's appropriate for dataset rows to appear. Based on all the above I think I would propose:

1. The reconfigurator only attempts to reconcile dataset rows after it has successfully sent a `PUT /omicron-zones` to all sleds. (There may be a separate issue here about what to do if reconfigurator fails when talking to some sleds but not all, but I don't think that only affects datasets.)
2. After all the `PUT /omicron-zones` have completed, reconfigurator runs an "insert if not exists" with a dataset for each crucible zone that was present on any sled. Critically, this is _not_ an upsert: we don't want to change any of the values for datasets that already existed; we only want to insert new datasets (with size_used = 0) for crucible zones that don't yet have one.

Does that look correct?

jgallagher:
On each invocation, for every Crucible zone present in the blueprint (all of which have already successfully been sent to sled-agent by this point), we attempt to insert a `dataset` record for that zone, but only if a record does not already exist. The datasets themselves are created by sled-agent when we send the zone configs.

Fixes #5118.


nexus/db-queries/src/db/datastore/dataset.rs
```diff
@@ -73,9 +82,262 @@ impl DataStore {
...
+            AsyncInsertError::CollectionNotFound => Error::ObjectNotFound {
+                type_name: ResourceType::Zpool,
+                lookup_type: LookupType::ById(zpool_id),
+            },
+            AsyncInsertError::DatabaseError(e) => public_error_from_diesel(
+                e,
+                ErrorHandler::Conflict(
+                    ResourceType::Dataset,
+                    &dataset_id.to_string(),
+                ),
```
davepacheco, line 125: We expect the `INSERT` to swallow this error, right?  i.e., I don't think you should ever get a conflict here, right?

nexus/blueprint-execution/src/datasets.rs
```diff
@@ -0,0 +1,311 @@
...
+/// record exists in `datastore`
+///
+/// Does not modify any existing dataset records. Returns the number of datasets
+/// inserted.
+pub(crate) async fn ensure_crucible_dataset_records_exist(
+    opctx: &OpContext,
+    datastore: &DataStore,
+    all_omicron_zones: impl Iterator<Item = &OmicronZoneConfig>,
+) -> anyhow::Result<usize> {
+    // Before attempting to insert any datasets, first query for any existing
```
davepacheco, line 32: I like this behavior (and the comment).

davepacheco (approved):
Nice!  I recommend another review from someone more familiar how we use datasets.

nexus/db-queries/src/db/datastore/dataset.rs
jgallagher: That's correct; this is some copy/paste from another query. I'll see if there's a way rewrite this without the misleading `Conflict`.

jgallagher: Replaced with `::Server` in f99fbdca2

nexus/db-queries/src/db/datastore/dataset.rs
```diff
@@ -73,9 +82,261 @@ impl DataStore {
...
             ),
         })
     }
+
+    /// Stores a new dataset in the database, but only if a dataset with the
+    /// given `id` does not already exist
+    ///
+    /// Does not update existing rows. If a dataset with the given ID already
+    /// exists, returns `Ok(None)`.
+    pub async fn dataset_insert_if_not_exists(
```
smklein, line 96: This seems like it'll fit in well with the existing "sled agent inserts dataset" structure, as we transition more towards "Nexus controls dataset creation".

nexus/db-queries/src/db/datastore/dataset.rs
```diff
@@ -73,9 +82,261 @@ impl DataStore {
...
+    /// This should generally not be used in API handlers or other
+    /// latency-sensitive contexts, but it can make sense in saga actions or
+    /// background tasks.
+    pub async fn dataset_list_all_batched(
+        &self,
+        opctx: &OpContext,
+        filter_kind: Option<DatasetKind>,
+    ) -> ListResultVec<Dataset> {
+        opctx.authorize(authz::Action::ListChildren, &authz::FLEET).await?;
+        opctx.check_complex_operations_allowed()?;
```
smklein, line 168: huh TIL about this opctx call, good to know!

nexus/db-queries/src/db/datastore/dataset.rs
```diff
@@ -73,9 +82,261 @@ impl DataStore {
...
+                p.found_batch(&batch, &|d: &nexus_db_model::Dataset| d.id());
+            all_datasets.extend(batch);
+        }
+
+        Ok(all_datasets)
+    }
+}
+
+#[cfg(test)]
+mod test {
```
smklein, line 186: Thanks for all the tests in this PR!

nexus/db-queries/src/db/datastore/dataset.rs
```diff
@@ -73,9 +82,261 @@ impl DataStore {
...
+            }
+        })
+    }
+
+    /// List one page of datasets
+    ///
+    /// If `filter_kind` is `Some(value)`, only datasets with a `kind` matching
+    /// `value` will be returned. If `filter_kind` is `None`, all datasets will
+    /// be returned.
+    pub async fn dataset_list(
```
smklein, line 130: Does this need to be `pub`?

smklein (approved):

nexus/db-queries/src/db/datastore/dataset.rs
jgallagher: It's quite new (from #4989)!

nexus/db-queries/src/db/datastore/dataset.rs
jgallagher: Not currently; I will make it not so.

PR #5144, opened 2024-02-26
Update Floating IP field name from address to ip

### LINKED ISSUE #5046: Change `FloatingIpCreate` `address` field to `ip` to match everything else
We should pick one or the other. To me, `ip` is clear and short and seems to be used in more places, but has the grammatical issue of being the IP of a `FloatingIp`. If that's fine we should probably stick with it. I don't like `address` as much.

https://github.com/oxidecomputer/omicron/blob/e1c3dd75b818147b545f9a46acfcf3fc19291473/nexus/types/src/external_api/params.rs#L879-L891

https://github.com/oxidecomputer/omicron/blob/e1c3dd75b818147b545f9a46acfcf3fc19291473/nexus/types/src/external_api/views.rs#L369-L379

The above is the user-facing disparity, but notably, the DB model and table use `ip`.

https://github.com/oxidecomputer/omicron/blob/e1c3dd75b818147b545f9a46acfcf3fc19291473/nexus/db-model/src/external_ip.rs#L131-L141

https://github.com/oxidecomputer/omicron/blob/e1c3dd75b818147b545f9a46acfcf3fc19291473/schema/crdb/dbinit.sql#L1726-L1727

charliepark:
Fixes #5046

Our API for Floating IPs currently has an `address` field, though in a few other places we use `ip`. This PR just makes the API more consistent across the API (and DB).

### DIFF
--- a/nexus/db-queries/src/db/datastore/external_ip.rs
+++ b/nexus/db-queries/src/db/datastore/external_ip.rs
@@ -256,7 +256,7 @@ impl DataStore {
 
         let pool_id = pool.id();
 
-        let data = if let Some(ip) = params.address {
+        let data = if let Some(ip) = params.ip {
             IncompleteExternalIp::for_floating_explicit(
                 ip_id,
                 &Name(params.identity.name),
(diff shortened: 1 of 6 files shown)


david-crespo (approved):
dab

PR #5145, opened 2024-02-27
Report physical disks through the inventory system

smklein:
- In Sled Agent, send info about physical storage up to Nexus
- In Nexus, process physical disks and add them to inventory in a new `inv_physical_disk` table
- This PR also adjusts inventory tests to account for this new "physical disk" information being added alongside sleds, and adds some disk to the "representative inventory data"

Fixes https://github.com/oxidecomputer/omicron/issues/5150


jgallagher (approved):
LGTM; just two tiny nits picking at memory usage that probably doesn't matter - feel free to ignore either or both.

schema/crdb/38.0.0/up01.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE IF NOT EXISTS omicron.public.inv_physical_disk (
```
davepacheco, line 1: Can/should we include the slot number here?  (if so, can/should we use that for pagination/uniqueness?)

sled-agent/src/sim/sled_agent.rs
```diff
@@ -750,6 +752,17 @@ impl SledAgent {
                 self.config.hardware.reservoir_ram,
             )
             .context("reservoir_size")?,
+            disks: self
+                .storage
+                .lock()
```
davepacheco, line 757: Are there any situations where this lock might be held for a long time (thinking tens of seconds or minutes)?

nexus/db-queries/src/db/datastore/inventory.rs
```diff
@@ -639,6 +659,20 @@ impl DataStore {
                 }
             }
 
+            // Insert rows for all the physical disks we found.
+            {
+                use db::schema::inv_physical_disk::dsl;
+
+                let mut iter =
+                    disks.chunks(SQL_BATCH_SIZE.get().try_into().unwrap());
```
davepacheco, line 667: In general I like the idea of inserting in batches.  But this is all going into one transaction so I'm not sure it's any better in that case.

davepacheco (approved):

nexus/db-queries/src/db/datastore/inventory.rs
```diff
@@ -125,6 +126,25 @@ impl DataStore {
...
+                    .disks
+                    .iter()
+                    .map(|disk| {
+                        InvPhysicalDisk::new(
+                            collection_id,
+                            *sled_id,
+                            disk.clone(),
+                        )
+                    })
+                    .collect::<Vec<_>>()
```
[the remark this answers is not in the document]
smklein: Fixed in ad6bf23b44a608412ed70358a3d90a40fc108c04

nexus/db-queries/src/db/datastore/inventory.rs
```diff
@@ -639,6 +659,20 @@ impl DataStore {
                 }
             }
 
+            // Insert rows for all the physical disks we found.
+            {
+                use db::schema::inv_physical_disk::dsl;
+
+                let mut iter =
+                    disks.chunks(SQL_BATCH_SIZE.get().try_into().unwrap());
+                while let Some(some_disks) = iter.next() {
```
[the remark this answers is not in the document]
smklein: On board with it, fixed in ad6bf23b44a608412ed70358a3d90a40fc108c04

schema/crdb/38.0.0/up01.sql
smklein: Sure, we'll probably want that out of inventory anyway. Also, I agree, this seems like a better PK ("sled id" + "slot") than the reported values of serial/model/vendor.

Updated to plumb through the slot from the sled agent, and using as pagination key in this PR.

sled-agent/src/sim/sled_agent.rs
smklein: I don't believe so, especially not since this is the simulated sled agent. At a cursory glance, i think it's mostly just getting locked to populate / grab contents from a group of hashmaps.

nexus/db-queries/src/db/datastore/inventory.rs
smklein: The batch size is still fairly large 1,000, so it's unlikely we'll be looping much regardless on a single rack or two. But I could see this blowing past the size of a "single SQL statement" if we consider it for larger multi-rack cases?

smklein:
`reconfigurator-cli::test_basic test_complex` looks like it's failing, and I think that requires a particular lab environment that's being used to update, so I guess this PR is blocked for the moment. This seems like a bit of a hard bottleneck (can only be safely edited from a single, specific deployment)?

nexus/db-queries/src/db/datastore/inventory.rs
```diff
@@ -125,6 +126,17 @@ impl DataStore {
...
+        // Pull disks out of all sled agents
+        let disks: Vec<_> = collection
+            .sled_agents
+            .iter()
+            .flat_map(|(sled_id, sled_agent)| {
+                sled_agent.disks.iter().map(|disk| {
+                    InvPhysicalDisk::new(collection_id, *sled_id, disk.clone())
+                })
+            })
+            .collect();
```
jgallagher, line 138: Tiny nit: I think you can drop this `collect()` (and the corresponding `: Vec<_>` type hint), make this `mut`, and then remove the `disks.into_iter()` call on line 659.

davepacheco:
> `reconfigurator-cli::test_basic test_complex` looks like it's failing, and I think that requires a particular lab environment that's being used to update, so I guess this PR is blocked for the moment. This seems like a bit of a hard bottleneck (can only be safely edited from a single, specific deployment)?

The purpose of that file is to exercise the CLI with data saved from a real system.  That seemed useful because part of the goal was to be able to import data from real systems.  But it could be more trouble than it's worth while we're still evolving the blueprint format.

Also, it doesn't have to be any particular system.  The test will dynamically pick an inventory collection and blueprints that it finds.

schema/rss-service-plan-v3.json
```diff
@@ -0,0 +1,828 @@
+{
```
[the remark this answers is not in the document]
smklein: It *is* used, and is stored, but it's not used particularly frequently.

- We always try to "load a plan if one already exists" before creating a new one: https://github.com/oxidecomputer/omicron/blob/145f25cd9b8665d1bc4906a06369dbcac7dea268/sled-agent/src/rack_setup/service.rs#L959-L971
- This tried to load with the latest version: https://github.com/oxidecomputer/omicron/blob/145f25cd9b8665d1bc4906a06369dbcac7dea268/sled-agent/src/rack_setup/plan/service.rs#L137-L172 , and in other cases, at least it tries to report a better error

In the limit: I think we may get more mileage out of "wiping everything and forcing RSS to restart" in the case where we restart, which would make these intermediate files removable.

nexus/db-queries/src/db/datastore/inventory.rs
jgallagher: Grrrr, this is very annoying. I checked that diff against `nexus-db-queries` and it was fine, but didn't check whether it cause compilation errors down the line. That _feels_ like a bug - if my function signature hasn't changed, why can implementation changes cause client breakage?

smklein:
> > `reconfigurator-cli::test_basic test_complex` looks like it's failing, and I think that requires a particular lab environment that's being used to update, so I guess this PR is blocked for the moment. This seems like a bit of a hard bottleneck (can only be safely edited from a single, specific deployment)?
> 
> The purpose of that file is to exercise the CLI with data saved from a real system. That seemed useful because part of the goal was to be able to import data from real systems. But it could be more trouble than it's worth while we're still evolving the blueprint format.
> 
> Also, it doesn't have to be any particular system. The test will dynamically pick an inventory collection and blueprints that it finds.

I logged onto a machine set up by a4x2, configured my ssh keys to be able to copy files into and out of that machine, used zlogin to access, the switch zone, and ran `omdb db reconfigurator-save` as the test recommended, but it is still failing because I didn't generate enough blueprints.

I can spin up this environment and take another lap - to clarify, do I need to do `omdb nexus blueprints regenerate` twice? I'm not sure if I'm being dumb and there are some docs I should be following for "how to do this the right way".

nexus/db-queries/src/db/datastore/inventory.rs
jgallagher: Thanks to @cbiffle who found we're hitting https://github.com/rust-lang/rust/issues/99492 here; this _ought_ to work. I found a couple weird looking workarounds using hints from that issue:

```diff
@@ -127,15 +127,10 @@ impl DataStore {
             })
             .collect::<Result<Vec<_>, Error>>()?;
         // Pull disks out of all sled agents
-        let disks: Vec<_> = collection
-            .sled_agents
-            .iter()
-            .flat_map(|(sled_id, sled_agent)| {
-                sled_agent.disks.iter().map(|disk| {
-                    InvPhysicalDisk::new(collection_id, *sled_id, disk.clone())
-                })
-            })
-            .collect();
+        let mut disks =
+            collection.sled_agents.iter().flat_map(|(sled_id, sled_agent)| {
+                std::iter::repeat(sled_id).zip(sled_agent.disks.iter())
+            });

         // Partition the sled agents into those with an associated baseboard id
         // and those without one.  We handle these pretty differently.
@@ -656,10 +651,18 @@ impl DataStore {
                 use db::schema::inv_physical_disk::dsl;

                 let batch_size = SQL_BATCH_SIZE.get().try_into().unwrap();
-                let mut disks = disks.into_iter();
                 loop {
-                    let some_disks =
-                        disks.by_ref().take(batch_size).collect::<Vec<_>>();
+                    let some_disks = disks
+                        .by_ref()
+                        .take(batch_size)
+                        .map(|(&sled_id, disk)| {
+                            InvPhysicalDisk::new(
+                                collection_id,
+                                sled_id,
+                                disk.clone(),
+                            )
+                        })
+                        .collect::<Vec<_>>();
                     if some_disks.is_empty() {
                         break;
                     }
```

or if we don't like `collection_id` floating around, it can go in the `repeat` too:

```diff
@@ -127,15 +127,10 @@ impl DataStore {
             })
             .collect::<Result<Vec<_>, Error>>()?;
         // Pull disks out of all sled agents
-        let disks: Vec<_> = collection
-            .sled_agents
-            .iter()
-            .flat_map(|(sled_id, sled_agent)| {
-                sled_agent.disks.iter().map(|disk| {
-                    InvPhysicalDisk::new(collection_id, *sled_id, disk.clone())
-                })
-            })
-            .collect();
+        let mut disks =
+            collection.sled_agents.iter().flat_map(|(sled_id, sled_agent)| {
+                std::iter::repeat(sled_id).zip(sled_agent.disks.iter())
+            });

         // Partition the sled agents into those with an associated baseboard id
         // and those without one.  We handle these pretty differently.
@@ -656,10 +651,18 @@ impl DataStore {
                 use db::schema::inv_physical_disk::dsl;

                 let batch_size = SQL_BATCH_SIZE.get().try_into().unwrap();
-                let mut disks = disks.into_iter();
                 loop {
-                    let some_disks =
-                        disks.by_ref().take(batch_size).collect::<Vec<_>>();
+                    let some_disks = disks
+                        .by_ref()
+                        .take(batch_size)
+                        .map(|(&sled_id, disk)| {
+                            InvPhysicalDisk::new(
+                                collection_id,
+                                sled_id,
+                                disk.clone(),
+                            )
+                        })
+                        .collect::<Vec<_>>();
                     if some_disks.is_empty() {
                         break;
                     }
```

I don't know that either of these are worth saving the intermediate `.collect()`, but at least we (kind of) understand it.

schema/rss-service-plan-v3.json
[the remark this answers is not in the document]
jgallagher: But those bits of code are still referencing the "rss-service-plan-v2.json" path, right? Should we bump that to v3? (I don't see anywhere where the schema json is checked, either? Sorry if I'm just missing it.)

smklein:
Okay, I ran:
- omdb nexus background-tasks show
Parsed that output to find the `last collection id`, and used that to run
- omdb -w nexus blueprints generate-from-collection 
I then waited for inventory to be collected again, and re-ran those commands, before running
- omdb db reconfigurator-save complex.json
- I then copied that file from my switch zone to my global zone, then added my ssh keys from my "hosting" illumos box to `~/.ssh/authorized_keys` on the "target" illumos box via the serial console, then I copied `complex.json` out to my host machine, and re-ran the complex test
- That failed, but re-running with EXPECTORATE=overwrite caused it to succeed

schema/rss-service-plan-v3.json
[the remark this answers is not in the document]
smklein: Ugh, thank you, good catch, I'm using this in https://github.com/oxidecomputer/omicron/pull/5172 , but not here yet. I'll pull this into the appropriate PR.

nexus/db-queries/src/db/datastore/inventory.rs
sunshowers: Yeah I ran into this yesterday. It's a pretty bad bug in rustc.

PR #5158, opened 2024-02-28
sled-agent needs to deal with SSD overprovisioning

### LINKED ISSUE #4619: sled-agent needs to deal with SSD overprovisioning
With upcoming hardware BOM changes, sled-agent is going to need to take care of dealing with SSD overprovisioning and namespace reformatting changes prior to setting up and using the SSDs for any particular purpose. In particular, for the WDC SN655 devices, we will need to issue the vendor-specific resize command to the devices. This process erases all data on the device, so we'll need to think through the lifecycle carefully.

While this is focused on the U.2s, this is a more general problem in sled-agent where there are certain things we need to be doing on a per-model basis for all storage. For example, for the M.2s, right now we are relying on the factory to partition them and to reformat the namespace. For the SN840s, we have ended up with 512-byte namespaces instead of 4 KiB (though unclear how much of a difference that strictly makes or not).

Note, this all must occur before or as part of making these items into `ParsedDisks` or the M.2 equivalent. However, we have to be careful. As the original SN840s were not formatted to have 4 KiB namespaces, if they have data on them, we **must not** perform this transition. That may mean it only makes sense to look at the SN855 or field replacements for the Micron 7300/7450 M.2s.

To support this technologically, we're working through [illumos IPD 43 NVMe 2.0, libnvme, and the nvme(4D) ioctl interface](https://github.com/illumos/ipd/blob/master/ipd/0043/README.adoc), which will provide a programmatic libnvme interface, which while still private to illumos, sled agent will be able to consume to perform these operations.

papertigers:
This introduces the ability for omicron's storage manager to resize and format NVMe disks based upon a lookup table keyed off of model number to preferred settings. 

Fixes #4619

### DIFF
--- a/sled-hardware/src/illumos/partitions.rs
+++ b/sled-hardware/src/illumos/partitions.rs
@@ -256,7 +256,7 @@ fn ensure_size_and_formatting(
         preferred_nvme_device_settings().get(identity.model.as_str())
     {
         let nvme = Nvme::new()?;
-        for controller in nvme.controller_discovery()?.into_iter() {
+        for controller in nvme.controller_discovery()? {
             let controller = controller?.write_lock().map_err(|(_, e)| e)?;
             let controller_info = controller.get_info()?;
             // Make sure we are operating on the correct NVMe device.
@@ -304,7 +304,6 @@ fn ensure_size_and_formatting(
                 .unwrap_or(DEFAULT_NVME_LBA_DATA_SIZE);
             let desired_lba = controller_info
                 .lba_formats()
-                .into_iter()
                 .collect::<Result<Vec<_>, _>>()?
                 .into_iter()
                 .find(|lba| {
--- a/sled-hardware/src/illumos/partitions.rs
+++ b/sled-hardware/src/illumos/partitions.rs
@@ -21,14 +21,18 @@ use illumos_utils::zpool::MockZpool as Zpool;
 #[cfg(not(test))]
 use illumos_utils::zpool::Zpool;
 
-// We always want the device to be formatted with 4k data and 0 byte metadata.
-// In the future if we want to override these values we can introduce an
-// optional value in `NvmeDeviceSettings`.
-static DEFAULT_NVME_LBA_META_SIZE: u32 = 0;
+/// NVMe devices use a meta size of 0 as we don't support writing addditional
+/// metadata
+static NVME_LBA_META_SIZE: u32 = 0;
+/// NVMe devices default to using 4k logical block addressing unless overriden.
 static DEFAULT_NVME_LBA_DATA_SIZE: u64 = 4096;
 
+/// NVMe device settings for a particular NVMe model.
 struct NvmeDeviceSettings {
-    size: u32,
+    /// The desired disk size for dealing with overprovisioning.
+    size: Option<u32>,
+    /// An override for the default 4k LBA formatting.
+    lba_data_size_override: Option<u64>,
 }
 
 static PREFERRED_NVME_DEVICE_SETTINGS: OnceLock<
@@ -39,11 +43,40 @@ fn preferred_nvme_device_settings(
 ) -> &'static HashMap<&'static str, NvmeDeviceSettings> {
     PREFERRED_NVME_DEVICE_SETTINGS.get_or_init(|| {
         HashMap::from([
-            ("WUS4C6432DSP3X3", NvmeDeviceSettings { size: 3200 }),
-            ("WUS5EA138ESP7E1", NvmeDeviceSettings { size: 3200 }),
-            ("WUS5EA138ESP7E3", NvmeDeviceSettings { size: 3200 }),
-            ("WUS5EA176ESP7E1", NvmeDeviceSettings { size: 6400 }),
-            ("WUS5EA176ESP7E3", NvmeDeviceSettings { size: 6400 }),
+            // This disk ships with a size of 3200 from the factory, but we
+            // still wish to apply 4K LBA formatting.
+            (
+                "WUS4C6432DSP3X3",
+                NvmeDeviceSettings { size: None, lba_data_size_override: None },
+            ),
+            (
+                "WUS5EA138ESP7E1",
+                NvmeDeviceSettings {
+                    size: Some(3200),
+                    lba_data_size_override: None,
+                },
+            ),
+            (
+                "WUS5EA138ESP7E3",
+                NvmeDeviceSettings {
+                    size: Some(3200),
+                    lba_data_size_override: None,
+                },
+            ),
+            (
+                "WUS5EA176ESP7E1",
+                NvmeDeviceSettings {
+                    size: Some(6400),
+                    lba_data_size_override: None,
+                },
+            ),
+            (
+                "WUS5EA176ESP7E3",
+                NvmeDeviceSettings {
+                    size: Some(6400),
+                    lba_data_size_override: None,
+                },
+            ),
         ])
     })
 }
@@ -252,26 +285,30 @@ fn ensure_size_and_formatting(
 
             // Resize the device if needed to ensure we get the expected
             // durability level in terms of drive writes per day.
-            if size != nvme_settings.size {
-                // TODO this also needs to be abstracted away.
-                controller.wdc_resize_set(nvme_settings.size)?;
-                info!(
-                    log,
-                    "Resized {} from {size} to {}",
-                    identity.serial,
-                    nvme_settings.size
-                )
+            if let Some(wanted_size) = nvme_settings.size {
+                if size != wanted_size {
+                    // TODO this also needs to be abstracted away.
+                    controller.wdc_resize_set(wanted_size)?;
+                    info!(
+                        log,
+                        "Resized {} from {size} to {wanted_size}",
+                        identity.serial,
+                    )
+                }
             }
 
             // Find the LBA format we want to use for the device.
+            let wanted_data_size = nvme_settings
+                .lba_data_size_override
+                .unwrap_or(DEFAULT_NVME_LBA_DATA_SIZE);
             let desired_lba = controller_info
                 .lba_formats()
                 .into_iter()
                 .collect::<Result<Vec<_>, _>>()?
                 .into_iter()
                 .find(|lba| {
-                    lba.meta_size() == DEFAULT_NVME_LBA_META_SIZE
-                        && lba.data_size() == DEFAULT_NVME_LBA_DATA_SIZE
+                    lba.meta_size() == NVME_LBA_META_SIZE
+                        && lba.data_size() == wanted_data_size
                 })
                 .ok_or_else(|| NvmeFormattingError::LbaFormatMissing)?;
 
@@ -291,9 +328,8 @@ fn ensure_size_and_formatting(
 
                 info!(
                     log,
-                    "Formatted {} using LBA with data size of {}",
+                    "Formatted {} using LBA with data size of {wanted_data_size}",
                     identity.serial,
-                    DEFAULT_NVME_LBA_DATA_SIZE
                 );
             }
 
--- a/sled-hardware/src/illumos/partitions.rs
+++ b/sled-hardware/src/illumos/partitions.rs
@@ -21,36 +21,29 @@ use illumos_utils::zpool::MockZpool as Zpool;
 #[cfg(not(test))]
 use illumos_utils::zpool::Zpool;
 
+// We always want the device to be formatted with 4k data and 0 byte metadata.
+// In the future if we want to override these values we can introduce an
+// optional value in `NvmeDeviceSettings`.
+static DEFAULT_NVME_LBA_META_SIZE: u32 = 0;
+static DEFAULT_NVME_LBA_DATA_SIZE: u64 = 4096;
+
 struct NvmeDeviceSettings {
     size: u32,
-    lba_meta_size: u32,
-    lba_data_size: u64,
 }
 
-static PREFERRED_NVME_DEVICE_FORMATS: OnceLock<
+static PREFERRED_NVME_DEVICE_SETTINGS: OnceLock<
     HashMap<&'static str, NvmeDeviceSettings>,
 > = OnceLock::new();
 
 fn preferred_nvme_device_settings(
 ) -> &'static HashMap<&'static str, NvmeDeviceSettings> {
-    PREFERRED_NVME_DEVICE_FORMATS.get_or_init(|| {
+    PREFERRED_NVME_DEVICE_SETTINGS.get_or_init(|| {
         HashMap::from([
-            (
-                "WUS4C6432DSP3X3",
-                NvmeDeviceSettings {
-                    size: 3200,
-                    lba_meta_size: 0,
-                    lba_data_size: 512,
-                },
-            ),
-            (
-                "NEWMODELHERE",
-                NvmeDeviceSettings {
-                    size: 3200,
-                    lba_meta_size: 0,
-                    lba_data_size: 512,
-                },
-            ),
+            ("WUS4C6432DSP3X3", NvmeDeviceSettings { size: 3200 }),
+            ("WUS5EA138ESP7E1", NvmeDeviceSettings { size: 3200 }),
+            ("WUS5EA138ESP7E3", NvmeDeviceSettings { size: 3200 }),
+            ("WUS5EA176ESP7E1", NvmeDeviceSettings { size: 6400 }),
+            ("WUS5EA176ESP7E3", NvmeDeviceSettings { size: 6400 }),
         ])
     })
 }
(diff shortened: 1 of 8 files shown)


sled-hardware/src/illumos/partitions.rs
```diff
@@ -17,6 +21,82 @@ use illumos_utils::zpool::MockZpool as Zpool;
...
+static NVME_LBA_META_SIZE: u32 = 0;
+/// NVMe devices default to using 4k logical block addressing unless overriden.
+static DEFAULT_NVME_LBA_DATA_SIZE: u64 = 4096;
+
+/// NVMe device settings for a particular NVMe model.
+struct NvmeDeviceSettings {
+    /// The desired disk size for dealing with overprovisioning.
+    size: Option<u32>,
+    /// An override for the default 4k LBA formatting.
+    lba_data_size_override: Option<u64>,
```
papertigers, line 35: Nothing uses this today, so I am open to removing it from this PR in favor of not having dead code.

papertigers:
I have tested this on my bench gimlet and observed the proper behavior.  I will run through that test one more time and paste the results on this PR.


Currently the blocker for this is updating the buildiomat image (cc @jclulow) which means we will introduce a new omicron dependency on libnvme.

sled-hardware/src/illumos/partitions.rs
```diff
@@ -17,6 +21,70 @@ use illumos_utils::zpool::MockZpool as Zpool;
...
+
+/// A mapping from model to desired settings.
+/// A device not found in this lookup table will not be modified by sled-agent.
+static PREFERRED_NVME_DEVICE_SETTINGS: OnceLock<
+    HashMap<&'static str, NvmeDeviceSettings>,
+> = OnceLock::new();
+
+fn preferred_nvme_device_settings(
+) -> &'static HashMap<&'static str, NvmeDeviceSettings> {
+    PREFERRED_NVME_DEVICE_SETTINGS.get_or_init(|| {
```
jgallagher, line 46: Tiny nit: if having to do this dance is annoying, you could use https://docs.rs/once_cell/latest/once_cell/sync/struct.Lazy.html instead, so you could use `PREFERRED_NVME_DEVICE_SETTINGS` directly instead of having to go through this helper. We use it elsewhere, and eventually it will land in the std lib.

sled-hardware/src/illumos/partitions.rs
```diff
@@ -17,6 +21,70 @@ use illumos_utils::zpool::MockZpool as Zpool;
...
+        HashMap::from([
+            (
+                "WUS4C6432DSP3X3",
+                NvmeDeviceSettings { size: 3200, lba_data_size_override: None },
```
jgallagher, line 50: Two thoughts, neither of which are necessarily relevant to landing this PR (and are kinda half baked; maybe @smklein has more thoughts here?):

* This list being built into sled-agent means adding support for more disks requires an OS update. Is that okay?
* If being baked into sled-agent is okay, do we want to define this in code, or would it make sense to put it in the sled-agent config.toml (which would require some work to pass it down into `sled-hardware`, admittedly).

sled-hardware/src/illumos/partitions.rs
```diff
@@ -119,6 +191,11 @@ fn internal_ensure_partition_layout<GPT: gpt::LibEfiGpt>(
             };
             match variant {
                 DiskVariant::U2 => {
+                    // First we need to check that this disk is of the proper
+                    // size and correct logical block address formatting.
+                    ensure_size_and_formatting(log, identity)?;
```
jgallagher, line 196: Mostly for my own education: what does this do with disks that already have data on them?

jgallagher (commented):
LGTM, but I think I'd vote for a review from @smklein too since he's currently down in the weeds of physical disk management from Nexus.

sled-hardware/src/illumos/partitions.rs
papertigers: We don't touch a disk with a label/zpool as changing the LBA format erases all data on the disk. This should only be preformed when StorageManger finds a new disk that has not been setup yet i.e first boot, add disk etc

sled-hardware/src/illumos/partitions.rs
andrewjstone: FWIW, I much prefer the nexus path laid out here to using the toml file. Thanks for the details @rmustacc.

sled-hardware/src/illumos/partitions.rs
rmustacc: And I expect that when we transform to an explicit adoption model then we'll want this to not occur until that adoption process is initiated.

andrewjstone: This only gets run if there is no GPT table. It's hidden by github above the diff.

sled-hardware/src/illumos/partitions.rs
```diff
@@ -154,13 +231,124 @@ fn internal_ensure_partition_layout<GPT: gpt::LibEfiGpt>(
     }
 }
 
+fn ensure_size_and_formatting(
```
andrewjstone, line 234: Is there any problem running this over after failure on a reboot or restart of sled-agent?

andrewjstone (approved):
@papertigers Nice work. This looks good to me, although I'm mostly trusting you (and @rmustacc) on how the actual NVME stuff works. I think that's probably a safe bet :)

As long as the answer to my question is "it's fine", this LGTM.

I don't think this will affect what @smklein is doing as I believe this is at a much lower level. But it would be good to get a quick check.

papertigers:
Just wanted to dump some testing notes.


## Setup

Two of the new disks that Robert added to my bench gimlet for me.
```
BRM42220027 # diskinfo | grep -vE 'Micron|WUS4C6'
TYPE    DISK                    VID      PID              SIZE          RMV SSD
NVME    c2t0014EE84015DD700d0   NVMe     WUS5EA138ESP7E3  3576.98 GiB   no  yes
NVME    c4t0014EE84015DD100d0   NVMe     WUS5EA138ESP7E3  3576.98 GiB   no  yes
```

I looped over them and configured them how we will likely see them in the wild from the manufacture:
```
BRM42220027 # for i in $(nvmeadm list | grep WUS5E | awk 'BEGIN {FS=":"}; {print $1}'); do controller="$i/1";  echo resetting "$controller";  nvmeadm detach "$controller";  nvmeadm wdc/resize -s 3840 "$controller"; nvmeadm format "$controller" 0; nvmeadm attach "$controller"; done
resetting nvme1/1
nvme1/1 resized to 3840 GB
resetting nvme3/1
nvme3/1 resized to 3840 GB
```

Where LBA 0 looks like:
```
    LBA Format 0
      Metadata Size:                        0 bytes
      LBA Data Size:                        512 bytes
      Relative Performance:                 Best
```

While I am here let us reset one of the deployed `WUS4C6432DSP3X3` drives with a 512 LBA to match what's in the field today:
```
BRM42220027 # nvmeadm detach nvme10/1
BRM42220027 # nvmeadm format nvme10/1 0
BRM42220027 # nvmeadm attach nvme10/1
```

Where LBA looks like:
```
    LBA Format 0
      Metadata Size:                        0 bytes
      LBA Data Size:                        512 bytes
      Relative Performance:                 Best
```

## 1st pass sled-agent

When we install omicron for the first time we expect to see the `WUS4C6432DSP3X3` device update to use a 4K LBA and  the two `WUS5EA138ESP7E3` devices should resize from 3840 down to 3200 and also be formatted using a 4K logical block address.

Install:
```
BRM42220027 # ./omicron-package -t bench install
```

Check sled-agent's logs:
```
BRM42220027 # cat $(svcs -L sled-agent) | grep -iE 'formatted|resized' | looker
02:56:51.459Z INFO SledAgent (StorageManager): Formatted disk with serial A079E3AF to an LBA with data size 4096
    file = sled-hardware/src/illumos/partitions.rs:316
02:56:57.009Z INFO SledAgent (StorageManager): Resized 23181L900156 from 3840 to 3200
    file = sled-hardware/src/illumos/partitions.rs:280
02:57:01.361Z INFO SledAgent (StorageManager): Formatted disk with serial 23181L900156 to an LBA with data size 4096
    file = sled-hardware/src/illumos/partitions.rs:316
02:57:09.718Z INFO SledAgent (StorageManager): Resized 23181L900102 from 3840 to 3200
    file = sled-hardware/src/illumos/partitions.rs:280
02:57:14.063Z INFO SledAgent (StorageManager): Formatted disk with serial 23181L900102 to an LBA with data size 4096
    file = sled-hardware/src/illumos/partitions.rs:316
```

## 2nd pass sled-agent

Remove omicron but don't modify the disks

```
BRM42220027 # ./omicron-package -t bench uninstall
Logging to: /tmp/omicron/out/LOG
About to delete the following datasets: [
    "oxi_b6014b47-c0cc-481f-8bdb-bbd543071616/cluster",
    "oxi_b6014b47-c0cc-481f-8bdb-bbd543071616/config",
    "oxi_b6014b47-c0cc-481f-8bdb-bbd543071616/debug",
    "oxi_b6014b47-c0cc-481f-8bdb-bbd543071616/install",
    "oxi_f369eded-643f-4a47-857f-e44c2a973c1f/cluster",
    "oxi_f369eded-643f-4a47-857f-e44c2a973c1f/config",
    "oxi_f369eded-643f-4a47-857f-e44c2a973c1f/debug",
    "oxi_f369eded-643f-4a47-857f-e44c2a973c1f/install",
    "oxp_1db0d263-c5e7-46ef-bf81-015f3a4af100/crypt",
    "oxp_22aa8eef-15f4-4769-8503-d25ede198c7d/crypt",
    "oxp_24df79f9-2aed-4b7a-93f8-047fbb1f2514/crypt",
    "oxp_410bc7c9-a23d-4a50-b537-9a80f4319b50/crypt",
    "oxp_4cd8e627-3a0a-4b35-a287-9ef51f91ec78/crypt",
    "oxp_67f9b83a-7ccb-4e50-b5ad-309128bdf014/crypt",
    "oxp_cfdc3eb7-6f02-410c-8daa-346868387c7e/crypt",
    "oxp_d17de75f-c329-417e-9cb1-10fc5d17ab14/crypt",
    "oxp_ed1efcc5-0c03-4b7b-9fbc-b5598ab7ea7a/crypt",
]
[yY to confirm] >> y
BRM42220027 # rm /var/svc/log/oxide-sled-agent:default.log
BRM42220027 # svcadm disable fmd
BRM42220027 # swap -l
swapfile             dev    swaplo   blocks     free
/dev/zvol/dsk/oxi_f369eded-643f-4a47-857f-e44c2a973c1f/swap 170,4         8 536870904 536870904
BRM42220027 # swap -d /dev/zvol/dsk/oxi_f369eded-643f-4a47-857f-e44c2a973c1f/swap
BRM42220027 # for zpool in $(zpool list -Hp | grep ox | awk '{print $1}'); do         zpool destroy "$zpool"; done
BRM42220027 # zpool list
NAME    SIZE  ALLOC   FREE  CKPOINT  EXPANDSZ   FRAG    CAP  DEDUP    HEALTH  ALTROOT
rpool  3.94G  1.31G  2.62G        -         -     5%    33%  1.00x    ONLINE  -
BRM42220027 # svcs sled-agent
svcs: Pattern 'sled-agent' doesn't match any instances
STATE          STIME    FMRI
```

Install omicron again and verify none of the disks are touched outside of zpool creation.

```
BRM42220027 # ./omicron-package -t bench install
... wait a bit ...
BRM42220027 # cat $(svcs -L sled-agent) | grep -iE 'formatted|resized' | looker
BRM42220027 #
BRM42220027 # zpool list
NAME                                       SIZE  ALLOC   FREE  CKPOINT  EXPANDSZ   FRAG    CAP  DEDUP    HEALTH  ALTROOT
oxi_61762044-15db-411c-8ffc-df3b9ea8eb7c   748G  1.43M   748G        -         -     0%     0%  1.00x    ONLINE  -
oxi_ebd965ff-a438-45cb-8f1b-cd6f0527a602   748G  2.36M   748G        -         -     0%     0%  1.00x    ONLINE  -
oxp_3c7bf7b8-1b8c-4d78-b478-35eb2097e8cf  2.91T  1.39M  2.91T        -         -     0%     0%  1.00x    ONLINE  -
oxp_40c404c2-5429-4b0f-88db-a99e5d0a923d  2.91T  1.42M  2.91T        -         -     0%     0%  1.00x    ONLINE  -
oxp_42f191fd-aa0e-43b7-b1bc-a960db7dca59  2.91T  1.42M  2.91T        -         -     0%     0%  1.00x    ONLINE  -
oxp_44972f33-648d-4b90-8e02-e1b94aea20a9  2.91T  1.42M  2.91T        -         -     0%     0%  1.00x    ONLINE  -
oxp_71cbbfee-4ea9-4a15-9acb-4015a156ce47  2.91T  1.70M  2.91T        -         -     0%     0%  1.00x    ONLINE  -
oxp_797f7bcf-387f-4f44-ac04-6396030f9e8a  2.91T  1.39M  2.91T        -         -     0%     0%  1.00x    ONLINE  -
oxp_91c965c6-1e7b-4c72-853c-81a806100ca0  2.91T  1.66M  2.91T        -         -     0%     0%  1.00x    ONLINE  -
oxp_a2a5efee-d300-4e05-8696-93fcbfbe68f5  2.91T  1.67M  2.91T        -         -     0%     0%  1.00x    ONLINE  -
oxp_b70628e2-86cf-4302-a64d-f20e46f06292  2.91T  1.40M  2.91T        -         -     0%     0%  1.00x    ONLINE  -
rpool                                     3.94G  1.31G  2.62G        -         -     5%    33%  1.00x    ONLINE  -
```

sled-hardware/src/illumos/partitions.rs
papertigers: I don't feel strongly about this one. I am okay leaving it as is if that works for you.

sled-hardware/src/illumos/partitions.rs
papertigers: Thanks for the helpful writeup @rmustacc.  What you outlined here makes sense to me and likely ties into some of the same work that's going to happen with nvme firmware upgrades. I can keep this in the back of my mind as I start to figure out what that's going to look like.

sled-hardware/src/illumos/partitions.rs
jgallagher: Yep, totally fine.

sled-hardware/src/illumos/partitions.rs
```diff
@@ -154,13 +231,124 @@ fn internal_ensure_partition_layout<GPT: gpt::LibEfiGpt>(
...
+            namespace.blkdev_attach()?;
+        }
+    } else {
+        info!(
+            log,
+            "There are no preferred NVMe settings for disk model {}; nothing to\
+             do for disk with serial {}",
+            identity.model,
+            identity.serial
+        );
```
smklein, line 334: Thanks for handling this case -- this means that all the file-backed vdevs in testing will still work.

sled-hardware/src/illumos/partitions.rs
papertigers: So while testing I found that if a disk is left in the unattached state then sled-agent will ignore it.  That's due to [these lines](https://github.com/oxidecomputer/omicron/blob/8697f39aac3323ffda3f33f4b5aa44fd7ad857e3/sled-hardware/src/illumos/mod.rs#L435-L437).  So, if we were to detach blkdev from the disk and then take a sled-agent reboot we wouldn't be notified of that disk again I believe.  

Now if we were in the process of formatting a disk and the box panicked or rebooted for some reason I don't know what would happen.  Perhaps @rmustacc could shed some light on that failure scenario.

Regardless I don't know what the best course of action is for the first failure mode, maybe we need an enhancement in how we detect disks from devinfo?

rmustacc: > Now if we were in the process of formatting a disk and the box panicked or rebooted for some reason I don't know what would happen. Perhaps @rmustacc could shed some light on that failure scenario.

There is nothing persistent about a blkdev attach/detach, so the kernel will attempt to if it can. Because we don't support namespace management and aren't doing anything here, this should likely be okay, though I can't comment on what happens if a device fails in the middle of say a format nvm command.

> Regardless I don't know what the best course of action is for the first failure mode, maybe we need an enhancement in how we detect disks from devinfo?

I would suggest we file a follow up bug that moves this to looking for the NVMe node and not the blkdev node. This is for a few reasons:

* An NVMe device may not have any blkdev instances attached as mentioned to start this.
* An NVMe device may have the wrong number of namespaces created if it came from another system. This would show up as two distinct block devices or possibly no block devices, which would likely be confusing.

test-utils/src/dev/mod.rs
```diff
@@ -145,3 +146,13 @@ pub fn process_running(pid: u32) -> bool {
     // only checks whether the process is running.
     0 == (unsafe { libc::kill(pid as libc::pid_t, 0) })
 }
+
+/// Returns a DiskIdentity that can be passed to ensure_partition_layout when
+/// not operating on a real disk.
+pub fn mock_device_identity() -> DiskIdentity {
```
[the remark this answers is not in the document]
papertigers: Fixed in ea09e03

PR #5162, opened 2024-02-28
env.sh: preer `readlink` to `cd && echo ${PWD}`

dancrossnyc:
A trivial change to make this script more robust for those of us who alias `cd` to something.


rcgoodfellow (approved):

jclulow:
Perhaps just use `builtin cd` to avoid alias expansion?

dancrossnyc:
> Perhaps just use `builtin cd` to avoid alias expansion?

I considered that, as well as using the `chdir` builtin, that's unlikely to be aliased. But I couldn't really see an advantage over `readlink`, which seems to be readily available pretty much everywhere. What's the advantage?

jclulow:
It requires a fork, and it depends on GNU coreutils I believe; we don't have a readlink in the base tools.

In this case I suppose it doesn't matter hugely, but the pattern used here is one we use in a lot of bash scripts all over the place -- so as long as we don't start changing all the other ones.

dancrossnyc:
The forks are the same (the subcommand for `$()` is already run in a child process, of course), but it does add an `exec`. Yeah, I wondered about the coreutils thing.

PR #5163, opened 2024-02-28
rename reconfigurator packages

davepacheco:
As discussed in oxide-update chat (though there's not much context there -- just this better describes what they do and reflects the relationship between these two packages).


sunshowers (approved):
Love it!

Now or as a followup, do you think it would be worth adding a readme to nexus/reconfigurator with a brief description? Even just a link to the RFDs, since reconfigurator feels like a name that may be slightly confusing to newcomers (but makes a ton of sense once you read the RFDs).

davepacheco:
Good idea!  For whatever reason I never expect to find READMEs in non-root directories but it's a good idea.  We can definitely iterate on the contents.

PR #5168, opened 2024-02-29
NTP self assembling zone

karencfv:


citrus-it:
For the DNS configuration, definitely take a look at https://github.com/oxidecomputer/omicron/blob/main/sled-agent/src/services.rs#L1396-L1429 which does this via the existing `svc:/network/dns/install:default` (part of illumos). It probably needs tweaking to handle multiple DNS servers, but it should be the right way to configure DNS in the zone.

karencfv:
Thanks @citrus-it! I'll take a look

karencfv:
I've tested manually on a Helios box and all services are up and running. `/etc/resolv.conf` is populated correctly.

```console
$ tail -n 1000000 $(svcs -L sled-agent) | grep "Profile for oxz_ntp" | looker
04:16:09.669Z INFO SledAgent (ServiceManager): Profile for oxz_ntp_a6d6b18d-aff0-47db-8bb6-9dac0b6e66a4:
    <!DOCTYPE service_bundle SYSTEM "/usr/share/lib/xml/dtd/service_bundle.dtd.1">
    <service_bundle type="profile" name="omicron">
      <service version="1" type="service" name="oxide/zone-network-setup">
          <property_group type="application" name="config">
            <propval type="astring" name="datalink" value="oxControlService4"/>
            <propval type="astring" name="gateway" value="fd00:1122:3344:101::1"/>
            <propval type="astring" name="static_addr" value="fd00:1122:3344:101::1b"/>
          </property_group>
        <instance enabled="true" name="default">
        </instance>
      </service>
      <service version="1" type="service" name="network/ssh">
        <instance enabled="false" name="default">
        </instance>
      </service>
      <service version="1" type="service" name="network/dns/install">
          <property_group type="application" name="install_props">
            <property type="net_address" name="nameserver">
              <net_address_list>
                <value_node value="1.1.1.1" />
                <value_node value="9.9.9.9" />
              </net_address_list>
            </property>
          </property_group>
        <instance enabled="true" name="default">
        </instance>
      </service>
      <service version="1" type="service" name="network/dns/client">
        <instance enabled="true" name="default">
        </instance>
      </service>
      <service version="1" type="service" name="oxide/ntp">
        <instance enabled="true" name="default">
          <property_group type="application" name="config">
            <propval type="astring" name="allow" value="fd00:1122:3344:100::/56"/>
            <propval type="boolean" name="boundary" value="true"/>
            <propval type="astring" name="server" value="0.pool.ntp.org"/>
          </property_group>
        </instance>
      </service>
      <service version="1" type="service" name="oxide/opte-interface-setup">
          <property_group type="application" name="config">
            <propval type="astring" name="interface" value="vopte0"/>
            <propval type="astring" name="gateway" value="172.30.3.1"/>
            <propval type="astring" name="ip" value="172.30.3.5"/>
          </property_group>
        <instance enabled="true" name="default">
        </instance>
      </service>
    </service_bundle>
    file = sled-agent/src/profile.rs:34
```

```console
root@oxz_ntp_a6d6b18d:~# svcs -a | grep dns
disabled        4:16:11 svc:/network/dns/multicast:default
disabled        4:16:14 svc:/network/dns/install:default
online          4:16:14 svc:/network/dns/client:default
root@oxz_ntp_a6d6b18d:~# svcs -a | grep oxide
online          4:16:15 svc:/oxide/zone-network-setup:default
online          4:16:15 svc:/oxide/ntp:default
online          4:16:21 svc:/oxide/opte-interface-setup:default
root@oxz_ntp_a6d6b18d:~# cat /etc/resolv.conf 
nameserver 1.1.1.1
nameserver 9.9.9.9
```

That said, I can only deploy single sled packages, which means I have not been able to manually test Internal NTP zones.

@smklein Should this be tested directly on dogfood or is testing somewhere like madrid enough? I'm not sure about the availability of madrid :/

I've still got the unit tests to get sorted, but figured I'd set this as ready to review for now

sled-agent/src/services.rs
```diff
@@ -1937,6 +1910,121 @@ impl ServiceManager {
...
+                if dns_servers.is_empty() {
+                    dns_client_service =
+                        ServiceBuilder::new("network/dns/client").add_instance(
+                            ServiceInstanceBuilder::new("default").disable(),
+                        );
+                } else {
+                    dns_client_service = ServiceBuilder::new(
+                        "network/dns/client",
+                    )
+                    .add_instance(ServiceInstanceBuilder::new("default"));
```
citrus-it, line 1996: I think you should move this up and make it more generic. Any zone that configures DNS via `dns/install` should also enable `dns/client` (cockroach zones aren't currently doing that), and any zone that does not configure DNS should disable `dns/client`.

This is similar in concept to disabling ssh, which we should do in all zones apart from the switch zone.
I suggest declaring a disabled `dns_client_service` at the same place as `disabled_ssh_service`
```rust
        let dns_client_service = ServiceBuilder::new("network/dns/client")
            .add_instance(ServiceInstanceBuilder::new("default").disable());
```
and have all zones use `dns_client_service` as part of their profile, but call `dns_client_service.enable()` 
 (new method) for any zone that also uses a `dns_install_service`.

Hopefully that makes sense.
The `network/dns/client` service itself doesn't actually do anything apart from serve as a dependency for the name services milestone, and so it's useful as a dependency for things like ntp - see comment elsewhere.

smf/ntp/manifest/manifest.xml
```diff
@@ -1,105 +1,93 @@
...
+  </dependency>
+
+  <dependency name="filesystem" grouping="require_all" restart_on="error"
+    type="service">
+  <service_fmri value="svc:/system/filesystem/minimal" />
+  </dependency>
+
+  <dependency name="name-services" grouping="optional_all" restart_on="none"
+    type="service">
+  <service_fmri value="svc:/milestone/name-services" />
```
citrus-it, line 29: Note that we have a dependency on `milestone/name-services`, which means we wait until the `network/dns/client` service is online. `network/dns/client` depends on both `dns/install` and the presence of `/etc/resolv.conf`, so everything works nicely and we'll block waiting for those.

I'd change `optional_all` here to `require_all` to make this more robust.

sled-agent/src/services.rs
```diff
@@ -1937,6 +1910,121 @@ impl ServiceManager {
...
+                    let opte_interface_setup =
+                        Self::opte_interface_set_up_install(&installed_zone)?;
+                    profile = profile.add_service(opte_interface_setup)
+                }
+
+                profile
+                    .add_to_zone(&self.inner.log, &installed_zone)
+                    .await
+                    .map_err(|err| {
+                        Error::io("Failed to setup NTP profile", err)
```
citrus-it, line 2023: nit: "set up"

citrus-it (commented):
This ended up being a lot less invasive than I expected, which is a good indication that the framework is right.

I had a small number of comments. Some of these are on pre-existing bits which you are just moving around, but it's a good opportunity to firm some things up.

smf/ntp/manifest/manifest.xml
```diff
@@ -1,105 +1,93 @@
...
+  </dependency>
+
+  <dependency name='zone_network_setup' grouping='require_all' restart_on='none'
+    type='service'>
+  <service_fmri value='svc:/oxide/zone-network-setup:default' />
+  </dependency>
+
+  <!-- 
+    Only Boundary NTP needs the opte interface. This means we cannot make the
+    'oxide/opte_interface_setup' service a dependency here. The service is
```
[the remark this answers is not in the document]
karencfv: ah gotcha, this is great to know thanks!

sled-agent/src/services.rs
karencfv: :woman_facepalming:

karencfv:
Thanks @citrus-it ! I think I've addressed your comments and here are the updated self assembling zones:

```console
root@oxz_cockroachdb_dedb2b05:~# svcs -a | grep dns
disabled        3:28:53 svc:/network/dns/multicast:default
disabled        3:28:57 svc:/network/dns/install:default
online          3:28:56 svc:/network/dns/client:default

root@oxz_crucible_0f35b386:~# svcs -a | grep dns
disabled        3:29:53 svc:/network/dns/install:default
disabled        3:29:54 svc:/network/dns/multicast:default
disabled        3:30:08 svc:/network/dns/client:default

root@oxz_oximeter_7a6f2330:~# svcs -a | grep dns
disabled        3:29:54 svc:/network/dns/install:default
disabled        3:29:56 svc:/network/dns/multicast:default
disabled        3:30:08 svc:/network/dns/client:default

root@oxz_crucible_pantry_271b260d:~# svcs -a | grep dns
disabled        3:30:01 svc:/network/dns/install:default
disabled        3:30:03 svc:/network/dns/multicast:default
disabled        3:30:12 svc:/network/dns/client:default

root@oxz_external_dns_5d69e2a7:~# svcs -a | grep dns
disabled        3:30:02 svc:/network/dns/install:default
disabled        3:30:02 svc:/network/dns/multicast:default
disabled        3:30:13 svc:/network/dns/client:default
online          3:30:19 svc:/oxide/external_dns:default

root@oxz_ntp_9c8cb92e:~# svcs -a | grep dns
disabled        3:28:18 svc:/network/dns/multicast:default
disabled        3:28:21 svc:/network/dns/install:default
online          3:28:21 svc:/network/dns/client:default

root@oxz_clickhouse_405b48f6:~# svcs -a | grep dns
disabled        3:30:10 svc:/network/dns/multicast:default
disabled        3:30:29 svc:/network/dns/install:default
online          3:30:29 svc:/network/dns/client:default
```

> This ended up being a lot less invasive than I expected, which is a good indication that the framework is right.

Yeah! I was very pleasantly surprised. I'll probably extract the [svc-site-ntp script](https://github.com/oxidecomputer/omicron/blob/main/smf/ntp/method/svc-site-ntp) into zone-network cli commands in a follow up PR though. I want to remove as many scripts as possible :)

citrus-it (approved):
Thanks for making those changes. Once whatever is going on with the tests is sorted out, looks good to me.

sled-agent/src/profile.rs
```diff
@@ -35,6 +35,13 @@ impl ProfileBuilder {
 
         let profile_path = installed_zone.site_profile_xml_path();
 
+        // Create the xml profile directory if it doesn't already exist.
+        // All self assembled zones already have this directory;
+        // this is mostly for our testing.
+        let mut path = profile_path.clone();
+        path.pop();
+        tokio::fs::create_dir_all(path).await?;
+
```
karencfv, line 44: I don't like this at all, but even though [`fake_install()`](https://github.com/oxidecomputer/omicron/blob/main/illumos-utils/src/running_zone.rs#L1301-L1304) creates the directory, the tests kept reporting the directory didn't exist :(

@smklein and/or @citrus-it , if you have a better suggestion on how to go on about this I'd love to hear it!

karencfv:
Gargh, now the job is failing with `files in /var/tmp/omicron_tmp (none expected on success):` (≖_≖ )

karencfv:
Hmmm... Looks like the tests are creating a temp directory, and `fake_install()` is creating another

sled-agent/src/profile.rs
karencfv: Never mind, fixed in https://github.com/oxidecomputer/omicron/pull/5168/commits/6718070742f09b9a5c343c20e6a48e05f32916b5 :)

karencfv:
This should work :crossed_fingers:

PR #5176, opened 2024-03-01
Enable Updates for Floating IPs

### LINKED ISSUE #5016: Need endpoint to update a floating IP
Noticed this small oversight while talking through the floating IPs UI. There is no update endpoint.

https://github.com/oxidecomputer/omicron/blob/f2f2bfae145f91486ac0444fa17aac51278fdecc/nexus/tests/output/nexus_tags.txt#L13-L20

I don't think you can change the pool or IP address (just make a new one) but you should be able to update the name and description.

https://github.com/oxidecomputer/omicron/blob/f2f2bfae145f91486ac0444fa17aac51278fdecc/nexus/types/src/external_api/params.rs#L877-L891

charliepark:
Fixes #5016 

This PR adds the ability to update names and descriptions for floating IPs.


nexus/tests/integration_tests/external_ips.rs
```diff
@@ -381,6 +382,50 @@ async fn test_floating_ip_create_name_in_use(
...
+    };
+
+    let new_fip =
+        NexusRequest::object_put(client, &floating_ip_url, Some(&updates))
+            .authn_as(AuthnMode::PrivilegedUser)
+            .execute()
+            .await
+            .expect("")
+            .parsed_body::<FloatingIp>()
+            .unwrap();
```
david-crespo, line 420: `object_put` is a little less noisy

https://github.com/oxidecomputer/omicron/blob/b993768e4d62515f33a058e4082dd7a721c9cee1/nexus/tests/integration_tests/ip_pools.rs#L157-L158

nexus/tests/integration_tests/external_ips.rs
```diff
@@ -381,6 +382,50 @@ async fn test_floating_ip_create_name_in_use(
...
+
+    assert_eq!(new_fip.identity.name.as_str(), new_fip_name);
+    assert_eq!(new_fip.identity.description, new_fip_desc);
+    assert_eq!(new_fip.project_id, project.identity.id);
+    assert_eq!(new_fip.identity.time_created, fip.identity.time_created);
+    assert_ne!(new_fip.identity.time_modified, fip.identity.time_modified);
```
david-crespo, line 426: I would also do a GET here on the floating IP and make sure what's returned also matches these asserts. Otherwise the endpoint could just be returning an updated view without writing to the DB.

nexus/src/external_api/http_entrypoints.rs
```diff
@@ -1921,6 +1922,43 @@ async fn floating_ip_create(
     apictx.external_latencies.instrument_dropshot_handler(&rqctx, handler).await
 }
 
+/// Update a Floating IP
```
david-crespo, line 1925:
```suggestion
/// Update floating IP
```

nexus/tests/integration_tests/external_ips.rs
```diff
@@ -1024,3 +1079,7 @@ async fn floating_ip_detach(
     .parsed_body()
     .unwrap()
 }
+
+fn assert_floating_ips_eq(first: &FloatingIp, second: &FloatingIp) {
+    assert_eq!(first.identity, second.identity);
+}
```
david-crespo, line 1085: Nitpick, but I think this could be inlined. It's not doing enough — I expected it to compare all fields or something.

david-crespo (approved):
woohoo!

FelixMcFelix (approved):
Looks good, thanks for adding this in! 😄

PR #5186, opened 2024-03-03
Add a show-empty option to oxlog

leftwo:
This adds an option flag (off by default) that will show matching log files even if they are
zero length.

By default, if the log file is empty, you won't see any output:
```
root@EVT22200005:/alan# /alan/bin/oxlog logs --current oxz_crucible_b2136649-3b43-4bd9-827d-127fe5fe757e
root@EVT22200005:/alan#
```

But, sometimes you want really want to know there is a log file that is indeed empty.
If so, add the `--show-empty` option, and volia, you can see the empty files:
```
root@EVT22200005:/alan# /alan/bin/oxlog logs --current --show-empty  oxz_crucible_b2136649-3b43-4bd9-827d-127fe5fe757e
/pool/ext/616b26df-e62a-4c68-b506-f4a923d8aaf7/crypt/zone/oxz_crucible_b2136649-3b43-4bd9-827d-127fe5fe757e/root/var/svc/log/oxide-crucible-agent:default.log
/pool/ext/616b26df-e62a-4c68-b506-f4a923d8aaf7/crypt/zone/oxz_crucible_b2136649-3b43-4bd9-827d-127fe5fe757e/root/var/svc/log/oxide-crucible-downstairs:downstairs-7b0f29d1-642c-4c60-bbc2-eddb6ace34a9.log
/pool/ext/616b26df-e62a-4c68-b506-f4a923d8aaf7/crypt/zone/oxz_crucible_b2136649-3b43-4bd9-827d-127fe5fe757e/root/var/svc/log/oxide-zone-network-setup:default.log
root@EVT22200005:/alan# ls -l /pool/ext/616b26df-e62a-4c68-b506-f4a923d8aaf7/crypt/zone/oxz_crucible_b2136649-3b43-4bd9-827d-127fe5fe757e/root/var/svc/log/oxide-crucible-agent:default.log
-rw-r--r--   1 root     root           0 Mar  3 01:45 /pool/ext/616b26df-e62a-4c68-b506-f4a923d8aaf7/crypt/zone/oxz_crucible_b2136649-3b43-4bd9-827d-127fe5fe757e/root/var/svc/log/oxide-crucible-agent:default.log
root@EVT22200005:/alan# ls -l /pool/ext/616b26df-e62a-4c68-b506-f4a923d8aaf7/crypt/zone/oxz_crucible_b2136649-3b43-4bd9-827d-127fe5fe757e/root/var/svc/log/oxide-crucible-downstairs:downstairs-7b0f29d1-642c-4c60-bbc2-eddb6ace34a9.log
-rw-r--r--   1 root     root           0 Mar  3 01:45 /pool/ext/616b26df-e62a-4c68-b506-f4a923d8aaf7/crypt/zone/oxz_crucible_b2136649-3b43-4bd9-827d-127fe5fe757e/root/var/svc/log/oxide-crucible-downstairs:downstairs-7b0f29d1-642c-4c60-bbc2-eddb6ace34a9.log
```

### DIFF
--- a/dev-tools/oxlog/src/lib.rs
+++ b/dev-tools/oxlog/src/lib.rs
@@ -75,6 +75,9 @@ pub struct Filter {
     /// standard paths or don't follow the naming conventions of SMF service
     /// files. e.g. `/pool/ext/e12f29b8-1ab8-431e-bc96-1c1298947980/crypt/zone/oxz_cockroachdb_8bbea076-ff60-4330-8302-383e18140ef3/root/data/logs/cockroach.log`
     pub extra: bool,
+
+    /// Show a log file even if is has zero size.
+    pub show_empty: bool,
 }
 
 /// Path and metadata about a logfile
@@ -264,17 +267,21 @@ impl Zones {
         // 'archived'. These files have not yet been migrated into the debug
         // directory.
         if filter.current || filter.archived {
-            load_svc_logs(paths.primary.clone(), &mut output);
+            load_svc_logs(
+                paths.primary.clone(),
+                &mut output,
+                filter.show_empty,
+            );
         }
 
         if filter.archived {
             for dir in paths.debug.clone() {
-                load_svc_logs(dir, &mut output);
+                load_svc_logs(dir, &mut output, filter.show_empty);
             }
         }
         if filter.extra {
             for (svc_name, dir) in paths.extra.clone() {
-                load_extra_logs(dir, svc_name, &mut output);
+                load_extra_logs(dir, svc_name, &mut output, filter.show_empty);
             }
         }
         output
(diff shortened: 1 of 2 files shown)


dev-tools/oxlog/src/lib.rs
```diff
@@ -344,7 +355,7 @@ fn load_svc_logs(dir: Utf8PathBuf, logs: &mut BTreeMap<ServiceName, SvcLogs>) {
             };
 
             logfile.read_metadata(&entry);
-            if logfile.size == Some(0) {
+            if logfile.size == Some(0) && show_empty {
```
andrewjstone, line 358: This condition is very confusing. If `show_empty` is false then we we show the file which is backwards. I believe this stems from the `ArgAction::SetFalse` for the `show_empty` option. While this makes sense on the command line, it doesn't make sense down here in these conditions. What we really want here is `if logfile.size == Some(0) && !show_empty`.

Maybe instead of using `action=ArgAction::SetFalse` for the clap argument, you use `default_value_t = false`. Then you can still pass the parameter down if it is not present, and I think it should also allow you to not have to pass `true` when you set the flag since it's a bool. 

While you could still use `action=ArgAction::SetFalse` and then flip the flag before passing it down, this also seems confusing as now the value the user passed is changing, which I dislike. Anyone who prints the args during debugging may be confused.

leftwo: Okay, I switched it around and now I think it lines up with what you suggested.
The command line behavior is still the same, but the logic and the variable names is
now consistent.

PR #5196, opened 2024-03-04
Add LLDP

Nieuwejaar:
This adds the `lldp` daemon and CLI to the switch zone and updates `sled-agent` to do the basic configuration and enablement of the service.

On its own, this change will not cause the switch zone to advertise or receive `lldp` packets, as no interfaces are configured.  That can be done using the CLI in the switch zone for testing on `dogfood` and/or the colo.  Adding runtime configuration by `nexus` will be follow-on work.

### DIFF
--- a/common/src/address.rs
+++ b/common/src/address.rs
@@ -50,6 +50,7 @@ pub const CLICKHOUSE_PORT: u16 = 8123;
 pub const CLICKHOUSE_KEEPER_PORT: u16 = 9181;
 pub const OXIMETER_PORT: u16 = 12223;
 pub const DENDRITE_PORT: u16 = 12224;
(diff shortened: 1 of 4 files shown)


citrus-it:
Depending on where the lldp CLI ends up, you may also want to update the PATH via https://github.com/oxidecomputer/omicron/blob/main/smf/profile/profile#L7

Nieuwejaar:
> Depending on where the lldp CLI ends up, you may also want to update the PATH via https://github.com/oxidecomputer/omicron/blob/main/smf/profile/profile#L7

It should already be on the PATH:
```
root@oxz_switch1:~# which lldpadm
/opt/oxide/bin/lldpadm
```

PR #5202, opened 2024-03-05
Teach `omdb db network list-eips` and `DataStore::vpc_resolve_to_sleds` about blueprints

jgallagher:
While working on #4886, I ran into a couple places that expect all services to have rows in the `service` table, but Reconfigurator does not add new rows to `service` when it spins up new services (see #4947 for rationale). This PR teaches these two users of `service` to _also_ check the current target blueprint when checking for services.

I also removed `LookupPath::service_id` and `lookup_resource! { ... Service ... }`; the only user of these was `omdb`, and I think their existence is misleading in the post-`service`-table world we're moving toward.


davepacheco:
I haven't reviewed this yet but the higher-level question came up on the update call of whether to use the blueprint, inventory, or something else to answer the question of "which sleds need to know about these firewall rules?".  I just wanted to write some of this down.

In an ideal world, we want these firewall rules to go to all the sleds that are currently running something that would use them -- i.e., running an instance or control plane zone with NIC on that VPC.  Right?  In general, that is hard to know for sure.  At the same time, we must already accept that (1) firewall changes are applied asynchronously and (2) those changes may be held up indefinitely in the face of transient failures.

The **current blueprint** reflects what control plane zones _should_ be running.  It could be wrong for this purpose in a few ways:

- It could contain zones that don't actually exist yet because they haven't been deployed yet.  Assuming it's no big deal to distribute rules to a sled that doesn't need it, this seems like a non-issue.
- It could be missing sleds running zones that _do_ exist, but shouldn't.  This presumably means they are about to be removed.  In an ideal world, I think we'd send the rules to these sleds too.  It seems conceivable this could even be a major problem -- if a firewall update is security-critical, and for some reason we don't manage to remove this zone for a while, might it be sitting there vulnerable for a while potentially without a way for the operator to know?  But per my note above, it can always happen that it takes a little while for us to propagate firewall rules everywhere that needs it so maybe this is fine.

The **inventory** reflects what control plane zones were running when we collected that inventory.  It too could be wrong for this purpose in a few ways:

- It could be missing zones that do exist because it's old.  That's not great.
- It could be missing zones that do exist because of a transient collection problem.  (This problem could conceivably even persist for a while, though I have trouble coming up with a case where that's likely _and_ the sled agent is alive enough to process the firewall update.) 
- It could contain zones that don't exist because it's old.  As above, this seems like a non-issue. 

Between the two of these, I think the blueprint is a better choice because it's less likely to be wrong (transient failures won't cause that) and ordinarily it would only be wrong for a short time (until the blueprint is realized).

---

I'd previously suggested that we update _all_ sleds every time.  You could look at this as letting the Sled Agent determine whether the update applies to anything it's running, which is better than both of these.  I realize this isn't going to scale that well in the long term.  But right now, it seems we have two tradeoffs: 

- Try to prune the set of sled agents that need to know about an update (what we're doing now and after this PR).  Risk: if we get this wrong, critical firewall rules might not be applied to something that needs it.  Depending on _why_ we got it wrong, this problem may or may not resolve itself on the next round.
- Just send the update to all sleds.  Risk: wasted effort for sleds that didn't need the update.

Is it expensive or otherwise impactful for a Sled Agent to receive an update that it doesn't need?  If not, these risks seem quite unbalanced.  I'd take the small extra resource utilization over the risk of getting this wrong at least until we have reason to believe that resource utilization is significant.

But I also think using the current target blueprint is probably fine.

schema/crdb/dbinit.sql
```diff
@@ -3159,6 +3159,10 @@ CREATE TABLE IF NOT EXISTS omicron.public.bp_target (
     time_made_target TIMESTAMPTZ NOT NULL
 );
 
+CREATE INDEX IF NOT EXISTS lookup_bp_target_by_id ON omicron.public.bp_target (
+    blueprint_id
+);
```
davepacheco, line 3164: I think I would add `version DESC` to the indexed columns.  For the query you're doing, that would allow you to always find the right record quickly, even if the current blueprint had been enabled/disabled a zillion times.  If at some point we wind up querying for the "target" record for a blueprint, we could in principle paginate through this by (blueprint_id, version), although that admittedly seems excessive.

nexus/db-queries/src/db/datastore/vpc.rs
```diff
@@ -650,8 +650,8 @@ impl DataStore {
         // Resolve each VNIC in the VPC to the Sled it's on, so we know which
```
davepacheco, line 650: (This belongs on L643 -- thanks GitHub)

This function returns the list of `Sled`s hosting instances _or control plane zones_ with network interfaces on the provided VPC, right?

nexus/db-queries/src/db/datastore/vpc.rs
```diff
@@ -681,6 +686,41 @@ impl DataStore {
...
             .select(Sled::as_select());
 
+        // ... and we also need to query for the current target blueprint to
+        // support systems that _are_ under Reconfigurator control. This query
+        // needs a join with a subquery, which isn't currently supported by
+        // diesel's DSL, so we fall back to a raw SQL string and rely on tests
+        // to ensure that this is valid.
+        //
+        // Diesel requires us to use aliases in order to refer to the
+        // `inv_collection` table twice in the same query.
```
davepacheco, line 696:
```suggestion
        // `bp_target` table twice in the same query.
```

nexus/db-queries/src/db/datastore/vpc.rs
```diff
@@ -681,6 +686,41 @@ impl DataStore {
             .filter(service_network_interface::time_deleted.is_null())
             .select(Sled::as_select());
 
+        // ... and we also need to query for the current target blueprint to
+        // support systems that _are_ under Reconfigurator control. This query
+        // needs a join with a subquery, which isn't currently supported by
+        // diesel's DSL, so we fall back to a raw SQL string and rely on tests
+        // to ensure that this is valid.
```
davepacheco, line 693: Where's the raw SQL that this refers to?

nexus/db-queries/src/db/datastore/vpc.rs
```diff
@@ -681,6 +686,41 @@ impl DataStore {
...
+                    .eq(bp_target1.field(bp_target::blueprint_id))),
+            )
+            .inner_join(sled::table.on(sled::id.eq(bp_omicron_zone::sled_id)))
+            .filter(
+                // This filters us down to the one current target blueprint (if
+                // it exists); i.e., the target with the maximal version.
+                bp_target1.field(bp_target::blueprint_id).eq_any(
+                    bp_target2
+                        .select(bp_target2.field(bp_target::blueprint_id))
+                        .order_by(bp_target2.field(bp_target::version).desc())
```
davepacheco, line 716: I thought a bunch about whether we want to add an `enabled = true` filter here.  I think we don't.  Regardless of whether we're trying to do anything with this blueprint, it _is_ the current intended state.

jgallagher:
> I'd previously suggested that we update _all_ sleds every time. You could look at this as letting the Sled Agent determine whether the update applies to anything it's running, which is better than both of these. I realize this isn't going to scale that well in the long term.

I'm not sure how to get to ground on this, but as another piece of input, these are all the call chains where we currently can land in `vpc_resolve_to_sleds`. `vpc_resolve_to_sleds` is only called by `Nexus::send_sleg_agents_firewall_rules()`, which is called by:

* The "VPC create" saga
* `PUT /v1/vpc-firewall-rules` external endpoint
* `Nexus::plumb_service_firewall_rules()`, which is called by:
  * `Nexus::rack_initialize()`, in the RSS-to-nexus handoff path
  * `POST /sled-agents/{sled_id}/firewall-rules-update` internal endpoint, but in this case we additionally restrict to only the specified sled_id (the use of `vpc_resolve_to_sleds` here can only move us from 1 sled to 0 sleds, if the specified sled_id has no NICs in the service VPC)
  * The reconfigurator execution RPW, as of #5157

I think of all of the above, the only one that is immediately problematic is the external endpoint `PUT /v1/vpc-firewall-rules` case. If we removed `vpc_resolve_to_sleds` altogether in favor of "send to all sleds", we'd presumably replace it with `DataStore::sled_list_all_batched()`. `sled_list_all_batched` has an `opctx.check_complex_operations_allowed()`, which specifically rejects opctxs coming from an external endpoint.

nexus/db-queries/src/db/datastore/vpc.rs
jgallagher: Oops, this comment is old. I found a different way to write the query that diesel does accept; will remove.

nexus/db-queries/src/db/datastore/vpc.rs
jgallagher: Agreed. I will add a comment to that effect.

dev-tools/omdb/src/bin/omdb/db.rs
```diff
@@ -1863,23 +1924,12 @@ async fn cmd_db_eips(
...
-                            "error looking up service with id {owner_id}: {e}"
-                        );
-                        continue;
-                    }
-                };
-                Owner::Service {
-                    id: owner_id,
-                    kind: format!("{:?}", service.1.kind),
-                }
+                let kind =
```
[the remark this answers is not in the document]
jgallagher: Also ended up being a little simpler, I think? (Or maybe that's just my diesel bias) Changed in f6220667e

nexus/db-queries/src/db/datastore/vpc.rs
jgallagher: Correct; updated comment in 4fa423bf0

schema/crdb/dbinit.sql
jgallagher: We looked at this offline for a while. I claimed (with `EXPLAIN` evidence) that we didn't need to add `version` to the index, which led to "why does CRDB need us to make an index here at all", which led to a very slight query modification that made the index unnecessary: instead of filtering by "the blueprint_id that matches the blueprint_id of the latest target version" (for which CRDB wants an index on `bp_target.blueprint_id`), we can filter by "the version that matches the latest target version" (for which CRDB can use the existing pkey index on `version`). This is implemented in ff89fa5e5

jgallagher:
> I'd previously suggested that we update _all_ sleds every time. You could look at this as letting the Sled Agent determine whether the update applies to anything it's running, which is better than both of these. I realize this isn't going to scale that well in the long term.

We talked about this offline. Briefly summarizing: this PR tries to tweak what we're doing today without changing direction (i.e., we're continuing to send firewall rules just those sleds we think need them instead of to all sleds). We can easily revisit this bigger question in the future, and _should_ revisit it if we start running into bugs where we've selected the wrong set of sleds.

davepacheco (approved):

Issue #4805, opened 2024-01-11
omdb could more safely support "write" commands


davepacheco:
illumos debugging tools like DTrace and MDB have a distinction between ordinary operations, which are generally read-only and safe, and "destructive" operations, which may modify the thing you're debugging or otherwise pose some risk.  I think this has proved a really successful model that we could use in omdb as well.  By default it should do no harm.  We could add other commands behind a top-level `-w`/`--destructive` flag.

It could be a little risky to do this purely by convention.  But I wonder if we could enforce it pretty well if:

- Could we add to progenitor a way to create a read-only client, that could only perform GET/HEAD operations?
- For the database connection pool, I wonder if we could add a mode that runs `set default_transaction_read_only = on` when establishing connections so that write operations would be rejected by the server?  We'd want to test this out -- I'm not clear from the [docs](https://www.cockroachlabs.com/docs/v23.1/show-vars) if this applies to implicit transactions or just those created with `BEGIN TRANSACTION`

If these both work out, we'd only have to check for the destructive flag in a few places where we create clients and the connection pool, instead of relying on everybody to check this whenever they add new commands.

davepacheco:
I filed https://github.com/seanmonstar/reqwest/issues/2160, thinking if reqwest supports this, we could use that in constructing our reqwest clients for our Progenitor clients.

jgallagher:
> It could be a little risky to do this purely by convention.

Just noting: we do attempt to enforce this by convention today (which does not obviate the desire for stronger enforcement as described in this issue!) via [check_allow_destructive](https://github.com/oxidecomputer/omicron/blob/88030ff7d9f9552bebe8129983d11b1671bc4064/dev-tools/omdb/src/bin/omdb/main.rs#L135-L146). Operations which perform writes should accept a `DestructiveOperationToken` which can only be created if the `--destructive` flag has been passed; e.g., https://github.com/oxidecomputer/omicron/blob/88030ff7d9f9552bebe8129983d11b1671bc4064/dev-tools/omdb/src/bin/omdb/nexus.rs#L2141

PR #5204, opened 2024-03-06
Expose uninitialized sleds and "add sled" via internal API and omdb

### LINKED ISSUE #5133: move (or copy?) "add sled" APIs to internal API
Right now, "add sled" involves two external API endpoints:

- list uninitialized sleds
- add sled

For our first intended customer use of this, we'll need access to these from the internal API also/instead.  We'll also need some way to invoke them, either with an internal API CLI or else omdb.

jgallagher:
This will let us drive the add sled process from a support context. (Prior to this change adding a sled required interaction with the external API.)

I want to test this before merging, but it's small enough that if there's anything that needs tweaking it's unlikely to affect any review comments. One change is that we had two `Baseboard` structs that were identical except for a minor different in field names (`{serial,part}` vs `{serial_number,part_number}`). I removed the internal one and replaced it with the external one to avoid unnecessary churn in the public API, but if I'm missing something intentional that led to having both, please say so.

Closes #5133.

### DIFF
--- a/dev-tools/omdb/src/bin/omdb/main.rs
+++ b/dev-tools/omdb/src/bin/omdb/main.rs
@@ -84,11 +84,24 @@ struct Omdb {
     #[arg(env = "OMDB_DNS_SERVER", long)]
     dns_server: Option<SocketAddr>,
 
+    /// allow potentially-destructive subcommands
+    #[arg(short = 'w', long = "destructive")]
+    allow_destructive: bool,
+
     #[command(subcommand)]
     command: OmdbCommands,
 }
 
 impl Omdb {
+    fn check_allow_destructive(&self) -> anyhow::Result<()> {
+        anyhow::ensure!(
+            self.allow_destructive,
+            "This command is potentially destructive. \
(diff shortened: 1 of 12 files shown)


andrewjstone (approved):
Awesome :)

jgallagher:
> Although we haven't really done #4805, what would you think about adding an opt-in mechanism here? Basically a `-w`/`--destructive` flag at the top level (`omdb -w`) that we'd check here and bail out if it wasn't set? Seems low cost and a potentially valuable safety.

No objection and very easy; added in [a1ff0c1](https://github.com/oxidecomputer/omicron/pull/5204/commits/a1ff0c1e1721d58ea3ab01224cf4cb0a505033d5). I attached the check to all the `nexus` subcommands that result in database modifications (please double check I didn't miss any). I don't think we have anything under `omdb {db,mgs,oximeter,sled-agent}` that qualifies as destructive?

davepacheco:
Above and beyond!  Thanks.  I did a quick pass over existing commands and found no others that I'd attach this too.  I wish it were easy to add a test for this but it doesn't seem like it is.

PR #5211, opened 2024-03-07
[WIP] [reconfigurator] start removing expunged zones

sunshowers:
Putting this up for early review. This implements internal bookkeeping for the blueprint builder, as well as database storage for expunged Nexus zones.

This does not implement any of the other things we discussed in Tuesday's meeting -- let's discuss this tomorrow.


nexus/reconfigurator/planning/src/blueprint_builder.rs
```diff
@@ -305,6 +362,7 @@ impl<'a> BlueprintBuilder<'a> {
             sled_ip_allocators: BTreeMap::new(),
             zones: BlueprintZones::new(parent_blueprint),
             zones_in_service: parent_blueprint.zones_in_service.clone(),
+            expunged_nexus_zones: parent_blueprint.expunged_nexus_zones.clone(),
```
jgallagher, line 365: Happy to discuss this during the meeting later today, but just jotting down a couple of thoughts here:

* Once a zone gets added to `expunged_nexus_zones`, how does the planner know when it's okay to remove it from a subsequent child blueprint?
* Does this need to be _Nexus_ zones specifically? Or could it be `zones_expunged` like `zones_in_service`? (I haven't thought about this much, but am assuming we're eventually going to want to track other kinds of zones that need execution work to be expunged.)

nexus/reconfigurator/planning/src/blueprint_builder.rs
```diff
@@ -142,6 +195,9 @@ impl<'a> BlueprintBuilder<'a> {
             .sleds
             .keys()
             .map(|sled_id| {
+                // XXX: do we need to handle, or check for, initially expunged
```
davepacheco, line 198: I would say no.  The only thing we could do here is mark zones expunged if we find them on sleds that are expunged.  But that's more of the planner's job.  The idea behind the initial blueprint is that if you executed it, it wouldn't do anything.  (That's why this is in BlueprintBuilder, which is not supposed to be that smart, and not Planner.)

nexus/reconfigurator/planning/src/blueprint_builder.rs
```diff
@@ -566,6 +721,88 @@ impl<'a> BlueprintBuilder<'a> {
...
+        let sled_zones = self.zones.change_sled_zones(sled_id);
+        let zones = std::mem::replace(&mut sled_zones.zones, Vec::new());
+        let (expunge, keep) = zones
+            .into_iter()
+            .partition::<Vec<_>, _>(|z| zone_ids_to_expunge.contains(&z.id));
+        sled_zones.zones = keep;
+
+        // Also add the list of expunged Nexus zones to self.expunged_nexus_zones.
+        self.expunged_nexus_zones.extend(expunge.iter().filter_map(|z| {
+            (z.zone_type.tag() == ZoneTag::Nexus).then_some(z.id)
```
davepacheco, line 793: FYI, I think there's already a `is_nexus()`.  (No reason to change this though.)

nexus/reconfigurator/planning/src/blueprint_builder.rs
```diff
@@ -566,6 +721,88 @@ impl<'a> BlueprintBuilder<'a> {
...
+        Ok(expunge)
+    }
+
+    // ---
+    // Helper methods
+    // ---
+
+    // Precondition: we must have checked that it is valid for the zone to be
+    // added to the sled. (TODO: actually check this here, or use the type
+    // system to statically verify this.)
```
davepacheco, line 805: Can you say more about this?  What does "valid" mean?  Does this panic or return an error or do something bad if this precondition is violated?

nexus/reconfigurator/planning/src/blueprint_builder.rs
```diff
@@ -566,6 +721,88 @@ impl<'a> BlueprintBuilder<'a> {
...
+        self.sled_expunge_zones(sled_id, |_| true)
+    }
+
+    /// Expunges a set of zones corresponding to this sled ID.
+    ///
+    /// Returns the list of expunged [`OmicronZoneConfig`] instances, or an
+    /// error if the sled ID wasn't found.
+    ///
+    /// Return true in `expunge_fn` to remove a zone, or false to retain it.
+    pub fn sled_expunge_zones(
```
davepacheco, line 746: Out of curiosity, why not just a `sled_expunge_zone(sled_id, zone_id)`?

nexus/reconfigurator/planning/src/planner.rs
```diff
@@ -63,9 +64,85 @@ impl<'a> Planner<'a> {
...
+
+                    return Some("decommissioned");
+                }
+            }
+
+            match sled_info.policy {
+                SledPolicy::InService { .. } => None,
+                SledPolicy::Expunged => Some("expunged"),
+            }
+        }
```
davepacheco, line 101: Is it possible to have a sled in state "decommissioned" that's not got policy "expunged"?  (If not, I think a lot of this could be simplified -- there'd only be one possible expunge reason and you'd only need to check one field to identify it.)

nexus/reconfigurator/planning/src/planner.rs
```diff
@@ -85,9 +162,9 @@ impl<'a> Planner<'a> {
         let mut sleds_ineligible_for_services = BTreeSet::new();
 
         for (sled_id, sled_info) in &self.policy.sleds {
-            // Decommissioned sleds don't get any services. (This is an
-            // explicit match so that when more states are added, this fails to
-            // compile.)
+            // Decommissioned and expunged sleds don't get any services. (These
+            // are explicit matches so that when more states are added, this
+            // fails to compile.)
```
davepacheco, line 167: Similarly I wonder if this could just look at "expunged" now.  It feels like only the policy should determine how we plan resources for the sled.

davepacheco (commented):
Nice.  (I know this is still a WIP.)

nexus/reconfigurator/planning/src/blueprint_builder.rs
```diff
@@ -566,6 +721,88 @@ impl<'a> BlueprintBuilder<'a> {
...
+        slog::debug!(
+            self.log,
+            "expunging zones for sled";
+            "sled_id" => %sled_id,
+            "zone_ids_to_expunge" => ?zone_ids_to_expunge,
+        );
+
+        // Check that all the zone IDs exist. It is an internal invariant
+        // violation for a zone to be in the list of current sled zones, but
+        // not in zones_in_service.
```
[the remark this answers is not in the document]
sunshowers: Yes -- I was thinking about this last night and what I did here is wrong. Thanks for catching this.

nexus/reconfigurator/planning/src/blueprint_builder.rs
andrewjstone: > Happy to discuss this during the meeting later today, but just jotting down a couple of thoughts here:
> 
> * Once a zone gets added to `expunged_nexus_zones`, how does the planner know when it's okay to remove it from a subsequent child blueprint?

I was wondering the same thing. Should there also be a 'decommissioned' column for when the zone has been completely removed?

> * Does this need to be _Nexus_ zones specifically? Or could it be `zones_expunged` like `zones_in_service`? (I haven't thought about this much, but am assuming we're eventually going to want to track other kinds of zones that need execution work to be expunged.)

I was also thinking the same thing. It seems somewhat silly to have a separate table per zone type unless each of those tables would have different rows because they store other "checkpoint" information while decommissioning a zone. I don't think we were planning to do this inside the zone table, but I'm not sure.

sunshowers: The latter we've taken into account by switching to a disposition enum.

sunshowers:
Turns out it's too hard to make this rebase cleanly, so I'm going to start from scratch and copy over bits as they make sense.

PR #5223, opened 2024-03-07
Unquarantine clickhouse binary on mac

david-crespo:
After #5127 (I assume, because this problem came up immediately after that) macOS Gatekeeper is putting the `clickhouse` binary in quarantine after download, which means we can't execute it unless we take it out of quarantine. This is apparently a new thing, as [this doc](https://github.com/ClickHouse/clickhouse-docs/blob/08d7a329d/knowledgebase/fix-developer-verification-error-in-macos.md) about how to do that was only [added Jan 9](https://github.com/ClickHouse/clickhouse-docs/pull/1835).

Here I am doing it in the simplest way possible: if mac, then remove quarantine attr. If anyone has a better way, I'm all ears.

### DIFF
--- a/tools/ci_download_clickhouse
+++ b/tools/ci_download_clickhouse
@@ -84,6 +84,12 @@ function main
 	# Unpack the tarball into a local directory
 	do_untar "$TARBALL_FILE" "$DEST_DIR"
 
+	# on macOS, we need to take the binary out of quarantine after download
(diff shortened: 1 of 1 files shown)


sunshowers (approved):
lgtm!

As mentioned in #oxide-control-plane, it's definitely quite weird that this (or homebrew for that matter) is hitting the quarantine bit. Generally only files downloaded via a web browser are supposed to trigger quarantining -- untarred files aren't supposed to. But figuring that out's for a future time, and there's nothing wrong with this.

PR #5229, opened 2024-03-08
async-await-ifying DumpSetup

lifning:
Replace the std::thread + Mutex originally used in DumpSetup with a tokio task and message-passing.
(Verified cores and kernel dumps and logs all getting archived appropriately on a bench gimlet)

### DIFF
--- a/sled-agent/src/dump_setup.rs
+++ b/sled-agent/src/dump_setup.rs
@@ -118,13 +118,13 @@ struct DebugDataset(Utf8PathBuf);
 #[derive(AsRef, Clone, Debug, Eq, From, Hash, Ord, PartialEq, PartialOrd)]
 struct CoreDataset(Utf8PathBuf);
 
-#[derive(AsRef, Clone, From)]
+#[derive(AsRef, Clone, Debug, From)]
 pub(super) struct CoreZpool {
     mount_config: MountConfig,
     name: ZpoolName,
 }
 
-#[derive(AsRef, Clone, From)]
+#[derive(AsRef, Clone, Debug, From)]
 pub(super) struct DebugZpool {
     mount_config: MountConfig,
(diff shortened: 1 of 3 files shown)


sled-agent/src/dump_setup.rs
```diff
@@ -492,24 +470,67 @@ impl DumpSetupWorker {
...
             coredumpadm_invoker,
             zfs_invoker,
             zone_invoker,
         }
     }
 
+    async fn poll_file_archival(mut self) {
+        info!(self.log, "DumpSetup poll loop started.");
+        loop {
+            match tokio::time::timeout(ARCHIVAL_INTERVAL, self.rx.recv()).await
```
smklein, line 483: This does look like it matches the old behavior, where we slept for `ARCHIVAL_INTERVAL`, but now that this is async -- do we want to `select` on this with a timeout?

Seems like we might want to react immediately if we get a request to `UpdateDumpdevSetup`, rather than waiting for the `ARCHIVAL_INTERVAL` to fire, and then we could reevaluate our choices + restart the timeout interval.

lifning: i think that's what this does, right? if `self.rx.recv()` gets something before `ARCHIVAL_INTERVAL` elapses, we break out of [tokio::time::timeout](https://docs.rs/tokio/latest/tokio/time/fn.timeout.html) early with the result. in both cases (got a message, or elapsed trying) we make a pass of archiving files - either right after we've re-evaluated our path choices or gone for five minutes without any changes to them

smklein: You're totally right, I misread this! Somehow didn't see the `self.rx.recv` within the `timeout` call

smklein (approved):

PR #5230, opened 2024-03-08
oxide/opte-interface-setup should default to disabled

citrus-it:
The `oxide/opte-interface-setup` service is present in all NTP zones, but only the boundary zones require an OPTE interface. sled-agent explicitly enables the service for just the boundary zones, but because the service is enabled by default it ends up attempting to start in the non-boundaries too, where it fails.

```
[ Mar  8 12:56:26 Executing start method ("/opt/oxide/zone-network-cli/bin/zone-networking opte-interface-set-up -i unknown -g unknown -p unknown"). ]
note: configured to log to "/dev/stderr"
error: invalid value 'unknown' for '--opte_interface <STRING>': ERROR: Missing OPTE interface

For more information, try '--help'.
[ Mar  8 12:56:26 Method "start" exited with status 2. ]
```

### DIFF
--- a/smf/opte-interface-setup/manifest.xml
+++ b/smf/opte-interface-setup/manifest.xml
@@ -4,7 +4,7 @@
 <service_bundle type='manifest' name='opte-interface-setup'>
 
 <service name='oxide/opte-interface-setup' type='service' version='1'>
-  <create_default_instance enabled='true' />
+  <create_default_instance enabled='false' />
 
   <!-- Run after the operating system's svc:/network/physical service is done. -->
   <dependency name='physical' grouping='require_all' restart_on='none'
@@ -20,7 +20,7 @@
   <exec_method type='method' name='start'
     exec='/opt/oxide/zone-network-cli/bin/zone-networking opte-interface-set-up -i %{config/interface} -g %{config/gateway} -p %{config/ip}'
     timeout_seconds='0' />
-  
+
   <property_group name='startd' type='framework'>
     <propval name='duration' type='astring' value='transient' />
   </property_group>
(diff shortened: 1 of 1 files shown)


citrus-it:
I deployed this commit on a bench gimlet and confirmed that the service came online in the single boundary NTP zone, and that time synchronised.

```
root@oxz_ntp_f27aeb1e:~# grep default /var/svc/manifest/site/opte-interface-setup/manifest.xml
  <create_default_instance enabled='false' />

root@oxz_ntp_f27aeb1e:~# svcs opte-interface-setup
STATE          STIME    FMRI
online         13:50:25 svc:/oxide/opte-interface-setup:default

root@oxz_ntp_f27aeb1e:~# ipadm
vopte0/public     dhcp     ok           172.30.3.5/32
oxControlService4/ll addrconf ok        fe80::8:20ff:fef0:5d89%oxControlService4/10
oxControlService4/omicron6 static ok    fd00:1122:3344:101::1b/64

root@oxz_ntp_f27aeb1e:~#
logout

[Connection to zone 'oxz_ntp_f27aeb1e-d65a-48fc-bdf4-286c6b73b6ad' pts/3 closed]

gimlet-sn06 # handoff
13:54:31.251Z INFO SledAgent (RSS): Handoff to Nexus is complete
    file = sled-agent/src/rack_setup/service.rs:686
```

smf/opte-interface-setup/manifest.xml
```diff
@@ -4,7 +4,7 @@
 <service_bundle type='manifest' name='opte-interface-setup'>
 
 <service name='oxide/opte-interface-setup' type='service' version='1'>
-  <create_default_instance enabled='true' />
+  <create_default_instance enabled='false' />
```
karencfv, line 7: aahh! I see what's going on there

smf/opte-interface-setup/manifest.xml
```diff
@@ -35,7 +35,7 @@
 
   <template>
     <common_name>
-      <loctext xml:lang='C'>Oxide OPTE interface set up</loctext>
+      <loctext xml:lang='C'>Oxide OPTE interface setup</loctext>
```
karencfv, line 38: gargh English :expressionless:. I seem to always get those two wrong

karencfv (approved):
Thanks for the fix!

PR #5231, opened 2024-03-08
Add internal API and omdb commands to toggle current target blueprint's enabled bit

### LINKED ISSUE #5210: blueprints: allow toggling current target's `enabled` bit
The only mechanism we have today for setting the current target blueprint is via `omdb nexus blueprints target set <UUID>`, which attempts to preserve the existing enabled-ness of the prior target blueprint (defaulting to `true` for the initial blueprint). However, we don't have any means of setting the enabled bit to false! We should be able to:

* Choose the `enabled` state when setting a target blueprint (enabled, disabled, or "inherit from parent")
* Toggle the `enabled` state without changing the target blueprint

The latter will require some work in Nexus and `nexus-db-queries`. Currently the only path by which a blueprint target can be set is through `DataStore::blueprint_target_set_current()`, which only allows _new_ targets, and this method fails if the current target is not the immediate parent of the new target. We'll need a way to flip the `enabled` bit (which will push a new `bp_target` row with the same `blueprint_id` but an incremented `version`) and a way to trigger that from `omdb` (presumably a new internal Nexus endpoint).

jgallagher:
I chose `PATCH` for the new endpoint; I think that's correct? Although it's the first use of `PATCH` in omicron AFAICT so maybe it should just be `PUT`?

There are a couple related-but-not-required changes as a part of this:

* `omdb nexus blueprints list` now has an `ENA: yes/no` column that is only populated for the current target
* We activate the blueprint execution bg task when setting an enabled target or changing the current target to be enabled. (The latter might be superfluous if the target was already enabled, but that's fine.)

I tried this locally against `omicron-dev run-all` and it all seemed to check out. If folks would be more comfortable with a test on a real system before merging I can do that; otherwise I'll test it next time I take a lap for some other reason.

Closes #5210.


jgallagher:
> I think it would be easiest to continue punting on this question and just using PUT.

:+1: done in [490e97a](https://github.com/oxidecomputer/omicron/pull/5231/commits/490e97a14566325336de2570c16a2bcbedb96675)

nexus/src/app/background/blueprint_load.rs
```diff
@@ -153,10 +153,34 @@ impl BackgroundTask for TargetBlueprintLoader {
...
+                                "time_created" => &time_created,
+                                "state" => status,
+                            );
+                            self.last =
+                                Some(Arc::new((new_bp_target, new_blueprint)));
+                            self.tx.send_replace(self.last.clone());
+                            json!({
+                                "target_id": target_id,
+                                "time_created": time_created,
+                                "time_found": chrono::Utc::now().to_string(),
```
sunshowers, line 177: Is the `to_string` really necessary?

nexus/db-queries/src/db/datastore/deployment.rs
```diff
@@ -641,6 +643,88 @@ impl DataStore {
...
+        // INSERT INTO bp_target
+        //   (SELECT
+        //        version + 1,
+        //        blueprint_id,
+        //        <target.enabled>,
+        //        <target.time_made_target>
+        //    FROM bp_target
+        //    WHERE
+        //        -- This part of the subquery restricts us to only the
+        //        -- current target (i.e., the bp_target with maximal versio)
```
davepacheco, line 677:
```suggestion
        //        -- current target (i.e., the bp_target with maximal version)
```

nexus/db-queries/src/db/datastore/deployment.rs
```diff
@@ -641,6 +643,88 @@ impl DataStore {
         Ok(())
     }
 
+    /// Set the current target blueprint's `enabled` field
+    ///
+    /// In order to change the enabled field, `target` must already be the
+    /// current target blueprint
+    pub async fn blueprint_target_set_current_enabled(
```
davepacheco, line 650: I feel like it's worth a comment explaining why this needs to be different from the other code path for setting the target.  I assume the reason is that that code path would fail because it verifies that the parent blueprint is the current target?  Assuming that's the reason, did you consider having that succeed if the _current_ blueprint matches the one we're trying to set?

davepacheco (approved):

nexus/db-queries/src/db/datastore/deployment.rs
jgallagher: Hmm, I thought I already addressed that (if briefly) in the doc comments stating the requirements of the `target`. I can expand those, but maybe you're asking for something different?

I did consider just funneling this through `blueprint_target_set_current` and changing its requirements, and opted not to for two reasons:

* These really are different operations, so I think it makes sense for them to have error conditions that are specific to each. In particular, I think it's useful to have a "you tried to set a new target blueprint but you can't because it already is the target"  error possibility. Maybe that's wrong and we'd like it to be more idempotent? I am not sure about this.
* The `InsertTargetQuery` CTE is kind of hairy already. Changing its checks to allow the current blueprint matching in addition to the parent is certainly doable, but would be more work than making them separate.

nexus/src/app/background/blueprint_load.rs
jgallagher: Doesn't look like it is, no. This was copy/pasted from nearby; I'll remove the unnecessary `to_string()`s.

nexus/db-queries/src/db/datastore/deployment.rs
davepacheco: That makes sense, although:

> These really are different operations, so I think it makes sense for them to have error conditions that are specific to each. In particular, I think it's useful to have a "you tried to set a new target blueprint but you can't because it already is the target" error possibility. Maybe that's wrong and we'd like it to be more idempotent? I am not sure about this.

I do think it'd be a little better if it were idempotent.  I raised this because as a caller it seems like it could be a little annoying that you can't just construct a `BlueprintTarget` matching whatever you (or the user) currently wants.  You have to invoke a different function (or not invoke it) based on what changed about the struct.  But given your other point about the hairy CTE (which I totally get), I don't think it makes sense to change any of this now.  We can always revisit if/when we make this part of the external API -- we're not committing to much here.

The doc comment explains the restriction, but not really why.  I'm just thinking something like:

```rust
// Although this function is like `blueprint_target_set_current()` in that both store the given `BlueprintTarget` into the table, the functions are distinct because the preconditions and error cases are different. 
```

(and it's not a big deal either way -- it was just a small point of confusion for me reading the code)

Issue #5134, opened 2024-02-23
Define public endpoint for marking sled as expunged


sunshowers:
Followup to #5032.

We need to define a public endpoint on the Nexus external API to mark a sled as expunged. This is straightforward in principle, but it is an irreversible operation and must be done with careful documentation.

davepacheco:
We may also need one on the internal API (see #5133).

sunshowers:
Thanks, updated description.

benjaminleonard:
Alongside the pattern, perhaps we take care that the API / CLI / Console have the appropriate (and equivalent?) level of friction and warning.

_Console_: Modal that perhaps requires the user to enter the name of the sled to confirm the action
_CLI_: Interactive y/n confirmation (maybe with a --y or --force flag)
_API_:  Is there a confirm deletion pattern that works well for RESTful APIs?

sunshowers:
Discussed in today's meeting: per @jgallagher, for C2 we can just do the internal API since we're going to trigger it from omdb.

jgallagher:
With #5234 we now have an internal API + omdb subcommand to expunge a sled. I changed the description and title back to just defining a public endpoint. I think we could also change the deliverable target now, but I'll defer to @davepacheco or @morlandi7 (I'm not entirely sure what it should be - do we have a target for "customer-driven sled add/remove"?)

davepacheco:
I'll move it to the general "add/remove sled" for now.  Thanks.

PR #5234, opened 2024-03-08
Add internal API and omdb command to expunge a sled

jgallagher:
The omdb interface might be overly cautious; feedback welcome. You can try this against `omicron-dev run-all`; it looks roughly like:

```
note: using Nexus URL http://[::1]:12221
note: using database URL postgresql://root@[::1]:42285/omicron?sslmode=disable
note: database schema version matches expected (39.0.0)
WARNING: sled b6d65341-167c-41df-9b5c-41cded99c229 IS PRESENT in the most recent inventory collection; are you sure you want to mark it expunged?
WARNING: This operation will PERMANENTLY and IRRECOVABLY mark sled b6d65341-167c-41df-9b5c-41cded99c229 (sim-b6d65341) expunged. To proceed, type the sled's serial number.
sled serial〉sim-b6d65341
expunged sled b6d65341-167c-41df-9b5c-41cded99c229 (previous policy: InService(Provisionable))
```

This handles the "internal" half of #5134


dev-tools/omdb/src/bin/omdb/nexus.rs
```diff
@@ -1048,3 +1074,102 @@ async fn cmd_nexus_sled_add(
...
+                    || collection.sps.keys().any(|baseboard| {
+                        baseboard.part_number == sled.part_number()
+                            && baseboard.serial_number == sled.serial_number()
+                    });
+            if sled_present {
+                eprintln!(
+                    "WARNING: sled {} IS PRESENT in the most recent inventory \
+                    collection; are you sure you want to mark it expunged?",
+                    args.sled_id,
+                );
```
jgallagher, line 1135: If "type the serial number in to continue" below is worth it, I wonder if we should also "type y/N to continue" here if we find the sled in the latest collection?

dev-tools/omdb/src/bin/omdb/db.rs
```diff
@@ -159,6 +158,71 @@ pub struct DbArgs {
     command: DbCommands,
 }
 
+#[derive(Debug, Args)]
+pub struct DbUrlOptions {
+    /// URL of the database SQL interface
+    #[clap(long, env("OMDB_DB_URL"))]
+    db_url: Option<PostgresConfigWithUrl>,
```
sunshowers, line 165: May be worth having a mutually exclusive option with `--db-host`/`OMDB_DB_HOST` here, which substitutes a `host` or `host:port` combo for the DNS lookup.

dev-tools/omdb/src/bin/omdb/db.rs
```diff
@@ -159,6 +158,71 @@ pub struct DbArgs {
...
+        }
+    }
+
+    pub async fn connect(
+        &self,
+        omdb: &Omdb,
+        log: &slog::Logger,
+    ) -> anyhow::Result<Arc<DataStore>> {
+        let db_url = self.resolve_pg_url(omdb, log).await?;
+        eprintln!("note: using database URL {}", &db_url);
```
sunshowers, line 208: Can this use the logger?

dev-tools/omdb/src/bin/omdb/nexus.rs
```diff
@@ -253,6 +273,12 @@ impl NexusArgs {
                 omdb.check_allow_destructive()?;
                 cmd_nexus_sled_add(&client, args).await
             }
+            NexusCommands::Sleds(SledsArgs {
+                command: SledsCommands::Expunge(args),
+            }) => {
+                omdb.check_allow_destructive()?;
```
sunshowers, line 279: Same comment as https://github.com/oxidecomputer/omicron/pull/5231#discussion_r1520283160 -- it would be nice to have a token here indicating that `check_allow_destructive` has been called.

dev-tools/omdb/src/bin/omdb/nexus.rs
```diff
@@ -1048,3 +1074,102 @@ async fn cmd_nexus_sled_add(
...
+    // couple of safeguards in place to ensure this cannot be called without
+    // due consideration:
+    //
+    // 1. We'll require manual input on stdin to confirm the sled to be removed
+    // 2. We'll warn sternly if the sled-to-be-expunged is still present in the
+    //    most recent inventory collection
+    use nexus_db_model::Sled;
+    use nexus_db_queries::context::OpContext;
+
+    // First, we need to look up the sled so we know it's serial number.
```
sunshowers, line 1095:
```suggestion
    // First, we need to look up the sled so we know its serial number.
```

dev-tools/omdb/src/bin/omdb/nexus.rs
```diff
@@ -1048,3 +1074,102 @@ async fn cmd_nexus_sled_add(
...
+        use diesel::SelectableHelper;
+        use nexus_db_queries::db::schema::sled::dsl;
+
+        let conn = datastore.pool_connection_for_tests().await?;
+        dsl::sled
+            .filter(dsl::id.eq(args.sled_id))
+            .select(Sled::as_select())
+            .first_async::<Sled>(&*conn)
+            .await
+            .with_context(|| format!("failed to find sled {}", args.sled_id))?
```
sunshowers, line 1113: Can this use `lookup_path` to look up the sled?

dev-tools/omdb/src/bin/omdb/nexus.rs
```diff
@@ -1048,3 +1074,102 @@ async fn cmd_nexus_sled_add(
...
+        .await
+        .context("loading latest collection")?
+    {
+        Some(collection) => {
+            let sled_present =
+                collection.sled_agents.contains_key(&args.sled_id)
+                    || collection.sps.keys().any(|baseboard| {
+                        baseboard.part_number == sled.part_number()
+                            && baseboard.serial_number == sled.serial_number()
+                    });
```
sunshowers, line 1129: May be worth making this a method on `Collection` since it seems generally useful. I was looking for something like this as part of #5211 myself.

dev-tools/omdb/src/bin/omdb/nexus.rs
```diff
@@ -1048,3 +1074,102 @@ async fn cmd_nexus_sled_add(
...
+            if sled_present {
+                eprintln!(
+                    "WARNING: sled {} IS PRESENT in the most recent inventory \
```
sunshowers, line 1132: Can this use the logger/`warn!`?

dev-tools/omdb/src/bin/omdb/nexus.rs
sunshowers: Interesting question -- I think asking for y/N makes sense, yeah. This is only going to be done a small number of times in practice, so putting some friction here seems fine.

dev-tools/omdb/src/bin/omdb/nexus.rs
```diff
@@ -1048,3 +1074,102 @@ async fn cmd_nexus_sled_add(
...
+        None => {
+            eprintln!("WARNING: no inventory collections present");
+        }
+    }
+
+    eprintln!(
+        "WARNING: This operation will PERMANENTLY and IRRECOVABLY mark sled \
+        {} ({}) expunged. To proceed, type the sled's serial number.",
+        args.sled_id,
+        sled.serial_number(),
```
sunshowers, line 1147: Love this check!

sunshowers (commented):
Looks really good overall, thanks for doing it! Just a few comments.

dev-tools/omdb/src/bin/omdb/db.rs
jgallagher: Sorry, what is `--db-host`/`OMDB_DB_HOST`? I think you can already override the DNS server used for any lookups via `--dns-server`/`OMDB_DNS_SERVER`.

(FWIW all of this was just refactoring this into its own type to reuse it from the `nexus` module; I made no changes to the implementation.)

dev-tools/omdb/src/bin/omdb/db.rs
jgallagher: I don't think it's supposed to, but I'm not positive. Probably a question for @davepacheco; I _think_ the logger is only passed into all the omicron types that need one, and output from `omdb` itself goes to stdout or stderr.

dev-tools/omdb/src/bin/omdb/nexus.rs
jgallagher: I think this also falls under the "where should omdb emit messages" question above. For this one in particular, if we're going to add a `y/N` that we expect the user to input on stdin, I think it needs to go to stdout or stderr, right?

sunshowers: Ah yes -- for prompts, absolutely. (May be worth checking out [dialoguer](https://docs.rs/dialoguer/latest/dialoguer/)).

dev-tools/omdb/src/bin/omdb/db.rs
sunshowers: Ah, by `--db-host` I mean something like:

* take the value over the cli/env
* interpolate it into `postgresql://root@{}/omicron?sslmode=disable`, just like the value from DNS is interpolated

dev-tools/omdb/src/bin/omdb/nexus.rs
jgallagher: Yeah, I think there are at least a few crates that are really targeted at what this was doing; we're already depending on reedline and I thought this was simple enough I'd avoid pulling in another dep. :man_shrugging:

dev-tools/omdb/src/bin/omdb/nexus.rs
jgallagher: Yes, yes it can. I always forget that path.

dev-tools/omdb/src/bin/omdb/db.rs
jgallagher: Ah, I see. I'm inclined to defer that to a separate PR (if it's needed? I basically always use `--dns-server`, either implicitly within a real switch zone or explicitly if I'm testing against `omicron-dev run-all`).

sunshowers: Sure, sounds good. It was just a thought that came to mind :)

dev-tools/omdb/src/bin/omdb/nexus.rs
sunshowers: same 😅

dev-tools/omdb/src/bin/omdb/nexus.rs
jgallagher: Ooh, `Collection` already interns baseboard IDs and stores them in `collection.baseboards`. I can just check that. (If you need something more than "does it exist", I agree adding a method to `Collection` seems reasonable.)

dev-tools/omdb/src/bin/omdb/nexus.rs
jgallagher: Implemented in 27812fde5. I'll let this land first, then use it in #5231 too.

dev-tools/omdb/src/bin/omdb/db.rs
sunshowers: Interesting. I think one way to handle this would be to have some kind of tag passed in as a kv to indicate that the message is user-facing, and in the log handler do filtering based on that. (You can also show such messages in a different manner, e.g. "**warn**: <text>" instead of `[timestamp WARN] text`.)

dev-tools/omdb/src/bin/omdb/main.rs
```diff
@@ -92,16 +92,28 @@ struct Omdb {
...
-    fn check_allow_destructive(&self) -> anyhow::Result<()> {
-        anyhow::ensure!(
-            self.allow_destructive,
-            "This command is potentially destructive. \
+mod check_allow_destructive {
+    /// Zero-size type that potentially-destructive functions can accept to
+    /// ensure `Omdb::check_allow_destructive` has been called.
+    // This is tucked away inside a module to prevent it from being constructed
+    // by anything other than `Omdb::check_allow_destructive`.
+    pub(crate) struct DestructiveOperationToken(());
```
sunshowers, line 100: Would put in `must_use` here.

dev-tools/omdb/src/bin/omdb/nexus.rs
```diff
@@ -1048,3 +1085,109 @@ async fn cmd_nexus_sled_add(
...
+            // there are some test environments (e.g., `omicron-dev run-all`)
+            // where that isn't guaranteed.
+            let sled_present_in_collection =
+                collection.sled_agents.contains_key(&args.sled_id)
+                    || collection.baseboards.contains(&sled_baseboard);
+            if sled_present_in_collection {
+                eprintln!(
+                    "WARNING: sled {} is PRESENT in the most recent inventory \
+                    collection; are you sure you want to mark it expunged?",
+                    args.sled_id,
```
davepacheco, line 1157: I'd suggest elaborating a bit more about the problem, just to give someone who doesn't know what they're doing a clue about the risk they're taking.  Maybe:

> WARNING: sled {} is PRESENT in the most recent inventory collection (spotted at {timestamp}).  Sleds should be removed prior to being expunged (and not re-inserted after).  It is dangerous to expunge a sled that is still running.  Are you sure you want to proceed anyway?

dev-tools/omdb/src/bin/omdb/nexus.rs
```diff
@@ -1048,3 +1085,109 @@ async fn cmd_nexus_sled_add(
...
+                );
+                let confirm = read_with_prompt("y/N")?;
+                if confirm != "y" {
+                    eprintln!("expungement not confirmed: aborting");
+                    return Ok(());
+                }
+            }
+        }
+        None => {
+            eprintln!("WARNING: no inventory collections present");
```
davepacheco, line 1167:
```suggestion
            eprintln!("WARNING: cannot verify that the sled is physically gone because there are no inventory collections present.  Please make sure that the sled has been physically removed.");
```

dev-tools/omdb/src/bin/omdb/nexus.rs
```diff
@@ -1048,3 +1085,109 @@ async fn cmd_nexus_sled_add(
...
+    let opctx = OpContext::for_tests(log.clone(), datastore.clone());
+    let opctx = &opctx;
+
+    // First, we need to look up the sled so we know its serial number.
+    let (_authz_sled, sled) = LookupPath::new(opctx, &datastore)
+        .sled_id(args.sled_id)
+        .fetch()
+        .await
+        .with_context(|| format!("failed to find sled {}", args.sled_id))?;
+
```
davepacheco, line 1116: Is this the only reason we need the database connection?  What about having an API instead to fetch sled info?  (This is not a big deal either way.)

davepacheco (approved):

dev-tools/omdb/src/bin/omdb/nexus.rs
```diff
@@ -1048,3 +1085,109 @@ async fn cmd_nexus_sled_add(
...
+    args: &SledExpungeArgs,
+    omdb: &Omdb,
+    log: &slog::Logger,
+    _destruction_token: DestructiveOperationToken,
+) -> Result<(), anyhow::Error> {
+    // This is an extremely dangerous and irreversible operation. We put a
+    // couple of safeguards in place to ensure this cannot be called without
+    // due consideration:
+    //
+    // 1. We'll require manual input on stdin to confirm the sled to be removed
```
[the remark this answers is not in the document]
jgallagher: Yeah, agreed. @benjaminleonard [made some suggestions](https://github.com/oxidecomputer/omicron/issues/5134#issuecomment-1966326485) about how we might design the _external_ API to mimic this kind of confirmation flow, but I did not implement any of that for the _internal_ API. I put these guards on omdb, but (a) we can relax them (or provide flags to skip them) if needed, or (b) tests may be able to hit the internal API directly.

dev-tools/omdb/src/bin/omdb/nexus.rs
```diff
@@ -1048,3 +1085,109 @@ async fn cmd_nexus_sled_add(
...
+    // 2. We'll warn sternly if the sled-to-be-expunged is still present in the
+    //    most recent inventory collection
+    use nexus_db_queries::context::OpContext;
+
+    let datastore = args.db_url_opts.connect(omdb, log).await?;
+    let opctx = OpContext::for_tests(log.clone(), datastore.clone());
+    let opctx = &opctx;
+
+    // First, we need to look up the sled so we know its serial number.
```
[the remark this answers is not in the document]
jgallagher: I admit I don't know enough about how PC-based racks are set up to answer this question. I think PCs still have a serial number in the `sled` table, right? If their sled-agent reports the same serial number, I think both this and the inventory check will work.

dev-tools/omdb/src/bin/omdb/nexus.rs
jgallagher: We use it both for this and fetching the latest inventory, which doesn't have APIs either (but could). Using the database connection in this subcommand seems a _little_ sketchy to me, but I wasn't sure about adding sled info + inventory internal API commands just to support guard rails. I'll keep this as-is for now but it's easy to revisit (or replace if/when we develop those APIs for some other consumers).

dev-tools/omdb/src/bin/omdb/nexus.rs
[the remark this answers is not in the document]
andrewjstone: I think this should still work on PCs. This is how we are able to add and remove sleds on the a4x2 testbed.

PR #5238, opened 2024-03-09
[reconfigurator] replace Blueprint::zones_in_service with a disposition enum next to each zone

sunshowers:
This greatly simplifies the handling of the expunged disposition in #5211. It also is more understandable, and less prone to making mistakes.

This PR only changes the in-memory representation of what is currently `zones_in_service`. I've skipped doing the DB migration in this PR because it'll be easier to just do it wholesale in #5211.

There are some small differences in diff output that I think make sense. But I took a bigger swing at it in #5270, which depends on this.

TODO:

- [x] Fix code.
- [x] Turn `Blueprint`'s `Debug` impl into a `display_detailed()` method (done in #5248)
- [x] Add snapshot tests for diff output (done in #5248)
- [x] Fix tests.


dev-tools/reconfigurator-cli/tests/output/cmd-complex-stdout
```diff
@@ -80,51 +80,51 @@ comment: from collection 5e1d6ae0-c5b4-4884-ac41-680cd4f5762d
...
+    0db88baf-911d-47bb-af71-3aebd7c96934 in service     nexus [underlay IP fd00:1122:3344:103::4]
+    213d7dd3-95a4-45a7-99ca-e4899991dab4 in service     crucible [underlay IP fd00:1122:3344:103::6]
+    218f36a5-9e51-4b88-ad1a-da4d85c246c4 in service     internal_dns [underlay IP fd00:1122:3344:3::1]
+    40c47ead-3be8-4724-a8ef-b66233fc5c65 in service     crucible [underlay IP fd00:1122:3344:103::7]
+    425d9d2f-72ea-4f53-80b6-04e002771cdb in service     crucible [underlay IP fd00:1122:3344:103::9]
+    46a050b6-37b8-47b9-b4ae-ccb736ecd6d5 in service     crucible [underlay IP fd00:1122:3344:103::8]
+    4ea8104f-2cfc-4599-9d2c-bcfd8f0b71fd in service     clickhouse [underlay IP fd00:1122:3344:103::5]
+    7611a549-effc-41b6-b6cb-c03e0c70a31f in service     internal_ntp [underlay IP fd00:1122:3344:103::b]
+    af7ccead-2bd8-4a8a-9fbf-3b788d180139 in service     cockroach_db [underlay IP fd00:1122:3344:103::3]
+    d4428e41-ad7b-41f0-bde6-7c7c955f20c1 in service     crucible [underlay IP fd00:1122:3344:103::a]
```
sunshowers, line 92: As part of this PR I unified the way we display zones between blueprints and diffs. I tried to make the output relatively clean.

I tried to add padding to align all the fields, but then realized this would just be better as a table. I'll do that in a followup since this PR has gotten large enough already.

sunshowers: Oh also, I was able to keep the sled IDs the same by using a migration script for the JSON schema.

sunshowers: Tabular output is in https://github.com/oxidecomputer/omicron/pull/5270 -- it looks so much better, wow.

sunshowers: (While this output has been removed in `cmd-complex-stdout`, the general principle still applies. I'd rather we just focus on #5270 to improve the output, because it is much better there.)

nexus/db-queries/src/db/datastore/vpc.rs
```diff
@@ -1599,6 +1600,12 @@ mod tests {
                         external_dns_servers: Vec::new(),
                     },
                 };
+                let zone_config = BlueprintZoneConfig {
+                    config,
+                    // XXX: NotInService retains the previous test behavior --
+                    // we may wish to change this to InService.
+                    zone_policy: BlueprintZonePolicy::NotInService,
```
jgallagher, line 1607: Two thoughts:

* I suspect changing this won't affect the test...
* ... but maybe it should?

When we're resolving sleds that need firewall rules, should we include sleds that have "not in service" services? I am a little worried we've added machinery here around a policy that we're not actually using, so are wandering a bit from where we'd go if we were using it. (This is not a complaint about this PR; I strongly agree with its restructuring choices.)

nexus/reconfigurator/execution/src/omicron_zones.rs
```diff
@@ -149,31 +155,36 @@ mod test {
...
+                            http_address: "some-ipv6-address".into(),
                         },
-                        dns_address: "oh-hello-internal-dns".into(),
-                        gz_address: "::1".parse().unwrap(),
-                        gz_address_index: 0,
-                        http_address: "some-ipv6-address".into(),
                     },
+                    // XXX: NotInService retains the previous test behavior --
+                    // we may wish to change this to InService.
+                    zone_policy: BlueprintZonePolicy::NotInService,
```
jgallagher, line 187: Similar to above - I suspect this was just a test oversight, and we should change it to `InService`.

nexus/reconfigurator/execution/src/resource_allocation.rs
```diff
@@ -619,65 +615,56 @@ mod tests {
 
         // Build the `zones` map needed by `ensure_zone_resources_allocated`,
         // with an arbitrary sled_id.
```
jgallagher, line 617: Can remove this bit about an arbitrary sled_id now.

nexus/reconfigurator/planning/src/planner.rs
```diff
@@ -642,7 +642,7 @@ mod test {
         let zones = sled_changes.zones_added().collect::<Vec<_>>();
         assert_eq!(zones.len(), policy.target_nexus_zone_count - 1);
         for zone in &zones {
-            let OmicronZoneType::Nexus { .. } = zone.zone_type else {
+            let OmicronZoneType::Nexus { .. } = zone.config.zone_type else {
```
jgallagher, line 645: Only because the diffs made this obvious: should this be

```suggestion
            if !zone.config.zone_type.is_nexus() {
```

instead?

nexus/reconfigurator/planning/src/planner.rs
```diff
@@ -727,7 +727,8 @@ mod test {
                 }
             }
             for zone in &zones {
-                let OmicronZoneType::Nexus { .. } = zone.zone_type else {
+                let OmicronZoneType::Nexus { .. } = zone.config.zone_type
+                else {
                     panic!("unexpectedly added a non-Crucible zone: {zone:?}");
```
jgallagher, line 732: Same suggestion as above about using `zone_type.is_nexus()`, and can we fix the comment string? (Both of these were problems prior to the PR, almost certainly from me - sorry!)

nexus/src/app/background/blueprint_execution.rs
```diff
@@ -38,64 +38,68 @@ impl BlueprintExecutor {
...
-    fn activate<'a>(
-        &'a mut self,
-        opctx: &'a OpContext,
-    ) -> BoxFuture<'a, serde_json::Value> {
-        async {
-            // Get the latest blueprint, cloning to prevent holding a read lock
-            // on the watch.
-            let update = self.rx_blueprint.borrow_and_update().clone();
+    /// Implementation for `BackgroundTask::activate` for `BlueprintExecutor`,
+    /// added here to produce better errors.
```
jgallagher, line 43: I don't follow this comment - it looks like the errors are the same?

nexus/types/src/deployment.rs
```diff
@@ -271,31 +295,159 @@ impl<'a> fmt::Display for BlueprintDisplay<'a> {
...
+impl BlueprintZonePolicy {
+    /// The maximum width of `Display` output.
+    const DISPLAY_WIDTH: usize = 14;
+}
+
+impl fmt::Display for BlueprintZonePolicy {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            // Neither `write!(f, "...")` nor `f.write_str("...")` obey fill
+            // and alignment (used above), but this does.
```
jgallagher, line 444: What! So writing a correct Display impl requires knowing whether your callers will use width and alignment modifiers?

jgallagher: TIL: yes, yes it does.

jgallagher (approved):
Thanks, this is much cleaner. And my apologies that most of my comments apply to the existing code this PR touches :grimacing:.

nexus/db-queries/src/db/datastore/vpc.rs
sunshowers: We discussed this in more detail in #oxide-update today. The conclusion that we came to is to fail to compile on the variants changing, for now, to be isomorphic to the current state. In #5211 we have to do a DB migration anyway, and there we may do a greater rethink of how this works (have a call about this tomorrow).

sunshowers: Changed this to `InService` though!

nexus/reconfigurator/execution/src/omicron_zones.rs
sunshowers: Also fixed, and added a check that will cause this code to fail to compile when a new variant is added.

nexus/reconfigurator/planning/src/planner.rs
sunshowers: Done, and also added `is_crucible` and changed another use of this pattern elsewhere.

nexus/reconfigurator/planning/src/planner.rs
sunshowers: Fixed.

nexus/types/src/deployment.rs
sunshowers: :)

nexus/src/app/background/blueprint_execution.rs
sunshowers: Fixed.

sunshowers:
I've put in the changes we discussed today in this PR:

* Renamed `BlueprintZonePolicy` to `BlueprintZoneDisposition`.
* Added a `BlueprintZoneFilter` enum which can be used to query sets of dispositions in various ways.

I also have a question: Should `all_omicron_zones` also take a `BlueprintZoneFilter` as an argument? I'd do that in a subsequent PR since the blast radius of that would be bigger.

I think this will benefit from another round of review since the changes are substantial. GitHub code review change tracking is completely useless, so [here's an interdiff](https://gist.github.com/sunshowers/91e49f2a70a3f7a1c517a259004f5e3a) generated by Jujutsu of the changes since last review.

nexus/reconfigurator/execution/src/omicron_zones.rs
```diff
@@ -21,8 +22,17 @@ use uuid::Uuid;
 pub(crate) async fn deploy_zones(
     opctx: &OpContext,
     sleds_by_id: &BTreeMap<Uuid, Sled>,
-    zones: &BTreeMap<Uuid, OmicronZonesConfig>,
+    zones: &BTreeMap<Uuid, BlueprintZonesConfig>,
 ) -> Result<(), Vec<anyhow::Error>> {
+    // Make this code fail to compile if a new zone state is added.
```
jgallagher, line 27: I don't think this needs to change on this PR, but related to the question about whether `all_omicron_zones` should take a filter: is the point of this check that when we add a new state, the implementation may need to change in unknown ways? Is there a way to restructure the inputs to this so that we don't feel the need for this guard?

nexus/reconfigurator/execution/src/omicron_zones.rs
```diff
@@ -259,26 +278,31 @@ mod test {
...
+                    zone_type: OmicronZoneType::InternalNtp {
+                        address: "::1".into(),
+                        dns_servers: vec!["::1".parse().unwrap()],
+                        domain: None,
+                        ntp_servers: vec!["some-ntp-server-addr".into()],
+                    },
                 },
+                // XXX: NotInService retains the previous test behavior -- we
+                // may wish to change this to InService.
+                disposition: BlueprintZoneDisposition::Quiesced,
```
jgallagher, line 296: Can we change this one to `InService` too?

nexus/reconfigurator/planning/src/blueprint_builder.rs
```diff
@@ -868,14 +874,12 @@ impl<'a> BlueprintZones<'a> {
...
                     // and we haven't added any zones to it yet.  Use the
                     // standard initial config.
-                    .unwrap_or_else(|| OmicronZonesConfig {
+                    .unwrap_or_else(|| BlueprintZonesConfig {
                         generation: Generation::new(),
                         zones: vec![],
                     });
 
-                // This is not strictly necessary.  But for testing, it's
-                // helpful for things to be in sorted order.
```
jgallagher, line 877: Can we keep this comment?

nexus/src/app/background/blueprint_execution.rs
```diff
@@ -229,26 +236,32 @@ mod test {
...
+                            http_address: "some-ipv6-address".into(),
                         },
-                        dns_address: "oh-hello-internal-dns".into(),
-                        gz_address: "::1".parse().unwrap(),
-                        gz_address_index: 0,
-                        http_address: "some-ipv6-address".into(),
                     },
+                    // XXX: NotInService retains the previous test behavior --
+                    // we may wish to change this to InService.
+                    disposition: BlueprintZoneDisposition::Quiesced,
```
jgallagher, line 260: Can we change this to `InService`?

nexus/types/src/deployment.rs
```diff
@@ -271,31 +300,231 @@ impl<'a> fmt::Display for BlueprintDisplay<'a> {
...
+        };
+        ret.sort();
+
+        ret
+    }
+
+    /// Sorts the list of zones stored in this configuration.
+    ///
+    /// This is not strictly necessary. But for testing, it's helpful for
+    /// things to be in sorted order.
```
jgallagher, line 366: Ah, I see the comment moved to here. I think it may make more sense at the call site? It's possible (although unlikely and I hope we don't) we could introduce some _necessary_ `.sort()` call at some point in the future.

nexus/types/src/deployment.rs
```diff
@@ -271,31 +300,231 @@ impl<'a> fmt::Display for BlueprintDisplay<'a> {
...
+/// Filters that apply to blueprint zones.
+///
+/// This logic lives here rather than within the individual components making
+/// decisions, so that this is easier to read.
+///
+/// The meaning of a particular filter should not be overloaded -- each time a
+/// new use case wants to make a decision based on the zone disposition, a new
+/// variant should be added to this enum.
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
+pub enum BlueprintZoneFilter {
```
jgallagher, line 514: Making this a separate type is interesting; I don't think I have strong feelings about it either way, but it's not what I expected. I thought we'd have methods on `BlueprintZoneDisposition` itself like `.in_internal_dns()`, `.needs_firewall_rules()`, etc., and that the filter we pass to `all_blueprint_zones` would be a `BlueprintZoneDisposition -> bool` closure that we'd expect to be a call to one of those methods. By taking an explicit filter, though, it does prevent callers from doing something goofy in that closure, and maybe forces them to add a specific filter if one of the existing ones isn't right for them?

jgallagher (approved):

jgallagher:
> I also have a question: Should `all_omicron_zones` also take a `BlueprintZoneFilter` as an argument? I'd do that in a subsequent PR since the blast radius of that would be bigger.

After reading the changes, I think I would very strongly vote "yes", and at that point we could probably drop `all_blueprint_zones` altogether? It looks like all the current callers of that are only using it because `all_omicron_zones` _doesn't_ take a filter.

nexus/types/src/deployment.rs
sunshowers: The separate type was not actually  the first implementation, but I realized it provided a really nice orthogonality to the API. With just a single implementation, we got 3/4 sets of methods:

* matches
* all_matching
* all_blueprint_zones
* and in the future, all_omicron_zones

You could do this with a callback, but the explicit filter makes it so that new callers have to add their condition here rather than doing something bespoke in the caller (e.g. == in-service).

nexus/reconfigurator/execution/src/omicron_zones.rs
sunshowers: Yeah, good question -- let me look at this a bit.

sunshowers:
> at that point we could probably drop all_blueprint_zones altogether?

Depends, I think some of the users want the full `BlueprintZoneConfig`. But it can definitely be de-emphasized.

nexus/reconfigurator/execution/src/omicron_zones.rs
sunshowers: I've changed `to_omicron_zones_config` to accept a filter argument, which feels quite satisfying.

nexus/reconfigurator/execution/src/omicron_zones.rs
sunshowers: Done. I guess we should test that quiesced zones get deployed too, though.

sunshowers: (so I've expanded the test to make it so that one zone is in-service and the other is quiesced -- in the future, we can also add a test for expunged.)

nexus/reconfigurator/planning/src/blueprint_builder.rs
sunshowers: Put back the part of the comment starting with "For testing..." -- left the full comment on the doc comment.

nexus/src/app/background/blueprint_execution.rs
sunshowers: Similar to `omicron_zones.rs`, expanded the test to check both in-service and quiesced zones.

nexus/types/src/deployment.rs
sunshowers: Left the comment in for now -- hopefully anyone adding the future necessary `sort()` will see the comment.

In #5270 I'm actually going to change the sort order to (zone kind, UUID), so all zones with the same kind are grouped together in output. That sort order may not make sense for non-test use cases.

PR #5239, opened 2024-03-09
test environment could be more realistic

davepacheco:
This PR makes a few changes to the environment that gets set up with a `ControlPlaneTestContext` in order to make it more realistic:

- Fix up the Omicron zones to match up with simulated sled agents: for the components that we start up (e.g., Clickhouse, CockroachDB, etc.), we now create associated `OmicronZoneConfig` objects and `PUT` them to the corresponding sled agent.  The net result of this is that an inventory collected from the system comes much closer to reflecting what's actually present.
- Fix up the switch zone components to match up with simulated sled agents:
    - Prior to this, we started up two Dendrite instances and two MGD instances.  But we only started one simulated sled agent and it reported itself as a Gimlet, not a Scrimlet.  We also started an MGS, but it was not lined up with any of this.
    - Now, we start up two simulated sled agents and they both report as Scrimlets.  (The second Scrimlet is marked non-provisionable because a bunch of tests assumed there was only one provisionable sled and I didn't think it was useful to fix that here.)
    - Now, there's one each of Dendrite, MGS, and MGD on each of the two simulated Scrimlet sled agents.

The net result is that inventory collections are more like real systems and so the resulting internal DNS is also more like a real system.  There's nothing in this PR that relies on that but there's a test in #5212 that does and that's how I tested this.


smklein (approved):
Makes a lot of sense to me, I was also poking at `nexus/test-utils/src/lib.rs` and wondering how to make the dependencies on the inventory "more real" in our simulated environment.

PR #5242, opened 2024-03-11
Add Error::not_found variant

jmpesp:
Add a variant to our external Error so that a 404 can be returned without requiring a ResourceType - adding one may not make sense for a particular endpoint.

### DIFF
--- a/common/src/api/external/error.rs
+++ b/common/src/api/external/error.rs
@@ -71,6 +71,11 @@ pub enum Error {
 
     #[error("Conflict: {}", .message.display_internal())]
     Conflict { message: MessagePair },
+
+    /// A generic 404 response. If there is an applicable ResourceType, use
+    /// ObjectNotFound instead.
+    #[error("Not found: {}", .message.display_internal())]
+    NotFound { message: MessagePair },
 }
 
 /// Represents an error message which has an external component, along with
@@ -199,6 +204,7 @@ impl Error {
             | Error::InsufficientCapacity { .. }
             | Error::InternalError { .. }
             | Error::TypeVersionMismatch { .. }
+            | Error::NotFound { .. }
             | Error::Conflict { .. } => false,
         }
     }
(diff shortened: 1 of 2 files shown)


common/src/api/external/error.rs
```diff
@@ -300,6 +306,15 @@ impl Error {
         Error::Conflict { message: MessagePair::new(message.into()) }
     }
 
+    /// Generates an [`Error::NotFound`] with a specific message.
+    ///
+    /// This is used in cases where a generic 404 is required. For cases where
+    /// there is a ResourceType, use a function that produces
+    /// [`Error::ObjectNotFound`] instead.
+    pub fn not_found(message: impl Into<String>) -> Error {
```
jgallagher, line 314: How picky/careful do we want to be about trying to help avoid misuse of this? E.g., would it make sense to name this something more annoying like `non_resource_not_found` to make it stand out at call sites?

jgallagher (approved):
I agree with @smklein's sidebar from chat:

> Part of me wonders if we're using the "external omicron error type" more than we should? Our inability to return reasonable info about internal errors through the "external error type" kinda makes sense, they don't necessarily represent the same thing

but this seems like a reasonable workaround that doesn't require major rework of errors :man_shrugging:

davepacheco:
This probably isn't a great place to put this but I'm not sure where else to put it.

> I agree with @smklein's sidebar from chat:
> 
> > Part of me wonders if we're using the "external omicron error type" more than we should? Our inability to return reasonable info about internal errors through the "external error type" kinda makes sense, they don't necessarily represent the same thing

I think it's probably true that we use that Error type much more ubiquitously than is useful.

I think it's really valuable for failures deep inside Nexus (like, in stuff that doesn't necessarily even know about Nexus, like the database connection pool) to be _able_ to produce a relatively uniform set of error kinds that allow us to uniformly classify any failure as "a transient blip" (503) vs. "an unexpected condition that we can't handle" (500), etc _and_ turn it into a fairly uniform log message / event / user-facing message.  This is of course useful for things that get used in API requests, but also for background automation like Reconfigurator to know how to respond to the failure.  The original thinking here was something like: these variants in `Error` are pretty broadly applicable -- everything can probably classify its failures as one of these things.

But as we're observing, that doesn't mean they have to all return this specific `Error` type directly.  Subsystems could totally define their own Error types and use those.  But I think most of those would need to be convertible to `Error` in a way that preserves the same level of information (e.g., was it a client problem or a server problem, is it a transient failure, was it an unhandleable unexpected condition, etc.).  Unfortunately, we've already had lots of places where we used our own error type and _didn't_ do this -- we just did `map_err(|e| Error::internal_error(...))`.  In my experience, failing to classify errors like this correctly has caused a lot of pain when we run into these conditions in the field.

The good news is that anybody can start using more fine-grained Error types with appropriate conversions today in any subsystem, new or old.  It doesn't have to be an all-or-nothing rewrite.

common/src/api/external/error.rs
jmpesp: agreed, done in 414298e89

PR #5244, opened 2024-03-12
RSS <-> Nexus handoff: Set an initial (disabled) target blueprint

### LINKED ISSUE #5222: every new system should have a target blueprint
Blueprints represent the canonical configuration of various subsystems, like the set of Omicron zones deployed on sleds.  Every system should have one.

On today's ad hoc update call we discussed a plan where RSS would construct a Blueprint describing its ServicePlan and pass that off to Nexus during the initial handoff.  This would then be made the initial target blueprint, presumably disabled.  This way, we could always assume a system has a target blueprint.

jgallagher:
I would like to test that after an RSS handoff, the initial blueprint we created in RSS and set in the Nexus handoff matches the blueprint we would generate from the first inventory collections. I can do that by hand on `madrid`, but I'm not sure there's a good way to do that in an automated test. Is this something I could lean on a4x2 to check?

This does not modify the `services` table or the `services` field of the RSS handoff. Removing those will come in a followup PR that will need an accompanying note for deployed systems (i.e., instructions to set a blueprint so the current users of `services` will continue to function by checking the current target blueprint).

Closes #5222.


nexus/db-queries/src/db/datastore/rack.rs
```diff
@@ -629,6 +641,46 @@ impl DataStore {
                         })?;
                     }
 
+                    // Insert the RSS-generated blueprint.
+                    Self::blueprint_insert_on_connection(
+                        &conn,
+                        opctx,
+                        &blueprint,
+                        ).await
```
smklein, line 649: Huh, interesting indentation choices by rustfmt

nexus/db-queries/src/db/datastore/rack.rs
```diff
@@ -629,6 +641,46 @@ impl DataStore {
...
+                    // when more of Reconfigurator is automated, but for now we
+                    // require a support operation to enable it.
+                    Self::blueprint_target_set_current_on_connection(
+                        &conn,
+                        opctx,
+                        BlueprintTarget {
+                            target_id: blueprint.id,
+                            enabled: false,
+                            time_made_target: Utc::now(),
+                        },
```
smklein, line 670: We wouldn't really expect anything to be able to concurrently set a *different* target while we're doing this operation, would we? As in, other than this pathway, is "setting the target" still an operator/developer-driven operation?

Why I ask: I'm just curious if some other background task in Nexus could attempt to concurrently set a blueprint and cause this `blueprint_target_set` call to fail

sled-agent/src/rack_setup/service.rs
```diff
@@ -1084,6 +1066,113 @@ impl ServiceInner {
...
+
+impl DeployStepVersion {
+    const V1_NOTHING: Generation = OmicronZonesConfig::INITIAL_GENERATION;
+    const V2_DNS_ONLY: Generation = Self::V1_NOTHING.next();
+    const V3_DNS_AND_NTP: Generation = Self::V2_DNS_ONLY.next();
+    const V4_COCKROACHDB: Generation = Self::V3_DNS_AND_NTP.next();
+    const V5_EVERYTHING: Generation = Self::V4_COCKROACHDB.next();
+}
+
+fn build_initial_blueprint_from_plan(
```
smklein, line 1102: Pretty stoked to see this here, bridging the gap between RSS and Blueprints.

Do we have any issues tracking "Eventually, we'd like the ServicePlan to be generated by the reconfigurator"? Do we still think that's possible?

sled-agent/src/rack_setup/service.rs
```diff
@@ -1084,6 +1066,113 @@ impl ServiceInner {
...
+        entry.insert(sled_config.clone());
+    }
+
+    Ok(build_initial_blueprint_from_sled_configs(
+        sled_configs,
+        internal_dns_version,
+    ))
+}
+
+pub(crate) fn build_initial_blueprint_from_sled_configs(
```
smklein, line 1135: This ended up much simpler than I expected it would be. Good to know that our blueprints aren't *that* far from the RSS outputs.

smklein (approved):

sled-agent/src/rack_setup/service.rs
jgallagher: I don't think we have an issue, but we should. I do think it's possible (and probably even _preferred_, especially if RSS could start a much more minimal set of services, handoff, and let Reconfigurator do the rest).

sled-agent/src/rack_setup/service.rs
jgallagher: Same! I was pleasantly surprised.

nexus/db-queries/src/db/datastore/rack.rs
jgallagher: Found it. This line went past 80 columns in an unbreakable way, so we hit https://github.com/rust-lang/rustfmt/issues/3863 and rustfmt gave up on everything around:

```
                       <Zpool as DatastoreCollection<Dataset>>::insert_resource(
```

`Zpool` only implements `DatastoreCollection` once, so I removed the cast and rustfmt did its thing in aac78b0b2

nexus/db-queries/src/db/datastore/rack.rs
```diff
@@ -629,6 +646,44 @@ impl DataStore {
                         })?;
                     }
 
+                    // Insert the RSS-generated blueprint.
+                    Self::blueprint_insert_on_connection(
+                        &conn, opctx, &blueprint,
+                    )
+                    .await
+                    .map_err(|e| {
+                        warn!(
```
davepacheco, line 655:
```suggestion
                        error!(
```

nexus/db-queries/src/db/datastore/rack.rs
```diff
@@ -629,6 +646,44 @@ impl DataStore {
...
+                        opctx,
+                        BlueprintTarget {
+                            target_id: blueprint.id,
+                            enabled: false,
+                            time_made_target: Utc::now(),
+                        },
+                    )
+                    .await
+                    .map_err(|e| {
+                        warn!(
```
davepacheco, line 678:
```suggestion
                        error!(
```

nexus/db-queries/src/db/datastore/rack.rs
```diff
@@ -629,6 +646,44 @@ impl DataStore {
...
+                            log,
+                            "Initializing Rack: \
+                                 Failed to set blueprint as target"
```
davepacheco, line 681: Can we include the error details here?

nexus/db-queries/src/db/datastore/rack.rs
```diff
@@ -629,6 +646,44 @@ impl DataStore {
...
 
+                    // Insert the RSS-generated blueprint.
+                    Self::blueprint_insert_on_connection(
+                        &conn, opctx, &blueprint,
+                    )
+                    .await
+                    .map_err(|e| {
+                        warn!(
+                            log,
+                            "Initializing Rack: Failed to insert blueprint"
```
davepacheco, line 657: Can we include the error details here?

sled-agent/Cargo.toml
```diff
@@ -46,6 +46,7 @@ macaddr.workspace = true
 mg-admin-client.workspace = true
 nexus-client.workspace = true
 nexus-config.workspace = true
+nexus-types.workspace = true
```
davepacheco, line 49: This feels like potentially a big deal...but seems fine!  (I say that because `nexus_types` I _think_ is mostly implementation details of Nexus and it seems a little sketchy to leak that outside?  But it doesn't seem worth creating a separate package for this right now.)

Come to think of it, it seems like this shouldn't be necessary because RSS would be using a progenitor-generated type.  Is this necessary because our internal Nexus client uses `replace` to use this type instead of generating one?  (And if I recall correctly, that's because we have functionality like `diff` that we want to use in omdb.  Ugh.)

sled-agent/src/rack_setup/service.rs
```diff
@@ -1084,6 +1066,113 @@ impl ServiceInner {
...
+            .context("invalid internal dns version")?;
+
+    let mut sled_configs = BTreeMap::new();
+    for sled_request in sled_plan.sleds.values() {
+        let sled_addr = get_sled_address(sled_request.body.subnet);
+        let sled_id = sled_request.body.id;
+        let entry = match sled_configs.entry(sled_id) {
+            btree_map::Entry::Vacant(entry) => entry,
+            btree_map::Entry::Occupied(_) => {
+                bail!("duplicate sled address deriving blueprint: {sled_addr}");
```
davepacheco, line 1117:
```suggestion
                bail!("duplicate sled address found while deriving blueprint: {sled_addr}");
```

sled-agent/src/sim/server.rs
```diff
@@ -450,7 +489,16 @@ pub async fn run_standalone_server(
         None => vec![],
     };
 
+    let services =
```
davepacheco, line 492: If I'm understanding right, we're still constructing `Service`s for these zones and still inserting those in the db.  Is that right?  (I think that's the right call until we finish #4947.)

davepacheco (approved):

sled-agent/src/sim/server.rs
jgallagher: That's right. I think there are three steps:

1. Ensure new systems start off with a blueprint set (this PR)
2. Ensure existing systems get a blueprint set (support operation)
3. Remove the `service` table and all the fallout from int (including dropping `Service`s from the RSS handoff) will be a followup PR that can be in flight or even landed concurrently with 2, as long as we do 2 before deploying 3

nexus/db-queries/src/db/datastore/rack.rs
jgallagher: I rolled all these log changes up (including other rack errors that weren't in the diff) into [f1acada](https://github.com/oxidecomputer/omicron/pull/5244/commits/f1acada1266311c7ae518ff570d69c842dcecb01)

sled-agent/Cargo.toml
jgallagher: Yes, it's the `replace` directive on `Blueprint` specifically that causes this. That is kind of a dicey thing to be `replace`ing, especially as we're actively changing it. :-/

jgallagher:
I ran through RSS on madrid and everything came out as expected. We had a disabled blueprint:

```
root@oxz_switch1:~# omdb nexus blueprints list
note: using Nexus URL http://[fd00:1122:3344:102::4]:12221
T ENA ID                                   PARENT TIME_CREATED
* no  30fc006d-5db7-4c9e-85d3-954100ad7c8e <none> 2024-03-14T16:20:51.377Z
```

Running `blueprints show` on it confirmed it came from RSS:

```
time_created: 2024-03-14T16:20:51.377863Z, creator: "RSS", comment: "initial blueprint from rack setup"
```

I then generated a blueprint from a collection, and the diff showed no changes between the blueprint from RSS and the blueprint from the collection.

sled-agent/src/rack_setup/service.rs
```diff
@@ -1084,6 +1066,123 @@ impl ServiceInner {
...
+        zones_in_service,
+        parent_blueprint_id: None,
+        internal_dns_version,
+        // We don't actually configure external DNS during RSS. During handoff,
+        // Nexus creates an initial empty external DNS configuration, then
+        // creates the recovery silo, which bumps the generation.
+        //
+        // TODO-john How should we do this? Baking this knowledge of how Nexus
+        // sets up external DNS here seems bad.
+        external_dns_version: Generation::new().next(),
```
jgallagher, line 1179: @davepacheco Do you have thoughts on this change I made getting caught up with `main`? A few obvious options:

* This is a little gross but not the end of the world
* We export an "external DNS generation after rack setup" from `nexus-db-queries` and use that here
* We set this to `Generation::new()` and let the planner sort out that there's a new external DNS later

davepacheco: What about:
- we set it to `Generation::new()` here
- during rack initialization, Nexus sets it to whatever it wants after receiving this blueprint but before writing it to the database

The explanation would be: as far as RSS is concerned, it's generation 1 -- the oldest possible generation should be a safe bet and it doesn't know any better.  When Nexus receives it, it's in a position to say: "okay, RSS handed me this initial blueprint, but it didn't know I was going to write an initial external DNS generation, so I'm going to adjust that here based on the actions I'm taking [which are: creating generation 2 of external DNS]".  What do you think?

jgallagher: I like that a lot - thanks! Done in [0b28298](https://github.com/oxidecomputer/omicron/pull/5244/commits/0b2829874a83445514e7124545f0866d796c8f3a)

PR #5245, opened 2024-03-12
Split out a `sled-hardware-types` crate

jgallagher:
On main as of #5158, we unexpectedly get some binaries depending on `libnvme` that shouldn't. In release builds:

```
installinator
omicron-dev
omicron-package
services-ledger-check-migrate
sled-agent
sled-agent-sim
wicketd
```

and in debug builds, all of the above plus `omdb`. We don't really care about this for binaries that don't run on the rack, so stripping the list down to binaries we do care about:

```
installinator
omdb (debug only)
sled-agent
wicketd
```

It's correct and expected that installinator and sled-agent depend on libnvme, but omdb shouldn't (and doesn't in release), and wicketd _must not_, as libnvme isn't available in the switch zone.

This PR fixes that incorrect dependency by splitting the parts of the `sled-hardware` crate that wicketd and omdb depended on (directly or transitively) into a new `sled-hardware-types` crate. On this PR, we are left with only the following needing libnvme, in both debug and release:

```
installinator
omicron-dev
omicron-package
services-ledger-check-migrate
sled-agent
sled-agent-sim
```

I assume with a bit more work we could trim out everything but `installinator` and `sled-agent`, and that might be worth doing, but we'd like to land this ASAP so as to not break any updates performed off of `main`. Separately, we could also imagine a CI check that we don't have unexpected library dependencies present in the final binaries; @papertigers is going to work on that.


papertigers (approved):
lgtm thanks again for the assist on this.

For future reference the way we were checking where this library was needed was with `elfedit -r -e 'dyn:tag NEEDED' /path/to/file`.

Then combined with find:
```
mike - atrium ~/src/omicron (git:john/sled-hardware-types) $ fd --max-depth=1 -t f . ./target/release/ -x bash -c "elfedit -r -e 'dyn:tag NEEDED' \{} 2>&1 | rg -v elfedit | rg libnvme && echo \{}"
       [1]  NEEDED            0x8b956a            libnvme.so.1
./target/release/sled-agent
       [1]  NEEDED            0x7e2b7a            libnvme.so.1
./target/release/services-ledger-check-migrate
       [0]  NEEDED            0x4d1be1            libnvme.so.1
./target/release/installinator
       [1]  NEEDED            0x895f0f            libnvme.so.1
./target/release/sled-agent-sim
       [9]  NEEDED            0x17d5ceb           libnvme.so.1
./target/release/omicron-dev
       [0]  NEEDED            0x3f5431            libnvme.so.1
./target/release/omicron-package
```

PR #5249, opened 2024-03-13
[nexus] Add zpool to inventory

smklein:
This PR adds zpools to the inventory, and updates the existing `omicron.public.zpool` table to only include aspects of the zpool which are fully controlled by Nexus.

This PR is intended to solve a problem introduced by https://github.com/oxidecomputer/omicron/pull/5172 . In that PR, physical disk and zpool provisioning will be decided by Nexus, and sent to the Sled Agent as a request. As a result, the `total_size` value of zpools is not known when a pool is provisioned. Here, we add zpools to the inventory, and JOIN with that table to access the "total_size" of the zpool if it is known.

Additionally, this PR adds a test to validate that the database (appropriately) throws "Insufficient capacity" errors if the Zpool exists, but has not appeared in the inventory.


nexus/db-model/src/zpool.rs
```diff
@@ -27,25 +27,28 @@ pub struct Zpool {
...
+    pub total_size: Option<ByteCount>,
 }
 
 impl Zpool {
     pub fn new(
         id: Uuid,
         sled_id: Uuid,
         physical_disk_id: Uuid,
-        total_size: ByteCount,
+        _total_size: ByteCount,
```
smklein, line 43: This callsite should be cleaned up, if we're really going the "get-size-from-inventory" route

nexus/db-queries/src/db/datastore/zpool.rs
```diff
@@ -32,6 +33,41 @@ use omicron_common::api::external::ResourceType;
 use uuid::Uuid;
 
 impl DataStore {
+    // TODO: Should this be paginated?
+    // TODO: Alternatively, should we have a better way of "inferring what needs
+    // to be updated" from inventory?
+    pub async fn zpool_update_all(
```
smklein, line 39: I feel a little uneasy about this invocation.

Basically, we have a big set of "all zpools" from inventory, and a big set of "all zpools" in the DB we're using as "control plane objects". This query updates all the live zpools using information from inventory, all at once.

I could see this as useful in the future for flagging zpools with "high" usage (regardless of their "provisioned" usage), or zpools that have faulted have have a state other than `ONLINE`. 

However, for just setting `total_size`, a value which **should** only be set once, it feels a little heavy-handed?

I think I'd only want to move forward with this PR if we think this other "dynamic zpool information informing database data" pattern seems useful generally.

smklein:
This PR basically implements the following:
image

In which Nexus can "PUT" zpools to the sled before knowing their size, and relies on the inventory (+ this background task) to update zpool information via scanning.

But I'm wondering if we'd prefer to just do the following instead:
image

In which we avoid the background task altogether, and simply await "all relevant zpool info" in the response to `omicron_physical_disks` PUT.

smklein:
Chatted with @davepacheco a little about this PR, here's our consensus:

- Keep "control plane zpool info" separate from "reported by inventory zpool info". This means no need for a background task to synchronize them.
- Make sure the inventory representation of "zpool info" has a timestamp, and an index by that timestamp.
- Anyone interested in observing the inventory info (like the region allocation query) can look for "the latest value of that zpool info from the inventory". It's okay if this is slightly stale.

nexus/db-queries/src/db/queries/region_allocation.rs
```diff
@@ -316,7 +316,11 @@ impl CandidateZpools {
         // into a BigInt, not a Numeric. I welcome a better solution.
         let it_will_fit = (old_zpool_usage::dsl::size_used
             + diesel::dsl::sql(&zpool_size_delta.to_string()))
-        .le(diesel::dsl::sql(zpool_dsl::total_size::NAME));
+        .le(diesel::dsl::sql(
```
smklein, line 319: Hours spent trying to make this work with Diesel, getting pushed away from subqueries, fighting aliases, and then fumbling with full-table scans as I tried to split this into a separate query:

~4

I am still motivated to make https://github.com/oxidecomputer/omicron/pull/5063 work after getting through this disk expungement stuff.

nexus/src/lib.rs
```diff
@@ -351,6 +353,17 @@ impl nexus_test_interface::NexusServer for Server {
             .unwrap();
     }
 
+    async fn inventory_collect_and_get_latest_collection(
```
smklein, line 356: This API is part of the `nexus_test_interface` API, and is basically punched through so tests which create synthetic disks can also "cause inventory collection to happen". They need it, to get the `inv_zpool` table populated.

nexus/db-queries/src/db/datastore/zpool.rs
smklein: I went ahead and removed this, instead, we're joining with the latest `inv_zpool` row.

smklein:
I need to take a lap through the instructions here for fixing the reconfigurator "test_complex" test (see: https://github.com/oxidecomputer/omicron/pull/5145#issuecomment-1992376219), but this is otherwise ready for review

smklein:
Also, for completeness, we ended up with an implementation like this, rather than the sequence diagrams above:

image

nexus/db-queries/src/db/datastore/mod.rs
```diff
@@ -1156,6 +1198,103 @@ mod test {
         logctx.cleanup_successful();
     }
 
+    #[tokio::test]
```
andrewjstone, line 1201: Very readable test!

schema/crdb/dbinit.sql
```diff
@@ -2986,6 +2986,28 @@ CREATE TABLE IF NOT EXISTS omicron.public.inv_physical_disk (
...
+    inv_collection_id UUID NOT NULL,
+    -- when this observation was made
+    time_collected TIMESTAMPTZ NOT NULL,
+
+    -- The control plane ID of the zpool
+    id UUID NOT NULL,
+    sled_id UUID NOT NULL,
+    total_size INT NOT NULL,
+
+    -- FK consisting of:
```
andrewjstone, line 3001:
```suggestion
    -- PK consisting of:
```

andrewjstone (approved):

nexus/db-queries/src/db/datastore/mod.rs
```diff
@@ -661,17 +661,59 @@ mod test {
...
+        .await;
+
+        add_test_zpool_to_inventory(datastore, zpool_id, sled_id).await;
+
+        zpool_id
+    }
+
+    // Creates a test zpool, returns its UUID.
+    //
+    // However, don't add the zpool to the inventory just yet.
```
sunshowers, line 678:
```suggestion
    // However, doesn't add the zpool to the inventory just yet.
```

sunshowers (approved):
Looks fantastic, thanks!

PR #5258, opened 2024-03-13
Current IP pool utilization endpoint

### LINKED ISSUE #5189: View current IP pool utilization (CRDB)
We want to be able to show IP pool utilization alongside compute, memory, and storage. Miniminally this could simply be a count of IPs allocated vs. the total number in a given pool, but ideally we would also break down ephemeral vs. floating vs. SNAT.

Note this issue does **not** call for tracking allocations in real time as IPs are allocated and deallocated, which would allow us to show a timeseries chart in the web console. That second stage is tracked in #5190.

david-crespo:
Closes #5189

The strangest thing here is that IPv6 `allocated` and `capacity` are both u128 that we put in the JSON spec as `"type": "string", "format": "uint128"`, where `uint128` is a custom format our client generators will have to look for and handle. `capacity` is huge because IPv6 ranges are huge. IPv4 allocated and capacity are `u32`.

~~The strangest thing here is the fact that we have to handle the unlikely possibility that an IPv6 range is too large for us to serialize its length in a JSON response. IPv6 range size is a `u128`, and we seem to be capping our numbers in API responses at `u32`. My proposal for handling this is simply to make the field nullable in the response, and we make clear that a null value means it was too large to serialize. I think this is fine because this should only happen if the operator does something really bad, and it is well understood that the operator can mess up networking with a misconfiguration. Another approach would be to always return the size as a string, but I think given the unlikelihood of the number being too large (and the fact that it's pathological if it is) means returning a numeric string puts undue burden on clients to handle it.~~


nexus/db-queries/src/db/datastore/utilization.rs
```diff
@@ -62,4 +64,191 @@ impl DataStore {
...
+            .count()
+            // TODO: how much do I have to worry about this being bigger than
+            // u32? it seems impossible, but do I have to handle it?
+            .first_async::<i64>(&*self.pool_connection_authorized(opctx).await?)
+            .await
+            .map_err(|e| public_error_from_diesel(e, ErrorHandler::Server))
+    }
+
+    // TODO: should this just return the vec of ranges and live in ip_pool.rs?
+    // TODO: generally we never retrieve all of anything, how bad is that
```
david-crespo, line 90: 👁️ 👁️

nexus/db-model/src/utilization.rs
```diff
@@ -55,3 +55,21 @@ impl From<SiloUtilization> for views::Utilization {
...
+
+impl From<IpPoolUtilization> for views::IpPoolUtilization {
+    fn from(util: IpPoolUtilization) -> Self {
+        Self {
+            allocated: util.allocated.try_into().ok(),
+            // if u128 is too big to convert to u32, fall back to None (null)
+            total: util.total.try_into().ok(),
+        }
+    }
+}
```
david-crespo, line 75: Will add tests here that exercise the boundary between convertible to u32 and not.

nexus/tests/integration_tests/ip_pools.rs
```diff
@@ -757,6 +759,60 @@ async fn create_pool(client: &ClientTestContext, name: &str) -> IpPool {
...
+    object_create::<IpRange, IpPoolRange>(client, &add_url, &range).await;
+
+    assert_ip_pool_utilization(client, "p0", 0, 5).await;
+
+    // now lets add the maximum number of addresses we can and still get a total
+    // in the response. that's FFFF FFFF minus five for the above range
+    let big_range = IpRange::V6(
+        Ipv6Range::new(
+            std::net::Ipv6Addr::new(0xfd00, 0, 0, 0, 0, 0, 0, 1),
+            std::net::Ipv6Addr::new(0xfd00, 0, 0, 0, 0, 0, 0xffff, 0xfffa),
```
david-crespo, line 793: Reconsidering whether this situation would really be all that unusual once we support IPv6 ranges.

bnaecker:
It probably will not be unusual at all. The IPv6 space is so huge it’s almost laughable. The most common subnet size is a /64, and technically you’re not even supposed to carve out smaller ones at all. We will certainly want a way to report the utilization of ranges up to the full /48 or whatever that we carve out for each VPC. On 13 Mar 2024, at 20:18, David Crespo ***@***.***> wrote:﻿
@david-crespo commented on this pull request.



In nexus/tests/integration_tests/ip_pools.rs:
> +        Ipv4Range::new(
+            std::net::Ipv4Addr::new(10, 0, 0, 1),
+            std::net::Ipv4Addr::new(10, 0, 0, 5),
+        )
+        .unwrap(),
+    );
+    object_create::<IpRange, IpPoolRange>(client, &add_url, &range).await;
+
+    assert_ip_pool_utilization(client, "p0", 0, 5).await;
+
+    // now lets add the maximum number of addresses we can and still get a total
+    // in the response. that's FFFF FFFF minus five for the above range
+    let big_range = IpRange::V6(
+        Ipv6Range::new(
+            std::net::Ipv6Addr::new(0xfd00, 0, 0, 0, 0, 0, 0, 1),
+            std::net::Ipv6Addr::new(0xfd00, 0, 0, 0, 0, 0, 0xffff, 0xfffa),

Reconsidering whether this situation would really be all that unusual once we support IPv6 ranges.

david-crespo:
Thanks. I will look into a more elaborate BigInt setup.

FelixMcFelix (approved):
Thanks a lot! I think this looks good on the omicron side, please ping me if you need to rework the `u128` handling to make it easier for consumers and I can take another look.

nexus/db-queries/src/db/datastore/ip_pool.rs
```diff
@@ -324,6 +324,66 @@ impl DataStore {
...
+        &self,
+        opctx: &OpContext,
+        authz_pool: &authz::IpPool,
+    ) -> Result<u128, Error> {
+        opctx.authorize(authz::Action::Read, authz_pool).await?;
+        opctx.authorize(authz::Action::ListChildren, authz_pool).await?;
+
+        use db::schema::ip_pool_range;
+
+        let ranges: Vec<IpPoolRange> = ip_pool_range::table
```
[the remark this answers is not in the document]
david-crespo: Interesting question! It hadn't occurred to me to give separate `allocated` and `capacity` counts for v4 and v6, but it makes a lot of sense, doesn't it. I will think about doing that here — it's always nice to avoid followup API changes where we can. One nice thing about that is that the IPv4 counts can be `u32`s.

Another way we could do this would be to have a given pool be IPv4 or IPv6 only, but I don't really like that. Right now you can add both IPv4 and IPv6 ranges to a pool, and it would be kind of complicated and/or confusing to require pools to have only one or the other.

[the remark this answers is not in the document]
david-crespo: All right, check out 39edb8ffad89148e40f7be298096e6b7aa5ac899. A little complicated, maybe, but the end result is 

```
{
  ipv4: { allocated, capacity },
  ipv6: { allocated, capacity },
}
```

Sound good?

nexus/src/app/utilization.rs
```diff
@@ -30,4 +33,55 @@ impl super::Nexus {
...
+                // below because an i64 could obviously be too big for u32.
+                // In addition to the fact that it is unlikely for anyone to
+                // allocate 4 billion IPs, we rely on the fact that there can
+                // only be 2^32 IPv4 addresses, period.
+                allocated: u32::try_from(allocated.ipv4).map_err(|_e| {
+                    Error::internal_error(&format!(
+                        "Failed to convert i64 {} IPv4 address count to u32",
+                        allocated.ipv4
+                    ))
+                })?,
```
david-crespo, line 69: I'm slightly nervous about this but I feel like it's going to be really, really hard to hit this 500!

nexus/db-queries/src/db/datastore/ip_pool.rs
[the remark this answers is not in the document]
FelixMcFelix: I think so, thanks for doing this! This also helps a lot since external v4s can be quite constrained in practice.

nexus/db-queries/src/db/datastore/ip_pool.rs
```diff
@@ -324,6 +334,80 @@ impl DataStore {
...
+            .select(IpPoolRange::as_select())
+            // This is a rare unpaginated DB query, which means we are
+            // vulnerable to a resource exhaustion attack in which someone
+            // creates a very large number of ranges in order to make this
+            // query slow. In order to mitigate that, we limit the query to the
+            // (current) max allowed page size, effectively making this query
+            // exactly as vulnerable as if it were paginated. If there are more
+            // than 10,000 ranges in a pool, we will undercount, but I have a
+            // hard time seeing that as a practical problem.
+            .limit(10000)
```
david-crespo, line 386: In [private chat](https://matrix.to/#/!YNYPOVxjAUeXksTcRj:oxide.computer/$NV_J25yvwwrIByPZw0QFaS7-Hl2AA3TfmSfkKwNNnfc?via=oxide.computer&via=unix.house&via=matrix.org) we discussed the alternative of doing this calculation in SQL so we don't have to limit it here. What we found is that neither CRDB nor Postgres support subtraction of IPv6 addresses when the result is greater than a 64 bit integer, and we couldn't figure out how to convert the addresses to 128 bit integers first and then subtract. So it seems we are forced to do this in Rust.

```
> select 'ffff::'::inet - '::1';
ERROR: result out of range
```

https://github.com/cockroachdb/cockroach/blob/91c1d5cbb5a56cd9c082d78814cd03c5471dddfa/pkg/util/ipaddr/ipaddr_test.go#L862-L881

https://github.com/cockroachdb/cockroach/blob/4deb9e3b77f38396a58bb8186fe51bb01cf9f007/pkg/util/ipaddr/ipaddr.go#L607-L632

PR #5259, opened 2024-03-13
Provide a way of detecting system library leakage

papertigers:
After https://github.com/oxidecomputer/omicron/pull/5158 was integrated @rain noticed that attempting to run a build of `omdb` in the switch zone suddenly stopped working and filed https://github.com/oxidecomputer/helios-omicron-brand/issues/15.  @jgallagher ended up fixing this by splitting out the sled-hardware types into their own crate in https://github.com/oxidecomputer/omicron/pull/5245. 

We decided it would be good if we added some sort of CI check to omicron to catch these library leakages earlier.  This PR introduces that check and adds it to the helios build and test buildomat job.  I have also added some notes to the readme for others that may end up adding a new library dependency.

Locally I modified the allow list so that it would produce errors, those errors end up looking like:
```
$ cargo xtask verify-libraries
    Finished dev [unoptimized + debuginfo] target(s) in 0.42s
     Running `target/debug/xtask verify-libraries`
    Finished dev [unoptimized + debuginfo] target(s) in 4.11s
Error: Found library issues with the following:
installinator
	UNEXPECTED dependency on libipcc.so.1
omicron-dev
	UNEXPECTED dependency on libipcc.so.1
	UNEXPECTED dependency on libresolv.so.2
sp-sim
	UNEXPECTED dependency on libipcc.so.1
	UNEXPECTED dependency on libresolv.so.2
sled-agent
	NEEDS libnvme.so.1 but is not allowed
mgs
	UNEXPECTED dependency on libipcc.so.1
	UNEXPECTED dependency on libresolv.so.2


If depending on a new library was intended please add it to xtask.toml
```


library-verification.toml
```diff
@@ -0,0 +1,12 @@
+# libnvme is a global zone only library and therefore we must be sure that only
+# programs running in the gz require it. Additionally only sled-agent should be
+# managing a sled's hardware.
+[libraries."libnvme.so.1"]
+binaries = [
```
sunshowers, line 5: So it isn't clear that this is actually an allowlist -- it could just as well be a denylist.

Could this be made clearer, e.g. by calling this `allowlist-binaries` or just `allowlist`?

README.adoc
```diff
@@ -106,6 +106,10 @@ When you're happy with things and want to make sure you haven't missed something
 cargo nextest run
 ```
 
+=== Adding a new system library dependency
+
+We check that certain system library dependencies are not leaked outside of their intended binaries via `cargo xtask verify-libraries` in CI. If you are adding a new dependency on a illumos/helios library it is recommended that you update xref:library-verification.toml[] with a whitelist of where you expect the dependency to show up. For example some libraries such as `libnvme.so.1` are only available in the global zone and therefore will not be present in any other zone. This check is here to help us catch any leakage before we go to deploy on a rack. You can inspect a compiled binary in the target directory for what it requires by using `elfedit` - for example `elfedit -r -e 'dyn:tag NEEDED' /path/to/omicron/target/debug/sled-agent`.
```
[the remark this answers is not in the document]
sunshowers: Hmm, thinking more about this -- can we also list out all the libraries we do expect to see everywhere, so if a new library is accidentally added we error out loudly instead of failing open?

dev-tools/xtask/src/main.rs
```diff
@@ -213,6 +237,86 @@ fn cmd_check_workspace_deps() -> Result<()> {
...
+                    }
+                }
+            }
+        }
+    }
+
+    if !offenders.is_empty() {
+        bail!(
+            "Found the following issues: {offenders:#?}\n\nIf you believe any \
+            of these issues are wrong please update the whitelist."
```
sunshowers, line 313: Instead of using the `Debug` impl, it would be best to generate a nicely formatted list as a string and print it out here.

dev-tools/xtask/src/main.rs
```diff
@@ -213,6 +237,86 @@ fn cmd_check_workspace_deps() -> Result<()> {
...
+    let config = read_library_verification_toml(&config_path)?;
+
+    let mut offenders = Vec::new();
+    for entry in fs::read_dir(&metadata.target_directory)
+        .with_context(|| format!("read_dir {:?}", &metadata.target_directory))?
+    {
+        let entry = entry.context("dirent")?;
+        if !entry.file_type()?.is_dir() {
+            continue;
+        }
```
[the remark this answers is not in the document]
sunshowers: Oh, and I forgot to mention that `cargo_metadata`, which xtask already pulls in, has support for this: https://docs.rs/cargo_metadata/0.18.1/cargo_metadata/enum.Message.html#method.parse_stream

library-verification.toml
```diff
@@ -0,0 +1,12 @@
+# libnvme is a global zone only library and therefore we must be sure that only
```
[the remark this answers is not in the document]
sunshowers: Oh also, can this be a more generically named file, e.g. `xtask.toml`? No reason to tie it to library verification specifically.

sunshowers (changes requested):
putting it back in your queue!

dev-tools/xtask/src/main.rs
```diff
@@ -213,6 +237,86 @@ fn cmd_check_workspace_deps() -> Result<()> {
     Ok(())
 }
 
+fn cmd_verify_library() -> Result<()> {
+    let metadata = load_workspace()?;
```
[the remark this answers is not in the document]
papertigers: Fixed in [dd71b07](https://github.com/oxidecomputer/omicron/pull/5259/commits/dd71b07e2d6b68fdeedb52cfd458366423d1c70c)

README.adoc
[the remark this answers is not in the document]
papertigers: Good suggestion, done!

dev-tools/xtask/src/main.rs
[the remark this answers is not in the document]
papertigers: Thanks for the tip! Fixed in [276e135](https://github.com/oxidecomputer/omicron/pull/5259/commits/276e13521b8782feff5b4d387ab86e56750e17b7)

dev-tools/xtask/src/illumos.rs
```diff
@@ -0,0 +1,144 @@
+use anyhow::{bail, Context, Result};
```
sunshowers, line 1: MPL files require a license header up top -- could you add one here?

dev-tools/xtask/src/illumos.rs
```diff
@@ -0,0 +1,144 @@
...
+    NotAllowed(String),
+}
+
+/// Verify that the binary at the provided path complies with the rules laid out
+/// in the xtask.toml config file. Errors are pushed to a hashmap so that we can
+/// display to a user the entire list of issues in one go.
+fn verify_executable(
+    config: &XtaskConfig,
+    path: &Utf8Path,
+    errors: &mut HashMap<String, Vec<LibraryError>>,
```
sunshowers, line 36: Should be some kind of ordered map -- either a `BTreeMap` or an `IndexMap` (probably a `BTreeMap`)

dev-tools/xtask/src/illumos.rs
```diff
@@ -0,0 +1,144 @@
...
+
+    Ok(())
+}
+pub fn cmd_verify_libraries() -> Result<()> {
+    let metadata = load_workspace()?;
+    let mut config_path = metadata.workspace_root;
+    config_path.push(".cargo/xtask.toml");
+    let config = read_xtask_toml(&config_path)?;
+
+    let mut command = Command::new("cargo")
```
sunshowers, line 91: This should use `$CARGO` if available (which it will be almost all of the time, since cargo is invoked via the rustup wrapper)

dev-tools/xtask/src/illumos.rs
```diff
@@ -0,0 +1,144 @@
...
+    let status = command.wait()?;
+    if !status.success() {
+        bail!("Failed to execute cargo build successfully {}", status);
+    }
+
+    if !errors.is_empty() {
+        let mut msg = String::new();
+        use std::fmt::Write;
+        errors.iter().for_each(|(binary, errors)| {
+            write!(msg, "{binary}\n").unwrap();
```
sunshowers, line 121: `writeln!`

Also you can use https://crates.io/crates/swrite to avoid the unwrap.

sunshowers (approved):
Looks fantastic, thanks! Just a few minor comments.

.github/buildomat/build-and-test.sh
```diff
@@ -76,6 +76,19 @@ export RUSTC_BOOTSTRAP=1
 # We report build progress to stderr, and the "--timings=json" output goes to stdout.
 ptime -m cargo build -Z unstable-options --timings=json --workspace --tests --locked --verbose 1> "$OUTPUT_DIR/crate-build-timings.json"
 
+# If we are running on illumos we want to verify that we are not requiring
+# system libraries outside of specific binaries. If we encounter this situation
+# we bail.
+# NB: `cargo xtask verify-libraries` runs `cargo build --bins` to ensure it can
```
sunshowers, line 82: Doesn't look like it runs `cargo build --bins`? Just `cargo build`.

dev-tools/xtask/src/illumos.rs
papertigers: Fixed in [fa5b905](https://github.com/oxidecomputer/omicron/pull/5259/commits/fa5b905ef38a8a5f50130329967961514346ec67)

dev-tools/xtask/src/illumos.rs
papertigers: Fixed in [fa5b905](https://github.com/oxidecomputer/omicron/pull/5259/commits/fa5b905ef38a8a5f50130329967961514346ec67)

papertigers: I left it as an error if the env::var("CARGO") fails, but happy to fall back on just "cargo" if you like that better.

dev-tools/xtask/src/illumos.rs
papertigers: Fixed in [fa5b905](https://github.com/oxidecomputer/omicron/pull/5259/commits/fa5b905ef38a8a5f50130329967961514346ec67)

dev-tools/xtask/src/illumos.rs
papertigers: Fixed in [fa5b905](https://github.com/oxidecomputer/omicron/pull/5259/commits/fa5b905ef38a8a5f50130329967961514346ec67)

dev-tools/xtask/src/illumos.rs
sunshowers: Should definitely fall back to `cargo`, yeah. This pattern is pretty standard, and is expressed in bash as e.g. `${CARGO:-cargo}`.

PR #5260, opened 2024-03-14
zone bundle tests fail on ZFS pool with 4K sectors (ashift=12)

jclulow:
There is a subtle and somewhat tedious to debug failure in the zone bundle tests due to our attempt to set a vanishingly small (32KB!) quota on a dataset.  This appears to work on a system with a ZFS pool that has an **ashift** value of **9** (i.e., 512 byte sectors); it fails catastrophically on a pool with an **ashift** of **12**, though.

The original test was trying to do this:

```
/usr/sbin/zfs create \
    -o quota=32768 \
    -o mountpoint=/pool/int/0aafdcd5-03ed-4d3f-ae3f-6a94cd313025/debug/bundle/zone \
     rpool/0aafdcd5-03ed-4d3f-ae3f-6a94cd313025
```

Dataset creation would fail with this error message:

```
cannot create 'rpool/0aafdcd5-03ed-4d3f-ae3f-6a94cd313025': out of space
```

This is clearly not accurate, as the pool was indeed mostly empty at the time.  If one were instead to create the dataset _without_ a quota, but then apply a quota as a second step, the error message is more revealing:

```
cannot set property for 'rpool/0aafdcd5-03ed-4d3f-ae3f-6a94cd313025': size is less than current used or reserved space
```

There is clearly room for improvement in the error message of `zfs create`, but at least we know what's wrong.  Bumping the quota to a much larger value will mean we don't see this error regardless of any sector size we are likely to see for a long time.

(I found this a few months ago as part of trying to get CI moved from AWS onto Propolis running on bench Gimlets, and am finally getting it cleaned up for integration.)

### DIFF
--- a/sled-agent/src/zone_bundle.rs
+++ b/sled-agent/src/zone_bundle.rs
@@ -2014,6 +2014,7 @@ mod tests {
     use super::ZoneBundleMetadata;
     use chrono::TimeZone;
     use chrono::Utc;
+    use rand::RngCore;
 
     #[test]
     fn test_sort_zone_bundle_cause() {
@@ -2176,7 +2177,9 @@ mod illumos_tests {
     use super::ZoneBundler;
     use super::ZFS;
     use anyhow::Context;
+    use chrono::DateTime;
     use chrono::TimeZone;
+    use chrono::Timelike;
     use chrono::Utc;
     use illumos_utils::zpool::ZpoolName;
     use sled_storage::disk::RawDisk;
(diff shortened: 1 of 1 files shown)


leftwo:
Just dropped by to point out:
`a subtle and somewhat tedious to debug failure ...due to our attempt to set a ... quota on a dataset.`
And, the solution:
`Bumping the quota to a much larger value`
That's all.

rcgoodfellow (approved):
LGTM

PR #5270, opened 2024-03-15
[reconfigurator] use tabled to display blueprints and diffs

sunshowers:
While developing #5238, I noticed that the output was getting significantly busier and less aligned. I decided to prototype out using `tabled` to display outputs, and I really liked the results.

Examples that cover all of the cases are included in the PR. In the future I'd also like to add color support on the CLI, and expand it to inventory and `omdb` (it's similar except it doesn't have the zone policy table).

Some other changes that are bundled into this PR:

* Sort by (zone type, zone ID) rather than zone ID, to keep zones of the same type grouped together.
* Moved unchanged data to the top to allow users to see less scrollback.
* Moved metadata to the bottom for the same reason.
* Add information about the zone config being changed.
* Change `Blueprint::diff_sleds` and `Blueprint::diff_sleds_from_collection` to `Blueprint::diff_since_blueprint` and `diff_since_collection` recently. 
* Reordered `diff_since_blueprint`'s arguments so that `self` is after and the argument is before, to align with `diff_since_collection`. (I found that surprising!)
* Renamed the diff type from `OmicronZonesDiff` to `BlueprintDiff`, since it's going to contain a lot more than zones.
* Return an error from the diff methods, specifically if the before and after have the same zone ID but different types.

Depends on #5238 and #5341.

TODO:

- [x] blueprints
- [x] diffs
- [x] show warning for modified sleds without a generation bump
- [x] move metadata to end
- [x] show DNS metadata in a table?
- [x] update PR summary above


nexus/reconfigurator/planning/tests/output/planner_nonprovisionable_1_2.txt
```diff
@@ -1,86 +1,102 @@
...
-+         87c299eb-470e-4b6d-b8c7-6759694e66b6 in service     nexus [underlay IP fd00:1122:3344:101::30] (added)
-+         c72b7930-0580-4f00-93b9-8cba2c8d344e in service     nexus [underlay IP fd00:1122:3344:101::2d] (added)
-+         d0095508-bdb8-4faf-b091-964276a20b15 in service     nexus [underlay IP fd00:1122:3344:101::31] (added)
-+         ff422442-4b31-4ade-a11a-9e5a25f0404c in service     nexus [underlay IP fd00:1122:3344:101::2f] (added)
+from: blueprint 55502b1b-e255-438b-a16a-2680a4b5f962
+-       external DNS version: 1
+-       internal DNS version: 1
+to:   blueprint 9f71f5d3-a272-4382-9154-6ea2e171a6c6
++       external DNS version: 1
++       internal DNS version: 1
```
sunshowers, line 6: Also included this kind of metadata.

clients/sled-agent-client/src/lib.rs
```diff
@@ -52,25 +53,62 @@ impl Eq for types::OmicronZoneConfig {}
 impl Eq for types::OmicronZoneType {}
 impl Eq for types::OmicronZoneDataset {}
 
+/// Like [`types::OmicronZoneType`], but without any associated data.
+///
+/// The [`fmt::Display`] impl for this type is a human-readable label, meant
+/// for testing and reporting.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, PartialOrd, Ord)]
+pub enum ZoneTag {
```
jgallagher, line 61: I have also wanted something like this. FWIW we also have `nexus_db_model::ServiceKind`, which is _almost_ the same (it has cases for a couple services within the switch zone, but not all, and doesn't distinguish boundary from internal NTP). I wonder if those can be unified. I don't think that fits in the scope of this PR, but curious if you have thoughts.

nexus/types/src/deployment.rs
```diff
@@ -828,3 +794,484 @@ pub struct UnstableReconfiguratorState {
...
+                            "{REMOVED_PREFIX}{SLED_SUBHEAD_INDENT}removed zones:"
+                        ));
+                    }
+                    builder.push_zone_record(
+                        format!("{REMOVED_PREFIX}{ZONE_INDENT}"),
+                        zone,
+                    );
+                }
+
+                // First print out all changed zones.
```
jgallagher, line 907: Tiny nit - this is second after printing all removed zones

nexus/types/src/deployment.rs
```diff
@@ -828,3 +794,484 @@ pub struct UnstableReconfiguratorState {
...
+        }
+
+        if before.config.zone_type != after.config.zone_type {
+            // There are two cases here:
+            // 1. The zone type changed (we don't expect this to happen)
+            // 2. The zone type did not change, but some of the configuration
+            //    parameters inside changed. We need to add additional text to
+            //    make that clearer.
+            //
+            // TODO: also print out the exact bits of configuration that
```
jgallagher, line 1122: Do you want to do this on this PR, in a followup, or file an issue for someone to pick up later? (I have no preference between these - the old format also didn't show the details here, right?)

jgallagher (commented):
Looks great to me!

nexus/reconfigurator/planning/tests/output/planner_nonprovisionable_1_2.txt
sunshowers: Sure, happy to move it to the bottom.

clients/sled-agent-client/src/lib.rs
sunshowers: Yeah I did find that enum. This felt like it was doing almost the same thing, but not quite, and I'm not entirely sure how to unify the two.

Maybe we should move `ZoneTag` next to `ServiceType` though, and document the differences, to make future archeologists wonder less.

nexus/types/src/deployment.rs
sunshowers: Definitely in a follow up, will file an issue for this. A question that immediately occurs to me is: why are diffs special, why not display all the configs for everything? Maybe we should have an extended mode in the displayer to make that possible.

clients/sled-agent-client/src/lib.rs
davepacheco: What do you think about calling this `Kind`?  That's what we call it elsewhere and "tag" could mean a lot of different hings.  ("kind" can too but at least I'd expect a thing to have only one "kind", whereas it may have many tags.)

According to dbinit.sql, the database `service_kind` enum ought to be used only for the services table, which I think does not contain switch zone services (this is from dogfood):

```
> select distinct kind from service;
       kind
-------------------
  clickhouse
  cockroach
  crucible
  crucible_pantry
  external_dns
  internal_dns
  nexus
  ntp
  oximeter
(9 rows)


```

and the certificates table (which I think currently only supports Nexus as the service kind).  Maybe we can drop `ServiceKind` when we drop the `services` table and be left with just this one?

nexus/reconfigurator/planning/tests/output/planner_basic_add_sled_2_3.txt
```diff
@@ -1,48 +1,61 @@
...
-          af322036-371f-437c-8c08-7f40f3f1403b in service     crucible [underlay IP fd00:1122:3344:102::23] (unchanged)
-          d637264f-6f40-44c2-8b7e-a179430210d2 in service     crucible [underlay IP fd00:1122:3344:102::25] (unchanged)
-          dce226c9-7373-4bfa-8a94-79dc472857a6 in service     crucible [underlay IP fd00:1122:3344:102::27] (unchanged)
-          edabedf3-839c-488d-ad6f-508ffa864674 in service     crucible [underlay IP fd00:1122:3344:102::24] (unchanged)
-+ sled b59ec570-2abb-4017-80ce-129d94e7a025 (added)
-+     zone config generation 2
-+         2d73d30e-ca47-46a8-9c12-917d4ab824b6 in service     internal_ntp [underlay IP fd00:1122:3344:104::21] (added)
+from: blueprint 979ef428-0bdd-4622-8a72-0719e942b415
+-       external DNS version: 1
+-       internal DNS version: 1
```
davepacheco, line 3: It seems like this shouldn't go in between the "from"/"to" labels.  (Or, why would DNS be different than the zones in this way?)

nexus/reconfigurator/planning/tests/output/planner_nonprovisionable_2_2a.txt
```diff
@@ -0,0 +1,106 @@
...
+                                                                                                                
+~      changed zones:                                                                                           
+~        19fbc4f8-a683-4f22-8f5a-e74782b935be  :  policy        in service             -> not in service        
+                                               :  zone type     crucible                  (unchanged)           
+                                               :  underlay IP   fd00:1122:3344:105::26    (unchanged)           
+~        6dff7633-66bb-4924-a6ff-2c896e66964b  :  policy        in service                (unchanged)           
+                                               :  zone type     nexus                     (config changed)      
+                                               :  underlay IP   fd00:1122:3344:105::22    (unchanged)           
+~        7f4e9f9f-08f8-4d14-885d-e977c05525ad  :  policy        in service                (unchanged)           
+                                               :  zone type     internal_ntp              (unchanged)
```
davepacheco, line 36: I don't think the zone type can ever change (I think we should mandate that)

nexus/reconfigurator/planning/tests/output/planner_nonprovisionable_bp2.txt
```diff
@@ -0,0 +1,95 @@
+blueprint  9f71f5d3-a272-4382-9154-6ea2e171a6c6
+parent:    55502b1b-e255-438b-a16a-2680a4b5f962
+created by add more Nexus
```
davepacheco, line 3: In whatever generates this, I think `"add more Nexus"` was probably meant for the `comment` field but got stuffed into the `creator` field.

nexus/types/src/deployment.rs
```diff
@@ -254,6 +254,18 @@ impl Blueprint {
     pub fn display(&self) -> BlueprintDisplay<'_> {
         BlueprintDisplay { blueprint: self }
     }
+
+    // ---
+    // Helper methods
+    // ---
+
+    fn diff_metadata(&self) -> BlueprintDiffMetadata {
```
davepacheco, line 262: What exactly is "metadata" (or: why are the internal and external DNS not just "another part of the blueprint" like zones and future stuff like which host OS should be running on which sleds, etc)?

davepacheco (commented):
Thanks for doing this.  I was ambivalent but I think I've come around.

I *really* like that we're testing the blueprint diff output.

Some parts definitely look nicer.

My complaints are all nitty preference on the output format (but then again, the point of the PR is to change the output format for clarity so hopefully this isn't too out of place).  My first reaction was: this looks a lot less like diff(1) output and I find it quite a lot harder to skim and scan it.  I had tried to write up why but then took another look at the raw files instead of the GitHub diff and I think it's not so bad.

I do prefer left-aligned (or even right-aligned) columns to center-aligned.  I think this is probably just convention burned into my brain more than anything -- I don't think I've seen many terminal tools that center-align columns and I'm not used to searching for tokens (like `nexus`) in the middle of a column.

sunshowers:
> My complaints are all nitty preference on the output format (but then again, the point of the PR is to change the output format for clarity so hopefully this isn't too out of place).

Absolutely! I really want to make things better for all of us, and feedback is really welcome. I've already incorporated some of your feedback. (And yes -- I do think it looks better in the terminal when I'm actually developing against it, the visual separation makes things really easy to parse.) And as always, we can keep iterating on this. Thanks for being so receptive :)

clients/sled-agent-client/src/lib.rs
sunshowers: Good point about kind vs tag -- I also think of entities as potentially having many tags.

I was hesitating to call it `ZoneKind` because of `ServiceKind`. And I saw a survey on some social media (maybe Fediverse) a while ago where Rust devs were being polled for how they name parallel enums without associated data -- `Kind` and `Tag` were the top two results. But if we just drop `ServiceKind` that becomes moot and `ZoneKind` is better.

sunshowers: Changed this to `ZoneKind.

nexus/types/src/deployment.rs
sunshowers: Fixed, thanks.

nexus/reconfigurator/planning/tests/output/planner_nonprovisionable_2_2a.txt
sunshowers: Done by eagerly computing the diff and returning an error if the zone type changes.

nexus/reconfigurator/planning/tests/output/planner_basic_add_sled_2_3.txt
sunshowers: Hmmm -- so are you saying this should also be a table?

davepacheco: I don't know?  It's more about the order of things.  Right now it's:

```
from: blueprint 1 label
     some stuff from blueprint1 that may have changed (the DNS stuff)
to: blueprint 2 label
     some stuff from blueprint2 that may have changed (the DNS stuff)

more stuff from blueprint1 that may have changed (the non-DNS stuff)
more stuff from blueprint2 that may have changed (the non-DNS stuff)
```

It makes it look like the DNS stuff is part of the identifying information about the blueprint, but I think of it as just another part of the blueprint contents.  I'd expect it to look more like:

```
from: blueprint1 label
to: blueprint2 label

DNS changes from blueprint1 to blueprint2
Omicron zone changes from blueprint1 to blueprint2
... (other changes from blueprint1 to blueprint2)
```

or more concretely instead of:

```
from: blueprint 979ef428-0bdd-4622-8a72-0719e942b415
-       external DNS version: 1
-       internal DNS version: 1
to:   blueprint 4171ad05-89dd-474b-846b-b007e4346366
+       external DNS version: 1
+       internal DNS version: 1

zones:
```

it'd be:

```
from: blueprint 979ef428-0bdd-4622-8a72-0719e942b415
to:   blueprint 4171ad05-89dd-474b-846b-b007e4346366

DNS:
    - internal DNS version: 1
    + internal DNS version: 2
    - external DNS version: 1
    + external DNS version: 2

zones:
```

(or something more concise if we want, like `internal DNS version: 1 -> 2`, but the point is that "from"/"to" would be the top two lines)

nexus/reconfigurator/planning/tests/output/planner_nonprovisionable_bp2.txt
sunshowers: Hmm, looks like this was set in the planner, which doesn't let you set comments externally. I'll change this to be clearer.

nexus/reconfigurator/planning/tests/output/planner_basic_add_sled_2_3.txt
sunshowers: Ahh, I see what you mean -- misunderstood earlier. That definitely makes sense, and I'll include it with @jgallagher's suggestion to move this metadata to the bottom.

nexus/types/src/deployment.rs
sunshowers: So I needed an abstraction to represent that:

* the "before" of a diff can be either a collection or a blueprint
* the "after" is always a blueprint
* the blueprint has stuff associated with it that collections don't.

So `DiffBlueprintMetadata` consists of all the blueprint-specific things, and there's also a `DiffBeforeMetadata` that has `Collection` and `Blueprint` variants.

nexus/reconfigurator/planning/tests/output/planner_basic_add_sled_2_3.txt
sunshowers: Done, and moved to the end.

nexus/reconfigurator/planning/tests/output/planner_nonprovisionable_1_2.txt
sunshowers: Done.

sunshowers:
This ended up being expanded quite substantially after all the feedback was incorporated, but I'm pretty happy overall with the way it generally looks.

I'll go over with some final ideas tomorrow morning, then put it up for another round of review.

nexus/reconfigurator/planning/tests/output/planner_nonprovisionable_bp2.txt
sunshowers: Changed these.

sunshowers:
Showed this to @smklein this morning, who had a bunch of really great suggestions. In particular, we figured out a much better way to show modified zones -- really happy with the current iteration.

nexus/reconfigurator/planning/tests/output/planner_nonprovisionable_2_2a.txt
```diff
@@ -0,0 +1,102 @@
...
+     crucible       e98cc0de-abf6-4da4-a20d-d05c7a9bb1d7   in service    fd00:1122:3344:105::2b             
+     crucible       f55e6aaf-e8fc-4913-9e3c-8cd1bd4bdad3   in service    fd00:1122:3344:105::29             
+-    crucible       4f1ce8a2-d3a5-4a38-be4c-9817de52db37   in service    fd00:1122:3344:105::2c   removed   
+-    crucible       19fbc4f8-a683-4f22-8f5a-e74782b935be   in service    fd00:1122:3344:105::26   modified  
++     └─                                                   quiesced      fd00:1122:3344:105::26             
+-    internal_ntp   7f4e9f9f-08f8-4d14-885d-e977c05525ad   in service    fd00:1122:3344:105::21   modified  
++     └─                                                   in service    fd01:1122:3344:105::21             
+-    nexus          6dff7633-66bb-4924-a6ff-2c896e66964b   in service    fd00:1122:3344:105::22   modified  
++     ├─                                                   in service    fd00:1122:3344:105::22             
+*     └─ zone type config changed
```
sunshowers, line 84: I skipped over all the columns that must inherently be the same over here.

jgallagher (commented):
Changes LGTM - just a few questions on some of the output files

nexus/reconfigurator/planning/tests/output/planner_nonprovisionable_2_2a.txt
```diff
@@ -0,0 +1,102 @@
...
+     crucible       93b137a1-a1d6-4b5b-b2cb-21a9f11e2883   in service    fd00:1122:3344:105::23           
+     crucible       9f0abbad-dbd3-4d43-9675-78092217ffd9   in service    fd00:1122:3344:105::25           
+     crucible       b0c63f48-01ea-4aae-bb26-fb0dd59d1662   in service    fd00:1122:3344:105::28           
+     crucible       c406da50-34b9-4bb4-a460-8f49875d2a6a   in service    fd00:1122:3344:105::24           
+     crucible       d660d7ed-28c0-45ae-9ace-dc3ecf7e8786   in service    fd00:1122:3344:105::2a           
+     crucible       e98cc0de-abf6-4da4-a20d-d05c7a9bb1d7   in service    fd00:1122:3344:105::2b           
+     crucible       f55e6aaf-e8fc-4913-9e3c-8cd1bd4bdad3   in service    fd00:1122:3344:105::29           
+-    crucible       4f1ce8a2-d3a5-4a38-be4c-9817de52db37   in service    fd00:1122:3344:105::2c  removed  
+-    crucible       19fbc4f8-a683-4f22-8f5a-e74782b935be   in service    fd00:1122:3344:105::26  modified 
++     └─                                                   quiesced      fd00:1122:3344:105::26
```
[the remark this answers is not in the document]
sunshowers: Hmm, so my philosophy here was to skip the things that are _necessarily_ the same, while showing the things that could possibly be different even if they were in reality the same.

If this is too confusing, maybe we should always add a second line explaining what changed.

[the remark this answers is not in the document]
jgallagher: Are there properties that we don't display at all that could still result in a `modified` zone?

> If this is too confusing, maybe we should always add a second line explaining what changed.

I _think_ I'm in favor of this, but I'm not completely sure - if you or others think it's overkill, I'd be fine with that.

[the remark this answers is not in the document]
sunshowers: > Are there properties that we don't display at all that could still result in a modified zone?

The only other case is zone type configs, and we display a "zone type config changed" message for that. So we'd potentially club together the other things that have changed as part of that.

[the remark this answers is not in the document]
sunshowers: Made this change, like it a lot how it makes what changed quite clear.

nexus/reconfigurator/planning/tests/output/planner_nonprovisionable_2_2a.txt
```diff
@@ -0,0 +1,102 @@
...
+     crucible       b0c63f48-01ea-4aae-bb26-fb0dd59d1662   in service    fd00:1122:3344:105::28           
+     crucible       c406da50-34b9-4bb4-a460-8f49875d2a6a   in service    fd00:1122:3344:105::24           
+     crucible       d660d7ed-28c0-45ae-9ace-dc3ecf7e8786   in service    fd00:1122:3344:105::2a           
+     crucible       e98cc0de-abf6-4da4-a20d-d05c7a9bb1d7   in service    fd00:1122:3344:105::2b           
+     crucible       f55e6aaf-e8fc-4913-9e3c-8cd1bd4bdad3   in service    fd00:1122:3344:105::29           
+-    crucible       4f1ce8a2-d3a5-4a38-be4c-9817de52db37   in service    fd00:1122:3344:105::2c  removed  
+-    crucible       19fbc4f8-a683-4f22-8f5a-e74782b935be   in service    fd00:1122:3344:105::26  modified 
++     └─                                                   quiesced      fd00:1122:3344:105::26           
+-    internal_ntp   7f4e9f9f-08f8-4d14-885d-e977c05525ad   in service    fd00:1122:3344:105::21  modified 
++     └─                                                   in service    fd01:1122:3344:105::21
```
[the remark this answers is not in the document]
sunshowers: Addressed via making it more obvious what changed.

nexus/reconfigurator/planning/tests/output/blueprint_builder_initial_diff.txt
```diff
@@ -0,0 +1,54 @@
...
+     crucible       b3583b5f-4a62-4471-9be7-41e61578de4c   in service    fd00:1122:3344:102::2a         
+     crucible       bac92034-b9e6-4e8b-9ffb-dbba9caec88d   in service    fd00:1122:3344:102::28         
+     crucible       d9653001-f671-4905-a410-6a7abc358318   in service    fd00:1122:3344:102::2b         
+     crucible       edaca77e-5806-446a-b00c-125962cd551d   in service    fd00:1122:3344:102::26         
+     internal_ntp   aac3ab51-9e2b-4605-9bf6-e3eb3681c2b5   in service    fd00:1122:3344:102::21         
+     nexus          29278a22-1ba1-4117-bfdb-39fcb9ae7fd1   in service    fd00:1122:3344:102::22         
+
+  METADATA:
++   internal DNS version:  1 (new)
++   external DNS version:  1 (new)
```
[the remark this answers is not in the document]
sunshowers: Yeah it is a quirk of that. Changed the note to "expected no non-trivial changes", and also made it clearer that these things are not present in collections.

nexus/types/src/deployment.rs
sunshowers: Updated this to the `BlueprintMetadata` that was added later.

sunshowers:
@jgallagher accepted this over chat (doesn't have access to GH right now), so landing.

PR #5274, opened 2024-03-15
make it a little harder to mess up new db schema migrations

davepacheco:
This is something we've talked about for a while.  Many ideas have been floated but as far as I know nobody's implemented any of them.  Here's a time-boxed attempt to make a tangible improvement.  It doesn't solve all the problems.

My goals here were:

* **When two people both make schema changes in separate branches, then one lands on "main", the other should be forced to resolve this _before_ they can land their PR on main.**  This basically means there has to be a git conflict when merging the first change into the second change's branch.  Recall that right now, there's not necessarily a git conflict.  In practice, it's possible for the second PR to land on "main" successfully, resulting in a broken "main".  The test suite on "main" will usually fail after that because for this to happen there must be a mix of files like "up1.sql" and "up01.sql" in the same version and we disallow that.  But this is only detected when running the test suite _after_ the merge, and as we said there are not necessarily any git conflicts generated by the merge, which means nothing prevents the PR from landing onto "main" and breaking it.
* **Further, when this happens, it should be easy to resolve this conflict.**  Right now, if you run into the above case, you have to create a new directory for your files and figure out which of the files in the previous version were yours vs. already present in upstream and move them manually.  It's fairly mechanical but tedious and error-prone.
* **Make sure a test catches the case of forgetting to update `SCHEMA_VERSION` in Rust or the schema version in `db_metadata`.**  This was already true via `dbinit_version_matches_version_known_to_nexus`.  Unfortunately I've found that if you get this wrong, many tests wind up hanging (because Nexus is sitting waiting for the database to be updated), so you'd have to be a little lucky to notice this one explicit failure.  We could have a Nexus config option that causes it to explode if the version doesn't match and then use that only in the test suite to try to catch this better but I decided to punt on this in this PR.

The approach I took:

* Instead of discovering all schema versions by listing the contents of the directory, we have an explicit list of versions in a Rust array called `KNOWN_VERSIONS`.  This means if two changes attempt to add the same item at the same spot in the array, they should cause a git conflict.  (I still want to test this.)  It is conceivable that _incorrect_ changes would _not_ generate a conflict (e.g., if somebody added their version in the wrong spot in the array) but I've added a lot of tests to verify that the array looks like it should (i.e., the versions can't be out of order, there can't be any gaps, etc.).
* Rather than assuming the directory of SQL files for each version is named for the version, the name is explicit in `KNOWN_VERSIONS`.  The developer's expected to create their own unique string for this (e.g., `drop-table-services`).  This way, if you do wind up conflicting with somebody, all you have to do is fix up the `KNOWN_VERSIONS` table.
* I hope the interfaces here make it hard to get any of this wrong (i.e., if you use `KnownVersion::new`, you _have_ to provide a directory name and the test suite makes sure it does not contain the version number in it).
* I did not change any of the existing migrations.  (There's a separate `KnownVersion::legacy()` constructor for those and you should get a test failure if you try to use that for any new migrations.)

I'm really interested to know if people feel like this would be a net improvement.  I know there were a lot of other ideas floated (and I'd be just as happy if someone implemented one of those instead!) but I'm trying to avoid letting the perfect be the enemy of the good.

Edit: there's an example migration in 89416f2.


davepacheco:
For the record, I've confirmed that if you make only this change to "main":

```
dap@ivanova omicron-play $ git diff
diff --git a/nexus/db-model/src/schema.rs b/nexus/db-model/src/schema.rs
index c31ce1677..08aab8ef7 100644
--- a/nexus/db-model/src/schema.rs
+++ b/nexus/db-model/src/schema.rs
 ///
-pub const SCHEMA_VERSION: SemverVersion = SemverVersion::new(43, 0, 0);
+pub const SCHEMA_VERSION: SemverVersion = SemverVersion::new(44, 0, 0);
 
 table! {
     disk (id) {
```

and run the test suite, after 25 minutes, 16 tests (the max that nextest runs concurrently on this machine?) appear hung: they've been running for 24+ minutes.  I think when I dug into this previously, they were waiting for the db version to catch up to Nexus's (which would never happen).

I don't expect this to be changed by this PR.

davepacheco:
To test the merge/conflict behavior, I did this:

- From the tip of this branch, create branch `dap/schema-migration-better-demo1` with a trivial migration from 45 to 46.  This passes the `schema_versions` tests.
- From the same commit, create branch `dap/schema-migration-better-demo2` with a different trivial migration from 45 to 46.  This passes the `schema_versions` tests.
- From the same commit, create branch `dap/schema-migration-better-demo-merged`:
    - This is our simulated "main" branch.
    - First, fast-forward merge `dap/schema-migration-better-demo1`.  This simulates the first branch landing on "main".
    - Next, merge `dap/schema-migration-better-demo2`.  This simulates the second branch trying to land on "main".
    - As expected, this generates a merge conflict that's pretty easy to resolve (more below).

Here's what the merge looks like.  Starting from this PR's branch:

```
dap@ivanova omicron-iter $ git status -uno
On branch dap/schema-migration-better
Your branch is up to date with 'origin/dap/schema-migration-better'.

nothing to commit (use -u to show untracked files)

dap@ivanova omicron-iter $ git log -1
commit 89416f23fd02f24ebdb5e1f2890debda27034248 (HEAD -> dap/schema-migration-better, origin/dap/schema-migration-better)
Author: David Pacheco <dap@oxidecomputer.com>
Date:   Fri Mar 15 15:42:15 2024 -0700

    add example migration
```

Create the simulated "main" branch:

```
dap@ivanova omicron-iter $ git switch -c dap/schema-migration-better-merged
Switched to a new branch 'dap/schema-migration-better-merged'
```

Merge in one branch:

```
dap@ivanova omicron-iter $ git merge --ff-only dap/schema-migration-better-demo1
Updating 89416f23f..d98746135
Fast-forward
 nexus/db-model/src/schema_versions.rs        | 3 ++-
 schema/crdb/dbinit.sql                       | 2 +-
 schema/crdb/migration-conflict-demo-1/up.sql | 2 ++
 3 files changed, 5 insertions(+), 2 deletions(-)
 create mode 100644 schema/crdb/migration-conflict-demo-1/up.sql
```

Merge in the other:

```
dap@ivanova omicron-iter $ git merge dap/schema-migration-better-demo2
Auto-merging nexus/db-model/src/schema_versions.rs
CONFLICT (content): Merge conflict in nexus/db-model/src/schema_versions.rs
Automatic merge failed; fix conflicts and then commit the result.
```

Good!  Here's what the conflict looks like (I have merge.conflictstyle=diff3 so it looks a little different than the default):

```
 22 /// List of all past database schema versions, in *reverse* order                 
 23 ///                                                                               
 24 /// If you want to change the Omicron database schema, you must update this.      
 25 static KNOWN_VERSIONS: Lazy<Vec<KnownVersion>> = Lazy::new(|| {                   
 26     vec![                                                                         
 27         // +- The next version goes here!  Duplicate this line, uncomment         
 28         // |  the *second* copy, then update that copy for your version,          
 29         // |  leaving the first copy as an example for the next person.           
 30         // v                                                                      
 31         // KnownVersion::new(next_int, "unique-dirname-with-the-sql-files"),      
 32 <<<<<<< HEAD                                                                      
 33         KnownVersion::new(46, "migration-conflict-demo-1"),                       
 34 ||||||| 89416f23f                                                                 
 35 =======                                                                           
 36         KnownVersion::new(46, "migration-conflict-demo-2"),                       
 37 >>>>>>> dap/schema-migration-better-demo2                                         
 38         KnownVersion::new(45, "migration-demo"),                                  
 39         // The first many schema versions only vary by major or patch number and  
 40         // their path is predictable based on the version number.  (This was      
 41         // historically a problem because two pull requests both adding a new     
 42         // schema version might merge cleanly but produce an invalid result.)     
 43         KnownVersion::legacy(44, 0),                                              
 44         KnownVersion::legacy(43, 0),                                              
```

The hand-merged result looks like this:

```
 27         // +- The next version goes here!  Duplicate this line, uncomment
 28         // |  the *second* copy, then update that copy for your version,
 29         // |  leaving the first copy as an example for the next person.
 30         // v
 31         // KnownVersion::new(next_int, "unique-dirname-with-the-sql-files"),
 32         KnownVersion::new(47, "migration-conflict-demo-2"),
 33         KnownVersion::new(46, "migration-conflict-demo-1"),
 34         KnownVersion::new(45, "migration-demo"),
 35         // The first many schema versions only vary by major or patch number and
 36         // their path is predictable based on the version number.  (This was
 37         // historically a problem because two pull requests both adding a new
 38         // schema version might merge cleanly but produce an invalid result.)
```

You also need to remember to update SCHEMA_VERSION above and dbinit.sql (which is annoying, but tests will at least not pass if you forget either or both of these).

All three of these branches are on GitHub if folks want to look or try this out.

schema/crdb/README.adoc
```diff
@@ -56,41 +56,75 @@ See RFD 319 for more discussion of the online upgrade plans.
...
+** Choose a unique, descriptive name for the directory.  It's suggested that
+   you stick to lowercase letters, numbers, and hyphen.  For example, if you're
+   adding a table called `widgets`, you might create a directory called
+   `create-table-widgets`.
+** Create the directory: `schema/crdb/NAME`.
+*** If only one SQL statement is necessary to get from `OLD_VERSION` to
+    `NEW_VERSION`, put that statement into `schema/crdb/NAME/up.sql`.  If
+    multiple statements are required, put each one into a separate file, naming
+    these `schema/crdb/NAME/upN.sql` for as many `N` as you need, staring with
+    `N=1`.
```
smklein, line 79:
```suggestion
    these `schema/crdb/NAME/upN.sql` for as many `N` as you need, starting with
    `N=1`.
```

schema/crdb/README.adoc
```diff
@@ -56,41 +56,75 @@ See RFD 319 for more discussion of the online upgrade plans.
...
+    statements.  But you should not mix schema-modifying statements and
+    data-modifying statements in one file.  And you should not include multiple
+    schema-modifying statements in one file.
+*** Beware that the entire file will be run in one transaction.  Expensive data-
+    modifying operations leading to long-running transactions are generally
+    to-be-avoided; however, there's no better way to do this today if you really
+    do need to update thousands of rows as part of the update.
+* Update `schema/crdb/dbinit.sql`:
+** Update the SQL statements to match what the database should look like
+   after your up*.sql files are applied.
```
smklein, line 91: Practically, I do this first, and save my schema changes for later. I wonder if we should recommend this first? "All tests except the schema tests" should pass if you only change `dbinit.sql`, and don't change the schema version.

nexus/db-model/src/schema_versions.rs
```diff
@@ -0,0 +1,798 @@
...
+///
+/// If you want to change the Omicron database schema, you must update this.
+static KNOWN_VERSIONS: Lazy<Vec<KnownVersion>> = Lazy::new(|| {
+    vec![
+        // +- The next version goes here!  Duplicate this line, uncomment
+        // |  the *second* copy, then update that copy for your version,
+        // |  leaving the first copy as an example for the next person.
+        // v
+        // KnownVersion::new(next_int, "unique-dirname-with-the-sql-files"),
+        KnownVersion::new(45, "migration-demo"),
```
smklein, line 32: Thanks for doing this in a way that encourages "only bumping major version" for now!

As a *super* nitpicky nitpick, can we change the name from `migration-demo` to like `first-named-migration`? At first glance, I thought it was test data, but I realize it actually is a durable long-lived migration.

nexus/db-model/src/schema_versions.rs
```diff
@@ -0,0 +1,798 @@
...
+
+/// The earliest supported schema version.
+pub const EARLIEST_SUPPORTED_VERSION: SemverVersion =
+    SemverVersion::new(1, 0, 0);
+
+/// Describes one version of the database schema
+#[derive(Debug, Clone)]
+struct KnownVersion {
+    /// All versions have an associated SemVer.  We only use the major number.
+    semver: SemverVersion,
```
smklein, line 98: > "We only use the major number"

This isn't *strictly* true, we do keep them ordered by minor/patch versions too, we just can't ensure backwards compatibility and recommend only bumping the major number.

I only bring this up because there already are legacy versions that only differ by minor versions.

smklein (approved):
Thanks for this, I hope it makes the process a little less conflict prone!

nexus/db-model/src/schema_versions.rs
jgallagher: I'll second the super nitpicky nitpick; the name threw me off also. (I was wondering if it was a demo that shoudl be removed before this lands.)

nexus/db-model/src/schema_versions.rs
```diff
@@ -0,0 +1,798 @@
...
+    ///
+    /// `major` should be the next available integer (one more than the previous
+    /// version's major number).
+    ///
+    /// `relative_path` is the path relative to "schema/crdb" (from the root of
+    /// this repository) where the SQL files live that will update the schema
+    /// from the previous version to this version.
+    fn new(major: u64, relative_path: &str) -> KnownVersion {
+        let semver = SemverVersion::new(major, 0, 0);
+        KnownVersion { semver, relative_path: relative_path.to_owned() }
```
jgallagher, line 116: I'm tempted to propose that we make the relative path for these `{semver}-{name}` so that the directory also remains sorted. But I don't feel super strongly about this.

nexus/db-queries/src/db/datastore/db_metadata.rs
```diff
@@ -160,147 +35,121 @@ impl DataStore {
...
-
+        ensure!(
+            all_versions.contains_version(&desired_version),
+            "Desired version {desired_version} was not found",
+        );
+
+        let target_versions: Vec<&SchemaVersion> = all_versions
+            .versions_range((
+                Bound::Excluded(&found_version),
+                Bound::Included(&desired_version),
```
jgallagher, line 110: Tiny nitpick and I realize this didn't change, but do you think this would be clearer as `.versions_range(&found_version..=&desired_version)`? I don't usually see explicit `Bound` tuples.

jgallagher (approved):

schema/crdb/README.adoc
davepacheco: I think this is a good idea and I looked at it but it makes the instructions _more_ complicated because up front you update the schema parts of dbinit.sql and later you go back to it.  Reading it straight through it reads like "why did you tell me to go back to the same file twice"?  I'd welcome edits here but otherwise I'm just going to leave it.

I also think this will be less necessary now.  I did the same as you because I was worried about dealing with merges, but now that's pretty easy so I will probably just go ahead and do the migrations earlier than I used to.

nexus/db-model/src/schema_versions.rs
davepacheco: Sorry, what I meant was not "only the major number is ever non-zero" but rather "we only use the major number to determine compatibility".  I'll clarified this comment.

nexus/db-model/src/schema_versions.rs
davepacheco: I thought about this but it makes the instructions for merging more complicated (you have to `git mv` your directory as part of resolving a conflict, and that wouldn't necessarily be obvious) and I'm not sure it helps that much because they don't appear in sorted order in `ls(1)` output and on GitHub anyway (because it's a lexicographic sort, not a semver-aware sort).

nexus/db-model/src/schema_versions.rs
```diff
@@ -0,0 +1,800 @@
...
+///
+/// If you want to change the Omicron database schema, you must update this.
+static KNOWN_VERSIONS: Lazy<Vec<KnownVersion>> = Lazy::new(|| {
+    vec![
+        // +- The next version goes here!  Duplicate this line, uncomment
+        // |  the *second* copy, then update that copy for your version,
+        // |  leaving the first copy as an example for the next person.
+        // v
+        // KnownVersion::new(next_int, "unique-dirname-with-the-sql-files"),
+        KnownVersion::new(46, "first-named-migration"),
```
sunshowers, line 32: Came up in `#oxide-update` today, but some schema changes will require multiple versions to be correct. (Came up for #5032 earlier, and also for #5287 which requires three migrations.) It would be nice to be able to express an ordered list of versions this way, maybe within a single directory have a bunch of subdirectories.

sunshowers: > (Though we think we can simplify this constraint and not require that, by changing the schema version after each file rather than each directory.)

@jgallagher is going to be doing this.

davepacheco: Sorry I didn't quite follow the problem.  @sunshowers Is this PR blocked on that work?  Does it make this situation worse?

> Came up in #oxide-update today, but some schema changes will require multiple versions to be correct. (Came up for https://github.com/oxidecomputer/omicron/pull/5032 earlier, and also for https://github.com/oxidecomputer/omicron/pull/5287 which requires three migrations.) It would be nice to be able to express an ordered list of versions this way, maybe within a single directory have a bunch of subdirectories.

For a PR that needs to land more than one separate schema update, is there any reason not to just define two new `KnownVersion` values with separate major numbers?  i.e., why does this use case not fit into the infrastructure we already have / what's added in this PR?

nexus/db-queries/src/db/datastore/db_metadata.rs
```diff
@@ -160,147 +34,118 @@ impl DataStore {
...
+            "Desired version {desired_version} was not found",
+        );
 
-        let target_versions = all_versions.range((
-            Bound::Excluded(&current_version),
-            Bound::Included(&desired_version),
-        ));
+        let target_versions: Vec<&SchemaVersion> = all_versions
+            .versions_range(&found_version..=&desired_version)
+            .collect();
```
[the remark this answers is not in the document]
jgallagher: Oops! That's definitely my bad; should've been `..` instead of `..=`.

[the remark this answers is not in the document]
smklein: I don't think that's exactly right, I think `..` is "start inclusive, end exclusive". I believe we actually want "start exclusive, end inclusive", which doesn't have a short-hand syntax. That's why I was using `std::ops::Bound::Excluded` explicitly before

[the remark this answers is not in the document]
jgallagher: Ooof. Doubly sorry then. :( Do you want to fix this on #5293 or should I open a fix just for this?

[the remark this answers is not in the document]
smklein: I'm patching it in 5293, so no need for something else unless the urgency seems like it should be higher

PR #5281, opened 2024-03-18
Update Propolis

### LINKED ISSUE #4851: Publish per-instance vCPU usage information to `oximeter`

bnaecker:
- Update propolis and related dep
- Pass instance metadata for vCPU statistics tracking
- Closes #4851

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -469,7 +469,7 @@ dependencies = [
 [[package]]
 name = "bhyve_api"
 version = "0.0.0"
-source = "git+https://github.com/oxidecomputer/propolis?rev=a9fdbc464db4cbd4ac0f43ffa0d045b4b02e0c1c#a9fdbc464db4cbd4ac0f43ffa0d045b4b02e0c1c"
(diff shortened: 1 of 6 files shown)


pfmooney (approved):

bnaecker:
Need to update `package-manifest.toml` with new Crucible / Propolis revisions, commit incoming.

leftwo (approved):

PR #5285, opened 2024-03-19
Add a target runner for running tests in VMs

andrewjstone:
This PR allows us to run tests inside falcon VMs via nextest. Due to the lack of [nextest per-test target runners](https://github.com/nextest-rs/nextest/issues/1358), we enable the target runner via environment variable and run a specific test at a time. 

In order to run a test, do the following, where the trailing `--nocapture` is optional but encouraged at this moment.

```
cd omicron
CARGO_TARGET_X86_64_UNKNOWN_ILLUMOS_RUNNER=./test-utils/src/bin/falcon_runner.sh cargo nextest run -p <package> <test name> --nocapture
```

An example: 
```
CARGO_TARGET_X86_64_UNKNOWN_ILLUMOS_RUNNER=./test-utils/src/bin/falcon_runner.sh cargo nextest run -p omicron-test-utils launch --nocapture
```

A few things remain to be done, although this is safe to merge without adding them in this PR.

* Right now, only one test can be run per test binary, although we can change this if we want. We should also error out in the case the user doesn't pass a specific test.
* We don't actually check the error result of the test in the VM. We should do this, and then automatically leave the VM running if the test fails.
* We need a small binary to allow us to login to a VM on a failed VM test. I hacked up the `solo` example in falcon for a demo, but we'll want one in Omicron for ease of use.
* We need a good way to get log files from failing tests out of the VMs. We can actually do this with our hypothetical binary and the falcon `exec` call like `<BIN> exec <VM_NAME> 'cat <FILE>'  > <LOCALFILE>`. However, a better mechanism would be to make the p9 filesystem read/write instead of read-only, or to inject ssh keys and pull the files that way.


test-utils/src/bin/falcon_runner.rs
```diff
@@ -0,0 +1,46 @@
...
+    let vm = d.node(&node_name, "helios-2.0", 2, gb(8));
+
+    let cargo_bay = tempdir()?;
+    eprintln!("Set cargo-bay to {} on host machine", cargo_bay.path());
+    let source_test_path = &args[1];
+    let test_name = &args[3];
+    let test_file_name = camino::Utf8Path::new(source_test_path)
+        .file_name()
+        .expect("Failed to get test file name");
+    let test_path = cargo_bay.path().to_path_buf().join(test_file_name);
```
smklein, line 29: When we say "cargo bay", is that just like, "convention for a thing we're mounting in the guest"?

test-utils/src/bin/falcon_runner.rs
```diff
@@ -0,0 +1,46 @@
...
+
+    d.mount(cargo_bay.path(), "/opt/cargo-bay", vm)?;
+
+    eprintln!("Launching test vm {node_name}");
+    d.launch().await?;
+    eprintln!("Launched test vm {node_name}");
+
+    eprintln!("Running test: {test_file_name}::{test_name}");
+    let run_test = format!("cd /opt/cargo-bay && chmod +x {test_file_name} && ./{test_file_name} --color never --exact {test_name} --nocapture");
+    let out = d.exec(vm, &run_test).await?;
```
smklein, line 42: This might be a falcon-ism, but what triggers the VM to be cleaned up?

test-utils/src/dev/falcon.rs
```diff
@@ -0,0 +1,13 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! Mechanisms to launch and control propolis VMs via falcon
```
smklein, line 5: Nit: Comment could probably be updated to indicate "this is an example falcon test"?

test-utils/src/dev/falcon.rs
```diff
@@ -0,0 +1,13 @@
...
+
+#[cfg(test)]
+mod test {
```
smklein, line 8: Idea, take it or leave it -- it might be kinda nice to have a way to validate "this test is being run from a VMM context, bail out if that isn't true?"

I dunno what we'd want to set in the `falcon_runner`, but just *something* to distinguish it from the host, since "getting the config right" is one of the trickiest bits to launch these VM tests right now

test-utils/src/bin/falcon_runner.rs
andrewjstone: Yeah, it's a convention for mounting. It's Ry's name of choice.

test-utils/src/bin/falcon_runner.rs
andrewjstone: When the destructor for the deployment (`d`) runs it gets cleaned up. You can change this behavior by setting `d.persistent = true`. I'm going to make that change so that we do that when tests fail.

test-utils/src/dev/falcon.rs
andrewjstone: I added a check for the hostname, which is quite unique. We can certainly inject a file though, which might work better once we start using different names for VMs. We'll want to use different names to run tests in parallel.

test-utils/src/dev/falcon.rs
andrewjstone: Done.

test-utils/src/bin/falcon_runner.rs
andrewjstone: Cleanup is now explicit, since we now leave the VM running on failure.

andrewjstone:
This is probably good enough for a real review now. I added a README to get started and also leave the VM up when a test fails.

andrewjstone:
> This is probably good enough for a real review now. I added a README to get started and also leave the VM up when a test fails.

There are few significant limitations listed in the README, but they can be resolved fairly quickly. It would probably be good to get this in to let people play with it, then figure out how we want to perform setup for more sophisticated tests that require injecting data into VMs, and for running in parallel.

test-utils/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Current Limitations
+
+* Tests must be self-contained and not rely on dependencies. E2E tests will
+not  currently work. We can easily install pre-reqs via our scripts, but this
+will delay each test run. We may instead want to build falcon base images with
+the pre-reqs included. There's a tradeoff here of test run time, vs getting the
+latest bits.
+* The name of the test `Runner` and `VM` is currently hardcoded. Therefore,
+tests can only be run serially for now. This is an easy limitation to lift, we
```
smklein, line 101: nit - the names are hard-coded, but these aren't the names used, right? It's the string including "launchpad_mcduck?"

test-utils/src/bin/falcon_runner.rs
```diff
@@ -0,0 +1,89 @@
...
+          ./{test_file_name} --color never --exact {test_name} --nocapture; \
+          res=$?; \
+        fi; \
+        echo $res"
+    );
+    let out = d.exec(vm, &run_test).await?;
+
+    // The last line of our output contains the exit code
+    let exit_code_index = out.rfind('\n').unwrap();
+    let exit_code: u8 = (&out[exit_code_index + 1..]).parse().unwrap_or(255);
```
smklein, line 64: Couple nitpicks here:

- If the `\n` character didn't exist in the output we might want to get better feedback from the test (via an `.expect` call, perhaps?)
- The indexing in `out` seems like it could easily go out-of-bounds -- what if `\n` was the last character emitted?

smklein (approved):

PR #5286, opened 2024-03-19
Networking rpw bugfixes

internet-diglett:


internet-diglett:
NOTE:

We are still `ensuring` the uplinks SMF property every cycle:

https://github.com/oxidecomputer/omicron/blob/5e9211496e93f8a0f3e751a0063eb02bf97c1d5a/nexus/src/app/background/sync_switch_configuration.rs#L384-L419

Currently we don't have any historical information to know if we've applied the update before, and sled-agent doesn't expose this SMF property directly, so we'd need to store the history (like we do with bootstore) or figure out a way to dig that out.

nexus/src/app/background/networking.rs
```diff
@@ -80,7 +80,11 @@ pub(crate) fn api_to_dpd_port_settings(
             LinkSettings {
                 params: LinkCreate {
                     autoneg: l.autoneg,
-                    lane: Some(LinkId(0)),
+                    // TODO: dendrite currently does nothing with this field.
+                    // Setting it to a value causes RPW reconciler passes to do
+                    // unnecessarily work, trying to set a field that dendrite
+                    // will not apply.
+                    lane: None,
```
internet-diglett, line 87: @rcgoodfellow Please sanity check me here, because `dendrite` _responds_ with `None` for this field whenever we check the config it has. I've started digging through dendrite's code to verify if we're really not using this field, or we're simply not sending the correct value back when we check the config.

rcgoodfellow: This does not seem right. Dendrite should be returning the lane assignment.

internet-diglett: Is it possibly due to this block?

https://github.com/oxidecomputer/dendrite/blob/41ddeab9d43d90a51e6fc1c236dc9982fc76f922/dpd/src/api_server.rs#L2761-L2770

nexus/src/app/background/sync_switch_configuration.rs
```diff
@@ -1265,6 +1341,17 @@ async fn apply_switch_port_changes(
                     }
                 };
 
+                if config_on_switch.into_inner() == dpd_port_settings {
```
rcgoodfellow, line 1344: I think this will have the desired effect. However overall implementation here is not as clear as it could be. What we want is to calculate two sets: `O` are the switch port settings in the omicron database, `D` are the switch port settings in dendrite. We can do the following operations hashing just on the switch/port names. This gives us the sets of port settings to add and remove. `/` denotes set difference and `&` denotes set intersection in what follows.

- `to_add = O / D`
- `to_del = D / O`

Additionally we have settings that may have been modified that need to be checked.

- `to_check = D & O`

We reduce these to the set that actually needs modification with a filter function

```
to_modify = to_check.filter(|d, o| d != 0)
```

I'm noting this as the filtering in this function (pretty late in the overall process) feels out of place. Looking at the `SwitchPortSettingsManager::Changes` function sheds light on why. That function is not actually computing changes, or a difference in the desired state versus the target state, it's just computing the desired state and then we get here and have to filter through things as we're making the changes.

Like I said above, I think this will have the desired effect, but I think we should consider some follow-on refactoring to simplify things a bit.

internet-diglett: I think that's a fair assessment. The initial implementation was focused on ensuring we're always enforcing what is in Nexus / Omicron, and then additional passes were made to make things more efficient (with this case unfortunately getting overlooked).

PR #5287, opened 2024-03-19
Drop the `service` table

### LINKED ISSUE #4947: retire "services" table

jgallagher:
I still need to test this on madrid and I do _not_ want to merge it before we cut the next release, but I believe this is ready for review.

Related changes / fallout also included in this PR:

* `omdb db services ...` subcommands are all gone. I believe this functionality is covered by `omdb`'s inspection of blueprints instead.
* I removed the `SledResourceKind::{Dataset,Service,Reserved}` variants that were unused. This isn't required, strictly speaking, but `SledResourceKind::Service` was intended to reference the `service` table, so I thought it was clearer to drop these for now (we can add them back when we get to the point of using them).

There are two major pieces of functionality that now _require_ systems to have a current target blueprint set:

* `DataStore::nexus_external_addresses()` now takes an explicit `Blueprint` instead of an `Option<Blueprint>`. Its callers (silo creation and reconfigurator DNS updates) fail if they cannot read the current target blueprint.
* `DataStore::vpc_resolve_to_sleds()` now _implicitly_ requires a target blueprint to be set, if and only if the VPC being queried involves control plane services. (In practice today, that means the VPC ID is exactly `SERVICES_VPC_ID`, although in the future we could have multiple service-related VPCs.) I didn't want to make this method take an explicit blueprint, because I think its most frequent callers are specifying instance VPCs, which do not need to access the blueprint.

These two together mean that deploying this change to a system without a target blueprint will result in (a) the inability to create silos or update external DNS via reconfigurator and (b) services (including Nexus) will not get the OPTE firewall rules they need to allow incoming traffic. All newly-deployed systems have a (disabled) blueprint set as of #5244, but we'll need to perform the necessary support operation to bring already-deployed systems in line.

Closes #4947.


schema/crdb/dbinit.sql
```diff
@@ -164,38 +164,32 @@ CREATE UNIQUE INDEX IF NOT EXISTS lookup_sled_by_rack ON omicron.public.sled (
...
-    -- omicron.public.service
-    'service',
     -- omicron.public.instance
-    'instance',
-    -- omicron.public.sled
-    --
-    -- reserved as an approximation of sled internal usage, such as "by the OS
-    -- and all unaccounted services".
-    'reserved'
+    'instance'
```
smklein, line 168: This makes sense, and the schema changes you have for it seem good to me.

schema/crdb/46.0.0/up2.sql
```diff
@@ -0,0 +1,17 @@
+-- Ensure there are no `sled_resource` rows with a `kind` other than 'instance'
+
+-- This is a full table scan, but the sled_resource table does not track
+-- historical, deleted resources, so is at most the size of the number of
+-- currently-running instances (which should be zero during a schema update).
+SET
+  LOCAL disallow_full_table_scans = OFF;
```
smklein, line 7: Should this be the first part of the schema change, before dropping the service table?

smklein (approved):
Looks great, happy to see this much code removed

nexus/db-queries/src/db/datastore/vpc.rs
```diff
@@ -1684,9 +1648,9 @@ mod tests {
             .await
             .expect("failed to insert blueprint");
 
-        // We haven't set a blueprint target yet, so we should still only see
-        // the two RSS-inserted service-running sleds.
-        assert_eq!(&harness.sled_ids[..2], fetch_service_sled_ids().await);
+        // We haven't set a blueprint target yet, so we should still fail to see
+        // any services on sleds.
+        assert_eq!(&[] as &[Uuid], fetch_service_sled_ids().await);
```
sunshowers, line 1653: Could you turn this into an `assert_service_sled_ids(expected: &[Uuid])`?

schema/crdb/46.0.0/up2.sql
jgallagher: Yes, yes it should. Alternatively, we could drop it entirely:

* I checked dogfood and Augustus checked colo to confirm all `sled_resource`s are `instance`.
* Removing the enum variants on the Rust side had no followon effects (i.e., we were never creating them in real code or tests)

so it's probably overly cautious to include?

schema/crdb/46.0.0/up3.sql
```diff
@@ -0,0 +1,3 @@
+-- We are dropping `kind` so that we can drop the `sled_resource` kind; we'll
+-- then recreate it (with some variants removed) and add this column back.
+ALTER TABLE omicron.public.sled_resource DROP COLUMN IF EXISTS kind;
```
sunshowers, line 3: Hmm, I think this needs to be in its own schema version (e.g. 46.0.1).

If `up3.sql` happens, then the migration fails, we'll attempt to repeat `up2.sql` -- which will fail forever.

schema/crdb/46.0.0/up5.sql
```diff
@@ -0,0 +1,3 @@
+CREATE TYPE IF NOT EXISTS omicron.public.sled_resource_kind AS ENUM (
+    'instance'
+);
```
sunshowers, line 3: I think this also needs to be its own schema version (46.0.2).

sunshowers (changes requested):
requesting changes only for the schema stuff

sunshowers:
We discussed this in `#oxide-update`, and we're going to change it so that we'll update the internal DB tracking of schema versions after each individual _file_ gets applied, not just after each _directory_. So folks in the future won't have to try and reason about idempotency across multiple files. (They will still have to think about idempotency for each individual file, but hopefully that will be much easier to reason about.)

schema/crdb/46.0.0/up2.sql
jgallagher: Swapped up1 and up2 in dc18f5758

nexus/db-queries/src/db/datastore/vpc.rs
jgallagher: Done in 7de458e49; that is much nicer - thanks!

jgallagher:
This is caught up with `main` which now includes #5293, so I believe it's okay to leave all these migrations in one major version / directory now.

nexus/src/app/silo.rs
```diff
@@ -100,7 +100,14 @@ impl super::Nexus {
             .blueprint_target_get_current_full(opctx)
             .await
             .internal_context("loading target blueprint")?;
-        let target = target_blueprint.as_ref().map(|(_, blueprint)| blueprint);
+        let target = target_blueprint
```
davepacheco, line 103: Maybe we want to have `blueprint_target_get_current_full()` return this 500 itself now that we expect all systems to have a target blueprint.

schema/crdb/drop-service-table/up7.sql
```diff
@@ -0,0 +1,3 @@
+SET
+  LOCAL disallow_full_table_scans = OFF;
+UPDATE omicron.public.sled_resource SET kind = 'instance';
```
davepacheco, line 3: Is this worse than having up6.sql add a `NOT NULL` and `DEFAULT` value and then removing the default here?  I would expect the approach here rewrites the `sled_resource` table, whereas setting a default and then removing it wouldn't.

davepacheco (approved):
Great!

In #4947 I wrote:

> Given that, I think we can just drop the services table altogether. There is only one thing that uses it as far as I know, and that's DNS propagation.

Boy did that turn out to be wrong.  Sorry!

sunshowers (approved):
Awesome!

nexus/src/app/silo.rs
jgallagher: This had more knock-on effects than I expected, but I think basically all positive: [78f9429](https://github.com/oxidecomputer/omicron/pull/5287/commits/78f942908af8d9535483f3e75466c24944a19de1)

schema/crdb/drop-service-table/up7.sql
jgallagher: ... yes, yes it is. Changed in [88aaf03](https://github.com/oxidecomputer/omicron/pull/5287/commits/88aaf03571656ed0020322a4bc7970a9a4f91d25)

dev-tools/omdb/tests/successes.out
```diff
@@ -71,45 +71,6 @@ loading inventory collections ... done.
 loading blueprints ... done.
 wrote <REDACTED>
 =============================================
-EXECUTING COMMAND: omdb ["db", "services", "list-instances"]
```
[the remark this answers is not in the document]
jgallagher: I think this is covered by our `blueprint.display()` output. I added a test for `nexus blueprint show <ID>` in [fcf547a](https://github.com/oxidecomputer/omicron/pull/5287/commits/fcf547adbacf1a644df5cab4ed478a5a28f243a7), then realized we were still using debug instead of `.display()` so tweaked that in [3fd0288](https://github.com/oxidecomputer/omicron/pull/5287/commits/3fd02881ba2fb47e4ebb804dc3cf8bda5fd03922).

Issue #5292, opened 2024-03-20
Save DB schema upgrade progress incrementally


smklein:
Came up in `#oxide-update` today, but some schema changes will require multiple versions to be correct. (Came up for #5032 earlier, and also for #5287 which requires three migrations.) It would be nice to be able to express an ordered list of versions this way, maybe within a single directory have a bunch of subdirectories.

_Originally posted by @sunshowers in https://github.com/oxidecomputer/omicron/pull/5274#discussion_r1531062927_

PR #5293, opened 2024-03-20
[nexus] During DB schema update, avoid re-applying sub-updates

smklein:
Fixes https://github.com/oxidecomputer/omicron/issues/5292


nexus/db-queries/src/db/datastore/db_metadata.rs
```diff
@@ -213,93 +215,157 @@ impl DataStore {
...
+            // update), but the "target_version" will keep shifting on each
+            // incremental step.
+            let mut prior_step_version = None;
+            for (i, SchemaUpgradeStep { path: _, sql }) in
+                schema_change.steps.iter().enumerate()
+            {
+                // Durably store our progress in the target version as the
+                // pre-release extension of "target_version".
+                target_version.0.pre =
+                    semver::Prerelease::new(&format!("step.{i}"))
```
jgallagher, line 290: It might be worth a comment that the `.` here is load-bearing (I had initially suggested `step-{i}`, which would be incorrect). From https://docs.rs/semver/1.0.22/semver/struct.Prerelease.html#examples:

> Be aware that if you are planning to number your own pre-releases, you should prefer to separate the numeric part from any non-numeric identifiers by using a dot in between. That is, prefer pre-releases alpha.1, alpha.2, etc rather than alpha1, alpha2 etc. The SemVer spec’s rule for pre-release precedence has special treatment of numeric components in the pre-release string, but only if there are no non-digit characters in the same dot-separated component. So you’d have alpha.2 < alpha.11 as intended, but alpha11 < alpha2.

nexus/db-queries/src/db/datastore/db_metadata.rs
```diff
@@ -213,93 +215,157 @@ impl DataStore {
...
         }
 
         let target_versions = all_versions.range((
             Bound::Excluded(&current_version),
             Bound::Included(&desired_version),
         ));
 
-        for target_version in target_versions.into_iter() {
+        // Iterate over semver versions (e.g., "1.0.0 -> 2.0.0").
+        for mut target_version in target_versions.into_iter().map(|v| v.clone())
```
jgallagher, line 259: Tiny nit:

```suggestion
        for mut target_version in target_versions.into_iter().cloned()
```

nexus/db-queries/src/db/datastore/db_metadata.rs
```diff
@@ -213,93 +215,157 @@ impl DataStore {
...
+                    // restart, for whatever reason. We skip all the incremental
+                    // steps that we know have completed.
+                    if found_target_version > &target_version {
+                        warn!(
+                            log,
+                            "Observed target version greater than this upgrade step. Skipping.";
+                            "observed_target_version" => %found_target_version,
+                            "attempted_target_version" => %target_version,
+                        );
+                        continue;
```
jgallagher, line 313: Do we need to set `prior_step_version` before `continue`ing? If so, could we rework the tests to fail without doing so? (That might be easier after #5274, when we can pass in the schema versions instead of a directory to read.) This sequence seems like it won't work:

* Update to `N.0.0-step.1` is successful
* Update to `N.0.0-step.2` fails; we bail out
* Next attempt, we land here and note that `N.0.0-step.2 > N.0.0-step.1`, so we `continue`; `prior_step_version` is still `None`
* We move on to `N.0.0-step.2`, and call `prepare_schema_update`. Because `prior_step_version` is `None`, its `valid_prior_targets` list only includes `N.0.0`, not `N.0.0-step.1`, so it fails.

nexus/db-queries/src/db/datastore/db_metadata.rs
```diff
@@ -324,44 +390,68 @@ impl DataStore {
             // Now that the schema change has completed, set the following:
             // - db_metadata.version = new version
             // - db_metadata.target_version = NULL
-            self.finalize_schema_update(&current_version, &target_version)
-                .await
-                .map_err(|e| e.to_string())?;
+            //
+            // The "version" field ignores all this pre-release versioning that we
+            // abuse in the "target_version" field.
```
jgallagher, line 395: At the top of this loop, should we check and bail out if the incoming `target_version` already has a prerelease tag? I don't think anyone is likely to put a prerelease tag on a schema migration version, but if they did it definitely wouldn't behave as they'd expect after these changes.

smklein:
> Just dropping some notes here from chat -- thanks @jgallagher.
> 
> Here's a somewhat more concrete example of the **problem** we're trying to solve: say we're at 46.0.0 and we want to move to 47.0.0. Say that transition has 8 steps and that step 6 invalidates the preconditions for step 2. (This is the key. It does not seem common but it might be easy to accidentally do.) Today if this happens, and then Nexus crashes, then when it resumes it'll try to re-execute all the steps, and so it'll re-execute step 2, but it'll fail (by construction, because step 6 has broken step 2's preconditions). Then it'll be stuck unable to move forward or backward.
> 
> Previous to this PR, we'd set target_version = 47.0.0 at the start of the migration from 46.0.0 to 47.0.0 Then we'd go apply a bunch of SQL transactions, each conditional on the target being 47.0.0. This PR addresses the problem above by including the step number in the target version (so that we store target version `47.0.0-step.6`) and, when resuming, skipping any steps that we know have been applied (because the target version is newer).
> 
> As I understand it we're **not** preventing the system from applying the same update step twice. That can still happen if two Nexus instances are doing it concurrently. The new guarantee that's being provided is that if you're updating from, say, 46.0.0 to 47.0.0 and you're talking about step 6, this step will only ever be applied to a database in one of two schema states:
> 
> * 46.0.0 plus steps 1-5 of the 46.0.0 -> 47.0.0 migration
> * 46.0.0 plus steps 1-6 of the 46.0.0 -> 47.0.0 migration
> 
> (whereas previously it could have been applied to a database at 46.0.0 plus steps 1-7 or steps 1-8 of the 46.0.0 -> 47.0.0 migration).

Totally agreed with this -- fortunately, we do have tests for "single steps being idempotent" already, just not "going backwards from step 6 to step 2".

nexus/db-queries/src/db/datastore/mod.rs
```diff
@@ -214,7 +214,7 @@ impl DataStore {
                 {
                     Ok(()) => return Ok(()),
                     Err(e) => {
-                        warn!(log, "Failed to ensure schema version: {e}");
+                        warn!(log, "Failed to ensure schema version"; "error" => ?e);
```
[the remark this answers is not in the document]
smklein: Ah, rad, happy to use `#%` - I tried using `%` and noticed that it didn't include the full error chain, but if `#%` does, I'll use it

nexus/db-queries/src/db/datastore/db_metadata.rs
```diff
@@ -167,9 +168,12 @@ impl DataStore {
         log: &Logger,
         desired_version: SemverVersion,
         config: Option<&SchemaConfig>,
-    ) -> Result<(), String> {
```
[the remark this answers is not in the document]
smklein: Already done, merging now

nexus/db-queries/src/db/datastore/db_metadata.rs
```diff
@@ -213,93 +215,157 @@ impl DataStore {
...
+            );
         }
 
         let target_versions = all_versions.range((
             Bound::Excluded(&current_version),
             Bound::Included(&desired_version),
         ));
 
-        for target_version in target_versions.into_iter() {
+        // Iterate over semver versions (e.g., "1.0.0 -> 2.0.0").
```
sunshowers, line 258: Worth making it clear that we're iterating over real versions here (i.e. directories), not prerelease versions (files)?

nexus/db-queries/src/db/datastore/db_metadata.rs
```diff
@@ -744,4 +864,103 @@ mod test {
         crdb.cleanup().await.unwrap();
         logctx.cleanup_successful();
     }
+
+    #[tokio::test]
+    async fn schema_version_subcomponents_save_progress() {
```
sunshowers, line 869: Does the `versions_have_idempotent_up` test need to be updated to test by file rather than by directory?

sunshowers (commented):
Thanks for doing this! So much easier to reason about.

nexus/db-queries/src/db/datastore/db_metadata.rs
```diff
@@ -370,9 +460,15 @@ impl DataStore {
         &self,
         from_version: &SemverVersion,
         to_version: &SemverVersion,
+        previous_step_version: Option<&SemverVersion>,
     ) -> Result<(), Error> {
         use db::schema::db_metadata::dsl;
 
+        let mut valid_prior_targets = vec![Some(to_version.to_string())];
```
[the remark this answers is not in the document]
smklein: Good catch, the options here don't need to make their way down to diesel. I can flatten this out.

nexus/db-queries/src/db/datastore/db_metadata.rs
smklein: Okay, keep me honest here - I'm on board with a way to make this simpler, but:

I think it's okay to keep the `prior_step_version` as "the latest version we think **our Nexus** has applied".

@jgallagher , in the case you mention, once we get to that fourth step (calling `prepare_schema_update`) , you're right that our `prior_step_version` is `None`, but the `to_version` parameter would still be `N.0.0-step.2` (we're trying to incrementally work on that step) so it would be included in `valid_prior_targets`. That would also be the target version stored in the DB, since that's the last thing we updated before failing on the prior attempt to update the schema.

(I think this was true before my merge as well, but less obvious, because I was mutating `target_version`. Now, I'm using a different variable named `target_step_version` to make this a little more clear).

smklein: Also, the test I added below really should be catching cases like this (and has, locally, when I've gotten it wrong) because it does trigger this "Fail a step and force retry" behavior.

jgallagher: > `to_version` parameter would still be `N.0.0-step.2`

Is this guaranteed? What if we succeed in updating to `N.0.0-step.1`, and then `prepare_schema_update` fails to set the new target version of `N.0.0-step.2`?

nexus/db-queries/src/db/datastore/mod.rs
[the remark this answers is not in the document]
smklein: Fixed

nexus/db-queries/src/db/datastore/db_metadata.rs
smklein: Good call, I'll do this to be defensive.

nexus/db-queries/src/db/datastore/db_metadata.rs
smklein: If that's the case, we'd re-apply "step 1", redoing the work of `N.0.0-step.1`, which should be fine since individual steps must be idempotent anyway, right?

Also, regardless: I'm taking this feedback, and refactored a bit of this to try to make it easier to follow -- though I think the functionality is the same.

nexus/db-queries/src/db/datastore/db_metadata.rs
smklein: Update the comments to make this more clear!

nexus/db-queries/src/db/datastore/db_metadata.rs
smklein: I think it's still okay as-is -- that particular test, as well as some of the other schema tests, are "testing the schema changes" more than they are "testing the mechanism to apply schema changes".

I say this referencing the following block of code:

https://github.com/oxidecomputer/omicron/blob/a655ff2a26b1c58335d4ecc17e8263df73045063/nexus/tests/integration_tests/schema.rs#L141-L150

Which appears to already be re-running each individual step, rather than "the whole directory at once".

nexus/db-queries/src/db/datastore/db_metadata.rs
```diff
@@ -13,13 +13,93 @@ use async_bb8_diesel::{AsyncRunQueryDsl, AsyncSimpleConnection};
...
+    // Drops the pre-release information about this target version.
+    //
+    // This is the version we are upgrading to, using these incremental steps.
+    fn without_prerelease(&self) -> SemverVersion {
+        let mut target_version = self.version.clone();
+        target_version.0.pre = semver::Prerelease::EMPTY;
+        target_version
+    }
+
+    fn previous(&self) -> anyhow::Result<Self> {
```
jgallagher, line 65: Tiny nit - I'd maybe make this return `Option<Self>` directly, since the only caller of it immediately calls `.ok()` to do that conversion.

nexus/db-queries/src/db/datastore/db_metadata.rs
jgallagher: Yes, you're right. Thanks for sticking with me on this one, and for the refactor - I find it more clearly correct for sure.

jgallagher (approved):

nexus/db-queries/src/db/datastore/db_metadata.rs
smklein: Sounds good, updated

nexus/db-queries/src/db/datastore/db_metadata.rs
sunshowers: Got it, thanks!

smklein: Although maybe that loop needs to be inverted? Seems like we want to apply each step `times_to_apply` times, not the whole set....

smklein: I updated this in 36fdb610b201a4195ae36740eea9edc2cdd1d56b - I think you're right, the test was checking the wrong thing before, and should be fixed now.

smklein: Okay, this requires more elaboration. When I updated this test, I noticed that it actually started failing.

Specifically, this version threw an error:

https://github.com/oxidecomputer/omicron/blob/a655ff2a26b1c58335d4ecc17e8263df73045063/schema/crdb/10.0.0/up05.sql#L1

As, once the value is dropped, it no longer exists.

Postgres supports syntax to ALTER TYPE DROP VALUE IF EXISTS: https://www.postgresql.org/docs/current/sql-altertype.html

But CockroachDB does not: https://www.cockroachlabs.com/docs/stable/alter-type

To mitigate:
- I have filed a bug against CockroachDB as a feature request: https://github.com/cockroachdb/cockroach/issues/120801
- In 28e6876910f3395e2109ebac53885ad8fdfcf182 , I make an exemption for this schema version

PR #5294, opened 2024-03-20
Restore instance tests to #5172

andrewjstone:
Restore instance tests added in #4691 so that they work with the latest updates in #5172. The tests now use the new `StorageManagerTestHarness` from #5172 instead of the since removed `FakeStorageManager`. Additionally, the dns server code from `sled-agent/src/fakes/nexus` is used rather than the reimplementation in #4691. This shrinks the amount of code added with these tests.

### DIFF
--- a/sled-agent/src/instance.rs
+++ b/sled-agent/src/instance.rs
@@ -1689,7 +1689,6 @@ mod tests {
         (srv, client)
     }
 
-    // make a FakeStorageManager with a "U2" upserted
     async fn setup_storage_manager(log: &Logger) -> StorageManagerTestHarness {
         let mut harness = StorageManagerTestHarness::new(log).await;
         let raw_disks =
--- a/sled-agent/src/instance.rs
+++ b/sled-agent/src/instance.rs
@@ -1520,7 +1520,7 @@ impl InstanceRunner {
     }
 }
 
-#[cfg(test)]
+#[cfg(all(test, target_os = "illumos"))]
 mod tests {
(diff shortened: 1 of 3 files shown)


sled-agent/src/instance.rs
```diff
@@ -1519,3 +1519,635 @@ impl InstanceRunner {
...
+
+        let client = propolis_client::Client::new(&format!(
+            "http://{}",
+            srv.local_addr()
+        ));
+
+        (srv, client)
+    }
+
+    // make a FakeStorageManager with a "U2" upserted
```
andrewjstone, line 1692: Need to remove this stale comment

sled-agent/src/instance.rs
```diff
@@ -1519,3 +1519,634 @@ impl InstanceRunner {
         out
     }
 }
+
+#[cfg(test)]
+mod tests {
```
smklein, line 1524: As discussed in chat - this should be illumos-only if it's using the `StorageManagerTestHarness`

smklein (approved):
Thank you for restoring this! I'm thankful to not need to revive the FakeStorageManager

sled-agent/src/instance.rs
andrewjstone: Done in [b7c1d67](https://github.com/oxidecomputer/omicron/pull/5294/commits/b7c1d67c7c215d6e231a2728e4d7b049e5706637)

PR #5297, opened 2024-03-21
BUGFIX: Bootstore BGP configuration

internet-diglett:
* Implement a more straightforward mechanism for generating bootstore bgp peer config from db data
* Filter bootstore interface peer config data by switch and port number


nexus/src/app/background/sync_switch_configuration.rs
```diff
@@ -301,7 +300,6 @@ impl BackgroundTask for SwitchPortSettingsManager {
                     },
                 };
 
-                info!(&log, "applying switch port config changes"; "changes" => ?changes);
```
rcgoodfellow, line 304: Just a high level note that `SwitchPortSettingsManager::activate` should be broken out into smaller functions as it's over 700 lines now.

rcgoodfellow:
I ran a diff between the bootstore config sled agent generates at v1 versus what's updated by omicron at v2, here is what that looks like.

```diff
git diff v1.json v2.json
diff --git a/v1.json b/v2.json
index fc760fd..0e85061 100644
--- a/v1.json
+++ b/v2.json
@@ -1,5 +1,5 @@
 {
-  "generation": 1,
+  "generation": 2,
   "schema_version": 1,
   "body": {
     "ntp_servers": [
@@ -24,11 +24,11 @@
               "asn": 47,
               "port": "qsfp0",
               "addr": "198.51.101.1",
-              "hold_time": null,
-              "idle_hold_time": null,
-              "delay_open": null,
-              "connect_retry": null,
-              "keepalive": null
+              "hold_time": 6,
+              "idle_hold_time": 3,
+              "delay_open": 0,
+              "connect_retry": 3,
+              "keepalive": 2
             }
           ],
           "autoneg": false
@@ -47,11 +47,11 @@
               "asn": 47,
               "port": "qsfp0",
               "addr": "198.51.101.9",
-              "hold_time": null,
-              "idle_hold_time": null,
-              "delay_open": null,
-              "connect_retry": null,
-              "keepalive": null
+              "hold_time": 6,
+              "idle_hold_time": 3,
+              "delay_open": 0,
+              "connect_retry": 3,
+              "keepalive": 2
             }
           ],
           "autoneg": false
@@ -70,11 +70,11 @@
               "asn": 48,
               "port": "qsfp0",
               "addr": "198.51.101.5",
-              "hold_time": null,
-              "idle_hold_time": null,
-              "delay_open": null,
-              "connect_retry": null,
-              "keepalive": null
+              "hold_time": 6,
+              "idle_hold_time": 3,
+              "delay_open": 0,
+              "connect_retry": 3,
+              "keepalive": 2
             }
           ],
           "autoneg": false
@@ -91,13 +91,13 @@
           "bgp_peers": [
             {
               "asn": 48,
-              "port": "qsfp0",
+              "port": "qsfp1",
               "addr": "198.51.101.13",
-              "hold_time": null,
-              "idle_hold_time": null,
-              "delay_open": null,
-              "connect_retry": null,
-              "keepalive": null
+              "hold_time": 6,
+              "idle_hold_time": 3,
+              "delay_open": 0,
+              "connect_retry": 3,
+              "keepalive": 2
             }
           ],
           "autoneg": false
```

nexus/src/app/background/sync_switch_configuration.rs
```diff
@@ -707,33 +694,54 @@ impl BackgroundTask for SwitchPortSettingsManager {
...
+                bgp.dedup();
+
                 let mut ports: Vec<PortConfigV1> = vec![];
 
                 for (location, port, change) in &changes {
                     let PortSettingsChange::Apply(info) = change else {
                         continue;
                     };
 
+                    let peer_configs = match self.datastore.bgp_peer_configs(opctx, *location, port.port_name.clone()).await {
```
rcgoodfellow, line 706: So this is where the primary fix is? Filtering the peers down based on switch location and port in this function which uses the new DB view?

nexus/src/app/background/sync_switch_configuration.rs
```diff
@@ -801,7 +809,9 @@ impl BackgroundTask for SwitchPortSettingsManager {
                             rack_subnet: subnet,
                             infra_ip_first,
                             infra_ip_last,
+                            // FIX THIS
```
rcgoodfellow: What is this and the one below in reference to? The things to fix in this PR?

nexus/src/app/background/sync_switch_configuration.rs
rcgoodfellow: But not in this PR.

nexus/src/app/background/sync_switch_configuration.rs
internet-diglett: Yep. Instead of trying to build it from `SwitchPortSettingsCombinedResult`, we make a view where we query exactly what we want, for exactly what we're working on.

In hindsight, I should have taken this approach from the beginning for BGP configurations, and maybe even basic dendrite switch port settings. I'll definitely lean towards this in the future refactor.

nexus/src/app/background/sync_switch_configuration.rs
internet-diglett: Yes, that's what they were for. I was getting really tired and wanted to make sure I had visual indicators for myself (and others) of which parameters were being corrupted / malformed by the old logic.

Issue #5284, opened 2024-03-18
Reconfigurator: Removing services should unregister them from oximeter


jgallagher:
The only service that currently registers itself as a metrics producer is Nexus, so at least for Nexus, we need to:

1. Remove the corresponding row in `metric_producer` for the service.
2. Notify the assigned Oximeter collector that it should stop polling the producer.

Expanding some details on these, from poking around and chatting with @bnaecker:

1.  `metric_producer.id` is the service ID, when `metric_producer.kind = 'service'`. I believe the only consumer of this table is when an Oximeter collector [posts itself to Nexus](https://github.com/oxidecomputer/omicron/blob/274edc009d979e4443112ab56095444219f60029/oximeter/collector/src/lib.rs#L258-L263), Nexus notifies it [of all the producers it should be polling](https://github.com/oxidecomputer/omicron/blob/274edc009d979e4443112ab56095444219f60029/nexus/src/app/oximeter.rs#L90-L147). 
2. Instances do this today by calling [`Nexus::unassign_producer`](https://github.com/oxidecomputer/omicron/blob/274edc009d979e4443112ab56095444219f60029/nexus/src/app/oximeter.rs#L202-L206). Reconfigurator will need to do the same (which will require some refactoring to make that method usable from the blueprint execution background task).

bnaecker:
@andrewjstone, @smklein, @jgallagher and I discussed the issue of deregistering an `oximeter` producer today in a video call.

#### The problem

The main issue brought up is how to deregister `oximeter` producers when we remove or move a service. Today, any services that want to produce metric data register with Nexus. Nexus finds an available `oximeter-collector` from a table in the DB; records the assignment of that collector to that producer in another table; and then notifies the `oximeter-collector` itself so that it can start collecting from the producer.

But what happens when a service wants to deregister? Or more problematically, what if a service simply goes away? Right now, the only time this happens is on instance shutdown, when Nexus undoes all the above (remove the records, and tell `oximeter-collector` to stop). But in general, services may not deregister or may not be able to deregister, and we'd like to have that happen automatically.

#### Proposal

The proposal we all agreed on uses the idea of a lease. Upon registering with Nexus, the producer gets back a lease interval from Nexus. The timestamp of this registration is also recorded in Nexus. (I actually think it already is.) The producer will be required to re-register within some fixed fraction of that interval (say 1/4th or 1/10th) in order for collection to continue.

[...]

PR #5311, opened 2024-03-22
Add a (disabled) RPW to garbage collect stale Oximeter producers

jgallagher:
This is the RPW portion of #5284. It's currently statically disabled (via the `if true {}` block at the top of `MetricProducerGc::activate()`); we should remove that block once [all Oximeter producers start refreshing leases](https://github.com/oxidecomputer/omicron/issues/5284#issuecomment-2008291919).

This picks a 10 minute lease expiration time - a complete WAG. Happy to raise or lower that if needed.


nexus-config/src/nexus_config.rs
```diff
@@ -395,6 +397,15 @@ pub struct DnsTasksConfig {
     pub max_concurrent_server_updates: usize,
 }
 
+#[serde_as]
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+pub struct MetricsProducerGcConfig {
+    /// period (in seconds) for periodic activations of the background task that
+    /// reads garbage collects metrics producers whose leases have expired
```
andrewjstone, line 404:
```suggestion
    /// garbage collects metrics producers whose leases have expired
```

nexus-config/src/nexus_config.rs
```diff
@@ -353,6 +353,8 @@ pub struct BackgroundTaskConfig {
     pub dns_internal: DnsTasksConfig,
     /// configuration for external DNS background tasks
     pub dns_external: DnsTasksConfig,
+    /// configuration for metrics producer GC background task
```
andrewjstone, line 356: worth spelling out "garbage collection" instead of `GC`?

nexus/metrics-producer-gc/src/lib.rs
```diff
@@ -0,0 +1,369 @@
...
+
+        // GC'ing expired producers with an expiration time _newer_ than our
+        // producer's `time_modified` should prune our one producer and notify
+        // the collector that it's doing so.
+        collector.expect(
+            Expectation::matching(request::method_path(
+                "DELETE",
+                format!("/producers/{}", producer.id()),
+            ))
+            .respond_with(status_code(404)),
```
bnaecker, line 349: I was surprised to see a 404 here. But it looks like the pruning method only logs the error, right?

nexus/src/app/background/metrics_producer_gc.rs
```diff
@@ -0,0 +1,116 @@
...
+                "Metric producer GC: statically disabled pending omicron#5284"
+            );
+            return json!({
+                "error": "metric producer gc disabled (omicron#5284)",
+            });
+        }
+
+        let Some(expiration) = TimeDelta::from_std(self.lease_duration)
+            .ok()
+            .and_then(|delta| Utc::now().checked_add_signed(delta))
```
bnaecker, line 48: Should this be a checked subtraction? I might be missing something, but this is always in the future since the lease duration is positive, right?

nexus/src/app/oximeter.rs
```diff
@@ -28,6 +28,12 @@ use std::num::NonZeroU32;
 use std::time::Duration;
 use uuid::Uuid;
 
+/// How long a metrics producer remains registered to a collector.
+///
+/// Producers are expected to renew their registration lease periodically, at
+/// some interval of this overall duration.
+pub const PRODUCER_LEASE_DURATION: Duration = Duration::from_secs(10 * 60);
```
bnaecker, line 35: Yeah, this is a WAG, but I think it's good for now

bnaecker (approved):
Looks solid to me! I've a few questions and one possibly-load-bearing issue with the expiration computation, but I think it's otherwise good to go. Thanks for doing this!

nexus/db-queries/src/db/datastore/oximeter.rs
```diff
@@ -159,4 +164,194 @@ impl DataStore {
...
+        paginated(dsl::metric_producer, dsl::id, pagparams)
+            .filter(dsl::time_modified.lt(expiration))
+            .order_by((dsl::oximeter_id, dsl::id))
+            .select(ProducerEndpoint::as_select())
+            .load_async(&*self.pool_connection_authorized(opctx).await?)
+            .await
+            .map_err(|e| {
+                public_error_from_diesel(
+                    e,
+                    ErrorHandler::Conflict(
```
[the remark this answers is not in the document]
jgallagher: Yeah good catch - lazy copy/paste mistake on my part.

nexus/metrics-producer-gc/src/lib.rs
jgallagher: Ah, I think this is leftover from when notifying the collector was more load bearing. If we try to delete a nonexistent producer, oximeter will return a 404 (reasonable), so I was testing that we succeeded whether we got a 204 or a 404, and then kept the wrong one when I was cleaning up. I'll switch this to 204.

You are correct that we only log an error - it's a one shot / best-effort optimization.

andrewjstone (approved):
Looks great!

nexus/src/app/background/metrics_producer_gc.rs
jgallagher: ... yes, yes it should. And it looks like I need at least a basic test here of `activate` to catch this.

jgallagher: Test added (and confirmed this should be `sub`) in [7e60b54](https://github.com/oxidecomputer/omicron/pull/5311/commits/7e60b54df48846c1265fdde17944d8214adc132e)

nexus/src/app/oximeter.rs
bnaecker, line 35: I'm working on the producer-renewal side of things, and I think we should find a way to make this public and easily-accessible in `oximeter-producer`. The producer needs to renew within some fraction of this time, and I'm a bit worried that we'll change this and forget to change that renewal period. `oximeter-producer` already depends on the grab-bag that is `omicron-common`, so that might be a good place.

jgallagher: I think at one point we mentioned having nexus return this when a producer registers - if we did that, would it remove the need to share the constant? I share your worry, but am also concerned about a future world where we can update services independently; a change to a constant like this wouldn't be an API break we'd detect at the progenitor level, so it could sneak in. But if Nexus told producers what to use, we'd only have to worry about changing it in Nexus, I think?

PR #5312, opened 2024-03-22
Add MGS `sp_sensor_read_value` endpoint

jgallagher:
This exposes the `faux-mgs read-sensor-value` command. Sensor IDs currently come from the host OS via topo.

Example on `madrid`:

```console
root@oxz_switch1:~# curl http://[::1]:12225/sp/sled/14/sensor/value/111
{"timestamp":1566837559,"result":{"kind":"success","value":60.0}}
```

(cc @rmustacc)


gateway/src/http_entrypoints.rs
```diff
@@ -298,6 +298,45 @@ struct UpdatePreparationProgress {
...
+    Clone,
+    Copy,
+    PartialEq,
+    PartialOrd,
+    Serialize,
+    Deserialize,
+    JsonSchema,
+)]
+pub struct SpSensorReading {
+    /// SP-centric timestamp.
```
bnaecker, line 313: Maybe a bit more detail? Does it have units? Is it only useful as to compute deltas, or is the absolute value meaningful?

gateway/src/http_entrypoints.rs
```diff
@@ -625,6 +674,28 @@ async fn sp_startup_options_set(
     Ok(HttpResponseUpdatedNoContent {})
 }
 
+/// Read the current value of a sensor by ID
+///
+/// Sensor IDs come from the host topo tree.
```
bnaecker, line 679: 🤘 this is awesome

gateway/src/http_entrypoints.rs
```diff
@@ -625,6 +674,28 @@ async fn sp_startup_options_set(
...
+#[endpoint {
+    method = GET,
+    path = "/sp/{type}/{slot}/sensor/value/{sensor_id}",
```
bnaecker, line 682: Slight API nit, but I kind of expected the `{sensor_id}` and `value` path components to be flipped. I think, anyway? Actually, is there anything to `GET` on the sensor _other_  than the value? Could we get rid of the `value` path component entirely?

gateway/src/http_entrypoints.rs
```diff
@@ -298,6 +298,45 @@ struct UpdatePreparationProgress {
...
+    Serialize,
+    JsonSchema,
+)]
+#[serde(tag = "kind", rename_all = "snake_case")]
+pub enum SpSensorReadingResult {
+    Success { value: f32 },
+    DeviceOff,
+    DeviceError,
+    DeviceNotPresent,
+    DeviceUnavailable,
```
bnaecker, line 336: What is the distinction between these two error types?

bnaecker (approved):
Couple of questions and a nit, but looks great to me! Thanks, this is going be excellent to collect through `oximeter`.

gateway/src/http_entrypoints.rs
jgallagher: I was intentionally vague here because it depends on what the SP is using as a timestamp mechanism, but maybe that's not very useful. Today, this is "milliseconds since the last SP boot", which I think is primarily useful to compute deltas (within the same SP as long as it hasn't rebooted), although the absolute value is incidentally meaningful (i.e., time since last boot). I don't know that this is a committed, stable guarantee on the SP side, though; if we started injecting NTP time into the SP, maybe this value would change, for example.

I'd be happy to expand the comment to explain what this value means today, maybe with a note that it could change with future SP releases?

gateway/src/http_entrypoints.rs
jgallagher: They definitely should be flipped - not sure what I was thinking there.

There's nothing to `GET` on the sensor today other than the value, but I wondered if we might add something like "get the description" or "get the units" in the future. I could see an argument for making this just `.../sensor/{sensor_id}`, and if we add metadata like that in the future, it could hang off of `.../sensor/{sensor_id/description` or whatever?

gateway/src/http_entrypoints.rs
jgallagher: The comments on these are directly lifted from the [hubris enum defining them](https://github.com/oxidecomputer/hubris/blob/1145c393818d70f9c73670540a7e90c556ef41ae/task/sensor-types/src/lib.rs#L70-L76) :sweat_smile:.

I think it is up to the tasks feeding data into the `sensor` task to decide what these mean. From the `thermal` task, I believe these are:

* `DeviceNotPresent` is sent when a "dynamic input" (e.g., a transceiver) has been removed
* `DeviceUnavailable` is sent when we expect there to be a sensor present, but we haven't gotten any data from it

bnaecker: Oh that makes sense, for sensors where we have an out-of-band presence mechanism like the QSFPs. Thanks!

gateway/src/http_entrypoints.rs
bnaecker: Yeah, if you'd like to leave it open to get future data under the sensor path, then  `.../sensor/{sensor_id}/value` makes sense!

gateway/src/http_entrypoints.rs
bnaecker: I think a quick note would be useful, maybe something about units and that it should probably only be used for deltas at this point. Out of curiosity, does the SP cache any measurements? Or is it reading a sensor directly in response to the request from the host?

jgallagher: It is caching measurements; under the hood, this request sends a [SensorRequestKind::LastReading](https://github.com/oxidecomputer/management-gateway-service/blob/fda8faaff9774dafed08c995e19a402efdad0cfb/gateway-sp-comms/src/single_sp.rs#L807), which ultimately lands in the hubris sensor task here: https://github.com/oxidecomputer/hubris/blob/8c42218796329eb48c9d50c93b5db8aba776310a/task/sensor/src/main.rs#L104-L112. The way the `sensors` task is structured, I believe it cannot cause a sensor to be read; it can only report readings (as we're doing here) or accept posts of updated values from other tasks.

All that to say - if you request a value from the same sensor multiple times faster than whatever task is checking that sensor, you can definitely expect to get back identical results (including identical timestamps, since it's the timestamp of the last data posted to the `sensor` task).

jgallagher: :+1: I put a longer comment in explaining what it is today, plus a note that it might change in a future SP release. Kinda vague, but maybe will serve as a hint if we change this in the future and it surprises someone.

PR #5313, opened 2024-03-22
add bfd to bootstore and early networking

rcgoodfellow:
BFD information is not currently in the bootstore, and early networking is not BFD aware. This means early networking will not set up BFD during upgrades for systems that have BFD configured, or during scrimlet cold-boot situations.

### DIFF
--- a/openapi/bootstrap-agent.json
+++ b/openapi/bootstrap-agent.json
@@ -229,6 +229,51 @@
           }
         ]
       },
+      "BfdMode": {
+        "description": "BFD connection mode.",
+        "type": "string",
+        "enum": [
+          "single_hop",
+          "multi_hop"
+        ]
+      },
+      "BfdPeerConfig": {
+        "type": "object",
+        "properties": {
+          "detection_threshold": {
+            "type": "integer",
+            "format": "uint8",
+            "minimum": 0
+          },
+          "local": {
+            "nullable": true,
+            "type": "string",
+            "format": "ip"
+          },
+          "mode": {
+            "$ref": "#/components/schemas/BfdMode"
+          },
+          "remote": {
+            "type": "string",
+            "format": "ip"
+          },
+          "required_rx": {
+            "type": "integer",
+            "format": "uint64",
+            "minimum": 0
+          },
+          "switch": {
+            "$ref": "#/components/schemas/SwitchLocation"
+          }
+        },
+        "required": [
+          "detection_threshold",
+          "mode",
+          "remote",
+          "required_rx",
+          "switch"
+        ]
+      },
       "BgpConfig": {
         "type": "object",
         "properties": {
@@ -691,6 +736,14 @@
         "description": "Initial network configuration",
         "type": "object",
         "properties": {
+          "bfd": {
+            "description": "BFD configuration for connecting the rack to external networks",
+            "default": [],
+            "type": "array",
+            "items": {
+              "$ref": "#/components/schemas/BfdPeerConfig"
+            }
+          },
           "bgp": {
             "description": "BGP configurations for connecting the rack to external networks",
             "type": "array",
--- a/sled-agent/src/bootstrap/early_networking.rs
+++ b/sled-agent/src/bootstrap/early_networking.rs
@@ -16,12 +16,13 @@ use internal_dns::resolver::{ResolveError, Resolver as DnsResolver};
 use internal_dns::ServiceName;
 use ipnetwork::Ipv6Network;
 use mg_admin_client::types::{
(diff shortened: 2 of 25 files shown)


sled-agent/src/bootstrap/early_networking.rs
```diff
@@ -804,6 +829,7 @@ impl RackNetworkConfigV0 {
                 .map(|uplink| PortConfigV1::from(uplink))
                 .collect(),
             bgp: vec![],
+            bfd: vec![],
```
andrewjstone, line 832: Is this actually good enough? Presumably we are already on `v1`. Will we serialize an existing `v1` into an empty array if the field isn't present?

andrewjstone: This comment really should have been on the `v1` type addition. But the question still stands. Do we need a v2?

rcgoodfellow: I've put a `#[serde(default)]` on the new field, so i'm trying to avoid a v2 if possible - but i'm not sure which way it's going to go yet

andrewjstone: Got it. Thanks!

rcgoodfellow:
## Testing Notes

I've been testing this in `a4x2`. I set up the enterprise topology. Once the topology is up I configure BFD via the omicron API and verify that the settings make it into the bootstore.

```
{
  "generation": 5,
  "schema_version": 1,
  "body": {
    "ntp_servers": [
      "time.cloudflare.com"
    ],
    "rack_network_config": {
      "rack_subnet": "fd00:1122:3344:100::/56",
      "infra_ip_first": "198.51.101.1",
      "infra_ip_last": "198.51.101.15",
      "ports": [
        {
          "routes": [
            {
              "destination": "0.0.0.0/0",
              "nexthop": "198.51.101.5"
            }
          ],
          "addresses": [
            "198.51.101.6/30"
          ],
          "switch": "switch1",
          "port": "qsfp0",
          "uplink_port_speed": "speed100_g",
          "uplink_port_fec": "none",
          "bgp_peers": [],
          "autoneg": false
        },
        {
          "routes": [
            {
              "destination": "0.0.0.0/0",
              "nexthop": "198.51.101.1"
            }
          ],
          "addresses": [
            "198.51.101.2/30"
          ],
          "switch": "switch0",
          "port": "qsfp0",
          "uplink_port_speed": "speed100_g",
          "uplink_port_fec": "none",
          "bgp_peers": [],
          "autoneg": false
        },
        {
          "routes": [
            {
              "destination": "0.0.0.0/0",
              "nexthop": "198.51.101.13"
            }
          ],
          "addresses": [
            "198.51.101.14/30"
          ],
          "switch": "switch1",
          "port": "qsfp1",
          "uplink_port_speed": "speed100_g",
          "uplink_port_fec": "none",
          "bgp_peers": [],
          "autoneg": false
        },
        {
          "routes": [
            {
              "destination": "0.0.0.0/0",
              "nexthop": "198.51.101.9"
            }
          ],
          "addresses": [
            "198.51.101.10/30"
          ],
          "switch": "switch0",
          "port": "qsfp1",
          "uplink_port_speed": "speed100_g",
          "uplink_port_fec": "none",
          "bgp_peers": [],
          "autoneg": false
        }
      ],
      "bgp": [],
      "bfd": [
        {
          "local": "0.0.0.0",
          "remote": "198.51.101.1",
          "detection_threshold": 3,
          "required_rx": 1000000,
          "mode": "single_hop",
          "switch": "switch0"
        },
        {
          "local": "0.0.0.0",
          "remote": "198.51.101.5",
          "detection_threshold": 3,
          "required_rx": 1000000,
          "mode": "single_hop",
          "switch": "switch1"
        }
      ]
    }
  }
}
```

I then inspect what the BFD state on one of the switches looks like.

```
root@oxz_switch:~# mgadm bfd get-peers
Peer          Listen   Required Rx  Detection Threshold  Mode       Status
198.51.101.1  0.0.0.0  1000000      3                    SingleHop  Init
```

Then I

1. Disable the nexus smf in each of the nexus zones. This is so the nexus BFD RPW does not accidentally help us out by detecting an mgd instance with an out of whack BFD configuration.
2. Restart one of the scrimlet sled-agents.
3. After the switch zone goes down and comes back up, i re-inspect the BFD state to make sure we are back where we started.

That whole process is working as expected.

openapi/bootstrap-agent.json
```diff
@@ -229,6 +229,51 @@
           }
         ]
       },
+      "BfdMode": {
```
rcgoodfellow, line 232: These API additions fall out from the addition of BGP configuration to `RackNetworkConfigV1`

andrewjstone (approved):
LGTM!

PR #5324, opened 2024-03-25
Run sled-agent in its own project

### LINKED ISSUE #5321: sled-agent should not be part of the "system" contract
#5318 and #5319 caused sled-agent to use up all of the available contracts in its project:

```
BRM44220001 # ls -1 /proc/`pgrep sled-agent`/contracts | wc -l
    9968
BRM44220001 # prctl -P `pgrep sled-agent` | egrep contract
project.max-contracts privileged 10000 - deny -
```

When this occurred, system services such as `cron` also began to fail, and it was not possible to connect to the server via `ssh`. Both of these were also failing to fork:

```
BRM44220001 # tail /var/cron/log
! cannot fork Mon Dec 29 06:14:47 1986
! rescheduling a cron job Mon Dec 29 06:14:47 1986
```

This is because `sled-agent` is in the `system` project, along with services such as cron and ssh.

```
BRM44220001 # ps -ef -o pid,project,args | egrep 'cron|sled-agent|ssh'
  686   system /opt/oxide/sled-agent/sled-agent run /opt/oxide/sled-agent/pkg/config.toml
  685   system ctrun -l child -o noorphan,regent /opt/oxide/sled-agent/sled-agent run /opt/oxi
  607   system /usr/sbin/cron
  637   system /usr/sbin/sshd
...
```

We should place `sled-agent` in its own project so that it is not sharing resource limits like this.

citrus-it:
Fixes #5321; also requires https://github.com/oxidecomputer/helios/pull/146


rmustacc (commented):
Where are we expecting to put the project's constraints? That is, I think some of this is to make sure that the project can't use up all the contracts. Is that inherent in the project's defaults somehow?

PR #5316, opened 2024-03-23
stop building rot-prod-rel/rot-staging-dev repos

iliana:

PR #5325, opened 2024-03-25
stop splitting the TUF repo on the 1 GiB boundary

iliana:
Follow-up to #5316, split to prevent breaking various scripts until v7 is out the door.

Per chat and https://github.com/oxidecomputer/omicron/pull/5316#issuecomment-2017325977, Buildomat has supported >1 GiB uploads for a while, but we've seen [issues with downloading files that large](https://github.com/oxidecomputer/buildomat/issues/36). There's now range request support, so retrying failed downloads is no longer catastrophic.

However a lot of scripts are out there that assume we must download a parta/partb. I don't want to prevent people from deploying quickly on dogfood while we're in the run up to a release, unless whoever's on dogfood updates wants to merge this and fix the script.

### DIFF
--- a/.github/buildomat/jobs/tuf-repo.sh
+++ b/.github/buildomat/jobs/tuf-repo.sh
@@ -5,7 +5,7 @@
 #: target = "helios-2.0"
 #: output_rules = [
 #:	"=/work/manifest*.toml",
-#:	"=/work/repo-*.zip.part*",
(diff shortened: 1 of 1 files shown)


labbott (approved):
It sounds like we've been given the green light to merge stuff for not R7. I support merging this soonish.

PR #5330, opened 2024-03-26
point maghemite to smf refresh branch

rcgoodfellow:
@jgallagher I created this branch from your branch to point at my updated maghemite branch with the smf refresh machinery added. I think the artifacts from this branch are probably worth a run on `madrid`.


jgallagher:
Great - thanks. I'm going to un-draft this and merge it so the artifacts can build.

PR #5331, opened 2024-03-26
net-rpw: ignore v6 loopback addrs not created by nexus

rcgoodfellow:
Maghemite creates randomized IPv6 tunnel addresses for tunnel routing endpoints on the underlay network. The network RPW detects these loopback addresses created by maghemite, does not see them in the nexus-tracked state, and determines that they must be destroyed. These addresses are required for egress traffic leaving the rack (see RFD 404 for more details).

This patch has the network RPW ignore loopback addresses that it did not create by filtering on the dpd tag it uses to create loopback addresses.


FelixMcFelix (approved):
Thanks for the debugging efforts on this one Ry, good catch!

Issue #5153, opened 2024-02-27
Add notion of physical disk "policy" and "state" to physical disks in the control plane


smklein:
This will be fixed by https://github.com/oxidecomputer/omicron/pull/3526, but is first blocked by:

 #5151 and #5150

At a first pass, this issue should track the ability to have "in-service" and "expunged" physical disks.

PR #5335, opened 2024-03-26
[nexus] Add Physical Disk Policy, State

smklein:
- Adds physical disk state, policy as defined in RFD 457
- Does not expose any way to modify this information
- In the future, these values will toggled by both sled expungement, and explicitly through a physical disk API

Fixes https://github.com/oxidecomputer/omicron/issues/5153


smklein:
This is a replacement for https://github.com/oxidecomputer/omicron/pull/3526 , which simplifies things a little bit:

- It does not expose an API for disabling physical disks individually, yet. Expunging physical disks associated with a whole sled is an easier prospect, so we'll tackle that first.
  - To safely expunge individual physical disks, we'll need to use the APIs introduced in https://github.com/oxidecomputer/omicron/pull/5172 , use https://github.com/oxidecomputer/omicron/pull/5050 , and otherwise ensure that all "sled-local disk usage" gets revoked for the disk being expunged.
  - ... But to safely expunge disks associated with a sled, we don't actually *need* to talk to the sled, which makes things easier.
- https://github.com/oxidecomputer/omicron/pull/3526 also relied on "deriving a control plane ID from vendor/serial/model" -- this is no longer the case, the UUID for physical disks is now controlled by RSS / Nexus.

- [ ] Before marking this as "Ready-for-review", I also need to update the region allocation query to check for these new columns.

schema/crdb/dbinit.sql
```diff
@@ -331,6 +331,40 @@ CREATE TYPE IF NOT EXISTS omicron.public.physical_disk_kind AS ENUM (
   'u2'
 );
 
+-- The disposition for a particular physical disk.
+-- This is update by the operator, either explicitly through an operator API,
```
jmpesp, line 335: s/update/updated/

jmpesp (approved):
:rocket:

schema/crdb/physical-disk-state-and-policy/up03.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE omicron.public.physical_disk
+    ADD COLUMN IF NOT EXISTS disk_policy omicron.public.physical_disk_policy
+        NOT NULL DEFAUlT 'in_service',
```
jmpesp, line 3: nit: lower case l here

schema/crdb/dbinit.sql
smklein: Done

smklein:
As an FYI, I added https://github.com/oxidecomputer/omicron/pull/5335/commits/f3e85c511cf7ce584230494aebf955b3fe231f53 , which addresses the review comments, but also:

- Checks for state / policy during region allocation
- Adds a test that only the "active / in-service" disks are used for region allocation

PR #5339, opened 2024-03-27
WIP: Allow control of tests run in VMs on the host

andrewjstone:
This is a reworking of `falcon_runner.rs` that splits the work between a `FalconTest` harness that allows direct manipulation of the runner so that the full falcon API is available, while also wrapping the bash nonsense to run a test and get the result back.

The harness is minimal at this point, and purposefully so, to allow for complete flexibility. This allows the user to launch multiple VMs, run arbitrary tests, mount arbitrary data directories into VMs, etc... We can provide more batteries included, perhaps with escape hatches, and we can also simplify further with attribute macros for the batteries included parts.

The example test can be run in the following manner, notably without a nextest target runner. `pfexec` may cause problems with permissions when forced to rebuild however.

```
 pfexec cargo nextest run -p omicron-test-utils run_launch_in_vm --run-ignored all --nocapture
```

Notice that we run the "host" test from nextest here, and that test spins up a VM and runs a test (`launch`) inside it.


test-utils/src/dev/falcon.rs
```diff
@@ -4,10 +4,62 @@
...
+        let mut test = super::FalconTest::new(runner_name, &log).unwrap();
+        eprintln!("args = {:?}", args);
+        // Create a falcon node (VM)
+        let vm = test.runner().node(node_name, "helios-2.0", 1, gb(2));
+
+        // Create a place(cargo-bay) to store data to be mounted into the VM
+        let cargo_bay = tempdir().unwrap();
+        info!(log, "Set cargo-bay to {} on host machine", cargo_bay.path());
+        let test_name = "launch";
+        let source_test_path = &args[0];
```
andrewjstone, line 31: This is a hack to easily find the source binary. It works only if the test to run in the VM is part of the same test binary generated by cargo.

test-utils/src/dev/falcon.rs
```diff
@@ -4,10 +4,62 @@
...
+        let test_path = cargo_bay.path().to_path_buf().join(test_file_name);
+
+        // Copy the test binary into the cargo-bay
+        info!(log, "Copying {source_test_path} to {test_path}");
+        std::fs::copy(camino::Utf8Path::new(source_test_path), &test_path)
+            .expect("failed to copy source test binary into cargo-bay");
+
+        // Mount the cargo-bay into the VM
+        test.runner()
+            .mount(cargo_bay.path(), "/opt/cargo-bay", vm)
```
andrewjstone, line 44: "/opt/cargo-bay", and the host path are both arbitrary, and can be anything. As an example, if a test wanted to launch multiple VMs they may all use the same VM mount path, but have different local directories. This is part of the reason for allowing full access to the falcon API.

andrewjstone:
I don't think we're ever going to merge this.

PR #5340, opened 2024-03-28
Oximeter producer registration should be more lax about success

bnaecker:
As part of https://github.com/oxidecomputer/omicron/issues/5284, we'd like to change the Nexus API used to register as a metric producer. It currently returns a 204 and no data, and we'd like it to return a 201 and a lease duration. However, if we change the server API first, Progenitor-based clients will see an unexpected response, which the `oximeter-producer` registration method currently returns as an error.

This commit relaxes the registration method, so that it still succeeds if it finds an unexpected-but-successful response from Nexus. This will let clients handle the desired changes on the server even before they truly recognize the new response as valid.

### DIFF
--- a/oximeter/producer/src/lib.rs
+++ b/oximeter/producer/src/lib.rs
@@ -251,8 +251,16 @@ pub async fn register(
 ) -> Result<(), Error> {
     let client =
         nexus_client::Client::new(&format!("http://{}", address), log.clone());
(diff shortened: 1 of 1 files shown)


bnaecker:
@ahl Gentle ping on this, should be pretty easy :)

jgallagher (approved):

PR #5341, opened 2024-03-28
[nexus] move UuidRng code out to a common crate, make CollectionBuilder use it

sunshowers:
In #5270, we need determinism not just from blueprints but also collections. So
move the UuidRng into a common place.

As part of that, I also decided to make it its own crate and write some
documentation about it, making it more generic along the way. I think this
should be a pretty clean representation of what this is trying to do.


typed-rng/src/lib.rs
```diff
@@ -0,0 +1,234 @@
...
+/// random values. So we expect that callers will also provide bytes of their
+/// own, specific to the callsite, and gently guide them towards doing the
+/// right thing.
+pub fn from_seed<R, H, H2>(seed: H, extra: H2) -> R
+where
+    R: SeedableRng,
+    H: Hash,
+    H2: Hash,
+{
+    // XXX: is Hash really the right thing to use here? That's what
```
andrewjstone, line 68: The guarantees around stability seem nice in stable-hash, especially if we want to persist these values. However, the backwards compatibility seems dangerous in that it ignores default values. Probably best to stick with this for now.

typed-rng/src/lib.rs
```diff
@@ -0,0 +1,234 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! Typed RNGs with support for trees seeds.
```
andrewjstone, line 5:
```suggestion
//! Typed RNGs with support for tree seeds.
```

andrewjstone (approved):
Looks good!

typed-rng/src/lib.rs
```diff
@@ -0,0 +1,234 @@
...
+}
+
+/// An RNG that can be used to generate values of a single type.
+///
+/// This is a convenience wrapper around a random number generator that
+/// generates values of a particular type. It works against any type that
+/// implements [`Generatable`], and any RNG that implements [`RngCore`].
+pub struct TypedRng<T, R> {
+    rng: R,
+    _marker: PhantomData<T>,
```
jgallagher, line 96: Tiny nit - should this be `PhantomData<fn() -> T>` or similar to avoid `TypedRng` losing `Send` etc if `T` isn't `Send`?

jgallagher (approved):

typed-rng/src/lib.rs
sunshowers: Fixed, and expanded on this comment. Thanks!

typed-rng/src/lib.rs
sunshowers: TIL, thanks! https://doc.rust-lang.org/nomicon/phantom-data.html#table-of-phantomdata-patterns

PR #5342, opened 2024-03-28
vnc proxy

lifning:
plumbing propolis' VNC server through nexus with a websocket endpoint.

it's can be considerably, unusably slow when accessed remotely, because current propolis code can only send uncompressed full-screen 32-bit color updates, so i'm also working on optimizing that in https://github.com/oxidecomputer/propolis/pull/673


david-crespo:
oooooooooh

PR #5344, opened 2024-03-28
[Reconfigurator] Introduce `PlanningInput` that includes external networking information from CRDB

jgallagher:
The primary change in this PR is that the blueprint planner now wants a `PlanningInput` (which contains a `Policy`) instead of a `Policy`. The bulk of the diff is adding new `DataStore` methods (and tests for them) that fetch all currently-allocated external IPs and NICs for services.

Some incidental changes that came along for the ride that I hope are not controversial, but could be backed out if they are:

* `nexus_db_model::NetworkInterface::slot` is now a `SqlU8` instead of an `i16`. I didn't have to change the queries here, so I think they're still converting this to an `i16`, which is probably okay? I could make a pass on them if needed though.
* I added an `omicron_uuid_kinds::ServiceKind` and started using it in this PR. I did not attempt to make a pass through all service UUIDs to start using this; I think this can be done incrementally?

Other notes:

* I'm not sure about the name `PlanningInput`. It feels vague; isn't every argument to the planner a kind of "planning input"? But I'm not sure what else to call "`Policy` plus extra CRDB state".
* This does not change execution at all. It's possible when I get to that there will need to be some changes here, but I think this is probably close enough that it can be reviewed, and any changes will be small and can be rolled into the execution work.


andrewjstone (approved):
Looks good.

nexus/src/app/deployment.rs
```diff
@@ -151,6 +156,47 @@ impl super::Nexus {
             NEXUS_REDUNDANCY,
         )?;
 
+        let service_external_ips = datastore
+            .service_ip_all_list_batched(opctx)
+            .await?
+            .into_iter()
+            .filter_map(|external_ip| {
+                if !external_ip.is_service {
```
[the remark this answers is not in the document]
jgallagher: It would definitely be a major bug in the datastore lookup, yes. We could omit this check entirely. The query for nics just below this doesn't have an equivalent check because the type we get back from the query doesn't even give us an opportunity to check it; arguably the datastore could also use more refined types for external IPs, too? I'm ambivalent about that, and almost left this check out in the first place.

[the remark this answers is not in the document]
jgallagher: I left this in because our current `ExternalIp` type doesn't really go for "make illegal states unrepresentable"; e.g., if `is_service` is true, then `parent_id` should always be `Some(_)`, but I feel like since that illegal state _is_ representable, it's safer to check here before converting to a type that does not allow that illegal state.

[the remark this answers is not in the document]
andrewjstone: Leave it in, make it an error?

nexus/types/src/deployment.rs
```diff
@@ -97,6 +101,47 @@ impl SledResources {
...
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct PlanningInput {
+    /// fleet-wide policy
+    pub policy: Policy,
+
+    /// external IPs allocated to services
+    pub service_external_ips: BTreeMap<TypedUuid<ServiceKind>, ExternalIp>,
+
+    /// vNICs allocated to services
+    pub service_nics: BTreeMap<TypedUuid<ServiceKind>, ServiceNetworkInterface>,
```
sunshowers, line 119: `ServiceKind` is a little generic -- is there a more specific term for it? I thought these were zone IDs (i.e. `OmicronZoneConfig::id`), is that actually the case?

sunshowers (approved):
Looks great, thanks!

nexus/types/src/deployment.rs
jgallagher: They are zone IDs, yeah. `OmicronZoneKind` maybe?

dev-tools/reconfigurator-cli/src/main.rs
```diff
@@ -41,6 +46,14 @@ struct ReconfiguratorSim {
     /// blueprints created by the user
     blueprints: IndexMap<Uuid, Blueprint>,
 
+    /// external IPs allocated to services
+    ///
+    /// In the real system, external IPs have IDs, but those IDs only live in
+    /// CRDB - they're not part of the zone config sent from Reconfigurator to
+    /// sled-agent. This mimics the minimal bit of the CRDB `external_ip` table
+    /// we need.
+    external_ips: IndexMap<IpAddr, Uuid>,
```
davepacheco, line 55: I'm trying to figure out the behavior of this -- like when and how it gets updated in `reconfigurator-cli`.  I see that we populate it when running `blueprint-plan` with a new Uuid if we haven't used this IP address before and we re-use the same id if we've seen this same IP address before.  Why is that correct?  Are we thinking of it like the parent blueprint has just been executed and we're starting from that state?  (I'm not seeing a problem here or anything -- just not quite getting it yet).

Will we need to have `omdb db reconfigurator-save` save this too?

nexus/db-queries/src/db/datastore/external_ip.rs
```diff
@@ -430,6 +433,49 @@ impl DataStore {
         self.allocate_external_ip(opctx, data).await
     }
 
+    /// List one page of all external IPs allocated to internal services
+    pub async fn service_ip_all_list(
```
davepacheco, line 437: I read this as a reversal of the usual convention of `service_ip_list_all` which would mean "list all the service IPs" but I guess it's really "list _a page_ of "all" service IPs"?  What's the "all" here?

nexus/types/src/deployment.rs
```diff
@@ -97,6 +101,47 @@ impl SledResources {
...
 }
 
+/// Policy and database inputs to the Reconfigurator planner
+///
+/// The primary inputs to the planner are the parent (either a parent blueprint
+/// or an inventory collection) and this structure. This type holds the
+/// fleet-wide policy as well as any additional information fetched from CRDB
+/// that the planner needs to make decisions.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct PlanningInput {
```
davepacheco, line 111: tl;dr: Don't feel like you have to do this now, but I wonder if we should:

- move some of the `Policy` stuff into `PlanningInput` (the sled state, zpools, and subnet)
- move the internal and external DNS generations that are accepted by all the same functions into this struct
- create some accessors/iterators around PlanningInput (e.g., methods to iterate the sleds) so that we don't have to make tons of tiny changes when we move stuff like this around inside this struct
- create a builder interface so that callers don't have to assemble all this by hand

I think this might be much more comprehensible and also eliminate quite a lot of code today that assembles these structs and pokes through their innards.  I'm thinking of the various bits in this PR where you had to change just a few characters to accommodate this extra level above `Policy`.  It might also reduce a lot of the small conflicts we have as arguments/fields are added here or there to these functions and structures.

---

How I got here: it seems like `PlanningInput` decomposes into:

- "policy" -- configuration and choices made by the operator
- "system state" -- miscellaneous stuff we have to fetch from the database for planning purposes

But right now `Policy` has `sleds` which contains sled policy (fine) but also sled state, zpools, and subnet, which by the definitions above are definitely "system state" and not "policy".  It seems like this second group should move into `PlanningInput`.  But then you have two maps of sleds, one for policy stuff and one for state.  Callers have to deal with one being present while the other's absent, etc.  If we put this behind an abstraction that ensures that these are consistent (e.g., constructor/builder + iterator that combines the information or whatever) then callers don't have to care how it's organized.

I also think we should add the internal and external DNS versions to `PlanningInput`.  Right now these are separate arguments to `BlueprintBuilder::build_initial_from_collection()`, `BlueprintBuilder::new_based_on()`, and `Planner::new_based_on()` but I don' see how they're any different from this other "system state" and it would clean up a lot of call sites to consolidate them into this struct.

There are also so many tests that make up bogus `Generation::new()` for the DNS versions because they don't care.  I didn't want to make these optional originally because I wanted people writing production code to be forced to think about what the generation should be.  But in practice there's basically one production caller and a ton of callers that variously care or don't care about this and it might be easier if it were just optional.

---

What do folks think?

nexus/types/src/deployment.rs
```diff
@@ -97,6 +101,47 @@ impl SledResources {
     }
 }
 
+/// Policy and database inputs to the Reconfigurator planner
```
davepacheco, line 104: (sorry, wasn't sure where to hang this comment)

In #5345 I added a common function for assembling `UnstableReconfiguratorState`.  This is used in omdb plus one of the reconfigurator-cli tests.  It's closely related but not quite the same as what Nexus assembles for planning.  I'm wondering if it would make sense to:

- have the common function assemble a `PlanningInput`
- `impl From<PlanningInput> for UnstableReconfiguratorState` -- maybe the `UnstableReconfiguratorState` is just an object with the contents of `PlanningInput` that impls `Serialize`/`Deserialize`?

Again, not a blocker for this PR.

dev-tools/reconfigurator-cli/src/main.rs
jgallagher: Good questions, and I admit I was kinda punting here (as I haven't actually used reconfigurator-cli in anger yet). For NICs we're in good shape: the network interface info included in `OmicronZoneType` includes the NIC ID, so we don't need to do anything special. But IPs are weird; `OmicronZoneType` includes the external IP for zones that need one, but does _not_ include the database ID of that external IP. I wonder if I'm putting a bandaid in the wrong place here, and instead we should send the external IP ID in the `OmicronZoneType` so we don't need to do anything at all here?

When I put this in, I was thinking of this as "some zones have external IPs; those have IDs that are effectively internal to CRDB, because they're not sent to the sleds as part of the zone config". To mimic that, I stashed them here.

> Will we need to have omdb db reconfigurator-save save this too?

Not sure. It may depend on how we use these in the planner, and how much execution-fakery reconfigurator-cli grows.

nexus/db-queries/src/db/datastore/external_ip.rs
jgallagher: There's already a `service_lookup_external_ips` which lists the IPs for a specific service. The "all" here is supposed to be "external IPs for _all_ services". I renamed this function 2 or 3 times and never landed on something that really felt right - any suggestions?

nexus/types/src/deployment.rs
jgallagher: I'm with you for the vast majority of that, and would be happy to take a pass at it as a followup to this PR (and maybe poke at your suggestion about the relationship between `PlanningInput` and `UnstableReconfiguratorState` while I'm at it). One bit I'm not sure about:

> I also think we should add the internal and external DNS versions to PlanningInput

These are inputs to `BlueprintBuilder::build_initial_from_collection()`, but `PlanningInput` is not (at least on this PR). The extra system state stashed in `PlanningInput` is not relevant to build a blueprint from a collection, and especially as that state grows it seems reasonable (I think?) to avoid having to supply it unnecessarily.

nexus/types/src/deployment.rs
jgallagher: Renamed in [91e3843](https://github.com/oxidecomputer/omicron/pull/5344/commits/91e3843678ea5ebf82c75b148f42ec21d1d23330)

nexus/src/app/deployment.rs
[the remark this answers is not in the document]
jgallagher: Now an `error!` [35b1e16](https://github.com/oxidecomputer/omicron/pull/5344/commits/35b1e16bbcc0e9feacca4242b8d344dbe57169d9)

dev-tools/reconfigurator-cli/src/main.rs
davepacheco: > But IPs are weird; `OmicronZoneType` includes the external IP for zones that need one, but does _not_ include the database ID of that external IP. I wonder if I'm putting a bandaid in the wrong place here, and instead we should send the external IP ID in the `OmicronZoneType` so we don't need to do anything at all here?

I'm not sure.  But if we did that, would that be a better way to solve our other problem of knowing which records to be deleted when a zone gets expunged?  I'm not actually sure it's any better or even if it would work.

I think this is probably fine for now and we may rethink it as we iterate on `reconfigurator-cli` -- its model of "state outside the blueprint" is getting a little creaky in other ways, too.

nexus/db-queries/src/db/datastore/external_ip.rs
davepacheco: I think from scratch I might choose `external_ip_list_all_services` and `external_ip_list_service` (for the one particular service).

Is this the "service" that we now call "zone"?  Maybe it would be `external_ip_list_all_zones` and `external_ip_list_zone`?

Whatever you think.  I also think it's fine as is.

jgallagher: `external_ip_list_all_service{,s}` sounds good to me. I'm less sure about calling it "zone"; external IPs can also be attached to instances, which also have zones. All the networking-related datastore bits try to label things as "service" or "instance".

dev-tools/reconfigurator-cli/src/main.rs
jgallagher: > I'm not sure. But if we did that, would that be a better way to solve our other problem of knowing which records to be deleted when a zone gets expunged? I'm not actually sure it's any better or even if it would work.

No, we'd still have to fetch all the external IPs for this to work. But we wouldn't need this extra state in `reconfigurator-cli` just to describe the in-service zones.

I'll go with "fine for now, iterate later", though. :-)

nexus/db-queries/src/db/datastore/external_ip.rs
davepacheco: > All the networking-related datastore bits try to label things as "service" or "instance".

Definitely outside the scope of this PR but I wonder if we want to revisit this, at least for "service".  I would guess this code uses "service" because that's what we used to call them.  We moved away from that term elsewhere.  RFD 248 defines "service" the way I think the industry usually uses that term, which is in terms of how the _consumer_ views it.  i.e., one "network service" might be a DNS name with many IPs corresponding to a potentially different number of backends.  It's super confusing to call each of those backends a separate "service".  There's also the confusion with SMF services, though SMF services themselves match the more consumer-oriented view (e.g., in having services like "filesystem/local" which means "persistent local filesystems are available", even though that's not necessarily a discrete OS process).

It's doubly unfortunate that broadly in the industry, "instance" means both "a virtual machine" and "one of many [basically fungible] backends for a service".  It'd be really nice if we had a notation or a more specific term we could use.

PR #5348, opened 2024-03-29
Add BlueprintZoneFilter to all_omicron_zones

sunshowers:
I'm not actually sure these are quite correct, but putting these up there for
an early look.

TODO:

- [x] Add some tests, there are a bunch of decisions we made here.
- [x] Are all these decisions correct?
- [x] Also need to do this filtering for direct joins against the `bp_omicron_zone` table -- but that will be easier once the zones are a simple column on #5211.


andrewjstone:
This is closer, I added a single test so far that ensures that filtering around external dns works. A few more tests may be warranted. With https://github.com/oxidecomputer/omicron/pull/5438, @sunshowers added the disposition field and filtered the one join on `bp_omicron_zones` we found. I also *think* the decisions made here are correct, so checked that box above as well.

andrewjstone:
I think this is good to go.

nexus/types/src/deployment.rs
```diff
@@ -400,13 +404,22 @@ impl BlueprintZoneDisposition {
         match self {
             Self::InService => match filter {
                 BlueprintZoneFilter::All => true,
-                BlueprintZoneFilter::SledAgentPut => true,
+                BlueprintZoneFilter::Crucible => true,
```
jgallagher, line 407: I don't love this name. This:

```rust
all_omicron_zones(BlueprintZoneFilter::Crucible)
```

looks like it's filtering to the zones that have zone _type_ Crucible, to me. I'm not even 100% sure what we're saying here, exactly. I would understand `SledFilter::ShouldHaveCrucibleZones`, but what does `BlueprintZoneFilter::Crucible` mean when applied to a non-crucible zone?

(I have similar concerns about `::External` and `::SledAgentPut`, but this seemed like the easiest one on which I could state the concern.)

nexus/types/src/deployment.rs
```diff
@@ -1593,3 +1615,45 @@ mod table_display {
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use proptest::prop_assert;
+    use test_strategy::proptest;
+
+    // This test is to ensure that the underlying filter semantics don't change
+    // and break higher level callers. It's fine to do that in some cases, but
+    // at least you'll have to think about it more.
```
jgallagher, line 1627: I don't object to this test (and love proptests in general), but I'm not entirely sure what it's adding - this looks like it's reimplementing `BlueprintZoneDisposition::matches()` with `prop_assert!` instead of bools, right?

jgallagher (commented):
~~Is this the PR where we're also supposed to fix up `DataStore::vpc_resolve_to_sleds()` (which implicitly uses `all_omicron_zones`, sorta, by directly querying the `bp_omicron_zone` table, but doesn't currently consider the zone disposition)?~~ Nevermind! The `vpc_resolve_to_sleds()` fix landed in #5238

nexus/types/src/deployment.rs
```diff
@@ -147,9 +147,12 @@ impl Blueprint {
     /// along with the associated sled id.
     pub fn all_omicron_zones(
         &self,
+        filter: BlueprintZoneFilter,
     ) -> impl Iterator<Item = (Uuid, &OmicronZoneConfig)> {
-        self.blueprint_zones.iter().flat_map(|(sled_id, z)| {
-            z.zones.iter().map(|z| (*sled_id, &z.config))
+        self.blueprint_zones.iter().flat_map(move |(sled_id, z)| {
```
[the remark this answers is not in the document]
sunshowers: Ahh I have a wip change that needs all_blueprint_zones.

nexus/types/src/deployment.rs
andrewjstone: >  I would understand SledFilter::ShouldHaveCrucibleZones, but what does BlueprintZoneFilter::Crucible mean when applied to a non-crucible zone?

I feel like this comment somewhat questions the overall premise of the `BlueprintZoneFilter`. We could indeed have `SledFilter` filter out any unnecessary zones by looking at zone disposition, but that would likely be somewhat incoherent as the sled is now filtering zones based on disposition disposition. I totally get what you are saying here though, and I think you have a point. You are talking about filtering on behaviors or what should happen as a result of a filter matched with a disposition, rather than individual zones, since a filter for an individual zone doesn't do anything for other zones. However, crucible zones are kinda special and they do go on all disks, unless the disposition is expunged. I wonder if instead of specifying crucible, we could cover other zones with this similar property, such as `SledAgentPut`, and even `VpcFirewall`. Those all get deployed unless expunged. The downside of that is lack of specificity in the caller.

I'm kinda ranting here, but I wonder if there's a larger change to be made...

nexus/types/src/deployment.rs
sunshowers: Yeah I agree with John that a higher-level test (many of which may require #5493) would be more useful.

nexus/reconfigurator/execution/src/dns.rs
```diff
@@ -706,7 +706,7 @@ mod test {
         // To start, we need a mapping from underlay IP to the corresponding
         // Omicron zone.
         let mut omicron_zones_by_ip: BTreeMap<_, _> = blueprint
-            .all_omicron_zones()
+            .all_omicron_zones(BlueprintZoneFilter::All)
             .filter(|(_, zone)| zone.id != out_of_service_id)
```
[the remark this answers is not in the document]
andrewjstone: Yes! I'm making a small change now.

nexus/types/src/deployment.rs
sunshowers: So... the main reason we introduced `BlueprintZoneFilter`, versus for example passing in a callback that returns a boolean, is to ensure that we have an accounting about every place that is making decisions about which zones it's considering.

In this case, `BlueprintZoneFilter::Crucible` is used to identify the set of zones for which we expect active dataset records to exist.

But you bring a really important point up: that `BlueprintZoneFilter::Crucible` can return non-Crucible zones in a manner that is somewhat surprising. So... what if `BlueprintZoneFilter::Crucible` also filtered based on zone kind? The first thing `ensure_crucible_dataset_records_exist` does is drop consideration for non-Crucible zones, so it would fit right in.

Also, I think it may be worth renaming this to `CrucibleDatasetRecords` or similar.

nexus/reconfigurator/execution/src/dns.rs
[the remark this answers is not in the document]
andrewjstone: Fixed in [965c25f](https://github.com/oxidecomputer/omicron/pull/5348/commits/965c25f2f1c715697c77955b76192e6e97e40fba)

nexus/types/src/deployment.rs
andrewjstone: Reverted.

nexus/types/src/deployment.rs
andrewjstone: I think it's worth changing the name as well. I will go ahead and do that. As for `SledAgentPut`, I think that one is actually extremely useful. It provided a solid mechanism for limiting which zones actually get included in the put message to sled agent. 

Right now, `External` really only maps to external DNS, unless I missed something. Maybe it's actually worth making that one *more* specific?

andrewjstone: I retract my statement on `External`. I like how it covers multiple things. How about renaming it to `ExternalResources` though?

andrewjstone: I ended up changing the name to `ExternallyReachable` in [514ff8c](https://github.com/oxidecomputer/omicron/pull/5348/commits/514ff8c9bcf6acc3a3e4d0f858fe8948f830adcd)

I'm happy to change it again if you all disagree.

andrewjstone: I also changed `Crucible` to `CrucibleDatasets`

dev-tools/omdb/src/bin/omdb/db.rs
```diff
@@ -686,15 +688,21 @@ async fn lookup_instance(
         .with_context(|| format!("loading instance {instance_id}"))
 }
 
+#[derive(Clone, Debug)]
+struct ServiceInfo {
+    service_kind: ServiceKind,
+    disposition: BlueprintZoneDisposition,
+}
+
 /// Helper function to look up the kind of the service with the given ID.
```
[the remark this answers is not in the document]
andrewjstone: Fixed in [514ff8c](https://github.com/oxidecomputer/omicron/pull/5348/commits/514ff8c9bcf6acc3a3e4d0f858fe8948f830adcd)

nexus/types/src/deployment.rs
jgallagher: > But you bring a really important point up: that BlueprintZoneFilter::Crucible can return non-Crucible zones in a manner that is somewhat surprising. So... what if BlueprintZoneFilter::Crucible also filtered based on zone kind? The first thing ensure_crucible_dataset_records_exist does is to not consider non-Crucible zones, so it would fit right in.

Yeah, I like this, although I think I'm still getting hung up on the fact that this primarily looks at the zone disposition. We want this call:

```rust
.all_omicron_zones(BlueprintZoneFilter::SomethingSomething)
```

to return (a) all crucible zones that (b) are still running in some capacity (either in service or quiesced). `::Crucible` sounds like it's just returning _all_ the crucible zones; `CrucibleDatasets` sounds like either "zones that provide datasets" or "zones that need datasets", but neither of those feel like they capture the "still running in some capacity" part of it.

I'm not sure where that leave me, other than vaguely unsatisfied with the names (with my apologies!). I feel like I want a word for "in service or quiesced". `Running` doesn't seem right, but if it were, I'd propose something like:

* `::RunningCrucible` (only non-expunged zones with type `::Crucible`)
* `::Running` (all non-expunged zones, which clients can additionally filter by type if they want to)

> Ooh, or we could change the name of External to ExternallyReachable.

I like this name a lot more; I feel like it captures the intent of what we want (i.e., we filter by this before applying any changes related to making zones reachable from outside the rack).

andrewjstone: I made the [changes]([0607575](https://github.com/oxidecomputer/omicron/pull/5348/commits/0607575165962707eec9a4516878e586da0358d9)) that @jgallagher, @sunshowers, and I discussed in a video chat.

nexus/types/src/deployment.rs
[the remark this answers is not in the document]
jgallagher: Ok, I'll go with door number 2 then - can we rewrite the implementation of this to use `all_blueprint_zones()`?

nexus/types/src/deployment.rs
```diff
@@ -461,14 +470,17 @@ pub enum BlueprintZoneFilter {
     /// All zones.
     All,
 
-    /// Filter by zones that should be in internal DNS.
-    InternalDns,
+    // Zones that are desired to be in the RUNNING state
```
jgallagher, line 473:
```suggestion
    /// Zones that are desired to be in the RUNNING state
```

andrewjstone: Done in [81df651](https://github.com/oxidecomputer/omicron/pull/5348/commits/81df6517be952e5baf1500135f34cf7e6d19ebcc)

nexus/types/src/deployment.rs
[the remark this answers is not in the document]
andrewjstone: Good idea! Done in [81df651](https://github.com/oxidecomputer/omicron/pull/5348/commits/81df6517be952e5baf1500135f34cf7e6d19ebcc)

PR #5351, opened 2024-03-29
[DO NOT MERGE] Repro bb8 issues

smklein:
As a part of https://github.com/oxidecomputer/omicron/pull/5172 , I hit a bug where several connections -- even ones from background jobs, that SHOULD NOT have been impacted -- returned "Timed out in bb8" errors.

This PR attempts to act as a reproduction case for that class of issues.

I'm also working on adding a reproduction case to https://github.com/oxidecomputer/async-bb8-diesel , but I haven't managed that quite yet.

### DIFF
--- a/nexus/db-queries/src/db/datastore/rack.rs
+++ b/nexus/db-queries/src/db/datastore/rack.rs
@@ -737,7 +737,35 @@ impl DataStore {
                     info!(log, "Inserted services");
 
                     for physical_disk in physical_disks {
-                        Self::physical_disk_upsert_on_connection(&conn, &opctx, physical_disk)
+                        // XXX This is wrong.
+                        //
+                        // Namely, calling this method on the datastore, instead
+                        // of re-using `conn`, checks out a new connection from
+                        // the connection pool and uses that. This new
+                        // connection is not participating in the transaction,
+                        // and does not benefit from transactional semantics.
+                        //
+                        // So, that's not right. But in:
+                        // https://github.com/oxidecomputer/omicron/pull/5172/commits/40a561e87d05d0938baf8f8023f26c0cbbffc8ff
+                        //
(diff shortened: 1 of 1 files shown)


smklein:
If this reproduces, I think I'm going to take the following tactics:

- [ ] Try to create a smaller reproduction. This is a big beefy transaction that needs a lot of stars aligned to work. Would be nice if we could re-create this with a smaller endpoint, or ideally, without a dropshot endpoint at all.
- [ ] Add a lot more inspection in bb8. What is the status of our transaction pool? How many connections do we *think* are open?
- [ ] Add more inspection on the Cockroach side. Is there any way to see which transactions are still open?

I have theories about how things could be going wrong, but need more data to validate.

- Is this triggered purely by "getting a new connection from within a transaction, which already has a transaction" -> I don't think so. This doesn't reproduce minimally (I tried, in async-bb8-diesel), and I *think* is happening implicitly with all the auth check calls that aren't using the same connection as the higher-level transaction.
- Could there be a deadlock here between transactions? -> This seems *possible* to me? Perhaps the transaction touches rows that are getting poked at by the "independently-checked out connection", and the inability of the transaction to complete hangs both?
- Could local state in Diesel (or bb8) be corrupted? This is possible, but I believe the "Transaction manager" semantics are "per-connection", so this seems **unlikely** to me?

smklein:
Confirmed, this has caused a repro:

https://buildomat.eng.oxide.computer/wg/0/artefact/01HT5GMRZHQ67AF5DV62HZA2ZF/nQYCIDW98iL5Ai4TyBjvrR1jOkHp3xKI4U6oLXnmap0Pz7DC/01HT5GN9A05911XEH4WDAXGNRN/01HT5NBJYC6KH0G7PBA967TSGY/oxide-nexus:default.log?format=x-bunyan

PR #5354, opened 2024-03-29
some initial blueprints could just be empty

davepacheco:
With an eye towards removing `BlueprintBuilder::build_initial_from_collection()`, this adds `BlueprintBuilder::build_initial_empty()` and uses that from several places we were using the other function.

This already eliminates some yak shaving in a few tests and I think that will improve further when fewer things create a collection just so they can construct the blueprint they want.

### DIFF
--- a/nexus/reconfigurator/planning/src/blueprint_builder.rs
+++ b/nexus/reconfigurator/planning/src/blueprint_builder.rs
@@ -91,15 +91,20 @@ pub enum EnsureMultiple {
 
 /// Helper for assembling a blueprint
 ///
-/// There are two basic ways to assemble a new blueprint:
+/// There are a few basic ways to assemble a new blueprint:
 ///
-/// 1. Build one directly from a collection.  Such blueprints have no parent
+/// 1. Build an initial empty one.  Such blueprints have no parent blueprint.
+///    They are not customizable.  Use
+///    [`BlueprintBuilder::build_initial_empty`] for this.  This is generally
+///    only used for testing.
+///
+/// 2. Build one directly from a collection.  Such blueprints have no parent
 ///    blueprint.  They are not customizable.  Use
-///    [`BlueprintBuilder::build_initial_from_collection`] for this.  This would
-///    generally only be used once in the lifetime of a rack, to assemble the
-///    first blueprint.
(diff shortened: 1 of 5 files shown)


nexus/reconfigurator/planning/src/blueprint_builder.rs
```diff
@@ -138,6 +143,48 @@ pub struct BlueprintBuilder<'a> {
...
 
 impl<'a> BlueprintBuilder<'a> {
+    /// Build a Blueprint describing a system with no Omicron zones
+    pub fn build_initial_empty(policy: &Policy, creator: &str) -> Blueprint {
+        Self::build_initial_empty_impl(
+            policy,
+            creator,
+            BlueprintBuilderRng::new(),
+        )
+    }
```
sunshowers, line 153: Can this return a `BlueprintBuilder` which `build()` can then be called on? Not sure how useful that might be.

nexus/types/src/deployment.rs
```diff
@@ -335,6 +335,11 @@ pub struct BlueprintZonesConfig {
 }
 
 impl BlueprintZonesConfig {
+    /// Constructs an empty [`BlueprintZonesConfig`]
+    pub fn empty() -> Self {
+        BlueprintZonesConfig { generation: Generation::new(), zones: vec![] }
+    }
```
sunshowers, line 341: Can this be the `Default` impl for `BlueprintZonesConfig`?

davepacheco: It could.  I usually prefer more explicit, named methods because `default()` doesn't convey semantics (this would not always be an appropriate default value -- really only if you know the corresponding sled has no zones).  But I get it's more idiomatic and I can do that.

nexus/reconfigurator/planning/src/blueprint_builder.rs
davepacheco: It could!  But we don't have any consumers for that yet so I'm not sure that'll make sense.  It will also add an extra step to all the callers but that's okay.

jgallagher (approved):
LGTM - sorry for all the conflicts on main now :(

davepacheco:
I think this is obviated by other ongoing work.

PR #5356, opened 2024-03-30
[oxlog] sort log files before printing them out

sunshowers:
Found this while attempting to debug core files during today's mupdate -- I was
confused about logs getting skipped for a bit. We should sort log files so
they're printed in ascending order, disregarding the directories they're from.

Sort by filename only, because log files can go in. Sorting by file name
sorts by (service name, timestamp), which is the order we want 99% of the
time.

There is one wrinkle here, which is that for log files that are written
before NTP sync, the timestamps are of the form:

```
/pool/ext/f522118c-5dcd-4116-8044-07f0cceec52e/crypt/debug/oxz_switch/oxide-wicketd:default.log.536112897
```

But for log files written after NTP sync, they're of the form:

```
/pool/ext/f522118c-5dcd-4116-8044-07f0cceec52e/crypt/debug/oxz_switch/oxide-wicketd:default.log.1699421382
```

The latter gets sorted before the former, which isn't ideal. But also it
works fine and achieves the goal of keeping related logs together.

(Note that sorting by file name is possible with `sort` but a bit hard. Something like `sort -t/ -k7` would work: the `7` here indicates the depth at which log files are found.)

### DIFF
--- a/dev-tools/oxlog/src/lib.rs
+++ b/dev-tools/oxlog/src/lib.rs
@@ -284,6 +284,15 @@ impl Zones {
                 load_extra_logs(dir, svc_name, &mut output, filter.show_empty);
             }
         }
+
+        // readdir traverses over directories in indeterminate order, so sort
+        // by filename (which is enough to sort by service name and timestamp
+        // in most cases).
+        for svc_logs in output.values_mut() {
+            svc_logs.archived.sort_unstable();
+            svc_logs.extra.sort_unstable();
+        }
+
         output
     }
 }
--- a/dev-tools/oxlog/src/lib.rs
+++ b/dev-tools/oxlog/src/lib.rs
@@ -98,6 +98,10 @@ impl LogFile {
             }
         }
     }
+
+    pub fn file_name_cmp(&self, other: &Self) -> std::cmp::Ordering {
+        self.path.file_name().cmp(&other.path.file_name())
+    }
 }
 
 impl PartialEq for LogFile {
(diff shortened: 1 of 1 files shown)


sunshowers:
Hmm so I'm not sure if this is totally right, because it'll sort log files where NTP hasn't been synced after all the other ones. Maybe that's okay though.

andrewjstone (commented):
I find it somewhat unexpected to sort by filenames and not paths. However, I suppose it makes sense since we typically are using oxlog just to list files in a specific zone, where the path doesn't really matter. I'm totally open to any options here, but let's ask around and get some opinions of others who are using oxlog heavily, like @leftwo and @david-crespo.

smklein (commented):
I'm totally in favor of sorting, but why disregard the path prefix? Why not sort by path *and* filename? (PS - asking this legitimately, rather than pushing for it necessarily)

Seems reasonable to have the sorted list of "earliest to newest" Nexus logs, but it seems weird to interleave those with logs from other services.

(Does this justify a "GROUP BY"-style flag?)

leftwo:
I think the sorting by file name makes sense when we have named the files with a datestamp as
the last field in the file. 

If we add the full path to the search, then logs that end up on different `/pool/ext/` will mess up
the sort order.  

99% of the time I want to see files from a single service, and I want them sorted from oldest to newest.
As long as I can still sort in post, I'll deal with any differences, but defaulting to searching by
* Name of the service (not the full path)
* Then, date on the file (which can be accomplished by use the file name itself).
Would make me happy :)

That's my 2c

smklein:
> If we add the full path to the search, then logs that end up on different `/pool/ext/` will mess up the sort order.
>
> 99% of the time I want to see files from a single service, and I want them sorted from oldest to newest.

Totally forgot about the `/pool/ext` prefix throwing things off, but yeah, I figured that "look within the context of a single service" would be the most common use-case.

sunshowers:
Right, exactly -- all the files go into different directories in `/pool/ext`. (Sorry, should have made this clearer -- in my defense it was Friday night :) )

> I think the sorting by file name makes sense when we have named the files with a datestamp as the last field in the file.
> 
> If we add the full path to the search, then logs that end up on different `/pool/ext/` will mess up the sort order.
> 
> 99% of the time I want to see files from a single service, and I want them sorted from oldest to newest. As long as I can still sort in post, I'll deal with any differences, but defaulting to searching by
> 
> * Name of the service (not the full path)
> * Then, date on the file (which can be accomplished by use the file name itself).
>   Would make me happy :)
> 
> That's my 2c

This is exactly what sorting by filename gets, yeah.

andrewjstone (approved):
Based on discussion, it seems this is preferable for some users. I suggest merging and trying it out!

sunshowers:
Thanks! I've updated the PR summary and I'm going to land this shortly.

PR #5358, opened 2024-03-30
[oxlog] add sigpipe::reset

sunshowers:
We have a number of CLI tools that panic when piped to things like `head`
instead of quietly exiting.  There's a long history about this within the Rust
community (see https://github.com/rust-lang/rust/issues/62569), but the long
and short of it is that SIGPIPE really should be set to its default handler
(`SIG_DFL`, terminate the process) for CLI tools.

Because oxlog doesn't make any network requests, reset the SIGPIPE handler to
`SIG_DFL`.

I looked at also potentially doing this for some of our other CLI tools that
wait on network services. This should be fine to do if and only if whenever we
send data over a socket, the `MSG_NOSIGNAL` flag is set. (This causes an
`EPIPE` error to be returned, but no `SIGPIPE` signal to be generated.)

Rust does set this flag [here]. **However, as of Rust 1.77 this flag is not
set on illumos.** That's a bug and I'll fix it in Rust upstream.

[here]: https://github.com/rust-lang/rust/blob/877d36b1928b5a4f7d193517b48290ecbe404d71/library/std/src/sys_common/net.rs#L32

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6293,6 +6293,7 @@ dependencies = [
  "chrono",
  "clap 4.5.1",
  "omicron-workspace-hack",
+ "sigpipe",
  "uuid 1.7.0",
 ]
 
@@ -8664,6 +8665,15 @@ checksum = "5e1788eed21689f9cf370582dfc467ef36ed9c707f073528ddafa8d83e3b8500"
 dependencies = [
  "digest",
  "rand_core 0.6.4",
+]
+
+[[package]]
+name = "sigpipe"
+version = "0.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5584bfb3e0d348139d8210285e39f6d2f8a1902ac06de343e06357d1d763d8e6"
+dependencies = [
+ "libc",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -378,6 +378,7 @@ sha3 = "0.10.8"
 shell-words = "1.1.0"
 signal-hook = "0.3"
 signal-hook-tokio = { version = "0.3", features = [ "futures-v0_3" ] }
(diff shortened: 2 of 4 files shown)


andrewjstone (approved):
Thanks @sunshowers!

If I understand this correctly, the problem is that when we output to something like head, `oxlog` is the write side of the pipe, and `head` is the read side. `head` only reads a subset of data and then exits, but `oxlog` continues to write which causes an `EPIPE`. This handler instead makes `oxlog` exit cleanly, right?

sunshowers:
> If I understand this correctly, the problem is that when we output to something like head, oxlog is the write side of the pipe, and head is the read side. head only reads a subset of data and then exits, but oxlog continues to write which causes an EPIPE. This handler instead makes oxlog exit cleanly, right?

Yes, that's right. When you fail to write to a pipe, two things happen:

1. The `SIGPIPE` signal is sent. The exception is to not do this if [`MSG_NOSIGNAL` is passed in](https://linux.die.net/man/2/send) -- as noted in the summary, with Rust programs it isn't set at all on illumos. Even on OSes where Rust does set the flag, it's only done for network streams, not local file descriptor writes.
2. The `EPIPE` error message is returned.

If you're a C program, then both 1 and 2 happen -- though 1 happens before 2 does. With SIGPIPE, the default signal disposition is for the program to terminate -- this is why CLI tools written in C exit cleanly.

But by default, Rust programs _ignore_ the `SIGPIPE` handler. This means that 2 happens. And `println!` doesn't return an error -- instead it just panics when that happens. Hence the ugly message.

The whole situation isn't great, but it also is tremendously hard to change at this point. I made a much more minor change in https://github.com/rust-lang/rust/pull/101077 to make it so that the signal mask was inherited rather than reset, and just that change broke a few downstream users.

Issue #5364, opened 2024-03-31
Oximeter producer deletion resulted in many 404 errors with request latencies of 10+ seconds


askfongjojo:
While investigating #5363 and [customer-support#116](https://github.com/oxidecomputer/customer-support/issues/116#issuecomment-2027923524), I saw a number of 500 errors resulted from oximeter producer removal errors, e.g.
```
23:27:59.435Z INFO 20b100d0-84c3-4119-aa9b-0c632b0b6a3a (dropshot_internal): request completed
    error_message_external = Internal Server Error
    error_message_internal = failed to delete producer from collector: Error Response: status: 404 Not Found; headers: {"content-type": "application/json", "x-request-id": "baf56a1b-7d8a-431d-87e2-9e4671d6c5bb", "content-length": "84", "date": "Sat, 30 Mar 2024 23:27:59 GMT"}; value: Error { error_code: None, message: "Not Found", request_id: "baf56a1b-7d8a-431d-87e2-9e4671d6c5bb" }
    file = /home/build/.cargo/git/checkouts/dropshot-a4a923d29dccc492/29ae98d/dropshot/src/server.rs:837
    latency_us = 9951084
    local_addr = [fd00:1122:3344:104::3]:12221
    method = PUT
    remote_addr = [fd00:1122:3344:10b::1]:54030
    req_id = 04a46873-ca82-47d2-aef0-60c726d20c19
    response_code = 500
    uri = /instances/b91a5c67-077e-4e05-a8c5-d85b997de150
```
They might have changed the overall responsiveness of stop-instance requests. It's unclear how they might cause other request timeouts.

askfongjojo:
There are more examples of this same error in catacomb:/staff/core/omicron-5363/oxide-nexus_default.log.1711843200. There are 54 occurrences of them when I was deleting 75 instances with Terraform (which used 10 concurrent threads by default).

askfongjojo:
I've uploaded the oximeter log file that corresponds to the Mar 30 22:30 to 23:30 time-window to catacomb:/staff/core/omicron-5363.

jgallagher:
> This might be caused by #5326: Oximeter now periodically refreshes its collectors and deletes those are gone

It looks like this is what happened, although maybe my analysis of this having to happen inside a window of `unassign_producer` is wrong, because these timestamps are pretty far part. Trimming down to just the relevant log lines:

1. Oximeter starts refreshing producers

```
[...]

PR #5366, opened 2024-04-01
Deleting a nonexistent producer should succeed

bnaecker:


bnaecker:
Partial fix for #5364

smklein (approved):

oximeter/collector/src/agent.rs
```diff
@@ -679,8 +679,10 @@ impl OximeterAgent {
         >,
         id: Uuid,
     ) -> Result<(), Error> {
-        let (_info, task) =
-            tasks.remove(&id).ok_or_else(|| Error::NoSuchProducer(id))?;
```
jgallagher, line 683: I think this was the only use of `Error::NoSuchProducer`, so we could probably drop this variant entirely now?

jgallagher (approved):

oximeter/collector/src/agent.rs
bnaecker: Yeah, good catch.

bnaecker: Done in f4736972c

Issue #5369, opened 2024-04-01
Expunge disks when sleds are expunged


smklein:
This is an "implicit" case of expungement. Physical disks are attached to sleds, so if the sled is removed, the disks should be removed too.

PR #5371, opened 2024-04-01
[nexus] Sled Expungement causes PhysicalDisk expungement

smklein:
Fixes https://github.com/oxidecomputer/omicron/issues/5369

This updates the "expunge sled" endpoint which is part of the internal API, but the same implementation should still apply when the endpoint is exposed to operators.

### DIFF
--- a/nexus/db-queries/src/db/datastore/sled.rs
+++ b/nexus/db-queries/src/db/datastore/sled.rs
@@ -329,6 +329,9 @@ impl DataStore {
     /// sufficient warning to the operator.
     ///
     /// This is idempotent, and it returns the old policy of the sled.
+    ///
+    /// Calling this function also implicitly marks the disks attached to a sled
+    /// as "expunged".
     pub async fn sled_set_policy_to_expunged(
         &self,
         opctx: &OpContext,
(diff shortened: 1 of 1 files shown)


nexus/db-queries/src/db/datastore/sled.rs
```diff
@@ -348,73 +351,127 @@ impl DataStore {
...
+                                ));
+                            }
+                        }
+                        #[cfg(test)]
+                        (ValidateTransition::No, _) => result.found.policy(),
+                    };
+
+                    // When a sled is expunged, the associated disks with that
+                    // sled should also be implicitly set to expunged.
+                    let new_disk_policy = match new_sled_policy {
```
andrewjstone, line 442: Just for my own understanding: This technically doesn't have to be in a transaction, right? It's just that we need someway to perpetually retry to expunge the disks when we expunge the sleds, and this is the most straightforward and robust?

nexus/db-queries/src/db/datastore/sled.rs
```diff
@@ -967,6 +1027,107 @@ mod test {
...
+                    .pool_connection_for_tests()
+                    .await
+                    .expect("No connection"),
+            )
+            .await
+            .expect("Failed to lookup physical disk")
+    }
+
+    #[tokio::test]
+    async fn test_sled_expungement_also_expunges_disks() {
```
andrewjstone, line 1050: Nice, easily readable test!

andrewjstone (approved):

nexus/db-queries/src/db/datastore/sled.rs
smklein: Yeah, it could also easily be a CTE (I think those are the only two options to make the UPDATEs of both tables atomic -- doing two separate UPDATEs back-to-back means that the sled could get expunged without  the disk being expunged, if we crashed halfway through).

andrewjstone: Yeah, that makes sense. I was more thinking that they don't need to be atomic updates at all neccesarrily, since we only soft-delete sleds. We *could* have an RPW that sat there and lazily expunged disks after detecting the sled was expunged. This is all theoretical though, as the implementation that exists looks great to me.

Issue #5372, opened 2024-04-01
expunge sled: planner must not use physical disks being removed


smklein:
This is probably blocked on oxidecomputer/omicron#3526, which I believe is adding physical disk policies described in RFD 457.  Once we have this, the blueprint planner will need to make sure that:

- [x] it doesn't try to put Crucible zones onto zpools whose disks have a policy other than "in-service" (will be resolved by https://github.com/oxidecomputer/omicron/pull/5601)
- [x] it doesn't try to put the dataset for any other kind of Omicron zone (like CockroachDB or Clickhouse or DNS) onto a zpool whose disk has a policy other than "in-service" (will be resolved by https://github.com/oxidecomputer/omicron/pull/5599)
- [x] it should identify any zones that already have persistent datasets on such disks and provision replacements instead (partially resolved by https://github.com/oxidecomputer/omicron/pull/5375 )
- [x] once we add the zone's (transient) root filesystem to the zone's config (oxidecomputer/omicron#5050), the planner will also need to identify zones whose root filesystem is on one of these disks and replace it (i.e., write a new config with a root filesystem that's elsewhere) and we'll need to make sure that sled agent does the right thing with that

smklein:
There are more details with actually plumbing the expungement command through (https://github.com/oxidecomputer/omicron/pull/5994), but from the planner perspective, this is complete

PR #5375, opened 2024-04-01
[reconfigurator] Only provision to zpools with in-service physical disks

smklein:
- Expand reconfigurator's `Policy`, `SystemDescription` to be aware of "what disks back zpools".
- Updates callsites iterating over "all zpools" to consider whether or not they only want provisionable zpools, using active + in-service disks.
- Adds a method `sled_ensure_zones_not_using_expunged_disks`, which can be used to remove zones using expunged disks.
  - This implementation leaves some TODOs, however, as it depends on https://github.com/oxidecomputer/omicron/pull/5211 , and partially depends on https://github.com/oxidecomputer/omicron/pull/5050

Partial fix of https://github.com/oxidecomputer/omicron/issues/5372

### DIFF
--- a/nexus/reconfigurator/planning/src/blueprint_builder.rs
+++ b/nexus/reconfigurator/planning/src/blueprint_builder.rs
@@ -41,6 +41,7 @@ use omicron_common::api::internal::shared::NetworkInterfaceKind;
 use rand::rngs::StdRng;
 use rand::SeedableRng;
 use slog::o;
+use slog::warn;
 use slog::Logger;
 use std::collections::BTreeMap;
 use std::collections::HashSet;
@@ -56,6 +57,8 @@ use uuid::Uuid;
 /// Errors encountered while assembling blueprints
 #[derive(Debug, Error)]
 pub enum Error {
+    #[error("sled {sled_id} not found")]
+    SledNotFound { sled_id: Uuid },
     #[error("sled {sled_id}: ran out of available addresses for sled")]
     OutOfAddresses { sled_id: Uuid },
     #[error("no Nexus zones exist in parent blueprint")]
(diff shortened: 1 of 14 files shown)


nexus/reconfigurator/planning/src/planner.rs
```diff
@@ -110,6 +110,24 @@ impl<'a> Planner<'a> {
...
+                info!(
+                    &self.log,
+                    "found sled with expunged disks (will remove the zones using them)";
+                    "sled_id" => %sled_id
+                );
+                self.blueprint.comment(&format!(
+                    "sled {}: remove expunged disks",
+                    sled_id
+                ));
+                continue;
```
smklein, line 128: This seems to match the pattern below of "if we have any changes to make, do them and stop making further changes".

It seems plausible that we could remove an expunged zone and also do other allocations, but I'm not sure what our motivation is for "updating the blueprints via small steps", and wasn't confident enough to break that mold.

nexus/reconfigurator/planning/src/blueprint_builder.rs
```diff
@@ -472,6 +475,44 @@ impl<'a> BlueprintBuilder<'a> {
         Ok(Ensure::Added)
     }
 
+    pub fn sled_ensure_zones_not_using_expunged_disks(
```
smklein, line 478: This method is currently acting as a no-op, because of the TODOs below.

But I think the right action here would be to add the zones to an "expunged zone" field within the blueprint?

@sunshowers , do you think #5211 will support that? Or will I need to add a `BTreeSet` to the `BlueprintBuilder` for each zone type that could be removed?

smklein:
This PR is in merge hell and has been replaced by: 

- https://github.com/oxidecomputer/omicron/pull/5600
- https://github.com/oxidecomputer/omicron/pull/5601
- https://github.com/oxidecomputer/omicron/pull/5599

PR #5379, opened 2024-04-03
[sled-agent] Don't block load_services behind requesting firewall rules when no services are requested

smklein:


sled-agent/src/sled_agent.rs
```diff
@@ -614,13 +614,25 @@ impl SledAgent {
...
+                // If there wasn't any work to do, we're done immediately.
+                if matches!(
+                    load_services_result,
+                    Ok(services::LoadServicesResult::NoServicesToLoad)
+                ) {
+                    info!(
+                        self.log,
+                        "load_services exiting early; no services to be loaded"
+                    );
+                    return Ok(());
```
jgallagher, line 632: I'm not sure this is correct for this sequence:

1. `load_services()` succeeds and starts at least one new service
2. `request_firewall_update()` fails, so we backoff and retry
3. `load_services()` succeeds and now returns `NoServicesToLoad`

If we return `Ok(())` after 3, we never successfully requested the firewall rules for the service started in 1, right?

smklein: Is this sequencing possible? That would basically require that:

- During step (1) and (2), we had a ledger of services to run, and tried to start them
- During step (3), somehow, the ledger disappeared entirely.

Even when a sled is instructed to run "no services", I believe it still creates a ledger. So this scenario would only happen if somehow both M.2s were lost or corrupted?

jgallagher: Oh! I misunderstood `NoServicesToLoad`, sorry; I read that as "no services started", not "the list of services is empty". Sorry!

jgallagher (approved):

PR #5381, opened 2024-04-03
[tools] Don't try to delete the vdev when destroying virtual hardware

smklein:


tools/virtual_hardware.sh
```diff
@@ -49,11 +49,9 @@ function try_destroy_zpools {
     for ZPOOL_TYPE in "${ZPOOL_TYPES[@]}"; do
         readarray -t ZPOOLS < <(zfs list -d 0 -o name | grep "^$ZPOOL_TYPE")
         for ZPOOL in "${ZPOOLS[@]}"; do
-            VDEV_FILE="${VDEV_DIR:-/var/tmp}/$VDEV"
             zfs destroy -r "$ZPOOL" && \
                     (zfs unmount "$ZPOOL" || true) && \
                     zpool destroy "$ZPOOL" && \
```
leftwo, line 54: You need to replace the `&&` with`||` here, otherwise you will always print
the `warn` message.

leftwo (commented):
If a swap device using one of the virtual devices exists, that has to be removed
as well to destroy the zpool:  

```
root@EVT22200005:/alan/omicron# swap -l
swapfile             dev    swaplo   blocks     free
/dev/zvol/dsk/oxi_17588470-2b87-45b9-a564-924ab23acb47/swap 171,1         8 134217720 134217720
root@EVT22200005:/alan/omicron# swap -d /dev/zvol/dsk/oxi_17588470-2b87-45b9-a564-924ab23acb47/swap
```

tools/virtual_hardware.sh
smklein: Done

smklein:
> If a swap device using one of the virtual devices exists, that has to be removed as well to destroy the zpool:
> 
> ```
> root@EVT22200005:/alan/omicron# swap -l
> swapfile             dev    swaplo   blocks     free
> /dev/zvol/dsk/oxi_17588470-2b87-45b9-a564-924ab23acb47/swap 171,1         8 134217720 134217720
> root@EVT22200005:/alan/omicron# swap -d /dev/zvol/dsk/oxi_17588470-2b87-45b9-a564-924ab23acb47/swap
> ```

Went ahead and added some scripting to find and remove the swap device.

Tested this manually with the following:

1. `./tools/create_virtual_hardware.sh`
2. `pfexec zpool create oxp_test /var/tmp/u2_0.vdev && pfexec zfs create oxp_test/test_dataset && pfexec zfs create -V 2GB oxp_test/test_swap && pfexec swap -a /dev/zvol/dsk/oxp_test/test_swap`
3. `./tools/destroy_virtual_hardware.sh`

This spins up virtual hardware, adds a volume with swap to one of the zpools, then destroys it. The swap device is removed as a part of this.

tools/virtual_hardware.sh
```diff
@@ -45,18 +45,46 @@ function ensure_vdevs {
...
+
+            success "Verified ZFS pool $ZPOOL does not exist"
+        done
+    done
 
-            success "Verified ZFS pool and vdev $ZPOOL does not exist"
+    VDEV_TYPES=('m2_' 'u2_')
+    for VDEV_TYPE in "${VDEV_TYPES[@]}"; do
+        readarray -t VDEVS < <( \
+                grep "\"$VDEV_TYPE" "$OMICRON_TOP/smf/sled-agent/non-gimlet/config.toml" | \
```
leftwo, line 78: This could still trip up if someone had commented out a vdev line like I did, but that
is probably a me only kind of problem.

Though, I guess it will be consistent with what create_virtual_hardware.sh does.

leftwo (approved):
We can probably kick the handling of  `#vdev` to another time.

tools/virtual_hardware.sh
smklein: Agreed that this case needs to be handled, but yeah, was going to handle it separately. Not sure if I want to do this in bash, or just commit to some xtasks + rust binaries.

PR #5393, opened 2024-04-03
[test-utils] perform extra redactions at the beginning, not the end

sunshowers:
Fix for a rare-but-clearly-possible flake seen in https://github.com/oxidecomputer/omicron/pull/5386/checks?check_run_id=23376436739.

It looks like what happened is that a temporary directory was generated with a
string that looked something like `/var/tmp/omicron_tmp/.tmp12msH9Y`. The
redactor matched `12ms` before it matched the path.

To avoid this, perform extra redactions before the usual ones. I believe this
is always correct -- if a user has passed in an extra redaction, it should
probably be handled before all others.


davepacheco (approved):
Ugh.  Thanks for fixing this.

PR #5421, opened 2024-04-04
remove cargo registry patch for samael

iliana:
Fixes the following warning introduced by #5408:

```
warning: Patch `samael v0.0.14 (https://github.com/oxidecomputer/samael?branch=oxide/omicron#9e609a8f)` was not used in the crate graph.
Check that the patched package version and available features are compatible
with the dependency requirements. If the patch has a different version from
what is locked in the Cargo.lock file, run `cargo update` to use the new
version. This may also occur with an optional dependency that is not enabled.
```

#5408 bumped samael to v0.0.15, which brought in the upstream fix for #4920, which is why we had the patch in the first place.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11541,8 +11541,3 @@ dependencies = [
  "quote",
  "syn 1.0.109",
 ]
-
-[[patch.unused]]
-name = "samael"
-version = "0.0.14"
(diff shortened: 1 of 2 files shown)


iliana:
Separately, I'm gonna go see if Renovate has an issue for doing something a little smarter with patched crates, and also see if we can fail CI if we get one of these warnings from Cargo. (Hopefully something less fragile than `! grep -qF '[[patch.unused]]' Cargo.lock`.)

david-crespo (approved):

PR #5422, opened 2024-04-05
Refactor xtask commands into separate files

smklein:
This is a no-op PR in preparation for https://github.com/oxidecomputer/omicron/pull/5423 . I want to split xtasks into separate files so it's easier to see what's what.


sunshowers (approved):
Looks great, thanks!

papertigers (approved):
Thanks for doing this!

PR #5423, opened 2024-04-05
Xtask virtual hardware

smklein:
Adds an xtask to create and destroy virtual hardware.

For simplicity, and to avoid breaking external repos, this PR does not remove the `./tools/create_virtual_hardware.sh` scripts. However, that is the goal.

Depends on https://github.com/oxidecomputer/omicron/pull/5422
Part of https://github.com/oxidecomputer/omicron/issues/3939
Fixes https://github.com/oxidecomputer/omicron/issues/5401

PR #5425, opened 2024-04-05
use virtual-hardware xtask in buildomat

smklein:
Depends on https://github.com/oxidecomputer/omicron/pull/5423


.github/buildomat/jobs/deploy.sh
```diff
@@ -201,11 +201,16 @@ routeadm -e ipv4-forwarding -u
...
 PXA_END="$EXTRA_IP_END"
 
-# These variables are used by softnpu_init, so export them.
-export GATEWAY_IP GATEWAY_MAC PXA_START PXA_END
-
 pfexec zpool create -f scratch c1t1d0 c2t1d0
-VDEV_DIR=/scratch ptime -m pfexec ./tools/create_virtual_hardware.sh
+
+ptime -m \
+    cargo xtask virtual-hardware \
```
jclulow, line 207: There's no cargo in the lab environment, right?  Presumably you need to build the xtask binary and include that in whatever job output become inputs for the deploy job

smklein: Good catch, updated to take this approach

leftwo (approved):

PR #5430, opened 2024-04-05
[nexus] use the SledFilter type in the database as well

sunshowers:
Now that we have the `SledFilter` type, we should be able to use it for
database queries as well.

I spent some time trying to figure out the most ergonomic way to use the
`sled_filter` method, and this is the one I settled on.

This also serves as an example for how we're going to do this for blueprint
zone filters.

### DIFF
--- a/nexus/db-queries/src/db/datastore/sled.rs
+++ b/nexus/db-queries/src/db/datastore/sled.rs
@@ -482,8 +482,8 @@ impl DataStore {
     /// # Errors
     ///
     /// This method returns an error if the sled policy is not a state that is
-    /// valid to decommission from (i.e. if, the cur,
-    /// [`SledPolicy::is_decommissionable`] returns `false`).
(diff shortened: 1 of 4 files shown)


nexus/db-queries/src/db/datastore/sled.rs
```diff
@@ -198,26 +200,22 @@ impl DataStore {
...
+                        )
+                        .group_by(sled_dsl::id)
+                        .having(
+                            sled_has_space_for_threads
+                                .and(sled_has_space_for_rss)
+                                .and(sled_has_space_in_reservoir),
+                        )
+                        .filter(sled_dsl::time_deleted.is_null())
+                        // Ensure that reservations can be created on the sled.
+                        .sled_filter(SledFilter::ReservationCreate)
```
andrewjstone, line 216: This is pretty cool :)

andrewjstone (approved):
Looks great. I didn't realize you could add a diesel extension like this so "easily" :D

PR #5433, opened 2024-04-05
[nexus-db-queries] make vpc_resolve_to_sled be aware of expunged and decommissioned sleds

sunshowers:
We shouldn't be sending firewall rules to sleds that are gone.

This is also a nice demonstration of the new `ApplySledFilterExt` added in
#5430.

Depends on #5430.


andrewjstone (approved):
Awesome!

jgallagher (approved):

PR #5438, opened 2024-04-05
[nexus] add BlueprintZoneDisposition::Expunged, make disposition a column in bp_omicron_zone

sunshowers:
This PR (finally) adds a `BlueprintZoneDisposition::Expunged` enum, along with database support for it by making the disposition a column.

We also drop the `bp_omicron_zones_not_in_service` table -- we don't bother to check for its contents because we aren't ever setting any zones as quiesced on customer sites currently.

I've also included a Diesel extension to allow easily filtering against `BlueprintZoneDisposition`s, along with an example query for it.


schema/crdb/blueprint-disposition-column/up2.sql
```diff
@@ -0,0 +1,9 @@
+-- Add the disposition column to the bp_omicron_zone table.
+ALTER TABLE omicron.public.bp_omicron_zone
+    ADD COLUMN IF NOT EXISTS disposition omicron.public.bp_zone_disposition
+    NOT NULL
+    -- The only currently-representable zones are in-service and quiesced
+    -- (represented by bp_omicron_zones_not_in_service, which we're going to
+    -- drop in the next statement). We don't actually have any quiesced zones
+    -- yet, so it's fine to just do this.
+    DEFAULT 'in_service';
```
jgallagher, line 9: Do we want another migration step that removes this default?

sunshowers: Yeah we do I'm pretty sure, thanks for catching!

sunshowers: Done.

andrewjstone: Ah, that's clever.

andrewjstone (approved):
Great stuff @sunshowers

PR #5444, opened 2024-04-05
Not an error to filter an empty table

### LINKED ISSUE #5439: OxQL: want better response when zero results for query
When I make this otherwise perfectly valid query with a time range small enough to exclude all existing data points, I get back a 400 `InvalidRequest` with a confusing message. Should this really be an error? If we think so, the error message could still be more helpful.

```
get collection_target:cpus_provisioned 
  | filter timestamp > @now() - 1h && id == "001de000-1334-4000-8000-000000000000"
```

```
errorCode: "InvalidRequest"
message: "Table contains no timeseries to filter"
```

image

bnaecker:
Fixes #5439


david-crespo (approved):
Looks good. If there isn’t a test that’s broken by this, there should be!

PR #5262, opened 2024-03-14
Plumbing for maghemite + oximeter

rcgoodfellow:
This PR hooks up omicron to maghemite oximeter stats. Beyond a maghemite version bump, this PR plumbs a few additional SMF properties to the `mgd`/`ddmd` services running in the switch zone - and calls a new maghemite API endpoint to enable stats for `ddmd` instances running in the GZ. That is needed as `ddmd` runs before omicron as a part of bootstrap.

Depends on

- oxidecomputer/maghemite#178

Example `oxdb` output

```
0x〉select * from bgp_session:keepalives_sent limit 10;

 timeseries_key       | hostname     | local_asn | peer                  | rack_id                                | sled_id                                | start_time                      | timestamp                       | datum
----------------------+--------------+-----------+-----------------------+----------------------------------------+----------------------------------------+---------------------------------+---------------------------------+-------
 14086356469394695481 | "oxz_switch" | 47        | "::ffff:198.51.101.1" | "4186465a-970d-4936-9637-eff6d513a799" | "cf6917a3-1f6d-4313-8223-a50f32c24b2c" | "2024-03-19 16:02:03.735964791" | "2024-03-19 16:09:28.790198103" | 217
 14086356469394695481 | "oxz_switch" | 47        | "::ffff:198.51.101.1" | "4186465a-970d-4936-9637-eff6d513a799" | "cf6917a3-1f6d-4313-8223-a50f32c24b2c" | "2024-03-19 16:02:03.735964791" | "2024-03-19 16:09:29.775027721" | 217
 14086356469394695481 | "oxz_switch" | 47        | "::ffff:198.51.101.1" | "4186465a-970d-4936-9637-eff6d513a799" | "cf6917a3-1f6d-4313-8223-a50f32c24b2c" | "2024-03-19 16:02:03.735964791" | "2024-03-19 16:09:31.355500296" | 218
 14086356469394695481 | "oxz_switch" | 47        | "::ffff:198.51.101.1" | "4186465a-970d-4936-9637-eff6d513a799" | "cf6917a3-1f6d-4313-8223-a50f32c24b2c" | "2024-03-19 16:02:03.735964791" | "2024-03-19 16:09:31.775022639" | 218
[...]

PR #5323, opened 2024-03-25
sled-agent: Configure mgd and mg-ddm even if we don't have rack/sled IDs yet

jgallagher:
Based on #5322; can be merged after it or we can close that one in favor of this one (this is just the first commit).

Fixes #5318.

PR #5453, opened 2024-04-07
Update pinned maghemite to one on maghemite main

david-crespo:
In https://github.com/oxidecomputer/omicron/pull/5262, the pinned maghemite versions were set to commits from https://github.com/oxidecomputer/maghemite/pull/178 which no longer exist after that PR was merged. This PR updates the pinned maghemite to the merge commit from that PR on main. We can probably go farther forward, but this is the conservative fix.


david-crespo:
The fact that every test failed is probably a bad sign.

david-crespo:
Looks like maghemite version in `package-manifest.toml` was bumped in https://github.com/oxidecomputer/omicron/pull/5323 to a later one than I was trying, so what I'm now doing in the second commit is changing the version in the prereqs scripts to match that.

david-crespo:
Will be fixed in #5649

PR #5478, opened 2024-04-09
Adding OxQL table operations for limiting data

### LINKED ISSUE #5436: OxQL: want single most recent data point from a time range
An OxQL op that will lower to `ORDER BY timestamp DESC LIMIT 1`.

bnaecker:
- Resolves #5436
- Adds a `Limit` table op, for taking either the first or last `k` points from any number of timeseries.
- Add OxQL grammar rules for parsing `first k` and `last k` into the table op types.
- Add methods for coalescing the limit operations in a query, similar to coalescing predicates. This lets us push filters through limits (and vice versa), to implement predicate pushdown.
- Include limits when fetching data from ClickHouse. This uses the `LIMIT BY` clause, which makes this extremely simple to implement in the database itself. That also reveals an unfortunate interaction between our table sort ordering and the `start_time` for cumulative metrics, specifically those which are created before NTP is synced. That can lead to incorrect results when using the `LIMIT BY` clause, since that takes the first few results. That means data with an early timestamp, and early (and random) start time, can come before the desired data with a later timestamp (but also later start time). In this one case, with a `last k` table op, we pay the cost of a CTE and an additional re-sorting of the data based on timestamp.

### DIFF
--- a/oximeter/db/src/oxql/ast/table_ops/limit.rs
+++ b/oximeter/db/src/oxql/ast/table_ops/limit.rs
@@ -1,0 +1,263 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! An AST node apply limiting timeseries operations.
+
+// Copyright 2024 Oxide Computer Company
+
+use crate::oxql::point::Points;
+use crate::oxql::point::ValueArray;
+use crate::oxql::point::Values;
+use crate::oxql::Error;
+use crate::oxql::Table;
+use crate::oxql::Timeseries;
+use std::num::NonZeroUsize;
+
+/// The kind of limiting operation
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum LimitKind {
+    /// Limit the timeseries to the first points.
+    First,
+    /// Limit the timeseries to the last points.
+    Last,
+}
+
+/// A table operation limiting a timeseries to a number of points.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct Limit {
+    /// The kind of limit
+    pub kind: LimitKind,
+    /// The number of points the timeseries is limited to.
+    pub count: NonZeroUsize,
+}
+impl Limit {
+    /// Apply the limit operation to the input tables.
+    pub(crate) fn apply(&self, tables: &[Table]) -> Result<Vec<Table>, Error> {
+        if tables.is_empty() {
+            return Ok(vec![]);
+        }
+
+        tables
+            .iter()
+            .map(|table| {
+                let timeseries = table.iter().map(|timeseries| {
+                    let input_points = &timeseries.points;
+
+                    // Compute the slice indices for this timeseries.
+                    let (start, end) = match self.kind {
+                        LimitKind::First => {
+                            // The count in the limit operation should not be
+                            // larger than the number of data points.
+                            let end = input_points.len().min(self.count.get());
+                            (0, end)
+                        }
+                        LimitKind::Last => {
+                            // When taking the last k points, we need to
+                            // subtract the count from the end of the array,
+                            // taking care that we don't panic if the count is
+                            // larger than the number of data points.
+                            let start = input_points
+                                .len()
+                                .saturating_sub(self.count.get());
+                            let end = input_points.len();
+                            (start, end)
+                        }
+                    };
+
+                    // Slice the various data arrays.
+                    let start_times = input_points
+                        .start_times
+                        .as_ref()
+                        .map(|s| s[start..end].to_vec());
+                    let timestamps =
+                        input_points.timestamps[start..end].to_vec();
+                    let values = input_points
+                        .values
+                        .iter()
+                        .map(|vals| {
+                            let values = match &vals.values {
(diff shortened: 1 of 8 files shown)


oximeter/db/src/oxql/ast/table_ops/limit.rs
```diff
@@ -0,0 +1,263 @@
...
+                            (0, end)
+                        }
+                        LimitKind::Last => {
+                            // When taking the last k points, we need to
+                            // subtract the count from the end of the array,
+                            // taking care that we don't panic if the count is
+                            // larger than the number of data points.
+                            let start = input_points
+                                .len()
+                                .saturating_sub(self.count.get());
```
david-crespo, line 62: A nitpick, but I had to go look a few things up to make sure I understood this, whereas this is self-explanatory and doesn't rely on the fact that it's a usize and therefore intrinsically 0-bounded.

```rs
let start = (end - self.count.get()).max(0)
```

david-crespo: Oh. On second thought, the fact that it's a usize is relevant because if you subtract more from it it will die immediately. 🤦 never mind

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=6f45cdc8e262c24d2377e39f10958ebb

bnaecker: Yeah, the whole `saturating_sub()` business is to avoid any conditionals or panics if the count is greater than the length of the array. Nice tool for this kind of thing.

oximeter/db/src/oxql/query/mod.rs
```diff
@@ -834,4 +903,131 @@ mod tests {
...
+    fn test_coalesce_limits() {
+        let query = Query::new("get a:b | last 5").unwrap();
+        let lim = query.coalesced_limits(None).expect("Should have a limit");
+        assert_eq!(
+            lim.kind,
+            LimitKind::Last,
+            "This limit op has the wrong kind"
+        );
+        assert_eq!(lim.count.get(), 5, "Limit has the wrong count");
+    }
```
david-crespo, line 917: great tests and very helpful for understanding the logic

david-crespo (approved):
This all looks good. The only thing I feel it could use is an end to end test that really jams everything in there at once, like multiple limits + rearranging + negation on the timestamp filter.

bnaecker:
While adding a test like the one @david-crespo mentioned, I noticed a subtle edge case in the code this PR introduces. It's possible to see different data depending on whether the `last` table operation (not `first`) can be pushed down into a query to the database directly. In general, I think it should be possible to work around this, which I'll explain later. It's also important to note that the results in both cases are **correct**, they're just different, as we'll see. The summary of all this is that **we've agreed to defer** the work to make these two queries return identical data.

#### Example

Consider the following data, from a _cumulative_ timeseries (this only applies to those):

start_time | timestamp | value
-----------|-----------|------
0          | 1         | 1
0          | 2         | 2
0          | 3         | 3

Now, suppose we want to take the last **2** samples, e.g, the query ends with `| last 2`. The values we get back depend on whether we can prove that it's safe to push this limit into the database. Suppose first that we can push it. First, we'll select the points:

start_time | timestamp | value
-----------|-----------|------
0          | 2         | 2
0          | 3         | 3

When selecting cumulative data like this, we automatically compute deltas from it. That's to make each point independent, which lets us group or aggregate them, and also plot them easily. After taking the delta, we get:

start_time | timestamp | value
-----------|-----------|------
0          | 2         | 2
2          | 3         | 1

Now, suppose we _cannot_ push the limit into the database. In that case, we fall back to selecting all (or generally just _more_) of the timeseries; computing deltas; and then taking the limit in Rust code. So we select the whole history, and then take the diff to get:

start_time | timestamp | value
-----------|-----------|------
0          | 1         | 1
1          | 2         | 1
2          | 3         | 1

And now we take the last 2, resulting in:

start_time | timestamp | value
-----------|-----------|------
1          | 2         | 1
2          | 3         | 1

The **first point** in the returned timeseries is different. They have different start times and values, though the timestamp remains the same.

#### Deferring this

There are a few reasons to defer this. The first is that I think it's possible to fix it. We still have the deltas when we do the `last` operation in Rust code, which means we can take a cumulative sum of the values that will be **removed** by the `last` operation and add that to the first value and start time.

Second, while we _can_ fix it, it's not clear we need to. It's critical to note here that the data is **not wrong** in either case. We are correctly reporting the value between the start time and timestamp in both cases, however the start times are different. Depending on what one wants to do with the data, the difference may not even be noticeable. For example, when graphing these results, one almost certainly wants to plot the second and later points as _relative to_ the first. So the absolute value of the first isn't critical.

Second, if one does need to consider the absolute value, one would also need to consider the start time. That's also different, so it's not really like one can draw conflicting conclusions from this. Each point just describes the delta in a different window.

Last, while it's possible to fix this, I think it'll be pretty complicated. It would require some careful accumulation, for example to avoid catastrophic cancellation for doubles. It also requires walking the start times carefully to be sure that we don't set it "too early." For example, we might miss a time when the underlying cumulative counter was restarted. Those can be detected from the deltas, but we need to iterate all of them, we can't just unconditionally use the first start time. These are all possible, it's just a lot of work, for an uncertain gain which only applies in a few circumstances, none of which we currently find ourselves in.

So long-story short, I'm going to merge this, and we can come back to this work if we need to.

PR #5481, opened 2024-04-09
[xtask] Convert ci_download bash scripts to Rust

smklein:
Implements `cargo xtask download`, which provides options to replace the existing scripts:

- ci_download_clickhouse                                                                                                                                                                                                                                                                                                                                                                                                                          
- ci_download_cockroachdb                                                                                                                                                                                                                                                                                                                                                                                                                         
- ci_download_console                                                                                                                                                                                                                                                                                                                                                                                                                             
- ci_download_dendrite_openapi
- ci_download_dendrite_stub
- ci_download_maghemite_mgd
- ci_download_maghemite_openapi
- ci_download_softnpu_machinery
- ci_download_thundermuffin
- ci_download_transceiver_control

This PR additionally introduces the `cargo xtask download all` option, which attempts to download all artifacts concurrently.

Somewhat related to https://github.com/oxidecomputer/omicron/issues/3939


smklein:
Example usage:

```
 $ cargo xtask download --help
    Finished dev [unoptimized + debuginfo] target(s) in 0.80s
     Running `target/debug/xtask download --help`
Download binaries, OpenAPI specs, and other out-of-repo utilities

Usage: xtask download [OPTIONS] <TARGETS>...

Arguments:
  <TARGETS>...
          The targets to be downloaded. This list is additive

          Possible values:
          - all:                 Download all targets
          - clickhouse:          Clickhouse binary
          - cockroach:           CockroachDB binary
          - console:             Web console assets
          - dendrite-openapi:    Dendrite OpenAPI spec
          - dendrite-stub:       Stub Dendrite binary tarball
          - maghemite-mgd:       Maghemite mgd binary
          - maghemite-openapi:   Maghemite OpenAPI spec
          - softnpu:             SoftNPU, an admin program (scadm) and a pre-compiled P4 program
          - thundermuffin:       Thundermuffin binary, for sending non-stop buffers between devices
          - transceiver-control: Transceiver Control binary

Options:
      --output-dir <OUTPUT_DIR>
          The path to the "out" directory of omicron
          
          [default: out]

      --versions-dir <VERSIONS_DIR>
          The path to the versions and checksums directory
          
          [default: tools]

  -h, --help
          Print help (see a summary with '-h')
```

david-crespo:
@rcgoodfellow did you say maghemite openapi might not be necessary anymore?

rcgoodfellow:
> @rcgoodfellow did you say maghemite openapi might not be necessary anymore?

It's still necessary to download `ddmd` and `mgd` from maghemite CI. But I do not believe downloading the openapi spec should be needed anymore.

smklein:
CI failing looks like test flake, filed https://github.com/oxidecomputer/omicron/issues/5498

dev-tools/xtask/src/download.rs
```diff
@@ -0,0 +1,982 @@
...
+                        downloader.download_thundermuffin().await
+                    }
+                    Target::TransceiverControl => {
+                        downloader.download_transceiver_control().await
+                    }
+                }.context("Failed to download {target:?}")?;
+                Ok(())
+            }
+        })
+        .collect::<futures::stream::FuturesUnordered<_>>();
```
hawkw, line 152: If we're downloading multiple things, might we want to `tokio::spawn` the `download_${FOO}` calls, and put their `JoinHandle`s in the `FuturesUnordered`, instead? That way, the various downloads may run in parallel on multiple CPU cores.

smklein: Great advice, done in 3144d5d02e93379cc084ddc6854282fea3e2103b

smklein:
> > @rcgoodfellow did you say maghemite openapi might not be necessary anymore?
> 
> It's still necessary to download `ddmd` and `mgd` from maghemite CI. But I do not believe downloading the openapi spec should be needed anymore.

Removed this in the rust implementation in 0cd628ca6c74dace2d60443246dbbc9282460d5d

smklein:
The follow-up to remove the bash scripts is in: https://github.com/oxidecomputer/omicron/pull/5482

dev-tools/xtask/src/download.rs
```diff
@@ -0,0 +1,935 @@
...
+            ChecksumAlgorithm::Sha2,
+            &sha2,
+        )
+        .await?;
+        set_permissions(&path, 0o755).await?;
+
+        Ok(())
+    }
+
+    async fn download_thundermuffin(&self) -> Result<()> {
```
rcgoodfellow, line 824: Strictly speaking I don't think we need thundermuffin here. It's needed as an [intermediate package](https://github.com/oxidecomputer/omicron/blob/dbf849401c08c62bfcc64c28292f048469b10ef8/package-manifest.toml#L658-L665) for building the probe zone image. But we can forgo downloading it here I think.

dev-tools/xtask/src/download.rs
```diff
@@ -0,0 +1,935 @@
...
+        tokio::fs::create_dir_all(&download_dir).await?;
+
+        let checksums_path = self.versions_dir.join("maghemite_mgd_checksums");
+        let [mgd_sha2, mgd_linux_sha2] = get_values_from_file(
+            ["CIDL_SHA256", "MGD_LINUX_SHA256"],
+            &checksums_path,
+        )
+        .await?;
+        let commit_path =
+            self.versions_dir.join("maghemite_mg_openapi_version");
```
rcgoodfellow, line 741: Using the [rev value from Cargo.toml](https://github.com/oxidecomputer/omicron/blob/dbf849401c08c62bfcc64c28292f048469b10ef8/Cargo.toml#L273) is fine here, and we can probably get rid of the `maghemite_mg_openapi_version` file.

dev-tools/xtask/src/download.rs
smklein: I added this to correspond to https://github.com/oxidecomputer/omicron/blob/main/tools/ci_download_thundermuffin , which is currently being invoked by `./tools/install_builder_prerequisites.sh`.

Just to confirm: Are you okay with us not invoking `ci_download_thundermuffin`, and also deleting that bash script?

rcgoodfellow: Yup, that should be fine.

dev-tools/xtask/src/download.rs
smklein: I agree that we probably should do this, but it's slightly sticky because the `flake.nix` file also depends on having this as a rev.

I agree that we should minimize version dups - but if it's okay with you, I'm going to punt this to a different PR, since my goal is "parity with the bash scripts" before changing this download process **too** much.

hawkw: The Nix flake could be rewritten to also parse the rev from `Cargo.toml`, FWIW, but I would love to have advance warning so I can make sure to actually do that. :)

rcgoodfellow: SGTM

iliana (approved):
LGTM given the goal of "parity with the bash scripts" until we remove those.

The additional deps in xtask increase the build time from 30s to 40s on my machine, but I don't think that's enough to warrant this being an external xtask.

PR #5482, opened 2024-04-09
[xtask] Redirect old bash ci_download scripts

smklein:
As of https://github.com/oxidecomputer/omicron/pull/5481 , we have a rust replacement with `cargo xtask download`.

This removes the bash scripts.


leftwo:
Give the masses time to digest the new way!

smklein:
> Give the masses time to digest the new way!

Will do, mostly putting this up here to ensure it doesn't break anything. I'll try to give appropriate heads-up before deleting anything in a backwards incompatible way.

david-crespo:
Well, it passed. I successfully used the new xtasks today to get things going on macos, so I am happy with this whenever everyone else is ready.

I also updated my macos mgd and dendrite workaround to use the xtasks.

https://gist.github.com/david-crespo/d8aa7ea1afb877ff585e6ad90fc5bc2c

iliana:
What about leaving the old ci_download scripts for now but having them `exec cargo xtask download ...` instead? Maybe with a nice deprecation message :) That way we can merge this and no longer need to have two scripts to maintain but it won't break folks.

iliana:
(I went ahead and did what I suggested.)

sunshowers (approved):

PR #5488, opened 2024-04-10
[nexus-types] use TypedUuid<SledKind> in blueprint-internal code

sunshowers:
Most APIs aren't changing, nor is the OpenAPI schema for now -- just code
internal to the blueprint.

PR #5493, opened 2024-04-10
[reconfigurator] turn expunged sleds into expunged zones

sunshowers:
It actually looks like this bit falls out nicely. Turns out we were already
disregarding Nexus zones on expunged sleds, so more of it works than I thought
it did!

Still to do:

- [x] Fix and add to blueprint diff tests
- [x] Split up handling of internal and external IPs while constructing a `BlueprintBuilder`
- [x] Add tests for this

Depends on #5488 and #5555.


nexus/reconfigurator/planning/src/blueprint_builder.rs
```diff
@@ -275,49 +278,67 @@ impl<'a> BlueprintBuilder<'a> {
...
-            let zone_type = &z.zone_type;
-            if let OmicronZoneType::Nexus { nic, .. } = zone_type {
-                match nic.ip {
-                    IpAddr::V4(ip) => {
-                        if !existing_nexus_v4_ips.insert(ip) {
-                            bail!("duplicate Nexus NIC IP: {ip}");
+            let zone_type = &z.config.zone_type;
+
+            // Running services have internal IPs.
+            if z.disposition.matches(BlueprintZoneFilter::ShouldBeRunning) {
```
sunshowers, line 307: Is this correct? I'm a little worried that this isn't quite what we want, because it's possible that the last blueprint hasn't yet finished execution, and that the resources haven't actually been freed up, by the time we're generating the new one. (In that case, should resources allocated to expunged zones just live forever until the zones are garbage collected -- a process we aren't doing for now?)

sunshowers: Yeah.... having thought more about it I'm pretty sure this is wrong, and that we actually need to get to the point where we've removed all records.

sunshowers: Discussed this in a meeting yesterday -- we need to check the planning input to see if the zones are still there.

sunshowers: In today's meeting we decided to just punting on garbage collection for release 8, which simplifies matters considerably.

sunshowers: https://github.com/oxidecomputer/omicron/issues/5552 tracks this.

nexus/reconfigurator/planning/src/blueprint_builder.rs
```diff
@@ -353,30 +366,26 @@ impl<'a> BlueprintBuilder<'a> {
         // of Nexus instances), but wouldn't be ideal if we have many resources
         // we need to skip. We could do something smarter here based on the sets
         // of used resources we built above if needed.
-        let nexus_v4_ips = Box::new(
+        let nexus_v4_ips = AvailableIterator::new(
```
sunshowers, line 369: I ended up not using the extra functionality provided by `AvailableIterator`, but this will be essential for testing https://github.com/oxidecomputer/omicron/issues/5552.

nexus/types/src/deployment/planning_input.rs
```diff
@@ -384,6 +410,18 @@ impl PlanningInput {
         self.sleds.get(sled_id).map(|details| &details.resources)
     }
 
+    pub fn omicron_zone_external_ips(
+        &self,
+    ) -> &BTreeMap<OmicronZoneUuid, OmicronZoneExternalIp> {
+        &self.omicron_zone_external_ips
+    }
+
```
davepacheco, line 418: This appears unused?

nexus/types/src/deployment/planning_input.rs
```diff
@@ -384,6 +410,18 @@ impl PlanningInput {
...
+    pub fn omicron_zone_nics(
+        &self,
+    ) -> &BTreeMap<OmicronZoneUuid, OmicronZoneNic> {
+        &self.omicron_zone_nics
+    }
+
```
davepacheco, line 424: This also appears unused?

nexus/types/src/deployment/planning_input.rs
```diff
@@ -333,6 +333,32 @@ pub struct SledDetails {
...
 
+impl SledDetails {
+    /// Returns `Some(reason)` if the sled needs its zones to be expunged,
+    /// based on the policy and state.
+    pub fn needs_zone_expungement(&self) -> Option<ZoneExpungeReason> {
+        match self.state {
+            SledState::Active => {}
+            SledState::Decommissioned => {
+                // A decommissioned sled that still has resources attached to
+                // it is an illegal state, but representable. If we see a sled
```
davepacheco, line 344: This comment trails off.

nexus/types/src/deployment/planning_input.rs
```diff
@@ -333,6 +333,32 @@ pub struct SledDetails {
...
+            SledPolicy::Expunged => Some(ZoneExpungeReason::SledExpunged),
+        }
+    }
+}
+
+#[derive(Copy, Clone, Debug)]
+pub enum ZoneExpungeReason {
+    SledDecommissioned,
+    SledExpunged,
+}
```
davepacheco, line 360: Why do we need this?  Is it just for debugging?  (I'm asking because trying to use it for correctness seems fraught.)

nexus/types/src/deployment/planning_input.rs
```diff
@@ -333,6 +333,32 @@ pub struct SledDetails {
     pub resources: SledResources,
 }
 
+impl SledDetails {
+    /// Returns `Some(reason)` if the sled needs its zones to be expunged,
+    /// based on the policy and state.
+    pub fn needs_zone_expungement(&self) -> Option<ZoneExpungeReason> {
```
davepacheco, line 339: This is pretty nitty, but: this behavior seems really specific to planning and it's a little surprising to me that it would live in nexus/types and under PlanningInput.  Why wouldn't the planner check this condition explicitly?

nexus/reconfigurator/planning/src/blueprint_builder.rs
```diff
@@ -436,6 +445,79 @@ impl<'a> BlueprintBuilder<'a> {
...
+    /// Builds a list of zones that are:
+    ///
+    /// * not *currently* expunged from the given sled
+    /// * but *need to be expunged* from this sled
+    ///
+    /// without actually performing the expungement.
+    ///
+    /// This method does not check whether the sled actually should be expunged
+    /// -- that is the responsibility of higher-level code.
+    pub fn build_pending_expunge_for_sled(
```
davepacheco, line 457: Why is this separated into two steps?  I'm wondering if that makes it possible to misuse it by accident.

nexus/reconfigurator/planning/tests/output/planner_nonprovisionable_1_2.txt
```diff
@@ -1,89 +1,137 @@
...
+     crucible       c406da50-34b9-4bb4-a460-8f49875d2a6a   in service    fd00:1122:3344:105::24           
+     crucible       d660d7ed-28c0-45ae-9ace-dc3ecf7e8786   in service    fd00:1122:3344:105::2a           
+     crucible       e98cc0de-abf6-4da4-a20d-d05c7a9bb1d7   in service    fd00:1122:3344:105::2b           
+     crucible       f55e6aaf-e8fc-4913-9e3c-8cd1bd4bdad3   in service    fd00:1122:3344:105::29           
+     internal_ntp   7f4e9f9f-08f8-4d14-885d-e977c05525ad   in service    fd00:1122:3344:105::21           
+     nexus          6dff7633-66bb-4924-a6ff-2c896e66964b   in service    fd00:1122:3344:105::22           
+                                                                                                          
+  MODIFIED SLEDS:                                                                                         
+                                                                                                          
+*  sled 48d95fef-bc9f-4f50-9a53-1e075836291d: zones at generation: 2 -> 3
```
davepacheco, line 26: pretty nitty and maybe not worth doing anything about now: it's a little weird to me that we go from generation 2 -> 3 to mark all the zones expunged.  I think that's because I think of this as the generation number of the list of zones that we tell the sled agent about.  When the sled is expunged, that list is non-existent.  Arguably we could say that's generation 3 and empty -- that'd be fine -- but it wouldn't contain expunged zones.

On the other hand, with the splitting of inventory zones vs. blueprint zones, I guess expunged zones may be in the blueprint but not the inventory or the set that we send the sled agent.  So maybe the generation number here covers not the list that we send to the sled agent, but the blueprint list.  And that does include expunged zones.  And then the list of zones that we would send to the sled is a filter over the blueprint list?  If that's how we're thinking of this (which seems reasonable), then I think I might have this output say "blueprint zones at generation: 2 -> 3" instead of just "zones at generation: 2 -> 3".

It might also be worth annotating the sled row to say that the sled itself is expunged.

nexus/reconfigurator/planning/tests/output/planner_nonprovisionable_1_2.txt
```diff
@@ -1,89 +1,137 @@
...
+     crucible       f55e6aaf-e8fc-4913-9e3c-8cd1bd4bdad3   in service    fd00:1122:3344:105::29           
+     internal_ntp   7f4e9f9f-08f8-4d14-885d-e977c05525ad   in service    fd00:1122:3344:105::21           
+     nexus          6dff7633-66bb-4924-a6ff-2c896e66964b   in service    fd00:1122:3344:105::22           
+                                                                                                          
+  MODIFIED SLEDS:                                                                                         
+                                                                                                          
+*  sled 48d95fef-bc9f-4f50-9a53-1e075836291d: zones at generation: 2 -> 3                                 
+-    crucible       094f27af-1acb-4d1e-ba97-1fc1377d4bf2   in service    fd00:1122:3344:103::2c  modified 
++     ├─                                                   expunged      fd00:1122:3344:103::2c           
+*     └─ changed: disposition
```
davepacheco, line 29: Can we include the expunge reason in this output?

nexus/reconfigurator/planning/src/blueprint_builder.rs
davepacheco: What is the extra functionality?

nexus/reconfigurator/planning/src/blueprint_builder.rs
```diff
@@ -149,14 +152,14 @@ pub struct BlueprintBuilder<'a> {
     comments: Vec<String>,
 
     // These fields mirror how RSS chooses addresses for zone NICs.
-    nexus_v4_ips: Box<dyn Iterator<Item = Ipv4Addr> + Send>,
-    nexus_v6_ips: Box<dyn Iterator<Item = Ipv6Addr> + Send>,
+    pub(crate) nexus_v4_ips: AvailableIterator<'static, Ipv4Addr>,
```
davepacheco, line 155: It makes me nervous to expose this (and the other iterators) to the whole crate.  Can we limit that somehow?  It seems like this is only used by the tests at least.

davepacheco (commented):
Nice!

nexus/reconfigurator/planning/src/blueprint_builder.rs
sunshowers: This lets you look at which resources were considered in-use at the time the iterator was constructed.

nexus/reconfigurator/planning/src/blueprint_builder.rs
sunshowers: Ah yeah this isn't needed right now, fixed. I needed it to figure out which resources were considered to be in use in the planner, but I'll figure out an alternative way to do that as part of #5552.

nexus/reconfigurator/planning/src/blueprint_builder.rs
```diff
@@ -436,6 +445,79 @@ impl<'a> BlueprintBuilder<'a> {
...
+    ///
+    /// This method does not check whether the sled actually should be expunged
+    /// -- that is the responsibility of higher-level code.
+    pub fn build_pending_expunge_for_sled(
+        &self,
+        sled_id: SledUuid,
+        reason: ZoneExpungeReason,
+    ) -> Option<PendingExpunge> {
+        let mut zones_to_expunge = Vec::new();
+        let sled_zones = self.zones.current_sled_zones(sled_id);
```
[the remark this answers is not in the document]
sunshowers: Good point, and not something the current model considers! Definitely worth guarding against.

nexus/reconfigurator/planning/src/blueprint_builder.rs
sunshowers: Good question -- the main reason is that I wanted to be able to unify the set of zones that should be expunged across all the possible sources of zone expungement, so it can be done in one go. The current ones are:

* sled expungement => all zones on the sled
* disk expungement => crucible zones backed by the disk

So the idea would be: first, build up the set of zones that should be expunged, then actually perform the expungement. It's a pattern I've generally found to be very good for introspection and testing against.

re misuse, one way is to add `#[must_use]` to a pending expunge, and I'll do so. But in general this should be caught pretty quickly by tests.

nexus/reconfigurator/planning/src/planner.rs
```diff
@@ -847,11 +919,27 @@ mod test {
         );
         assert_eq!(diff.sleds_added().len(), 0);
         assert_eq!(diff.sleds_removed().len(), 0);
-        let sleds = diff.sleds_modified().collect::<Vec<_>>();
+        let mut sleds = diff.sleds_modified().collect::<HashMap<_, _>>();
+
+        // The expunged and decommissioned sleds should have had all zones
```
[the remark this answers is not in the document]
sunshowers: Sorry, yes, it should be "all zones expunged". I'll fix this comment as well as L915.

`sleds_removed` means sleds removed from the blueprint, not merely marked as expunged. That's not something we're doing at the moment and we should clarify that in this test.

nexus/types/src/deployment/planning_input.rs
sunshowers: Yes, just for logging/debugging/testing. I'll make a note about this.

nexus/types/src/deployment/planning_input.rs
sunshowers: Indeed, this and the below one was part of the half-done code in #5552. No reason to keep it around for now. Removed.

nexus/types/src/deployment/planning_input.rs
sunshowers: Removed as well.

nexus/reconfigurator/planning/tests/output/planner_nonprovisionable_1_2.txt
sunshowers: Yeah, re this and the next one -- we're currently only storing the fact that the zone or sled is expunged at the moment. We aren't storing information about why the zone or sled was expunged in any persistent form. While writing this I started encountering the issue you did as well -- a sled or a bunch of zones being marked expunged without that context can be confusing.

I'd like to revisit this after release 8, potentially adding a comment field to the database. Is that okay?

nexus/reconfigurator/planning/tests/output/planner_nonprovisionable_1_2.txt
sunshowers: See above. Ideally yes, currently it's hard.

nexus/reconfigurator/planning/src/blueprint_builder.rs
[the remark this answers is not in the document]
sunshowers: (It's kind of like how `git rm` to a newly added file requires you to pass in `-f`)

nexus/types/src/deployment/planning_input.rs
sunshowers: Haha that was because as I was writing this comment I started thinking about the situations where this could happen, and went down a long rabbit hole. I'll fix this.

nexus/types/src/deployment/planning_input.rs
sunshowers: I see your point here. I think originally, I was thinking of sharing this across planning and some other code -- but that's not necessary for now. I'll move it to the planner.

sunshowers: Done.

nexus/reconfigurator/planning/src/blueprint_builder.rs
```diff
@@ -300,17 +303,26 @@ impl<'a> BlueprintBuilder<'a> {
...
+        //
+        // * We don't unassign a resource from an expunged sled/zone and
+        //   reassign it to a different zone.
+        // * We also don't perform any kind of garbage collection here: a sled
+        //   or zone that's marked expunged will always stay in that state.
+        //
+        // https://github.com/oxidecomputer/omicron/issues/5552 tracks
+        // implementing this kind of garbage collection, and we should do it
+        // very soon.
+
```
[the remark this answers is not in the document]
sunshowers: Done, and made it flow better. Thanks.

nexus/reconfigurator/planning/src/planner.rs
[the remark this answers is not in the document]
sunshowers: Fixed.

nexus/types/src/deployment/planning_input.rs
sunshowers: Done.

nexus/reconfigurator/planning/src/blueprint_builder.rs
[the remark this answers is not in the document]
sunshowers: To achieve this we need to track what happened in the builder previously -- https://github.com/oxidecomputer/omicron/pull/5555 has that code.

sunshowers:
Want to add a few more tests tomorrow, but hopefully we can get some cycles going with this.

nexus/reconfigurator/planning/tests/output/planner_nonprovisionable_1_2.txt
davepacheco: Sounds good.  Are you okay with adding the word "blueprint" to the "zones at generation", just to clarify that we're now talking about the blueprint zones and not the set we're sending to the sled agent?

nexus/reconfigurator/planning/src/blueprint_builder.rs
davepacheco: I think you'd get the same behavior from a function that was just `let summary = builder.expunge_sled(sled_id, reason)` because as far as I can tell, the caller always:

- calls `build_pending_expunge_for_sled()` 0 or more times
- for each returned value, call `apply_pending_expunge`

i.e., there's no merging or other processing of the returned values before calling `apply()`.

I can definitely see separating it for testing but callers could use a wrapper that always composes the two functions correctly.

It makes sense to use `#[must_use]` to ensure that the caller doesn't forget to call `apply()`.  There's also the risk that they try to call it more than once.  I think this is impossible today because `apply()` takes ownership and `PendingExpunge` doesn't impl `Clone` (though I can see somebody wanting to add `Clone` for some other reason and then introducing this risk).  I think what I'm most worried about is actually that separating the computation of `PendingExpunge` from applying it means there's a window in which a caller could do any number of other operations that invalidate the `PendingExpunge`.  What if they use other `BlueprintBuilder` functions to deploy more zones to that sled in the meantime?  Or use `sled_ensure_disks()` or something like that.  Or it might be easy to do something like:

```
// there are 3 nexus zones, one of which is on $sled_id
builder.build_pending_expunge_for_sled($sled_id, ...)
// go calculate how many Nexus zones should exist and erroneously include the one on $sled_id because it hasn't actually been expunged yet
builder.apply_pending_expunge()
```

Of course we can add validation to various functions to prevent this stuff but it just seems more complicated than structurally ensuring that external callers _can't_ do anything in the meantime because they have no opportunity to.

That's my pitch but this isn't that important either way and I'm fine if you want to keep it as is!

jgallagher: > I can definitely see separating it for testing but callers could use a wrapper that always composes the two functions correctly.

I think I would vote for this, with the caveat that it doesn't look like we currently have any tests these two in isolation, so maybe the split might be premature even for this case vs joining them into one function that returns a summary as you suggested?

nexus/reconfigurator/planning/src/blueprint_builder.rs
```diff
@@ -1025,6 +1165,26 @@ mod builder_zones {
...
+            zones: &[OmicronZoneUuid],
+        ) -> Result<(), BuilderZonesConfigError> {
+            for zone_id in zones {
+                // A linear search for each zone is fine, each sled has a pretty
+                // small number of zones.
+                if let Some(zone) =
+                    self.zones.iter_mut().find(|z| z.zone.id == *zone_id)
+                {
+                    // Just check that the zone is still expungeable.
+                    is_already_expunged(&zone.zone, zone.state)?;
```
jgallagher, line 1179: I don't entirely follow what this is guarding here. In particular, what would a return value of `Ok(true)` mean? We already expunged the zone but forgot to mark the state as `Modified` somehow?

nexus/reconfigurator/planning/src/blueprint_builder.rs
sunshowers: > i.e., there's no merging or other processing of the returned values before calling apply().

So... this is how the code is today, but that's only because the only source of zone expungement is *sled* expungement. In the very near future (release 9 I believe), we'll have a second source of zone expungement: individual *disk* expungement. In that case, it is possible that a zone is expunged for multiple reasons -- e.g. for a crucible zone, more than one of the backing disks was expunged.

In that case, there are two ways to address this:

1. Compute and apply each expungement directly to the builder, relying on the fact that expungement is idempotent.
2. Build a set of pending expunges, merge all of them building up the reasons for each, and then apply all of them together.

From experience, absent other overriding concerns the solution I generally tend to gravitate towards in these situations is 2. 1 is definitely harder to misuse, but 2 allows for many things, including dry-runs and easier testing. Many, many times in my life I've rewritten code from 1 to 2, and so I just start from pattern 2 these days. (In this case, one of the immediate things pattern 2 does is make us able to produce better comments for the blueprint.)

---

But let me try and make the case along the same lines for pattern 1:

* A blueprint itself is pattern 2, a dry-run. A blueprint builder is a dry-run of a dry-run. This is a dry-run^3, which maybe is too much dry-run.
* Maybe the solution to providing better comments is to actually do the merging *within the comment layer*. So instead of storing raw strings, we store an operation log that we can merge intelligently, and then resolve to the list of comments while finishing up the blueprint.

That is on top of the undeniable argument that pattern 1 is harder to misuse. So given that both of you feel this way, I think I'll switch to pattern 1.

nexus/reconfigurator/planning/src/blueprint_builder.rs
sunshowers: This is making sure that any zones passed in at this level are in a state where marking them as expunged is valid. This is an unavoidable check that happens a second time. The list of zones to expunge has to be passed in -- it cannot be constructed here, because a `BuilderZonesConfig` is only created after we've decided to change the sled. If there are no zones to expunge then a `BuilderZonesConfig` shouldn't be created.

nexus/reconfigurator/planning/src/planner.rs
```diff
@@ -73,9 +76,76 @@ impl<'a> Planner<'a> {
...
+                         allocated to it; will expunge them \
+                         (sled policy is \"{}\")", sled_details.policy;
+                    );
+                }
+                ZoneExpungeReason::SledExpunged => {
+                    // This is the expected situation.
+                    info!(
+                        &log,
+                        "expunged sled with non-expunged zones found \
+                         (will expunge all zonewith non-expungeds)";
```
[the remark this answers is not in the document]
sunshowers: Fixed, thanks.

nexus/reconfigurator/planning/src/planner.rs
```diff
@@ -72,9 +76,79 @@ impl<'a> Planner<'a> {
...
+                continue;
+            };
+
+            match reason {
+                ZoneExpungeReason::SledDecommissioned => {
+                    // A sled marked as decommissioned should have no resources
+                    // allocated to it. If it does, it's an illegal state,
+                    // possibly introduced by a bug elsewhere in the system --
+                    // we need to warn on this, and also remove the zones.
+                    warn!(
```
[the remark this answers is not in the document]
sunshowers: Oh huh, I thought this was an error already, that was definitely the intention. Fixed, thanks.

nexus/reconfigurator/planning/tests/output/planner_nonprovisionable_1_2.txt
sunshowers: Sure, that makes sense.

nexus/reconfigurator/planning/src/blueprint_builder.rs
jgallagher: Hm, but am I right that `Ok(true)` would be a nonsensical return value? Would it make sense to separate "validate that this zone is expungeable" from "has the zone already been marked expunged", and only call the first one here?

nexus/reconfigurator/planning/src/blueprint_builder.rs
sunshowers: I've switched to pattern 1 in this case, though (as outlined above) I think it'll have some repercussions down the road that we'll have to deal with.

nexus/reconfigurator/planning/tests/output/planner_nonprovisionable_1_2.txt
sunshowers: Done.

nexus/reconfigurator/planning/src/blueprint_builder.rs
sunshowers: `Ok(true)` indicates a no-op (idempotence). I changed the definition of `is_already_expunged` slightly, and especially with this new definition -- but even with the old one -- I don't think it makes sense to split them out. The simplest definition of a `validate_expungeable` function would be to just call `is_already_expunged()?;` and drop the return value.

(I went through this series of arguments myself while writing this code. Options I considered included renaming `is_already_expunged` to also include validation -- I decided that was too wordy, and already indicated by the fact that it returns a `Result`.)

nexus/reconfigurator/planning/tests/output/planner_nonprovisionable_1_2.txt
sunshowers: Filed https://github.com/oxidecomputer/omicron/issues/5586 for this.

andrewjstone (approved):
Great work @sunshowers. I [tested this on a4x2](https://github.com/oxidecomputer/omicron/issues/5480#issuecomment-2067547066) and it seems to properly mark zones as expunged. I suggest we get it in asap and build upon it.

sunshowers:
Thanks @andrewjstone! Going to land this and we can take care of any remaining comments in followups.

PR #5501, opened 2024-04-10
Use TypedUuid for Zpools in Sled Agent

smklein:
Starts the conversion of using "typed UUIDs" within Sled Agent. This is a thin vertical slice, but if we're happy with how this looks, I can keep propagating types to other UUIDs in this area.

- I tried to convert effectively all usage within the sled agent and family of crates
- I intentionally did not apply any conversions to the DB schema within Nexus -- so the changes propagating up there are incidental to using Sled Agent APIs

### DIFF
--- a/uuid-kinds/src/lib.rs
+++ b/uuid-kinds/src/lib.rs
@@ -20,16 +20,20 @@ use schemars::JsonSchema;
 macro_rules! impl_typed_uuid_kind {
     ($($kind:ident => $tag:literal),* $(,)?) => {
         $(
-            #[cfg_attr(feature = "schemars08", derive(JsonSchema))]
-            pub enum $kind {}
+            paste::paste! {
+                #[cfg_attr(feature = "schemars08", derive(JsonSchema))]
+                pub enum [< $kind Kind>] {}
 
-            impl TypedUuidKind for $kind {
-                #[inline]
-                fn tag() -> TypedUuidTag {
-                    // `const` ensures that tags are validated at compile-time.
-                    const TAG: TypedUuidTag = TypedUuidTag::new($tag);
-                    TAG
+                impl TypedUuidKind for [< $kind Kind >] {
+                    #[inline]
+                    fn tag() -> TypedUuidTag {
+                        // `const` ensures that tags are validated at compile-time.
+                        const TAG: TypedUuidTag = TypedUuidTag::new($tag);
+                        TAG
+                    }
                 }
+
+                pub type [< $kind Uuid>] = TypedUuid::<[< $kind Kind >]>;
             }
         )*
     };
@@ -45,13 +49,14 @@ macro_rules! impl_typed_uuid_kind {
 // Please keep this list in alphabetical order.
 
 impl_typed_uuid_kind! {
-    DownstairsKind => "downstairs",
-    DownstairsRegionKind => "downstairs_region",
-    LoopbackAddressKind => "loopback_address",
-    OmicronZoneKind => "service",
-    SledKind => "sled",
-    TufRepoKind => "tuf_repo",
-    UpstairsKind => "upstairs",
-    UpstairsRepairKind => "upstairs_repair",
-    UpstairsSessionKind => "upstairs_session",
+    Downstairs => "downstairs",
+    DownstairsRegion => "downstairs_region",
+    LoopbackAddress => "loopback_address",
+    OmicronZone => "service",
+    Sled => "sled",
+    TufRepo => "tuf_repo",
+    Upstairs => "upstairs",
+    UpstairsRepair => "upstairs_repair",
+    UpstairsSession => "upstairs_session",
+    Zpool => "zpool",
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3511,6 +3511,7 @@ dependencies = [
  "macaddr",
  "mockall",
  "omicron-common",
+ "omicron-uuid-kinds",
  "omicron-workspace-hack",
  "opte-ioctl",
  "oxide-vpc",
@@ -4857,6 +4858,7 @@ dependencies = [
  "omicron-passwords",
  "omicron-sled-agent",
  "omicron-test-utils",
+ "omicron-uuid-kinds",
  "omicron-workspace-hack",
  "oximeter",
  "oximeter-collector",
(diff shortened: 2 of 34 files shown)


uuid-kinds/src/lib.rs
```diff
@@ -20,16 +20,20 @@ use schemars::JsonSchema;
...
-
-            impl TypedUuidKind for $kind {
-                #[inline]
-                fn tag() -> TypedUuidTag {
-                    // `const` ensures that tags are validated at compile-time.
-                    const TAG: TypedUuidTag = TypedUuidTag::new($tag);
-                    TAG
+            paste::paste! {
+                #[cfg_attr(feature = "schemars08", derive(JsonSchema))]
+                pub enum [< $kind Kind>] {}
```
smklein, line 25: This change is worth reviewing -- these types are named the same thing as before, but I'm just automatically applying the `Kind` suffix with `paste`

uuid-kinds/src/lib.rs
```diff
@@ -20,16 +20,20 @@ use schemars::JsonSchema;
...
+                impl TypedUuidKind for [< $kind Kind >] {
+                    #[inline]
+                    fn tag() -> TypedUuidTag {
+                        // `const` ensures that tags are validated at compile-time.
+                        const TAG: TypedUuidTag = TypedUuidTag::new($tag);
+                        TAG
+                    }
                 }
+
+                pub type [< $kind Uuid>] = TypedUuid::<[< $kind Kind >]>;
```
smklein, line 36: I'm also making a type alias for the type, which is where `ZpoolUuid` comes from. IMO this is slightly easier to parse than `TypedUuid::<ZpoolKind>`, but if there are strong opinions against this pattern, I can switch back.

jgallagher: There was some discussion about this on the initial PR: https://github.com/oxidecomputer/omicron/pull/4934#issuecomment-1918078400, https://github.com/oxidecomputer/omicron/pull/4934#issuecomment-1918081744. I think I'd vote with that and not use type aliases, although `ZoneUuid` is definitely clearer than `ZoneId` (which was the proposal in that discussion).

jgallagher: To be clear I'm weakly voting for `TypedUuid<FooKind>`; if there's disagreement here I'm happy to go with the majority.

smklein: If I am in the minority, I'm willing to "disagree and commit" to this approach. I totally get that this is a choice which the `newtype_uuid` crate didn't want to have, but `omicron_uuid_kinds` seems like it could be more opinionated.

(it just feels so wordy! `TypedUuid<ZpoolKind>` is like double the characters of `ZpoolUuid`)

sunshowers: Having written some code that uses `TypedUuid`, I think I've changed my mind here, and in particular I like the `ZoneUuid` pattern a lot. I'd just want to only use it for this commonly-used type.

If we like it, we should probably commit to this now since the more folks use it the harder it will be to make changes.

It would also be worth updating the readme to include this information.

uuid-kinds/src/lib.rs
```diff
@@ -20,16 +20,20 @@ use schemars::JsonSchema;
...
-            #[cfg_attr(feature = "schemars08", derive(JsonSchema))]
-            pub enum $kind {}
-
-            impl TypedUuidKind for $kind {
-                #[inline]
-                fn tag() -> TypedUuidTag {
-                    // `const` ensures that tags are validated at compile-time.
-                    const TAG: TypedUuidTag = TypedUuidTag::new($tag);
-                    TAG
+            paste::paste! {
```
sunshowers, line 23: Hmm, `paste` is a proc macro which increases build times, but that's probably ok for this set of things. The alternative would be to have the macro be something like:

```rust
impl_typed_uuid_kind! {
    (DownstairsKind, DownstairsUuid) => "downstairs",
    ...
}
```

which feels worse.

sunshowers (commented):
Love it!

uuid-kinds/src/lib.rs
sunshowers: Basically the risk I'm concerned about is that e.g. I just added `TypedUuidRng<T>` in #5488 -- I don't want us to grow another set of aliases for things that aren't commonly used. But I think the UUID types themselves are above the threshold where a type alias is useful, and `TypedUuidRng<T>`, `DbTypedUuid<T>` etc are below that threshold.

uuid-kinds/src/lib.rs
smklein: While it is a proc macro, shouldn't it only run once for the `omicron-uuid-kinds` crate and then "be done" for other crates which depend on `omicron-uuid-kinds`?

(Very possible I'm wrong on this -- also happy to change it to the version with more arguments if we think the compilation cost would spread)

sunshowers: Yes, it's just a one-time thing, and plenty of other crates in omicron have much harsher dependencies on proc macros. I'm just a bit paranoid about proc macro build times, in this case unreasonably so.

uuid-kinds/src/lib.rs
andrewjstone: If @sunshowers is cool with it, I'm cool with it :)

sunshowers (approved):
Thanks! Only real comment I have is just adding some notes about the type aliases to the readme.

uuid-kinds/src/lib.rs
jgallagher: > Basically the risk I'm concerned about is that e.g. I just added `TypedUuidRng<T>` in #5488 -- I don't want us to grow another set of aliases for things that aren't commonly used. But I think the UUID types themselves are above the threshold where a type alias is useful, and `TypedUuidRng<T>`, `DbTypedUuid<T>` etc are below that threshold.

This balance makes sense to me. So we create aliases for `TypedUuid<FooKind> -> FooUuid` (and the change from `FooId` to `FooUuid` really helps here I think!), but we _don't_ create aliases for other the other typed UUID supporting cast?

sunshowers: Yeah.

sunshowers:
Sean's busy, so with his permission I've fixed the tests, added a section to the readme, and am going to go ahead and land this. Thanks again Sean!

Issue #5503, opened 2024-04-10
Nexus must create "Control Plane Disks" for disks automatically


smklein:
This is the root case of https://github.com/oxidecomputer/omicron/issues/5502 .

(The decision to create a new issue is a little arbitrary, I just wanted to deep-dive onto the root cause here, while letting the symptoms stay visible on #5502).

As of https://github.com/oxidecomputer/omicron/pull/5172 , RSS / Nexus are responsible for managing "control plane disks" and "control plane zpools". In this world, the sled agent no longer autonomously formats disks, but waits for the "go-ahead" from Nexus.

For bootstrapping, RSS also has this responsibility:
- It creates a cofiguration based on inventory: https://github.com/oxidecomputer/omicron/blob/20299e0b472586014054bc2ebc325e0f4d44b511/sled-agent/src/rack_setup/plan/service.rs#L345-L370
- Sends the configuration to sleds: https://github.com/oxidecomputer/omicron/blob/20299e0b472586014054bc2ebc325e0f4d44b511/sled-agent/src/rack_setup/service.rs#L348-L350
- And tells Nexus about it in the RSS handoff, where the info is stored in CRDB: https://github.com/oxidecomputer/omicron/blob/20299e0b472586014054bc2ebc325e0f4d44b511/nexus/db-queries/src/db/datastore/rack.rs#L739-L760

So, this works in the initialization case.

However, in the case of "sled addition", we're missing the follow-up parts:

- [ ] Nexus needs to assemble the `OmicronPhysicalDisksConfig` that gets sent to sleds
  - [ ] For "new disks": This means scanning the inventory for physical disks, and creating a new control plane object if one does not exist. Note that "expunged disks" still exist, and should not be automatically re-added.
  - [ ] For "existing disks": This means using control plane disks that are not expunged.
- [ ] As a part of blueprint execution, Nexus needs to actually send `OmicronPhysicalDisksConfig` to individual sleds

Regarding new disk addition, there's a bit of uncertainty about "when we should perform this action".
- We could do it once during "sled add" - take a specific point-in-time view of inventory, adopt everything then, and subsequently refuse to auto-online disks
- ... or we could have "auto-online" be the default policy, until the disk is expunged, at which time it must be explicitly re-added

[...]

Issue #5502, opened 2024-04-10
NTP zone failed to start due to U2NotFound error 


askfongjojo:
This happened to a sled that was added to the cluster on rack2.

Steps taken before hitting the error:
1. Mupdated sled 7 (BRM27230045) to the same SP/RoT/OS version as the cluster current runs on.
2. From switch zone, executed `omdb --destructive nexus sleds add BRM27230045 913-0000019`.
3. Confirmed that after step 2, sled 7 was no longer listed in `oxide system hardware sled list-uninitialized` and started to show up in `oxide system hardware sled list`.
4. Regenerate blueprint and set the new one as target (see the bottom of the description for the detailed output)
5. Checked sled-agent log file on sled 7 and saw that the ntp zone was unable to start due to an `U2NotFound` error:
```
00:46:40.868Z WARN SledAgent (ServiceManager): Zone failed to start
    file = sled-agent/src/services.rs:3001
    zone = oxz_ntp_30bfad3a-fab5-4589-a11d-63f1ea076cdb
00:46:40.868Z INFO SledAgent (dropshot (SledAgent)): request completed
    error_message_external = Internal Server Error
    error_message_internal = Failed to initialize zones: [("oxz_ntp_30bfad3a-fab5-4589-a11d-63f1ea076cdb", U2NotFound)]
    file = /home/build/.cargo/git/checkouts/dropshot-a4a923d29dccc492/29ae98d/dropshot/src/server.rs:837
    latency_us = 19705
    local_addr = [fd00:1122:3344:123::1]:12345
    method = PUT
    remote_addr = [fd00:1122:3344:102::3]:44382
    req_id = 085ac5d0-dddf-4833-8356-76c68d20027a
    response_code = 500
    uri = /omicron-zones
```
----
Blueprint commands ran and their output:
```
root@oxz_switch1:~# omdb nexus blueprints list
T ENA ID                                   PARENT TIME_CREATED             
* no  95c3f06b-4dbf-4614-ae7c-507c1193bde9 <none> 2024-03-22T18:37:17.291Z 
```
```
root@oxz_switch1:~# omdb -w nexus blueprints regenerate
generated new blueprint 0ff40c05-188e-4690-ab15-a63d737d550f
```
```
root@oxz_switch1:~# omdb nexus blueprints list
T ENA ID                                   PARENT                               TIME_CREATED             
[...]

PR #5506, opened 2024-04-11
Nexus/Reconfigurator: Automatically adopt disks, deploy them to sleds via blueprints

smklein:
Automatically adopt disks and deploy them to sleds using blueprints

- Queries for physical disk info during reconfigurator planning phase
- Adds "physical disks" to blueprint, in-memory as well as the database schema
- Blueprint planning now ensures that in-service physical disks appear in the blueprint
- Blueprint execution sends a request to sled agents via `omicron-physical-disks PUT`
- A background task has been added to automatically adopt new physical disks as control plane objects, and to insert them into the database
- "Physical disk upsert" has largely been changed to "Physical disk insert", to avoid potential overwriting issues
- The physical disk "vendor/serial/model" uniqueness constraint has been removed for decommissioned disks. This will provide a pathway to eventually re-provisioning deleted disks, if an operator asks for it.

Still needs:
- [ ] Lots of tests.
  - [x] Fix the failing ones?
  - [x] Test background task
  - [ ] Test reconfigurator db changes
  - [x] Test reconfigurator planning
  - [x] Test disk deployment
  - [x] Test on real system?
- [x] Do I need to update how the "diff" tooling works? I updated the output of reconfigurator-cli, hopefully that's enough? 
- [x] This info should probably be plumbed into the "initial blueprint" via RSS (edit: it is now)

Fixes https://github.com/oxidecomputer/omicron/issues/5503 , https://github.com/oxidecomputer/omicron/issues/5502


jgallagher:
I realize there is urgency here and I don't want to create unnecessary work, but given this PR is already large-ish and there's a nontrivial list of things still to add (particularly tests of all the different pieces arriving here), would it make sense to break this into a few smaller PRs for easier reviewing / landing?  Just from reading the PR description it sounds like there might be some natural break points (planning input changes in one PR, database changes in one PR, new background task in one PR, maybe?), but if this isn't as natural as it sounds or is a bunch of work to organize feel free to veto this suggestion.

nexus/db-queries/src/db/datastore/deployment.rs
```diff
@@ -412,6 +554,7 @@ impl DataStore {
         Ok(Blueprint {
             id: blueprint_id,
             blueprint_zones,
+            blueprint_disks,
             parent_blueprint_id,
             internal_dns_version,
```
jgallagher, line 559: github won't let me put a comment on something that isn't in the diff, but `blueprint_delete` also needs to be updated to remove the disks associated with the blueprint-being-deleted.

nexus/db-queries/src/db/datastore/deployment.rs
```diff
@@ -403,6 +495,56 @@ impl DataStore {
...
+                            d.sled_id,
+                        ))
+                        .ok_or_else(|| {
+                            // This error means that we found a row in
+                            // bp_omicron_physical_disk with no associated record in
+                            // bp_sled_omicron_physical_disks.  This should be
+                            // impossible and reflects either a bug or database
+                            // corruption.
+                            Error::internal_error(&format!(
+                                "disk {:?}: unknown sled: {:?}",
```
jgallagher, line 533: Small nit:

```suggestion
                                "disk {}: unknown sled: {}",
```

For `Uuid` this doesn't matter (its display and debug impls are [identical](https://docs.rs/uuid/1.8.0/src/uuid/fmt.rs.html#22-33)), but if/when we move these to `TypedUuid` it does matter (TypedUuid [includes the tag](https://docs.rs/newtype-uuid/1.0.1/src/newtype_uuid/lib.rs.html#155-158) in its debug impl, which I think we generally don't want in logs / error messages).

nexus/db-queries/src/db/datastore/deployment.rs
```diff
@@ -403,6 +495,56 @@ impl DataStore {
...
+                        })?;
+                    let disk = d.into_blueprint_disk_config();
+                    sled_disks.disks.push(disk);
+                }
+            }
+        }
+
+        // Sort all zones to match what blueprint builders do.
+        for (_, disks_config) in blueprint_disks.iter_mut() {
+            disks_config.disks.sort_by(|a, b| a.id.partial_cmp(&b.id).unwrap());
```
jgallagher, line 545: I think this would be equivalent and faster (allow unstable sorting, no unwrap needed):

```suggestion
            disks_config.disks.sort_unstable_by_key(|d| d.id);
```

nexus/db-queries/src/db/datastore/deployment.rs
```diff
@@ -403,6 +495,56 @@ impl DataStore {
...
+                                d.id, d.sled_id
+                            ))
+                        })?;
+                    let disk = d.into_blueprint_disk_config();
+                    sled_disks.disks.push(disk);
+                }
+            }
+        }
+
+        // Sort all zones to match what blueprint builders do.
```
jgallagher, line 543:
```suggestion
        // Sort all disks to match what blueprint builders do.
```

nexus/db-queries/src/db/datastore/physical_disk.rs
```diff
@@ -34,42 +38,30 @@ impl DataStore {
...
-                .do_update()
-                .set((
-                    dsl::sled_id.eq(excluded(dsl::sled_id)),
-                    dsl::time_deleted.eq(Option::<DateTime<Utc>>::None),
-                    dsl::time_modified.eq(now),
-                )),
+            diesel::insert_into(dsl::physical_disk).values(disk.clone()),
         )
         .insert_and_get_result_async(conn)
         .await
```
jgallagher, line 67: Should the `ErrorHandler` in the .map_err after this line use `ErrorHandler::Conflict` now that this is an insert instead of an upsert?

nexus/db-queries/src/db/datastore/physical_disk.rs
```diff
@@ -126,6 +118,82 @@ impl DataStore {
...
+        use db::schema::inv_physical_disk::dsl as inv_physical_disk_dsl;
+        use db::schema::physical_disk::dsl as physical_disk_dsl;
+        use db::schema::sled::dsl as sled_dsl;
+
+        // This is kind of frustrating - we'd like to be able to apply this
+        // condition to all "in service" sleds, regardless of provisionability,
+        // but the sled policy crate explicitly avoids exporting that type.
+        //
+        // As a workaround, we must create the corresponding external types
+        // and manually convert them back into the database types.
```
jgallagher, line 143: I think this is intentionally difficult because we want to steer queries to using `SledFilter` is to avoid exactly this kind of "build up exactly the right policy / state combination" work. I think we can drop both of these, and change the filter to

```rust
.filter(SledFilter::InService)
```

nexus/db-queries/src/db/datastore/physical_disk.rs
```diff
@@ -126,6 +118,82 @@ impl DataStore {
...
+                        .eq(inventory_collection_id)
+                        .and(inv_physical_disk_dsl::sled_id.eq(sled_id))
+                        .and(
+                            inv_physical_disk_dsl::variant
+                                .eq(PhysicalDiskKind::U2),
+                        ),
+                ),
+            )
+            // Filter out any disks in the inventory for which we have ever had
+            // a control plane disk.
```
jgallagher, line 172: This is probably a naive question because I don't grok the disk lifecycle yet, but: does this prevent us from reinitializing a disk that has previously been expunged?

nexus/src/app/background/physical_disk_adoption.rs
```diff
@@ -0,0 +1,110 @@
...
+                    let disk = PhysicalDisk::new(
+                        Uuid::new_v4(),
+                        inv_disk.vendor,
+                        inv_disk.serial,
+                        inv_disk.model,
+                        inv_disk.variant,
+                        inv_disk.sled_id,
+                    );
+
+                    let result = self.datastore.physical_disk_insert(opctx, disk).await;
```
jgallagher, line 90: Just making sure I understand the flow; is this right?

1. A new sled is added
2. The sled-agent on the new sled creates zpools on all its physical disks
3. An inventory collection is taken when includes the new sled, which includes zpools+physical disks from step 2
4. This background task executes, which adopts any zpool+physicaldisk pairs, turning them into control plane `PhysicalDisks` via this call
5. The next time the planner runs, its `PlanningInput` includes the new sled with its zpool+physicaldisk pairs (it already would have gotten the new zpools; getting the physical disk alongside it is new)
6. The planner uses the new disk input to build `blueprint_disks`
7. The executor sends the list of zpool+disk pairs for each sled from `blueprint_disks` to every sled-agent (an idempotent operation)

This gets at the question of autoadoption vs operator-driven adoption - it feels a little backwards to me that the disks are adopted _before_ reconfigurator gets involved, and then execution is only responsible for telling sled-agents about it. I think the feeling is because for external networking resources, reconfigurator chooses whether to assign resources during planning, and then execution is responsible for both (a) record that decision in the database and then (b) tell sled-agent about it, whereas for disks (a) is happening ahead of time. Maybe that's fine or even better in this case? I'm not sure, but just wanted to jot down the feeling.

nexus/db-queries/src/db/datastore/physical_disk.rs
andrewjstone: Yes, I believe that is intentional based on last nights discussion, and possibly RFD 457.

nexus/reconfigurator/execution/src/omicron_physical_disks.rs
```diff
@@ -0,0 +1,265 @@
...
+use uuid::Uuid;
+
+/// Idempotently ensure that the specified Omicron disks are deployed to the
+/// corresponding sleds
+pub(crate) async fn deploy_disks(
+    opctx: &OpContext,
+    sleds_by_id: &BTreeMap<Uuid, Sled>,
+    sled_configs: &BTreeMap<TypedUuid<SledKind>, BlueprintPhysicalDisksConfig>,
+) -> Result<(), Vec<anyhow::Error>> {
+    let errors: Vec<_> = stream::iter(sled_configs)
```
andrewjstone, line 29: I think this is purely sequential. Do we want to instead make deployment concurrent via `FuturesUnordered` or the like? My main concern is multiple timeouts adding up consecutively, rather than amortized to a single timeout.

nexus/reconfigurator/planning/src/blueprint_builder.rs
```diff
@@ -405,6 +438,89 @@ impl<'a> BlueprintBuilder<'a> {
         self.comments.push(String::from(comment));
     }
 
+    /// Ensures that the blueprint contains disks for a sled which already
```
andrewjstone, line 441: This method is super clear and well documented. Thanks!

nexus/reconfigurator/planning/src/blueprint_builder.rs
```diff
@@ -405,6 +438,89 @@ impl<'a> BlueprintBuilder<'a> {
...
+        let disks = &mut self.disks.change_sled_disks(sled_id).disks;
+
+        disks.append(&mut additions);
+        disks.retain(|config| {
+            !removals.contains(
+                &TypedUuid::<PhysicalDiskKind>::from_untyped_uuid(config.id),
+            )
+        });
+
+        Ok(Ensure::Added)
```
andrewjstone, line 521: Nit: maybe not worth changing right now since this is used elsewhere, but `Ensure::Modified` is perhaps a better name.

nexus/reconfigurator/planning/src/blueprint_builder.rs
```diff
@@ -863,6 +981,102 @@ impl<'a> BlueprintZonesBuilder<'a> {
     }
 }
 
+/// Helper for working with sets of disks on each sled
+///
+/// Tracking the set of disks is slightly non-trivial because we need to bump
+/// the per-sled generation number iff the disks are changed.  So we need to
+/// keep track of whether we've changed the disksrelative to the parent
```
andrewjstone, line 988:
```suggestion
/// keep track of whether we've changed the disks relative to the parent
```

nexus/src/app/background/physical_disk_adoption.rs
andrewjstone: I think it makes sense to auto adopt in this case. We need a mechanism to create the physical control plane disks so that reconfigurator can see them. This is somewhat of a stop-gap because we don't have APIs to list raw disks and allow the operator to adopt them. However, as an intermediate step, we could choose to have the physical control plane disks not put in service and then adopted by reconfigurator. I wouldn't worry about that for this PR though.

andrewjstone (commented):
Thanks for jumping on this so quickly @smklein. It's a great start!

smklein:
Also - I'm on board to split this up! But I might do it as a late-stage step taking things out of draft, because it's easier to tell when I'm "green on CI" with a single PR, even if logically I can split things into smaller chunks.

nexus/reconfigurator/execution/src/omicron_physical_disks.rs
jgallagher: I think I'd vote for keeping these sequential on this PR, and have a separate issue for blueprint realization concurrency / optimization. I'd maybe also group "error handling" in that (right now, there are several ways an error can prevent further blueprint realization; I think that's probably overly cautious, but relaxing it might be tricky).

smklein:
Tested this on a4x2, appears to be working! @andrewjstone is following-up on an issue to get inventory to re-trigger more quickly when sled add completes (It *is* triggering right now, but before the sled agent is in the DB, so we don't actually query the new sled for inventory info yet)

andrewjstone (approved):
Ship It! Tested on a4x2 and huzzah!

andrewjstone:
> Tested this on a4x2, appears to be working! @andrewjstone is following-up on an issue to get inventory to re-trigger more quickly when sled add completes (It _is_ triggering right now, but before the sled agent is in the DB, so we don't actually query the new sled for inventory info yet)

Work for this is in https://github.com/oxidecomputer/omicron/pull/5526

smklein:
One last minute change, FYI -- I changed zpool / physical disk "upsert" to "insert" a while ago, because they are now creating control plane objects, rather than "updating a view of info from sleds". That's the responsibility of the inventory now.

Using INSERT here is the right call, **most** of the time, except for one spot: RSS, which wants to "INSERT unless it already exists". Normally, INSERT throws an error if there's a conflict. RSS would want to explicitly keep going. I fixed that here: https://github.com/oxidecomputer/omicron/pull/5506/commits/c985ddcb72f3c1bdb0aec029605770e5db1d207d when I saw another a4x2 run get stuck after trying to re-insert these devices.

PR #5511, opened 2024-04-11
[meta] TypedUuid<SledKind> -> SledUuid, OmicronZoneKind -> OmicronZoneUuid

sunshowers:
Followup from #5501 -- use `SledUuid` and `OmicronZoneUuid` everywhere. I wanted to try it out to see how it feels, and I think I like it a lot.

Depends on #5501.

### DIFF
--- a/nexus/db-queries/src/db/datastore/deployment.rs
+++ b/nexus/db-queries/src/db/datastore/deployment.rs
@@ -1092,7 +1092,8 @@ mod tests {
     use omicron_common::api::external::Generation;
     use omicron_test_utils::dev;
     use omicron_uuid_kinds::GenericUuid;
-    use omicron_uuid_kinds::TypedUuid;
+    use omicron_uuid_kinds::SledUuid;
+    use omicron_uuid_kinds::ZpoolUuid;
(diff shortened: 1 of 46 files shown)


nexus/db-queries/src/db/datastore/deployment.rs
```diff
@@ -1202,7 +1202,7 @@ mod tests {
             );
             for (sled_id, agent) in &collection.sled_agents {
                 // TODO-cleanup use `TypedUuid` everywhere
-                let sled_id = TypedUuid::from_untyped_uuid(*sled_id);
+                let sled_id = SledUuid::from_untyped_uuid(*sled_id);
```
sunshowers, line 1205: Particularly like that this is now explicit in a way that's still convenient, and much nicer than `TypedUuid::<SledKind>::from_untyped_uuid`.

jgallagher (approved):
Yeah, this is great. If I was at all on the fence still, I'm not anymore

sunshowers:
Accidentally landed this on the wrong branch -- put up https://github.com/oxidecomputer/omicron/pull/5513 which targets main.

PR #5512, opened 2024-04-11
Periodically renew oximeter producer leases

bnaecker:
- Changes Nexus metric producer API. This used to return a 204, and now returns a 201 along with a new lease duration for an oximeter producer.
- Reworks the producer server. The API endpoints it exposes were previously available as public functions. The goal was to support using those endpoints in an existing dropshot server, rather than a special-purpose one just for metric production. These endpoints are now private, so all producers use the built-in server. The point of this is to centralize the logic for renewing a producer's lease periodically from Nexus. The server now spawns a task for doing that internally, so consumers only need to make the server and register their own metric producer types (which they already did). One downside of this is that the `base_route` field is no longer needed. That allowed placing the collection endpoint anywhere in the Dropshot server's router. Since the endpoint must now be served by the `oximeter_producer::Server`, that's extraneous. New clients will have that value ignored, and set to the expected one internally. We'll remove the field and its database column in the future, once all clients are on the new version where it does not matter.
- Updates the Nexus and Sled Agent metric production to use the new server. They previously had the needed endpoints bolted onto their existing Dropshot servers. The upshot of this is that neither needs to bother with explicitly registering themselves with Nexus -- the producer server does that internally, both initially and when renewing the lease.
- Fixup integration tests. These no longer need to explicitly ask Nexus to register as a metric producer in each test, since that's done when a Nexus object is created.


bnaecker:
This is the last bit of #5284. It internalizes a lot of the complexity around producer registration inside the `oximeter_producer::Server` type that most code uses to register with Nexus and serve data to the `oximeter` collector. Previously, creating a server would block until it registered with Nexus, which meant most code needed to run that in a background task of some kind, and store the server in something like an `Arc<Mutex<Option<Server>>>` so that it can be replaced upon successful registration.

In this PR, the server itself spawns a task to register with Nexus, and to periodically _renew_ that registration. Nexus's registration API now returns the lease duration, and the task waits for (a fraction of) that period before renewing the lease.

oximeter/producer/src/lib.rs
```diff
@@ -64,56 +85,64 @@ pub struct Config {
...
 
     /// Serve requests for metrics.
     pub async fn serve_forever(self) -> Result<(), Error> {
+        self.registration_task.abort();
         self.server.await.map_err(Error::Server)
     }
 
     /// Close the server
     pub async fn close(self) -> Result<(), Error> {
+        self.registration_task.abort();
```
jgallagher, line 145: Should we also abort this task in `Drop`? This isn't a leading question - it would only matter if someone drops a server without calling `.close()`, which is a little sketchy, so I'm not sure we should.

oximeter/producer/src/lib.rs
```diff
@@ -208,57 +291,123 @@ impl Server {
...
+enum FindNexus {
+    /// An explicit address was provided at creation time, just use it.
+    ByAddr(SocketAddr),
+    /// An address was not provided, we'll resolve it on each attempt to renew
+    /// the lease.
+    WithResolver(Resolver),
+}
+
+/// The default period within which we renew our lease from Nexus.
+const DEFAULT_LEASE_DURATION: Duration = Duration::from_secs(60 * 2);
```
jgallagher, line 309: Do we need a default lease duration? This PR also changes Nexus to always return the lease duration, right?

oximeter/producer/src/lib.rs
```diff
@@ -274,21 +423,152 @@ pub async fn register(
...
+            _ => BackoffError::permanent(e.to_string()),
+        })
+    };
+    backoff::retry_notify(
+        backoff::retry_policy_internal_service(),
+        do_register,
+        log_failure,
+    )
+    .await
+    .expect("Expected infinite retry loop registering as a producer with")
```
jgallagher, line 490: I don't think we can `.expect` here - the two `BackoffError::permanent` points above mean it _isn't_ an infinite retry loop. I'm not sure what we're supposed to do with a permanent registration error, though?

oximeter/producer/src/lib.rs
```diff
@@ -274,21 +423,152 @@ pub async fn register(
...
+                bind_address: address,
+                ..Default::default()
+            },
+            log: LogConfig::Logger(log),
+        };
+
+        // Ideally, we would check pretty carefully that there are exactly N
+        // registrations after N renewal periods. That's brittle, especially on
+        // a loaded system. Instead, we'll run for a number of intervals, and
+        // just check that we have more than one.
```
jgallagher, line 563: Should the `.times(1..)` above be `.times(2..)` if we want to check "more than one"?

nexus/src/lib.rs
```diff
@@ -417,6 +418,46 @@ pub async fn run_server(config: &NexusConfig) -> Result<(), String> {
...
+            address,
+            base_route: String::from("/metrics/collect"),
+            interval: std::time::Duration::from_secs(10),
+        },
+        // Some(_) here prevents DNS resolution, using our own address to
+        // register.
+        registration_address: Some(nexus_addr),
+        dropshot: dropshot::ConfigDropshot {
+            bind_address: address,
+            request_body_max_bytes: 1024 * 1024 * 10,
```
jgallagher, line 452: I see sled-agent also uses this value. Is this something producer-specific, or does every producer want the same value? If the latter, should `oximeter_producer::Config` maybe only take `bind_address` and then construct the rest of the `ConfigDropshot` internally? I think `default_handler_task_mode` should also be a decision of `oximeter_producer`, not its users, right?

nexus/src/lib.rs
```diff
@@ -417,6 +418,46 @@ pub async fn run_server(config: &NexusConfig) -> Result<(), String> {
...
+
+/// Create a new metric producer server.
+fn start_producer_server(
+    log: &Logger,
+    registry: &ProducerRegistry,
+    nexus_addr: SocketAddr,
+) -> Result<ProducerServer, String> {
+    // The producer server should listen on any available port, using the
+    // same IP as the main Dropshot server.
+    let address = SocketAddr::new(nexus_addr.ip(), 0);
```
jgallagher, line 432: I don't object to this, but just making sure I understand - this is a fairly significant change to producers, right? Now every time the producer restarts it will listen on a different port, so _every_ producer is now effectively ephemeral and dependent on the garbage collection task to clean up past instances of itself?

Should this PR also remove the `disabled` bit from the gc task? Or do the producers outside of `omicron` need to be updated first to reregister themselves?

jgallagher (commented):
Sorry for the delay on the review here! Thanks for taking all of this on.

oximeter/producer/src/lib.rs
```diff
@@ -64,56 +85,64 @@ pub struct Config {
...
+            config.registration_address.as_ref(),
             &config.dropshot,
             &config.log,
         )
-        .await
     }
 
     /// Serve requests for metrics.
     pub async fn serve_forever(self) -> Result<(), Error> {
+        self.registration_task.abort();
```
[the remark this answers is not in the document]
bnaecker: I think that's right, and not what I intended! I think that slipped through because we don't actually need to call `serve_forever()` to start serving requests. That happens when we start the Dropshot server itself. In any case, I'll remove this, thanks for catching.

oximeter/producer/src/lib.rs
```diff
@@ -198,8 +245,44 @@ impl Server {
...
+                );
+                FindNexus::ByAddr(*addr)
+            }
+            None => {
+                debug!(
+                    log,
+                    "Nexus IP not provided, will use DNS to resolve it"
+                );
+                let IpAddr::V6(our_addr) = server_info.address.ip() else {
+                    unreachable!();
```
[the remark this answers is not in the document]
bnaecker: Sure, that seems reasonable, will do.

oximeter/producer/src/lib.rs
bnaecker: I don't know, but it does seem like a reasonable idea. Yes, it's odd to drop the server without calling close. But if that happens, we probably want the registration task to bail as well, since the server will never be reachable anyway.

bnaecker: Hmm, ok I think we actually _cannot_ implement `Drop` for this type. The issue is that the Dropshot server's `close()` method takes `self`. If we implement `Drop`, then we can't move out of the fields of the `Server`, such as when we call `self.server.close()`. This is [E0509](https://doc.rust-lang.org/error_codes/E0509.html?highlight=e0509#).

oximeter/producer/src/lib.rs
bnaecker: It does, yeah. I was trying to handle the case where we're registering with an older Nexus. That might have been possible during development, but it might be impossible now?

nexus/src/lib.rs
bnaecker: The requests size depends on the data being produced, so I think there still needs to be a knob for it here. The producer server should probably control the handler task mode, though. We could either move the bind address and request size to the `oximeter_producer::Config` directly; or just overwrite the task mode internally? Neither seems awesome.

nexus/src/lib.rs
bnaecker: > I don't object to this, but just making sure I understand - this is a fairly significant change to producers, right? Now every time the producer restarts it will listen on a different port

This part is correct. Some producers, like Nexus, previously attached their metric endpoints to their "main" Dropshot server. One of the big changes in this PR is that that's not supported anymore. You _have_ to run a new Dropshot server. The point of that is that it provides a central place to run the periodic registration task, inside the `oximeter_producer::Server`.

> so every producer is now effectively ephemeral and dependent on the garbage collection task to clean up past instances of itself?

I don't think this part follows. Suppose a producer restarts, so that it is now running the Dropshot server on a new port. The next time it registers with Nexus, that will _update_ the database record for that producer. There is no past instance of itself, there's only one record. (This is assuming the producer is restarted with the same UUID! That's true for Nexus, and if it wasn't true before, then this was always an issue. This PR doesn't make it worse, unless I'm missing something.)

> Should this PR also remove the disabled bit from the gc task? Or do the producers outside of omicron need to be updated first to reregister themselves?

I think we need to update all the out-of-tree consumers to use this new producer server first. Once that's rolled out, we can turn on the GC task.

oximeter/producer/src/lib.rs
bnaecker: Good point, this is not an infinite loop any more. And I'm also not sure what to do with a permanent error. I mean, what _can_ we do? I think the only things are (1) panicking, not great; or (2) logging it, and stopping the registration loop. If we do that, is it worth adding methods for fetching the registration state, such as the result of the last registration (and timestamp)? I don't know how we'd use it, but adding a way to kick the registration task to try again might be useful too.

oximeter/producer/src/lib.rs
bnaecker: Done in 2c0afef2ebf1321aeca50dffd86df08894436b39

oximeter/producer/src/lib.rs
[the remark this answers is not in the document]
bnaecker: Done in 2c0afef2e

oximeter/producer/src/lib.rs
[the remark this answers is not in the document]
bnaecker: Moved in 2c0afef2e

oximeter/producer/src/lib.rs
jgallagher: I think it's impossible since the Nexus change landed first, yeah. The way we do updates it's not possible to have a newer service talk to a Nexus any older than the one that existed when it was built.

nexus/src/lib.rs
jgallagher: > I don't think this part follows. Suppose a producer restarts, so that it is now running the Dropshot server on a new port. The next time it registers with Nexus, that will _update_ the database record for that producer. There is no past instance of itself, there's only one record. (This is assuming the producer is restarted with the same UUID! That's true for Nexus, and if it wasn't true before, then this was always an issue. This PR doesn't make it worse, unless I'm missing something.)

Ah hah, yes, I was forgetting/missing that it restarts with the same ID. Thanks.

> I think we need to update all the out-of-tree consumers to use this new producer server first. Once that's rolled out, we can turn on the GC task.

:+1:

oximeter/producer/src/lib.rs
jgallagher: I think the two `_ =>`s in this call chain make me a little nervous (one on line 481, and one on line 418). Even if all the variants covered by `_` today really are permanent errors, will that be true for new error variants added in the future?

I'm moderately tempted to suggest we remove the permanent errors altogether here and make all errors transient. E.g., if we get some response from Nexus that indicates a permanent error, but it came from a Nexus that's borked, could we succeed if we resolve to a different Nexus?

oximeter/producer/src/lib.rs
bnaecker: Ok, I think that makes sense. I'll remove the const and propagate an error in the one place it's currently used. Thanks.

oximeter/producer/src/lib.rs
bnaecker: That's a good question. I wrote it this way a long time ago to avoid retrying operations that could never succeed, e.g., making the same bad request to Nexus forever. I get your point though, that the "same request" may actually succeed in the future, e.g., when we support online updates of Nexus or blue-green deployments in which we resolve to a new Nexus that _can_ process the request for us.

It does make me a bit nervous to continually try bad requests. I'm worried about the availability hit we might take, really. But I agree that there's nothing obviously better to do in this case, and backoff should protect us from the worst of that.

nexus/src/lib.rs
bnaecker: Sounds good, I'll move the configuration parameters around. Thanks!

oximeter/producer/src/lib.rs
bnaecker: Ok, this is done in 1413ce8ed. I've also removed the `Error::RegistrationError` variant entirely, since it was only used internally, to communicate if we should retry a registration request. We always do that now, so no need for the variant at all.

oximeter/producer/src/lib.rs
bnaecker: Done in 1413ce8ed

bnaecker:
@jgallagher Thanks for your input. I've addressed everything, but let me know if you have other questions or concerns.

nexus/src/lib.rs
```diff
@@ -417,6 +418,42 @@ pub async fn run_server(config: &NexusConfig) -> Result<(), String> {
...
+    //
+    // Note that because we're registering with _ourselves_, the listening
+    // address for the producer server and the registration address use the
+    // same IP.
+    let config = oximeter_producer::Config {
+        server_info: ProducerEndpoint {
+            id: registry.producer_id(),
+            kind: ProducerKind::Service,
+            address,
+            base_route: String::from("/metrics/collect"),
```
jgallagher, line 444: Just double-checking - `base_route` is irrelevant now right?

jgallagher (approved):
Looks good to me! :+1:

nexus/src/lib.rs
jgallagher: Yeah, it jumped out at me because the rest of the ones in the diff have become an empty string with a comment about it being unused.

bnaecker: Cool, I'll change this one too, thanks for catching.

bnaecker:
I'm rebasing to address the new conflicts with `main`, then will automerge. Thanks for the review @jgallagher!

PR #5518, opened 2024-04-12
[internal-dns] move sled and zone IDs to TypedUuid

sunshowers:
I wanted to understand how internal DNS works wrt the reconfigurator, so I
tried to port it over to using TypedUuid to basically get an overview via
compile errors. Here's the result.

I realized that we actually were overloading the `Host::Zone`'s ID to mean
either the sled ID (for dendrite) or the zone ID (for the `Other` variant).
I've addressed that in the PR by making the variant also carry the ID. (This
was always an expected outcome of the transition to typed UUIDs -- if we really
are storing more than one kind of UUID within a field, we should make that
clear by specifying our own enum variants.)

Probably depends on #5517 due to shared imports.


jgallagher (approved):
LGTM. The test failure looks legit but trivial to fix. Up to you if you want to wait for Dave to take a look when he gets back.

sunshowers:
Fixed the issues, and in the process also ended up moving sled IDs to newtype-uuid (I started on that process and the two got kinda mixed up.)

PR #5523, opened 2024-04-12
Separate `BlueprintZoneConfig` and `OmicronZoneConfig`

jgallagher:
This PR should make no changes to behavior and is solely a beefy refactor (although the bulk of the diff volume is just in the openapi spec; the actual changes are closer to +1000,-500). The motivating case here is that `OmicronZoneConfig` (and particularly `OmicronZoneType`) does not have all the information Reconfigurator needs to clean up expunged zones; in particular, we need the CRDB ID of external IP addresses to delete them, and `OmicronZoneType` only stores the raw IP address. In this case we could clean this up by querying by IP to get the ID and then delete the ID, but the plan has always been that these types would diverge over time. We already passed that line when we added `disposition`, but we didn't really separate the types at that point.

This ended up a little messier than I expected, because there are several places that want to construct a `BlueprintZoneConfig` from an `OmicronZoneConfig`+disposition. The "+ disposition" will grow to "+ disposition + external IP ID" soon, which may make this conversion unwieldy enough to consider doing something else in all those places? I'm not sure about this and plan to cross that bridge over the next day or two.


nexus/reconfigurator/execution/src/datasets.rs
```diff
@@ -45,14 +45,17 @@ pub(crate) async fn ensure_crucible_dataset_records_exist(
         .await
         .context("failed to list all datasets")?
         .into_iter()
-        .map(|dataset| dataset.id())
+        .map(|dataset| OmicronZoneUuid::from_untyped_uuid(dataset.id()))
```
sunshowers, line 48: https://github.com/oxidecomputer/omicron/pull/5532 should take care of this

sled-agent/src/rack_setup/service.rs
```diff
@@ -1329,6 +1329,31 @@ pub(crate) fn build_initial_blueprint_from_sled_configs(
...
+                kind,
+                addr,
+                err,
+            }) => {
+                // This error is impossible, because our `z.into()` converts all
+                // `SocketAddr`s into strings, which are all therefore valid.
+                unreachable!(
+                    "RSS produced {kind} zone with invalid address \
+                    {addr}: {err}"
+                )
```
sunshowers, line 1352: Can this return an error instead? I'm worried that some other caller will come about and call this with untrusted data -- instead, the caller which knows it is trusted can panic if this errors out.

sunshowers (approved):
Looks great!

sled-agent/src/rack_setup/service.rs
jgallagher: Yeah, changed in [5b290f1](https://github.com/oxidecomputer/omicron/pull/5523/commits/5b290f12b8236f6dcd8fb61660a80845fd6eaff4). This was simpler than I expected because there are already other ways RSS could construct an invalid plan (and those return errors already); I think I was a little tunnel visioned on this progenitor socketaddr format annoyance.

PR #5524, opened 2024-04-12
[nix flake] Updates and maintainance

hawkw:
This commit updates the Nix flake so that it builds with the current dependencies on `mgd` and Clickhouse. The Clickhouse version we download has changed, so the SHA256 hash in the Nix flake must be updated (this can't be depended on automatically from the fs, currently, because we store its hash as MD5 in the `clickhouse_version` file, rather than SHA256). `mgd` now depends on OpenSSL, so I've changed the build inputs for its derivation to include OpenSSL headers for `autoPatchelfHook`.

I think I might be the only user of the Nix flake, so I don't know if it really makes sense to ask anyone else to review this. Rest assured that it does, in fact, fix it. :)

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -224,6 +224,7 @@
             buildInputs = [
               glibc
               gcc-unwrapped
+              openssl.dev
             ];
 
             installPhase =
@@ -259,7 +260,7 @@
           # can't give Nix those hashes and must instead determine it ourselves.
           # this means that we will have to update this SHA if the clickhouse
           # version changes.
-          sha256 = "1lgxwh67apgl386ilpg0iy5xkyz12q4lgnz08zswjbxv88ra0qxj";
+          sha256 = "0wx8w9sdms5hsc9f835ivsissf15wjzdb9cvxr65xdi384i9pkzx";
           src = builtins.fetchurl
             {
               inherit sha256;
@@ -428,10 +429,3 @@
           };
     };
 }
-
-
-
-
-
-
-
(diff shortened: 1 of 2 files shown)


sunshowers (approved):
Looks reasonable enough!

flake.nix
```diff
@@ -259,7 +260,7 @@
           # can't give Nix those hashes and must instead determine it ourselves.
           # this means that we will have to update this SHA if the clickhouse
           # version changes.
-          sha256 = "1lgxwh67apgl386ilpg0iy5xkyz12q4lgnz08zswjbxv88ra0qxj";
+          sha256 = "0wx8w9sdms5hsc9f835ivsissf15wjzdb9cvxr65xdi384i9pkzx";
```
smklein, line 263: lowkey would love to switch these from md5 to sha2 instead, so we can drop our md5 dependency. But I realize that's not just a "nix flake" problem, that's like, a "fix how we're building our third party software" problem

hawkw: I would also really like that --- then, the Nix flake would require substantially less manual intervention when we update Clickhouse. For the deps that we have SHA2 hashes for, the Nix stuff should "just work" when the hash files change (modulo cases where other library deps appear, like `mgd` adding an `openssl-dev` dep).

faithanalog (approved):

hawkw:
Huh, looks like the `helios-deploy` CI job [ran for about a day and then failed](https://github.com/oxidecomputer/omicron/runs/23772793830). I can't think of any reason why this PR would have broken that, as it only touches Nix stuff and shouldn't affect anything outside of my Nix dev environment. I guess I'll try restarting it.

PR #5066, opened 2024-02-14
Trigger inventory collection on sled add

andrewjstone:
This is based on discussion in https://github.com/oxidecomputer/omicron/issues/5058

I'm still testing it.

PR #5526, opened 2024-04-12
Collect inventory when a new sled-agent comes online

andrewjstone:
Previously, we collected inventory when a sled was added through the nexus external API. However, this is too early to gather the physical disks in inventory for the sled, since those require a sled in the `sled` table for each corresponding disk, before they can be added as a control plane physical disk. This specific disk related change is part of https://github.com/oxidecomputer/omicron/pull/5506 and without a proper trigger on sled addition, we'd have to wait another 10 minutes for inventory collection in order to see the disks show up. An alternate workaround would be  to manually generate an "extra" blueprint, which also triggers inventory collection, although this would also have to be after the sled-agent registered itself via the nexus internal API and got entered into the `sled` table.

Since sled-agent-put is idempotent, and inventory collection is somewhat expensive, we only trigger collection if the sled actually gets updated in the db. We do this via a check on the modified timestamp, which is somewhat hacky but serves our purpose well.

H/T to @davepacheco for being [prescient](https://github.com/oxidecomputer/omicron/pull/5066#pullrequestreview-1881455062).


andrewjstone:
I'm going to test this out on the a4x2 testbed now.

nexus/db-queries/src/db/datastore/sled.rs
```diff
@@ -82,7 +86,10 @@ impl DataStore {
                         &sled_update.id().to_string(),
                     ),
                 )
-            })
+            })?;
+
+        let was_modified = now == *sled.time_modified();
```
smklein, line 91: Did you check that this exact equality check actually works? I thought Cockroachdb truncates timestamps, so even if we did update this sled, it would be losing precision, and thus not appear equal

(IMO we should prefer using inequality operators with time, if we can frame things that way)

andrewjstone: I am launching a testbed now to see if the collection fires

andrewjstone: I got a clippy error, so I'll also rework this then test. I knew we talked about this, but wanted to see what would happen. I thought there was a special DB type for these comparisons we used. I'll look into it.

nexus/db-queries/src/db/datastore/sled.rs
```diff
@@ -43,22 +43,26 @@ use uuid::Uuid;
...
+        let insertable_sled = sled_update.clone().into_insertable();
+        let now = insertable_sled.time_modified().clone();
+
+        let sled = diesel::insert_into(dsl::sled)
+            .values(insertable_sled)
             .on_conflict(dsl::id)
             .do_update()
             .set((
-                dsl::time_modified.eq(Utc::now()),
                 dsl::ip.eq(sled_update.ip),
```
smklein, line 66: One other realization - are we actually updating `time_modified` in the on-conflict case? Looks like no?

nexus/db-model/src/sled.rs
```diff
@@ -46,6 +46,7 @@ pub struct Sled {
     #[diesel(embed)]
     identity: SledIdentity,
```
smklein, line 47: The "identity" struct already contains a `time_modified` field, fwiw. 

That's why this is currently in the DB, otherwise it would require a schema change: https://github.com/oxidecomputer/omicron/blob/00971809e1326a35f41206413ae87bd0db0d338c/schema/crdb/dbinit.sql#L113-L119

nexus/db-queries/src/db/datastore/sled.rs
smklein: I realized I commented this while looking at the right hand side, but it is here on the left hand side.

Why are we no longer updating time modified here? Doesn't that mean that even if we do the update, `time_modified` won't change?

andrewjstone: Oh, whoops. I totally biffed that. I was thinking everything in the` insertable_sled` would go in. I'll add that back. I'm actually going to revert most of these changes, since as you point out below the SledIdentity already contains the modified value. That means none of the tests need to change.

nexus/db-model/src/sled.rs
andrewjstone: I had no idea that was a thing. A little bit too much magic for my personal taste.

nexus/db-queries/src/db/datastore/sled.rs
andrewjstone: Oy, I'm tired. Tests still need to change. Anyway, have a fix. Thanks for pointing all this out.

nexus/db-model/src/sled.rs
andrewjstone: Changed in [9c6efef](https://github.com/oxidecomputer/omicron/pull/5526/commits/9c6efef90c4c01fc3a74d27dbeb7ae8f13517c0b) to use the `identity` field. I removed the redundant `time_modified`.

nexus/db-queries/src/db/datastore/sled.rs
andrewjstone: > Why are we no longer updating time modified here? Doesn't that mean that even if we do the update, time_modified won't change?

When I added the redundant `time_modified` field to sled, I assumed that would take precedence. It probably would have just been completely ignored. Thanks for pointing out `identity`. I reverted this behavior in [9c6efef](https://github.com/oxidecomputer/omicron/pull/5526/commits/9c6efef90c4c01fc3a74d27dbeb7ae8f13517c0b)
except the `now` value is reused so we don't have different value on the conflict case.

nexus/db-queries/src/db/datastore/sled.rs
andrewjstone: I reworked this to compare seconds since the epoch instead. About to test.

andrewjstone:
I tested this on the a4x2 testbed by adding sled `g2` and verifying that a new inventory collection was taken immediately after. The duel terminal screenshot below shows this in action.

Screenshot 2024-04-12 at 11 17 42 PM

nexus/src/app/background/inventory_collection.rs
```diff
@@ -224,18 +224,28 @@ mod test {
...
         // can assume the rest of the collections came from the instance that
         // we're testing.
+        //
+        // This test setup also contains 2 sled-agents. Each of those
+        // sled-agents will upsert themselves via nexus, which will trigger
+        // another collection.
+        //
+        // Therefore we would expect to have to wait for a total of 3
+        // collections to appear before we run manual activations.
+        // TODO: Why then are we only seeing two collections?
```
andrewjstone, line 238: I am stumped here. I saw that both upserts were triggering collections. But I can also see that when nexus gets a handoff it activates the collector again. That should be 3 collections and not 2.

jgallagher: I think I understand this. On this branch we see 5 activations of the inventory collection bg task.

The first activation is when the task is created; tokio timeouts always fire immediately. This always fails pre-rack-setup with `Forbidden` (I think because the database hasn't been populated with roles, etc., needed for queries to work?), but I think trying to activate immediately is still correct because on a Nexus start with a populated DB this ought to work.

The second and third activations are racing each other. One is the activation Nexus fires [after it populates the built-in data](https://github.com/oxidecomputer/omicron/blob/6ed661734385d90f388e66c51400165caadce4cb/nexus/src/app/mod.rs#L466-L494); this is `tokio::spawn`'d as a part of [starting the internal Nexus server](https://github.com/oxidecomputer/omicron/blob/6ed661734385d90f388e66c51400165caadce4cb/nexus/test-utils/src/lib.rs#L1338-L1341); I strongly suspect this one basically always wins the race, but since it's `spawn`'d it technically is racing. The other is from the first `sled_upsert` added on this PR.

The fourth activation is the second `sled_upsert` added on this PR.

The fifth activation is the collection triggered after the RSS handoff.

When I run this branch, I see the first three collections fail: the first fails with `Forbidden` (as expected), and the second and third both fail with a DNS resolution error:

```
14:01:01.187Z WARN ca2462e3-3f7e-454f-9c45-2afce70a6863 (ServerContext): inventory collection failed
    background_task = inventory_collection
    error = failed to collect inventory: looking up MGS addresses: no record found for Query { name: Name("_mgs._tcp.control-plane.oxide.internal."), query_type: SRV, query_class: IN }
```

The fourth and fifth activations succeed, matching the 2 you're seeing here.

However, I'll claim that activations 2, 3, and 4 are all racing against internal DNS being populated. To test this, I added a 10 second sleep just before `nexus-test-utils` [populates internal DNS](https://github.com/oxidecomputer/omicron/blob/6ed661734385d90f388e66c51400165caadce4cb/nexus/test-utils/src/lib.rs#L773); with that sleep, activations 2, 3, and 4 _all_ fail with DNS resolution errors, and we only see 1 collection at this point instead of 2 or 3. The final collection triggered by RSS handoff always succeeds, because nexus-test-utils doesn't kick that off until after populating internal DNS (and other things), so even with my artificial sleep this one is fine.

We could make this deterministic by reordering the init steps in the test context; e.g., by moving [`populate_internal_dns`](https://github.com/oxidecomputer/omicron/blob/6ed661734385d90f388e66c51400165caadce4cb/nexus/test-utils/src/lib.rs#L1362-L1365) up before [`start_sled1`](https://github.com/oxidecomputer/omicron/blob/6ed661734385d90f388e66c51400165caadce4cb/nexus/test-utils/src/lib.rs#L1342-L1349). I'm not sure if that's reasonable or not. In real RSS, if a sled is able to _find_ Nexus to register itself, that means DNS is up and the triggered inventory collection should work. But other parts of sleds coming online have to work even before internal DNS is up, since part of "sleds coming online" is bringing up internal DNS.

I think I'd be inclined to actually leave this test the way it was, and maybe test that sled_upsert triggers collections separately (e.g., via the add sled flow?).

andrewjstone: John and I talked a bunch about this in chat and agree that there will pretty much always be a race. John implemented a patch that tolerates races in this test that I just pushed up.

nexus/db-queries/src/db/datastore/sled.rs
```diff
@@ -82,7 +87,12 @@ impl DataStore {
                         &sled_update.id().to_string(),
                     ),
                 )
-            })
+            })?;
+
+        // We compare only seconds since the epoch, because writing to and
+        // reading from the database causes us to lose precision.
+        let was_modified = now.timestamp() == sled.time_modified().timestamp();
+        Ok((sled, was_modified))
```
davepacheco, line 95: This seems probably alright but not great.  This would produce false positives if the value was recently modified and we _didn't_ change it.  That's not a problem for the current caller -- it'll just trigger an extra inventory collection, which isn't that big a deal.  I'm more worried someone will look at the docs and rely on the boolean being correct for correctness (whereas right now it's just an optimization).

I _think_ in this PR there's only one caller of this function that needs this behavior, but a lot of the delta here is around changing the existing callers to ignore this second boolean.

So what do you think of this: don't change this function at all, and have that one caller check the modification time on the returned sled?  That way the limitation is specific to that one caller and the general code is not providing a boolean that might not be accurate.  (edit: oh, all callers would need to do the `sled_update.into_insertable()` themselves to do this.  That still seems better to me but definitely more debatable.)

Also I think the call that triggers this function only happens when a sled agent starts up, right?  It doesn't seem that bad to just always trigger a collection in that case.

davepacheco (approved):

nexus/db-queries/src/db/datastore/sled.rs
jgallagher: I had a couple similar comments staged but I forgot to hit "submit": one complained that this was icky and wondered if we could get an accurate answer from the database instead of having to check `time_modified` as a proxy, and the other was a request to not use `bool` because it's not obvious to me what it means at a call site like this:

```rust
let (_, was_upserted) = sled_upsert(...)?;
if was_upserted {
    // was the sled updated or inserted?
}
```

Your suggestion would address both of my concerns.

andrewjstone: I hate the bool in a tuple return value, but I also didn't want to force all callers to call `into_insertable`. That seems to make the `SledUpdate` type somewhat of a weird intermediate state, and no longer part of the API. We only turn it into an insertable to use the magic id stuff and grab the mod time, etc.... I also find that whole [macro transformation](https://github.com/oxidecomputer/omicron/blob/main/nexus/db-macros/src/lib.rs#L182) to be way too much magic for my liking, but changing that is outside the scope of this PR.

The other alternative I thought of when writing this code was to pass in an optional timestamp and set it in `sled_update`, but that seems nearly equivalent to the bool, except now it can't be ignored by the caller. They have to pass in a `None`, which I find more distasteful than returning a bool. 

All this is how I ended up where I did, and a other than changing the return value to a struct or adding a newtype wrapper around the bool with a doc comment, I don't think there's a better way forward.

andrewjstone: In light of my counter arguments, and moving onto more critical things, I'm going to keep this as is. We can always revisit.

PR #5532, opened 2024-04-15
[nexus] move dataset IDs to TypedUuid

sunshowers:
Datasets are zones, so they're `OmicronZoneUuid` instances. Was quite
straightforward in the end.

There is some messiness around Diesel, and honestly I'm half-tempted to just
fork and publish our own version of Diesel, with support for typed UUIDs +
other things we've run into.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4827,6 +4827,7 @@ dependencies = [
  "nexus-config",
  "nexus-types",
  "omicron-common",
+ "omicron-uuid-kinds",
  "omicron-workspace-hack",
  "slog",
  "uuid 1.8.0",
--- a/clients/nexus-client/src/lib.rs
+++ b/clients/nexus-client/src/lib.rs
@@ -37,6 +37,7 @@ progenitor::generate_api!(
         NetworkInterfaceKind = omicron_common::api::internal::shared::NetworkInterfaceKind,
         TypedUuidForCollectionKind = omicron_uuid_kinds::CollectionUuid,
         TypedUuidForDownstairsKind = omicron_uuid_kinds::TypedUuid<omicron_uuid_kinds::DownstairsKind>,
+        TypedUuidForOmicronZoneKind = omicron_uuid_kinds::OmicronZoneUuid,
         TypedUuidForUpstairsKind = omicron_uuid_kinds::TypedUuid<omicron_uuid_kinds::UpstairsKind>,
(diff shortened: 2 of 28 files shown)


jgallagher:
> Datasets are zones, so they're `OmicronZoneUuid` instances.

Is this true? I thought a zone could have 0 or more datasets. Today it looks like it's "0 or exactly 1", and if 1, the dataset ID does match the zone ID, but I'm not sure it will always be that way?

sunshowers:
Oh interesting, I didn't realize that. Thoughts on how this should be represented then? I feel like we should use an enum that covers both the N = 1 and N = many cases.

davepacheco:
Yeah, I believe that right now, there is at most one dataset associated with each zone.  And you know based on the zone type whether it has one.  In principle we could have more datasets later, but that's pretty hypothetical and I think it's a pretty useful simplification for the time being that the zone id == dataset id.  That's especially useful when debugging.  I'd be inclined _not_ to frontload the work now to generalize this for something that we don't know whether/when/why we'll do later.

I also think it makes sense to use a separate type like `DatasetUuid`.  It's a separate id that happens to always have the same value.  Is this self-inconsistent?

Forking diesel sounds pretty extreme.  There's a whole ecosystem of crates related to it and I think we also want to be able to take new updates to it pretty easily.

sunshowers:
> I also think it makes sense to use a separate type like DatasetUuid. It's a separate id that happens to always have the same value. Is this self-inconsistent?

I'm okay with that, I think.

> Forking diesel sounds pretty extreme. There's a whole ecosystem of crates related to it and I think we also want to be able to take new updates to it pretty easily.

Oh I was imagining just a patchset on top that we'd use a `[patch]` directive to pull in, we wouldn't lose any of the ecosystem. Then we could have our patched Diesel implement support for newtype-uuid just like it implements support for uuid today. But this isn't a fleshed-out idea. The Rust orphan rules cause this to really suck :(

sunshowers:
> Oh I was imagining just a patchset on top that we'd use a `[patch]` directive to pull in, we wouldn't lose any of the ecosystem. Then we could have our patched Diesel implement support for newtype-uuid just like it implements support for uuid today. But this isn't a fleshed-out idea. The Rust orphan rules cause this to really suck :(

Made a post on the Diesel discussion board about this, https://github.com/diesel-rs/diesel/discussions/3993 -- hoping to have Diesel take in optional support for newtype-uuid, which will solve this problem.

sunshowers:
We'll have to do this from scratch -- going to abandon this and come back to it when I have more time.

PR #5540, opened 2024-04-16
[nexus-types] make PlanningInput's ExternalIp only support a single IP

sunshowers:
This simplifies bookkeeping within the blueprint builder considerably --
services always get a single IP, and I don't think that's going to change any
time soon.


nexus/types/src/deployment/planning_input.rs
```diff
@@ -466,10 +475,32 @@ impl PlanningInputBuilder {
...
+    /// Like `add_omicron_zone_external_ip`, but can accept an [`IpNetwork`],
+    /// validating that the IP is a single address.
+    pub fn add_omicron_zone_external_ip_network(
+        &mut self,
+        zone_id: OmicronZoneUuid,
+        ip_id: ExternalIpUuid,
+        ip: IpNetwork,
+    ) -> Result<(), PlanningInputBuildError> {
+        let size = match ip.size() {
+            NetworkSize::V4(n) => n as u128,
```
jgallagher, line 487: Could we use 

```suggestion
            NetworkSize::V4(n) => u128::from(n),
```

instead? `as` has always made me skittish.

sunshowers: good point -- in this case it's okay because it's lossless. Wonder if we should activate some of the lints in https://rust-lang.github.io/rust-clippy/master/index.html#/cast.

nexus/types/src/deployment/planning_input.rs
```diff
@@ -97,11 +100,12 @@ impl SledResources {
...
 ///
 /// This is a slimmer `nexus_db_model::ExternalIp` that only stores the fields
-/// necessary for blueprint planning.
+/// necessary for blueprint planning, and requires that the service have a
+/// single IP.
 #[derive(Debug, Clone, Serialize, Deserialize)]
-pub struct ExternalIp {
-    pub id: Uuid,
-    pub ip: IpNetwork,
+pub struct ServiceExternalIp {
```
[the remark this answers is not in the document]
sunshowers: Sure.

nexus/types/src/deployment/planning_input.rs
sunshowers: Added a clippy warning in https://github.com/oxidecomputer/omicron/pull/5544.

sunshowers: And fixed in this commit separately as well.

nexus/types/src/deployment/planning_input.rs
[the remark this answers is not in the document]
sunshowers: Done, and also changed `ServiceNetworkInterface` to `OmicronZoneNic` to match this name.

PR #5541, opened 2024-04-16
[nexus-types] more validation for PlanningInput

sunshowers:
While reconciling the PlanningInput with the blueprint, I was a bit concerned
about getting nonsensical inputs (e.g. multiple zones get the same IPs) and the
system going haywire. In particular, I was concerned about the Serialize and
Deserialize impls which were seemingly added for reconfigurator-cli, since they
didn't check for internal invariants.

So I split up the `PlanningInput` into unvalidated and validated forms, such
that the unvalidated form can be deserialized separately. I also added some checks to
ensure that zones, resource IDs and resources are all 1:1:1 maps. (This is
where #5540 is helpful -- it ensures that the only state representable for each
service is as an IP.)

Having these checks here simplifies the life of the blueprint builder
tremendously, meaning it doesn't have to carefully guard against invalid
input.

I haven't added validation tests, wanted to get feedback about this being the
right direction. Just personally speaking, I spent a fair bit of time trying to
figure this out today (filled up two pages of my notebook with a bunch of
tables), and having this logic be explicit would have helped me tremendously.

Some questions:

- [x] ~~Should the reconfigurator CLI switch to deserializing this form and
      showing validation errors?~~ Have decided "no".
- [x] ~~Worth adding validation tests before r8? The validation logic currently
      seems correct, and I can come back to it right after r8 to finish adding
      tests.~~ Decided to postpone since this doesn't block r8.

Depends on #5540.


sunshowers:
Hmm, I just realized that we may need similar checks for blueprints as well.

jgallagher:
> While reconciling the PlanningInput with the blueprint, I was a bit concerned
> about getting nonsensical inputs (e.g. multiple zones get the same IPs) and the
> system going haywire. In particular, I was concerned about the Serialize and
> Deserialize impls which were seemingly added for reconfigurator-cli. However,
> they were liabilities that didn't check for internal invariants.

This is a fair concern that I think we're maybe not particularly careful with in _many_ places. I'm in favor of adding the validation in `Deserialize`, but I'm not sure we need the separate pub `UnvalidatedPlanningInput` type - I don't think any consumer really wants to work with that, right? Even reconfigurator-cli presumably will want to turn that into a validated `PlanningInput` ASAP and bail out if something is wrong.

It might make sense to keep `UnvalidatedPlanningInput` as a private type for clarity of implementation, but I think dropping the pub might allow some nontrivial cleanup (e.g., we probably wouldn't need `impl_planning_input!` because only `PlanningInput` would need those methods?).

sunshowers:
Sure -- I'm fine with that, and if we do decide to make it public in the future we can just do that.

sunshowers:
Going to come back to this after release 8.

sunshowers:
Re-did this in https://github.com/oxidecomputer/omicron/pull/5644.

PR #5544, opened 2024-04-16
[clippy] warn on lossless casts

sunshowers:
Originally brought up in #5540.

It might seem strange to only warn on lossless casts for now, but:

- if the somewhat-illegible "from" type changes, they can suddenly become
  lossy, and
- these are the casts that are easiest to convert over

In the future, it would be nice to also warn on lossy casts.


davepacheco (approved):
I love this.  (I've basically stopped using `as` altogether.)

PR #5555, opened 2024-04-18
[reconfigurator-planning] track more state while building zones

sunshowers:
In #5493 we'd like to track invariants like: a zone should not be added and
expunged in the same blueprint. In order to do that, we need to track this
state. (There are probably other ways to do it, but this is the most
explicit method and I really like that.)

This lives in a submodule because I don't want the rest of the blueprint
builder to reach into the internals here.

I split this from #5493 because it became somewhat complex in its own right,
with its own tests.


nexus/reconfigurator/planning/src/blueprint_builder.rs
```diff
@@ -1285,6 +1394,182 @@ pub mod test {
...
+        // One sled was added.
+        let sleds: Vec<_> = diff.sleds_added().collect();
+        assert_eq!(sleds.len(), 1);
+        let (sled_id, new_sled_zones) = sleds[0];
+        assert_eq!(sled_id, new_sled_id);
+        // The generation number should be newer than the initial default.
+        assert_eq!(new_sled_zones.generation, Generation::new().next());
+        assert_eq!(new_sled_zones.zones.len(), 1);
+
+        // Two sled was modified: existing_sled_id and control_sled_id.
```
jgallagher, line 1543:
```suggestion
        // Two sleds were modified: existing_sled_id and control_sled_id.
```

jgallagher (approved):
LGTM! :+1:

nexus/reconfigurator/planning/src/blueprint_builder.rs
```diff
@@ -945,29 +934,144 @@ impl<'a> BlueprintZonesBuilder<'a> {
...
+                    )
+                }
             })
             .collect()
     }
 }
 
+// This is a sub-module to hide implementation details from the rest of
+// blueprint_builder.
+mod builder_zones {
```
jgallagher, line 964: This seemed fine reviewing this PR, but when I got to #5493, I think I'd prefer if this were in a separate file. It's really a dependency of the parent module, but when reviewing file-by-file that leaves it a little out of order. If it were a separate file, it'd be a little easier to jump to it first then review its caller. Not a big deal if you disagree.

nexus/reconfigurator/planning/src/blueprint_builder.rs
sunshowers: lol, fixed -- thanks!

nexus/reconfigurator/planning/src/blueprint_builder.rs
sunshowers: Hmm, while writing this code it was actually really nice to be able to make changes within the same file, but from the reviewer standpoint I can see that argument. I don't have a strong opinion, so I'll move it into its own file in #5493.

Issue #5561, opened 2024-04-18
nexus oid cache is broken across schema migrations


davepacheco:
We updated dogfood today and ran into a new problem with instance provisions:

```
error_message_internal = saga ACTION error at node "sled_id": unexpected database error: type with ID 218 does not exist
```

After some digging, it appears that Nexus's OID caching is hanging onto a stale value that was changed by the schema migration.  Namely, some sequence like this happens during the upgrade:

- After mupdate, new Nexus starts up, establishes its connections to CockroachDB, and populates its OID cache (which maps enum types like `sled_resource_kind` to their numeric database [OID](https://www.postgresql.org/docs/current/datatype-oid.html)).
- We run the schema migration that drops and re-creates the `sled_resource_kind` enum.  This invalidates the cache entry because now the name `sled_resource_kind` points to a different OID.
- In whatever context Diesel uses that cache (which appears to include at least `INSERT` statements), it uses the old OID, which does not correspond to any existing type any more, and we get this error from the database.

The workaround is to restart Nexus instances after this happens because when they come back up they will re-populate their cache with the correct value.  The real fix will be to somehow invalidate this cache after schema migrations but we're still figuring out how to do that.

davepacheco:
## Summary

**Synopsis:** Any time the database OID for a particular type name changes, as happened with the schema migration in #5287, queries that use this type that are executed by any running Nexus instances will fail with a 500 error or equivalent.  This impact is pretty severe.

This problem was introduced in #4735 and recently triggered by #5287.  Other schema migrations that change the OID for a type name will trigger this again.

**Workaround:** Restart all Nexus instances.

**Plan:** Fixing this properly is non-trivial.  More on this below.  For now, we're going to disable the proactive cache fill, which should mitigate the problem in practice at the expected cost of an extra second or so of wall clock time for Nexus integration tests that use the database.  See #5565.

Props to @andrewjstone for [foreseeing this](https://github.com/oxidecomputer/omicron/pull/4735#pullrequestreview-1800838076) class of problem.

## Problem details

### Background: OIDs, prepared statements, Diesel's OID caching

Database users typically refer to tables, views, and types by their name.  For example, Omicron uses an enum type called `sled_resource_kind`.  Internally, CockroachDB (like PostgreSQL) maps these names to [OIDs](https://www.postgresql.org/docs/current/datatype-oid.html), or object identifiers.

[...]

PR #5565, opened 2024-04-18
Remove diesel OID priming

jgallagher:
See #5561 for rationale. We could add this back if we could figure out how to correctly invalidate the cache after a schema upgrade, but there don't seem to be any easy ways to do that. This will ensure correctness for now and give us time to figure out a different path.

### DIFF
--- a/nexus/db-queries/src/db/pool_connection.rs
+++ b/nexus/db-queries/src/db/pool_connection.rs
@@ -4,207 +4,26 @@
 
 //! Customization that happens on each connection as they're acquired.
 
-use async_bb8_diesel::AsyncConnection;
-use async_bb8_diesel::AsyncRunQueryDsl;
 use async_bb8_diesel::AsyncSimpleConnection;
 use async_bb8_diesel::Connection;
 use async_bb8_diesel::ConnectionError;
 use async_trait::async_trait;
 use bb8::CustomizeConnection;
-use diesel::pg::GetPgMetadataCache;
-use diesel::pg::PgMetadataCacheKey;
-use diesel::prelude::*;
 use diesel::PgConnection;
 use diesel_dtrace::DTraceConnection;
-use std::collections::HashMap;
-use tokio::sync::Mutex;
 
 pub type DbConnection = DTraceConnection<PgConnection>;
-
-// This is a list of all user-defined types (ENUMS) in the current DB schema.
-//
-// Diesel looks up user-defined types as they are encountered, and loads
-// them into a metadata cache. Although this cost is amortized over the lifetime
-// of a connection, this can be slower than desired:
-// - Diesel issues a round-trip database call on each user-defined type
-// - The cache of OIDs for user-defined types is "per-connection", so when
-// using a connection pool, we redo all these calls for new connections.
-//
-// To mitigate: We look up a list of user-defined types here on first access
-// to the connection, and pre-populate the cache. Furthermore, we save this
-// information and use it to populate other connections too, without incurring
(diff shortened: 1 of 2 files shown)


nexus/db-model/src/schema_versions.rs
```diff
@@ -12,6 +12,9 @@ use omicron_common::api::external::SemverVersion;
 use once_cell::sync::Lazy;
 use std::collections::BTreeMap;
 
+pub const SCHEMA_DIR: &str =
```
davepacheco, line 15: Why did you need to move this?

davepacheco (approved):

nexus/db-model/src/schema_versions.rs
davepacheco: Nevermind -- found it!

nexus/db-queries/src/db/pool_connection.rs
```diff
@@ -224,90 +39,152 @@ impl CustomizeConnection<Connection<DbConnection>, ConnectionError>
...
+        {
+            let client = db.connect().await.expect("Failed to connect to CRDB");
+            client
+                .batch_execute(DISALLOW_FULL_TABLE_SCAN_SQL)
+                .await
+                .expect("failed to disallow full table scans");
+
+            let all_versions = AllSchemaVersions::load(SCHEMA_DIR.into())
+                .expect("Failed to load schema versions");
+            for version in all_versions.iter_versions() {
```
davepacheco, line 101: It does make me a little nervous to duplicate this logic.  I wonder if we should modify the existing tests that do this instead?

nexus/db-queries/src/db/pool_connection.rs
```diff
@@ -224,90 +39,152 @@ impl CustomizeConnection<Connection<DbConnection>, ConnectionError>
...
+        // Attempt to insert a sled resource. This involves the
+        // `sled_resource_kind` enum, whose OID was changed by the schema
+        // migration in version 53.0.0. If the diesel OID cache was populated
+        // when we the datastore started polling for the version, this would
+        // fail with a `type with ID $NUM does not exist` error.
+        //
+        // This test isn't perfect: (a) it's possible on an overloaded system
+        // the datastore doesn't start polling CRDB until after we've finished
+        // applying updates (although that's slow, so this seems very unlikely),
+        // and (b) it's possible some _other_ OID has been cached that we're not
```
davepacheco, line 166: (a) seems like a real problem that will lead to a flaky test.

I'm not sure how much plumbing you'd need to do, but what if you create the pool first here, check out a connection (so you know you've got one), then use that pool to create a new datastore?  Then you don't need a background task and there's no race.

davepacheco (changes requested):

nexus/db-queries/src/db/pool_connection.rs
davepacheco: (This one's not a blocker for me.)

jgallagher: The existing tests that do this don't use a connection pool. I guess they could? I'll look and see (taking the other comment into consideration too while I'm there).

jgallagher:
With apologies for the force push (it was easier to start back on `main` to ensure the new test failed while we still had caching in place), both your comments should be addressed now. I moved the test to the existing schema migration test, which muddies it up a bit but is probably better than duplicating the schema migration logic, and it builds a `Pool` by hand to avoid the datastore raciness.

davepacheco (approved):

PR #5574, opened 2024-04-19
add pumpkind to switch and global zones

Nieuwejaar:
This adds the `pumpkind` daemon and service to the switch zone and the global zone, and connects the wires to start it in both places.

In addition, this adds a `-b <branch>` option to each of the `tools/update_<service>` scripts, making it easier to test something other than `main`.  The `-c <commit>` option is still present, but I suspect the `-b` will end up being more commonly used.

I've added Ry as the reviewer, as I _think_ he's the only other person who has added the same service to both the global and in switch zones before.

### DIFF
--- a/.github/buildomat/jobs/package.sh
+++ b/.github/buildomat/jobs/package.sh
@@ -29,6 +29,9 @@ set -o xtrace
 cargo --version
 rustc --version
 
+WORK=/work
+pfexec mkdir -p $WORK && pfexec chown $USER $WORK
+
 #
 # Generate the version for control plane artifacts here. We use `0.git` as the
 # prerelease field because it comes before `alpha`.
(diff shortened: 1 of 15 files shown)


.github/buildomat/jobs/package.sh
```diff
@@ -29,6 +29,9 @@ set -o xtrace
 cargo --version
 rustc --version
 
+WORK=/work
+pfexec mkdir -p $WORK && pfexec chown $USER $WORK
```
Nieuwejaar, line 33: Changing `/work` to `$WORK` makes it easier to run this job by hand on a build machine.  The `mkdir` is moved up to the top because it gets used just below.  I assume this hasn't bitten us before because the directory is created for us in CI.

sled-agent/src/services.rs
```diff
@@ -2760,6 +2762,18 @@ impl ServiceManager {
                             }
                             smfh.refresh()?;
                         }
+                        SwitchService::Pumpkind { asic } => {
+                            // The pumpkin daemon is only needed when running on
+                            // with real sidecar.
+                            if asic == &DendriteAsic::TofinoAsic {
```
Nieuwejaar, line 2768: This has been tested in regular CI, on `a4x2`, and on real hardware, covering both cases for this `if`.

rcgoodfellow (approved):
LGTM

PR #5580, opened 2024-04-19
Allow different versions for differently signed RoT images

labbott:
Currently we restrict all RoT archives for {Gimlet, PSC, Sidecar} to be the same version. This restricts us to having staging and production images to be the same version. What we actually want is all archives for a particular kind _signed with the same key_ to have the same version.

### DIFF
--- a/update-common/src/artifacts/update_plan.rs
+++ b/update-common/src/artifacts/update_plan.rs
@@ -33,6 +33,7 @@ use std::collections::btree_map;
 use std::collections::BTreeMap;
 use std::collections::HashMap;
 use std::io;
+use tokio::io::AsyncReadExt;
 use tufaceous_lib::HostPhaseImages;
 use tufaceous_lib::RotArchives;
 
@@ -112,6 +113,7 @@ pub struct UpdatePlanBuilder<'a> {
     // The by_id and by_hash maps, and metadata, used in `ArtifactsWithPlan`.
     by_id: BTreeMap<ArtifactId, Vec<ArtifactHashId>>,
     by_hash: HashMap<ArtifactHashId, ExtractedArtifactDataHandle>,
+    by_sign: HashMap<(KnownArtifactKind, String), Vec<ArtifactId>>,
     artifacts_meta: Vec<TufArtifactMeta>,
 
     // extra fields we use to build the plan
@@ -144,6 +146,7 @@ impl<'a> UpdatePlanBuilder<'a> {
 
             by_id: BTreeMap::new(),
             by_hash: HashMap::new(),
+            by_sign: HashMap::new(),
             artifacts_meta: Vec::new(),
 
             extracted_artifacts,
@@ -316,6 +319,44 @@ impl<'a> UpdatePlanBuilder<'a> {
                 RotArchives::extract_into(reader, out_a, out_b)
             },
         )?;
+
+        // We need to get all the signing information now to properly check
+        // version at builder time (builder time is not async)
+        let image_a_stream = rot_a_data
+            .reader_stream()
+            .await
+            .map_err(RepositoryError::CreateReaderStream)?;
+        let mut image_a = Vec::with_capacity(rot_a_data.file_size());
+        tokio_util::io::StreamReader::new(image_a_stream)
+            .read_to_end(&mut image_a)
+            .await
+            .expect("failed to read RoT image a");
+
+        let (artifact_id, image_a_sign) =
+            read_hubris_sign_from_archive(artifact_id, image_a)?;
+
+        self.by_sign
+            .entry((artifact_kind, image_a_sign))
+            .or_default()
+            .push(artifact_id.clone());
+
+        let image_b_stream = rot_b_data
+            .reader_stream()
+            .await
+            .map_err(RepositoryError::CreateReaderStream)?;
+        let mut image_b = Vec::with_capacity(rot_b_data.file_size());
+        tokio_util::io::StreamReader::new(image_b_stream)
+            .read_to_end(&mut image_b)
+            .await
+            .expect("failed to read RoT image b");
+
+        let (artifact_id, image_b_sign) =
+            read_hubris_sign_from_archive(artifact_id, image_b)?;
+
+        self.by_sign
+            .entry((artifact_kind, image_b_sign))
+            .or_default()
+            .push(artifact_id.clone());
 
         // Technically we've done all we _need_ to do with the RoT images. We
         // send them directly to MGS ourself, so don't expect anyone to ask for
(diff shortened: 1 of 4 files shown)


update-common/src/artifacts/update_plan.rs
```diff
@@ -317,6 +321,50 @@ impl<'a> UpdatePlanBuilder<'a> {
             },
         )?;
 
+        // We need to get all the signing information now to properly check
+        // version at builder time (builder time is not async)
+        let image_a = rot_a_data
+            .reader_stream()
+            .and_then(|stream| async {
```
sunshowers, line 328: Could this use a `.await` rather than a future combinator? In general, `.await` is preferable -- it's easier to understand, particularly because `.and_then` exists both on futures and on results. (There are exceptions where future combinators make sense, but this doesn't seem to be one of them.)

update-common/src/artifacts/update_plan.rs
```diff
@@ -317,6 +321,50 @@ impl<'a> UpdatePlanBuilder<'a> {
...
+                let mut buf = Vec::with_capacity(rot_a_data.file_size());
+                tokio_util::io::StreamReader::new(stream)
+                    .read_to_end(&mut buf)
+                    .await
+                    .unwrap();
```
sunshowers, line 333: These are extracted as files on disk, so I believe they can produce errors. Even if they can't, it would be good to replace the `unwrap()` with an `expect("justification")`.

update-common/src/artifacts/update_plan.rs
```diff
@@ -317,6 +321,50 @@ impl<'a> UpdatePlanBuilder<'a> {
...
+            .and_then(|stream| async {
+                let mut buf = Vec::with_capacity(rot_b_data.file_size());
+                tokio_util::io::StreamReader::new(stream)
+                    .read_to_end(&mut buf)
+                    .await
+                    .unwrap();
+                Ok(buf)
+            })
+            .await
+            .unwrap();
```
sunshowers, line 358: Same comments about both using `.await` and bubbling up errors here.

update-common/src/artifacts/update_plan.rs
```diff
@@ -895,11 +958,11 @@ mod tests {
...
     }
 
-    fn make_random_rot_image() -> RandomRotImage {
+    fn make_random_rot_image(sign: &str, board: &str) -> RandomRotImage {
         use tufaceous_lib::CompositeRotArchiveBuilder;
 
-        let archive_a = make_random_bytes();
-        let archive_b = make_random_bytes();
+        let archive_a = make_fake_rot_image(sign, board);
+        let archive_b = make_fake_rot_image(sign, board);
```
sunshowers, line 965: haha I was wondering when this would become less random, thanks for doing this :)

sunshowers (commented):
Had a quick look, generally looks good!

update-common/src/artifacts/update_plan.rs
labbott: I was going off of what wicketd was doing (https://github.com/oxidecomputer/omicron/blob/main/wicketd/src/update_tracker.rs#L1679-L1696) but I'm fine with changing it for readability.

update-common/src/artifacts/update_plan.rs
labbott: wicketd mentions this can be an I/O error https://github.com/oxidecomputer/omicron/blob/main/wicketd/src/update_tracker.rs#L1690 . I'll match that here.

update-common/src/errors.rs
```diff
@@ -140,6 +140,8 @@ pub enum RepositoryError {
         "duplicate hash entries found in artifacts.json for kind `{}`, hash `{}`", .0.kind, .0.hash
     )]
     DuplicateHashEntry(ArtifactHashId),
+    #[error("error creating reader stream")]
+    CreateReaderStream(#[source] anyhow::Error),
```
labbott, line 144: Not sure the opinion on adding new errors here

jgallagher: Adding new errors (when needed) is good :+1:

update-common/src/artifacts/update_plan.rs
```diff
@@ -317,6 +320,44 @@ impl<'a> UpdatePlanBuilder<'a> {
...
+        // version at builder time (builder time is not async)
+        let image_a_stream = rot_a_data
+            .reader_stream()
+            .await
+            .map_err(RepositoryError::CreateReaderStream)?;
+        let mut image_a = Vec::with_capacity(rot_a_data.file_size());
+        tokio_util::io::StreamReader::new(image_a_stream)
+            .read_to_end(&mut image_a)
+            .await
+            .expect("failed to read RoT image a");
```
jgallagher, line 333: I don't think we can `.expect()` these - `reader_stream()` returns streams that are backed by `tokio::fs::File`s, which can fail at runtime. Probably need something like a 

```rust
RepositoryError::ReadExtractedArchive {
    artifact: ArtifactHashId,
    #[source]
    error: std::io::Error,
}
```

variant to use here?

update-common/src/artifacts/update_plan.rs
```diff
@@ -700,38 +741,21 @@ impl<'a> UpdatePlanBuilder<'a> {
...
-            // the A artifacts and one from the B artifacts, checked above) so
-            // we can safely unwrap the first.
-            let version =
-                &single_board_rot_artifacts.next().unwrap().id.version;
-            for artifact in single_board_rot_artifacts {
-                if artifact.id.version != *version {
+        // Ensure that all A/B RoT images for each board kind and same
+        // signing key have the same version.
+        for ((kind, _), versions) in self.by_sign {
+            let version = &versions.first().unwrap().version;
```
jgallagher, line 747: Can you add a comment noting why this `.unwrap()` is safe? (The guarantee that there is at least one element happens pretty far away from this up in `add_rot_artifact`, right?)

update-common/src/artifacts/update_plan.rs
```diff
@@ -803,6 +827,38 @@ pub struct UpdatePlanBuildOutput {
...
+            return Err(RepositoryError::ReadHubrisCaboose { id, error });
+        }
+    };
+    let sign = match caboose.sign() {
+        Ok(sign) => sign,
+        Err(error) => {
+            return Err(RepositoryError::ReadHubrisCabooseBoard { id, error });
+        }
+    };
+    let sign = match std::str::from_utf8(sign) {
```
jgallagher, line 853: Do we use the fact that we're storing `sign` as a `String` anywhere? I'm wondering if this conversion is necessary at all, or if we could store `(ArtifactId, Vec<u8>)` instead.

sunshowers (commented):
General logic looks great! Just a few comments + co-signing John's comments.

update-common/src/artifacts/update_plan.rs
```diff
@@ -112,6 +113,7 @@ pub struct UpdatePlanBuilder<'a> {
     // The by_id and by_hash maps, and metadata, used in `ArtifactsWithPlan`.
     by_id: BTreeMap<ArtifactId, Vec<ArtifactHashId>>,
     by_hash: HashMap<ArtifactHashId, ExtractedArtifactDataHandle>,
+    by_sign: HashMap<(KnownArtifactKind, String), Vec<ArtifactId>>,
```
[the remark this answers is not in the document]
labbott: I'll clarify this with some rework and comments

update-common/src/artifacts/update_plan.rs
```diff
@@ -803,6 +827,38 @@ pub struct UpdatePlanBuildOutput {
...
+            return Err(RepositoryError::ParsingHubrisArchive { id, error });
+        }
+    };
+    let caboose = match archive.read_caboose().map_err(Box::new) {
+        Ok(caboose) => caboose,
+        Err(error) => {
+            return Err(RepositoryError::ReadHubrisCaboose { id, error });
+        }
+    };
+    let sign = match caboose.sign() {
```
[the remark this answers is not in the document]
labbott: It's a `&[u8]` (https://github.com/oxidecomputer/hubtools/blob/main/hubtools/src/caboose.rs#L47-L49) although as @jgallagher noted below the extra step of conversion to a string is probably unnecessary except for error cases.

jgallagher (approved):
LGTM :+1:

sunshowers (approved):
agreed, looks great!

PR #5583, opened 2024-04-19
Remove ability to generate a Blueprint from an inventory collection

jgallagher:
We want to add information in blueprints, particularly `BlueprintZoneConfig` and `BlueprintZoneType`, that is not present in the sled-agent types `OmicronZoneConfig` and `OmicronZoneType`. Today on main conversion between those types is bidirectional. As we add to blueprints, we will continue to be able to convert a `BlueprintZoneConfig` into an `OmicronZoneConfig`, but the opposite direction will become more and more difficult as callers need to provide the additional information required for blueprints.

We have enough users of the inventory -> blueprint direction that it's quite painful to try to add to blueprints, so this PR attempts to knock out one of the more common uses: converting an inventory collection into a blueprint via `BlueprintBuilder::build_initial_from_collection()`. This method is removed, and all the remaining changes are fallout from that. Most uses of this have been replaced by either the blueprint produced by `ExampleSystem` or the new `BlueprintBuilder::build_empty_with_sleds()` helper for constructing an empty blueprint. One test needed the full blueprint-from-inventory, so that one actually gained a new use of converting OmicronZoneConfig -> BlueprintZoneConfig. (Not ideal, but fine for now!)


nexus/db-queries/src/db/datastore/deployment.rs
```diff
@@ -1632,10 +1595,23 @@ mod tests {
...
+                .blueprint_disks
+                .values()
+                .map(|c| c.disks.len())
+                .sum::<usize>()
+                + new_sled_zpools.len(),
+            blueprint2
+                .blueprint_disks
+                .values()
+                .map(|c| c.disks.len())
+                .sum::<usize>()
```
jgallagher, line 1614: This assertion change is, I believe, a fix for a test that was passing incorrectly. `blueprint_disks` is a `BTreeMap<SledUuid, PhysicalDisksConfig>`, which means `blueprint_disks.len()` reported "for how many sleds do we have a `PhysicalDisksConfig`?", not "how many disks are in the blueprint" as the check seemed to mean. I think this was passing purely by coincidence:

* `blueprint1.blueprint_disks` was empty, because a blueprint created from an inventory collection didn't populate it at all
* `new_sled_zpools` was 4, because we added 4 disks
* `blueprint2.blueprint_disks.len()` was 4, because there are 4 sleds in blueprint2, and going through the builder populates `blueprint_disks`

so we were previously passing because `0 + 4 == 4`, even though the units of these values made the comparison nonsensical.

nexus/reconfigurator/planning/src/blueprint_builder.rs
```diff
@@ -494,7 +445,7 @@ impl<'a> BlueprintBuilder<'a> {
                 })
                 .collect();
 
-            (additions, removals)
+            dbg!((additions, removals))
```
davepacheco, line 448: stray `dbg!`?

davepacheco (approved):

nexus/reconfigurator/planning/src/blueprint_builder.rs
jgallagher: Yes, nice catch. This was trying to figure out why https://github.com/oxidecomputer/omicron/pull/5583#discussion_r1572910123 was failing :sweat_smile:

PR #5596, opened 2024-04-22
Reconfigurator: Don't talk to expunged sleds from executor

### LINKED ISSUE #5594: Reconfigurator: Executor should not try to talk to expunged sleds
As part of testing in https://github.com/oxidecomputer/omicron/issues/5480, we found that the executor still tries, and fails, to send `OmicronPhysicalDisksConfig` to expunged sleds. This causes the rest of the executor to not complete on the error and so zones don't get deploye.

andrewjstone:
We confirmed on a4x2 that Nexus gets deployed on the new sled with this change.

Fixes #5594

### DIFF
--- a/nexus/reconfigurator/execution/src/omicron_zones.rs
+++ b/nexus/reconfigurator/execution/src/omicron_zones.rs
@@ -10,6 +10,7 @@ use anyhow::Context;
 use futures::stream;
 use futures::StreamExt;
 use nexus_db_queries::context::OpContext;
+use nexus_types::deployment::BlueprintZoneDisposition;
 use nexus_types::deployment::BlueprintZoneFilter;
 use nexus_types::deployment::BlueprintZonesConfig;
 use omicron_uuid_kinds::GenericUuid;
@@ -33,6 +34,16 @@ pub(crate) async fn deploy_zones(
             {
                 Some(sled) => sled,
                 None => {
+                    if config.zones.iter().all(|c| {
+                        c.disposition == BlueprintZoneDisposition::Expunged
+                    }) {
+                        info!(
+                            opctx.log,
+                            "Skipping zone deployment to expunged sled";
+                            "sled_id" => %sled_id
+                        );
+                        return None;
+                    }
                     let err = anyhow!("sled not found in db list: {}", sled_id);
                     warn!(opctx.log, "{err:#}");
                     return Some(err);
--- a/nexus/reconfigurator/execution/src/omicron_zones.rs
+++ b/nexus/reconfigurator/execution/src/omicron_zones.rs
@@ -10,7 +10,6 @@ use anyhow::Context;
 use futures::stream;
 use futures::StreamExt;
 use nexus_db_queries::context::OpContext;
-use nexus_types::deployment::BlueprintZoneDisposition;
 use nexus_types::deployment::BlueprintZoneFilter;
 use nexus_types::deployment::BlueprintZonesConfig;
 use omicron_uuid_kinds::GenericUuid;
--- a/nexus/reconfigurator/execution/src/omicron_zones.rs
+++ b/nexus/reconfigurator/execution/src/omicron_zones.rs
@@ -34,9 +34,7 @@ pub(crate) async fn deploy_zones(
             {
                 Some(sled) => sled,
                 None => {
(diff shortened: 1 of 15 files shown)


sunshowers:
Makes a ton of sense, thanks for doing this!

nexus/reconfigurator/execution/src/omicron_zones.rs
```diff
@@ -33,6 +34,16 @@ pub(crate) async fn deploy_zones(
             {
                 Some(sled) => sled,
                 None => {
+                    if config.zones.iter().all(|c| {
+                        c.disposition == BlueprintZoneDisposition::Expunged
```
jgallagher, line 38: Do you think we'll do this frequently enough it warrants something like `config.are_all_zones_expunged()`?

nexus/src/app/background/inventory_collection.rs
```diff
@@ -180,7 +181,11 @@ impl<'a> nexus_inventory::SledAgentEnumerator for DbSledAgentEnumerator<'a> {
             while let Some(p) = paginator.next() {
                 let records_batch = self
                     .datastore
-                    .sled_list(&self.opctx, &p.current_pagparams())
+                    .sled_list(
+                        &self.opctx,
+                        &p.current_pagparams(),
+                        SledFilter::All,
```
jgallagher, line 187: No objection on this PR, but this might not be right in the limit. (Some discussion in chat after #5480 noted inventory collection timeouts trying to talk to expunged sleds.)

nexus/src/app/rack.rs
```diff
@@ -737,7 +738,10 @@ impl super::Nexus {
         };
 
         debug!(self.log, "Listing sleds");
-        let sleds = self.db_datastore.sled_list(opctx, &pagparams).await?;
+        let sleds = self
+            .db_datastore
+            .sled_list(opctx, &pagparams, SledFilter::All)
```
jgallagher, line 743: Should this be `SledFilter::InService`? I think with `::All` we wouldn't allow re-initializing a previously-expunged sled (because we'd still see its baseboard here)

nexus/src/app/sled.rs
```diff
@@ -115,7 +116,7 @@ impl super::Nexus {
         opctx: &OpContext,
         pagparams: &DataPageParams<'_, Uuid>,
     ) -> ListResultVec<db::model::Sled> {
-        self.db_datastore.sled_list(&opctx, pagparams).await
+        self.db_datastore.sled_list(&opctx, pagparams, SledFilter::All).await
```
jgallagher, line 119: I'm not sure where this is called, but from context I would have guessed this should be `::InService` (e.g., why would Nexus list expunged sleds?). Or maybe this method could take a filter as an argument?

jgallagher (approved):

nexus/src/app/rack.rs
andrewjstone: I wasn't sure about this, but I'm pretty sure you are right. Good catch!

nexus/src/app/sled.rs
andrewjstone: Hmm. I mainly was sticking with the existing functionality here, but I think you are correct. I'll change this.

nexus/src/app/background/inventory_collection.rs
andrewjstone: Yeah, I think based on the level of discussion that should be left to a follow up.

nexus/reconfigurator/execution/src/omicron_zones.rs
andrewjstone: I don't think it will be used that frequently, but I like the method better. Added in 8b07fa885bdf94ed2009e1bda4da5d34dd6b4c72

nexus/src/app/rack.rs
andrewjstone: Fixed in 8b07fa885bdf94ed2009e1bda4da5d34dd6b4c72

nexus/src/app/sled.rs
andrewjstone: Fixed in 8b07fa885bdf94ed2009e1bda4da5d34dd6b4c72

PR #5599, opened 2024-04-23
[nexus] Remove zones on expunged disks

smklein:
Expunges zones for which the corresponding physical disk, holding durable data, has been removed.

A follow-up PR will be necessary to expunge zones which have had their transient zone filesystems removed,
but Nexus is not yet aware of where zone placement decisions are made.

Partial fix of #5372


nexus/reconfigurator/planning/src/planner.rs
```diff
@@ -840,6 +862,61 @@ mod test {
...
+
+        // Avoid churning on the quantity of Nexus zones - we're okay staying at
+        // one.
+        builder.policy_mut().target_nexus_zone_count = 1;
+
+        let (_, sled_details) = builder.sleds_mut().iter_mut().next().unwrap();
+        let (_, disk) =
+            sled_details.resources.zpools.iter_mut().next().unwrap();
+        disk.policy =
+            nexus_types::external_api::views::PhysicalDiskPolicy::Expunged;
```
sunshowers, line 884: Could you just import the `PhysicalDiskPolicy` type?

sunshowers (approved):
I was wondering when we'd do this! Thought it would only be required for r9 but I guess we have to do it for r8. Thanks for doing it :)

nexus/reconfigurator/planning/src/planner.rs
smklein: Done in https://github.com/oxidecomputer/omicron/pull/5599/commits/09f416202b93c263aee224fe3a0bee813fff8490

nexus/reconfigurator/planning/src/blueprint_builder/builder.rs
```diff
@@ -404,6 +404,19 @@ impl<'a> BlueprintBuilder<'a> {
...
+
+    /// Expunges all zones from a sled where `filter_fn` returns `true`.
+    ///
+    /// Returns a list of zone IDs expunged (excluding zones that were already
+    /// expunged). If the list is empty, then the operation was a no-op.
+    pub(crate) fn expunge_all_zones_for_sled_where(
+        &mut self,
+        sled_id: SledUuid,
+        reason: ZoneExpungeReason,
+        filter_fn: impl Fn(&BlueprintZoneConfig) -> bool,
```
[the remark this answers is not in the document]
smklein: Attempted in https://github.com/oxidecomputer/omicron/pull/5599/commits/9e1e18a3b11b2cd27f5ee5fd1e21a5f13c54792c - though FYI @sunshowers this is a more significant refactor of the "whole sled" expungement code too.

nexus/reconfigurator/planning/src/blueprint_builder/builder.rs
```diff
@@ -392,33 +393,68 @@ impl<'a> BlueprintBuilder<'a> {
...
                         "for sled {sled_id}, error computing zones to expunge"
                     )))
                 })?;
 
             if !is_expunged {
-                zones_to_expunge.insert(z.id);
+                match reason {
+                    ZoneExpungeReason::DiskExpunged => {
+                        info!(
+                            &log,
```
sunshowers, line 432: Worth creating another log instance here with the zone ID, maybe the zone type etc, and then using that here.

nexus/reconfigurator/planning/src/blueprint_builder/builder.rs
```diff
@@ -430,51 +466,43 @@ impl<'a> BlueprintBuilder<'a> {
...
+        let count_and_reason = [
+            (count_disk_expunged, "zone was using expunged disk"),
+            (count_sled_decommissioned, "sled state is decommissioned"),
+            (count_sled_expunged, "sled policy is expunged"),
+        ];
+        for (count, reason) in count_and_reason {
+            if count > 0 {
+                self.comment(format!(
+                    "sled {sled_id} ({reason}): {count} zones expunged",
+                ));
```
sunshowers, line 503: Yeah this is exactly the comment handling I was talking about in #5493 -- thanks for doing this!

I think I might consider doing this at a higher level. Instead of storing a list of comments in the `BlueprintBuilder`, store a log of operations. Then, while converting a `BlueprintBuilder` into a `Blueprint`, turn the log of operations into comments (after possibly doing some merging of operations). That would separate out the data layer from the presentation layer -- and you can test against this.

nexus/reconfigurator/planning/src/planner.rs
```diff
@@ -841,6 +861,60 @@ mod test {
...
+        // We should be removing a single zone, associated with the Crucible
+        // using that device.
+        assert_eq!(diff_modified.zones_added().len(), 0);
+        assert_eq!(diff_modified.zones_removed().len(), 0);
+        assert_eq!(diff_modified.zones_modified().count(), 1);
+        assert!(diff_modified
+            .zones_modified()
+            .next()
+            .unwrap()
+            .disposition_changed());
```
sunshowers, line 913: Can we check the exact disposition here rather than just that it changed?

nexus/reconfigurator/planning/src/blueprint_builder/builder.rs
smklein: I'm interested in this proposal, but it seems like it might extend a fair bit beyond the scope of this PR. Right now, the `BlueprintBuilder` is modifying structures in-place, rather than having a clear delineation of the "before" + "applied operations" .

If I was to implement what you're suggesting, presumably we'd alter the `change_sled_zones` and `change_sled_disks` - rather than mutating in place, we'd need some way to represent operations as "additions, updates, or removals" that are stored separately, and applied when we build the blueprint?

(That's my interpretation of your comment at least, am I reading that right?)

nexus/reconfigurator/planning/src/blueprint_builder/builder.rs
smklein: Done in e80756b00b689e939a2b265472eeffe32d3a5a68

nexus/reconfigurator/planning/src/planner.rs
smklein: Good call, I felt like I was taking a little bit of a shortcut when writing this too.

Updated to be more explicit on the zone type and disposition in e80756b00b689e939a2b265472eeffe32d3a5a68

nexus/reconfigurator/planning/src/blueprint_builder/builder.rs
sunshowers: > If I was to implement what you're suggesting, presumably we'd alter the change_sled_zones and change_sled_disks - rather than mutating in place, we'd need some way to represent operations as "additions, updates, or removals" that are stored separately, and applied when we build the blueprint?

Ah so we decided not to do this in #5493. Instead, what we can do is to both apply the changes, and store a structured log of the changes.

smklein: I've attempted to resolve this in https://github.com/oxidecomputer/omicron/pull/5834, or at least *start* this, with an operation log. It is currently being stringified when the blueprint is built, but it could remain a stronger type eventually, if that would be useful.

PR #5600, opened 2024-04-23
[common] Share a common implementation of ZpoolName

smklein:
This removes dependencies on `illumos-utils`, and obviates some unnecessary parsing.


smklein:
This PR is used by https://github.com/oxidecomputer/omicron/pull/5599

nexus/reconfigurator/execution/src/datasets.rs
```diff
@@ -70,19 +69,7 @@ pub(crate) async fn ensure_crucible_dataset_records_exist(
 
         // Map progenitor client strings into the types we need. We never
         // expect these to fail.
```
jgallagher, line 71: This comment can go away too I think? It doesn't look like we're doing any other progenitor client type conversion after this change.

nexus/reconfigurator/planning/src/blueprint_builder/builder.rs
```diff
@@ -112,13 +111,8 @@ pub enum EnsureMultiple {
...
 }
 
-fn zpool_id_to_external_name(zpool_id: ZpoolUuid) -> anyhow::Result<ZpoolName> {
-    let pool_name_generated =
-        illumos_utils::zpool::ZpoolName::new_external(zpool_id).to_string();
-    let pool_name = ZpoolName::from_str(&pool_name_generated).map_err(|e| {
-        anyhow!("Failed to create zpool name from {zpool_id}: {e}")
-    })?;
-    Ok(pool_name)
+fn zpool_id_to_external_name(zpool_id: ZpoolUuid) -> ZpoolName {
```
jgallagher, line 114: Not sure this function is useful anymore - maybe remove and inline `ZpoolName::new_external()` everywhere?

jgallagher (approved):

nexus/reconfigurator/execution/src/datasets.rs
smklein: Done

nexus/reconfigurator/planning/src/blueprint_builder/builder.rs
smklein: Done

PR #5601, opened 2024-04-23
[reconfigurator] Only place Crucible zones on provisionable zpools

smklein:
Previously, the planner placed Crucible zones on zpools unconditionally. Now, we respect the policy and state of disks to
limit which disks are targets to receive these zones.

Partial fix of https://github.com/oxidecomputer/omicron/issues/5372


nexus/types/src/deployment/planning_input.rs
```diff
@@ -87,6 +96,25 @@ pub struct SledResources {
...
 
 impl SledResources {
+    /// Returns if the zpool is provisionable (known, in-service, and active).
+    pub fn zpool_is_provisionable(&self, zpool: &ZpoolUuid) -> bool {
+        let Some(disk) = self.zpools.get(zpool) else { return false };
+        disk.provisionable()
+    }
+
+    /// Returns all in-service, active zpools
+    pub fn provisionable_zpools(
```
sunshowers, line 106: Are we going to need a `ZpoolFilter` soon similar to the `SledFilter` type? One advantage of that type is that it can be used everywhere from this kind of thing to database queries. Regardless, not important here.

sunshowers (approved):
Love how straightforward this is!

nexus/types/src/deployment/planning_input.rs
jgallagher: IIRC we considered specific methods like this for zones and decided to go with the filter because there were going to be a bunch of them (and the consistency between this and db queries as you mention). I'd default to making the same decision here to use a filter, but I could see an argument for methods if we don't expect very many.

smklein: I went ahead and added a `ZpoolFilter` here for consistency. It's simple, and it's implemented very similarly to the `DiskFilter`, but I'm okay being slightly verbose if it keeps us the code consistent.

PR #5606, opened 2024-04-23
Support env variables in virtual-hardware xtask

andrewjstone:
Some of these environment variables were not carried over when transitioning from the `create-virtual-hardware.sh` bash script to the xtask. This prevented the a4x2 testbed softnpu setup from being correctly setup, which resulted in bootstrap addresses not being routable. This commit fixes that.

Once this is merged, then the corresponding [fix](https://github.com/oxidecomputer/testbed/pull/43) in testbed should work.

### DIFF
--- a/dev-tools/xtask/src/virtual_hardware.rs
+++ b/dev-tools/xtask/src/virtual_hardware.rs
@@ -19,7 +19,7 @@ enum Commands {
         /// The physical link over which Chelsio links are simulated
         ///
         /// Will be inferred by `dladm show-phys` if unsupplied.
-        #[clap(long)]
+        #[clap(long, env)]
         physical_link: Option<String>,
 
         /// Sets `promisc-filtered` off for the sc0_1 vnic.
@@ -33,6 +33,10 @@ enum Commands {
         /// Will be inferred via `netstat` if unsupplied.
         #[clap(long)]
         gateway_ip: Option<String>,
+
+        /// The configured mode for softnpu
+        #[clap(long, env, default_value = "zone")]
+        softnpu_mode: String,
 
         /// The MAC address of your gateway IP
         ///
@@ -153,6 +157,7 @@ pub fn run_cmd(args: Args) -> Result<()> {
         Commands::Create {
             physical_link,
             promiscuous_filter_off,
+            softnpu_mode,
             gateway_ip,
             gateway_mac,
             pxa,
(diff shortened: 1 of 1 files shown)


rcgoodfellow (approved):
Thanks!

dev-tools/xtask/src/virtual_hardware.rs
```diff
@@ -34,6 +34,10 @@ enum Commands {
         #[clap(long)]
         gateway_ip: Option<String>,
 
+        /// The configured mode for softnpu
```
smklein, line 37: Should this be a bool? we only know the value `zone`, and all other values are ignored

rcgoodfellow: If we change this to a bool, i think the environment variable name would need to change to something like softnpu_zone_mode. The idea behind using a string was to be able to accommodate multiple modes, it just turns out right now that there is nothing special the virtual hardware setup needs to do for "propolis" mode.

dev-tools/xtask/src/virtual_hardware.rs
```diff
@@ -34,6 +34,10 @@ enum Commands {
...
+        #[clap(long, env, default_value = "zone")]
```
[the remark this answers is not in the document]
sunshowers: Should we have a prefix? i.e. instead of accepting `SOFTNPU_MODE`, use something like `OMICRON_SOFTNPU_MODE`.

dev-tools/xtask/src/virtual_hardware.rs
```diff
@@ -34,6 +34,10 @@ enum Commands {
...
+        softnpu_mode: String,
```
sunshowers, line 39: Can this be an arbitrary string? If it's only one of a small number of values, it'd be better to use `ValueEnum`.

dev-tools/xtask/src/virtual_hardware.rs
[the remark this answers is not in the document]
andrewjstone: I don't think it's necessary. We'd also need to change it in a4x2 if we were going to expand it.

dev-tools/xtask/src/virtual_hardware.rs
andrewjstone: I believe it's determined by the caller, and may have uses outside this xtask. If for some reason some using software changed it, I wouldn't want deployment to break. However, I'm also not sure it's that big of a deal to require updating this script.

andrewjstone:
Thanks for the comments all. My preference right now is to merge in as is and keep moving, since testbed is currently broken without this.

PR #5609, opened 2024-04-24
Make network interface deletion query idempotent

jgallagher:
I ran into this working on the Reconfigurator executor trying to clean up external IPs and network interfaces. Deleting external IPs is already idempotent, but deleting NICs is not, and trying to delete the same NIC twice fails with an incorrect error (specifically, that the interface cannot be deleted because secondary interfaces still exist, which doesn't make sense because the interface is _already_ deleted and there never were any secondary interfaces).

The same query is used to delete instance network interfaces, so I'd appreciate a careful look here that I didn't break something. I think this change does fix a very minor bug for instance NICs, though, where clients could see the same nonsensical error about secondary interfaces. Specifically, today on main, if you delete an interface twice from the CLI serially, the second attempt 404s as expected:

```console
% ./oxide instance nic delete --interface 561bf0bd-acd3-4be9-87fc-5ad3f5df1d51
null
% ./oxide instance nic delete --interface 561bf0bd-acd3-4be9-87fc-5ad3f5df1d51
error
Error Response: status: 404 Not Found; headers: {"content-type": "application/json", "x-request-id": "4eab6025-0678-4302-81aa-6ca04958b091", "content-length": "195", "date": "Wed, 24 Apr 2024 14:46:01 GMT"}; value: Error { error_code: Some("ObjectNotFound"), message: "not found: instance-network-interface with id \"561bf0bd-acd3-4be9-87fc-5ad3f5df1d51\"", request_id: "4eab6025-0678-4302-81aa-6ca04958b091" }
```

But the 404 is coming from [the authz lookup](https://github.com/oxidecomputer/omicron/blob/aec34001a1a9df2c503a7c8cee51530d9c43af2e/nexus/src/app/network_interface.rs#L168-L169), not the database query, so if we issue multiple requests simultaneously, they can TOCTOU past the authz lookup, at which point one succeeds and the other fails with the nonsense 400:

```console
% ./oxide instance nic delete --interface dbea0c30-9389-4aaf-bbcc-582de02e6736 &
% ./oxide instance nic delete --interface dbea0c30-9389-4aaf-bbcc-582de02e6736 &
null
error
Error Response: status: 400 Bad Request; headers: {"content-type": "application/json", "x-request-id": "1e3d8c7e-db60-4ee4-8c2c-5c46f0dae691", "content-length": "195", "date": "Wed, 24 Apr 2024 14:48:40 GMT"}; value: Error { error_code: Some("InvalidRequest"), message: "The primary interface may not be deleted while secondary interfaces are still attached", request_id: "1e3d8c7e-db60-4ee4-8c2c-5c46f0dae691" }
```

This PR changes the query to return a `Result<bool, ...>` indicating whether or not the delete was performed (using an almost identical CTE as our [update_and_check gadget](https://github.com/oxidecomputer/omicron/blob/aec34001a1a9df2c503a7c8cee51530d9c43af2e/nexus/db-queries/src/db/update_and_check.rs#L200-L216), which I don't think is directly usable here because we already have a CTE of our own, not a diesel update statement). In the public API, I'm mapping `Ok(false)` to a 404, which should fix the above 400.

There's one other caller of this function I didn't touch: the instance create saga has an undo action [that uses this query to delete a NIC](https://github.com/oxidecomputer/omicron/blob/aec34001a1a9df2c503a7c8cee51530d9c43af2e/nexus/src/app/sagas/instance_create.rs#L437-L445). I don't think this is idempotent today - is it supposed to be? It will be as of this PR, but if that's wrong and it needs to do something special for `Ok(false)`, I'm happy to make changes there.

### DIFF
--- a/nexus/db-queries/src/db/queries/network_interface.rs
+++ b/nexus/db-queries/src/db/queries/network_interface.rs
@@ -1692,7 +1692,7 @@ impl Query for DeleteQuery {
 impl RunQueryDsl<DbConnection> for DeleteQuery {}
 
 /// Errors related to deleting a network interface
-#[derive(Debug, thiserror::Error, SlogInlineError)]
+#[derive(Debug, thiserror::Error, SlogInlineError, PartialEq)]
 pub enum DeleteError {
     /// Attempting to delete the primary interface, while there still exist
     /// secondary interfaces.
@@ -1728,6 +1728,24 @@ impl DeleteError {
                     e,
                     query.parent_id,
                 )
+            }
+            // Faithfully plumb through `NotFound`
+            DieselError::NotFound => {
+                let type_name = match query.kind {
+                    NetworkInterfaceKind::Instance => {
+                        external::ResourceType::InstanceNetworkInterface
+                    }
+                    NetworkInterfaceKind::Service => {
+                        external::ResourceType::ServiceNetworkInterface
+                    }
+                    NetworkInterfaceKind::Probe => {
+                        external::ResourceType::ProbeNetworkInterface
+                    }
+                };
+                DeleteError::External(external::Error::ObjectNotFound {
+                    type_name,
+                    lookup_type: external::LookupType::ById(query.interface_id),
+                })
             }
             // Any other error at all is a bug
             _ => DeleteError::External(error::public_error_from_diesel(
@@ -1823,6 +1841,7 @@ fn decode_delete_network_interface_database_error(
 mod tests {
     use super::first_available_address;
     use super::last_address_offset;
+    use super::DeleteError;
     use super::InsertError;
     use super::MAX_NICS_PER_INSTANCE;
     use super::NUM_INITIAL_RESERVED_IP_ADDRESSES;
@@ -1857,7 +1876,6 @@ mod tests {
     use omicron_common::api::external::MacAddr;
     use omicron_test_utils::dev;
     use omicron_test_utils::dev::db::CockroachInstance;
-    use slog_error_chain::InlineErrorChain;
     use std::collections::HashSet;
     use std::convert::TryInto;
     use std::net::IpAddr;
@@ -2177,22 +2195,24 @@ mod tests {
         assert!(!second_deleted, "second delete did nothing");
 
         // Attempting to delete a nonexistent interface should fail.
+        let bogus_id = Uuid::new_v4();
         let err = context
             .db_datastore
             .service_delete_network_interface(
                 &context.opctx,
                 service_id,
-                Uuid::new_v4(),
+                bogus_id,
             )
             .await
             .expect_err(
                 "unexpectedly succeeded deleting nonexistent interface",
             );
-        let err = InlineErrorChain::new(&err).to_string();
-        assert!(
-            err.contains("Record not found"),
-            "expected `Record not found`; got {err:?}"
-        );
+        let expected_err =
+            DeleteError::External(external::Error::ObjectNotFound {
+                type_name: external::ResourceType::ServiceNetworkInterface,
+                lookup_type: external::LookupType::ById(bogus_id),
+            });
+        assert_eq!(err, expected_err);
         context.success().await;
     }
 
--- a/nexus/db-queries/src/db/queries/network_interface.rs
+++ b/nexus/db-queries/src/db/queries/network_interface.rs
@@ -11,15 +11,16 @@ use crate::db::pool::DbConnection;
 use crate::db::queries::next_item::DefaultShiftGenerator;
 use crate::db::queries::next_item::NextItem;
 use crate::db::schema::network_interface::dsl;
+use async_bb8_diesel::AsyncRunQueryDsl;
 use chrono::DateTime;
 use chrono::Utc;
 use diesel::pg::Pg;
 use diesel::prelude::Column;
-use diesel::query_builder::AstPass;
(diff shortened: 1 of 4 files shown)


nexus/db-queries/src/db/queries/network_interface.rs
```diff
@@ -1502,18 +1504,39 @@ fn push_instance_state_verification_subquery<'a>(
 ///                     parent_id = <parent_id> AND
 ///                     kind = <kind> AND
 ///                     time_deleted IS NULL
-///             ) = 1,
+///             ) <= 1,
```
jgallagher, line 1507: This `= 1` is where we were triggering the incorrect "can't delete while secondaries exist" error. Specifically, one of two things has to be true in this check; either:

a) this interface is a secondary interface that is not deleted (lines 1489-1495), or
b) there is exactly 1 interface for this instance/service that is not deleted (lines 1499-1507)

If we're only considering instances/services with a single, primary interface, (a) is never true, so we always check (b). If we've already deleted the only interface, this count returns 0, not one, triggering the `'secondaries'` error sentinel. The change here is to allow "at most 1" instead of "exactly 1", so we get past this check and proceed to the rest of the CTE if there are no undeleted interfaces left.

nexus/db-queries/src/db/queries/network_interface.rs
```diff
@@ -2042,6 +2123,78 @@ mod tests {
         }
     }
 
+    #[tokio::test]
+    async fn test_delete_service_is_idempotent() {
+        let context =
+            TestContext::new("test_insert_running_instance_fails", 2).await;
```
bnaecker, line 2129: Looks like the test name was copy-pasted from another test?

luqmana:
> There's one other caller of this function I didn't touch: the instance create saga has an undo action [that uses this query to delete a NIC](https://github.com/oxidecomputer/omicron/blob/aec34001a1a9df2c503a7c8cee51530d9c43af2e/nexus/src/app/sagas/instance_create.rs#L437-L445). I don't think this is idempotent today - is it supposed to be? It will be as of this PR, but if that's wrong and it needs to do something special for `Ok(false)`, I'm happy to make changes there.

The saga already seems to swallow a NotFound from the authz check with just a warning. It probably makes sense to do the same with the actual query

nexus/db-queries/src/db/queries/network_interface.rs
```diff
@@ -1545,6 +1568,37 @@ impl DeleteQuery {
...
+                Ok(true)
+            }
+            (Some(_), None) => Ok(false),
+            (None, Some(deleted)) => {
+                panic!(
+                    "internal query error: \
+                     deleted nonexisted interface {deleted}"
+                )
+            }
+            (None, None) => Err(DieselError::NotFound),
```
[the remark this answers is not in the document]
jgallagher: Good call - fixed in [c382a28](https://github.com/oxidecomputer/omicron/pull/5609/commits/c382a28e822697627e8c765058cac4feb68a8860)

nexus/db-queries/src/db/queries/network_interface.rs
```diff
@@ -1608,25 +1670,43 @@ impl QueryFragment<Pg> for DeleteQuery {
...
 
+impl Query for DeleteQuery {
+    type SqlType = (Nullable<sql_types::Uuid>, Nullable<sql_types::Uuid>);
+}
+
 impl RunQueryDsl<DbConnection> for DeleteQuery {}
 
 /// Errors related to deleting a network interface
-#[derive(Debug)]
+#[derive(Debug, thiserror::Error, SlogInlineError, PartialEq)]
```
luqmana, line 1695: Nit: you can drop the `thierror::Error, SlogInlineError` now

nexus/src/app/sagas/instance_create.rs
```diff
@@ -428,38 +428,42 @@ async fn sic_create_network_interface_undo(
...
             datastore
                 .instance_delete_network_interface(
                     &opctx,
                     &authz_instance,
                     &authz_interface,
                 )
                 .await
-                .map_err(|e| e.into_external())?;
-            Ok(())
+                .map_err(|e| e.into_external())?
```
luqmana, line 447: Hmm, so we could still return `ObjectNotFound` here (NIC doesn't exist at all) thus skipping the warning below. In theory we shouldn't end up there since the authz lookup should fail first and we don't do hard deletes. I guess failing the saga makes sense at that point.

luqmana (approved):
Thanks for fixing this wart!

nexus/db-queries/src/db/queries/network_interface.rs
jgallagher: Ah, this is actually so I can slog these errors from a Nexus RPW in a subsequent PR, unless there's an objection to this other than that it isn't used yet?

nexus/src/app/sagas/instance_create.rs
jgallagher: I think that's right, yeah - an `ObjectNotFound` on this call means the NIC ID has either been hard deleted or never existed, neither of which seems like a "just warn and keep going" case.

nexus/db-queries/src/db/queries/network_interface.rs
luqmana: Ah, carry on then!

PR #5610, opened 2024-04-24
Prevent chrony from authoritatively advertising times from the 1980s.

citrus-it:


.github/buildomat/jobs/deploy.sh
```diff
@@ -139,6 +139,9 @@ fi
 #
 pfexec /sbin/zfs create -o mountpoint=/zone rpool/zone
 
+# Ensure we have the latest chrony package
+pfexec pkg update chrony
```
smklein, line 143: This doesn't change the version of chrony installed in our production images, does it? I thought this was only for our lab tests (still worthwhile keeping up to date, admittedly!).

citrus-it: The production images already have chrony 4.5, but buildomat images don't yet, and that's what I was seeing if I could work around for testing this. It doesn't work as pkg is not in the image.

zone-setup/src/bin/zone-setup.rs
```diff
@@ -333,8 +333,14 @@ allow @ALLOW@
...
-# in this mode, we report as stratum 10 to clients.
-local stratum 10
+# in this mode, we report as stratum 10 to clients. The `distance' parameter
+# controls when we will decide to abandon the upstreams and switch to the local
+# reference. By setting `activate`, we prevent the server from ever activating
+# its local reference until it has synchronised with upstream at least once and
+# the root distance has dropped below the provided threshold. This prevents
+# a boundary server in a cold booted rack from authoritatively advertising a
+# time from the 1980s prior to gaining external connectivity.
+local stratum 10 distance 0.4 activate 0.5
```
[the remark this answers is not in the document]
citrus-it: That's a great summary, I will work those into the comment.

The numbers are a very conservative - the default value for distance is 1.0, all of the deployed racks I've surveyed have distances in the < 0.01 range. There's no hard and fast requirement for distance < activate, but that seems to be a reasonable constraint to me. Activate only triggers once, after that we switch to and from the local reference as the root distance crosses the `distance` threshold.

.github/buildomat/jobs/deploy.sh
jclulow: This is done now, FYI.

Issue #4872, opened 2024-01-23
need a way to trigger cleanup and next steps for vanished instances


davepacheco:
There are various cases where the control plane might discover that Instances that we thought were running are, in fact, not.  Examples (from concrete to speculative):

- when Nexus is notified by Sled Agent (in turn notified from Propolis) that a guest is no longer running (because it was stopped from inside the guest)
- when a sled reboots, its instances are all gone and this process needs to be applied to each of them (this is #3633)
- when we implement support for (ungraceful) sled removal, we'll need to apply this to all of the instances that we thought were on the system that's been removed
- we probably want some periodic task that asks Sled Agent what instances it's running and compares that to what we think it should be running
- we _may_ want a periodic task that monitors sleds and, if we find that they seem to be unexpectedly not responding, takes steps to make sure that they're offline (i.e., power cycles them via Ignition) and then applies this process for any instances on them

Relaying some notes from @gjcolombo (my apologies where I've got these details wrong):

- this code path exists today where we handle the first case above (where Nexus handles a message from Sled Agent saying the instance is no longer running)
- this code path has some issues and is not well factored use in other contexts like we need here
- all of this would be reworked by #3742
- the "cleanup" referred to here includes: foreign key from instance to vmm table needs to be reaped; networking state needs to be cleaned up; provisioning counters and sled resources need to be cleaned up; etc.
- the "next steps" referred to here include: looking at some per-instance "fault discipline" (currently the "boot_on_fault" boolean) to determine if we should start the instance running again somewhere else and, if so, doing that

We may view this issue as a dup of #3742 but I wanted to file this separate ticket to reflect more precisely what's needed for sled removal.  The implementation may well just be #3742.

gjcolombo:
#4226 is also tangentially related. Today Nexus uses the "Failed" instance state to mean, more or less, "something went wrong during some operation on this instance," and the only recovery action we support is deleting the instance (see #2825). #4226 proposes distinguishing between different values of "something" so that we can recover appropriately: transient error conditions might resolve on retry, but if a sled reports that a VMM is definitely gone then the error is permanent.

4226 could possibly be resolved without handling cases like "the operator says this sled is gone, so all the living VMMs there should immediately be marked as gone," so these issues aren't duplicates. But I think they're all dancing around a single concept--Nexus should have an idea of a VMM's "health," which I think could start out along these lines:

- this VMM was healthy the last time we checked on it
[...]

PR #5611, opened 2024-04-24
[nexus] add RPW for proactively pulling instance state from sled-agents

hawkw:
Currently, instance states are communicated by the sled-agents of the
sleds on which an instance is running calling into a Nexus HTTP API.
This is a "push"-shaped communication pattern, where the monitored
resources (the sled-agents) act as clients of the monitor (Nexus) to
publish their state to it. While push-shaped communications are
generally more efficient than pull-shaped communications (as they only
require communication when a monitored resource actively changes state),
they have a substantial disadvantage: if the sled-agent is gone, stuck,
or otherwise failed, Nexus will never become aware of this, and will
continue happily chugging along assuming the instances are in the last
state reported by their sled-agents. 

In order to allow Nexus to determine when a sled-agent or VMM has
failed, this branch introduces a background task ("instance_watcher")
that periodically attempts to pull the states of a sled-agent's managed
instances. This way, if calls to sled agents fail, the Nexus process is
made aware that a sled-agent it expected to publish the state of its
monitored instances may no longer be doing so, and we can (eventually)
take corrective action. 

Presently, the `instance_watcher` task does not perform corrective
action, such as restarting failed instances or advancing instances to
the "failed" state in CRDB, as affirmatively detecting failures will
require other work. For example, if a sled-agent is unreachable when
checking on a particular instance, the instance itself might be fine,
and we shouldn't decide that it's gone unless we also can't reach the
`propolis-server` process directly. Furthermore, there's the split-brain
problem to consider: the Nexus process performing the check might be on
one side of a network partition...and it might not be on the same side
of that partition as the majority of the rack, so we may not want to
consider an unreachable sled-agent "failed" unless *all* the Nexus
replicas can't talk to it. And, there are probably even more
considerations I haven't thought of yet. I'm planning on starting an RFD
to propose a comprehensive design for instance health checking.

In the meantime, though, the `instance_watcher` task just emits a bunch
of Oximeter metrics describing the results of instance checks. These
metrics capture:
- Cases where the instance check *failed* in a way that might indicate
  that something is wrong with the instance or its sled-agent or sled
- Cases where the instance check was *unsuccessful* due to an internal
  error, such as a client-side request error or an error updating the
  instance's database records
- Cases where the instance check was successful, and what state updates
  occurred as a result of the check:
  - Whether the instance state in the `instances` table was updated as a
    result of the state returned by sled-agent
  - Whether the VMM state in the `vmms` table was updated as a result of
    the state returned by sled-agent


nexus/src/app/background/instance_watcher.rs
```diff
@@ -0,0 +1,222 @@
...
+                    //    "failed"
+                    // 4. this might mean that the whole sled is super gone,
+                    //    figure that out too.
+                    //
+                    // for now though, we'll just log a really big error.
+                    slog::error!(
+                        opctx.log,
+                        "instance seems to be in a bad state: {error}"
+                    );
+                    return;
```
smklein, line 112: For the moment, we are avoiding "automatically marking a whole sled as failed" -- this is an operation that an operator can perform, but we're avoiding automating that behavior.

https://github.com/oxidecomputer/rfd/pull/691#discussion_r1463879601

However, operators will have the ability to mark sled `policy`, which can lead to expungement.

So:
- We could read from the `omicron.public.sled` table
- If the sled is no longer in service, we've decided that the sled **should** be gone
- If that sled still has instances, they **should** fail

As an alternative, when we make the decision to "expunge" a sled, we could automatically mark all instances on that sled "failed" there, too, rather than deferring to this background task.

smklein: Also, https://github.com/oxidecomputer/omicron/issues/4872 seems relevant!

I *think* that if a sled reboots, it should come back up and explicitly "not return any info about the instance" Nexus queries the Sled Agent (thought it may intermittently "not respond" to Nexus while it's booting up)

nexus/src/app/background/instance_watcher.rs
```diff
@@ -0,0 +1,222 @@
...
+        let client = client.clone();
+
+        async move {
+            let InstanceWatcher { datastore, resolver, max_retries } = watcher;
+            slog::trace!(opctx.log, "checking on instance...");
+            let backoff = backoff::retry_policy_internal_service();
+            let mut retries = 0;
+            let rsp = backoff::retry_notify(
+                backoff,
+                || async {
```
smklein, line 67: I kinda agree with @andrewjstone 's comments here -- I'm not sure that we should be retrying in this background task.

I don't think that this task -- instance monitoring -- is ultimately responsible for deciding whether or not a sled is alive. It *is*, however, responsible for updating the state of an instance if the sled rebooted. In this case, if we see:

- Sled policy: not expunged
- Sled agent: not responding

... and we wait... later, we should see:

- Sled policy: not expunged
- Sled agent: responding! and I don't know about your instance.

I don't *think* this requires an inner retry loop, nor a "max retries" check.

As an alternative, if we expunge the sled, we'll see:

- Sled policy: expunged
- Sled agent: <we shouldn't talk to it>

We'll need to sort through #4872 to decide what to do here, but at least, we'll know that the sled being out-to-lunch is an explicitly marked operation, not something we need to infer

smklein (commented):
Great start -- this is a first batch of comments, and I can come back and do a deeper review in a moment

nexus-config/src/nexus_config.rs
```diff
@@ -519,6 +522,18 @@ pub struct RegionReplacementConfig {
...
 
+#[serde_as]
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+pub struct InstanceWatcherConfig {
+    /// period (in seconds) for periodic activations of this background task
+    #[serde_as(as = "DurationSeconds<u64>")]
+    pub period_secs: Duration,
+
+    /// maximum number of retries to attempt before considering a sled-agent
+    /// dead.
```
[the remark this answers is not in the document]
hawkw: Yeah, that makes sense. Right now, we're not doing _anything_ with this information besides logging a warning, but I think that the eventual end goal would be to use this as a _signal_ that we should look in other places to figure out what's going on. E.g. I was also thinking that if the sled-agent isn't responding, Nexus may also want to try talking to the VMMs directly to see if _they're_ interested in talking to it --- it might just be the sled-agent that's wedged. And, before we make determinations about the whole sled, we should absolutely look elsewhere as well.

nexus/src/app/background/instance_watcher.rs
hawkw: Okay, regarding this and your comments from https://github.com/oxidecomputer/omicron/pull/5611#discussion_r1578505777 and https://github.com/oxidecomputer/omicron/pull/5611#discussion_r1578523951, here's what I'm currently thinking I'm going to do as part of this PR:

- remove all the retry logic
- add a new table of "sled instance check failures" to the database. for now, i'm imagining that this would be shaped sort of like `(sled_id, instance_id, id_of_the_nexus_performing_the_health_check, time_updated, failure_count)`. we might also want to add a notion of failure kinds to this as well.
- if we ask a sled-agent about an instance we believe it has, and the sled-agent responds with an HTTP error that proactively indicates that it does not know about that instance, mark that instance as `Failed` (as I believe this would indicate that the sled has been rebooted and the instance is gone[^1])
- for all other errors (the ones that currently get retried), such as TCP connections refused/timing out, etc, increment the failure count table for that `(sled_id, instance_id, our_nexus_id)` tuple
- _for now_, all we do is increment that error count. eventually, the error count might be exposed to an operator and/or used to drive automated decisions (e.g. if the failure count for a particular instance is > _n_, but the sled-agent has responded successfully for other instances, restart the instance...if the failure count for all instances on a sled is > _n_, power cycle the sled, etc)

What do you think?

[^1]: Potentially, if we wanted to be maximally distrustful of the `sled-agent`'s understanding of the state of the world, we could also try pinging the actual `propolis-server` process for that instance, to see if _it_ still exists. If the `sled-agent` doesn't believe it exists, it _shouldn't_, but unfortunately, as you probably already know, distributed systems...

hawkw: It does seem like emitting both these failures and failures from clients to sled-agent endpoints elsewhere in Nexus as Oximeter metrics is potentially a good idea. The table of failure counts was just sort of a strawman proposal so that we're doing _something_ with the information here...I'm also happy to just log them and punt on doing anything else in this PR!

bnaecker: If you would like to spit these out as a timeseries, it might be helpful to look at the [`oximeter` self-stats](https://github.com/oxidecomputer/omicron/blob/main/oximeter/collector/src/self_stats.rs) module, which reports statistics about its own collections. A very similar kind of reporting could be done here, I think.

hawkw:
Hmm, upon further consideration, I think this change actually _does_ depend on the instance-update saga that Greg proposed being implemented first. The reason for this is that i realized that in the current code, where instance states are only pushed from sled-agents _to_ Nexus, the sled agent only calls `cpapi_instances_put` on _one_ Nexus replica, essentially giving it a lock on the instance row for _that_ state update. On the other hand, in a design where states are pulled by Nexuses, we might end up in a situation where multiple Nexus instances race to update the instance in CRDB, possibly with different states, so I think we need to introduce Greg's "instance updater lock" before we update instance states from data pulled by Nexus.

As an aside, that _does_ also highlight another reason pulling instance states is, in the long run, the right thing. In a push-based model, it's possible that a sled-agent might push an instance state update to a Nexus replica that crashes before it gets to write the state update. If that happens, that state update is basically lost forever, because we only ever send it to one Nexus. That _could_ also be solved by adding retries to the Nexus client in sled-agent,[^1] provided that we would retry with a different Nexus replica, but pulling instance states has other advantages, so it's probably the better long term solution.

[^1]: and it's possible that they're already there, since the client implementation is hidden in Progenitor-generated code? I'm still trying to figure that out...

hawkw:
> Hmm, upon further consideration, I think this change actually _does_ depend on the instance-update saga that Greg proposed being implemented first. The reason for this is that i realized that in the current code, where instance states are only pushed from sled-agents _to_ Nexus, the sled agent only calls `cpapi_instances_put` on _one_ Nexus replica, essentially giving it a lock on the instance row for _that_ state update. On the other hand, in a design where states are pulled by Nexuses, we might end up in a situation where multiple Nexus instances race to update the instance in CRDB, possibly with different states, so I think we need to introduce Greg's "instance updater lock" before we update instance states from data pulled by Nexus.

Or, perhaps the instance's generation number already guards against this? I'm not totally sure, I'll have to think a bit more about it...

nexus/src/app/background/instance_watcher.rs
hawkw: A quick update on this: I'm currently working on emitting these failure counts as Oximeter metrics. I'll probably want @bnaecker's feedback on how I'm doing that once I push it up for review --- I can ping you then!

hawkw:
Okay, I've updated this branch so that the `instance_watcher` background task now records a bunch of Oximeter metrics, rather than attempting to make any determinations of instance failure other than those reported explicitly by the sled-agent. I've changed the PR description to reflect this.

I'd love another review from @smklein and @bnaecker, when you have ha chance!

nexus/src/app/background/instance_watcher.rs
```diff
@@ -0,0 +1,635 @@
...
+                }
+            }
+        }
+    }
+}
+
+#[derive(
+    Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, oximeter::Target,
+)]
+struct InstanceTarget {
```
bnaecker, line 190: I would probably eschew the use of "target" in the name here, it's mostly `oximeter`-internal jargon. Something like the `VirtualMachine` struct we use in Propolis might be good. The exact same name with different fields will not raise conflict errors, so long as the _full timeseries name_ (target name + metric name) has a unique and consistent set of fields. So don't name any metric here `VcpuUsage` and you should be good :)

In any case, we should definitely add the project and silo IDs as fields to these timeseries. I might also recommend adding the sled's model / revision / serial as well, if they're easily accessible at the point we build  type.

nexus/src/app/background/instance_watcher.rs
```diff
@@ -0,0 +1,635 @@
...
+    sled_agent_id: Uuid,
+    /// The sled agent's IP address.
+    sled_agent_ip: IpAddr,
+    /// The sled agent's port.
+    sled_agent_port: u16,
+}
+
+struct CheckResult {
+    target: InstanceTarget,
+    /// `Some` if the instance's state was up
```
bnaecker, line 203: Spelling, I think? Did you mean "state was up-**dated**"?

nexus/src/app/background/instance_watcher.rs
```diff
@@ -0,0 +1,635 @@
...
+    instance_updated: Option<crate::app::instance::InstanceUpdated>,
+
+    /// `Some` if the instance check indicated that the instance is in a bad state.
+    ///
+    /// This is a result that indicates that something is *wrong* with either the
+    /// sled on which the instance is running, the sled-agent on that sled, or the
+    /// instance itself. This is distinct from a [`CheckError`], which indicates
```
bnaecker, line 210: Does this enum really reflect the instance state itself? From the variants below, it appears to be only about errors the sled-agent may return to us, not "something wrong with the instance itself."

nexus/src/app/background/instance_watcher.rs
```diff
@@ -0,0 +1,635 @@
...
+
+                    let mut client = mk_client(&curr_sled);
+                    tasks.spawn(self.check_instance(
+                        opctx,
+                        &curr_sled,
+                        &client,
+                        sled_instance,
+                    ));
+
+                    for (sled, sled_instance) in batch {
```
bnaecker, line 334: I'm a bit confused why we have a `batch.next()` call _in addition_ to the loop here. Could these be collapsed into one loop, like a `while let` loop? A note about why it's structured this way may also be sufficient!

nexus/src/app/background/instance_watcher.rs
```diff
@@ -0,0 +1,635 @@
...
+
+        pub(super) fn prune(&mut self) -> usize {
+            let mut pruned = 0;
+            self.instances.retain(|_, instance| {
+                let touched = std::mem::replace(&mut instance.touched, false);
+                if !touched {
+                    pruned += 1;
+                }
+                touched
+            });
```
bnaecker, line 507: This is a bit of a style nit, but you could avoid the variable `pruned` here by computing the size at the start, then after the call to `retain()`, and then returning their difference.

nexus/src/app/background/instance_watcher.rs
```diff
@@ -0,0 +1,635 @@
...
+    /// pair.
+    #[derive(Clone, Debug, Metric)]
+    struct InstanceChecks {
+        /// `true` if the instance state changed as a result of this check.
+        instance_updated: bool,
+        /// `true` if the VMM state changed as a result of this check.
+        vmm_updated: bool,
+        /// The number of successful checks for this instance and sled agent.
+        datum: Cumulative<u64>,
+    }
```
bnaecker, line 607: This metric structure seems a bit funky to me. I'm not sure I can put my finger on it just yet, so bear with me as I talk / write through this a bit.

As it stands, one would view this through the timeseries `instance_target:instance_checks`. The datum of that timeseries would count the total number of successful checks, with different timeseries for those where the instance or VMM was updated. (So 4 total timeseries.)

It seems like one might want to instead report the _state_ of the instance and VMM itself as part of the timeseries, or more concretely the count of times we successfully observed each state. This mirrors for example the vCPU usage, which reports the cumulative time in each of the tracked vCPU microstates. That way, I can ask questions like, "When was the instance running?", "How long was the instance stopped on this day?", or "Which instances are currently failed?".

I'm not sure how one would answer those questions with the knowledge of state _changes_ only. Also, if one really wants to know that, we can build methods for finding those points of change from the raw data itself. (Such a thing doesn't yet exist in OxQL, but I can squint and see how one might implement it.)

Concretely, I think we might prefer something like this:

```rust
#[derive(oximeter::Metric)]
struct InstanceState {
    // Stringified form of the same enum we report in the API
    state: String,
    // Count of times we observed this state
    datum: Cumulative<u64>,
}

#[derive(oximeter::Metric)]
struct VmmState {
    // Stringified form of the same enum we track internally about the VMM state
    state: String,
    // Count of times we observed this state
    datum: Cumulative<u64>,
}
```

Now that I think about it, a lot of this is assuming that this is _the place_ we report instance state. It's possible that we want Propolis to report these states directly, alongside the vCPU metrics it already publishes. If that is the intention, then most of what I've written here isn't relevant! I am still unclear on how one might use the boolean fields around the instance / VMM state changes, but it's much less important if the example questions I wrote earlier can be answered from another timeseries!

bnaecker (commented):
I've had to limit my review to the metrics-related portions for now, so sorry about that. I'm sure Sean can do a better job than me with the rest of it anyway :)

The overall structure looks great! I've some comments on the timeseries themselves (meaning, fields, use-cases), but very little to say about the code itself 🤘 Nice work, this looks really helpful.

nexus/src/app/background/instance_watcher.rs
hawkw: Ah, I didn't realize the name of the struct was reflected in the generated metrics, thanks! I'll change it.

hawkw: Out of curiosity, is it possible for other fields like the project/silo/sled serial/etc to be hydrated from the sled and instance UUIDs when the metrics are actually consumed? Or do we need to record everything we might ever care about now?

nexus/src/app/background/instance_watcher.rs
hawkw: We take the first element from the iterator by calling `batch.next()` before the `for` loop so that we can construct the initial sled-agent client with the sled address from the first item. I thought this seemed neater than making it an `Option` and having to handle the `None` case on every iteration.

I can add a comment explaining that!

nexus/src/app/background/instance_watcher.rs
hawkw: I think labeling the timeseries with the actual observed state of the instance and the VMM seems like a good idea! I'll change this code to do that.

nexus/src/app/background/instance_watcher.rs
bnaecker: >  I didn't realize the name of the struct was reflected in the generated metrics

I thought I was being clever when I wrote these macros, but this has been a source of confusion for a long time! I am planning some work in the near future that's mostly to support _updates_ to these schemas, and I will be making the name of these things a very explicit part of that.

> Out of curiosity, is it possible for other fields like the project/silo/sled serial/etc to be hydrated from the sled and instance UUIDs when the metrics are actually consumed?

I don't quite follow this. Could you expand on your question? Do you mean something like, looking up the project ID elsewhere at the time we query the metrics?

> Or do we need to record everything we might ever care about now?

Without any mechanisms for updating these at the moment, I would err on the side of more is better. These are quite cheap to store, we only have one copy of them in the database per _timeseries_, not per sample. (Though they are emitted in each sample.)

nexus/src/app/background/instance_watcher.rs
hawkw: Yup, good catch — I think this was an artifact of a previous version of this code where I was `retain`ing a nested set of `BTreeMap`s and doing it that way was more complicated. I'll change it.

nexus/src/app/background/instance_watcher.rs
hawkw: > > Out of curiosity, is it possible for other fields like the project/silo/sled serial/etc to be hydrated from the sled and instance UUIDs when the metrics are actually consumed?
> 
> I don't quite follow this. Could you expand on your question? Do you mean something like, looking up the project ID elsewhere at the time we query the metrics?

Yeah, that's what I was wondering about. My prior experience with metrics systems has mostly been with Prometheus, where it's possible for a query to join additional label fields with a timeseries, so the set of labels added to a particular timeseries need only be enough to uniquely identify that timeseries, and other labels can be hydrated at display time. On the other hand, the number of labels on a timeseries as it's recorded was expensive enough that we generally tried to avoid adding labels that could be joined in a query based on a unique key. If the constraints for Clickhouse are different, that's good to know!

nexus/src/app/background/instance_watcher.rs
hawkw: @bnaecker if we were to start recording metrics for instance and VMM states, would it make more sense to:

1. add the VMM's UUID as a field on the `VirtualMachine` target, so the queries would be `virtual_machine:instance_state` and `virtual_machine:vmm_state`, where the `VirtualMachine` target includes a VMM UUID (and thus would not be equal if an instance migrated from one VMM to another), or,
2. create separate `VirtualMachine` and `VirtualMachineManager` targets, where only the `VirtualMachineManager` target has the VMM UUID?

Really, I guess, what I'm asking is, is it possible to query a target based on only one of its fields (such as the instance ID)? If not, the current structure might not make sense.

nexus/src/app/background/instance_watcher.rs
```diff
@@ -0,0 +1,635 @@
...
+            let rsp = client.instance_get_state(&instance_id).await;
+            let state = match rsp {
+                Ok(rsp) => rsp.into_inner(),
+                Err(ClientError::ErrorResponse(rsp)) => {
+                    let status = rsp.status();
+                    if status == StatusCode::NOT_FOUND
+                        && rsp.as_ref().error_code.as_deref()
+                            == Some("NO_SUCH_INSTANCE")
+                    {
+                        slog::info!(opctx.log, "instance is wayyyyy gone");
```
smklein, line 88: rip

nexus/src/app/background/instance_watcher.rs
```diff
@@ -0,0 +1,635 @@
...
+                "not_found": not_found,
+                "unreachable_instances": unreachable_instances,
+                "sled_agent_errors": sled_agent_errors,
+                "check_errors": check_errors,
+                "pruned_instances": pruned,
+            })
+        }
+        .boxed()
+    }
+}
```
smklein, line 436: I know this was a lot of work, but I'm really grateful you structured things this way -- nice to see the built-in concurrency, and great to see we're plumbing out all this data. Thanks for making this task so observable!

nexus/src/app/background/instance_watcher.rs
```diff
@@ -0,0 +1,635 @@
...
+        datum: Cumulative<u64>,
+    }
+
+    /// The number of failed checks for an instance and sled agent pair.
+    #[derive(Clone, Debug, Metric)]
+    struct InstanceCheckFailures {
+        /// The reason why the check failed.
+        ///
+        /// # Note
+        /// This must always be generated from a `CheckFailure` enum.
```
smklein, line 615: (Here and below) I see this *happens* to come from a `.to_string()` call on `CheckFailure`, but why **must** it be this way? I wanna make sure I grok this, I'm not sure where that requirement is coming from?

smf/nexus/multi-sled/config-partial.toml
```diff
@@ -55,6 +55,8 @@ blueprints.period_secs_execute = 60
 sync_service_zone_nat.period_secs = 30
 switch_port_settings_manager.period_secs = 30
 region_replacement.period_secs = 30
+instance_watcher.period_secs = 30
+instance_watcher.max_retries = 5
```
smklein, line 59: Does this field still exist? I think no?

smklein (approved):
This looks great, thanks for putting in all the effort to make this accessible through omdb too!

smf/nexus/multi-sled/config-partial.toml
hawkw: ah, whoops, thanks for catching that!

nexus/src/app/background/instance_watcher.rs
hawkw: Thanks for saying so!

nexus/src/app/background/instance_watcher.rs
```diff
@@ -0,0 +1,635 @@
...
+                Ok(rsp) => rsp.into_inner(),
+                Err(ClientError::ErrorResponse(rsp)) => {
+                    let status = rsp.status();
+                    if status == StatusCode::NOT_FOUND
+                        && rsp.as_ref().error_code.as_deref()
+                            == Some("NO_SUCH_INSTANCE")
+                    {
+                        slog::info!(opctx.log, "instance is wayyyyy gone");
+                        // TODO(eliza): eventually, we should attempt to put the
+                        // instance in the `Failed` state here.
```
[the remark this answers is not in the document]
hawkw: I was a bit uncertain about whether it was safe to do that here. I was a bit worried about potentially racing with other Nexus replicas that may have received a different result from checking on the same instance and sled-agent, which I thought Greg's "instance updater lock" would resolve; I thought it was better to wait until I'd implemented that. For the other case where we update instance states, we've _received_ a `Generation` from the sled-agent, so that prevents racy updates, but here, we would have to increment it ourselves.

If you think it would be safe to do it now, though, I could be convinced to go ahead and add it here!

nexus/db-queries/src/db/datastore/sled_instance.rs
```diff
@@ -31,4 +36,37 @@ impl DataStore {
...
             .map_err(|e| public_error_from_diesel(e, ErrorHandler::Server))
     }
+
+    pub async fn sled_instance_list_by_sled_agent(
+        &self,
+        opctx: &OpContext,
+        collection: CollectionUuid,
+        pagparams: &DataPageParams<'_, DbTypedUuid<SledKind>>,
+    ) -> ListResultVec<(InvSledAgent, SledInstance)> {
+        // TODO(eliza): should probably paginate this?
```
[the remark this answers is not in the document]
hawkw: whoops, i thought i deleted it, my bad!

nexus/src/app/background/instance_watcher.rs
hawkw: > In any case, we should definitely add the project and silo IDs as fields to these timeseries. I might also recommend adding the sled's model / revision / serial as well, if they're easily accessible at the point we build type.

Currently, the CRDB table I'm querying to get the list of instances/sled-agents to check (the `sled_instance` table) has silo and project _names_, but it doesn't have their UUIDs. Do you think it's okay to use the name, or should I change the DB query to do more joins to get the UUIDs instead?

hawkw:
Okay, I've made a bunch of changes to the structure of the metrics based on @bnaecker's suggestions --- I would love another review!

Also, @bnaecker, if you have a chance, I've opened a smaller PR #5670 to add support for using `Cow<'static, str>` in Oximeter string fields. Since most of the "state" and "reason" fields are static string constants, I thought this was a good idea, since it would let us avoid making a separate copy of the actual string on the heap for each instance that has a given state. As Nexus is now storing at least O(_instances_) metrics in memory, I thought reducing the size of each individual metric was worthwhile --- let me know what you think?

nexus/src/app/background/instance_watcher.rs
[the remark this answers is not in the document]
smklein: I think that is a fear which is warranted -- we'll need to fix it eventually, but we don't *need* to fix it here. Let's punt to a different PR.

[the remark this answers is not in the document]
hawkw: Great, let's do that, then.

nexus/src/app/background/instance_watcher.rs
bnaecker: > it's possible for a query to join additional label fields with a timeseries

We current support inner joins in OxQL, but not outer joins, which you'd need to combine disparate fields from two timeseries. It shouldn't be too much trouble to do that, if we decide we need it. I can appreciate the desire for a compact representation, but I have consistently found that ClickHouse is able to handle enormous amounts of data. We deduplicate the fields of each timeseries on disk, so you have one copy of them for each timeseries. ClickHouse will further compress them, so the fields themselves are extremely cheap. I would err on the side of attaching all the fields you want, to the timeseries where you'd like them.

> Currently, the CRDB table I'm querying to get the list of instances/sled-agents to check (the sled_instance table) has silo and project names, but it doesn't have their UUIDs. Do you think it's okay to use the name, or should I change the DB query to do more joins to get the UUIDs instead?

I would recommend paying the cost of the extra lookup to get the UUIDs, at some point when you're fetching the data. The names can change at any time in response to a user API request, and I think we'd prefer to be unaware or unencumbered by that in the timeseries system. It is not clear how you'd follow an instance that's renamed, for example, since CRDB stores only the _current_ name, not its history.

hawkw: Okay, that makes sense, thank you! As an aside, though, it occurs to me that, because we can't do outer joins, there isn't presently a way for OxQL to turn a project/silo UUID into a user-understandable name?

nexus/db-queries/src/db/datastore/instance.rs
```diff
@@ -385,6 +389,44 @@ impl DataStore {
...
+    pub async fn instance_and_vmm_list_by_sled_agent(
+        &self,
+        opctx: &OpContext,
+        pagparams: &DataPageParams<'_, Uuid>,
+    ) -> ListResultVec<(Sled, Instance, Vmm, Project)> {
+        use crate::db::schema::{
+            instance::dsl as instance_dsl, project::dsl as project_dsl,
+            sled::dsl as sled_dsl, vmm::dsl as vmm_dsl,
+        };
+        opctx.authorize(authz::Action::Read, &authz::INVENTORY).await?;
```
smklein, line 401: I'm not necessarily suggesting a change, just pointing out something quirky here.

In `nexus/db-queries/src/authz/api_resources.rs`, INVENTORY is labelled as "Synthetic resource used for modeling access to low-level hardware inventory data". Does that fit the bill here?

Should this be authorizing read access to `FLEET`?

I'm not necessarily pushing for a change here because within the context of this change, I think the behavior will be largely the same, and "how we bucket our authz resources" also seems a little arbitrary

dev-tools/omdb/src/bin/omdb/nexus.rs
```diff
@@ -895,75 +895,58 @@ fn print_task_details(bgtask: &BackgroundTask, details: &serde_json::Value) {
...
-
-            /// number of unexpected errors returned by sled-agent
-            sled_agent_errors: usize,
+            /// number of stale instance metrics that were deleted
+            pruned_instances: usize,
 
-            /// number of instances for which the sled agent was unreachable
-            unreachable_instances: usize,
+            /// instance states from completed checks
+            instance_states: BTreeMap<String, usize>,
```
smklein, line 902: It's hard to tell from looking at this struct -- what are the String key values here, and in the other fields below?

smklein: I think this is a mapping of "stringified state" -> "count of instances in that state"? This seems fine, but it might be nice to document that here. I guess that's why you implement "Display" below?

It might be nice to include some words of caution that "any additional context added to states/failures/checks, which causes the struct to print differently, will also be counted as a distinct {state,check,failure}"

smklein (approved):
PR still LGTM, the refactors seem very reasonable!

nexus/db-queries/src/db/datastore/instance.rs
hawkw: Ah, yeah, this should probably be `FLEET`,  I think this is residue from an earlier version of this change where we started by reading sleds from the inventory. I'll update it!

dev-tools/omdb/src/bin/omdb/nexus.rs
hawkw: let me know what you think of the docs I've added in 70b8485b28a8b7047cdde8576bedf2afb8ea722b?

smklein: Looks great, thank you!

nexus/src/app/background/instance_watcher.rs
```diff
@@ -0,0 +1,593 @@
...
+    /// The number of unsuccessful checks for an instance and sled agent pair.
+    #[derive(Clone, Debug, Metric)]
+    struct IncompleteCheck {
+        /// The reason why the check was unsuccessful.
+        ///
+        /// This is generated from the [`Incomplete`] enum's `Display` implementation.
+        reason: String,
+        /// The number of failed checks for this instance and sled agent.
+        datum: Cumulative<u64>,
+    }
```
hawkw, line 592: Here is where I could use a little advice on the Oximeter idioms: it's occurred to me that we could potentially roll these into a single metric, now that the `Check` metric has separate "state" and "reason" fields. I wanted to make sure that the metric structure could represent both "check has marked the instance as failed due to this error" and "check has not made a determination about the instance due to an internal check error", so I've started by having two separate metrics `Check` and `IncompleteCheck`. However, it occurs to me that incomplete checks _could_, alternatively, be represented by a `state` string like `state: "unknown"` or `state: "incomplete"`, with the `reason` field indicating why we weren't able to do a check successfully. 

@bnaecker (or someone else who's good at Oximeter), what do you think? Is it better to emit more separate timeseries, to represent these, or combine them into one that can represent both cases?

nexus/src/app/background/instance_watcher.rs
```diff
@@ -0,0 +1,593 @@
...
+        /// If an instance was marked as "failed" due to a check failure, this
+        /// will be a string representation of the failure reason. Otherwise, if
+        /// the check was successful, this will be "success". Note that this may
+        /// be "success" even if the instance's state is "failed", which
+        /// indicates that we successfully queried the instance's state from the
+        /// sled-agent, and the *sled-agent* reported that the instance has
+        /// failed --- which is distinct from the instance watcher marking an
+        /// instance as failed due to a failed check.
+        reason: String,
+        /// this will be a string representation of the failure reason.
```
bnaecker, line 579: This looks like a misplace comment, I think?

nexus/src/app/background/instance_watcher.rs
```diff
@@ -0,0 +1,593 @@
...
+                            sled_agent_port: curr_sled.address().port(),
+                        };
+                        tasks.spawn(self.check_instance(opctx, &client, target));
+                    }
+                }
+            }
+
+            // All requests fired off! While we wait for them to come back,
+            // let's prune old instances.
+            let pruned = self.metrics.lock().unwrap().prune();
```
bnaecker, line 399: IIUC, here we prune instances based on whether they've been marked touched by the tasks we spawn above. Does that mean we're racing them? For example, suppose we have the following:

- We spawn the checking task, it marks the instance touched.
- The prune call swaps that with `false` so it's marked "untouched"
- On the next pass, we spawn the checking task, but before it runs...
- We get here, pruning that still-untouched instance.

I think the actual data we report to ClickHouse will be handled correctly, in either case, but it does mean there will appear to be many restarts of that cumulative timeseries. Logically, it seems like we should be using the database to tell us which instances to report data for: if we have it in our list from the database, keep it, otherwise prune it. That might not play well with the pagination you have here, but I think it could be made to work and avoids the race above.

bnaecker (commented):
I'm still responding to your main question, but GH decided I can't post single comments anymore, only reviews!

nexus/src/app/background/instance_watcher.rs
bnaecker: This is a good question, and though I've been thinking for a while, I'm very sorry to say I don't have a good answer for you :/

In general, I'd probably make this decision based on how one wants to query the data. If one wants to easily break out or group by a field, they should probably be part of the same timeseries. If that's rare, then I'd probably put these mostly-unrelated metrics (instance-state vs instance-state-fetch-failures) into two different timeseries.

My only hesitation with grouping them is that it sort of complicates the meaning of the data. IIUC, then the `reason` field is non-empty iff the state is `"unknown"`, i.e., we failed the check. So even the successful samples carry this now-empty field. To be clear, I'm not worried about resource usage here, just confusion about what the field means, and when it's valid. Maybe changing that field name to `failure_reason` is enough to mitigate that confusion, though. And this is a pretty non-specific worry -- I think it should mechanically work fine to do that, even if one wants to separately query for things like known states or failed checks. In either case, we can answer the same kinds of questions:

- what fraction of time was the instance in some state (say, stopped or running), over some period of time?
- how often did the instance change states?
- was the instance rebooted in this period?
- why did Nexus fail to determine the instance's state?

So I'd just say, choose what feels most natural to you at this point! I wish I had more concrete advice for you, but I guess we'll have to develop that together as we use it more :)

bnaecker (commented):
Ok, I'm very confused what's going on with GH's UI here. This was supposed to be a one-off comment answering your question. It appears I can only add it as part of a review now. Let me know if anything is unclear!

nexus/src/app/background/instance_watcher.rs
hawkw: After thinking about this for a minute, I don't actually believe this is racy.
The checking task itself does *not* actually mutate the metrics, it just returns
a value to *this* task, which updates the metric and sets the touched flag. This is the only task that
mutates the metrics, and there can only be one of these tasks executing at any
given time, as activating the background task borrows it mutably. And, we won't
activate the background task again until *all* in-flight checks have completed,
because this task waits for all tasks spawned on the `JoinSet` to complete
before this future completes. There would be
a potential race if the checking tasks were responsible for actually updating
the metrics, but they're not.

I think there *is* an issue here, though, but it's much simpler: the issue is
just that we should call `prune` *after* all tasks spawned on the `JoinSet` have
completed, rather than before.

nexus/src/app/background/instance_watcher.rs
hawkw: @bnaecker 
> IIUC, then the reason field is non-empty iff the state is `"unknown"`, i.e., we failed the check. So even the successful samples carry this now-empty field. To be clear, I'm not worried about resource usage here, just confusion about what the field means, and when it's valid. Maybe changing that field name to failure_reason is enough to mitigate that confusion, though.

So, it's a little bit more complex than that, actually. There are (currently) three possible outcomes for an instance check:
- The check was "successful", indicating that we got a state back from the sled-agent (which may be "failed", if the sled-agent has decided that the instance has failed)
- The check was a "failure", indicating that we got an _error_ from the sled-agent that affirmatively indicates that the instance has failed. Presently, this occurs when we get back an HTTP error indicating that the sled agent doesn't know that it it has the instance we expected it to know about, which means that the sled-agent has crashed and the instance is now lost, or if the sled-agent was totally unreachable.
- The check was "incomplete", indicating that it failed in a way that made it impossible for us to determine what was up with the instance. This includes HTTP client errors and other things that indicate the *nexus* process messed up.

In the first two cases, successful and failed checks, we will emit both a "state" field, representing the determination about the instance's state, and a "reason" field, indicating _why_ we made that determination, which will be "success" if the check succeeded with an affirmative response from the sled-agent[^1], or a representation of the error that caused the check to fail. In the incomplete case, we instead increment the `virtual_machine:incomplete_check` metric, indicating that we couldn't figure out what was going on. 

What I'm proposing here is representing the "incomplete check" case using the same `virtual_machine:check` metric, instead, by emitting "unknown" as the instance state.

bnaecker: Ok, thanks for clarifying. I still don't have a great answer, and thing both approaches have their merits. Keeping them separate seems reasonable because the second metric really tracks a failure to communicate with the sled agent for some reason, but complicates checking the instance state itself in that case (unknown). Merging them makes tracking the state simpler, but complicates the meaning of the metric itself (it includes information about communication with the sled agent).

I think I have a slight preference for merging them, as you're suggesting. However, I think if we do keep them separate, I would emit a metric for the instance state as "unknown" in addition to the metric describing why the check failed. That is, for every sample period we should have a new value for the state of the instance itself, even if it's "unknown".

nexus/src/app/background/instance_watcher.rs
bnaecker: > the checking task itself does not actually mutate the metrics

Ah, I didn't grok that, thanks.

> we should call prune after all tasks spawned on the JoinSet have completed

I think that makes sense. It seems logically equivalent to pruning any instances that are not returned from the database, right? If so, then that seems good to me.

nexus/src/app/background/instance_watcher.rs
hawkw: I think I'm going to go ahead and merge them, then, based on the rationale that if we emit a `state: "unknown"` for every incomplete check, then the `incomplete_check` metric isn't really capturing any information we don't already have.

bnaecker:
Thanks for all your patience on the back-and-forth on this PR! I've one other question, which I feel does not block integrating this, just something I'm curious about. How does the fact that each Nexus instance runs these checks impact the resulting metrics?

nexus/src/app/background/instance_watcher.rs
hawkw: Actually, upon further consideration, I think having a separate `incomplete_check` metric _is_ the right thing, because it captures additional information about another possible kind of incomplete check: one where we _do_ determine the instance's state, but fail to update the database for that instance. In that case, we would want to record the instance state we determined, but _also_ record that we didn't successfully update the instance in CRDB. I think that emitting a `state: "unknown"` for any check that _didn't_ successfully determine the state is a good idea though.

hawkw:
@bnaecker 

> Thanks for all your patience on the back-and-forth on this PR! I've one other question, which I feel does not block integrating this, just something I'm curious about. How does the fact that each Nexus instance runs these checks impact the resulting metrics?

Yeah, that's why I asked you whether Oximeter automatically included producer metadata or not --- I was, once again, suffering from Prometheus brain damage (prometheus always adds the IP and port of the scrape target, as well as metadata about how it was discovered). Commit 08e01adccc0c5a028c5dcca2554ee234ec76fc2c adds the UUID of the Nexus process to the metric target. We could also add the rack UUID if you think that's worth including now as well?

bnaecker:
@hawkw My current stance is that more fields are usually better :) That might change in the future, but they can always be ignored in the query and storing them is inexpensive. So if you think the rack / cluster ID is something you might want to know, then go for it!

hawkw:
> @hawkw My current stance is that more fields are usually better :) That might change in the future, but they can always be ignored in the query and storing them is inexpensive. So if you think the rack / cluster ID is something you might want to know, then go for it!

Given that it sounds like metric schema changes are painful, I'll throw it in just in case we need it! :)

bnaecker:
I wouldn't call them painful, so much as impossible :) I hope to change that in the coming months.

PR #5620, opened 2024-04-25
[omdb] show sled policy and state, allow application of filter

sunshowers:
I realized that we have this wonderful `SledFilter` enum lying around, and we
can just use it in omdb.

I decided to shorten "eligible-for-discretionary-services" to "discretionary",
which I hope communicates the same meaning but in a shorter manner.

Depends on #5615.


jgallagher (approved):
Nice!

PR #5622, opened 2024-04-25
[nexus] factor out instance management into free functions

hawkw:
Currently, the instance state management code in nexus is implemented as methods on the `Nexus` type. This is problematic, as these methods can only be called in HTTP endpoints, and not from background tasks or sagas. For example, as part of #5611, I needed to call the `Nexus::notify_instance_updated` method from a background task, which isn't possible with the current factoring. Therefore, this PR moves that method and its dependencies, as well as all the methods in the `instance_network` module into free functions that can be called without a `Nexus` reference.

The `Nexus` methods are left in place in order to avoid having to update all the callers as part of this PR, but they now just call into the free functions. The free functions take a few more arguments that are provided by the `&self` reference in the `Nexus` methods, such as the internal DNS `Resolver` and `opctx_alloc`, so the methods pass these in from `&self`, making them a bit more convenient to call when a `Nexus` reference _is_ available. For both of those reasons, I thought keeping them around was the right thing to do.

I had briefly considered moving this code to a new `nexus-instances` crate, but it seemed a bit iffy to me as `notify_instance_updated` also unregisters an Oximeter producer, and I didn't love the idea of moving Oximeter-related code to a `nexus-instances` crate --- perhaps it also ought to go in a `nexus-oximeter` crate, or something, but that felt like a rabbit hole that I was afraid I might just end up going down forever, so I'll save it for later.

If there are any functional changes as part of this PR, I've made a very bad mistake; this _should_ be a pure refactor...

### DIFF
--- a/nexus/src/app/instance.rs
+++ b/nexus/src/app/instance.rs
@@ -28,6 +28,7 @@ use nexus_db_queries::db::datastore::InstanceAndActiveVmm;
 use nexus_db_queries::db::identity::Resource;
 use nexus_db_queries::db::lookup;
 use nexus_db_queries::db::lookup::LookupPath;
+use nexus_db_queries::db::DataStore;
 use nexus_types::external_api::views;
 use omicron_common::api::external::http_pagination::PaginatedBy;
 use omicron_common::api::external::ByteCount;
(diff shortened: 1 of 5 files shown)


smklein (approved):
Thanks for refactoring this into a separate PR! I appreciate this mechanical stuff being easier to look at

PR #5623, opened 2024-04-25
Use SledUuid keys in Blueprint::blueprint_zones

jgallagher:
Should all be mechanical changes, and lets us drop a couple of workarounds (`all_omicron_zones_types()` and the `Cow`).


sunshowers (approved):
awesome, the dream!

PR #5626, opened 2024-04-25
inventory collector should not query sled agents on expunged sleds

davepacheco:
This came up under https://github.com/oxidecomputer/omicron/issues/5480#issuecomment-2073586139.

While it's useful for the collector to notice expunged sleds that are still visible to MGS, we probably ought not be trying to reach out to their sled agents.  (If nothing else, that sled's underlay IP should be re-useable, so this could concievably be a different sled now!)


nexus/src/app/background/inventory_collection.rs
```diff
@@ -168,45 +161,29 @@ async fn inventory_activate(
...
     datastore: &'a DataStore,
-    page_size: NonZeroU32,
 }
 
 impl<'a> nexus_inventory::SledAgentEnumerator for DbSledAgentEnumerator<'a> {
     fn list_sled_agents(
         &self,
     ) -> BoxFuture<'_, Result<Vec<String>, InventoryError>> {
         async {
-            let mut all_sleds = Vec::new();
```
davepacheco, line 179: Some time after this code was written, the same behavior was factored into `sled_list_all_batched()`, so I switched to that here.

sunshowers (approved):
Looks great, really straightforward. Thanks!

nexus/types/src/deployment/planning_input.rs
```diff
@@ -249,6 +249,9 @@ pub enum SledFilter {
     /// Sleds that are eligible for discretionary services.
     EligibleForDiscretionaryServices,
 
+    /// Sleds whose sled agents should be queried for inventory
+    InInventory,
```
andrewjstone, line 253: I initially misunderstood this value to mean sleds that are in inventory, and not ones that should be queried for it. I'm not sure if that's a reason to rename it to something more specific though. I was thinking something like, `QueryForInventory`

davepacheco: Yeah, I hear that.  I went back and forth on a few names and didn't like any of them.

- `ForInventory`, as in "this is the _filter_ used _for_ collecting inventory" but that felt weird -- the other names are not phrases that get attached to the phrase "sled filter"
- `QueryForInventory` made me a little worried someone might think these sleds are not in inventory at all since they're not being queried (but they can be in inventory if MGS can see them)
- `InInventory` (what I have) -- what you said
- `IncludedInInventory` -- same problem, it's misleading
- `QuerySledAgentDurringInventory` -- accurate, but a mouthful

I'm open to these or other ideas but I don't feel especially great about any of these.

sunshowers: I'm fine with `InInventory` personally, but what about `QueryDuringInventory`?

davepacheco: I think that's about the same as `QueryForInventory`.  The confusion I'm worried about there is: you have to know how inventory works (namely, that we also query MGS) to understand that even if something is not queried, it may appear in the inventory.  But this isn't a big deal.  And `InInventory` isn't really better.  I'll go ahead and use `QueryForInventory` for now.

PR #5634, opened 2024-04-26
Enable the SFTP subsystem in the switch zone ssh server

citrus-it:
We have historically accessed the rack's switch zone from systems running a version of OpenSSH from Helios 1.0, which uses the legacy scp/rcp protocol for `scp(1)` operations. More recent versions of OpenSSH default to using SFTP  and only fall back to the legacy protocol when the `scp -O` flag is used. SFTP is considered generally better and we should enable the subsystem for this SSH server instance.

### DIFF
--- a/wicket/zone-etc/ssh/sshd_config
+++ b/wicket/zone-etc/ssh/sshd_config
@@ -22,6 +22,11 @@ AllowAgentForwarding no
 AllowTcpForwarding no
 PrintMotd no
 
+# Enable the SFTP sub-system. This allows clients such as scp(1) to use
+# SFTP rather than the legacy scp/rcp protocol, which is deprecated. SFTP has
+# been the default protocol for OpenSSH's scp(1) since April 2022.
+Subsystem sftp internal-sftp
+
 #
 # Allow the user to pass through a limited set of environment variables for
 # locale and shell preferences:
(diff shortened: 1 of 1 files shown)


wicket/zone-etc/ssh/sshd_config
```diff
@@ -30,7 +30,11 @@ AcceptEnv LANG LC_* OXIDE_PREF_*
 
 AllowUsers wicket support
 
-# Allow "wicket" to log in without a password, and constrain it to the captive shell
+# Allow the use of the SFTP transport protocol
```
rmustacc, line 33: I think it might help to explain why as the what is somewhat straightforward given the command line. Even if not in the config file, perhaps mentioning it in the description of the issue at least might be useful.

rmustacc (commented):
Thanks for putting this together. I think in general I'm happy with us enabling this given the scp history and the fact that from an external observer perspective the net set of things you're able to do isn't changing terribly and we get additional ways to lock down sftp in the future.

wicket/zone-etc/ssh/sshd_config
citrus-it: I've added more to the configuration file and the PR description.

rmustacc (approved):

Issue #5625, opened 2024-04-25
Adding a previously-expunged sled leaves it expunged


jgallagher:
On an a4x2 testbed, I went through this sequence:

* Set up with three sleds (g0, g1, g3)
* Add sled g2
* Step through blueprints to give g2 NTP and crucible
* Use `omdb db reconfigurator-save`, `reconfigurator-cli`, and `omdb nexus blueprints import` to craft a blueprint that gives g2 a Nexus zone
* Confirm the Nexus zone comes online and is externally reachable
* Manually clean-slate the sled (disable sled-agent and mg-ddm, remove relevant files in `/pool/int/*`)
* `a4x2 hyperstop g2` to shut off the sled
* `omdb db nexus sleds expunge ...` to expunge the sled
* `a4x2 hyperstart g2` to bring the sled back online
* Reenable sled-agent and mg-ddm

At this point the sled shows up as uninitialized:

```
root@oxz_switch:~# omdb nexus sleds list-uninitialized
RACK_ID                              CUBBY SERIAL PART  REVISION
73d99599-d25c-41d6-a78b-246ceba80faa 2     g2     i86pc 0
```

And adding it appears to succeed:

```
root@oxz_switch:~# omdb -w nexus sleds add g2 i86pc
added sled g2 (i86pc)
```

However, after doing so, it still shows up in `list-uninitialized`:

```
root@oxz_switch:~# omdb nexus sleds list-uninitialized
RACK_ID                              CUBBY SERIAL PART  REVISION
73d99599-d25c-41d6-a78b-246ceba80faa 2     g2     i86pc 0
```

and we can even apparently add it again (again successfully):

```
root@oxz_switch:~# omdb -w nexus sleds add g2 i86pc
added sled g2 (i86pc)
```

but it remains in the `list-uninitialized` output.

We expect this to work, and the add should create a new `sled_id` for this same baseboard we've seen before. But looking in CRDB, we only have one row for g2 (the one created when we added it the first time), and it is still `expunged`:

```
root@[fd00:1122:3344:103::3]:32221/omicron> select id,time_created,time_modified,serial_number,sled_policy from sled;
                   id                  |         time_created          |         time_modified         | serial_number | sled_policy
---------------------------------------+-------------------------------+-------------------------------+---------------+--------------
[...]

PR #5637, opened 2024-04-26
Add `omdb expunge-testing fudge-serial <SLED_ID>`

jgallagher:
This subcommand should unblock looped sled add/expunge testing on madrid. After expunging, run this command to overwrite the serial numbers of the now-expunged sled, allowing it to be added again with its actual serial number.

This is obviously gross and should be removed once we solve #5625, which requires transitioning sleds to the decommissioned state.

There are some very primitive checks in place. Trying to fudge the serials for a sled that isn't expunged fails:

```
root@oxz_switch:~# ./omdb -w db expunge-testing fudge-serial 7bd6b802-e2d6-48bb-a53c-a7ee31f6e216
Error: serial number fudging only supported for expunged sleds; sled 7bd6b802-e2d6-48bb-a53c-a7ee31f6e216 has policy InService { provision_policy: Provisionable }
```

And trying to fudge the serials of the same sled twice fails:

```
root@oxz_switch:~# ./omdb -w db expunge-testing fudge-serial ac4a39a0-3dc5-4fc8-a7ad-5217af1b8f3d
Error: sled serial number (omdb-fudged-g2-0-1714162798) already appears fudged
```

### DIFF
--- a/dev-tools/omdb/src/bin/omdb/db.rs
+++ b/dev-tools/omdb/src/bin/omdb/db.rs
@@ -15,6 +15,7 @@
 // NOTE: emanates from Tabled macros
 #![allow(clippy::useless_vec)]
 
+use crate::check_allow_destructive::DestructiveOperationToken;
 use crate::Omdb;
 use anyhow::anyhow;
 use anyhow::bail;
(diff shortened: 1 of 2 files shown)


andrewjstone (approved):
Just gross enough to work!

jgallagher:
Closing this in favor of #5733 which should allow us to test this without this workaround

PR #5644, opened 2024-04-27
[nexus-types] more validation for PlanningInput

sunshowers:
(Re-do of #5541.)

While reconciling the PlanningInput with the blueprint, I was a bit concerned
about getting nonsensical inputs (e.g. multiple zones get the same IPs) and the
system going haywire. In particular, I was concerned about the Serialize and
Deserialize impls which were seemingly added for reconfigurator-cli. However,
they were liabilities that didn't check for internal invariants.

To address this, this PR introduces the notion of a `TriMap`. A `TriMap` is
a 1:1:1 map which can be looked up by any one of three keys. Instead of storing
just a regular map, the `PlanningInput` now stores a couple of `TriMap`
instances via an intermediate `OmicronZoneNetworkResources` struct.

A `TriMap` is implemented as a vector with three indexes. It's a pretty
straightforward implementation, and I've also added property-based tests
to ensure that a `TriMap` is always valid (including always deserializing
to a valid structure).

At the moment, a `TriMap` does not allow removing entries. If necessary,
removals can be implemented by just marking the entry as dead and removing
the keys from the indexes. This is fine for our use case, since `TriMap`s
aren't long-lived.

I've also factored out the omicron zone IP and NIC code into an
`OmicronZoneNetworkResources` struct. In the future, we'll make the
`BlueprintBuilder` use this struct, so that blueprints are validated
whenever they're being mutated. We can also add this to the general code
that validates blueprints.


nexus/types/src/deployment/planning_input.rs
```diff
@@ -5,34 +5,136 @@
...
+/// the add/remove sled use case.
+///
+/// The planning input has some internal invariants that code outside of this
+/// module can rely on. They include:
+///
+/// - Each Omicron zone has at most one external IP and at most one vNIC.
+/// - A given external IP or vNIC is only associated with a single Omicron
+///   zone.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct PlanningInput {
```
sunshowers, line 52: Decided to move this up-top because this file is called `planning_input.rs`, and generally I think the main part of a struct should live at the top. (Also I was getting really confused :) )

The only differences here are that we now have a `network_resources` field and method.

nexus/types/src/deployment/tri_map.rs
```diff
@@ -0,0 +1,509 @@
...
+    borrow::Borrow,
+    collections::{hash_map, BTreeSet, HashMap},
+    fmt,
+    hash::Hash,
+};
+
+use derive_where::derive_where;
+use serde::{Deserialize, Serialize, Serializer};
+
+/// An append-only 1:1:1 (trijective) map for three keys and a value.
```
sunshowers, line 15: This is append-only for now -- we may add the ability to remove entries in the future if really necessary. (I think at least at the moment it isn't necessary either here or in `BlueprintBuilder`.)

nexus/types/src/deployment/tri_map.rs
```diff
@@ -0,0 +1,509 @@
...
+    entries: Vec<T>,
+    // Invariant: the value (usize) in these maps are valid indexes into
+    // `entries`, and are a 1:1 mapping.
+    k1_to_entry: HashMap<T::K1, usize>,
+    k2_to_entry: HashMap<T::K2, usize>,
+    k3_to_entry: HashMap<T::K3, usize>,
+}
+
+// Note: Eq and PartialEq are not implemented for TriMap. Implementing them
+// would need to be done with care, because
```
jgallagher, line 33: This comment cuts off just when it's getting good. :)

nexus/types/src/deployment/network_resources.rs
```diff
@@ -0,0 +1,283 @@
...
+use thiserror::Error;
+
+/// Tracker and validator for network resources allocated to Omicron-managed zones.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct OmicronZoneNetworkResources {
+    /// external IPs allocated to Omicron zones
+    omicron_zone_external_ips: TriMap<OmicronZoneExternalIpEntry>,
+
+    /// vNICs allocated to Omicron zones
+    omicron_zone_nics: TriMap<OmicronZoneNicEntry>,
```
jgallagher, line 27: Would it make sense to squish these two maps together? I ended up combining external IP + NIC in the [BlueprintZoneType helper method](https://github.com/oxidecomputer/omicron/blob/bca1f2d63b70e71aef509b94e190e54c6eb3a6be/nexus/types/src/deployment/zone_type.rs#L36-L38) because all zone types either have both or neither. (Now that I ask that, that might require a 5-way map instead of a 3-way map? :grimacing:)

nexus/types/src/deployment/network_resources.rs
```diff
@@ -0,0 +1,283 @@
...
+    pub ip: OmicronZoneExternalIp,
+}
+
+/// Specification for the tri-map of Omicron zone external IPs.
+impl TriMapEntry for OmicronZoneExternalIpEntry {
+    type K1 = OmicronZoneUuid;
+    type K2 = ExternalIpUuid;
+
+    // Note: cannot use IpAddr here, because SNAT IPs can overlap as long as
+    // their port blocks are disjoint.
```
jgallagher, line 224: I think this is a pretty strong argument for your suggestion that we change SNAT configs to take an enum describing the block instead of a pair of `u16`s. We should no longer be able to construct SNAT configs that don't exactly match one of the port blocks, but this code is relying on that I think? (E.g., if we could still construct invalid port blocks that overlapped but weren't equal, that wouldn't be detected by this map's dup checks.)

jgallagher (approved):

nexus/types/src/deployment/network_resources.rs
sunshowers: That's right, yeah. Maybe I should fix that then.

nexus/types/src/deployment/network_resources.rs
sunshowers: Sure, I think this pattern scales up to any N.

nexus/types/src/deployment/tri_map.rs
sunshowers: Oof, meant to write something like:

```
// Note: Eq and PartialEq are not implemented for TriMap. Implementing them
// would need to be done with care, because TriMap is not semantically like an
// IndexMap: two maps are equivalent even if their entries are in a different
// order.
```

nexus/types/src/deployment/network_resources.rs
sunshowers: Ehh, I think for now I'll just rely on `SourceNatConfig` externally guaranteeing that the port blocks are disjoint.

nexus/types/src/deployment/network_resources.rs
sunshowers: Chatted with @jgallagher a bit more. While it is true that these are 1:1 in the blueprint:

* that might not be the case in the database itself
* within the planning input we're not depending on the 1:1 property across IPs and NICs right now
* the future need for this map is for expunged zone garbage collection. That doesn't require 1:1 across IPs and NICs.

So based on all this I'll stick with the pair of tri-maps and add a comment explaining this.

sunshowers: Added a comment.

nexus/types/src/deployment/tri_map.rs
sunshowers: Fixed.

Issue #5247, opened 2024-03-13
Tight coupling of "HTTP servers, as interface" to "Implementations" is painful


smklein:
This may be an issue related to the dropshot repo too, as Omicron isn't the sole user of this pattern.

This issue describes a process which is documented in our README, but which is still painful: https://github.com/oxidecomputer/omicron?tab=readme-ov-file#generated-service-clients-and-updating

# Background

- We use dropshot to generate a JSON OpenAPI spec, that is used to generate client crates
- When using dropshot, we write servers (see: `http_entrypoints.rs` throughout the codebase) which utilize dropshot macros and ingest types to define the specification.
- These endpoints are capable of emitting an openapi specification, **if and only if they can compile**. This often takes the from of an EXPECTORATE test, which spits out that aforementioned JSON file.

# When is this a problem

Mostly: When things don't compile, it can be exceptionally difficult to "regenerate the interface we want".

In the cleanest scenario, you'd make changes to your implementation *AND* interface together, then run the "openapi spec generation tests", then apply those changes to all clients which exist in the codebase. However, there are scenarios where this breaks:

1. Circular dependencies. For example, Nexus exposes an internal API to sled agent, and sled agent exposes an API to Nexus. This isn't necessarily bad! They both have reasons to communicate with each other. However, updating interactions between the two servers often involves re-running the JSON generation several times, then dealing with breakages, and then re-building. At an interface level, this is fine, but it's painful that "the rest of the crate (nexus or sled agent)'s implementation must also compile first, before the openapi test can run". This isn't too bad for adding new endpoints, but for altering interfaces that exist, it often means filling up a codebase with `todo!()` and commented-out code to get to a compiling state, then undoing all those changes immediately afterwards.
2. Merge conflicts. This forces you to deal with an openapi spec that is out-of-sync from main, but also potentially out-of-sync from your current tree. Again, this forces the same type of solution: find everyone using types that are generated by the openapi spec, and "force them to do the wrong thing, as long as it compiles".

# A better world?

It would be really nice to decouple the implementation of these interfaces from the interface declarations themselves.
For example, if the Sled Agent API was not actually part of the sled-agent crate, but rather, a separate "sled-agent-interface" crate, we could do the following:

- `EXPECTORATE=overwrite cargo nt -p sled-agent-interface` could re-generate the client bindings, regardless of whether or not "the rest of sled agent compiles"
[...]

PR #5653, opened 2024-04-29
[WIP] example of trait-based Dropshot server

sunshowers:
As discussed in https://github.com/oxidecomputer/omicron/issues/5247.

This PR converts the Nexus internal API to being a trait-based server, using the patched copy of Dropshot in https://github.com/oxidecomputer/dropshot/pull/984. Thought converting a non-trivial but still relatively simple API would be able to give us a better sense of what we want the eventual API to look like.

The main advantage is that this lets you generate the internal API without having to compile all of Nexus. This is not only faster, I believe it completely solves issues where there are circular dependencies between different APIs. (For example, currently there's a Nexus internal API -> sled-agent -> Nexus internal API dependency, and changing the APIs can be extremely painful. With this approach, the API crates don't depend on each other -- only the implementation crates do.)

Another advantage is that it's pretty easy to write multiple implementations of the same API. So, for example, the simulated sled-agent can implement the same trait as the real one.

### DIFF
--- a/nexus/internal-api/src/lib.rs
+++ b/nexus/internal-api/src/lib.rs
@@ -145,15 +145,14 @@ pub struct ProbePathParam {
 }
 
 #[dropshot_server]
-pub trait NexusInternalApi: Send + Sync + 'static {
+pub trait NexusInternalApi: Send + Sync + Sized + 'static {
     /// Return information about the given sled agent
     #[endpoint {
         method = GET,
         path = "/sled-agents/{sled_id}",
     }]
     async fn sled_agent_get(
-        &self,
-        rqctx: RequestContext<()>,
+        rqctx: RequestContext<Self>,
         path_params: Path<SledAgentPathParam>,
     ) -> Result<HttpResponseOk<SledAgentInfo>, HttpError>;
 
@@ -163,8 +162,7 @@ pub trait NexusInternalApi: Send + Sync + 'static {
         path = "/sled-agents/{sled_id}",
     }]
     async fn sled_agent_put(
-        &self,
-        rqctx: RequestContext<()>,
+        rqctx: RequestContext<Self>,
         path_params: Path<SledAgentPathParam>,
         body: TypedBody<SledAgentInfo>,
     ) -> Result<HttpResponseUpdatedNoContent, HttpError>;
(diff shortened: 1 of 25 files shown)


nexus/internal-api/src/lib.rs
```diff
@@ -0,0 +1,587 @@
...
+
+#[dropshot_server]
+pub trait NexusInternalApi: Send + Sync + Sized + 'static {
+    /// Return information about the given sled agent
+    #[endpoint {
+        method = GET,
+        path = "/sled-agents/{sled_id}",
+    }]
+    async fn sled_agent_get(
+        rqctx: RequestContext<Self>,
```
sunshowers, line 155: Static dispatch is nicer because you don't have to prove that the trait is object-safe -- you can also add non-endpoint methods to the trait.

nexus/src/internal_api/http_entrypoints.rs
```diff
@@ -51,1060 +68,779 @@ use omicron_common::api::internal::nexus::RepairProgress;
...
-        api.register(blueprint_target_set_enabled)?;
-        api.register(blueprint_regenerate)?;
-        api.register(blueprint_import)?;
-
-        api.register(sled_list_uninitialized)?;
-        api.register(sled_add)?;
-        api.register(sled_expunge)?;
-
-        api.register(probes_get)?;
-
```
sunshowers, line 113: This is now automatically handled!

nexus/src/internal_api/http_entrypoints.rs
```diff
@@ -51,1060 +68,779 @@ use omicron_common::api::internal::nexus::RepairProgress;
...
-    path = "/switch/{switch_id}",
-}]
-async fn switch_put(
-    rqctx: RequestContext<Arc<ServerContext>>,
-    path_params: Path<SwitchPathParam>,
-    body: TypedBody<SwitchPutRequest>,
-) -> Result<HttpResponseOk<SwitchPutResponse>, HttpError> {
-    let apictx = rqctx.context();
-    let handler = async {
+#[async_trait::async_trait]
```
sunshowers, line 85: For now I've chosen to use `async_trait`. But note that we don't need dynamic dispatch, so it should be possible to use the `-> impl Future` support in current versions of Rust.

Main thing to watch out for, however, is that we'd have to add `Send` bounds to all the returned Futures in the trait. That would change the types to `impl Future<...> + Send`, which would no longer make it possible to just write `async fn`. (Hopefully this is addressed soon.)

sunshowers: Removed async-trait -- works great!

sunshowers:
Closed in favor of #5844.

PR #5660, opened 2024-04-29
[nexus] Fix overzealous authz check on floating IP create

### LINKED ISSUE #5508: Floating IP create by non-admin fails when pool is specified
This permission check fails when the pool is specified. I think we should check a different permission, but it's conceivable we should instead change the permissions on this resource.

https://github.com/oxidecomputer/omicron/blob/9123e4b7cd7c270bc8c362821ca97f9e48a1d499/nexus/src/app/external_ip.rs#L114-L122

david-crespo:
Fixes #5508 

I think and hope that this is all that's required!

### DIFF
--- a/nexus/src/app/external_ip.rs
+++ b/nexus/src/app/external_ip.rs
@@ -114,7 +114,11 @@ impl super::Nexus {
         let pool = match pool {
             Some(pool) => Some(
                 self.ip_pool_lookup(opctx, &pool)?
-                    .lookup_for(authz::Action::Read)
+                    // every authenticated user has CreateChild on IP pools
+                    // because they need to be able to allocate IPs from them.
+                    // The check that the pool is linked to the current silo
+                    // happens inside allocate_floating_ip
+                    .lookup_for(authz::Action::CreateChild)
                     .await?
                     .0,
             ),
--- a/nexus/tests/integration_tests/external_ips.rs
+++ b/nexus/tests/integration_tests/external_ips.rs
@@ -22,8 +22,10 @@ use nexus_test_utils::resource_helpers::create_default_ip_pool;
 use nexus_test_utils::resource_helpers::create_floating_ip;
 use nexus_test_utils::resource_helpers::create_instance_with;
 use nexus_test_utils::resource_helpers::create_ip_pool;
+use nexus_test_utils::resource_helpers::create_local_user;
 use nexus_test_utils::resource_helpers::create_project;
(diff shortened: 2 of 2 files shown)


nexus/tests/integration_tests/external_ips.rs
```diff
@@ -249,6 +252,128 @@ async fn test_floating_ip_create(cptestctx: &ControlPlaneTestContext) {
...
+        },
+        pool: Some(NameOrId::Name("other-pool".parse().unwrap())),
+        ip: None,
+    };
+    let fip: views::FloatingIp =
+        NexusRequest::objects_post(client, &create_url, &body)
+            .authn_as(AuthnMode::SiloUser(user.id))
+            .execute_and_parse_unwrap()
+            .await;
+    assert_eq!(fip.identity.name.to_string(), "another-soda");
```
david-crespo, line 351: This request fails as expected without the fix.

nexus/src/app/external_ip.rs
```diff
@@ -114,7 +114,11 @@ impl super::Nexus {
         let pool = match pool {
             Some(pool) => Some(
                 self.ip_pool_lookup(opctx, &pool)?
-                    .lookup_for(authz::Action::Read)
+                    // every authenticated user has CreateChild on IP pools
+                    // because they need to be able to allocate IPs from them.
+                    // The check that the pool is linked to the current silo
+                    // happens inside allocate_floating_ip
+                    .lookup_for(authz::Action::CreateChild)
```
luqmana, line 121: Doing a `CreateChild` instead of a `Read` makes sense but `Datastore::allocate_floating_ip` also then does the same lookup? I'd say just drop this lookup and have `allocate_floating_ip` take the `Option<NameOrId>` pool parameter from the floating IP create args directly and only do the lookup there. That's basically what `allocate_probe_ephemeral_ip` does. (`allocate_instance_ephemeral_ip` should do the same in that case as well)

david-crespo: Will try it. I noticed it was redundant but wanted to avoid changing much.

david-crespo: I'm going to merge this to make sure the fix is in the next dogfood deploy, but I will do that in a followup.

nexus/tests/integration_tests/external_ips.rs
FelixMcFelix: I do really like this test for running through (default, linked, unlinked) pools on a more typical user. Thanks for putting it together.

PR #5661, opened 2024-04-29
Fix flaky `oximeter` reregistration test

### LINKED ISSUE #5645: test flake: oximeter-producer tests::test_producer_registration_task
Saw it here: https://buildomat.eng.oxide.computer/wg/0/details/01HWF3FQYGD9FXPMNK03VQWRMS/TKYHnLgC2vm5idnTWgPNdLA4jCSSluaR4V1HCKdLAjo5OYYC/01HWF3G1G84300N8BW78V8MJ5M#S5495

bnaecker:
- Fixes #5645
- Reworks the registration task to use `wait_for_condition`. This extends the duration of the test, but is much less flaky when the system is loaded.


oximeter/producer/src/lib.rs
```diff
@@ -481,60 +487,89 @@ mod tests {
         log
     }
 
+    // Re-registration interval for tests.
+    const INTERVAL: Duration = Duration::from_secs(1);
+
+    // Number of requests made to the fake Nexus server.
+    static N_REQUESTS_MADE: AtomicU32 = AtomicU32::new(0);
```
jgallagher, line 494: Tiny nit, but I'd maybe make this a field of `FakeNexus` (I guess behind an `Arc` so it can also be moved into the handler closure). As written this is fine, but no other test can interact with this correctly, I think, so it seems a little hazardous to leave it as a `static` that they _could_ touch.

oximeter/producer/src/lib.rs
```diff
@@ -481,60 +487,89 @@ mod tests {
...
-
-        // Drop the server to check its requests, rather than calling
-        // `verify_and_clear()`. It's possible for requests to be in-flight
-        // between when we called `verify_and_clear()` and when we exit this
-        // test, at which point the drop impl of the `TestServer` checks the
-        // expectations _again_. Let's just do the check by dropping here.
-        drop(fake_nexus);
+        const N_REQUESTS: u32 = 10;
+        const POLL_INTERVAL: Duration = Duration::from_millis(100);
+        const POLL_DURATION: Duration = Duration::from_secs(10);
```
jgallagher, line 559: Do we maybe need this to be longer? I'm not sure what was causing the old test to falke, but it was also 10x the reregistration interval, right?

jgallagher (approved):

oximeter/producer/src/lib.rs
bnaecker: Sure, we can make it longer. The old test flaked because we were basically racing with the registration task. Here, we're just waiting until we get at least the expected number, so the mechanism is quite different.

oximeter/producer/src/lib.rs
bnaecker: So, turns out we've written this nice, easy web server called `dropshot`, and that works much better :)

jgallagher (approved):
Nice! :+1:

PR #5663, opened 2024-04-29
Blueprints: store sled state

jgallagher:
This adds the `sled_state` map described in RFD 457 to `Blueprint`. The motivating driver here is decommissioning expunged sleds, which is on the path to addressing #5625, but this PR only adds the field - the planner carries forward states from its inputs, but otherwise makes no changes, and the executor doesn't read the new field at all.


nexus/db-queries/src/db/datastore/vpc.rs
```diff
@@ -1603,6 +1604,12 @@ mod tests {
         }
     }
 
+    fn sled_states_active(
```
davepacheco, line 1607: The repetition of this reinforces the idea that we could use a nice low-level builder interface that could take care of this.

nexus/types/src/deployment.rs
```diff
@@ -115,6 +116,9 @@ pub struct Blueprint {
     /// A map of sled id -> disks in use on each sled.
     pub blueprint_disks: BTreeMap<SledUuid, BlueprintPhysicalDisksConfig>,
 
+    /// A map of sled id -> state of the sled.
```
davepacheco, line 119:
```suggestion
    /// A map of sled id -> desired state of the sled
```

schema/crdb/blueprint-add-sled-state/up2.sql
```diff
@@ -0,0 +1,13 @@
+set local disallow_full_table_scans = off;
+
+-- All sleds are considered active (this migration is in support of
+-- transitioning active-but-expunged sleds to 'decommissioned'). We'll fill in
```
davepacheco, line 4:
```suggestion
-- At this point in history, all sleds are considered active and this table only exists to support
-- transitioning active-but-expunged sleds to 'decommissioned'). We'll fill in
```

(easy to misread as "this [database] migration is in support of ...", in which case it's surprising for the new state to always be 'active')

davepacheco (approved):
I am generally skeptical of landing interfaces without consumers, but I appreciate the desire to separate changes that are separable.  It might be useful to wait to land this until we've fleshed out the stuff that needs to use it but I defer to you on that!

nexus/db-queries/src/db/datastore/vpc.rs
jgallagher: I thought this would require some serious rework of `BlueprintBuilder` and the planner to make the builder less plan-y, but I was wrong (although I think there is still a lot of room for improvement there). I reworked this test to use BlueprintBuilder in [4949eac](https://github.com/oxidecomputer/omicron/pull/5663/commits/4949eac70c7ccfda81d5fd8dea8704f19e4a0f65) and it went pretty well. There were two spots (creating `bp2` and `bp4`) where it was easier to clone the parent blueprint and manually poke it than to go through the builder, but it still seems like a pretty significant cleanup overall.

andrewjstone (approved):

PR #5649, opened 2024-04-28
oxpop

rcgoodfellow:
- [x] Maghemite bump
- [x] Plumb new BGP features
- [x] Incorporate VLANs
- [x] Bootstore / early networking integration
- [x] Wicket / rss integration

PR #5666, opened 2024-04-30
[sled-agent-client] automatically derive Eq/Hash for a few structs

sunshowers:
These impls are currently correct, but https://github.com/oxidecomputer/omicron/pull/5649 was going to add new fields to these structs. That would make these impls incorrect in a somewhat terrifying fashion. Having `PartialEq` and `Hash` disagree for a type is a recipe for
difficult-to-debug bugs and security vulnerabilities. See https://users.rust-lang.org/t/what-hapens-if-hash-and-partialeq-dont-match-when-using-hashmap/98052/19 for some discussion.

In this case we can simply tell progenitor to derive these traits. But in the future, one way to handle this at compile time is to write:

```rust
let BgpConfig {
    asn,
    originate,
} = self;
asn.hash(hasher);
originate.hash(hasher);
```

If a new field is added, then this will fail loudly.

### DIFF
--- a/clients/sled-agent-client/src/lib.rs
+++ b/clients/sled-agent-client/src/lib.rs
@@ -12,9 +12,6 @@ use std::fmt;
 use std::hash::Hash;
 use std::net::IpAddr;
 use std::net::SocketAddr;
-use types::{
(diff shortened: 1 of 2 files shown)


rcgoodfellow (approved):
Thanks!

PR #5668, opened 2024-04-30
[wicketd] Pass RSS a /56 rack subnet instead of a raw IP address (/128)

### LINKED ISSUE #5665: Incorrect rack subnet after RSS?
Poking around a tangentially related test failure, I looked at the setup on madrid (which ran through RSS about a week ago), and the rack subnet there is a `/128` instead of a `/56`:

```
root@oxz_switch1:~# omdb db rack list
ID                                   INITIALIZED TUF_BASE_URL RACK_SUBNET
d12f892b-e723-4f1f-932d-ad9821770d44 true        -            fd00:1122:3344:100::/128
```
```
root@[fd00:1122:3344:104::4]:32221/omicron> select * from rack;
                   id                  |         time_created          |         time_modified         | initialized | tuf_base_url |     rack_subnet
---------------------------------------+-------------------------------+-------------------------------+-------------+--------------+-----------------------
  d12f892b-e723-4f1f-932d-ad9821770d44 | 2024-04-24 00:11:21.644128+00 | 2024-04-24 00:11:57.688143+00 |    true     | NULL         | fd00:1122:3344:100::
(1 row)
```

I think this is coming from what's in the bootstore:

```
BRM42220081 # curl http://[fd00:1122:3344:104::1]:12345/network-bootstore-config
{"generation":2,"schema_version":1,"body":{"ntp_servers":["ntp.eng.oxide.computer"],"rack_network_config":{"rack_subnet":"fd00:1122:3344:100::/128", ...
```

Everything looks fine on dogfood (the values for all the above end in `/56`), so this presumably broke at some intermediate point. I haven't had time yet to look at where this might be going wrong, and I also don't know whether this is actually _problematic_ (e.g., does everything that reads the rack subnet assume it's a `/56` and not actually care what's in the bootstore or database?). I'll investigate more tomorrow unless this rings bells for anyone else in the meantime.

jgallagher:
I think this should fix #5665. I checked a4x2 and it has a `/56`, so I think #5665 is specific to RSS when it's been run via wicket. I'll try this on madrid once a TUF repo is built.

I opened #5669 for the fact that our types allow this mistake; e.g., I think both https://github.com/oxidecomputer/omicron/blob/9c90e4b54694e8b4bec1884306d2626dcd062246/common/src/api/internal/shared.rs#L162 and https://github.com/oxidecomputer/omicron/blob/9c90e4b54694e8b4bec1884306d2626dcd062246/nexus/db-model/src/rack.rs#L19 are incorrect in that they allow any network size, and both should probably be `Ipv6Net<RACK_PREFIX>` instead. Fixing that is not trivial because at least the former is serialized in the bootstore.

### DIFF
--- a/wicketd/src/rss_config.rs
+++ b/wicketd/src/rss_config.rs
@@ -26,6 +26,9 @@ use gateway_client::types::SpType;
 use omicron_certificates::CertificateError;
 use omicron_common::address;
 use omicron_common::address::Ipv4Range;
+use omicron_common::address::Ipv6Subnet;
+use omicron_common::address::RACK_PREFIX;
+use once_cell::sync::Lazy;
 use sled_hardware_types::Baseboard;
 use slog::warn;
 use std::collections::BTreeSet;
@@ -38,8 +41,10 @@ use wicket_common::rack_setup::UserSpecifiedRackNetworkConfig;
 // TODO-correctness For now, we always use the same rack subnet when running
 // RSS. When we get to multirack, this will be wrong, but there are many other
 // RSS-related things that need to change then too.
-const RACK_SUBNET: Ipv6Addr =
-    Ipv6Addr::new(0xfd00, 0x1122, 0x3344, 0x0100, 0, 0, 0, 0);
+const RACK_SUBNET: Lazy<Ipv6Subnet<RACK_PREFIX>> = Lazy::new(|| {
+    let ip = Ipv6Addr::new(0xfd00, 0x1122, 0x3344, 0x0100, 0, 0, 0, 0);
+    Ipv6Subnet::new(ip)
+});
 
 const RECOVERY_SILO_NAME: &str = "recovery";
 const RECOVERY_SILO_USERNAME: &str = "recovery";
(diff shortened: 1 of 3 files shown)


andrewjstone (approved):

wicketd/src/rss_config.rs
```diff
@@ -498,7 +503,7 @@ fn validate_rack_network_config(
     // TODO Add more client side checks on `rack_network_config` contents?
 
     Ok(bootstrap_agent_client::types::RackNetworkConfigV1 {
-        rack_subnet: RACK_SUBNET.into(),
+        rack_subnet: RACK_SUBNET.net(),
```
sunshowers, line 506: Thanks for catching this.

The original code makes me really uncomfortable to be honest, I don't like the fact that there's a From impl here. (Feels like the opposite of misuse-resistant.) Is that something we can fix?

jgallagher:
Tested on madrid; looks good:

```
root@oxz_switch1:~# omdb db rack list
ID                                   INITIALIZED TUF_BASE_URL RACK_SUBNET
45e3bc39-e5e9-40e3-adea-a3b9cf70aa73 true        -            fd00:1122:3344:100::/56

PR #5670, opened 2024-04-30
[oximeter] use `Cow<'static, str>` for `FieldValue::String`

### LINKED ISSUE #5664: oximeter: want to be able to use `&'static str`s (and `Cow`s) in `derive(Metric)`
While working on adding Oximeter metrics in #5611, I noticed that `&'static str`s (or any `&str`s) cannot currently be used as fields in a struct deriving the `Metric` trait, as there is no `From<&&str>` impl for `oximeter::FieldType`. It would be nice if there was a way to use static strings as metric labels as well as heap-allocated ones. The motivating use case is that I have an enum of values that I would like to use for labeling metrics, and none of the variants have fields that require dynamic formatting. It seems like a shame to require formatting and allocating a `String` for each sample I want to emit, so being able to use a `&'static str` instead would be nice.

Similarly, I have another enum where only _one_ variant requires dynamic formatting. In that case, it would be nice to be able to use `Cow<'static, str>` as a `Metric` field, so that only the dynamically formatted variant requires a heap allocation.

I'm happy to open a PR adding this.

hawkw:
Fixes #5664.

Currently, Oximeter `FieldValue::String` fields are always represented
by a `String` value. This is somewhat unfortunate, as it requires
allocating a new `String` for every string field value that's emitted to
Oximeter, even when the string value is fixed at compile-time --- for
example, when an enum is used to generate a metric label with `&'static
str` values, they must be alloced into new `String`s to record the
metric.

This commit changes `FieldValue::String` from holding a `String` to
holding a `Cow<'static, str>`, so that static string constants need not
be heap-allocated.

Additionally, I've changed Oximeter's `self_stats` module to use
`Cow<'static, str>` to represent the field value generated by the
`FailureReason` enum. This way, a heap-allocated string is only
necessary for dynamically formatted variants (the `FailureReason::Other`
variant, which owns an HTTP `StatusCode`). For the
`FailureReason::Unreachable` and `FailureReason::Deserialization`
variants, we can just emit a `&'static str` without allocating.
Hopefully, this reduces resident memory for the collector process
somewhat when encountering a lot of failures.

### DIFF
--- a/nexus/db-queries/src/transaction_retry.rs
+++ b/nexus/db-queries/src/transaction_retry.rs
@@ -353,7 +353,9 @@ mod test {
             assert_eq!(
                 target_fields["name"].value,
                 FieldValue::String(
-                    "test_transaction_retry_produces_samples".to_string()
+                    "test_transaction_retry_produces_samples"
+                        .to_string()
+                        .into()
                 )
             );
 
--- a/oximeter/collector/src/self_stats.rs
+++ b/oximeter/collector/src/self_stats.rs
@@ -14,6 +14,7 @@ use oximeter::MetricsError;
 use oximeter::Sample;
 use oximeter::Target;
 use reqwest::StatusCode;
+use std::borrow::Cow;
 use std::collections::BTreeMap;
 use std::net::IpAddr;
 use std::time::Duration;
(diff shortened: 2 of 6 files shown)


bnaecker (approved):
Thanks for your patience while I was handling some new customer features. This all looks good to me, thanks!

PR #5671, opened 2024-04-30
[wicketd-client] use more replace directives

sunshowers:
I'm trying to change the format of these structs slightly as part of r8 work,
and the whole process is really, unnecessarily cumbersome. It doesn't spark
joy. This is an internal API (wicket <-> wicketd) that has no cross-version
concerns in any case.


wicket/src/cli/rack_setup/config_toml.rs
```diff
@@ -400,79 +388,7 @@ mod tests {
                 .collect(),
             external_dns_ips: value.external_dns_ips,
             ntp_servers: value.ntp_servers,
-            rack_network_config: InternalUserSpecifiedRackNetworkConfig {
```
sunshowers, line 403: There's simply no point to this...

andrewjstone (approved):

PR #5674, opened 2024-04-30
Clean up IP pool resolution from `NameOrId`

david-crespo:
Followup to #5660, specifically https://github.com/oxidecomputer/omicron/pull/5660#discussion_r1583752697. I didn't do it quite as suggested, basically because I think datastore functions being aware of `NameOrId` is a bad thing as that is, to me, an API concept that we should have resolved to something real by the time we get to the datastore functions. This is still a little odd and a little redundant, but it is cleaner and more consistent.

In each case (floating IP create, probe create, instance ephemeral IP allocated) we are doing the following:

1. In the Nexus layer, turn `Option<NameOrId>` into `Option<authz::IpPool>` (this is a DB query of course)
2. In the datastore, call the new `resolve_pool_for_allocation`, which
  a. If a pool is specified, makes sure it's linked to the current silo, otherwise 404
  b. If a pool is not specified, get the default pool for the current silo
  c. Either way, make sure we have `CreateChild` on the pool (we always do if we're authenticated at all, but that could change in the future)
3. At this point we definitely have an `authz::IpPool` and can get on with whatever allocation

### DIFF
--- a/nexus/db-queries/src/db/datastore/external_ip.rs
+++ b/nexus/db-queries/src/db/datastore/external_ip.rs
@@ -140,33 +140,9 @@ impl DataStore {
         // - At most MAX external IPs per instance
         // Naturally, we now *need* to destroy the ephemeral IP if the newly alloc'd
         // IP was not attached, including on idempotent success.
-        let pool = match pool {
-            Some(authz_pool) => {
-                let (.., pool) = LookupPath::new(opctx, &self)
-                    .ip_pool_id(authz_pool.id())
-                    // any authenticated user can CreateChild on an IP pool. this is
-                    // meant to represent allocating an IP
-                    .fetch_for(authz::Action::CreateChild)
-                    .await?;
 
-                // If this pool is not linked to the current silo, 404
-                // As name resolution happens one layer up, we need to use the *original*
-                // authz Pool.
-                if self.ip_pool_fetch_link(opctx, pool.id()).await.is_err() {
-                    return Err(authz_pool.not_found());
-                }
-
-                pool
-            }
-            // If no name given, use the default logic
-            None => {
-                let (.., pool) = self.ip_pools_fetch_default(&opctx).await?;
-                pool
-            }
-        };
-
-        let pool_id = pool.identity.id;
-        let data = IncompleteExternalIp::for_ephemeral(ip_id, pool_id);
+        let authz_pool = self.resolve_pool_for_allocation(&opctx, pool).await?;
+        let data = IncompleteExternalIp::for_ephemeral(ip_id, authz_pool.id());
 
         // We might not be able to acquire a new IP, but in the event of an
         // idempotent or double attach this failure is allowed.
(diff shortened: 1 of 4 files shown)


nexus/db-queries/src/db/datastore/external_ip.rs
```diff
@@ -50,7 +50,6 @@ use omicron_common::api::external::Error;
 use omicron_common::api::external::IdentityMetadataCreateParams;
 use omicron_common::api::external::ListResultVec;
 use omicron_common::api::external::LookupResult;
-use omicron_common::api::external::NameOrId;
```
david-crespo, line 53: getting this out of here is a good thing

nexus/db-queries/src/db/datastore/probe.rs
```diff
@@ -15,7 +15,6 @@ use diesel::{ExpressionMethods, QueryDsl, SelectableHelper};
 use nexus_db_model::IncompleteNetworkInterface;
 use nexus_db_model::Probe;
 use nexus_db_model::VpcSubnet;
-use nexus_types::external_api::params;
```
david-crespo, line 18: getting this out of here is a good thing

nexus/db-queries/src/db/datastore/external_ip.rs
```diff
@@ -239,29 +222,7 @@ impl DataStore {
...
         let ip_id = Uuid::new_v4();
 
-        // This implements the same pattern as in `allocate_instance_ephemeral_ip` to
-        // check that a chosen pool is valid from within the current silo.
-        let pool = match pool {
-            Some(authz_pool) => {
-                let (.., pool) = LookupPath::new(opctx, &self)
-                    .ip_pool_id(authz_pool.id())
-                    .fetch_for(authz::Action::CreateChild)
-                    .await?;
```
david-crespo, line 249: this fetch was not necessary at all — replaced by `opctx.authorize` call in `resolve_pool_for_allocation`

nexus/src/app/probe.rs
```diff
@@ -60,9 +60,22 @@ impl super::Nexus {
...
+        // resolve NameOrId into authz::IpPool
+        let pool = match &new_probe_params.ip_pool {
+            Some(pool) => Some(
+                self.ip_pool_lookup(opctx, &pool)?
+                    .lookup_for(authz::Action::CreateChild)
+                    .await?
+                    .0,
+            ),
+            None => None,
+        };
```
david-crespo, line 72: While overall I think this PR is a net improvement already, I might try to roll up this logic further with what's in `resolve_pool_for_allocation` because it is awkward to split up this part of the resolution of the pool from the rest.

nexus/db-queries/src/db/datastore/external_ip.rs
```diff
@@ -84,33 +83,14 @@ impl DataStore {
         opctx: &OpContext,
         ip_id: Uuid,
         probe_id: Uuid,
-        pool_name: Option<NameOrId>,
+        pool: Option<authz::IpPool>,
```
luqmana, line 86: Mind giving `allocate_instance_snat_ip` similar treatment so we're consistent across everything here

luqmana (approved):
Thanks for cleaning that up! LGTM w/ one suggestion

nexus/db-queries/src/db/datastore/external_ip.rs
david-crespo: Ah yes. I didn’t look hard enough.

PR #5675, opened 2024-04-30
Reject add sled requests for sleds that already exist

jgallagher:
I ran into this as a part of #5625, where adding a previously-expunged sled appeared to succeed, but didn't actually add anything new.

Today if we try to add a sled that is already _running_, we get a 500, because Nexus fails when it tries to tell the sled-agent to start. But with this PR, we fail earlier: adding a sled that already has a subnet allocation fails before we even try to talk to the sled-agent, because it means someone has already added this sled:

```
root@oxz_switch:~# omdb -w nexus sleds add g2 i86pc
added sled g2 (i86pc): 90413e40-8139-43b4-9081-365dab6e5579
root@oxz_switch:~# omdb -w nexus sleds add g2 i86pc
Error: adding sled

Caused by:
    Error Response: status: 400 Bad Request; headers: {"content-type": "application/json", "x-request-id": "9eb95a9f-3fe0-4f75-8846-13490b95500e", "content-length": "188", "date": "Tue, 30 Apr 2024 20:54:49 GMT"}; value: Error { error_code: Some("ObjectAlreadyExists"), message: "already exists: sled \"g2 / i86pc (90413e40-8139-43b4-9081-365dab6e5579)\"", request_id: "9eb95a9f-3fe0-4f75-8846-13490b95500e" }
```

This does change the external API slightly (204 -> 201 created, and we now return the ID), but I think (?) that's probably fine since we have no real consumers of that yet.

### DIFF
--- a/nexus/db-model/src/sled_underlay_subnet_allocation.rs
+++ b/nexus/db-model/src/sled_underlay_subnet_allocation.rs
@@ -3,6 +3,8 @@
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 use crate::schema::sled_underlay_subnet_allocation;
+use crate::typed_uuid::DbTypedUuid;
+use omicron_uuid_kinds::SledKind;
 use uuid::Uuid;
 
 /// Underlay allocation for a sled added to an initialized rack
@@ -10,7 +12,7 @@ use uuid::Uuid;
 #[diesel(table_name = sled_underlay_subnet_allocation)]
 pub struct SledUnderlaySubnetAllocation {
     pub rack_id: Uuid,
-    pub sled_id: Uuid,
+    pub sled_id: DbTypedUuid<SledKind>,
     pub subnet_octet: i16,
     pub hw_baseboard_id: Uuid,
 }
--- a/nexus/test-utils/src/lib.rs
+++ b/nexus/test-utils/src/lib.rs
@@ -61,6 +61,9 @@ use omicron_uuid_kinds::ZpoolUuid;
 use oximeter_collector::Oximeter;
 use oximeter_producer::LogConfig;
 use oximeter_producer::Server as ProducerServer;
+use sled_agent_client::types::EarlyNetworkConfig;
+use sled_agent_client::types::EarlyNetworkConfigBody;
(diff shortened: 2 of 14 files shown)


nexus/db-model/src/sled_underlay_subnet_allocation.rs
```diff
@@ -3,14 +3,16 @@
...
+use omicron_uuid_kinds::SledKind;
 use uuid::Uuid;
 
 /// Underlay allocation for a sled added to an initialized rack
 #[derive(Queryable, Insertable, Debug, Clone, Selectable)]
 #[diesel(table_name = sled_underlay_subnet_allocation)]
 pub struct SledUnderlaySubnetAllocation {
     pub rack_id: Uuid,
-    pub sled_id: Uuid,
+    pub sled_id: DbTypedUuid<SledKind>,
```
andrewjstone, line 15: Nit: Can you use the `SledUuid` alias here?

nexus/test-utils/src/lib.rs
```diff
@@ -911,6 +914,26 @@ impl<'a, N: NexusServer> ControlPlaneTestContextBuilder<'a, N> {
             })
             .await
             .expect("Failed to configure sled agent with our zones");
+        client
```
andrewjstone, line 917: Why are these bootstore related changes in here?

andrewjstone (approved):
Thanks for the fix!

nexus/db-model/src/sled_underlay_subnet_allocation.rs
andrewjstone: Ah, got it. Thanks.

nexus/test-utils/src/lib.rs
andrewjstone: Thank you for your sufferi^W^Wfix.

andrewjstone: Hmm, I'm thinking that we probably shouldn't read this from the bootstore. I don't know why we wouldn't use what's in `rack_initialize`. This seems like a bug waiting to happen, although I also don't see how that bug would arise exactly, given what's in the bootstore and rack_initialize should be the same.

PR #5678, opened 2024-05-01
[nexus] SNAT IP comes from same pool as ephemeral IP

### LINKED ISSUE #5043: SNAT IP should come from chosen pool instead of default pool
Right now the SNAT IP always comes from the default pool for the current silo, even when the user has specified they want an external IP from a different pool. We want to change the following logic to use the specified pool instead.

That change is simple, but there’s another wrinkle: what to do about existing instances where the SNAT IP and ephemeral IP come from different pools. This could lead to confusion later when an operator tries to delete an IP range with outstanding IPs. There’s a good chance there aren’t any such instances because nearly all instances are created with the default pool. But it might be worth checking somehow, just so we know. Longer term we will want (and have plans to build) ways of telling exactly where allocated IPs actually are.

https://github.com/oxidecomputer/omicron/blob/ce7c8d3c179b34fb1b1e19d91902fd7e59d05e62/nexus/src/app/sagas/instance_create.rs#L646-L655

david-crespo:
Closes #5043

Like I said [here](https://github.com/oxidecomputer/omicron/issues/5043#issuecomment-2087903321), this is arguably a little strange when you think about it: why should the IP pool for the ephemeral IP also govern where the SNAT IP comes from? I'll definitely wait for confirmation from a networking knower that this is actually the desired behavior.

### DIFF
--- a/nexus/db-queries/src/db/queries/external_ip.rs
+++ b/nexus/db-queries/src/db/queries/external_ip.rs
@@ -1038,15 +1038,6 @@ mod tests {
             instance_id
         }
 
-        async fn default_pool_id(&self) -> Uuid {
-            let (.., pool) = self
-                .db_datastore
-                .ip_pools_fetch_default(&self.opctx)
-                .await
-                .expect("Failed to lookup default ip pool");
-            pool.identity.id
-        }
-
         async fn success(mut self) {
             self.db.cleanup().await.unwrap();
             self.logctx.cleanup_successful();
@@ -1075,7 +1066,7 @@ mod tests {
                     &context.opctx,
                     id,
                     instance_id,
-                    context.default_pool_id().await,
+                    None,
                 )
                 .await
                 .expect("Failed to allocate instance external IP address");
@@ -1092,7 +1083,7 @@ mod tests {
                 &context.opctx,
                 Uuid::new_v4(),
                 instance_id,
-                context.default_pool_id().await,
+                None,
             )
             .await
             .expect_err(
@@ -1148,7 +1139,7 @@ mod tests {
                 &context.opctx,
                 Uuid::new_v4(),
                 instance_id,
-                context.default_pool_id().await,
+                None,
             )
             .await;
         assert!(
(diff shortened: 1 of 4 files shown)


nexus/db-queries/src/db/queries/external_ip.rs
```diff
@@ -1075,7 +1066,7 @@ mod tests {
                     &context.opctx,
                     id,
                     instance_id,
-                    context.default_pool_id().await,
+                    None,
```
david-crespo, line 1069: All of these work as expected when given `None` to make it keep choosing the default pool, but there should probably now be a test for a non-`None` value.

nexus/tests/integration_tests/instances.rs
```diff
@@ -4175,7 +4172,10 @@ async fn test_instance_ephemeral_ip_no_default_pool_error(
     };
     let error =
         object_create_error(client, &url, &body, StatusCode::NOT_FOUND).await;
-    assert_eq!(error.message, msg);
+    assert_eq!(
+        error.message,
+        "not found: ip-pool with name \"nonexistent-pool\""
+    );
```
david-crespo, line 4178: This used to complain about the default pool despite another pool being specified because the SNAT IP came from the default pool, and that failed before any ephemeral IP stuff came up.

nexus/tests/integration_tests/instances.rs
```diff
@@ -3958,13 +3958,10 @@ async fn test_instance_ephemeral_ip_from_correct_pool(
...
 
-    // SNAT comes from default pool, but count does not change because
-    // SNAT IPs can be shared. https://github.com/oxidecomputer/omicron/issues/5043
-    // is about getting SNAT IP from specified pool instead of default.
     assert_ip_pool_utilization(client, "pool1", 2, 5, 0, 0).await;
 
-    // ephemeral IP comes from specified pool
-    assert_ip_pool_utilization(client, "pool2", 1, 5, 0, 0).await;
+    // ephemeral IP and SNAT IP come from specified pool, so +2
+    assert_ip_pool_utilization(client, "pool2", 2, 5, 0, 0).await;
```
david-crespo, line 3964: This is what is supposed to happen, so that's good.

PR #5679, opened 2024-05-01
Bump gateway-messages to expose stage0 information

labbott:
This is a dependency for some upcoming work to be able update stage0 (RoT bootloader)

This change exposes the stage0 information

### DIFF
--- a/gateway/src/http_entrypoints.rs
+++ b/gateway/src/http_entrypoints.rs
@@ -72,9 +72,37 @@ pub struct SpState {
     Serialize,
     JsonSchema,
 )]
+#[serde(rename_all = "snake_case")]
+pub enum RotImageError {
+    Unchecked,
+    FirstPageErased,
+    PartiallyProgrammed,
+    InvalidLength,
+    HeaderNotProgrammed,
+    BootloaderTooSmall,
+    BadMagic,
+    HeaderImageSize,
+    UnalignedLength,
+    UnsupportedType,
+    ResetVectorNotThumb2,
+    ResetVector,
+    Signature,
+}
+
+#[derive(
+    Debug,
+    Clone,
+    PartialEq,
+    Eq,
+    PartialOrd,
+    Ord,
+    Deserialize,
+    Serialize,
+    JsonSchema,
+)]
 #[serde(tag = "state", rename_all = "snake_case")]
 pub enum RotState {
-    Enabled {
+    V2 {
         active: RotSlot,
         persistent_boot_preference: RotSlot,
         pending_persistent_boot_preference: Option<RotSlot>,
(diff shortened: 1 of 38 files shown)


jgallagher (approved):

gateway/src/http_entrypoints.rs
```diff
@@ -1224,6 +1290,40 @@ async fn sp_rot_cfpa_get(
     Ok(HttpResponseOk(RotCfpa { base64_data, slot }))
 }
 
+/// Read the RoT boot state from a root of trust
```
labbott, line 1293: This gets called implicitly as part of the sp state. I had this exposed as part of an earlier draft and kept it around. Might be better to remove if it's not being used?

nexus/db-model/src/inventory.rs
```diff
@@ -532,6 +661,13 @@ pub struct InvRootOfTrust {
     pub slot_boot_pref_persistent_pending: Option<HwRotSlot>,
     pub slot_a_sha3_256: Option<String>,
     pub slot_b_sha3_256: Option<String>,
+    pub stage0_fwid: Option<String>,
+    pub stage0next_fwid: Option<String>,
+
+    pub slot_a_status: Option<RotImageError>,
+    pub slot_b_status: Option<RotImageError>,
+    pub stage0_status: Option<RotImageError>,
+    pub stage0next_status: Option<RotImageError>,
```
jgallagher, line 670: Small nit - I'm not sure about naming these fields `*_status` when they can only contain errors. Is a value of `None` good? Would this be clearer either as `*_error` or by keeping `*_status` but changing the types to `Result<(), RotImageError>`?

sp-sim/Cargo.toml
```diff
@@ -4,6 +4,10 @@ version = "0.1.0"
 edition = "2021"
 license = "MPL-2.0"
 
+[features]
+old-state = []
+no-caboose = []
```
jgallagher, line 9: Do these need to be cargo features? I worry that they won't get tested as thoroughly or used as frequently if they aren't the default.

jgallagher (approved):
LGTM - just a few minor clarifications / questions that could be deferred

gateway/src/http_entrypoints.rs
labbott: I made a first pass attempt and it was kind of a mess because of backwards compatibility for checking where the RoT info lives. I do think it's the right approach and I may see about doing that again with more of the wicket changes.

nexus/db-model/src/inventory.rs
labbott: The actual gateway message does have this as `Result<(), ImageError>` but iirc I ran into some weirdness with trying to get that into the database so I made it an Option instead. I can change the name to make it clear it's only an error.

sp-sim/Cargo.toml
labbott: These are intended for testing some of the backwards compatibility state we need to support.

jgallagher: Could they be runtime / config parameters instead? (I'm thinking about something like a4x2, where changing a cargo feature requires a rebuild + redeploy, but changing a config file could be done without that lap.)

labbott: Good suggestion, I converted them over to take an option in the config file

PR #5681, opened 2024-05-01
Update propolis version

pfmooney:
Included changes:
Generate SMBIOS tables guest consumption
use for_internal_error for 500 error (#692)

### DIFF
--- a/package-manifest.toml
+++ b/package-manifest.toml
@@ -517,10 +517,10 @@ service_name = "propolis-server"
 only_for_targets.image = "standard"
 source.type = "prebuilt"
 source.repo = "propolis"
-source.commit = "27e2789d381c0fcc237fbe30cec2ec66bd750c12"
+source.commit = "0b71410d3a12045d34fa1c6c1d9ba1d8dd652564"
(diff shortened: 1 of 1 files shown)


leftwo (commented):
Do you also need to update `Cargo.toml` as well?  So Omicron will compile with
the same git rev as the package will have?  I don't think there is any
actual required changes, but I know I always bump both files.

pfmooney:
There have been no changes to the external API which would be reflected in what `propolis-client` exposes.  If I update those git revs, it drags in a bunch of other changes via `cargo update`.

leftwo (approved):
I do now believe that there is another Crucible bump coming, but that does not need to hold this back.

pfmooney:
> I do now believe that there is another Crucible bump coming, but that does not need to hold this back.

If there's a crucible bump coming, I'll just defer to that, rather than tying up the CI queue for another build.

leftwo:
The crucible bump was my mis-understanding, it was just new work to crucible main, not for release 8. So this can go go go!

Issue #4331, opened 2023-10-24
all disk deletes hang while Crucible downstairs is unreachable


augustuswm:
When the `disk_delete` saga runs, one of its responsibilities is to clean up regions that are freed as a result of a volume being deleted (`svd_delete_freed_crucible_regions`). This uses the following query to find regions that need to be cleaned up:

```sql
select * from region
inner join volume on region.volume_id = volume.id
inner join dataset on region.dataset_id = dataset.id
left join region_snapshot on region_snapshot.region_id = region.id or region_snapshot.dataset_id = dataset.id
where (region_snapshot.volume_references = 0 or region_snapshot.volume_references is null)
and not volume.time_deleted is null;
```

The results of this looks something like (keys redacted):
```
                   id                  |         time_created          |         time_modified         |              dataset_id              |              volume_id               | block_size | blocks_per_extent | extent_count |                  id                  |         time_created         |        time_modified         |         time_deleted          | rcgen |                                                                                                                                                                                                                     data                                                                                                                                                                                                                     |                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       resources_to_clean_up                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        |                  id                  |         time_created          |         time_modified         | time_deleted | rcgen |               pool_id                |          ip           | port  |   kind   |  size_used  | dataset_id | region_id | snapshot_id | snapshot_addr | volume_references
[...]

PR #5683, opened 2024-05-01
Implement region replacement for Volumes

### LINKED ISSUE #3886: Need API and support in Omicron to replace a crucible downstairs.
Crucible upstairs supports replacement of a specific downstairs.
propolis-server support is in progress.
Once Propolis server supports it, we need to add that support in Omicron.

### LINKED ISSUE #5191: Spin up an upstairs to handle downstairs replacement if no propolis is available
If we require a downstairs replacement in a disk that has no instance attached to it, then
we need to spin up an upstairs (a pantry instance?) to attach to the downstairs and handle the replacement.

This should take a VCR with the new downstairs already attached (because the old one may
be gone), then do the reconciliation to make all three downstairs consistent, then exit.

Care will need to be taken to make the transition between the two downstairs, and made to
happen in a way that:
* The disk/volume has recorded that a downstairs is being replaced (so no others can be replaced)
* The operations will recover if interrupted while in progress.
* We don't attach this disk to an instance till it is completed.
* On completion, we can destroy the old downstairs if it still exists and is present.

jmpesp:
When a disk is expunged, any region that was on that disk is assumed to be gone. A single disk expungement can put many Volumes into degraded states, as one of the three mirrors of a region set is now gone. Volumes that are degraded in this way remain degraded until a new region is swapped in, and the Upstairs performs the necessary repair operation (either through a Live Repair or Reconciliation). Nexus can only initiate these repairs - it does not participate in them, instead requesting that a Crucible Upstairs perform the repair.

These repair operations can only be done by an Upstairs running as part of an activated Volume: either Nexus has to send this Volume to a Pantry and repair it there, or Nexus has to talk to a propolis that has that active Volume. Further complicating things is that the Volumes in question can be activated and deactivated as a result of user action, namely starting and stopping Instances. This will interrupt any on-going repair. This is ok! Both operations support being interrupted, but as a result it's then Nexus' job to continually monitor these repair operations and initiate further operations if the current one is interrupted.

Nexus starts by creating region replacement requests, either manually or as a result of disk expungement. These region replacement requests go through the following states:

        Requested   <--
                      |
            |         |
            v         |
                      |
        Allocating  --

            |
            v

         Running    <--
                      |
            |         |
            v         |
                      |
         Driving    --

            |
            v

     ReplacementDone  <--
                        |
            |           |
            v           |
                        |
        Completing    --

            |
            v

        Completed

A single saga invocation is not enough to continually make sure a Volume is being repaired, so region replacement is structured as series of background tasks and saga invocations from those background tasks.

Here's a high level summary:

- a `region replacement` background task:

  - looks for disks that have been expunged and inserts region replacement requests into CRDB with state `Requested`

  - looks for all region replacemnt requests in state `Requested` (picking up new requests and requests that failed to transition to `Running`), and invokes a `region replacement start` saga.

- the `region replacement start` saga:

  - transitions the request to state `Allocating`, blocking out other invocations of the same saga

  - allocates a new replacement region

  - alters the Volume Construction Request by swapping out the old region for the replacement one

  - transitions the request to state `Running`

  - any unwind will transition the request back to the `Requested` state.

- a `region replacement drive` background task:

  - looks for requests with state `Running`, and invokes the `region replacement drive` saga for those requests

  - looks for requests with state `ReplacementDone`, and invokes the `region replacement finish` saga for those requests

- the `region replacement drive` saga will:

  - transition a request to state `Driving`, again blocking out other invocations of the same saga

  - check if Nexus has taken an action to initiate a repair yet. if not, then one is needed. if it _has_ previously initiated a repair operation, the state of the system is examined: is that operation still running? has something changed? further action may be required depending on this observation.

  - if an action is required, Nexus will prepare an action that will initiate either Live Repair or Reconciliation based on the current observed state of the system.

  - that action is then executed. if there was an error, then the saga unwinds. if it was successful, it is recorded as a "repair step" in CRDB and will be checked the next time the saga runs.

  - if Nexus observed an Upstairs telling it that a repair was completed or not necessary, then the request is placed into the `ReplacementDone` state, otherwise it is placed back into the `Running` state. if the saga unwinds, it unwinds back to the `Running` state.

- finally, the `region replacement finish` saga will:

  - transition a request into `Completing`

  - delete the old region by deleting a transient Volume that refers to it (in the case where a sled or disk is actually physically gone, expunging that will trigger #4331, which needs to be fixed!)

  - transition the request to the `Complete` state

More detailed documentation is provided in each of the region replacement saga's beginning docstrings.

Testing was done manually using the Canada region using the following test cases:

- a disk needing repair is attached to a instance for the duration of the repair

- a disk needing repair is attached to a instance that is migrated mid-repair

- a disk needing repair is attached to a instance that is stopped mid-repair

- a disk needing repair is attached to a instance that is stopped mid-repair, then started in the middle of the pantry's repair

- a detached disk needs repair

- a detached disk needs repair, and is then attached to an instance that is then started

- a sled is expunged, causing region replacement requests for all regions on it

Fixes #3886
Fixes #5191


nexus/src/app/sagas/region_replacement_drive.rs
```diff
@@ -0,0 +1,1521 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! # first, some Crucible background #
```
andrewjstone, line 5: great comment!

andrewjstone (commented):
James, this is epic work. I've only given it a cursory look so far, and will need to spend much more time digging in. Given how far behind current main this is, and the slight alleviation of urgency, I was wondering if you could split this up into multiple logical PRs to make it easier to review. I think this should be feasible by splitting along datastore queries and then saga / background task lines. Each of those can be added to the code and tested without being used. The background tasks for instance don't need to be enabled immediately and the sagas don't need to be triggered by the background tasks and or omdb. The OMDB change can come in last. My gut feeling is that this would also make it easier to test things in isolation, as you may see issues while doing the split and writing individual commit messages.

jmpesp:
Closing this, will split it up!

PR #5684, opened 2024-05-01
Read rack_subnet from RSS handoff request instead of the bootstore

### LINKED ISSUE #5680: `rack_initialize` ignores the rack subnet it's given and reads it from the bootstore instead
Hmm, I'm thinking that we probably shouldn't read this from the bootstore. I don't know why we wouldn't use what's in `rack_initialize`. This seems like a bug waiting to happen, although I also don't see how that bug would arise exactly, given what's in the bootstore and rack_initialize should be the same.

_Originally posted by @andrewjstone in https://github.com/oxidecomputer/omicron/pull/5675#discussion_r1586473854_

More context in https://github.com/oxidecomputer/omicron/pull/5675#discussion_r1586461240 - we pass a `rack_subnet` to `rack_initialize`, but it ignores it in favor of dynamically reading the bootstore's current config.

jgallagher:
Fixes #5680.

### DIFF
--- a/nexus/src/app/mod.rs
+++ b/nexus/src/app/mod.rs
@@ -441,12 +441,6 @@ impl Nexus {
 
         // TODO-cleanup all the extra Arcs here seems wrong
         let nexus = Arc::new(nexus);
-        let bootstore_opctx = OpContext::for_background(
-            log.new(o!("component" => "Bootstore")),
-            Arc::clone(&authz),
-            authn::Context::internal_api(),
-            Arc::clone(&db_datastore),
-        );
         let opctx = OpContext::for_background(
             log.new(o!("component" => "SagaRecoverer")),
             Arc::clone(&authz),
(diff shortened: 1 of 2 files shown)


nexus/src/app/mod.rs
```diff
@@ -486,12 +480,6 @@ impl Nexus {
                     for task in task_nexus.background_tasks.driver.tasks() {
                         task_nexus.background_tasks.driver.activate(task);
                     }
-                    if let Err(e) = task_nexus
```
andrewjstone, line 489: To future readers: This was needed in order to backfill a new `rack_subnet` row in the CRDB `rack` table for existing customer racks, since the bootstore had that subnet information from wicket, but CRDB did not. This was needed for BGP. However, reading from the bootstore is a always suspect, as it is eventually consistent, and can return different results. This worked for this case since we require all nodes to ack during RSS, and because the `rack_subnet` is immutable. It does not work in the add sled case, because a new node would not have this information yet. This wouldn't actually be a problem in production, because the `rack_subnet` would not get overwritten after RSS on existing racks. But it is a problem in testing, given the ordering of our fake startup. 

Since this is no longer needed, and since reading from the bootstore is wrong in general, it is safe to remove.

andrewjstone (commented):
Thanks for the fix @jgallagher!

PR #5688, opened 2024-05-02
Warn on `clippy::declare_interior_mutable_const`

jgallagher:
It seems dubious that this is classified as a `style` warning, since it could lead to incorrectness. (No incorrectness around `Lazy`, just completely destroying the whole point of `Lazy` in the first place.)


hawkw:
IMHO, this warning has a really above-average number of false positives; e.g., any time you want to initialize an array of atomics, the recommended approach is to use a constant, as `Atomic{foo}` is not `Copy`, as in:
```rust
static MY_ARRAY_OF_ATOMICS: [AtomicUsize; 4] = {
    const INIT: AtomicUsize = AtomicUsize::new(0);
    [INIT; 4]
};
```
which is a correct use of an interior mutable constant that nevertheless triggers this lint.

The 'known problems' section at https://rust-lang.github.io/rust-clippy/master/index.html#/declare_interior_mutable_const does mention this, but (IMO) doesn't really indicate sufficiently that there are plenty of non-"legacy" reasons you might want a const initializer.

jgallagher:
Fair! But it appears we don't have any of those in omicron, and it does catch an actual problem (and @bnaecker mentioned it would have caught something he ran into but fixed in dev last week). I'm inclined to turn this on and if it becomes a nuisance revisit, given that.

bnaecker:
Yeah, I hit this while fixing #5661. I was not at all aware of the behavior of `const`s being "copied" everywhere they're referred to. I had a `const AtomicU64`, where I just typed `const` instead of `static` without thinking about it, and was _extremely_ confused why no operations ever did anything to the value!

sunshowers:
I think to me it seems like @hawkw's use case is valid, but less common than the issue @bnaecker hit. Would it be reasonable to enable this lint, and document it in the config as something that you should disable if you really want to?

bnaecker:
That would be my suggestion too. It was a really confusing learning experience that this lint would have caught immediately, and it can always be ignored explicitly at specific sites for those who need to work around it.

jgallagher:
I added a comment on the warning in [92b6ff8](https://github.com/oxidecomputer/omicron/pull/5688/commits/92b6ff874aaeb441a1851374405b20bef086b25d)

dev-tools/xtask/src/clippy.rs
```diff
@@ -68,6 +68,14 @@ pub fn run_cmd(args: ClippyArgs) -> Result<()> {
         .arg("clippy::len_zero")
         .arg("--warn")
         .arg("clippy::redundant_field_names")
+        // `declare_interior_mutable_const` is classified as a style lint, but
+        // it can identify real bugs (e.g., declarying a `const Atomic` and
+        // using it like a `static Atomic`). However, it is also subject to
+        // false positives (e.g., idiomatically declarying a static array of
```
sunshowers, line 74:
```suggestion
        // false positives (e.g., idiomatically declaring a static array of
```

sunshowers (approved):

PR #5689, opened 2024-05-02
remove the newtype uuid from the public API

ahl:
I don't **think** we want to include newtype UUIDs in the public API. There's only one and it was added (I think by accident) recently.

### DIFF
--- a/nexus/src/external_api/http_entrypoints.rs
+++ b/nexus/src/external_api/http_entrypoints.rs
@@ -83,7 +83,7 @@ use omicron_common::api::external::{
     http_pagination::data_page_params_for, AggregateBgpMessageHistory,
 };
 use omicron_common::bail_unless;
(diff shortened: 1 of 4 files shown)


jgallagher (approved):

sunshowers:
Good idea for now. I think that at some point there was definitely discussion of being able to namespace uuids within the public API.

ahl:
> Good idea for now. I think that at some point there was definitely discussion of being able to namespace uuids within the public API.

There was! I don't think we reached any determination, and I don't think the newtype uuids match my recollection of that discussion (though they would be a useful mechanism on the way there).

karencfv:
Perhaps discussions about the public API could be discussed/brought up in the Product eng sync? Some decisions could change the way our language SDKs look like, and have customer impact.

PR #5700, opened 2024-05-05
Insert a default allowlist during schema migrations

bnaecker:
On new racks, an allowlist for user-facing services is populated during RSS. Nexus waits for the allowlist to exist before starting its external server, to ensure that the rules exist from the first request. On existing racks, however, RSS has already run and the allowlist is never inserted, causing Nexus to wait forever.

This commit adds a one-time insertion of the default allowlist (allowing any external traffic) as a data-only DB migration. An existing allowlist is not modified to preserve existing allowlists during further migrations.


rcgoodfellow (approved):
Works in the testbed. Thanks!

andrewjstone (approved):
Makes sense to me. Thanks for all the heroic work getting this right @bnaecker

PR #5706, opened 2024-05-06
Expose disk firmware in inventory

papertigers:
The necessary bits to track a disk's firmware version over its lifetime. This also plumbs up the firmware metadata to the inventory http endpoint, although nothing is consuming it just yet as there will be a follow up PR that actually inserts this data into CRDB.

### DIFF
--- a/sled-hardware/src/disk.rs
+++ b/sled-hardware/src/disk.rs
@@ -132,7 +132,6 @@ impl DiskPaths {
     }
 }
 
-// XXX MTZ: Make this an enum with DiskFirmware::Nvme?
 #[derive(
     Debug, Clone, PartialEq, Eq, Hash, Ord, PartialOrd, Deserialize, Serialize,
 )]
@@ -140,7 +139,11 @@ pub struct DiskFirmware {
     active_slot: u8,
     next_active_slot: Option<u8>,
     slot1_read_only: bool,
-    slots: Vec<Option<String>>,
+    // NB: This vec is 0 indexed while active_slot and next_active_slot are
+    // referring to "slots" in terms of the NVMe spec which defines slots 1-7.
+    // If the active_slot is 1, then it will be slot_firmware_versions[0] in the
+    // vector.
+    slot_firmware_versions: Vec<Option<String>>,
 }
 
 impl DiskFirmware {
@@ -157,7 +160,7 @@ impl DiskFirmware {
     }
 
     pub fn slots(&self) -> &[Option<String>] {
-        self.slots.as_slice()
+        self.slot_firmware_versions.as_slice()
     }
 }
 
@@ -168,7 +171,12 @@ impl DiskFirmware {
         slot1_read_only: bool,
         slots: Vec<Option<String>>,
     ) -> Self {
-        Self { active_slot, next_active_slot, slot1_read_only, slots }
+        Self {
+            active_slot,
+            next_active_slot,
+            slot1_read_only,
+            slot_firmware_versions: slots,
+        }
     }
 }
 
--- a/sled-hardware/src/disk.rs
+++ b/sled-hardware/src/disk.rs
@@ -132,6 +132,46 @@ impl DiskPaths {
     }
 }
 
+// XXX MTZ: Make this an enum with DiskFirmware::Nvme?
+#[derive(
+    Debug, Clone, PartialEq, Eq, Hash, Ord, PartialOrd, Deserialize, Serialize,
+)]
+pub struct DiskFirmware {
+    active_slot: u8,
+    next_active_slot: Option<u8>,
+    slot1_read_only: bool,
+    slots: Vec<Option<String>>,
+}
+
+impl DiskFirmware {
+    pub fn active_slot(&self) -> u8 {
+        self.active_slot
+    }
+
+    pub fn next_active_slot(&self) -> Option<u8> {
+        self.next_active_slot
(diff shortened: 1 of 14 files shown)


sled-hardware/src/disk.rs
```diff
@@ -132,6 +132,46 @@ impl DiskPaths {
...
 }
 
+// XXX MTZ: Make this an enum with DiskFirmware::Nvme?
+#[derive(
+    Debug, Clone, PartialEq, Eq, Hash, Ord, PartialOrd, Deserialize, Serialize,
+)]
+pub struct DiskFirmware {
+    active_slot: u8,
+    next_active_slot: Option<u8>,
+    slot1_read_only: bool,
```
smklein, line 142: It feels odd that all the other fields imply an arbitrary number of slots, but this field specifically embeds the notion of "slot1"? Also, is this one or zero-indexed?

sled-hardware/src/disk.rs
```diff
@@ -132,6 +132,46 @@ impl DiskPaths {
...
+    slots: Vec<Option<String>>,
```
smklein, line 143: What's the value of this String? This feels a little bit like a more specific type that has been coerced into a string?

sled-hardware/src/illumos/mod.rs
```diff
@@ -250,17 +264,38 @@ impl HardwareView {
...
     ) {
-        // In old set, not in new set.
-        let removed = self.disks.difference(&polled_hw.disks);
-        // In new set, not in old set.
-        let added = polled_hw.disks.difference(&self.disks);
+        let mut added = Vec::new();
+        let mut removed = Vec::new();
+        let mut updated = Vec::new();
+
+        // Find new or udpated disks.
```
smklein, line 271:
```suggestion
        // Find new or updated disks.
```

sled-hardware/src/illumos/mod.rs
```diff
@@ -492,15 +534,45 @@ fn poll_blkdev_node(
...
+                log,
+                "NVMe Controller is already locked so we will try again
+                in the next hardware snapshot"
+            );
+            return Err(Error::NvmeControllerLocked);
+        }
+        libnvme::controller::TryLockResult::Err(err) => {
+            return Err(Error::from(err))
+        }
+    };
```
smklein, line 554: :+1:  this makes sense, given our prior conversation

sled-storage/src/disk.rs
```diff
@@ -404,6 +426,26 @@ impl Disk {
             Self::Synthetic(disk) => disk.raw.slot,
         }
     }
+
+    // Today the only update we expect to be able to apply to
+    // a `Disk` is firmware.
+    pub(crate) fn update_disk(&mut self, raw_disk: &RawDisk) {
```
smklein, line 432: I wonder if we could program this a little bit more defensively, in case there are other parts of these structures that can be updated. The comment indicates the true behavior ("we ignore everything but the firmware") but the calllsites don't really make that clear.

Can we either call this `update_firmware`, or have some other way to make sure "we don't drop updates on the floor" if something other than the firmware field changes?

smklein (commented):
Overall structure makes sense to me, just a couple nitpicks!

sled-hardware/src/disk.rs
papertigers: Welcome to the NVMe spec! The spec states that devices have slot 1-7 (and potentially more in the future as there's unused padding at the end of the C struct). The spec also says that only slot 1 can be read only. Not every device will have all 7 slots available, for example the C bhyve nvme devices shows only one slot that is read only, while some of our WDC devices show 4 slots.  And not every available slot will necessarily contain firmware.  The firmware version itself is defined to be an ascii string hence the `Vec<Option<String>>`.

After having chatted with @jgallagher about this we landed on `NvmeSlot` being a wrapping type to address the array index being 0 but corresponding with slot1 - this can be seen [here] (https://github.com/oxidecomputer/libnvme/blob/mike/libnvme-firmware/libnvme/src/firmware.rs#L112-L135) and has yet to be officially reviewed.  The thought was that one could gain access to all the slots via an iterator and upstream of the library you could deal with the 0 indexed array. 

I originally had this represented as an enum but was unsure what the underlying CRDB type should look like.  Very open to suggestions here as I just went with _something_ for this PR as a placeholder.

sled-hardware/src/disk.rs
papertigers: See the above comment.

sled-storage/src/disk.rs
papertigers: I can fix this - I had originally left this as `update_disk` as there might be more fields in the future we wish to support. But I am happy to make this an explicit `update_firmware` for the time being. (side note... maybe `update_firmware_metadata` is a better name).

papertigers:
This is ready for review now that https://github.com/oxidecomputer/libnvme/pull/4 has landed.

sled-storage/src/manager_test_harness.rs
```diff
@@ -300,6 +300,15 @@ impl StorageManagerTestHarness {
             .expect("Failed to remove vdev");
     }
 
+    // XXX MTZ: Provide a vdev update aka new firmware.
```
smklein, line 303: Can we address this this "XXX" before merging?

smklein (approved):

Issue #5566, opened 2024-04-18
Don't exit with an error if virtual hardware has already been destroyed


karencfv:
Currently when running `cargo xtask virtual-hardware destroy` and the hardware has already been destroyed you get the following error:

```console
$ cargo xtask virtual-hardware destroy
   Compiling xtask v0.1.0 (/home/coatlicue/src/omicron/dev-tools/xtask)
    Finished dev [unoptimized + debuginfo] target(s) in 1.69s
     Running `target/debug/xtask virtual-hardware destroy`
destroying virtual hardware
Disabling svc:/system/fmd:default
Re-enabling svc:/system/fmd:default
unloading xde driver
ensuring softnpu zone destroyed
Error: "/usr/bin/pfexec" failed: exit status: 1 (stderr: failed to halt zone config for sidecar_softnpu: Failed to parse command output: exit code 1
stdout:

stderr:
zoneadm: sidecar_softnpu: No such zone configured
failed to uninstall zone config for sidecar_softnpu: Failed to parse command output: exit code 1
stdout:

stderr:
zoneadm: sidecar_softnpu: No such zone configured
failed to delete zone config for sidecar_softnpu: Failed to parse command output: exit code 1
stdout:

stderr:
sidecar_softnpu: No such zone configured
sidecar_softnpu: No such zone configured
Error: No such file or directory (os error 2)
)
```

It would be a nicer experience to just get a message that the virtual hardware has already been destroyed.

Additionally, when trying to run the command again after the previous attempt to destroy virtual hardware was interrupted for some reason, you get this error again. Which means you have to run the destroy virtual hardware script instead.

PR #5707, opened 2024-05-06
[xtask] Make virtual-hardware destroy idempotent

smklein:
Tested on atrium: re-running `cargo xtask virtual-hardware destroy` no longer throws errors

Fixes https://github.com/oxidecomputer/omicron/issues/5566

### DIFF
--- a/dev-tools/xtask/src/virtual_hardware.rs
+++ b/dev-tools/xtask/src/virtual_hardware.rs
@@ -275,6 +275,8 @@ fn remove_softnpu_zone(npu_zone: &Utf8Path) -> Result<()> {
         if output.to_string().contains("No such zone configured") {
             println!("zone {npu_zone} already destroyed");
             return Ok(());
+        } else {
+            return Err(output);
         }
     }
     Ok(())
--- a/dev-tools/xtask/src/virtual_hardware.rs
+++ b/dev-tools/xtask/src/virtual_hardware.rs
@@ -270,7 +270,12 @@ fn remove_softnpu_zone(npu_zone: &Utf8Path) -> Result<()> {
         "--ports",
         "sc0_1,tfportqsfp0_0",
     ]);
-    execute(cmd)?;
+    if let Err(output) = execute(cmd) {
+        // Don't throw an error if the zone was already removed
+        if output.to_string().contains("No such zone configured") {
+            return Ok(());
+        }
+    }
     Ok(())
 }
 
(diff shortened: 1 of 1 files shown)


dev-tools/xtask/src/virtual_hardware.rs
```diff
@@ -270,7 +270,12 @@ fn remove_softnpu_zone(npu_zone: &Utf8Path) -> Result<()> {
         "--ports",
         "sc0_1,tfportqsfp0_0",
     ]);
-    execute(cmd)?;
+    if let Err(output) = execute(cmd) {
+        // Don't throw an error if the zone was already removed
+        if output.to_string().contains("No such zone configured") {
```
karencfv, line 275: Nit: Can we print out the zone that was already destroyed? Something like:

```suggestion
        if output.to_string().contains("No such zone configured") {
            println!("zone {npu_zone} already destroyed");
```

karencfv (approved):
Nice! Thanks for working on this so quickly :bowing_woman: 

Just one optional comment, feel free to ignore if you disagree :)

dev-tools/xtask/src/virtual_hardware.rs
smklein: Sure, added

dev-tools/xtask/src/virtual_hardware.rs
```diff
@@ -270,7 +270,13 @@ fn remove_softnpu_zone(npu_zone: &Utf8Path) -> Result<()> {
...
```
bnaecker, line 275: It seems like we should be returning an error if the conditional is false, right? This will fall through to `return Ok(())`, unless I'm missing something.

karencfv: I was under the impression this was by design. We want to keep destroying things even if there is an error destroying the zone. Or at least that was my take on it

karencfv: But I guess perhaps we do want to print out the error message so people know something didn't quite go the way we wanted it to.

bnaecker: I think it depends on the error. If we failed to remove it because it's already gone, then yes, we should continue. If we failed to remove it for an unexpected reason, then I'm not so sure. I'd at least want to know _why_ it failed, even if we ultimately decide to continue.

smklein: whoops, silly oversight -- agreed, I need to actually propagate most errors. updated!

bnaecker (approved):
Thanks, this looks great

PR #5714, opened 2024-05-07
[xtask] allow overriding `clippy` message format

hawkw:
Currently, I've configured VS Code to run checks on save using the same
list of lints checked by `cargo xtask clippy` with a
`.vscode/settings.json` like this:

```json
{
    "rust-analyzer.check.overrideCommand": [
        "cargo",
        "clippy",
        "--all-targets",
        "--message-format=json",
        "--",
        "--allow",
        "clippy::style",
        "--warn",
        "clippy::disallowed_macros",
        "--warn",
        "clippy::disallowed_methods",
        "--warn",
        "clippy::disallowed_names",
        "--warn",
        "clippy::disallowed_script_idents",
        "--warn",
        "clippy::disallowed_types",
        "--warn",
        "clippy::iter_cloned_collect",
        "--warn",
        "clippy::iter_next_slice",
        "--warn",
        "clippy::iter_nth",
        "--warn",
        "clippy::iter_nth_zero",
        "--warn",
        "clippy::iter_skip_next",
        "--warn",
        "clippy::len_zero",
        "--warn",
        "clippy::redundant_field_names",
        "--warn",
        "clippy::cast_lossless"
    ]
}
```

Having to reproduce the list of lints in the `settings.json` is a bit
unfortunate, though, because if a developer wishes to keep their list of
enabled lints in sync with those run by `cargo xtask clippy`, they must
manually update the VS Code config when new lints are added. It would be
nicer if the VS Code config could just run `cargo xtask clippy` on save.

However, this isn't currently possible, as rust-analyzer relies on the
JSON output format for `cargo check`/`cargo clippy`, so it needs to pass
the `--message-format=json` flag to `cargo clippy`. This flag is not
currently propagated by `cargo xtask clippy`.

Therefore, I've updated the xtask to accept a `--message-format`
argument that can be passed through to the underlying `cargo clippy`
command. Now, the VS Code config can be changed to just:

```json
{
    "rust-analyzer.check.overrideCommand": [
        "cargo",
        "xtask",
        "clippy",
        "--message-format=json",
    ]
}
```


hawkw:
I'm also open to checking in a `.vscode/config.json` that configures VS Code to run `cargo xtask clippy` on save, but IIRC, someone mentioned that there are some issues with committing VS Code config files? Alternatively, we could add this to developer documentation someplace.

smklein (approved):

iliana:
Disabled auto merge for merge freeze

hawkw:
> Disabled auto merge for merge freeze

ah, whoops, sorry!

dev-tools/xtask/src/clippy.rs
```diff
@@ -13,6 +13,9 @@ pub struct ClippyArgs {
     /// Automatically apply lint suggestions.
     #[clap(long)]
     fix: bool,
+    /// Error format passed to `cargo clippy`.
+    #[clap(long)]
+    message_format: Option<String>,
```
sunshowers, line 18: Maybe `value_name = "FMT"` to match what cargo produces?

dev-tools/xtask/src/clippy.rs
```diff
@@ -25,6 +28,17 @@ pub fn run_cmd(args: ClippyArgs) -> Result<()> {
...
 
+    // Pass along the `--message-format` flag if it was provided.
+    //
+    // We don't really care about validating that it's a valid argument to
+    // `cargo check --message-format`, because `cargo check` will error out if
+    // it's unrecognized, and repeating the validation here just presents an
+    // opportunity to get out of sync with what Cargo actually accepts should a
+    // new message format be added.
+    if let Some(fmt) = args.message_format {
+        command.args(&["--message-format", &fmt]);
```
sunshowers, line 39: Don't need the first & here right?

sunshowers (approved):
Awesome! Just a couple small nits.

hawkw:
I think @iliana's PR #5715 kind of obviates this (and honestly, kind of obviates `cargo xtask clippy` entirely, honestly), so I'll probably close this. But, if folks think this is worth having regardless, I'm happy to merge it instead.

sunshowers:
I think I'd still want this tbh

hawkw:
Mmkay, I'll get it ready to merge, then.

PR #5715, opened 2024-05-07
move clippy args to `[workspace.lints.clippy]`

iliana:
This moves our Clippy configuration out of xtask and into Cargo's relatively-new lint configuration functionality. https://blog.rust-lang.org/2023/11/16/Rust-1.74.0.html#lint-configuration-through-cargo

This seeks to solve a similar root issue as #5714, which is "please, rust-analyzer, do not tell me about clippy lints we don't care about in CI".

Unfortunately this requires littering `[lints] workspace = true` in every Cargo.toml, at least until [implicit workspace inheritance](https://github.com/rust-lang/cargo/issues/12208) is figured out, which is most of the noise of this PR. We may want a CI check to ensure that this is set in all packages.

As part of this I removed the `#![allow(clippy::style)]` attributes from nexus and sled-agent, which revealed that this was overriding any style lints we re-enabled in the xtask code.

One issue is that workspace crates can't override these lints yet (https://github.com/rust-lang/cargo/issues/13157), but they can continue to override them with `#![allow(clippy::whatever)]` attributes.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -192,6 +192,11 @@ redundant_field_names = "warn"
 # idiomatically declaring a static array of atomics uses `const Atomic`). We
 # warn on this to catch the former, and expect any uses of the latter to allow
 # this locally.
+#
+# Note: any const value with a type containing a `bytes::Bytes` hits this lint,
+# and you should `#![allow]` it for now. This is most likely to be seen with
+# `http::header::{HeaderName, HeaderValue}`. This is a Clippy bug which will be
+# fixed in the Rust 1.80 toolchain (rust-lang/rust-clippy#12691).
 declare_interior_mutable_const = "warn"
 # Also warn on casts, preferring explicit conversions instead.
 #
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -162,6 +162,43 @@ default-members = [
 ]
 resolver = "2"
 
+#
+# Tree-wide lint configuration.
+# https://doc.rust-lang.org/stable/cargo/reference/manifest.html#the-lints-section
+#
+# For a list of Clippy lints, see
+# https://rust-lang.github.io/rust-clippy/master.
+#
+[workspace.lints.clippy]
+# Clippy's style nits are useful, but not worth keeping in CI.
+style = { level = "allow", priority = -1 }
+# But continue to warn on anything in the "disallowed_" namespace.
+disallowed_macros = "warn"
+disallowed_methods = "warn"
+disallowed_names = "warn"
+disallowed_script_idents = "warn"
+disallowed_types = "warn"
+# Warn on some more style lints that are relatively stable and make sense.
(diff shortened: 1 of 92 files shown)


Cargo.toml
```diff
@@ -162,6 +162,43 @@ default-members = [
...
+iter_nth_zero = "warn"
+iter_skip_next = "warn"
+len_zero = "warn"
+redundant_field_names = "warn"
+# `declare_interior_mutable_const` is classified as a style lint, but it can
+# identify real bugs (e.g., declarying a `const Atomic` and using it like
+# a `static Atomic`). However, it is also subject to false positives (e.g.,
+# idiomatically declaring a static array of atomics uses `const Atomic`). We
+# warn on this to catch the former, and expect any uses of the latter to allow
+# this locally.
```
hawkw, line 194: take it or leave it: maybe worth also mentioning `http::HeaderName` explicitly, since that's probably one of the most common false positives we'll hit in omicron? it would be nice if someone who adds a `const HeaderName` in the future to immediately see "no, just allow that lint" written down here.

Cargo.toml
```diff
@@ -162,6 +162,43 @@ default-members = [
...
+#
+# Tree-wide lint configuration.
+# https://doc.rust-lang.org/stable/cargo/reference/manifest.html#the-lints-section
+#
+# For a list of Clippy lints, see
+# https://rust-lang.github.io/rust-clippy/master.
+#
+[workspace.lints.clippy]
+# Clippy's style nits are useful, but not worth keeping in CI.
+style = { level = "allow", priority = -1 }
```
hawkw, line 174: what does `priority = -1` do? i assume it allows this to be overridden, but why do we only need to set it here?

hawkw (approved):

jgallagher (approved):
Ditto Eliza's question about `priority = -1`, but other than that LGTM!

Cargo.toml
iliana: Until `[lints]`, all methods for modifying lint levels had inherent ordering; either command line flags, or attributes in code. TOML tables are explicitly unordered, so Rust had to do something different here. [Per the RFC](https://rust-lang.github.io/rfcs/3389-manifest-lint.html#reference-level-explanation) any sorting beyond user-defined priorities is undefined ([more detail](https://rust-lang.github.io/rfcs/3389-manifest-lint.html#lint-precedence)); `style` has to come first because otherwise it would re-allow the `style` lints that happen to come before it once sorted.

The default `priority` for any key in this table is 0, so setting this to -1 means it is ordered before the rest of the keys.

sunshowers:
Ooh just saw this one — really concerned that we're going to miss workspace = true in new crates. How hard would it be to address this before landing?

iliana:
Probably!

sunshowers (approved):
Awesome, thanks!

PR #5726, opened 2024-05-09
Update pinned maghemite

charliepark:
The pinned sha for maghemite was out of date, pointing to an invalid sha (I believe because of some force pushes that removed the specific commit).

This updates omicron to point to [the most recent sha for maghemite](https://github.com/oxidecomputer/maghemite/pull/199) and updates the necessary checksums so the tests pass.

Some context on why this is even needed: because there is no mac build of mgd, we build it from source for macs, and need the correct sha for that.


charliepark:
I'm not totally confident there isn't a step I've missed in this; if @rcgoodfellow or others have things to point out, I'm all ears.

david-crespo:
Heh just caught this. It would be nice if either this mismatch could be caught in CI and/or it was easier to make the change.

https://github.com/oxidecomputer/omicron/blob/4dda5cbba6c7bd4d47df79cf0001a7873abd21b8/package-manifest.toml#L563-L567

david-crespo:
Since it seems like `package-manifest.toml` is more likely to be correct, I updated the hack script we use to pull the dendrite and maghemite commits from there. But this mismatch will probably affect someone else eventually. The nix flake uses the commit from the files changed in this PR, though because it can pull the linux binary, it will at least be able to find a file on buildomat. However, it may not be the correct file.

https://gist.github.com/david-crespo/d8aa7ea1afb877ff585e6ad90fc5bc2c

rcgoodfellow (approved):
Thanks for fixing this. Ultimately it would be nice to get rid of these environment variable files and just use what's in Cargo.toml.

PR #5729, opened 2024-05-09
Check full source IP allow-list, not just the last entry

bnaecker:


wicket/src/ui/panes/rack_setup.rs
```diff
@@ -1086,6 +1086,7 @@ fn rss_config_text<'a>(
     );
 
     // Add the allowlist for connecting to user-facing rack services.
+    println!("{:?}", allowed_source_ips);
```
bnaecker, line 1089: Woops, debugging cruft.

wicket/tests/output/example_non_empty.toml
```diff
@@ -47,19 +47,19 @@ internal_services_ip_pool_ranges = [
 # Confirm this list contains all expected sleds before continuing!
 bootstrap_sleds = [
     1, # serial 1 2 3 (model model1 revision 3, IP address UNKNOWN)
-    5, # serial 4 5 6 (model model2 revision 5, ::1)
```
bnaecker, line 50: Same, these should be untouched, mistakenly did `git add -u`

rcgoodfellow (approved):

PR #5730, opened 2024-05-09
Plumb IP allow list all the way to the front door

### LINKED ISSUE #5728: Source IP allowlist is not making it from RSS config file to Wicket UI
During customer install, the correct IP allowlist in the RSS config TOML file included many IPs. The `wicket` UI showed the default of `Any`, indicating it doesn't seem to be populated from the file in some way. No further details yet.

bnaecker:
- Fixes #5728
- Store the uploaded allowlist in the `wicketd` server context
- Spit the allowlist out in the TOML document we get back from the `wicket` CLI `setup get-config` subcommand

### DIFF
--- a/wicket/tests/output/example_non_empty.toml
+++ b/wicket/tests/output/example_non_empty.toml
@@ -50,6 +50,20 @@ bootstrap_sleds = [
     5, # serial 4 5 6 (model model2 revision 5, ::1)
 ]
 
+# Allowlist of source IPs that can make requests to user-facing services.
+#
+# Use the key:
+#
+# allow = "any"
+#
+# to indicate any external IPs are allowed to make requests. This is the default.
+#
+# Use the below two lines to only allow requests from the specified IP subnets.
+# Requests from any other source IPs are refused. Note that individual addresses
+# must include the netmask, e.g., "1.2.3.4/32".
+#
+# allow = "list"
+# ips = [ "1.2.3.4/5", "5.6.7.8/10" ]
 [allowed_source_ips]
 allow = "any"
 
--- a/wicket/tests/output/example_non_empty.toml
+++ b/wicket/tests/output/example_non_empty.toml
@@ -67,12 +67,6 @@ bootstrap_sleds = [
 [allowed_source_ips]
 allow = "any"
 
-# Use the below two lines to only allow requests from the specified IP subnets.
-# Requests from any other source IPs are refused. Note that individual addresses
-# must include the netmask, e.g., `1.2.3.4/32`.
-# allow = "list"
-# ips = [ "1.2.3.4/5", "5.6.7.8/10" ]
-
 # TODO: docs on network config
 [rack_network_config]
 infra_ip_first = "172.30.0.1"
--- a/wicket/tests/output/example_non_empty.toml
+++ b/wicket/tests/output/example_non_empty.toml
@@ -50,9 +50,7 @@ bootstrap_sleds = [
     5, # serial 4 5 6 (model model2 revision 5, ::1)
 ]
 
-# Allowlist of source IPs that can make requests to user-facing services.
 [allowed_source_ips]
-# Any external IPs to make requests. This is the default.
 allow = "any"
 
 # Use the below two lines to only allow requests from the specified IP subnets.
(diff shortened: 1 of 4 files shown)


bnaecker:
See the corresponding issue for details. I missed a few places we needed to actually store the allow-list on upload to `wicketd`, and the part that serializes it to a TOML document for the `wicket` CLI. I showed in the issue that the `wicket` TUI now includes the list, and we can also see here that it's included in the TOML that we get back from the CLI:

```
bnaecker@shale : ~/omicron $ rg -A 10 allowed_source wicket/tests/output/example_non_empty.toml
54:[allowed_source_ips]
55-# Any external IPs to make requests. This is the default.
56-#allow = "any"
57-
58-# Use the below two lines to only allow requests from the specified IP subnets.
59-# Requests from any other source IPs are refused. Note that individual addresses
60-# must include the netmask, e.g., `1.2.3.4/32`.
61-allow = "list"
62-ips = [ "1.2.3.4/5", "5.6.7.8/10" ]
63-
64-# TODO: docs on network config
bnaecker@shale : ~/omicron $ SSH_ORIGINAL_COMMAND="setup set-config" cargo run -p wicket < wicket/tests/output/example_non_empty.toml
    Finished dev [unoptimized + debuginfo] target(s) in 0.90s
     Running `target/debug/wicket`
May 09 23:13:42.128 INFO reading config from stdin..., file: wicket/src/cli/rack_setup.rs:110
May 09 23:13:42.129 INFO parsing config..., file: wicket/src/cli/rack_setup.rs:115
May 09 23:13:42.131 INFO uploading config to wicketd..., file: wicket/src/cli/rack_setup.rs:120
May 09 23:13:42.133 INFO config upload complete, file: wicket/src/cli/rack_setup.rs:126
bnaecker@shale : ~/omicron $ SSH_ORIGINAL_COMMAND="setup get-config" cargo run -p wicket | rg -A 10 allowed_source
    Finished dev [unoptimized + debuginfo] target(s) in 0.72s
     Running `target/debug/wicket`
[allowed_source_ips]
allow = "list"
ips = ["1.2.3.4/5", "5.6.7.8/10"]

# Use the below two lines to only allow requests from the specified IP subnets.
# Requests from any other source IPs are refused. Note that individual addresses
# must include the netmask, e.g., `1.2.3.4/32`.
# allow = "list"
# ips = [ "1.2.3.4/5", "5.6.7.8/10" ]

# TODO: docs on network config
```

wicket/tests/output/example_non_empty.toml
```diff
@@ -50,9 +50,7 @@ bootstrap_sleds = [
     5, # serial 4 5 6 (model model2 revision 5, ::1)
 ]
 
-# Allowlist of source IPs that can make requests to user-facing services.
 [allowed_source_ips]
-# Any external IPs to make requests. This is the default.
```
sunshowers, line 55: Any particular reason the comments were removed?

sunshowers (approved):
Looks great, thanks!

wicket/tests/output/example_non_empty.toml
sunshowers: Also I think this is separate from this PR, but I really would like the format to be either `"any"` or a list of IPs.

bnaecker: The comments were removed to placate the tests. I wasn't clear on how one could emit comments in the `TomlTemplate` alongside the allow-list itself. It could be that just moving them all prior to the table would solve it, I'm not sure.

> I really would like the format to be either "any" or a list of IPs/subnets.

Do you mean that the TOML itself would look like this?

```toml
allow = "any"
# OR
# allow = ["1.2.3.4/5"]
```

If so, then there's a tension here with the JSON schema for these API types. While it is the most natural for Rust, the generated schema is quite complicated and a bit awkward, since we effectively get an untagged enum of completely different types. I feel like this is a reasonable middle ground, but we can definitely bike-shed it a bit when it becomes important.

bnaecker: > Any particular reason the comments were removed?

I've moved the comments all prior to the actual table itself, and now they all round-trip just fine. I'll put them back like that for documentation.

bnaecker: Added the comments to the test in 0be6930b6

sunshowers: > If so, then there's a tension here with the JSON schema for these API types.

Agreed. I think we should use separate deserializers for toml and JSON.

PR #5734, opened 2024-05-10
Adding ip pool id to view

### LINKED ISSUE #5716: Add `ip_pool_id` to floating IP response
It's on the model:

https://github.com/oxidecomputer/omicron/blob/c1f9e8fde9284d9e0ee5ccb9d228c3bf3cec199c/nexus/db-model/src/external_ip.rs#L204-L214

But not on the view:

https://github.com/oxidecomputer/omicron/blob/c1f9e8fde9284d9e0ee5ccb9d228c3bf3cec199c/nexus/types/src/external_api/views.rs#L448-L458

It would be trivial to add and we think we want this info on the frontend. For example, imagine there are two pools linked to a silo, one with public IPs and one with private IPs. You would want to know which pool a floating IP came from. Right now you'd have to put it in the name or description.

charliepark:
Fixes #5716

This PR adds the IP Pool's ID to the Floating IP view, so we can use it on the frontend. Adds a few test assertions as well to existing tests.

### DIFF
--- a/nexus/db-model/src/external_ip.rs
+++ b/nexus/db-model/src/external_ip.rs
@@ -590,6 +590,7 @@ impl From<FloatingIp> for views::FloatingIp {
 
         views::FloatingIp {
             ip: ip.ip.ip(),
+            ip_pool_id: ip.ip_pool_id,
             identity,
             project_id: ip.project_id,
             instance_id: ip.parent_id,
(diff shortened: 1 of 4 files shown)


david-crespo (approved):
yeehaw

david-crespo:
Whoops! We changed the API and forgot to regen the schema. Run this:

```sh
EXPECTORATE=overwrite cargo t -p omicron-nexus -E 'test(=integration_tests::commands::test_nexus_openapi)'
```

image

charliepark:
Ah, thank you for catching that!

PR #5738, opened 2024-05-11
move dpd-client dependency out of omicron common

rcgoodfellow:
This breaks downstreams that import omicron_common as the way dpd-client gets built is specific to the omicron build machinery.

For example, this happens when trying to bump omicron in maghemite.

```
error: failed to run custom build command for `dpd-client v0.1.0 (https://github.com/oxidecomputer/omicron?branch=main#c472464a)`

Caused by:
  process didn't exit successfully: `/Users/ry/src/maghemite/target/debug/build/dpd-client-8e2b35d52fea36a5/build-script-build` (exit status: 1)
  --- stdout
  cargo:rerun-if-changed=../../package-manifest.toml

  --- stderr
  Error: ../../out/downloads/dpd-3b84ea6516cafb4595a6f2a668df16c1a501b687.json doesn't exist; rerun `tools/ci_download_dendrite_openapi` (after updating `tools/dendrite_openapi_version` if the dendrite commit in package-manifest.toml has changed)
```


common/src/api/external/mod.rs
```diff
@@ -2894,39 +2894,6 @@ pub struct SwitchPortAddressConfig {
     pub interface_name: String,
 }
 
-/// Opaque object representing link state. The contents of this object are not
-/// yet stable.
-#[derive(Clone, Debug, Deserialize, Serialize)]
-pub struct SwitchLinkState {
```
rcgoodfellow, line 2900: The bulk of this PR is moving this type to `nexus/types/src/external_api/shared.rs`

common/Cargo.toml
```diff
@@ -27,7 +27,6 @@ http.workspace = true
 ipnetwork.workspace = true
 macaddr.workspace = true
 mg-admin-client.workspace = true
-dpd-client.workspace = true
```
rcgoodfellow, line 30: This causes downstream breakage, as compilation of this crate requires omicron-specific build machinery.

sunshowers: ah yeah... I feel like CI should catch this, maybe by building a stub project with omicron-common as its only dependency. Could you file an issue for this?

sunshowers (approved):
thanks!

common/Cargo.toml
rcgoodfellow: https://github.com/oxidecomputer/omicron/issues/5748

PR #5742, opened 2024-05-11
Update the uuid kinds to emit a JSON Schema that includes the x-rust-type extension

ahl:
Unfortunately this doesn't yet have a visible change, but I thought it would be useful to break things up into smaller pieces. This will have visible changes after https://github.com/oxidecomputer/newtype-uuid/pull/47 and become useful after https://github.com/oxidecomputer/typify/pull/584.

### DIFF
--- a/uuid-kinds/Cargo.toml
+++ b/uuid-kinds/Cargo.toml
@@ -14,11 +14,12 @@ workspace = true
 [dependencies]
 newtype-uuid.workspace = true
 schemars = { workspace = true, optional = true }
+serde_json = { workspace = true, optional = true }
 paste.workspace = true
 
 [features]
 default = ["std"]
 serde = ["newtype-uuid/serde"]
-schemars08 = ["newtype-uuid/schemars08", "schemars"]
+schemars08 = ["newtype-uuid/schemars08", "dep:schemars", "dep:serde_json"]
 std = ["newtype-uuid/std"]
 uuid-v4 = ["newtype-uuid/v4"]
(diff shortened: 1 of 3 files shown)


uuid-kinds/Cargo.toml
```diff
@@ -14,11 +14,12 @@ workspace = true
...
 newtype-uuid.workspace = true
 schemars = { workspace = true, optional = true }
+serde_json = { workspace = true, optional = true }
 paste.workspace = true
 
 [features]
 default = ["std"]
 serde = ["newtype-uuid/serde"]
-schemars08 = ["newtype-uuid/schemars08", "schemars"]
+schemars08 = ["newtype-uuid/schemars08", "dep:schemars", "dep:serde_json"]
```
sunshowers, line 23: Good catch re using `dep:`, missed that somehow.

uuid-kinds/src/lib.rs
```diff
@@ -14,15 +14,63 @@ pub use newtype_uuid::{
...
 macro_rules! impl_typed_uuid_kind {
     ($($kind:ident => $tag:literal),* $(,)?) => {
         $(
             paste::paste! {
-                #[cfg_attr(feature = "schemars08", derive(JsonSchema))]
-                pub enum [< $kind Kind>] {}
+                pub enum [< $kind Kind >] {}
+
+                #[cfg(feature = "schemars08")]
+                impl schemars::JsonSchema for [< $kind Kind >]{
```
sunshowers, line 24: Yeah this is about what I expected -- let's chat today about seeing how we can get this into newtype-uuid instead.

PR #5744, opened 2024-05-13
rearrange buildomat jobs; rewrite releng process in rust and aggressively parallelize

iliana:
**Note**: documentation says `cargo xtask releng` but I am going to wire that up in a follow-up PR; the current equivalent is `cargo run --release --bin omicron-releng`.

---

Prior to this change we have five main "release engineering" Buildomat jobs that do operations beyond running the test suite:
- a **package** job which runs omicron-package in various configurations,
- a **build OS images** job which builds the host and trampoline images,
- a **TUF repo** job which builds the final TUF repo *(this is the build artifact we actually want)*,
- a **deploy** job which uses the single-sled packages to test that a VM boots to SSH *(this is a test we actually want)*,
- and a **CI tools** job which builds common tools used by multiple jobs.

This looks like:

```mermaid
graph LR
    package --> host-image["build OS images"]
    package --> deploy
    package --> tuf-repo["TUF repo"]
    host-image --> tuf-repo
    ci-tools["CI tools"] --> deploy
    ci-tools --> tuf-repo
```

(There are also the currently-disabled a4x2 jobs but those are independent of this particular graph.)

I think the initial idea behind this was to reuse build artifacts where possible, but this is pretty complicated and adds a lot more output upload/download overhead than expected, which slows down the time to get the end artifact we actually want.

This PR changes the graph to:

```mermaid
graph LR
    package --> deploy
    tuf-repo["TUF repo"]
```

And the **TUF repo** job primarily runs a new **releng** binary, which runs all of the steps required to download and build all the components of the TUF repo in a single task, using a terrible job runner I wrote.

CI taking too long? busy, but can't get stuff done? try 30 to 40 async rust functions. 30 to 40 async rust fns: an easy iteration time winner. put them directly on a runtime with a half-baked job server. you will certainly not regret writing 30 to 40 async fns<br><sup>(thank you to @lifning for the image)</sup>

The primary goal here was to reduce the time from pushing a commit to getting a TUF repo out the other end; this drops time-to-TUF-repo from ~80 minutes to ~45. In the process this also made it much easier to build a TUF repo (and iterate on that process) locally: just run `cargo xtask releng` (TODO: soon). It also deleted a lot of Bash.

One thing to note is that, in service of the mission to get time-to-TUF-repo down as much as possible, that job _only_ uploads the TUF repo (and some logs). I also put all of the outputs for the **package** job into a single tarball for the **deploy** job to unpack. There are no longer separate uploads for the OS images and each zone; these can be extracted from the repo as we normally do.


dev-tools/releng/src/hubris.rs
```diff
@@ -0,0 +1,145 @@
...
+        .get(format!("{}/artifact/{}", base_url, hash))
+        .send()
+        .and_then(|response| response.json())
+        .await
+        .with_context(|| {
+            format!(
+                "failed to fetch hubris artifact {} from {}",
+                hash, base_url
+            )
+        })
```
labbott, line 119: it's unfortunate that permslip is still a private repo but I guess it's mostly just a wrapper around reqwest anyway

dev-tools/releng/src/hubris.rs
```diff
@@ -0,0 +1,145 @@
...
+}
+
+#[derive(Deserialize)]
+struct Manifest {
+    #[serde(rename = "artifact")]
+    artifacts: HashMap<KnownArtifactKind, Vec<Artifact>>,
+}
```
labbott, line 126: Is it at all beneficial to pull this and some of the other structures I have in the hubris release repo to a third repo to try and keep things coordinated?

labbott (commented):
This looks like it means it's really easy to build a TUF repo via xtask?

dev-tools/releng/src/hubris.rs
iliana: In an earlier revision (not pushed) I ended up copying the openapi.json out of the repo and generating a client with Progenitor, and then I realized how much that was slowing down rust-analyzer just for a single API call. So I went with this relatively-small wrapper.

dev-tools/releng/src/hubris.rs
iliana: That would be great!

iliana:
> This looks like it means it's really easy to build a TUF repo via xtask?

That, and running it locally re-uses all the Omicron artifacts too, so iterating on this process is a lot easier. (I think it takes about 8 minutes total if the control plane is previously built; that time is mostly building the OS images.)

david-crespo:
Thanks for the great description, and looking forward to `docs/releng.adoc`. I never really knew what these jobs were doing.

.github/buildomat/jobs/package.sh
```diff
@@ -3,24 +3,11 @@
 #: name = "helios / package"
 #: variety = "basic"
 #: target = "helios-2.0"
-#: rust_toolchain = "1.72.1"
+#: rust_toolchain = "1.77.2"
```
sunshowers, line 6: Is this supposed to stay in sync with `rust-toolchain.toml`? Can we read from that file directly instead? To be honest I didn't know this existed.

.github/buildomat/jobs/package.sh
```diff
@@ -52,88 +28,33 @@ ptime -m cargo run --locked --release --bin omicron-package -- \
...
 
 # Build the xtask binary used by the deploy job
 ptime -m cargo build --locked --release -p xtask
 
-# Assemble some utilities into a tarball that can be used by deployment
-# phases of buildomat.
+# Build the end-to-end tests
+# Reduce debuginfo just to line tables.
+export CARGO_PROFILE_DEV_DEBUG=1
+export CARGO_PROFILE_TEST_DEBUG=1
```
sunshowers, line 40: https://doc.rust-lang.org/cargo/reference/profiles.html#debug suggests that `1` is more than just line-tables-only.

Can we use the full string here?

Cargo.toml
```diff
@@ -336,6 +338,7 @@ nexus-test-utils = { path = "nexus/test-utils" }
 nexus-types = { path = "nexus/types" }
 num-integer = "0.1.46"
 num = { version = "0.4.2", default-features = false, features = [ "libm" ] }
+num_cpus = "1.16.0"
```
sunshowers, line 341: This could use https://doc.rust-lang.org/std/thread/fn.available_parallelism.html instead.

dev-tools/releng/src/cmd.rs
```diff
@@ -0,0 +1,130 @@
...
+            if let Some(value) = value {
+                write!(
+                    command_str,
+                    "{}={} ",
+                    shell_words::quote(&name.to_string_lossy()),
+                    shell_words::quote(&value.to_string_lossy())
+                )
+                .unwrap();
+            }
+        }
```
sunshowers, line 53: Oh good spot.

dev-tools/releng/src/hubris.rs
sunshowers: Hmm, how are these structures different from those in `tufaceous-lib`? Could you add a comment talking about how?

dev-tools/releng/src/job.rs
```diff
@@ -0,0 +1,284 @@
...
+        command: &mut Command,
+    ) -> Selector<'_> {
+        let name = name.as_ref().to_owned();
+        assert!(!self.map.contains_key(&name), "duplicate job name {}", name);
+        self.map.insert(
+            name.clone(),
+            Job {
+                future: Box::pin(spawn_with_output(
+                    // terrible hack to deal with the `Command` builder
+                    // returning &mut
```
sunshowers, line 108: oh no haha

dev-tools/releng/src/main.rs
```diff
@@ -0,0 +1,669 @@
...
+
+/// The base version we're currently building. Build information is appended to
+/// this later on.
+///
+/// Under current policy, each new release is a major version bump, and
+/// generally referred to only by the major version (e.g. 8.0.0 is referred
+/// to as "v8", "version 8", or "release 8" to customers). The use of semantic
+/// versioning is mostly to hedge for perhaps wanting something more granular in
+/// the future.
+const BASE_VERSION: Version = Version::new(8, 0, 0);
```
sunshowers, line 45: Is this going to become the sole source of truth for the system version?

dev-tools/releng/src/main.rs
```diff
@@ -0,0 +1,669 @@
...
+            );
+            preflight_ok = false;
+        }
+    }
+
+    // Ensure the Helios checkout exists
+    if args.helios_dir.exists() {
+        if !args.ignore_helios_origin {
+            // check that our helios clone is up to date
+            Command::new("git")
```
sunshowers, line 242: Should generally consult the `GIT` env var first, then use `git` if that isn't found (applies to rest of this file as well)

dev-tools/releng/src/main.rs
```diff
@@ -0,0 +1,669 @@
...
+            // need for building the OS itself (we are just building an image
+            // from an already-built OS).
+            .env("BUILD_OS", "no"),
+    );
+
+    // Download the toolchain for phbl before we get to the image build steps.
+    // (This is possibly a micro-optimization.)
+    jobs.push_command(
+        "phbl-toolchain",
+        Command::new("cargo")
```
sunshowers, line 349: Same, consult `CARGO` first (this is important because rustup will set it).

dev-tools/releng/src/main.rs
```diff
@@ -0,0 +1,669 @@
...
+            "-m",
+            "cargo",
+            "build",
+            "--locked",
+            "--release",
+            "--bin",
+            "omicron-package",
+        ]),
+    );
+    let omicron_package = WORKSPACE_DIR.join("target/release/omicron-package");
```
sunshowers, line 367: So ideally we'd use `--message-format json-render-diagnostics` to retrieve the path to the binary. That would be resilient to things like a custom target dir being set up. Are you inclined to make that change? (We use the JSON format elsewhere including in xtask.)

dev-tools/releng/src/main.rs
```diff
@@ -0,0 +1,669 @@
...
+        "omicron-package",
+        Command::new("ptime").args([
+            "-m",
+            "cargo",
+            "build",
+            "--locked",
+            "--release",
+            "--bin",
+            "omicron-package",
+        ]),
```
sunshowers, line 365: Consider accepting an `OMICRON_PACKAGE` environment variable for when you want to bring your own binary.

dev-tools/releng/src/main.rs
```diff
@@ -0,0 +1,669 @@
...
+                    .args($image_build_args)
+                    .current_dir(&args.helios_dir)
+                    .env("IMAGE_DATASET", &$image_dataset),
+            )
+            .after("helios-setup")
+            .after(concat!($target_name, "-proto"));
+        };
+    }
+
+    os_image_jobs! {
```
sunshowers, line 473: Nice abstraction! Can this be a function call? If not, then can the reason for that be documented in a comment?

dev-tools/releng/src/job.rs
```diff
@@ -0,0 +1,284 @@
...
+            sender.send(()).ok();
+        }
+        Ok(())
+    }
+}
+
+impl<'a> Selector<'a> {
+    #[track_caller]
+    pub(crate) fn after(self, other: impl AsRef<str>) -> Self {
+        let (sender, receiver) = oneshot::channel();
```
sunshowers, line 157: I've used this pattern before, and I love how it makes respecting the DAG Tokio's responsibility. Can that be made explicit in a comment for those unfamiliar with it?

dev-tools/releng/src/job.rs
```diff
@@ -0,0 +1,284 @@
...
+        tokio::io::stdout(),
+        log_file_1,
+    );
+    let stderr = reader(
+        &name,
+        child.stderr.take().unwrap(),
+        tokio::io::stderr(),
+        log_file_2,
+    );
+    match tokio::try_join!(child.wait(), stdout, stderr) {
```
sunshowers, line 246: This will cancel immediately if any of the operations fail. Is that desirable?

* If so, then can this be documented in a comment?
* If not, then https://docs.rs/cancel-safe-futures/0.1.5/cancel_safe_futures/macro.join_then_try.html exists.

dev-tools/releng/src/job.rs
```diff
@@ -0,0 +1,284 @@
...
+    let prefix_len = buf.len();
+    loop {
+        buf.truncate(prefix_len);
+        let size = reader.read_until(b'\n', &mut buf).await?;
+        if size == 0 {
+            return Ok(());
+        }
+        terminal_writer.write_all(&buf).await?;
+        logfile_writer.write_all(&buf[prefix_len..]).await?;
+    }
```
sunshowers, line 283: Is this the same as https://docs.rs/tokio-stream/0.1.15/tokio_stream/wrappers/struct.LinesStream.html?

dev-tools/releng/src/job.rs
```diff
@@ -0,0 +1,284 @@
...
+
+pub(crate) struct Jobs {
+    logger: Logger,
+    permits: Arc<Semaphore>,
+    log_dir: Utf8PathBuf,
+    map: HashMap<String, Job>,
+}
+
+struct Job {
+    future: Pin<Box<dyn Future<Output = Result<()>>>>,
```
sunshowers, line 46: 1. Can this use https://docs.rs/futures/latest/futures/future/type.BoxFuture.html instead?
2. Should this be `Send`? (`BoxFuture` is `Send`.)

dev-tools/releng/src/job.rs
```diff
@@ -0,0 +1,284 @@
...
+        assert!(!self.map.contains_key(&name), "duplicate job name {}", name);
+        self.map.insert(
+            name.clone(),
+            Job {
+                future: Box::pin(run_job(
+                    self.logger.clone(),
+                    self.permits.clone(),
+                    name.clone(),
+                    future,
+                )),
```
sunshowers, line 88: If you use `BoxFuture` you can switch this to `boxed`: https://docs.rs/futures/latest/futures/future/trait.FutureExt.html#method.boxed

dev-tools/releng/src/job.rs
```diff
@@ -0,0 +1,284 @@
...
+            .wait_for
+            .into_iter()
+            .collect::<FuturesUnordered<_>>()
+            .try_collect::<()>()
+            .await;
+        result.map_err(|_| anyhow!("dependency failed"))?;
+
+        self.future.await?;
+        for sender in self.notify {
+            sender.send(()).ok();
```
sunshowers, line 148: Can we log the situation where the receiver is closed instead of ignoring it?

dev-tools/releng/src/job.rs
```diff
@@ -0,0 +1,284 @@
...
+where
+    F: Future<Output = Result<()>> + 'static,
+{
+    if !PERMIT_NOT_REQUIRED.contains(&name.as_str()) {
+        let _ = permits.acquire_owned().await?;
+    }
+
+    info!(logger, "[{}] running task", name);
+    let start = Instant::now();
+    let result = future.await;
```
sunshowers, line 199: Surprised this doesn't spawn a new Tokio task for the future.

sunshowers (commented):
First-pass review, generally looks good!

.github/buildomat/jobs/package.sh
iliana: Yes:

https://github.com/oxidecomputer/omicron/blob/d2ed4524f17ae025fb2e55956d7c82d377872a3e/rust-toolchain.toml#L1-L7

If it's not in sync it doesn't actually break anything, but the job runs a bit slower (Buildomat downloads 1.72.1 and then rustup's Cargo proxy sees you actually wanted 1.77.2 and downloads that).

Buildomat feature request: https://github.com/oxidecomputer/buildomat/issues/55

dev-tools/releng/src/main.rs
iliana: That's the plan, until we eventually find a better one. (The former source of truth was `.github/buildomat/jobs/package.sh`, which generated a `version.txt` file.)

dev-tools/releng/src/main.rs
iliana: In this particular case (and many other cases) we explicitly _do not_ want that; we are deliberately calling the rustup proxy here to pull down [the toolchain that phbl wants](https://github.com/oxidecomputer/phbl/blob/88977630a62b27f45b6d3ccc394eb3c708f95a44/rust-toolchain). (This is why `CARGO`, `CARGO_MANIFEST_DIR`, and `RUSTUP_TOOLCHAIN` are unset early; the Helios build process uses a menagerie of toolchain versions.)

dev-tools/releng/src/job.rs
iliana: the Command builder is my nemesis.

dev-tools/releng/src/job.rs
iliana: I think the only reason a receiver would close is if the tasks are all cancelled, which would happen if any task fails or if the user hits Ctrl-C. Assuming that's the case, this would only log in a situation where it would be unnecessarily noisy (we already log why a task failed or bubble the error all the way back up).

.github/buildomat/jobs/package.sh
iliana: I think this changed since I first wrote this line (this was moved from ci-tools.sh). I'll investigate.

dev-tools/releng/src/main.rs
iliana: Is your thinking that setting `OMICRON_PACKAGE` would skip adding this job?

dev-tools/releng/src/job.rs
iliana: (thinking about this again has resulted in a slightly less bad hack for this particular problem; I'm going to give it a shot)

dev-tools/releng/src/main.rs
sunshowers: Yeah. It can be pretty useful to copy a binary around without having to pay the price of building it.

dev-tools/releng/src/main.rs
sunshowers: I guess the main alternative is that this lives somewhere in a toml file? I think it's fine as is though.

dev-tools/releng/src/main.rs
sunshowers: Ohhhh yeah. Missed that. Agreed that clear variable separation would be ideal.

dev-tools/releng/src/job.rs
iliana: I don't think it needs to? We poll it immediately here by awaiting, so I'm not sure what creating a new task gets us.

sunshowers: One issue is that, while one future is being executed the others will stay suspended. This can especially be a problem if there are awaits in a loop body. https://without.boats/blog/poll-progress/

Tasks execute independently, so one of them won't be blocked on others.

iliana: Oh, yeah, there are certainly awaits in a loop body here:

https://github.com/oxidecomputer/omicron/blob/6dbca3c52b5355b542824a20baf05eee95fa0e71/dev-tools/releng/src/hubris.rs#L44-L46

I'll put this (and I suppose `async fn reader` later in this file?) in some tasks.

dev-tools/releng/src/job.rs
iliana: It appears the answer to "should this be `Send`" is "yes, because I am about to give it to `tokio::spawn`. (This makes it equivalent to `BoxFuture` so I'll use that.)

dev-tools/releng/src/job.rs
sunshowers: Ah ok -- that seems reasonable then.

dev-tools/releng/src/job.rs
sunshowers: Aha, good catch. Agreed that we shouldn't outright fail on non-UTF-8 output. (Again, can this be put in a comment?)

dev-tools/releng/src/job.rs
sunshowers: I think the readers can generally be within the same task as the jobs themselves, but yeah putting each job in a task would make sense.

docs/releng.adoc
```diff
@@ -0,0 +1,79 @@
...
+Each of these has their own build processes that produce some sort of
+usable artifact, whether that is an illumos zone or a tarball of static
+assets.
+
+The release engineering process builds the control plane and combines
+it with the many external artifacts into a final artifact -- a Zip
+archive of a TUF repository -- that contains everything necessary for
+the product to operate. This process is run on each commit to ensure it
+is always functional. You can also run the process locally with `cargo
+xtask releng`.
```
david-crespo, line 32: this wrapping doesn't render nicely

docs/releng.adoc
```diff
@@ -0,0 +1,79 @@
...
+  This is controlled by the `tools/permslip_production` and
+  `tools/permslip_staging` files.
+. `omicron-package` is the heart of the release engineering process; it
+  reads the manifest from `package-manifest.toml`, runs an appropriate
+  `cargo build` command, downloads any additional artifacts, and
+  packages them into a series of illumos zones and tarballs. (It can
+  also manage installation and uninstallation of these zones; see
+  how-to-run.adoc.)
+. Some of the packaged artifacts are installed directly in the OS
+  images; `cargo xtask releng` unpacks these into a temporary directory
```
david-crespo, line 53: It took me a bit of thinking to understand this — I assume it refers to stuff like putting the console assets directly in the file system of the control plane zone in the host image. But the line as-is is pretty cryptic; it could stand to be two or three sentences.

docs/releng.adoc
```diff
@@ -0,0 +1,79 @@
...
+. Finally, `cargo xtask releng` generates a Zip archive of a
+  https://theupdateframework.io/[TUF] repository, which contains the
+  host and trampoline OS images, the ROT and SP images, and all the
+  illumos zones that are not installed into the OS images. This archive
+  can be uploaded to Wicket to perform an upgrade of the rack while the
+  control plane is not running.
+
+`cargo xtask releng` performs all of these steps in parallel (with
+the temporary exception of artifact downloads handled by
+`tools/install_builder_prerequisites.sh`.)
```
david-crespo, line 68: I would put this at the top. The mention of `cargo xtask releng` in some tasks but not others left me initially wondering how the xtask fits in. I did not realize at first it is doing all of these steps.

docs/releng.adoc
jclulow: Where?

iliana: (I fixed it about three seconds before David posted his note)

dev-tools/releng/src/main.rs
iliana: Using `--message-format json-render-diagnostics` isn't something we can drop into this job runner, since we would need to read or redirect the output and then somehow use that as the program name for Command structs we've already created.

It appears an alternative that would work with custom target directories would be to get the target directory via `cargo metadata --format-version=1 --no-deps | jq -r '.target_directory'` and append `release/omicron-package` to that.

dev-tools/releng/src/main.rs
iliana: I ended up refactoring this particular macro out but another macro slipped in. But I did comment why the new one isn't a function call.

dev-tools/releng/src/cmd.rs
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    async fn ensure_stdout(&mut self, logger: &Logger) -> Result<String> {
+        let output = xtrace(self, logger, Command::output).await?;
+
+        // Obnoxiously, `tokio::process::Command::output` overrides
+        // your stdout and stderr settings (because it doesn't use
+        // std::process::Command::output).
+        //
+        // Compensate by dumping whatever is in `output.stderr` to stderr.
```
sunshowers, line 83: ah interesting, I didn't realize the behavior difference here. Looks like it's this: https://docs.rs/tokio/1.37.0/tokio/process/struct.Command.html#method.output

Can this use the `spawn/wait_for_output` pattern?

dev-tools/releng/src/hubris.rs
sunshowers: Thank you, that's really helpful!

dev-tools/releng/src/job.rs
sunshowers: So on second thought, this does worry me. Current users don't ever use the passed-in `Command` afterwards, but future users might. What do you think of something like:

```rust
push_command("foo-bar", Command::new("foo"), |c| c.arg("bar"))
```

sunshowers (commented):
Looked through again, just a few more comments but pretty happy otherwise!

dev-tools/releng/src/main.rs
iliana: So it turns out you should not trust the `CARGO` set by rustup without also keeping `RUSTUP_TOOLCHAIN` set. The resulting problem is somewhat unintuitive to debug but very funny once you figure out what's going on:

- `CARGO` is set by the rustup proxy to `/home/iliana/.rustup/toolchains/1.77.2-x86_64-unknown-illumos/bin/cargo`.
- When this is run, Cargo searches for rustc in `PATH`, and finds the rustup proxy at `~/.cargo/bin/rustc`.
- The rustup-proxied rustc is invoked to compile some dependency at some path in `~/.cargo/registry`.
- Cargo invokes rustc on a crate that ships a rust-toolchain.toml, or a Git checkout that has a rust-toolchain.toml.
- The rustc proxy, not having any override set in RUSTUP_TOOLCHAIN, proceeds to fetch and use the specified toolchain, then is extremely confused when it sees a dependency that was compiled by somebody else:

```
[ omicron-package] error[E0514]: found crate `syn` compiled by an incompatible version of rustc
[ omicron-package]   --> /home/iliana/.cargo/git/checkouts/opte-c3062ea19cee8fd4/7ee353a/crates/kstat-macro/src/lib.rs:61:13
[ omicron-package]    |
[ omicron-package] 61 |             syn::Fields::Unit => {
[ omicron-package]    |             ^^^
[ omicron-package]    |
[ omicron-package]    = note: the following crate versions were found:
[ omicron-package]            crate `syn` compiled by rustc 1.77.2 (25ef9e3d8 2024-04-09): /home/iliana/git/omicron/target/release/deps/libsyn-39c2edc7c7bde633.rlib
[ omicron-package]    = help: please recompile that crate using this compiler (rustc 1.76.0 (07dca489a 2024-02-04)) (consider running `cargo clean` first)
```

Overall given our dependence on using the rustup proxy to correctly build the Helios build utilities, I think we should stick to using the "cargo" on the user's path universally instead of attempting to consult `CARGO`.

dev-tools/releng/src/job.rs
iliana: I mean at that point I'd rather re-create the surface area of the Command interface we use that instead returns `self`, because I've already defined five new methods in an extension trait.

dev-tools/releng/src/cmd.rs
iliana: Oh I didn't notice that. Probably, yeah.

dev-tools/releng/src/main.rs
iliana: Actually, now I have a commit I'm testing locally that doesn't unset any environment at the start, and unsets environment variables when needed when running particular commands. This should allow for, say `cargo +nightly xtask releng` while still using the correct toolchains for the Helios stuff.

iliana:
I think I've addressed all outstanding review feedback; I still need to figure out a pattern for external xtasks (clap is currently making me unhappy), but I think that's the last TODO here.

dev-tools/releng/src/cmd.rs
```diff
@@ -0,0 +1,167 @@
...
+
+    pub(crate) fn into_parts(self) -> (Description, tokio::process::Command) {
+        (Description { str: self.to_string() }, self.inner)
+    }
+}
+
+impl std::fmt::Display for Command {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let command = self.inner.as_std();
+        for (name, value) in command.get_envs() {
```
sunshowers, line 102: How noisy is this? I think this would print out the entire environment, which would be a lot: on my machine, `export | wc -c` returns 4923.

dev-tools/releng/src/main.rs
```diff
@@ -0,0 +1,734 @@
...
+/// used to update a rack.
+///
+/// For more information, see `docs/releng.adoc` in the Omicron repository.
+///
+/// Note that `--host-dataset` and `--recovery-dataset` must be set to different
+/// values to build the two OS images in parallel. This is strongly recommended.
+struct Args {
+    /// ZFS dataset to use for `helios-build` when building the host image
+    #[clap(long, default_value_t = Self::default_dataset("host"))]
+    host_dataset: String,
```
sunshowers, line 110: nit: is there some validation we can do on the dataset names? Would that be worth it at all?

sunshowers (approved):
love it, thanks for doing this work!

dev-tools/releng/src/cmd.rs
iliana: It only prints the environment variables explicitly set in the command. https://doc.rust-lang.org/stable/std/process/struct.Command.html#method.get_envs

dev-tools/releng/src/main.rs
iliana: I think verifying the datasets already exist (one of the preflight checks we run) is sufficient.

iliana:
Despite the documentation all saying `cargo xtask releng` I think I am going to go ahead and merge this so we get the CI time improvements. For now folks can use `cargo run --release --bin omicron-releng`. I'm still toying with a few different options for external xtasks and I think that probably deserves its own separate review.

PR #5745, opened 2024-05-13
[omdb] Show physical disks in blueprints and inventory

### LINKED ISSUE #5624: [omdb] Add support for showing physical disks in blueprints and inventory

andrewjstone:
This PR changes the construction and display of blueprint diffs significantly. The need for the changes arose out of the tight coupling between zones and sleds. When adding disks to the mix we needed to also take them into account when determininig when a sled was modified. We also want to display physical disks as tables along with the zone tables under the sleds. This turned out to be somewhat trickier than necesary, and so I changed how the tables were constructed and rendered. Hopefully this will also make it easier to add new tables in the future.

One possibly controversial change is that I changed the way zone modifications are rendered. They are no longer three lines long. Currently only `disposition` is allowed to change, and so I used an arrow from "old" to "new" versions. Additionally, I removed the (added/ modified/removed) suffixes as they seem redundant to me, make the lines longer, and make things harder to read IMO. I'm open to discussion about both of these changes. My guess is that eventually, we'll want to be able to do row filtering, and I plan to also add some colored output, also in this PR most likely.

It should be noted that the output mechanism is decoupled from the representation of the tables, `BpSledSubtable`. This allows output in other formats if desired in the future.

As for the inventory collection output, I added a requirement when filtering on sled-id, that we also filter on collection-id, because filtering on sled-id does a full table scan and requires a new index. I can add the index instead if we feel it's important.

Fixes #5624


nexus/types/src/deployment.rs
```diff
@@ -268,9 +379,64 @@ impl<'a> fmt::Display for BlueprintDisplay<'a> {
...
+                        zones.bp_generation(),
+                        zones.rows(BpDiffState::Unchanged).collect(),
+                    )
+                    .to_string()
+                }
+                None => "".to_string(),
+            };
+            writeln!(
+                f,
+                "\n  sled: {sled_id}\n\n{disks_table}\n\n{zones_table}\n"
```
jgallagher, line 414: Ah, this might be the source of all the blank space I saw above - maybe move this into the `match` and omit `\n\n{zones_table}` in the `None` case?

nexus/types/src/deployment/blueprint_diff.rs
```diff
@@ -0,0 +1,804 @@
...
+
+                for (zone_id, zone_before) in before_by_id {
+                    if let Some(zone_after) = after_by_id.remove(&zone_id) {
+                        // Are the zones equal?
+                        if zone_before == zone_after {
+                            unchanged.push(zone_after.into());
+                        } else {
+                            // The zones are different. They are only allowed to differ in terms
+                            // of `disposition`, otherwise we have an error.
+                            match zone_before.is_valid_modification(&zone_after)
```
jgallagher, line 166: Nit - I'd expect `is_*` to return a bool. In the spirit of "parse, don't validate", could we make this `ModifiedZone`'s constructor instead? That would also centralize the decision for "what is allowed to change".

nexus/types/src/deployment/blueprint_diff.rs
```diff
@@ -0,0 +1,804 @@
...
+    /// 3. Modified
+    /// 4. Added
+    ///
+    /// The idea behind the order is to (a) group all changes together
+    /// and (b) put changes towards the bottom, so people have to scroll
+    /// back less.
+    ///
+    /// Errors are printed in a more freeform manner after the table is
+    // displayed.
+
```
jgallagher, line 307:
```suggestion
    /// displayed.
```

(and probably also trim the newline between the doc comment and the function?)

nexus/types/src/deployment/blueprint_diff.rs
```diff
@@ -0,0 +1,804 @@
...
+            Some(BpSledSubtable::new(
+                BpPhysicalDisksSubtableSchema {},
+                generation,
+                rows,
+            ))
+        }
+    }
+}
+
+// TODO: Make this pub and remove the other from deployment.rs
```
jgallagher, line 502: this is done I assume?

nexus/types/src/deployment/blueprint_display.rs
```diff
@@ -0,0 +1,331 @@
...
+    pub fn new(state: BpDiffState, columns: Vec<String>) -> Self {
+        BpSledSubtableRow { state, columns }
+    }
+}
+
+/// Metadata about all instances of specific type of [`BpSledSubtable`],
+/// such as omicron zones or physical disks.
+pub trait BpSledSubtableSchema {
+    fn table_name(&self) -> &'static str;
+    fn column_names(&self) -> Vec<&'static str>;
```
jgallagher, line 116: This requires a new `Vec` every time it's called. Probably fine for diffs, but could it be `&[&'static str]` instead?

oximeter/collector/tests/output/self-stat-schema.json
```diff
@@ -39,7 +39,7 @@
       }
     ],
     "datum_type": "cumulative_u64",
-    "created": "2024-05-03T22:37:51.326086935Z"
+    "created": "2024-05-11T09:41:23.361298682Z"
```
jgallagher, line 42: Was this supposed to be committed?

nexus/reconfigurator/planning/src/blueprint_builder/zones.rs
```diff
@@ -346,6 +346,11 @@ mod tests {
                 .expect("new zone ID should be present");
         }
 
+        // TODO: FIXME: AJS - Delete use of the control sled? - We don't
```
jgallagher, line 349: Do you want to address this on this PR?

nexus/reconfigurator/planning/src/blueprint_builder/zones.rs
```diff
@@ -346,6 +346,11 @@ mod tests {
...
+        // actually compute changes based on generation number anymore. Zones or
+        // physical disks must change for a sled to be modified. It's not clear
+        // to me that the way we use this in this test is actually useful.
+        //
         // Also call change_sled_zones without making any changes. This
         // currently bumps the generation number, but in the future might
         // become smarter and not do so (in which case this test will break).
```
jgallagher, line 356: Should we make this change and make `change_sled_zones` smarter? (Happy to defer this also, but it seems like it would address your TODO?)

nexus/reconfigurator/planning/src/planner.rs
```diff
@@ -483,20 +492,18 @@ mod test {
...
+        assert!(sled_zones.generation_after.unwrap() > Generation::new());
         assert_eq!(sled_id, new_sled_id);
         assert_eq!(sled_zones.zones.len(), 1);
-        assert!(matches!(
-            sled_zones.zones[0].zone_type,
-            BlueprintZoneType::InternalNtp(_),
-        ));
-        assert_eq!(diff.sleds_removed().len(), 0);
-        assert_eq!(diff.sleds_modified().count(), 0);
+        assert!(matches!(sled_zones.zones[0].kind(), ZoneKind::InternalNtp,));
```
jgallagher, line 504:
```suggestion
        assert!(matches!(sled_zones.zones[0].kind(), ZoneKind::InternalNtp));
```

nexus/types/src/deployment.rs
andrewjstone: Done!

dev-tools/omdb/tests/successes.out
```diff
@@ -501,27 +501,38 @@ stdout:
...
-    nexus             ..........<REDACTED_UUID>...........   in service   ::ffff:127.0.0.1 
-
-METADATA:
-  created by:            nexus-test-utils        
-  created at:            <REDACTED     TIMESTAMP>
-  comment:               initial test blueprint  
-  internal DNS version:  1                       
-  external DNS version:  2                       
+!..........<REDACTED_UUID>...........
+WARNING: Zones exist without physical disks!
```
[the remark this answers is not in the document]
andrewjstone: Good catch! Fixed.

nexus/types/src/deployment/blueprint_diff.rs
andrewjstone: Excellent idea. Done.

nexus/types/src/deployment/blueprint_diff.rs
andrewjstone: fixed.

nexus/types/src/deployment/blueprint_diff.rs
andrewjstone: yep. good catch. removed.

dev-tools/omdb/tests/successes.out
```diff
@@ -501,27 +501,38 @@ stdout:
...
+    omicron zones at generation 2:
+    -----------------------------------------------
+    zone type   zone id   disposition   underlay IP
+    -----------------------------------------------
+
+
```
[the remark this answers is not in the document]
andrewjstone: Fixed I think :)

nexus/reconfigurator/planning/src/planner.rs
andrewjstone: Fixed.

nexus/types/src/deployment/blueprint_display.rs
andrewjstone: Fixed.

oximeter/collector/tests/output/self-stat-schema.json
andrewjstone: Hmm, it changed. But maybe I shouldn't have committed it?

nexus/reconfigurator/planning/src/blueprint_builder/zones.rs
andrewjstone: I removed this section of code, as it's unused.

nexus/reconfigurator/planning/src/blueprint_builder/zones.rs
andrewjstone: I didn't make this any smarter. I don't think it matters for my todo, and is definitely outside the scope of this PR. This is especially true because I don't know how to make it smarter :)

andrewjstone:
There's actually a bug here where we're generating new output on every test run because UUIDs for physical disks aren't fixed.

There's another one, where we want to sort by zone type rather than uuid, but we don't. 

I'll fix both of these up.

andrewjstone:
> There's actually a bug here where we're generating new output on every test run because UUIDs for physical disks aren't fixed.
> 
> There's another one, where we want to sort by zone type rather than uuid, but we don't.
> 
> I'll fix both of these up.

These are now fixed.

I think this PR is good to go. I haven't yet changed back the diffs to be multiline, but I can do that in a follow up immediately. There are some changes that make writing tests slightly different which I'd like to get in before the blueprint stuff gets refactored. There are also a bunch of other changes to make the test system more deterministic by using seeded TypedUuids everywhere I noticed an issue.

oximeter/collector/tests/output/self-stat-schema.json
jgallagher: I think I would vote "don't commit it" along with possibly file an issue? I'm surprised this isn't an issue on main...

bnaecker: This works because we explicitly ignore timestamps when comparing the schema on disk with the current one. It's fine to commit, or not, whatever you'd like.

andrewjstone: That makes sense. Thanks @bnaecker. I believe it changed when I ran all the tests with `cargo nextest run`. I'll leave it out of sheer laziness.

jgallagher (approved):
:tada:

PR #5750, opened 2024-05-13
re-use reqwest::Client in disk_manual_import

### LINKED ISSUE #5717: nexus does not re-use CruciblePantryClient during disk imports.
While attempting to upload 64 400-MiB disk images concurrently to **madrid** (4-sled racklette), I caused Nexus to consume 100% of the sled's CPU resources. These uploads were extremely slow, and over 50% of them failed to fully upload their disks.

**madrid** is running https://github.com/oxidecomputer/omicron/commit/fcf7980c37328b5c5b27d864d476957fbc210a19

You can read the full adventure here https://github.com/oxidecomputer/artemiscellaneous/blob/main/journal/2024-05-06-slow-disk-uploads.md#adventures-in-parallel-boot-disk-uploads

But here's the important bits:

I uploaded the images like this:

```
seq 1 64 | xargs -I '%' -P 64 sh -c '
    oxide disk import \
    --project iodriver --description "iodriver boot medium" \
    --path "/dev/shm/iodriver.iso" \
    --disk "iodriver-boot-%"
'
```

Each image uploaded at about 300 KiB/s, or about 18MiB/s aggregate average. This generated tremendous resource contention in nexus.

**prstat from the sled with nexus**:

```
root@BRM42220046:~# prstat
Total: 332 processes, 12302 lwps, load averages: 127.20, 120.51, 83.41
   PID USERNAME  SIZE   RSS STATE  PRI NICE      TIME  CPU PROCESS/NLWP
 12198 root     1007M  943M cpu96    0    0  84:49:00  98% nexus/145
   163 root     7156K 5616K sleep   59    0   0:40:44 0.2% devfsadm/14
  4207 root      792M  606M sleep   59    0   0:16:42 0.1% cockroach/136
     5 root        0K    0K sleep   98  -20   0:02:22 0.0% zpool-rpool/327
  1496 root        0K    0K sleep   98  -20   0:04:13 0.0% zpool-oxp_4a1be/327
   652 root       92M   57M sleep   59    0   0:05:02 0.0% sled-agent/135
  1569 root        0K    0K sleep   98  -20   0:01:30 0.0% zpool-oxp_bc8f5/327
 10782 root       57M   47M sleep   59    0   0:00:02 0.0% dtrace/2
  1546 root        0K    0K sleep   98  -20   0:01:00 0.0% zpool-oxp_8ada9/327
   142 root       33M   29M sleep   59    0   0:00:02 0.0% syseventd/19
   587 root     4540K 2392K sleep   59    0   0:00:02 0.0% syslogd/10
   543 root     2152K  728K sleep   59    0   0:00:00 0.0% utmpd/1
   484 root     7444K 4944K sleep   59    0   0:00:06 0.0% nscd/23
    48 root     3076K 1996K sleep   59    0   0:00:00 0.0% dlmgmtd/7
   732 root        0K    0K sleep   98  -20   0:00:00 0.0% zpool-oxi_f32c2/327
```

**flamegraph of nexus** (view raw for interactivity)

disk-upload-flamegraph-nexus

I was uploading these images from an instance in dogfood which is located in the same physical place as madrid, meaning I had a high-speed connection. You can see the graph of network traffic **from the dogfood instance** is very sporadic despite this.

image

The problem, ultimately, is that Nexus is not caching its `CruciblePantryClient`. The `oxide` cli splits the image up into 512KiB blocks. Each time one of these 512KiB blocks is uploaded, Nexus creates a new `CruciblePantryClient` from scratch, and pays all the overhead of setting up that client. The flamegraph shows the overhead is mostly down to Reqwest preparing a TLS-capable connector (though it won't use TLS at all for this connection), which causes openssl to read & parse the system trusted certificate roots.

We are managing to upload 37 of these blocks per second, but we have `64 CLI invocations * 8 worker threads per invocation =` **512** client worker threads fighting over this throughput, which leads to a number of them failing, presumably due to connection timeouts.

This problem is probably not unique to massively parallel uploads, and uploading a few images in parallel likely generates a considerable load (though much less contention), but I have not measured this.

Caching the `CruciblePantryClient` between block uploads should make this a lot better.

iliana:
Fixes #5717. @faithanalog will follow-up with performance numbers she's running now.

We should probably do this to other clients we generate on-the-fly in Nexus, too, but this is probably the worst offender to fix.

### DIFF
--- a/nexus/src/app/disk.rs
+++ b/nexus/src/app/disk.rs
@@ -488,10 +488,10 @@ impl super::Nexus {
             // that user's program can act accordingly. In a way, the user's
             // program is an externally driven saga instead.
 
-            let client = crucible_pantry_client::Client::new(&format!(
-                "http://{}",
-                endpoint
-            ));
+            let client = crucible_pantry_client::Client::new_with_client(
+                &format!("http://{}", endpoint),
+                self.reqwest_client.clone(),
+            );
             let request = crucible_pantry_client::types::BulkWriteRequest {
                 offset: param.offset,
                 base64_encoded_data: param.base64_encoded_data,
--- a/nexus/src/app/mod.rs
+++ b/nexus/src/app/mod.rs
@@ -152,6 +152,13 @@ pub struct Nexus {
     /// The metric producer server from which oximeter collects metric data.
     producer_server: std::sync::Mutex<Option<ProducerServer>>,
 
+    /// Reusable `reqwest::Client`, to be cloned and used with the Progenitor-
+    /// generated `Client::new_with_client`.
+    ///
+    /// (This does not need to be in an `Arc` because `reqwest::Client` uses
+    /// `Arc` internally.)
+    reqwest_client: reqwest::Client,
+
     /// Client to the timeseries database.
     timeseries_client: LazyTimeseriesClient,
 
@@ -343,6 +350,12 @@ impl Nexus {
             }
         }
 
+        let reqwest_client = reqwest::ClientBuilder::new()
+            .connect_timeout(std::time::Duration::from_secs(15))
+            .timeout(std::time::Duration::from_secs(15))
+            .build()
+            .map_err(|e| e.to_string())?;
+
         // Connect to clickhouse - but do so lazily.
         // Clickhouse may not be executing when Nexus starts.
         let timeseries_client = if let Some(address) =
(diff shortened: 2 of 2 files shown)


faithanalog:
Here's a look at things uploading 64 disks in parallel from dogfood -> london

In aggregate I'm seeing a pretty stable 1.1 to 1.2gbit/s on my dogfood VM

saturating the link

On london, nexus is only using 5-6% CPU. previously it was using 100% of a sled's CPU (for a fraction of the upload throughput)

nexus_6pct_2

Interesting to note that since we are processing more requests now, cockroach CPU usage is a bit higher on one of the other sleds. I think this is probably auth handling? Doesn't seem to be a problem at the moment.

cockroachforsomereason

Basically we are uploading like 10x faster while using 1/20th the CPU, and probably not bottlenecking in this part of the process anymore. You love a `+18 -4` that makes the thing be good.

PR #5751, opened 2024-05-14
Add the server kind to the Nexus API context

### LINKED ISSUE #5651: Restrict IP pool size to prevent CRDB memory budget exceeded error
I hit the error accidentally while blindly plugging in a large CIDR start/end IP range, e.g. 240.0.0.1 - 240.4.255.254.

Nexus failed the request during floating IP create request against such a large IP range in the pool I chose:
```
00:47:16.243Z INFO 65a11c18-7f59-41ac-b9e7-680627f996e7 (dropshot_external): request completed
    error_message_internal = unexpected database error: root: memory budget exceeded: 4218880 bytes requested, 133364712 currently allocated, 134217728 bytes in budget
    file = /home/build/.cargo/git/checkouts/dropshot-a4a923d29dccc492/29ae98d/dropshot/src/server.rs:837
    latency_us = 624875
    local_addr = 172.30.2.5:443
    method = POST
    remote_addr = 172.20.17.42:50856
    req_id = 2db527c4-aa8f-4af4-96c7-581e2e5db726
    response_code = 500
    uri = https://silo11.sys.rack2.eng.oxide.computer/v1/floating-ips?project=bob
```

This is very much a corner case but having some upfront validation on the pool size will be a good thing. I poked a bit into what that maximum is and it looks like `/14` works fine and covers an insanely large number of addresses.

### LINKED ISSUE #5735: Check for remote address in allow-list update should ignore underlay
When attempting to update the allow-list for user-facing services, Nexus ensures that the address of the peer sending the request is contained in the allow-list itself. This is a safety mechanism to prevent locking oneself out. However, there are still ways one can get locked out. As an example, if a user sets the allow-list to enable traffic from some external proxy, but that proxy is issued a new IP address without notice, they won't be able to get back into the rack to change the list.

To address this, we could make use of the Nexus external API proxy offered by `wicketd`. This works at the transport layer, and runs over the _underlay_ network. That means it's not subject to the allow-list checks at all, since it's over a completely different interface. However, clients will never be able to actually update the allow-list from the proxy, because the containment mentioned at the beginning will fail! Users don't know (and can't know) the underlay address, so Nexus will always reject the list.

We should relax this check in Nexus. The remote address must either be on the allow-list, or on the underlay network. It's not yet clear to me if that means "same rack subnet" or "any rack subnet", or if we want to do an explicit lookup in the database to make sure it's in a known-valid rack.

bnaecker:
- Adds a server kind enum, used to distinguish which API server is running any particular handler.
- Wraps the existing `ServerContext` into a higher-level `ApiContext` type, which includes the former in addition to the kind of server handling the request.
- Fixes #5735

### DIFF
--- a/nexus/src/app/allow_list.rs
+++ b/nexus/src/app/allow_list.rs
@@ -57,6 +57,10 @@ impl super::Nexus {
             // Note that we elide this check when handling a request proxied
             // from `wicketd`. This is intentional and used as a safety
             // mechanism in the even of lockout or other recovery scenarios.
+            let check_remote_addr = match server_kind {
+                ServerKind::External => true,
+                ServerKind::Techport | ServerKind::Internal => false,
+            };
             let mut contains_remote = false;
             for entry in list.iter() {
                 contains_remote |= entry.contains(remote_addr);
@@ -74,8 +78,7 @@ impl super::Nexus {
                     ));
                 }
             }
-            if !contains_remote && !matches!(server_kind, ServerKind::Techport)
-            {
+            if check_remote_addr && !contains_remote {
                 return Err(Error::invalid_request(
                     "The source IP allow list would prevent access \
                     from the current client! Ensure that the allowlist \
--- a/nexus/src/app/allow_list.rs
+++ b/nexus/src/app/allow_list.rs
@@ -13,6 +13,8 @@ use omicron_common::api::external;
 use omicron_common::api::external::Error;
 use std::net::IpAddr;
 
+use crate::context::ServerKind;
+
 impl super::Nexus {
     /// Fetch the allowlist of source IPs that can reach user-facing services.
     pub async fn allow_list_view(
@@ -30,6 +32,7 @@ impl super::Nexus {
         &self,
         opctx: &OpContext,
         remote_addr: IpAddr,
+        server_kind: ServerKind,
         params: params::AllowListUpdate,
     ) -> Result<AllowList, Error> {
         if let external::AllowedSourceIps::List(list) = &params.allowed_ips {
@@ -50,6 +53,10 @@ impl super::Nexus {
             // the request came from is on the allowlist. This is our only real
             // guardrail to prevent accidentally preventing any future access to
             // the rack!
+            //
+            // Note that we elide this check when handling a request proxied
+            // from `wicketd`. This is intentional and used as a safety
+            // mechanism in the even of lockout or other recovery scenarios.
             let mut contains_remote = false;
             for entry in list.iter() {
                 contains_remote |= entry.contains(remote_addr);
@@ -67,7 +74,8 @@ impl super::Nexus {
                     ));
                 }
             }
-            if !contains_remote {
+            if !contains_remote && !matches!(server_kind, ServerKind::Techport)
+            {
                 return Err(Error::invalid_request(
                     "The source IP allow list would prevent access \
                     from the current client! Ensure that the allowlist \
--- a/nexus/src/app/sagas/disk_create.rs
+++ b/nexus/src/app/sagas/disk_create.rs
@@ -1093,7 +1093,7 @@ pub(crate) mod test {
         let log = &cptestctx.logctx.log;
 
         let client = &cptestctx.external_client;
-        let nexus = &cptestctx.server.apictx.context.nexus;
+        let nexus = &cptestctx.server.server_context().nexus;
         let project_id =
             create_project(&client, PROJECT_NAME).await.identity.id;
         let opctx = test_opctx(&cptestctx);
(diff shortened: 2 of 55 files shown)


bnaecker:
The proximal reason for doing this is #5651, to elide the check when setting the source IP allow-list that ensures you can't lock yourself out. That should help in recovery situations.

I'm a bit on the fence about this. It feels like the nuclear option for a nearly trivial fix (adding a check on the server kind in the allow-list update handler). While it does make the server kind available in an API handler, that has to be passed in as an argument to any other function which needs it. That's nice and explicit, but is a bit counter to the shared state we've got smeared around Nexus in most other situations.

Still, it does solve the original problem, and might be useful in the future for code that needs to know more about which server is handling any particular request.

nexus/src/app/allow_list.rs
```diff
@@ -67,7 +74,8 @@ impl super::Nexus {
                     ));
                 }
             }
-            if !contains_remote {
+            if !contains_remote && !matches!(server_kind, ServerKind::Techport)
```
jgallagher, line 77: Tiny nit - I've started to really dislike `matches!` because it's effectively a match with a `_ => false` arm, and I don't love catchall arms like that. What would you think about lifting this up above the loop into a full match; something like

```rust
let require_remote_addr = match server_kind {
    // can move the comment about why we treat this to here
    ServerKind::TechPort => false,
    ServerKind::External => true,
};
```

and now that I write that, I'm not sure what `ServerKind::Internal` should be? Probably `false` - if that's right, it's another ding against `matches!`.

nexus/src/context.rs
```diff
@@ -29,6 +29,61 @@ use std::str::FromStr;
...
+    /// The kind of server.
+    pub kind: ServerKind,
+    /// Shared state available to all endpoint handlers.
+    pub context: Arc<ServerContext>,
+}
+
+impl ApiContext {
+    /// Create a new context with a rack ID and logger. This creates the
+    /// underlying `Nexus` as well.
+    pub async fn new(
```
jgallagher, line 60: I think I'd name this `for_internal` (without changing the arguments). Then it's consistent with the other names and obvious that it creates a `ServerKind::Internal` context.

jgallagher (approved):
I get why you're on the fence with this (and thank you for doing such a monotonous set of changes to do this), but I think I'm on board. I would not be offended if you want to get a second or third opinion, though! :)

nexus/src/context.rs
bnaecker: Yeah, I had that originally and switched to make it clear we have to start with an internal kind. But that's obvious from the signature, I'll switch it back!

nexus/src/context.rs
```diff
@@ -29,6 +29,61 @@ use std::str::FromStr;
...
+        Self { kind: ServerKind::External, context: self.context.clone() }
+    }
+
+    /// Clone self for use by the techport Dropshot server.
+    pub fn for_techport(&self) -> Self {
+        Self { kind: ServerKind::Techport, context: self.context.clone() }
+    }
+}
+
+impl std::borrow::Borrow<ServerContext> for ApiContext {
```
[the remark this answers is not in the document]
bnaecker: We do need borrow here, to satisfy this trait bound:

https://github.com/oxidecomputer/omicron/blob/8c38ad1f07f90b6caca0062a6be5a3aabb5e6e79/nexus/db-queries/src/authn/external/mod.rs#L49-L54

I agree that it seems a bit weird, and I'm not sure I like the trait bound itself. I kept it simply to minimize churn, but I'm happy to reevaluate.

I also considered implementing `Deref`. I think it might make some other changes unnecessary, but there actually didn't seem to be many `ServerContext` methods used themselves. Most of the fields are visible and used directly. I also agree it's a bit magical sometimes, so I think it makes sense to defer it.

nexus/src/app/sagas/disk_create.rs
```diff
@@ -1093,7 +1093,7 @@ pub(crate) mod test {
         let log = &cptestctx.logctx.log;
 
         let client = &cptestctx.external_client;
-        let nexus = &cptestctx.server.apictx.nexus;
+        let nexus = &cptestctx.server.apictx.context.nexus;
```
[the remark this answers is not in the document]
bnaecker: Done in 85d66c0ed

nexus/src/app/allow_list.rs
bnaecker: Done in 85d66c0ed

nexus/src/context.rs
bnaecker: Renamed in 85d66c0ed

bnaecker:
Whoops, I've linked to the wrong issue here. This fixes #5735. I'll amend the commit message to reflect that now.

nexus/src/context.rs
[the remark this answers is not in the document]
jgallagher: TIL! I'd definitely defer any reevaluation of that for some followup work (which is probably not necessary at all - implementing `Borrow` is surprising but actually required - that's fine).

PR #5698, opened 2024-05-03
[reconfigurator] Decommission sleds

jgallagher:
This builds on (and is currently pointed at) #5663, and is on the path to fixing #5625, although there is still more work to do there. (Should be small, and I'll start on it while this is in review to make sure it really fixes it.) I don't plan on landing this before R8 is out the door, but wanted to go ahead and open it up to review.

Much of this is fallout from our discussions about what it means to decommission a sled. Major changes are:

* `SledFilter::All` has been renamed to `SledFilter::Commissioned`, and no longer returns sleds that are in `SledState::Decommissioned`.
* The blueprint planner will now update the desired sled state to `Decommissioned` for sleds which satisfy our conditions. (See `do_plan_decommission()`.)
* The blueprint planner will carry forward the Omicron zones of decommissioned sleds to child blueprints. Pruning these is #5552.
* The blueprint planner will _not_ carry forward a desired sled state of `Decommissioned` once the inputs report that the sled has already been decommissioned.
* The blueprint executor will decommission sleds that the planner said to.
* Decommissioning a sled implicitly decommissions all its disks. (This matches what happens with sled expungement, and should not interfere with region replacement, which keys off of policy, not state.)

PR #5765, opened 2024-05-14
fix #5698 vs. #5751 semantic merge conflict

iliana:
Semantic merge conflict between:
- #5698
- #5751


iliana:
Got the first pull request wrong, fixed in the PR title so that will end up in the final commit.

sunshowers (approved):

PR #5733, opened 2024-05-10
Allow recommissioning of previously-decommissioned sleds

jgallagher:
This is a pretty small delta on top of #5698 that I've been sitting on for a few days before getting a chance to test it on a4x2 this afternoon. I set up a4x2 with three sleds (g0, g1, g3), and went through 3.5 add/remove cycles on sled g2 (3 add/remove pairs and a 4th add). g2 is present in `omdb db sleds`:

```
root@oxz_switch:~# omdb db sleds
 SERIAL  IP                             ROLE      POLICY      STATE   ID
 g3      [fd00:1122:3344:103::1]:12345  scrimlet  in service  active  13b8d30d-b66d-4333-b156-9aa2527e130b
 g2      [fd00:1122:3344:124::1]:12345  -         in service  active  218e4e9c-0a27-4460-a65c-7e93bef531c4
 g1      [fd00:1122:3344:102::1]:12345  -         in service  active  661e9e3e-0beb-43fe-9606-109b7145b258
 g0      [fd00:1122:3344:101::1]:12345  scrimlet  in service  active  9f19235f-bfa4-4032-9cad-608448b4f1a0
```

I added `SledFilter::Decommissioned` specifically for use with omdb, where we can see the 3 previous times `g2` was present (each with a different sled ID):

```
root@oxz_switch:~# omdb db sleds -F decommissioned
 SERIAL  IP                             ROLE  POLICY    STATE           ID
 g2      [fd00:1122:3344:121::1]:12345  -     expunged  decommissioned  a7b09236-c8ba-4025-934d-b5f0539d9379
 g2      [fd00:1122:3344:123::1]:12345  -     expunged  decommissioned  ec0e81bf-6366-4273-97bc-47223334dc90
 g2      [fd00:1122:3344:122::1]:12345  -     expunged  decommissioned  f5f8ba44-9681-4e6f-84f2-fd654c83dd23
```

The current blueprint shows history of all four times the sled has been present, because we don't currently prune expunged zones:

```
root@oxz_switch:~# omdb nexus blueprints show current
blueprint  5bfb182e-df8a-41e5-a8a3-862bb42f8feb
parent:    5b2823c8-ca41-46b5-9efd-ffaae0e6e028

  -----------------------------------------------------------------------------------------------
    zone type         zone ID                                disposition  underlay IP
  -----------------------------------------------------------------------------------------------

  sled 13b8d30d-b66d-4333-b156-9aa2527e130b: blueprint zones at generation 5
[...]

PR #5767, opened 2024-05-14
fix #5733 vs #5751 semantic merge conflict

jgallagher:
same problem and fix as #5765


iliana (approved):

sunshowers:
Think we may need to combine this with https://github.com/oxidecomputer/omicron/pull/5765 into one PR.

sunshowers:
Oh maybe not, maybe just rebase this onto https://github.com/oxidecomputer/omicron/pull/5765.

PR #5768, opened 2024-05-14
[nexus] deflake `test_instance_watcher_metrics`

### LINKED ISSUE #5752: test flake: omicron-nexus::test_all integration_tests::metrics::test_instance_watcher_metrics
Example:

https://buildomat.eng.oxide.computer/wg/0/details/01HXT52WBZWPKP4HRFE5BZEJ90/TzkBQ5mcdy8cXJtRPqbP8sTUggD48AxTs1uwGGPeMg15q7e7/01HXT53AB77VSYQFJ33G6K5FSE

I feel like I saw something similar go by at some point.

hawkw:
Presently, `test_instance_watcher_metrics` will wait for the
`instance_watcher` background task to have run before making assertions
about metrics, but it does *not* ensure that oximeter has actually
collected those metrics. This can result in flaky failures --- see
#5752.

This commit adds explicit calls to `oximeter.force_collect()` prior to
making assertions, to ensure that the latest metrics have been
collected.

Fixes #5752

### DIFF
--- a/nexus/tests/integration_tests/metrics.rs
+++ b/nexus/tests/integration_tests/metrics.rs
@@ -329,6 +329,7 @@ async fn test_instance_watcher_metrics(
     let client = &cptestctx.external_client;
     let internal_client = &cptestctx.internal_client;
     let nexus = &cptestctx.server.server_context().nexus;
+    let oximeter = &cptestctx.oximeter;
 
     // TODO(eliza): consider factoring this out to a generic
     // `activate_background_task` function in `nexus-test-utils` eventually?
@@ -399,6 +400,8 @@ async fn test_instance_watcher_metrics(
         )
         .await
         .unwrap();
+        // Make sure that the latest metrics have been collected.
+        oximeter.force_collect().await;
     };
 
     #[track_caller]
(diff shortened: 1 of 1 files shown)


iliana:
(should be good to rebase now)

bnaecker (approved):

nexus/tests/integration_tests/metrics.rs
```diff
@@ -399,6 +400,8 @@ async fn test_instance_watcher_metrics(
         )
         .await
         .unwrap();
+        // Make sure that the latest metrics have been collected.
+        oximeter.force_collect().await;
     };
```
iliana, line 405: I'm looking at this change since it didn't seem to help with the flakiness -- I don't know that this would do anything? The line immediately following each use of the `activate_instance_watcher` closure is an awaited call to `timeseries_query`, which performs a `force_collect` first thing:

https://github.com/oxidecomputer/omicron/blob/e0a118496b05a30bf7a5a8a0820e0dda415c485a/nexus/tests/integration_tests/metrics.rs#L283-L288

hawkw: Huh, okay, I'm not sure if I understand why the test was flaky in the first place, then! I'll have to keep digging.

bnaecker: Oh, yeah I definitely missed that. I've honestly found it frustrating to write metrics tests that assert exact equalities like this. There are just too many programs that the data has to move through, all of which are running a mess of Tokio tasks and the like (the producer, `oximeter`, ClickHouse, the test program, etc.). I've generally used inequalities if one wants to check things like the amount of data; or checks that there are samples with timestamps after a certain point; or some value exists in the data stream; etc.

PR #5769, opened 2024-05-15
Remove unused `base_route` field from oximeter producer info

### LINKED ISSUE #5658: Remove `base_route` from oximeter producer information
As part of https://github.com/oxidecomputer/omicron/issues/5284, we've updated `oximeter` producers so that they periodically and automatically renew their registrations with Nexus. The best way to do that removed the need for the `base_route` field in the `ProducerEndpoint` struct:

https://github.com/oxidecomputer/omicron/blob/bca1f2d63b70e71aef509b94e190e54c6eb3a6be/common/src/api/internal/nexus.rs#L113-L115

This issue tracks removing that entirely from the API and database.

bnaecker:
- Remove from DB schema, add migration to drop the column
- Remove all references, except in the timeseries schema reporting oximeter self-statistics. That needs to wait for functionality to update timeseries schema.
- Fixes #5658

### DIFF
--- a/nexus/db-model/src/schema_versions.rs
+++ b/nexus/db-model/src/schema_versions.rs
@@ -17,7 +17,7 @@ use std::collections::BTreeMap;
 ///
 /// This must be updated when you change the database schema.  Refer to
 /// schema/crdb/README.adoc in the root of this repository for details.
-pub const SCHEMA_VERSION: SemverVersion = SemverVersion::new(62, 0, 0);
+pub const SCHEMA_VERSION: SemverVersion = SemverVersion::new(63, 0, 0);
 
 /// List of all past database schema versions, in *reverse* order
 ///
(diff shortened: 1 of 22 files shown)


nexus/db-model/src/schema_versions.rs
```diff
@@ -29,6 +29,7 @@ static KNOWN_VERSIONS: Lazy<Vec<KnownVersion>> = Lazy::new(|| {
         // |  leaving the first copy as an example for the next person.
         // v
         // KnownVersion::new(next_int, "unique-dirname-with-the-sql-files"),
+        KnownVersion::new(63, "remove-producer-base-route-column"),
```
jgallagher, line 32: I don't see the migration files for this directory - missing in the commit maybe?

jgallagher (approved):

nexus/db-model/src/schema_versions.rs
bnaecker: `git add -u` strikes again!

bnaecker: Added in a54ce97ac

Issue #5552, opened 2024-04-17
Add garbage collection for expunged sleds and zones


sunshowers:
Specifically, the blueprint planner needs to:

- if a zone is expunged, and has no resources attached to it according to the planning input, then remove it from the blueprint
- if a sled is expunged, and has no zones attached to it, then mark it decommissioned

This should likely be done both at construction time against the parent blueprint, and incrementally.

Resources include:

- external IPs
- vNICs
- for Nexus instances, sagas
- ???

We've deferred this for release 8 but would like to do this for release 9.

sunshowers:
Some half-finished code which needs to be updated to take #5541 etc into account:

```rust
/// Builds a map of resources that either the planning input or the parent
/// blueprint says are in-use.
fn build_external_in_use_map<T: Clone + Hash + Eq>(
    parent_blueprint: &Blueprint,
    input_map: &BTreeMap<OmicronZoneUuid, T>,
    accessor: impl Fn(&OmicronZoneConfig) -> Option<T>,
) -> Result<HashMap<T, ResourceInUseDetails>, Error> {
    // The basic table is, for each resource:
    //
    // ```
    //    in blueprint?     in input_map?    result
    //  yes, reachable          yes          in use
    //  yes, non-reachable      yes          in use      [1]
    //         no               yes        in use, warn  [2]
    //  yes, reachable           no          in use      [3]
    //  yes, non-reachable       no         not in use
    //         no                no         not in use   [4]
    // ```
    //
    // "in blueprint?" means whether that resource is allocated to a zone
    // that's described in the blueprint (desired state). "reachable" is
    // defined by the `ShouldBeExternallyReachable` filter.
    //
    // "in input_map?" means whether that resource is allocated to a zone
    // that's described in the planning input (actual state).
    //
    // [1]: This is a zone that the executor will aim to remove external
    //      resources for. However, it is *currently* in use, and we can't just
    //      pave over and reallocate its resources.
    // [2]: This should never happen. The blueprint is meant to be a complete
    //      description of the state. This is a bug -- warn about it (and maybe
    //      consider erroring out?)
    // [3]: This is a zone that the executor will aim to add external resources
    //      for. We consider it to be in use.
    // [4]: This resource will never be considered in the below code.
    //
    // There are also some error cases:
    //
    // 1. Two zones in the input map are using the same resource.
    // 2. Two zones in the blueprint are using the same resource.
    // 3. For the same zone, the blueprint and the input map disagree on what
[...]

PR #5776, opened 2024-05-15
Blueprint builder: Guard against unknown external networking inputs

jgallagher:
This started as a small cleanup to make `builder.rs` a little less noisy (the first commit, that just moves all the external networking "what IPs are in use and what IP should I pick next" logic into a separate module). As I was doing that, though, I noticed we only look at the parent blueprint when deciding what IPs are used, and we completely ignore `PlanningInput` (which tells us what IPs were in use in the database "recently").

I _think_ we should never have an IP or NIC in the database that isn't also in the parent blueprint: blueprints are the source of how IPs/NICs are created, and we shouldn't prune zones from the blueprint until we've confirmed that external networking records for those zones have been deleted. I made this a planning error: if we get an input that reports external IPs or NICs that the parent blueprint doesn't know about, we bail out and fail to plan.

Assuming all of the above is correct (please correct me if any of it seems wrong!), it is still possible for this to fail, I think, if we had a sequence like this:

1. Nexus A generates a `PlanningInput` by reading from CRDB
2. Nexus B executes on a target blueprint that removes IPs/NICs from CRDB
3. Nexus B regenerates a new blueprint and prunes the zone(s) associated with the IPs/NICs from step 2
4. Nexus B makes this new blueprint the target
5. Nexus A attempts to generate a new blueprint with its `PlanningInput` from step 1

This isn't possible today for at least a couple reasons: we don't do prune zones as in step 3 at all, and the timing for steps 3-5 probably require automated blueprint regeneration. I think it could eventually be possible, but seems fine? Nexus A would fail to plan in step 5, but it would succeed the next time it tried (because it would generate a new `PlanningInput` that reflected the changes from step 2).


andrewjstone (approved):
You're commit message makes sense to me.

For an expunged zone, we'd first remove the network resources from the DB, then from the blueprint. The next blueprint could add a new zone reusing those resources as necessary. That resource wouldn't be in the DB because it was already removed. This is the converse of what you mention, but tracks with all that IIRC.

jgallagher:
Relaying some offline chat for posterity: @sunshowers pointed out that the change here exactly matches case 2 from [this comment](https://github.com/oxidecomputer/omicron/issues/5552#issuecomment-2062141829):

```
    // [2]: This should never happen. The blueprint is meant to be a complete
    //      description of the state. This is a bug -- warn about it (and maybe
    //      consider erroring out?)
```

and this PR did make the decision to error out on it.

Issue #5780, opened 2024-05-15
Make MGS available in nexus test suite


rcgoodfellow:
The `test_port_settings_basic_crud` test is currently disabled because it requires that nexus talk to MGS to figure out which switches it needs to talk to land port settings on a particular switch.

https://github.com/oxidecomputer/omicron/blob/e0a118496b05a30bf7a5a8a0820e0dda415c485a/nexus/tests/integration_tests/switch_port.rs#L28-L33

When this test was originally written, I believe we only had support for one switch and MGS was not involved in helping nexus make decisions. But now that we have multi-switch support, things are more complicated. This may ultimately not be necessary depending on how the following issues pan out e.g., if nexus determines what switches to talk to in ways that do not involve MGS.

- https://github.com/oxidecomputer/omicron/issues/5092
- https://github.com/oxidecomputer/omicron/issues/5201

sunshowers:
I think for this and other related issues, we should have a setup script that can ensure that the right version of mgd (etc) is present on disk, and if not then download it.

davepacheco:
Is it MGD or MGS that we need to be available in the test suite?  MGS has been available in the test suite for some time (since at least #5239).

rcgoodfellow:
This test was [pulled](https://github.com/oxidecomputer/omicron/pull/4390/files#r1380688472) in #4390, so we can likely retry this test now with the additional infrastructure from #5239 and #4291.

PR #3640, opened 2023-07-15
test-utils: Handle EOF mid-line when parsing ClickHouse log file

jgallagher:
I believe (:crossed_fingers:) this should fix the ubuntu CI hangs we've been seeing.

Closes #3580.

PR #5782, opened 2024-05-16
tests: fix mgd and dpd tcp port regex

rcgoodfellow:
The regular expressions for extracting `mgd` and `dpd` TCP ports from logs have issues that are fixed by this PR.

I considered short-circuiting the launching of `mgd` from the nexus test suite entirely, since it's not really being used due to #5780 (and that is why we did not catch this regex issue earlier). However, that would involve unwinding a fair amount of test infrastructure that we ultimately want to keep in place, we just need to come up with a solution for #5780. So I've left `mgd` in place for now and just fixed the regular expressions.

### DIFF
--- a/test-utils/src/dev/dendrite.rs
+++ b/test-utils/src/dev/dendrite.rs
@@ -139,7 +139,7 @@ async fn discover_port(logfile: String) -> Result<u16, anyhow::Error> {
 async fn find_dendrite_port_in_log(
     logfile: String,
 ) -> Result<u16, anyhow::Error> {
-    let re = regex::Regex::new(r#""local_addr":"\[::1\]:?([0-9]+)""#).unwrap();
+    let re = regex::Regex::new(r#""local_addr":"\[::1\]:([0-9]+)""#).unwrap();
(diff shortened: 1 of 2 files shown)


bnaecker:
Fixes for the regexes themselves look good to me. I think we should add a test or two for each of them though.

While we're here, we may also want to pull in a similar fix as that in https://github.com/oxidecomputer/omicron/pull/3640, specifically these lines:

https://github.com/oxidecomputer/omicron/blob/509bf0727e8bb98d88ed7346ce6469d70bd3c848/test-utils/src/dev/clickhouse.rs#L712-L715

That will ensure that, should we hit EOF partway through the needle the regex is searching for, we can still eventually recover it. That will catch an edge case that's pretty difficult to debug, because post-hoc the log files all look correct -- it's only when reading them as they're being written that the error appeared.

rcgoodfellow:
Thanks for the feedback @bnaecker. I've added a port discovery test for `mgd` based on the one for `dpd`. Also added the reader fix for both `dpd` and `mgd`.

bnaecker (approved):
Looks good, thanks for adding the test!

PR #5783, opened 2024-05-16
"external xtasks" pattern; wire up `cargo xtask releng`

iliana:
I finally found the set of Clap options to do what I wanted!

```
$ cargo xtask releng --help
    Finished dev [unoptimized + debuginfo] target(s) in 0.58s
     Running `target/debug/xtask releng --help`
    Finished release [optimized] target(s) in 1.02s
     Running `target/release/omicron-releng --help`
Run the Oxide release engineering process and produce a TUF repo that can be used to update a rack.

For more information, see `docs/releng.adoc` in the Omicron repository.

Note that `--host-dataset` and `--recovery-dataset` must be set to different values to build the two OS images in parallel. This is strongly recommended.

Usage: omicron-releng [OPTIONS]

Options:
      --host-dataset <HOST_DATASET>
          ZFS dataset to use for `helios-build` when building the host image

[... and so on]
```

### DIFF
--- a/dev-tools/releng/src/main.rs
+++ b/dev-tools/releng/src/main.rs
@@ -104,7 +104,7 @@ static WORKSPACE_DIR: Lazy<Utf8PathBuf> = Lazy::new(|| {
 /// Note that `--host-dataset` and `--recovery-dataset` must be set to different
 /// values to build the two OS images in parallel. This is strongly recommended.
 #[derive(Parser)]
-#[command(name = "cargo xtask releng")]
+#[command(name = "cargo xtask releng", bin_name = "cargo xtask releng")]
 struct Args {
     /// ZFS dataset to use for `helios-build` when building the host image
     #[clap(long, default_value_t = Self::default_dataset("host"))]
--- a/dev-tools/releng/src/main.rs
+++ b/dev-tools/releng/src/main.rs
@@ -96,7 +96,6 @@ static WORKSPACE_DIR: Lazy<Utf8PathBuf> = Lazy::new(|| {
     dir
 });
 
-#[derive(Parser)]
 /// Run the Oxide release engineering process and produce a TUF repo that can be
 /// used to update a rack.
 ///
@@ -104,6 +103,8 @@ static WORKSPACE_DIR: Lazy<Utf8PathBuf> = Lazy::new(|| {
 ///
 /// Note that `--host-dataset` and `--recovery-dataset` must be set to different
 /// values to build the two OS images in parallel. This is strongly recommended.
+#[derive(Parser)]
+#[command(name = "cargo xtask releng")]
 struct Args {
     /// ZFS dataset to use for `helios-build` when building the host image
     #[clap(long, default_value_t = Self::default_dataset("host"))]
--- a/dev-tools/xtask/src/external.rs
+++ b/dev-tools/xtask/src/external.rs
@@ -1,0 +1,68 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! External xtasks. (extasks?)
+
+use std::ffi::{OsStr, OsString};
+use std::os::unix::process::CommandExt;
+use std::process::Command;
+
+use anyhow::{Context, Result};
+use clap::Parser;
+
+/// Argument parser for external xtasks.
+///
+/// In general we want all developer tasks to be discoverable simply by running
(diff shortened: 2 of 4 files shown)


jclulow:
> Note that `--host-dataset` and `--recovery-dataset` must be set to different values to build the two OS images in parallel. This is strongly recommended.

Could you just have the tool create child datasets for each of these under a single parent?

iliana:
I wanted to avoid writing anything in this tool that would pfexec directly. Also the default values for each is `rpool/images/$LOGNAME/host` and `rpool/images/$LOGNAME/recovery` respectively, so by default the behavior is as-recommended.

dev-tools/releng/src/main.rs
```diff
@@ -96,14 +96,15 @@ static WORKSPACE_DIR: Lazy<Utf8PathBuf> = Lazy::new(|| {
...
-#[derive(Parser)]
 /// Run the Oxide release engineering process and produce a TUF repo that can be
 /// used to update a rack.
 ///
 /// For more information, see `docs/releng.adoc` in the Omicron repository.
 ///
 /// Note that `--host-dataset` and `--recovery-dataset` must be set to different
 /// values to build the two OS images in parallel. This is strongly recommended.
+#[derive(Parser)]
+#[command(name = "cargo xtask releng")]
```
smklein, line 107: Is this here to give better error messages when this command is invoked via xtask?

It's possible that the command is still `cargo run` explicitly, instead of being invoked by xtask, right?

dev-tools/xtask/src/external.rs
```diff
@@ -0,0 +1,68 @@
...
+    disable_help_flag(true),
+    disable_help_subcommand(true),
+    disable_version_flag(true)
+)]
+pub struct External {
+    #[clap(trailing_var_arg(true), allow_hyphen_values(true))]
+    args: Vec<OsString>,
+
+    #[clap(skip)]
+    command: Option<Command>,
```
smklein, line 38: If these args are being skipped, does it make sense to just omit this from the parsed struct?

dev-tools/xtask/src/external.rs
```diff
@@ -0,0 +1,68 @@
...
+}
+
+impl External {
+    pub fn cargo_args(
+        mut self,
+        args: impl IntoIterator<Item = impl AsRef<OsStr>>,
+    ) -> External {
+        self.command.get_or_insert_with(default_command).args(args);
```
smklein, line 46: Related to my comment above, this is calling `.get_or_insert_with`, but it's *always* using `default_command`, right?

dev-tools/xtask/src/external.rs
```diff
@@ -0,0 +1,68 @@
...
+        self
+    }
+
```
smklein, line 49: Rather than calling `cargo_args` and `exec`, we could just call:

```rust
pub struct External {
  args: Vec<OsString>,
}

impl External {
  pub fn exec(self, binary: impl AsRef<OsStr>) -> Result<()> {
    default_command()  // I might rename this to "cargo_run_command()"
      .arg("--bin")
      .arg(binary)
      .arg("--")
      .args(self.args)
      .exec();
    ...
  }
}
```

I think this has the same functionality, but makes it a little clear that the "default" command is not actually modifiable.

smklein (approved):

dev-tools/xtask/src/external.rs
sunshowers: Yeah, this is what I was wondering too -- any particular reason to have `command` be part of the struct rather than creating it here?

sunshowers (approved):

dev-tools/releng/src/main.rs
iliana: I thought this adjusted the usage string because this is the same thing that's in xtask, but it actually doesn't -- this appears to just be the name of the tool, which defaults to the crate name. So I'm going to try and find the correct thing and adjust that here and in xtask.

dev-tools/xtask/src/external.rs
iliana: What this is missing is the ability to add `--release` to the command, which is particularly useful for the releng tooling. (I don't expect that all external xtasks should be built and run in release mode.)

iliana: Specifically I imagine the usual flow in main.rs would be:

```rust
Cmds::Whatever(external) => external.exec("omicron-whatever"),
```

What the `Option<Command>`, `defualt_command`, etc chicanery gets us is the ability to optionally add additional arguments to cargo itself, which we use for `Cmds::Releng`.

iliana:
I'll add some comments to the `cargo_args` machinery to make it clearer what's going on.

PR #5784, opened 2024-05-16
[nexus] really deflake test_instance_watcher_metrics

### LINKED ISSUE #5752: test flake: omicron-nexus::test_all integration_tests::metrics::test_instance_watcher_metrics

hawkw:
The test `integration_tests::metrics::test_instance_watcher_metrics`
remains flaky even after adding an explicit call to
`Oximeter::force_collect` to ensure that metrics have been collected. I
believe this is due to the fact that, if the test runs long enough, the
`instance_watcher` background may be activated by its timer, causing
metrics to be collected another time, in addition to the test's explicit
activations. This can cause flaky failures when we then assert that
there is exactly a certain number of timeseries counted.

This branch changes the test to make assertions based on inequality,
instead. Now, we assert that the timeseries has *at least* the expected
count, so if the `instance_watcher` task has collected instance metrics
an additional time, we can tolerate that. We're still able to assert
that at least the expected counts are present. This is based on the
approach suggested by @bnaecker in [this comment][1].

I've re-run the test five times on my machine, and it appears to always
pass. Hopefully, this should actually fix #5752, but we probably
shouldn't close the issue until this has made it through a few CI
runs...

[1] https://github.com/oxidecomputer/omicron/pull/5768#discussion_r1602328362

### DIFF
--- a/nexus/tests/integration_tests/metrics.rs
+++ b/nexus/tests/integration_tests/metrics.rs
@@ -318,6 +318,28 @@ pub async fn timeseries_query(
 async fn test_instance_watcher_metrics(
     cptestctx: &ControlPlaneTestContext<omicron_nexus::Server>,
 ) {
+    macro_rules! assert_gte {
+        ($a:expr, $b:expr) => {{
+            let a = $a;
+            let b = $b;
+            assert!(
+                $a >= $b,
+                concat!(
+                    "assertion failed: ",
+                    stringify!($a),
+                    " >= ",
+                    stringify!($b),
+                    ", ",
+                    stringify!($a),
+                    " = {:?}, ",
+                    stringify!($b),
+                    " = {:?}",
+                ),
+                a,
+                b
+            );
+        }};
+    }
     use oximeter::types::FieldValue;
     const INSTANCE_ID_FIELD: &str = "instance_id";
     const STATE_FIELD: &str = "state";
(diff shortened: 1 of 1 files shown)


sunshowers:
(I'd like to add some kind of stress testing to nextest at some point, to make it easier to reproduce flakiness)

bnaecker (approved):
I think this all makes sense, thanks for dealing with this frustrating corner of the metrics system.

bnaecker:
> Looks good to me code-wise, but I'll let someone else accept it since I'm not totally sure it's the right thing to do policy-wise. (And do other tests need to be updated as well?)

It's a good question. I think most tests in the same module already work like this, asserting that there are _some_ samples after a timestamp, rather than there is an exact number of them. It seems like the right way to go right now, since there are many independent avenues for these metrics to get into the database, all of which can be sequenced in any order. It's not ideal, but I think it captures the intent of the test well -- making sure that the data we expect exists, even if there is also other data as well.

PR #5785, opened 2024-05-16
Add multiline diffs to blueprint display output

andrewjstone:


nexus/types/src/deployment/blueprint_display.rs
```diff
@@ -204,16 +242,54 @@ impl fmt::Display for BpSledSubtable {
...
+                        after,
+                        ..
+                    } = column
+                    {
+                        format!("{ADDED_PREFIX} {after}")
+                    } else {
+                        "".to_string()
+                    };
+
+                    if i != 0 {
```
jgallagher, line 282: Nit so trivial I have to apologize - can we make this `if i == 0` and flip the arms? Seems more natural for the first block to be executed on the first iteration.

nexus/types/src/deployment/blueprint_display.rs
```diff
@@ -204,16 +242,54 @@ impl fmt::Display for BpSledSubtable {
...
                 if i != 0 {
                     write!(f, "{:<COLUMN_GAP$}{column:<width$}", "")?;
                 } else {
                     write!(f, "{column:<width$}")?;
                 }
             }
             write!(f, "\n")?;
+
+            // Do we need any multiline output?
+            if multiline_row {
```
jgallagher, line 266: It's not obvious to me how this will scale to multiple multiline rows. We don't need that now so that seems fine, but I'm curious if you have an idea what that will look like?

jgallagher (approved):

nexus/types/src/deployment/blueprint_display.rs
andrewjstone: Honestly, i can only see the need for one more row, with a description of what happened. Anything else will be ridiculous clutter. I think we'll scale it by adding another boolean and loop to the rendering function, along with more detail in a column. It's done this way because of the way the rendering is done line by line. Another way to do it would be to compute the output into another intermediate format and then render that form. That really seems like overkill right now for what we likely won't need. I suggest re-evaluating when we have more details about what we need beyond a "reason or multiple reason" lines.

nexus/types/src/deployment/blueprint_display.rs
andrewjstone: Ha, it's not unreasonable. I actually put some rendering code in the wrong branch earlier. I have no idea why I wrote it this way in the first place. Will change.

andrewjstone: Fixed in [1ebc6c0](https://github.com/oxidecomputer/omicron/pull/5785/commits/1ebc6c0c3d0668f090c86096dc51a4dfa72de0a5)

PR #5794, opened 2024-05-20
Update propolis and crucible deps

bnaecker:


leftwo (approved):
This looks good. 
I assume openAPI tests did not complain about any changes?  I thought we had one
in extent count, but I guess the tests would complain if there was a problem.

bnaecker:
That change in the extent count from a `u64` to a `u32` was handled here:

https://github.com/oxidecomputer/omicron/blob/3965c15e8474dc7a377d0618e1edb43135cbdbe8/nexus/src/app/sagas/common_storage.rs#L40-L44

PR #5797, opened 2024-05-20
Reconfigurator: Add cockroachdb zones as needed

jgallagher:
This builds on #5788 to add support to the planner to add new CRDB zones to blueprints (if the current count is below the policy's target count). No changes were needed on the execution side, and the CRDB zones already bring themselves up as part of the existing cluster by looking up other node names in DNS.

A big chunk of the diff comes from expectorate output - our simulated test system was producing sleds that all had the same physical disk and zpool UUIDs, which messed with the test I wrote for the builder's zpool allocation. I tweaked the `TypedUuidRng` that the sled uses to include the `sled_id` (which itself comes from a "parent" `TypedUuidRng`) as the second seed argument. If that seems unreasonable, I am very open to other fixes!


nexus/reconfigurator/planning/src/blueprint_builder/builder.rs
```diff
@@ -614,15 +621,19 @@ impl<'a> BlueprintBuilder<'a> {
         Ok(Ensure::Added)
     }
 
-    /// Return the number of Nexus zones that would be configured to run on the
-    /// given sled if this builder generated a blueprint
+    /// Return the number of zones of a given kind that would be configured to
+    /// run on the given sled if this builder generated a blueprint
```
sunshowers, line 625:
```suggestion
    /// Return the number of zones of a given kind that would be configured to
    /// run on the given sled if this builder generated a blueprint.
```

nexus/reconfigurator/planning/src/planner.rs
```diff
@@ -328,70 +329,121 @@ impl<'a> Planner<'a> {
...
-        }
+        // We usually don't need to construct an `OmicronZonePlacement` to add
+        // discretionary zones, so defer its creation until it's needed.
+        let mut zone_placement = None;
+
+        for zone_kind in [
+            DiscretionaryOmicronZone::Nexus,
+            DiscretionaryOmicronZone::CockroachDb,
+        ] {
+            // Count the number of `kind` zones on all in-service sleds. This
```
sunshowers, line 347: What do you think about moving this block to another function? Might make it easier to test/call.

nexus/reconfigurator/planning/src/system.rs
```diff
@@ -446,7 +453,7 @@ impl Sled {
         let serial = format!("serial{}", unique);
         let revision = 0;
         let mut zpool_rng =
-            TypedUuidRng::from_seed("SystemSimultatedSled", "ZpoolUuid");
+            TypedUuidRng::from_seed("SystemSimultatedSled-ZpoolUuid", sled_id);
```
sunshowers, line 456: Ah good catch. Though what about:

```suggestion
            TypedUuidRng::from_seed("SystemSimultatedSled", (sled_id, "ZpoolUuid"));
```

sunshowers (approved):
Looks fantastic, thanks!

nexus/reconfigurator/planning/src/blueprint_builder/builder.rs
```diff
@@ -1081,6 +1178,33 @@ pub mod test {
                 );
             }
         }
+
+        // On any given zpool, we should have at most one zone of any given
```
andrewjstone, line 1182: Is this function actually used anywhere?

nexus/reconfigurator/planning/src/system.rs
```diff
@@ -121,6 +122,10 @@ impl SystemDescription {
 
         // Policy defaults
         let target_nexus_zone_count = NEXUS_REDUNDANCY;
+        // TODO-cleanup This is wrong, but we don't currently set up any CRDB
+        // nodes in our fake system, so this prevents downstream test issues
+        // with the planner thinking our system is out of date from the gate.
+        let target_cockroachdb_zone_count = 0;
```
andrewjstone, line 128: I was also wondering about this when looking at our tests the other day.  Should we actually deploy CRDB zones and everything else as part of our `ExampleSystem`? It feels like we should. Not in this PR though.

andrewjstone (approved):
Looks great!

nexus/reconfigurator/planning/src/planner.rs
jgallagher: I originally had it that way, but the `zone_placement` option that's reused across loop iterations was a little awkward. I've done some other rework since then; let me try it again...

nexus/reconfigurator/planning/src/blueprint_builder/builder.rs
jgallagher: Yeah, most/all of the tests in this file and `planner.rs` call this after generating blueprints to do basic sanity checking

nexus/reconfigurator/planning/src/system.rs
jgallagher: Agreed, and initially I set this to `COCKROACHDB_REDUNDANCY`. That broke basically all the nontrivial tests, though, so I took the easy way out and punted on this.

nexus/reconfigurator/planning/src/planner.rs
jgallagher: I tried extracting the full body of this loop into a separate function, and didn't like two things:

* The aforementioned `&mut Option<OmicronZonePlacement>` it has to take
* The function ended with an unconditional call to `add_discretionary_zones`, which felt like I was artificially splitting things up into two, which doesn't leave them independently testable

I took a second shot in `81b78ee64`: I kept the `Option<OmicronZonePlacement>` (and its construction, if needed) in the body of this loop, but moved the calculation of how many zones are needed into a new method. I think I like this better than either what was here before or moving the whole body of the loop into a method - what do you think?

nexus/reconfigurator/planning/src/blueprint_builder/builder.rs
```diff
@@ -791,6 +847,46 @@ impl<'a> BlueprintBuilder<'a> {
         allocator.alloc().ok_or(Error::OutOfAddresses { sled_id })
     }
 
+    fn sled_alloc_zpool(
```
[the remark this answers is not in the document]
jgallagher: Yeah, understandable. I'm not sure how we'd address this if we wanted to do something different here - maybe keep a separate zpool -> zone map in the builder? Might be worth it if we end up needing to do this more than just here, but seems annoying to keep in sync with zone changes being made.

nexus/reconfigurator/planning/src/system.rs
andrewjstone: Makes sense.

andrewjstone: I opened an issue for this: https://github.com/oxidecomputer/omicron/issues/5808

PR #5749, opened 2024-05-13
Perform instance state transitions in `instance-update` saga

hawkw:
A number of bugs relating to guest instance lifecycle management have been observed. These include:

- Instances getting "stuck" in a transient state, such as `Starting` or `Stopping`, with no way to forcibly terminate them (#4004)
-  Race conditions between instances starting and receiving state updates, which cause provisioning counters to underflow (#5042)
- Instances entering and exiting the `Failed` state when nothing is actually wrong with them, potentially leaking virtual resources (#4226)

These typically require support intervention to resolve.

Broadly , these issues exist because the control plane's current mechanisms for understanding and managing an instance's lifecycle state machine are "kind of a mess". In particular:

- **(Conceptual) ownership of the CRDB `instance` record is currently split between Nexus and sled-agent(s).** Although Nexus is the only entity that actually reads or writes to the database, the instance's runtime state is also modified by the sled-agents that manage its active Propolis (and, if it's migrating, it's target Propolis), and written to CRDB on their behalf by Nexus. This means that there are multiple copies of the instance's state in different places at the same time, which can potentially get out of sync. When an instance is migrating, its state is updated by two different sled-agents, and they may potentially generate state updates that conflict with each other. And, splitting the responsibility between Nexus and sled-agent makes the code more complex and harder to understand: there is no one place where all instance state machine transitions are performed.
- **Nexus doesn't ensure that instance state updates are processed reliably.** Instance state transitions triggered by user actions, such as `instance-start` and `instance-delete`, are performed by distributed sagas, ensuring that they run to completion even if the Nexus instance executing them comes to an untimely end. This is *not* the case for operations that result from instance state transitions reported by sled-agents, which just happen in the HTTP APIs for reporting instance states. If the Nexus processing such a transition crashes, gets network partition'd, or encountering a transient error, the instance is left in an incomplete state and the remainder of the operation will not be performed.

This branch rewrites much of the control plane's instance state management subsystem to resolve these issues. At a high level, it makes the following high-level changes:

-  **Nexus is now the sole owner of the `instance` record.** Sled-agents no longer have their own copies of an instance's `InstanceRuntimeState`, and do not generate changes to that state when reporting instance observations to Nexus. Instead, the sled-agent only publishes updates to the `vmm` and `migration` records (which are never modified by Nexus directly) and Nexus is the only entity responsible for determining how an instance's state should change in response to a VMM or migration state update.
[...]

PR #5812, opened 2024-05-23
[nexus] add background task for cleaning up abandoned VMMs

hawkw:
**Note**: This change is part of the ongoing work on instance lifecycle
management that I'm working on in PR #5749. It's not actually necessary
on its own, it's just a component of the upcoming instance updater saga.
However, I thought it would be easier to review if I factored out this
change into a separate PR that can be reviewed and merged on its own.

The instance update saga (see PR #5749) will only clean up after VMMs
whose IDs appear in an `instance` record. When a live migration finishes
(successfully or not), we want to allow a new migration to begin as soon
as possible, which means we have to unlink the “unused” side of the
migration --- the source if migration succeeded, or the target if it
failed --- from the instance, even though that VMM may not be fully
destroyed yet. Once this happens, the instance update saga will no
longer be able to clean up these VMMs, so we’ll need a separate task
that cleans up these "abandoned" VMMs in the background.

This branch introduces an `abandoned_vmm_reaper` background task that's
responsible for doing this. It queries the database to list VMMs which
are:

- in the `Destroyed` state
- not deleted yet (i.e. `time_deleted` IS NOT NULL)
- not pointed to by their corresponding instances (neither the
  `active_propolis_id` nor the `target_propolis_id` equals the VMM's ID)
  
For any VMMs returned by this query, the `abandoned_vmm_reaper` task
will:
- remove the `sled_resource` reservation for that VMM
- sets the `time_deleted` on the VMM record if it was not already set.

This cleanup process will be executed periodically in the background.
Eventually, the background task will also be explicitly triggered by the
instance update saga when it knows it has abandoned a VMM.

As an aside, I noticed that the current implementation of
`DataStore::vmm_mark_deleted` will always unconditionally set the
`time_deleted` field on a VMM record, even if it's already set. This is
"probably fine" for overall correctness: the VMM remains deleted, so the
operation is still idempotent-ish. But, it's not *great*, as it means
that any queries for VMMs deleted before a certain timestamp may not be
strictly correct, and we're updating the database more frequently than
we really need to. So, I've gone ahead and changed it to only set
`time_deleted` if the record's `time_deleted` is null, using
`check_if_exists` so that the method still returns `Ok` if the record
was already deleted --- the caller can inspect the returned `bool` to
determine whether or not they were the actual deleter, but the query
still doesn't fail.

### DIFF
--- a/nexus/db-queries/src/db/datastore/vmm.rs
+++ b/nexus/db-queries/src/db/datastore/vmm.rs
@@ -63,9 +63,15 @@ impl DataStore {
         let updated = diesel::update(dsl::vmm)
             .filter(dsl::id.eq(*vmm_id))
             .filter(dsl::state.eq_any(valid_states))
+            .filter(dsl::time_deleted.is_null())
             .set(dsl::time_deleted.eq(Utc::now()))
-            .execute_async(&*self.pool_connection_authorized(opctx).await?)
+            .check_if_exists::<Vmm>(*vmm_id)
+            .execute_and_check(&*self.pool_connection_authorized(opctx).await?)
             .await
+            .map(|r| match r.status {
+                UpdateStatus::Updated => true,
+                UpdateStatus::NotUpdatedButExists => false,
+            })
             .map_err(|e| {
                 public_error_from_diesel(
                     e,
@@ -76,7 +82,7 @@ impl DataStore {
                 )
             })?;
 
-        Ok(updated != 0)
+        Ok(updated)
     }
 
     pub async fn vmm_fetch(
--- a/nexus/db-queries/src/db/datastore/vmm.rs
+++ b/nexus/db-queries/src/db/datastore/vmm.rs
@@ -9,8 +9,10 @@ use crate::authz;
 use crate::context::OpContext;
 use crate::db::error::public_error_from_diesel;
 use crate::db::error::ErrorHandler;
+use crate::db::model::InstanceState as DbInstanceState;
 use crate::db::model::Vmm;
 use crate::db::model::VmmRuntimeState;
+use crate::db::pagination::paginated;
 use crate::db::schema::vmm::dsl;
 use crate::db::update_and_check::UpdateAndCheck;
 use crate::db::update_and_check::UpdateStatus;
@@ -18,7 +20,10 @@ use async_bb8_diesel::AsyncRunQueryDsl;
 use chrono::Utc;
 use diesel::prelude::*;
 use omicron_common::api::external::CreateResult;
+use omicron_common::api::external::DataPageParams;
 use omicron_common::api::external::Error;
+use omicron_common::api::external::InstanceState as ApiInstanceState;
+use omicron_common::api::external::ListResultVec;
 use omicron_common::api::external::LookupResult;
 use omicron_common::api::external::LookupType;
 use omicron_common::api::external::ResourceType;
@@ -50,9 +55,6 @@ impl DataStore {
         opctx: &OpContext,
         vmm_id: &Uuid,
     ) -> UpdateResult<bool> {
-        use crate::db::model::InstanceState as DbInstanceState;
-        use omicron_common::api::external::InstanceState as ApiInstanceState;
-
         let valid_states = vec![
             DbInstanceState::new(ApiInstanceState::Destroyed),
             DbInstanceState::new(ApiInstanceState::Failed),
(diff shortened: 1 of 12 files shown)


nexus/db-queries/src/db/datastore/vmm.rs
```diff
@@ -164,4 +172,52 @@ impl DataStore {
...
+        use crate::db::schema::instance::dsl as instance_dsl;
+        let destroyed = DbInstanceState::new(ApiInstanceState::Destroyed);
+        paginated(dsl::vmm, dsl::id, pagparams)
+            // In order to be considered "abandoned", a VMM must be:
+            // - in the `Destroyed` state
+            .filter(dsl::state.eq(destroyed))
+            // - not deleted yet
+            .filter(dsl::time_deleted.is_null())
+            // - not pointed to by their corresponding instances
+            .left_join(
```
smklein, line 202: (as mentioned in the hypervisor meeting) the fact that this is a `left_join`, not an `inner_join`, I *think* means we would also find any "VMMs that are 'destroyed', but not 'deleted', as long as instances aren't pointing at them".

I think this is actually a nice-to-have feature -- could (should?) this background task be responsible for destroying VMMs that are in a terminal state?

nexus/src/app/background/abandoned_vmm_reaper.rs
```diff
@@ -0,0 +1,450 @@
...
+        vmms: &[Vmm],
+    ) {
+        results.found += vmms.len();
+        slog::debug!(opctx.log, "Found abandoned VMMs"; "count" => vmms.len());
+
+        for vmm in vmms {
+            let vmm_id = vmm.id;
+            slog::trace!(opctx.log, "Deleting abandoned VMM"; "vmm" => %vmm_id);
+            // Attempt to remove the abandoned VMM's sled resource reservation.
+            match self.datastore.sled_reservation_delete(opctx, vmm_id).await {
```
smklein, line 97: The one spot of this PR that feels weird to me is the overlap with functionality in https://github.com/oxidecomputer/omicron/blob/c2f35151c7ac345305f396bc01f088083ccae7e0/nexus/src/app/instance.rs#L1970 

When a VMM is destroyed, that function seems to:

- Update counters in the project / fleet about "virtual provisioning" (see: `virtual_provisioning_collection_delete_instance`)
- Unassign producers from oximeter (see: `unassign_producer`)
- Update instance and VMM states to indicate that the VMM has been destroyed
- Delete the sled reservation (physical usage)
- Mark the VMM deleted

(We have previously identified that it is a problem these steps aren't atomic! But that's an issue which exists independently of this PR)

In contrast, this background task seems to:

- Delete the sled reservation (physical usage)
- Mark the VMM deleted

So, I have a smattering of follow-up questions here:

- Should this background task take over this responsibility for destroying VMMs from `notify_instance_updated` ? e.g., should that function start kicking this background task?
- That function also seems to destroy VMMs in the `Failed` state. Is that scope we would want this task to handle too?
- Should the "physical usage" tracking (sled reservation) and "virtual usage" tracking (`virtual_provisioning_collection...`) updates be more tightly coupled? I think they're both already atomic.

nexus/db-queries/src/db/datastore/vmm.rs
```diff
@@ -164,4 +172,52 @@ impl DataStore {
...
         Ok(vmm)
     }
+
+    /// Lists VMMs which have been abandoned by their instances after a
+    /// migration and are in need of cleanup.
+    ///
+    /// A VMM is considered "abandoned" if (and only if):
+    ///
+    /// - It is in the `Destroyed` state.
+    /// - It has previously been asigned to an instance.
```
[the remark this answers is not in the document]
hawkw: Yeah, that's correct, I'll update this comment to reflect that!

nexus/src/app/background/abandoned_vmm_reaper.rs
```diff
@@ -0,0 +1,450 @@
...
+                    slog::warn!(
+                        opctx.log,
+                        "Failed to delete sled reservation for abandoned VMM";
+                        "vmm" => %vmm_id,
+                        "error" => %e,
+                    );
+                    results.error_count += 1;
+                    *last_err = Err(e).with_context(|| {
+                        format!(
+                            "failed to delete sled reservation for {vmm_id}"
```
bnaecker, line 116: I might add something to the error string that indicates the UUID refers to the VMM, something like `"...reservation for VMM {vmm_id}"`. Unless that context is already obvious from the existing error `e`...

nexus/src/app/background/abandoned_vmm_reaper.rs
```diff
@@ -0,0 +1,450 @@
...
+                Err(e) => {
+                    slog::warn!(
+                        opctx.log,
+                        "Failed to mark abandoned VMM as deleted";
+                        "vmm" => %vmm_id,
+                        "error" => %e,
+                    );
+                    results.error_count += 1;
+                    *last_err = Err(e).with_context(|| {
+                        format!("failed to mark {vmm_id} as deleted")
```
bnaecker, line 149: Same comment here, might want to make clear what the UUID refers to.

bnaecker (approved):
Nice, clean work, looks good to me! Just one comment, but quite minor.

nexus/src/app/background/abandoned_vmm_reaper.rs
hawkw: yeah, good call, thanks!

nexus/db-queries/src/db/datastore/vmm.rs
hawkw: Yeah, we discussed this during the hypervisor sync --- I've updated the comments for this to more accurately reflect that this task will also clean up VMMs that are `Destroyed` and have *never* been assigned to an instance. The one case that this task is _not_ responsible for is for VMMs that are `Destroyed`/`Failed` while being used actively to run an instance, which is the responsibility of the (future) `instance-update` saga.

nexus/src/app/background/abandoned_vmm_reaper.rs
hawkw: I *think* our discussion in the hypervisor sync and on Matrix today has cleared this up --- for readers other than @smklein, though, the goal is that eventually:

- The behavior currently run in `notify_instance_updated` will eventually move to a saga (see #5749), and the Nexus upcall will just update the instance's state and kick off that saga
- When a VMM that's _actively in use by an instance_ is destroyed/failed, a bunch of cleanup needs to happen, including virtual provisioning, Oximeter producer, and networking; the update saga will do this
- When a VMM is abandoned after a migration (either because it was a failed migration target, or because it was the old VMM leftover after a successful migration), the cleanup is much simpler, as the _instance_ is still around and still using the virtual provisioning resources. Just the physical sled resources consumed by the individual Propolis process need to be released, the other resources are still in use by the instance, just not via *this* VMM.

This background task is only responsible for abandoned VMMs, so it only needs to do the simple cleanup process for them; it doesn't deallocate resources owned by instances when an in-use VMM fails or is stopped.

smklein (approved):

nexus/db-queries/src/db/datastore/vmm.rs
smklein: Sounds good - I think I was a bit narrowly focused on "what is the data", rather than "how did we get in this state" in my review, but your comments here make a lot of sense, especially in the context of the upcoming `instance-update` saga.

hawkw:
Okay, I've added some more comments clarifying which VMMs are candidates for cleanup by this task, and what cleanup is handled elsewhere (e.g. when an active VMM is destroyed). I'm going to merge this once it gets through CI --- thanks @smklein and @bnaecker for the reviews! <3

PR #5777, opened 2024-05-15
VPC Subnet Routing [1/2] -- RPW and System Routers

FelixMcFelix:
This PR wires up all the backing machinery for VPC subnet routing, and automatically resolves and pushes updated rules to sleds using an RPW. This allows instances in all subnets of a VPC to talk with one another -- assuming no firewall rules have been configured otherwise. At a high level, this works by a few changes:
* During the VPC create saga, we now push two rules explicitly to the system router -- default routes from `(0.0.0.0/0, ::/0) -> inetgw:outbound`.
* Any CRUD operation on a VPC subnet will reconcile the set of VPC subnet routes within the system router to have one entry per subnet. This takes the form `subnet:{name} -> subnet:{name}` for each subnet, which are later resolved to both v4 and v6 entries.
* Ports are created using route information known to sled-agent -- this defaults to an empty route set for instances/probes, and an internet gateway rule for services to enable early NTP sync.
* Routes are sync'd with sleds using a new background task. Broadly, this asks each sled for the set of VPCs and subnets it has ports on, and a version for the current route set installed in each. The background task will use this information to determine which routes must be rebuilt, and will send updated versions out in response.

The most immediate consequence in this PR is that hosts within a subnet -- on different VPCs -- will be able to talk with one another at last. The user facing API (#2116) will be re-enabled in a concurrent PR -- #5823 -- as will NIC spoof detection hole-punching.

Depends on oxidecomputer/opte#490.

Closes #2232, Fixes #1336.

---

A few pieces will block tests passing & merge-readiness:
- [x] Creation of a `lab-2.0-opte-0.32` image.
- [x] Merge of oxidecomputer/maghemite#274 (and updating all the right SHAs in this PR).

PR #5823, opened 2024-05-24
VPC Subnet Routing [2/2] -- Custom Routers and NIC 'transit IP' lists

### LINKED ISSUE #2116: Want Nexus API for managing VPC routing
We need to implement the customer-facing Nexus API for managing the virtual routing primitives defined in RFD 21:

- The VPC-wide system router and its contained routes
- Custom routers and routes that are defined per-VPC Subnet.

FelixMcFelix:
This PR builds on #5777 to provide the Custom routers for subnets as described in RFD21. This entails a few things:
* We remove the `unpublished = true` tag from the user API for VPC routers and routes.
* Custom routers may be attached/detached to a VPC subnet using the `custom_router` field in subnet `POST` and `PUT` requests.
* NICs now individually have a `transit_ips` list, which denotes an additional set of CIDR blocks that a NIC is allowed to send and receive traffic on. This is set during `POST` and/or `PUT` on instances which are stopped. This is a key feature to enable software routing by instances, as today's default behaviour drops any packets not matching an assigned IP for an instance.
   * I suspect there will be some discussion over the shape of this API, so there isn't yet test coverage here until we know we're happy with it.
* Revisited which router routes can be created by users, e.g., better validation on v4/v6 dest/target pairs.

There are some allowances around currently non-existent features:
* **Internet Gateways.** We allow unlimited use of one pseudo-gateway, `inetgw:outbound`, which appears in our existing rules. Using this target sends packets upstream as it does today.
* **VPC peering.** VPCs as destinations/targets are currently disallowed in router routes.

Closes #2116.


FelixMcFelix:
With the first go at this, I've tentatively named the anti-anti-spoof feature 'transit IPs' for want of a better name -- these can currently be set on NIC create/update. This is a lot less interactive than VPC routing in that it can only be configured while an instance is stopped. I'd like to fold this and other NIC-related parameters into another RPW in future so we can have that flexibility (and maybe also simplify, e.g., external IP updates), but I'm assuming that will be out of scope here.

Below I have a simple two instance setup -- `cafe` and `targeted`. Each is in a separate subnet of the same VPC, where `targeted`'s NIC permits transit of all v4/v6 traffic. Both share a router, with rules directing `6.6.6.6 -> targeted` and `7.7.7.7 -> cafe`.

<details>

<summary>Config details for 'cafe' and 'targeted'.</summary>

```bash
kyle@KyleOxide ~ % oxide instance list --project myproj
[
  {
    "description": "Test test test.",
    "hostname": "cafe",
    "id": "80e3b191-b4f8-4cf9-b41c-acf5277d1d35",
    "memory": 4294967296,
    "name": "cafe",
    "ncpus": 2,
    "project_id": "38f73344-edbc-4490-bae6-2ef60e9a38df",
    "run_state": "running",
    "time_created": "2024-05-29T11:25:38.052186Z",
    "time_modified": "2024-05-29T11:25:38.052186Z",
    "time_run_state_updated": "2024-05-29T11:25:52.142743Z"
  }, {
    "description": "",
    "hostname": "targeted",
    "id": "cf074e08-b198-4d11-969d-a255ef246239",
    "memory": 8589934592,
    "name": "targeted",
    "ncpus": 2,
    "project_id": "38f73344-edbc-4490-bae6-2ef60e9a38df",
    "run_state": "running",
    "time_created": "2024-05-29T11:30:11.007976Z",
    "time_modified": "2024-05-29T11:30:11.007976Z",
    "time_run_state_updated": "2024-05-29T11:35:01.667806Z"
  }
]
kyle@KyleOxide ~ % oxide instance nic list --instance 80e3b191-b4f8-4cf9-b41c-acf5277d1d35
[
  {
    "description": "default primary interface for cafe",
    "id": "54fa1bda-c474-4831-b9d5-d0a9dbdc77fa",
    "instance_id": "80e3b191-b4f8-4cf9-b41c-acf5277d1d35",
    "ip": "172.30.0.5",
    "mac": "A8:40:25:F5:AA:6E",
    "name": "net0",
    "primary": true,
    "subnet_id": "621f29b2-c8c8-4203-9ae3-69eec4b622b6",
    "time_created": "2024-05-29T11:25:38.193813Z",
    "time_modified": "2024-05-29T11:25:38.193813Z",
    "transit_ips": [],
    "vpc_id": "167f5077-4220-4098-91ec-0e99dcce253d"
  }
]
kyle@KyleOxide ~ % oxide instance nic list --instance cf074e08-b198-4d11-969d-a255ef246239
[
  {
    "description": "",
    "id": "89dc3edd-b015-43f5-8bb5-38f579d28787",
    "instance_id": "cf074e08-b198-4d11-969d-a255ef246239",
    "ip": "192.168.0.5",
    "mac": "A8:40:25:F5:62:18",
    "name": "niccy",
    "primary": true,
    "subnet_id": "30e2ddc6-5c29-4e43-b8a3-e69e734f14e5",
    "time_created": "2024-05-29T11:30:11.162952Z",
    "time_modified": "2024-05-29T11:34:29.072612Z",
    "transit_ips": [
      "0.0.0.0/0",
      "::/0"
    ],
    "vpc_id": "167f5077-4220-4098-91ec-0e99dcce253d"
  }
]
[
  {
    "custom_router_id": "275bca9f-d2be-4130-8cc2-cebc266963ca",
    "description": "",
    "id": "30e2ddc6-5c29-4e43-b8a3-e69e734f14e5",
    "ipv4_block": "192.168.0.0/24",
    "ipv6_block": "fd37:faf4:cc25:a950::/64",
    "name": "a-new-net",
    "time_created": "2024-05-29T11:27:17.600426Z",
    "time_modified": "2024-05-29T12:16:44.806520Z",
    "vpc_id": "167f5077-4220-4098-91ec-0e99dcce253d"
  }, {
    "custom_router_id": "275bca9f-d2be-4130-8cc2-cebc266963ca",
    "description": "The default subnet for default",
    "id": "621f29b2-c8c8-4203-9ae3-69eec4b622b6",
    "ipv4_block": "172.30.0.0/22",
    "ipv6_block": "fd37:faf4:cc25::/64",
    "name": "default",
    "time_created": "2024-05-29T11:25:37.460287Z",
    "time_modified": "2024-05-29T11:40:09.620637Z",
    "vpc_id": "167f5077-4220-4098-91ec-0e99dcce253d"
  }
]
oxide vpc route list --router 275bca9f-d2be-4130-8cc2-cebc266963ca
[
  {
    "description": "",
    "destination": {
      "type": "ip",
      "value": "7.7.7.7"
    },
    "id": "230747b7-b9fb-43b2-9b0b-68fefef04d8a",
    "kind": "custom",
    "name": "to-cafe",
    "target": {
      "type": "instance",
      "value": "cafe"
    },
    "time_created": "2024-05-29T12:01:57.323551Z",
    "time_modified": "2024-05-29T12:01:57.323551Z",
    "vpc_router_id": "275bca9f-d2be-4130-8cc2-cebc266963ca"
  }, {
    "description": "",
    "destination": {
      "type": "ip",
      "value": "6.6.6.6"
    },
    "id": "76879081-fb2a-4663-8594-32997816837b",
    "kind": "custom",
    "name": "to-targeted",
    "target": {
      "type": "instance",
      "value": "targeted"
    },
    "time_created": "2024-05-29T11:42:09.748739Z",
    "time_modified": "2024-05-29T11:42:09.748739Z",
    "vpc_router_id": "275bca9f-d2be-4130-8cc2-cebc266963ca"
  }
]
```

</details>

When `cafe` pings 6.6.6.6, `targeted` receives the traffic:

```bash
localhost:~# tcpdump -i eth0 dst 6.6.6.6
tcpdump: verbose output suppressed, use -v[v]... for full protocol decode
listening on eth0, link-type EN10MB (Ethernet), snapshot length 262144 bytes
12:10:19.492267 IP 172.30.0.5 > 6.6.6.6: ICMP echo request, id 2446, seq 0, length 64
12:10:20.492511 IP 172.30.0.5 > 6.6.6.6: ICMP echo request, id 2446, seq 1, length 64
^C
2 packets captured
2 packets received by filter
0 packets dropped by kernel
```


When `targeted` pings 7.7.7.7, `cafe` does not receive any traffic (as occurs today):

```bash
localhost:~# timeout 10 tcpdump -i eth0 dst 7.7.7.7
tcpdump: verbose output suppressed, use -v[v]... for full protocol decode
listening on eth0, link-type EN10MB (Ethernet), snapshot length 262144 bytes

0 packets captured
0 packets received by filter
0 packets dropped by kernel
localhost:~# 
```

FelixMcFelix:
As a quick sanity test, we can now use this to deliver packets to a tunnel endpoint as suggested in [RFD 21](https://rfd.shared.oxide.computer/rfd/21#example_vpc_custom_router). 

`subnet-tunnel-setup.sh` creates two instances (`node`@10.32.0.5, `gateway`@10.32.0.8) in a single VPC, where the latter is intended as a provide access to to 240.0.0.0/4. This is done by installing a custom router with one entry (`{dest: IpNet(240.0.0.0/4), target: Instance("gateway")}`), and adding the same IP block to the transit IP list for `gateway`'s primary NIC.

(It should be noted that this is using a modified `oxide.rs` client based on the new OpenAPI spec.)

<details>

<summary><code>subnet-tunnel-setup.sh</code></summary>

```bash
IP_POOL_START=10.1.222.11
IP_POOL_END=10.1.222.254

# Initial setup: set quotas and IP pool on external IP range.
oxide ip-pool create --name default --description "aha!"
oxide ip-pool range add --pool default --first $IP_POOL_START --last $IP_POOL_END
oxide ip-pool silo link --pool default --is-default true --silo recovery
oxide silo quotas update --silo recovery --cpus 8 --memory 17179869184 --storage 68719476736

# Create and start two instances, each in its own subnet in the default VPC.
oxide project create --name=myproj --description demo

oxide api /v1/images?project=myproj --method POST --input - <<EOF
{
  "name": "alpine",
  "description": "boot from propolis zone blob!",
  "os": "linux",
  "version": "1",
  "source": {
    "type": "you_can_boot_anything_as_long_as_its_alpine"
  }
}
EOF

# gateway testing.
oxide vpc create \
 --project myproj \
 --description "It has been upgraded!" \
 --dns-name tunnel-test \
 --name tunnel-test

oxide vpc subnet delete \
 --project myproj \
 --vpc tunnel-test \
 --subnet default

# Create router, install rule, bind to subnet.
oxide vpc router create \
 --project myproj \
 --vpc tunnel-test \
 --name phonebook \
 --description "How can we dial out over a tunnel?"

oxide api "/v1/vpc-router-routes?project=myproj&vpc=tunnel-test&router=phonebook" --method POST --input - <<EOF
{
  "name": "call-out-of-state",
  "description": "This is (probably) how.",
  "destination": {"type": "ip_net", "value": "240.0.0.0/4"}, 
  "target": {"type": "instance", "value": "gateway"}
}
EOF

oxide vpc subnet create \
 --project myproj \
 --vpc tunnel-test \
 --name call-centre \
 --description "Where we do gateway testing." \
 --ipv4-block "10.32.0.0/16" \
 --custom-router phonebook

for name in node gateway; do
  oxide instance from-image \
   --image alpine \
   --project myproj \
   --description "Test test test." \
   --hostname $name \
   --name $name \
   --memory 2g \
   --ncpus 2 \
   --size 2g

  oxide instance nic delete \
   --project myproj \
   --instance $name \
   --interface net0

  if [ "$name" = "gateway" ]; then
    addr="10.32.0.8"
  else
    addr="10.32.0.5"
  fi

  oxide instance nic create \
   --project myproj \
   --instance $name \
   --name $name-nic \
   --description "Another NIC!" \
   --vpc-name tunnel-test \
   --subnet-name call-centre \
   --ip $addr

  if [ "$name" = "gateway" ]; then
    oxide api "/v1/network-interfaces/$name-nic?project=myproj&instance=$name" --method PUT --input - <<EOF
{
  "primary": false,
  "transit_ips": ["240.0.0.0/4"]
}
EOF
  fi

  oxide instance start \
   --project myproj \
   --instance $name
done

```

</details>

If we now try to ping an address in this range from `node`:

```bash
localhost:~# ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP qlen 1000
    link/ether a8:40:25:fd:32:74 brd ff:ff:ff:ff:ff:ff
    inet 10.32.0.5/32 scope global eth0
       valid_lft forever preferred_lft forever
    inet6 fe80::aa40:25ff:fefd:3274/64 scope link 
       valid_lft forever preferred_lft forever
localhost:~# ping 240.1.0.1
PING 240.1.0.1 (240.1.0.1): 56 data bytes
^C
--- 240.1.0.1 ping statistics ---
6 packets transmitted, 0 packets received, 100% packet loss
```

It will be delivered to `gateway`.

```bash
localhost:~# ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP qlen 1000
    link/ether a8:40:25:f5:9e:ca brd ff:ff:ff:ff:ff:ff
    inet 10.32.0.8/32 scope global eth0
       valid_lft forever preferred_lft forever
    inet6 fe80::aa40:25ff:fef5:9eca/64 scope link 
       valid_lft forever preferred_lft forever
localhost:~# tcpdump -i eth0 dst 240.1.0.1
tcpdump: verbose output suppressed, use -v[v]... for full protocol decode
listening on eth0, link-type EN10MB (Ethernet), snapshot length 262144 bytes
21:54:42.170120 IP 10.32.0.5 > 240.1.0.1: ICMP echo request, id 2451, seq 0, length 64
21:54:43.170465 IP 10.32.0.5 > 240.1.0.1: ICMP echo request, id 2451, seq 1, length 64
21:54:44.170537 IP 10.32.0.5 > 240.1.0.1: ICMP echo request, id 2451, seq 2, length 64
21:54:45.170689 IP 10.32.0.5 > 240.1.0.1: ICMP echo request, id 2451, seq 3, length 64
21:54:46.170979 IP 10.32.0.5 > 240.1.0.1: ICMP echo request, id 2451, seq 4, length 64
^C
5 packets captured
5 packets received by filter
0 packets dropped by kernel
```

We also see that `gateway` can now originate traffic on this subnet, and replies will be routed to it. Source addresses outside this block will *not* be permitted.

```bash
localhost:~# ip a add 240.2.0.2 dev eth0
localhost:~# ping -I 240.2.0.2 10.32.0.5
PING 10.32.0.5 (10.32.0.5) from 240.2.0.2: 56 data bytes
64 bytes from 10.32.0.5: seq=0 ttl=64 time=0.388 ms
64 bytes from 10.32.0.5: seq=1 ttl=64 time=0.273 ms
64 bytes from 10.32.0.5: seq=2 ttl=64 time=0.469 ms
64 bytes from 10.32.0.5: seq=3 ttl=64 time=0.499 ms
64 bytes from 10.32.0.5: seq=4 ttl=64 time=0.523 ms
64 bytes from 10.32.0.5: seq=5 ttl=64 time=0.522 ms
64 bytes from 10.32.0.5: seq=6 ttl=64 time=0.490 ms
64 bytes from 10.32.0.5: seq=7 ttl=64 time=0.323 ms
^C
--- 10.32.0.5 ping statistics ---
8 packets transmitted, 8 packets received, 0% packet loss
round-trip min/avg/max = 0.273/0.435/0.523 ms

localhost:~# ip a add 10.32.0.9 dev eth0
localhost:~# ping -I 10.32.0.9 10.32.0.5
PING 10.32.0.5 (10.32.0.5) from 10.32.0.9: 56 data bytes
^C
--- 10.32.0.5 ping statistics ---
5 packets transmitted, 0 packets received, 100% packet loss
```

common/src/api/internal/shared.rs
```diff
@@ -57,6 +57,7 @@ pub struct NetworkInterface {
     pub vni: external::Vni,
     pub primary: bool,
     pub slot: u8,
+    pub transit_ips: Option<Vec<external::IpNet>>,
```
FelixMcFelix, line 60: This is `Option`al instead of just a `Vec` because this type seems to be used in stored blueprints etc., and this seemed the easiest way not to break deserialisation. This field should only be used on instance NICs.

I don't know whether we'd prefer a `#[serde(default)]` here, or something else entirely since service NICs, zone config etc. should never use this property.

common/src/api/external/mod.rs
```diff
@@ -1979,6 +1979,10 @@ pub struct InstanceNetworkInterface {
     /// True if this interface is the primary for the instance to which it's
     /// attached.
     pub primary: bool,
+
+    /// A set of additional networks that this interface may send and
+    /// receive traffic on.
+    pub transit_ips: Vec<IpNet>,
```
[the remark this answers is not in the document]
FelixMcFelix: I think it couldn't hurt to have it, especially considering that this is not a property that most instance NICs will need to care about. Only tunnel endpoints and/or instances acting as designated software routers should use this.

Done in 992b9b515.

nexus/types/src/external_api/params.rs
```diff
@@ -1252,7 +1269,9 @@ pub struct VpcRouterUpdate {
 pub struct RouterRouteCreate {
     #[serde(flatten)]
     pub identity: IdentityMetadataCreateParams,
+    /// The location that matched packets should be forwarded to.
     pub target: RouteTarget,
+    /// A CIDR block (or named subnet) which this route will apply to.
```
david-crespo, line 1274: Is this accurate? I see a destination can be an IP, IP net, VPC, or subnet.

nexus/types/src/external_api/params.rs
```diff
@@ -1220,13 +1225,25 @@ pub struct VpcSubnetCreate {
...
     /// existing subnet in the VPC.
     pub ipv6_block: Option<Ipv6Net>,
+
+    /// An optional router, used to direct packets sent from hosts in this subnet
+    /// to any destination address.
+    ///
+    /// Custom routers apply in addition to the VPC-wide *system* router, and have
+    /// higher priority than the system router for an otherwise
+    /// equal-prefix-length match.
+    pub custom_router: Option<NameOrId>,
```
david-crespo, line 1235: I was wondering if this could just be called `router` and then on the `VpcSubnet` view we can call it `router_id`, and we always include the ID. But I see now, "Custom routers apply in addition to the VPC-wide *system* router" means that doesn't make sense, right? Because when there is a custom router, there is still also the system router. It's not a default vs. custom situation. So I think this is good.

david-crespo (approved):
The API side looks good to me. `transit_ips` vec seems fine. We can always tweak it after we see how it feels in practice.

nexus/types/src/external_api/params.rs
FelixMcFelix: Thanks for looking over this!

It can be any of those, except for VPCs -- that will likely relax in future with VPC peering, but it's not something that makes sense right now. I'll try and rework the wording for this, I maybe want to make it clearer that 'destination' matches on the address in routed packets.

david-crespo: Ok, sounds good. I was asking because I saw more stuff on the `RouteDestination` type. But it's fine if there are things that are allowed by the type that are nonetheless rejected by the endpoint logic as invalid, and I see you're doing that nicely [here](https://github.com/oxidecomputer/omicron/pull/5823/files/d747d01467d4be7797a562f0956489d0541c3f72#diff-0017315ba1221fe0d101b6602d359187f4b40e5c3a79b8aa2be94bd85758b3fcR224-R234).

common/src/api/external/mod.rs
```diff
@@ -1407,14 +1407,14 @@ pub struct RouterRoute {
...
     /// The ID of the VPC Router to which the route belongs
     pub vpc_router_id: Uuid,
-
     /// Describes the kind of router. Set at creation. `read-only`
     pub kind: RouterRouteKind,
-
+    /// The location that matched packets should be forwarded to.
     pub target: RouteTarget,
+    /// The set of destination IP addresses or subnets that this route
+    /// will match packets against.
```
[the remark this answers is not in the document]
FelixMcFelix: Yeah, the existing `RouteDestination` text is good there. I think if we lean into just showing its role and delegating the details to the type, I'd go for:
```rust
/// Selects which traffic this routing rule will apply to.
pub destination: RouteDestination,
```

FelixMcFelix:
I've mupdated london from ff0c914 to the most recent commit on this PR, without any ledger-related issues stemming from `NetworkInterface`. Nexus has come back up (after a schema update), so I think we're covered on the sled-agent changes and good to go right after #5777.

PR #5827, opened 2024-05-28
fix `clippy::redundant_closure_call` warning

hawkw:
The `diff_row!` macro in `blueprint_diff.rs` has recently started
generating [`clippy::redundant_closure_call`][1] warnings, as
invocations without a `display` function pass in a closure that performs
an identity operation (e.g. just returns the value). Rather than
allowing the warning, which would also have been fine, I've changed this
arm of the macro to pass the named function `std::convert::identity`
instead, which is equivalent but eliminates the closure.

[1]: https://rust-lang.github.io/rust-clippy/master/index.html#redundant_closure_call

### DIFF
--- a/nexus/types/src/deployment/blueprint_diff.rs
+++ b/nexus/types/src/deployment/blueprint_diff.rs
@@ -667,7 +667,7 @@ impl<'diff> BlueprintDiffDisplay<'diff> {
     ) -> impl IntoIterator<Item = KvListWithHeading> {
         macro_rules! diff_row {
             ($member:ident, $label:expr) => {
-                diff_row!($member, $label, |value| value)
+                diff_row!($member, $label, std::convert::identity)
             };
 
             ($member:ident, $label:expr, $display:expr) => {


david-crespo (approved):

iliana:
Hm, I had that initially but something about the ownership was wrong. Although I may have papered over that with the double-referencing bullshit I did elsewhere in that macro:

https://github.com/oxidecomputer/omicron/blob/68ac8762f1c0539aa36071d3df13cf1ad5358234/nexus/types/src/deployment/blueprint_diff.rs#L682

It is strange that this wasn't seen in the CI job, although I have noticed a number of other clippy lints that my editor is seeing but that CI isn't catching.

hawkw:
> Hm, I had that initially but something about the ownership was wrong. Although I may have papered over that with the double-referencing bullshit I did elsewhere in that macro:
> 
> https://github.com/oxidecomputer/omicron/blob/68ac8762f1c0539aa36071d3df13cf1ad5358234/nexus/types/src/deployment/blueprint_diff.rs#L682

It certainly compiles with `convert::identity` _now_, at least for me, but that may not have always been the case.

> It is strange that this wasn't seen in the CI job, although I have noticed a number of other clippy lints that my editor is seeing but that CI isn't catching.

Yeah, surprisingly, I haven't seen it on CI either...I'm definitely seeing it locally though.

hawkw:
[This timeout](https://buildomat.eng.oxide.computer/wg/0/details/01HZ0CAZGE210DGK2G8VAZZY53/UeLNcjW86HYxzsEIGMhrT2MYtVsX1YmmZCBhIVTvWsFhDEdx/01HZ0CBEXJEYPHS951NWP54M0S#S4128) really _looks_ like a test flake...I'm restarting it.

PR #5829, opened 2024-05-28
add `oxlog services` to print filterable service names

faithanalog:
`oxlog logs` can filter to logs from a specific service. This is convenient, but I wasn't sure what the name of the service was that I wanted to find.

I read the oxlog logic and came up with this command:

```
$ ./oxlog zones | xargs -n1 ./oxlog logs --current --archived | awk -F/ '{ print $NF }' | awk -F: '{ svc=$1; gsub(/^[^-]+-/, "", svc); print svc }' | sort -u
chrony-setup
cockroachdb
crucible-agent
crucible-downstairs
crucible-pantry
dendrite
illumos-propolis-server    <- note the error in my prefix-stripping, oops! should be propolis-server
internal_dns
lldpd
mg-ddm
mgd
mgs
nexus
ntp
opte-interface-setup
pumpkind
sled-agent
switch_zone_setup
tfport
uplink
wicketd
zone-network-setup
```

But that's a bit cumbersome I think, so this PR adds `oxlog services`, which generates the same list based on oxlog's own understanding of the service names:

```
$ ./oxlog services
chrony-setup
cockroachdb
crucible-agent
crucible-downstairs
crucible-pantry
dendrite
internal_dns
lldpd
mg-ddm
mgd
mgs
nexus
ntp
opte-interface-setup
propolis-server
pumpkind
sled-agent
switch_zone_setup
tfport
uplink
wicketd
zone-network-setup
```

### DIFF
--- a/dev-tools/oxlog/src/bin/oxlog.rs
+++ b/dev-tools/oxlog/src/bin/oxlog.rs
@@ -22,7 +22,7 @@ enum Commands {
 
     /// List logs for a given service
     Logs {
-        // The name of the zone
+        /// The name of the zone
         zone: String,
 
         /// The name of the service to list logs for
@@ -36,10 +36,13 @@ enum Commands {
         filter: FilterArgs,
     },
 
-    /// List the names of all services on the system, from the perspective of
-    /// oxlog. Use these names with `oxlog logs` to filter output to logs from a
+    /// List the names of all services in a zone, from the perspective of oxlog.
+    /// Use these names with `oxlog logs` to filter output to logs from a
     /// specific service.
-    Services,
+    Services {
+        /// The name of the zone
+        zone: String,
+    },
 }
 
 #[derive(Args, Debug)]
@@ -130,7 +133,7 @@ fn main() -> Result<(), anyhow::Error> {
             }
             Ok(())
         }
-        Commands::Services => {
+        Commands::Services { zone } => {
             let zones = Zones::load()?;
 
             // We want all logs that exist, anywhere, so we can find their
@@ -142,13 +145,10 @@ fn main() -> Result<(), anyhow::Error> {
                 show_empty: true,
             };
 
-            // Collect a unique set of services, based on the logs in all zones
-            let services: BTreeSet<String> = zones
-                .zones
-                .keys()
-                .flat_map(|zone| zones.zone_logs(&zone, filter).into_iter())
-                .map(|(name, _)| name)
-                .collect();
+            // Collect a unique set of services, based on the logs in the
+            // specified zone
+            let services: BTreeSet<String> =
+                zones.zone_logs(&zone, filter).into_keys().collect();
 
             for svc in services {
                 println!("{}", svc);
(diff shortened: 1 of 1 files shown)


dev-tools/oxlog/src/bin/oxlog.rs
```diff
@@ -124,5 +130,31 @@ fn main() -> Result<(), anyhow::Error> {
...
+            // We want all logs that exist, anywhere, so we can find their
+            // service names.
+            let filter = Filter {
+                current: true,
+                archived: true,
+                extra: true,
+                show_empty: true,
+            };
+
+            // Collect a unique set of services, based on the logs in all zones
```
andrewjstone, line 145: This could potentially be really slow in production. In the original incarnation of `oxlog`, I didn't require passing a zone at all and would search across all of them. Unfortunately, each zone is on a separate disk, and some have lots of files. This resulted in really slow behavior, on the order of seconds to 10s of seconds. I could also see this potentially interfering in production workloads, and because of that I forced a specific zone to be passed when looking for logs. In practice this isn't a huge deal, as you typically know where you want to look I believe.

I'd prefer we stick to that pattern here as well and force the user to pass a specific zone, and then look for the file names in that zone.

While I sympathize with wanting to see everything, I'm worried about the cost of accidentally doing this repeatedly on a production system. I think over time, we can possibly make some of this logic async and try to run it concurrently. While making the command faster, that may actually make the load worse though. It's possible I'm being too conservative here, but a command that takes 10 seconds to complete is also no fun to work with.

faithanalog: makes sense. I'll switch it taking a zone argument then. `oxlog zones | xargs -n1 oxlog services` would then do the trick for looking at all zones, if I really wanted to

andrewjstone: Thanks @faithanalog. Conditionally approved on that change. 

Thanks for making this addition. It's super helpful!

andrewjstone (approved):

faithanalog:
Changed to take a zone arg. Also added a missing `/` on one of the arg comments. looks like it's working to me, going to hit the auto-merge

```
BRM44220007 # ./oxlog services oxz_crucible_0585e502-147e-41f8-9129-38d92c5b95df
crucible-agent
crucible-downstairs
zone-network-setup
BRM44220007 # ./oxlog services oxz_nexus_b7332b1d-155e-441f-bf59-3321a154d463
nexus
opte-interface-setup
zone-network-setup
BRM44220007 # ./oxlog services oxz_ntp_86e921b5-56a1-42c7-ad02-3829f94a010f
chrony-setup
ntp
opte-interface-setup
zone-network-setup
BRM44220007 # ./oxlog zones | xargs -n1 ./oxlog services | sort -u
chrony-setup
cockroachdb
crucible-agent
crucible-downstairs
crucible-pantry
mg-ddm
nexus
ntp
opte-interface-setup
propolis-server
pumpkind
sled-agent
zone-network-setup
```

Issue #5525, opened 2024-04-12
Duplicate resource accounting update when multiple stop/deletion requests are made on the same instance  


askfongjojo:
We have a project on rack2 that shows negative compute usage which manifested as an error during disk deletion:
```
{
  "msg": "request completed",
  "v": 0,
  "name": "65a11c18-7f59-41ac-b9e7-680627f996e7",
  "level": 30,
  "time": "2024-04-12T14:28:00.793206966Z",
  "hostname": "oxz_nexus_65a11c18-7f59-41ac-b9e7-680627f996e7",
  "pid": 10277,
  "uri": "//v1/disks/hubris-disk-lpc55xpresso-279c894b-ed57-4d14-8100-5928cf03924c?project=hubris-build",
  "method": "DELETE",
  "req_id": "aa0543d3-0c79-4faa-9b86-8ada4e0ca201",
  "remote_addr": "172.20.16.182:47642",
  "local_addr": "172.30.2.5:443",
  "component": "dropshot_external",
  "file": "/home/build/.cargo/git/checkouts/dropshot-a4a923d29dccc492/29ae98d/dropshot/src/server.rs:837",
  "error_message_external": "Internal Server Error",
  "error_message_internal": "saga ACTION error at node \"no_result1\": unexpected database error: value is too small for a byte count",
  "latency_us": 133503,
  "response_code": "500"
}
```
A database query again the `virtual_resource_collection` table for the related project showed that it had negative compute usage:
```
  a494b278-d167-4940-9ffb-6d900ea0e47d | 2024-04-12 14:28:00.75635+00  | Project         |                              0 |               -4 |     -8589934592
```
Here are the vmm/instances that were in the project previously:
```
root@[fd00:1122:3344:105::3]:32221/omicron> select id, name, ncpus, memory, state, time_created, time_deleted from instance where project_id = 'a494b278-d167-4940-9ffb-6d900ea0e47d' order by time_created;
                   id                  |                              name                               | ncpus |   memory   |   state   |         time_created          |         time_deleted
---------------------------------------+-----------------------------------------------------------------+-------+------------+-----------+-------------------------------+--------------------------------
  c306cd25-bc22-4b40-be8a-d1f7ed3850fd | hubris-build-oxide-rot-1-ad95a39e-de73-43f6-ad83-053fb8b91fad   |     4 | 8589934592 | destroyed | 2024-04-04 19:54:39.735178+00 | 2024-04-04 20:00:43.863439+00
[...]

Issue #5042, opened 2024-02-10
Race between start saga, start saga unwind, and sled agent instance teardown can cause provisioning counter underflow


gjcolombo:
## Summary

There appears to be an extremely subtle race between the instance start saga, its unwind path, and the instance cleanup logic in sled agent that can cause an instance's virtual provisioning counters to underflow. It goes something like this:

1. Instance gets created; the instance record has generation 2
2. Task T1 begins start saga A1
3. A1 sets the VMM ID V1 in the instance, advancing its record to generation 3
4. A1 registers the V1 with sled S1, telling the sled the instance is at generation 3
5. A1 allocates virtual provisioning counters for the now-starting instance
6. A1 calls S1 to ask it to set the instance to running
7. In S1's sled agent, T2 takes the instance's lock and begins setting up a Propolis process (`setup_propolis_locked`) and telling it to create a VM (`ensure_propolis_and_tasks`)
8. A1 times out waiting for T2 to finish its work; start saga unwind begins
9. A1 releases virtual provisioning counters as part of unwind
10. A1 issues an `ensure_unregistered` call to S1
11. S1 spawns T3 to handle this call; this blocks waiting for the instance lock (which T2 owns)
12. T2 spawns instance monitoring task T4; this calls into Propolis, gets an initial state from it, returns, and then blocks on the instance lock (which T2 still owns)
13. T2 drops the lock; T3 acquires it, then unregisters the instance; this terminates the Propolis zone and marks the relevant VMM as `Destroyed`, then returns to Nexus
14. T1 finishes unwinding saga S1: this clears the VMM ID and moves the instance generation to 4, then returns
15. A new start saga begins for the same instance; call it A2. A2 registers a new VMM for the instance, advances the instance generation to 5, re-registers the instance (note that its previous monitoring task T4 is not gone yet!), 
16. A2 registers a VMM, advancing the instance to generation 5; it then tells sled S2 to register the instance and charges the virtual provisioning collection
17. Back over on S1, T4 gets the lock and calls sled agent's `InstanceInner::observe_state` to apply its state observation from step 12. This is a no-op because the instance's zone was destroyed in step 13. However, the monitoring task still notifies Nexus of the state change!
18. T4 publishes to Nexus an update that says that the instance's VMM is destroyed; this has an instance generation of 3, because that's what was registered with S1 all the way back in step 4
[...]

PR #5830, opened 2024-05-28
[nexus] Add virtual provisioning idempotency tests, prevent underflow

smklein:
Builds on https://github.com/oxidecomputer/omicron/pull/5081 and https://github.com/oxidecomputer/omicron/pull/5089 , but more out of convenience than necessity.

# Summary

This PR attempts to validate that, for the "virtual provisioning collection {insert, delete}" operations, they are idempotent. Currently, our usage of `max_instance_gen` only **partially** prevents updates during instance provisioning deletions:
- If `max_instance_gen` is smaller than the observed instance generation number...
- ... we avoid deleting the `virtual_provisioning_resource` record (which is great)
- ... but we still decrement the `virtual_provisioning_collection` values (which is really not great).

This basically means that we can "only cause the project/silo/fleet usage values to decrement arbitrarily, with no other changes". This has been, mechanically, the root cause of our observed underflows (e.g, https://github.com/oxidecomputer/omicron/issues/5525). 

# Details of this change

- All the changes in `nexus/db-queries/src/db/datastore/virtual_provisioning_collection.rs` are tests validating idempotency of these operations.
- All the changes in `nexus/db-queries/src/db/queries/virtual_provisioning_collection_update.rs` are actually changes to the query which change functionality. The objective of these changes is to preserve idempotency of the newly added tests, and to prevent undercounting of virtual provisioning resources. If these changes are reverted, the newly added tests start failing, showing a lack of coverage.


nexus/db-queries/src/db/queries/virtual_provisioning_collection_update.rs
```diff
@@ -239,11 +239,52 @@ WITH
...
+                // resources if the instance generation was already advanced to or
+                // past G.
+                //
+                // If the relevant instance ID is not in the database, then some
+                // other operation must have ensured the instance was previously
+                // stopped (because that's the only way it could have been deleted),
+                // and that operation should have cleaned up the resources already,
+                // in which case there's nothing to do here.
+                query.sql("
+  do_update
```
smklein, line 262: I basically tried to move all the validation of the state_generation into this boolean `do_update?` subquery, then re-use that check in multiple spots where we can perform updates below.

nexus/db-queries/tests/output/virtual_provisioning_collection_update_delete_instance.sql
```diff
@@ -40,25 +40,17 @@ WITH
             1
         )
         = 1
+        AND EXISTS(
+            SELECT 1 FROM instance WHERE instance.id = $5 AND instance.state_generation < $6 LIMIT 1
+          )
```
smklein, line 45: The problem here is that (before this PR) `do_update` could be "true", but the state generation could be below our "max", which would result in that buggy partial deletion state.

Now:
- If `do_update` is true, we really believe we should perform the insertion/deletion
- If that's the case, we can use it as the deciding factor to both delete the `virtual_provisioning_resource` **and** decrement the usage in the `virtual_provisioning_collection` table

gjcolombo (approved):
Thanks for doing this! LGTM as far as the use of the `do_update` clause goes--this is more or less the fix I had in mind for #5042, which issue I think this will resolve.

I know the conversion to raw SQL is part of a different PR, but I also looked at the entire expected query output in `virtual_provisioning_collection_update_delete_instance.sql` and the structure there also looks right to me (I didn't check the parameter ordering too closely though).

nexus/db-queries/src/db/datastore/virtual_provisioning_collection.rs
```diff
@@ -325,3 +325,614 @@ impl DataStore {
...
+        datastore
+            .silo_update_quota(&opctx, &authz_silo, quotas_update)
+            .await
+            .unwrap();
+
+        // Actually provision the instance
+
+        let instance_id = Uuid::new_v4();
+        let cpus = 12;
+        let ram = ByteCount::try_from(1 << 30).unwrap();
```
hawkw, line 464: nit, take it or leave it: it looks like creating the instance record is factored out into a function, but the rest of this test setup code is repeated for all the tests. why not just have one `test_setup` (or something) that does all of this?

hawkw (approved):

nexus/db-queries/src/db/datastore/virtual_provisioning_collection.rs
smklein: Good feedback, I'll incorporate this. I was being lazy!

PR #5831, opened 2024-05-29
[nexus] add instance-updater lock

hawkw:
In order to simplify the instance lifecycle state machine and ensure that instance state updates are processed reliably, we intend to perform instance state updates in a saga (which will be added in PR #5749). This saga will require a notion of mutual exclusion between update sagas for the same instance, in order to avoid race conditions like the following:

1. Sagas `S1` and `S2` start at the same time and observe the same instance/VMM states, which indicate that the instance’s active VMM has shut down
2. `S1` clears all the resources/provisioning counters and marks the instance as `Stopped``
3. User restarts the instance
4. `S2` clears the same instance provisioning counters again

Presently, these races are avoided by the fact that instance state updates are performed partially in `sled-agent`, which serves as an "actor" with exclusive ownership over the state transition. Moving these state transitions to Nexus requires introducing mutual exclusion.

This commit adds a distributed lock on instance state transitions to the datastore. We add the following fields to the `instance` table:

- `updater_id`, which is the UUID of the saga currently holding the update lock on the instance (or `NULL` if no saga has locked the instance)
- `updater_gen`, a generation counter that is incremented each time the lock is acquired by a new saga

Using these fields, we can add new datastore methods to try and acquire an instance update lock by doing the following:

1. Generate a UUID for the saga, `saga_lock_id`. This will be performed in the saga itself and isn't part of this PR.
2. Read the instance record and interpret the value of the `updater_id` field as follows:
  - `NULL`: lock not held, we can acquire it by incrementing the `updater_gen` field and setting the `updater_id` field to the saga's UUID.
  - `updater_id == saga_id`: the saga already holds the lock, we can proceed with the update.
  - `updater_id != saga_id`: another saga holds the lock, we can't proceed with the update. Fail the operation.
3. Attempt to write back the updated instance record with generation incremented and the `updater_id` set to the saga's ID, conditional on the `updater_gen` field being equal to the ID that was read when read the instance record. This is equivalent to the atomic compare-and-swap operation that one might use to implement a non-distributed lock in a single address space.
  - If this fails because the generation number is outdated, try again (i.e. goto (2)).
  - If this succeeds, the lock was acquired successfully.

Additionally, we can add a method for unlocking an instance record by clearing the `updater_id` field and incrementing the `updater_gen`. This query is conditional on the `updater_id` field being equal to the saga's UUID, to prevent cases where we accidentally unlock an instance that was locked by a different saga.

Introducing this distributed lock is considered fairly safe, as it will only ever be acquired in a saga, and the reverse action for the saga action that acquires the lock will ensure that the lock is released if the saga unwinds. Essentially, this is equivalent to a RAII guard releasing a lock when a thread panics in a single-threaded Rust program.

Presently, none of these methods are actually used. The saga that uses them will be added in PR #5749. I've factored out this change into its own PR so that we can merge the foundation needed for that branch. Hopefully this makes the diff a bit smaller and easier to review, as well as decreasing merge conflict churn with the schema changes.


nexus/db-queries/src/db/datastore/instance.rs
```diff
@@ -529,4 +530,369 @@ impl DataStore {
...
+    /// # Returns
+    ///
+    /// - [`Ok`]`((`[Instance`]`, true))` if the lock was acquired.
+    /// - [`Ok`]`((`[Instance`]`, false))` if the lock was not acquired because
+    ///   the instance was already locked by another saga, or because the
+    ///   generation has advanced since `current_gen` was captured.
+    /// - [`Err`]`(`[`Error::ObjectNotFound`]`)` if no instance with the
+    ///   provided ID exists (or the instance record has been deleted).
+    /// - [`Err`]`(other)` if the database query failed for another reason.
+    pub async fn instance_updater_try_lock(
```
smklein, line 668: I noticed `instance_updater_lock` is pub, and looks like a higher-level interface. Did you want this be pub too?

In particular: This function **does not check** that the value of `updater_id` is currently `None`, so it seems like a caller could invoke this method and overwrite an update lock that is already held.

I think the invocation of this function by `instance_updater_lock` is safe, but that usage could be dangerous if the association of `current_gen` -> `updater_id = None` is not carefully respected

nexus/db-queries/src/db/datastore/instance.rs
```diff
@@ -529,4 +530,369 @@ impl DataStore {
...
+                    authz_instance,
+                    gen,
+                    saga_lock_id,
+                )
+                .await?;
+        }
+    }
+
+    /// Attempts to lock an instance's record to apply state updates in an
+    /// instance-update saga, if the pprovided `current_gen` matches the
```
smklein, line 633:
```suggestion
    /// instance-update saga, if the provided `current_gen` matches the
```

nexus/db-model/src/instance.rs
```diff
@@ -179,6 +179,18 @@ pub struct InstanceRuntimeState {
...
+    /// If this is [`None`] the instance is not locked. Otherwise, if this is
+    /// [`Some`], this contains the UUID of the saga holding the lock.
+    ///
+    /// This field is guarded by the instance's `updater_gen`
+    #[diesel(column_name = updater_id)]
+    pub updater_id: Option<Uuid>,
+
+    /// The generation number for the updater lock.
+    #[diesel(column_name = updater_gen)]
+    pub updater_gen: Generation,
```
smklein, line 193: Can we add docs specifying that this is incremented when the lock is *successfully* acquired or released?

(On first read, I thought this was "on acquire only", but that's not the case!)

nexus/db-queries/src/db/datastore/instance.rs
```diff
@@ -529,4 +530,369 @@ impl DataStore {
...
+
+        let locked = diesel::update(dsl::instance)
+            .filter(dsl::time_deleted.is_null())
+            .filter(dsl::id.eq(instance_id))
+            // If the generation is the same as the captured generation when we
+            // read the instance record to check if it was not locked, we can
+            // lock this instance. Otherwise, the update will fail. This query
+            // is equivalent to an atomic compare-and-swap instruction in a
+            // non-distributed single-process mutex.
+            .filter(dsl::updater_gen.eq(current_gen))
```
smklein, line 690: In other words: There's an assumption that if `updater_gen` is the same, the **value** of `updater_id` is `None`?

nexus/db-queries/src/db/datastore/instance.rs
```diff
@@ -529,4 +530,369 @@ impl DataStore {
...
+    use nexus_types::external_api::params;
+    use omicron_common::api::external::ByteCount;
+    use omicron_common::api::external::IdentityMetadataCreateParams;
+
+    use omicron_test_utils::dev;
+
+    #[tokio::test]
+    async fn test_instance_updater_acquires_lock() {
+        // Setup
+        let logctx = dev::test_setup_log("test_empty_blueprint");
```
smklein, line 783: Nit: test name copy-pasted?

nexus/db-queries/src/db/datastore/instance.rs
```diff
@@ -529,4 +530,369 @@ impl DataStore {
...
+        datastore
+            .instance_updater_unlock(&opctx, &authz_instance, &saga1)
+            .await
+            .expect("instance must be unlocked by saga 1");
+
+        // now, locking the instance from saga 2 should succeed.
+        assert_locked!(saga2);
+
+        // trying to lock the instance again from saga 1 should fail
+        assert_not_locked!(saga1);
```
smklein, line 892: This is great, I appreciate how readable the test is!

Could we add a test for double-locking / double-unlocking by the same saga?

smklein (approved):
Looks great. I know there has been a bunch of hesitation around the usage of distributed locks, but this implementation is clear, narrowly focused around sagas, and well thought-out!

nexus/db-model/src/instance.rs
```diff
@@ -179,6 +179,18 @@ pub struct InstanceRuntimeState {
     /// This field is guarded by the instance's `gen`.
     #[diesel(column_name = migration_id)]
     pub migration_id: Option<Uuid>,
+
+    /// The UUID of the saga currently holding the update lock on this instance.
+    /// If this is [`None`] the instance is not locked. Otherwise, if this is
+    /// [`Some`], this contains the UUID of the saga holding the lock.
```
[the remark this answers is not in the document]
hawkw: it was *originally* intended to be the saga's UUID, but it turned out that
there`s actually no way for a saga to get its own UUID. so, it's "a UUID
generated by a saga", instead. in practice, it's not super important that it be
the actual ID of the saga. i'll update the comment regardless.

nexus/db-queries/src/db/datastore/instance.rs
hawkw: Yeah, you're right, this shouldn't be public --- honestly, it can probably just
be folded into the `instance_updater_lock` method. I'll go ahead and do that.

nexus/db-queries/src/db/datastore/instance.rs
```diff
@@ -529,4 +530,369 @@ impl DataStore {
...
+    /// - [`Err`]`([`Error::InternalError`])` if the instance was locked by
+    ///   another saga.
+    /// - [`Err`]`(`[`Error::ObjectNotFound`]`)` if no instance with the
+    ///   provided ID exists (or the instance record has been deleted).
+    /// - [`Err`]`(other)` if the database query failed for another reason.
+    pub async fn instance_updater_lock(
+        &self,
+        opctx: &OpContext,
+        authz_instance: &authz::Instance,
+        saga_lock_id: &Uuid,
```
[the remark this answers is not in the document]
hawkw: Hmm, yeah...I also thought about having it return `Result<Option<Instance>,
Error>` to distinguish between "you can't get ye lock" and "something unrelated
went wrong", but I felt like the "lock not acquired" case should _probably_ be
a `Result` so that you can't easily ignore it...

nexus/db-queries/src/db/datastore/instance.rs
```diff
@@ -529,4 +540,397 @@ impl DataStore {
...
+                dsl::updater_gen.eq(dsl::updater_gen + 1),
+                dsl::updater_id.eq(None::<Uuid>),
+            ))
+            .check_if_exists::<Instance>(instance_id)
+            .execute_and_check(&*self.pool_connection_authorized(opctx).await?)
+            .await
+            .map(|r| match r.status {
+                UpdateStatus::Updated => true,
+                // TODO(eliza): should this be an error?
+                UpdateStatus::NotUpdatedButExists => false,
```
gjcolombo, line 722: I think this case does indicate a programming error in Nexus. I would be sort of inclined to convert it to an error. 

IIUC, we'll hit this case whenever the instance ID appears in the instance table, but one of the filter conditions didn't hold. This implies that either that

1. someone is dropping a lock for a deleted instance, or
2. the lock ID is different from what the caller supplied

Case 2 seems manifestly erroneous to me--the caller is trying to drop a lock it doesn't own. I think case 1 is also going to be a bug, but it technically depends on what the instance update saga ends up looking like. If we ignore the Failed state for a minute, and the saga guarantees the following, then it shouldn't be possible to drop a lock for a destroyed instance:

a. The saga only acquires the lock when an instance has an active VMM.
b. If the saga decides to retire an instance's VMMs and move the instance to Stopped, then it will drop the lock in the same atomic database operation.

If these rules hold, an instance can only be destroyed when the lock's not held, and so an instance should never get deleted out from under a lock owner.

The Failed state makes all this more complicated because Nexus will happily move an instance to Failed of its own volition, outside of anything the instance update functions might be doing, and it's possible to delete a Failed instance. We'll need to consider this carefully as we draft the instance update saga (and we should also make Nexus use the Failed state more judiciously, but that's for another set of PRs).

gjcolombo (approved):
LGTM with one minor comment fix. I also left a comment discussing what should happen if the unlock statement gets a `NotUpdatedButExists` error, but until there are some callers of this function I'm not totally committed to making that case return an error (or blocking the PR for it).

nexus/db-model/src/instance.rs
```diff
@@ -179,6 +179,23 @@ pub struct InstanceRuntimeState {
...
+    /// rather, a UUID *generated by* that saga. Therefore, it may not be
+    /// useable to look up which saga holds the lock.
+    ///
+    /// This field is guarded by the instance's `updater_gen`
+    #[diesel(column_name = updater_id)]
+    pub updater_id: Option<Uuid>,
+
+    /// The generation number for the updater lock. This is updated whenever the
+    /// lock is acquired or released, and is used in attempts to set the
+    /// `updater_id` field to
```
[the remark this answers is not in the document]
hawkw: Whoops, good catch!

nexus/db-queries/src/db/datastore/instance.rs
hawkw: Hmm, my thinking was just that:
  1. we (will) have a saga action that releases a lock
  2. saga actions must be idempotent
  3. therefore, releasing a lock must be idempotent

But, I didn't think *too* hard about it, just treating "saga actions must be idempotent" as a hard requirement. 

What I don't want, though, is a situation where a Nexus instance releases the lock but then crashes before it has recorded execution of that saga node, and then a subsequent Nexus resumes execution of that saga, attempts to release the lock again, and fails because it has already been unlocked, and then unwinds the whole saga, undoing a bunch of otherwise fine state transitions. Is that a valid concern?

gjcolombo: You're right, I missed the idempotency angle here. Without further refining this query I think that does require us to get an `Ok` value here. (The query seems to handle this case correctly: if someone presenting some lock ID gets a result of `Updated`, then repeats the query, the filter on line 711 should prevent the lock from being released again.)

hawkw: Okay, I _think_ the change from ceebdbb1abcd47f8ddf58638827a9f7206f48063 is a reasonable compromise between idempotency and "trying to unlock someone else's lock is Always Wrong And Bad". Now, `instance_updater_unlock` takes a generation number as well as a saga lock ID, and we return `Ok(false)` if the unlock query didn't do anything because the *generation number* has changed, but we return an error if the generation number is correct but the lock is held by a different saga. This way, if you unlock your lock twice, that doesn't error, but if you try to unlock someone *else*'s lock at the correct generation, you get an error.

What do you think?

nexus/db-queries/src/db/datastore/instance.rs
```diff
@@ -529,4 +542,525 @@ impl DataStore {
...
+            // The generation has advanced past the generation at which the
+            // lock was held. This means that we have already released the
+            // lock. Return `Ok(false)` here for idempotency.
+            UpdateAndQueryResult {
+                status: UpdateStatus::NotUpdatedButExists,
+                ref found,
+            } if found.runtime_state.updater_gen > current_gen => Ok(false),
+            // The instance exists, but the lock ID doesn't match our lock ID.
+            // This means we were trying to release a lock we never held, whcih
+            // is almost certainly a programmer error.
```
gjcolombo, line 751: Or that someone stole the lock from the logically correct owner. (Should we also assert/log an error in `instance_updater_lock` if it observes that its attempt to lock returned `UpdateStatus::Updated` but the lock owner wasn't previously NULL?)

nexus/db-queries/src/db/datastore/instance.rs
```diff
@@ -529,4 +542,525 @@ impl DataStore {
...
+    ///
+    /// This method will unlock the instance if (and only if) the lock is
+    /// currently held by the provided `saga_lock_id`. If the lock is held by a
+    /// different saga UUID, the instance will remain locked. If the instance
+    /// has already been unlocked, this method will return `false`.
+    ///
+    /// # Arguments
+    ///
+    /// - `authz_instance`: the instance to attempt to unlock
+    /// - `current_gen`: the generation number for the lock being released.
```
gjcolombo, line 695: I would possibly rename this to something like `after_lock_acquired_gen` and note explicitly that it needs to be the lock generation number that was produced by acquiring the lock. (For many sagas I expect it will be very easy to reach for the generation number that was in the instance record when the saga started.)

Maybe we should leverage the type system here: have `instance_updater_lock` return an opaque struct that contains the lock ID and the post-acquisition generation and make `instance_updater_unlock` take that type. Harder to mishandle the values that way. WDYT?

hawkw: > Maybe we should leverage the type system here: have `instance_updater_lock` return an opaque struct that contains the lock ID and the post-acquisition generation and make `instance_updater_unlock` take that type. Harder to mishandle the values that way. WDYT?

I actually did this, originally, on the other branch! I removed it because I wasn't sure if a saga node's reverse action would be able to access the output of its forward action, so I wasn't sure if there was a way for the "release lock on unwind" reverse action to get the opaque token or not. If that's possible, I'll definitely try this approach.

nexus/db-queries/src/db/datastore/instance.rs
hawkw: We could probably just make `.filter(dsl::updater_id.is_null())` a condition of the query as well, honestly. The generation number *should* guard it, but there's probably no reason *not* to also check that the ID is null in the query.

nexus/db-queries/src/db/datastore/instance.rs
gjcolombo: > I removed it because I wasn't sure if a saga node's reverse action would be able to access the output of its forward action

I believe you can; see e.g. the `CREATE_VMM_RECORD` step in the instance start saga, where `sis_destroy_vmm_record` refers to the `vmm_record` output from `sis_create_vmm_record`.

hawkw: Ah, wonderful. In that case, I'll see about making this change.

PR #5833, opened 2024-05-29
Fix incorrect timeseries fields in error message

### LINKED ISSUE #5832: OxQL error message for unknown filter fields is wrong
When applying a filter table op, we match the field names in the query with the fields in the schema of the timeseries. If the filter references non-existent fields, we return an error message saying so. We specify the _valid_ fields here:

https://github.com/oxidecomputer/omicron/blob/1fe55e9e44f8ff5bd415ba8246a40b68d4ded0d0/oximeter/db/src/oxql/ast/table_ops/filter.rs#L319

But that's wrong. That lists the fields in the _query_, not in the timeseries _schema.

As a side note, this is done too late. We should do this check _before_ fetching any data from the database, if possible. I'm testing this out on dogfood, and the query takes a _long_ time (~15s). That's because we appear to be fetching all the data from the database, and _then_ doing this check. We can do it earlier to avoid any of that work.

bnaecker:
- Fixes #5832
- Uses timeseries fields in error message rather than filter identifiers

### DIFF
--- a/oximeter/db/src/oxql/ast/table_ops/filter.rs
+++ b/oximeter/db/src/oxql/ast/table_ops/filter.rs
@@ -298,11 +298,11 @@ impl Filter {
             // You give nothing, you get nothing.
             return Ok(tables.to_vec());
         };
-        let ident_names = self.ident_names();
 
         // There are extra, implied names that depend on the data type of the
         // timeseries itself, check those as well.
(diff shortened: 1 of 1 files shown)


bnaecker:
Friendly ping on this @david-crespo, should be quick!

david-crespo (approved):

PR #5834, opened 2024-05-30
Store comments as strongly-typed operations

smklein:
Attempts to resolve the discussion in https://github.com/oxidecomputer/omicron/pull/5599#discussion_r1576940017 by using a strongly-typed `Operation` enum instead of arbitrary String comments.


sunshowers (approved):
Thanks!

Issue #4511, opened 2023-11-17
Stop request for an instance with an abandoned VMM succeeds, returns state of "Running," and never stops the instance


gjcolombo:
Observed in dogfood with an instance that was running and not stopped before the rack was parked for update. This is very closely related to #725, but we might be able to mitigate it tactically without solving the entire "what to do when a sled reboots" problem described in that issue.

The specific problem here is that if an instance is Running, but isn't registered with its purported sled agent (e.g. because the sled rebooted and the agent lost the instance), a request to stop the instance will succeed but not change the instance's state. I originally did this intending to make stop requests idempotent:

https://github.com/oxidecomputer/omicron/blob/781ed1236489bb536506fb03ba1be39b34ed81a5/sled-agent/src/instance_manager.rs#L335-L351

But arguably this would be better to handle in Nexus instead: if sled agent can promise that a call to `ensure_state` returns a specific HTTP status code if and only if the instance is not known to that sled agent, then Nexus can reason as follows: "this sled doesn't know about this instance, so I, Nexus, own its state; and this was a request to stop, so I can try to update the state to reflect that the instance is actually stopped now."

gjcolombo:
Discussed at the 9/26/24 hypervisor huddle. Sled agent's `InstanceManager::ensure_state` still has the problematic behavior here.

> But arguably this would be better to handle in Nexus instead: if sled agent can promise that a call to ensure_state returns a specific HTTP status code if and only if the instance is not known to that sled agent, then Nexus can reason as follows: "this sled doesn't know about this instance, so I, Nexus, own its state; and this was a request to stop, so I can try to update the state to reflect that the instance is actually stopped now."

Sled agent has such an HTTP error code now, so it just needs to return it in this particular case; Nexus can then decide whether to handle that code specially if the request was a request to stop (though I suspect it shouldn't--the relevant VMM should just be marked as Failed as with any other state change that tries to target an unexpectedly-nonexistent VMM).

gjcolombo:
The `test_instance_failed_after_sled_agent_forgets_vmm_can_be_stopped` instance integration test actually checks this case, but it uses the simulated sled agent, which handles missing VMMs differently from the sled agent InstanceManager. If we make the instance manager's behavior match the simulated agent's I think we'll be on the right track here.

hawkw:
Oh no...does the real life sled-agent not actually do the right thing? :grimacing:

gjcolombo:
[...]

PR #5838, opened 2024-05-30
[nexus] Add a schema change to fix instance counter underflow

smklein:
This is a corollary PR to https://github.com/oxidecomputer/omicron/pull/5830 , which fixed the root cause.

Due to a bug in the virtual provisioning query, it was possible to undercount virtual provisioning information for
instances, which would result in an integer underflow for "total CPU/RAM provisioned" for a {project, silo, fleet}.

Although #5830 fixed the root cause, it's possible that in-field systems have an invalid value if they experienced this bug.

This PR uses a schema change, exploiting the fact that schema changes occur with instances offline, to reset these values to a known value.


schema/crdb/fix-provisioning-counters/up.sql
```diff
@@ -0,0 +1,21 @@
+-- This change fixes provisioning counters, alongside the
+-- underflow fix provided in https://github.com/oxidecomputer/omicron/pull/5830.
+-- Although this underflow has been fixed, it could have resulted
+-- in invalid accounting, which is mitigated by this schema change.
+--
+-- This update is currently occurring offline, so we exploit
+-- that fact to identify that all instances should be off.
```
gjcolombo, line 7: nit: IMO the word "should" is doing more work than average here (there could be a Running instance if the servicing operator failed to stop it or hit something like #4511 that prevented them from stopping it). I think this is safe anyway for the reasons we discussed in chat (provisioning counter cleanup is idempotent now, so this won't impede other attempts to tell Nexus that the VMM is gone/drive the instance to Stopped), but I might leave a short note here to that effect.

gjcolombo (approved):
LGTM, thanks for doing this!

schema/crdb/fix-provisioning-counters/up.sql
smklein: Sounds good, updated the comment here to elaborate on this point.

PR #5602, opened 2024-04-23
Clear XDE underlay when destroying virtual hardware

FelixMcFelix:
OPTE now prevents itself from being unloaded if its underlay state is set. Currently, underlay setup is performed only once, and it seems to be the case that XDE can be unloaded in some scenarios (e.g., `a4x2` setup).

However, a consequence is that removing the driver requires an extra operation to explicitly clear the underlay state. This PR adds this operation to the `cargo xtask virtual-hardware destroy` command.

This is ready for review, but we want to merge after:
- [x] New lab image for OPTE v0.31.x is built and ready.
- [x] #5722 is merged.
- [x] oxidecomputer/maghemite#258 is merged, and SHAs are updated in this PR.

Closes #5314.

PR #5842, opened 2024-05-31
a few updates to get things to come up in a4x2

rcgoodfellow:


Nieuwejaar (commented):
I have the database fix in my workspace, but I'm happy to just pull it from here.  

Do you know how the opte/buildomat relationship is managed?  Presumably Josh would update the buildomat systems, but that would break everybody else.  So, it seems like the OPTE update should go straight to the main branch, where everybody else can get it.

FelixMcFelix:
> Do you know how the opte/buildomat relationship is managed? Presumably Josh would update the buildomat systems, but that would break everybody else. So, it seems like the OPTE update should go straight to the main branch, where everybody else can get it.

I have this OPTE update tee-d up in #5602 (since device teardown is now pretty different in a breaking way). It's currently waiting on the open lab ticket for a `lab-2.0-opte-0.31` image to be built. I think @jclulow still has to do this manually, but creation of a new image doesn't break `main` since each is pinned to an exact OPTE version.

rcgoodfellow:
The maghemite being used in this branch depends on OPTE API version 31, so I think one way or another OPTE needs to be pulled in here, or we explicitly roll back maghemite which seems less appealing.

rcgoodfellow:
Since #5602 has landed, once the `vlan` branch is rebased, this PR becomes obviated. Closing.

PR #5844, opened 2024-05-31
[nexus] switch nexus internal API to a trait, add scaffolding to manage it

sunshowers:
Part of the work on [RFD 479 Dropshot API traits](https://rfd.shared.oxide.computer/rfd/0479).

As a first, impactful example, this PR switches over the Nexus internal API to being defined by a trait. Some minor refactoring was required but overall it took me around half an hour to do this.

Included also is a tool, `cargo xtask openapi`, which is used to manage this and future documents, as well as instructions that folks can hopefully follow.


dev-tools/openapi-manager/src/check.rs
```diff
@@ -0,0 +1,185 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use std::{io::Write, process::ExitCode};
```
sunshowers, line 5: A lot of this code is quite generic, and I think can live separately. What do you think about moving this out to the dropshot repo or its own repo?

sunshowers: In a meeting yesterday we decided to wait and see -- leave it in omicron for now but be open to the possibility of putting this code in a central location later.

openapi/nexus-internal.json
```diff
@@ -3955,6 +3947,14 @@
           "last_port"
         ]
       },
+      "ProbeExternalIpKind": {
```
ahl, line 3950: was this renamed on purpose?

workspace-hack/Cargo.toml
```diff
@@ -25,7 +25,8 @@ bit-set = { version = "0.5.3" }
 bit-vec = { version = "0.6.3" }
 bitflags-dff4ba8e3ae991db = { package = "bitflags", version = "1.3.2" }
 bitflags-f595c2ba2a3f28df = { package = "bitflags", version = "2.5.0", default-features = false, features = ["serde", "std"] }
-bstr = { version = "1.9.1" }
+bstr-6f8ce4dd05d13bba = { package = "bstr", version = "0.2.17" }
+bstr-dff4ba8e3ae991db = { package = "bstr", version = "1.9.1" }
```
ahl, line 29: is this intentional or do we want a single bstr?

ahl (approved):

workspace-hack/Cargo.toml
sunshowers: hah good question -- this happened because I pulled in the `similar` library to do diffs, which depends on bstr 0.2. I filed https://github.com/mitsuhiko/similar/issues/66 about this.

openapi/nexus-internal.json
sunshowers: Yeah -- there's already an `ExternalIp` type at https://github.com/oxidecomputer/omicron/blob/468e699bacc2b5d0b76e4990e9edfba191049247/nexus/types/src/external_api/views.rs#L428 which is exported by the Nexus external API. Basically we have several types all called something like `ExternalIp` that are mostly but not entirely isomorphic (e.g. some of them carry around database UUIDs). We should probably revisit this at some point.

sunshowers:
Will fix the test failures and rebase the PR onto main once #6056 lands.

PR #5847, opened 2024-05-31
Fix IP pool silos pagination bug

### LINKED ISSUE #5837: IP pool linked silos endpoint pagination broken
The linked silo tab only looks to be displaying the first page of results: https://oxide.sys.r3.oxide-preview.com/system/networking/ip-pools/default?tab=silos

Clicking next page results in an empty list of results, and we know that there are more than 25 linked silos.

Looking at the API calls that are being made this may be an issue in Nexus itself. The API itself with the page token is not returning any results, whereas if we request more than 25 results we can the expected amount back.

david-crespo:
Closes #5837

- [x] Write test reproducing the bug
- [x] Fix the bug

### DIFF
--- a/nexus/tests/integration_tests/ip_pools.rs
+++ b/nexus/tests/integration_tests/ip_pools.rs
@@ -724,6 +724,54 @@ async fn test_ip_pool_pagination(cptestctx: &ControlPlaneTestContext) {
     assert_eq!(get_names(next_page.items), &pool_names[5..8]);
 }
 
+#[nexus_test]
+async fn test_ip_pool_silos_pagination(cptestctx: &ControlPlaneTestContext) {
+    let client = &cptestctx.external_client;
+
+    // one pool, and there should be no linked silos
+    create_pool(client, "p0").await;
+    let silos_p0 = silos_for_pool(client, "p0").await;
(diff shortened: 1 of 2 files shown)


nexus/tests/integration_tests/ip_pools.rs
```diff
@@ -724,6 +724,54 @@ async fn test_ip_pool_pagination(cptestctx: &ControlPlaneTestContext) {
...
+        "{}?limit=5&page_token={}",
+        base_url,
+        first_five.next_page.unwrap()
+    );
+    let next_page =
+        objects_list_page_authz::<IpPoolSiloLink>(client, &next_page_url).await;
+    assert_eq!(
+        next_page.items.iter().map(|s| s.silo_id).collect::<Vec<_>>(),
+        &silo_ids[5..8]
+    );
```
david-crespo, line 772: This assert fails because `next_page` is empty. This is the bug.

```
assertion `left == right` failed
  left: []
 right: [d9184cf8-578d-44e7-9d21-925e8f20fb2f, e0518aa9-664d-4f5e-8259-c7ba3d663c4e, f605f0d5-3759-4ccd-b7a2-d19f3e19eefc]
```

PR #5849, opened 2024-05-31
[nexus] Split authn/authz and db-fixed-data into new crates

smklein:
As a part of the ongoing effort to split Nexus into smaller pieces, this PR splits out two new crates:

- `nexus-auth` takes the contents of `nexus/db-queries/src/auth{n,z}`, as well as `nexus/db-queries/src/context.rs`, and separates this logic into a new bespoke crate. Although this crate **does** have a dependency on the datastore itself, it only actually invokes a single method, and can be abstracted via a new trait, defined in `nexus/auth/storage`.
- `nexus-db-fixed-data` takes the contents of `nexus/db-queries/src/db/fixed-data` and separates this logic into a new crate.


nexus/auth/src/authn/silos.rs
```diff
@@ -68,57 +62,6 @@ pub enum IdentityProviderType {
     Saml(SamlIdentityProvider),
 }
 
-impl IdentityProviderType {
```
smklein, line 71: This method was moved into `nexus/db-queries/src/db/datastore/identity_provider.rs` - it doesn't need to be defined here, and is tightly coupled with the datastore, not auth logic.

nexus/auth/src/authz/api_resources.rs
```diff
@@ -103,27 +101,27 @@ pub trait ApiResourceWithRoles: ApiResource {
...
-impl<T: ApiResource + oso::PolarClass + Clone> AuthorizedResource for T {
-    fn load_roles<'a, 'b, 'c, 'd, 'e, 'f>(
+impl<T> AuthorizedResource for T
+where
+    T: ApiResource + oso::PolarClass + Clone,
+{
+    fn load_roles<'a, 'b, 'c, 'd, 'e>(
         &'a self,
         opctx: &'b OpContext,
-        datastore: &'c DataStore,
```
smklein, line 114: You'll see this a lot in this PR: Places where we pass both `opctx` and `datastore` actually don't *need* to pass both, since `opctx` itself contains a reference to the `datastore`.

By refactoring this logic, I cut down on the number of arguments, here and elsewhere in this PR.

Since we're abstracting away access to the datastore as a `dyn Storage` trait now, this also will be easier to refactor in the future.

nexus/auth/src/authz/context.rs
```diff
@@ -51,7 +52,6 @@ impl Authz {
         self.oso.is_allowed(actor.clone(), action, resource.clone())
     }
 
-    #[cfg(test)]
```
smklein, line 54: The `policy_test` (look for the files changed in this PR) was previously a "unit test" in `db-queries` that used auth and datastore methods extensively.

It has stayed in `db-queries` to access the database methods, but still needs to access methods defined in `nexus-auth`. As a result, some `cfg(test)` methods are exposed as `pub` to make this work. This is one such example.

nexus/auth/src/authz/context.rs
```diff
@@ -66,18 +66,22 @@ impl Authz {
 pub struct Context {
     authn: Arc<authn::Context>,
     authz: Arc<Authz>,
-    datastore: Arc<DataStore>,
+    datastore: Arc<dyn Storage>,
```
smklein, line 69: This is another common pattern through this PR: Wherever `nexus/auth` depends on the `DataStore`, I basically use dependency injection via `dyn Storage` to not **directly** depend on the datastore. Then, the implementation of `Storage` allows `nexus/auth` to access a very limited subset of the datastore.

Turns out, that "limited interface" is only a single method. See: `nexus/auth/src/storage.rs`

nexus/auth/src/authz/context.rs
```diff
@@ -196,17 +196,46 @@ pub trait AuthorizedResource: oso::ToPolar + Send + Sync + 'static {
...
+            &self,
+            _opctx: &OpContext,
+            _identity_type: IdentityType,
+            _identity_id: Uuid,
+            _resource_type: ResourceType,
+            _resource_id: Uuid,
+        ) -> Result<Vec<RoleAssignment>, Error> {
+            todo!();
+        }
+    }
```
smklein, line 233: This is an example of a test that actively does not need access to the database, and can be replaced by a stub.

nexus/auth/src/authz/mod.rs
```diff
@@ -179,9 +181,8 @@ pub use context::Context;
 
 mod oso_generic;
 pub use oso_generic::Action;
+pub use oso_generic::Database;
 pub use oso_generic::DATABASE;
 
 mod roles;
-
-#[cfg(test)]
-mod policy_test;
```
smklein, line 187: This test still exists, it has just been kept in `db-queries`

nexus/auth/src/authz/oso_generic.rs
```diff
@@ -172,8 +172,7 @@ pub fn make_omicron_oso(log: &slog::Logger) -> Result<OsoInit, anyhow::Error> {
 ///
 /// There's currently just one enum of Actions for all of Omicron.  We expect
 /// most objects to support mostly the same set of actions.
-#[derive(Clone, Copy, Debug, Eq, PartialEq)]
-#[cfg_attr(test, derive(strum::EnumIter))]
```
smklein, line 176: Another example of "I need this trait in `policy_test`, but it's not part of the same crate anymore"

nexus/auth/src/context.rs
```diff
@@ -346,19 +353,42 @@ mod test {
...
+            &self,
+            _opctx: &OpContext,
+            _identity_type: IdentityType,
+            _identity_id: Uuid,
+            _resource_type: ResourceType,
+            _resource_id: Uuid,
+        ) -> Result<Vec<RoleAssignment>, Error> {
+            todo!();
+        }
+    }
```
smklein, line 385: Another example of a test that previously required a whole database, but actually, never queries anything from the database

nexus/auth/src/storage.rs
```diff
@@ -0,0 +1,27 @@
...
+
+use crate::context::OpContext;
+use nexus_db_model::IdentityType;
+use nexus_db_model::RoleAssignment;
+use omicron_common::api::external::Error;
+use omicron_common::api::external::ResourceType;
+use uuid::Uuid;
+
+#[async_trait::async_trait]
+pub trait Storage: Send + Sync {
```
smklein, line 18: I'd be interested in figuring out a way to make this not use async-trait.

I basically went down this route because:

- authz::Context contains one of these objects
- If it uses generics, then `OpContext` also needs a generic parameter
- That changes a *lot* of code
- ... so I went down the `dyn Storage` route instead, as the DataStore is already wrapped in an Arc

nexus/db-fixed-data/src/project.rs
```diff
@@ -2,12 +2,14 @@
...
 
-use crate::db;
-use crate::db::datastore::SERVICES_DB_NAME;
+use nexus_db_model as model;
 use nexus_types::external_api::params;
 use omicron_common::api::external::IdentityMetadataCreateParams;
 use once_cell::sync::Lazy;
 
+/// The name of the built-in Project and VPC for Oxide services.
+pub const SERVICES_DB_NAME: &str = "oxide-services";
```
smklein, line 11: This is just moved, not new

nexus/db-queries/src/db/datastore/auth.rs
```diff
@@ -0,0 +1,81 @@
...
+use nexus_db_model::RoleAssignment;
+use omicron_common::api::external::Error;
+use omicron_common::api::external::ResourceType;
+use uuid::Uuid;
+
+#[async_trait::async_trait]
+impl Storage for super::DataStore {
+    /// Return the built-in roles that the given built-in user has for the given
+    /// resource
+    async fn role_asgn_list_for(
```
smklein, line 25: This method was moved exactly from `nexus/db-queries/src/db/datastore/role.rs`

nexus/db-queries/src/db/datastore/identity_provider.rs
```diff
@@ -11,18 +11,66 @@ use crate::db;
...
+use nexus_auth::authn::silos::IdentityProviderType;
 use omicron_common::api::external::http_pagination::PaginatedBy;
 use omicron_common::api::external::CreateResult;
 use omicron_common::api::external::ListResultVec;
+use omicron_common::api::external::LookupResult;
 use omicron_common::api::external::ResourceType;
 use ref_cast::RefCast;
 
 impl DataStore {
+    pub async fn identity_provider_lookup(
```
smklein, line 30: This method was moved exactly from `IdentityProvider::lookup` to here (and `DataStore` is now a `&self` parameter instead of the first argument)

nexus/auth/src/authn/external/mod.rs
```diff
@@ -9,6 +9,7 @@ use super::SiloAuthnPolicy;
 use crate::authn;
 use async_trait::async_trait;
 use authn::Reason;
+use slog::trace;
```
hawkw, line 12: i'm assuming this is because there used to be a `#[macro_use] somewhere?

nexus/auth/src/authz/api_resources.rs
hawkw: Oh, lovely --- I hadn't realized `OpContext` also had a `DataStore` in it, that would have made some of my refactor-things-into-free-functions change a bit nicer. Whoops!

nexus/auth/src/authz/api_resources.rs
```diff
@@ -103,27 +101,27 @@ pub trait ApiResourceWithRoles: ApiResource {
...
     where
-        'a: 'f,
-        'b: 'f,
-        'c: 'f,
-        'd: 'f,
-        'e: 'f,
+        'a: 'e,
+        'b: 'e,
+        'c: 'e,
+        'd: 'e,
```
hawkw, line 122: Nit, take it or leave it: personally, I'd prefer to give these lifetimes semantically meaningful names, like
```rust
    fn load_roles<'resource, 'opctx, 'authn, 'roles, 'future>(
        &'resource self,
        opctx: &'opctx OpContext,
        authn: &'authn authn::Context,
        roleset: &' mut RoleSet,
    ) -> futures::future::BoxFuture<'future, Result<(), Error>>
    where
        'resource: 'future,
        'opctx: 'future,
        'authn 'future,
        'roleset: 'future,
    {
```

because, honestly, seeing
```rust
    where
        'a: 'e,
        'b: 'e,
        'c: 'e,
        'd: 'e,
```
just kinda feels like nonsense to me.

Up to you, though; I agree that single-letter lifetime param names does seem to be the more common idiom... :woman_shrugging:

(as an aside, I'm not convinced we even need so many lifetime params --- can't they all just be `'future` references, and the compiler will coerce them? but maybe that doesn't actually work, I dunno...)

nexus/auth/src/authz/context.rs
```diff
@@ -196,17 +196,46 @@ pub trait AuthorizedResource: oso::ToPolar + Send + Sync + 'static {
...
+    impl crate::storage::Storage for FakeStorage {
+        async fn role_asgn_list_for(
+            &self,
+            _opctx: &OpContext,
+            _identity_type: IdentityType,
+            _identity_id: Uuid,
+            _resource_type: ResourceType,
+            _resource_id: Uuid,
+        ) -> Result<Vec<RoleAssignment>, Error> {
+            todo!();
```
hawkw, line 230: perhaps we should change this to something like 
```suggestion
            unimplemented!("this test shouldn't need to actually access the database");
```
or something, so that if someone introduces a change that invalidates this assumption, it's more obvious what went wrong?

nexus/auth/src/authz/oso_generic.rs
hawkw: We could, potentially, make the `strum` dependency optional and make this:
```rust
#[cfg_attr(feature = "strum", derive(strum::EnumIter)]
```
and then have `policy_test` enable the `strum` feature in its dev-dependency on `authz`...but i'm not sure if that's *really* worth the effort, up to you.

nexus/db-fixed-data/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "nexus-db-fixed-data"
+version = "0.1.0"
+edition = "2021"
+license = "MPL-2.0"
+
```
hawkw, line 6: it could be nice to have a `description` here saying what this crate is for?

hawkw (approved):
Overall, this seems good to me, although I had a few small style suggestions --- take them or leave them! And, I've mostly tried to avoid commenting on the code that was moved around but hasn't actually changed.

nexus/auth/src/authz/api_resources.rs
hawkw: (this comment also applies, you know...everywhere else)

nexus/auth/src/authn/external/mod.rs
smklein: Exactly: https://github.com/oxidecomputer/omicron/blob/b0dfd535386cb93361e08a17d926e91867d9cb08/nexus/db-queries/src/lib.rs#L14-L17

nexus/auth/src/authz/api_resources.rs
smklein: Completely agree with you on both accounts. I'll try to get rid of these, seems a little surprising that they ever existed in the first place.

nexus/db-queries/src/db/datastore/auth.rs
andrewjstone: nit: could spell out `assign`, but I also understand not wanting to introduce that to the diff if this is used in a lot of places.

nexus/db-queries/src/policy_test/resource_builder.rs
```diff
@@ -242,7 +242,39 @@ macro_rules! impl_dyn_authorized_resource_for_global {
     };
 }
 
-impl_dyn_authorized_resource_for_global!(authz::oso_generic::Database);
+impl_dyn_authorized_resource_for_resource!(authz::AddressLot);
```
andrewjstone, line 245: Nifty :)

andrewjstone (approved):
Looks great!

nexus/auth/src/authz/api_resources.rs
smklein: Okay, I have:

- Reduced this to a single lifetime parameter
- Renamed it `'fut`, to make it more clear "this is here for the BoxFuture return value"

nexus/auth/src/authz/context.rs
smklein: Sounds good, updated!

nexus/db-fixed-data/Cargo.toml
smklein: Added

nexus/auth/src/authz/api_resources.rs
hawkw: oh, that's _way_ better! thanks! :)

nexus/auth/src/authz/oso_generic.rs
smklein: Yeah, I have mixed feelings on this, because that feels "logically correct", but it's also a pain-in-the-butt to need to test an inflated matrix of possible features.

My belief in *this* case is that the value of (feature flag + reduced API) is smaller than the value of (no feature, slightly larger API). But that's totally an opinion. Would be willing to update that if you disagree!

nexus/db-queries/src/db/datastore/auth.rs
smklein: Grepping for `role_asgn`, this appears to be a pattern used beyond this method. I agree with you that I'd rather it be fully spelled out, but I'm gonna punt on this to avoid touching several files that aren't currently part of this PR?

andrewjstone: That works for me

PR #5593, opened 2024-04-22
[sled-agent] Self assembling switch zone

karencfv:
## Overview

This PR migrates the switch zone to a self assembling format. There are a few bits of old code I'll be cleaning up, more logs I'll be adding and some documentation about how the switch zone flow works, but I'll do this in follow up PRs to keep this one as compact as possible.

## Caveats

I've tested this in a local single node deployment and in the a4x2 testbed. Unfortunately, this is not enough testing to make sure all of the services play nice together on a real rack. We'll have to keep an eye on dogfood when this is deployed.

The only services that depend on the [common networking](https://github.com/oxidecomputer/omicron/blob/30eb1ee38987201ac71f2115fdd89da4b08710c7/zone-setup/src/bin/zone-setup.rs#L578-L690) [service](https://github.com/oxidecomputer/omicron/blob/30eb1ee38987201ac71f2115fdd89da4b08710c7/smf/zone-network-setup/manifest.xml) are dendrite and MGS. While this makes sense on the a4x2 testbed, I'd like to verify that these dependencies make sense when running on a real rack.

As several people have worked on different parts of this zone,  I've tagged a whole bunch of people for review, sorry if this is overkill! Just want to make sure I've got the right eyes on each service of the zone.

Related: https://github.com/oxidecomputer/omicron/issues/1898
Closes: https://github.com/oxidecomputer/omicron/issues/2884

TODO:

- [ ] Update Dendrite hashes after merging https://github.com/oxidecomputer/dendrite/pull/990

PR #5853, opened 2024-06-04
Add optional VLAN ID to uplink address configuration

Nieuwejaar:


common/src/api/internal/shared.rs
```diff
@@ -299,12 +299,81 @@ pub struct RouteConfig {
...
+}
+
 #[derive(Clone, Debug, Deserialize, Serialize, PartialEq, Eq, JsonSchema)]
 pub struct PortConfigV1 {
     /// The set of routes associated with this port.
     pub routes: Vec<RouteConfig>,
-    /// This port's addresses.
-    pub addresses: Vec<IpNet>,
+    /// This port's addresses and optional vlan IDs
+    pub addresses: Vec<UplinkAddressConfig>,
```
rcgoodfellow, line 376: I think this will be a breaking change for the bootstore, e.g. if we try to parse existing bootstore entries into this data structure, the serde deserialization will fail. We could take this on a few different ways.

1. Create a `PortConfigV2` and the associated conversion machinery.
2. Instead of changing the structure of `addresses`, add a `vlans` member that's a map indexed on address list which defaults to the empty set.

Nieuwejaar: Good catch.  I've tested install and reboot, but not install, reboot, upgrade, and reboot.

common/src/api/internal/shared.rs
```diff
@@ -339,7 +430,29 @@ impl From<UplinkConfig> for PortConfigV1 {
     }
 }
 
-/// Deprecated, use PortConfigV1 instead. Cannot actually deprecate due to
+// Deprecated - still present to support importing from older bootstores
```
FelixMcFelix, line 433: Nit: if this is a doc-comment it will be immediately visible in rust-analyzer.
```suggestion
/// Deprecated - still present to support importing from older bootstores
```

nexus/src/app/background/sync_switch_configuration.rs
```diff
@@ -51,8 +51,8 @@ use omicron_common::{
 use serde_json::json;
 use sled_agent_client::types::{
     BgpConfig as SledBgpConfig, BgpPeerConfig as SledBgpPeerConfig,
-    EarlyNetworkConfig, EarlyNetworkConfigBody, HostPortConfig, PortConfigV1,
-    RackNetworkConfigV1, RouteConfig as SledRouteConfig,
+    EarlyNetworkConfig, EarlyNetworkConfigBody, HostPortConfig, PortConfigV2,
+    RackNetworkConfigV2, RouteConfig as SledRouteConfig, UplinkAddressConfig,
```
FelixMcFelix, line 55: I'd probably prefer `RackNetworkConfig` here over an explicit version so that any new versions force changes to be mirrored here. Ditto with `PortConfig`. I think this should play nicely with the existing replace directives pointing at `api::internal::shared` -- if not, then it's not a deal-breaker.

common/src/api/internal/shared.rs
```diff
@@ -154,10 +154,29 @@ pub enum SourceNatConfigError {
 
 // We alias [`RackNetworkConfig`] to the current version of the protocol, so
 // that we can convert between versions as necessary.
-pub type RackNetworkConfig = RackNetworkConfigV1;
+pub type RackNetworkConfig = RackNetworkConfigV2;
```
FelixMcFelix, line 157: Can we have another type alias for `PortConfig = PortConfigV2` here too?

rcgoodfellow (approved):
Thanks Nils. Updates LGTM. We'll definitely want to take this on an upgrade test run on madrid or london.

nexus/src/app/background/sync_switch_configuration.rs
Nieuwejaar: I should have moved it back to "draft", as I expect this to take a few iterations before it's ready for re-review.

Anyway, thanks for taking a look.  This sounds like a good idea.  I'll give it a try and see how it shakes out.

common/src/api/internal/shared.rs
Nieuwejaar: Done.

common/src/api/internal/shared.rs
Nieuwejaar: Done.

nexus/src/app/background/sync_switch_configuration.rs
Nieuwejaar: I took a look at this, and I think it would actually have the opposite effect.  Changing the structure without changing the name would make it easy to overlook a change on the client side - especially if we `#[derive(Default)]`.  

I've made the older versions of the structs non-public, which will force any current users to adapt to the new structure.

FelixMcFelix: Okay, that makes a lot of sense to me -- I think that's a good way to go about enforcing the correct behaviour. Thanks for the extra time spent here!

FelixMcFelix (approved):

sled-agent/src/rack_setup/service.rs
```diff
@@ -734,16 +734,16 @@ impl ServiceInner {
...
             let config = &config.rack_network_config;
-            NexusTypes::RackNetworkConfigV1 {
+            NexusTypes::RackNetworkConfigV2 {
                 rack_subnet: config.rack_subnet,
                 infra_ip_first: config.infra_ip_first,
                 infra_ip_last: config.infra_ip_last,
                 ports: config
                     .ports
                     .iter()
-                    .map(|config| NexusTypes::PortConfigV1 {
```
andrewjstone, line 744: For lack of a better place to put this comment: You should also change the `schema_version` to `2` in the `EarlyNetworkConfig` on line 1139. As we discussed, the schema version doesn't work as I expected or really matter for correctness right now, due to our use of json and not having an intermediates serialized body. But for consistency, it should still be changed I think.

andrewjstone (approved):
Looks great.

sled-agent/src/services.rs
```diff
@@ -3801,6 +3801,7 @@ impl ServiceManager {
             }
         };
 
+        info!(self.inner.log, "Setting up uplinkd service");
```
karencfv, line 3804: I'm going to put https://github.com/oxidecomputer/omicron/pull/5593 for review soon and will be making substantial changes to this file. To avoid merge conflicts (I just finished fixing quite a few! ), do you mind removing these log messages from this PR and I promise to add them in mine?

sled-agent/src/services.rs
```diff
@@ -3811,6 +3812,7 @@ impl ServiceManager {
 
         for port_config in &our_ports {
             for addr in &port_config.addrs {
+                info!(self.inner.log, "configuring port: {port_config:?}");
```
karencfv, line 3815: same as above

sled-agent/src/services.rs
Nieuwejaar: Done

karencfv: Thanks! :bowing_woman:

karencfv: Added here https://github.com/oxidecomputer/omicron/pull/5593/commits/2ea71b8fd613c676e9a17e4ac0453a5750da1092

sled-agent/tests/old-rss-sled-plans/madrid-rss-sled-plan.json
```diff
@@ -1 +1 @@
...
...40:2504:3d2::1","fdb0:a840:2504:355::1","fdb0:a840:2504:396::1","fdb0:a840:2504:157::1"]},"ntp_servers":["ntp.eng.oxide.computer"],"dns_servers":["1.1.1.1","9.9.9.9"],"internal_services_ip_pool_ranges":[{"first":"172.20.28.1","last":"172.20.28.10"}],"external_dns_ips":["172.20.28.1"],"external_dns_zone_name":"madrid.eng.oxide.computer","external_certificates":[{"cert":"<censored>","key":"<censored>"}],"recovery_silo":{"silo_name":"recovery","user_name":"recovery","user_password_hash":"$argon2id$v=19$m=98304,t=13,p=1$RUlWc0ZxaHo0WFdrN0N6ZQ$S8p52j85GPvMhR/ek3GL0el/oProgTwWpHJZ8lsQQoY"},"rack_network_config":{"rack_subnet":"fd00:1122:3344:1::/56","infra_ip_first":"172.20.15.37","infra_ip_last":"172.20.15.38","ports":[{"routes":[{"destination":"0.0.0.0/0","nexthop":"172.20.15.33"}],"addresses":[{"address":"172.20.15.38/29"}],"switch":"switch0","port":"qsfp0","uplink_port_speed":"speed40_g","uplink_port_fec":"none","bgp_peers":[],"autoneg":false},{"routes":[{"destination":"0.0.0.0/0","nexthop":"172.20.15.33"}],"addresses":[{"address":"172.20.15.37/29"}],"switch":"switch1","port":"qsfp0","uplink_port_speed":"speed40_g","uplink_port_fec":"none","bgp_peers":[],"autoneg":false}],"bgp":[]}}}
```
smklein, line 1: Is it safe that we updated this? This file was intended to represent the state of old, in-field deployments. When we update this file, we're losing state about configs that are deployed at customer sites

jgallagher: I think it is not safe, for exactly the reason you defined. We should probably revert this change - if that breaks something else, that's almost certainly meaningful.

Nieuwejaar: I'll take a look.

Nieuwejaar: Yeah, this is busted.  Sorry.  I'll fix it.

https://github.com/oxidecomputer/omicron/issues/5902

PR #5855, opened 2024-06-04
Add `VmmState::SagaUnwound`

hawkw:
This branch is part of ongoing work on the `instance-update` saga (see PR #5749); I've factored it out into a separate PR. This is largely because this branch makes mechanical changes to a bunch of different files that aren't directly related to the core change of #5749, and I'd like to avoid the additional merge conflicts that are likely when these changes remain un-merged for a long time.

---

Depends on #5854 and should merge only after that PR.

As part of #5749, it is desirable to distinguish between *why* a VMM record was marked as `Destroyed`, in order to determine which saga is responsible for cleaning up that VMM's resources. The `instance-update` saga must be the only entity that can set an instance's VMM IDs (active and target) to NULL. Presently, however, the `instance-start` and `instance-migrate` sagas may also do this when they unwind. This is a requirement to avoid situations where a failed `instance-update` saga cannot unwind, because the instance's generation number has changed while the saga was executing.

We want to ensure that if a start or migrate request fails, the instance will appear to be in the correct post-state as soon as the relevant API call returns. In order to do this, without having to also guarantee that an instance update has occurred, we introduce a new VMM state, `SagaUnwound`, with the following rules:

- It is legal to start or migrate an instance if the `active_propolis_id` or `destination_propolis_id` (respectively) is either `NULL` or refers to a VMM that’s in the `SagaUnwound` state (the new VMM ID directly replaces the old ID).
- If an instance has an `active_propolis_id` in the `SagaUnwound` state, then the instance appears to be `Stopped`.
- If an instance has a `destination_propolis_id` in the `SagaUnwound` state, nothing special happens–the instance’s state is still derived from its active VMM’s state.
- The update saga treats `SagaUnwound` VMMs as identical to `Destroyed` VMMs for purposes of deciding whether to remove a VMM ID from an instance.

This branch adds a new `VmmState::SagaUnwound` variant. The `SagaUnwound` state is an internal implementation detail that shouldn't be exposed to an operator or in the external API. Sled-agents will never report that a VMM is in this state. Instead, this state is set my the `instance-start` and `instance-migrate` sagas when they unwind. When determining the API instance state from an instance and active VMM query so that the `SagaUnwound` state is mapped to `Destroyed`.

Closes #5848, which this replaces.

### DIFF
--- a/nexus/db-model/src/schema_versions.rs
+++ b/nexus/db-model/src/schema_versions.rs
@@ -17,7 +17,7 @@ use std::collections::BTreeMap;
 ///
 /// This must be updated when you change the database schema.  Refer to
 /// schema/crdb/README.adoc in the root of this repository for details.
-pub const SCHEMA_VERSION: SemverVersion = SemverVersion::new(70, 0, 0);
+pub const SCHEMA_VERSION: SemverVersion = SemverVersion::new(71, 0, 0);
 
 /// List of all past database schema versions, in *reverse* order
 ///
@@ -29,6 +29,7 @@ static KNOWN_VERSIONS: Lazy<Vec<KnownVersion>> = Lazy::new(|| {
         // |  leaving the first copy as an example for the next person.
         // v
         // KnownVersion::new(next_int, "unique-dirname-with-the-sql-files"),
+        KnownVersion::new(71, "add-saga-unwound-vmm-state"),
         KnownVersion::new(70, "separate-instance-and-vmm-states"),
         KnownVersion::new(69, "expose-stage0"),
         KnownVersion::new(68, "filter-v2p-mapping-by-instance-state"),
--- a/nexus/db-model/src/vmm_state.rs
+++ b/nexus/db-model/src/vmm_state.rs
@@ -24,6 +24,7 @@ impl_enum_type!(
     Migrating => b"migrating"
     Failed => b"failed"
     Destroyed => b"destroyed"
+    SagaUnwound => b"saga_unwound"
 );
 
 impl VmmState {
@@ -37,6 +38,7 @@ impl VmmState {
             VmmState::Migrating => "migrating",
             VmmState::Failed => "failed",
             VmmState::Destroyed => "destroyed",
+            VmmState::SagaUnwound => "saga_unwound",
         }
     }
 }
@@ -58,7 +60,7 @@ impl From<VmmState> for omicron_common::api::internal::nexus::VmmState {
             VmmState::Rebooting => Output::Rebooting,
             VmmState::Migrating => Output::Migrating,
             VmmState::Failed => Output::Failed,
-            VmmState::Destroyed => Output::Destroyed,
+            VmmState::Destroyed | VmmState::SagaUnwound => Output::Destroyed,
         }
     }
 }
@@ -74,7 +76,7 @@ impl From<VmmState> for sled_agent_client::types::VmmState {
             VmmState::Rebooting => Output::Rebooting,
             VmmState::Migrating => Output::Migrating,
             VmmState::Failed => Output::Failed,
-            VmmState::Destroyed => Output::Destroyed,
+            VmmState::Destroyed | VmmState::SagaUnwound => Output::Destroyed,
         }
     }
 }
@@ -108,7 +110,7 @@ impl From<VmmState> for omicron_common::api::external::InstanceState {
             VmmState::Rebooting => Output::Rebooting,
             VmmState::Migrating => Output::Migrating,
             VmmState::Failed => Output::Failed,
-            VmmState::Destroyed => Output::Destroyed,
+            VmmState::Destroyed | VmmState::SagaUnwound => Output::Destroyed,
         }
     }
 }
--- a/nexus/db-model/src/vmm_state.rs
+++ b/nexus/db-model/src/vmm_state.rs
@@ -106,6 +106,11 @@ impl From<VmmState> for omicron_common::api::external::InstanceState {
             VmmState::Starting => Output::Starting,
             VmmState::Running => Output::Running,
             VmmState::Stopping => Output::Stopping,
+            // `SagaUnwound` should map to `Stopped` so that an `instance_view`
+            // API call that produces an instance with an unwound VMM will appear to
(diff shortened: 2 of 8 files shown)


nexus/db-model/src/schema_versions.rs
```diff
@@ -29,6 +29,7 @@ static KNOWN_VERSIONS: Lazy<Vec<KnownVersion>> = Lazy::new(|| {
         // |  leaving the first copy as an example for the next person.
         // v
         // KnownVersion::new(next_int, "unique-dirname-with-the-sql-files"),
+        KnownVersion::new(69, "add-saga-unwound-vmm-state"),
```
gjcolombo, line 32: s/69/70

nexus/src/app/instance.rs
```diff
@@ -1664,7 +1664,7 @@ impl super::Nexus {
                         vmm.runtime.state,
                     )))
                 }
-                DbVmmState::Destroyed => Err(Error::invalid_request(
+                DbVmmState::Destroyed | DbVmmState::SagaUnwound => Err(Error::invalid_request(
                     "cannot connect to serial console of destroyed instance",
```
gjcolombo, line 1668: I think we should tweak this so that the error message says the instance is Stopped in this case. (It's not Destroyed from the user's perspective since they were able to refer to it to ask for its serial console.)

The message here isn't new, of course; I think the difference is that today, when a VMM dies or a start saga gets unwound, the instance's active Propolis ID is cleared before the VMM actually gets destroyed. When the start saga starts using this state, it's expected that an instance will have a non-NULL Propolis ID that points to a VMM in this state, so this match arm will be much more reachable than before.

gjcolombo (commented):
This looks right to me (I'm happy to recheck it once the state enum split merges and the PR is targeting main). Just a couple of small comments.

nexus/src/app/instance.rs
hawkw: yup, makes sense. will update it!

hawkw: hmm, perhaps we should rephrase this to match the wording of the error message in the previous match arm, so that we don't have to different messages that both suggest that the instance is "stopped", but with different wording (which implies to the user that they are Secretly Different...)

nexus/db-model/src/vmm_state.rs
```diff
@@ -108,7 +110,7 @@ impl From<VmmState> for omicron_common::api::external::InstanceState {
             VmmState::Rebooting => Output::Rebooting,
             VmmState::Migrating => Output::Migrating,
             VmmState::Failed => Output::Failed,
-            VmmState::Destroyed => Output::Destroyed,
+            VmmState::Destroyed | VmmState::SagaUnwound => Output::Destroyed,
```
gjcolombo, line 113: SagaUnwound should map to `Stopped` so that an `instance_view` API call that produces an instance with an unwound VMM will appear to be in the correct external state. (Possibly `VmmState::Destroyed` should also map to `Stopped`, but today we don't ever expect an `active_propolis_id` to point to a destroyed VMM.)

The start saga test that covers this case (`test_ensure_running_unwind`) currently passes anyway because `sis_move_to_starting_undo` erases the active Propolis ID before marking the VMM as destroyed. IIRC that will change when the rest of the instance update saga lands and Nexus begins to enforce the rule that only that saga can set `active_propolis_id` to NULL.

nexus/src/app/sagas/instance_common.rs
```diff
@@ -326,7 +329,10 @@ pub async fn instance_ip_get_instance_state(
                 ),
             )));
         }
-        (InstanceState::Vmm, Some(VmmState::Destroyed)) => {
+        (
+            InstanceState::Vmm,
+            Some(VmmState::Destroyed | VmmState::SagaUnwound),
```
gjcolombo, line 334: I think we should treat `VmmState::SagaUnwound` the same as `InstanceState::NoVmm` in this match. They describe the same condition: the instance has no active VMM and isn't even prospectively running anywhere; the difference is just in whether the updater task has garbage-collected the VMM ID left behind by the failed attempt to start.

hawkw: Yup, this makes sense to me. Will change it!

nexus/db-model/src/vmm_state.rs
```diff
@@ -104,7 +106,7 @@ impl From<VmmState> for omicron_common::api::external::InstanceState {
             VmmState::Starting => Output::Starting,
             VmmState::Running => Output::Running,
             VmmState::Stopping => Output::Stopping,
-            VmmState::Stopped => Output::Stopped,
+            VmmState::Stopped | VmmState::SagaUnwound => Output::Stopped,
```
gjcolombo, line 109: nit: I might throw a comment in here too just to point out what this state means in the grand scheme of things

gjcolombo (approved):
LGTM, thanks for the change!

I think I still accept my past self's reason for introducing this state (to recap for other readers: we want to introduce an "instance state update" task; that task requires the exclusive right to clear an instance's Propolis ID fields; unwinding start/migrate sagas therefore need a way to mark that a VMM needs to be GC'ed by the update task). That said, it's sort of a bummer that there are now two states that represent a stopped instance (`InstanceState::NoVmm` and `InstanceState::Vmm` + `VmmState::SagaUnwound`). As we author and review the update task, it might be worth a little extra elbow grease to make absolutely sure we can't avoid the "updater task is the only thing that can set Propolis IDs to NULL" restriction. But I'm confident enough in that requirement that I'm good with merging this PR.

nexus/db-model/src/vmm_state.rs
hawkw: how about this:
```suggestion
            // `SagaUnwound` should map to `Stopped` so that an `instance_view`
            // API call that produces an instance with an unwound VMM will appear to
            // be `Stopped`. This is because we instances with unwound VMMs can
            // be started by a subsequent instance-start saga, just like
            // instances whose internal state actually is `Stopped`.
            VmmState::Stopped | VmmState::SagaUnwound => Output::Stopped,
```

gjcolombo: Except for the extra "we" on line 116, lgtm ship it :)

PR #5856, opened 2024-06-05
Decompose Nexus into multiple types

andrewjstone:
This is a mostly mechanical change that takes nexus methods implemented in `nexus/src/app/*.rs` and creates concrete types for those methods rather than impling them on `Nexus`. This is the first part of an effort to make `Nexus` less of a "god" type. Specifically, this change allows the new types to be passed independently around nexus and into background tasks as needed. Each type is full of mostly immutable methods and wraps other types. All types can be freely cloned with the implementation being a shallow clone with `Arc`s at the bottom. In order to allow the http endpoints to have access to existing functionality and minimize the diff, all the new types are attached to `Nexus` and accessed directly rather than using the prior `Nexus` methods.

There is stil a circular build dependency that is not broken, such that background tasks cannot directly execute and recover sagas. This will be fixed in a follow up PR.


andrewjstone:
This is a draft only because I need to go through and rename all the methods to reduce stuttering. There may be some other cleanup I'm forgetting also, but tests pass and I'd like to get eyes on it.

david-crespo:
net +1700 lines sounds bad for a mechanical change, but it looks like it's all this kind of thing. is that right?

image

andrewjstone:
> net +1700 lines sounds bad for a mechanical change, but it looks like it's all this kind of thing. is that right?
> 
> image

Mostly, yes and putting things into the new types that hang off nexus.

nexus/src/app/mod.rs
```diff
@@ -190,20 +176,62 @@ pub struct Nexus {
...
+    pub switch_interface: switch_interface::SwitchInterface,
+    pub switch_port: switch_port::SwitchPort,
+    pub switch: switch::Switch,
+    pub update: update::Update,
+    pub utilization: utilization::Utilization,
+    pub volume: volume::Volume,
+    pub vpc_router: vpc_router::VpcRouter,
+    pub vpc_subnet: vpc_subnet::VpcSubnet,
+    pub vpc: vpc::Vpc,
+    pub saga_context: saga_interface::SagaContext,
```
smklein, line 234: While I'm on-board with refactoring portions of code to only use dependencies that are actually necessary, this ends up carrying some structural / lifetime costs that seem non-ideal to me. Basically: For each module of code we add, we now require additional Arc clones, we increase the size of this structure indefinitely, etc.

Is that actually necessary? Is it important to make each one of these objects that have a lifetime tied to Nexus?

Let's say an operation arrives that wants to access something from the `SwitchInterface` submodule. Would it be possible to create a new `SwitchInterface` struct, containing *references* to the fields borrowed from `Nexus`, rather than cloning them? (E.g., grab a `&Arc<Datastore>`, rather than cloning?)

I think this is slightly closer to the approach @jgallagher took when creating https://github.com/oxidecomputer/omicron/tree/main/nexus/networking -- basically, pull out the functionality from Nexus, make it accessible to Nexus, but without creating a new object that "owns" these fields indefinitely. Even though free functions were used in that module, it seems possible to create on-the-fly "structs containing references to borrowed data" -- that seems effectively the same, as the lifetime of the borrowed fields would still be "in the big Nexus object".

andrewjstone: > While I'm on-board with refactoring portions of code to only use dependencies that are actually necessary, this ends up carrying some structural / lifetime costs that seem non-ideal to me. Basically: For each module of code we add, we now require additional Arc clones, we increase the size of this structure indefinitely, etc.
> 

I view this all as an intermediate state to get to where I want to go, which is figuring out which of these types can actually be merged into one type or more functional types, and then broken out into their own crates. I don't want to keep adding new types here, but the only way to make progress towards not having Nexus itself be required everywhere but the HTTP handlers was do this all in one shot. I definitely tried not doing it, as this was a lot of tedious work.

> Is that actually necessary? Is it important to make each one of these objects that have a lifetime tied to Nexus?
> 
> Let's say an operation arrives that wants to access something from the `SwitchInterface` submodule. Would it be possible to create a new `SwitchInterface` struct, containing _references_ to the fields borrowed from `Nexus`, rather than cloning them? (E.g., grab a `&Arc<Datastore>`, rather than cloning?)
> 

This may be possible to do in some cases, but I don't see how it works with async tokio tasks. Those all need static lifetimes on references as they can move across threads. This is especially problematic for background tasks.

> I think this is slightly closer to the approach @jgallagher took when creating https://github.com/oxidecomputer/omicron/tree/main/nexus/networking -- basically, pull out the functionality from Nexus, make it accessible to Nexus, but without creating a new object that "owns" these fields indefinitely. Even though free functions were used in that module, it seems possible to create on-the-fly "structs containing references to borrowed data" -- that seems effectively the same, as the lifetime of the borrowed fields would still be "in the big Nexus object".

I'm not particularly a fan of the free-function pattern, and based on lifetime issues, I don't see how we'd do what you are suggesting unfortunately.

davepacheco:
Thanks for taking a swing at this.  I'm still working through the big questions here but I thought it might be useful to write this down and share it anyway.

First, I'm wrestling with the meta question.  I gather the stack here is something like:

1. When a Nexus zone is expunged, we need to assign its sagas to another Nexus.
2. That would be happening through one of the Reconfigurator background tasks.
3. That requires that a background task be able to trigger saga recovery.
4. That kicked over an issue around circular initialization of `Nexus` and `BackgroundTasks`.
5. One approach to solving this is: divide Nexus functionality into smaller subsystems, then hand the "saga" subsystem over to background tasks, which avoids those having to know about `Nexus`
6. That led to this PR and the much broader question of: how do we organize the large body of code that makes up Nexus's top level?

Is this PR the place to discuss other approaches for solving (4) (the circular initialization issue), or is this really about (6) (how do we better organize Nexus), independent of solving the specific problem here?

---

tl;dr: As I understand it, this PR takes a swing at the approach of: divide Nexus functionality into top-level subsystems.  I've generally liked this idea, and there are specific things I like here, too.  But overall I don't like this incremental step and it makes me doubt whether the subsystem approach itself can work.  But it's hard to tell, and that's what led me to the meta question above.  The question of "how to better organize the top level of Nexus" seems really broad, likely to generate a bunch of different views, and it seems likely to me that we won't have a clear way to choose among the opinions.  That's not a reason not to discuss it but then how do we avoid churning on the question (or, worse, the code) indefinitely?

On this PR: I **like** about this that a lot of code now accepts a much smaller type that only has access to the few things it needs.  This makes it much easier to see the intrinsic relationships between things.  And it should stay up-to-date since Rust will tell us if some field is unused.

But there's some stuff I don't like here.  A lot of it boils down to "these types don't really feel like coherent subsystems":

* They don't abstract or encapsulate anything.  e.g., it's not really an `InstanceManager` -- it's just a collection of functions that happen to work on instances.
* They're not even close to independent -- they remain quite interdependent.  It's nice that we can see those interdependencies, but for me, they don't simplify the mental model at all the way I want when I create subsystems.
* Many of the types have the same fields as each other -- several _just_ have a `DataStore`.  I'm not sure why this bothers me -- it wouldn't bother me if they _were_ `InstanceManager` and `DeploymentManager` or whatever, but for some reason the fact so many are mostly the same highlights that they're more of a namespace than an abstraction.
* As an example, "project" and "instance" and "disk" are under different types because they were in different files before.  But they're closely related and I assume they were separated before only because _anything_ is separable when it's just a question of what file it's in and we wanted to keep files small.  But these areas are much more closely related than they are to, say, metrics or deployment.  That's hidden in the structure.  (This was already true.  But before, the structure was just "the file the functions are defined in".  Putting these into separate types suggests to me encapsulation or abstraction that's not really what's going on here.)
* It sometimes seems unclear where a thing belongs.  I was surprised glancing at [nexus/src/app/sagas/disk_delete.rs](https://github.com/oxidecomputer/omicron/pull/5856/files#diff-4fa3605fa6742d4cfdc09250948610c40c85b8fb0cc3b65c58976a1e8d4ceae9) that it's `nexus.project_lookup` (not `nexus.project.lookup()`) but later it's `nexus.disk.project_create_disk()`.
* There's a lot of boilerplate / repetition.  See [this block](https://github.com/oxidecomputer/omicron/pull/5856/files#diff-a25f83014afb84b325de4f6f4ff6b4c8a1bfccfbdf533ce0691a410a67b33b1bR198-R234).
* In that same block: it feels like there's too many things there to expect someone to know where to put a new hunk of code.

This is smaller and probably easy to fix, but I got really confused for a while when I saw that an `Instance` holds a `Disk` and that `Certificate` holds a `Silo`.  It took me a good few minutes (thinking I'd completely misunderstood this) to realize that these are holding references to what is basically a module, not a thing representing a specific disk or silo.  Given that we already have authz types, view types, param types, and db model types -- all with the same name -- it might be nice to not overload the same names again.  Maybe this is easy to fix by referring to things as `app::Disk` everywhere?

Aside from that, the bullet points above are all sort of saying the same thing, which is complaining about the specific ways these things have been grouped.  I get that maybe that's just because this is a first step towards the "coherent subsystems with top-level types" approach, and we're starting by at least separating what can be easily separated.  But if that's the approach, I wonder if a better incremental step would be: instead of creating types that are just namespaces (not abstractions) and then figuring out better abstractions later, how about creating fuller abstractions, one at a time, for areas as we want/need them.  Concretely, if the goal is to do be able to do saga recovery without a full Nexus, maybe just do what you did with saga.rs (only) to make just that subsystem now?  Arguably, this is already what we're doing.  It's not like Nexus is all one big package today -- we've already created abstractions in subpackages like `nexus-db-queries` (`DataStore`), `nexus-auth` (`OpContext`), and others.

I said above this actually makes me doubt the whole subsystem-oriented approach altogether, and what I mean is: if these areas are forever going to have deep interdependencies, then I think it's confusing to create separations between them.  But it might just be a question of how we slice it up.  If the subsystems are Projects, Instances, Networking, etc., well those _are_ all related.  But maybe that means the orthogonal subsystems should be something more like: DataStore, IAM, Sagas, Low-level networking (by which I mean the inward-facing RPWs), "Cloud Infra" (for lack of a better term; this would include all of: instances, disks, user-facing networking, etc.).

Anyway, sorry for the ramble.  As I said I don't feel like I've got a clear idea of what we should do here but I wanted to at least share what I'm thinking about.  I'm very curious what others feel about all this!

andrewjstone:
> Is this PR the place to discuss other approaches for solving (4) (the circular initialization issue), or is this really about (6) (how do we better organize Nexus), independent of solving the specific problem here?

I think that both questions are valid, and what I want to discuss today. I think likely, I'll deal with the circular initialization issue first so that I can implement saga migration. As it turns out, even with this change, it didn't work as I expected.

I agree with all your other comments also @davepacheco. The types are totally intertwined here, and the more I look at this the more I dislike it. Letting it sit for a few weeks was not a bad thing! I strongly recommend not merging this PR, but I do hope this spawns a conversation about how we want to move forward with Nexus. That may also mean leaving things mostly the way they are.

andrewjstone:
We just had a team meeting where we decided to focus first on fixing the circular dependency between background tasks and nexus. We proposed to do this by creating the activation handles for the background tasks independently from the tasks themselves, such that the handles can then be passed to things that need to activate tasks without those tasks existing yet.

This strategy may extend to saga execution and recovery as well, but doesn't directly solve the Nexus contains all functionality and takes a long time to build issue.

PR #5859, opened 2024-06-05
add `migration` table and explicit migration tracking in sled-agent

### LINKED ISSUE #2948: want mechanism to track & report on live migration statuses
See discussion here: https://github.com/oxidecomputer/omicron/pull/2892#discussion_r1180683094

hawkw:
As part of ongoing work on improving instance lifecycle management (see #5749) , we intend to remove the `InstanceRuntimeState` tracking from `sled-agent`, and make Nexus the sole owner of `instance` records in CRDB, with a new `instance-update` saga taking over the responsibility of managing the instance's state transitions. 

In order to properly manage the instance state machine, Nexus will need information about the status of active migrations that are currently only available to sled-agents. For example, if an instance is migrating, and a sled agent reports that the source VMM is `Destroyed`, Nexus doesn't presently have the capability to determine whether the source VMM was destroyed because the migration completed successfully, or that the source shut down prior to starting the migration, resulting in a failure. 

In order for Nexus to correctly manage state updates during live migration, we introduce a new `migration` table to the schema for tracking the state of ongoing migrations. The `instance-migrate` saga creates a `migration` record when beginning a migration. The Nexus and sled-agent APIs are extended to include migration state updates from sled-agents to Nexus. In *this* branch, the `migration` table is (basically) write-only; Nexus doesn't really read from it, and just stuffs updates into it. In the future, however, this will be used by the `instance-update` saga.

It occurred to me that, in addition to using the migration table for instance updates, it might also be useful to add an OMDB command to look up the status of a migration using this table. However, I decided that made more sense as a follow-up change, as I'd like to get back to integrating this into #5749.

Fixes #2948


schema/crdb/dbinit.sql
```diff
@@ -4063,6 +4063,20 @@ VALUES (
...
+  'completed'
+);
+
+-- A table of the states of current migrations.
+CREATE TABLE IF NOT EXISTS omicron.public.migration (
+    id UUID PRIMARY KEY,
+    source_state omicron.public.migration_state NOT NULL,
+    source_propolis_id UUID NOT NULL,
+    target_state omicron.public.migration_state NOT NULL,
+    target_propolis_id UUID NOT NULL
```
gjcolombo, line 4078: We mostly covered this offline, but for posterity: I think we'll also want generation numbers and update times, and they need to be recorded separately for the source and target.

My mental model is that migration state is adjacent to VMM state and will be pushed to (or pulled by) Nexus alongside VMM state updates. That is:

1. on a migration phase change, Propolis emits new migration state to any callers of its instance watcher endpoint
2. sled agent observes these changes and reflects them up to Nexus via the same "vmm state changed" endpoint it uses for "major" state changes
3. in response, Nexus writes the new vmm/migration states to CRDB
4. Nexus decides whether the change it recorded was momentous enough to require an instance update saga

It's only safe to do this if we can guarantee that Nexus receives and processes migration state changes in order. I don't think that's the case today because of the instance watcher task, but even if that task weren't there, I think it's better to give sled agent and Nexus some flexibility in how these messages are sent and processed (e.g. we might someday want to make the "write back to the DB" part of the sled agent upcall asynchronous, which would allow sled agent to come back, read some even newer state from Propolis, and send that up to Nexus before the prior state was saved).

The source- and target-reported states are managed by independent sleds that won't be able to see each other's generation numbers, so they should be separate in the DB as well (as should their update times).

We'll have a little bit of work to do to merge states and update times if/when we expose migration status via API, but I don't think it will be anything intractable.

hawkw: Yup, I agree; working on that now. I'm not totally convinced we need both update times _and_ generation counts if the generation is owned uniquely by the sled-agent, but I suppose we may as well add the timestamps as well.

gjcolombo: If we don't have separate timestamps, I think things can get out of whack if there's a delay between the moment the timestamp is captured for some record we're about to write and the moment Nexus writes it back:

t=1: Nexus captures time for a write that says "source = in_progress"
t=2: Nexus captures time for a write that says "target = done"
t=3: Nexus writes "source = in_progress, updated_time = 1"
<time passes>
t=100: Nexus prepares to write "source = done"
t=101: Nexus writes "source = done, updated_time = 100"
t=102: Nexus writes "target = done, updated_time = 2"

I'm not sure if using CRDB's `now()` function as a time source would help with this. There's still the possibility of a delay between the time CRDB gets the timestamp and the time it commits it, but it's possible that the conflicting writes would cause a transaction retry (giving CRDB a chance to refresh the timestamp).

Probably simpler, I think, just to have two timestamps (with the added advantage that we'll be able to see exactly how long one or the other side of a migration appears to have been out to lunch).

hawkw: I totally agree that we may as well record timestamps, especially for diagnostic purposes, and that if we're going to have them, we need to have two of them. What I meant was that, AFAIK, the generation numbers are sufficient to ensure *correctness* here (such as in the case you described). Am I overlooking something?

gjcolombo: Oh, my mistake, I misread you. Yeah, the generation numbers are really all you need for this, at least insofar as "older states shouldn't overtake newer ones" is concerned.

common/src/api/internal/nexus.rs
```diff
@@ -108,6 +109,80 @@ pub struct SledInstanceState {
...
+/// role of the VMM whose migration state was updated.
+#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema)]
+pub struct MigrationRuntimeState {
+    pub migration_id: Uuid,
+    pub state: MigrationState,
+    pub role: MigrationRole,
+    pub gen: Generation,
+
+    /// Timestamp for the migration state update.
+    // TODO(eliza): could this just be the VMM state timestamp?
```
hawkw, line 128: I think I ought to remove this comment. If we intend to use these timestamps for tracking the duration of migrations, we probably want to separate them from the "time state updated" timestamp.
```suggestion
```

common/src/api/internal/nexus.rs
```diff
@@ -108,6 +109,80 @@ pub struct SledInstanceState {
...
+    pub time_updated: DateTime<Utc>,
+}
+
+/// The state of an instance's live migration.
+#[derive(
+    Clone, Copy, Debug, PartialEq, Eq, Deserialize, Serialize, JsonSchema,
+)]
+#[serde(rename_all = "snake_case")]
+pub enum MigrationState {
```
hawkw, line 137: N.B. that currently we don't have a default state for "nexus has created the migration record in the DB but the sled-agent hasn't actually reported in that the migration is in progress. I'm not *entirely* sure whether we need that or not, but we might...

hawkw: Upon further consideration, I think I'm just going to go ahead and add this now, to avoid having to do another DB migration later.

schema/crdb/dbinit.sql
```diff
@@ -4064,6 +4064,55 @@ VALUES (
 ON CONFLICT (id)
 DO NOTHING;
 
+CREATE TYPE IF NOT EXISTS omicron.public.migration_state AS ENUM (
+  'pending',
+  'in_progress',
```
smklein, line 4069: Out of curiosity, what's the distinction between these two? Is it when the source stops running, that the sled agent reports a new state and we transition to `in_progress`?

nexus/db-model/src/migration.rs
```diff
@@ -0,0 +1,80 @@
...
+    pub source_propolis_id: Uuid,
+
+    /// The generation number for the source state.
+    pub source_gen: Generation,
+
+    /// The time the source VMM state was most recently updated.
+    pub time_source_updated: Option<DateTime<Utc>>,
+
+    /// The state of the migration target VMM.
+    pub target_state: MigrationState,
```
smklein, line 48: I'm not opposed to starting with the "state" being the same at source and destination, but I wonder if we'll eventually want different types here. Seems possible that there's a source-only or target-only state in the future.

(No change needed, just musing)

nexus/db-queries/src/db/datastore/instance.rs
```diff
@@ -385,12 +386,14 @@ impl DataStore {
         new_instance: &InstanceRuntimeState,
         vmm_id: &Uuid,
         new_vmm: &VmmRuntimeState,
-    ) -> Result<(bool, bool), Error> {
+        migration: &Option<MigrationRuntimeState>,
+    ) -> Result<(bool, bool, Option<bool>), Error> {
```
smklein, line 390: This changes makes the `Return value` comment out-of-date -- could we use a named type here instead?

nexus/db-queries/src/db/queries/instance.rs
```diff
@@ -165,24 +183,90 @@ impl InstanceAndVmmUpdate {
...
+                        diesel::update(migration_dsl::migration)
+                            .filter(migration_dsl::id.eq(migration_id))
+                            .filter(
+                                migration_dsl::source_propolis_id.eq(vmm_id),
+                            )
+                            .filter(migration_dsl::source_gen.lt(gen))
+                            .set((
+                                migration_dsl::source_state.eq(state),
+                                migration_dsl::time_target_updated
+                                    .eq(time_updated),
```
smklein, line 226: This should be `time_source_updated`, right?

nexus/src/app/instance.rs
```diff
@@ -563,9 +563,12 @@ impl super::Nexus {
         // outright fails, this operation fails. If the operation nominally
         // succeeds but nothing was updated, this action is outdated and the
         // caller should not proceed with migration.
-        let (updated, _) = match instance_put_result {
+        let (updated, _, _) = match instance_put_result {
             Ok(state) => {
                 self.write_returned_instance_state(&instance_id, state).await?
+
+                // TODO(eliza): this is where we would also want to insert to the
```
smklein, line 570: Is it a problem we aren't doing so in this PR?

(I'm okay splitting things out, just want to make sure we don't cause regressions by leaving this as a comment)

smklein (commented):
Looks solid, great idea to split out the source/destination columns so explicitly. Already seems easier to follow, and I think the omdb commands are a great idea.

nexus/src/app/instance.rs
smklein: ^ this is basically my only blocking comment, if we're okay with this, the whole PR LGTM

gjcolombo, line 570: Is this comment accurate? This looks like it's already happening in the migration saga proper (which is where I'd expect it to happen).

nexus/src/app/sagas/instance_migrate.rs
```diff
@@ -127,6 +132,7 @@ impl NexusSaga for SagaInstanceMigrate {
 
         builder.append(reserve_resources_action());
         builder.append(allocate_propolis_ip_action());
+        builder.append(create_migration_record_action());
```
gjcolombo, line 135: I would consider doing this step and the next in the other order--insert the destination VMM and then the migration record so that it's always safe to interpret the target ID in the migration as an FK into the VMM table.

sled-agent/src/sim/instance.rs
```diff
@@ -360,6 +367,24 @@ impl SimInstanceInner {
...
+            let role = self
+            .state
+            .migration()
+            .expect(
+                "we just got a `put_migration_ids` request with `Some` IDs, \
+                so we should have a migration"
+            )
+            .role;
+            if role == MigrationRole::Source {
+                self.queue_successful_migration(MigrationRole::Target)
```
gjcolombo, line 384: Should this be `Source`?

hawkw:
Thanks @gjcolombo and @smklein for the very thorough reviews, I really appreciate it! 🖤

schema/crdb/dbinit.sql
hawkw: `Pending` is the initial state when the migration record is inserted into the database, *before* a sled-agent has confirmed that the migration has actually started; the state advances to `InProgress` only after a sled-agent reports back that its Propolis has actually started the migration. Initially, I didn't differentiate between these two states (see https://github.com/oxidecomputer/omicron/pull/5859/commits/631ba0f18edd9f1ad83e8d64073bcc243b9bac3f), but I thought it might be nice to be able to capture a difference between cases where one side of the migration fails before the actual migration has started, and cases where one side fails *during* the migration.

Right now, none of the code that will use this in the `instance-update` saga actually needs to distinguish this. Mostly, I thought it would be useful for debugging tools like a potential future `omdb db migration` command, but I figured we may as well be recording it.

nexus/db-queries/src/db/datastore/migration.rs
```diff
@@ -0,0 +1,58 @@
...
+            .on_conflict(dsl::id)
+            .do_update()
+            .set(dsl::time_created.eq(dsl::time_created))
+            .returning(Migration::as_returning())
+            .get_result_async(&*self.pool_connection_authorized(opctx).await?)
+            .await
+            .map_err(|e| public_error_from_diesel(e, ErrorHandler::Server))
+    }
+
+    pub async fn migration_mark_deleted(
```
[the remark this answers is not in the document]
hawkw: Good question! You're absolutely right that currently, migration records don't actually get marked as deleted when the migration completes successfully. This is because...I haven't actually figured out when they can safely be deleted --- since I haven't written the part of the update saga that will consume information from the migration table yet, I'm not totally sure whether migration records can be marked as deleted by the migrate saga as soon as the migration completes, or whether they will need to stick around for a subsequent update saga, and the update saga will have to be responsible for deleting them. I probably should have mentioned that in the PR description.

I didn't want  to add code to the migration saga  to mark migration records as deleted that will then get removed in my update saga branch, so I just didn't make it delete them here. Now that I think through it, though, it occurs to me that this might not be a good idea: if we cut a release that includes this change, but *not* the update saga, we might end up shipping code where we create a bunch of migration records that never get deleted and can't be easily pruned from the DB. I think it's probably better to make a completed migration saga remove the migration record _for now_, and revisit it later when writing the update saga.

sled-agent/src/common/instance.rs
```diff
@@ -256,56 +279,82 @@ impl InstanceStates {
...
-
-                    assert_eq!(self.propolis_role(), PropolisRole::Retired);
+            ObservedMigrationStatus::Succeeded => {
+                if let Some(ref mut migration) = self.migration {
+                    migration.state = MigrationState::Completed;
+                    migration.gen = migration.gen.next();
+                    migration.time_updated = observed.time;
+                } else {
+                    // XXX(eliza): we don't have an active migration state, but
+                    // we saw a migration succeed. this is almost certainly a
```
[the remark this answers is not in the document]
hawkw: Ah, yeah, I think you're right that we shouldn't be able to reach this case as a
result of Propolis misbehaving. It should probably be a debug assert (or perhaps
we should just `expect` it to be `Some`, honestly...)

nexus/src/app/instance.rs
hawkw: yeah, this is out of date, my bad --- `write_returned_instance_state` now calls
into `instance_and_vmm_update_runtime`, which does write to the `migration` table.

hawkw: I explained this in response to one of Greg's comments: the comment was just stale, I've removed it: https://github.com/oxidecomputer/omicron/pull/5859#discussion_r1633845748

sled-agent/src/common/instance.rs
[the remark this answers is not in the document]
hawkw: Huh..turns out we were wrong about this, at least in the simulated sled-agent: https://buildomat.eng.oxide.computer/wg/0/details/01J025F2Z61DTCC4HXS6MFWZXB/a2v6jXzafrXSXZxlVhUaz7pMhGxO8yGnPcmi9sRlIIWYqTay/01J025FBZT929NBEJKMF0JFY5J#S4656

That seems like it could be a test bug though, will investigate further.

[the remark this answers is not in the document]
hawkw: Oh, yeah, that's a test bug, these are unit tests that never call into the code that would create the `MigrationState`

nexus/db-model/src/migration.rs
hawkw: Right now, the intention for this table is to store pretty low-resolution data about the migration state, basically just whether it's completed or failed, so we're not including more detailed migration status reported by Propolis. So, we don't *really* expect these to include states that are specific to one side of the migration currently, although I suppose we *could*...

nexus/db-queries/src/db/datastore/migration.rs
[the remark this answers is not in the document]
hawkw: Actually, looking closer, I think it's `notify_instance_updated` that would be responsible for marking these records as deleted, rather than the `instance-migrate` saga; since that saga just *starts* the migration.

sled-agent/src/common/instance.rs
```diff
@@ -256,58 +279,90 @@ impl InstanceStates {
...
+                if let Some(ref mut migration) = self.migration {
+                    migration.state = MigrationState::InProgress;
+                    migration.gen = migration.gen.next();
+                    migration.time_updated = observed.time;
+                }
+            }
             ObservedMigrationStatus::NoMigration
-            | ObservedMigrationStatus::InProgress
             | ObservedMigrationStatus::Pending => {}
         }
```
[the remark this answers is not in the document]
hawkw: Hmm, my thinking was that the target observing a failure here would be sufficient to handle this case, but your point that the target sled could *also* have crashed and might not see the source's failure... I'll see about having this fail the migration.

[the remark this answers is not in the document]
hawkw: Okay, as of f6c98753e4020ed285310b65fde203c44e1fbec5 we should report failure immediately if a VMM fails/is destroyed while it's part of an in-progress migration.

sled-agent/src/common/instance.rs
```diff
@@ -321,22 +332,16 @@ impl InstanceStates {
 
                     // This is a migration source that previously reported success
                     // and removed itself from the active Propolis position. Don't
-                    // touch the instance.
+                    // touch the instance.'
```
gjcolombo, line 335: turbo nit: extra `'`

nexus/db-queries/src/db/datastore/migration.rs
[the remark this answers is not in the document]
gjcolombo: Thanks! I think this behavior looks good for now. We may want to change it later, e.g. to support an API that lets an operator migrate an instance and poll to determine when migration has finished, but I don't think we know enough about the necessity or shape of that API to try to design for it just yet. I mention it here just so we can keep it in the backs of our heads.

sled-agent/src/common/instance.rs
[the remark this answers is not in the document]
gjcolombo: Looks good. I suspect that in the long run we're probably going to need Nexus to deal with this in a health check task: if Nexus discovers that a VMM is unexpectedly gone, then the VMM's state changes *and* any migrations that it was a part of fail. But we'll address that when we add those checks.

gjcolombo (approved):
Thanks for going another lap on this!

sled-agent/src/common/instance.rs
hawkw:
```suggestion
                    // touch the instance.
```

PR #5860, opened 2024-06-05
honor `XTASK_BIN` in `tools/install_*_prerequisites.sh`

iliana:
Merging #5482 introduced breakage I didn't account for: being able to run `install_runner_prerequisites.sh` without having Cargo available, which is used by the a4x2 testbed. This change introduces the `XTASK_BIN` environment variable which will be used instead of `cargo xtask` if set.

The companion PR in a4x2 is https://github.com/oxidecomputer/testbed/pull/49/files.

Tested that all these work:
- `./tools/install_builder_prerequisites.sh -y`
- `XTASK_BIN=target/debug/xtask ./tools/install_builder_prerequisites.sh -y` (after building xtask)
- `XTASK_BIN=target/debug/nonexistent-file ./tools/install_builder_prerequisites.sh -y` (correctly fails with ENOENT)

### DIFF
--- a/tools/install_builder_prerequisites.sh
+++ b/tools/install_builder_prerequisites.sh
@@ -85,6 +85,15 @@ function retry
   exit $retry_rc
 }
 
+function xtask
+{
+  if [ -z ${XTASK_BIN+x} ]; then
+    cargo xtask "$@"
+  else
+    "$XTASK_BIN" "$@"
+  fi
+}
+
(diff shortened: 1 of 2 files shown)


iliana:
(I also checked that there aren't any other shell scripts in the repo that call `cargo xtask`.)

jgallagher (approved):
Thank you!

PR #5862, opened 2024-06-05
add tool for experimenting with argon2 hash parameters

davepacheco:
I had this sitting around from last time we were tweaking the argon2 hash parameters.  I figured it was worth having something in the repo for this.  It's quite simple but without this you have to recompile every time you want to try a different set of parameters.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5642,7 +5642,9 @@ dependencies = [
 name = "omicron-passwords"
 version = "0.1.0"
 dependencies = [
+ "anyhow",
  "argon2",
+ "clap",
  "criterion",
  "omicron-workspace-hack",
  "rand 0.8.5",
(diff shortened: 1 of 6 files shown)


sunshowers:
Thanks! Could this be made an external xtask command for greater discoverability as well?

davepacheco:
What would that mean, exactly?

sunshowers:
> What would that mean, exactly?

Basically you'd be able to run cargo xtask argon2, which would run this binary. See e.g. https://github.com/oxidecomputer/omicron/blob/e48fd96e68088630c62ef33fa46df632234dffd5/dev-tools/xtask/src/main.rs#L75.

davepacheco:
Thanks for the specifics.  I thought you were asking to rework the implementation as an `xtask` subcommand, not just adding an xtask subcommand that invokes this program.  I took a swing in ee3262a5c1fb6f81b74b50c2eef3ef2ab3785e32.

sunshowers (approved):
Awesome, thanks!

PR #5863, opened 2024-06-06
[wicket] Support capital letters for some commands

### LINKED ISSUE #5796: wicket ignition control shows 'I', but only responds to 'i'
While using ignition to try to cycle things, it took me a bit to get to the right keypress. It was shown as `Ignition <I> `, which led me to think I had to use a capital I to use it. However, only a lowercase 'i' would get there based on the current contents of `keymap.rs`. That took me a while to get. It'd be nice if either:

* The capital `I` actually worked for ignition like shown
* The message was changed so a lowercase `i` was shown

andrewjstone:
We exclude vim commands.

Fixes #5796

### DIFF
--- a/wicket/src/keymap.rs
+++ b/wicket/src/keymap.rs
@@ -182,36 +182,36 @@ impl KeyHandler {
         if let Some(seq) = self.seq {
             match seq {
                 MultiKeySeqStart::g => match event.code {
-                    KeyCode::Char('g') => {
+                    KeyCode::Char('g') | KeyCode::Char('G') => {
                         self.seq = None;
                         return Some(Cmd::GotoTop);
                     }
-                    KeyCode::Char('e') => {
+                    KeyCode::Char('e') | KeyCode::Char('E') => {
                         self.seq = None;
                         return Some(Cmd::GotoBottom);
                     }
                     _ => (),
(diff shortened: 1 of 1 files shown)


wicket/src/keymap.rs
```diff
@@ -230,10 +230,10 @@ impl KeyHandler {
...
             KeyCode::Esc => Cmd::Exit,
             KeyCode::Char(' ') => Cmd::Toggle,
-            KeyCode::Char('e') => Cmd::Expand,
-            KeyCode::Char('c') => Cmd::Collapse,
-            KeyCode::Char('d') => Cmd::Details,
-            KeyCode::Char('i') => Cmd::Ignition,
+            KeyCode::Char('e') | KeyCode::Char('E') => Cmd::Expand,
+            KeyCode::Char('c') | KeyCode::Char('C') => Cmd::Collapse,
+            KeyCode::Char('d') | KeyCode::Char('D') => Cmd::Details,
+            KeyCode::Char('i') | KeyCode::Char('I') => Cmd::Ignition,
```
citrus-it, line 236: Oh yes, thank you!

sunshowers (approved):
Awesome!

PR #5866, opened 2024-06-06
[nexus] Put a `SagaContext` in a once cell

andrewjstone:
This allows the ability to run sagas from background tasks. The `SagaContext` already has a copy of an `Arc<BackgroundTasks>` which means that individual background tasks must have been created/ initialized before the `SagaContext`. We get around this build time circular dependency by putting the `SagaContext` inside a static `OnceCell` where the background tasks can access it at runtime.

Builds upon #5856

### DIFF
--- a/nexus/src/app/mod.rs
+++ b/nexus/src/app/mod.rs
@@ -669,6 +669,12 @@ impl Nexus {
             Arc::clone(&background_tasks),
         );
 
+        // Initialize a global `SagaContext`. This provides access to
+        // background tasks.
+        saga_interface::SAGA_CONTEXT
+            .set(saga_context.clone())
+            .expect("SAGA_CONTEXT already initialized");
+
         let nexus = Nexus {
             id: config.deployment.id,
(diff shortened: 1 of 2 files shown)


davepacheco:
Using a global this seems like it would break anything that tried to create multiple Nexus instances in the same process, wouldn't it?  I'm a little surprised we don't already have something that does that in the test suite.  Could put the cell in a more specific place that needs it (e.g., stored in a background task's struct or something)?

smklein:
> Using a global this seems like it would break anything that tried to create multiple Nexus instances in the same process, wouldn't it? I'm a little surprised we don't already have something that does that in the test suite. Could put the cell in a more specific place that needs it (e.g., stored in a background task's struct or something)?

I think the reason we're getting away with this is because nextest launches each test in a different process. It does seem like a test that tries to create multiple nexuses would struggle though.

andrewjstone:
This is no longer relevant.

Issue #4192, opened 2023-10-03
Revisit decision to use "bb8" as a connection pool?


smklein:
AFAICT, the decision to use bb8 in Omicron comes from https://github.com/oxidecomputer/omicron/pull/57 , where it was added with the comment: 

> Service discovery and connection pooling: I used bb8 here because it was easy. Based on my survey, I think we'll probably want to build our own cueball-like library.

We may want to revisit this decision, weighing the pros and cons of bb8, other connection pools, and building our own.

PR #5876, opened 2024-06-10
Qorb integration as connection pool for database

smklein:
Replaces all usage of bb8 with a new connection pooling library called [qorb](https://github.com/oxidecomputer/qorb).

qorb, detailed in RFD 477, provides the following benefits over bb8:
- It allows lookup of multiple backends via DNS SRV records
- It dynamically adjusts the number of connections to each bakend based on their health, and prioritizes vending out connections to healthy backends
- It should be re-usable for both our database and progenitor clients (using a different "backend connector", but the same core library and DNS resolution mechanism).

Fixes https://github.com/oxidecomputer/omicron/issues/4192
Part of https://github.com/oxidecomputer/omicron/issues/3763 (fixes CRDB portion)


nexus-config/src/postgres_config.rs
```diff
@@ -32,6 +33,29 @@ impl PostgresConfigWithUrl {
...
+    ///
+    /// # Safety
+    ///
+    /// This method makes the assumption that the hostname has at least one
+    /// "host IP / port" pair which can be extracted. If the supplied URL
+    /// does not have such a pair, this function will panic.
+    // Yes, panicking in the above scenario sucks. But this type is already
+    // pretty ubiquitous within Omicron, and integration with the qorb
+    // connection pooling library requires access to database by SocketAddr.
+    pub fn address(&self) -> SocketAddr {
```
smklein, line 47: I don't love that this method makes this assumption, though this appears to be embedded mostly within tests.

It's necessary within `nexus/db-queries/src/db/pool.rs` to implement `SingleHostResolver`, which is our way of identifying backends as `SocketAddr` types.

I can try to make the "address" of the backend generic through qorb, so we could leave it as a URL?

nexus/db-queries/src/db/collection_attach.rs
```diff
@@ -859,7 +857,7 @@ mod test {
             dev::test_setup_log("test_attach_missing_collection_fails");
         let mut db = test_setup_database(&logctx.log).await;
         let cfg = db::Config { url: db.pg_config().clone() };
-        let pool = db::Pool::new(&logctx.log, &cfg);
+        let pool = db::Pool::new_qorb_single_host(&cfg);
```
smklein, line 860: There are two flavors of `db::Pool` that are exposed now:

1. One that uses DNS to resolve backends (this will be used in prod, and recommended), and
2. One that uses a hard-coded address for resolution and skips DNS (this is used by a lot of our tests)

I'm using the function name `new_qorb_single_host` to identify case (2) explicitly.

nexus/src/context.rs
```diff
@@ -210,7 +207,7 @@ impl ServerContext {
         // nexus in dev for everyone
 
         // Set up DNS Client
-        let resolver = match config.deployment.internal_dns {
+        let (resolver, dns_addrs) = match config.deployment.internal_dns {
```
smklein, line 210: We need to pull out the DNS addresses here to help bootstrap the qorb pool below.

We will eventually be able to dynamically discover DNS servers with qorb, but we need to supply a value of https://github.com/oxidecomputer/qorb/blob/bb432c9574c393bf2eb7b19516b996aa5584573e/src/resolvers/dns.rs#L387-L390 first

nexus/db-queries/src/db/pool.rs
```diff
@@ -3,108 +3,143 @@
...
-    }
+fn make_dns_resolver(
+    bootstrap_dns: Vec<SocketAddr>,
+) -> qorb::resolver::BoxedResolver {
+    Box::new(DnsResolver::new(
+        service::Name(internal_dns::ServiceName::Cockroach.srv_name()),
+        bootstrap_dns,
+        DnsResolverConfig {
+            query_interval: tokio::time::Duration::from_secs(10),
+            hardcoded_ttl: Some(std::time::Duration::from_secs(60)),
```
smklein, line 63: This `hardcoded_ttl` config is load-bearing. The default is `None`, and the `DnsResolver` uses a TTL from the DNS records themselves to identify the the records as present or not. **However**, all our DNS records are being saved with a TTL of "zero", so they are immediately discarded.

nexus-config/src/postgres_config.rs
```diff
@@ -32,6 +33,29 @@ impl PostgresConfigWithUrl {
     pub fn url(&self) -> String {
         self.url_raw.clone()
     }
+
+    /// Accesses the first ip / port pair within the URL.
+    ///
+    /// # Safety
```
hawkw, line 39: nit, take it or leave it: I would, generally, avoid the use of "Safety" in comments that don't specifically deal with _memory_ safety. personally, i would expect to see this documentation under a heading like
```suggestion
    /// # Panics
```

nexus/db-queries/src/db/pool.rs
```diff
@@ -3,108 +3,144 @@
...
 }
 
-impl Pool {
-    pub fn new(log: &slog::Logger, db_config: &DbConfig) -> Self {
-        // Make sure diesel-dtrace's USDT probes are enabled.
-        usdt::register_probes().expect("Failed to register USDT DTrace probes");
-        Self::new_builder(log, db_config, bb8::Builder::new())
-    }
+// Provides an alternative to the DNS resolver for cases where we want to
+// contact the pool directly.
```
hawkw, line 34: "contact the pool directly" doesn't make a ton of sense to me in this context --- I'm assuming this is for the case where we use a single DB address, such as in tests?

nexus/db-queries/src/db/pool.rs
```diff
@@ -3,108 +3,144 @@
...
-                super::pool_connection::ConnectionCustomizer::new(),
-            ))
-            .error_sink(Box::new(error_sink))
-            .build_unchecked(manager);
-        Pool { pool }
+impl Resolver for SingleHostResolver {
+    fn monitor(&mut self) -> watch::Receiver<AllBackends> {
+        self.tx.subscribe()
     }
+}
```
hawkw, line 54: not a blocker or anything, but: it occurs to me that `qorb` could provide an `impl Resolver for watch::Receiver<AllBackends>` that just clones the rx, which would kind of obviate the need for this type..

nexus/db-queries/src/db/pool.rs
```diff
@@ -3,108 +3,144 @@
...
 }
 
-impl bb8::ErrorSink<ConnectionError> for LoggingErrorSink {
-    fn sink(&self, error: ConnectionError) {
-        error!(
-            &self.log,
-            "database connection error";
-            "error_message" => #%error
-        );
+impl Pool {
```
hawkw, line 88: nit, not important: personally, i'm not super convinced that these all need `_qorb_` in their names; that seems kinda like an implementation detail. as a user of this module, I feel like i mostly just care about the behavior of the pool, not which library is used to provide it.

nexus/db-queries/src/db/pool.rs
```diff
@@ -3,108 +3,144 @@
...
+        usdt::register_probes().expect("Failed to register USDT DTrace probes");
+
+        let resolver = make_single_host_resolver(db_config);
+        let connector = make_postgres_connector();
+
+        let policy = Policy {
+            claim_timeout: tokio::time::Duration::from_millis(1),
+            ..Default::default()
+        };
+        Pool { inner: qorb::pool::Pool::new(resolver, connector, policy) }
```
hawkw, line 137: perhaps these should be gated behind a `test_util` feature or something, which can be enabled by dev-dependencies on this crate?

nexus/db-queries/src/db/pool_connection.rs
```diff
@@ -4,46 +4,91 @@
...
 pub type DbConnection = DTraceConnection<PgConnection>;
 
 pub const DISALLOW_FULL_TABLE_SCAN_SQL: &str =
     "set disallow_full_table_scans = on; set large_full_scan_rows = 0;";
 
-/// A customizer for all new connections made to CockroachDB, from Diesel.
-#[derive(Debug)]
-pub(crate) struct ConnectionCustomizer {}
+/// A [backend::Connector] which provides access to [PgConnection].
+pub struct DieselPgConnector {
```
hawkw, line 22: does this need to be public? or is it internal to `db-queries`?

nexus/db-queries/src/db/pool_connection.rs
```diff
@@ -4,46 +4,91 @@
...
+    /// swaps out the IP address at runtime depending on the selected backend.
+    ///
+    /// Format of the url is:
+    ///
+    /// - postgresql://{user}@{address}/{db}
+    ///
+    /// Or, if arguments are supplied:
+    ///
+    /// - postgresql://{user}@{address}/{db}?{args}
+    pub fn new(user: &str, db: &str, args: Option<&str>) -> Self {
```
hawkw, line 38: nitpicky: a function taking multiple positional arguments of the same type with different semantics is always kind of a red flag to me...it seems easy to accidentally swap the ordering of the username and db. if this is going to be a public API, i might consider a struct which passes these arguments as named fields, or newtypes or something. on the other hand, if this is internal to the `db-queries` crate... :woman_shrugging:

hawkw (commented):
Overall, this looks really nice! I left a few minor suggestions --- I hope they're helpful!

nexus/db-queries/src/db/pool.rs
smklein: Yeah, this was incorrect phrasing -- it's for cases where we want to contact the **database** directly without performing resolution first, when there's a single DB as you mentioned.

smklein: (I'll update the comment)

nexus-config/src/postgres_config.rs
smklein: Sure thing, will update

nexus/db-queries/src/db/pool.rs
smklein: Hrmmm I suppose so, but the choice of resolver still needs to be caller-controlled. I think we could probably adjust the API of `Pool::new` to take a `watch::Receiver<AllBackends>` instead of a Boxed `Resolver`, but they feel pretty similar functionally, right?

nexus/db-queries/src/db/pool.rs
smklein: Fair enough, I think it was way more relevant during implementation when I had a combination of bb8 and qorb (briefly) but that is no longer the case. I'll update these names.

nexus/db-queries/src/db/pool.rs
smklein: Good idea -- this is a little harder to do on `new_single_host`, because of the current clients and nexus configs, but I can definitely do this for the failfast one at least

nexus/db-queries/src/db/pool_connection.rs
smklein: I can make it crate-local! Done

nexus/db-queries/src/db/pool_connection.rs
smklein: Updated to use a struct explicitly (even though it is, as you suggested, crate-internal)

nexus/db-queries/src/db/pool.rs
hawkw: :woman_shrugging: it's up to you, i don't feel terribly strongly about this!

smklein: I think it was the right call! Updated 'em

nexus/db-queries/src/db/pool.rs
```diff
@@ -3,108 +3,153 @@
...
+
+        let resolver = make_single_host_resolver(db_config);
+        let connector = make_postgres_connector(log);
+
+        let policy = Policy::default();
+        Pool { inner: qorb::pool::Pool::new(resolver, connector, policy) }
+    }
+
+    /// Creates a new qorb-backed connection pool which returns an error
+    /// if claims are not quickly available.
```
hawkw, line 128: nit, take it or leave it: maybe instead of "quickly" we should just say the actual duration?

nexus/db-queries/src/db/pool.rs
```diff
@@ -3,108 +3,153 @@
...
+    ///
+    /// This is intended for test-only usage.
```
hawkw, line 130: it might be nice if this comment explained when test code should use `new_single_host` and when it should use this constructor?

hawkw (approved):
Overall, this looks good to me! Up to you whether it's worth waiting for a review from someone else, as well.

nexus/db-queries/src/db/pool_connection.rs
```diff
@@ -4,46 +4,128 @@
...
-        conn: &mut Connection<DbConnection>,
-    ) -> Result<(), ConnectionError> {
-        conn.batch_execute_async(DISALLOW_FULL_TABLE_SCAN_SQL).await?;
-        Ok(())
+    fn to_url(&self, address: std::net::SocketAddr) -> String {
+        format!(
+            "{prefix}{address}{suffix}",
+            prefix = self.prefix,
+            suffix = self.suffix,
+        )
```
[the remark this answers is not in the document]
smklein: - https://crates.io/crates/uri looks deprecated, so probably not that?
- https://docs.rs/url/2.5.2/url/ looks popular , and like it could be used for this purpose, though it doesn't have exactly that API you described
- https://docs.rs/http/latest/http/uri/struct.Builder.html also looks like it's a builder for URLs

I'm gonna go with the `url` crate, and use it primarily for the query parameters.

nexus/db-queries/src/db/pool.rs
smklein: Updated (it's one millisecond)

nexus/db-queries/src/db/pool.rs
smklein: Sure thing, added.

nexus/db-queries/src/db/pool_connection.rs
```diff
@@ -47,20 +49,29 @@ impl DieselPgConnector {
...
-            suffix = self.suffix,
-        )
+    fn to_url(
+        &self,
+        address: std::net::SocketAddr,
+    ) -> Result<String, anyhow::Error> {
+        let user = &self.user;
+        let db = &self.db;
+        let mut url =
+            Url::parse(&format!("postgresql://{user}@{address}/{db}"))?;
```
hawkw, line 68: i really thought there was a builder-style interface for constructing URLs from parts like this, but looking at the `url` docs, i think it must be the _other_ Rust URL library that has that...

smklein:
This is still pretty much ready to go, but I'm interested in getting https://github.com/oxidecomputer/omicron/pull/5912 resolved first so we don't fragment our DNS client usage -- especially considering that we've been bumping into mysterious timeouts with hickory DNS that still need resolution.

davepacheco:
#5912 has now been resolved and the hickory-dns timeouts understood so I think this can move forward.

PR #5880, opened 2024-06-11
Randomize read-only parent UUIDs during migration

jmpesp:
During an instance migration, the migration destination is given a new volume checkout for a disk during the migration saga. Checkout increases the generation number in the sub volume portion of the Volume but not the read-only parent portion. This means that the migration destination can activate and take over the read/write ownership of the sub volumes, as it should.

However, the read-only side of the hierarchy does not have a generation number bump. This is ok, it doesn't need one: there's no exclusive access that needs to be maintained for the read-only data, and many different Upstairs can connect to a read-only downstairs without issues.

But for read-only connections that share an upstairs ID, the read-only Crucible Downstairs will kick out which ever connected first, and subsequent connections will kick out previous ones. This is a problem in the migration case, because the read-only parents will be the same for the source and the destination instance's disks - volume IDs are _not_ randomized during checkout.

This is only problem during migration - for other cases, the read-only parent's UUIDs are properly randomized. This commits adds a randomization step during `instance_ensure_registered` for the migration case only: the source and destination read-only Upstairs should be uniquely identified.

Also, this commit truncates the long strings that were preventing `cargo fmt` from formatting instance.rs, and bundles those changes in.

### DIFF
--- a/nexus/src/app/instance.rs
+++ b/nexus/src/app/instance.rs
@@ -26,6 +26,7 @@ use nexus_db_queries::authz;
 use nexus_db_queries::context::OpContext;
 use nexus_db_queries::db;
 use nexus_db_queries::db::datastore::InstanceAndActiveVmm;
+use nexus_db_queries::db::datastore::VolumeCheckoutReason;
 use nexus_db_queries::db::identity::Resource;
 use nexus_db_queries::db::lookup;
 use nexus_db_queries::db::lookup::LookupPath;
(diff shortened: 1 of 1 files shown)


jmpesp:
There's a bug here that just occurred to me: if the migration destination needs a region or snapshot replacement, the UUIDs that were used in `Volume::construct` will be randomized from what is in the db: `compare_vcr_for_replace` will fail due to these different UUIDs.

jmpesp:
After talking with @gjcolombo I'm closing this. I was unnecessarily worried!

The intention of this PR was to prevent the case where the a propolis has activated a Volume, then is used as a migration source, and somehow kicks out the migration destination's read-only connections. Importantly, activating Volumes would be the mechanism of this kicking out, and today this only occurs when a migration succeeds enough that the destination propolis will take over. In this case, we do want the destination to take over the read-only connections too.

Even in the case that there's some sort of roll-back that could occur after the destination performs this activation, like:

- activate on source
- run most of migration protocol
- target activates volumes
- target says "whoops never mind"
- source reactivates and resumes

The source would need a activate using a new checkout to get the increased generation number for the R/W part of the Volume anyway, and would then take over from the failed migration destination's activation.

PR #5881, opened 2024-06-11
(likely) fix #5865 test flake

### LINKED ISSUE #5865: Test flake in `test_unauthorized` due to concurrent discovery of timeseries schema
We hit a test flake [here](https://buildomat.eng.oxide.computer/wg/0/details/01HZQBGQ3NNZAT8MTZKX9G2VKG/PIgxSEsnPmEQQ8LLsNudpE5iel4VTZfORvIRj10mfjwlY7qT/01HZQBHMDQ8N73F8QH823CK3M7#S4454), in which the test that all endpoints are authorized detected a change between two successive `GET` calls to the `/v1/timeseries/schema` endpoint. We've not seen that before, but seems plausible that this can happen. The timeseries schema are derived from the data that `oximeter` collects. That means we see a schema in the output of this endpoint after the producer has started; registered with Nexus; one collection interval has elapsed; and `oximeter` has successfully derived and inserted the timeseries schema in ClickHouse. Each of these steps has some jitter in them as well, so there are lots of opportunities for successive get requests to return different data.

One approach suggested by @davepacheco is to add a variant `AllowedMethod::GetVolatile`. That would mark the endpoint as possibly returning different data each time. The authorization checks on the endpoint could still be done, but we would elide the check for consistent return values.

iliana:
Ran across https://github.com/oxidecomputer/omicron/issues/5865 while looking at a test flake; this implements the suggested fix.

### DIFF
--- a/nexus/tests/integration_tests/endpoints.rs
+++ b/nexus/tests/integration_tests/endpoints.rs
@@ -979,6 +979,12 @@ pub enum AllowedMethod {
     /// always fail in the correct way.
     #[allow(dead_code)]
     GetUnimplemented,
+    /// HTTP "GET" method, but where the response data may change for reasons
+    /// other than successful user interaction.  This should be uncommon; in
+    /// most cases resources do not change merely due to the passage of time,
+    /// although one common case is when the response data is updated by a
+    /// background task.
+    GetVolatile,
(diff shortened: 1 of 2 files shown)


bnaecker (approved):
Nice, thanks!

PR #5882, opened 2024-06-12
Support updating the RoT bootloader (stage0)

labbott:


wicketd/src/update_tracker.rs
```diff
@@ -946,6 +977,78 @@ impl UpdateDriver {
                 },
             )
             .register();
+
+        // Send the bootloader update to the RoT.
+        // XXX is this where we mutex?
```
lzrd, line 982: Wicket is all manually driven, right? By that I mean that each stage0 update needs to be triggered by a human.
In any case, we don't want more than one stage0 update in process at any one time and we want manual intervention if any fail. If we loose a device because of a bricked RoT, we don't want to proceed.
This isn't a real issue until the chain of trust and trust quorum are fully built out, but bricked RoTs are going to at least require field service to touch the machines.

wicketd/src/update_tracker.rs
```diff
@@ -1844,6 +1950,76 @@ impl UpdateContext {
...
+                        format!(
+                            "{message} (failed to parse RoT bootloader version: {err})"
+                        ),
+                    )
+                    .into(),
+                }
+            }
+            None => StepSuccess::new(make_result(None))
+                .with_message(
+                    "Stage0 has no caboose, proceeding with update anyway",
```
lzrd, line 2017: There are few enough stage0 versions released that we could include an FWID to known version map here until we've purged them from existence.

wicketd/src/update_tracker.rs
```diff
@@ -2434,13 +2673,132 @@ impl<'a> SpComponentUpdateContext<'a> {
...
+                        },
+                    )
+                    .register();
+
+                registrar
+                    .new_step(
+                        SpComponentUpdateStepId::Resetting,
+                        "Waiting for RoT to boot".to_string(),
+                        move |_cx| async move {
+                            const WAIT_FOR_BOOT_TIMEOUT: Duration =
```
lzrd, line 2702: There are multiple WAIT_FOR_BOOT_TIMEOUT declarations. Can they be combined?

wicketd/src/update_tracker.rs
jgallagher: > Wicket is all manually driven, right? By that I mean that each stage0 update needs to be triggered by a human.

Wicket is only manually driven to a point. A human kicks off "do all the updates for sled N", but then wicket drives RoT -> SP -> host updates autonomously. Multiple sleds can be updated in parallel.

wicketd/src/update_tracker.rs
labbott: I can think of 7 off the top of my head (staging/dev and prod/rel for each of gimlet psc rot plus bart). Anything that isn't one of those should probably just be updated manually via faux-mgs

wicketd/src/update_tracker.rs
```diff
@@ -1588,6 +1705,43 @@ struct RotInterrogation {
...
+                SpComponent::SP_ITSELF.const_as_str(),
+                0,
+            )
+            .await
+            .ok()
+            .map(|v| v.into_inner());
+
+        // Older versions of the SP have a bug that prevents setting
+        // the active slot for the RoT bootloader. Check for these
+        // and skip the update until the SP gets updated
```
jgallagher, line 1725: Just double-checking: we only care about SP version and not RoT version for this?

wicketd/src/update_tracker.rs
```diff
@@ -1844,6 +2001,151 @@ impl UpdateContext {
...
+            "ee688a237a480e9fd111a7f70cc4c6f9ac837dcac65a01e7cfa29f7c28545d07",
+            "f31442015da37523a13ffaa173b4dfe0b069c6d890cf1c9748a898001fe4110e",
+            "fa73f26fb73b27b5db8f425320e206df5ebf3e137475d40be76b540ea8bd2af9",
+        ];
+
+        // We already validated at repo-upload time there is at least one RoT
+        // artifact available and that all available RoT artifacts are the same
+        // version, so we can unwrap the first artifact here and assume its
+        // version matches any subsequent artifacts.
+        // TODO this needs to be fixed for multi version to work!
```
jgallagher, line 2034: I assume this TODO is for a subsequent PR? Will that be similar to the changes to allow multi-version non-bootloader RoT images?

wicketd/src/update_tracker.rs
```diff
@@ -1844,6 +2001,151 @@ impl UpdateContext {
...
+                SpComponent::ROT.const_as_str(),
+                &GetRotBootInfoParams {
+                    version:
+                        gateway_messages::RotBootInfo::HIGHEST_KNOWN_VERSION,
+                },
+            )
+            .await
+        {
+            Ok(v) => match v.into_inner() {
+                // the minimum we will ever return is 3
```
jgallagher, line 2056: What guarantees this?

wicketd/src/update_tracker.rs
```diff
@@ -2434,13 +2799,130 @@ impl<'a> SpComponentUpdateContext<'a> {
...
+                        move |_cx| async move {
+                            update_cx
+                                .set_component_active_slot(
+                                    component_name,
+                                    firmware_slot,
+                                    true,
+                                )
+                                .await
+                                .map_err(|error| {
+                                    SpComponentUpdateTerminalError::SetRotActiveSlotFailed {
```
jgallagher, line 2862: Tiny nit - should this be a new `SetRotBootloaderActiveSlotFailed` variant to get a more precise error message?

wicketd/src/update_tracker.rs
```diff
@@ -2518,6 +3000,7 @@ impl<'a> SpComponentUpdateContext<'a> {
             }
             UpdateComponent::Sp => {
                 // Nothing special to do on the SP - just reset it.
+                // TODO fixup the SP to also set the active slot
```
jgallagher, line 3003: Is this for oxidecomputer/hubris#1808? Do we need to put this fix in before R9 and/or were you planning to do it on this PR?

jgallagher (commented):
Changes LGTM, just a handful of questions

wicketd/src/update_tracker.rs
```diff
@@ -2434,13 +2799,130 @@ impl<'a> SpComponentUpdateContext<'a> {
...
+                                    SpComponentUpdateTerminalError::SetRotActiveSlotFailed {
+                                        error,
+                                    }
+                                })?;
+                            StepSuccess::new(()).into()
+                        },
+                    )
+                    .register();
+
+                // Now reset (again) to boot into the new stage0
```
jgallagher, line 2871: Do we need to mutex this substep (or one of the other ones) (https://github.com/oxidecomputer/omicron/pull/5882#discussion_r1636913858)?

wicketd/src/update_tracker.rs
labbott: Correct. If the RoT version supports `rot-boot-info` it should support stage0 update.

wicketd/src/update_tracker.rs
labbott: Yes, those changes should be very similar.

wicketd/src/update_tracker.rs
labbott: we request the `HIGHEST_KNOWN_VERSION` which is 3. If this message return success we should always return `V3` (see https://github.com/oxidecomputer/hubris/blob/master/drv/lpc55-update-server/src/main.rs#L349, https://github.com/oxidecomputer/management-gateway-service/blob/main/gateway-messages/src/sp_to_mgs.rs#L616) . If we return an error that gets handled below and we should skip the update.

wicketd/src/update_tracker.rs
labbott: I couldn't figure out how to place the mutex. We want to mutex at the rack level and it seemed like this gets spawned off per SP. This is probably my complete inexperience with Rust synchronization though.

wicketd/src/update_tracker.rs
labbott: Yes, this is related. It's not necessary for this PR to work and I was wary about adding more logic changes for backwards compatability since this PR is already pretty large.

wicketd/src/update_tracker.rs
jgallagher: It's going to be a moderate-to-large mess no matter what, I think. How bad is landing this without it?

wicketd/src/update_tracker.rs
labbott: the `rot-boot-info` is a separate message that gets sent to the RoT. If the SP doesn't know about the message it will return an error that it can't be deserialized and we will skip the bootloder update
```
 │                  │ Message: Failed to run `rot_boot_info`: Error Response: status: 503 Service Unavailable; headers: {"content-type": "application/json",                │                   │ 
 │                  │ "x-request-id": "8afc7433-17e1-40c8-ac22-edb184b79063", "content-length": "242", "date": "Sun, 28 Dec 1986 00:08:31 GMT"}; value: Error { error_code: │                   │ 
 │                  │ Some("SpCommunicationFailed"), message: "error communicating with SP SpIdentifier { typ: Sled, slot: 15 }: Error response from SP: sprot: failed to   │                   │ 
 │                  │ deserialize message", request_id: "8afc7433-17e1-40c8-ac22-edb184b79063" }. Will not proceed with update.                                             │                   │ 
 │                  │                                                                                                                     
 ``` 
If the SP _does_ know about the message it will send it on to the RoT

Comment (jgallagher) [reply wicketd/src/update_tracker.rs:2056]: Ahh thanks!

File: wicketd/src/update_tracker.rs (hunk 9 above)
Comment (labbott) [reply wicketd/src/update_tracker.rs:2871]: For this particular set of stage0 updates it's pretty low risk. We're going to have some subsequent stage0 updates to add epochs to avoid roll backs so I think we could target it there.

**Comment by labbott** [comment]
I just did a test on london going from release 8 to this repo. What's going to happen is the first time through we're going to get an error because the SP is too old
```
 │                  │ Message: Failed to run `rot_boot_info`: Error Response: status: 503 Service Unavailable; headers: {"content-type": "application/json",                │                   │ 
 │                  │ "x-request-id": "8afc7433-17e1-40c8-ac22-edb184b79063", "content-length": "242", "date": "Sun, 28 Dec 1986 00:08:31 GMT"}; value: Error { error_code: │                   │ 
 │                  │ Some("SpCommunicationFailed"), message: "error communicating with SP SpIdentifier { typ: Sled, slot: 15 }: Error response from SP: sprot: failed to   │                   │ 
 │                  │ deserialize message", request_id: "8afc7433-17e1-40c8-ac22-edb184b79063" }. Will not proceed with update.                                             │                   │ 
 │                  │                                                                                                                     
 ```
 
 And then we're going to have to run the update again which should give a warning such as `Message: fwid c23a53858e94932a95945f28730e41ae4a2d1a8db4776283245eda143b6b2994 is known to be missing a caboose.` before we finally update stage0.
 
 The only thing I have not been able to test is running the update on a prod/rel signed sled. There is one in dogfood so once this merges I'm going to plan to do an off cycle dogfood update to make sure this works.

wicketd/src/update_tracker.rs
labbott: I'll make that change

jgallagher (approved):

PR #5884, opened 2024-06-12
Fix underflow in wicket in a cross-screen interaction

### LINKED ISSUE #5877: Wicket TUI panic
Managed to hit a panic:

```
thread 'main' panicked at wicket/src/state/rack.rs:110:55:                                                                                                                       
             attempt to subtract with overflow
```

which corresponds to
```
    pub fn left_or_right(&mut self) {
        match self.selected {
            ComponentId::Sled(i) => {
                if self.left_column {
                    self.left_column = false;
                    self.selected = ComponentId::Sled(i + 1);
                } else { 
                    self.left_column = true;
                    self.selected = ComponentId::Sled(i - 1);
                }
            }
            _ => (),
        }
    }
```

unclear what I did but I feel like I hit this before and couldn't reproduce it

jgallagher:
Closes #5877 (see https://github.com/oxidecomputer/omicron/issues/5877#issuecomment-2163349194 for the reproduction steps)

### DIFF
--- a/wicket/src/state/rack.rs
+++ b/wicket/src/state/rack.rs
@@ -102,13 +102,8 @@ impl RackState {
     pub fn left_or_right(&mut self) {
         match self.selected {
             ComponentId::Sled(i) => {
-                if self.left_column {
-                    self.left_column = false;
-                    self.selected = ComponentId::Sled(i + 1);
-                } else {
-                    self.left_column = true;
-                    self.selected = ComponentId::Sled(i - 1);
-                }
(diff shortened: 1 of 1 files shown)


andrewjstone (approved):
Nice fix! It took me a second to recall what this was supposed to do!

jgallagher:
Yeah, I started by just deleting `left_column` entirely, but then discovered its main goal is to remember which column we _were_ in if you navigate onto and then away from the dual-width rows (switch/psc). So this was the easiest fix, I think.

PR #5888, opened 2024-06-12
[nexus] Add `DataStore::instance_fetch_all`

hawkw:
In order to implement the `instance-update` saga ongoing in #5749, it's necessary to have a way to query the database for the state of an instance, its active VMM, its target VMM, and active migration in a single, atomic query. This will be used as a snapshot of the instance state that the update saga will operate on.

This commit adds an implementation of such a query in the form of a `DataStore::instance_fetch_all` method, returning an `InstanceSnapshot` struct that bundles together all these records. We do this by `LEFT JOIN`ing the `instance` table with the `vmm` and `migration` table on the `instance`'s various ID columns, with a query that should look something like:

```sql
SELECT * FROM instance
LEFT JOIN vmm
    ON ((instance.active_propolis_id = vmm.id)
    OR (instance.target_propolis_id = vmm.id))
LEFT JOIN migration ON (instance.migration_id = migration.id)
WHERE instance.id = instance_id
```

Ideally, we would instead do two subqueries that join `instance` with `vmm` on the active and destination IDs separately and return a single row which contains both in a way that can be distinguished in the result set, but I wasn't able to figure out how to do that in Diesel. Instead, the query may return two rows, if the instance has both an active and target VMM, and iterate over the result set to find both VMMs in Rust code. This is a bit uglier, but it works fine.

### DIFF
--- a/nexus/db-queries/src/db/datastore/instance.rs
+++ b/nexus/db-queries/src/db/datastore/instance.rs
@@ -1253,7 +1253,7 @@ mod tests {
     async fn test_instance_fetch_all() {
         // Setup
         let logctx = dev::test_setup_log("test_instance_fetch_all");
-        let db = test_setup_database(&logctx.log).await;
+        let mut db = test_setup_database(&logctx.log).await;
         let (opctx, datastore) = datastore_test(&logctx, &db).await;
         let authz_instance = create_test_instance(&datastore, &opctx).await;
         let snapshot =
--- a/nexus/db-queries/src/db/datastore/instance.rs
+++ b/nexus/db-queries/src/db/datastore/instance.rs
@@ -1409,5 +1409,9 @@ mod tests {
             Some(dbg!(migration.id)),
             "fetched migration must be the instance's migration"
         );
+
+        // Clean up.
+        db.cleanup().await.unwrap();
+        logctx.cleanup_successful();
     }
 }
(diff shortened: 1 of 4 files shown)


nexus/db-queries/src/db/datastore/instance.rs
```diff
@@ -330,6 +351,123 @@ impl DataStore {
...
+        // Because we join with the `vmm` table on VMMs whose ID is the active
+        // *or* target VMM ID in the instance record, this query may return a
+        // result set with multiple rows. If it does, we have to iterate over
+        // them and determine which VMM result is the active VMM and which is the
+        // target. The instance and migration records should be identical, so we
+        // only need the first ones.
+        //
+        // Yes, this code is a bit unfortunate, but Diesel doesn't like the idea
+        // of joining with the same table twice on different columns...so, this,
+        // at least, works.
```
gjcolombo, line 425: Yeah, this is probably the best way of doing this while coloring inside the lines Diesel has set out. (I could imagine having some kind of CTE that conjures up a temporary table that contains the VMM IDs and a "sequence" column that can be used to order the returned rows unambiguously, but I've found it very hard to do this kind of thing without dropping out to raw SQL.)

gjcolombo (approved):

nexus/db-queries/src/db/datastore/instance.rs
hawkw: That's sort of what I thought. Writing the CTE felt like it would be enough of a pain that I didn't really think it would be worth the effort, but we can revisit this later.

PR #5889, opened 2024-06-12
Define timeseries schema in TOML

bnaecker:
This is the first commit in a thread of work to support updates to timeseries schema. In our first step, we're moving the definition of timeseries from the Rust structs we use today, to TOML text files. Each file describes one target and all the metrics associated with it, from which we generate exactly the same Rust code that it replaces.

This opens the door to a lot of downstream work, including well-defined updates to schema; detection of conflicting schema; improved metadata for timeseries; and generation of documentation. As an example, this moves exactly one (1) timeseries into the new format, the physical datalink statistics tracked through illumos kstats.

- Move the `oximeter` crate into a private implementation crate, and re-export it at its original path
- Add intermediate representation of timeseries schema, and code for deserializing TOML timeseries definitions, checking them for conflicts, and handling updates. Note that **no actual updates** are currently supported. The ingesting code includes a check that version numbers are exactly 1, except in tests. We include tests that check updates in a number of ways, but until the remainder of the work is done, we limit timeseries in the wild to their current version of 1.
- Add a macro for consuming timeseries definitions in TOML, and generating the equivalent Rust code. Developers should use this new `oximeter::use_timeseries!()` proc macro to create new timeseries.
- Add an integration test in Nexus that pulls in timeseries schema ingested with with `oximeter::use_timeseries!()`, and checks them all for conflicts. As developers add new timeseries in TOML format, they must be added here as well to ensure we detect problems at CI time.
- Updates `kstat-rs` dep to simplify platform-specific code. This is part of this commit because we need the definitions of the physical-data-link timeseries for our integration test, but those were previously only compiled on illumos platforms. They're now included everywhere, while the tests remain illumos-only.


bnaecker:
Note to self: make issues for sequence of work, and refer to them in this PR instead of `TODO(ben)`.

oximeter/impl/src/schema/ir.rs
```diff
@@ -0,0 +1,1288 @@
...
+
+    fn fields(&self) -> Option<&BTreeSet<String>> {
+        match self {
+            CurrentVersion::Active { fields, .. } => Some(fields),
+            CurrentVersion::Inactive { .. } => None,
+        }
+    }
+}
+
+fn defn_error<T>(s: String) -> Result<T, MetricsError> {
```
jgallagher, line 397: Tiny nit - maybe I'm Stockholm-syndromed into long errors, but I'd probably not have written this function at all, and inlined all the error returns. If you really like this, it could maybe take `s: Into<String>` to save even more call site boilerplate (in the cases where we pass a static string and not a `format!()`'d one).

oximeter/impl/src/schema/ir.rs
```diff
@@ -0,0 +1,1288 @@
...
+    Err(MetricsError::SchemaDefinition(s))
+}
+
+/// Load the list of timeseries schema from a schema definition in TOML format.
+pub fn load_schema(
+    contents: &str,
+) -> Result<Vec<TimeseriesSchema>, MetricsError> {
+    toml::from_str::<TimeseriesDefinition>(contents)
+        .map_err(|e| MetricsError::Toml(e.to_string()))
```
jgallagher, line 406: Tiny nit - can we preserve the full error type here, instead of squashing it into `e.to_string()`? If it's important to keep it a string, maybe consider using `InlineErrorChain::new(&e).to_string()` so we get any upstream errors included?

oximeter/impl/src/schema/ir.rs
```diff
@@ -0,0 +1,1288 @@
...
+                "Metric '{}' has '{}' authorization scope, and so must \
+                contain a field '{}' of UUID type",
+                metric_name, scope, key,
+            ))
+        }
+    };
+    match authz_scope {
+        AuthzScope::Silo => check_for_key("silo")?,
+        AuthzScope::Project => check_for_key("project")?,
+        _ => {}
```
jgallagher, line 460: Nit - could we replace `_` with an explicit list of variants, so new variants added in the future have to be considered?

oximeter/impl/src/schema/mod.rs
```diff
@@ -153,18 +159,58 @@ fn validate_timeseries_name(s: &str) -> Result<&str, MetricsError> {
     }
 }
 
+/// Text descriptions for the target and metric of a timeseries.
+#[derive(Clone, Debug, Default, Deserialize, JsonSchema, Serialize)]
+pub struct TimeseriesDescription {
+    pub target: String,
+    pub metric: String,
```
jgallagher, line 166: These fields being `pub` means a client can create `TimeseriesDescription`s that aren't valid (e.g., a target containing `:`). Is that okay? (Same question for the derived `Deserialize` impl, which will allow any strings)

oximeter/impl/src/schema/mod.rs
```diff
@@ -153,18 +159,58 @@ fn validate_timeseries_name(s: &str) -> Result<&str, MetricsError> {
...
+// TODO-completeness: Decide whether and how to handle dimensional analysis
+// during queries, if needed.
+pub enum Units {
+    Count,
+    Bytes,
+}
+
+impl Units {
+    /// Construct a `Units::Count`.
+    pub const fn count() -> Self {
```
jgallagher, line 182: I'm not sure this method is worth it - `Units::count()` is two characters longer than just `Units::Count`

oximeter/impl/src/schema/mod.rs
```diff
@@ -153,18 +159,58 @@ fn validate_timeseries_name(s: &str) -> Result<&str, MetricsError> {
     }
 }
 
+/// Text descriptions for the target and metric of a timeseries.
+#[derive(Clone, Debug, Default, Deserialize, JsonSchema, Serialize)]
```
jgallagher, line 163: Does `Default` make sense here? That would just be an empty description, right?

oximeter/impl/src/schema/mod.rs
```diff
@@ -153,18 +159,58 @@ fn validate_timeseries_name(s: &str) -> Result<&str, MetricsError> {
...
 ///
 /// This includes the name of the timeseries, as well as the datum type of its metric and the
 /// schema for each field.
 #[derive(Clone, Debug, Deserialize, Serialize, JsonSchema)]
 pub struct TimeseriesSchema {
     pub timeseries_name: TimeseriesName,
+    // TODO-cleanup: This default should be removed once schema are tracked in
+    // CockroachDB. Same for the version and scope fields.
+    #[serde(default)]
+    pub description: TimeseriesDescription,
```
jgallagher, line 197: How is this related to `timeseries_name`? If it's expected to always match exactly, could we drop it entirely and just use `timeseries_name` as needed?

oximeter/impl/src/schema/mod.rs
```diff
@@ -153,18 +159,58 @@ fn validate_timeseries_name(s: &str) -> Result<&str, MetricsError> {
...
     pub field_schema: BTreeSet<FieldSchema>,
     pub datum_type: DatumType,
+    #[serde(default = "default_schema_version")]
+    pub version: NonZeroU8,
+    #[serde(default = "AuthzScope::fleet")]
+    pub authz_scope: AuthzScope,
+    #[serde(default = "Units::count")]
+    pub units: Units,
```
jgallagher, line 205: Why do these fields need `#[serde(default = ..)]` tags? The defaults don't seem super intuitive to me, but maybe that's just my unfamiliarity.

oximeter/impl/src/schema/mod.rs
```diff
@@ -263,6 +330,45 @@ impl PartialEq for TimeseriesSchema {
...
+#[serde(rename_all = "snake_case")]
+pub enum AuthzScope {
+    /// Timeseries data is limited to fleet readers.
+    Fleet,
+    /// Timeseries data is limited to the authorized silo for a user.
+    Silo,
+    /// Timeseries data is limited to the authorized projects for a user.
+    Project,
+    /// The timeseries is viewable to all without limitation.
+    None,
```
jgallagher, line 362: Tiny nit - do you think a more descriptive name would be useful here? E.g. based on the comment, maybe `ViewableToAll`?

oximeter/impl/src/schema/mod.rs
```diff
@@ -263,6 +330,45 @@ impl PartialEq for TimeseriesSchema {
...
+}
+
+impl AuthzScope {
+    /// Construct a fleet authz scope.
+    pub const fn fleet() -> Self {
```
jgallagher, line 367: Similar nit as above - not sure this method is worth it

oximeter/instruments/schema/physical-data-link.toml
```diff
@@ -0,0 +1,90 @@
+[target]
+name = "physical_data_link"
+description = "A physical network link on a compute sled"
+authz_scope = "fleet"
+versions = [
+    { version = 1, fields = [ "rack_id", "sled_id", "hostname", "serial", "link_name" ] },
+    #{ version = 2, fields = [ "rack_id", "sled_id", "serial", "model", "revision", "link_name" ] },
```
jgallagher, line 7: Is this leftover from testing, or left intentionally? Maybe add a comment if the latter?

oximeter/instruments/src/kstat/mod.rs
```diff
@@ -262,7 +275,35 @@ impl<'a> ConvertNamedData for NamedData<'a> {
...
 }
 
-#[link(name = "c")]
-extern "C" {
-    fn gethrtime() -> i64;
+#[cfg(target_os = "illumos")]
+mod time {
+    #[link(name = "c")]
+    extern "C" {
+        fn gethrtime() -> i64;
```
jgallagher, line 282: No objection here but just curious about the difference - is there a preference for this over using `clock_gettime()`? (I guess the constant is different for illumos vs linux, which is a little annoying, but it'd be _less_ of a difference?) Or, is there a reason we can't use `Instant` and forego the platform-specific bits altogether?

oximeter/oximeter/src/bin/oximeter-schema.rs
```diff
@@ -0,0 +1,97 @@
...
+                (Some(name), Some(version)) => {
+                    for each in schema.into_iter().filter(|s| {
+                        s.timeseries_name == name && s.version == version
+                    }) {
+                        println!("{each:#?}");
+                    }
+                }
+            }
+        }
+        Cmd::Emit => {
```
jgallagher, line 89: Thank you! This is much nicer than using something like `cargo expand` to see the emitted code.

oximeter/oximeter/src/lib.rs
```diff
@@ -18,33 +20,111 @@
...
+//! [[metrics]]
+//! name = "total_requests"
+//! description = "The cumulative number of requests served"
+//! datum_type = "cumulative_u64"
+//! units = "count"
+//! versions = [
+//!     { added_in = 1, fields = ["route", "method", "response_code"],
+//! ]
+//!
+//! [fields.name]
```
jgallagher, line 48: Is it okay for this to come after the `[[metrics]]` section? Or, how do we define fields on the target vs fields on the metric(s)?

jgallagher (commented):
Sorry for the many nits and questions / thanks for your patience! I haven't done a lot with oximeter so I hope these aren't way off base.

schema/rss-sled-plan.json
```diff
@@ -195,6 +195,7 @@
         },
         "checker": {
           "description": "Checker to apply to incoming messages.",
+          "default": null,
```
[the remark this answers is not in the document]
bnaecker: This one is out of our handles actually. `schemars` introduced a bug which removed NULL values from optional fields with a default. I updated the `schemars` dep, which pulled in a fix that put it _back_. [Here](https://github.com/GREsau/schemars/issues/295) is the original `schemars` bug report, which links to the fix.

oximeter/impl/src/schema/codegen.rs
```diff
@@ -0,0 +1,391 @@
...
+/// Provided with a TOML-formatted schema definition, this emits Rust types for
+/// the target and metric from the latest version; and a function that returns
+/// the `TimeseriesSchema` for _all_ versions of the timeseries.
+///
+/// Both of these items are emitted in a module with the same name as the
+/// target.
+pub fn use_timeseries(contents: &str) -> Result<TokenStream, MetricsError> {
+    let schema = load_schema(contents)?;
+    let latest = find_schema_version(schema.iter().cloned(), None);
+    let mod_name = quote::format_ident!("{}", latest[0].component_names().0);
```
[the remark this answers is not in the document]
bnaecker: Yeah, that's a good idea. I hadn't actually used this call much before this PR, but now that it's getting more mileage, adding a struct with names is a good call.

oximeter/impl/src/lib.rs
```diff
@@ -0,0 +1,45 @@
...
+pub use types::MetricsError;
+pub use types::Sample;
+
+/// Construct the timeseries name for a Target and Metric.
+pub fn timeseries_name<T, M>(target: &T, metric: &M) -> TimeseriesName
+where
+    T: Target,
+    M: Metric,
+{
+    TimeseriesName(format!("{}:{}", target.name(), metric.name()))
```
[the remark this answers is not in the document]
bnaecker: Maybe. the `Target::name` and `Metric::name` methods all return the snake_case name of the struct right now (assuming they're derived, which they all are, I checked). So since `:` isn't valid in a Rust identifier, we can't have the names contain that character.

That said, it's definitely _possible_, in that folks could implement the traits themselves and create invalid names. I don't have a strong preference, so would be happy to add that check if you think it's worthwhile.

oximeter/impl/src/schema/ir.rs
bnaecker: I'm definitely not wedded to this. It was mostly evolution rather than design anyway! I can change it up.

oximeter/impl/src/schema/ir.rs
bnaecker: IIRC, I tried keeping the error directly but hit some trait bound. I'm not sure what that would have been though, now that I see the error implements the common traits like `Clone` and `Display`. I'll try it again, and possibly use the inline-error chain trick if that doesn't work.

oximeter/impl/src/schema/ir.rs
bnaecker: Yeah, good call, will update.

oximeter/impl/src/schema/mod.rs
bnaecker: The _description_ is just freeform text. The _name_ cannot have `:` in it, but that's checked in other ways. I don't have any reason to restrict the format of the description at this time, except possibly limiting the total length?

oximeter/impl/src/schema/mod.rs
bnaecker: I need it for the `#[serde(default = "...")]` attribute, which requires a callable function rather than a variant name, unfortunately. I agree, it's silly.

oximeter/impl/src/schema/mod.rs
bnaecker: Correct, it's an empty description, and I think that does make sense. The intent was to support the sequence of changes where we start _adding_ these extra pieces of metadata to existing schema that don't currently have them. This type needs `Default` implemented so we can ask serde to create us a default on deserialization.

This is one of the areas I do appreciate extra eyes on though. I am added default values for deserialization to all the new fields of a schema:

- version (default = 1)
- description (default is empty, which you noted here)
- authz_scope (defaults to fleet)
- units (defaults to count)

I _think_ this is the right step. The plan is to add defaults for all these so that we can send them across serialization boundaries; add them into the CRDB and ClickHouse schema; and then start rewriting schema to use the TOML format where these will be provided with "real" values. But I need the defaults first, to support serialization.

oximeter/impl/src/schema/mod.rs
bnaecker: Ah, I see the confusion of some earlier comments. The _description_ is a freeform textual description of what the timeseries is supposed to measure, with a component for the target and metric. That provides extra context and details, beyond the existing _name_. It's a docstring, basically.

oximeter/impl/src/schema/mod.rs
bnaecker: I had previously convinced myself that we need defaults for these, to support implementing this whole thing in pieces. I don't think that's right anymore.

Mostly as a note to myself, I've written out why this is the case. The default is useful when the client expects different data from what the server produces. For example, the client expects a `units` field, but the serer doesn't provide one. We're not in that situation. Here, the server produces new fields that the client isn't aware of. Older clients can ignore these new fields. And newer clients which expect this field can only exist once we've deployed the code that produces it. Defaults aren't needed in either case.

Part of my confusion is around the databases. Currently `oximeter` derives a schema for each sample, and inserts it in ClickHouse. But we don't have columns for things like the `units` yet. However, we still don't need a `#[serde(default)]`, because we use the internal `DbTimeseriesSchema` type to model the database field. In the conversion from that to the real schema type, we can insert the default, but that's not a `#[serde(default)]`.

Anyway, sorry for the confusion on this, and thanks for bringing this up again. I'll remove all the default attributes and implementations that I've added here.

oximeter/impl/src/schema/mod.rs
bnaecker: Sure, that sounds good.

oximeter/impl/src/schema/mod.rs
bnaecker: Same comment -- I used this for `#[serde(default = "...")]`. Since that's going away, I'll remove this field too.

oximeter/instruments/schema/physical-data-link.toml
bnaecker: That was intentional. I have a check in non-test code that the version number is never greater than 1, so that even though we have code to handle and tests to verify multiple versions, you can't actually _create_ multiple versions right now. I was going to remove that as the very last step.

This is a breadcrumb to show what an update would look like. I'll add a comment to that effect.

oximeter/instruments/src/kstat/mod.rs
bnaecker: Yeah, I considered making the `cfg`-detection be around the clock type. I don't have a strong preference, so that might be more reasonable. `man clock_gettime` is clear that the high-res clock serves as the source for `gethrtime()` anyway, so it would be clearer and simpler to do it that way. Thanks.

oximeter/oximeter/src/lib.rs
bnaecker: In this TOML format, the fields are independent of the metrics and the target, so at the same "level" of the TOML hierarchy. The target and metric definitions refer to the fields by name. So I think they can go before or after, since we deserialize and then look up each field by name in the deserialized type.

oximeter/impl/src/schema/mod.rs
bnaecker: My previous comment isn't quite right. We don't actually need defaults for deserialization, so I'm going to remove this trait derivation.

openapi/bootstrap-agent.json
```diff
@@ -328,6 +328,7 @@
           "checker": {
             "nullable": true,
             "description": "Checker to apply to incoming messages.",
+            "default": null,
```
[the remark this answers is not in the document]
bnaecker: Same thing, this was fixing a previous bug introduced by `schemars`.

oximeter/impl/src/schema/codegen.rs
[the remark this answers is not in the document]
bnaecker: I opted instead for methods which return the relevant element of the tuple instead. See 5610bb467

oximeter/impl/src/schema/mod.rs
bnaecker: Removed in 5610bb467

oximeter/impl/src/schema/mod.rs
bnaecker: I removed the `#[serde(default)]` attributes and the constructors used to create the defaults in 5610bb467

oximeter/impl/src/schema/mod.rs
bnaecker: Removed in 5610bb467

oximeter/impl/src/schema/mod.rs
bnaecker: Renamed in dca0355f3

oximeter/instruments/schema/physical-data-link.toml
bnaecker: Added a comment in dca0355f3

oximeter/instruments/src/kstat/mod.rs
bnaecker: Reworked this in dca0355f3 so only the clock-source is platform-dependent

oximeter/impl/src/schema/ir.rs
bnaecker: Done in dca0355f3

oximeter/impl/src/schema/ir.rs
bnaecker: Ah, this type shows up in our API, so it needs serialization and `JsonSchema` traits. I'll use the error-chain approach you recommend, thanks.

oximeter/impl/src/schema/ir.rs
bnaecker: Removed in a410171f4

oximeter/impl/src/schema/ir.rs
bnaecker: Done in a410171f4

oximeter/impl/src/schema/mod.rs
bnaecker: D'oh, actually renamed in a410171f4

openapi/nexus.json
```diff
@@ -9344,6 +9344,39 @@
           }
         ]
       },
+      "AuthzScope": {
```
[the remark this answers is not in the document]
bnaecker: I think they are, yes. We're adding a bunch of new concepts to the public API, so they need to show up here.

oximeter/impl/src/schema/mod.rs
bnaecker: Hmm, so we _do_ actually need the defaults for [this test](https://buildomat.eng.oxide.computer/wg/0/details/01J0PT87YHXSPPPTX7ZXQ3NZCT/m6KJAqWpc4klzlu2Akw78hwN89jbzUThRPeG33pW6mN4Htb5/01J0PT8KF46B6ERA5ZZRBECTQS#S5148). That test will be obviated by a lot of this work here, so I might remove it now. But I will dig into it a bit first.

bnaecker: I fixed this by just overwriting the test file to include the new fields, in 13048c1333cd8b4f828a45522fd0f886879be6ec

bnaecker:
Ok, I believe I've addressed most of @jgallagher's comments. The big outstanding chunk I'd like to do is bite the bullet and centralize everything in one crate, probably `oximeter` itself. That will change the `use_timeseries!()` macro, since it'll just refer to the target and possibly metric, rather than the full file path. It will also obviate the integration test I have in Nexus, which I'll replace with one or more similar tests in `oximeter` itself.

oximeter/impl/src/lib.rs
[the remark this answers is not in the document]
jgallagher: I guess there are three options:

1. Leave as-is
2. Change this to `try_from()` and unwrap the result
3. Change this to `try_from()` and return an error

I think 1 and 2 are almost the same: derived implementations will be fine, an incorrect manual implementation would panic here in 2, and an incorrect manual implementation would (probably?) panic somewhere else later? 3 would be the safest but might also be annoying since this function is currently infallible.

If 3 is super easy I'd do that, but if it's annoying I might go with option 2, and make it something like `.try_from().expect("illegal name: implementations should be derived")` or whatever to make it clear that the only reason it's failing is because someone went out of their way to do it wrong.

oximeter/impl/src/schema/mod.rs
jgallagher: Oh, yeah, I completely misunderstood this (which you realized in the subsequent comment, haha). Sorry! Ignore this suggestion.

oximeter/oximeter/src/lib.rs
jgallagher: Ok. I've had weird experiences with table orderings and TOML before, but I can never remember the details.

bnaecker: That's definitely true. Any array of tables always refers to the most recent entry. So if you have `[[foo]]` and then do `[[foo.bar]]`, that's a sub-table in the most recent entry in the `foo` array. There are probably other situations too, but thankfully they don't apply here!

oximeter/impl/src/lib.rs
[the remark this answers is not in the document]
bnaecker: I made this fallible in d358d61c8

bnaecker:
Ok @jgallagher I've also centralized the timeseries definitions into the `oximeter/oximeter/schema` subdirectory. The `use_timeseries!()` proc-macro now takes the name of a file only, inside that subdirectory, and generates the Rust code from its contained timeseries. We can now also easily run a test in `oximeter` that checks the consistency across all timeseries defined in that directory. That should make it easy to detect conflicts, without developers needing to remember to add files anywhere.

oximeter/timeseries-macro/src/lib.rs
```diff
@@ -0,0 +1,60 @@
...
+/// timeseries. It will generate a submodule named by the target in the file,
+/// with a Rust struct for the target and each metric defined in the file.
+#[proc_macro]
+pub fn use_timeseries(
+    tokens: proc_macro::TokenStream,
+) -> proc_macro::TokenStream {
+    match syn::parse::<syn::LitStr>(tokens) {
+        Ok(token) => {
+            let filename = token.value();
+            let path = std::path::Path::new(SCHEMA_DIRECTORY).join(&filename);
```
jgallagher, line 19: Not sure how much we care about this, but `.join()` is pretty flexible; e.g., if someone calls this with a filename of `/foo/bar/baz`, `SCHEMA_DIRECTORY` will be discarded entirely. If we want to _require_ that this be inside `SCHEMA_DIRECTORY`, we maybe want to call `filename.file_name()` to strip any parent directories or `../` escapes?

jgallagher (approved):
LGTM! Just one note / question about how strict we want to be with schema location.

oximeter/timeseries-macro/src/lib.rs
bnaecker: Yeah, nice catch, thanks. I might choose to be even more conservative here at this point, just disallowing any slashes in the name at all, and / or checking that the `file_name()` output is the same as the input literal string itself. Thanks!

bnaecker: Done in 6a7b8ae7f

bnaecker:
Thanks for the review @jgallagher! All the comments were really helpful. I'm going to give Adam a chance to take a look, and then I'll fix up the conflicts on the way in.

oximeter/impl/src/schema/codegen.rs
```diff
@@ -0,0 +1,392 @@
...
+                schema.description.metric.as_str(),
+            )
+        }
+    };
+    quote! {
+        #[doc = #type_docstring]
+        #[derive(Debug, Clone, #oximeter_trait, PartialEq)]
+        pub struct #struct_name {
+            #(#field_defs),*
+            #maybe_datum
```
ahl, line 121: Is this right? I *think* that `#( #field_defs ),*` will omit the trailing comma. So if we have one or more items in `field_defs` and then `maybe_datum` is `Some(_)` we'd have:

```rust
pub struct Foo {
    pub x: Y
    pub datum: D
}
```

I'd suggest:


```suggestion
            #( #field_defs, )*
            #maybe_datum ,
```

ahl (commented):
maybe I found a bug... or I'm confused

oximeter/impl/src/schema/codegen.rs
bnaecker: Nice catch, thanks. This does look like a bug that works by accident. The one set of metrics I'm moving now all have no fields themselves, only those from the target. So we never hit the case you describe. I'll add tests for that stuff now.

bnaecker: I added tests that catch this and fixed it up in 7cefed7a7

bnaecker:
Aight, I'm going to start rebasing on main to resolve the conflicts before smashing that merge button.

PR #5891, opened 2024-06-13
Make `DataStore::instance_fetch_all` less ugly

hawkw:
This commit rewrites the `DataStore::instance_fetch_all` query added in
#5888 to perform two separate left joins with the `vmm` table, so that
the result set only contains a single row. This is a bit more
aesthetically pleasing than having to iterate over a multi-row result
set, but should still return the same values.

Thanks to @jgallagher and @luqmana for figuring out how to do the Diesel
alias!

### DIFF
--- a/nexus/db-queries/src/db/datastore/instance.rs
+++ b/nexus/db-queries/src/db/datastore/instance.rs
@@ -373,19 +373,32 @@ impl DataStore {
 
         use db::schema::instance::dsl as instance_dsl;
         use db::schema::migration::dsl as migration_dsl;
-        use db::schema::vmm::dsl as vmm_dsl;
+        use db::schema::vmm;
 
-        let mut results = instance_dsl::instance
+        // Create a Diesel alias to allow us to LEFT JOIN the `instance` table
+        // with the `vmm` table twice; once on the `active_propolis_id` and once
+        // on the `target_propolis_id`.
+        let (active_vmm, target_vmm) =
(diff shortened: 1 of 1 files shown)


nexus/db-queries/src/db/datastore/instance.rs
```diff
@@ -373,19 +373,32 @@ impl DataStore {
...
-        use db::schema::vmm::dsl as vmm_dsl;
+        use db::schema::vmm;
+
+        // Create a Diesel alias to allow us to LEFT JOIN the `instance` table
+        // with the `vmm` table twice; once on the `active_propolis_id` and once
+        // on the `target_propolis_id`.
+        let (active_vmm, target_vmm) =
+            diesel::alias!(vmm as active_vmm, vmm as target_vmm);
+        let vmm_selection =
+            <Vmm as Selectable<diesel::pg::Pg>>::construct_selection();
```
jgallagher, line 384: Great find! I looked for something like this hanging off of `Alias` and came up empty.

jgallagher (approved):

nexus/db-queries/src/db/datastore/instance.rs
hawkw: Yeah, there was, as far as I can tell, no actual documentation explaining this usage  I found it by looking at the implementation of `as_select()` and noticed that there was a comment on the `Selectable` trait saying that the return value from `construct_selection()` was "usually a tuple of fields".

PR #5896, opened 2024-06-13
Begin using typed instance and Propolis UUIDs in sled agent and Nexus

gjcolombo:
One goal of the ongoing Great Instance State Rework of 2024 is to clarify that, in broad strokes, Nexus manages instances and sled agents manage VMMs. Nexus decides whether and when an instance is connected to a VMM and how to convert VMM states to instance states; sled agents manage individual Propolis processes and send their state changes to Nexus so that it has the relevant VMM states available.

Today, sled agent's instance manager tracks VMMs in a map with type `BTreeMap<Uuid, (Uuid, Instance)>`, where the keys are instance IDs and the values contain Propolis IDs. In our proposed brave new world, sled agent can use Propolis IDs as keys and may not need instance IDs at all--i.e., its map type should be `BTreeMap<PropolisUuid, Vmm>`, where the `Vmm` may or may not contain an `InstanceUuid`.

To support this future change, change sled agent's instance map to a `BTreeMap<InstanceUuid, (PropolisUuid, Instance)>` and work through all the compiler-error fallout. This causes the difference between instance and Propolis UUIDs to filter all the way up through Nexus's instance management routines and sagas. This should make it much easier to refactor sled agent's APIs to take Propolis IDs, since the compiler can then help catch instance and Propolis ID transpositions.

Nexus's instance management code often deals with instance IDs, Propolis IDs, and sled IDs together, so also start using `SledUuid` in some of these paths.

This change draws an arbitrary line at using typed UUIDs in database model types and in the generated resource objects in the authz layer; calls into/out of these layers use `GenericUuid` and `from_untyped_uuid`/`into_untyped_uuid` as needed.(This is not because these changes are necessarily impossible, merely because the line had to go somewhere.)

Tests: `cargo nextest`; while this change is large, it should by and large be a rote refactoring exercise.

### DIFF
--- a/uuid-kinds/src/lib.rs
+++ b/uuid-kinds/src/lib.rs
@@ -57,6 +57,7 @@ impl_typed_uuid_kind! {
     LoopbackAddress => "loopback_address",
     OmicronZone => "service",
     PhysicalDisk => "physical_disk",
+    Propolis => "propolis",
     Sled => "sled",
     TufRepo => "tuf_repo",
     Upstairs => "upstairs",
--- a/clients/nexus-client/src/lib.rs
+++ b/clients/nexus-client/src/lib.rs
@@ -38,6 +38,7 @@ progenitor::generate_api!(
         NewPasswordHash = omicron_passwords::NewPasswordHash,
         TypedUuidForCollectionKind = omicron_uuid_kinds::CollectionUuid,
         TypedUuidForDownstairsKind = omicron_uuid_kinds::TypedUuid<omicron_uuid_kinds::DownstairsKind>,
+        TypedUuidForPropolisKind = omicron_uuid_kinds::TypedUuid<omicron_uuid_kinds::PropolisKind>,
         TypedUuidForSledKind = omicron_uuid_kinds::TypedUuid<omicron_uuid_kinds::SledKind>,
         TypedUuidForUpstairsKind = omicron_uuid_kinds::TypedUuid<omicron_uuid_kinds::UpstairsKind>,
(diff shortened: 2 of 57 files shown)


uuid-kinds/src/lib.rs
```diff
@@ -57,6 +57,7 @@ impl_typed_uuid_kind! {
     LoopbackAddress => "loopback_address",
     OmicronZone => "service",
     PhysicalDisk => "physical_disk",
+    Propolis => "propolis",
```
hawkw, line 60: In some places, we refer to the UUID of a Propolis process as a "Propolis ID", elsewhere, we call it a "VMM ID". Perhaps we should commit to being more consistent with naming and always refer to them as one or the other (at least, when it doesn't require a database migration...)

hawkw (approved):

uuid-kinds/src/lib.rs
gjcolombo: The general rule I've tried to apply is to use "Propolis ID" to refer to these UUIDs, except when working directly with DB queries where we're working with the primary key to the VMM table.

(When I created the VMM table in #4194, `propolis_id` was a much more common field in slog messages that contained a Propolis/VMM ID, so when it came time to choose an option it won. IMO the structured logging field name should carry a lot of weight--it should be easy to filter for messages concerning a single VMM without having to write `propolis_id = "foo" || vmm_id = "foo"` to get all the messages.)

hawkw: Yup, that makes sense. I may have to rename some `vmm_id` fields in logs I've added...

Issue #4913, opened 2024-01-28
`oximeter` histograms could maintain more information about their inputs


bnaecker:
`oximeter::Histogram` is a pretty simple type that maintains a set of buckets and a counter for each. As samples are added, the counter for bucket containing that sample is incremented. We don't currently store any other information about the input distribution, but we probably should.

- We maintain the number of samples already, but we should add the sum of the inputs so we can compute the mean
- We could use [Welford's algorithm](https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Welford's_online_algorithm) to keep an estimate of standard deviation. It's pretty simple and uses constant space.
- I recently spent some time digging into the [P**2 algorithm](https://www.cse.wustl.edu/~jain/papers/ftp/psqr.pdf) for estimating quantiles also using constant space. It's a very interesting algorithm, mostly in that it has no business being as effective as it is. We could probably stand to keep the median, p90, and p99 for all distributions.

Also, all of these should go into each `Sample` and into the database.

PR #5897, opened 2024-06-13
Additional tracking on histograms for means, std dev, p* quantile estimates

zeeshanlakhani:
Closes https://github.com/oxidecomputer/omicron/issues/4913.

This is part 1 of tracking more information within a histogram to compute min, max, means, std_dev, and estimates for p50, p90, and p99 and how to store/retrieve the associated info from Clickhouse.


oximeter/oximeter/src/histogram.rs
```diff
@@ -508,66 +704,136 @@ where
...
+                    / (self.n_samples() as f64);
+                Some(variance.sqrt())
+            })
+        } else {
+            None
+        }
+    }
+
+    /// Return the sample standard deviation for inputs to the histogram.
+    pub fn sample_std_dev(&self) -> Option<f64> {
```
bnaecker, line 774: I think we might want to improve the documentation of this method and the `std_dev()` method. `std_dev()` actually returns the standard deviation of the samples seen by the histogram. This is a biased estimator of the standard deviation of the population underlying those samples. The value returned from `sample_std_dev()` is technically the _corrected sample standard deviation_, though we do often call it just the "sample standard deviation".

I'd recommend making clear that `std_dev()` is a biased estimate of the population deviation, and that `sample_std_dev()` is an unbiased estimate applying Bessel's correction. I might also suggest we make a recommendation that people usually call `sample_std_dev`, as that's more often what people want: an unbiased estimator of the deviation in the underlying distribution.

oximeter/oximeter/src/quantile.rs
```diff
@@ -0,0 +1,393 @@
...
+/// Structure for estimating the p-quantile of a population.
+///
+/// This is based on the P² algorithm for estimating quantiles using
+/// constant space.
+///
+/// The algorithm consists of maintaining five markers: the
+/// minimum, the p/2-, p-, and (1 + p)/2 quantiles, and the maximum.
+///
+/// Read <https://www.cs.wustl.edu/~jain/papers/ftp/psqr.pdf> for more.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, JsonSchema)]
```
bnaecker, line 35: We can make this type `Copy`, for an easy performance improvement.

oximeter/oximeter/src/quantile.rs
```diff
@@ -0,0 +1,393 @@
...
+            self.marker_hghts[3],
+        ];
+
+        let len = self.len() as usize;
+        float_ord::sort(&mut heights[..len]);
+
+        let desired_index = (len as f64) * self._p() - 1.;
+        let mut index = desired_index.ceil();
+        if desired_index == index && index >= 0. {
+            let index = index.round_ties_even() as usize;
```
bnaecker, line 140: This variable shadowing is quite confusing, let's rename the one inside this branch.

bnaecker, line 140: I don't follow why we're using `round_ties_even()` rather than `round()`. Could you make a note explaining that?

oximeter/oximeter/src/quantile.rs
```diff
@@ -0,0 +1,393 @@
...
+            self.marker_hghts[0],
+            self.marker_hghts[1],
+            self.marker_hghts[2],
+            self.marker_hghts[3],
+        ];
+
+        let len = self.len() as usize;
+        float_ord::sort(&mut heights[..len]);
+
+        let desired_index = (len as f64) * self._p() - 1.;
```
bnaecker, line 137: This paragraph is pretty confusing. For one thing, `desired_index` is always <= 0.0, so the check on L139 degenerates to a check that both it and `index` are equal to 0. It also looks like none of the tests exercise this branch. There are also a few shadowed variables and things clamped in ways I don't quite follow.

To be clear, I don't know if anything is _wrong_, but it's hard for me to verify it. Could you take another pass here and maybe provide some details about what we're doing and why? Renaming the variables would help, and possibly crafting a test that hits this check, if it's not too onerous.

oximeter/oximeter/src/quantile.rs
```diff
@@ -0,0 +1,393 @@
...
+        self.len() == 0
+    }
+
+    /// Append a value/observation to the population and adjust the heights.
+    pub fn append<T>(&mut self, value: T) -> Result<(), QuantileError>
+    where
+        T: HistogramSupport,
+    {
+        if !value.is_finite() {
+            return Err(QuantileError::NonFiniteValue(format!("{:?}", value)));
```
bnaecker, line 170: Nit, but I don't think we need to format the value. It's either -inf, +inf, or NaN, and I'm not sure knowing which would provide much.

oximeter/oximeter/src/quantile.rs
```diff
@@ -0,0 +1,393 @@
...
+                let q_new = self.parabolic(i, d);
+                if self.marker_hghts[i - 1] < q_new
+                    && q_new < self.marker_hghts[i + 1]
+                {
+                    self.marker_hghts[i] = q_new;
+                } else {
+                    self.marker_hghts[i] = self.linear(i, d);
+                }
+                let delta = d.round_ties_even() as i64;
+                debug_assert_eq!(delta.abs(), 1);
```
bnaecker, line 230: If this is really an invariant, I'd just `assert!()` it.

oximeter/oximeter/src/quantile.rs
```diff
@@ -0,0 +1,393 @@
...
+                let d = Float::signum(d);
+                let q_new = self.parabolic(i, d);
+                if self.marker_hghts[i - 1] < q_new
+                    && q_new < self.marker_hghts[i + 1]
+                {
+                    self.marker_hghts[i] = q_new;
+                } else {
+                    self.marker_hghts[i] = self.linear(i, d);
+                }
+                let delta = d.round_ties_even() as i64;
```
bnaecker, line 229: Same note here, why `round_ties_even()` as opposed to `round()`? Actually, isn't `delta == d` by construction, since `d = d.signum()`? Why round at all?

oximeter/oximeter/src/quantile.rs
```diff
@@ -0,0 +1,393 @@
...
+                    k = i;
+                    break;
+                }
+            }
+            if self.marker_hghts[4] < value {
+                self.marker_hghts[4] = value;
+            }
+        };
+
+        // Increment all positions greater than k.
```
bnaecker, line 204: The comment seems to disagree with the code, since we're starting _at_ `k`. I think this is an artifact of the paper using 1-based indexing, possibly?

oximeter/oximeter/src/quantile.rs
```diff
@@ -0,0 +1,393 @@
...
+                debug_assert_eq!(delta.abs(), 1);
+                self.marker_pos[i] += delta;
+            }
+        }
+
+        Ok(())
+    }
+
+    /// Subtract another `Quantile` instance from this one.
+    pub fn sub(&self, other: &Quantile) -> Quantile {
```
bnaecker, line 239: What does it mean to subtract two quantiles?

oximeter/oximeter/src/quantile.rs
```diff
@@ -0,0 +1,393 @@
...
+mod tests {
+    use super::*;
+    use oximeter::test_util::assert_almost_eq;
+
+    #[test]
+    fn test_float_observations() {
+        let observations = [
+            0.02, 0.5, 0.74, 3.39, 0.83, 22.37, 10.15, 15.43, 38.62, 15.92,
+            34.60, 10.28, 1.47, 0.40, 0.05, 11.39, 0.27, 0.42, 0.09, 11.37,
+        ];
```
bnaecker, line 301: Let's refer to the data in the paper that this comes from.

oximeter/oximeter/src/quantile.rs
```diff
@@ -0,0 +1,393 @@
...
+                100.0
+            ]
+        );
+        assert_eq!(q.len(), 100);
+        assert_eq!(q.p(), 30);
+        assert_eq!(q.estimate().unwrap(), 30.0);
+    }
+
+    #[test]
+    fn test_p50() {
```
bnaecker, line 338: The following tests are the same internals, could we factor out the guts to something like:

```rust
test_quantile_impl(p: u64) {
    let mut q = Quantile::new(p).unwrap();
    for o in 1..=100 {
        q.append(o).unwrap();
    }
    assert_eq!(q.p(), p);
    assert_eq!(q.estimate().unwrap(), f64::try_from(p).unwrap());
}
```

bnaecker (commented):
Thanks for taking all this on, it's a big chunk of work! I've not been able to work through everything yet, but have some questions and comments around the guts of the P2 algorithm. Maybe we can take a pass at those, and then I'll work through the rest once we're happy? Please let me know if my questions aren't clear.

Thanks!

oximeter/oximeter/src/quantile.rs
```diff
@@ -0,0 +1,393 @@
...
+///
+/// The algorithm consists of maintaining five markers: the
+/// minimum, the p/2-, p-, and (1 + p)/2 quantiles, and the maximum.
+///
+/// Read <https://www.cs.wustl.edu/~jain/papers/ftp/psqr.pdf> for more.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, JsonSchema)]
+pub struct Quantile {
+    /// The heights of the markers.
+    #[serde(rename = "marker_heights")]
+    pub marker_hghts: [f64; 5],
```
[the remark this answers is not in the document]
zeeshanlakhani: Yep. I've planned to do this, and so will.

oximeter/oximeter/src/quantile.rs
```diff
@@ -0,0 +1,393 @@
...
+            marker_hghts: [0.; 5],
+            marker_pos: [1, 2, 3, 4, 0],
+            desired_marker_pos: [1., 1. + 2. * p, 1. + 4. * p, 3. + 2. * p, 5.],
+            desired_marker_incrs: [0., p / 2., p, (1. + p) / 2., 1.],
+        })
+    }
+
+    /// Create a new `Quantile` instance from the given marker heights and
+    /// positions.
+    pub fn from(
```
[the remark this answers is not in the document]
zeeshanlakhani: Yeah, I'll look to work around this.

oximeter/oximeter/src/histogram.rs
```diff
@@ -333,12 +354,137 @@ pub struct Bin<T> {
...
+                    // Ensure that the sum of samples and sum of squares don't overflow
+                    let on_unwrap = || {
+                        if value < <$type>::zero() {
+                            i64::MIN
+                        } else {
+                            i64::MAX
+                        }
+                    };
+
+                    self.sum_of_samples = self.sum_of_samples.checked_add(value as i64).unwrap_or_else(on_unwrap);
```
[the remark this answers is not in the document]
zeeshanlakhani: Yeah, I'll double check that, but I think so.

oximeter/oximeter/src/quantile.rs
zeeshanlakhani: ah, will remove this actually.

oximeter/oximeter/src/quantile.rs
zeeshanlakhani: This is a good question. We should chat offline on this.

oximeter/oximeter/src/quantile.rs
zeeshanlakhani: yep, just did this today actually :).

oximeter/oximeter/src/quantile.rs
zeeshanlakhani: Yep. I don't think it's strictly necessary, I'll update it.

oximeter/oximeter/src/quantile.rs
```diff
@@ -0,0 +1,393 @@
...
+    /// Create a new `Quantile` instance.
+    pub fn new(p: u64) -> Result<Self, QuantileError> {
+        let p = p as f64 / 100.;
+        if p < 0. || p > 1. {
+            return Err(QuantileError::InvalidPValue);
+        }
+
+        Ok(Self {
+            marker_hghts: [0.; 5],
+            marker_pos: [1, 2, 3, 4, 0],
```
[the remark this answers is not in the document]
zeeshanlakhani: Yeah, I'll double-check on this bit.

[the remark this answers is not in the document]
zeeshanlakhani: Ok, reworked to code to not do this.

oximeter/oximeter/src/quantile.rs
zeeshanlakhani: I decided to empty these out;  none them, as it doesn't really make sense.

oximeter/oximeter/src/quantile.rs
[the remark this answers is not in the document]
zeeshanlakhani: Revisited this in a few implementations and the paper. It actually makes complete sense as the 5 marker position mirrors the observation after it because useful once filled.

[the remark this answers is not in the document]
zeeshanlakhani: This position gets updated accordingly after we've filled it. Will document.

oximeter/oximeter/src/histogram.rs
```diff
@@ -508,66 +704,136 @@ where
...
+    pub fn min(&self) -> T {
+        self.min
+    }
+
+    /// Return the maximum value of all inputs to the histogram.
+    pub fn max(&self) -> T {
+        self.max
+    }
+
+    /// Return the mean of all inputs/samples in the histogram.
```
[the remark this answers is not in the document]
zeeshanlakhani: This is now updated, as it won't be done but something or 0.

oximeter/oximeter/src/histogram.rs
```diff
@@ -508,66 +704,136 @@ where
...
+    }
+
+    /// Return the standard deviation for inputs to the histogram.
+    pub fn std_dev(&self) -> Option<f64> {
+        if self.n_samples() > 1 {
+            self.mean().and_then(|mean| {
+                let sum_of_squares = self.sum_of_squares.to_f64()?;
+                let sum_of_samples = self.sum_of_samples.to_f64()?;
+
+                let variance = (sum_of_squares - (sum_of_samples * mean))
```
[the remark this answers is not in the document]
zeeshanlakhani: This has now been updated to track squared_mean as an f64 throughout.

oximeter/oximeter/src/histogram.rs
```diff
@@ -480,22 +637,61 @@ where
...
+        p99: Quantile,
+    ) -> Result<Self, HistogramError> {
+        if bins.len() != counts.len() {
+            return Err(HistogramError::ArraySizeMismatch {
+                n_bins: bins.len(),
+                n_counts: counts.len(),
+            });
+        }
+
+        let mut hist = Self::new(&bins)?;
```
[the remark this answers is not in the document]
zeeshanlakhani: Adding this shortly with extra docs.

oximeter/db/src/oxql/point.rs
```diff
@@ -1548,14 +1601,30 @@ impl<T: DistributionSupport> Distribution<T> {
...
+            .zip(rhs.counts.iter())
+            .map(|(x, y)| x.checked_sub(*y))
             .collect::<Option<_>>()
             .context("Underflow subtracting distributions values")?;
-        Ok(Self { bins: self.bins.clone(), counts })
+
+        // Subtract sum_of_samples and squared_mean directly.
+        // We allow underflow here.
+        let sum_of_samples = self.sum_of_samples - rhs.sum_of_samples;
+        let squared_mean = self.squared_mean - rhs.squared_mean;
```
zeeshanlakhani, line 1615: As per convo with @bnaecker, I am fixing this.

oximeter/oximeter/src/quantile.rs
```diff
@@ -0,0 +1,569 @@
...
+const FILLED_MARKER_LEN: usize = 5;
+
+/// Errors related to constructing a `Quantile` instance or estimating the
+/// p-quantile.
+#[derive(
+    Debug, Clone, Error, JsonSchema, Serialize, Deserialize, PartialEq,
+)]
+#[serde(tag = "type", content = "content", rename_all = "snake_case")]
+pub enum QuantileError {
+    /// The p value must be in the range [0, 1].
```
bnaecker, line 33: Just checking, does the algorithm work correctly if someone passes `0.0` or `1.0` to the constructor? That is, can we actually "estimate" the minimum and maximum? The paper says:

> For the algorithm to work correctly, the marker heights must always be in a nondecreasing order

That implies it _should_ work to estimate the 0 or 100th quantile, but they also don't give examples that I can see. Those values are also sometimes not considered "true" quantiles, so we might want to disallow them anyway. Up to you, but something to check.

oximeter/oximeter/src/quantile.rs
```diff
@@ -0,0 +1,569 @@
...
+    pub fn new(p: f64) -> Result<Self, QuantileError> {
+        if p < 0. || p > 1. {
+            return Err(QuantileError::InvalidPValue);
+        }
+
+        Ok(Self {
+            p,
+            marker_heights: [0.; FILLED_MARKER_LEN],
+            // We start with a sample size of 0.
+            //marker_positions: [0, 1, 2, 3, 0],
```
bnaecker, line 96: Dead code?

oximeter/oximeter/src/quantile.rs
```diff
@@ -0,0 +1,569 @@
...
+    where
+        T: HistogramSupport,
+    {
+        if !value.is_finite() {
+            return Err(QuantileError::NonFiniteValue);
+        }
+        // We've already checked that the value is finite.
+        let value_f = value.to_f64().unwrap();
+
+        // if !self.is_filled {
```
bnaecker, line 253: Is this dead code, or a comment? I like the method name or field, if you want to keep it. It does make clear the intent much better than the check against the const.

oximeter/oximeter/src/quantile.rs
```diff
@@ -0,0 +1,569 @@
...
+            }
+            Some(i) => i,
+            None => {
+                self.marker_heights[0] = value_f;
+                0
+            }
+        };
+
+        // Handle rounding issues as described in
+        // <https://aakinshin.net/posts/p2-quantile-estimator-rounding-issue>.
```
bnaecker, line 278: Nice catch by you and this person! The continual summing will add more error as the sums get much larger than the desired increments. Nice.

oximeter/oximeter/src/quantile.rs
```diff
@@ -0,0 +1,569 @@
...
+        }
+
+        Ok(())
+    }
+
+    /// Find the higher marker cell whose height is lower than the observation.
+    ///
+    /// Returns `None` if the value is less than the initial marker height.
+    fn find_cell(&mut self, value: f64) -> Option<usize> {
+        if value < self.marker_heights[0] {
```
bnaecker, line 311: It feels like we're reimplementing `slice::partition_point()` but I'm not sure it's worth changing with such small arrays and the 1-based indexing differences.

bnaecker (commented):
Looks great! I really like the new docstrings and doctests, and the cleanup of the quantile internals is all awesome. I've got a few remaining comments, most of which are pretty small at this point. Hopefully things are clear, but as always let me know if not! Thanks, this is really great to see.

oximeter/db/src/oxql/point.rs
```diff
@@ -1517,15 +1520,30 @@ pub trait DistributionSupport:
...
 pub struct Distribution<T: DistributionSupport> {
     bins: Vec<T>,
     counts: Vec<u64>,
+    min: Option<T>,
+    max: Option<T>,
+    sum_of_samples: T,
+    squared_mean: f64,
+    p50: Option<Quantile>,
+    p90: Option<Quantile>,
+    p99: Option<Quantile>,
```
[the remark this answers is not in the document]
zeeshanlakhani: I'll add this last part as a comment in the code too. I fixed up the squared means sub.

oximeter/db/src/oxql/point.rs
```diff
@@ -1548,14 +1601,30 @@ impl<T: DistributionSupport> Distribution<T> {
...
-            .map(|(x, y)| x.checked_sub(y))
+            .zip(rhs.counts.iter())
+            .map(|(x, y)| x.checked_sub(*y))
             .collect::<Option<_>>()
             .context("Underflow subtracting distributions values")?;
-        Ok(Self { bins: self.bins.clone(), counts })
+
+        // Subtract sum_of_samples and squared_mean directly.
+        // We allow underflow here.
+        let sum_of_samples = self.sum_of_samples - rhs.sum_of_samples;
```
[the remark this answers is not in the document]
zeeshanlakhani: Yep. Will checked sub this, even though it will either be int64 or f64.

oximeter/db/schema/single-node/5/up.sql
```diff
@@ -0,0 +1,139 @@
+ALTER TABLE oximeter.measurements_histogrami8 ADD COLUMN min Int8;
```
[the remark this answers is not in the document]
zeeshanlakhani: Makes sense.

oximeter/db/schema/single-node/5/up.sql
```diff
@@ -0,0 +1,139 @@
...
+ALTER TABLE oximeter.measurements_histogrami8 ADD COLUMN max Int8;
```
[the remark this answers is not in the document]
zeeshanlakhani: I figured they might, but good to be explicit.

oximeter/oximeter/src/quantile.rs
zeeshanlakhani: It should, but I'll double check these bounds.

oximeter/oximeter/src/quantile.rs
zeeshanlakhani: Ah, dead code. Will remove.

oximeter/oximeter/src/quantile.rs
zeeshanlakhani: I'll double-check.

oximeter/db/src/oxql/point.rs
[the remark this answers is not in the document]
bnaecker: Oh, right, these are signed, but the checked sub above is because the counts can't be negative. I was wrong then, this can definitely be negative. Suppose we get a huge negative sample in a later histogram? We want to reflect that in this subtraction.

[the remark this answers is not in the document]
zeeshanlakhani: Ok, yeah, will leave this.

oximeter/db/schema/replicated/5/up.sql
```diff
@@ -0,0 +1,279 @@
+ALTER TABLE oximeter.measurements_histogrami8_local ON CLUSTER oximeter_cluster ADD COLUMN min Int8;
```
bnaecker, line 1: I forgot that we made all scalar measurement columns nullable in version 4, to support missing samples. We did _not_ do that for the histogram types, because arrays cannot be nullable in ClickHouse, e.g., you can't have a type `Nullable(Array(Int8))`. Since we know that all `Histogram`s are non-empty by construction (they have at least one bin), then we use an empty array to represent those missing samples.

So, what do we do here? For the scalar columns, `max`, `sum_of_samples` and `squared_mean`, I think we want them to be `Nullable(T)`. For their defaults, I think it unfortunately depends on whether the sample is missing or not (whether the bins / counts arrays are empty). I _think_ you can write that like this:

```
DEFAULT if(empty(bins), NULL, 0)
```

I would do the same thing for the array columns, but using `if(empty(bins), [], [0, 0, 0, 0, 0])` instead.

There is actually one other option, which I think might be better. We could just assert that it's fraught to set a default on these columns, even in the interest of preserving the existing data. Instead, one could drop those tables and recreate them with the desired structure. I just double-checked, and the only existing histogram timeseries are the Nexus request latencies, which are not consumed anywhere at all. You could have one SQL file which drops each table and one which recreates it, this time _without_ defaults. (Those are only there to support the existing data, which we're dropping.)

oximeter/db/src/client/mod.rs
```diff
@@ -3671,12 +3675,12 @@ mod tests {
...
         // Write out the upgrading SQL files.
         //
-        // Note that all of these statements are going in the version 2 schema
+        // Note that all of these statements are going in the version 6 schema
         // directory.
         let (schema_dir, version_dirs) =
             create_test_upgrade_schema_directory(replicated, &[NEXT_VERSION])
                 .await;
-        const NEXT_VERSION: u64 = 2;
+        const NEXT_VERSION: u64 = OXIMETER_VERSION + 1;
```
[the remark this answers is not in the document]
zeeshanlakhani: Yep. reverted.

oximeter/db/schema/single-node/5/up.sql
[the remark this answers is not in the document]
zeeshanlakhani: As discussed, will do a drop/create series for these.

oximeter/db/schema/single-node/5/up.sql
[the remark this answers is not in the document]
zeeshanlakhani: As discussed, will do a drop/create series for these.

oximeter/db/schema/replicated/5/up.sql
zeeshanlakhani: As discussed, will do a drop/create series for these.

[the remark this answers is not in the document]
zeeshanlakhani: As discussed, will do a drop/create series for these.

oximeter/oximeter/src/quantile.rs
zeeshanlakhani: tests added for these and one with a comparison to what another impl returns at 1.0.

oximeter/oximeter/src/quantile.rs
zeeshanlakhani: Updated.

zeeshanlakhani:
Last piece is the sql updates (will make those in the next commit).

bnaecker (approved):
I took a look at the SQL migrations assuming that's the only thing that changed. Those LGTM, though there's one opportunity I think we should take for simplifying the definition of a distributed table on top of the local ones. That will really reduce the difficulty of changing the replicated table definitions.

Other than that, LGTM! Thanks for all the back and forth on this, it'll be cool to see this work merged!

oximeter/db/schema/replicated/5/up04.sql
```diff
@@ -0,0 +1,23 @@
...
+    p50_marker_heights Array(Float64),
+    p50_marker_positions Array(UInt64),
+    p50_desired_marker_positions Array(Float64),
+    p90_marker_heights Array(Float64),
+    p90_marker_positions Array(UInt64),
+    p90_desired_marker_positions Array(Float64),
+    p99_marker_heights Array(Float64),
+    p99_marker_positions Array(UInt64),
+    p99_desired_marker_positions Array(Float64)
+)
```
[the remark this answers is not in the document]
zeeshanlakhani: Making updates that look like this:

```sql
CREATE TABLE IF NOT EXISTS oximeter.measurements_histogramu16 ON CLUSTER oximeter_cluster
AS oximeter.measurements_histogramu16_local
ENGINE = Distributed('oximeter_cluster', 'oximeter', 'measurements_histogramu16_local', xxHash64(splitByChar(':', timeseries_name)[1]));
```

PR #5901, opened 2024-06-14
Account for region snapshots during allocation

jmpesp:
When replacing a snapshot (aka a read-only downstairs somewhere), region allocation must occur for a snapshot volume. A snapshot volume's region set is currently only composed of read-only downstairs which all contain the same data, and any newly allocated region should take those into account when looking to meet the redundancy criteria for a region set: in production, allocate the new region to a distinct sled.

This is done by slightly changing the region allocation query to accept an optional snapshot id, and, if that is supplied, will add the region snapshot's pools to the `existing_zpools` temporary table in order to prevent region allocation there.


nexus/db-queries/src/db/datastore/region.rs
```diff
@@ -240,6 +241,76 @@ impl DataStore {
         Ok(dataset_and_regions)
     }
 
+    /// Allocate new regions for a snapshot volume
+    #[allow(clippy::too_many_arguments)]
+    pub async fn arbitrary_region_allocate_for_snapshot(
```
smklein, line 246: Nit: What's the word `arbitrary_` doing for us in these functions?

nexus/db-queries/src/db/queries/region_allocation.rs
```diff
@@ -116,24 +122,42 @@ pub fn allocation_query(
...
-    .sql("
-  existing_zpools AS (
-    SELECT
-      dataset.pool_id
-    FROM
-      dataset INNER JOIN old_regions ON (old_regions.dataset_id = dataset.id)
-  ),")
+    FROM dataset WHERE ((dataset.size_used IS NOT NULL) AND (dataset.time_deleted IS NULL)) GROUP BY dataset.pool_id),");
+
+    let builder = if let Some(snapshot_id) = snapshot_id {
```
smklein, line 127: (No action needed) nice to see, based on tests/output, that this doesn't change the SQL for the existing snapshot-less queries

nexus/db-queries/src/db/queries/region_allocation.rs
```diff
@@ -116,24 +122,42 @@ pub fn allocation_query(
...
+        builder.sql("
+      existing_zpools AS ((
+        SELECT
+          dataset.pool_id
+        FROM
+          dataset INNER JOIN old_regions ON (old_regions.dataset_id = dataset.id)
+      ) UNION (
+       select dataset.pool_id from
+ dataset inner join region_snapshot on (region_snapshot.dataset_id = dataset.id)
+ where region_snapshot.snapshot_id = ").param().sql(")),")
```
smklein, line 139: :+1: makes sense

smklein (approved):
Looks solid, thanks for the tests!

nexus/db-queries/src/db/datastore/region.rs
jmpesp: The original function (before adding the redundancy level as an parameter) was `disk_region_allocate`, which allocated `REGION_REDUNDANCY_THRESHOLD` regions. `arbitrary_region_allocate` was meant to accept an arbitrary redundancy parameter. The `_direct` suffix was used to differentiate the function that accepted

```rust
        disk_source: &params::DiskSource,
        size: external::ByteCount,
```

and

```rust
        block_size: u64,
        blocks_per_extent: u64,
        extent_count: u64,
```

Adding `_snapshot_` to the name in this PR is further meant to differentiate when allocating for snapshot volumes...

Maybe something like builder pattern could work here instead:

```rust
        let query = crate::db::queries::region_allocation::AllocationQuery()
            .for_disk(volume_id)
            .with_disk_source(disk_source)
            .with_size(size)
            .with_allocation_strategy(allocation_strategy)
            .with_redundancy(3)
            .build()?;

        let query = crate::db::queries::region_allocation::AllocationQuery()
            .for_snapshot(volume_id, snapshot_id)
            .with_block_size(block_size)
            .with_blocks_per_extent(blocks_per_extent)
            .with_extent_count(extent_count)
            .with_redundancy(1)
            .build()?;
```

nexus/db-queries/src/db/queries/region_allocation.rs
jmpesp: :100:

nexus/db-queries/src/db/queries/region_allocation.rs
jmpesp: :+1:

nexus/db-queries/src/db/datastore/region.rs
smklein: I do like the idea of a builder pattern, if it doesn't add too much overhead. I definitely did not grok that `arbitrary` referred to the "redundancy" aspect from the function invocation.

smklein: Follow-up from rust huddle: seems like builder pattern might be overkill (dynamically verifying the right params kinda sucks, because we'd introduce a new error that should never happen, and statically verifying params would result in type explosion). However, using a combinations of structs / enums to group together argument types does seem sufficient to reduce the calling complexity here too.

jmpesp: This ended up working very well, see 9c4487b0d

smklein: Ship it!

PR #5905, opened 2024-06-17
UplinkAddress change broke deserialization of old test data

Nieuwejaar:
Use "use as" to avoid BgpPeer name collision with maghemite


Nieuwejaar:
- This restores the test data that was broken by the initial UplinkAddress change.  It also fixes the bugs that that test data would have caught.
- This makes the backwards compatibility paths for the bootstore and rss mechanisms more similar to one another.  It might be worth recasting this as a Trait at some point, but I've deferred that in the interest of getting the repo unfucked as quickly as possible.
- This also fixes a bgp-specific issue with the bootstore

sled-agent/src/bootstrap/early_networking.rs
```diff
@@ -905,186 +896,246 @@ impl From<EarlyNetworkConfig> for bootstore::NetworkConfig {
...
 }
 
-#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema)]
-struct EarlyNetworkConfigBodyV1 {
-    /// The external NTP server addresses.
-    pub ntp_servers: Vec<String>,
+/// Structures and routines used to maintain backwards compatibility.  The
+/// contents of this module should only be used to convert older data into the
+/// current format, and not for any ongoing run-time operations.
+pub mod back_compat {
```
smklein, line 902: Good idea to throw this into a separate module.

(For any other reviewers, the contents of this file appear to mostly be re-shuffling these structs, rather than changing them significantly)

sled-agent/src/bootstrap/params.rs
```diff
@@ -96,6 +183,24 @@ pub struct RackInitializeRequest {
...
+            Ok(req) => return Ok(req),
+            Err(e) => e,
+        };
+        if let Ok(v1) =
+            toml::from_str::<back_compat::RackInitializeRequestV1>(&data)
+        {
+            return Ok(v1.into());
+        }
+
+        Err(v2_err.into())
```
smklein, line 200: I'm assuming the intent is: If we can't parse this struct as any known format, return the parsing error for the most-recent version?

sled-agent/src/rack_setup/plan/sled.rs
```diff
@@ -53,6 +55,42 @@ pub struct Plan {
...
+            }
+            Err(e) => format!("unable to parse Plan: {e:?}"),
+        };
+        serde_json::from_str::<ShadowPlanV1>(&value)
+            .map(|v1| Plan {
+                rack_id: v1.rack_id,
+                sleds: v1.sleds,
+                config: v1.config.into(),
+            })
+            .map_err(|_| v2_err)
```
smklein, line 90: ... and looks like the same behavior here, I assume that if we added a new version you'd want us to do:

- try to parse v3
- try to parse v2
- try to parse v1
- if all those fail, return the parse error from v3

smklein (approved):

sled-agent/src/bootstrap/params.rs
Nieuwejaar: Exactly.  I'll add a comment to that effect.

sled-agent/src/rack_setup/plan/sled.rs
Nieuwejaar: Right.  It's the same pattern in both places, which makes turning this into a Trait potentially useful.

PR #5906, opened 2024-06-17
pick up Propolis 50cb28f5

gjcolombo:
Update to Propolis commit 50cb28f5 to obtain the following updates:

```
50cb28f server: separate statuses of inbound and outbound migrations (#661)
61e1481 Don't return None from BlockData::block_for_req when attachment is paused (#711)
79e243b Build and clippy fixes for Rust 1.79 (#712)
3c0f998 Modernize fw_cfg emulation
7a65be9 Update and prune dependencies
```

50cb28f5 changes the Propolis instance status and migration status APIs to report both inbound and outbound migration status. Update sled agent accordingly.

Tests: `cargo nextest`; will also run a couple of instances on a dev cluster before merging.

### DIFF
--- a/sled-agent/src/common/instance.rs
+++ b/sled-agent/src/common/instance.rs
@@ -114,7 +114,7 @@ impl ObservedPropolisState {
     /// runtime state and an instance state monitor response received from
     /// Propolis.
     pub fn new(
-        migration: Option<&MigrationRuntimeState>,
+        instance_runtime: &InstanceRuntimeState,
         propolis_state: &InstanceStateMonitorResponse,
     ) -> Self {
         // If there's no migration currently registered with this sled, report
@@ -122,7 +122,14 @@ impl ObservedPropolisState {
         // even if Propolis has some migration data to share. (This case arises
         // when Propolis returns state from a previous migration that sled agent
         // has already retired.)
-        let Some(migration) = migration else {
+        //
+        // N.B. This needs to be read from the instance runtime state and not
+        //      the migration runtime state to ensure that, once a migration in
+        //      completes, the "completed" observation is reported to
+        //      `InstanceStates::apply_propolis_observation` exactly once.
+        //      Otherwise that routine will try to apply the "inbound migration
+        //      complete" instance state transition twice.
+        let Some(migration_id) = instance_runtime.migration_id else {
             return Self {
                 vmm_state: PropolisInstanceState(propolis_state.state),
                 migration_status: ObservedMigrationStatus::NoMigration,
@@ -136,12 +143,8 @@ impl ObservedPropolisState {
             &propolis_state.migration.migration_in,
             &propolis_state.migration.migration_out,
         ) {
-            (Some(inbound), _) if inbound.id == migration.migration_id => {
-                inbound
-            }
-            (_, Some(outbound)) if outbound.id == migration.migration_id => {
-                outbound
-            }
+            (Some(inbound), _) if inbound.id == migration_id => inbound,
+            (_, Some(outbound)) if outbound.id == migration_id => outbound,
             _ => {
                 // Sled agent believes this instance should be migrating, but
                 // Propolis isn't reporting a matching migration yet, so assume
--- a/sled-agent/src/common/instance.rs
+++ b/sled-agent/src/common/instance.rs
@@ -114,82 +114,57 @@ impl ObservedPropolisState {
     /// runtime state and an instance state monitor response received from
     /// Propolis.
     pub fn new(
-        instance_runtime: &InstanceRuntimeState,
-        propolis_id: PropolisUuid,
+        migration: Option<&MigrationRuntimeState>,
         propolis_state: &InstanceStateMonitorResponse,
     ) -> Self {
-        // Decide which of Propolis's reported migrations to pay attention to.
-        // If this Propolis is currently a migration target (evidenced by its ID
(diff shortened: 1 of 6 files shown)


sled-agent/src/common/instance.rs
```diff
@@ -115,15 +115,40 @@ impl ObservedPropolisState {
...
+        // look at the current migration in. Otherwise, look at the reported
+        // migration out.
+        let role = if instance_runtime
+            .dst_propolis_id
+            .is_some_and(|id| id == propolis_id)
+        {
+            MigrationRole::Target
+        } else {
+            MigrationRole::Source
+        };
```
hawkw, line 133: This logic is going to have to change substantially once sled-agent no longer has an `InstanceRuntimeState` to make this determination. I think we ought to be able to do this instead using the `migration_state` in `InstanceStates`, which should also have the ID of the currently active migration (which would let us check whether the migration in or migration out refers to the current migration).

hawkw:
these [clippy failures](https://github.com/oxidecomputer/omicron/actions/runs/9554123138/job/26334490984?pr=5906) should be fixed by #5907, btw, so rebasing should fix that.

sled-agent/src/common/instance.rs
gjcolombo: Good suggestion, thanks! d9b4ae8 should clean this up a bit.

gjcolombo:
This looks good in local testing (I've gone through a few instance state transitions and checked to make sure that instance reboot works properly now that the fix for oxidecomputer/propolis#709 is in place). Still need to pick up the fix for #5907 in this branch, but otherwise this is ready for review.

hawkw (approved):

hawkw:
looks great! thanks for changing it to use the `MigrationState` instead of the `InstanceRuntimeState` — that'll make my life easier later!

gjcolombo:
I appear to have upset the migration integration tests; will investigate.

gjcolombo:
The problem is an assertion failure caused by d9b4ae8 (newest commit in this PR). The simplest fix may just be to revert that commit, but I'll think about it some more.

The basic issue is that `InstanceStates::apply_propolis_observation` assumes that the "migration in succeeded" branch of its logic will only be reached once per migration target. (This assumption is encoded in `InstanceStates::switch_propolis_id_to_target`, which asserts that the new active Propolis ID it's writing isn't `None`.) Previously, this was ensured by having `ObservedPropolisState::new` key off of the migration ID in sled agent's copy of the `InstanceRuntimeState`: once migration succeeds, sled agent clears the instance's migration ID, so subsequent observations will report that no migration is in progress even if Propolis continues to report on the completed migration.

d9b4ae8 breaks this assumption by making `ObservedPropolisState` use sled agent's `MigrationRuntimeState` to decide if a migration is in progress. This migration state is *not* cleared by `InstanceStates::clear_migration_ids`, so the existing "migration succeeded" sequence no longer closes the door on itself properly. The fix is unfortunately not as simple as clearing more fields in `clear_migration_ids`--the `MigrationRuntimeState` needs to be `Some` at the end of `apply_propolis_observation`, or sled agent won't send updated migration state to Nexus.

This is, of course, way more complicated than it needs to be: sled agent shouldn't be in the business of updating instance runtime state *at all*, and if it weren't there would be no assertions to trip. Unfortunately, we're not there quite yet, so we'll need to stitch something together to keep this working in the meantime.

gjcolombo:
I haven't rebased this PR on top of #5907 yet (so as not to lose my place in the buildomat queue), but I've done so locally, run `cargo xtask clippy`, and see no errors.

hawkw:
> The problem is an assertion failure caused by [d9b4ae8](https://github.com/oxidecomputer/omicron/commit/d9b4ae893e8a46517bfee3516414decb47ea7596) (newest commit in this PR). The simplest fix may just be to revert that commit, but I'll think about it some more.
> 
> The basic issue is that `InstanceStates::apply_propolis_observation` assumes that the "migration in succeeded" branch of its logic will only be reached once per migration target. (This assumption is encoded in `InstanceStates::switch_propolis_id_to_target`, which asserts that the new active Propolis ID it's writing isn't `None`.) Previously, this was ensured by having `ObservedPropolisState::new` key off of the migration ID in sled agent's copy of the `InstanceRuntimeState`: once migration succeeds, sled agent clears the instance's migration ID, so subsequent observations will report that no migration is in progress even if Propolis continues to report on the completed migration.
> 
> [d9b4ae8](https://github.com/oxidecomputer/omicron/commit/d9b4ae893e8a46517bfee3516414decb47ea7596) breaks this assumption by making `ObservedPropolisState` use sled agent's `MigrationRuntimeState` to decide if a migration is in progress. This migration state is _not_ cleared by `InstanceStates::clear_migration_ids`, so the existing "migration succeeded" sequence no longer closes the door on itself properly. The fix is unfortunately not as simple as clearing more fields in `clear_migration_ids`--the `MigrationRuntimeState` needs to be `Some` at the end of `apply_propolis_observation`, or sled agent won't send updated migration state to Nexus.
> 
> This is, of course, way more complicated than it needs to be: sled agent shouldn't be in the business of updating instance runtime state _at all_, and if it weren't there would be no assertions to trip. Unfortunately, we're not there quite yet, so we'll need to stitch something together to keep this working in the meantime.

As we discussed on Matrix, I'm fine with doing it this way for now so that everything can keep working, and I'll get rid of the use of `InstanceRuntimeState` when removing it from sled-agent in #5749.

PR #5908, opened 2024-06-17
in-progress update to steno with UndoActionPermanentError

davepacheco:
(This is not yet remotely ready.)

In this branch I've started updating Omicron for oxidecomputer/steno#283.  See discussion there.  At this point I've updated a dozen or more call sites but there are a few dozen left and I'm not sure this is the right way to go.


davepacheco:
Closing because this branch has rotted.  If we pick this up again, we'll probably start over.

PR #5909, opened 2024-06-17
Add instance ID column to migration table

hawkw:
Currently, the `migration` table does not contain a way to directly
associate migrations with instances. An `instance` row's `migration_id`
is a foreign key into the `migration` table pointing at the currently
active migration, if one exists, but past migrations cannot be easily
associated with migrations. This is unfortunate, as it seems useful to
be able to easily list completed/failed migrations for an instance. This
is particularly important as we would like to change the behavior around
marking migration records as deleted, so that deleting an instance also
deletes any corresponding migration records. It may also be useful for
debugging tools and UIs that expose a migration's history.

This branch adds an instance ID column to the migration table. To avoid
having `NULL`s in this column, I've added a migration[^1] that deletes
the old `migration` table and creates a new one, rather than backfilling
migration IDs by trying to figure them out using the `vmm` records for
the migration records. This seemed fine, because (1) the `migration`
table hasn't been released to customers yet, and (2), we don't actually
use the data in that table for anything yet, so nuking it should be
okay. I've also added a query for listing migration records by instance
ID.

Now, we no longer mark migrations as deleted when they complete or fail,
and instead mark all migrations for an instance as deleted when the
instance itself is deleted.

Migration records are still marked as deleted when an `instance-migrate`
saga unwinds. I'm on the fence as to whether this is the Right Thing or
not, but I think it's fine, as those migration states won't be needed by
the `instance-update` saga (AFAIK so far). If that doesn't end up being
the case, we might consider adding a `SagaUnwound` state, similar to the
one for VMMs, so that we can distinguish between migrations that failed
because a VMM failed and migrations whose sagas unwound...

[^1]: A database migration, not a live migration migration. :)


gjcolombo (approved):

nexus/db-queries/src/db/datastore/migration.rs
```diff
@@ -38,6 +42,22 @@ impl DataStore {
             .map_err(|e| public_error_from_diesel(e, ErrorHandler::Server))
     }
 
+    /// List all migrations associated with the provided instance ID.
+    pub async fn migration_list_by_instance(
```
bnaecker, line 46: Just curious, is there ever more than one entry here? We're filtering to non-deleted migrations, so I naively expected to see just a single value returned here.

schema/crdb/add-instance-id-to-migrations/up02.sql
```diff
@@ -0,0 +1,45 @@
...
+    id UUID PRIMARY KEY,
+
+    /* The ID of the instance that was migrated */
+    instance_id UUID NOT NULL,
+
+    /* The time this migration record was created. */
+    time_created TIMESTAMPTZ NOT NULL,
+
+    /* The time this migration record was deleted. */
+    time_deleted TIMESTAMPTZ,
```
bnaecker, line 12: Does this need a modification time?

bnaecker: Ah, I think this is covered by the two "updated" time fields below, is that right?

bnaecker (approved):
Looks good to me! I have one question about the count of migrations per instance we expect, but it's mostly curiosity.

schema/crdb/add-instance-id-to-migrations/up02.sql
hawkw: Yeah, that's correct; once a migration record is inserted, updates will touch either the source OR target fields. I should probably add a comment about this.

nexus/db-queries/src/db/datastore/migration.rs
hawkw: There's never more than one entry *yet*.  Right now, migrations are marked as deleted as soon as the migration completes. However, this will probably change in the future, and migrations will be deleted only when the instance itself is deleted (and possibly once they have been completed for a 'fairly long' period of time). I didn't make that part of the change in this PR to keep the diff small, but I suppose I could...

bnaecker: Interesting, thanks for the explanation! Is the idea to keep them around in the future as an auditable log of migrations? I have no opinion on whether that should be added now, but I do think a quick comment that there may be more than one in the future might be helpful. Thanks!

hawkw:
Okay, @bnaecker and @gjcolombo, if you have a moment, I'd love another review, as I've made some additional changes: now, we no longer mark migrations as deleted when they complete or fail, and instead mark all migrations for an instance as deleted when the instance itself is deleted.

Migration records are still marked as deleted when an `instance-migrate` saga unwinds. I'm on the fence as to whether this is the Right Thing or not, but I think it's fine, as those migration states won't be needed by the `instance-update` saga (AFAIK so far). If that doesn't end up being the case, we might consider adding a `SagaUnwound` state, similar to the one for VMMs, so that we can distinguish between migrations that failed because a VMM failed and migrations whose sagas unwound...

schema/crdb/add-instance-id-to-migrations/up02.sql
hawkw: Added a comment explaining this in 3ec590483b33943bf0eb821a3ec0069ad77e6c30

bnaecker (approved):

nexus/db-queries/src/db/datastore/migration.rs
hawkw: That's part of the motivation, yeah. Additionally, though, migration records will be used by the `instance-update` saga to do things like determining which Propolis processes are no longer in use because their instance has been migrated out.

gjcolombo (approved):

PR #5913, opened 2024-06-18
400 instead of 500 on duplicate firewall rule names

### LINKED ISSUE #5725: Firewall rule update 500s if two rules have the same name
Bad! Could be fixed along with #5662?

image

david-crespo:
Closes #5725

This does the check in a helper function in the model file, which feels appropriate. I wasn't sure about raising external API validation errors from the model file, but we do that in a bunch of places:

```console
$ rg 'external::Error' nexus/db-model
nexus/db-model/src/allow_list.rs
17:use omicron_common::api::external::Error;

nexus/db-model/src/role_assignment.rs
9:use omicron_common::api::external::Error;

nexus/db-model/src/ipv6.rs
17:use omicron_common::api::external::Error;

nexus/db-model/src/saga_types.rs
21:use omicron_common::api::external::Error;

nexus/db-model/src/vpc_subnet.rs
76:    ) -> Result<(), external::Error> {
81:        .map_err(|e| external::Error::invalid_request(e.to_string()))

... there were a bunch more
```

### DIFF
--- a/nexus/db-model/src/vpc_firewall_rule.rs
+++ b/nexus/db-model/src/vpc_firewall_rule.rs
@@ -14,6 +14,8 @@ use nexus_types::identity::Resource;
 use omicron_common::api::external;
 use serde::Deserialize;
 use serde::Serialize;
+use serde_json::json;
+use std::collections::HashSet;
 use std::io::Write;
 use uuid::Uuid;
 
@@ -253,13 +255,38 @@ impl VpcFirewallRule {
     pub fn vec_from_params(
         vpc_id: Uuid,
         params: external::VpcFirewallRuleUpdateParams,
-    ) -> Vec<VpcFirewallRule> {
-        params
+    ) -> Result<Vec<VpcFirewallRule>, external::Error> {
+        ensure_no_duplicates(&params)?;
+        Ok(params
             .rules
             .iter()
             .map(|rule| VpcFirewallRule::new(Uuid::new_v4(), vpc_id, rule))
-            .collect()
+            .collect())
     }
+}
+
+fn ensure_no_duplicates(
+    params: &external::VpcFirewallRuleUpdateParams,
+) -> Result<(), external::Error> {
+    // we could do this by comparing set(names).len() to names.len(), but this
+    // way we can say what the duplicate names are, and that's nice!
+    let mut names = HashSet::new();
+    let mut dupes = HashSet::new();
+    for r in params.rules.iter() {
+        if !names.insert(r.name.clone()) {
+            // insert returns false if already present
+            dupes.insert(r.name.clone());
+        }
+    }
+
+    if !dupes.is_empty() {
+        return Err(external::Error::invalid_value(
+            "rules",
+            format!("Rule names must be unique. Duplicates: {}", json!(dupes)),
+        ));
+    }
+
+    Ok(())
 }
 
 impl Into<external::VpcFirewallRule> for VpcFirewallRule {
--- a/nexus/db-model/src/vpc_firewall_rule.rs
+++ b/nexus/db-model/src/vpc_firewall_rule.rs
@@ -14,7 +14,6 @@ use nexus_types::identity::Resource;
 use omicron_common::api::external;
 use serde::Deserialize;
 use serde::Serialize;
-use serde_json::json;
 use std::collections::HashSet;
 use std::io::Write;
 use uuid::Uuid;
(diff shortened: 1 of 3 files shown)


nexus/tests/integration_tests/vpc_firewall.rs
```diff
@@ -162,20 +170,6 @@ async fn test_vpc_firewall(cptestctx: &ControlPlaneTestContext) {
...
-    NexusRequest::object_get(client, url)
-        .authn_as(AuthnMode::PrivilegedUser)
-        .execute()
-        .await
-        .unwrap()
-        .parsed_body::<VpcFirewallRules>()
-        .unwrap()
-        .rules
-}
-
```
david-crespo, line 178: The changes from here up in this file are noise — I got rid of `get_rules` because it was equivalent to an `object_get` one-liner.

nexus/tests/integration_tests/vpc_firewall.rs
```diff
@@ -292,3 +286,38 @@ fn is_default_firewall_rules(
...
+        &format!("/v1/vpc-firewall-rules?vpc=default&project={}", project_name),
+        &VpcFirewallRuleUpdateParams {
+            rules: vec![rule.clone(), rule.clone()],
+        },
+        StatusCode::BAD_REQUEST,
+    )
+    .await;
+    assert_eq!(error.error_code, Some("InvalidValue".to_string()));
+    assert_eq!(error.message, "unsupported value for \"rules\": Rule names must be unique. Duplicates: [\"dupe\"]");
+}
```
david-crespo, line 323: Per the issue, this was 500ing with a unique constraint violation before the fix. We could double-fix this by also adding an error handler to the query, though this PR's change is sufficient to close the issue.

https://github.com/oxidecomputer/omicron/blob/7fb6cb47c7b194b7088744ede0572d5e5be703a6/nexus/db-queries/src/db/datastore/vpc.rs#L598-L644

nexus/db-model/src/vpc_firewall_rule.rs
```diff
@@ -253,15 +255,40 @@ impl VpcFirewallRule {
...
+        if !names.insert(r.name.clone()) {
+            // insert returns false if already present
+            dupes.insert(r.name.clone());
+        }
+    }
+
+    if !dupes.is_empty() {
+        return Err(external::Error::invalid_value(
+            "rules",
+            format!("Rule names must be unique. Duplicates: {}", json!(dupes)),
```
hawkw, line 285: it feels a bit odd to me that we're stuffing JSON into what looks like a human-readable string. does the console actually parse the JSON from this message?

hawkw: ...oh, I see, it looks like the `json!` macro is just used as a convenient way to format a comma-delimited list. that's clever --- but maybe it's worth commenting on?

david-crespo: I'll see what the more literal map and join equivalent looks like and I'll go with that if it's not too ugly, otherwise will comment.

nexus/db-model/src/vpc_firewall_rule.rs
```diff
@@ -253,15 +254,42 @@ impl VpcFirewallRule {
...
+    if dupes.is_empty() {
+        return Ok(());
+    }
+
+    let dupes_str =
+        dupes.iter().map(|d| format!("\"{d}\"")).collect::<Vec<_>>().join(", ");
+    return Err(external::Error::invalid_value(
+        "rules",
+        format!("Rule names must be unique. Duplicates: [{}]", dupes_str),
+    ));
```
hawkw, line 290: Honestly, the `fmt::Debug` implementation for `HashSet<String>` is [not _that_ far off from this, except that it'll use `{`/`}` instead of `[`/`]`](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=2c713dbee1f44a6a50643b43a8fb4c5f), so you probably _could_ just use that, unless you think having square brackets is significantly clearer for the reader. Up to you, though!

hawkw (approved):

nexus/db-model/src/vpc_firewall_rule.rs
david-crespo: Cool idea. I tried it and it is tolerable, but it gives me a funny feeling that I traced to not liking that error message output would depend on the choice of iterable data structure.

PR #5914, opened 2024-06-18
Limit number of firewall rules per VPC

### LINKED ISSUE #5662: Nexus API should limit the number of firewall rules
There does not appear to be a limit on the number of firewall rules that a user can express in an API call to Nexus. We should probably add one to avoid huge requests or blowing up OPTE itself with a massive number of rules to match.

### LINKED ISSUE #6032: [firewall rules] Additional documentation for filters in the API
Not sure if this is the right repo for this issue, happy to move it if it belongs elsewhere.

I have some firewall rules that don't specify any ports in the filters:

```console
$ oxide vpc firewall-rules view --vpc fa59febf-92a0-4eb7-8eae-0ff62258502b
success
VpcFirewallRules {
    rules: [
        VpcFirewallRule {
            action: Allow,
            description: "allow all connections from anywhere",
            direction: Inbound,
            filters: VpcFirewallRuleFilter {
                hosts: None,
                ports: None,
                protocols: Some(
                    [
                        Icmp,
                        Tcp,
                        Udp,
                    ],
                ),
            },
            id: 9714b9bc-b4d5-4023-bc83-123222ada61e,
            name: Name(
                "allow-all",
            ),
            priority: 65533,
            status: Enabled,
            targets: [
                Vpc(
                    Name(
                        "calico-test",
                    ),
                ),
            ],
            time_created: 2024-07-10T05:32:26.683745Z,
            time_modified: 2024-07-10T05:32:26.683745Z,
            vpc_id: fa59febf-92a0-4eb7-8eae-0ff62258502b,
        },
        VpcFirewallRule {
            action: Allow,
            description: "allow all outbound traffic to anywhere",
            direction: Outbound,
            filters: VpcFirewallRuleFilter {
                hosts: None,
                ports: None,
                protocols: Some(
                    [
                        Icmp,
                        Tcp,
                        Udp,
                    ],
                ),
            },
            id: 29d53e48-b79d-444a-b91f-3e5996798e98,
            name: Name(
                "allow-internal-outbound",
            ),
            priority: 65534,
            status: Enabled,
            targets: [
                Vpc(
                    Name(
                        "calico-test",
                    ),
                ),
            ],
            time_created: 2024-07-10T05:32:26.683753Z,
            time_modified: 2024-07-10T05:32:26.683753Z,
            vpc_id: fa59febf-92a0-4eb7-8eae-0ff62258502b,
        },
    ],
}
```

When i try to connect using port 80 the connection is refused

```console
$ nc -v 45.154.216.153 80
nc: connect to 45.154.216.153 port 80 (tcp) failed: Connection refused
```

When I try to connect using port 22 the connection succeeds

```console
$ nc -v 45.154.216.153 22
Connection to 45.154.216.153 22 port [tcp/ssh] succeeded!
SSH-2.0-OpenSSH_8.9p1 Ubuntu-3ubuntu0.3
```

Is this expected? It feels a bit odd to have some ports enabled by default and others not

david-crespo:
Closes #5662 
Closes #6032

- [x] Max firewall rules
- [x] Enforce max length on filters (hosts, protocols, and ports) and targets too
- [x] Improve doc comments on firewall rules types and API endpoint


nexus/db-model/src/vpc_firewall_rule.rs
```diff
@@ -264,6 +265,20 @@ impl VpcFirewallRule {
     }
 }
 
+const MAX_FIREWALL_RULES: usize = 32;
```
karencfv, line 268: Is this an arbitrary number? Perhaps it'd be  good to have a comment that explains why we chose this limit?

rmustacc: This number feels way too low. I would expect a value in the hundreds at least, especially since we don't have tags for grouping firewall rules.

david-crespo: Yeah, definitely. Was going to ask about the number next. How about 512?

nexus/db-model/src/vpc_firewall_rule.rs
```diff
@@ -211,20 +211,55 @@ pub struct VpcFirewallRule {
...
+
+        if let Some(hosts) = rule.filters.hosts.as_ref() {
+            ensure_max_len(&hosts, "filters.hosts", MAX_FW_RULE_PARTS)?;
+        }
+        if let Some(ports) = rule.filters.ports.as_ref() {
+            ensure_max_len(&ports, "filters.ports", MAX_FW_RULE_PARTS)?;
+        }
+        if let Some(protocols) = rule.filters.protocols.as_ref() {
+            ensure_max_len(&protocols, "filters.protocols", MAX_FW_RULE_PARTS)?;
+        }
```
karencfv, line 260: If a customer is exceeding 2 or more of these limits, it'd be really nice to get a single error message that tells them every limit they've exceeded. WDYT?

nexus/db-model/src/vpc_firewall_rule.rs
```diff
@@ -211,20 +211,55 @@ pub struct VpcFirewallRule {
...
 
+/// Cap on the number of rules in a VPC
+///
+/// The choice of value is somewhat arbitrary, but the goal is to have a
+/// large number that customers are unlikely to actually hit, but which still
+/// meaningfully limits the ability to overload the DB with a single request.
+const MAX_FW_RULES_PER_VPC: usize = 1024;
+
+/// Cap on targets and on each type of filter
+const MAX_FW_RULE_PARTS: usize = 128;
```
karencfv, line 222: I took a look at GCP's limits https://cloud.google.com/firewall/docs/quotas#per-firewall-rule . Perhaps we should increase this one, or make it more nuanced?

karencfv (commented):
Thanks for this PR! It'll really help set customer's expectations on what is allowed. Could you add documentation about these limits to the relevant API endpoints?

nexus/db-model/src/vpc_firewall_rule.rs
david-crespo: I’d rather just make it bigger. I’d want to see what people’s needs are before introducing a more granular set of limits that aren’t grounded in any real limitations of the system.

david-crespo:
Will add a line to the descriptions, and where possible add a max length to the OpenAPI schema.

openapi/nexus.json
```diff
@@ -20438,7 +20439,7 @@
         ]
       },
       "VpcFirewallRuleFilter": {
-        "description": "Filter for a firewall rule. A given packet must match every field that is present for the rule to apply to it. A packet matches a field if any entry in that field matches the packet.",
+        "description": "Filters reduce the scope of a firewall rule. Without filters, the rule applies to all packets to the targets (or from the targets, if it's an outbound rule). With multiple filters, the rule applies only to packets matching ALL filters. The maximum number of each type of filter is 256.",
```
david-crespo, line 20442: @karencfv I brought in something more like the console explanation here. I think it's an improvement, the only issue is that it's hard to find on the docs site, buried in the request body.

image

What might be nicer would be to put some basic summary of all of it here at the top... OR we could just link to the guide, which is pretty nice: https://docs.oxide.computer/guides/configuring-guest-networking#_firewall_rules

image

nexus/db-model/src/vpc_firewall_rule.rs
david-crespo: I bumped the parts to 256.

nexus/db-model/src/vpc_firewall_rule.rs
```diff
@@ -211,20 +211,55 @@ pub struct VpcFirewallRule {
...
 
+/// Cap on the number of rules in a VPC
+///
+/// The choice of value is somewhat arbitrary, but the goal is to have a
+/// large number that customers are unlikely to actually hit, but which still
+/// meaningfully limits the ability to overload the DB with a single request.
+const MAX_FW_RULES_PER_VPC: usize = 1024;
+
+/// Cap on targets and on each type of filter
+const MAX_FW_RULE_PARTS: usize = 256;
```
karencfv, line 222: Just occurred to me. We don't support port ranges (e.g. "3000-4000") as a value, do we? If so, we should increase this number. Or better yet, add support for port ranges at some point.

If we do support port ranges, then ignore this comment :)

openapi/nexus.json
karencfv: Oh nice! This is great. I'd like to keep this here so the SDKs inherit the descriptions, but it would be really great to add the text from the second image as well for the docs. Would it be too weird to have it all?

nexus/db-model/src/vpc_firewall_rule.rs
david-crespo: We do support ranges. You just put in `“123-456”`.

openapi/nexus.json
david-crespo: Yeah, I’ll add more. I don’t think anyone will be mad at us for having too good of documentation.

nexus/db-model/src/vpc_firewall_rule.rs
karencfv: Sweet! Can we add that to the docs?

david-crespo: It's sort of in there, but I will improve the wording.

image

david-crespo:
I've added a nice description to the update endpoint that includes some of the explanation.

### What it will look like if we don't do anything else

image

### What it will look like after we fix a couple of things in dropshot and the docs site

image

Issue #5872, opened 2024-06-08
Marking a sled non_provisionable causes existing instances to lose their VPC v2p mappings 


askfongjojo:
I ran into the issue after marking a sled as non_provisionable by running
```
oxide system hardware sled set-provision-policy --sled-id 1efda86b-caef-489f-9792-589d7677e59a --state provisionable
```
My intention was to drain instances from the sled so that it could be expunged later on. After the sled policy was updated, an instance in it was no longer able to reach other private IPs in its VPC subnet.  Running `opteadm dump-v2p` showed that the v2p mapping of this instance was gone. Once I put the sled back to provisionable, the mapping was populated again and the instance was back to being reachable from/to other instances on the same subnet.

Although setting a sled as non_provisionable could express an intent to eliminate all instances on the sled, it's also possible that the sled has some temporary problems (e.g. pending faulty component replacement) and the operator doesn't intend to remove all its existing instances (by stopping/starting them or live-migrating them). Also, it appears that the v2p mapping drop is solely because the sled query excludes all non_provisionable sleds. It's unclear if the behavior is by design.

@davepacheco / @internet-diglett - Thoughts?

davepacheco:
This does seem like a bug to me and it would certainly suck to hit this unexpectedly during whatever support activity might cause us to use this flag.  But I think @internet-diglett and @jgallagher understand the considerations here better than I do.

jgallagher:
I think that's right; v2p mappings should still go to sleds that are in-service but not provisionable, because our definition for that policy is "still in service but new resources should not be provisioned onto it".

[...]

PR #5917, opened 2024-06-19
BUGFIX: use sled filter to apply v2p mappings

internet-diglett:
It has been reported that marking a sled as "non_provisionable" causes it to lose the instance v2p mappings. This is a regression. It was determined that the scoping of the mappings is too strict, relying on the sled_policy to be set to "in service" via the db view.

The preferred option is to rely on the sled policy machinery to do the filtering for us, so the view has been removed and the querying logic updated accordingly.

Also, based on a comment in a previous related PR, we are now selecting target sleds for instances using the active_propolis_id to avoid potential conflicts during live migrations.

## Related
---
#5872

### DIFF
--- a/nexus/db-queries/src/db/datastore/v2p_mapping.rs
+++ b/nexus/db-queries/src/db/datastore/v2p_mapping.rs
@@ -7,11 +7,15 @@ use crate::context::OpContext;
 use crate::db;
 use crate::db::datastore::SQL_BATCH_SIZE;
 use crate::db::error::{public_error_from_diesel, ErrorHandler};
+use crate::db::model::ApplySledFilterExt;
 use crate::db::model::V2PMappingView;
 use crate::db::pagination::paginated;
 use crate::db::pagination::Paginator;
 use async_bb8_diesel::AsyncRunQueryDsl;
-use diesel::{QueryDsl, SelectableHelper};
+use diesel::{ExpressionMethods, QueryDsl, SelectableHelper};
+use diesel::{JoinOnDsl, NullableExpressionMethods};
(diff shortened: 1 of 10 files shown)


internet-diglett:
@gjcolombo can you look over the new diesel query to see if it lines up with your feedback from your previous comment? https://github.com/oxidecomputer/omicron/pull/5845#discussion_r1623000807

gjcolombo (approved):
LGTM for the instance/VMM join. I had forgotten about that comment, so thanks for picking that up!

(The sled policy changes look right to me as well, but you may also want an LGTM from someone who's more familiar with the sled state machine than I am.)

nexus/db-queries/src/db/datastore/v2p_mapping.rs
```diff
@@ -7,34 +7,132 @@ use crate::context::OpContext;
...
+                .select((
+                    NetworkInterface::as_select(),
+                    Sled::as_select(),
+                    Vpc::as_select(),
+                ))
+                .load_async(&*self.pool_connection_authorized(opctx).await?)
+                .await
+                .map_err(|e| public_error_from_diesel(e, ErrorHandler::Server))?
+                .into_iter()
+                .map(|i: (NetworkInterface, Sled, Vpc)| V2PMappingView {
```
jgallagher, line 79: Tiny nit - I think since these have fields with the same name (`.identity`, `.ip`) I'd probably give these full names instead of `i.N`:

```suggestion
                .map(|(nic, sled, vpc): (NetworkInterface, Sled, Vpc)| V2PMappingView {
```

(same comment in the next `paginated` block below)

jgallagher (approved):
Sled state/policy changes LGTM! :+1:

PR #5919, opened 2024-06-20
Rename `project_ip_pools_list` and `_view`

david-crespo:
These endpoints were named when we thought IP pools might be project-scoped, but they are not; they are silo-scoped, with no plans I can see to change that. So it would be cool if we could get `project_` out of the endpoint names. However, there are complications that make it hard to pick a natural alternative. There are three (3) different IP pool list endpoints:

| Op ID | Path | Description | Used for
| --- | --- | --- | --- |
| `ip_pool_list` | `/v1/system/ip-pools` | All IP pools in fleet | IP pools list (operator) |
| `silo_ip_pool_list` | `/v1/system/silos/{silo}/ip-pools` | Pools linked to specified silo | Silo detail IP pools tab (operator) |
| `project_ip_pool_list` | `/v1/ip-pools` | Pools linked to user's current silo | End-user IP pool picking |

If it weren't for `silo_ip_pool_list` already existing, I would probably just rename `project_ip_pool_list` to `silo_ip_pool_list` and leave it at that. But if we want to keep `silo_ip_pool_list` as-is, we need a different name to distinguish. Usually we treat the silo-scoped endpoint as the conceptual default and therefore leave off a qualifier. So I've decided to add `system_` to the fleet-scoped list and view. The weird thing is that there are a bunch of other system-level IP pool endpoints that will never have analogues inside the silo and which it feels silly to add `system_` to.

https://github.com/oxidecomputer/omicron/blob/76e140be8680615d83a12309d19cf7df82b348ba/nexus/tests/output/nexus_tags.txt#L154-L172


karencfv:
I'm a bit confused by all of this :sweat_smile: Would there be much harm in leaving the system and silo endpoints as they are, and changing `project_ip_pool_list` to `ip_pool_list_local` or something like that? Or am I making it even more confusing?

david-crespo:
I think I would prefer that too if we can come up a decent name. Since `silo_ip_pool_list` is taken, maybe `current_silo_ip_pool_list`.

karencfv:
I like `current_silo_ip_pool_list`

david-crespo:
Sympathetic to all points, definitely tempted to punt. My counter-feeling is that "project_" is straight up wrong, so there _has_ to be something we change it to that's better than straight up wrong.

ahl:
> Sympathetic to all points, definitely tempted to punt.

You know that's my go-to move.

> My counter-feeling is that "project_" is straight up wrong, so there _has_ to be something we change it to that's better than straight up wrong.

I'm less convinced that it's wrong, but if we're going to change something I'd suggest we change the system-level endpoints to have a prefix and leave the silo-context ones as `ip_pools_{list,view}`

david-crespo:
I think you're succeeding at getting me to just give up here. I guess `project_` is not so bad. 🙃 

--- 

I considered putting the prefix on the fleet ones, but I was surprised to find that we only have the `system_` prefix on 3 endpoints.

https://github.com/oxidecomputer/omicron/blob/c36ada99955f4f7ff8378cb29a349110e3212f1c/nexus/tests/output/nexus_tags.txt

It felt a bit heavy to put the prefix on like 15 endpoints here.

https://github.com/oxidecomputer/omicron/blob/c36ada99955f4f7ff8378cb29a349110e3212f1c/nexus/tests/output/nexus_tags.txt#L154-L172

